PR #3412, opened 2026-06-06
feat(middleware): add TokenBudgetMiddleware for per-run token budget e…

chetan655:
## Summary

Adds `TokenBudgetMiddleware` to enforce configurable token limits per agent run, preventing runaway API costs from long-horizon or looping agents.

## How it works

- **Soft warning** — at `warn_threshold` (default 80%), injects a `[TOKEN BUDGET WARNING]` into the next model call prompting the agent to wrap up.
- **Hard stop** — at `hard_stop_threshold` (default 100%), strips `tool_calls` from the response, forcing LangGraph to route to `END` cleanly without crashing.
- **Subagent-aware** — dynamically sums `usage_metadata` across all `AIMessages` on every turn, automatically capturing tokens injected retroactively by `TokenUsageMiddleware` for spawned subagents.
- Tracks **three independent dimensions**: total, input, and output tokens.

## Configuration

Disabled by default. Enable in `config.yaml`:

```yaml
token_budget:
  enabled: true
  max_tokens: 200000        # total (input + output) per run
  max_input_tokens: null    # optional input-only limit
  max_output_tokens: null   # optional output-only limit
  warn_threshold: 0.8       # warn at 80%
  hard_stop_threshold: 1.0  # hard stop at 100%
```

## Tests

Added **8 unit tests** in `test_token_budget_middleware.py` covering:

| Test | Coverage |
|---|---|
| `test_no_usage_metadata_returns_none` | Skips gracefully if model returns no usage metadata |
| `test_below_threshold_returns_none` | No action under budget |
| `test_warning_threshold_injects_warning_and_returns_none` | Soft warning queued & multi-message token accumulation |
| `test_warn_injected_at_next_model_call` | Deferred injection via `wrap_model_call` |
| `test_warn_only_once_per_run` | Warning fires at most once per run |
| `test_hard_stop_strip_tool_calls` | Tool calls stripped on hard stop |
| `test_input_tokens_trigger_limit` | Independent input token limit |
| `test_output_tokens_trigger_limit` | Independent output token limit |

## Changes

| File | Description |
|---|---|
| `token_budget_config.py` | Pydantic config with validation |
| `token_budget_middleware.py` | Core middleware implementation |
| `app_config.py` | Added `token_budget` field to `AppConfig` |
| `features.py` | Added `token_budget` flag to `RuntimeFeatures` |
| `factory.py` / `agent.py` | Registered middleware in both SDK and app builder |
| `config.example.yaml` | Documented new config section |


backend/packages/harness/deerflow/agents/middlewares/token_budget_middleware.py
```diff
@@ -0,0 +1,260 @@
...
+        messages = state.get("messages", [])
+        if not messages:
+            return None
+
+        last_msg = messages[-1]
+        if not isinstance(last_msg, AIMessage):
+            return None
+
+        current = TokenUsage()
+        for msg in messages:
```
willem-bd, line 188: **Budget is non-monotonic — summarization can reset it below the hard-stop threshold, defeating the cost cap.**
`_apply` re-sums `usage_metadata` across `state["messages"]` every turn, but `SummarizationMiddleware` removes old messages (`RemoveMessage(id=REMOVE_ALL_MESSAGES)`, `summarization_middleware.py:215`) when context grows. After summarization the summed total collapses, so an agent that already burned most of the budget keeps running far past the limit and the hard stop never fires — exactly the runaway-cost case this feature exists to prevent.
**Suggestion:** track a monotonic max-watermark (e.g. on `ThreadState` / persisted per run) instead of re-summing mutable history.

backend/packages/harness/deerflow/agents/factory.py
```diff
@@ -282,7 +282,17 @@ def _assemble_from_features(
...
-    # --- [13] Clarification (always last among built-ins) ---
+    # --- [13] TokenBudget ---
+    if feat.token_budget is not False:
+        if isinstance(feat.token_budget, AgentMiddleware):
+            chain.append(feat.token_budget)
+        else:
+            from deerflow.agents.middlewares.token_budget_middleware import TokenBudgetMiddleware
+            from deerflow.config.token_budget_config import TokenBudgetConfig
+
+            chain.append(TokenBudgetMiddleware.from_config(TokenBudgetConfig()))
```
willem-bd, line 293: **SDK path enforces the budget by default and ignores both `enabled` and the user's `config.yaml` limits.**
`_assemble_from_features` appends `TokenBudgetMiddleware.from_config(TokenBudgetConfig())` whenever `feat.token_budget is not False` (`features.py` default is `True`), and the middleware never checks `config.enabled`. So the SDK/embedded factory path is always *on* with the stock 200k limit, while `build_middlewares` (Gateway + `DeerFlowClient`) is *off* by default (`enabled=False`). A legitimate long run via the SDK gets a surprise hard stop with the final message rewritten.
**Suggestion:** honor `enabled` in the middleware, or have the factory read the real `AppConfig.token_budget` (and default `features.token_budget` to `False`).

config.example.yaml
```diff
@@ -32,6 +32,22 @@ log_level: info
...
   enabled: true
 
+# ============================================================================
+# Token Budget — Per-run token limits
+# ============================================================================
+# Prevents runaway API costs by enforcing hard token limits per run.
+# When warn_threshold is crossed, the agent receives an in-context warning.
+# When hard_stop_threshold is crossed, tool_calls are stripped and the agent
+# is forced to produce a final answer immediately.
+token_budget:
```
willem-bd, line 42: **`config_version` not bumped for the new schema section.**
`backend/CLAUDE.md` requires: *"When changing the config schema, bump `config_version` in `config.example.yaml`."* It's still `14`. Without the bump, existing deployments get no startup warning that they're missing `token_budget`, and `make config-upgrade` won't merge the new fields.

backend/packages/harness/deerflow/config/token_budget_config.py
```diff
@@ -0,0 +1,41 @@
...
+    return _token_budget_config
+
+
+def set_token_budget_config(config: TokenBudgetConfig) -> None:
+    """Set the token budget configuration."""
+    global _token_budget_config
+    _token_budget_config = config
+
+
+def load_token_budget_config_from_dict(config_dict: dict) -> None:
```
willem-bd: **Module-level singleton (`get_/set_/load_token_budget_config_from_dict`) is dead code and a silent no-op trap.**
`AppConfig` holds its own instance (`token_budget: TokenBudgetConfig = Field(default_factory=...)`), separate from `_token_budget_config`. None of `get/set/load_token_budget_config_from_dict` are called anywhere, so calling `set_token_budget_config(...)` / `load_token_budget_config_from_dict(...)` has no effect — the runtime keeps the original config. Either wire these to `AppConfig` or remove them.

backend/packages/harness/deerflow/agents/middlewares/token_budget_middleware.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @override
+    def before_agent(self, state: AgentState, runtime: Runtime) -> None:
+
+        thread_id = self._get_thread_id(runtime)
+        run_id = self._get_run_id(runtime)
+
+        with self._lock:
+            # clear any stale state from prior runs on this thread
+            stale_keys = [k for k in self._warned if k[0] == thread_id and k[1] != run_id]
```
willem-bd: **`_warned`/`_pending_warnings` grow unbounded across threads — abandoned runs on other threads are never evicted.**
`before_agent` only sweeps keys whose `thread_id` matches the current run; `after_agent` only clears the current key. A thread that is warned then abandoned (user leaves, interrupt, crash without `after_agent`) leaves its entry forever, since no later run lands on the same `thread_id`. On a long-lived multi-tenant server this is unbounded growth keyed by user/thread ids.
**Suggestion:** bound the dicts (LRU/TTL) or clear by a run epoch.

backend/packages/harness/deerflow/agents/middlewares/token_budget_middleware.py
```diff
@@ -0,0 +1,260 @@
...
+            stale_keys.extend([k for k in self._pending_warnings if k[0] == thread_id and k[1] != run_id])
+            for key in stale_keys:
+                self._clear_run_state(key)
```
willem-bd: **Two concurrent runs on the same thread clobber each other's state.**
`before_agent` clears every other `run_id` on the same `thread_id`. If run A is active on thread T and run B starts on T, B's `before_agent` wipes A's `_warned`/`_pending_warnings` mid-run → A re-warns (duplicate banner) and loses queued warnings. Reachable via re-entrant/streaming runs or multitask strategies.
**Suggestion:** key state so sibling runs don't need eviction, or only evict truly-terminal runs.

backend/packages/harness/deerflow/agents/middlewares/token_budget_middleware.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @staticmethod
+    def _get_thread_id(runtime: Runtime) -> str:
+        ctx = getattr(runtime, "context", None)
+        if isinstance(ctx, dict):
+            return ctx.get("thread_id", "default")
+        return "default"
+
+    @staticmethod
+    def _get_run_id(runtime: Runtime) -> str:
```
willem-bd, line 91: **`run_id` defaults to `"default"`, colliding across runs when the context has no `run_id`.**
`worker.py` sets `run_id` for the Gateway path, but `DeerFlowClient._get_runnable_config` (`client.py:209–217`) sets only `thread_id`. With the budget enabled over the embedded client, every run on a thread keys to `(thread_id, "default")`, so `before_agent` can't distinguish runs and the warned flag leaks across sequential/concurrent runs.
**Suggestion:** require/derive a run id, or key state off something always present.

backend/packages/harness/deerflow/agents/middlewares/token_budget_middleware.py
```diff
@@ -0,0 +1,260 @@
...
+                return None
+
+        return None
+
+    @override
+    def after_model(self, state: AgentState, runtime: Runtime) -> dict | None:
+        return self._apply(state, runtime)
+
+    @override
+    async def aafter_model(self, state: AgentState, runtime: Runtime) -> dict | None:
```
willem-bd, line 252: **Blocking `threading.Lock` on the async path.**
`aafter_model` → `_apply` acquires `self._lock` (a `threading.Lock`) inside an async hook; under contention this stalls the event loop. The critical section is small (the message-sum loop is outside the lock, so only the fractions/dict update is under it), so impact is limited, but consider an `asyncio`-friendly approach or `run_in_executor` to stay consistent with the repo's blocking-IO stance.

willem-bd (commented):
High-effort review of the new `TokenBudgetMiddleware`. The hard-stop *mechanism* is sound — stripping `tool_calls` routes to END correctly, and the `ModelRequest.override()`-based warning injection is valid against langchain 1.2.15. Findings are inline below.

**Top two I'd want addressed before merge:**
1. **Hard stop can be defeated by summarization** — the budget is recomputed by re-summing `usage_metadata` over the mutable message history, so it's non-monotonic and the cost cap won't reliably fire once `SummarizationMiddleware` trims history. Track a monotonic watermark instead.
2. **SDK factory path is on by default and ignores `enabled` / config limits** — diverges from the documented "disabled by default" and from the `agent.py` (Gateway/embedded) path.

**Also:** `config_version` not bumped (CLAUDE.md); dead/misleading config singleton; unbounded per-thread state growth; same-thread concurrent-run state clobbering; `run_id`-missing collision on the embedded client; blocking lock on the async path.

**Verified and ruled out (not posting):** `ModelRequest.override(messages=...)` exists → warning injection works; the injected `HumanMessage` does *not* pollute checkpoint state (only the returned `AIMessage` is appended); `after_model` always sees an `AIMessage` last, so the `isinstance` guard isn't a "hard stop never fires" bug.

chetan655:
Thanks for the thorough review! All points have been addressed and pushed.

- **Monotonic Per-Run Budget**: Replaced the stateless re-summing approach with a difference accumulator keyed by `msg.id`, making the budget strictly monotonic even if history is trimmed. Additionally, updated `before_agent` to baseline old messages from previous runs so they aren't double-charged to the current run.
- **SDK Path Config**: `features.token_budget` now defaults to `False`, and the middleware explicitly checks `self._config.enabled` to ensure it stays off unless intentionally activated.
- **State Growth & Clobbering**: Bounded the state dictionaries using `BoundedDict(1000)` to prevent memory leaks, and keyed them strictly by `run_id` to prevent concurrent runs on the same thread from clobbering each other.
- **Embedded Client Collision**: Added a fallback to `str(id(runtime))` to guarantee a unique ID if `run_id` is missing.
- **Async Lock**: Because state is now safely isolated per-run, we completely deleted the blocking `threading.Lock` from the async path.
- **Config & Dead Code**: Bumped `config.example.yaml` to v15 and deleted the dead module-level singleton code from `token_budget_config.py`.
- **Unit Tests**: Updated `test_token_budget_middleware.py` and `test_create_deerflow_agent.py` to properly mock the new difference logic and reflect the updated middleware order. All tests are passing!

Ready for another look!

WillemJiang (approved):

PR #2565, opened 2026-04-26
fix(runtime): protect sync singleton init/reset with threading.Lock

GoDiao:
## Summary

`get_checkpointer()` and `get_store()` use lazy-init singletons without locks. Concurrent calls can race past the `if instance is None` check and create multiple instances, leaking DB connections/file handles. `reset_*()` interleaved with `get_*()` can also cause state inconsistency.

Applied double-check locking pattern to both providers:
- Added `threading.Lock` to `checkpointer/provider.py` and `store/provider.py`
- `get_*()` checks without lock first (fast path), then re-checks inside lock
- `reset_*()` holds lock during cleanup to prevent interleaving with init

Closes #2549

## Test plan
- [x] Added `TestCheckpointerSingletonConcurrency` (3 tests)
  - 8 concurrent threads get the same instance
  - Reset during init is serialized without errors
  - Get after reset returns a valid new instance
- [x] Added `TestStoreSingletonConcurrency` (3 tests) — same coverage
- [x] `ruff check` and `ruff format --check` pass

PR #3413, opened 2026-06-06
fix(runtime): protect sync singleton init and reset

### LINKED ISSUE #2549: [harness][runtime] Sync checkpointer/store singletons initialize without locks
## Problem

Synchronous checkpointer and store provider singletons are lazily initialized without locks. Concurrent access can race and create multiple instances or observe partially initialized/reset state.

## Impact

In multithreaded gateway/runtime contexts, callers can get inconsistent singleton instances or leak resources during concurrent initialization/reset.

## Suggested Fix

Protect singleton initialization and reset with a lock and double-check pattern. Ensure reset and initialization cannot interleave.

## Tests

- Concurrently call the sync checkpointer provider from multiple threads and verify a single instance is created.
- Repeat for the sync store provider.
- Verify reset during initialization is serialized.

## References

- `backend/packages/harness/deerflow/agents/checkpointer/provider.py:103`
- `backend/packages/harness/deerflow/runtime/store/provider.py:103`

ggnnggez:
Closes #2549
Closes #2565

## Summary

Fixes the synchronous checkpointer and store singleton race described in #2549.

This PR builds on #2565 by @GoDiao and keeps their original singleton locking contribution in the commit history. The follow-up changes strengthen the implementation and test coverage:

- serialize sync checkpointer/store initialization and reset with the same lock
- avoid publishing partially initialized context-manager state by assigning singleton globals only after `__enter__()` succeeds
- use deterministic concurrency tests that block singleton initialization to prove only one instance is created
- verify reset during initialization is serialized for both checkpointer and store
- add explicit store reset identity coverage so reset is proven to create a new singleton

## Attribution

This work includes and builds on the original PR #2565 by @GoDiao, whose commit introduced the initial double-check locking approach for sync checkpointer/store singleton initialization and reset.

## Validation

```bash
cd backend
uv run pytest tests/test_checkpointer.py -q
uv run ruff check tests/test_checkpointer.py packages/harness/deerflow/runtime/checkpointer/provider.py packages/harness/deerflow/runtime/store/provider.py
```

Both checks passed locally.


ggnnggez:
@WillemJiang 
Changes in [af116f8](https://github.com/bytedance/deer-flow/pull/3413/commits/af116f82818c623c5cb0d50081576db57111092c) :
Swtiched both provider from `threading.RLock`to `threading.Lock`.
Moved the config loading outside the provider singleton locks.
Added regression coverage to assert that `get_app_config()` is not called while the provider singleton lock is held.

WillemJiang:
@ggnnggez, here is a comment for the code 
1. Identical helper functions duplicated across providers

  _ensure_checkpointer_config_loaded() and _ensure_store_config_loaded() are byte-for-byte identical:

  ```
def _ensure_checkpointer_config_loaded() -> None:
      from deerflow.config.app_config import _app_config
      from deerflow.config.checkpointer_config import get_checkpointer_config
      config = get_checkpointer_config()
      if config is not None or _app_config is not None:
          return
      try:
          get_app_config()
      except FileNotFoundError:
          pass

```
  Same body in store/provider.py. Extract to a shared utility (e.g.,
  deerflow.config.checkpointer_config.ensure_config_loaded()) to avoid divergence over time.

ggnnggez:
@WillemJiang 
Addressed in 4d93e5e

WillemJiang (approved):

PR #3417, opened 2026-06-07
fix(middleware): externalize oversized tool output into sandbox for non-mounted sandboxes

### LINKED ISSUE #3416: [bug] ToolOutputBudgetMiddleware中间件tool结果写在本地了
### Before you start

### Problem summary

ToolOutputBudgetMiddleware的_externalize会把超长工具结果写在本地，但是后续是通过readfile工具去使用的，这就会导致找不到文件

### Affected area(s)

### What happened?

ToolOutputBudgetMiddleware的_externalize会把超长工具结果写在本地，
但是后续是通过readfile工具去使用的
这就会导致找不到文件，

### Expected behavior

希望是通过sandbox的write file函数写到沙箱里，比如：
sandbox_provider = get_sandbox_provider()
        sandbox_id = sandbox_provider.acquire(thread_id)
        sandbox = sandbox_provider.get(sandbox_id)
        if sandbox is None:
            return None
        # 沙箱写入文件
        sandbox.write_file(filepath, content)

### Steps to reproduce

1. 调用一个结果超长的tool，比如查询数据
2. _externalize写入本地文件
3. 模型根据提示词后续去readfile读找不到

### Relevant logs

```shell
暂无记录
```

Local (make dev)

### Operating system

macOS

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

Huixin615:
Fixes #3416

## Why

The trigger: issue #3416 reported that on a remote AIO sandbox, after a tool returned an oversized result, the model would try to `read_file` the path it was given and get "file not found".

Root cause: `ToolOutputBudgetMiddleware._externalize` was writing the full tool output to the **host** filesystem via `open(filepath, "w")` and returning a `/mnt/user-data/outputs/...` virtual path. When `SandboxProvider.uses_thread_data_mounts = True` (local Docker sandbox, AIO + LocalContainerBackend) the host outputs dir is bind-mounted to the same virtual path inside the sandbox, so this happened to work. When `uses_thread_data_mounts = False` (remote AIO) there is no such mount, so the file exists on the host but is invisible to the sandbox-side `read_file` tool — the model is handed an unreadable path.

The pain: any oversized tool result on a non-mounted sandbox is effectively unrecoverable — the model has the preview only, can never read the full result, and downstream reasoning silently degrades.

## What changed

From a caller's perspective, nothing changes for users running a mounted sandbox — that code path is byte-for-byte equivalent.

For users running a non-mounted (remote) sandbox: oversized tool results now actually become readable. The middleware writes the full content into the sandbox's own `/mnt/user-data/outputs/.tool-results/` (with `mkdir -p` + `write_file` + `test -s` validation), so the `read_file` instruction in the preview now works. If anything goes wrong (subdir rejected, mkdir/write/validate failure), the middleware falls back to the existing inline head+tail truncation — the model is **never** handed an unreadable path.

Public middleware API, config schema, and mounted-sandbox behavior are all unchanged.

## Surface area

## Screenshots / Recording

N/A — backend middleware fix, no UI surface.

## Bug fix verification

- Test path that reproduces the bug: `backend/tests/test_tool_output_budget_middleware.py::TestWrapToolCallSandboxIntegration::test_oversized_output_lands_in_sandbox_not_host`
- Did it go red on `main` and green on this branch? **yes** — on `main`, `_externalize` writes to the host `tmp_path` and the test's assertion `assert not (tmp_path / ".tool-results").exists()` fails; on this branch, the non-mounted dispatch routes the write into the (fake) sandbox and the host path is untouched.
- Supporting unit tests added: `TestExternalizeToSandbox` (write success / write failure / `test -s` MISSING / unsafe subdir rejection / default extension), `TestBudgetContentSandboxDispatch` (mounted vs non-mounted vs no-sandbox fallback), `TestResolveSandbox` (no state / no sandbox dict / no sandbox_id / provider success / provider exception).

## Validation

Ran locally:

- `cd backend && make format` — clean, no files reformatted, no lint findings.
- `cd backend && PYTHONPATH=. uv run pytest tests/test_tool_output_budget_middleware.py -v` — **94 passed** in 3.79s (50 original + 44 new across the 4 new test classes).
- `cd backend && make test` (full suite) — **3919 passed, 16 skipped, 0 failed** in 167s. No regressions in any other module.

## AI assistance

**Tool(s) used:** Claude (assisted with reading the existing middleware, validating assumptions against `SandboxProvider` / `Sandbox` / `SandboxMiddleware` source, drafting the dispatch and the test class structure)

**How you used it:** I described the bug from #3416 and the constraints (mounted vs non-mounted dispatch, no blocking I/O in the tool-call hot path, must never hand the model an unreadable path). The model proposed an initial design that I rejected for over-fitting (a 3-way fallback in `_resolve_sandbox` and a host-path `acquire()` in the sync wrapper), then iterated to the current simpler version that only reads `runtime.state["sandbox"]` and avoids `provider.acquire()` entirely. I reviewed every line, wrote the `mkdir -p` + `test -s` validation requirement myself after reading `aio_sandbox.py` and confirming `execute_command` returns `"Error: …"` as a string instead of raising, and verified each test class against the existing `_make_request` / fixture conventions in the file.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/tool_output_budget_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/tool_output_budget_middleware.py
@@ -11,10 +11,11 @@ from __future__ import annotations
 import asyncio
 import logging
 import os
+import shlex
 import uuid
 from collections.abc import Awaitable, Callable
 from dataclasses import replace as dc_replace
-from typing import Any, override
+from typing import TYPE_CHECKING, Any, override
 
 from langchain.agents import AgentState
 from langchain.agents.middleware import AgentMiddleware
@@ -24,8 +25,18 @@ from langgraph.prebuilt.tool_node import ToolCallRequest
 from langgraph.types import Command
 
 from deerflow.config.tool_output_config import ToolOutputConfig
+from deerflow.sandbox.sandbox_provider import get_sandbox_provider
+
+if TYPE_CHECKING:
+    from deerflow.sandbox.sandbox import Sandbox
 
 logger = logging.getLogger(__name__)
+
+# Virtual outputs root inside the sandbox. Host-mounted sandboxes map this to
+# the thread outputs dir on the host; for non-mounted (remote) sandboxes the
+# same path is written directly into the sandbox filesystem so the model's
+# ``read_file`` tool can read it back (issue #3416).
+_VIRTUAL_OUTPUTS_BASE = "/mnt/user-data/outputs"
 
 
 def _default_config() -> ToolOutputConfig:
@@ -94,6 +105,18 @@ def _sanitize_tool_name(name: str) -> str:
     return safe or "unknown"
 
 
+def _build_externalized_filename(*, tool_name: str, tool_call_id: str) -> str:
+    """Build the on-disk filename for an externalized tool output.
+
+    Shared by the host-disk and sandbox externalization paths so both
+    produce the identical naming scheme.
+    """
+    safe_name = _sanitize_tool_name(tool_name)
+    ext = _EXT_MAP.get(tool_name, "txt")
+    short_id = uuid.uuid4().hex[:12]
+    return f"{safe_name}-{short_id}.{ext}"
+
+
 def _externalize(
     content: str,
     *,
@@ -111,10 +134,7 @@ def _externalize(
     except OSError:
         return None
 
-    safe_name = _sanitize_tool_name(tool_name)
-    ext = _EXT_MAP.get(tool_name, "txt")
-    short_id = uuid.uuid4().hex[:12]
-    filename = f"{safe_name}-{short_id}.{ext}"
+    filename = _build_externalized_filename(tool_name=tool_name, tool_call_id=tool_call_id)
     filepath = os.path.join(storage_dir, filename)
 
     if not os.path.abspath(filepath).startswith(os.path.abspath(storage_dir)):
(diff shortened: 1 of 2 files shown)


WillemJiang:
@Huixin615, please take a look at the unit test error.

Huixin615:
> @Huixin615, please take a look at the unit test error.

Hi @WillemJiang  thanks for the review, I have pushed two commits to solve them.  Could you please take an another view when you have time, thank you very much!

fancyboi999:
Verified this against a real AIO Docker sandbox (the `all-in-one-sandbox` image), not just the fake-sandbox unit tests. I wrote the original middleware (#3303) and #3416 is assigned to me, so I wanted to close the one thing the unit tests can't reach: the actual container write→read round-trip.

Three checks, all green:

1. **Core round-trip.** `_externalize_to_sandbox` wrote a 208,919-char payload into the container, then `sandbox.read_file` on the returned virtual path read it back byte-for-byte, head and tail intact. This is exactly what was broken before; the file now lands where the sandbox-side `read_file` tool looks for it.
2. **Non-mounted dispatch.** I forced `uses_thread_data_mounts=False` to drive the remote-sandbox branch against real container I/O, ran the full `_budget_content`, parsed the path out of the preview, and read the full content back. Exact match.
3. **Mounted path unchanged.** With `uses_thread_data_mounts=True` the dispatch wrote host-side and never touched the sandbox, so existing local/mounted setups behave as before.

The fix holds up and I didn't need to change anything. LGTM from me. @Huixin615 nice work — when you get a chance, a sanity check on your own sandbox setup would be good to have on record too. @WillemJiang this closes the AIO gap I left open in #3303.

fancyboi999:
@WillemJiang ran a full end-to-end check on this before you merge, posting it for the record.

Setup: real AIO Docker container + real LLM, with `uses_thread_data_mounts` forced to False so the agent loop actually takes the non-mounted branch this PR adds. A plain mounted run would skip the new code entirely, so this is the path that matters.

- Turn 1: the agent ran `seq 1 5000` in the container. The oversized result (19,999 chars) was externalized into the sandbox and the agent only saw the preview.
- Turn 2: the agent called `read_file` on the saved path and correctly answered "5000". The full content came back through the sandbox.

That's the exact #3416 failure (agent handed an unreadable path) reproduced and then fixed in a real run. Code works as-is, nothing to change. Good to merge from my side.

WillemJiang (approved):

PR #3195, opened 2026-05-24
fix(#3189): prevent write_file streaming timeout on long reports

Huixin615:
Closes #3189

## Why

Long reports written in a single `write_file` tool call produced an oversized tool-call JSON payload. The model's streaming chunk-gap (silence between SSE chunks) then grew past langchain-openai's 60 s default and tripped `StreamChunkTimeoutError`, after which the retry loop re-sent the same oversized payload — stacking 6–12 minutes of dead air per agent turn. Two real-world causes had to be addressed together: the chunk-gap timeout itself was too tight for reasoning models, and there was no upstream guard preventing the agent from emitting payloads that practically guarantee a timeout.

## What changed

- **factory (`backend/packages/harness/deerflow/models/factory.py`)**: default `stream_chunk_timeout=240s` for `langchain_openai:ChatOpenAI` clients only. Users override per-model via `ModelConfig.stream_chunk_timeout` in `config.yaml`. An explicit `null` is excluded by `model_dump(exclude_none=True)` and treated as "unset", so the default is injected. For non-OpenAI providers the key is popped before kwargs reach the constructor, so it never raises `TypeError: unexpected keyword argument`.
- **sandbox (`backend/packages/harness/deerflow/sandbox/tools.py`)**: server-side **80 KB UTF-8 byte cap** on non-append `write_file` calls. Configurable via env var `DEERFLOW_WRITE_FILE_MAX_BYTES` (set to `0` to disable). Oversized payloads are rejected with a structured error pointing the model at `str_replace` or `append=True`. Note: the cap is **byte-based**, not word-based.
- **middleware**: `StreamChunkTimeoutError` is classified as transient but retries are capped at **1** via per-exception `_RETRY_BUDGET_OVERRIDES`. A same-payload retry on a chunk-gap timeout buffers the same way upstream; the full 3-attempt loop would stack 6–12 min of dead air.
- **middleware**: stream-drop exceptions now surface an actionable user-facing message instead of a raw LangChain stack.
- **prompts (`lead_agent`, `general_purpose` subagent)**: added a routing-style **File Editing Workflow** hint pointing the model at `str_replace` for incremental edits (mirrors Claude Code's `Edit` and OpenAI Codex's `apply_patch`).
- **tests**: behavioural coverage for the size guard, retry budget override, stream-drop user message, factory default injection, and the non-OpenAI provider pop (regression for CR feedback).

## Surface area

[...]

PR #3430, opened 2026-06-08
fix(web_fetch): support proxy for Jina reader in restricted networks (#3418)

18062706139fcz:
## Background

`web_fetch` (the Jina Reader tool) builds a bare `httpx.AsyncClient()` with no proxy awareness:

```python
async with httpx.AsyncClient() as client:
    response = await client.post("https://r.jina.ai/", ...)
```

For users behind a corporate proxy, or running inside Docker / WSL2 where outbound traffic must go through a host proxy, the client cannot reach `https://r.jina.ai` and `web_fetch` simply times out. There is currently no way to point the Jina request at a proxy, even though the rest of the user's environment is proxy-configured.

This addresses #3418.

### Root cause

Two independent gaps:

1. **Application layer** — `JinaClient.crawl()` never reads or honors any proxy setting, and the `web_fetch` tool config has no proxy option to pass down.
2. **Container layer** — the `gateway` service in both compose files does not inherit the host proxy environment, and even when it would, internal service-to-service calls (`gateway` → `provisioner`, sandbox health checks, etc.) must be exempted from the proxy or they break.

### 1. Application layer (`jina_client.py`, `tools.py`)

- Add optional `proxy` and `trust_env` parameters to `JinaClient.crawl()`. When `proxy` is set it is passed to `httpx.AsyncClient(proxy=...)`; `trust_env` (default `True`) lets httpx pick up `HTTP(S)_PROXY` from the environment otherwise.
- The `web_fetch` tool reads `proxy` / `trust_env` / `timeout` from its tool config and forwards them. Because YAML values can arrive as strings (e.g. `"false"`, `"20"`), three small coercion helpers (`_coerce_bool`, `_coerce_timeout`, `_coerce_proxy`) normalize them safely and fall back to defaults on junk input.

### 2. Container layer (`docker-compose*.yaml`, `scripts/docker.sh`)

- Proxy values (`HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY`) are inherited from `../.env` via the existing `env_file`, so no extra wiring is needed.
- Only `NO_PROXY` / `no_proxy` are declared in `environment`, appending internal hostnames (`gateway`, `frontend`, `nginx`, `provisioner`, `host.docker.internal`, loopback) to whatever the user already has, so in-cluster calls never get routed through the external proxy.
- `scripts/docker.sh` loads proxy vars from `.env` into the shell before invoking compose, so the `NO_PROXY` interpolation can merge user-provided values on the `make docker-start` path.

### 3. Docs

- `config.example.yaml` documents the new `proxy` / `trust_env` options under the Jina `web_fetch` entry, including the `host.docker.internal` hint for host-side proxies.

## Alternatives considered

| Option | Why not |
| --- | --- |
| Inject `HTTP_PROXY`/`HTTPS_PROXY` as bare `- HTTP_PROXY` keys in compose `environment` | **Verified to regress**: when the host shell has the var unset, compose resolves the bare key to `null`, which then *overrides* the value coming from `env_file`/`.env`. Real `docker compose config` showed `HTTP_PROXY: null` for users who configured the proxy only in `.env`. Dropping the bare keys and relying on `env_file` keeps `.env` transparent. |
| Only rely on `trust_env` + host env vars, no explicit `proxy` config | Doesn't help users who want a different proxy for Jina than their global env, and doesn't solve the in-cluster `NO_PROXY` exemption problem. |
| Set proxy globally in the gateway entrypoint | Too broad: would force every outbound client through the proxy and is harder to opt out of per-tool. The tool-config approach keeps it scoped to `web_fetch`. |

### Unit tests (`backend/tests/test_jina_client.py`)
- `44 passed` (added 4 behavior tests for proxy/trust_env forwarding + 3 parametrized helper test groups, 27 new cases total).
- Coverage for `jina_client.py` and `tools.py`: **100%**.
- `ruff check`: passed.

### Real `docker compose` validation
Validated with a standalone Docker Compose v2.32.4 binary (`config` subcommand, no daemon needed), parsing the actual compose files with a temporary `.env`:

- **Proxy only in `.env`, shell unset** → `HTTP_PROXY` / `HTTPS_PROXY` correctly transparent from `.env` (the bare-key approach regressed this to `null`; current approach fixes it).
- **Shell sets `HTTPS_PROXY` + custom `NO_PROXY`** → value honored; `NO_PROXY` correctly appends internal hostnames.
- **`make` path (`docker.sh` exports `.env` proxy vars)** → `NO_PROXY` merges to `dotenv.internal,localhost,127.0.0.1,...,host.docker.internal`.
- Production `docker-compose.yaml` validated with the same scenarios.

`bash -n scripts/docker.sh`: passed.

### DIFF
--- a/backend/packages/harness/deerflow/community/jina_ai/jina_client.py
+++ b/backend/packages/harness/deerflow/community/jina_ai/jina_client.py
@@ -9,7 +9,7 @@ _api_key_warned = False
 
 
 class JinaClient:
-    async def crawl(self, url: str, return_format: str = "html", timeout: int = 10) -> str:
+    async def crawl(self, url: str, return_format: str = "html", timeout: int = 10, proxy: str | None = None, trust_env: bool = True) -> str:
         global _api_key_warned
         headers = {
             "Content-Type": "application/json",
@@ -23,7 +23,10 @@ class JinaClient:
             logger.warning("Jina API key is not set. Provide your own key to access a higher rate limit. See https://jina.ai/reader for more information.")
         data = {"url": url}
         try:
-            async with httpx.AsyncClient() as client:
+            client_kwargs: dict[str, object] = {"trust_env": trust_env}
+            if proxy:
+                client_kwargs["proxy"] = proxy
+            async with httpx.AsyncClient(**client_kwargs) as client:
                 response = await client.post("https://r.jina.ai/", headers=headers, json=data, timeout=timeout)
 
             if response.status_code != 200:
--- a/backend/packages/harness/deerflow/community/jina_ai/tools.py
+++ b/backend/packages/harness/deerflow/community/jina_ai/tools.py
@@ -9,6 +9,38 @@ from deerflow.utils.readability import ReadabilityExtractor
 readability_extractor = ReadabilityExtractor()
 
 
+def _coerce_bool(value: object, default: bool) -> bool:
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, str):
+        normalized = value.strip().lower()
+        if normalized in {"1", "true", "yes", "on"}:
+            return True
+        if normalized in {"0", "false", "no", "off"}:
+            return False
+    return default
+
+
+def _coerce_timeout(value: object, default: int) -> int:
+    if isinstance(value, bool):
+        return default
+    if isinstance(value, int):
+        return value
+    if isinstance(value, str):
+        try:
+            return int(value)
+        except ValueError:
+            return default
+    return default
+
+
+def _coerce_proxy(value: object) -> str | None:
+    if not isinstance(value, str):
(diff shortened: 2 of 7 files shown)


fancyboi999:
Heads-up: the failing **Layer 2 — full-stack render** check here is a false positive from the replay-e2e gate (#3365), **not** an issue with this PR. The committed replay fixture went stale after an unrelated system-prompt edit (#3195), so the gate red-fails on every PR that touches its trigger paths. Root-cause fix in #3436 (it excludes the system prompt from the replay match key, so prompt edits no longer make fixtures stale). Once #3436 merges this check goes green again — nothing to change on your side.

WillemJiang (approved):

Issue #3431, opened 2026-06-08
refactor(lead-agent): make _build_middlewares public to drop the last cross-module private import


ShenAC-SAC:
### Context

Found during review of #3341 (extend deferred MCP tool loading to subagents).

`deerflow/client.py` imports the **private** `_build_middlewares` from `deerflow/agents/lead_agent/agent.py` across a module boundary, and calls it as if it were public API. `DeerFlowClient` (an embedded lead-agent variant) genuinely needs the lead agent's full middleware composition, but reaching into a `_`-prefixed name means any future rename or signature change to `_build_middlewares` silently breaks the client. (Python does not enforce private names, and the test suite already monkeypatches `deerflow.client._build_middlewares`, baking in the assumption that the private name is cross-module-visible.)

#3341 removed one such cross-module private import (`_assemble_deferred` → public `assemble_deferred_tools`) but intentionally left this one out of scope to keep that PR focused.

### Proposal

Promote `_build_middlewares` to a public, documented entry point (drop the underscore), or extract the client-needed composition into a public factory in `agents/middlewares/tool_error_handling_middleware.py` alongside the existing `build_lead_runtime_middlewares` / `build_subagent_runtime_middlewares`. (Note: `build_lead_runtime_middlewares` is a different, lower layer — not a drop-in replacement.)

### Blast radius

`agent.py` + `client.py` + ~5 test files (incl. `test_build_middlewares_*` and the `deerflow.client._build_middlewares` monkeypatches). Mechanical rename, **no behavior change**.

### Out of scope

Any behavior change. This is a pure maintainability refactor.

PR #3432, opened 2026-06-08
feat(subagents): extend deferred MCP tool loading to subagents

### LINKED ISSUE #3341: tool_search: extend deferred MCP tool loading to subagents
### Background

#3272 reworked deferred MCP tool loading to hide schemas reliably via build-time closures + per-thread graph state. That work wired deferral into the **lead** and **client** construction paths only.

### Gap

Subagents built via `SubagentExecutor._create_agent` call `get_available_tools()` (which tags MCP tools) but do **not** run the deferred assembly, so they attach no `tool_search` tool and no `DeferredToolFilterMiddleware`. With `tool_search.enabled: true` a subagent binds full MCP schemas and can call them directly — the context savings don't apply inside subagents.

This is **pre-existing** (the filter middleware was only ever attached to the lead agent) and **not a regression** from #3272 — flagging it so it isn't lost.

### Proposed fix

In `subagents/executor.py`, after the subagent's `filtered_tools` are resolved:

- `build_deferred_tool_setup(filtered_tools, enabled=...)`
- append the returned `tool_search` tool
- attach `DeferredToolFilterMiddleware(setup.deferred_names, setup.catalog_hash)`
- inject the `<available-deferred-tools>` section into the subagent system prompt

### Test

A subagent with MCP tools + `tool_search.enabled: true` must not bind MCP schemas before promotion (mirror `test_deferred_promotion_integration.py` for the subagent path).

ShenAC-SAC:
Fixes #3341. Follow-up to #3370.

## What

Extends the lead agent's deferred MCP tool mechanism (#3272 / #3342) to **subagents**. When `tool_search.enabled`, a subagent no longer binds full MCP tool schemas onto the model. Instead:

- MCP tools are surfaced **by name only** in an `<available-deferred-tools>` system-prompt section, and fetched on demand via the generated `tool_search` helper.
- `DeferredToolFilterMiddleware` deterministically rewrites `request.tools` to physically hide the deferred schemas — the prompt section is discovery only, **not** enforcement.

Token-cost optimization: large MCP schemas stay out of the `tools` array until the model actually asks for them.

## How

- **Consolidation.** `deerflow.tools.builtins.tool_search` is now the single home for `assemble_deferred_tools` (centralized fail-closed guard, replacing the lead-only private `_assemble_deferred`) and the relocated `get_deferred_tools_prompt_section`. Shared by every build path: lead agent, embedded client, and subagent executor.
- **Subagent wiring.** `executor._build_initial_state` assembles deferral once (after policy filtering), injects the prompt section into the single `SystemMessage`, and threads `deferred_setup` through `build_subagent_runtime_middlewares` → `DeferredToolFilterMiddleware`.
- **Safe by construction.** `tool_search` is appended *after* the subagent's name-level tool policy (`config.tools` / `disallowed_tools`) and is treated as infrastructure: its catalog is built from the already-policy-filtered list, so it can never surface a tool the policy denied.

## Tests

- New end-to-end `test_subagent_deferred_promotion_integration.py`: runs the subagent recipe through a real graph, asserting the hide→promote loop.
- New builder/executor tests: deferral on/off, policy interaction (denied MCP tool never enters the catalog; `tool_search` present as infra), and middleware threading.
- Migrated existing cross-context / prompt-section tests to the public names.

Full suite: **4017 passed, 19 skipped**; `ruff check` + `format` clean.

## Follow-up

A **pre-existing** cross-module private import (`client.py` → lead's `_build_middlewares`) was found during review and is tracked separately in #3431 — out of scope here.

## Note for reviewers

This touches function-calling internals (deferred-tool assembly, the tool-filter middleware, and the subagent executor). The executor + middleware diff is worth a careful read before merge.

### DIFF
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -305,6 +305,7 @@ Proxied through nginx: `/api/langgraph/*` → Gateway LangGraph-compatible runti
 **Concurrency**: `MAX_CONCURRENT_SUBAGENTS = 3` enforced by `SubagentLimitMiddleware` (truncates excess tool calls in `after_model`), 15-minute timeout
 **Flow**: `task()` tool → `SubagentExecutor` → background thread → poll 5s → SSE events → result
 **Events**: `task_started`, `task_running`, `task_completed`/`task_failed`/`task_timed_out`
+**Deferred MCP tools** (if `tool_search.enabled`): `SubagentExecutor._build_initial_state` assembles deferral after policy filtering via the shared `assemble_deferred_tools` (fail-closed), appends the `tool_search` tool, injects the `<available-deferred-tools>` section into the subagent's `SystemMessage`, and threads the setup to `_create_agent`, which attaches `DeferredToolFilterMiddleware` through `build_subagent_runtime_middlewares(deferred_setup=...)`. Subagents thus withhold full MCP schemas until promotion, same as the lead agent; each task run gets a fresh `ThreadState` so promotion is isolated per run
 
 ### Tool System (`packages/harness/deerflow/tools/`)
 
--- a/backend/packages/harness/deerflow/agents/lead_agent/agent.py
+++ b/backend/packages/harness/deerflow/agents/lead_agent/agent.py
@@ -21,7 +21,6 @@ middleware, and the async path inside ``TitleMiddleware``. Any new in-graph
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
 
 from langchain.agents import create_agent
 from langchain.agents.middleware import AgentMiddleware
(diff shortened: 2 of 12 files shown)


fancyboi999:
Heads-up: the failing **Layer 2 — full-stack render** check here is a false positive from the replay-e2e gate (#3365), **not** an issue with this PR. The committed replay fixture went stale after an unrelated system-prompt edit (#3195), so the gate red-fails on every PR that touches its trigger paths. Root-cause fix in #3436 (it excludes the system prompt from the replay match key, so prompt edits no longer make fixtures stale). Once #3436 merges this check goes green again — nothing to change on your side.

WillemJiang (approved):

WillemJiang (approved):

PR #3433, opened 2026-06-08
fix(frontend): restructure Memory settings toolbar into two rows

hetaoBackend:
## Summary

The Memory settings page toolbar packed the search input, the All/Facts/Summaries filter tabs, and the four action buttons (Import / Export / Add fact / Clear all memory) into one horizontal row. At the settings-dialog width this collapsed the search field into an unusable empty square and truncated the **"Summaries"** filter tab to **"Summarie"**. The destructive *Clear all memory* button also sat inline next to the normal actions.

## Changes

Restructured the toolbar in `frontend/src/components/workspace/settings/memory-settings-page.tsx` into two clear rows:

- **Row 1 — search + filters:** the search input now uses `min-w-0 flex-1` (capped at `sm:max-w-md`) so it always keeps a usable width; filter tabs are `shrink-0` and right-aligned (`sm:ml-auto`).
- **Filter tabs no longer truncate:** each `ToggleGroupItem` got `whitespace-nowrap`, so "Summaries" renders in full.
- **Row 2 — actions:** Import / Export / Add fact stay grouped on the left; the destructive **Clear all memory** is pushed to the far right with `ml-auto`, visually separated from the constructive actions.
- Everything stacks/wraps gracefully on narrow widths.

The empty-state behavior (the "No memory saved yet." box and the summary scaffold) is intentionally left unchanged.

## Testing

- `pnpm check` (ESLint + `tsc --noEmit`) passes clean.
- Verified visually in **Settings → Memory**.


hetaoBackend:
After a patched fix.
image

fancyboi999:
Heads-up: the failing **Layer 2 — full-stack render** check here is a false positive from the replay-e2e gate (#3365), **not** an issue with this PR. The committed replay fixture went stale after an unrelated system-prompt edit (#3195), so the gate red-fails on every PR that touches its trigger paths (this one touches `frontend/**`). Root-cause fix in #3436 (it excludes the system prompt from the replay match key, so prompt edits no longer make fixtures stale). Once #3436 merges this check goes green again — nothing to change on your side.

LittleChenLiya (approved):

MagicCube (approved):

Issue #3380, opened 2026-06-04
[feat] support PG schema


houmaodou:
### Before you start

- [x] I searched [existing issues](https://github.com/bytedance/deer-flow/issues?q=is%3Aissue) and this is not a duplicate.

### Problem / motivation

suggest supporting pg schema

### Proposed solution

suggest supporting pg schema

### Affected area(s)

Backend API (gateway / endpoints / SSE)

Huaian666:
came here to say the same thing. we've been working around the absence of this feature for months. would simplify our codebase a lot.

PR #3442, opened 2026-06-08
feat(persistence): support custom postgres schema

18062706139fcz:
## Summary

This PR implements GitHub issue #3380 (`suggest supporting pg schema`) and adds first-class PostgreSQL schema support for DeerFlow persistence.

The change makes DeerFlow able to place all PostgreSQL persistence artifacts into a caller-selected schema instead of always relying on the server default schema/search_path.

### New / enhanced capabilities

- Adds `postgres_schema` support to the unified `database` config.
- Adds `postgres_schema` support to the legacy `checkpointer` config for backward compatibility.
- Ensures DeerFlow ORM tables are created in the configured schema when `database.backend: postgres` is used.
- Ensures LangGraph checkpointer tables are created in the configured schema.
- Ensures LangGraph store tables are created in the configured schema.
- Automatically creates the target PostgreSQL schema at startup when it does not already exist.
- Preserves existing libpq `options` values while injecting or replacing `search_path`.
- Supports both PostgreSQL URL DSNs and libpq keyword DSNs for psycopg-based sync paths.
- Resets sync persistence singletons correctly when unified `database` persistence config changes during config reload.

### Scope covered by this PR

- Async ORM engine path via SQLAlchemy + asyncpg.
- Async LangGraph checkpointer path.
- Sync LangGraph checkpointer path.
- Async LangGraph store path.
- Sync LangGraph store path.
- Unified `database` configuration path.
- Legacy `checkpointer` configuration path.

### Core implementation

- Adds a shared PostgreSQL schema validator and helper utilities.
- Uses asyncpg `connect_args={"server_settings": {"search_path": ...}}` for SQLAlchemy engine connections.
- Uses psycopg/libpq `options=-c search_path=...` for LangGraph checkpointer/store connections.
- Creates schema with `CREATE SCHEMA IF NOT EXISTS` before ORM or LangGraph tables are initialized.
- Preserves non-`search_path` libpq options such as `statement_timeout` when schema injection is applied.
- Replaces existing `search_path` options instead of duplicating them.
- Handles libpq URL encoding correctly with `%20` instead of `+` for spaces in `options` query parameters.

### Files updated

- Config model and validation:
  - `backend/packages/harness/deerflow/config/database_config.py`
  - `backend/packages/harness/deerflow/config/checkpointer_config.py`
  - `backend/packages/harness/deerflow/config/postgres_schema.py`
- Persistence helpers and ORM engine wiring:
  - `backend/packages/harness/deerflow/persistence/postgres_schema.py`
  - `backend/packages/harness/deerflow/persistence/engine.py`
- LangGraph runtime wiring:
  - `backend/packages/harness/deerflow/runtime/checkpointer/async_provider.py`
  - `backend/packages/harness/deerflow/runtime/checkpointer/provider.py`
  - `backend/packages/harness/deerflow/runtime/store/async_provider.py`
  - `backend/packages/harness/deerflow/runtime/store/provider.py`
- Config reload behavior:
  - `backend/packages/harness/deerflow/config/app_config.py`
- Tests and examples:
  - `backend/tests/test_postgres_schema_helper.py`
  - `backend/tests/test_pg_schema_integration.py`
  - `backend/tests/test_app_config_reload.py`
  - `backend/tests/test_checkpointer.py`
  - `backend/tests/test_persistence_scaffold.py`
  - `config.example.yaml`

### Environment variables

You can provide the PostgreSQL connection string through an environment variable referenced by `config.yaml`:

```bash
export DATABASE_URL='postgresql://user:password@host:5432/deerflow'
```

Then reference it in config:

```yaml
database:
  backend: postgres
  postgres_url: $DATABASE_URL
  postgres_schema: deerflow_app
```

### Unified database configuration

Recommended configuration path:

```yaml
database:
  backend: postgres
  postgres_url: $DATABASE_URL
  postgres_schema: deerflow_app
  pool_size: 5
  echo_sql: false
```

Parameter reference:

- `database.backend`: persistence backend. Must be `postgres` to enable PostgreSQL persistence.
- `database.postgres_url`: PostgreSQL connection URL used by both app persistence and LangGraph persistence.
- `database.postgres_schema`: target schema for app ORM tables and LangGraph checkpointer/store tables.
- `database.pool_size`: SQLAlchemy async engine pool size for the app ORM engine.
- `database.echo_sql`: enables SQL logging for debugging.

Notes:

- `postgres_schema: ""` keeps the database server default `search_path` behavior, usually `public`.
- Schema names must be plain PostgreSQL identifiers matching `^[A-Za-z_][A-Za-z0-9_]{0,62}$`.
- Quoted identifiers are intentionally not supported in this version.

### Legacy checkpointer configuration

Backward-compatible path:

```yaml
checkpointer:
  type: postgres
  connection_string: $DATABASE_URL
  postgres_schema: deerflow_graph
```

Parameter reference:

- `checkpointer.type`: set to `postgres`.
- `checkpointer.connection_string`: psycopg connection string / DSN.
- `checkpointer.postgres_schema`: target schema for LangGraph checkpointer/store tables when using legacy config.

Notes:

- Legacy `checkpointer` config continues to take precedence over the unified `database` section.
- This path is preserved for compatibility, but the unified `database` config is the recommended setup for new deployments.

### Recommended setup steps

1. Install PostgreSQL dependencies.

```bash
cd backend
uv sync --all-packages --extra postgres
```

2. Set your connection string.

```bash
export DATABASE_URL='postgresql://user:password@host:5432/deerflow'
```

3. Configure DeerFlow.

```yaml
database:
  backend: postgres
  postgres_url: $DATABASE_URL
  postgres_schema: deerflow_app
```

4. Start DeerFlow normally.
- DeerFlow creates the schema automatically if needed.
- DeerFlow creates ORM tables in the configured schema.
- DeerFlow creates LangGraph checkpointer/store tables in the same configured schema.

5. Verify in PostgreSQL.

```sql
SHOW search_path;
SELECT table_schema, table_name
FROM information_schema.tables
WHERE table_schema = 'deerflow_app'
ORDER BY table_name;
```

### Example 1: Unified app + LangGraph persistence in one schema

```yaml
database:
  backend: postgres
  postgres_url: $DATABASE_URL
  postgres_schema: deerflow_prod
```

Expected behavior:

- DeerFlow ORM tables such as `runs`, `run_events`, `threads_meta`, `feedback`, and `users` are created in `deerflow_prod`.
- LangGraph checkpoint/store tables are also created in `deerflow_prod`.

### Example 2: Keep default server schema behavior

```yaml
database:
  backend: postgres
  postgres_url: $DATABASE_URL
  postgres_schema: ""
```

Expected behavior:

- DeerFlow does not override `search_path`.
- Tables are created according to the PostgreSQL server/user default schema behavior.

### Example 3: Legacy checkpointer config

```yaml
checkpointer:
  type: postgres
  connection_string: $DATABASE_URL
  postgres_schema: deerflow_legacy
```

Expected behavior:

- LangGraph checkpointer/store tables use `deerflow_legacy`.
- Legacy config precedence is preserved.

### Automated tests run

```bash
cd backend
uv run pytest tests/test_checkpointer.py tests/test_postgres_schema_helper.py tests/test_persistence_scaffold.py tests/test_app_config_reload.py tests/test_pg_schema_integration.py -q
```

Observed result in this environment:

- `112 passed`
- `2 skipped`
- `1 warning`

### What the tests cover

- Config validation for allowed PostgreSQL schema names.
- SQLAlchemy async engine wiring for schema-specific table creation.
- Async checkpointer/store PostgreSQL schema routing.
- Sync checkpointer/store PostgreSQL schema routing.
- Correct `%20` URL encoding for libpq `options` values.
- Preservation of existing libpq `options` values.
- Replacement of an existing `search_path` option without clobbering unrelated options.
- Support for libpq keyword DSNs.
- Runtime config reload behavior for sync persistence singletons.
- Optional live PostgreSQL integration coverage when `DEERFLOW_TEST_POSTGRES_URL` is set.

### Optional live integration test

To run the real PostgreSQL integration test locally:

```bash
export DEERFLOW_TEST_POSTGRES_URL='postgresql://user:password@host:5432/deerflow'
cd backend
uv run pytest tests/test_pg_schema_integration.py -q
```

## Migration / Upgrade Notes

If you already have DeerFlow tables in `public` and want to move to a dedicated schema:

1. Stop writers / stop the service.
2. Create the new schema if needed.
3. Move existing tables into the new schema.
4. Update `config.yaml` with `postgres_schema`.
5. Restart DeerFlow.
6. Verify new connections are using the target schema.

Example SQL:

```sql
CREATE SCHEMA IF NOT EXISTS deerflow_app;
ALTER TABLE public.runs SET SCHEMA deerflow_app;
ALTER TABLE public.run_events SET SCHEMA deerflow_app;
ALTER TABLE public.threads_meta SET SCHEMA deerflow_app;
ALTER TABLE public.feedback SET SCHEMA deerflow_app;
ALTER TABLE public.users SET SCHEMA deerflow_app;
```

LangGraph tables should be migrated similarly based on the actual table names in your deployment.

## Risks / Compatibility

- This PR keeps backward compatibility with the legacy `checkpointer` configuration path.
- Existing deployments that leave `postgres_schema` empty retain current behavior.
- The live PostgreSQL integration test is optional and only runs when `DEERFLOW_TEST_POSTGRES_URL` is configured.
- No plan/spec document is included in this PR.


WillemJiang:
@18062706139fcz, please take a look at the lint error and unit test error.

backend/packages/harness/deerflow/config/postgres_schema.py
```diff
@@ -0,0 +1,17 @@
+"""Shared validation for PostgreSQL schema names."""
+
+from __future__ import annotations
+
+import re
+
+POSTGRES_SCHEMA_PATTERN = r"^[A-Za-z_][A-Za-z0-9_]{0,62}$"
```
willem-bd: **🔴 Root cause of the mixed-case bug.** This regex allows uppercase letters and reserved words (`MySchema`, `Orders`, `user`), but nothing lowercases the value downstream. The schema is then created **quoted** (`CREATE SCHEMA IF NOT EXISTS "MySchema"`, case-preserved) while `search_path` is applied **unquoted** — which PostgreSQL folds to lowercase. The two never match, and PG silently ignores the non-existent path entry, so tables land in `public`.

Simplest fix: tighten to lowercase-only:
```python
POSTGRES_SCHEMA_PATTERN = r"^[a-z_][a-z0-9_]{0,62}$"
```
(Alternatively, quote the `search_path` value everywhere to match the quoted `CREATE SCHEMA`.) Either way, please add a mixed-case schema (`MySchema`) to the valid/invalid parametrize — all current tests use lowercase `deerflow`, so this is uncaught.

backend/packages/harness/deerflow/persistence/postgres_schema.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def build_asyncpg_connect_args(schema: str) -> dict:
+    """Return SQLAlchemy ``connect_args`` that pin asyncpg's search_path.
+
+    Empty *schema* yields ``{}`` so the engine keeps the server default.
+    """
+    if not schema:
+        return {}
+    return {"server_settings": {"search_path": schema}}
```
willem-bd, line 33: These two helpers (`build_asyncpg_connect_args` here, `build_psycopg_options` at L43) interpolate the schema into `search_path` **unquoted**, while `create_schema_sql` (L89) emits it **quoted**. For any non-lowercase name PostgreSQL down-cases the unquoted `search_path` token, so it points at a different (non-existent) schema than the one created — tables silently fall back to `public`. If you keep the regex permissive, the `search_path` value must be double-quoted to match (`search_path="MySchema"`); the cleaner fix is the lowercase-only regex.

backend/packages/harness/deerflow/persistence/engine.py
```diff
@@ -146,19 +155,38 @@ def _enable_sqlite_wal(dbapi_conn, _record):  # noqa: ARG001 — SQLAlchemy cont
...
         logger.debug("deerflow.persistence.models not found; skipping auto-create tables")
 
+    async def _ensure_schema_and_tables(conn) -> None:
+        # CREATE SCHEMA is DDL and is unaffected by search_path, so it is
+        # safe even though the connection's search_path already points at
+        # the (not-yet-existing) target schema.
+        if backend == "postgres" and postgres_schema:
+            from sqlalchemy.schema import CreateSchema
+
+            await conn.execute(CreateSchema(postgres_schema, if_not_exists=True))
```
willem-bd: Same mixed-case root cause in the ORM path: `CreateSchema(postgres_schema, ...)` is emitted **quoted** by SQLAlchemy's identifier preparer (case-preserved), but the engine's connection pins `search_path` via the **unquoted** asyncpg `server_settings` (`build_asyncpg_connect_args`). The models declare no explicit `schema=`, so `Base.metadata.create_all` relies entirely on `search_path`. With a mixed-case name the folded `search_path` misses the created schema and ORM tables (`runs`, `users`, ...) land in `public`. Fixed by the lowercase regex / consistent quoting above.

backend/packages/harness/deerflow/persistence/postgres_schema.py
```diff
@@ -0,0 +1,137 @@
...
+            index += 2
+            continue
+        if token.startswith("-csearch_path="):
+            index += 1
+            continue
+        merged.append(token)
+        index += 1
+
+    merged.extend(shlex.split(new_option))
+    return shlex.join(merged)
```
willem-bd: **🟠 `shlex.join` uses POSIX shell quoting, but libpq `options` does not.** A preserved option value containing a space (e.g. a DSN with `options=-c application_name=My App`) gets re-quoted by `shlex.join` as `-c 'application_name=My App'`. libpq's `options` parameter separates args on spaces **unless backslash-escaped** — single quotes are literal there, so libpq sees broken tokens and drops/corrupts the option. Existing tests only cover space-free values (`statement_timeout=5000`). Consider backslash-escaping spaces instead of `shlex.join`.

backend/packages/harness/deerflow/runtime/checkpointer/async_provider.py
```diff
@@ -47,26 +48,40 @@ def _prepare_database_sqlite_checkpointer_path(db_config) -> str:
...
+        "prepare_threshold": 0,
+        "row_factory": dict_row,
+        "keepalives": 1,
+        "keepalives_idle": 60,
+        "keepalives_interval": 10,
+        "keepalives_count": 6,
+    }
+    options = build_psycopg_options(schema)
+    if options is not None:
+        kwargs["options"] = options
```
willem-bd: **🟠 This overrides any `options` already in the DSN.** psycopg applies the explicit `kwargs["options"]` on top of the conninfo string, so a `connection_string`/`postgres_url` carrying e.g. `?options=-c%20statement_timeout%3D5000` loses `statement_timeout` on this async-checkpointer path only. The sync and DSN paths use `_merge_search_path_option`, which preserves other `-c` settings. Please route this through the same merge logic so behavior is consistent across paths.

backend/packages/harness/deerflow/runtime/store/async_provider.py
```diff
@@ -24,10 +24,29 @@
...
+    """Create the configured schema before LangGraph creates its store tables."""
+    statement = create_schema_sql(schema)
+    if statement is None:
+        return
+    try:
+        import psycopg
+    except ImportError as exc:
+        raise ImportError(POSTGRES_STORE_INSTALL) from exc
+
+    conn = await psycopg.AsyncConnection.connect(conn_string, autocommit=True)
```
willem-bd: **🟠 `+asyncpg` suffix not stripped here.** `_ensure_postgres_schema` hands the **raw** `conn_string` to `psycopg.AsyncConnection.connect`, but `dsn_with_search_path` (used 2 lines below for the store DSN) deliberately strips a SQLAlchemy `+driver` suffix because libpq only understands the bare `postgresql://` scheme. If `database.postgres_url` is set to `postgresql+asyncpg://...` (nothing validates against it), `CREATE SCHEMA` raises an opaque libpq parse error at startup. Same pattern in `store/provider.py:57`, `checkpointer/provider.py:57`, and `checkpointer/async_provider.py` (`_build_postgres_pool` at L69, which never routes through `dsn_with_search_path` at all). Normalize the scheme once before connecting, or validate/reject the `+driver` form on the config field.

backend/packages/harness/deerflow/runtime/store/provider.py
```diff
@@ -43,6 +45,21 @@
 )
 POSTGRES_CONN_REQUIRED = "checkpointer.connection_string is required for the postgres backend"
 
+
+def _ensure_postgres_schema(conn_string: str, schema: str) -> None:
```
willem-bd, line 49: **🟡 Cleanup (low).** `_ensure_postgres_schema` is duplicated near-verbatim across 4 sites (here, `checkpointer/provider.py:47`, `store/async_provider.py:33`, plus the pool variant `checkpointer/async_provider.py:76`), differing only by the install-hint constant and sync/async connect style. A future fix to the CREATE-SCHEMA flow has to be applied in all 4 or one path silently diverges. Consider a shared helper in `persistence/postgres_schema.py` parameterized by install-hint + a sync/async connector.

willem-bd (commented):
## Code review — custom postgres schema support

Thanks for the thorough PR and tests. The feature works for the common case (lowercase schema name, bare `postgresql://` URL, no embedded libpq `options`). I ran a high-recall pass and found a few defects that all fail **silently**, so they're worth closing before merge. Grouped by root cause:

### 🔴 1. Mixed-case / reserved-word schema names silently break (high)
The validation regex `^[A-Za-z_][A-Za-z0-9_]{0,62}$` allows uppercase + reserved words, but the schema is **CREATEd quoted** while `search_path` is **applied unquoted**. PostgreSQL folds unquoted `search_path` tokens to lowercase and silently ignores a non-existent entry, so for any name with a capital letter (e.g. `MySchema`) the schema is created as `MySchema` but `search_path` resolves to `myschema` → all ORM/checkpointer/store tables silently land in `public`. See inline comments on `config/postgres_schema.py`, `persistence/postgres_schema.py`, and `persistence/engine.py`.
**Fix:** restrict the regex to lowercase-only `^[a-z_][a-z0-9_]{0,62}$` (simplest, matches PG folding + the "plain identifiers only" contract), or quote the `search_path` value everywhere to match the quoted `CREATE SCHEMA`. Add a mixed-case case to the valid-identifier parametrize — every test currently uses lowercase `deerflow`, so this boundary is untested.

### 🟠 2. `+asyncpg` driver suffix stripped inconsistently (medium)
`dsn_with_search_path` deliberately strips a SQLAlchemy `+driver` suffix, but the raw-URL paths (`_ensure_postgres_schema` in both store/checkpointer providers, and `_build_postgres_pool`) do not — so a `postgresql+asyncpg://` URL crashes at startup in those paths. Inline on `store/async_provider.py`.

### 🟠 3. Async pool clobbers pre-existing libpq `options` (medium)
`checkpointer/async_provider.py` sets `kwargs["options"]` directly, overriding any `options` already in the DSN (e.g. `statement_timeout`), unlike the sync/DSN paths which merge. Inline comment.

### 🟠 4. `shlex.join` ≠ libpq `options` escaping (medium)
`_merge_search_path_option` reassembles with `shlex.join` (POSIX single-quote quoting), but libpq's `options` needs backslash-escaped spaces. Space-containing option values round-trip into a form libpq misparses. Inline comment.

### 🟡 5. Cleanup (low)
`_ensure_postgres_schema` is duplicated near-verbatim across 4 sites — consider a shared helper in `persistence/postgres_schema.py`. Also, `CREATE SCHEMA` runs on a separate dedicated connection in checkpointer + store + ORM at every startup (3+ extra handshakes); the idempotent DDL could ride the saver/engine's own connection.

**Must-fix:** #1. **Should-fix:** #2–#4 (all cheap, edge-config robustness).

WillemJiang:
@18062706139fcz  Please take a look at the latest code review.

WillemJiang:
@18062706139fcz  please resolve the conflicts between the main branch.

backend/packages/harness/deerflow/config/postgres_schema.py
```diff
@@ -0,0 +1,21 @@
...
+
+from __future__ import annotations
+
+import re
+
+# Lowercase-only on purpose. The schema is created *quoted*
+# (``CREATE SCHEMA IF NOT EXISTS "<schema>"``, case-preserved) but pinned via an
+# *unquoted* ``search_path`` token, which PostgreSQL folds to lowercase. Allowing
+# uppercase here would let the two diverge so tables silently land in ``public``.
+POSTGRES_SCHEMA_PATTERN = r"^[a-z_][a-z0-9_]{0,62}$"
```
willem-bd: The `$` here matches just before a single trailing `\n` (Python regex semantics), so `validate_postgres_schema("deerflow\n")` currently **passes** — verified: `re.match(r"^[a-z_][a-z0-9_]{0,62}$", "deerflow\n")` returns a match; `re.fullmatch(...)` does not.

This produces exactly the quoted-create-vs-unquoted-search_path divergence the comment above (lines 7–10) warns about:
- `create_schema_sql` emits `CREATE SCHEMA IF NOT EXISTS "deerflow\n"` — a *quoted* identifier, so PostgreSQL keeps the LF and creates a schema literally named `deerflow\n`.
- The `search_path` is injected *unquoted* (`-c search_path=deerflow\n`, asyncpg `server_settings`). PostgreSQL folds unquoted identifiers to lowercase and strips trailing whitespace, so it resolves to `deerflow` — which does not exist.
- Net: tables silently fall back to `public` with no error.

Realistic trigger: a YAML block scalar (`postgres_schema: |` / `  deerflow`) or an env-var substitution whose value carries a trailing LF. CRLF is correctly rejected (the `\r` breaks the char class), so only a bare LF slips through.

Fix: `re.fullmatch(r"[a-z_][a-z0-9_]{0,62}", value)` (or `\Z` instead of `$`), plus a regression test for `"deerflow\n"` and a block-scalar case.

backend/packages/harness/deerflow/persistence/postgres_schema.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def _join_libpq_options(tokens: list[str]) -> str:
+    """Join tokens into a libpq ``options`` string.
+
+    Spaces and backslashes inside a token are backslash-escaped so libpq
+    keeps each token intact. ``shlex.join`` cannot be used: it emits POSIX
+    shell quoting (single quotes), which libpq treats as literal characters.
+    """
+    escaped = [re.sub(r"([\\ ])", r"\\\1", token) for token in tokens]
```
willem-bd: The escaper only re-escapes `\` and space. `_split_libpq_options` correctly keeps a backslash-escaped whitespace byte as part of one token, but if that byte is a TAB / CR / LF it is left unescaped on re-join, and libpq then re-tokenizes on it — corrupting a user's pre-existing `options`.

Verified with a real TAB byte: input `-c application_name=My` + `\` + `<TAB>` + `App` splits to one token, rejoins with a bare TAB, and re-splits to two tokens (`['-c', 'application_name=My', 'App']`), so the round-trip is lossy.

Trigger is edge-case (escaped non-space whitespace inside an existing `options` value, e.g. an `application_name` containing a tab), hence medium severity. Fix: `re.sub(r"([\\\s])", r"\\\1", token)` so all whitespace is re-escaped.

backend/packages/harness/deerflow/persistence/postgres_schema.py
```diff
@@ -0,0 +1,235 @@
...
+
+    merged.extend(_split_libpq_options(new_option))
+    return _join_libpq_options(merged)
+
+
+def create_schema_sql(schema: str) -> str | None:
+    """Return a safe CREATE SCHEMA statement for a validated plain identifier."""
+    if not schema:
+        return None
+    return f'CREATE SCHEMA IF NOT EXISTS "{schema}"'
```
willem-bd, line 143: Defense-in-depth: this f-string is the SQL boundary, but the "validated plain identifier" contract is enforced only by the distant pydantic validator, not here. `ensure_postgres_schema` / `create_schema_sql` are publicly exported, and psycopg3's `Connection.execute` accepts multiple `;`-separated statements — so a future caller that bypasses the validator (e.g. a test helper) could inject.

No current call site bypasses the validator (every path routes through `DatabaseConfig` / `CheckpointerConfig`), so there is no live exploit. Suggestion: add `assert re.fullmatch(r"[a-z_][a-z0-9_]*", schema)` at the top of `create_schema_sql` (or render via SQLAlchemy's identifier quoting) so the safety property is local to the SQL-emitting function rather than enforced at a distance.

backend/packages/harness/deerflow/persistence/migrations/env.py
```diff
@@ -72,7 +72,22 @@ def do_run_migrations(connection):
...
+    # When a custom Postgres schema is configured, pin the alembic-spawned
+    # engine's search_path to it. This engine is built from the bare URL and
+    # does NOT inherit the asyncpg ``server_settings`` the app engine sets via
+    # ``connect_args``, so without this both ``alembic_version`` and every
+    # migration's DDL would land in the default (``public``) schema while the
+    # ORM tables land in the custom schema. ``init_engine`` has already created
+    # the schema (``CREATE SCHEMA IF NOT EXISTS``) before bootstrap runs.
+    pg_schema = config.get_main_option("deerflow_pg_schema")
+    connect_args: dict = {}
+    if pg_schema and url and url.startswith("postgresql"):
```
willem-bd: Minor: `url.startswith("postgresql")` misses the `postgres://` short scheme that libpq accepts. If a user set `postgres_url: postgres://...`, `DatabaseConfig.app_sqlalchemy_url`'s `postgresql://`→`postgresql+asyncpg://` rewrite wouldn't fire either, so alembic would skip `search_path` injection here and `alembic_version` + migration DDL would land in the default schema.

Low impact in practice — SQLAlchemy 2.x rejects `postgres://` for an async engine (so it surfaces as a connection error rather than silent misrouting) and every documented example uses `postgresql://`. Cheap belt-and-braces fix: also accept `postgres`, or normalize the scheme once at the `DatabaseConfig` boundary.

willem-bd (commented):
Reviewed the postgres-schema implementation end to end (validator + helpers, `engine.py`, `bootstrap.py`, alembic `env.py`, all four checkpointer/store providers, and `app_config` reload) plus the test suite. The design is sound and coverage is broad, but a Python regex gotcha in the validator silently breaks the core safety property — see the inline comment on `config/postgres_schema.py`. Three other inline notes below.

Two items not tied to a single line:

**PR scope.** `git diff main...pr-3442` also includes the entire `backend/packages/harness/deerflow/workspace_changes/` package (~1300 lines), its `runtime/runs/worker.py` wiring, the OpenAI-thinking setup-wizard changes (`scripts/wizard/`), matching frontend, and their tests — none on `main`, none mentioned in the title/body. They came in via the merge commits `aec979c6` / `f722861c` (commits `00161811` #3945 and `f0f9dd66` #3428). A reviewer of "support custom postgres schema" would otherwise ship a second feature unreviewed — consider rebasing onto current `main` so those land under their own PRs.

**Upgrade data-stranding (documented; enhancement only).** Switching an existing deployment from default `public` to `postgres_schema` makes `_reflect_state` (search_path now pinned) see the target schema as empty, so bootstrap re-runs `create_all + stamp head` while the user's real data sits stranded in `public`. The migration steps in the PR body cover it; an optional startup `logger.warning` when `postgres_schema` is set and pre-existing DeerFlow tables are detected in the default schema would catch operators who miss that section.

Happy to follow up with fix patches for the two confirmed bugs if useful.

backend/packages/harness/deerflow/persistence/postgres_schema.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+    statement = create_schema_sql(schema)
+    if statement is None:
+        return
+    try:
+        import psycopg
+    except ImportError as exc:
+        raise ImportError(install_hint) from exc
+
+    with psycopg.connect(normalize_libpq_dsn(conn_string), autocommit=True) as conn:
```
willem-bd: In psycopg 3 (the project pins `psycopg[binary]>=3.3.3`), `Connection.__exit__` only runs `commit()` / `rollback()` — it does **not** call `close()`. That is the documented psycopg2→3 migration pitfall (https://www.psycopg.org/psycopg3/docs/api/connections.html#connection-as-a-context-manager), so this `with` block commits the `CREATE SCHEMA` but leaves the underlying libpq connection open.

The async counterpart just below already gets this right with `try / finally: await conn.close()`. The sync path should mirror it:

```python
conn = psycopg.connect(normalize_libpq_dsn(conn_string), autocommit=True)
try:
    conn.execute(statement)
finally:
    conn.close()
```

On CPython the leaked connection is collected promptly by refcounting when the function returns, so in the common runtime this is a correctness/robustness gap rather than an active leak — but `_ensure_postgres_schema` is reached on every sync checkpointer/store setup and every CLI `checkpointer_context()` call, and on non-refcounting runtimes (or if a reference cycle retains the object) the idle connections would accumulate against the server's `max_connections`.

backend/packages/harness/deerflow/config/app_config.py
```diff
@@ -343,8 +345,12 @@ def _apply_singleton_configs(cls, config: Self, acp_agents: dict[str, ACPAgentCo
         load_stream_bridge_config_from_dict(config.stream_bridge.model_dump() if config.stream_bridge is not None else None)
         load_acp_config_from_dict({name: agent.model_dump() for name, agent in acp_agents.items()})
 
-        if previous_checkpointer_config != config.checkpointer:
-            # These runtime singletons derive their backend from checkpointer config.
+        database_drives_persistence = previous_checkpointer_config is None and config.checkpointer is None
+        database_config_changed = database_drives_persistence and previous_database_config != config.database
```
willem-bd: This new branch resets the sync checkpointer and store singletons when `config.database` changes — but the ORM engine is intentionally **not** rebuilt. `reload_boundary.py` documents `database` as restart-required with: `"init_engine_from_config() runs once during langgraph_runtime() startup; the SQLAlchemy engine holds the connection pool and is not rebuilt on config.yaml edits."`

The asymmetry produces a split-brain on a runtime `postgres_schema` edit. Operator starts with `postgres_schema: v1`, edits to `v2`, and touches the file:
- this branch fires `database_config_changed` → `reset_checkpointer()` / `reset_store()` → the next `get_checkpointer()` / `make_store()` rebuilds against `v2`;
- but the ORM engine still carries `server_settings={"search_path": "v1"}` (pinned in `engine.py` via `build_asyncpg_connect_args`);
- so new checkpoint/store tables land in `v2` while new ORM rows (`runs`, `threads_meta`, …) keep landing in `v1`, with no error surfaced.

Since `database` is already classified restart-required, the cleaner fix is probably to **not** partially reset here either — drop the `or database_config_changed` clause and let a `database` / `postgres_schema` change require the documented restart, so the deployment stays self-consistent instead of half-migrating. (If partial reset is intentional, the ORM engine would have to be reset too, which is the riskier path.)

willem-bd (commented):
Two follow-up findings from a deeper second pass over the engine / provider wiring. Both verified against the code on this branch.

WillemJiang:
@18062706139fcz, please resolve the conflicts.

fancyboi999 (changes requested):
Thanks @18062706139fcz. I found one migration issue that can strand existing production data when this option is enabled.

[P1] Move every DeerFlow table and Alembic state during the schema migration

- Location: `config.example.yaml:1660`
- Problem: The documented upgrade moves only `runs`, `run_events`, `threads_meta`, `feedback`, and `users`. Existing channel connection data, scheduled tasks/runs, and Alembic state stay in `public`; after restart DeerFlow creates/replays empty counterparts in the target schema, so those existing rows disappear from the application's view.
- Evidence: Current `origin/main` also owns `channel_connections`, `channel_credentials`, `channel_oauth_states`, `channel_conversations`, `scheduled_tasks`, and `scheduled_task_runs`, and bootstrap uses the schema-local `alembic_version` table to decide whether the database is versioned. The instructions omit all of those (and the PR body repeats the same five-table SQL). With the five listed tables moved but no `alembic_version`, `_reflect_state` classifies the target schema as legacy, backfills empty baseline channel tables, stamps `0001_baseline`, and applies the scheduled-task revision while the real rows remain in `public`.
- Suggested fix: Replace the partial hard-coded migration with a complete, version-aware procedure that moves all DeerFlow ORM tables, LangGraph tables, and `alembic_version` (or provide a checked migration script that discovers the owned set). Explicitly cover the four `channel_*` tables and both `scheduled_*` tables present today.
- Test: Seed rows in every DeerFlow table in `public`, run the documented migration, restart/bootstrap against `postgres_schema`, and assert the seeded channel/scheduler/application rows remain visible and no DeerFlow-owned tables are left behind in `public`.


Reviewed head SHA: `f3ebdc681cc7c2d269cc5e3815cf557de6de3c22`

backend/packages/harness/deerflow/runtime/checkpointer/provider.py
```diff
@@ -140,6 +148,24 @@ def _sync_checkpointer_cm(config: CheckpointerConfig) -> Iterator[Checkpointer]:
     raise ValueError(f"Unknown checkpointer type: {config.type!r}")
 
 
+@contextlib.contextmanager
+def _sync_checkpointer_from_database(db_config) -> Iterator[Checkpointer]:
```
willem-bd: **Suggestion (cleanup):** This helper is only reached by the optional, env-gated integration test (`test_pg_schema_integration.py`, skipped unless `DEERFLOW_TEST_POSTGRES_URL` is set) - it isn't exported from `runtime/checkpointer/__init__.py` and no production path calls it. It also re-implements the `DatabaseConfig -> CheckpointerConfig` backend resolution that `_resolve_checkpointer_config` (L57) already owns, so the backend-selection logic now lives in two places.

The store side of the same integration test already avoids a dedicated helper by reusing the production path directly (`_resolve_store_config(...)` + `_sync_store_cm(...)`). The checkpointer side could mirror that with `_resolve_checkpointer_config(app_config)` + `_sync_checkpointer_cm(config)`, and this helper could then be dropped.

Why it matters: the only caller is skipped in CI, so a future backend added to `_resolve_checkpointer_config` wouldn't be reflected here and the drift would stay silent. If this is intended as a public entry point, export it and add a non-skipped test; otherwise removing it keeps the backend-resolution contract in one place.

willem-bd (commented):
Reviewed head `0dfb826b`. The prior review rounds look fully addressed at this head: the validator is lowercase-only with `re.fullmatch` (mixed-case / trailing-newline fixed), `_merge_search_path_option` backslash-escapes whitespace (shlex.join fixed), the async pool routes through `dsn_with_search_path(normalize_libpq_dsn(...))` (options-clobber + `+asyncpg` fixed), `ensure_postgres_schema` uses `try/finally: close()`, `env.py` accepts the `postgres` short scheme, `app_config` no longer half-resets on a `database` change (covered by the new reload test), and the migration note now moves every DeerFlow table plus `alembic_version`.

One new minor suggestion inline: `_sync_checkpointer_from_database` is test-only code that duplicates `_resolve_checkpointer_config`. Also note the PR body still quotes the schema pattern as `^[A-Za-z_][A-Za-z0-9_]{0,62}$` (allows uppercase), but the implemented/validated pattern is lowercase-only `[a-z_][a-z0-9_]{0,62}` - worth updating the body so operators aren't surprised that `MySchema` is rejected.

WillemJiang (approved):

PR #3453, opened 2026-06-08
fix(replay-e2e): key fixtures by caller and conversation

### LINKED ISSUE #3452: [bug] Replay E2E should key model fixtures by caller + conversation, not conversation alone
## Problem

PR #3436 correctly excludes the living system prompt from replay matching to avoid false CI failures when lead-agent prompt text changes.

However, the new conversation-only key can collapse distinct model callers into the same replay bucket when they share the same non-system conversation input. For example, `lead_agent`, `middleware:title`, `suggest_agent`, or a future `subagent:*` call could have identical human/ai/tool-visible input after system messages are removed.

That weakens the original replay design goal: matching by stable input rather than depending on model-call order.

## Why It Matters

If two callers share one conversation hash, replay falls back to bucket order via `deque.popleft()`. Then unrelated changes such as middleware timing, suggestion trigger timing, or a new model call can cause the wrong recorded output to be returned.

This turns a local caller/timing change into fixture churn or confusing replay failures.

## Proposed Fix

Include a stable caller identity in the replay key:

```text
input_hash = hash(caller + conversation_hash)
```

Caller identity should come from existing LangChain callback metadata:

- prefer explicit tags such as `middleware:title`, `subagent:*`, `lead_agent`
- otherwise use callback `name` / `run_name`, such as `suggest_agent`
- default to `lead_agent` when no caller marker exists

Keep system prompt excluded from the conversation hash so prompt wording changes do not stale fixtures.

## Fixture Format

New fixture turns should include:

```json
{
  "caller": "lead_agent",
  "conversation_hash": "<conversation-only hash>",
  "input_hash": "<caller-aware hash>",
  "output": {}
}
```

Existing conversation-only fixtures should remain replayable through a legacy fallback.

## Acceptance Criteria

- Recorder writes `caller`, `conversation_hash`, and caller-aware `input_hash`.
- Replayer matches by caller-aware key.
- Legacy fixtures with only `input_hash` still work.
- A focused test proves two callers with identical conversation text return different recorded outputs.
- Existing `write_read_file.ultra` replay golden remains green.
- Replay E2E docs explain caller + conversation keying.

ggnnggez:
Closes #3452 
## What changed

- Namespaces replay fixture matching by stable model caller plus conversation hash.
- Keeps the living system prompt excluded from the conversation hash, so prompt wording edits do not stale fixtures.
- Records `caller`, `conversation_hash`, and caller-aware `input_hash` for new fixtures, while retaining legacy conversation-only fixture fallback.
- Migrates the existing `write_read_file.ultra` fixture and updates replay-e2e docs.
- Adds focused tests for identical conversations from different callers.

## Why

#3436 fixed prompt drift by excluding system messages from replay matching, but a conversation-only key can collapse distinct model callers into the same replay bucket. That can make replay depend on call order again when callers share identical non-system messages.

This keeps replay stable against system prompt edits while preserving caller isolation for `lead_agent`, `middleware:title`, `suggest_agent`, and future `subagent:*` calls.

## Validation

- `uv run ruff check tests/replay_provider.py tests/test_replay_provider.py scripts/record_gateway.py scripts/build_fixture_from_jsonl.py`
- `PYTHONPATH=. uv run pytest tests/test_replay_provider.py tests/test_replay_golden.py -q`


fancyboi999 (commented):
Nice, keying by caller is the right move here. It closes the same-conversation / different-caller collision that the conversation-only key left open (the thing Copilot and I had flagged on #3436).

I think the red Layer 2 is a real catch, not a flake. Layer 1 (the backend golden) is green, and it still asserts zero replay misses, so the `middleware:title` turn matches fine when the run is driven straight through the gateway. Layer 2 drives the actual frontend and misses on that same title turn, so the auto-title never renders. The caller you resolve for the title call comes out different depending on how the run started: `middleware:title` from `drive_gateway`, something else from the browser/SDK run.

Reading the code: `_caller_from_run_manager` first uses whatever `_ReplayCallerCapture.on_chat_model_start` stashed by run_id (that one sees `name` + `tags`), and if there's nothing there it falls back to `caller_identity(tags=run_manager.tags)` with tags only. For the title turn those two don't agree across paths. My guess is the `middleware:title` tag isn't on the model call's run_manager/callbacks when the run comes from the frontend SDK (different stream_mode/config), so it falls through to the tags-only branch and lands on something other than `middleware:title`. The `Caller:` value your miss message already prints in the Layer 2 log will say exactly what it resolved to.

Separate from the title, one thing worth tightening while you're in here: record and replay resolve the caller differently. Recording goes through `caller_identity(name=name, tags=tags)`, but the replay fallback is `caller_identity(tags=...)` with no `name`. Your docstring says route-level callers like `suggest_agent` come from `name`, so the suggestions turn would also miss on replay any time the capture-by-run-id path doesn't get populated. Probably worth routing both sides through one derivation so they can't drift apart.

Happy to pull the exact frontend-path caller for the title if it saves you time. The fix is making the resolved caller identical across entry paths, and it'd be worth a small unit test pinning that (driven vs frontend), since that's the gap Layer 2 is catching.

WillemJiang:
@ggnnggez please take a look at the test failure of e2e

WillemJiang (approved):

PR #3456, opened 2026-06-09
Feat/stepfun support

hata33:
## Why

StepFun (阶跃星辰) provides OpenAI-compatible reasoning models (step-3.7-flash, step-3.5-flash), but LangChain's `ChatOpenAI` silently drops the non-standard `reasoning` / `reasoning_content` fields from both streaming deltas and non-streaming responses. This breaks reasoning visibility in multi-turn conversations, and causes API errors when the provider expects `reasoning_content` to be echoed back on historical assistant messages.

## What changed

- **New adapter**: `PatchedChatStepFun` captures reasoning from all response paths (streaming + non-streaming) and replays it on historical assistant messages via `_get_request_payload`, mirroring the approach in `PatchedChatMiMo`.
- **Dual field support**: Handles both `reasoning` (StepFun default) and `reasoning_content` (deepseek-style) field names.
- **17 unit tests**: Cover streaming capture, non-streaming capture, payload replay, SDK object handling, and noop paths.
- **Config example**: Added StepFun configuration block to `config.example.yaml`.

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [x] **Agents / LangGraph** — new model adapter under `packages/harness/deerflow/models/`
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies** — no new dependencies
- [x] **Docs / tests / CI only** — config example + tests

## Validation

```bash
cd backend && PYTHONPATH=. uv run pytest tests/test_patched_stepfun.py -v
# 17 passed


WillemJiang:
@hata33 thanks for the PR, but it contains a lot of changes which are not related to the PR topic. 
Please pick the right change set in this PR.

PR #3084, opened 2026-05-20
fix(runtime): harden JSONL async I/O and DB put_batch thread validation

64johnlee:
Closes #2816

## Problem
Two related reliability issues in the run event store:

1. **JSONL store** (`JsonlRunEventStore`) used synchronous file I/O (`open()`, `f.read()`, `f.write()`) inside `async` methods. Under asyncio this blocks the event loop during every read/write, causing latency spikes under concurrent runs.

2. **DB store** (`DbRunEventStore.put_batch()`) accepted cross-thread event batches silently, which could corrupt seq ordering when callers accidentally mixed thread IDs.

## Fix
- `JsonlRunEventStore` now offloads all blocking file I/O to a thread pool via `asyncio.to_thread`, keeping the event loop free. Per-thread `asyncio.Lock` objects serialise concurrent writes within a single process.
- `DbRunEventStore.put_batch()` validates that all events in a batch share the same `thread_id` and raises `ValueError` immediately for cross-thread batches.
- `delete_by_thread()` now pops the per-thread lock from `_write_locks` (alongside `_seq_counters`) to prevent unbounded memory growth in workloads that repeatedly create and delete threads.

## Tests
`test_jsonl_event_store_async_io.py` — 13 tests covering:
- `asyncio.to_thread` is invoked for `_write_record` (monkeypatch regression guard)
- Concurrent puts produce unique, monotonic seq values
- seq counter and write lock are both cleared after `delete_by_thread`
- DB batch validation rejects mixed-thread event batches

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

Issue #1917, opened 2026-04-06
load_skills() uses blocking os.walk inside LangGraph async event loop — breaks tool calls under `langgraph dev`


JerryChaox:
## Summary

Every turn, the lead agent rebuilds its system prompt via `_get_enabled_skills()` in [`backend/packages/harness/deerflow/agents/lead_agent/prompt.py`](backend/packages/harness/deerflow/agents/lead_agent/prompt.py), which calls `load_skills(enabled_only=True)`. `load_skills()` walks the skills directory with `os.walk(category_path, followlinks=True)` in [`backend/packages/harness/deerflow/skills/loader.py`](backend/packages/harness/deerflow/skills/loader.py) (around line 66). This is a synchronous filesystem scan executed on the asyncio event loop, which `blockbuster` (shipped with `langgraph dev`) intercepts and raises:

```
blockbuster.blockbuster.BlockingError: Blocking call to ScandirIterator.__next__
  File ".../deerflow/skills/loader.py", line 66, in load_skills
    for current_root, dir_names, file_names in os.walk(category_path, followlinks=True):
  File ".../deerflow/agents/lead_agent/prompt.py", line 13, in _get_enabled_skills
    return list(load_skills(enabled_only=True))
```

Because this happens during prompt construction, the resulting exception cascades into the middleware chain: **the tool call the agent just made appears to fail**, and the model surfaces a generic "blocking call issue" to the user — even though the tool itself (e.g. a web fetch tool) executed fine. From the end user's perspective, the agent is suddenly unable to use any tool after a cold start or an update to the `skills/` directory.

## Why it appears intermittently

With a flat `skills/custom/*/SKILL.md` layout the scandir count is small and some runs slip past, so users may not hit it. Adding **any** custom skill that ships with subdirectories (e.g. `skill-name/scripts/`, `skill-name/references/`) increases the scandir iteration count and makes the error reliably reproducible on every turn — i.e. adding a well-structured skill bundle (the standard Anthropic skill layout) turns an occasional warning into a hard failure.

## Reproduction

1. `make dev` (starts `langgraph dev --no-browser --no-reload …`).
2. Add a custom skill with nested directories:
   ```
   skills/custom/example-skill/
     SKILL.md
     scripts/helper.py
     references/notes.md
   ```
3. Open the web UI, send any message that causes a tool call (e.g. ask the agent to fetch a URL).
[...]

PR #3457, opened 2026-06-09
fix(agents): offload blocking filesystem IO in the custom-agent router off the event loop

ly-wang19:
## Why

`make detect-blocking-io` flags the custom-agent management routes: both async handlers touch the filesystem directly on the event loop, blocking it for the duration of every create/delete:

- `create_agent_endpoint` — resolves + creates the agent directory, writes `config.yaml` and `SOUL.md`, reads the config back, and `shutil.rmtree`s on failure.
- `delete_agent` — resolves the agent directory (`Paths.base_dir` → `Path.resolve`), probes it (`Path.exists`), and `shutil.rmtree`s it.

Under concurrency these stall every other request sharing the event loop — the recurring class of issue DeerFlow already guards against in `tests/blocking_io/` (e.g. #1917, #1912, #3084).

## What changed

Both handlers now run their filesystem section in a small sync helper dispatched via `asyncio.to_thread`. Externally observable behavior is unchanged — same `201/409/500` (create) and `204/404/409/500` (delete) responses, same cleanup-on-failure semantics.

## Surface area

- [x] **Backend API** — `app/gateway/routers/agents.py` (custom-agent routes) + tests. No request/response shape change.

## Validation

- `cd backend && uv run pytest tests/blocking_io -q` → **9 passed**, including a new `tests/blocking_io/test_agents_router.py` anchor that drives both handlers under the strict Blockbuster gate (it fails if the IO regresses back onto the event loop — verified red→green while developing).
- `cd backend && uv run pytest tests/test_custom_agent.py -q` → **55 passed** (create/delete behavior unchanged).
- `ruff check` + `ruff format` → clean.

## AI assistance

**How you used it:** AI ran `make detect-blocking-io`, verified the findings against the strict Blockbuster gate, implemented the `asyncio.to_thread` offload for both handlers, and wrote the regression anchor. I reviewed the change.


ShenAC-SAC:
@ly-wang19 thanks for the PR.

Please address Copilot's review comments first.

Additional review point:
- Please add `tests/blocking_io/` coverage for:
  - create failure after the agent directory is created, to verify cleanup runs off the event loop;
  - delete missing-agent -> 404;
  - delete legacy-only -> 409.

These tests are mainly to let runtime detection prevent regressions.

WillemJiang (approved):

WillemJiang:
> @ly-wang19 thanks for the PR.
> 
> Please address Copilot's review comments first.
> 
> Additional review point:
> 
> * Please add `tests/blocking_io/` coverage for:
>   
>   * create failure after the agent directory is created, to verify cleanup runs off the event loop;
>   * delete missing-agent -> 404;
>   * delete legacy-only -> 409.
> 
> These tests are mainly to let runtime detection prevent regressions.

It would be great if we had a skill that could generate tests by looking up these entry points.

PR #3461, opened 2026-06-09
feat(models): add StepFun reasoning model adapter

hata33:
This PR replaces #3456 (closed). Previous PR contained unrelated changes; this one is scoped to the StepFun adapter only.

## Why

StepFun provides OpenAI-compatible reasoning models (step-3.7-flash, step-3.5-flash). However, LangChain's ChatOpenAI drops the non-standard `reasoning` / `reasoning_content` fields. This causes:
- Loss of reasoning visibility in multi-turn conversations
- API errors when historical assistant messages miss `reasoning_content`

## What changed

- Add `PatchedChatStepFun` adapter that captures reasoning from both streaming and non-streaming responses
- Replay reasoning onto historical assistant messages in `_get_request_payload`
- Support both `reasoning` and `reasoning_content` field names
- Add 17 unit tests

## Surface area

- [ ] Frontend UI
- [ ] Backend API
- [x] **Agents / LangGraph** — new model adapter
- [ ] Sandbox
- [ ] Skills
- [ ] Dependencies
- [ ] Default behavior change
- [x] **Docs / tests / CI only** — config example + tests

## Validation

```bash
cd backend && PYTHONPATH=. uv run pytest tests/test_patched_stepfun.py -v
# 17 passed

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -274,6 +274,32 @@ models:
   #       thinking:
   #         type: disabled
 
+  # Example: StepFun (阶跃星辰) reasoning models
+  # StepFun provides OpenAI-compatible API with reasoning models.
+  # With reasoning_format: deepseek-style, the API returns reasoning_content
+  # (same field as DeepSeek), which must be replayed on historical assistant
+  # messages in multi-turn tool-call conversations.
+  # Use PatchedChatStepFun instead of plain ChatOpenAI.
+  # Docs: https://platform.stepfun.com/docs/api-reference/chat-completions
+  # - name: step-3.7-flash
+  #   display_name: Step 3.7 Flash
+  #   use: deerflow.models.patched_stepfun:PatchedChatStepFun
+  #   model: step-3.7-flash
+  #   api_key: $STEPFUN_API_KEY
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: config.example.yaml [hunk 1]
```diff
@@ -274,12 +274,38 @@ models:
...
+  #     extra_body:
+  #       reasoning_format: deepseek-style
+
   # Example: MiniMax (OpenAI-compatible) - International Edition
   # MiniMax provides high-performance models with 512K context window and 128K max output
   # Docs: https://platform.minimax.io/docs/api-reference/text-openai-api
   # - name: minimax-m3
   #   display_name: MiniMax M3
-  #   use: deerflow.models.patched_minimax:PatchedChatMiniMax
+  #   use: langchain_openai:ChatOpenAI
```
WillemJiang: Please revert the below change which is not related to your PR.

hata33: @WillemJiang Thanks for the review. I've fixed it in commit ebe615d — now only the StepFun config section is added, and the MiniMax config remains untouched. PTAL again. Thanks!

WillemJiang (approved):

PR #3018, opened 2026-05-17
[codex] support slash skill activation

hetaoBackend:
> Follow-up hardening and latest-main integration are continued in #3466.

## Summary

Implements the MVP for slash skill activation from RFC #3017.

This PR keeps DeerFlow's base skill prompt progressive-disclosure friendly: startup/reload still exposes skill metadata only, while an explicit `/skill-name ...` request deterministically injects that skill's full `SKILL.md` into the current turn as hidden context.

## Changes

- Add `deerflow.skills.slash` for slash skill parsing and enabled/whitelisted skill resolution.
- Add `SkillActivationMiddleware` to inject full `SKILL.md` content when the latest user message starts with `/skill-name`.
- Wire slash activation into the lead-agent runtime and embedded `DeerFlowClient`, preserving custom-agent skill whitelists.
- Update skill prompt guidance to document explicit slash activation.
- Let IM channel commands fall back to chat handling when an unknown slash command matches an enabled skill.
- Add unit coverage for resolver behavior, middleware injection, whitelist enforcement, processed-message deduping, and channel fallback.

## Notes

- Slash activation is current-turn only and does not mutate thread or agent skill config.
- Existing control commands such as `/new`, `/help`, and `/bootstrap` retain precedence.
- Disabled skills are not loaded because activation resolution reads `load_skills(enabled_only=True)`.
- `allowed-tools` policy is not bypassed; tool filtering remains based on the current runtime-visible skill set.

## Validation

- `uv run ruff check packages/harness/deerflow/skills/slash.py packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py packages/harness/deerflow/agents/lead_agent/agent.py packages/harness/deerflow/client.py app/channels/manager.py tests/test_slash_skills.py tests/test_channels.py`
- `uv run pytest -q tests/test_slash_skills.py`
- `uv run pytest -q tests/test_channels.py -k 'slash_skill or command_help or command_new or command_bootstrap'`
- `uv run pytest -q tests/test_lead_agent_prompt.py tests/test_subagent_skills_config.py`
- `uv run pytest -q tests/test_skills_custom_router.py tests/test_skills_installer.py tests/test_skills_bundled.py`
- `uv run pytest -q tests/test_client.py -k 'SkillsManagement or ScenarioSkillInstallAndUse or InstallSkillSecurity'`
- `uv run pytest -q` => 3392 passed, 19 skipped

Refs #3017

Issue #3017, opened 2026-05-17
RFC: Support Slash Skill activation


hetaoBackend:
## Background

DeerFlow already supports Agent Skills. Today the lead-agent system prompt exposes enabled skill metadata (`name`, `description`, and `location`) and asks the model to `read_file` the relevant `SKILL.md` on demand. This matches the progressive disclosure pattern: metadata stays resident, the full skill body is loaded only when activated, and referenced resources are loaded incrementally.

The missing piece is a deterministic explicit activation path. When a user types `/skill-name ...`, DeerFlow should activate that skill directly instead of relying only on model judgment. IM channels currently treat every leading slash as a channel command, so an unknown `/skill-name` is rejected before it can reach the agent.

## Goals

- Support `/skill-name task` as an explicit, deterministic skill activation syntax.
- Keep startup/reload prompt injection limited to skill `name + description + path`.
- Load full `SKILL.md` only when a user explicitly activates a skill or when the model decides a listed skill is relevant and reads it through the existing progressive loading path.
- Respect global skill enabled state and custom-agent skill whitelists.
- Preserve existing skill install/edit/enable behavior and custom-over-public override semantics.
- Keep the base system prompt static where possible to preserve prefix-cache friendliness.

## Non-goals for MVP

- No skill marketplace, signing, native platform command registration, or tool-direct command dispatch in the first iteration.
- No persistent mutation of a thread or agent's default skill config when a slash skill is used.
- No automatic execution of skill scripts or support resources. `references/`, `templates/`, `scripts/`, and `assets/` remain progressively loaded as needed.

## Current integration points

- Skill prompt metadata: `backend/packages/harness/deerflow/agents/lead_agent/prompt.py`
- Skill storage and enabled state: `backend/packages/harness/deerflow/skills/storage/skill_storage.py`
- Skill parser/types: `backend/packages/harness/deerflow/skills/parser.py`, `backend/packages/harness/deerflow/skills/types.py`
- Lead agent runtime and custom-agent skill whitelist: `backend/packages/harness/deerflow/agents/lead_agent/agent.py`
- IM slash command dispatch: `backend/app/channels/manager.py`, `backend/app/channels/commands.py`
- Skill management API: `backend/app/gateway/routers/skills.py`

### 1. Add a slash skill resolver

Create a shared resolver that takes user text plus the runtime-visible skills and returns either no match or a validated activation:

```text
[...]

PR #3466, opened 2026-06-09
fix(skills): harden slash skill activation across chat channels

hetaoBackend:
## Summary

Continues and hardens the slash skill activation work from #3018 after rebasing/merging it onto the latest `main`.

This PR keeps the progressive-disclosure model from RFC #3017: DeerFlow exposes skill metadata by default, while an explicit `/skill-name ...` turn injects that skill's full `SKILL.md` as hidden model context. The additional work here tightens parsing, IM-channel routing, upload preservation, frontend transcript/export safety, and regression coverage.

## Changes

- Align the slash skill activation branch with latest `main` and resolve integration conflicts.
- Add strict slash skill parsing for `/skill-name <task>` and reserve built-in channel commands such as `/new`, `/help`, `/status`, `/models`, `/memory`, and `/bootstrap`.
- Resolve installed, enabled, and agent-whitelisted skills explicitly, with clear errors for disabled or unavailable skills.
- Harden `SkillActivationMiddleware` with safe skill-file reads, XML escaping, audit events, upload-aware original user text, hidden/summary filtering, and duplicate activation prevention.
- Preserve slash skill activation through uploaded-file middleware so `/skill-name ...` plus attachments still activates the intended skill.
- Update IM channel routing so unknown slash text is no longer blindly treated as a control command; slash skills can flow through chat while known commands keep precedence.
- Cover channel-specific edge cases for Telegram bot suffixes, Slack bot mentions, Discord mentions, topic-thread replies, and disabled/uninstalled skills.
- Prevent hidden slash activation context from leaking into frontend message display, grouping, or JSON/Markdown exports.
- Add Playwright coverage for slash skill chat submission, slash skill plus attachment metadata, and pending attachment visibility.

## User Impact

Users can invoke enabled skills with `/skill-name ...` reliably from the web UI and supported IM channels. Existing channel commands still work, while accidental command execution from malformed slash text, paths, or non-bot mentions is avoided. Hidden `SKILL.md` activation context is kept out of visible transcripts and exports.

## Validation

- `PYTHONPATH=. uv run pytest -q` => 4185 passed, 15 skipped
- `pnpm run check`
- `pnpm test` => 205 passed
- `pnpm exec playwright test --project=chromium` => 23 passed
- `git diff --check`

Refs #3017
Builds on #3018


LittleChenLiya (commented):
## 问题：Shift+Enter 与斜杠技能自动完成冲突

**文件：** `frontend/src/components/workspace/input-box.tsx` — `handleSkillSuggestionKeyDown`

当技能建议下拉菜单可见时，用户按 `Shift+Enter`（意图插入换行符），但 `event.key === "Enter"` 分支未检查 `event.shiftKey`，导致选中技能而非插入换行。

### 调用链路

```
PromptInputTextarea.handleKeyDown (prompt-input.tsx:894)
  └─ onKeyDown?.(e)  →  handleSkillSuggestionKeyDown (input-box.tsx:440)
       └─ event.key === "Enter" 匹配（含 Shift+Enter）           ← 第 463 行
       └─ event.preventDefault() 执行，defaultPrevented = true    ← 第 464 行
       └─ applySkillSuggestion() 执行，选中技能                   ← 第 467 行
  └─ e.defaultPrevented === true → 直接 return                   ← 第 896 行
  └─ (永远执行不到) if (e.shiftKey) { return; }                  ← 第 903 行
```

### 根因

`handleSkillSuggestionKeyDown` 的 Enter/Tab 分支没有对 `event.shiftKey` 做守卫。子组件调用了 `preventDefault()`，导致父组件 `prompt-input.tsx` 中第 903 行正确的 Shift+Enter 换行逻辑被跳过。

### 修复建议

在 `handleSkillSuggestionKeyDown` 的 Enter/Tab 分支开头添加 `event.shiftKey` 守卫：

```typescript
if (event.key === "Enter" || event.key === "Tab") {
  if (event.shiftKey) return;  // 让 Shift+Enter 回退到 textarea 的换行逻辑
  event.preventDefault();
  const selectedSkill = skillSuggestions[skillSuggestionIndex];
  if (selectedSkill) {
    applySkillSuggestion(selectedSkill);
  }
  return;
}
```

### 测试建议

在 `frontend/tests/e2e/chat.spec.ts` 中补充 e2e 测试，验证下拉可见时 Shift+Enter 不会选中技能，而是正常插入换行符。

WillemJiang:
@hetaoBackend  Here are some other review comments beside @LittleChenLiya  point 
1: _resolve_slash_skill_command in manager.py loads skills twice

backend/app/channels/manager.py — _resolve_slash_skill_command

```
def _resolve_slash_skill_command(text, ...):
    if parse_slash_skill_reference(text) is None:
        return None
    try:
        resolved_storage = storage() if callable(storage) else storage or get_or_new_skill_storage()
        skills = resolved_storage.load_skills(enabled_only=False)       # ← first load
        reference = parse_slash_skill_reference(text)                   # ← parses again
        ...
        skill = next((candidate for candidate in skills ...), None)     # ← manual lookup
        ...
        resolved = resolve_slash_skill(text, skills, ...)               # ← second load via resolve

```
parse_slash_skill_reference is called twice, and resolve_slash_skill internally calls it a third time. The function also loads all skills and manually filters them before calling resolve_slash_skill, which does the same work. This is redundant.

Suggestion: Simplify to a single call path — either use the manual checks OR call resolve_slash_skill, not both.

---
2: _handle_command in manager.py has duplicate "Unknown command" response blocks

backend/app/channels/manager.py — _handle_command

There are now three separate blocks that construct an "Unknown command" OutboundMessage with nearly identical code:

1. Empty parts → "Unknown command. Available commands: ..."
2. not raw_text.startswith("/") → "Unknown command: /{command}. Available commands: ..."
3. Final else after slash skill resolution → "Unknown command: /{command}. Available commands: ..."

This should be extracted into a helper to reduce duplication and the risk of divergent error formatting.

---
3: _has_existing_activation_for_target has a subtle dual-path dedup strategy

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py

```
def _has_existing_activation_for_target(messages, target_index, target):
    if target_index <= 0:
        return False

    if target.id:
        for previous in messages[:target_index]:
            if not is_slash_skill_activation_reminder(previous):
                continue
            target_id = previous.additional_kwargs.get(...)
            if target_id == target.id or previous.id == f"{target.id}__slash_activation":
                return True
        return False

    # Fallback: check only the immediately preceding message
    previous = messages[target_index - 1]
    return is_slash_skill_activation_reminder(previous)
```

When target.id is truthy, the function returns False after the loop without checking the immediately preceding message. This means if the target has an id but the activation reminder was placed by the fallback path (no target ID match), it won't be detected. Always check the preceding message as a final fallback, regardless of target.id.

ShenAC-SAC:
@hetaoBackend I found one more issue that looks separate from the existing slash-skill comments:

 Preserve slash-skill activation when IM messages include attachments

- Location: `backend/app/channels/manager.py` around `_handle_chat`, where uploaded file context is prepended before calling `runs.wait` / `runs.stream`.
- Problem: For IM channels, a user can send `/data-analysis ...` with an attachment. The message is routed to chat correctly, but `_handle_chat()` prepends `<uploaded_files>...</uploaded_files>` directly to `msg.text` before sending the human message to LangGraph. `SkillActivationMiddleware` later parses the human content and only activates when the text starts with `/skill-name`, so the uploaded-files prefix prevents slash-skill activation.
- Evidence: The current tests cover slash-skill routing without attachments and attachment handling without slash-skill activation, but not the combined path. The same parser accepts `/data-analysis analyze report.pdf` and rejects the same text once it is prefixed by `<uploaded_files>`.
- Suggested fix: Preserve the original IM user text separately when adding uploaded-file context, similar to `UploadsMiddleware` using `original_user_content`. For both `runs.wait` and `runs.stream`, pass the rewritten content plus `additional_kwargs.original_user_content` containing the original `/skill-name ...` text, so `SkillActivationMiddleware` can resolve the intended skill.
- Test: Add a channel-manager regression test for `/data-analysis analyze report.pdf` with `files` present, and verify the original slash command survives into the LangGraph human message metadata or that the activation middleware still resolves the skill.

hetaoBackend:
Thanks all, addressed the review feedback in the latest commits.

@WillemJiang
- Simplified `backend/app/channels/manager.py::_resolve_slash_skill_command` to a single manual validation path: parse once, load skills once, and avoid calling `resolve_slash_skill` after duplicating the same lookup/filtering.
- Extracted shared unknown-command formatting into `_unknown_command_reply()` so `_handle_command` no longer has repeated response construction blocks.
- Updated `_has_existing_activation_for_target()` so targets with ids still fall back to checking the immediately preceding activation reminder. Added a regression test for the legacy/no-target-id reminder case.

@LittleChenLiya
- Added the `event.shiftKey` guard before skill autocomplete consumes Enter/Tab, so `Shift+Enter` now falls through to textarea newline handling.
- Added Playwright coverage for the suggestion menu being visible while `Shift+Enter` inserts a newline instead of selecting the skill.

@ShenAC-SAC
- Preserved the original IM slash command when uploaded file context is prepended. Both `runs.wait` and `runs.stream` now send the rewritten content plus `additional_kwargs.original_user_content`, so `SkillActivationMiddleware` can still resolve `/skill-name ...` after `<uploaded_files>`.
- Added regressions for non-streaming and streaming IM slash-skill messages with attachments.

Local validation:
- `uv run python -m pytest -q` => 4188 passed, 15 skipped
- `pnpm run check`
- `pnpm test` => 205 passed
- `pnpm exec playwright test --project=chromium` => 26 passed
- Manual local service check on http://localhost:3000 for slash autocomplete Enter and Shift+Enter behavior

Latest pushed commits: `c773f23f` and `f1107ada`.

WillemJiang (approved):

PR #3467, opened 2026-06-09
feat(memory): implement fact decay and relevance-based scoring

chetan655:
## Why

- **The trigger**: We needed to implement a next-generation memory architecture featuring temporal decay and category weighting.
- **The pain being addressed**: Previously, facts were stored and evicted based on raw confidence alone. This caused old, outdated facts with high confidence to linger indefinitely in the agent's memory, eventually evicting newer, highly relevant facts. By using a composite relevance score, the agent's memory stays fresh and contextually relevant.

## What changed

- **Config settings**: Added optional `decay_half_life_days` (default `90`) and `category_weights` configurations to `MemoryConfig` with backward-compatible defaults.
- **Composite relevance scoring**: Introduced a composite relevance score formula: $\text{relevance} = \text{confidence} \times 0.5^{(\text{age} / \text{half-life})} \times \text{weight}$.
- **Relevance-based eviction**: Pruning now drops the lowest relevance facts (rather than confidence) when `max_facts` is exceeded.
- **Consistent injection ranking**: `format_memory_for_injection` now sorts injected prompt facts by relevance.

## Surface area

- [x] **Default behavior change** — changes existing behavior without the user opting in (memory facts are now sorted/evicted by relevance score by default)

## Screenshots / Recording

*N/A (Backend logic changes)*

## Validation

Ran the backend tests:
```bash
uv run pytest
```

All backend tests passed.

## AI assistance

**Tool(s) used:** Antigravity (Gemini and opus)

**How you used it:** AI checked that the new changes didn't break existing functionality and wrote the test cases.


xg-gh-25:
The composite relevance scoring formula is sound — temporal decay via exponential half-life plus category weighting addresses the "high-confidence stale facts evicting fresh relevant ones" problem cleanly. The backward-compatible defaults (90-day half-life, 1.0 baseline weights) are defensible for gradual rollout.

A few observations from our production memory systems:

**On the decay formula:** The exponential decay `0.5^(age_days / half_life)` is standard (we use variants of this), but one edge case to consider: **facts with no `createdAt` timestamp**. Your code defaults `age_days=0.0` (full relevance), which means timestamp-less facts never decay. This is safe for backward compatibility, but it creates an invisible class of "immortal facts" that can't be aged out. Consider either (a) logging a warning when this happens, or (b) injecting a synthetic `createdAt` at import time for legacy facts. Otherwise you'll have a bifurcated memory: new facts decay, old facts don't.

**On category weights:** You mention `correction=1.5, goal=1.2` in the docstring, but the defaults in `MemoryConfig` aren't shown in the diff. Are those hardcoded in config? If so, how do users discover them? The "default 1.0" language suggests most categories get baseline weight, but then which categories are upweighted by default? Documentation gap here — users will want to know *what* is weighted higher and *why*.

**On relevance-based eviction:** The old code evicted by raw confidence; the new code evicts by composite relevance. This is correct, but it changes *which facts survive* in long-running agents. If a user has tuned their `max_facts` to keep roughly N days of history under the old regime, the new regime might keep a different temporal window (because decay bends the curve). Consider flagging this as a "default behavior change" in the PR description (you marked it, good) and documenting a migration path — e.g., "if you had max_facts=500 tuned to ~60 days under confidence-only eviction, you may need to increase it under relevance-based eviction to maintain the same coverage."

**On injection ranking:** You're now sorting injected facts by relevance in `format_memory_for_injection`. This is the right call (highest relevance first = most likely to stay in context before truncation), but it introduces a subtle ordering dependency: if the model learns to treat "facts near the top" as more trusted, this creates a feedback loop where high-relevance facts get reinforced. Not necessarily bad (that's how human memory works), but worth documenting as an architectural property.

**On the test coverage:** The PR description says "AI checked that the new changes didn't break existing functionality and wrote the test cases." But I don't see test files in the diff excerpt. Where are the tests? Specifically:
- Does a test verify that facts with `createdAt=None` don't decay?
- Does a test verify the eviction order changes (old: by confidence, new: by relevance)?
- Does a test verify category weights are applied correctly (e.g., `goal` category with weight=1.2 should rank higher than `context` with weight=1.0 at equal age/confidence)?

Without these, the risk is subtle regression when someone changes the decay formula later.

Minor: The docstring has a typo: "per-category multiplier from config (default 1.0; correction=1.5, goal=1.2)" — should clarify "example weights" to avoid confusion about what's actually in the config.

---
*Memory relevance architecture recognized via [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-MEM](https://github.com/xg-gh-25/SwarmAI/discussions/13)*

chetan655:
@xg-gh-25   Thank you for the feedback.

Based on your suggestions, warning logs are now emitted when facts are missing timestamps, and I've added a new test case (`test_relevance_score_missing_date_fallback`) to verify the fallback behavior. The injection feedback loop has also been documented in the codebase.

To clarify one point, the default category weights are defined directly in the `MemoryConfig` schema.

The initial commits already included a dedicated unit test suite (`TestMemoryRelevanceScoring` in `test_memory_updater.py`) covering relevance scoring, eviction ordering, and category weight scaling. The latest commits build on that by adding coverage for the missing-date fallback logic as well.

Addressed all of the nits flagged by Copilot: decoupling the local import helper logic, fixing docstring typos, extracting the pruning sort for better readability, reformatting the config schema, and ensuring deterministic datetime usage in both the injection ranking and unit tests.

PR #3471, opened 2026-06-10
fix: align auth-disabled mode and mock history loading

hetaoBackend:
## Summary
- make the Gateway honor `DEER_FLOW_AUTH_DISABLED=1` with the same synthetic admin user the frontend uses
- prevent auth-disabled mode from overriding valid session/internal identities, and ignore it in explicit production environments
- let auth-disabled requests pass the full write path by covering both Gateway CSRF middleware and LangGraph auth CSRF checks
- disable real run-history loading for `?mock=true` chats so demo/mock pages do not call backend run-message APIs
- fix the new-chat `Load more` regression and guard in-flight history loads from writing stale real-backend messages into mock chats
- address follow-up review nits around E2E scoping, mock=true request coverage, and auth-disabled startup visibility

## Problem
E2E testing found two related user-facing failures. First, the frontend could render the workspace as an auth-disabled `e2e-user`, while the Gateway still required a cookie for protected APIs such as `/api/models`, `/api/skills`, and `/api/langgraph/threads/search`. That left fresh browsers stuck in a split-auth state: the workspace shell rendered, but its API calls returned 401/403, and `/login` or `/setup` immediately redirected back to the broken workspace.

Second, landing-page case studies opened with `?mock=true` loaded the static/mock LangGraph history, but `useThreadHistory` still used the default real client path and then fetched `/api/threads/:thread_id/runs/:run_id/messages` from the real backend. In auth-required environments that request returned 401 and could collapse the demo page back to a new workspace chat.

Review also caught several follow-up issues in the first patch: auth-disabled writes were still blocked by CSRF, new non-mock chats displayed a perpetual `Load more`, valid cookies/internal auth could be clobbered by the synthetic e2e user, the flag had no explicit production guard, and an in-flight history fetch could write stale real-backend messages after switching to mock mode.

## Solution
- Added shared backend auth-disabled helpers/constants for the deterministic `e2e-user` admin identity and auth source labels.
- Hardened `is_auth_disabled()` so `DEER_FLOW_AUTH_DISABLED=1` is ignored when `DEER_FLOW_ENV` or `ENVIRONMENT` is `prod`/`production`.
- Added a startup WARNING when auth-disabled mode is actually active, so accidental self-hosted deployments surface the bypass loudly in logs.
- Updated `AuthMiddleware` to prefer internal auth, then a valid session cookie, then auth-disabled fallback, stamping `request.state.user`, `request.state.auth_source`, `request.state.auth`, and the runtime user context consistently.
- Updated `get_current_user_from_request` to reuse trusted middleware state for `session`, `internal`, and `auth_disabled` instead of creating a half-authenticated e2e fallback in the dependency layer.
- Skipped CSRF checks for auth-disabled requests in both `CSRFMiddleware` and `langgraph_auth.authenticate`, with regression tests using the production-style middleware stack.
- Returned a clearer `change-password` error for the synthetic auth-disabled user.
- Added a frontend `AUTH_DISABLED_USER` constant next to the auth server code so the synthetic-user contract is named and tested.
- Enabled auth-disabled mode on the real-backend replay gateway test server and added a no-cookie `/api/v1/auth/me` E2E contract check against `AUTH_DISABLED_USER`.
- Added an `enabled` option to `useThreadHistory`/`useThreadRuns` and pass `enabled: !isMock` from `useThreadStream`, preventing mock chats from issuing real run-list/run-message requests.
- Strengthened the `mock=true` E2E to fail on both real runs-list and run-messages requests, and removed the fixed sleep by waiting for a deterministic demo render signal.
- Added a history-load generation guard so stale in-flight fetches cannot write messages after mock mode/thread state resets.
- Made `hasMore` false when there is no thread id, removing the new-chat `Load more` regression.
- Scoped real-backend render assertions to `#chat` and stopped relying on `getByTestId("chat")` mirroring behavior from a third-party component.

## Verification
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pytest tests/test_auth_middleware.py::test_auth_disabled_startup_warning_when_effective tests/test_auth_middleware.py::test_auth_disabled_startup_warning_suppressed_in_explicit_production_env tests/test_auth_middleware.py::test_auth_disabled_is_ignored_in_explicit_production_env -q` → 3 passed
- `pnpm format`
- `pnpm lint`
- `pnpm typecheck`
- `CI=1 pnpm test:e2e --reporter=list --retries=0` → 27 passed
- `CI=1 pnpm exec playwright test --config=playwright.real-backend.config.ts --reporter=list --retries=0` → 3 passed

### DIFF
--- a/backend/app/gateway/auth_disabled.py
+++ b/backend/app/gateway/auth_disabled.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import logging
 import os
 from types import SimpleNamespace
 
@@ -16,6 +17,8 @@ AUTH_SOURCE_AUTH_DISABLED = "auth_disabled"
 _PRODUCTION_ENV_VARS: tuple[str, ...] = ("DEER_FLOW_ENV", "ENVIRONMENT")
 _PRODUCTION_ENV_VALUES: frozenset[str] = frozenset({"prod", "production"})
 
+logger = logging.getLogger(__name__)
+
 
 def is_explicit_production_environment() -> bool:
     return any(os.environ.get(name, "").strip().lower() in _PRODUCTION_ENV_VALUES for name in _PRODUCTION_ENV_VARS)
@@ -27,6 +30,17 @@ def is_auth_disabled_requested() -> bool:
 
 def is_auth_disabled() -> bool:
     return is_auth_disabled_requested() and not is_explicit_production_environment()
+
+
+def warn_if_auth_disabled_enabled() -> None:
+    if not is_auth_disabled():
+        return
+
+    logger.warning(
+        "%s=1 is active: authentication is bypassed and anonymous requests run as synthetic admin user %r. Do not enable this in shared or production deployments.",
+        AUTH_DISABLED_ENV_VAR,
+        AUTH_DISABLED_USER_ID,
+    )
 
 
 def get_auth_disabled_user():
--- a/backend/app/gateway/auth_disabled.py
+++ b/backend/app/gateway/auth_disabled.py
@@ -9,9 +9,24 @@ AUTH_DISABLED_ENV_VAR = "DEER_FLOW_AUTH_DISABLED"
 AUTH_DISABLED_USER_ID = "e2e-user"
 AUTH_DISABLED_USER_EMAIL = "e2e@test.local"
 
+AUTH_SOURCE_SESSION = "session"
+AUTH_SOURCE_INTERNAL = "internal"
+AUTH_SOURCE_AUTH_DISABLED = "auth_disabled"
+
+_PRODUCTION_ENV_VARS: tuple[str, ...] = ("DEER_FLOW_ENV", "ENVIRONMENT")
+_PRODUCTION_ENV_VALUES: frozenset[str] = frozenset({"prod", "production"})
+
+
+def is_explicit_production_environment() -> bool:
+    return any(os.environ.get(name, "").strip().lower() in _PRODUCTION_ENV_VALUES for name in _PRODUCTION_ENV_VARS)
+
+
+def is_auth_disabled_requested() -> bool:
+    return os.environ.get(AUTH_DISABLED_ENV_VAR) == "1"
+
 
 def is_auth_disabled() -> bool:
-    return os.environ.get(AUTH_DISABLED_ENV_VAR) == "1"
+    return is_auth_disabled_requested() and not is_explicit_production_environment()
 
 
 def get_auth_disabled_user():
--- a/backend/app/gateway/auth_disabled.py
+++ b/backend/app/gateway/auth_disabled.py
@@ -1,0 +1,25 @@
+"""Shared helpers for local/E2E auth-disabled mode."""
+
+from __future__ import annotations
+
+import os
+from types import SimpleNamespace
+
+AUTH_DISABLED_ENV_VAR = "DEER_FLOW_AUTH_DISABLED"
+AUTH_DISABLED_USER_ID = "e2e-user"
+AUTH_DISABLED_USER_EMAIL = "e2e@test.local"
+
+
+def is_auth_disabled() -> bool:
+    return os.environ.get(AUTH_DISABLED_ENV_VAR) == "1"
+
+
+def get_auth_disabled_user():
+    return SimpleNamespace(
+        id=AUTH_DISABLED_USER_ID,
+        email=AUTH_DISABLED_USER_EMAIL,
+        password_hash=None,
+        system_role="admin",
+        needs_setup=False,
+        token_version=0,
+    )
--- a/backend/app/gateway/auth_middleware.py
+++ b/backend/app/gateway/auth_middleware.py
@@ -17,6 +17,7 @@ from starlette.responses import JSONResponse
 from starlette.types import ASGIApp
 
 from app.gateway.auth.errors import AuthErrorCode, AuthErrorResponse
+from app.gateway.auth_disabled import get_auth_disabled_user, is_auth_disabled
 from app.gateway.authz import _ALL_PERMISSIONS, AuthContext
 from app.gateway.internal_auth import INTERNAL_AUTH_HEADER_NAME, get_internal_user, is_valid_internal_auth_token
 from deerflow.runtime.user_context import reset_current_user, set_current_user
@@ -80,8 +81,16 @@ class AuthMiddleware(BaseHTTPMiddleware):
         if is_valid_internal_auth_token(request.headers.get(INTERNAL_AUTH_HEADER_NAME)):
             internal_user = get_internal_user()
 
+        auth_source = "session"
+
         # Non-public path: require session cookie
-        if internal_user is None and not request.cookies.get("access_token"):
+        if internal_user is not None:
+            user = internal_user
+            auth_source = "internal"
+        elif is_auth_disabled():
+            user = get_auth_disabled_user()
+            auth_source = "auth_disabled"
+        elif not request.cookies.get("access_token"):
             return JSONResponse(
                 status_code=401,
                 content={
@@ -91,23 +100,20 @@ class AuthMiddleware(BaseHTTPMiddleware):
                     ).model_dump()
                 },
             )
-
-        # Strict JWT validation: reject junk/expired tokens with 401
-        # right here instead of silently passing through. This closes
-        # the "junk cookie bypass" gap (AUTH_TEST_PLAN test 7.5.8):
-        # without this, non-isolation routes like /api/models would
-        # accept any cookie-shaped string as authentication.
-        #
-        # We call the *strict* resolver so that fine-grained error
-        # codes (token_expired, token_invalid, user_not_found, …)
-        # propagate from AuthErrorCode, not get flattened into one
-        # generic code. BaseHTTPMiddleware doesn't let HTTPException
-        # bubble up, so we catch and render it as JSONResponse here.
-        from app.gateway.deps import get_current_user_from_request
-
-        if internal_user is not None:
-            user = internal_user
         else:
+            # Strict JWT validation: reject junk/expired tokens with 401
+            # right here instead of silently passing through. This closes
+            # the "junk cookie bypass" gap (AUTH_TEST_PLAN test 7.5.8):
+            # without this, non-isolation routes like /api/models would
+            # accept any cookie-shaped string as authentication.
+            #
+            # We call the *strict* resolver so that fine-grained error
+            # codes (token_expired, token_invalid, user_not_found, …)
+            # propagate from AuthErrorCode, not get flattened into one
+            # generic code. BaseHTTPMiddleware doesn't let HTTPException
+            # bubble up, so we catch and render it as JSONResponse here.
+            from app.gateway.deps import get_current_user_from_request
+
             try:
                 user = await get_current_user_from_request(request)
             except HTTPException as exc:
@@ -118,6 +124,7 @@ class AuthMiddleware(BaseHTTPMiddleware):
         # None" branch short-circuits instead of running the entire
         # JWT-decode + DB-lookup pipeline a second time per request).
         request.state.user = user
+        request.state.auth_source = auth_source
         request.state.auth = AuthContext(user=user, permissions=_ALL_PERMISSIONS)
         token = set_current_user(user)
         try:
--- a/backend/app/gateway/auth_middleware.py
+++ b/backend/app/gateway/auth_middleware.py
@@ -17,7 +17,13 @@ from starlette.responses import JSONResponse
 from starlette.types import ASGIApp
 
 from app.gateway.auth.errors import AuthErrorCode, AuthErrorResponse
-from app.gateway.auth_disabled import get_auth_disabled_user, is_auth_disabled
+from app.gateway.auth_disabled import (
+    AUTH_SOURCE_AUTH_DISABLED,
+    AUTH_SOURCE_INTERNAL,
+    AUTH_SOURCE_SESSION,
+    get_auth_disabled_user,
+    is_auth_disabled,
+)
 from app.gateway.authz import _ALL_PERMISSIONS, AuthContext
 from app.gateway.internal_auth import INTERNAL_AUTH_HEADER_NAME, get_internal_user, is_valid_internal_auth_token
 from deerflow.runtime.user_context import reset_current_user, set_current_user
@@ -81,26 +87,14 @@ class AuthMiddleware(BaseHTTPMiddleware):
         if is_valid_internal_auth_token(request.headers.get(INTERNAL_AUTH_HEADER_NAME)):
             internal_user = get_internal_user()
 
-        auth_source = "session"
+        auth_source = AUTH_SOURCE_SESSION
+        access_token = request.cookies.get("access_token")
 
         # Non-public path: require session cookie
         if internal_user is not None:
             user = internal_user
-            auth_source = "internal"
-        elif is_auth_disabled():
-            user = get_auth_disabled_user()
-            auth_source = "auth_disabled"
-        elif not request.cookies.get("access_token"):
-            return JSONResponse(
-                status_code=401,
-                content={
-                    "detail": AuthErrorResponse(
-                        code=AuthErrorCode.NOT_AUTHENTICATED,
-                        message="Authentication required",
-                    ).model_dump()
-                },
-            )
-        else:
+            auth_source = AUTH_SOURCE_INTERNAL
+        elif access_token:
             # Strict JWT validation: reject junk/expired tokens with 401
             # right here instead of silently passing through. This closes
             # the "junk cookie bypass" gap (AUTH_TEST_PLAN test 7.5.8):
@@ -117,7 +111,23 @@ class AuthMiddleware(BaseHTTPMiddleware):
             try:
                 user = await get_current_user_from_request(request)
             except HTTPException as exc:
-                return JSONResponse(status_code=exc.status_code, content={"detail": exc.detail})
+                if not is_auth_disabled():
+                    return JSONResponse(status_code=exc.status_code, content={"detail": exc.detail})
+                user = get_auth_disabled_user()
+                auth_source = AUTH_SOURCE_AUTH_DISABLED
+        elif is_auth_disabled():
+            user = get_auth_disabled_user()
+            auth_source = AUTH_SOURCE_AUTH_DISABLED
+        else:
+            return JSONResponse(
+                status_code=401,
+                content={
+                    "detail": AuthErrorResponse(
+                        code=AuthErrorCode.NOT_AUTHENTICATED,
+                        message="Authentication required",
+                    ).model_dump()
+                },
+            )
 
         # Stamp both request.state.user (for the contextvar pattern)
         # and request.state.auth (so @require_permission's "auth is
--- a/backend/app/gateway/deps.py
+++ b/backend/app/gateway/deps.py
@@ -331,8 +331,24 @@ async def get_current_user_from_request(request: Request):
 
     Raises HTTPException 401 if not authenticated.
     """
+    state = getattr(request, "state", None)
+    state_user = getattr(state, "user", None)
+    if state_user is not None and getattr(state, "auth_source", None) in {
+        "session",
+        "auth_disabled",
+    }:
+        return state_user
+
     from app.gateway.auth import decode_token
     from app.gateway.auth.errors import AuthErrorCode, AuthErrorResponse, TokenError, token_error_to_code
+    from app.gateway.auth_disabled import get_auth_disabled_user, is_auth_disabled
+
+    if is_auth_disabled():
+        user = get_auth_disabled_user()
+        if state is not None:
+            state.user = user
+            state.auth_source = "auth_disabled"
+        return user
 
     access_token = request.cookies.get("access_token")
     if not access_token:
--- a/backend/app/gateway/deps.py
+++ b/backend/app/gateway/deps.py
@@ -333,22 +333,17 @@ async def get_current_user_from_request(request: Request):
     """
     state = getattr(request, "state", None)
     state_user = getattr(state, "user", None)
+    from app.gateway.auth_disabled import AUTH_SOURCE_AUTH_DISABLED, AUTH_SOURCE_INTERNAL, AUTH_SOURCE_SESSION
+
     if state_user is not None and getattr(state, "auth_source", None) in {
-        "session",
-        "auth_disabled",
+        AUTH_SOURCE_SESSION,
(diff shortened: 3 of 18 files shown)


backend/app/gateway/auth_middleware.py
```diff
@@ -80,8 +81,16 @@ async def dispatch(self, request: Request, call_next: Callable) -> Response:
...
             internal_user = get_internal_user()
 
+        auth_source = "session"
+
         # Non-public path: require session cookie
-        if internal_user is None and not request.cookies.get("access_token"):
+        if internal_user is not None:
+            user = internal_user
+            auth_source = "internal"
+        elif is_auth_disabled():
```
hetaoBackend: **[High] Auth-disabled mode bypasses AuthMiddleware, but `CSRFMiddleware` still rejects all state-changing requests.**

`CSRFMiddleware` is mounted globally (`app.py:336`) and has no knowledge of `is_auth_disabled()`. The `csrf_token` cookie is only minted on POSTs to the four auth endpoints (`csrf_middleware.py:207-218`), which the auth-disabled flow never calls — `frontend/src/core/auth/server.ts:23` short-circuits with the synthetic user without logging in.

Reproduced empirically with the real middleware stack in the same order as `app.py`, with `DEER_FLOW_AUTH_DISABLED=1`:
- `GET /api/threads` → 200 ✅
- `POST /api/threads` → **403 `CSRF token missing. Include X-CSRF-Token header.`** ❌

So the PR's stated goal (fresh browser works without login) still fails for every write: sending a chat message, creating a run, uploading files.

The PR's tests don't catch this because: the backend test app in `test_auth_middleware.py::_make_app` mounts only `AuthMiddleware`; the e2e specs use `mockLangGraphAPI`/`page.route()` interception; and the real-backend suite explicitly registers a throwaway account, which mints the csrf cookie (`real-backend-render.spec.ts:59-63`).

Suggested fix: have `CSRFMiddleware.should_check_csrf` skip when `is_auth_disabled()` (or mint the csrf cookie on the auth-disabled path), and mount the same middleware stack as `app.py` in the backend test app so middleware-composition gaps like this are testable.

hetaoBackend (commented):
Reviewed this PR with a multi-agent pass (7 finder angles, every candidate independently verified against the actual code; 4 false positives were discarded during verification). Overall the direction is right and the test coverage added is solid, but there are **2 high-severity issues** worth addressing before merge, plus a few follow-ups.

**Must fix:**
1. **CSRF still blocks all writes in auth-disabled mode** — the PR's own target scenario (fresh browser, no login) gets 403 on every POST/PUT/DELETE. See inline comment on `auth_middleware.py`.
2. **`hasMore` regression on new chats** — every brand-new (non-mock) chat now renders a perpetual "Load more" indicator. See inline comment on `hooks.ts`.

**Strongly recommended:** add a production safeguard for `DEER_FLOW_AUTH_DISABLED` (inline comment on `auth_disabled.py`).

The remaining 5 inline comments (in-flight fetch race, internal-auth identity clobber, misleading change-password error, triple auth-disabled resolution paths, frontend/backend identity contract drift) are non-blocking follow-ups.

frontend/src/core/threads/hooks.ts
```diff
@@ -992,14 +1006,14 @@ export function useThreadHistory(threadId: string) {
...
-  }, [threadId, runs.data, loadMessages]);
+  }, [enabled, threadId, runs.data, loadMessages]);
 
   const appendMessages = useCallback((_messages: Message[]) => {
     setMessages((prev) => {
       return dedupeMessagesByIdentity([...prev, ..._messages]);
     });
   }, []);
-  const hasMore = indexRef.current >= 0 || !runs.data;
+  const hasMore = enabled && (indexRef.current >= 0 || !runs.data);
```
hetaoBackend: **[High] Regression for all non-mock users: `hasMore` is now permanently `true` for new chats.**

`useThreadRuns` now has `enabled: enabled && Boolean(threadId)` (line 1108). Previously the query ran even with `threadId === ""` and the queryFn early-returned `[]`, so `runs.data = []` and `hasMore = (-1 >= 0 || ![]) = false`. Now the query is disabled for empty threadId, `runs.data` stays `undefined` forever, and `hasMore = enabled && (false || !undefined) = true`.

This hits every brand-new chat: `page.tsx:77` passes `threadId: undefined` for new threads, so `useThreadHistory("", { enabled: true })` is called. `LoadMoreHistoryIndicator` is rendered unconditionally at the top of the conversation and its null-guard is `if (!hasMore && !isLoading) return null;` (`message-list.tsx:129`) — so a perpetual "Load more" button appears on every empty chat, and its IntersectionObserver keeps firing a no-op `loadMore` until a real thread id arrives.

Suggested one-line fix:
```ts
const hasMore = enabled && Boolean(threadId) && (indexRef.current >= 0 || !runs.data);
```

backend/app/gateway/auth_disabled.py
```diff
@@ -0,0 +1,25 @@
...
+import os
+from types import SimpleNamespace
+
+AUTH_DISABLED_ENV_VAR = "DEER_FLOW_AUTH_DISABLED"
+AUTH_DISABLED_USER_ID = "e2e-user"
+AUTH_DISABLED_USER_EMAIL = "e2e@test.local"
+
+
+def is_auth_disabled() -> bool:
+    return os.environ.get(AUTH_DISABLED_ENV_VAR) == "1"
```
hetaoBackend, line 28: **[Security hardening] No production safeguard on this flag, and it takes precedence over valid session cookies.**

Two consequences:

1. If `DEER_FLOW_AUTH_DISABLED=1` ever leaks into a real deployment (copied `.env`, compose file, CI image), every anonymous remote request becomes `system_role="admin"` with `_ALL_PERMISSIONS` — including `PUT /api/mcp/config` (`routers/mcp.py:89`), which registers stdio MCP commands, i.e. an RCE surface.
2. Because the middleware checks `is_auth_disabled()` **before** cookie validation, a real logged-in user with a valid `access_token` is silently remapped to `e2e-user` — their per-user data under `users/{real-id}/` vanishes from view and new writes are attributed to `e2e-user`. A footgun for local devs with existing data.

Note the sibling dangerous flag in this codebase is hardened: `internal_auth.py:31-33` guards its bypass with `secrets.compare_digest` against a per-process secret. Suggest following that convention here — gate this flag on a non-production check (or at minimum log a loud startup warning), and consider honoring a valid cookie when present.

frontend/src/core/threads/hooks.ts
```diff
@@ -965,12 +975,12 @@ export function useThreadHistory(threadId: string) {
       loadingRunIdRef.current = null;
       setLoading(false);
     }
-  }, []);
+  }, [enabled]);
```
hetaoBackend, line 985: **[Medium] In-flight `loadMessages` fetch is not invalidated when `enabled` flips to `false`.**

The post-`await` guard (line ~937) only checks `threadIdRef.current !== requestThreadId`. The new `if (!enabled) return;` at the top doesn't help an in-flight invocation, which captured the old closure where `enabled === true`.

Scenario: user on `/workspace/chats/X` with a run-messages fetch in flight soft-navigates to `/workspace/chats/X?mock=true` (same threadId; `isMock` comes from `useSearchParams()` in `use-thread-chat.ts:38`, and query-only changes don't remount — `page.tsx:89` explicitly relies on no-remount semantics via `history.replaceState`). The effect resets state with `setMessages([])`, then the resumed fetch passes the threadId-only guard and writes **real-backend history into the mock chat**.

Narrow window, but the defect is unambiguous — there's no AbortController or generation counter. A simple fix is a monotonically increasing load-generation ref checked after the `await`, bumped in the reset effect.

backend/app/gateway/deps.py
```diff
@@ -331,8 +331,24 @@ async def get_current_user_from_request(request: Request):
...
+        "session",
+        "auth_disabled",
+    }:
+        return state_user
+
     from app.gateway.auth import decode_token
     from app.gateway.auth.errors import AuthErrorCode, AuthErrorResponse, TokenError, token_error_to_code
+    from app.gateway.auth_disabled import get_auth_disabled_user, is_auth_disabled
+
+    if is_auth_disabled():
```
hetaoBackend: **[Medium, E2E-mode only] Internal-auth requests get their identity clobbered when auth-disabled is on.**

With `DEER_FLOW_AUTH_DISABLED=1` and a valid `X-DeerFlow-Internal-Token`, middleware stamps `state.user = internal_user` / `auth_source = "internal"`. A route dep then skips the fast path above (`"internal"` is not in the trusted set) and lands here, returning `e2e-user` **and overwriting** `state.user`/`state.auth_source` — while the `set_current_user` contextvar still holds the internal user.

Result: handler-level identity (`e2e-user`) diverges from repository-layer owner filtering (internal user), e.g. channel-worker runs through `thread_runs.py:207` get scoped to `"e2e-user"`. Only reachable in E2E mode, but the identity overwrite is newly introduced here. Consider either including `"internal"` in the trusted set (with a comment) or making this fallback not overwrite middleware-stamped state.

backend/app/gateway/auth_disabled.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def get_auth_disabled_user():
+    return SimpleNamespace(
```
hetaoBackend, line 47: **[Low] `password_hash=None` makes the Account Settings password form return a misleading error in auth-disabled mode.**

`POST /api/v1/auth/change-password` (`routers/auth.py:347-348`) deterministically returns 400 `"OAuth users cannot change password"` for this synthetic user — and that endpoint is wired into the real UI (`account-settings-page.tsx:41`), so a dev clicking around Settings under `DEER_FLOW_AUTH_DISABLED=1` will hit it even though no OAuth is involved.

(Checked the other consumers of the current-user object — `/me`, `mcp.py`, `authz.py` — they only use the six provided attributes, so no AttributeError risk beyond this.) Worth either a clearer error for the auth-disabled case or a doc note on which flows are unsupported in this mode.

backend/app/gateway/deps.py
```diff
@@ -331,8 +331,24 @@ async def get_current_user_from_request(request: Request):
 
     Raises HTTPException 401 if not authenticated.
     """
+    state = getattr(request, "state", None)
+    state_user = getattr(state, "user", None)
+    if state_user is not None and getattr(state, "auth_source", None) in {
```
hetaoBackend, line 338: **[Cleanup] Auth-disabled resolution now has three touchpoints that must stay in sync, keyed by untyped magic strings.**

(1) the middleware's `elif is_auth_disabled()` branch, (2) this state-reuse fast path with a hand-maintained `{"session", "auth_disabled"}` allowlist (silently excluding `"internal"`, uncommented), and (3) the `is_auth_disabled()` fallback below, which is unreachable for any request that went through the middleware and — when it does fire in middleware-less contexts — stamps only `state.user`/`state.auth_source`, **not** `state.auth` (`AuthContext`) and **not** the `set_current_user` contextvar, producing a half-authenticated request state.

The middleware already delegates the JWT path to this function, so extending that single-resolver direction to auth-disabled (and lifting the `auth_source` values to module-level constants) would collapse this to one path. A future fourth auth source that forgets to join the set literal here gets silently 401'd despite being middleware-authenticated.

backend/app/gateway/auth_disabled.py
```diff
@@ -0,0 +1,25 @@
+"""Shared helpers for local/E2E auth-disabled mode."""
+
+from __future__ import annotations
+
+import os
+from types import SimpleNamespace
+
+AUTH_DISABLED_ENV_VAR = "DEER_FLOW_AUTH_DISABLED"
+AUTH_DISABLED_USER_ID = "e2e-user"
```
hetaoBackend, line 10: **[Cleanup] The synthetic-user contract is duplicated between backend and frontend with nothing pinning them together.**

The same literals (`"e2e-user"`, `"e2e@test.local"`, `system_role: "admin"`, `needs_setup: false`) are hardcoded inline in `frontend/src/core/auth/server.ts:23-33`, with no comment on either side pointing at the other. No test catches drift: the new backend tests assert only the backend values, `frontend/tests/unit/core/auth/server.test.ts` never exercises the `DEER_FLOW_AUTH_DISABLED` branch, and the real-backend e2e suite (which runs with the flag set) only asserts rendered chat content.

If either side changes the id, the SSR-displayed user and the gateway identity silently diverge and user-scoped data lands under a different `users/{id}/` than displayed — the same split-auth bug class this PR fixes. Suggest a named constant on the frontend next to the existing `STATIC_WEBSITE_USER` pattern, plus one e2e assertion that `/api/v1/auth/me` returns the SSR user's id.

backend/app/gateway/auth_middleware.py
hetaoBackend: Fixed in `fea7994a`: `CSRFMiddleware.should_check_csrf` now skips when `is_auth_disabled()`, and LangGraph auth has the same CSRF skip. Added a production-style AuthMiddleware+CSRFMiddleware regression test plus LangGraph auth coverage; full backend and real-backend E2E are green.

frontend/src/core/threads/hooks.ts
hetaoBackend: Fixed in `fea7994a`: `hasMore` now requires `enabled && Boolean(threadId)` before treating missing run data as loadable history. Added an E2E assertion that a brand-new chat hides the Load more button.

backend/app/gateway/auth_disabled.py
hetaoBackend: Fixed in `fea7994a`: `is_auth_disabled()` is ignored when `DEER_FLOW_ENV` or `ENVIRONMENT` is `prod`/`production`, and `AuthMiddleware` now honors internal auth and valid session cookies before falling back to the e2e user. Added regression tests for valid session preservation and explicit production env.

frontend/src/core/threads/hooks.ts
hetaoBackend: Fixed in `fea7994a`: `useThreadHistory` now bumps a load generation on reset and checks it after the awaited run-message fetch before writing messages. The `finally` path is generation-guarded too, so stale requests cannot clear a newer load state.

backend/app/gateway/deps.py
hetaoBackend: Fixed in `fea7994a`: auth source labels are constants and `get_current_user_from_request` now trusts `session`, `internal`, and `auth_disabled` middleware state. The old dependency-layer auth-disabled fallback was removed, and a regression test covers internal identity preservation.

backend/app/gateway/auth_disabled.py
hetaoBackend: Fixed in `fea7994a`: change-password now checks `request.state.auth_source == AUTH_SOURCE_AUTH_DISABLED` first and returns a mode-specific error instead of the misleading OAuth password message.

backend/app/gateway/deps.py
hetaoBackend: Fixed in `fea7994a`: lifted auth source strings to constants, expanded the trusted middleware-state fast path to include internal auth, and removed the dependency-layer auth-disabled fallback that could create half-authenticated request state.

backend/app/gateway/auth_disabled.py
hetaoBackend: Fixed in `fea7994a` and `d4ef1ffd`: frontend now has a named `AUTH_DISABLED_USER` constant with unit coverage, and the real-backend Playwright suite now starts the replay gateway with `DEER_FLOW_AUTH_DISABLED=1` and asserts no-cookie `/api/v1/auth/me` returns that same contract.

frontend/tests/e2e-real-backend/real-backend-render.spec.ts
```diff
@@ -101,7 +101,8 @@ test.describe("real backend render (replay, no API key)", () => {
       EXPECTED_SUGGESTION,
       "fixture should contain a suggestions turn (re-record; the record spec waits for /suggestions)",
     ).not.toBe("");
-    await expect(page.getByText(EXPECTED_TITLE)).toBeVisible({
+    const chat = page.getByTestId("chat");
```
fancyboi999: `getByTestId("chat")` 能 work 其实是靠 react-resizable-panels 把 `id="chat"` 镜像成了 `data-testid`——`frontend/src` 里并没有真正的 `data-testid`。同一个 panel 别处(`artifact-preview.spec`)是用 `locator("#chat")` 的,换过来更稳,也不依赖第三方库的实现细节:

```suggestion
    const chat = page.locator("#chat");
```

另外下面第 108 行的 `EXPECTED_SUGGESTION` 断言还是全页范围的,跟这次修的 title 是同一类 strict-mode 碰撞隐患,建议也一并 scope 到 `chat`。

frontend/tests/e2e/thread-history.spec.ts
```diff
@@ -61,6 +62,50 @@ test.describe("Thread history", () => {
...
 
+  test("mock thread does not load run messages from the real backend", async ({
+    page,
+  }) => {
+    const backendRunMessageUrls: string[] = [];
+    page.on("request", (request) => {
+      const url = request.url();
+      if (
+        request.method() === "GET" &&
+        url.includes(`/api/threads/${DEMO_THREAD_ID}/runs/`)
```
fancyboi999: 这个过滤带了尾斜杠 `/runs/`,只能匹配 run-messages(`/api/threads/<id>/runs/<run>/messages`)。但 `useThreadRuns` 真正打的 runs-list 是 `/api/langgraph/threads/<id>/runs`(前缀不同、无尾斜杠),匹配不到。也就是说单独 revert `useThreadRuns` 的 `enabled` 这一半,这个用例守不住。产品行为这次已经修对了(`hasMore` + `enabled`),所以不影响功能,只是这个断言覆盖不到 runs-list 那条路径——要兜全可以把 runs-list 也纳入过滤,或者直接 `page.route` 命中即 fail。

frontend/tests/e2e/thread-history.spec.ts
```diff
@@ -61,6 +62,50 @@ test.describe("Thread history", () => {
...
+        },
+      ],
+    });
+
+    await page.goto(`/workspace/chats/${DEMO_THREAD_ID}?mock=true`);
+
+    await expect(
+      page.getByText("What might be the trends and opportunities in 2026?"),
+    ).toBeVisible({ timeout: 15_000 });
+    await page.waitForTimeout(500);
```
fancyboi999: 小提一句:这是整个 e2e 套件里唯一的硬 `waitForTimeout`。负向断言(不该发某请求)用固定 500ms 窗口,慢 CI 上理论上能漏(请求晚于 500ms 落地仍判过)。可以换成等一个确定的完成信号(比如等 demo 的某个后续渲染出现)再断言,把这个 sleep 去掉。非阻塞。

backend/app/gateway/auth_disabled.py
```diff
@@ -0,0 +1,40 @@
...
+
+def is_explicit_production_environment() -> bool:
+    return any(os.environ.get(name, "").strip().lower() in _PRODUCTION_ENV_VALUES for name in _PRODUCTION_ENV_VARS)
+
+
+def is_auth_disabled_requested() -> bool:
+    return os.environ.get(AUTH_DISABLED_ENV_VAR) == "1"
+
+
+def is_auth_disabled() -> bool:
```
fancyboi999, line 31: 生产守卫这条加得挺好。一个加固建议:现在是黑名单式,只有显式设了 `DEER_FLOW_ENV`/`ENVIRONMENT=prod/production` 才拦截。不少自托管 docker 部署默认不设这俩变量,那种环境里 `DEER_FLOW_AUTH_DISABLED=1` 仍会放行成全权限 admin。考虑到这个开关的破坏力,建议在它真正生效时打一条显眼的启动 WARNING 日志(类似 internal_auth 留痕的做法),让误开的人能从日志一眼看出来。

fancyboi999 (commented):
先说结论:这版相比最初那次改了非常多。CSRF、新对话的 `hasMore` 回归、internal 身份被覆盖、langgraph compat、改密码报错、`AUTH_SOURCE_*` 常量、前后端 synthetic user 统一,基本都处理了,而且配了挺到位的回归测试——按生产中间件顺序挂 Auth+CSRF 验 POST 不再 403、断言 internal 请求下 contextvar 与 state 身份一致、`chat.spec.ts` 直接断言新对话不出现 "Load more" 按钮、`auth-disabled-contract.spec.ts` 拿前端常量打真实 gateway `/auth/me` 做跨端 parity。这几条核心的我对着测试逐个看过,是真修,不是测试假绿。另外我读了 `langgraph_auth.authenticate` 的改动,auth-disabled 返回 `AUTH_DISABLED_USER_ID` 和正常路径返回 `payload.sub` 都是字符串,`add_owner_filter` 用 `ctx.user.identity` 也能正确拿到,类型一致,没引入新的不一致。

剩下几条都是 nit,不阻塞合并,行内标了。整体我觉得 OK。

frontend/tests/e2e-real-backend/real-backend-render.spec.ts
hetaoBackend: Fixed in `7571bc99`: switched the real-backend render spec to `page.locator("#chat")` and scoped both title and suggestion assertions to `chat`.

frontend/tests/e2e/thread-history.spec.ts
hetaoBackend: Fixed in `7571bc99`: strengthened the `mock=true` E2E so it fails on both `/api/langgraph/threads/:id/runs` and `/api/threads/:id/runs/:run_id/messages` GETs for the demo thread.

frontend/tests/e2e/thread-history.spec.ts
hetaoBackend: Fixed in `7571bc99`: removed the fixed `waitForTimeout(500)` and now wait for a deterministic later demo message before asserting no real backend run-history request was made.

backend/app/gateway/auth_disabled.py
hetaoBackend: Fixed in `7571bc99`: added a startup WARNING when `DEER_FLOW_AUTH_DISABLED=1` actually takes effect, plus tests for warning emission and production-env suppression. The real-backend E2E startup log also shows the warning.

hetaoBackend:
Follow-up nits are addressed in `7571bc99`: scoped the real-backend render assertions to `#chat`, expanded the `mock=true` E2E guard to fail on both runs-list and run-messages requests without a fixed timeout, and added the auth-disabled startup WARNING plus tests. All review threads are resolved and the latest GitHub checks are green. @fancyboi999 could you re-review/approve when you have a moment?

WillemJiang (approved):

Issue #3260, opened 2026-05-26
Gateway mode streaming reconnect is not multi-worker safe due to in-memory RunManager/StreamBridge


AnoobFeng:
1. In Gateway mode, runs are executed in background asyncio.Tasks and tracked by in-memory RunManager.

2. SSE replay/join depends on in-memory StreamBridge.

3. In a multi-pod deployment without sticky sessions, refreshing the page may route joinStream to another pod, which cannot find the run or stream buffer, while the original run may still be executing.

Are there any plans to support distributed environments (e.g., using Redis for shared state)? If so, what is the roadmap?

CrazyCollin:
I’d like to help move cross-worker stream join support forward.

I noticed #3191 is already working on RedisStreamBridge. If that PR needs help, I can help rebase/test it. Alternatively, I can open a smaller preparatory PR around `StreamBridge` capabilities and `store_only` gateway behavior.

Could someone advise the preferred path? I’d be happy to take this on if useful.

Issue #3239, opened 2026-05-26
多 worker 部署下取消对话及取消后再次对话 返回 409 Conflict


sfsflm:
**问题描述**：

在 K8s 环境中使用 `uvicorn --workers 4` 部署后端服务时，出现以下问题：

1. **取消对话时报错**：在对话过程中点击取消按钮，有时会返回 409 Conflict，错误信息为 `Run xxx is not active on this worker and cannot be cancelled`

2. **取消后发送消息报错**：取消当前对话后，再次发送新消息时，有时会返回 409 Conflict，错误信息为 `Thread xxx already has an active run`

上述问题不是必现的，存在一定的随机性，可能与请求被路由到哪个 worker 有关。

**环境信息**：
- 部署方式：K8s Deployment，replicas: 1
- uvicorn workers: 4
- 数据库：Postgres

- 配置文件：
`database:
  backend: postgres
  postgres_url: $DATABASE_URL

run_events:
  backend: db
  max_trace_content: 10240
  track_token_usage: true`
- 错误日志：
- INFO:     10.233.127.36:39972 - "POST /api/threads/d33ea779-edc8-4895-b3a5-042b6822592d/runs/7ef9024c-8aca-47b1-9278-e0708a4416f3/cancel?wait=0&action=interrupt HTTP/1.1" 409 Conflict
- INFO:     10.233.127.36:49192 - "POST /api/threads/d33ea779-edc8-4895-b3a5-042b6822592d/runs/d114e03d-3fa2-4847-a5c7-72d913dd3920/cancel?wait=0&action=interrupt HTTP/1.1" 409 Conflict
 
- 启动日志：

-    Building deerflow-harness @ file:///app/backend/packages/harness
      Built deerflow-harness @ file:///app/backend/packages/harness
Uninstalled 1 package in 0.76ms
Installed 1 package in 1ms
INFO:     Uvicorn running on http://0.0.0.0:8001 (Press CTRL+C to quit)
INFO:     Started parent process [118]
/app/backend/.venv/lib/python3.12/site-packages/langgraph/checkpoint/serde/encrypted.py:5: LangChainPendingDeprecationWarning: The default value of `allowed_objects` will change in a future version. Pass an explicit value (e.g., allowed_objects='messages' or allowed_objects='core') to suppress this warning.
  from langgraph.checkpoint.serde.jsonplus import JsonPlusSerializer
[...]

PR #3191, opened 2026-05-24
feat(gateway): add redis stream bridge

Janlaywss:
## Summary

This PR adds a Redis Streams-backed `StreamBridge` implementation for Docker and multi-worker Gateway deployments.

The existing in-memory bridge works only within one process. In Docker production, the Gateway can run with multiple Uvicorn workers, so SSE consumers may land on a different worker than the producer and fail to receive live run events. The Redis bridge stores per-run events in Redis Streams so workers can publish and subscribe across process boundaries while preserving `Last-Event-ID` replay behavior.

## Changes

- Add `RedisStreamBridge` backed by Redis Streams.
- Wire `make_stream_bridge()` to initialize Redis when `stream_bridge.type: redis` is configured or `DEER_FLOW_STREAM_BRIDGE_REDIS_URL` is set.
- Add Redis service, health check, and persistent volume to Docker dev/prod compose files.
- Enable Docker Gateway containers to use Redis bridge automatically via `DEER_FLOW_STREAM_BRIDGE_REDIS_URL`.
- Allow store-only run streaming only when the active bridge supports cross-process subscriptions.
- Add Redis bridge unit tests and cross-process bridge route coverage.
- Add `make docker-logs-redis`.

## Tests

- `uv run ruff format app/gateway/routers/thread_runs.py app/gateway/services.py packages/harness/deerflow/runtime/stream_bridge packages/harness/deerflow/config/stream_bridge_config.py tests/test_stream_bridge.py tests/test_thread_run_messages_pagination.py`
- `uv run ruff check app/gateway/routers/thread_runs.py app/gateway/services.py packages/harness/deerflow/runtime/stream_bridge packages/harness/deerflow/config/stream_bridge_config.py tests/test_stream_bridge.py tests/test_thread_run_messages_pagination.py`
- `uv run pytest tests/test_stream_bridge.py tests/test_thread_run_messages_pagination.py tests/test_gateway_lifespan_shutdown.py tests/test_gateway_runtime_cleanup.py tests/test_gateway_run_recovery.py`
- `docker compose -f docker/docker-compose-dev.yaml config`
- `docker compose -f docker/docker-compose.yaml config`

PR #3475, opened 2026-06-10
fix(docker): default Gateway to a single worker to prevent multi-worker breakage

fancyboi999:
## Problem

The default production deployment (`make up`) starts the Gateway with `--workers 4`. But the Gateway holds run state — `RunManager._runs` and the stream bridge — as **in-process singletons**, and nginx in front of it uses **no sticky sessions**. So with the default config, requests for the same run scatter across workers that each keep their own copy of run state. The result, out of the box:

- **Run cancellation** returns 409 when the cancel request lands on a worker that does not own the run.
- **SSE reconnect** (`join`/stream) rebuilds an empty stream on a worker that never ran the request, so clients hang on heartbeats instead of receiving a terminal event.
- **`multitask=reject` de-duplication** only checks the local process, so the same thread can run concurrently on two workers and write the same checkpoint.
- **IM channels** start one channel service per worker, causing duplicate consumption / duplicate replies (and Telegram `getUpdates` 409s).

This is the default `docker compose` configuration, so a fresh `make up` is broken for anything beyond a single in-flight conversation. Reported in #3239 and #3260.

## Root cause

`docker/docker-compose.yaml` defaulted `GATEWAY_WORKERS` to `4`, while the runtime keeps all run state in process and the shared cross-worker stream bridge (e.g. the Redis bridge proposed in #3191) is not yet implemented — the redis path currently raises `NotImplementedError`. Multi-worker safety therefore does not exist yet; the default just silently enabled an unsafe topology.

## Solution (m1 stop-gap)

Make the **default** deployment correct by running a single Gateway worker:

- `docker-compose.yaml`: default `GATEWAY_WORKERS` to `1` (was `4`), with a comment explaining why.
- `README.md`: document the single-worker boundary in the production deployment section so operators understand why and what raising the count breaks.
- Regression test pinning the default to `1` while asserting it stays overridable via `GATEWAY_WORKERS`.

**Scope / trade-offs.** This is a stop-gap, not a multi-worker implementation. It does **not** make the Gateway multi-worker safe — it only stops the default config from shipping a broken topology. The worker count remains operator-overridable via `GATEWAY_WORKERS` for when a shared stream bridge exists. The full fix (shared run state + stream bridge + distributed cancel) is tracked separately in #3191. The Gateway is async I/O–bound, so a single Uvicorn worker comfortably handles concurrent SSE streams; scale a single worker up with more CPU/RAM (or move the database and sandbox onto dedicated tiers) instead of raising the worker count.

Hence `Refs` rather than `Closes`: #3239 / #3260 are only fully resolved once cross-worker run state lands.

## Validation

`docker compose config` renders the effective command (no build / up needed):

```
# default (GATEWAY_WORKERS unset)
... uvicorn app.gateway.app:app --host 0.0.0.0 --port 8001 --workers 1

# override still works
GATEWAY_WORKERS=4 ... --workers 4
```

Regression test (`backend/tests/test_compose_default_workers.py`, no Docker dependency — pure YAML parse, runs in CI):

- `test_gateway_defaults_to_single_worker` — default worker count is `1`.
- `test_gateway_worker_count_remains_overridable` — command uses `${GATEWAY_WORKERS:-1}`, so operators can still override it.

Both pass; `make lint` clean.

Refs #3239, #3260

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,9 @@ Access: http://localhost:2026
 
 The unified nginx endpoint is same-origin by default and does not emit browser CORS headers. If you run a split-origin or port-forwarded browser client, set `GATEWAY_CORS_ORIGINS` to comma-separated exact origins such as `http://localhost:3000`; the Gateway then applies the CORS allowlist and matching CSRF origin checks.
 
+> [!IMPORTANT]
+> The Gateway holds run state (RunManager and the stream bridge) in process, so production defaults to a single Gateway worker (`GATEWAY_WORKERS=1`). Raising the worker count without a shared cross-worker stream bridge — which is not yet available — breaks run cancellation, SSE reconnects, request de-duplication, and IM channels, because nginx uses no sticky sessions and each worker keeps its own run state. Scale a single worker up with more CPU/RAM (or move the database and sandbox onto dedicated tiers) instead of raising `GATEWAY_WORKERS`.
+
 See [CONTRIBUTING.md](CONTRIBUTING.md) for detailed Docker development guide.
 
 #### Option 2: Local Development
--- a/backend/tests/test_compose_default_workers.py
+++ b/backend/tests/test_compose_default_workers.py
@@ -1,0 +1,45 @@
+"""Regression test for the Docker Compose default Gateway worker count.
+
+The Gateway holds run state (RunManager and the stream bridge) in process, so
+the default deployment must run a single Uvicorn worker. Running more than one
+worker without a shared cross-worker stream bridge breaks run cancellation, SSE
+reconnects, request de-duplication, and IM channels (nginx has no sticky
+sessions, so requests scatter across workers that each keep their own run
+state). This test pins the safe default so it cannot silently regress to a
+multi-worker default, while still allowing operators to override it once a
+shared stream bridge exists.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import yaml
+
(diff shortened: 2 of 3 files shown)


WillemJiang:
@fancyboi999  https://github.com/bytedance/deer-flow/pull/3191 supports sharing the runner status across the gateway instances. It would be much easier for us to address the problem by merging the patch #3191.

fancyboi999:
Thanks @WillemJiang — agreed #3191 is the right direction for cross-worker streaming, and I read through its diff and left review notes there.

One relevant point: #3191 resolves the SSE side (#3260) but doesn't touch `RunManager` (`cancel` / `create_or_reject` dedup) or per-worker IM channel startup. So with `GATEWAY_WORKERS>1`, even after #3191 merges, `cancel` still hits the local process's `_runs` (the #3239 409) and each worker still starts its own IM channel service — the live SSE path is fixed, but cancel/dedup/IM are not yet cross-worker.

So this PR isn't really an alternative to #3191 — it's a zero-risk default (`workers=1`) covering that remaining gap until a cross-worker `RunManager` lands, and it can be reverted the moment multi-worker is fully safe. Happy to fold the default change into #3191 or a follow-up, or close this if you'd rather track it there — your call.

WillemJiang:
@fancyboi999, as we are close to the 2.0-m1 release, we can move the stop runner support of multiple workers in the next version of DeerFlow.  I will merge this quick fix first.

WillemJiang (approved):

PR #3480, opened 2026-06-10
feat(runtime): Redis Streams stream bridge for multi-worker streaming

18062706139fcz:
## Summary

Adds a **Redis-backed `StreamBridge`** so run events can be streamed across multiple gateway workers / replicas, not just within the single process that produced them. Today `MemoryStreamBridge` (in-process `asyncio.Queue`) is the only backend, so a run started on worker A cannot be joined/streamed from worker B — which blocks any horizontally-scaled deployment behind a load balancer.

This PR also adds the **terminal-reconciliation plumbing** required to make late joins and lost-`END` recovery correct over a shared backend, since with Redis the subscriber and producer are no longer guaranteed to share memory or lifecycle.

Default behavior is unchanged: `stream_bridge.type` defaults to `memory`. Redis is opt-in via config + the `deerflow-harness[redis]` extra.

## Usage

Enable in `config.yaml`:

```yaml
stream_bridge:
  type: redis
  redis_url: $REDIS_URL          # only the whole-value "$VAR" form is resolved
  queue_maxsize: 256             # approx MAXLEN per run
  redis_max_command_connections: 64
  redis_max_blocking_connections: 1024
  redis_pool_timeout: 1.0
  redis_ttl_seconds: 86400       # fallback expiry when cleanup is missed (>=60)
```

Install the optional dependency:

```bash
pip install "deerflow-harness[redis]"
```

`stream_bridge` is a **startup-only** field (registered in the reload boundary) — changing it requires a process restart, like `database` / `checkpointer`.

## Design

**Producer / consumer decoupling.** `StreamBridge` mirrors LangGraph Platform's Queue + StreamManager split: the worker `publish()`es events, any number of SSE endpoints `subscribe()`. The new `supports_cross_process_subscribe` property lets routers allow `store_only` runs (active on another worker) to be joined when the backend can actually see them.

**Redis Streams layout.** One stream key per run. `XADD` appends events; the Redis entry id (`<ms>-<seq>`) is used **directly** as the SSE `id:`, so `Last-Event-ID` reconnection is just an `XRANGE` from the cursor. `END` is written as a normal entry (`event=__end__`) rather than out-of-band, so ordering and late-join replay are preserved for free.

**Best-effort publish with bounded retry.** Data-event `publish` retries transient connection/timeout errors with short backoff, then drops + logs the frame rather than raising — Redis availability must not couple to run success rate. Note: redis-py's `ConnectionError`/`TimeoutError` do **not** inherit from builtin `OSError`, so all three are caught explicitly (this was a real latent bug — the original `except (ConnectionError, TimeoutError, OSError)` never matched redis-py's exceptions).

**Dual connection pools.** A command pool (publish / cleanup / refresh_ttl / boundary probes) is isolated from a blocking-subscribe pool (one connection per active SSE subscription), so a flood of subscribers cannot starve publishes.

**TTL retention + active-run keeper.** Keys carry a fallback TTL so missed cleanup cannot leak streams forever. On the publish hot path `refresh_ttl` is sampled at most once / 60s; for sparse long-idle runs the worker also runs a TTL keeper task that refreshes independently of event rate, cancelled when the run goes terminal.

**Three convergent `end` paths.** Once subscriber and producer can live on different workers, a lost best-effort `publish_end` could hang a subscription on heartbeats forever. Handled by:
1. **real END** — normal `__end__` entry replayed/streamed;
2. **terminal short-circuit** (router) — a late join on a terminal run whose retained stream is already gone returns a single `end` immediately instead of subscribing;
3. **synthetic end** (consumer) — during sustained silence the consumer reconciles against the `RunStore`; if the run is terminal *and* no event exists after the cursor, it synthesises an `end`.

All three reuse `build_end_payload()` from the new `runtime/runs/terminal.py` (also the single source of `TERMINAL_STATUSES`) so they share one `{status, error}` schema and never drift.

**Honest cross-worker degradation.** Cross-worker *cancel* is not supported (the task lives on the owner worker and the Redis bridge does not forward control commands), so it now returns `409` with an actionable message instead of silently no-op'ing; `on_disconnect=cancel` on a `store_only` run logs the skip rather than swallowing it.

**RunStore H-compat.** Legacy rows with an empty `user_id` are no longer hard-filtered out of `get()`; only a non-empty mismatching `user_id` is rejected (owner_check + thread_id validation remain the access backstop).

## Files

- `runtime/stream_bridge/redis.py` — `RedisStreamBridge` (new)
- `runtime/stream_bridge/{base,async_provider,__init__}.py` — protocol extensions + factory wiring
- `runtime/runs/terminal.py` — shared terminal helpers (new)
- `runtime/runs/worker.py` — TTL keeper + best-effort end/error/cleanup
- `app/gateway/{routers/thread_runs,services}.py` — cross-process join, short-circuit, reconciliation
- `config/stream_bridge_config.py`, `config.example.yaml` (config_version → 12), reload boundary, pyproject extras
- `runs/store/memory.py`, `persistence/run/sql.py` — H-compat user_id handling
- `backend/CLAUDE.md` — Stream Bridge architecture section

## Testing

- `tests/test_redis_stream_bridge.py` — fakeredis unit tests (publish/replay/resume/heartbeat/TTL/oversized/cleanup/config validation)
- `tests/test_terminal_reconcile.py` — short-circuit + synthetic-end reconciliation
- `tests/test_worker_stream_ttl.py` — TTL keeper + safe cleanup
- `tests/test_run_store_user_scope.py` — H-compat legacy user_id
- Verified end-to-end against a **real `redis-server`** (not fakeredis), including a cross-worker producer→consumer round trip.

### DIFF
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -462,6 +462,23 @@ LangSmith and Langfuse are both supported. The wiring lives in two layers:
 
 Returns `{}` when Langfuse is not in the enabled providers — LangSmith-only deployments are unaffected. Set `DEER_FLOW_ENV` (or `ENVIRONMENT`) to tag traces by deployment environment. Tests live in `tests/test_tracing_factory.py`, `tests/test_tracing_metadata.py`, `tests/test_worker_langfuse_metadata.py`, and `tests/test_client_langfuse_metadata.py`.
 
+### Stream Bridge (`packages/harness/deerflow/runtime/stream_bridge/`)
+
+Decouples agent workers (producers) from SSE endpoints (consumers), aligning with LangGraph Platform's Queue + StreamManager architecture. A run's events are published by the worker and replayed/streamed to any number of SSE subscribers (with `Last-Event-ID` reconnection support).
+
+- **Abstract protocol** (`base.py`): `StreamBridge` with `publish` / `publish_end` / `subscribe` / `cleanup` / `refresh_ttl` / `has_retained_stream` / `has_events_after` / `ping`. `subscribe()` yields `StreamEvent` plus `HEARTBEAT_SENTINEL` (idle) and `END_SENTINEL` (producer done). `supports_cross_process_subscribe` tells routers whether `store_only` runs on another worker can be joined.
+- **`MemoryStreamBridge`** (`memory.py`): in-process `asyncio.Queue`, single-process only (`supports_cross_process_subscribe=False`).
+- **`RedisStreamBridge`** (`redis.py`, extra `deerflow-harness[redis]`): Redis Streams backend for multi-worker / multi-replica deployments (`supports_cross_process_subscribe=True`). Any replica can subscribe to a run regardless of which worker produced the events. Key behaviors:
+  - Data-event `publish` is **best-effort with bounded retry**: transient `redis.ConnectionError` / `redis.TimeoutError` / `OSError` are retried with short backoff (`_PUBLISH_RETRY_BACKOFFS`), then the frame is dropped + logged rather than raised, so Redis availability does not couple to run success rate. Note redis-py's `ConnectionError` / `TimeoutError` do **not** inherit from builtin `OSError`, so all three are caught explicitly (`self._retryable_errors`).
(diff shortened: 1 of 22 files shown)


WillemJiang:
The issue was addressed in #3191

PR #3483, opened 2026-06-10
fix(docker): auto-create .env files from examples on start

imkratos:
---
  # Fixes 

  ## Why

  First-time users who clone the repo and jump straight to `./scripts/docker.sh start`
  (or follow the Docker quick-start path) hit an immediate startup failure if they
  haven't manually copied `.env.example → .env` and `frontend/.env.example → frontend/.env`.
  This is a common friction point that produces a confusing error with no clear remediation
  hint — especially for users who are not yet familiar with the project layout.

  ## What changed

  `scripts/docker.sh` now auto-provisions missing env files before starting the stack:

  - A reusable `ensure_env_file <example> <target> <label>` helper was added.
    It is a no-op when the target already exists, copies from the example when it doesn't,
    and exits with a clear message if the example itself is also absent.
  - `start()` calls the helper twice — once for the root `.env` and once for `frontend/.env` —
    so both locations are covered in a single startup pass.

  Existing users with `.env` files already in place are unaffected.

  ## Surface area

  - [x] **Sandbox** — change is inside `docker/` / `scripts/docker.sh`

  ## Validation

  Simulate a clean-clone state

  mv .env .env.bak && mv frontend/.env frontend/.env.bak

  ./scripts/docker.sh start
  Expected: two "Created … from example" lines, then normal container bring-up

  Verify idempotence — no duplicate copies on second run

  ./scripts/docker.sh start
  Expected: no "Created" lines, normal startup

  Restore

  mv .env.bak .env && mv frontend/.env.bak frontend/.env

  ## AI assistance

  **Tool(s) used:** Claude Code
  **How you used it:** Generated the PR description from the commit diff and existing code context.

### DIFF
--- a/scripts/docker.sh
+++ b/scripts/docker.sh
@@ -41,6 +41,24 @@ load_proxy_env_from_dotenv() {
     done
 }
 
+ensure_env_file() {
+    local example_file="$1"
+    local env_file="$2"
+    local label="$3"
+
+    if [ -f "$env_file" ]; then
+        return
+    fi
+
+    if [ ! -f "$example_file" ]; then
+        echo -e "${YELLOW}✗ $label not found and no example file to copy from: $example_file${NC}"
+        exit 1
+    fi
+
+    cp "$example_file" "$env_file"
+    echo -e "${BLUE}Created $label from example${NC}"
+}
+
 detect_sandbox_mode() {
     local config_file="$PROJECT_ROOT/config.yaml"
     local sandbox_use=""
@@ -245,6 +263,9 @@ start() {
             echo -e "${BLUE}Created empty extensions_config.json${NC}"
         fi
     fi
+
+    ensure_env_file "$PROJECT_ROOT/.env.example" "$PROJECT_ROOT/.env" ".env"
+    ensure_env_file "$PROJECT_ROOT/frontend/.env.example" "$PROJECT_ROOT/frontend/.env" "frontend/.env"
 
     load_proxy_env_from_dotenv
 


frontend/src/app/workspace/agents/[agent_name]/edit/page.tsx
```diff
@@ -0,0 +1,213 @@
+"use client";
```
Comment (WillemJiang) [inline frontend/src/app/workspace/agents/[agent_name]/edit/page.tsx]: @imkratos  This commit is not related with PR topic, please remove it from this PR.

imkratos:
@WillemJiang  done

WillemJiang:
@imkratos, please click the CLA assistant button to sign the CLA before we keep reviewing your PR.

imkratos:
@WillemJiang  done

scripts/docker.sh
```diff
@@ -41,6 +41,24 @@ load_proxy_env_from_dotenv() {
...
+    if [ -f "$env_file" ]; then
+        return
+    fi
+
+    if [ ! -f "$example_file" ]; then
+        echo -e "${YELLOW}✗ $label not found and no example file to copy from: $example_file${NC}"
+        exit 1
+    fi
+
+    cp "$example_file" "$env_file"
```
willem-bd, line 58: **[P2] Freshly created `.env` inherits `.env.example`'s mode (typically 0644).**

`cp` preserves the source's mode bits by default. `.env.example` is checked into the repo, so it's almost always 0644 (world-readable). The user is then expected to paste real `OPENAI_API_KEY` / `TAVILY_API_KEY` / etc. into that file. On a shared dev host, multi-user box, or shared CI runner, any other local user can `cat .env` and exfiltrate the keys.

The natural place to harden this is right here in the helper, scoped to `.env` specifically (config.yaml at line 232 has the same gap but is less secret-sensitive):

```bash
cp "$example_file" "$env_file"
chmod 600 "$env_file"
echo -e "${BLUE}Created $label from example${NC}"
```

Or restrict the chmod to the `.env` label only if you want config.yaml-style files to keep their group-readable mode.

Severity: P2 — the script only does the initial copy; it doesn't write secrets itself. But this PR is the first time the project creates `.env` for the user, so it's also the first time the project has any say in what mode it lands on disk with.

willem-bd (commented):
**Code review (max effort) — 2 findings**

Reviewed the 21-line shell change line-by-line, walked the enclosing `start()` flow, and verified no regression for users with a pre-existing `.env`. The core fix is correct: early `return` makes it a true no-op when `.env` is present, and the new helper sequences before `load_proxy_env_from_dotenv` so the proxy loader sees the freshly created file.

Two non-blocking observations below — one security hardening (P2), one altitude flag (no diff-introduced bug). The change is safe to merge as-is.

**Also verified (no findings):** `set -e` + bare `return` interaction; symlink/broken-link handling; ordering vs proxy loader; idempotence on second run; missing-example exit semantics.

scripts/docker.sh
```diff
@@ -41,6 +41,24 @@ load_proxy_env_from_dotenv() {
     done
 }
 
+ensure_env_file() {
```
willem-bd, line 44: **Altitude — four near-identical "if missing, copy from example" blocks now exist with subtly different error semantics.**

Not a diff-introduced bug — flagging because this PR is the natural place to consolidate before the divergence calcifies.

After this PR, `start()` has four blocks that all answer "if the target file is missing, copy from an example," each with different semantics:

| Target | Missing file behavior | Missing example behavior |
| --- | --- | --- |
| `config.yaml` (L230–249) | copies + prints yellow banner + `exit 0` | `exit 1` |
| `extensions_config.json` (L253–261) | copies + blue line | falls back to writing `{}` |
| `.env` (new) | copies + blue line | `exit 1` |
| `frontend/.env` (new) | copies + blue line | `exit 1` |

Three different missing-example behaviors for what is fundamentally the same operation. Future authors have to remember which file gets which behavior, and a fix in one block — e.g. adding `chmod 600` per the other comment, or improving the error message — won't propagate.

A single helper that takes the per-file divergence as parameters (e.g. `--on-missing-example=exit1|fallback|banner-exit0`) would let all four call sites share one implementation. Not blocking this PR; raising as a follow-up direction so the next file-bootstrap block doesn't add a fifth shape.

PR #3488, opened 2026-06-10
fix(frontend): prevent user message bubble overflow with long unbreakable strings

### LINKED ISSUE #3484: [bug] Horizontal overflow from long chat messages not constrained in container.
### Before you start

### Problem summary

在消息输入框里输入一段长消息并且不换行，会导致文本溢出。

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

在我尝试复制一段长文本提示词粘贴到对话框时，发现发送出去的提示词会横向溢出，只能通过底部横向滚动条去浏览发送出去的内容。

### Expected behavior

发送长文本提示词应该限制应约束在消息容器宽度内，应自动换行或截断，不应水平溢出。

### Steps to reproduce

在底部输入框输入一段长文本，且不允许换行。可以稳定复现。

### Relevant logs

```shell
无
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

Windows 11 (x86_64)

### Python version

Python: 3.11.5

### Node.js version

Node.js: v20.14.0

### pnpm version

pnpm: 10.33.4

### uv version

uv: 0.11.19

### Git state

最新提交 SHA 值:  0fb18e3

### Additional context

Image

Image

ech0hol:
fix(frontend): prevent user message bubble overflow with long unbreakable strings

- Add max-w-full min-w-0 to constrain message container width
- Change bubble width from w-fit to conditional w-full max-w-full (user) / w-fit (assistant)
- Move break-words from outer wrapper to AIElementMessageResponse for precise text wrapping

Fixes #3484

## Why

用户输入包含长不可换行字符串（如 `=====================================`、长 URL、无空格连续字符）时，用户消息气泡会被内容无限撑宽，超出页面容器导致**页面出现横向滚动条**。

触发场景：用户粘贴长代码注释分隔线、长 URL、连续符号等无换行机会的字符串。

## What changed

- 用户消息气泡宽度现在与 AI 回复气泡保持一致，不再随内容无限扩展
- 长不可换行字符串现在会在气泡边界处自动换行，不再撑破容器
- 页面不再因用户消息出现横向滚动条

## Surface area

- [x] **Frontend UI** — page / component / setting / interaction under frontend/

## Screenshots / Recording
改动后的代码块或Markdown文本可以正常展示
QQ_1781108821784
QQ_1781108769790

## Bug fix verification

- 问题可通过在前端输入框中粘贴长不可换行字符串（如 50+ 个 `=` 号）复现
- 属于纯 CSS 样式 bug，不涉及逻辑层，不适合用单元测试覆盖
- 验证方式：Docker 开发环境中实测长字符串气泡换行正常，横向滚动条消失

## Validation

```bash
# 前端单元测试
# 结果：23/23 test files passed, 167/167 tests passed
```

### DIFF
--- a/frontend/src/components/workspace/messages/message-list-item.tsx
+++ b/frontend/src/components/workspace/messages/message-list-item.tsx
@@ -302,6 +302,7 @@ function MessageContent_({
   if (isHuman) {
     const messageResponse = contentToDisplay ? (
       <AIElementMessageResponse
+        className="break-words"
         remarkPlugins={humanMessagePlugins.remarkPlugins}
         rehypePlugins={humanMessagePlugins.rehypePlugins}
         components={components}
@@ -311,10 +312,15 @@ function MessageContent_({
       </AIElementMessageResponse>
     ) : null;
     return (
-      <div className={cn("ml-auto flex flex-col gap-2", className)}>
+      <div
+        className={cn(
+          "ml-auto flex max-w-full min-w-0 flex-col gap-2",
+          className,
+        )}
+      >
         {filesList}
         {messageResponse && (
-          <AIElementMessageContent className="w-fit">
+          <AIElementMessageContent className="w-full max-w-full">
             {messageResponse}
           </AIElementMessageContent>
         )}


WillemJiang:
@snaplap  thanks for your contribution. Please fix the lint error and here are some review comments on the PR.

1. overflow-x-clip replaces overflow-visible globally — In message.tsx:48, the change from overflow-visible to overflow-x-clip affects all message types (user and assistant), not just user messages. If assistant messages had any intentional overflow behavior (e.g., tooltips or absolutely positioned elements), this could cause them to clip unexpectedly.

Suggestion: Consider scoping the overflow change to user messages only:
`"overflow-visible group-[.is-user]:overflow-x-clip"`

2. break-words on the wrapper vs. the content — The break-words class is added to the outer wrapper MessageContent, but the actual text is inside MessageResponse. Depending on the nesting, break-words on the parent may not propagate correctly to the deepest text node if there are intermediate elements with white-space: nowrap or similar. Worth verifying that the word-breaking actually applies where the text renders.

3. w-full max-w-full on bubble vs. assistant messages — In message-list-item.tsx:319, the change to AIElementMessageContent className="w-full max-w-full" affects the bubble for both user and assistant messages (this component is used for both paths in the same file). The original w-fit gave assistant messages a natural-width bubble. Now both will stretch to w-full, which may visually change how assistant responses look (they'll take full width even for short replies).

Suggestion: Apply w-full max-w-full conditionally:
`<AIElementMessageContent className={cn(isHuman ? "w-full max-w-full" : "w-fit")}>`

ech0hol:
@WillemJiang Thanks for the detailed review! All three points addressed in this update:

1.`overflow-x-clip` scoping — You're right that it shouldn't be global. In fact, after deeper analysis, I realized `message.tsx `doesn't need any changes at all. My initial fix was over-scoped — I added `overflow-x-clip` there trying to clip overflow at the `MessageContent` level, but the real root cause is that the user message bubble in `message-list-item.tsx` lacked width constraints. Once `max-w-full` and `min-w-0` are added to the wrapper div there, the original `overflow-visible` works fine for both message types. So `message.tsx` has been reverted to match main with no modifications.

2.`break-words` moved to inner element — Great suggestion. Moved from the outer `MessageContent` wrapper to `AIElementMessageResponse`, which directly wraps the rendered text. This ensures `overflow-wrap: break-word` applies at the text rendering layer with no risk of intermediate CSS interference.

3.`w-full max-w-full` conditional — Appreciate the suggestion! I initially implemented the ternary as you suggested, but upon closer inspection, this `AIElementMessageContent` is already inside an `if (isHuman)` block (line 302), so `isHuman` is always `true` here and the `"w-fit" ` branch is unreachable. The assistant path uses a separate `AIElementMessageContent` (line 333) with its own className, so it's unaffected regardless. Simplified to just `className="w-full max-w-full"` to avoid dead code.

WillemJiang (approved):

PR #3494, opened 2026-06-11
[codex] Fix stale AIO sandbox cache reuse

### LINKED ISSUE #3474: [bug] AioSandboxProvider warm pool holds stale container reference after crash, causing persistent Connection refused until gateway restart
### Before you start

### Problem summary

When a sandbox container dies unexpectedly (OOM, manual removal, host reboot, etc.), the warm pool in `AioSandboxProvider` continues to hold a stale in-process reference to the dead container. Every subsequent tool call (`bash`, `read_file`, `write_file`) against that thread receives `[Errno 111] Connection refused` and the provider never attempts to recreate the sandbox. The only recovery is a full gateway restart.

### Affected area(s)

Sandbox / `deerflow.community.aio_sandbox`

### What happened?

A sandbox container was running normally. At some point the container disappeared (exact cause unknown — possibly killed by the OS). The gateway's warm pool was unaware:

```
INFO  - Reusing in-process sandbox 7b85d00d for thread 697632fa-...
ERROR - Failed to write file in sandbox: [Errno 111] Connection refused

INFO  - Reusing in-process sandbox 7b85d00d for thread 697632fa-...
ERROR - Failed to execute command in sandbox: [Errno 111] Connection refused

INFO  - Reusing in-process sandbox 7b85d00d for thread 697632fa-...
ERROR - Failed to read file in sandbox: [Errno 111] Connection refused

# Eventually the idle timeout kicks in and tries to clean up:
WARNING - Failed to stop container deer-flow-sandbox-7b85d00d: Error response from daemon: No such container: deer-flow-sandbox-7b85d00d
INFO  - Destroyed idle warm-pool sandbox 7b85d00d
```

After the container is "destroyed" from the pool, a new request for the same thread still cannot get a working sandbox — the thread is effectively stuck until gateway restart.

### Expected behavior

When `Connection refused` is returned on a sandbox operation, the provider should:

1. Mark the sandbox as unhealthy and remove it from the warm pool immediately
2. Provision a new sandbox container for the thread
3. Retry the failed operation (or at minimum surface a recoverable error so the agent can retry)

This is essentially a **testOnBorrow** pattern common in connection pool implementations.

### Steps to reproduce

1. Start a conversation thread that uses sandbox tools (bash, read_file, etc.)
2. While the sandbox container is running, manually remove it: `docker rm -f deer-flow-sandbox-<id>`
3. Continue sending messages in the same thread
4. Observe that all sandbox tool calls return `Connection refused` indefinitely

### Environment

- Deployment: Docker Compose (self-hosted)
- Sandbox provider: `deerflow.community.aio_sandbox:AioSandboxProvider`
- Gateway version: latest main (post `a57d05fe`)

hetaoBackend:
## Summary
- Validate cached active AIO sandboxes and warm-pool entries before handing them back to a thread.
- Drop stale sandbox state from all provider maps so dead containers do not keep returning stale clients.
- Keep async sandbox lookup paths off the event loop after cached lookup can perform backend health checks.

## Root Cause
`AioSandboxProvider` reused `_sandboxes` and `_warm_pool` entries by in-memory id only. If the backing container disappeared, the provider kept returning the old `AioSandbox` client, so tool calls stayed on `[Errno 111] Connection refused` until the gateway restarted.

## Validation
- `PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run pytest tests/test_aio_sandbox_provider.py tests/test_sandbox_middleware.py -q`
- `PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run pytest tests/test_sandbox_orphan_reconciliation.py tests/test_remote_sandbox_backend.py tests/test_aio_sandbox_local_backend.py -q`
- `PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run ruff check packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py packages/harness/deerflow/sandbox/tools.py tests/test_aio_sandbox_provider.py tests/test_sandbox_middleware.py`
- `git diff --check`

Fixes #3474


backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
```diff
@@ -777,6 +840,17 @@ def get(self, sandbox_id: str) -> Sandbox | None:
...
             The sandbox instance if found, None otherwise.
         """
+        with self._lock:
+            sandbox = self._sandboxes.get(sandbox_id)
+            info = self._sandbox_infos.get(sandbox_id)
+
+        if sandbox is None:
+            return None
+
+        if info is not None and not self._is_tracked_sandbox_alive(sandbox_id, info):
```
hetaoBackend: **Blocking I/O on the event loop via `get()`.** This change turns `get()` from a pure in-memory lookup into a call that performs blocking backend I/O (`docker inspect` subprocess in `LocalContainerBackend.is_alive`; a provisioner HTTP request in the remote backend). The PR wraps the two `get()` call sites in `ensure_sandbox_initialized_async` in `asyncio.to_thread`, but other callers still invoke `get()` directly on the event loop:

- `ToolOutputBudgetMiddleware.awrap_tool_call` → `_resolve_sandbox` → `get_sandbox_provider().get(sandbox_id)` (`tool_output_budget_middleware.py:612` / `:319`). The comment right above that call relies on the old invariant: *"_resolve_sandbox only touches runtime.state and the provider's in-memory sandbox registry, so it is safe to call on the event loop."* That invariant is now broken — every budgeted tool result stalls the event loop for a subprocess/HTTP round trip.

Consider keeping `get()` I/O-free (validate only in acquire paths, or reactively on connection error), or adding a `get_async` provider hook mirroring `acquire_async` and updating these call sites.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
```diff
@@ -527,6 +550,46 @@ def _register_created_sandbox(self, thread_id: str | None, sandbox_id: str, info
         logger.info(f"Created sandbox {sandbox_id} for thread {thread_id} at {info.sandbox_url}")
         return sandbox_id
 
+    def _is_tracked_sandbox_alive(self, sandbox_id: str, info: SandboxInfo) -> bool:
+        """Return whether a tracked sandbox still appears alive."""
+        try:
+            return self._backend.is_alive(info)
+        except Exception as e:
+            logger.warning(f"Failed to check sandbox {sandbox_id} health: {e}")
+            return False
```
hetaoBackend: **A transient health-check failure destroys a healthy container.** Any exception from `_backend.is_alive()` is treated as "dead", and all three callers then run `_drop_unhealthy_sandbox`, which calls `self._backend.destroy(info)`. So a momentary Docker-daemon hiccup or provisioner HTTP timeout — an error in the *check*, not the container — tears down a healthy container that may have in-flight executions and per-thread state. (`LocalContainerBackend._is_container_running` only catches `CalledProcessError`/`TimeoutExpired`, so e.g. `OSError` from a busy daemon lands here.) Before this PR, transient backend errors never destroyed anything, and this now runs on every cached acquire/get.

Suggest distinguishing "check failed" from "definitely dead": on exception, keep the entry (or drop it from the in-process maps without calling `backend.destroy`), and only destroy when `is_alive` returns a definitive `False`.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
```diff
@@ -527,6 +550,46 @@ def _register_created_sandbox(self, thread_id: str | None, sandbox_id: str, info
...
+
+    def _drop_unhealthy_sandbox(self, sandbox_id: str, reason: str) -> None:
+        """Remove a dead sandbox from all in-process tracking maps."""
+        sandbox = None
+        info = None
+        thread_ids_to_remove: list[str] = []
+
+        with self._lock:
+            sandbox = self._sandboxes.pop(sandbox_id, None)
```
hetaoBackend, line 593: **TOCTOU: `_drop_unhealthy_sandbox` can destroy a freshly recreated sandbox.** The health check runs on an `info` snapshot taken with `self._lock` released, but the drop is keyed by id only and removes/destroys whatever is *currently* registered under that id. Sandbox ids are deterministic per thread, and `get()` is not serialized by the per-thread lock (only `acquire` is), so this interleaving is possible:

1. Thread A (`get(S)`) snapshots `info_old`; the container dies.
2. Thread B (acquire path, holding the per-thread lock) detects the death, drops, and re-creates the sandbox under the same id `S` with `info_new`.
3. Thread A's health check on `info_old` fails → `_drop_unhealthy_sandbox(S)` pops the new entry and calls `backend.destroy(info_new)`, killing B's brand-new healthy container.

Consider passing the checked `info` in and only dropping when `self._sandbox_infos.get(sandbox_id) is info` (with a warm-pool equivalent), so a failed check can only evict the exact entry it validated.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -1195,7 +1196,7 @@ async def ensure_sandbox_initialized_async(runtime: Runtime | None = None) -> Sa
 
     runtime.state["sandbox"] = {"sandbox_id": sandbox_id}
 
-    sandbox = provider.get(sandbox_id)
+    sandbox = await asyncio.to_thread(provider.get, sandbox_id)
```
hetaoBackend: **Every sandbox tool call now pays backend health checks — and this one is redundant.** `ensure_sandbox_initialized(_async)` runs per tool invocation (`bash`, `ls`, `read_file`, `write_file`, …), and `provider.get()` now does an unconditional `is_alive` round trip (a `docker inspect` subprocess with a 5s timeout locally; provisioner HTTP remotely), with no TTL/debounce. On this specific path the sandbox was just validated (or freshly created) by `acquire_async` a few lines above — this `get()` health-checks it a second time milliseconds later. It also adds a new failure mode: if `is_alive` flaps once here, the just-acquired sandbox is destroyed and we raise `SandboxNotFoundError` even though acquisition succeeded.

Consider a short TTL on health verdicts (skip re-check within N seconds of the last successful check/registration), or validating only in the acquire/reuse paths rather than in `get()`.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
```diff
@@ -527,6 +550,46 @@ def _register_created_sandbox(self, thread_id: str | None, sandbox_id: str, info
...
 
+    def _is_tracked_sandbox_alive(self, sandbox_id: str, info: SandboxInfo) -> bool:
+        """Return whether a tracked sandbox still appears alive."""
+        try:
+            return self._backend.is_alive(info)
+        except Exception as e:
+            logger.warning(f"Failed to check sandbox {sandbox_id} health: {e}")
+            return False
+
+    def _drop_unhealthy_sandbox(self, sandbox_id: str, reason: str) -> None:
```
hetaoBackend: **Near-verbatim duplicate of `destroy()`.** The lock-guarded five-map cleanup here is byte-for-byte identical to `destroy()` (including the warm-pool `if/else` info recovery), and the close/destroy tail differs only by a try/except around `backend.destroy` and log wording. This tricky multi-map cleanup now exists twice — a future map (or a fix to the warm-pool branch) has to be applied in both places. Suggest implementing this as a thin wrapper around `destroy()` (with the try/except here), or extracting a shared private helper both call.

ShenAC-SAC:
@hetaoBackend Thanks for the fix — the testOnBorrow direction and the `expected_info` TOCTOU guard look good. One issue worth addressing before merge:

**Both real backends report transient failures as `is_alive() == False`, so the "unknown ≠ dead" safety valve in the provider never actually engages.**

`_check_tracked_sandbox_alive` only returns `None` (treat-as-alive) when `backend.is_alive()` *raises*, and `_drop_unhealthy_sandbox` destroys on a definitive `False`. But neither backend ever raises:

- `local_backend.py` → `_is_container_running` runs `docker inspect` without `check=True` and catches `TimeoutExpired`, so a busy/unreachable Docker daemon (non-zero returncode, or the 5s timeout) returns `False` — indistinguishable from "container is gone".
- `remote_backend.py` → `_provisioner_is_alive` catches every `RequestException` (provisioner restart, network blip, timeout) and returns `False`.

Consequence: a transient infra hiccup at acquire time now destroys a **healthy** sandbox (in-container state like installed packages and running processes is lost, plus a cold start; pod churn in provisioner mode). That's a regression risk this PR introduces — before it, no health check ran on the cached-reuse path, so there was no misfire to have.

The guard test `test_acquire_keeps_cached_sandbox_when_health_check_errors` mocks `is_alive` with `side_effect=OSError(...)`, a code path the shipped backends can never take — so it stays green without protecting the real scenario.

Suggested fix: make `is_alive` distinguish "definitively not running" from "check failed":
- local: `returncode == 0 && Running=false` or a "No such object" stderr → `False`; daemon errors / timeouts → raise (or return `None` if you widen the contract to `bool | None`);
- remote: 404 or `status != Running` → `False`; `RequestException` → raise/`None`.

(`is_alive` has other callers, e.g. orphan reconciliation, so worth re-checking those when changing the contract.)

hetaoBackend:
Good catch — fixed in a4d42cb8. `LocalContainerBackend.discover()` now catches the `RuntimeError` from `_is_container_running` and returns `None`, so an unverifiable container is simply not adopted and acquire falls through to create instead of hard-failing. Added `test_discover_returns_none_when_runtime_check_fails` / `..._times_out` to pin the behavior. (`RemoteSandboxBackend._provisioner_discover` already swallowed `RequestException` → `None`, so the remote path was unaffected.)

hetaoBackend:
@ShenAC-SAC Addressed in 0b5a0858: both backends now distinguish "definitively not running" from "check failed" — local returns `False` only for `Running=false` or a no-such-container error and raises `RuntimeError` on daemon errors/timeouts; remote returns `False` only on 404/`status != Running` and raises on `RequestException`. The provider treats a raise as unknown and keeps the entry. Backend-level tests now exercise the real code paths (`test_is_container_running_raises_on_runtime_error` / `..._on_timeout`, and the remote equivalents), not just a mocked `side_effect`.

Re the other-callers caveat: orphan reconciliation uses `list_running()` (unaffected), and the one caller that did slip through — `discover()` sharing `_is_container_running` — is fixed in a4d42cb8 so a transient blip during discovery falls through to create instead of failing acquire.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
hetaoBackend: Addressed in 7829a853: `get()` is back to a pure in-memory lookup (validation moved to the acquire/reuse paths), the `asyncio.to_thread` wraps in `ensure_sandbox_initialized_async` were reverted, and `test_get_uses_in_memory_registry_only` pins that `get()` never calls `backend.is_alive`.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
hetaoBackend: Addressed in 7829a853 + 0b5a0858: `_check_tracked_sandbox_alive` returns `bool | None` and entries are only dropped on a definitive `False`; both backends now raise on transient errors instead of reporting dead (local: no-such-container → `False`, daemon error/timeout → raise; remote: 404 → `False`, `RequestException` → raise). Pinned by `test_acquire_keeps_cached_sandbox_when_health_check_errors` and backend-level tests.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
hetaoBackend: Addressed in 7829a853: `_drop_unhealthy_sandbox` now takes `expected_info`, and `_remove_tracked_sandbox` only removes the entry when the currently tracked `SandboxInfo` (active or warm-pool) is the exact object that was health-checked. Pinned by `test_drop_unhealthy_sandbox_skips_recreated_entry`.

backend/packages/harness/deerflow/sandbox/tools.py
hetaoBackend: Addressed in 7829a853 by moving validation out of `get()` entirely: steady-state tool calls hit only the in-memory registry, backend checks run once per acquire/reuse, and the redundant post-acquire re-check (and its `SandboxNotFoundError` flap mode) is gone.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
hetaoBackend: Addressed in 7829a853: the five-map cleanup is extracted into `_remove_tracked_sandbox`, shared by both `destroy()` and `_drop_unhealthy_sandbox` (which adds the `expected_info` guard and swallowed `backend.destroy`).

WillemJiang:
@hetaoBackend， here is one review comment for the code

 _is_no_such_container_error — overly broad "not found" check

  local_backend.py:172-174

  ```
def _is_no_such_container_error(stderr: str) -> bool:
      message = stderr.lower()
      return "no such object" in message or "no such container" in message or "not found" in message
```

  "not found" is very broad. Docker error messages like command not found, path not found, or even image-pull errors contain this substring. A transient error like:

  `Error: command not found in $PATH`

  …would be treated as "container definitely doesn't exist" (return False) rather than raising RuntimeError for a retry. Consider narrowing to "not found" in the message and ("container" in the message or "object" in the message), or require it to match a more specific pattern.

hetaoBackend:
@WillemJiang Good point — fixed in 547d5a50. `_is_no_such_container_error` now keeps Docker's specific phrases (`No such object` / `No such container`) and only trusts a generic `not found` (Apple Container's wording) when the message also names the inspected container or refers to a container/object. Transient errors like `command not found` / `credential helper not found` now stay on the `RuntimeError` path. Pinned by `test_is_container_running_raises_on_unrelated_not_found_error` and `test_is_container_running_false_on_apple_container_not_found`.

WillemJiang (approved):

PR #3498, opened 2026-06-11
fix(scripts):start with make start-daemon,can not stop next-server with make stop

airene:
It worked fine initially but started failing after a certain version update—this has been going on for a while.
On Linux, when starting the services with `make start-daemon` and stopping it with `make stop`, the `next-server` fails to shut down correctly 100% of the time.

### DIFF
--- a/scripts/serve.sh
+++ b/scripts/serve.sh
@@ -28,6 +28,11 @@ set -e
 
 REPO_ROOT="$(builtin cd "$(dirname "${BASH_SOURCE[0]}")/.." >/dev/null 2>&1 && pwd -P)"
 cd "$REPO_ROOT"
+PID_DIR="$REPO_ROOT/logs/pids"
+STOP_TIMEOUT_SECONDS="${DEERFLOW_STOP_TIMEOUT_SECONDS:-10}"
+case "$STOP_TIMEOUT_SECONDS" in
+    ''|*[!0-9]*) STOP_TIMEOUT_SECONDS=10 ;;
+esac
 
 # ── Load .env ────────────────────────────────────────────────────────────────
 
@@ -95,6 +100,12 @@ DEERFLOW_ROOTS="$(
 # the ".../deer-flow" root.
 _is_deerflow_pid() {
     local pid=$1 files root
+
+    if [ -r "/proc/$pid/environ" ] &&
+        tr '\0' '\n' < "/proc/$pid/environ" 2>/dev/null | grep -Fxq "DEERFLOW_DAEMON_ROOT=$REPO_ROOT"; then
+        return 0
+    fi
+
     files=$(lsof -p "$pid" 2>/dev/null) || return 1
     while IFS= read -r root; do
         [ -n "$root" ] || continue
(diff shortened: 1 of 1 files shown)


scripts/serve.sh
```diff
@@ -95,6 +100,12 @@ DEERFLOW_ROOTS="$(
 # the ".../deer-flow" root.
 _is_deerflow_pid() {
     local pid=$1 files root
+
+    if [ -r "/proc/$pid/environ" ] &&
+        tr '\0' '\n' < "/proc/$pid/environ" 2>/dev/null | grep -Fxq "DEERFLOW_DAEMON_ROOT=$REPO_ROOT"; then
```
fancyboi999, line 104: This env tag reads like the actual root-cause fix: it's what finally lets `_is_deerflow_pid` recognise the `next-server` child no matter what `lsof` reports. But once the child carries it, the `_kill_repo_port 3000` (a `kill -9` on whoever holds the port) and `_kill_repo_processes "next-server"` already in `stop_all` should reap it too. Is the recorded-PGID / setsid path covering a case those two can't, like the dev server respawning its worker mid-stop, or wanting a graceful TERM over `kill -9`? If the env tag alone fixed your repro, a fair chunk of the new code may be doing what `stop_all` already does.

scripts/serve.sh
```diff
@@ -423,7 +550,30 @@ run_service() {
...
+            nohup env DEERFLOW_DAEMON_ROOT="$REPO_ROOT" DEERFLOW_DAEMON_SERVICE="$service_slug" \
+                bash -c 'echo "$$" > "$1"; shift; exec sh -c "$1"' bash "$pid_file" "$cmd" > /dev/null 2>&1 &
+        fi
+
+        for _ in 1 2 3 4 5; do
+            [ -s "$pid_file" ] && break
+            sleep 0.1
+        done
+        if [ ! -s "$pid_file" ]; then
+            echo "$!" > "$pid_file"
```
fancyboi999: Minor: in the `setsid` branch just above, `setsid` forks, so `$!` here is the `setsid` parent rather than the session leader that became the service. The in-band `echo "$$" > pid_file` normally wins the 0.5s wait so this fallback rarely fires, but when it does the recorded pid (and the pgid derived from it) point at a process that's already exited, and the group teardown quietly no-ops for that service. The non-setsid branch is fine since nothing forks there.

fancyboi999 (commented):
Good find. The failure is real: `next start` / `next dev` fork a `next-server` child that holds the port, and when the launcher's identity check doesn't recognise that child, every stop path walks right past it.

Recording each daemon's pid/pgid and tearing the whole group down with TERM then KILL is a sound way to make sure the children go with the parent, and the defensive bits are good: every kill re-checks `_is_deerflow_pid` so a recycled PID can't be hit, the `pgid != current_pgid` guard stops the script from signalling its own group, and persisting the recovered PGID before the wait/KILL phases (the comment at L322) is the kind of detail people usually get wrong.

What I'd like to pin down before this much code lands (inline note left at the env tag): on `main`, `stop_all` already tries to reap `next-server` both by name (`_kill_repo_processes "next-server"`) and by port (`_kill_repo_port 3000`, which is a `kill -9` on whoever holds 3000). The reason those miss the child today looks like `_is_deerflow_pid` not recognising it, and the new `DEERFLOW_DAEMON_ROOT` env tag fixes exactly that. So once the child is tagged, the existing port/name reap should already kill it. If the PGID/setsid layer is there for something the simpler path can't handle (a `next dev` worker respawning mid-stop, or wanting a graceful TERM instead of `kill -9`), that's a fair reason, but it's worth spelling out, because ~150 lines of process-group bookkeeping is a lot of surface to own if the ~10-line env tag is what actually closed the bug.

Also +1 to the bot on `STOP_TIMEOUT_SECONDS`: it's read before `.env` is sourced, so a `DEERFLOW_STOP_TIMEOUT_SECONDS` set there won't take effect.

I went through this on macOS (the no-setsid path); I didn't have a Linux box to reproduce your exact case, so this is read off the code rather than a repro. Correct me if the env tag alone didn't do it for you.

scripts/serve.sh
airene: ok I‘ll recheck it.

airene:
@fancyboi999 
I have resubmitted the changes; this version addresses the current situation on Linux and has been manually verified.

The previous version was somewhat complex, as it actually employed two separate mechanisms to ensure proper termination, while also including robust compatibility support for macOS.

Thanks for the great work on deer-flow.

WillemJiang (approved):

PR #3499, opened 2026-06-11
perf(runtime): index runs by thread_id to avoid O(n) scans in RunManager

ly-wang19:
## Why

`RunManager` keeps all live + recently-finished runs in a single `self._runs` dict (each run is removed only after a ~5-minute `cleanup` delay). Three per-thread queries scanned **every** in-memory run to filter by `thread_id`:

- `list_by_thread` (thread run history)
- `create_or_reject` (inflight check before starting a run)
- `has_inflight`

Each is `O(total in-memory runs)`, so under concurrency the cost grows with overall gateway traffic rather than the (usually tiny) number of runs in the queried thread.

## What changed

- Add a `thread_id -> run_ids` secondary index, maintained in lockstep with `_runs` under the existing lock at every mutation site (`create`, `create_or_reject`, both rollback paths, and `cleanup`).
- The three queries above now iterate only the queried thread's runs — `O(runs-in-thread)`.
- The index is an insertion-ordered `dict`-as-ordered-set, so `list_by_thread` keeps its existing stable tie-breaking when `created_at` values tie.
- Externally observable behavior is unchanged. A `self._runs.get(...)` guard in the lookup keeps results correct even under any transient index/`_runs` disagreement.

## Surface area

- [x] **Backend** — `packages/harness/deerflow/runtime/runs/manager.py` + tests. No API / schema change; an internal data-structure optimization.

## Validation

- `cd backend && uv run pytest tests/test_run_manager.py -q` → **51 passed** (45 existing + 6 new: index/`_runs` consistency across create/cleanup/rollback, per-thread scoping, insertion-order preservation, inflight detection)
- Full RunManager-related suite (7 files) → **146 passed**
- `ruff check` + `ruff format` → clean

## AI assistance

**How you used it:** AI audited the runtime for algorithmic hotspots, implemented the secondary index with lock-correct maintenance at all mutation sites, and wrote the consistency/ordering regression tests. I reviewed the change.

### DIFF
--- a/backend/packages/harness/deerflow/runtime/runs/manager.py
+++ b/backend/packages/harness/deerflow/runtime/runs/manager.py
@@ -143,8 +143,14 @@ class RunManager:
         """Return live in-memory records for *thread_id*. Caller must hold ``self._lock``.
 
         Uses the ``_runs_by_thread`` index for O(runs-in-thread) lookup instead of
-        scanning every in-memory run. The ``self._runs.get`` guard keeps the result
-        correct even if the index and ``_runs`` momentarily disagree.
+        scanning every in-memory run. Correctness rests on the index and ``_runs``
+        being mutated in lockstep under ``self._lock`` (no ``await`` between the two
+        writes), so any holder of the lock sees them agree. The ``self._runs.get``
+        filter is defense-in-depth, not reconciliation: it drops a stale id still in
+        the index but already gone from ``_runs``, yet it cannot recover a run that is
+        in ``_runs`` but missing from the index (such a run would be silently
+        omitted). It guards only that one direction, should a future refactor ever
+        break the lockstep invariant.
         """
         run_ids = self._runs_by_thread.get(thread_id)
         if not run_ids:
--- a/backend/packages/harness/deerflow/runtime/runs/manager.py
+++ b/backend/packages/harness/deerflow/runtime/runs/manager.py
@@ -118,9 +118,38 @@ class RunManager:
         persistence_retry_policy: PersistenceRetryPolicy | None = None,
     ) -> None:
         self._runs: dict[str, RunRecord] = {}
+        # Secondary index: thread_id -> insertion-ordered run_id set (a dict is
+        # used as an ordered set), maintained in lockstep with ``_runs`` so
+        # per-thread queries avoid O(total in-memory runs) full scans while
+        # preserving ``_runs`` iteration order (see ``_thread_records_locked``).
+        self._runs_by_thread: dict[str, dict[str, None]] = {}
         self._lock = asyncio.Lock()
         self._store = store
         self._persistence_retry_policy = persistence_retry_policy or PersistenceRetryPolicy()
+
+    def _index_run_locked(self, record: RunRecord) -> None:
+        """Register *record* in the thread index. Caller must hold ``self._lock``."""
+        self._runs_by_thread.setdefault(record.thread_id, {})[record.run_id] = None
+
+    def _unindex_run_locked(self, run_id: str, thread_id: str) -> None:
+        """Drop *run_id* from the thread index. Caller must hold ``self._lock``."""
+        bucket = self._runs_by_thread.get(thread_id)
+        if bucket is not None:
+            bucket.pop(run_id, None)
+            if not bucket:
+                self._runs_by_thread.pop(thread_id, None)
+
+    def _thread_records_locked(self, thread_id: str) -> list[RunRecord]:
+        """Return live in-memory records for *thread_id*. Caller must hold ``self._lock``.
+
+        Uses the ``_runs_by_thread`` index for O(runs-in-thread) lookup instead of
+        scanning every in-memory run. The ``self._runs.get`` guard keeps the result
+        correct even if the index and ``_runs`` momentarily disagree.
+        """
+        run_ids = self._runs_by_thread.get(thread_id)
+        if not run_ids:
+            return []
+        return [record for run_id in run_ids if (record := self._runs.get(run_id)) is not None]
 
     @staticmethod
     def _store_put_payload(record: RunRecord, *, error: str | None = None) -> dict[str, Any]:
@@ -338,6 +367,7 @@ class RunManager:
         )
         async with self._lock:
             self._runs[run_id] = record
+            self._index_run_locked(record)
             persisted = False
             try:
                 await self._persist_new_run_to_store(record)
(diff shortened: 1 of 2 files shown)


fancyboi999 (commented):
Took a careful pass focused on index/`_runs` consistency and the locking, and ran it locally on `ebae7dd`.

The core invariant holds: every site that mutates `_runs` maintains the index under the same lock and with no `await` in between — `create` (insert at L370 + rollback unindex at L382), `create_or_reject` (insert L584 + rollback unindex L596), and `cleanup` (L679). The two callers that scan `_runs` but never insert into it (`reconcile_orphaned_inflight_runs`, `shutdown`) correctly stay on the existing full iteration and don't need the index. So there's no "inserted into `_runs` but invisible to per-thread queries" path — the one direction that would actually be a bug.

Two things I think are done right and worth calling out:
- The `dict`-as-ordered-set keeps `list_by_thread`'s tie-break identical to the old full-scan order (same per-thread creation order), so the sort stays stable on equal `created_at`. Easy to get subtly wrong with a plain `set`.
- `_unindex_run_locked` pruning the empty bucket keeps `_runs_by_thread` bounded by live `_runs` instead of leaking a key per thread ever seen.

Local verification:
- `uv run pytest tests/test_run_manager.py -q` -> 51 passed. The 6 new index tests also pass in isolation (`-k "thread_index or has_inflight_reflects or create_or_reject_inflight_is_thread or failed_create_unindexes"` -> 6 passed).
- `ruff check` + `ruff format --check` -> clean. (`ruff.toml` sets `line-length = 240`, so the long assertion line flagged below is within limits for this repo.)

On the multi-worker angle (#3239 / #3475 making `--workers 1` the default): this is a per-process optimization. `_runs` and the index live in one RunManager instance, so the O(n) being removed is over a single worker's in-memory runs. It's orthogonal to the cross-worker shared-state story — neither helps nor regresses it. Just noting the scope is intentional.

Two non-blocking notes inline. Nothing here blocks from my side; clean, surgical change.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -118,10 +118,39 @@ def __init__(
...
+        """Return live in-memory records for *thread_id*. Caller must hold ``self._lock``.
+
+        Uses the ``_runs_by_thread`` index for O(runs-in-thread) lookup instead of
+        scanning every in-memory run. The ``self._runs.get`` guard keeps the result
+        correct even if the index and ``_runs`` momentarily disagree.
+        """
+        run_ids = self._runs_by_thread.get(thread_id)
+        if not run_ids:
+            return []
+        return [record for run_id in run_ids if (record := self._runs.get(run_id)) is not None]
```
fancyboi999, line 158: Building on the existing note about this guard — the asymmetry matters and is worth pinning down precisely:

`self._runs.get(...)` only covers one direction of divergence. A stale id present in the index but already gone from `_runs` gets filtered to `None`. The inverse — a run that is in `_runs` for this thread but missing from the index bucket — is silently dropped from the result, and the guard does nothing for it.

The reassuring part: under the current code neither can happen at a suspension point. Every `_runs`/index mutation is paired under `self._lock` with no `await` between them, so any coroutine that takes the lock sees both structures in agreement. That makes this guard defense-in-depth against a future refactor breaking the lockstep invariant — and it only catches half of what such a break could produce. I'd reword the docstring to say that explicitly (lockstep under the lock; the `.get` filter is defensive, not a reconciliation) rather than implying it actively keeps results correct under real disagreement. Cheap to keep the `.get`; just don't let the comment oversell it.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -562,6 +593,7 @@ async def create_or_reject(
                 # Also covers cancellation, which bypasses ``except Exception``.
                 if not persisted:
                     self._runs.pop(run_id, None)
+                    self._unindex_run_locked(run_id, record.thread_id)
```
fancyboi999, line 602: `test_failed_create_unindexes_run` exercises the `create()` rollback path nicely, but this symmetric rollback in `create_or_reject` isn't covered by any test. Same shape, but since the whole point of the PR is index/`_runs` lockstep across mutation sites, a parallel test that forces a persist failure here (e.g. reuse `_FailingPutRunStore` against `create_or_reject`) would close the last untested mutation path. Non-blocking.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -118,10 +118,39 @@ def __init__(
...
+        """Register *record* in the thread index. Caller must hold ``self._lock``."""
+        self._runs_by_thread.setdefault(record.thread_id, {})[record.run_id] = None
+
+    def _unindex_run_locked(self, run_id: str, thread_id: str) -> None:
+        """Drop *run_id* from the thread index. Caller must hold ``self._lock``."""
+        bucket = self._runs_by_thread.get(thread_id)
+        if bucket is not None:
+            bucket.pop(run_id, None)
+            if not bucket:
+                self._runs_by_thread.pop(thread_id, None)
```
fancyboi999, line 140: Nice — pruning the bucket when it empties is the right call. Without it, `_runs_by_thread` would retain a key for every thread ever seen even after all its runs are cleaned up, slowly leaking memory under churn. Good that `test_thread_index_cleanup_prunes_run_and_empty_bucket` locks this in.

WillemJiang:
@ly-wang19  thanks for your contriubtion, could you spend some time to take a look at the review comments?

ly-wang19:
Thanks for the careful pass and the local run on `ebae7dd` — much appreciated.

Addressed the inline notes in `972ed99`:

- **`_thread_records_locked` docstring** — reworded so it no longer oversells the `.get` filter. It now states the real guarantee is lockstep mutation under `self._lock` (no `await` between the two writes), and that `self._runs.get` is one-directional defense-in-depth: it drops a stale id still in the index but already gone from `_runs`, and **cannot** recover the inverse (a run in `_runs` but missing from the index — that would be silently omitted). Framed as a guard against a future refactor breaking lockstep, not as active reconciliation.
- **`create_or_reject` rollback coverage** — added `test_failed_create_or_reject_unindexes_run`, symmetric to the `create()` one: reuses `_FailingPutRunStore` to force a persist failure inside `create_or_reject` and asserts neither `_runs` nor the index retains the rolled-back run. Closes the last untested mutation path you flagged.
- **`_FailingPutRunStore` docstring** — "initial put" → "every put" to match behavior.

Left the long-assertion line as-is per your note (`ruff.toml` sets `line-length = 240`).

`tests/test_run_manager.py` → 52 passed; `ruff check` + `ruff format --check` clean.

WillemJiang (approved):

PR #3503, opened 2026-06-11
feat(skill): add blocking-io-guard — SOP skill for blocking-IO triage and runtime anchors

ShenAC-SAC:
## Summary

Adds a developer-facing `blocking-io-guard` skill (`.agent/skills/blocking-io-guard/`) that turns the existing blocking-IO tooling (static detector + Blockbuster runtime gate) into a deterministic, repeatable SOP:

- **Diff-scoped scanner**: `backend/tests/support/detectors/blocking_io_changed.py` + `scripts/scan_changed_blocking_io.py` — reports the blocking-IO candidates *this* change introduces (Mode A), as the union of (a) findings on the added lines of `git diff <base>...HEAD` and (b) findings new versus the merge base, matched by the stable key (path, function, symbol). The base comparison catches exposure created without touching the blocking line itself — e.g. a new async caller making an old sync helper async-reachable; the changed-line selection covers the same-key collision case (a second identical symbol added to an already-flagged function). Mode B (triage round) reuses `make detect-blocking-io` unchanged.
- **Router SOP** (`SKILL.md`): scope → judge (GUARD / FIX+ANCHOR / NO-ACTION) → fix + re-scan disappearance (matched by path/function/symbol, never line numbers) → anchor in `tests/blocking_io/` → mandatory teeth check (reintroduce the block → gate must go RED; restore → GREEN) → deliver. The teeth check doubles as the anchor-vs-rule discriminator; a new Blockbuster rule has strict admission criteria and always ships alone. An empty Mode A result is documented as "nothing the static detector can see", with the residual same-file-only blind spot (cross-file async callers) called out explicitly instead of silently treated as zero exposure — and the scanner now prints that caveat itself on every empty report, so the guidance reaches contributors who never read the SOP.
- **PR batching policy**: one fix unit per PR while any HIGH remains; MEDIUM/LOW batched (~5) once HIGHs are clear.
- References: a generic detect→judge→generate→verify skeleton (future detector domains), good-anchor rules distilled from `backend/docs/BLOCKING_IO_DETECTION.md`, and an anchor template following the `tests/blocking_io/` conventions.
- Docs: pointers added to `backend/docs/BLOCKING_IO_DETECTION.md` and `backend/CLAUDE.md`.

CI is unchanged — the skill produces anchors that the existing `backend-blocking-io` gate consumes.

## Validation

- Scanner built TDD: 8 unit tests in `backend/tests/test_scan_changed_blocking_io.py`; integration tests stub only the git boundary and run the real static scan. One test pins the false-negative scenario directly: the blocking line is not in the diff, only the new async caller is, and the finding must still be reported.
- Review round 2 (thanks @willem-bd): union dedup now matches by the stable key instead of dict identity; `parse_changed_lines` handles any unified diff (context lines advance the counter, per-file reset); base-version sources are scanned in memory via a new `blocking_io_static.scan_source` instead of a temp-file round-trip; empty reports carry the blind-spot caveat. Declined as out of scope for this PR: batching the per-file `git show` calls (cold-path dev tooling, per the review's own note), deduplicating the three `scripts/` CLI shims, and replacing the pre-existing `parents[4]` repo-root pattern (cross-cutting refactor across all detectors) — the latter two are tracked in #3510 as a follow-up cleanup once this PR lands.
- Real-git end-to-end: CLI run against an older base (57 changed files, 48 base versions materialized via `git show <merge-base>:<path>`, 53 base findings) exercises the full merge-base comparison path.
- `cd backend && make test-blocking-io` green; `make lint` green.
- **End-to-end acceptance**: the skill was executed for real on a triage round and produced companion fix PR #3505 — full SOP evidence (scan → fix → re-scan → anchor → red/green teeth) is in that PR's description.

## Test Plan
- [x] `backend/tests/test_scan_changed_blocking_io.py` (8 passed; 26 passed together with the static-detector suite)
- [x] `make test-blocking-io` (13 passed incl. companion anchor)
- [x] CLI smoke: text + JSON modes, empty and non-empty cases, real merge-base comparison

### DIFF
--- a/backend/tests/support/detectors/blocking_io_changed.py
+++ b/backend/tests/support/detectors/blocking_io_changed.py
@@ -16,7 +16,6 @@ import json
 import re
 import subprocess
 import sys
-import tempfile
 from collections import defaultdict
 from collections.abc import Sequence
 from pathlib import Path
@@ -36,10 +35,11 @@ _HUNK_RE = re.compile(r"^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@")
 def parse_changed_lines(diff_text: str) -> dict[str, set[int]]:
     """Map repo-relative path -> set of added line numbers in the new file.
 
-    Expects `git diff --unified=0` output. Records only added lines (`+`, not
-    the `+++` header), numbered from each hunk's new-file start line. Deletions
-    (`-`) do not advance the new-file counter; deleted files (`+++ /dev/null`)
-    are skipped.
+    Accepts any unified diff (with or without `--unified=0`): context lines
+    advance the new-file counter, deletions (`-`) and `\\ No newline` markers
+    do not. Records only added lines (`+`, not the `+++` header), numbered
+    from each hunk's new-file start line; deleted files (`+++ /dev/null`) are
+    skipped.
     """
     changed: dict[str, set[int]] = defaultdict(set)
     current_path: str | None = None
@@ -51,13 +51,18 @@ def parse_changed_lines(diff_text: str) -> dict[str, set[int]]:
                 current_path = None
             else:
                 current_path = target[2:] if target.startswith("b/") else target
+            next_line = 0
             continue
         match = _HUNK_RE.match(raw)
         if match:
             next_line = int(match.group(1))
             continue
-        if current_path and raw.startswith("+") and not raw.startswith("+++"):
+        if not current_path:
+            continue
+        if raw.startswith("+"):
             changed[current_path].add(next_line)
+            next_line += 1
+        elif raw.startswith(" ") or raw == "":
             next_line += 1
     return dict(changed)
 
@@ -120,17 +125,10 @@ def base_python_contents(base: str, paths: Sequence[str], repo_root: Path = REPO
 
 def scan_python_contents(contents: dict[str, str]) -> list[dict[str, object]]:
     """Run the static detector over in-memory sources (repo-relative path -> code)."""
-    if not contents:
-        return []
-    with tempfile.TemporaryDirectory() as td:
-        root = Path(td)
-        files = []
-        for rel_path, source in contents.items():
-            target = root / rel_path
-            target.parent.mkdir(parents=True, exist_ok=True)
-            target.write_text(source, encoding="utf-8")
-            files.append(target)
-        return [finding.to_dict() for finding in static.scan_paths(files, repo_root=root)]
+    findings: list[dict[str, object]] = []
+    for rel_path in sorted(contents):
+        findings.extend(finding.to_dict() for finding in static.scan_source(contents[rel_path], rel_path))
+    return findings
 
 
 def _stable_key(finding: dict[str, object]) -> tuple[str, str, str]:
@@ -170,13 +168,18 @@ def find_changed_blocking_io(base: str, repo_root: Path = REPO_ROOT) -> list[dic
     on_changed_lines = select_findings_on_changed_lines(head_findings, changed_lines)
     base_findings = scan_python_contents(base_python_contents(base, sorted(changed_lines), repo_root))
     new_vs_base = select_findings_new_vs_base(head_findings, base_findings)
-    selected = {id(finding) for finding in on_changed_lines} | {id(finding) for finding in new_vs_base}
-    return [finding for finding in head_findings if id(finding) in selected]
+    selected_keys = {_stable_key(finding) for finding in (*on_changed_lines, *new_vs_base)}
+    return [finding for finding in head_findings if _stable_key(finding) in selected_keys]
 
 
 def format_report(findings: Sequence[dict[str, object]], base: str) -> str:
     if not findings:
-        return f"No blocking-IO candidates introduced by this change (base: {base})."
+        return (
+            f"No blocking-IO candidates introduced by this change (base: {base}).\n"
+            "Note: async reachability is resolved within each file only. If this change\n"
+            "adds an async call into a sync helper defined in another file, check that\n"
+            "helper manually (codegraph or git grep) before relying on this empty result."
+        )
     lines = [
         f"Blocking-IO candidates introduced/touched by this change (base: {base}): {len(findings)}",
         "",
--- a/backend/tests/support/detectors/blocking_io_changed.py
+++ b/backend/tests/support/detectors/blocking_io_changed.py
@@ -110,11 +110,7 @@ def format_report(findings: Sequence[dict[str, object]], base: str) -> str:
     for finding in sorted(findings, key=lambda f: order.get(str(f["priority"]), 9)):
         location = finding["location"]  # type: ignore[index]
         call = finding["blocking_call"]  # type: ignore[index]
-        lines.append(
-            f"{finding['priority']} {call['category']}/{call['operation']} "
-            f"{location['path']}:{location['line']} in {location['function']} "
-            f"exposure={finding['event_loop_exposure']}"
-        )
+        lines.append(f"{finding['priority']} {call['category']}/{call['operation']} {location['path']}:{location['line']} in {location['function']} exposure={finding['event_loop_exposure']}")
         lines.append(f"  symbol: {call['symbol']}")
         if finding.get("code"):
             lines.append(f"  code: {finding['code']}")
@@ -122,9 +118,7 @@ def format_report(findings: Sequence[dict[str, object]], base: str) -> str:
 
 
 def main(argv: Sequence[str] | None = None) -> int:
-    parser = argparse.ArgumentParser(
-        description="List blocking-IO candidates on this change's added lines (diff against --base)."
-    )
+    parser = argparse.ArgumentParser(description="List blocking-IO candidates on this change's added lines (diff against --base).")
     parser.add_argument("--base", default="origin/main", help="Base ref to diff against (default: origin/main).")
     parser.add_argument("--format", choices=("text", "json"), default="text", help="Output format.")
     args = parser.parse_args(argv)
--- a/backend/tests/support/detectors/blocking_io_changed.py
+++ b/backend/tests/support/detectors/blocking_io_changed.py
@@ -1,0 +1,141 @@
+"""Intersect a git diff with static blocking-IO findings.
+
+Wraps the static detector (`blocking_io_static`) to answer a narrower question:
+which blocking-IO candidates does THIS change introduce or touch on its added
+lines? Used by the `blocking-io-guard` skill as the deterministic L1 scope step.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import subprocess
+import sys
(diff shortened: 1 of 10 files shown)


willem-bd (commented):
## Code Review: blocking-io-guard skill

8 findings from a high-effort review across correctness, efficiency, altitude, reuse, and robustness angles.

**Top findings:**
1. 🔴 `id()`-based dedup in `find_changed_blocking_io` is a maintenance landmine — would silently return empty results if selectors ever return copied dicts (line 173)
2. 🟡 `parse_changed_lines` silently misbehaves for non-unified=0 diffs — latent bug in public API (line 47)
3. 🟡 N sequential `git show` subprocesses grows linearly with PR size (line 110)
4. 🔵 Temp-file workaround for Path-only API — underlying visitor already accepts in-memory content (line 125)
5. 🔵 Third copy of identical script boilerplate (triplicates maintenance burden)
6. 🔵 Cross-file async reachability blind spot emits no warning on empty results (line 26)
7. 🔵 `parents[4]` REPO_ROOT perpetuates fragile pattern (line 26, low priority — existing)
8. 🔵 `next_line` not reset across file boundaries (line 48, minor robustness)

Overall: well-structured PR with good test coverage. The id()-dedup and parse_changed_lines issues are worth addressing before merge; the rest are improvement opportunities.

backend/tests/support/detectors/blocking_io_changed.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    changed_lines = changed_python_lines(base, repo_root)
+    if not changed_lines:
+        return []
+    files = [repo_root / path for path in changed_lines]
+    head_findings = [finding.to_dict() for finding in static.scan_paths(files, repo_root=repo_root)]
+    on_changed_lines = select_findings_on_changed_lines(head_findings, changed_lines)
+    base_findings = scan_python_contents(base_python_contents(base, sorted(changed_lines), repo_root))
+    new_vs_base = select_findings_new_vs_base(head_findings, base_findings)
+    selected = {id(finding) for finding in on_changed_lines} | {id(finding) for finding in new_vs_base}
```
willem-bd: **🔴 `id()`-based dedup is fragile.** `find_changed_blocking_io` uses `id()` of dict objects for set-union dedup instead of the logical `_stable_key`. This works today because both selectors return references from `head_findings`, but if either is ever refactored to return copied dicts (e.g., `dict(finding)` for isolation), the `id()` set no longer matches objects in `head_findings` and the final list comprehension silently returns an empty list.

**Suggestion:** Replace `id()`-based dedup with `_stable_key`-based dedup:
```python
selected_keys = {_stable_key(f) for f in on_changed_lines} | {_stable_key(f) for f in new_vs_base}
return [f for f in head_findings if _stable_key(f) in selected_keys]
```
This is robust to future refactoring and reuses the same logical identity already defined in the module.

backend/tests/support/detectors/blocking_io_changed.py
```diff
@@ -0,0 +1,210 @@
...
+
+    Expects `git diff --unified=0` output. Records only added lines (`+`, not
+    the `+++` header), numbered from each hunk's new-file start line. Deletions
+    (`-`) do not advance the new-file counter; deleted files (`+++ /dev/null`)
+    are skipped.
+    """
+    changed: dict[str, set[int]] = defaultdict(set)
+    current_path: str | None = None
+    next_line = 0
+    for raw in diff_text.splitlines():
```
willem-bd, line 47: **🟡 `parse_changed_lines` silently misbehaves for non-unified=0 diffs.** The function is public but never advances `next_line` for context lines (space-prefixed) or deletion lines. It works only because `changed_python_lines` hardcodes `--unified=0`. If any future caller passes a standard diff, every `+` line after a context line gets an undercounted line number.

**Suggestion:** Either (a) make the function private (`_parse_changed_lines`) and document the unified=0 requirement more prominently, or (b) advance `next_line` for context lines too:
```python
if current_path:
    if raw.startswith("+") and not raw.startswith("+++"):
        changed[current_path].add(next_line)
        next_line += 1
    elif not raw.startswith("-"):
        next_line += 1  # advance for context lines
```
This makes the function correct for any unified diff, not just unified=0.

backend/tests/support/detectors/blocking_io_changed.py
```diff
@@ -0,0 +1,210 @@
...
+    finding in them counts as new.
+    """
+    merge_base = subprocess.run(
+        ["git", "-C", str(repo_root), "merge-base", base, "HEAD"],
+        capture_output=True,
+        text=True,
+        check=True,
+    ).stdout.strip()
+    contents: dict[str, str] = {}
+    for path in paths:
```
willem-bd, line 115: **🟡 N sequential `git show` subprocesses.** `base_python_contents` spawns one `git show` per changed file in a sequential loop. For a PR touching 50 files, that is 51 subprocess invocations (1 merge-base + 50 shows) adding ~0.5–1s of pure spawn latency before any AST analysis.

**Suggestion:** Batch with `git cat-file --batch` to reduce N+1 subprocess calls to 2, regardless of file count. Alternatively, since this is a developer tooling script (not a hot-path API), the current approach is acceptable for typical PR sizes but worth noting for future optimization if it becomes a bottleneck.

backend/tests/support/detectors/blocking_io_changed.py
```diff
@@ -0,0 +1,210 @@
...
+        if shown.returncode == 0:
+            contents[path] = shown.stdout
+    return contents
+
+
+def scan_python_contents(contents: dict[str, str]) -> list[dict[str, object]]:
+    """Run the static detector over in-memory sources (repo-relative path -> code)."""
+    if not contents:
+        return []
+    with tempfile.TemporaryDirectory() as td:
```
willem-bd: **🔵 Temp-file workaround for Path-only API.** `scan_python_contents` writes in-memory base-version source to temporary files on disk purely because `scan_paths` only accepts `Path` objects. The underlying `BlockingIOStaticVisitor` already accepts `source_lines` (a list of strings), but there's no public entry point that takes in-memory content. For a PR touching 30 files, this creates 30 temp files only for `scan_paths` to read them back immediately.

**Suggestion:** Add a `scan_source(source: str, relative_path: str, repo_root: Path)` function to `blocking_io_static.py` that wraps the visitor without going through `path.read_text()`. This eliminates the temp-file round-trip and also benefits unit testing.

scripts/scan_changed_blocking_io.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
+"""CLI wrapper for the changed-lines blocking IO scanner."""
+
+from __future__ import annotations
+
+import sys
+from collections.abc import Sequence
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
```
willem-bd, line 10: **🔵 Third copy of identical script boilerplate.** This script is character-for-character identical to `scripts/detect_blocking_io_static.py` and `scripts/detect_thread_boundaries.py`, differing only in the docstring and the deferred import target. If the `sys.path` computation ever changes (e.g., `backend/tests` moves during a refactor), all three must be updated independently.

**Suggestion:** Extract the shared boilerplate into a `scripts/_detector_cli.py` helper, or add a `--mode changed` subcommand to the existing `detect_blocking_io_static.py` script.

backend/tests/support/detectors/blocking_io_changed.py
```diff
@@ -0,0 +1,210 @@
...
+    on_changed_lines = select_findings_on_changed_lines(head_findings, changed_lines)
+    base_findings = scan_python_contents(base_python_contents(base, sorted(changed_lines), repo_root))
+    new_vs_base = select_findings_new_vs_base(head_findings, base_findings)
+    selected = {id(finding) for finding in on_changed_lines} | {id(finding) for finding in new_vs_base}
+    return [finding for finding in head_findings if id(finding) in selected]
+
+
+def format_report(findings: Sequence[dict[str, object]], base: str) -> str:
+    if not findings:
+        return f"No blocking-IO candidates introduced by this change (base: {base})."
```
willem-bd: **🔵 Cross-file async reachability blind spot with no warning.** The static detector resolves async reachability same-file only (`ASYNC_REACHABLE_SAME_FILE`). If this change adds an async caller of a sync blocking helper in **another file**, the scanner reports zero findings and gives false confidence. The SOP documents this limitation, but the scanner itself emits no warning when it might be affected.

**Suggestion:** When Mode A returns empty, emit an explicit caveat in the report output (i.e., in `format_report` when `findings` is empty):
```
Note: The scanner only checks async reachability within the same file. If your diff
adds an async call into a sync helper defined in another file, check that helper
manually (codegraph or `git grep`) before relying on this empty result.
```
This surfaces the SOP guidance at the point of use rather than requiring the contributor to have read the SOP.

backend/tests/support/detectors/blocking_io_changed.py
```diff
@@ -0,0 +1,210 @@
...
+import subprocess
+import sys
+import tempfile
+from collections import defaultdict
+from collections.abc import Sequence
+from pathlib import Path
+
+from support.detectors import blocking_io_static as static
+
+REPO_ROOT = Path(__file__).resolve().parents[4]
```
willem-bd, line 25: **🔵 `parents[4]` depth-indexed REPO_ROOT perpetuates fragile pattern.** `REPO_ROOT = Path(__file__).resolve().parents[4]` is the same fragile pattern as `blocking_io_static.py`. It breaks silently when the file is moved to a different directory depth — `scan_paths` would scan nothing and return zero findings with no error.

This is a pre-existing pattern that the PR reproduces rather than introduces, so low priority. A more robust approach would walk upward looking for a repo-root marker (`.git`, `pyproject.toml`), but that is a cross-cutting refactor best done separately for all detectors at once.

backend/tests/support/detectors/blocking_io_changed.py
```diff
@@ -0,0 +1,210 @@
...
+    Expects `git diff --unified=0` output. Records only added lines (`+`, not
+    the `+++` header), numbered from each hunk's new-file start line. Deletions
+    (`-`) do not advance the new-file counter; deleted files (`+++ /dev/null`)
+    are skipped.
+    """
+    changed: dict[str, set[int]] = defaultdict(set)
+    current_path: str | None = None
+    next_line = 0
+    for raw in diff_text.splitlines():
+        if raw.startswith("+++ "):
```
willem-bd, line 48: **🔵 `next_line` not reset across file boundaries.** After processing `+++ b/file_b.py`, `next_line` retains its value from the previous file's last hunk. In well-formed git output a `@@` hunk header always follows `+++`, so this is not a practical issue today. However, adding `next_line = 0` in the `+++` branch would be a defensive safeguard against malformed input with no downside.

WillemJiang (approved):

PR #3504, opened 2026-06-11
fix(skills): offload blocking filesystem IO in async skill installation

ly-wang19:
## Why

`LocalSkillStorage.ainstall_skill_from_archive` is an async method (the `POST /api/skills/install` path), but it ran its entire filesystem workload directly on the event loop:

- archive probing (`is_file` / `exists`), `mkdir`, `tempfile.TemporaryDirectory`
- zip open + `safe_extract_skill_archive` (extract every member)
- `shutil.copytree` of the staged skill into `custom/`

So a skill install blocks the event loop for its full duration — flagged by `make detect-blocking-io`. Same class of issue DeerFlow already guards against in `tests/blocking_io/` (e.g. #1917, the skill *load* path).

## What changed

Split the blocking work into two sync helpers dispatched via `asyncio.to_thread`:

- `_extract_and_validate` — probe / mkdir / extract / frontmatter-validate. Returns the extraction tempdir (created with `mkdtemp`, not auto-cleaned) so it survives the **awaited** security scan that runs between the two phases.
- `_stage_and_commit` — `copytree` + atomic move.

The extraction tempdir is removed in a `finally` (also offloaded). Externally observable behavior is unchanged: same `FileNotFoundError` / `ValueError` / `SkillAlreadyExistsError` paths, same cleanup-on-failure, same result dict.

> Note: the security scanner `_scan_skill_archive_contents_or_raise` has its own `rglob` enumeration on the event loop — a separate, smaller blocking spot left for a focused follow-up (it's security-sensitive and best changed on its own).

## Surface area

- [x] **Backend** — `packages/harness/deerflow/skills/storage/local_skill_storage.py` + tests. No API / schema change.

## Validation

- New `tests/blocking_io/test_skill_install.py` anchor under the strict Blockbuster gate (verified **red→green**: fails without the offload), plus a tempdir-cleanup-on-failure regression test → **2 passed**
- Full `tests/blocking_io` suite → **8 passed**
- Skill install/storage regression (`test_skills_installer`, `test_local_skill_storage_write`, `test_skills_custom_router`, `test_skill_manage_tool`) → **65 passed**
- `ruff check` + `ruff format` → clean

## AI assistance

**How you used it:** AI located the finding via `make detect-blocking-io`, implemented the two-phase `asyncio.to_thread` offload (handling the tempdir lifetime across the awaited scan), and wrote the gate anchor + cleanup regression test. I reviewed the change.

### DIFF
--- a/backend/packages/harness/deerflow/skills/storage/local_skill_storage.py
+++ b/backend/packages/harness/deerflow/skills/storage/local_skill_storage.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import errno
 import json
 import logging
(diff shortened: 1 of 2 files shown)


ly-wang19:
Thanks for the thorough review.

**On the scan:** you're right that `_scan_skill_archive_contents_or_raise` (`installer.py:156`/`182`/`183` — `read_text` / `rglob` / `is_file`) still runs on the loop between the two offloaded halves. I scoped that out of this PR deliberately — it's the security-sensitive path and I wanted the extract/copytree offload to land cleanly first — and called it out in the PR body. But folding the enumeration + per-file reads into a worker thread (read bytes off-loop, then `await` the content scan) is clearly the right completion; `make detect-blocking-io` won't be clean on this install path until that's done.

**On #3505:** thanks for the heads-up. It overlaps this PR and additionally offloads exactly that scan, so it's the more complete fix. I don't want to duplicate work or race a parallel contributor's PR. I'm happy to go either way:

1. **Defer to #3505** — close this once it lands, or
2. **Extend this PR** to also offload the `installer.py` scan (matching #3505's coverage), if you'd rather land this one.

Whichever is less work to consolidate on your side — just say the word. 🙏

ShenAC-SAC:
@ly-wang19 Thanks for your contribution. If you are interested in issues like blocking-io, you can consider submitting a PR using #3503 Skills. There are still some unresolved issues.

ly-wang19:
Thanks @ShenAC-SAC, and congrats on #3505 landing 🎉 — it's the more complete fix (it also offloads the `installer.py` scan I'd deliberately scoped out of this PR), so as I offered above I'm closing this one in its favor to keep things consolidated.

Appreciate the pointer to the #3503 `blocking-io-guard` skill for the remaining cases — I'll use it to triage the other unresolved blocking-IO spots. Thanks both for the reviews here.

PR #3505, opened 2026-06-11
fix(skills): keep skill archive installation off the event loop

ShenAC-SAC:
## Summary

`LocalSkillStorage.ainstall_skill_from_archive` — the async entry point awaited by the gateway `POST /api/skills/install` route — ran its entire filesystem pipeline inline on the event loop: zip extraction, frontmatter validation, `rglob` enumeration, per-file `read_text`, `shutil.copytree` staging, and tempdir lifecycle. During an install, every other request on the loop stalls behind this work.

The fix restructures the pipeline into offloaded phases while keeping the security decision logic and exception contract unchanged:

- `_prepare_skill_archive` (extract + validate + name checks) and `_commit_skill_install` (staging copytree + move) run via `asyncio.to_thread`; the temp directory is created/removed off-loop too.
- In `installer.py`, the security scanner's file enumeration (`rglob` + `is_file` filter) and each `read_text` are offloaded; only the per-file LLM scan (`scan_skill_content`, genuinely async) stays awaited on the loop.
- One deliberate deviation: temp-dir cleanup is now best-effort (`_cleanup_install_tmp`). The previous `TemporaryDirectory` context would raise on cleanup failure, which could mask a scan/commit exception raised inside the block; the helper swallows `OSError` to preserve the original exception, but logs a warning (review follow-up — the initial version used `ignore_errors=True`, which leaked the temp dir silently). Per @WillemJiang's review, the cleanup await is additionally bounded with `asyncio.wait_for` (5s) so a hung filesystem in the `finally` cannot stall the install outcome, and `_collect_scannable_files` takes `skill_dir` as an explicit argument instead of a closure capture.

## Process note

This PR was produced by executing the `blocking-io-guard` skill from #3503 end-to-end (Mode B triage round), as its acceptance run. SOP evidence:

- **Step 0 (scope)**: `make detect-blocking-io` on `2d5f0787` → 79 findings, 3 HIGH. Two HIGHs share one root cause (this install pipeline): `installer.py _scan_skill_archive_contents_or_raise | skill_dir.rglob` and `local_skill_storage.py ainstall_skill_from_archive | shutil.copytree` → one fix unit per the batching policy.
- **Step 2 (re-scan)**: after the fix, findings drop 79 → 71; all 8 install-path findings (2 HIGH + MEDIUM/LOW neighbours) disappear, matched by (path, function, symbol).
- **Step 5 (teeth)**: anchor `tests/blocking_io/test_skills_install.py` drives the real install pipeline (real `.skill` archive, real filesystem; only the LLM boundary `scan_skill_content` stubbed). Against pre-fix code the strict Blockbuster gate goes RED (`BlockingError: Blocking call to os.stat`); with the fix, GREEN.
- **Dispositions**: remaining HIGH (`channels/manager.py _ingest_inbound_files`) is a separate fix unit, queued for the next round. NO-ACTION: `LocalSkillStorage.__init__` path resolution (one-time, cached via `get_or_new_skill_storage`; sits in the static detector's cross-file blind spot) — noted, not fixed here.

## Test Plan
- [x] `make test-blocking-io` → 13 passed (12 existing + new anchor); teeth verified red→green
- [x] Existing behavior suites: installer/router/storage/validation (84 passed), client install path (15 passed)
- [x] Full backend suite: 4208 passed, 18 skipped
- [x] `make lint` / `make format` clean

### DIFF
--- a/backend/packages/harness/deerflow/skills/installer.py
+++ b/backend/packages/harness/deerflow/skills/installer.py
@@ -153,7 +153,7 @@ async def _scan_skill_file_or_raise(skill_dir: Path, path: Path, skill_name: str
     rel_path = path.relative_to(skill_dir).as_posix()
     location = f"{skill_name}/{rel_path}"
     try:
-        content = path.read_text(encoding="utf-8")
+        content = await asyncio.to_thread(path.read_text, encoding="utf-8")
     except UnicodeDecodeError as e:
         raise SkillSecurityScanError(f"Security scan failed for skill '{skill_name}': {location} must be valid UTF-8") from e
 
@@ -179,10 +179,10 @@ async def _scan_skill_archive_contents_or_raise(skill_dir: Path, skill_name: str
     skill_md = skill_dir / "SKILL.md"
     await _scan_skill_file_or_raise(skill_dir, skill_md, skill_name, executable=False)
 
-    for path in sorted(skill_dir.rglob("*")):
-        if not path.is_file():
-            continue
+    def _collect_scannable_files() -> list[Path]:
+        return [candidate for candidate in sorted(skill_dir.rglob("*")) if candidate.is_file()]
 
+    for path in await asyncio.to_thread(_collect_scannable_files):
         rel_path = path.relative_to(skill_dir)
         if rel_path == Path("SKILL.md"):
             continue
--- a/backend/packages/harness/deerflow/skills/installer.py
+++ b/backend/packages/harness/deerflow/skills/installer.py
@@ -174,15 +174,17 @@ async def _scan_skill_file_or_raise(skill_dir: Path, path: Path, skill_name: str
         raise SkillSecurityScanError(f"Security scan failed for {location}: invalid scanner decision {decision!r}")
 
 
+def _collect_scannable_files(skill_dir: Path) -> list[Path]:
+    """Enumerate archive files for scanning (blocking; run off the event loop)."""
+    return [candidate for candidate in sorted(skill_dir.rglob("*")) if candidate.is_file()]
+
+
 async def _scan_skill_archive_contents_or_raise(skill_dir: Path, skill_name: str) -> None:
     """Run the skill security scanner against all installable text and script files."""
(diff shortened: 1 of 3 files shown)


WillemJiang:
@ShenAC-SAC, here are some comments for this PR.
1. _cleanup_install_tmp should arguably run even if earlier phases raise (local_skill_storage.py)

  The current structure is:
  ```
tmp = await asyncio.to_thread(tempfile.mkdtemp)
  try:
      skill_dir, skill_name, target = await asyncio.to_thread(self._prepare_skill_archive, ...)
      await _scan_skill_archive_contents_or_raise(skill_dir, skill_name)
      await asyncio.to_thread(self._commit_skill_install, ...)
  finally:
      await asyncio.to_thread(self._cleanup_install_tmp, tmp)
```

  The finally block correctly runs cleanup even on exception. ✓ However, if _cleanup_install_tmp itself hangs (e.g. NFS stall), the
  await in finally would prevent the exception from propagating. Consider a timeout:

```
  finally:
      try:
          await asyncio.wait_for(asyncio.to_thread(self._cleanup_install_tmp, tmp), timeout=5.0)
      except (asyncio.TimeoutError, Exception):
          logger.warning("Timed out cleaning up temp dir %s", tmp)
```

  This is a defensive enhancement, not a blocker — the current behavior is already better than the original TemporaryDirectory context manager.

  2. _collect_scannable_files closure captures skill_dir from enclosing scope (installer.py:182-183)

 ```
 def _collect_scannable_files() -> list[Path]:
      return [candidate for candidate in sorted(skill_dir.rglob("*")) if candidate.is_file()]
```

  This is a closure over skill_dir which is a Path argument to the enclosing function. Since Path is immutable and the function
  runs in a thread, this is safe. But making it an explicit argument would be slightly cleaner for readability and testability:

```
  def _collect_scannable_files(sd: Path = skill_dir) -> list[Path]:
      return [candidate for candidate in sorted(sd.rglob("*")) if candidate.is_file()]
```

WillemJiang (approved):

PR #3506, opened 2026-06-11
feat(auth): add generic OIDC/SSO authentication with Keycloak support

zhfeng:
## Why

DeerFlow currently supports only local email/password authentication. This PR adds generic OIDC SSO support, compatible with Keycloak, Google, Azure AD, Okta, and any OIDC 2.0 compliant provider. The user model already anticipated OAuth via `oauth_provider`/`oauth_id` fields — this implementation wires them up.

## What changed

- **OIDC service** (`backend/app/gateway/auth/oidc.py`): provider discovery, token exchange, JWKS-based ID token validation, userinfo fetching
- **Signed state cookies** (`backend/app/gateway/auth/oidc_state.py`): stateless multi-worker deployment via signed HttpOnly cookies for OIDC state/nonce/PKCE verifier
- **User provisioning** (`backend/app/gateway/auth/user_provisioning.py`): auto-create users, email domain restriction, verified-email requirement, optional email linking
- **Config models** (`backend/packages/harness/deerflow/config/auth_config.py`): per-provider settings for PKCE, nonce, provisioning, endpoint overrides
- **Backend endpoints** (`backend/app/gateway/routers/auth.py`): `GET /providers`, `GET /oauth/{provider}`, `GET /callback/{provider}`
- **Frontend login page**: fetches SSO providers, renders "Continue with {provider}" buttons
- **Frontend callback page** (`/auth/callback`): detects auth after SSO redirect, redirects to workspace
- **Account settings**: hides password change for SSO users, shows provider name, i18n support (en-US/zh-CN)
- **Docs**: `backend/docs/SSO.md` with setup guide and local Keycloak example, updated `AUTH_DESIGN.md`

## Surface area

- [x] **Frontend UI** — login page SSO buttons, auth callback page, account settings for SSO users
- [x] **Backend API** — OIDC endpoints, auth middleware public path updates
- [x] **Dependencies** — `pyjwt[crypto]` for RS256/ES256 ID token signature validation
- [x] **Default behavior change** — OIDC is opt-in (disabled by default), no change to existing auth
- [x] **Docs / tests / CI only** — SSO.md, updated AUTH_DESIGN.md, commented sections in config.example.yaml

## Validation

Will validate after merge:
- `cd backend && make lint`

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Implementation was pair-programmed through Claude Code CLI — design decisions (signed cookies over server-side state, PKCE+nonce defaults, provisioning flags) were made collaboratively.

### DIFF
--- a/backend/app/gateway/auth/oidc.py
+++ b/backend/app/gateway/auth/oidc.py
@@ -1,0 +1,433 @@
+"""OIDC (OpenID Connect) authentication service.
+
+Provides provider-agnostic OIDC operations: discovery, authorization URL
+generation, token exchange, ID token validation, and userinfo retrieval.
+"""
+
+from __future__ import annotations
+
+import logging
+import secrets
+import time
+from dataclasses import dataclass
+from typing import Any
+from urllib.parse import urlencode
+
+import httpx
+import jwt
+from jwt import PyJWK
+
+logger = logging.getLogger(__name__)
+
+# ── Data types ────────────────────────────────────────────────────────────
+
+OIDC_DISCOVERY_PATH = "/.well-known/openid-configuration"
+METADATA_CACHE_TTL = 300  # 5 minutes
+JWKS_CACHE_TTL = 300
+
+
+@dataclass(frozen=True)
+class OIDCMetadata:
+    """Resolved OIDC provider metadata after discovery."""
+
+    issuer: str
+    authorization_endpoint: str
+    token_endpoint: str
+    userinfo_endpoint: str | None
+    jwks_uri: str
+
+
+@dataclass(frozen=True)
+class OIDCIdentity:
+    """Normalized identity extracted from an OIDC provider response."""
+
+    provider: str
+    subject: str
+    email: str
+    email_verified: bool
+    name: str | None
+    claims: dict[str, Any]
+
+
+class OIDCError(Exception):
+    """Base error for OIDC operations. Message is safe for API responses."""
+
+
+class OIDCProviderError(OIDCError):
+    """The OIDC provider returned an error (e.g. access_denied)."""
+
+
+class OIDCValidationError(OIDCError):
+    """ID token validation failed."""
+
+
+class OIDCUserInfoMismatch(OIDCError):
+    """UserInfo sub does not match ID token sub."""
+
+
+# ── Service ────────────────────────────────────────────────────────────────
+
+
+class OIDCService:
+    """OIDC authentication service.
+
+    Uses in-process caching for provider metadata and JWKS. The cache is
+    keyed by the provider's ``issuer`` — different providers get separate
+    entries. TTLs are configurable via constructor arguments.
+    """
+
+    def __init__(
+        self,
+        metadata_cache_ttl: float = METADATA_CACHE_TTL,
+        jwks_cache_ttl: float = JWKS_CACHE_TTL,
+    ) -> None:
+        self._metadata_cache: dict[str, tuple[float, dict[str, Any]]] = {}
+        self._jwks_cache: dict[str, tuple[float, dict[str, Any]]] = {}
+        self._metadata_ttl = metadata_cache_ttl
+        self._jwks_ttl = jwks_cache_ttl
+        self._http = httpx.AsyncClient(timeout=httpx.Timeout(15.0))
+
+    async def close(self) -> None:
+        """Close the underlying HTTP client."""
+        await self._http.aclose()
+
+    # ── Discovery ──────────────────────────────────────────────────────────
+
+    async def discover(self, issuer: str, overrides: dict[str, str | None] | None = None) -> OIDCMetadata:
+        """Fetch and cache OIDC discovery metadata from the issuer.
+
+        ``overrides`` may contain endpoint URIs to override discovery values
+        (e.g. for providers with non-standard endpoints).
+        """
+        now = time.time()
+        cached = self._metadata_cache.get(issuer)
+        if cached and now - cached[0] < self._metadata_ttl:
+            return self._metadata_from_dict(cached[1], overrides)
+
+        discovery_url = issuer.rstrip("/") + OIDC_DISCOVERY_PATH
+        try:
+            resp = await self._http.get(discovery_url)
+            resp.raise_for_status()
+            data: dict[str, Any] = resp.json()
+        except httpx.HTTPStatusError as exc:
+            raise OIDCError(f"OIDC discovery failed for issuer {issuer}: HTTP {exc.response.status_code}") from exc
+        except httpx.RequestError as exc:
+            raise OIDCError(f"OIDC discovery failed for issuer {issuer}: {exc}") from exc
+
+        discovered_issuer = data.get("issuer")
+        if not discovered_issuer:
+            raise OIDCError(f"OIDC discovery response from {issuer} is missing the issuer field")
+
+        # RFC 8414 §4: the metadata issuer must equal the configured issuer.
+        # Pinning it prevents a tampered/rogue discovery document from steering
+        # the accepted `iss` (and thus the ID-token forgery surface) to an
+        # attacker-chosen value.
+        if discovered_issuer.rstrip("/") != issuer.rstrip("/"):
+            raise OIDCError(f"OIDC discovered issuer '{discovered_issuer}' does not match configured issuer '{issuer}'")
+
+        self._metadata_cache[issuer] = (now, data)
+        return self._metadata_from_dict(data, overrides)
+
+    def _metadata_from_dict(self, data: dict[str, Any], overrides: dict[str, str | None] | None) -> OIDCMetadata:
+        """Build OIDCMetadata from a discovery dict, applying endpoint overrides."""
+        overrides = overrides or {}
+        return OIDCMetadata(
+            issuer=data["issuer"],
+            authorization_endpoint=overrides.get("authorization_endpoint") or data["authorization_endpoint"],
+            token_endpoint=overrides.get("token_endpoint") or data["token_endpoint"],
+            userinfo_endpoint=overrides.get("userinfo_endpoint") or data.get("userinfo_endpoint"),
+            jwks_uri=overrides.get("jwks_uri") or data["jwks_uri"],
+        )
+
+    # ── Authorization URL ──────────────────────────────────────────────────
+
+    def build_authorization_url(
+        self,
+        metadata: OIDCMetadata,
+        client_id: str,
+        redirect_uri: str,
+        scopes: list[str],
+        state: str,
+        nonce: str | None = None,
+        code_challenge: str | None = None,
+    ) -> str:
+        """Build the OIDC authorization URL for the provider.
+
+        Returns a URL the browser should be redirected to.
+        """
+        params: dict[str, str] = {
+            "response_type": "code",
+            "client_id": client_id,
+            "redirect_uri": redirect_uri,
+            "scope": " ".join(scopes),
+            "state": state,
+        }
+        if nonce:
+            params["nonce"] = nonce
+        if code_challenge:
+            params["code_challenge"] = code_challenge
+            params["code_challenge_method"] = "S256"
+
+        return f"{metadata.authorization_endpoint}?{urlencode(params)}"
+
+    # ── Token exchange ─────────────────────────────────────────────────────
+
+    async def exchange_code(
+        self,
+        metadata: OIDCMetadata,
+        client_id: str,
+        client_secret: str | None,
+        code: str,
+        redirect_uri: str,
+        code_verifier: str | None = None,
+        auth_method: str = "client_secret_post",
+    ) -> dict[str, Any]:
+        """Exchange the authorization code for tokens at the token endpoint."""
+        data: dict[str, str] = {
+            "grant_type": "authorization_code",
+            "code": code,
+            "redirect_uri": redirect_uri,
+            "client_id": client_id,
+        }
+        if code_verifier:
+            data["code_verifier"] = code_verifier
+
+        headers: dict[str, str] = {"Accept": "application/json"}
+
+        if auth_method == "client_secret_basic" and client_secret:
+            import base64
+
+            creds = base64.b64encode(f"{client_id}:{client_secret}".encode()).decode("ascii")
+            headers["Authorization"] = f"Basic {creds}"
+        elif auth_method == "client_secret_post" and client_secret:
+            data["client_secret"] = client_secret
+
+        try:
+            resp = await self._http.post(metadata.token_endpoint, data=data, headers=headers)
+            resp.raise_for_status()
+            return resp.json()
+        except httpx.HTTPStatusError as exc:
+            body = "unknown"
+            try:
+                body = exc.response.text[:200]
+            except Exception:
+                pass
+            raise OIDCError(f"Token exchange failed: HTTP {exc.response.status_code} — {body}") from exc
+        except httpx.RequestError as exc:
+            raise OIDCError(f"Token exchange failed: {exc}") from exc
+
+    # ── JWKS loading ───────────────────────────────────────────────────────
+
+    async def _load_jwks(self, jwks_uri: str, force_refresh: bool = False) -> dict[str, Any]:
+        """Load (and cache) JWKS from the provider.
+
+        Set ``force_refresh=True`` to bypass the cache (e.g. on a kid miss).
+        """
+        now = time.time()
+        cached = self._jwks_cache.get(jwks_uri)
+        if not force_refresh and cached and now - cached[0] < self._jwks_ttl:
+            return cached[1]
+
+        try:
+            resp = await self._http.get(jwks_uri)
+            resp.raise_for_status()
+            data: dict[str, Any] = resp.json()
+        except httpx.HTTPStatusError as exc:
+            raise OIDCError(f"JWKS fetch failed: HTTP {exc.response.status_code}") from exc
+        except httpx.RequestError as exc:
+            raise OIDCError(f"JWKS fetch failed: {exc}") from exc
+
+        self._jwks_cache[jwks_uri] = (now, data)
+        return data
+
+    async def _resolve_signing_key(
+        self,
+        jwks_data: dict[str, Any],
+        kid: str | None,
+        algorithm: str,
+        jwks_uri: str,
+    ) -> Any | None:
+        """Find the signing key matching ``kid`` in the JWKS.
+
+        Returns the key object or ``None`` if no match is found. Catches
+        invalid JWK entries (e.g. wrong key type for the algorithm) and
+        logs a warning so a single bad entry does not crash validation.
+        """
+        for jwk_dict in jwks_data.get("keys", []):
+            if kid and jwk_dict.get("kid") != kid:
+                continue
+            try:
+                jwk = PyJWK(jwk_dict, algorithm=algorithm)
+                return jwk.key
+            except jwt.PyJWTError as exc:
+                logger.warning("Skipping invalid JWK (kid=%s) from %s: %s", kid, jwks_uri, exc)
+                if not kid:
+                    # No kid in token — try next key
+                    continue
+                # kid was specified and this key is the one — fail fast
+                raise OIDCValidationError(f"JWK for kid={kid} is invalid: {exc}") from exc
+        return None
+
+    # ── ID token validation ────────────────────────────────────────────────
+
+    async def validate_id_token(
+        self,
+        metadata: OIDCMetadata,
+        client_id: str,
+        id_token: str,
+        nonce: str | None = None,
+    ) -> dict[str, Any]:
+        """Validate the ID token and return its claims.
+
+        Validates: signature (via JWKS), issuer, audience, expiration,
+        issued-at, and nonce (if provided).
+        """
+        jwks_data = await self._load_jwks(metadata.jwks_uri)
+
+        # Resolve the signing key from the JWKS using the token's kid header
+        jwt_header = jwt.get_unverified_header(id_token)
+        kid = jwt_header.get("kid")
+        alg = jwt_header.get("alg", "RS256")
+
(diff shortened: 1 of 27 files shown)


WillemJiang:
@zhfeng, thanks for your contribution. Please add unit test of this new added feature.

willem-bd:
## Code Review Findings

Reviewed at high effort (7 angles × 6 candidates → 1-vote verify). 14 candidates verified, 1 refuted, 13 confirmed. Top 10 ranked by severity below.

---

### 🔴 #1 — Security: `email_verified` string truthiness bypasses verification
**File:** `backend/app/gateway/auth/oidc.py` ~line 465

`email_verified = bool(merged.get("email_verified", False))` — Python's `bool("false")` evaluates to `True`. A non-standard OIDC provider returning `email_verified` as the JSON string `"false"` (instead of boolean `false`) would pass the check, making the system treat the email as verified when it isn't.

**Failure scenario:** A non-standard OIDC provider returns `email_verified` as the string `"false"` → `bool("false")` is `True` → `OIDCIdentity.email_verified` is `True` → `user_provisioning` allows auto-creation or email linking for an unverified email, enabling account takeover when `allow_email_linking` is enabled.

**Fix:** Replace `bool(merged.get(...))` with an explicit check: `merged.get("email_verified") is True`.

---

### 🔴 #2 — Bug: Email case normalization mismatch causes duplicate accounts
**File:** `backend/app/gateway/auth/user_provisioning.py` ~line 682

`identity.email` is used raw (potentially mixed-case) for `get_user_by_email` lookup and `create_oauth_user`, while local registration normalizes via `EmailStr` (lowercases the domain). The SQLite repo does case-sensitive comparison.

**Failure scenario:** OIDC provider returns `"Admin@Example.COM"` → `get_user_by_email` fails to find existing local account `"admin@example.com"` → if `auto_create_users=True`, a duplicate account is created; if `allow_email_linking=True`, the link silently fails.

**Fix:** Normalize `identity.email` to lowercase before lookups and account creation (e.g., `identity.email.lower()`).

---

### 🟡 #3 — Bug: `_metadata_from_dict` crashes when `overrides=None`
**File:** `backend/app/gateway/auth/oidc.py` ~line 173

`_metadata_from_dict` calls `overrides.get(...)` without guarding against `overrides=None`. The `discover()` method has `overrides=None` as default and passes it through on cache hits.

**Failure scenario:** A future caller invokes `service.discover(issuer)` without overrides → cache hit passes `None` to `_metadata_from_dict` → `AttributeError: 'NoneType' object has no attribute 'get'`.

**Fix:** Add `overrides = overrides or {}` at the top of `_metadata_from_dict` or in `discover()` before calling it.

---

### 🟡 #4 — Resource leak: OIDCService httpx client never closed on shutdown
**File:** `backend/app/gateway/routers/auth.py` ~line 848

The `_get_oidc_service()` singleton creates an `httpx.AsyncClient` in `OIDCService.__init__()`, but `OIDCService.close()` is never called from the app lifespan shutdown hook.

**Failure scenario:** On worker shutdown/restart, the connection pool is abandoned without graceful close, producing `ResourceWarning` logs and potentially hanging the worker with lingering TCP connections.

**Fix:** Register `_get_oidc_service().close()` in the `lifespan()` shutdown path in `app.py`.

---

### 🟡 #5 — Dead code: colon check in `validate_next_param` is unreachable
**File:** `backend/app/gateway/routers/auth.py` ~line 1141

The condition `if ":" in next_param and not next_param.startswith("/")` can never be `True` because line 1137 already returns `None` for anything not starting with `/`. The intended protection against protocol-bearing paths is silently not enforced.

**Failure scenario:** A path like `/:evil` passes all checks. While `urllib.parse.quote()` mitigates the redirect, the intended validation is dead code and misleads future auditors.

**Fix:** Change the colon check to `if ":" in next_param:` (remove the redundant `not startswith("/")`) or restructure the validation to handle colon-in-path cases.

---

### 🟡 #6 — Reuse: `_constant_time_compare` reimplements `secrets.compare_digest`
**File:** `backend/app/gateway/auth/oidc.py` ~line 477

A hand-rolled constant-time comparison that re-implements `secrets.compare_digest` from the Python stdlib. The same PR already imports and uses `secrets.compare_digest` at `auth.py:1049`.

**Fix:** Replace the body with `return secrets.compare_digest(a, b)` and remove the 7-line custom implementation.

---

### 🔵 #7 — Dead code: redundant second `alg=none` check in `validate_id_token`
**File:** `backend/app/gateway/auth/oidc.py` ~line 367

The method has two `alg=none` rejection checks — one before key resolution (line ~334) and another after `jwt.decode` (line ~367). The second is unreachable: the first blocks it, and `jwt.decode`'s explicit asymmetric-only allowlist also rejects it. The second check also re-parses the unverified header unnecessarily.

**Fix:** Remove the second check (lines ~367-370).

---

### 🔵 #8 — Code quality: regex validation runs after provider config lookup
**File:** `backend/app/gateway/routers/auth.py` ~line 942

`_OIDC_PROVIDER_KEY_RE.match(provider)` runs after `oidc_config.providers.get(provider)`, so the provider string is used as a dict key and embedded in cookie names/paths before being validated.

**Fix:** Move the regex check before the config lookup.

---

### 🔵 #9 — Inconsistency: `get_me` uses `getattr` while `register`/`initialize_admin` access `user.oauth_provider` directly
**File:** `backend/app/gateway/routers/auth.py` ~line 829

`get_me` uses `getattr(user, 'oauth_provider', None)` (defensive) while `register` and `initialize_admin` use `user.oauth_provider` directly. If a code path ever returns a user-like object without the attribute, `get_me` degrades gracefully while the others crash.

**Fix:** Use consistent access — either `getattr` everywhere or direct access everywhere (since the `User` model always has the field).

---

### 🔵 #10 — Duplication: `validate_next_param` implemented independently in backend and frontend
**Files:** `backend/app/gateway/routers/auth.py` ~line 1129, `frontend/src/app/(auth)/auth/callback/page.tsx` ~line 14

The same open-redirect protection logic is duplicated across the stack. A fix to one copy must be manually replicated in the other or the protection diverges.

**Fix:** Consider extracting a shared utility or at minimum documenting the cross-stack dependency clearly.

---

<details><summary>ℹ️ Verification summary</summary>

- **14 unique candidates** from 6 finder angles (Angle B hit rate limit; old placeholder OAuth endpoints were the only deletions, so limited removed-behavior surface)
- **15 parallel verifiers** run against the actual PR diff
- **1 REFUTED**: middleware path traversal via `../` — route handler validates `{provider}` against `^[a-zA-Z0-9_-]+$`, blocking traversal before any protected endpoint is reached
- **13 CONFIRMED**: top 10 reported above; 3 additional confirmed findings cut for lower severity: `_set_csrf_cookie` duplicates `CSRFMiddleware.dispatch`, `sso_cancelled` error code defined in frontend but never produced by backend, singleton race condition theoretically possible but safe under CPython GIL

</details>

zhfeng:
@WillemJiang is it good for merging?

zhfeng:
Thanks for the thorough review! All 10 findings have been addressed. Summary below:

| # | Finding | Status | Resolution |
|---|---------|--------|------------|
| 🔴 1 | `email_verified` string truthiness bypass | ✅ Fixed | `oidc.py` now uses `merged.get("email_verified") is True` — only a real JSON boolean `true` counts as verified. |
| 🔴 2 | Email case normalization mismatch | ✅ Fixed | `user_provisioning.py` normalizes `email = identity.email.lower()` before lookup, linking, and creation. |
| 🟡 3 | `_metadata_from_dict` crash on `overrides=None` | ✅ Fixed | Guarded with `overrides = overrides or {}`. |
| 🟡 4 | OIDCService httpx client never closed | ✅ Fixed | `close_oidc_service()` is now registered in the `lifespan()` shutdown path in `app.py`. |
| 🟡 5 | Dead colon check in `validate_next_param` | ✅ Fixed | Simplified to `if ":" in next_param:` so the protocol guard is actually enforced. |
| 🟡 6 | `_constant_time_compare` reimplements stdlib | ✅ Fixed | Body replaced with `secrets.compare_digest(a, b)`. |
| 🔵 7 | Redundant second `alg=none` check | ✅ Fixed | Removed; the pre-decode allowlist plus pyjwt's asymmetric-only `algorithms` arg are the single source of truth. |
| 🔵 8 | Regex validation runs after config lookup | ✅ Fixed | `_OIDC_PROVIDER_KEY_RE.match(provider)` now runs before the config lookup in both `oauth_login` and `oauth_callback`. |
| 🔵 9 | Inconsistent `oauth_provider` access in `get_me` | ✅ Fixed | `get_me` now uses direct `user.oauth_provider` access, consistent with `register`/`initialize_admin`. |
| 🔵 10 | `validate_next_param` duplicated FE/BE | ✅ Fixed | The frontend `validateNextParam` now includes the same colon guard (`next.includes(":")`) as the backend, so the two copies enforce identical open-redirect rules. |

The REFUTED finding (middleware path traversal) and the 3 lower-severity items noted in the verification summary were reviewed as well. Branch rebased on latest `main` before pushing.

backend/app/gateway/auth/oidc.py
```diff
@@ -0,0 +1,425 @@
...
+                    expected_sub=claims["sub"],
+                )
+            except OIDCError as exc:
+                logger.warning("OIDC userinfo fetch failed (continuing with ID token): %s", exc)
+
+        # Merge userinfo into claims (userinfo takes precedence for email)
+        merged = {**claims, **userinfo}
+
+        email = merged.get("email") or ""
+        email_verified = merged.get("email_verified") is True
```
willem-bd, line 419: **[Security] `email_verified` is taken from the userinfo-overridden merge, which can upgrade an unverified email to verified.**

`merged = {**claims, **userinfo}` lets the **unsigned** UserInfo response override the **signed** ID token's `email_verified` (the comment only notes precedence 'for email', but it applies to this flag too). The ID token is cryptographically signed; UserInfo is only bearer-protected. So a provider whose UserInfo returns `email_verified: true` for a claim the signed ID token marked `false` will be treated as verified here.

That flag then gates the two most security-sensitive decisions in `user_provisioning.py`: `require_verified_email` and the email-**linking** verified gate. With `allow_email_linking: true` and any provider that asserts verified emails the user doesn't actually own (a misconfigured/rogue IdP — this PR explicitly supports arbitrary providers), an unverified email can be linked onto an existing local account → account takeover.

Suggested fix: take `email`/`email_verified` from the signed `claims` first, falling back to UserInfo only when the claim is absent — never let UserInfo override a signed `email_verified`.

backend/app/gateway/auth/user_provisioning.py
```diff
@@ -0,0 +1,119 @@
...
+
+    # 5. Auto-create
+    if not provider_config.auto_create_users:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Automatic account creation is disabled. Contact your administrator.",
+        )
+
+    role = _resolve_role(email, provider_config.admin_emails)
+    user = await local_provider.create_oauth_user(
```
willem-bd: **[Correctness] Concurrent / duplicate provisioning raises an uncaught `ValueError` → HTTP 500.**

`create_oauth_user` → `repositories/sqlite.py:create_user` raises `ValueError` on `IntegrityError` (duplicate `email` or `(oauth_provider, oauth_id)` collision). Neither `get_or_provision_oidc_user` nor `oauth_callback` catches it — `oauth_callback` only catches `OIDCError`/`HTTPException`.

Two concurrent callbacks that resolve to the same email (double-click, replayed code, or one identity arriving via two configured providers) both pass the `get_user_by_email == None` check; the loser fails the unique index, `create_user` raises `ValueError`, and it bubbles to a raw 500 — the user is left stranded on the callback URL and the state cookie is never cleared.

Suggested fix: catch the integrity conflict here and treat it as 'already exists' (re-lookup and link/return), or surface a clean 409 / `sso_account_exists` redirect.

backend/app/gateway/auth/user_provisioning.py
```diff
@@ -0,0 +1,119 @@
...
+
+        # Check for conflicting OAuth link on the existing user
+        if local_user.oauth_provider and local_user.oauth_provider != provider_id:
+            raise HTTPException(
+                status_code=status.HTTP_409_CONFLICT,
+                detail="This email is already linked to a different SSO provider.",
+            )
+
+        # Link the existing user to this OIDC identity
+        local_user.oauth_provider = provider_id
```
willem-bd: **[Correctness/Security] Linked SSO users keep their `password_hash`, so `/change-password` still works for 'SSO' users.**

Linking sets `oauth_provider`/`oauth_id` but never clears the local user's `password_hash`. The frontend then hides the password form and `SSO.md` states 'OAuth users cannot change password' — but `change_password` only rejects users with `password_hash is None`. Since the hash is still present, a linked 'SSO' user can still change their password via the API, and the original password still authenticates at `/login/local`. That contradicts the advertised SSO-only account semantics (and silently changes account semantics during a linking-based takeover).

Suggested fix: clear `password_hash` when linking, or tighten `change_password` to also reject users where `oauth_provider` is set.

backend/app/gateway/auth/oidc.py
```diff
@@ -0,0 +1,425 @@
...
+        try:
+            resp = await self._http.get(discovery_url)
+            resp.raise_for_status()
+            data: dict[str, Any] = resp.json()
+        except httpx.HTTPStatusError as exc:
+            raise OIDCError(f"OIDC discovery failed for issuer {issuer}: HTTP {exc.response.status_code}") from exc
+        except httpx.RequestError as exc:
+            raise OIDCError(f"OIDC discovery failed for issuer {issuer}: {exc}") from exc
+
+        if not data.get("issuer"):
```
willem-bd: **[Security] Discovery doesn't pin the discovered `issuer` to the configured `issuer`.**

`discover` only checks the document has an `issuer` field, but never compares it to the configured `issuer`. `_metadata_from_dict` uses `data["issuer"]` verbatim, and `validate_id_token` then trusts it as the expected `iss`. RFC 8414 §4 requires the metadata issuer to equal the discovery URL — not enforcing it means a tampered/rogue discovery doc (MITM, DNS rebinding, or a misconfigured provider) can set the accepted `iss` to an attacker-chosen value, widening the ID-token forgery surface.

Suggested fix: add `if data['issuer'] != issuer: raise OIDCError('discovered issuer does not match configured issuer')` right after this check.

backend/app/gateway/routers/auth.py
```diff
@@ -499,39 +522,297 @@ async def initialize_admin(request: Request, response: Response, body: Initializ
...
-    if provider not in ["github", "google"]:
-        raise HTTPException(
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail=f"Unsupported OAuth provider: {provider}",
+    if provider_config.redirect_uri:
+        return provider_config.redirect_uri
+
+    # Development fallback: build from the current request's base URL.
+    scheme = request.url.scheme
+    host = request.headers.get("host", "localhost:8001")
```
willem-bd: **[Security] `redirect_uri` dev fallback trusts the raw `Host` header.**

When `redirect_uri` isn't configured, the fallback builds it from `request.headers.get('host')` and `request.url.scheme`. With a naive reverse proxy that passes the client `Host` through, and a loosely-configured IdP client (the SSO.md examples lean toward wildcard dev redirects), a spoofed `Host` can steer the IdP's `redirect_uri` to an attacker host. Separately, behind nginx without `X-Forwarded-Proto` handling, `scheme` is `http` in production so `secure` cookies never get set.

Suggested fix: reuse the proxy-aware origin resolution already in `csrf_middleware._request_origin` instead of the raw `Host` header, and require `redirect_uri` to be set explicitly in production.

willem-bd (commented):
Reviewed the OIDC/SSO implementation at **high effort**. Overall it's solid — alg restriction (no `none`/HS), `require: [exp,iss,sub,aud]`, constant-time nonce check, PKCE S256, state bound to a signed HttpOnly cookie, and a UserInfo `sub` match.

The **5 inline comments** below are what I'd ask to address before merge. The top three, in priority order:

1. **`email_verified` is read from the UserInfo-overridden merge** — the unsigned source can flip it false→true and defeats the email-linking verified gate (account-takeover surface under `allow_email_linking`).
2. **Concurrent / duplicate provisioning raises an uncaught `ValueError` → HTTP 500** (state cookie left uncleared, user stranded on the callback URL).
3. **Email-linked SSO users keep their `password_hash`** — so `/change-password` still works for 'SSO' users, contradicting the UI and `SSO.md`.

#6 and #7 are cheap security hardening: **pin the discovered `issuer` to the configured one**, and **use proxy-aware origin resolution for the `redirect_uri` fallback** instead of the raw `Host` header.

**Also worth a follow-up** (not blocking, not inlined):
- `admin_emails` role is evaluated only on first auto-create and never reconciled on later logins — admin stays sticky (and a one-time grant is permanent).
- SSO users are created with `token_version=0` and have no runtime revocation path, since `change_password` (the only API path that bumps it) is blocked for them. A stolen SSO JWT is valid for the full `token_expiry_days`.
- `validateNextParam` exists in three already-drifted copies (frontend `callback`/`login` + backend `validate_next_param`) — worth unifying into one strict shared helper.
- `validate_next_param` rejects **all** colons, which silently drops any legitimate `next` path containing `:` to `/workspace`.

**Verified NOT issues** (so they don't get 'fixed' into bugs): `get_user_by_oauth` exists on `LocalAuthProvider`/repos; `$ENV_VAR` secrets are expanded by `AppConfig.resolve_env_variables`; `(auth)/layout.tsx` sets `dynamic = "force-dynamic"` so `useSearchParams()` needs no `<Suspense>`; login-CSRF is mitigated by the state↔cookie binding; `alg=none`/HS-key confusion is blocked by the `allowed_algorithms` allowlist; the `_get_oidc_service` singleton is event-loop-safe.

backend/app/gateway/auth/oidc.py
zhfeng: Thanks — agreed on the takeover risk. Rather than harden the verified-email gate around linking, we removed email-match auto-linking entirely (commit b44b302).

`get_or_provision_oidc_user` now only ever (a) reuses an account already linked by `(provider, subject)`, or (b) auto-creates a brand-new account. If the IdP-reported email collides with a pre-existing local account, the SSO login is blocked with a **409** instead of linking onto that row — so an SSO login can never seize a password account, regardless of how much we trust the IdP's `email_verified`. The `allow_email_linking` config field is gone.

This closes the account-takeover half of this finding, and also resolves the related "linked SSO users keep their `password_hash`" comment (no linking → no orphaned hash).

Note on the remaining half: the `merged = {**claims, **userinfo}` ordering at `oidc.py:408-411` still lets unsigned UserInfo override the signed `email_verified`, which still feeds the `require_verified_email` gate on the auto-create path. That's a narrower issue now (it can let an auto-created account be treated as verified, but no longer enables takeover). We'll address the signed-claims-first precedence separately so this PR's change stays scoped to the linking removal.

backend/app/gateway/auth/user_provisioning.py
zhfeng: Resolved by removing email-match auto-linking in commit b44b302. Linking is gone entirely, so there's no longer a path that sets `oauth_provider`/`oauth_id` on an existing local account while leaving `password_hash` in place. A pre-existing local account now blocks the SSO login with a 409 rather than being converted into a hybrid account.

backend/app/gateway/auth/oidc.py
zhfeng: Fixed in commit 58060ae. `discover()` now pins the discovered metadata issuer to the configured issuer (trailing-slash tolerant) right after the presence check:

```python
if discovered_issuer.rstrip("/") != issuer.rstrip("/"):
    raise OIDCError(f"OIDC discovered issuer '{discovered_issuer}' does not match configured issuer '{issuer}'")
```

A tampered/rogue discovery document can no longer steer the accepted `iss` to an attacker-chosen value. Covered by `test_oidc_discover_rejects_mismatched_issuer` and `test_oidc_discover_accepts_issuer_with_trailing_slash_difference`.

backend/app/gateway/routers/auth.py
zhfeng: Fixed in commit 58060ae. The dev fallback now builds the redirect URI from the proxy-aware origin resolver (`csrf_middleware._request_origin`) instead of the raw `Host` header:

```python
origin = _request_origin(request)
if not origin:
    origin = f"{request.url.scheme}://{request.headers.get('host', 'localhost:8001')}"
return f"{origin}/api/v1/auth/callback/{provider_id}"
```

`_request_origin` honors `Forwarded` / `X-Forwarded-Host` / `X-Forwarded-Proto` exactly like the CSRF origin checks, so a spoofed `Host` can no longer steer the IdP `redirect_uri`, and the scheme reflects the real client-facing protocol behind a proxy (fixing the `http`-behind-nginx `secure` cookie issue too). An explicitly configured `redirect_uri` still takes precedence. Covered by three new tests (`test_oidc_redirect_uri_*`).

backend/app/gateway/auth/user_provisioning.py
zhfeng: Fixed in commit 58060ae. `get_or_provision_oidc_user` now catches the `ValueError` that `create_user` raises on the unique-index conflict, so a concurrent/duplicate provisioning no longer bubbles to a raw 500:

```python
try:
    user = await local_provider.create_oauth_user(...)
except ValueError:
    existing = await local_provider.get_user_by_oauth(provider_id, identity.subject)
    if existing:
        return {"user": existing, "created": False}
    raise HTTPException(status_code=409, detail="An account with this email already exists. ...") from None
```

If the race winner created this same `(provider, subject)` identity, we re-resolve and return it; if the collision was on email only (a different account), it surfaces a clean 409 → `sso_account_exists` redirect, which `oauth_callback` already maps and which clears the state cookie. Covered by `test_oidc_provision_recovers_existing_user_on_create_race` and `test_oidc_provision_create_race_on_email_only_raises_409`.

WillemJiang (approved):

PR #3508, opened 2026-06-11
fix(frontend): isolate new chat thread messages

zwj110610:
## Why

Fix a New Chat state leak where messages from the previous thread could remain visible above the welcome screen after client-side navigation to `/workspace/chats/new`.

## What changed
The chat page reused thread-local UI state across client-side navigation. In particular, optimistic user messages created during send were stored on the stream hook instance without being scoped to the currently displayed thread. When the user switched to New Chat, those optimistic messages could still be merged into the rendered message list, making the previous prompt appear on the empty welcome page.

This was especially visible when the previous prompt contained an artifact markdown response such as `shared artifact`.

## Surface area

## Screenshots / Recording

bug2

## Bug fix verification

- Test: `new chat does not show previous optimistic user message after client-side navigation`
- Also added coverage for persisted history leakage: `new chat does not show previous thread messages after client-side navigation`

- Expected: yes.

## Validation

Ran:

- `cd frontend && ./node_modules/.bin/tsc --noEmit`
- `cd frontend && ./node_modules/.bin/eslint . --ext .ts,.tsx --rule 'import/order: off'`
- `cd frontend && ./node_modules/.bin/prettier --check src/app/workspace/chats/[thread_id]/page.tsx src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx src/components/workspace/chats/use-thread-chat.ts src/core/threads/hooks.ts tests/e2e/thread-history.spec.ts`
- `cd frontend && ./node_modules/.bin/playwright test tests/e2e/thread-history.spec.ts --project=chromium --grep optimistic --list`

## AI assistance
**Tool(s) used:** Codex

**How you used it:** Investigated the New Chat message leakage, identified that optimistic messages and persisted thread state were not scoped to the displayed thread, implemented the fix, and added E2E regression coverage for the leak.

### DIFF
--- a/frontend/src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx
+++ b/frontend/src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx
@@ -72,20 +72,21 @@ export default function AgentChatPage() {
     loadMoreHistory,
   } = useThreadStream({
     threadId: isNewThread ? undefined : threadId,
+    displayThreadId: threadId,
     context: { ...settings.context, agent_name: agent_name },
     isMock,
     onSend: () => {
       setIsWelcomeMode(false);
     },
     onStart: (createdThreadId) => {
-      setThreadId(createdThreadId);
-      setIsNewThread(false);
       // ! Important: Never use next.js router for navigation in this case, otherwise it will cause the thread to re-mount and lose all states. Use native history API instead.
       history.replaceState(
         null,
         "",
         `/workspace/agents/${agent_name}/chats/${createdThreadId}`,
       );
+      setThreadId(createdThreadId);
+      setIsNewThread(false);
     },
     onFinish: (state) => {
       if (document.hidden || !document.hasFocus()) {
--- a/frontend/src/app/workspace/chats/[thread_id]/page.tsx
+++ b/frontend/src/app/workspace/chats/[thread_id]/page.tsx
@@ -75,6 +75,7 @@ export default function ChatPage() {
     loadMoreHistory,
   } = useThreadStream({
     threadId: isNewThread ? undefined : threadId,
+    displayThreadId: threadId,
     context: settings.context,
     isMock,
     // onSend only animates the UI; do NOT flip `isNewThread` here — the
@@ -84,10 +85,10 @@ export default function ChatPage() {
       setIsWelcomeMode(false);
     },
     onStart: (createdThreadId) => {
-      setThreadId(createdThreadId);
-      setIsNewThread(false);
       // ! Important: Never use next.js router for navigation in this case, otherwise it will cause the thread to re-mount and lose all states. Use native history API instead.
       history.replaceState(null, "", `/workspace/chats/${createdThreadId}`);
+      setThreadId(createdThreadId);
+      setIsNewThread(false);
     },
     onFinish: (state) => {
       if (document.hidden || !document.hasFocus()) {
--- a/frontend/src/components/workspace/chats/use-thread-chat.ts
+++ b/frontend/src/components/workspace/chats/use-thread-chat.ts
@@ -1,29 +1,44 @@
 "use client";
 
 import { useParams, usePathname, useSearchParams } from "next/navigation";
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 
 import { uuid } from "@/core/utils/uuid";
 
 export function useThreadChat() {
   const { thread_id: threadIdFromPath } = useParams<{ thread_id: string }>();
   const pathname = usePathname();
+  const actualPathname =
+    typeof window === "undefined" ? pathname : window.location.pathname;
+  const isNewPath = actualPathname.endsWith("/new");
+  const newThreadIdRef = useRef<string | null>(
+    threadIdFromPath === "new" ? uuid() : null,
+  );
+
+  if (isNewPath && !newThreadIdRef.current) {
+    newThreadIdRef.current = uuid();
+  }
 
   const searchParams = useSearchParams();
-  const [threadId, setThreadId] = useState(() => {
-    return threadIdFromPath === "new" ? uuid() : threadIdFromPath;
+  const [threadId, setThreadIdState] = useState(() => {
+    return threadIdFromPath === "new"
+      ? (newThreadIdRef.current ?? uuid())
+      : threadIdFromPath;
   });
 
-  const [isNewThread, setIsNewThread] = useState(
+  const [isNewThreadState, setIsNewThreadState] = useState(
     () => threadIdFromPath === "new",
   );
 
   useEffect(() => {
-    if (pathname.endsWith("/new")) {
-      setIsNewThread(true);
-      setThreadId(uuid());
+    if (isNewPath) {
+      const nextThreadId = newThreadIdRef.current ?? uuid();
+      newThreadIdRef.current = nextThreadId;
+      setIsNewThreadState(true);
+      setThreadIdState(nextThreadId);
       return;
     }
+    newThreadIdRef.current = null;
     // Guard: after history.replaceState updates the URL from /chats/new to
     // /chats/{UUID}, Next.js useParams may still return the stale "new" value
     // because replaceState does not trigger router updates.  Avoid propagating
@@ -32,9 +47,28 @@ export function useThreadChat() {
     if (threadIdFromPath === "new") {
       return;
     }
-    setIsNewThread(false);
-    setThreadId(threadIdFromPath);
-  }, [pathname, threadIdFromPath]);
+    setIsNewThreadState(false);
+    setThreadIdState(threadIdFromPath);
+  }, [isNewPath, threadIdFromPath]);
+
+  const setThreadId = useCallback((nextThreadId: string) => {
+    newThreadIdRef.current = null;
+    setThreadIdState(nextThreadId);
+  }, []);
+
+  const setIsNewThread = useCallback((nextIsNewThread: boolean) => {
+    if (!nextIsNewThread) {
+      newThreadIdRef.current = null;
+    }
+    setIsNewThreadState(nextIsNewThread);
+  }, []);
+
   const isMock = searchParams.get("mock") === "true";
-  return { threadId, setThreadId, isNewThread, setIsNewThread, isMock };
+  return {
+    threadId: isNewPath ? (newThreadIdRef.current ?? threadId) : threadId,
+    setThreadId,
+    isNewThread: isNewPath ? true : isNewThreadState,
+    setIsNewThread,
+    isMock,
+  };
 }
--- a/frontend/src/core/threads/hooks.ts
+++ b/frontend/src/core/threads/hooks.ts
@@ -54,6 +54,13 @@ type SendMessageOptions = {
   additionalKwargs?: Record<string, unknown>;
 };
 
+const EMPTY_THREAD_VALUES: AgentThreadState = {
+  title: "",
+  messages: [],
+  artifacts: [],
+  todos: [],
+};
+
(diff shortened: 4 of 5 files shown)


willem-bd:
## Code Review: High-Effort Findings

Reviewed with 7-angle sweep (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude) + 1-vote verification.

---

### 🔴 #1 — `messagesRef` starved during new-thread streaming (correctness)

**File:** `frontend/src/core/threads/hooks.ts:955`

`persistedMessages` returns `[]` when `threadId` is falsy (new thread). The guard `if (persistedMessages.length >= messagesRef.current.length)` evaluates to `0 >= N` (false), so `messagesRef` is never updated during the window between `submit()` and the `onStart` callback. During this gap, `onUpdateEvent` (line 536: `const _currentMessages = [...messagesRef.current]`) reads stale data from the previous thread, causing the summarization middleware to compute incorrect `_movedMessages` and potentially push old data into the new thread's history via `appendMessages`.

**Failure scenario:** User on thread A with 10 messages → navigates to `/chats/new` → sends a message → stream starts → `onUpdateEvent` fires inside the SDK's `for-await` loop before React re-renders → `messagesRef.current` still has thread A's 10 messages → summarization operates on wrong data.

**Suggested fix:** Reset `messagesRef` in the `threadId`-change effect (around line 691) alongside the other reset logic, or change the guard to also clear when `persistedMessages` is empty:

```ts
useEffect(() => {
  // existing resets…
  messagesRef.current = []; // clear stale messages on thread switch
}, [threadId]);
```

---

### 🟡 #2 — `window.location.pathname` is a non-reactive effect dependency (fragility)

**File:** `frontend/src/components/workspace/chats/use-thread-chat.ts:11`

`actualPathname = window.location.pathname` is read as a snapshot during render, then `isNewPath` is derived from it and used as a `useEffect` dependency. This works today only because `onStart` always pairs `history.replaceState` with `setThreadId`/`setIsNewThread` (which trigger the re-render that re-evaluates `actualPathname`). A future code path that changes the URL without a state update would leave `isNewPath` stale — the effect would not re-fire.

**Consider:** Documenting this coupling explicitly, or deriving `isNewPath` from the `pathname` returned by `usePathname()` (which is reactive) and only reading `window.location.pathname` when the two disagree after a known `replaceState`.

---

### 🟡 #3 — Shallow spread in `mergedThread.values` only clears known fields (correctness, minor)

**File:** `frontend/src/core/threads/hooks.ts:981`

When `threadId` is falsy, the code does `{ ...thread.values, title: undefined, messages: [], artifacts: [], todos: [] }`. This only clears 4 explicitly-listed fields. Any custom agent state field added by a sub-agent graph would leak through from the previous thread's SDK state for one render cycle (before `useStream` re-initializes). The `AgentThreadState` type is `Record<string, unknown>`, so the type system does not prevent this.

**Consider:** Either clearing all keys (`Object.fromEntries(Object.keys(thread.values).map(k => [k, undefined]))`) or replacing with a static empty state object.

---

### 🟡 #4 — Ref read inside state updater in `handleStreamStart` (correctness, mitigated)

**File:** `frontend/src/core/threads/hooks.ts:451`

`setOptimisticThreadId` reads `currentViewThreadIdRef.current` inside a state setter callback. Refs are mutable and not part of React's reactive data flow — in concurrent mode, the ref value may differ between when the updater was queued and when it executes. Currently mitigated by the `visibleOptimisticMessages` guard (`optimisticThreadId === currentViewThreadId ? optimisticMessages : []`) which protects the rendering path regardless of the updater's behavior.

**Low risk today** because concurrent mode is not heavily used, but worth noting for future React upgrades.

---

### 🟢 #5 — Duplicated `onStart`/`onFinish` callbacks between ChatPage and AgentChatPage (maintenance)

**Files:** `frontend/src/app/workspace/chats/[thread_id]/page.tsx:86` and `frontend/src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx:80`

The `onStart` callback body (`setThreadId + setIsNewThread + history.replaceState`) and the `onFinish` notification callback are nearly verbatim duplicated, differing only in URL path prefix. Drift is already visible: `onFinish` uses `.at(-1)` in the chats page but `[state.messages.length - 1]` in the agent page.

**Consider:** Extracting a shared helper (e.g., `useThreadNavigation(agentName?)`) to hold the `onStart`/`onFinish` logic, parameterized by the URL path prefix.

---

### 🟢 #6 — 7 thread-identity variables increase cognitive load (altitude)

**File:** `frontend/src/core/threads/hooks.ts:401+`

The PR adds `displayThreadId`, `currentViewThreadId`, `currentViewThreadIdRef`, and `optimisticThreadId` on top of the existing `threadId`, `onStreamThreadId`, and `threadIdRef` — totalling 7 variables for thread identity. Each serves a purpose, but the overall system is harder to reason about. During rapid navigation, these variables can briefly disagree, and a developer must trace which is authoritative at any point.

**Consider:** Whether `displayThreadId` could be the single source of truth, with `threadId` derived from it (i.e., `displayThreadId when !isNewThread, else undefined`), reducing the number of independent variables.

---

**Overall:** The PR effectively fixes the message leak bug with good E2E coverage. Finding #1 (`messagesRef` starvation) is the most actionable — it is a regression where the new `persistedMessages` guard prevents `messagesRef` from being cleared on thread switch, and should be fixed before merge. The other findings are lower-risk fragility/maintenance concerns.

zwj110610:
> ## Code Review: High-Effort Findings
> Reviewed with 7-angle sweep (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude) + 1-vote verification.
> 
> ### 🔴 #1 — `messagesRef` starved during new-thread streaming (correctness)
> **File:** `frontend/src/core/threads/hooks.ts:955`
> 
> `persistedMessages` returns `[]` when `threadId` is falsy (new thread). The guard `if (persistedMessages.length >= messagesRef.current.length)` evaluates to `0 >= N` (false), so `messagesRef` is never updated during the window between `submit()` and the `onStart` callback. During this gap, `onUpdateEvent` (line 536: `const _currentMessages = [...messagesRef.current]`) reads stale data from the previous thread, causing the summarization middleware to compute incorrect `_movedMessages` and potentially push old data into the new thread's history via `appendMessages`.
> 
> **Failure scenario:** User on thread A with 10 messages → navigates to `/chats/new` → sends a message → stream starts → `onUpdateEvent` fires inside the SDK's `for-await` loop before React re-renders → `messagesRef.current` still has thread A's 10 messages → summarization operates on wrong data.
> 
> **Suggested fix:** Reset `messagesRef` in the `threadId`-change effect (around line 691) alongside the other reset logic, or change the guard to also clear when `persistedMessages` is empty:
> 
> ```ts
> useEffect(() => {
>   // existing resets…
>   messagesRef.current = []; // clear stale messages on thread switch
> }, [threadId]);
> ```
> 
> ### 🟡 #2 — `window.location.pathname` is a non-reactive effect dependency (fragility)
> **File:** `frontend/src/components/workspace/chats/use-thread-chat.ts:11`
> 
> `actualPathname = window.location.pathname` is read as a snapshot during render, then `isNewPath` is derived from it and used as a `useEffect` dependency. This works today only because `onStart` always pairs `history.replaceState` with `setThreadId`/`setIsNewThread` (which trigger the re-render that re-evaluates `actualPathname`). A future code path that changes the URL without a state update would leave `isNewPath` stale — the effect would not re-fire.
> 
> **Consider:** Documenting this coupling explicitly, or deriving `isNewPath` from the `pathname` returned by `usePathname()` (which is reactive) and only reading `window.location.pathname` when the two disagree after a known `replaceState`.
> 
> ### 🟡 #3 — Shallow spread in `mergedThread.values` only clears known fields (correctness, minor)
> **File:** `frontend/src/core/threads/hooks.ts:981`
> 
> When `threadId` is falsy, the code does `{ ...thread.values, title: undefined, messages: [], artifacts: [], todos: [] }`. This only clears 4 explicitly-listed fields. Any custom agent state field added by a sub-agent graph would leak through from the previous thread's SDK state for one render cycle (before `useStream` re-initializes). The `AgentThreadState` type is `Record<string, unknown>`, so the type system does not prevent this.
> 
> **Consider:** Either clearing all keys (`Object.fromEntries(Object.keys(thread.values).map(k => [k, undefined]))`) or replacing with a static empty state object.
> 
> ### 🟡 #4 — Ref read inside state updater in `handleStreamStart` (correctness, mitigated)
> **File:** `frontend/src/core/threads/hooks.ts:451`
> 
> `setOptimisticThreadId` reads `currentViewThreadIdRef.current` inside a state setter callback. Refs are mutable and not part of React's reactive data flow — in concurrent mode, the ref value may differ between when the updater was queued and when it executes. Currently mitigated by the `visibleOptimisticMessages` guard (`optimisticThreadId === currentViewThreadId ? optimisticMessages : []`) which protects the rendering path regardless of the updater's behavior.
> 
> **Low risk today** because concurrent mode is not heavily used, but worth noting for future React upgrades.
> 
> ### 🟢 #5 — Duplicated `onStart`/`onFinish` callbacks between ChatPage and AgentChatPage (maintenance)
> **Files:** `frontend/src/app/workspace/chats/[thread_id]/page.tsx:86` and `frontend/src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx:80`
> 
> The `onStart` callback body (`setThreadId + setIsNewThread + history.replaceState`) and the `onFinish` notification callback are nearly verbatim duplicated, differing only in URL path prefix. Drift is already visible: `onFinish` uses `.at(-1)` in the chats page but `[state.messages.length - 1]` in the agent page.
> 
> **Consider:** Extracting a shared helper (e.g., `useThreadNavigation(agentName?)`) to hold the `onStart`/`onFinish` logic, parameterized by the URL path prefix.
> 
> ### 🟢 #6 — 7 thread-identity variables increase cognitive load (altitude)
> **File:** `frontend/src/core/threads/hooks.ts:401+`
> 
> The PR adds `displayThreadId`, `currentViewThreadId`, `currentViewThreadIdRef`, and `optimisticThreadId` on top of the existing `threadId`, `onStreamThreadId`, and `threadIdRef` — totalling 7 variables for thread identity. Each serves a purpose, but the overall system is harder to reason about. During rapid navigation, these variables can briefly disagree, and a developer must trace which is authoritative at any point.
> 
> **Consider:** Whether `displayThreadId` could be the single source of truth, with `threadId` derived from it (i.e., `displayThreadId when !isNewThread, else undefined`), reducing the number of independent variables.
> 
> **Overall:** The PR effectively fixes the message leak bug with good E2E coverage. Finding #1 (`messagesRef` starvation) is the most actionable — it is a regression where the new `persistedMessages` guard prevents `messagesRef` from being cleared on thread switch, and should be fixed before merge. The other findings are lower-risk fragility/maintenance concerns.


Thanks for the detailed review.

Addressed the actionable issues in `7259726a`:

- Reset `messagesRef.current` when the displayed thread changes, so summarization middleware cannot read messages from the previous thread during the new-thread streaming window.
- Also reset `summarizedRef` and the pending usage baseline with the same thread-local reset.
- Replaced the shallow-spread fallback for new-chat `values` with a static empty thread state, so custom graph state fields cannot leak through from the previous thread.

I left the other notes unchanged for now:
- `window.location.pathname` is still intentionally used to handle the native `history.replaceState` path where Next `usePathname()` can lag behind.
- The ref read in `handleStreamStart`, duplicated page callbacks, and thread identity naming are lower-risk cleanup items and seem better as follow-up refactors outside this bug fix.

Validation after the update:

- `tsc --noEmit`
- `eslint . --ext .ts,.tsx --rule 'import/order: off'`
- `prettier --check src/core/threads/hooks.ts tests/e2e/thread-history.spec.ts`
- `playwright test --project=chromium`

Full E2E result: `32 passed`.

WillemJiang (approved):

PR #3509, opened 2026-06-11
fix(frontend): resolve build permission issues introduced by node:22-alpine Alpine 3.24 upgrade

guyuecode:
## Summary

Pin the frontend base image from `node:22-alpine` to `node:22-alpine3.22` to avoid build failures introduced by the upstream Alpine 3.24 upgrade.

## Background

The `node:22-alpine` image now tracks Alpine 3.24:

[https://github.com/nodejs/docker-node/blob/c94b4249a1a1bfb2ebb05fac9012cd34fecc2db7/22/alpine3.24/Dockerfile](https://github.com/nodejs/docker-node/blob/c94b4249a1a1bfb2ebb05fac9012cd34fecc2db7/22/alpine3.24/Dockerfile)

After the underlying Alpine version changed from 3.22 to 3.24, the frontend image build started failing with permission-related errors during the pnpm setup stage:

```text
EPERM: operation not permitted, write
```

This issue is caused by changes in the base image behavior rather than any DeerFlow application code changes.

## Changes

```diff
-FROM node:22-alpine AS base
+FROM node:22-alpine3.22 AS base
```

## Impact

* Keeps the Alpine runtime version consistent with the previously working environment.
* Avoids permission-related build failures introduced by Alpine 3.24.
* No application code or runtime behavior changes.
* Reduces the risk of unexpected regressions caused by upstream base image updates.

## Validation

* Verified that the frontend image builds successfully after pinning the base image to `node:22-alpine3.22`.
* Confirmed no functional changes to the application runtime.

## PR Title

`fix(frontend): pin node:22-alpine to Alpine 3.22 to prevent build permission issues`

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -7,7 +7,7 @@ ARG PNPM_STORE_PATH=/root/.local/share/pnpm/store
 ARG NPM_REGISTRY
 
 # ── Base: shared setup ────────────────────────────────────────────────────────
-FROM node:22-alpine AS base
+FROM node:22-alpine3.22 AS base
 ARG PNPM_STORE_PATH
 ARG NPM_REGISTRY
 # Configure corepack registry before installing pnpm so the download itself
@@ -35,7 +35,7 @@ RUN cd /app/frontend && pnpm install --frozen-lockfile
 RUN cd /app/frontend && SKIP_ENV_VALIDATION=1 pnpm build
 
 # ── Prod: minimal runtime with pre-built output ───────────────────────────────
-FROM node:22-alpine AS prod
+FROM node:22-alpine3.22 AS prod
 ARG PNPM_STORE_PATH
 ARG NPM_REGISTRY
 RUN if [ -n "${NPM_REGISTRY}" ]; then \


WillemJiang:
@guyuecode It looks like the permission issue has been fixed. I cannot reproduce the error here https://github.com/WillemJiang/deer-flow/actions/runs/27359089601/job/80841305289#logs. 
Can you double-check it?

guyuecode:
> @guyuecode It looks like the permission issue has been fixed. I cannot reproduce the error here https://github.com/WillemJiang/deer-flow/actions/runs/27359089601/job/80841305289#logs. Can you double-check it?

Thanks for your reply.

I was able to reproduce this issue in different build environments and found that it does not appear to be caused by the Alpine 3.24 upgrade itself.

### Build Environment Comparison

#### Environment 1 (Build Succeeds)

* OS: Debian 12
* Docker: 29.5.3

Build result:

```text
[+] build 3/3
 ✔ Image deer-flow-frontend    Built                                                              158.6s
 ✔ Image deer-flow-gateway     Built                                                              158.6s
 ✔ Image deer-flow-provisioner Built                                                              158.6s

==========================================
  ✓ Images built successfully
==========================================
```

#### Environment 2 (Build Fails)

* OS: CentOS 7
* Docker: 26.1.4

Build result:

```text
=> ERROR [frontend base 3/6] RUN pnpm config set store-dir /root/.local/share/pnpm/store
------
 > [frontend base 3/6] RUN pnpm config set store-dir /root/.local/share/pnpm/store:
0.835  EPERM  EPERM: operation not permitted, write
0.835
0.835
0.835
------
failed to solve: process "/bin/sh -c pnpm config set store-dir ${PNPM_STORE_PATH}" did not complete successfully: exit code: 255
```

### Observation

The same source code and Dockerfile build successfully on Debian 12 with Docker 29.5.3, but fail consistently on CentOS 7 with Docker 26.1.4.

This suggests that the issue may be related to differences in the Docker runtime, BuildKit implementation, filesystem permissions, or compatibility between older Docker/CentOS environments and the current Node.js/Pnpm setup, rather than being directly caused by the Alpine 3.24 upgrade.

Further investigation would be needed to determine the exact root cause.

WillemJiang:
@guyuecode  thanks for the reply.  Please keep us posted if you dig into the root cause.

PR #3517, opened 2026-06-12
fix(security): mount host Docker socket only in aio (DooD) sandbox mode

### LINKED ISSUE #3516: security: host Docker socket mounted into gateway by default (DooD host-escape surface)
## Summary

The default Docker Compose stack mounts the host Docker socket
(`/var/run/docker.sock`) read-write into the `gateway` container, which runs as
root and ships the Docker CLI. Because the gateway executes model-generated
shell/code, any in-container code-execution primitive (prompt injection, MCP /
tool misuse, or a future RCE) can reach the socket and obtain
**root-equivalent control of the host** — e.g. `docker run -v /:/host …`.

The key point: this happens **even in the default `local` sandbox mode, which
does not need the socket at all**. So the host-escape channel is open on the
default attack surface for zero functional benefit. The deploy script already
"knows" this — `scripts/deploy.sh` only checks for the socket when
`sandbox_mode != local` — yet the Compose files mount it unconditionally.

Thanks to **@greatmengqi** for the original report that surfaced this. 🙏

## Where

- `docker/docker-compose.yaml` — `gateway.volumes: ${DEER_FLOW_DOCKER_SOCKET}:/var/run/docker.sock`
- `docker/docker-compose-dev.yaml` — `gateway.volumes: /var/run/docker.sock:/var/run/docker.sock`
- `backend/Dockerfile` — no `USER` directive (runs as root); Docker CLI baked in (`COPY --from=docker:cli …`)

## Two corrections to the original report (for the record)

Confirming the finding, but keeping it precise:

1. **There is no separate `langgraph` container.** The Compose services are
   `nginx / frontend / gateway / provisioner`; only `gateway` mounts the
   socket. The LangGraph runtime is embedded *inside* the gateway process, not
   a distinct container.
2. **CVSS 9.8 / `AV:N` overstates it.** Mounting the socket is not a
   network-triggerable vuln on its own — it is a privilege-escalation amplifier
   that requires a separate code-execution precondition (which the report
   itself notes). Real severity is conditional on sandbox mode:
   - `local` mode — agent `bash` runs *inside* the gateway container, so with
     the socket present a prompt injection → host escape is fairly direct
     (still needs the injection vector).
   - `aio` (Docker sandbox) mode — agent commands run in an isolated sandbox
     container without the socket; only the gateway holds it.

   None of this lowers the priority — it's a real hardening gap — but the fix
   and its scoping should reflect the accurate model rather than a standalone
   AV:N 9.8.

## Proposed fix

Make the socket mount **opt-in for `aio` (pure-DooD) mode only**, via a small
`docker/docker-compose.dood.yaml` overlay that `scripts/deploy.sh` /
`scripts/docker.sh` append automatically when `detect_sandbox_mode()` returns
`aio`. The default (`local`) and `provisioner`/Kubernetes modes then never
expose the host daemon. `SECURITY.md` documents the DooD threat model.

PR #3517 implements this with real-Docker validation evidence.

## Validation (local, real Docker)

| Scenario | `/var/run/docker.sock` in gateway | `docker ps` inside container |
|---|---|---|
| Default `local` (no overlay) | absent (`No such file`) | fails — cannot reach host daemon |
| `aio` (with overlay) | present (`srw-rw---- … docker.sock`) | succeeds — lists host containers (DooD escape reproduced) |

`detect_sandbox_mode()` resolves `local → local`, `aio → aio` (overlay added),
`aio + provisioner_url → provisioner` (no socket), missing config → `local`.

## Out of scope (noted, not fixed here)

The gateway also bind-mounts `$HOME/.claude` and `$HOME/.codex` (read-only) for
CLI auto-auth in all modes; these hold long-lived CLI credentials. Documented
in `SECURITY.md` as a hardening note; left for a separate change to avoid
breaking auto-auth.

fancyboi999:
## Problem

The default Docker Compose stack mounts the host Docker socket
(`/var/run/docker.sock`) read-write into the root `gateway` container **in every
sandbox mode**, including the default `local` mode that never uses it. Since the
gateway executes model-generated commands, any in-container code-execution
(prompt injection, tool/MCP misuse, future RCE) can use the socket to gain
root-equivalent control of the host. This is an unnecessary default attack
surface — `scripts/deploy.sh` already gates the socket *check* on
`sandbox_mode != local`, but the Compose files mounted it unconditionally.

Reported by **@greatmengqi** — thank you. 🙏 (Original report's "two containers
(gateway, langgraph)" and CVSS 9.8/AV:N are addressed in the linked issue: there
is only the `gateway` container, and this is a privilege-escalation amplifier,
not a standalone network-triggerable critical.)

## Root cause

Only `aio` (AioSandboxProvider without `provisioner_url`, i.e. pure
Docker-out-of-Docker) needs the host socket. `local` (default) runs commands in
the gateway container's own filesystem; `provisioner`/Kubernetes talks to the
provisioner over HTTP. Both run fine without the socket — but it was mounted for
all of them.

## Solution

- Add `docker/docker-compose.dood.yaml` — an **opt-in** overlay that mounts the
  socket onto the `gateway` service (works for both prod and dev compose).
- Remove the unconditional socket mount from `docker/docker-compose.yaml` and
  `docker/docker-compose-dev.yaml`.
- `scripts/deploy.sh` and `scripts/docker.sh` append the overlay **only** when
  `detect_sandbox_mode()` returns `aio`, and the socket existence check is
  tightened from `!= local` to `== aio` (so `provisioner` mode no longer
  demands a socket it doesn't use).
- `SECURITY.md` documents the DooD threat model and the per-mode isolation
  table.

Scope is deliberately limited to the socket. The root `USER` and the
`$HOME/.claude` / `$HOME/.codex` credential mounts are noted in `SECURITY.md`
but left untouched (changing them risks breaking auto-auth / volume ownership
and belongs in separate changes). A Docker API proxy is intentionally not added:
once a container can `containers/create` with bind mounts it can still escape,
so a proxy gives limited benefit over simply not mounting the socket by default.

## Validation (local, real Docker — Docker Desktop 28.3.3)

Ran a throwaway `docker:cli` container against the **project's real**
`docker-compose.dood.yaml` overlay:

```
Scenario A — default local (no overlay):
  $ ls -l /var/run/docker.sock     → No such file or directory
  $ docker ps                      → cannot connect to /var/run/docker.sock   ✅ no escape

Scenario B — aio (base + docker-compose.dood.yaml):
  $ ls -l /var/run/docker.sock     → srw-rw---- 1 root root … /var/run/docker.sock
  $ docker ps                      → lists host containers                     ✅ socket works
```

`docker compose config` rendering (authoritative engine merge):

```
prod/dev × local : gateway has NO docker.sock volume
prod/dev × aio   : docker.sock volume present
```

`detect_sandbox_mode()` truth table:

```
local                       → local        (no socket)
aio (no provisioner_url)    → aio          (overlay appended)
aio + provisioner_url       → provisioner  (no socket, HTTP)
missing config.yaml         → local        (safe default)
```

`bash -n` clean on both scripts. `shellcheck` clean on the added lines (one
pre-existing `SC2086` on the unrelated `$services` line in `docker.sh` is left
as-is — not introduced here).

## Notes

- `make build` / `deploy.sh down` paths are unaffected (they run before
  `detect_sandbox_mode` and never append the overlay).
- Manual DooD without the scripts: `docker compose -f docker-compose.yaml -f docker-compose.dood.yaml up -d`.

Closes #3516

/cc @greatmengqi — would appreciate your review since you reported the original issue.


SECURITY.md
```diff
@@ -10,3 +10,42 @@ Currently, we have two branches to maintain:
 ## Reporting a Vulnerability
 
 Please go to https://github.com/bytedance/deer-flow/security to report the vulnerability you find.
+
+## Sandbox Isolation and the Docker Socket (DooD)
```
WillemJiang, line 14: @fancyboi999  Current SECURITY.md is used for the security policy with GitHub.  Please update this information in the backend/docs/CONFIGURATION.md file.

WillemJiang (approved):

PR #3464, opened 2026-06-09
fix(sandbox): persist lazily-acquired sandbox state via Command

Huixin615:
Fixes #3463

## Why

While using the sandbox, `ensure_sandbox_initialized` is invoked and
acquires a sandbox successfully, but the freshly-acquired `sandbox_id`
cannot be observed from `state["sandbox"]` in any subsequent graph step.

Root cause: `ensure_sandbox_initialized` /
`ensure_sandbox_initialized_async` in `deerflow/sandbox/tools.py`
persist the sandbox id by **mutating `runtime.state["sandbox"]` in
place**. LangGraph's state is an immutable snapshot that can only be
updated by returning a `dict` or `Command(update=...)` so the channel
reducer can merge it. In-place mutation is local to the current tool
invocation and is silently dropped — so:

- subsequent graph steps don't see the sandbox id
- downstream consumers (`ToolOutputBudgetMiddleware`, sub-agent
  `task_tool`, checkpoint/resume) cannot read the sandbox from state
- the sandbox may be unnecessarily re-acquired

The reporter asked for a fix that goes through "a proper middleware /
update path", which matches the approach taken here.

## What changed

- `SandboxMiddleware` now overrides the official `wrap_tool_call` /
  `awrap_tool_call` hooks. It snapshots `runtime.state["sandbox"]
  ["sandbox_id"]` before the handler runs; if it transitions from
  `None` to a value, the handler triggered a lazy init and the wrapper
  emits a proper `Command(update={"sandbox": {"sandbox_id": <id>}})` so
  the channel reducer persists it.
- Three handler return shapes are handled explicitly:
  - `ToolMessage` → wrapped into
    `Command(update={"sandbox": ..., "messages": [msg]})`
  - `Command` with a dict `update` → merge the `sandbox` key while
    preserving `messages` / `goto` / `graph` / `resume`
  - `Command` with a non-dict `update` (e.g. list of tuples) →
    passthrough unchanged, to avoid silent data loss on unknown shapes
- If sandbox id was already in state (`prev is not None`), the wrapper
  short-circuits — no duplicate emits.
- The production `ensure_sandbox_initialized` body is intentionally
  not touched. The fix is confined to the middleware, keeping the
  change surface minimal and avoiding ripple effects across the ~8
  tool call-sites that depend on it.

From a caller's perspective: after this PR, any tool that triggers
lazy sandbox initialization will correctly propagate the
`sandbox.sandbox_id` to graph state, and all downstream middleware /
[...]

PR #3518, opened 2026-06-12
fix(sandbox): merge idempotent sandbox state updates

liuchuan01:
## Why

  #3464 fixed missing sandbox state persistence by having sandbox tool calls emit `Command(update={"sandbox": ...})` when lazy initialization happens.
  I also report this bug in #3464 comment
  That solves the original problem, but it also exposed a follow-up concurrent update case: when multiple sandbox tools are triggered in the same LangGraph step, more than one tool
  can now emit the same `sandbox` state update.

  In that case, the sandbox tool calls themselves may succeed, but LangGraph rejects the run while merging graph state:

  ``` text
  InvalidUpdateError: At key 'sandbox': Can receive only one value per step.
  Use an Annotated key to handle multiple values.
 ```
  This PR adds the missing reducer for ThreadState.sandbox, so identical lazy-init sandbox updates are treated as idempotent instead of failing the run. It still fails fast if
  different sandbox_id values are merged, because that would indicate a sandbox lifecycle or isolation bug.

  ## What changed

  ThreadState.sandbox now has an explicit reducer.

  The reducer treats repeated writes of the same sandbox_id as idempotent and merges them safely. If two different sandbox_id values appear in the same state merge, it fails fast
  instead of silently choosing one, since that would indicate a sandbox lifecycle or isolation bug.

  Regression tests cover:

  - preserving existing sandbox state when a node does not update it
  - accepting a first sandbox update
  - merging identical concurrent sandbox_id updates
  - rejecting conflicting sandbox ids
  - ensuring ThreadState.sandbox remains wired to the reducer

  ## Surface area

  - [ ] Frontend UI — page / component / setting / interaction under frontend/
  - [ ] Backend API — endpoint / SSE event / request-response shape under backend/app
  - [x] Agents / LangGraph — agent node, graph wiring, langgraph.json, or prompt change
  - [x] Sandbox — docker/ or sandboxed execution
  - [ ] Skills — change under skills/
  - [ ] Dependencies — new/upgraded entry in backend/pyproject.toml or frontend/package.json (say what it buys us)
  - [ ] Default behavior change — changes existing behavior without the user opting in (default model, default setting, data shape)
  - [ ] Docs / tests / CI only — no runtime behavior change

  ## Screenshots / Recording

  N/A. No frontend UI changes.

  ## Bug fix verification

  backend/tests/test_thread_state_reducers.py

  The new reducer tests verify that duplicate same-id sandbox updates are accepted and conflicting sandbox ids fail closed.

  I also verified the runtime behavior with a minimal LangGraph graph where two parallel nodes write the same sandbox_id: it now merges successfully. A conflicting-id variant raises
  the expected ValueError.

  ## Validation

  cd backend && uv run pytest tests/test_thread_state_reducers.py -q
  17 passed

  cd backend && uv run pytest tests/test_sandbox_middleware.py -q
  17 passed

  Pre-commit also passed for the committed files:

  ruff lint
  ruff format

  ## AI assistance

  Tool(s) used: Codex

  How you used it: Investigated the LangGraph sandbox state merge failure, proposed the reducer strategy, implemented the reducer and regression tests, and ran focused validation. I
  reviewed the change and test behavior before submitting.


liuchuan01:
@WillemJiang since this appears to be a follow-up edge case from #3464. You have the most context on the lazy sandbox state persistence change, so your review would be
  especially helpful.

WillemJiang:
@liuchuan01  Here are some comments for the PR

1. Minor: Consider using pytest.raises instead of try/except in the conflict test.
The existing tests in this file use a mix, but pytest.raises(ValueError, match="Conflicting") would be more idiomatic and concise:

```
def test_conflicting_sandbox_ids_raise(self):
    with pytest.raises(ValueError, match="Conflicting sandbox state updates"):
        merge_sandbox({"sandbox_id": "sandbox-1"}, {"sandbox_id": "sandbox-2"})
```

2. Minor: The None == None edge case deserves a test. When both existing and new have sandbox_id set to None (or omitted entirely — both produce None from .get()), the reducer returns existing. This is correct, but an explicit test would document the intent:

```
def test_both_none_sandbox_id_is_idempotent(self):
    assert merge_sandbox({"sandbox_id": None}, {"sandbox_id": None}) == {"sandbox_id": None}
```

3. Consider: Should the reducer also handle SandboxState({}) (key omitted entirely)?
Since sandbox_id is NotRequired, SandboxState() is valid. existing.get("sandbox_id") returns None in that case, so it merges identically to sandbox_id=None. This is fine, but worth confirming this is the intended behavior for "sandbox not yet initialized."

liuchuan01:
Thanks for the review. I updated the tests accordingly:

  - Replaced the manual `try/except` conflict assertion with `pytest.raises(..., match=...)`.
  - Added coverage for both `{"sandbox_id": None}` updates.
  - Added coverage for omitted `sandbox_id` via `{}` updates to document that this represents uninitialized sandbox state and is treated idempotently.

WillemJiang (approved):

PR #3519, opened 2026-06-12
fix(frontend): reset active chat after deletion

zwj110610:
## Why

Deleting the currently open chat could remove it from the sidebar while the main chat panel still displayed the deleted conversation. This was especially visible after creating a new chat, sending a message, waiting for the response, and then deleting that active chat without selecting another conversation.

The fix ensures the active chat view immediately returns to the new chat screen once the remote thread deletion succeeds, without waiting for local cleanup to finish.

## What changed

- Active chat deletion now resets the chat page state and navigates back to the new chat screen.
- The chat state hook now listens for active-thread deletion events and creates a fresh empty thread state.
- Deleted or inaccessible thread URLs now redirect back to the new chat screen on refresh.
- The same stale-thread redirect behavior is applied to agent chat routes.

## Surface area

## Screenshots / Recording

Before deleting the conversation
bug4
After deleting the conversation
bug5

## Bug fix verification

- Test: `deleting the active newly created chat returns to the new chat screen`

The test creates a new chat, sends a message, waits for the response, deletes the active chat from the sidebar, and verifies the main panel immediately returns to the new chat screen without showing the deleted conversation. It also verifies that navigating back to the deleted thread URL redirects to `/workspace/chats/new`.

The test additionally simulates local cleanup failure after the remote thread deletion succeeds, to ensure the UI reset is not blocked by `/api/threads/{id}` cleanup.
## Validation

- `prettier --check` on changed frontend files
- `eslint` on changed frontend files
- `tsc --noEmit`

Note: local `git commit` used `--no-verify` because this machine does not have the `pre-commit` command installed; the frontend checks above were run manually.

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Helped inspect the deletion/navigation flow, implement the frontend state reset, review the diff, and run local validation.

### DIFF
--- a/frontend/src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx
+++ b/frontend/src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx
@@ -25,7 +25,11 @@ import { useI18n } from "@/core/i18n/hooks";
 import { useModels } from "@/core/models/hooks";
 import { useNotification } from "@/core/notification/hooks";
 import { useLocalSettings, useThreadSettings } from "@/core/settings";
-import { useThreadStream, useThreadTokenUsage } from "@/core/threads/hooks";
+import {
+  useThreadMetadata,
+  useThreadStream,
+  useThreadTokenUsage,
+} from "@/core/threads/hooks";
 import { threadTokenUsageToTokenUsage } from "@/core/threads/token-usage";
 import { textOfMessage } from "@/core/threads/utils";
 import { env } from "@/env";
@@ -54,6 +58,10 @@ export default function AgentChatPage() {
     isNewThread || isMock ? undefined : threadId,
     { enabled: tokenUsageEnabled && !isMock },
   );
+  const threadMetadata = useThreadMetadata(threadId, {
+    enabled: !isNewThread && !isMock,
+    isMock,
+  });
   const backendTokenUsage = threadTokenUsageToTokenUsage(threadTokenUsage.data);
 
   const { showNotification } = useNotification();
@@ -105,6 +113,34 @@ export default function AgentChatPage() {
       }
     },
   });
+
+  const hasThreadMessages = thread.messages.length > 0;
+
+  useEffect(() => {
+    if (
+      !isNewThread &&
+      !isMock &&
+      threadMetadata.data === null &&
+      !threadMetadata.isLoading &&
+      !threadMetadata.isFetching &&
+      !isHistoryLoading &&
+      !hasMoreHistory &&
+      !hasThreadMessages
+    ) {
+      router.replace(`/workspace/agents/${agent_name}/chats/new`);
+    }
+  }, [
+    agent_name,
+    hasMoreHistory,
+    hasThreadMessages,
+    isHistoryLoading,
+    isMock,
+    isNewThread,
+    router,
+    threadMetadata.data,
+    threadMetadata.isFetching,
+    threadMetadata.isLoading,
+  ]);
 
   const handleSubmit = useCallback(
     (message: PromptInputMessage) => {
--- a/frontend/src/app/workspace/chats/[thread_id]/page.tsx
+++ b/frontend/src/app/workspace/chats/[thread_id]/page.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { useRouter } from "next/navigation";
 import { useCallback, useEffect, useRef, useState } from "react";
 
 import { type PromptInputMessage } from "@/components/ai-elements/prompt-input";
@@ -24,7 +25,11 @@ import { useI18n } from "@/core/i18n/hooks";
 import { useModels } from "@/core/models/hooks";
 import { useNotification } from "@/core/notification/hooks";
 import { useLocalSettings, useThreadSettings } from "@/core/settings";
-import { useThreadStream, useThreadTokenUsage } from "@/core/threads/hooks";
+import {
+  useThreadMetadata,
+  useThreadStream,
+  useThreadTokenUsage,
+} from "@/core/threads/hooks";
 import { threadTokenUsageToTokenUsage } from "@/core/threads/token-usage";
 import { textOfMessage } from "@/core/threads/utils";
 import { env } from "@/env";
@@ -32,6 +37,7 @@ import { cn } from "@/lib/utils";
 
 export default function ChatPage() {
   const { t } = useI18n();
+  const router = useRouter();
   const { threadId, setThreadId, isNewThread, setIsNewThread, isMock } =
     useThreadChat();
   // `isNewThread` tracks whether the backend has the thread yet — gates the
@@ -47,6 +53,10 @@ export default function ChatPage() {
     isNewThread || isMock ? undefined : threadId,
     { enabled: tokenUsageEnabled && !isMock },
   );
+  const threadMetadata = useThreadMetadata(threadId, {
+    enabled: !isNewThread && !isMock,
+    isMock,
+  });
   const backendTokenUsage = threadTokenUsageToTokenUsage(threadTokenUsage.data);
   const mountedRef = useRef(false);
   useSpecificChatMode();
@@ -107,6 +117,33 @@ export default function ChatPage() {
       }
     },
   });
+
+  const hasThreadMessages = thread.messages.length > 0;
+
+  useEffect(() => {
+    if (
+      !isNewThread &&
+      !isMock &&
+      threadMetadata.data === null &&
+      !threadMetadata.isLoading &&
+      !threadMetadata.isFetching &&
+      !isHistoryLoading &&
+      !hasMoreHistory &&
+      !hasThreadMessages
+    ) {
+      router.replace("/workspace/chats/new");
+    }
+  }, [
+    hasMoreHistory,
+    hasThreadMessages,
+    isHistoryLoading,
+    isMock,
+    isNewThread,
+    router,
+    threadMetadata.data,
+    threadMetadata.isFetching,
+    threadMetadata.isLoading,
+  ]);
 
   const handleSubmit = useCallback(
     (message: PromptInputMessage) => {
--- a/frontend/src/components/workspace/chats/use-thread-chat.ts
+++ b/frontend/src/components/workspace/chats/use-thread-chat.ts
@@ -5,6 +5,25 @@ import { useCallback, useEffect, useRef, useState } from "react";
 
 import { uuid } from "@/core/utils/uuid";
 
+export const THREAD_CHAT_RESET_EVENT = "deer-flow:thread-chat-reset";
+
+type ThreadChatResetDetail = {
+  deletedThreadId: string;
+  nextPath: string;
+  force?: boolean;
+};
+
+export function resetThreadChatAfterDelete(detail: ThreadChatResetDetail) {
+  if (typeof window === "undefined") {
+    return;
+  }
+  window.dispatchEvent(
+    new CustomEvent<ThreadChatResetDetail>(THREAD_CHAT_RESET_EVENT, {
+      detail,
+    }),
+  );
+}
+
 export function useThreadChat() {
   const { thread_id: threadIdFromPath } = useParams<{ thread_id: string }>();
   const pathname = usePathname();
@@ -30,6 +49,13 @@ export function useThreadChat() {
     () => threadIdFromPath === "new",
   );
 
+  const resetToNewThread = useCallback(() => {
+    const nextThreadId = uuid();
+    newThreadIdRef.current = nextThreadId;
+    setIsNewThreadState(true);
+    setThreadIdState(nextThreadId);
+  }, []);
+
   useEffect(() => {
     if (isNewPath) {
       const nextThreadId = newThreadIdRef.current ?? uuid();
@@ -50,6 +76,35 @@ export function useThreadChat() {
     setIsNewThreadState(false);
     setThreadIdState(threadIdFromPath);
   }, [isNewPath, threadIdFromPath]);
+
+  useEffect(() => {
+    const handleReset = (event: Event) => {
+      const detail = (event as CustomEvent<ThreadChatResetDetail>).detail;
+      if (!detail?.nextPath) {
+        return;
+      }
+
+      const currentPathname = window.location.pathname;
+      const isDeletingCurrentThread =
+        detail.force === true ||
+        detail.deletedThreadId === threadId ||
+        detail.deletedThreadId === threadIdFromPath ||
+        currentPathname.endsWith(`/${detail.deletedThreadId}`);
+
+      if (!isDeletingCurrentThread) {
+        return;
+      }
+
+      // URL replacement is owned by the caller's Next router action; this hook
+      // only resets local chat state so the router state and browser URL stay
+      // in sync.
+      resetToNewThread();
+    };
+
+    window.addEventListener(THREAD_CHAT_RESET_EVENT, handleReset);
+    return () =>
+      window.removeEventListener(THREAD_CHAT_RESET_EVENT, handleReset);
+  }, [resetToNewThread, threadId, threadIdFromPath]);
 
   const setThreadId = useCallback((nextThreadId: string) => {
     newThreadIdRef.current = null;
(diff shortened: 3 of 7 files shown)


zwj110610:
> Nice fix for the active-chat case — the reset-to-new-chat flow and the 403/404-only `isThreadMissingError` guard (so a transient 5xx doesn't bounce you off a valid thread) both look right. One thing before it lands, though: deleting a _non-active_ conversation looks like it can drag you off the chat you're currently in.
> 
> `resetThreadChatAfterDelete()` always runs `window.history.replaceState(null, "", detail.nextPath)`, and `handleDelete`'s `onSuccess` (in `recent-chat-list.tsx`) calls it for **every** deletion, not just the active one — `onRemoteDeleted` is gated on `isCurrentThread`, but the `onSuccess` reset isn't:
> 
> ```ts
> onSuccess: () => {
>   resetThreadChatAfterDelete({
>     deletedThreadId: thread.thread_id,
>     nextPath: nextThreadPath,   // always /workspace/chats/new
>   });
> },
> ```
> 
> So if I'm viewing thread A and delete thread B from the sidebar, the `CustomEvent` listener correctly bails (`isDeletingCurrentThread` is false, so no state reset) — but the `replaceState` inside `resetThreadChatAfterDelete` still fires unconditionally and rewrites the address bar from `/workspace/chats/A` to `/workspace/chats/new`. And since `useThreadChat` derives `isNewThread`/`threadId` from `window.location.pathname` at render time, the next re-render then sees `isNewPath === true` and resets the _open_ chat to a fresh new thread — so deleting an unrelated background conversation can kick me out of the one I'm actually in.
> 
> The e2e only covers deleting the active chat, so this path isn't exercised. Simplest fix is to gate the `onSuccess` reset on `isCurrentThread` the same way `onRemoteDeleted` is (or make the `replaceState` in `resetThreadChatAfterDelete` conditional on the deleted thread being the current one) — for a non-active delete, nothing about the current view should change.
> 
> 有问题或异议欢迎 ping [@fancyboi999](https://github.com/fancyboi999) — the human behind this account.

Thank you for your review. The issue has been fixed.
The helper no longer calls `replaceState` unconditionally. It now only dispatches the reset event, and the URL/state reset happens inside the listener after confirming the deleted thread is the active one.

Added an e2e regression test for deleting an inactive sidebar chat while keeping the current chat open.

Validated with prettier, eslint, typecheck, and `thread-history.spec.ts` passing.

WillemJiang:
The e2e tests have failed. @zwj110610, could you double-check them out?

zwj110610:
> The e2e tests have failed. @zwj110610, could you double-check them out?

Thanks for the heads-up! I pushed a follow-up commit to address the e2e failures.

The failures came from two frontend test/runtime paths:

the mock thread history could overwrite the streamed Hello from DeerFlow! response after a newly-created chat was persisted in the mock thread list;
the missing-thread redirect was too eager and could redirect before the real-backend run history had a chance to hydrate.
I updated the mock stream/thread history data and made the missing-thread redirect wait until metadata/history loading is complete and there are no messages to display.

I also reran the CI-aligned validation locally with Node 22, pnpm 10.26.2, Python 3.12, and uv:

backend/frontend lint
frontend unit tests
backend unit tests with CI=true
frontend e2e: 40 passed
replay Layer 1: 1 passed
replay Layer 2: 3 passed
GitHub Actions have also been rerun and are green now.

willem-bd (commented):
## Overview

Fixes the case where deleting the currently-active chat removed it from the sidebar but left the deleted conversation in the main panel. The fix:

1. `useDeleteThread` gains an `onRemoteDeleted` callback that fires **after** the remote LangGraph delete but **before** the local `/api/threads/{id}` cleanup — so the UI can reset even if local cleanup fails.
2. `recent-chat-list.tsx` detects when the deleted thread is the current view and, on remote-delete success, dispatches a global reset event + `router.replace("…/new")`.
3. `use-thread-chat.ts` listens for the reset event, resets to a fresh empty-thread state, and does a raw `window.history.replaceState`.
4. Both chat pages add a `useThreadMetadata` hook; if the thread GET returns 403/404 (→ `null`), the page redirects to the new-chat screen on refresh.
5. E2E coverage added for inactive-delete (keeps current), active-delete (resets + tolerates local-cleanup failure), and stale-URL redirect.

The intent is good and the test coverage is solid. A few real concerns below — the History-API / Next-router mixing is the one I'd want addressed before merge.

## P1 — Mixing raw `window.history.replaceState` with the Next.js App Router

In `use-thread-chat.ts`, `handleReset` calls `window.history.replaceState(null, "", detail.nextPath)`, while `handleDelete` *also* calls `router.replace(nextThreadPath)` for the same target path. Mutating the browser History API directly behind the App Router is a known footgun: Next tracks its own router state separately from `window.location`, so they can desync, producing back-button breakage and occasional "router recovered from a bad state" behavior.

Recommendation: pick one mechanism. If the state reset needs to happen without a full route transition, do the state reset on the reset event and let a *single* `router.replace` own the URL change — drop the `window.history.replaceState`. If you truly need the raw replaceState, document why and don't also call `router.replace` on the same path.

## P2 — 403 is treated as "thread missing" → silent redirect, no feedback

`isThreadMissingError` returns true for both `403` and `404`. A 403 (forbidden / not owner) now silently redirects the user to the new-chat screen with no message. If the intent is to avoid leaking thread existence, that's defensible — but the UX cost is that a user who legitimately lacks access just sees themselves bounced to a blank chat with no explanation. Consider a brief toast/notification for the 403 path (404 can stay silent). At minimum, worth a code comment noting the 403 is intentional.

## P2 — `useThreadHistory` `hasMore`/`loading` change is broader than the deletion fix

`hasMore` and `loading` now incorporate run-loading/unresolved states for **all** threads, not just deleted ones:

```ts
const hasMore =
  enabled && hasThreadId &&
  (indexRef.current >= 0 || hasUnloadedRuns || isRunsLoading || isRunsUnresolved);
...
loading: loading || isRunsLoading,
```

This is plausibly needed so the new redirect effect (`!hasMoreHistory && !hasThreadMessages`) doesn't fire prematurely while runs are still loading. But it changes the "Load more" / infinite-scroll gating for every chat, and the new e2e tests don't directly exercise long-history pagination. This is bundled into a PR titled "reset active chat after deletion." Suggest adding (or pointing to) coverage for the "Load more" button visibility on a multi-run history, or splitting this into its own change.

## P2 — `onSuccess` dispatch of `resetThreadChatAfterDelete` looks like dead code

Tracing the paths:
- **Current-thread delete:** `onRemoteDeleted` fires with `force: true` and resets the URL to `/new`. By the time `onSuccess` runs (after local cleanup), `handleReset`'s non-force guard (`deletedThreadId === threadId` / path suffix) no longer matches → no-op.
- **Inactive-thread delete:** `onRemoteDeleted` is `undefined`; `onSuccess` dispatches without `force`, but `deletedThreadId` ≠ current thread → no-op.

So the `onSuccess` `resetThreadChatAfterDelete` call doesn't appear to fire meaningfully in any path. Either remove it, or leave a comment documenting the case where it's actually needed.

## P3 — Minor

- **Side-effect callback inside mutation args.** `onRemoteDeleted` embedded in `useMutation` input couples the data layer to UI behavior. Justified by the "reset must survive local-cleanup failure" requirement (since `onSuccess` runs after the whole `mutationFn`), but consider whether `onMutate` or splitting remote/local into separate steps reads cleaner. Code-smell only.
- **`getHttpStatus` message-string regex fallback.** `/\b(?:HTTP|status)\s*(\d{3})\b/i` parses a status out of free-text error messages — the `error.status` / `error.response.status` paths are robust; the string fallback adds a small risk of false positives for marginal gain. Prefer dropping it or tightening.
- **Extra request on every chat view.** `useThreadMetadata` adds a GET-thread to the critical path of every chat load, used only to detect the deleted/inaccessible case. Acceptable tradeoff for the refresh-redirect behavior, but consider whether the existing history/stream error could be piggybacked instead.

## Verified accurate

- ✅ `apiClient.threads.get` and `apiClient.threads.delete` both exist (LangGraph SDK; confirmed in `api-client.ts`).
- ✅ `onRemoteDeleted` ordering (after remote delete, before local cleanup) matches the stated goal — the e2e test that forces a 500 on local cleanup confirms the reset still happens.
- ✅ Deleting an inactive chat correctly leaves the current view open (matches the new "deleting an inactive chat keeps the current chat open" test).
- ✅ Redirect effect uses a strong conjunction of guards (`threadMetadata.data === null && !isLoading && !isFetching && !isHistoryLoading && !hasMoreHistory && !hasThreadMessages`) to avoid premature redirect.

## Note

The PR body states the commit used `--no-verify` (no local `pre-commit`). Frontend lint/format/tsc were run manually per the body; CI will be the authoritative check. Just flagging the local-gate gap.

### Summary

Good, well-tested fix for a real UX bug. Before merge I'd want: (1) the History-API/Next-router mixing resolved, (2) a decision on the 403 UX, and (3) confirmation that the `hasMore`/`loading` refactor doesn't regress long-history pagination. The rest are cleanups.

zwj110610:
> ## Overview
> Fixes the case where deleting the currently-active chat removed it from the sidebar but left the deleted conversation in the main panel. The fix:
> 
> 1. `useDeleteThread` gains an `onRemoteDeleted` callback that fires **after** the remote LangGraph delete but **before** the local `/api/threads/{id}` cleanup — so the UI can reset even if local cleanup fails.
> 2. `recent-chat-list.tsx` detects when the deleted thread is the current view and, on remote-delete success, dispatches a global reset event + `router.replace("…/new")`.
> 3. `use-thread-chat.ts` listens for the reset event, resets to a fresh empty-thread state, and does a raw `window.history.replaceState`.
> 4. Both chat pages add a `useThreadMetadata` hook; if the thread GET returns 403/404 (→ `null`), the page redirects to the new-chat screen on refresh.
> 5. E2E coverage added for inactive-delete (keeps current), active-delete (resets + tolerates local-cleanup failure), and stale-URL redirect.
> 
> The intent is good and the test coverage is solid. A few real concerns below — the History-API / Next-router mixing is the one I'd want addressed before merge.
> 
> ## P1 — Mixing raw `window.history.replaceState` with the Next.js App Router
> In `use-thread-chat.ts`, `handleReset` calls `window.history.replaceState(null, "", detail.nextPath)`, while `handleDelete` _also_ calls `router.replace(nextThreadPath)` for the same target path. Mutating the browser History API directly behind the App Router is a known footgun: Next tracks its own router state separately from `window.location`, so they can desync, producing back-button breakage and occasional "router recovered from a bad state" behavior.
> 
> Recommendation: pick one mechanism. If the state reset needs to happen without a full route transition, do the state reset on the reset event and let a _single_ `router.replace` own the URL change — drop the `window.history.replaceState`. If you truly need the raw replaceState, document why and don't also call `router.replace` on the same path.
> 
> ## P2 — 403 is treated as "thread missing" → silent redirect, no feedback
> `isThreadMissingError` returns true for both `403` and `404`. A 403 (forbidden / not owner) now silently redirects the user to the new-chat screen with no message. If the intent is to avoid leaking thread existence, that's defensible — but the UX cost is that a user who legitimately lacks access just sees themselves bounced to a blank chat with no explanation. Consider a brief toast/notification for the 403 path (404 can stay silent). At minimum, worth a code comment noting the 403 is intentional.
> 
> ## P2 — `useThreadHistory` `hasMore`/`loading` change is broader than the deletion fix
> `hasMore` and `loading` now incorporate run-loading/unresolved states for **all** threads, not just deleted ones:
> 
> ```ts
> const hasMore =
>   enabled && hasThreadId &&
>   (indexRef.current >= 0 || hasUnloadedRuns || isRunsLoading || isRunsUnresolved);
> ...
> loading: loading || isRunsLoading,
> ```
> 
> This is plausibly needed so the new redirect effect (`!hasMoreHistory && !hasThreadMessages`) doesn't fire prematurely while runs are still loading. But it changes the "Load more" / infinite-scroll gating for every chat, and the new e2e tests don't directly exercise long-history pagination. This is bundled into a PR titled "reset active chat after deletion." Suggest adding (or pointing to) coverage for the "Load more" button visibility on a multi-run history, or splitting this into its own change.
> 
> ## P2 — `onSuccess` dispatch of `resetThreadChatAfterDelete` looks like dead code
> Tracing the paths:
> 
> * **Current-thread delete:** `onRemoteDeleted` fires with `force: true` and resets the URL to `/new`. By the time `onSuccess` runs (after local cleanup), `handleReset`'s non-force guard (`deletedThreadId === threadId` / path suffix) no longer matches → no-op.
> * **Inactive-thread delete:** `onRemoteDeleted` is `undefined`; `onSuccess` dispatches without `force`, but `deletedThreadId` ≠ current thread → no-op.
> 
> So the `onSuccess` `resetThreadChatAfterDelete` call doesn't appear to fire meaningfully in any path. Either remove it, or leave a comment documenting the case where it's actually needed.
> 
> ## P3 — Minor
> * **Side-effect callback inside mutation args.** `onRemoteDeleted` embedded in `useMutation` input couples the data layer to UI behavior. Justified by the "reset must survive local-cleanup failure" requirement (since `onSuccess` runs after the whole `mutationFn`), but consider whether `onMutate` or splitting remote/local into separate steps reads cleaner. Code-smell only.
> * **`getHttpStatus` message-string regex fallback.** `/\b(?:HTTP|status)\s*(\d{3})\b/i` parses a status out of free-text error messages — the `error.status` / `error.response.status` paths are robust; the string fallback adds a small risk of false positives for marginal gain. Prefer dropping it or tightening.
> * **Extra request on every chat view.** `useThreadMetadata` adds a GET-thread to the critical path of every chat load, used only to detect the deleted/inaccessible case. Acceptable tradeoff for the refresh-redirect behavior, but consider whether the existing history/stream error could be piggybacked instead.
> 
> ## Verified accurate
> * ✅ `apiClient.threads.get` and `apiClient.threads.delete` both exist (LangGraph SDK; confirmed in `api-client.ts`).
> * ✅ `onRemoteDeleted` ordering (after remote delete, before local cleanup) matches the stated goal — the e2e test that forces a 500 on local cleanup confirms the reset still happens.
> * ✅ Deleting an inactive chat correctly leaves the current view open (matches the new "deleting an inactive chat keeps the current chat open" test).
> * ✅ Redirect effect uses a strong conjunction of guards (`threadMetadata.data === null && !isLoading && !isFetching && !isHistoryLoading && !hasMoreHistory && !hasThreadMessages`) to avoid premature redirect.
> 
> ## Note
> The PR body states the commit used `--no-verify` (no local `pre-commit`). Frontend lint/format/tsc were run manually per the body; CI will be the authoritative check. Just flagging the local-gate gap.
> 
> ### Summary
> Good, well-tested fix for a real UX bug. Before merge I'd want: (1) the History-API/Next-router mixing resolved, (2) a decision on the 403 UX, and (3) confirmation that the `hasMore`/`loading` refactor doesn't regress long-history pagination. The rest are cleanups.

Thanks for the careful review!

I pushed an update  addressing the main points:

- Removed the raw `window.history.replaceState` from the delete-reset event handler. The event now only resets local chat state, and the URL change is owned by the single `router.replace(...)` call from the delete flow.
- Removed the `onSuccess` `resetThreadChatAfterDelete` dispatch since, as you pointed out, it was not doing useful work in either active-thread or inactive-thread deletion paths.
- Kept the 403 -> blank-chat redirect behavior, but added a code comment documenting that this is intentional to avoid disclosing whether an inaccessible thread exists.
- Removed the `getHttpStatus` free-text message regex fallback.
- Narrowed the `useThreadHistory` change so run-loading/unresolved state affects `loading` only, not `hasMore`. That avoids broadening Load More / pagination visibility for ordinary long-history chats while still preventing the stale-thread redirect from firing too early.

I left `onRemoteDeleted` in place because the reset still needs to happen after the remote LangGraph delete succeeds but before local cleanup, so the UI reset survives local `/api/threads/{id}` cleanup failures.

Validation run locally:

- `prettier --check`
- `eslint`
- `tsc --noEmit`
- `next build`
- frontend unit tests: `288 passed`
- frontend e2e: `40 passed`
- backend `make lint`
- backend unit tests in CI mode: `4353 passed, 19 skipped`
- backend blocking-io tests: `15 passed`

WillemJiang (approved):

PR #3521, opened 2026-06-12
fix(security): do not bind-mount host CLI auth dirs by default

### LINKED ISSUE #3520: security: gateway bind-mounts entire ~/.claude and ~/.codex into container by default (history + credential exposure)
## Summary

The Compose stack bind-mounts the **entire** `~/.claude` and `~/.codex`
directories (read-only) into the root `gateway` container in **every**
configuration. This exposes far more than credentials: `~/.claude` typically
contains full conversation history (`history.jsonl`), per-project session data
(`projects/`), and the global `CLAUDE.md` — alongside the
`.credentials.json` token; `~/.codex` is similar. A gateway compromise (prompt
injection, tool/MCP misuse, RCE) leaks all of it.

The default setup does **not** need these mounts at all: the default model
providers are OpenAI-compatible (`api_key`) and the default sandbox is `local`.
The mounts are only needed when you opt into using the Claude/Codex CLI
subscription login as a model provider (`ClaudeChatModel` / Codex provider) or
for ACP agents — all opt-in config.

Thanks to **@greatmengqi**, whose report (see #3516) flagged the
`$HOME/.claude` / `$HOME/.codex` read-only mounts alongside the Docker socket
issue. 🙏 This issue tracks the credential-mount half; #3516 / #3517 cover the
Docker socket.

## Where

- `docker/docker-compose.yaml` and `docker/docker-compose-dev.yaml` — the
  gateway `volumes:` bind-mount `${HOME}/.claude` → `/root/.claude` and
  `${HOME}/.codex` → `/root/.codex` (read-only), unconditionally.

## Why it's more than "credential exposure"

A developer's `~/.claude` usually holds `history.jsonl` (all Claude Code
conversations), `projects/` (per-project sessions), and `CLAUDE.md` (global
instructions, which frequently name where other credentials live), plus the
`.credentials.json` token. Mounting the whole directory hands all of that to
the container, not just an auth token.

## Proposed fix

- Stop mounting the dirs by default.
- Add an opt-in `docker/docker-compose.cli-auth.yaml` overlay for setups that
  genuinely need the full CLI config (e.g. ACP adapters running the CLI
  in-container, which inherit `HOME=/root` and read the whole dir).
- Document env-token paths in `.env.example`: the Gateway credential loader
  reads `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_AUTH_TOKEN` (and `CODEX_AUTH_PATH`
  for a single file) **before** the default credential files, so model-provider
  setups need no directory mount at all.
- Document the threat model and per-mode options in `SECURITY.md`.

PR #3521 implements this.

## Note (breaking, with migration)

This changes a default. Setups currently relying on the auto-mounted
`~/.claude` / `~/.codex` for a Claude/Codex model provider must, after upgrade,
either set `CLAUDE_CODE_OAUTH_TOKEN` (etc.) in `.env` — recommended — or opt in
with `-f docker-compose.cli-auth.yaml`. Unlike the Docker socket (which keys off
the reliable `detect_sandbox_mode`), credential need can't be reliably inferred
from config, so this is explicit opt-in rather than auto-detected.

## Validation (local, real Docker)

| Scenario | `/root/.claude` in gateway |
|----------|----------------------------|
| Default (no overlay) | absent (`No such file`) |
| With `cli-auth.yaml` overlay | present, read-only (writes rejected) |

`docker compose config` confirms the default prod/dev stacks render **zero**
`.claude`/`.codex` mounts, and the overlay adds them back on `/root/.claude`
and `/root/.codex`.

fancyboi999:
## Problem

The Compose stack bind-mounts the **entire** `~/.claude` and `~/.codex`
directories (read-only) into the root `gateway` container in **every**
configuration. That exposes not just credentials but full conversation history
(`history.jsonl`), per-project session data (`projects/`), and global CLI config
(`CLAUDE.md`). The default OpenAI-compatible model providers and the local
sandbox never use these mounts — so it is an unnecessary default exposure, the
credential-mount counterpart to the Docker socket issue in #3516 / #3517.

Reported by **@greatmengqi** — the original report flagged the `$HOME/.claude` /
`$HOME/.codex` read-only mounts alongside the socket. 🙏

## Root cause

The mounts are only needed for opt-in setups: a Claude/Codex CLI login used as a
model provider (`ClaudeChatModel` / Codex provider, which read
`~/.claude/.credentials.json` / `~/.codex/auth.json`), or ACP agents that run the
CLI in-container (they inherit `HOME=/root` and read the whole dir). None of
these are on by default, yet the dirs were mounted for everyone.

## Solution

- Remove the unconditional `.claude` / `.codex` mounts from
  `docker/docker-compose.yaml` and `docker/docker-compose-dev.yaml`.
- Add an opt-in `docker/docker-compose.cli-auth.yaml` overlay (works with both
  prod and dev compose) for setups that need the full CLI config.
- `.env.example`: document `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_AUTH_TOKEN`
  and `CODEX_AUTH_PATH`. The Gateway credential loader
  (`credential_loader.py`) reads env vars **before** the default credential
  files, so model-provider setups need no bind mount at all.
- `SECURITY.md`: document the exposure and the per-mode least-exposure options.

This is explicit opt-in (not auto-detected like the socket): unlike sandbox
mode, "does this deployment need the CLI credential dir?" can't be reliably
inferred from `config.yaml`, so wiring it into the scripts would be guesswork.

## Validation (local, real Docker — Docker Desktop 28.3.3)

`docker compose config` (authoritative engine merge), exit 0:

```
prod/dev default            : gateway has ZERO .claude/.codex mounts
prod/dev + cli-auth.yaml    : mounts /root/.claude and /root/.codex (read-only)
```

Throwaway container against the **project's real** overlay:

```
Default (no overlay):   ls /root/.claude → No such file ; ls /root/.codex → No such file
With cli-auth overlay:  /root/.claude present; touch → "Read-only file system"  (read_only OK)
```

ACP note: `invoke_acp_agent_tool.py` spawns the adapter inheriting `HOME=/root`
with no `CLAUDE_CONFIG_DIR` override, so an in-container CLI *would* read
`/root/.claude` if it relied on the CLI login. In practice many ACP adapters
(e.g. claude-agent-acp) authenticate via a standard env API key
(`ANTHROPIC_API_KEY` / `OPENAI_API_KEY`) and need no mount at all — so the
overlay mounts the full dir only for adapters that genuinely read the CLI config
directory. The exact credential source is adapter-defined, not constrained by
DeerFlow.

## Breaking change + migration

Setups relying on the auto-mounted dirs for a Claude/Codex model provider must,
after upgrade, set `CLAUDE_CODE_OAUTH_TOKEN` (etc.) in `.env` (recommended) or
add `-f docker-compose.cli-auth.yaml`. Documented in `.env.example` and
`SECURITY.md`.

Closes #3520

/cc @greatmengqi — would appreciate your review; this is the credential-mount
half of your original report.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -32,8 +32,10 @@ with the least exposure that fits your setup:
 
 The Gateway credential loader checks environment variables **before** the
 default credential files, so the env-token paths need no bind mount at all. ACP
-adapters authenticate independently of DeerFlow — their credential source is
-adapter-defined, and some require a standard API key via env rather than the CLI
-subscription login. Prefer the adapter's documented env auth, and reach for the
-`docker-compose.cli-auth.yaml` overlay only when an adapter genuinely needs the
-full CLI config directory.
+adapters authenticate independently of DeerFlow via their own documented env —
+for example the common `claude-code-acp` adapter starts as
+`ANTHROPIC_API_KEY=… claude-code-acp` and honors `CLAUDE_CONFIG_DIR` to redirect
+its config directory, so it needs no `~/.claude` mount at all. Prefer the
+adapter's documented env auth, and reach for the
+`docker-compose.cli-auth.yaml` overlay only as a fallback for an adapter that
+genuinely reads the full CLI config directory.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -26,11 +26,14 @@ with the least exposure that fits your setup:
 
 | Need | How | Exposure |
 |------|-----|----------|
-| Claude model provider | env `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_AUTH_TOKEN` (via `.env`) | none — token only |
+| Claude model provider | env `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_AUTH_TOKEN` (via `.env`), or `CLAUDE_CODE_CREDENTIALS_PATH` → a single mounted `.credentials.json` | none / one file |
 | Codex model provider | env `CODEX_AUTH_PATH` pointing at a single mounted `auth.json` | one file |
-| ACP agent running the CLI in-container | opt-in `docker/docker-compose.cli-auth.yaml` overlay | full dir (read-only) |
+| ACP agent | the adapter's own auth — many ACP adapters take an env API key (e.g. `ANTHROPIC_API_KEY` / `OPENAI_API_KEY`) and need no mount; use the opt-in `docker/docker-compose.cli-auth.yaml` overlay only if your adapter reads the full CLI config dir | none / full dir |
 
 The Gateway credential loader checks environment variables **before** the
-default credential files, so the env-token paths need no bind mount at all. Use
-the `docker-compose.cli-auth.yaml` overlay only when the in-container CLI
-genuinely needs the full config directory (e.g. some ACP adapters).
+default credential files, so the env-token paths need no bind mount at all. ACP
+adapters authenticate independently of DeerFlow — their credential source is
+adapter-defined, and some require a standard API key via env rather than the CLI
+subscription login. Prefer the adapter's documented env auth, and reach for the
+`docker-compose.cli-auth.yaml` overlay only when an adapter genuinely needs the
+full CLI config directory.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -10,3 +10,27 @@ Currently, we have two branches to maintain:
 ## Reporting a Vulnerability
 
 Please go to https://github.com/bytedance/deer-flow/security to report the vulnerability you find.
+
+## CLI Credential Mounts (Claude Code / Codex)
+
+DeerFlow can reuse your Claude Code / Codex CLI subscription login as a model
+provider (`ClaudeChatModel`, the Codex provider) or for ACP agents that run the
+CLI in-container. The Compose stack used to bind-mount the **entire** `~/.claude`
+and `~/.codex` directories (read-only) into the gateway container in **every**
+configuration — exposing not just credentials but full conversation history,
+per-project session data, and global CLI config. A gateway compromise (prompt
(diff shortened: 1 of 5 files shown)


fancyboi999:
### ACP auth — verification backing the "no mount needed" guidance

To support the `SECURITY.md` note that ACP agents don't need the `~/.claude`
mount, I checked the common adapter's source (static evidence, not a live ACP
handshake):

- `@zed-industries/claude-code-acp@0.16.2` does **not** shell out to the `claude`
  CLI — it depends on `@anthropic-ai/claude-agent-sdk@0.2.44`.
- Its README documents startup as `ANTHROPIC_API_KEY=sk-... claude-code-acp`
  (env API key, no directory).
- The underlying SDK's `cli.js` also reads `CLAUDE_CODE_OAUTH_TOKEN`,
  `~/.claude/.credentials.json`, and OAuth, so a Claude subscription token works
  too — not only a raw API key.
- `claude-code-acp` honors `CLAUDE_CONFIG_DIR`, so even a file-based setup can
  point at a minimal directory instead of the whole `~/.claude`.

Conclusion: the opt-in `docker-compose.cli-auth.yaml` overlay is a fallback
only; env auth covers the common ACP adapters with zero directory exposure,
which is why this PR removes the default mount rather than keeping it for ACP.

SECURITY.md
```diff
@@ -10,3 +10,32 @@ Currently, we have two branches to maintain:
 ## Reporting a Vulnerability
 
 Please go to https://github.com/bytedance/deer-flow/security to report the vulnerability you find.
+
```
WillemJiang, line 13: @fancyboi999  As the SECURITY.md is used in GitHub for special usage, please move the below content to the backend/doc/CONFIGURE.md file.

WillemJiang (approved):

PR #3526, opened 2026-06-12
fix(channels): surface WeCom WebSocket connection failures (#2000)

18062706139fcz:
## What

Fixes #2000.

The WeCom channel started the SDK connection with a fire-and-forget `asyncio.create_task(self._ws_client.connect())` and never inspected the task result. As a result, connection failures (e.g. the gateway WebSocket handshake to `wss://openws.work.weixin.qq.com` failing) were silently swallowed:

- the channel still logged `WeCom channel started`, falsely reporting success;
- SDK `error` / `disconnected` events went unobserved, so the real cause never reached the logs;
- the unawaited connect task produced `Task exception was never retrieved` noise.

## How

- Add a done-callback on `_ws_task` that logs a clear, actionable error when the connect task fails (pointing users to check that the network/proxy allows `wss://openws.work.weixin.qq.com` and that `bot_id`/`bot_secret` are valid), and stays silent on cancellation (`stop()`) and on clean exit — eliminating the "Task exception was never retrieved" noise.
- Subscribe to the SDK's `error` / `disconnected` events in `start()` and translate them into DeerFlow's own readable logs.

This does not change the async/background-reconnect model; it only makes failures visible instead of silent.

> Note: the underlying `websockets` `AttributeError` (when `connection_lost` runs before `connection_made`) is a third-party library issue and is intentionally not touched here, but such exceptions are now captured by the task done-callback and surfaced as readable logs.

## Test

- `uv run ruff check` passes.
- Added 6 unit tests in `backend/tests/test_channels.py` covering: connect-task exception → error log, cancellation → silent, clean exit → silent, `error` event → error log, `disconnected` event → warning, and `start()` subscribing to the lifecycle events. Full `TestWeComChannel` suite (11 tests) passes.

### DIFF
--- a/backend/app/channels/wecom.py
+++ b/backend/app/channels/wecom.py
@@ -79,11 +79,33 @@ class WeComChannel(Channel):
             self._ws_client.on("message.mixed", self._on_ws_mixed)
             self._ws_client.on("message.image", self._on_ws_image)
             self._ws_client.on("message.file", self._on_ws_file)
+            self._ws_client.on("error", self._on_ws_error)
+            self._ws_client.on("disconnected", self._on_ws_disconnected)
             self._ws_task = asyncio.create_task(self._ws_client.connect())
+            self._ws_task.add_done_callback(self._on_ws_task_done)
 
             self._running = True
             self.bus.subscribe_outbound(self._on_outbound)
         logger.info("WeCom channel started")
+
+    def _on_ws_task_done(self, task: asyncio.Task) -> None:
+        if task.cancelled():
+            return
+        exc = task.exception()
+        if exc is None:
+            return
+        logger.error(
+            "WeCom WebSocket connection task failed: %s. "
+            "Check that the network/proxy allows wss://openws.work.weixin.qq.com "
+            "and that bot_id/bot_secret are valid.",
+            exc,
+        )
+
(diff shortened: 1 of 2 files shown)


WillemJiang:
@18062706139fcz 

 If the SDK passes a reason or context in the disconnect event, it's silently discarded. Consider logging at least the first arg if present:

```
def _on_ws_disconnected(self, *args: Any) -> None:
    detail = f" ({args[0]})" if args else ""
    logger.warning("WeCom WebSocket disconnected%s; SDK will attempt to reconnect", detail)

```

18062706139fcz:
> @18062706139fcz
> 
> If the SDK passes a reason or context in the disconnect event, it's silently discarded. Consider logging at least the first arg if present:
> 
> ```
> def _on_ws_disconnected(self, *args: Any) -> None:
>     detail = f" ({args[0]})" if args else ""
>     logger.warning("WeCom WebSocket disconnected%s; SDK will attempt to reconnect", detail)
> ```

@WillemJiang 

Thanks for the review! Good catch — I've updated _on_ws_disconnected to include the SDK-provided reason/context (the first event arg) in the warning log instead of discarding it, so disconnect causes are now visible. Also added a test covering the case where a reason is present. PTAL.

WillemJiang (approved):

PR #3534, opened 2026-06-12
feat(telegram): stream agent replies by editing the placeholder message in place

hetaoBackend:
## Summary
- Telegram channel now streams agent replies like Feishu: `CHANNEL_CAPABILITIES`/`supports_streaming` flipped so the manager routes telegram chats through the existing `runs.stream()` pipeline (no manager logic changes).
- The "Working on it..." placeholder is registered as the stream target; non-final updates `editMessageText` it in place with a 1s channel-side throttle, 4096-char truncation, and silent drops on 429 / not-modified (the guaranteed `is_final=True` message backstops completeness).
- The final update performs the last edit (guarded one-shot retry on 429), splits >4096 texts into follow-up messages delivered through the 3-retry send path, clears stream state, and keeps the threaded-reply chain (`_last_bot_message`) intact; direct sends without stream state behave exactly as before.
- Manager tests that used telegram as the non-streaming sample now use slack; docs (backend/CLAUDE.md) updated; design spec and implementation plan included under `docs/superpowers/`.

## Test Plan
- [x] `tests/test_channels.py` (175) + `tests/test_telegram_channel_connections.py` (2): 177 passed, 0 failed
- [x] New `TestTelegramStreaming` suite (15 tests): placeholder registration, in-place edits, throttling, truncation, 429 drop/retry/fallback, not-modified handling, >4096 overflow split with retries, no-state direct-send regression guard
- [x] `make lint` clean
- [x] Manual smoke test against a real Telegram bot (long-running prompt → placeholder morphs into streamed answer; >4096 answer splits into follow-ups)


backend/app/channels/telegram.py
```diff
@@ -5,6 +5,7 @@
 import asyncio
```
willem-bd, line 5: ## Code Review Findings

7 findings survived verification — 1 confirmed correctness bug, 2 confirmed/plausible maintenance hazards, and 4 plausible cleanup/robustness improvements.

---

### 1. 🔴 Unthreaded messages in `_send_stream_update` (CONFIRMED)

**File:** `backend/app/channels/telegram.py` ~line 142

`_send_stream_update` calls `bot.send_message` without `reply_to_message_id` in two code paths:
1. **State-is-None initial send** (placeholder failed) — sends an unthreaded message
2. **Edit-failure fallback** — sends a replacement message but unthreaded

Both should thread under the user's original message. The `key` parameter (`f"{chat_id}:{thread_ts}"`) contains the user's message id and could be extracted, or the thread_ts could be passed explicitly.

**Failure scenario:** Placeholder send fails (transient 429) → first streaming update creates a new unthreaded message. Later, `edit_message_text` fails (message deleted by admin) → fallback sends another unthreaded message. The user sees orphaned messages not nested under their original query, breaking conversation threading.

---

### 2. 🟡 Event name coupling in manager (CONFIRMED)

**File:** `backend/app/channels/manager.py` lines 1132 & 1149

`stream_mode=["messages-tuple", "values"]` and `if event in ("messages-tuple", "messages")` are coupled but maintained independently with no mechanical link. Adding a new stream mode to one without the other silently breaks streaming. Consider deriving the event check from `stream_mode` or using a shared constant.

---

### 3. 🟡 Divergent error classification in streaming vs final edit (PLAUSIBLE)

**File:** `backend/app/channels/telegram.py` ~lines 176 & 219

The `_is_not_modified` → `_is_retry_after` pattern is duplicated between `_send_stream_update` and `_edit_final_chunk` with different recovery strategies. Adding a new Telegram error class to one path but not the other causes behavioral inconsistency. A shared error classifier (`_classify_edit_error(exc) -> str`) would collapse the duplication.

---

### 4. 🟡 Orphaned `_stream_messages` entries (PLAUSIBLE)

**File:** `backend/app/channels/telegram.py` ~line 95

Entries are only removed on `is_final=True`. If the final outbound never arrives (asyncio task cancellation during `ChannelManager.stop()`, or error in the manager's `finally` block), entries leak permanently. Consider adding a TTL-based sweep or max-size cap to bound memory growth.

---

### 5. 🟢 Retry-loop duplication across 6 channels (PLAUSIBLE — reuse)

**File:** `backend/app/channels/telegram.py` ~line 215

The exponential-backoff retry loop (`_max_retries=3`, `delay=2**attempt`) is copy-pasted across feishu, slack, telegram, dingtalk, wechat, and wecom — ~30 lines of identical boilerplate per channel. A shared async helper in `base.py` would eliminate this.

---

### 6. 🟢 `_log_future_error` copy-pasted with divergent fixes (PLAUSIBLE — reuse)

**File:** `backend/app/channels/telegram.py` ~line 193

`_log_future_error` is copy-pasted across feishu, telegram, and dingtalk. DingTalk added `CancelledError` handling but feishu/telegram didn't — so bot restarts flood the log with spurious CancelledError tracebacks in those channels.

---

### 7. 🟢 `CHANNEL_CAPABILITIES` dual source of truth (PLAUSIBLE — robustness)

**File:** `backend/app/channels/manager.py` line 48

The static `CHANNEL_CAPABILITIES` dict duplicates the `supports_streaming` property on each `Channel` subclass. If `_channel_supports_streaming` is called before the channel service initializes (service is `None`), the static dict wins — updating one without the other silently changes behavior at startup.

hetaoBackend: Thanks — addressed in dd898301:

- **#1 (unthreaded fallback messages)**: fixed. `send()` now parses the user message id from `thread_ts` and passes it down; both the state-is-None initial send and the edit-failure fallback in `_send_stream_update` carry `reply_to_message_id`. Pinned by `test_stream_update_without_placeholder_sends_new_message` (extended) and the new `test_stream_edit_fallback_message_threads_under_user_message`.
- **#2 (event-name coupling)**: fixed. `STREAM_MODES` / `MESSAGE_STREAM_EVENTS` module constants now sit next to each other with a comment linking them; both call sites use the constants.
- **#4 (orphaned registry entries)**: bounded. Registrations go through `_register_stream_message`, which caps the registry at `MAX_TRACKED_STREAM_MESSAGES = 256` and evicts the oldest entry — same in-memory-tracking model as Feishu's card map, now with a hard bound. Test: `test_stream_message_registry_is_bounded`.
- **#3 (divergent error classification)**: keeping as-is. The classification predicates (`_is_not_modified` / `_is_retry_after`) are already shared helpers; what differs between the streaming and final paths is intentionally the *recovery policy* (drop vs wait-and-retry), which a shared classifier wouldn't collapse without obscuring that difference.
- **#5/#6/#7**: real, but pre-existing and cross-channel (all six channel impls) — tracked in #3536 rather than expanding this PR's blast radius.

hetaoBackend:
@fancyboi999 re the empty-final edge: agreed it's effectively unreachable — the manager's streaming path guarantees a non-empty final (`response_text` falls back to artifact text, the busy/error message, the last streamed text, or `"(No response from agent)"` in `_handle_streaming_chat`'s `finally` block), so `_finalize_stream_message` never sees `""` in practice. Left it without an explicit short-circuit to keep the channel code free of dead branches; if a second producer of final outbounds ever appears, the guard belongs in the manager invariant rather than per-channel.

Review findings from the other thread are addressed in dd898301 (threading on fallback sends, shared stream-event constants, bounded stream registry); cross-channel items spun out to #3536.

WillemJiang (approved):

PR #3551, opened 2026-06-13
fix(artifacts): offload blocking filesystem IO in artifact serving

ly-wang19:
## Why

`get_artifact` (`GET /api/threads/{id}/artifacts/{path}`) — the endpoint that serves agent-generated artifacts to the web UI — ran its entire filesystem workload directly on the event loop:

- virtual-path resolution (`resolve_thread_virtual_path` → `.resolve()` → `os.path.abspath`)
- `exists()` / `is_file()` probes
- MIME sniffing (`mimetypes.guess_type` lazily stats the system MIME database on first use)
- full-file `read_text()` / `read_bytes()` (loads the whole artifact into memory)
- `is_text_file_by_content()` (open + read)
- `.skill` archive `zipfile` open + member extract

So serving any artifact blocks the event loop for the full read. `make detect-blocking-io` flags 8 calls here — same class of issue as the offloads in #3457 / #3529.

## What changed

Offload each branch's filesystem work via `asyncio.to_thread`, one worker hop per branch:

- `_load_skill_archive_member(...)` — `.skill` branch: stat probes + ZIP open/extract + MIME sniff → returns `(content, mime_type)`.
- `_read_artifact_payload(...)` — regular branch: stat probes + MIME sniff + read → returns a `(kind, mime_type, payload)` plan (`"file"` lets `FileResponse` stream it; `"text"`/`"bytes"` carry the inline payload).
- The two `resolve_thread_virtual_path` calls are offloaded too.

`FileResponse` (explicit `download` / active HTML/SVG content) keeps streaming the file itself — Starlette handles that IO outside the handler. Branching, error codes (404/400/413), the active-content force-download rule, and the path-traversal security boundary are all unchanged.

> The runtime gate also caught a blocking call the static scanner missed: `resolve_thread_virtual_path`'s `.resolve()` (`os.path.abspath`). It's offloaded too.

## Surface area

- [x] **Backend** — `app/gateway/routers/artifacts.py` + new `tests/blocking_io` anchor. No API / schema / response change.

## Validation

- New `tests/blocking_io/test_artifacts_router.py` anchor (text / binary / `.skill` member) under the strict Blockbuster gate, verified **red→green** (removing the offload fails with `BlockingError`) → **3 passed**
- Full `tests/blocking_io` suite → **19 passed**
- Existing artifact behavior tests (`tests/test_artifacts_router.py`) → **10 passed**; artifact e2e (`test_client_e2e.py -k artifact`) → **5 passed**
- `make detect-blocking-io` → `artifacts.py` no longer reported
- `ruff check` + `ruff format --check` → clean

## AI assistance

**How you used it:** AI located the finding via `make detect-blocking-io`, implemented the per-branch `asyncio.to_thread` offload (folding resolve + stat + MIME + read into one worker hop), and wrote the gate anchor — which surfaced an extra blocking call (`resolve`'s `os.path.abspath`) the static scan had missed. I reviewed the change.

### DIFF
--- a/backend/app/gateway/routers/artifacts.py
+++ b/backend/app/gateway/routers/artifacts.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import mimetypes
 import zipfile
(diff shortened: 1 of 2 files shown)


fancyboi999 (commented):
Read through both files at `8806629c`. The offload looks correct and behavior-preserving — went branch by branch rather than skimming:

**Regular branch** — `_read_artifact_payload` (`artifacts.py:120-142`) reproduces the original ladder exactly: `exists`→404 (`:131`), `is_file`→400 (`:133`), then `download or mime in ACTIVE_CONTENT_MIME_TYPES`→`("file", …)` (`:136`), `text/*`→`read_text` (`:138`), `is_text_file_by_content`→`read_text` (`:140`), else `read_bytes` (`:142`). Collapsing `download` and active-content into one `"file"` kind is safe: both originally returned a `FileResponse` with `_build_attachment_headers`, and the handler still does exactly that at `:221-224`, so active content stays force-downloaded. Returning `payload=None` for the file kind and letting `FileResponse(path=actual_path, …)` stream it keeps large/active files off both the event loop and the worker's memory — good call not to `read_bytes` those into a buffer.

**`.skill` branch** — `_load_skill_archive_member` (`artifacts.py:100-116`) keeps the `exists`→404 (`:110`) / `is_file`→400 (`:112`) / extract→404 (`:115`) order and the MIME sniff, and the 413 size-cap inside `_extract_file_from_skill_archive` (`:54`, `:62`) now runs in the worker too. The in-memory text-decode + `UnicodeDecodeError` fallback stay on the loop, which is fine — they operate on already-extracted bytes, no FS IO.

**Error codes through `to_thread`** — every `HTTPException` (404/400/413, plus whatever `resolve_thread_virtual_path` raises for traversal, offloaded at `:192`/`:212`) is raised inside the worker and re-raised by `await asyncio.to_thread(...)`, so status codes survive the offload intact.

**Tests** — `test_artifacts_router.py` exercises all three kinds (text / binary / `.skill` member) through `get_artifact.__wrapped__` under the strict Blockbuster gate, so any IO regressing back onto the loop trips `BlockingError`. Solid anchors — the binary case deliberately uses a null byte to force the `read_bytes` path.

One non-blocking note: each branch now does two `to_thread` hops (resolve, then load/read) rather than one, since `actual_path` is needed back on the loop for the `logger.info` and the `FileResponse` path. Reasonable tradeoff, not worth folding together.

Looks good to merge.

WillemJiang (approved):

PR #3552, opened 2026-06-13
fix(mcp): offload blocking filesystem IO in MCP config update

ly-wang19:
## Why

`PUT /api/mcp/config` (`update_mcp_configuration`) — the endpoint the web UI calls to add/edit MCP servers — ran its whole filesystem workload on the event loop:

- resolve the extensions config path
- `config_path.exists()` probe
- `open(...)` + `json.load` to read the raw on-disk config
- `open(..., "w")` + `json.dump` to write the merged config
- `reload_extensions_config()` to re-read it into the cache

So every MCP config save blocks the event loop for the full read-modify-write. `make detect-blocking-io` flags it — same class as #3457 / #3529 / #3551.

## What changed

The admin check (`await _require_admin_user`) is async; everything after it — path resolve, existence probe, raw read, the secret-preserving merge, the write, and the reload — is a single contiguous block with no interleaved awaits. Move it into one `_apply_mcp_config_update(body)` helper dispatched via `asyncio.to_thread`, returning the reloaded server configs; the masked `McpConfigResponse` is built back on the loop.

The `$VAR`/secret-preserving merge (`_merge_preserving_secrets`), preservation of extra top-level keys (e.g. `mcpInterceptors`), error codes, and the stdio command allowlist (`_validate_mcp_update_request`) are all unchanged.

## Surface area

- [x] **Backend** — `app/gateway/routers/mcp.py` + new `tests/blocking_io` anchor. No API / schema / response change.

## Validation

- New `tests/blocking_io/test_mcp_router.py` anchor under the strict Blockbuster gate, verified **red→green** (removing the offload fails with `BlockingError: os.stat`) → **1 passed**
- Full `tests/blocking_io` suite → **17 passed**
- Existing MCP behavior tests (`test_mcp_client_config`, `test_mcp_config_secrets`, `test_mcp_custom_interceptors`, `test_mcp_oauth`, `test_mcp_session_pool`, `test_mcp_sync_wrapper`) → **87 passed**
- `make detect-blocking-io` → `mcp.py` no longer reported
- `ruff check` + `ruff format --check` → clean

## AI assistance

**How you used it:** AI located the finding via `make detect-blocking-io`, moved the contiguous read-modify-write into one `asyncio.to_thread` helper (keeping the async admin check + masked response on the loop), and wrote the gate anchor. I reviewed the change.


ly-wang19:
Good call — addressed in `e43c75d4`.

Offloading the read-modify-write to a thread did drop the within-process serialization the single-threaded event loop gave it for free, so I've guarded the offloaded RMW with a module-level `asyncio.Lock`: concurrent `PUT /api/mcp/config` calls now serialize within the process. Cross-process writers remain the separate, pre-existing concern you flagged (no file lock) — out of scope here.

Added a serialization regression test that's red without the lock (tracked max concurrency > 1) and green with it. `tests/blocking_io/test_mcp_router.py` → 2 passed; existing MCP behavior suites still green; ruff clean.

fancy-agent:
Confirmed — the `asyncio.Lock` fix is correct and well-anchored.

The lock at module level (`_mcp_config_write_lock = asyncio.Lock()`) is the right shape: it's acquired as `async with` on the event loop, so it serialises concurrent `PUT /api/mcp/config` calls *before* they reach `to_thread`. That restores the implicit serialisation the single-threaded event loop used to give you for free. Cross-process writers remain the pre-existing concern you flagged, and keeping them out of scope here is the right call.

The regression test is a good anchor: `state_lock` + the `active` counter correctly tracks max concurrency inside the worker thread (threading-safe), and the 20 ms sleep is enough to expose overlap without the lock. Red→green confirms the fix holds.

Looks good to merge on this front.

---
<sub>有问题或异议欢迎 ping [@fancyboi999](https://github.com/fancyboi999) — the human behind this account.</sub>

backend/app/gateway/routers/mcp.py
```diff
@@ -322,54 +393,14 @@ async def update_mcp_configuration(request: Request, body: McpConfigUpdateReques
...
-        with open(config_path, "w", encoding="utf-8") as f:
-            json.dump(config_data, f, indent=2)
-
-        logger.info(f"MCP configuration updated and saved to: {config_path}")
+        # Offload the blocking read-modify-write of extensions_config.json
+        # (path resolve, existence probe, raw read, merged write, reload). The
+        # lock serializes concurrent updates within this process so the RMW stays
+        # atomic now that it no longer runs inline on the event loop.
+        async with _mcp_config_write_lock:
+            reloaded_servers = await asyncio.to_thread(_apply_mcp_config_update, body)
```
willem-bd, line 401: **Defeats the offload: the duplicate `reload_extensions_config()` below (line 407) is still on the event loop, and it discards this `servers` build.**

`_apply_mcp_config_update()` already calls `reload_extensions_config()` inside the worker thread and returns `reloaded_servers` — and you build the masked `servers` here from it. But the next three lines (404–409, just below this hunk) reload a second time **on the event loop** and rebuild `servers` from that second reload, throwing away this dict. The second reload path goes through `ExtensionsConfig.from_file()` → `Path.exists()` + `open(..., encoding="utf-8")` + `json.load(...)` — the exact blocking IO this PR is meant to remove. `make detect-blocking-io` likely doesn't catch it because the blocking call now sits inside `reload_extensions_config()` rather than literally in `mcp.py`.

Besides re-blocking the loop, the two reloads create a window where the masked response can disagree with what was just written if anything mutates the cache between them.

Suggested fix — collapse to:

```python
async with _mcp_config_write_lock:
    reloaded_servers = await asyncio.to_thread(_apply_mcp_config_update, body)

reset_mcp_tools_cache()
servers = {
    name: _mask_server_config(McpServerConfigResponse(**server.model_dump()))
    for name, server in reloaded_servers.items()
}
return McpConfigResponse(mcp_servers=servers)
```

i.e. drop the trailing `reloaded_config = reload_extensions_config()` and the second `servers = ...` rebuild; keep `reset_mcp_tools_cache()` (cheap, in-memory).

backend/tests/blocking_io/test_mcp_router.py
```diff
@@ -0,0 +1,92 @@
...
+async def test_update_mcp_configuration_does_not_block_event_loop(tmp_path: Path, monkeypatch) -> None:
+    config_path = tmp_path / "extensions_config.json"
+    # resolve_config_path() requires the env-pointed file to exist; seed a minimal one.
+    await asyncio.to_thread(config_path.write_text, '{"mcpServers": {}, "skills": {}}', encoding="utf-8")
+    monkeypatch.setenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH", str(config_path))
+
+    async def _noop_admin(_request) -> None:
+        return None
+
+    monkeypatch.setattr(mcp_router, "_require_admin_user", _noop_admin)
```
willem-bd: **This patch target does not exist — the test never exercises the handler.**

`backend/app/gateway/routers/mcp.py` does `from app.gateway.deps import require_admin_user`, so the module attribute is `mcp_router.require_admin_user` (no leading underscore). `monkeypatch.setattr` defaults to `raising=True`, so this call raises `AttributeError` at setup and the test errors before `update_mcp_configuration` is called. The same bug is on line 67.

Even if `raising=False` were passed, the real `require_admin_user(request=None, ...)` then does `getattr(request.state, ...)` on `None` → `AttributeError` → caught by the handler's broad `except Exception` → `HTTPException(500)`, and the asserts on `resp.mcp_servers` / `max_active == 1` would fail.

Fix:

```python
monkeypatch.setattr(mcp_router, "require_admin_user", _noop_admin)
```

This also means the validation summary (red→green, 1 passed / 17 passed) needs re-running after the fix — right now the gate isn't wired up.

backend/app/gateway/routers/mcp.py
```diff
@@ -259,6 +266,70 @@ async def get_mcp_configuration(request: Request) -> McpConfigResponse:
...
+
+    # Load current config to preserve skills
+    current_config = get_extensions_config()
+
+    # Load raw (un-resolved) JSON from disk to use as the merge source.
+    # This preserves $VAR placeholders in env values and top-level keys
+    # like mcpInterceptors that would otherwise be lost.
+    raw_servers: dict[str, dict] = {}
+    raw_other_keys: dict = {}
+    if config_path is not None and config_path.exists():
```
willem-bd, line 294: Minor cleanup: by this point `config_path` is unconditionally set — either from `ExtensionsConfig.resolve_config_path()` or the `Path.cwd().parent / "extensions_config.json"` fallback five lines above. The `config_path is not None` half of this guard is dead and reads as if there were a path where it could still be `None`. Just `if config_path.exists():`.

willem-bd (commented):
Reviewed the blocking-IO offload for `PUT /api/mcp/config`. The helper extraction and the in-process `_mcp_config_write_lock` are good, but two things keep the PR from delivering what the title says:

1. **The handler still calls `reload_extensions_config()` on the event loop after the offload, defeating the fix.** `_apply_mcp_config_update()` already reloads inside the worker thread and returns `reloaded_servers`, and the response is built from it on the new line 401. But the lines immediately after (404–409, unchanged context just below this hunk) reload a second time on the loop — that path goes through `ExtensionsConfig.from_file()` → `Path.exists()` + `open(...)` + `json.load(...)`, exactly the blocking IO the PR set out to remove — and then rebuild `servers` from the second reload, discarding line 401. `make detect-blocking-io` likely missed it because the blocking call now lives inside `reload_extensions_config()` rather than literally in `mcp.py`.
2. **The new `tests/blocking_io/test_mcp_router.py` anchors don't actually run the handler.** Both `monkeypatch.setattr(mcp_router, "_require_admin_user", _noop_admin)` calls target an attribute that does not exist on the module — the import is `from app.gateway.deps import require_admin_user` (no leading underscore). With `monkeypatch.setattr`'s default `raising=True`, both calls raise `AttributeError` at setup before `update_mcp_configuration` ever runs. The claimed `1 passed` / `17 passed` validation needs re-running after the patch target is fixed before this is a real regression gate.

One minor cleanup is inline as well. With #1 fixed, the surrounding offload comment block becomes accurate as written.

WillemJiang (approved):

Issue #3550, opened 2026-06-13
[context] Context Lifecycle Evidence — applied components and reduction boundaries


ShenAC-SAC:
> **Status: redesign proposal, proceeding against a provisional dependency.**
>
> This issue assumes the deliberately narrow **Problem 1 RunEvent contract proposed in #4243**: a machine-readable envelope, identity and ordering rules, compatibility rules, and an event-type registry. That proposal is not settled yet. We will use it as the working substrate and adapt this design if the contract changes after review.
>
> #3550 is the first extension of that substrate. It must not define a second event envelope, store, completeness model, or tracing system.

## Decision

Replace the previous “Experience Ledger / visible → evaluable → learnable” roadmap with one narrow vertical slice:

> Record compact, content-free evidence of which covered DeerFlow context components were present at each real lead-agent model-call boundary, and record deterministic reductions where the provider can prove them.

The first covered components are **Memory** and **Summary**. Memory preserves the compatibility baseline merged in #3556; Summary proves that the design can represent a transformed and lossy context component rather than only a hash.

After Memory + Summary work through the existing run-events API, this issue stops. Additional providers or evaluation features must earn their own scope through a named consumer.

## Problem

Agent outcomes depend on hidden context, but DeerFlow cannot currently answer these questions from durable, structured evidence:

1. Which covered context components were present in a specific lead-agent model call?
2. Did two model calls receive byte-identical Memory or Summary components?
3. Was a Summary reduced at DeerFlow's final render boundary before that call?
4. Can an operator join that evidence to the run and its existing feedback/evaluation data without persisting private prompt content?

External tracing shows model/tool execution. Checkpoints preserve state. Neither is a compact, provider-owned statement of DeerFlow's context application and deterministic reduction semantics.

This is **Context Lifecycle Evidence**, not generic tracing and not prompt logging.

## Current code truth

Verified against `upstream/main` at `693507870cae26dadf3af487f811eb2bcfe18f87`:

- `DynamicContextMiddleware._record_effective_memory()` hashes the exact effective hidden Memory `HumanMessage.content`.
- `RunJournal.record_memory_context()` writes `0..1` `context:memory` events per Gateway run:

  ```json
  {
    "event_type": "context:memory",
    "category": "context",
    "content": {
      "content_sha256": "..."
    }
  }
  ```

- A later run records the frozen Memory block actually effective in that run. This is run-level evidence, not one thread-level snapshot row.
[...]

PR #3556, opened 2026-06-13
feat(context): record effective memory identity per run

ShenAC-SAC:
## Why

Issue #3550 calls for making the context used by a run visible before building evaluation or learning on top of it. Today DeerFlow checkpoints the hidden memory block, but the run ledger cannot answer a simpler debugging question: **did two runs use exactly the same effective memory?**

This PR records that identity without copying user memory into `run_events` and without inventing a broader snapshot schema. It is a rewrite on current `main` after #4122; the `MemoryManager` and DeerMem provider contracts remain unchanged.

## What Changed

- `DynamicContextMiddleware` hashes the exact effective hidden memory `HumanMessage.content`, including the `<memory>` wrapper.
- Each Gateway run with an effective memory block records one `context:memory` event through its existing run-scoped `RunJournal`.
- First-run injection records the block created by the current middleware update. Later runs and checkpoint-based branches record the frozen block already present before the run, without reloading memory.
- Goal continuations are deduplicated to one memory identity event per run.
- The existing debug/audit API is the consumer: `GET /api/threads/{thread_id}/runs/{run_id}/events?event_types=context:memory`.
- Untrusted Gateway input cannot mark a caller-supplied message as server-created dynamic context. Trusted internal calls keep their existing metadata behavior.

The event payload intentionally has one field:

```json
{
  "content_sha256": "<sha256 of the effective hidden memory block>"
}
```

`content_sha256` is consumed by operators comparing effective memory identity across runs and can be recomputed from the checkpointed thread state when deeper inspection is needed.

## Why Only One Field

The previous implementation exposed source IDs, counts, sections, token budgets, truncation state, status fields, and a schema version. Those fields do not have a reliable current consumer. In particular, memory fact IDs are mutable/deletable, do not cover summary sections, and custom-agent memory is not resolvable through the current Memory API.

This rewrite removes that speculative surface instead of preserving it for hypothetical future evaluators. The hash answers the current debugging question; future fields should be added only with a concrete consumer.

## Behavior and Boundaries

- No event is emitted when there is no effective memory block or no run journal.
- Event-write failures never block memory injection or the agent run.
- The full memory text remains only in checkpoint state; it is not duplicated into the event store.
- The anti-forgery boundary applies prospectively. Checkpoints already polluted before deployment cannot be authenticated retroactively.
- Event retention and durability follow the configured `run_events` backend.

## Out of Scope

- Source-level memory provenance or historical fact resolution.
- Skills/config context identity.
- UI or evaluator consumers.
- Embedded `DeerFlowClient` ledger support when no Gateway run journal exists.

## Testing

- Focused middleware, Gateway, journal, endpoint, boundary, and blocking-I/O tests: `183 passed`.
- Full backend suite: `7524 passed, 28 skipped`.
- Ruff format and lint: passed.
- `git diff --check`: passed.

## Risk / Rollback

- Risk is limited to one fixed-size event per run that has effective memory, plus stripping two server-owned dynamic-context markers from untrusted input.
- Rollback by reverting this commit; existing consumers can also ignore `context:memory` events.

Related: #3550


WillemJiang:
@ShenAC-SAC  cloud you resolve the conflicts with the main branch?

ShenAC-SAC:
@WillemJiang Conflicts are resolved — merge commit 7a4c7075, and the PR is mergeable again. The conflicts were not mechanical: `main` has evolved the exact code paths this PR instruments, so I took `main`'s newer implementations as the base and re-wove the M1 provenance capture on top. Details below.

## What changed in the merge

**`memory/prompt.py`** — the selection logic this PR refactors was rewritten on `main` by #3592 (guaranteed injection for correction facts: a dual-budget selection with `_select_fact_lines` / `_fallback_format_facts` helpers and a structure-aware safety truncation). The resolution keeps #3592's implementation as authoritative and rebuilds `_build_injection_parts` on top of it:

- `_select_fact_lines` and `_fallback_format_facts` now additionally return the selected facts' stable identities, so the snapshot's `fact_ids` covers all three selection paths — guaranteed, regular, and the exception fallback — in final emission order.
- `format_memory_for_injection`, `build_injected_memory_snapshot`, and `build_injection_text_and_snapshot` all pass through the new `guaranteed_categories` / `guaranteed_token_budget` kwargs; the single-pass text/snapshot lockstep guarantee is unchanged.
- One semantic improvement falls out of #3592's structure-aware truncation: the Facts block is now a *protected suffix*, so `fact_ids` stays exact even when the safety-net fires. `truncated` now means "the preceding user-context/history sections were clipped" (the snapshot payload schema is unchanged, still `schema_version: 1`).

**`dynamic_context_middleware.py`** — `main`'s #3630 split the reminder into a date-only `SystemMessage` plus a separate role-isolated memory `HumanMessage`. `_build_full_reminder` now returns `(date_reminder, memory_block, snapshot)`, fully preserving that role isolation; the snapshot is still built inside the timed `_inject` pass from the same memory load as the text, carried out via `_InjectResult`, and recorded once per thread on the first-turn branch.

**Docs / tests** — `backend/CLAUDE.md` is a stub on `main` now, so the snapshot doc block moved to `backend/AGENTS.md`. Tests that patched `_get_memory_context` (dead seam after the merge — the middleware reads `_get_memory_context_with_snapshot`) were repointed, and message-shape assertions were made layout-agnostic against the #3630 three-message layout. The PR-side summary-target test was dropped in favor of `main`'s equivalent `test_legacy_summary_message_is_not_injection_target`.

## Latent bug on `main`, fixed here

This PR's truncation test exposed a pre-existing bug introduced by #3592: in `format_memory_for_injection`, `facts_header` and `all_fact_lines` are initialized *inside* the `if isinstance(facts_data, list) and facts_data:` block, but the structure-aware truncation at the bottom of the function references them unconditionally. A payload with **no facts** whose user-context/history sections alone exceed `max_tokens` (e.g. an oversized `workContext.summary`) skips the facts block and then crashes with `UnboundLocalError: all_fact_lines` in the truncation branch — memory injection fails entirely instead of clipping.

Fixed in the merge by hoisting both initializations above the facts block, so the truncation path is well-defined in the facts-less case and collapses to the plain prefix-clip. The case is pinned by this PR's `test_snapshot_flags_final_truncation_and_keeps_hash_faithful` (facts-less over-budget payload → clip fires, `truncated=True`, hash matches the clipped text).

## Validation

- Full backend suite on the merged tree: **6371 passed, 26 skipped, 0 failed**
- `ruff check` + `ruff format --check`: clean

ggnnggez:
### Follow-up: no functional blocker, one docs/naming nit

Re-reviewed the snapshot path in detail. Confirming there is **no functional blocker** — the single-pass design, `resolve_run_journal` extraction, once-per-thread emission, and bounded payload are all correct, and the `content_hash` fully serves its stated drift-detection purpose today (the `<memory>` wrapper and `<system-reminder>` envelope are deterministic constants, so hashing the body is a valid drift key).

One **non-blocking docs/naming** nit worth a fixup: `content_hash` is computed over the formatter **body** (`parts.text`), not over the `<memory>`-wrapped / `<system-reminder>`-enveloped bytes that actually land in the message.

Hash is over `parts.text` (body only):

https://github.com/bytedance/deer-flow/blob/7a4c7075755cffa6e9c735118b127bf49b45cd37/backend/packages/harness/deerflow/agents/memory/prompt.py#L858-L878

But the injected string is wrapped one layer up (and then wrapped again in `<system-reminder>` by the middleware):

https://github.com/bytedance/deer-flow/blob/7a4c7075755cffa6e9c735118b127bf49b45cd37/backend/packages/harness/deerflow/agents/lead_agent/prompt.py#L699-L708

The docstrings/comments call this "the exact injected text" / "what was actually injected", and the e2e test comment says the hash is "the text that was actually injected" while asserting against the bare body:

https://github.com/bytedance/deer-flow/blob/7a4c7075755cffa6e9c735118b127bf49b45cd37/backend/packages/harness/deerflow/agents/memory/prompt.py#L824-L830

https://github.com/bytedance/deer-flow/blob/7a4c7075755cffa6e9c735118b127bf49b45cd37/backend/tests/test_dynamic_context_middleware_snapshot.py#L294-L296

The risk is only latent: a future M2/M3 consumer that trusts the "exact injected text" wording could try to byte-match `content_hash` against the raw `llm_request` message content and find it doesn't line up (wrapper not stripped).

Suggested fix (preferred): clarify the wording / rename to something like `memory_body_hash`, documenting it as "hash of the formatted memory body, before the `<memory>` wrapper and `<system-reminder>` envelope", and add an assertion pinning that the body is a substring of the final message. I'd avoid re-hashing the wrapped text — that folds a deterministic presentation envelope into a value whose purpose is memory-content drift, which reads worse.

ShenAC-SAC:
@ggnnggez Addressed in a1c0c9b7 — wording + test pinning; the hash input itself is unchanged. Point by point:

1. **Wording clarified at every site that made the claim.** Everywhere that called `content_hash` "the exact injected text" now says it keys the *formatted memory body, before the `<memory>` wrapper and `<system-reminder>` envelope*: the `InjectedMemorySnapshot` class docstring and the `truncated` field comment, the `build_injected_memory_snapshot` docstring (which now also states explicitly that consumers comparing the hash against a captured request must strip those envelopes first), the `build_injection_text_and_snapshot` summary, the snapshot doc block in `backend/AGENTS.md`, and the test docstrings/comments (`test_snapshot_hash_matches_injected_text` → `test_snapshot_hash_matches_formatted_memory_body`).

2. **Envelope relation pinned in the e2e test.** `test_end_to_end_records_real_context_event_in_store` previously asserted the bare body was a substring of the message; it now asserts the wrapped form — `f"<memory>\n{expected_memory_text}\n</memory>"` — is a substring of the injected message content, right next to the hash assertion over the bare body. So the test itself documents "hash input = body, message = wrapped body", and any change to the wrapper shape breaks loudly instead of silently invalidating the wording.

3. **Did not re-hash the wrapped text** — agreed with your reasoning. The hash's purpose is a memory-*content* drift key; folding a deterministic presentation envelope into it would make the key churn on unrelated envelope refactors (another #3630-style message-shape change would invalidate every stored hash with zero actual memory drift).

4. **Kept the `content_hash` field name** rather than renaming to `memory_body_hash`. Renaming would be free right now (`schema_version: 1` is unshipped, and a whole-tree grep shows only this PR's own code and tests reference the snapshot's field), so cost wasn't the reason. The reason is forward shape: M1.5 plans to add `context:skills` / `context:config` snapshots to the same ledger, and a uniform `content_hash` key across `context:*` payload types is friendlier to generic ledger consumers than a per-category name like `memory_body_hash`. The self-description concern you raised is real, but it's a docs-precision problem rather than a naming one, and (1) closes it — the payload docs now state exactly what the hash covers.

Validation: all snapshot-related test files pass (63 tests across the five touched/adjacent files), `ruff check` and `ruff format --check` clean.

WillemJiang:
@ShenAC-SAC There are some memory manager refactorings on the main branch https://github.com/bytedance/deer-flow/pull/4122;  you may need to apply the patch on top of the new main branch.

backend/packages/harness/deerflow/agents/middlewares/dynamic_context_middleware.py
```diff
@@ -305,3 +310,51 @@ async def abefore_agent(self, state, runtime: Runtime) -> dict | None:
...
+        A first-run block must come from this middleware's update. A reused
+        block must have existed in the checkpoint before the run; the Gateway
+        strips the reminder marker from untrusted input so a caller cannot
+        replace a known checkpoint ID with forged provenance.
+        """
+        if isinstance(update, dict):
+            update_messages = update.get("messages")
+            if isinstance(update_messages, list):
+                for message in update_messages:
+                    message_id = str(message.id or "") if isinstance(message, HumanMessage) else ""
```
willem-bd: Nit (non-blocking): the `else ""` branch here is dead — the next line re-checks `isinstance(message, HumanMessage)` before `message_id` is used, so the ternary never changes the outcome. The same pattern repeats at line 339 in the state-scan loop. Slightly cleaner:

```python
for message in update_messages:
    if not isinstance(message, HumanMessage):
        continue
    message_id = str(message.id or "")
    if message_id.endswith("__memory") and is_dynamic_context_reminder(message) and isinstance(message.content, str):
        return message
```

Separate micro-nit on the same method: `_effective_memory_message` is `@staticmethod` but only ever called from the instance method `_record_effective_memory` — a plain method (or module-level helper) would be a bit easier to mock/extend.

backend/app/gateway/services.py
```diff
@@ -57,6 +57,13 @@
     RunStatus.interrupted,
 }
 
+_SERVER_OWNED_DYNAMIC_CONTEXT_KEYS = frozenset(
```
willem-bd, line 61: Nit (maintainability): this frozenset hardcodes the *values* of `_DYNAMIC_CONTEXT_REMINDER_KEY` / `_REMINDER_DATE_KEY` defined in `dynamic_context_middleware.py`, and `test_normalize_input_strips_external_dynamic_context_metadata` hardcodes the same literals. If either constant's value ever changes, this strip list silently drifts and the anti-forgery boundary reopens with no test failure. Consider importing those constants here (`services.py` already imports from `deerflow.*`) so the coupling is structural rather than by string — this is the one nit with a real future-bug tail.

backend/packages/harness/deerflow/runtime/journal.py
```diff
@@ -630,6 +631,23 @@ def record_middleware(self, tag: str, *, name: str, hook: str, action: str, chan
...
+        """Record the effective hidden memory block for this run.
+
+        The full block already lives in checkpoint state and may contain user
+        data, so the event stores only its exact SHA-256 identity. Operators
+        consume it through the existing run-events debug API to compare the
+        effective memory used by different runs without copying that content.
+        """
+        if self._memory_context_recorded:
+            return
+        self._memory_context_recorded = True
```
willem-bd: Nit (non-blocking): the dedup flag is flipped before `self._put(...)`. If `_put` ever raised, the event would be lost for the run with no retry. `_put` only does an in-memory buffer append today so this is academic, but the conventional ordering is to set the guard *after* the write succeeds.

backend/AGENTS.md
```diff
@@ -583,6 +583,11 @@ The cached value is reused for both the blocking (`runs.wait`) and streaming (`_
...
 - Next interaction injects selected facts + context into `<memory>` tags in the system prompt when `injection_enabled` is true.
 
+**Run-level memory identity**:
+- Every Gateway run with an effective hidden memory block hashes the exact `HumanMessage.content`, including the `<memory>` wrapper, and records one `context:memory` event through its run-scoped `RunJournal`. Later runs and checkpoint-based branches reuse the frozen message without reloading memory; goal continuations are deduplicated to one event per run.
```
willem-bd, line 588: Nit (doc precision): "Every Gateway run with an effective hidden memory block … records one `context:memory` event" isn't quite true on the timeout path — on a later run the frozen block stays in `state["messages"]` and is served to the model, but `_record_effective_memory` is skipped when `abefore_agent` times out, so no event is emitted despite memory being effectively present. This is a very narrow case (timeouts are essentially a first-run cold-start concern, where there's no frozen block yet) and the event is best-effort by design — a half-sentence caveat ("except when injection times out") would make the doc match behavior exactly.

willem-bd (commented):
Reviewed the full diff against current `main`. Overall this is solid: the anti-forgery boundary (pre-existing-ids + Gateway marker stripping) is defense-in-depth and tested, the one-event-per-run dedup holds across goal continuations (the continuation config carries `__pregel_runtime`, so the same `RunJournal` is reused and `_memory_context_recorded` persists), and persisting only a SHA-256 keeps user memory text out of the event store. `category="context"` is sound (only `category="message"` feeds `list_messages`).

Approve with the minor nits below — none are blockers. Issue #2 (duplicated key-name literals) is the only one with a real future-bug tail, so that's the one I'd most like to see addressed; the rest are readability/precision polish.

backend/packages/harness/deerflow/runtime/journal.py
ShenAC-SAC: Addressed in 942691d1. The dedup flag is now set only after _put succeeds. I added a regression test where the first append fails and the second call successfully records exactly one event.

backend/AGENTS.md
ShenAC-SAC: Addressed in 942691d1 by closing the behavior gap rather than weakening the invariant. On timeout, the middleware now records trusted checkpoint memory with update=None; late worker output is still ignored, so no phantom event is possible. Tests cover both the later-run checkpoint case and the first-run late-result case, and the timeout log and comment wording is now precise.

backend/packages/harness/deerflow/agents/middlewares/dynamic_context_middleware.py
ShenAC-SAC: Addressed in 942691d1. Both loops now narrow to HumanMessage with an early continue before reading the ID. I kept @staticmethod intentionally because the helper has no instance state and there is no current override or extension consumer.

backend/app/gateway/services.py
ShenAC-SAC: Addressed in 942691d1. The Gateway strip policy now imports the canonical _DYNAMIC_CONTEXT_REMINDER_KEY and _REMINDER_DATE_KEY from the middleware, and the external and trusted-input tests use the same constants. The trusted path now asserts that both metadata keys are preserved.

ShenAC-SAC:
## Clarifying the current scope and value of this PR

This PR is intentionally a narrow run-observability change. It is not intended to provide complete context provenance, memory evaluation, or a user-facing memory inspection feature.

### The problem it solves

DeerFlow stores the hidden memory block in checkpoint state, but the run event ledger currently cannot answer a basic debugging question:

> Did two runs use exactly the same effective memory?

This matters when two runs from the same thread produce different results. Without a run-level memory identity, an operator cannot quickly determine whether the difference may have come from memory assembly or whether that part of the input was unchanged.

The same question also appears when debugging continuations, checkpoint branches, resumes, and injection timeouts: did the run actually continue with the expected frozen memory block?

### What this PR records

For a Gateway run with an effective hidden-memory block and a run journal, the middleware records a deduplicated `context:memory` event containing one payload field:

```json
{
  "content_sha256": "<sha256 of the exact effective memory message content>"
}
```

The hash is computed from the exact `HumanMessage.content` supplied to the model, including the `<memory>` wrapper.

The effective block is selected from:

- the memory created by the current middleware update on the first run;
- the trusted frozen memory already present in checkpoint state on later runs or checkpoint branches;
- the trusted checkpoint memory that remains effective when a new injection attempt times out.

A timed-out worker result is not recorded after it arrives, because that result was never applied to the run.

The full memory text remains in checkpoint state and is not copied into the event store.

### How it can be used today

The immediate consumer is the existing run-events API:

```text
GET /api/threads/{thread_id}/runs/{run_id}/events?event_types=context:memory
```

This supports several concrete debugging workflows:

1. **Compare runs with different outputs**

   If two runs have the same hash, the effective memory content was the same, so memory assembly can usually be ruled out as the source of the behavioral difference.

   If the hashes differ, the operator knows that the model received different memory content and can inspect the corresponding checkpoint states.

2. **Verify continuation and branch behavior**

   An operator can check whether a resumed run or checkpoint branch reused the expected frozen memory instead of silently loading or constructing a different block.

3. **Diagnose timeout behavior**

   When memory injection times out on a later run, the event identifies the checkpoint memory that still remained effective. Late background results are deliberately ignored.

4. **Correlate memory identity with other run events**

   The memory fingerprint can be examined alongside tool calls, token usage, errors, and other events already associated with the same `thread_id` and `run_id`.

The event is best-effort: event-write failures do not block the agent run. Therefore, the presence of an event proves which memory identity was recorded, while absence should not be treated as an authentication or compliance guarantee.

### What it deliberately does not solve

This PR does not record:

- memory fact IDs or source-level provenance;
- sections, token counts, budgets, or truncation state;
- skills or configuration identity;
- the reason why memory changed;
- the full memory text;
- a schema reserved for hypothetical evaluators.

It also does not add a UI, an automated evaluator, a cross-run index, or event support for embedded `DeerFlowClient` runs without a Gateway `RunJournal`.

The hash is a content-identity fingerprint, not proof of origin or authenticity. Existing checkpoints that were already polluted before this boundary was introduced cannot be authenticated retroactively.

### Why the scope is intentionally small

Earlier versions attempted to expose a broader snapshot schema, but most of those fields did not have a reliable current consumer and some could not be reconstructed consistently across memory providers.

This version keeps only the field needed to answer the current debugging question. It reuses the existing checkpoint, `RunJournal`, event store, and events API, with no new table, migration, provider contract, or persistence abstraction.

The value of this PR is therefore specific:

> It makes the effective hidden memory used by a run comparable across runs.

It does not explain why the memory changed, but it provides the first concrete signal needed to determine whether memory divergence is relevant to an observed run difference.

WillemJiang (approved):

PR #3557, opened 2026-06-13
fix(frontend): improve chat math rendering

Boooobby:
## Why

LLM responses often use LaTeX delimiters such as `\(...\)` and multiline `\[...\]` for math. The chat renderer currently relies on remark-math/KaTeX paths that are most reliable with `$...$` / `$$...$$`, so common generated equations can render as raw text or fail during streaming. Long display equations can also overflow the chat column awkwardly.

This PR makes math rendering more robust for common chat output without changing code blocks or normal Markdown text.

## What changed

- Normalize common LaTeX math delimiters before rendering:
  - `\(...\)` -> `$...$`
  - multiline `\[ ... \]` -> `$$ ... $$`
- Apply the same delimiter normalization in the shared Streamdown wrapper so direct Streamdown usage benefits too.
- Make KaTeX rendering tolerant with `throwOnError: false` and `strict: false`.
- Ensure `MarkdownContent` always includes the default Streamdown rehype plugins, while still allowing callers to add extra rehype plugins.
- Add KaTeX display CSS polish so wide equations scroll horizontally instead of being squeezed or clipped.
- Add unit coverage for inline math conversion, multiline display math conversion, and code-block preservation.

## Surface area

- [x] **Default behavior change** — improves default chat math rendering behavior

## Screenshots / Recording

Not attached. Verified locally with chat responses containing inline math, multiline display math, `aligned`, `cases`, and matrix formulas.

## Bug fix verification

- Test path that reproduces the delimiter conversion bug:
  - `frontend/tests/unit/core/streamdown/preprocess.test.ts`
  - The new tests encode behavior that was not supported on `main`; they pass on this branch.
  - Full visual regression coverage for KaTeX layout would require browser screenshot tests. I added unit coverage for the delimiter normalization logic and manually verified the rendered chat output locally.

## Validation

Ran in the Docker development frontend container:

```bash
cd /app/frontend
pnpm exec vitest run tests/unit/core/streamdown/preprocess.test.ts tests/unit/core/streamdown/plugins.test.ts
pnpm typecheck
pnpm lint
pnpm format -- src/components/ai-elements/streamdown.tsx src/components/workspace/messages/markdown-content.tsx src/components/workspace/messages/message-list-item.tsx src/core/streamdown/plugins.ts src/core/streamdown/preprocess.ts src/styles/globals.css tests/unit/core/streamdown/preprocess.test.ts
NODE_ENV=production BETTER_AUTH_SECRET=*** pnpm build
pnpm test
```

Note: for the local Docker container, I copied the repo-level contract fixture into `/app/contracts/subagent_status_contract.json` before `pnpm test`, because the frontend dev container only mounts `/app/frontend` while that unit test resolves `../../../../../contracts/subagent_status_contract.json`. CI should have the full repository checkout.

## AI assistance

**Tool(s) used:** Hermes Agent

**How you used it:** AI assisted with investigating the rendering issue, implementing the delimiter normalization/CSS changes, and running validation commands. I manually reviewed the resulting diff and local behavior.


frontend/src/components/workspace/messages/markdown-content.tsx
```diff
@@ -40,6 +40,11 @@ export function MarkdownContent({
     () => preprocessStreamdownMarkdown(content),
     [content],
   );
+  const effectiveRehypePlugins = useMemo(() => {
+    const base = streamdownPlugins.rehypePlugins ?? [];
```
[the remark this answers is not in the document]
Boooobby: Thanks for calling this out. Confirmed — I agree chat assistant messages should not start parsing raw HTML as part of this fix.

I refined the change so `MarkdownContent` now uses `streamdownPluginsWithoutRawHtml` as the default plugin set and only appends caller-provided extra rehype plugins. That keeps the KaTeX/remark-math path shared, but preserves the previous no-`rehypeRaw` behavior for chat messages and avoids widening the security surface here.

The latest push also keeps `reasoningPlugins` derived from the same no-raw-html plugin set for consistency.

WillemJiang:
@Boooobby, could you please fix the lint and unit test errors? The conflicts of the main branch also need fixing.

Boooobby:
@WillemJiang Thanks for the reminder. I have updated the PR and addressed the requested items:

- Resolved the conflicts with the latest `main` branch.
- Fixed the frontend formatting/lint issue.
- Fixed the frontend unit test failure by updating the math preprocessing test expectation after display-math compaction.

I also re-ran the frontend validation locally in the project Docker frontend environment:

```bash
pnpm format
pnpm test
pnpm lint
pnpm typecheck
```

The local results are green: 33 unit test files passed, 317 tests passed, and format/lint/typecheck passed. The latest PR head is `3f8bf53f`.

frontend/src/core/streamdown/preprocess.ts
```diff
@@ -100,10 +100,169 @@ export function capMarkdownNesting(markdown: string): string {
...
 }
 
-export function preprocessStreamdownMarkdown(markdown: string): string {
-  if (!MERMAID_BLOCK_HINT_RE.test(markdown) || !markdown.includes("-.->")) {
+type MathDelimiter = {
+  close: "\\)" | "\\]";
+  replacement: "$" | "$$";
+};
+
+function convertLatexDelimitersInLine(
```
willem-bd, line 121: **Escaped-backslash blind spot: `\(` / `\[` is treated as a math delimiter even when the backslash is itself escaped.**

`convertLatexDelimitersInLine` matches `\(`/`\[` as a math open without checking whether the backslash is itself escaped. When an assistant explains LaTeX syntax in prose and writes `Use \( ... \) for inline math.` (markdown that should render the literal text `\( ... \)`), the regex guard `/[\\][([\])]/` matches, and scanning consumes the first `\` char-by-char, then pairs the *second* `\` with `(` — opening an inline-math block.

Verified by reproduction (exact transcription of the shipped function):

    input:  `Use \( to start inline math.`
    output: `Use \$ to start inline math.`   # literal \( destroyed, dangling unclosed `$`

The dangling `$` then swallows following lines as math until a stray `\)` appears. Fix: detect a preceding unescaped backslash (or consume `\\` as a unit) before treating `\(`/`\[` as a delimiter.

frontend/src/components/ai-elements/streamdown.tsx
```diff
@@ -56,21 +59,28 @@ class StreamdownFallbackBoundary extends Component<
...
-  const safeChildren = useMemo(
-    () =>
-      typeof children === "string" ? capMarkdownNesting(children) : children,
-    [children],
-  );
+  const safeChildren = useMemo(() => {
+    if (typeof children !== "string") {
+      return children;
+    }
+    return normalizeStreamdownMathMarkdown(capMarkdownNesting(children));
```
willem-bd: **Math is normalized twice here on the main message path — redundant, and not idempotent for math containing inner delimiters.**

`MarkdownContent` computes `normalizedContent = preprocessStreamdownMarkdown(content)` (which this PR made math-aware) and renders `<MessageResponse>{normalizedContent}</MessageResponse>`. `MessageResponse` wraps `ClipboardSafeStreamdown`, whose `useMemo` runs `normalizeStreamdownMathMarkdown(capMarkdownNesting(children))` on that already-normalized string — so every rendered message pays for math normalization twice.

The second pass is not idempotent for math containing literal `\(`/`\[`/`\]`/`\)`: the first pass preserves them verbatim inside the math block, and the second pass re-interprets them as delimiters:

    input:  `Equation \(f\[0\]) here.`
    1x:     `Equation $f\[0\]) here.`
    2x:     `Equation $f$$0$$) here.`        # display-math $$ injected mid-equation

`ClipboardSafeStreamdown`'s normalization is still needed for its direct callers (reasoning / artifact / subtask), which skip `preprocessStreamdownMarkdown`. Fix: keep math normalization in only one of the two chained layers — simplest is to revert `preprocessStreamdownMarkdown` to mermaid-only and let `ClipboardSafeStreamdown` own math normalization.

willem-bd (commented):
## Review: improve chat math rendering

The CSS-import reorder (so the new `.katex` rules in `globals.css` win over `katex.min.css`), the KaTeX options (`throwOnError` / `strict`), the centralized katex plugin (removing the duplicate per-message `[rehypeKatex]`), and the remark/rehype ordering are all correct. I verified `rehypeSplitWordsIntoSpans` only splits direct text children of `p`/`h*`/`li`/`strong` and leaves element children (math/KaTeX nodes) untouched, so the plugin reorder is benign.

The real issues are in the new preprocessing/streamdown plumbing — **all four reproduced against an exact transcription of the shipped functions**:

1. `convertLatexDelimitersInLine` treats `\(` / `\[` as math delimiters even when the backslash is itself escaped, corrupting prose that displays literal delimiters.
2. Math is normalized **twice** on the main message path (`MarkdownContent` -> `ClipboardSafeStreamdown`), which is redundant and not idempotent for math containing inner `\(`/`\[`.
3. `inInlineCode` is a per-line local and doesn't survive line breaks, so delimiters inside multi-line backtick code spans get converted.
4. `parseIncompleteMarkdown` is forwarded to Streamdown but never set by any caller, so the streaming-parse improvement is never active.

No crash-level bug, but #1 and #2 visibly corrupt legitimate math/technical content. Details inline.

frontend/src/core/streamdown/preprocess.ts
```diff
@@ -100,10 +100,169 @@ export function capMarkdownNesting(markdown: string): string {
...
+  line: string,
+  openBlock: MathDelimiter | null,
+): { line: string; openBlock: MathDelimiter | null } {
+  let result = "";
+  let i = 0;
+  let inInlineCode = false;
```
willem-bd: **`inInlineCode` is a per-line local, so it doesn't survive line breaks — LaTeX delimiters inside a multi-line backtick code span get converted.**

`convertLatexDelimitersInLine` initializes `let inInlineCode = false` on every call and only threads `openBlock` (the math state) across lines. An inline code span that contains a newline is valid CommonMark (e.g. `` `code spans\na line with \(x\) inside` ``); on line 2 `inInlineCode` resets to `false`, so the `\(x\)` *inside* the code is converted:

    input:  `` `code spans\na line with \(x\) inside` ``
    output: `` `code spans\na line with $x$ inside` ``

The math `openBlock` is already threaded across lines; `inInlineCode` should be threaded the same way (return it alongside `openBlock`), or the line scan should track code state across the whole document.

frontend/src/components/ai-elements/streamdown.tsx
```diff
@@ -56,21 +59,28 @@ class StreamdownFallbackBoundary extends Component<
...
+  }, [children]);
   return (
     <StreamdownFallbackBoundary raw={children}>
-      <Streamdown {...props}>{safeChildren}</Streamdown>
+      <Streamdown
+        parseIncompleteMarkdown={parseIncompleteMarkdown ?? false}
```
willem-bd: **`parseIncompleteMarkdown` is forwarded to `Streamdown` but never set by any caller, so it is always `false`.**

A repo-wide search finds no caller passing `parseIncompleteMarkdown` (`MessageResponse`, `reasoning`, `artifact-file-detail`, `subtask-card` all leave it unset). For a PR titled "improve chat math rendering", the `Streamdown` `parseIncompleteMarkdown` option — which helps parse partial markdown/math mid-stream (e.g. an unclosed `$$` while tokens are still arriving) — is wired but never enabled, so streaming-incomplete math does not benefit from it.

Either enable it on the streaming path (e.g. tie it to `isLoading`) or drop the dead plumbing so the public surface doesn't imply a capability that isn't active.

WillemJiang:
@Boooobby, please take a look at the recent review comments.

Boooobby:
@willem-bd Thanks for the thorough review. All four issues are addressed in the latest commit:

1. **Escaped-backslash blind spot** — `convertLatexDelimitersInLine` now consumes `\\` as a unit before checking for math delimiters, so `\\(` / `\\[` with an escaped backslash passes through unchanged.

2. **Double normalization** — `preprocessStreamdownMarkdown` is reverted to mermaid-only. Math normalization now lives solely in `ClipboardSafeStreamdown`, single pass.

3. **`inInlineCode` across lines** — Introduced a `DelimiterState` type that threads both `openBlock` and `inInlineCode` across line boundaries, same as the existing multi-line math state.

4. **`parseIncompleteMarkdown` dead plumbing** — `MarkdownContent` now passes `parseIncompleteMarkdown={isLoading}` to `MessageResponse`, so Streamdown's streaming-incomplete parsing is active during streaming.

Added test coverage for escaped-backslash preservation and multi-line code span protection. All tests / typecheck / lint pass.

Boooobby:
@WillemJiang Thanks for the reminder. I have addressed the latest review comments and pushed the updates. The related frontend tests, typecheck, and lint now pass locally.

WillemJiang:
@Boooobby, could you please check the lint error?  You can set up the pre-commit check by running the 'make install' in the root directory.

frontend/src/core/streamdown/preprocess.ts
```diff
@@ -100,8 +100,182 @@ export function capMarkdownNesting(markdown: string): string {
...
+      }
+      const converted = convertLatexDelimitersInLine(line, mathState);
+      mathState = converted.state;
+      return converted.line;
+    })
+    .join("\n");
+}
+
+function flattenDisplayMathBody(lines: string[]): string {
+  return lines.map((line) => line.trim()).join(" ");
```
willem-bd: [P1] `flattenDisplayMathBody` silently drops anything after a `%` comment in multi-line display math

- Problem: `flattenDisplayMathBody` joins display-math body lines with a single space (`lines.map(l => l.trim()).join(" ")`). KaTeX honors TeX-style `%` line comments — confirmed in the installed `katex@0.16.28` `Lexer.js`: `"%": 14 // comment character` and `if (this.catcodes[text] === 14)` strips through the next `\n`. Flattening removes the newline, so the comment now extends to the end of the (now single-line) body, swallowing everything after it.
- Evidence: input `$$\na % step 1\n+ b\n$$` → body becomes `a % step 1 + b` → KaTeX renders only `a` instead of `a + b`. Before this PR, the newline survived and KaTeX rendered `a + b` correctly. This is a new regression from `compactDisplayMathBlocks`.
- Realism: low frequency but real — LLMs trained on LaTeX source emit `% step 1`, `% definition of X`, etc., particularly when walking through derivations. Once you hit it, the failure is silent (no KaTeX error; equation just looks wrong).
- Suggested fix: when flattening, either (a) skip flattening if any body line contains an unescaped `%`, or (b) strip TeX comments before flattening (`line.replace(/(?<!\\)%.*$/, '')`), or (c) join with `\n` instead of space — Streamdown's block splitter is what motivated compaction; keeping newlines inside `$$...$$` may still work if the splitter only cares about the delimiters being on their own lines.
- Test: extend `compactDisplayMathBlocks keeps display math as display math` with a body line containing `%`, assert the post-`%` content survives.

frontend/src/core/streamdown/preprocess.ts
```diff
@@ -100,8 +100,182 @@ export function capMarkdownNesting(markdown: string): string {
...
+  state: DelimiterState,
+): { line: string; state: DelimiterState } {
+  let result = "";
+  let i = 0;
+  let inInlineCode = state.inInlineCode;
+  let currentBlock = state.openBlock;
+
+  while (i < line.length) {
+    // Handle backtick for inline code tracking
+    if (line[i] === "`") {
```
willem-bd, line 131: [P2] `inInlineCode` toggles on every single backtick, ignoring CommonMark multi-backtick code-span runs

- Problem: CommonMark allows code-span delimiters to be runs of N backticks (`` `` ``, `` ``` ``, etc.), with the closing delimiter required to be a run of the SAME length. The new converter toggles `inInlineCode` on every individual backtick character, so a 2-backtick code span like `` ``\(literal\)`` `` flips state on the first backtick (entering code), then immediately flips back on the second backtick (exiting code) — and now treats the code-span contents as prose and rewrites the `\(literal\)` inside.
- Evidence: input `` Use ``\(literal\)`` here`` → output `` Use ``$literal$`` here``. Distinct from willem-bd's multi-LINE finding (which the PR addressed by threading `inInlineCode` through `mathState`); this is multi-BACKTICK on a single line, which CommonMark uses to write code spans that themselves contain a literal backtick.
- Realism: a user asking the assistant to explain LaTeX syntax ("to write inline math, type `` ``\(x\)`` `` …") is a realistic chat input that hits this exactly.
- Suggested fix: detect a backtick RUN (greedy `` `+ ``) and toggle code state only when a matching-length closing run appears; until then, keep `inInlineCode=true`. Or treat code-span tokenization at a higher layer (remark already does this — the conversion could happen as a remark plugin walking only text nodes, sidestepping the whole code-span problem).
- Test: add a case with multi-backtick code spans containing `\(...\)`.

frontend/src/components/ai-elements/streamdown.tsx
willem-bd: [P2] Flipping `parseIncompleteMarkdown` default to `false` regresses streaming for direct callers that don't opt in

- Problem: this PR adds `parseIncompleteMarkdown={parseIncompleteMarkdown ?? false}` to fix willem-bd's earlier finding ("the prop is forwarded but never set"), but Streamdown's library default is `true` — confirmed in `node_modules/streamdown/README.md` (`parseIncompleteMarkdown | boolean | true`) and in `dist/index.js` (`parseIncompleteMarkdown:t=!0`). Flipping the wrapper default to `false` means every direct `ClipboardSafeStreamdown` caller that doesn't pass the prop now has incomplete-markdown parsing DISABLED for the first time, even though they were getting `true` before this PR.
- Evidence: `MarkdownContent` was updated to forward `={isLoading}` (good), but `ReasoningContent` (`reasoning.tsx:230`) is `<ClipboardSafeStreamdown {...reasoningPlugins}>` with no `parseIncompleteMarkdown`, and `<Reasoning>` is rendered with `isStreaming={isLoading}` from `message-list-item.tsx`. During streaming, partial markdown like `**Analyzing the` now renders literal asterisks until the closing `**` arrives, instead of styled bold like before. Same for `artifact-file-detail.tsx`, `memory-settings-page.tsx`, `subtask-card.tsx`.
- Suggested fix: default to Streamdown's own default (`true`) instead of `false` — i.e. `parseIncompleteMarkdown={parseIncompleteMarkdown ?? true}`, or just don't pre-destructure the prop and let it pass through `{...props}` naturally. The streaming-math benefit the PR actually wants is achieved by `MarkdownContent`'s explicit `={isLoading}`; that doesn't require flipping the wrapper default.

frontend/src/components/ai-elements/streamdown.tsx
willem-bd: [P2] Manual edit to `ai-elements/` violates the registry-clean rule in `frontend/CLAUDE.md`

- Problem: `frontend/CLAUDE.md:84` says verbatim: *"Components: `ui/` and `ai-elements/` are generated from registries (Shadcn, MagicUI, React Bits, Vercel AI SDK) — don't manually edit these."* This PR adds new logic (`normalizeStreamdownMathMarkdown(capMarkdownNesting(children))` plus `parseIncompleteMarkdown` extraction) inside `ai-elements/streamdown.tsx`. A registry re-sync wipes the math normalization wiring.
- Evidence: pre-existing `capMarkdownNesting` already lives here, so this PR is not creating the convention break — but it's making it worse by adding more logic to a file that's marked don't-edit. The pre-processing is reusable across all `ClipboardSafeStreamdown` callers, but it doesn't have to live INSIDE the registry-managed component.
- Suggested fix: extract the math+nesting pre-processor into a hook (e.g. `useSafeStreamdownChildren(children)`) in `src/core/streamdown/` and call it from `MarkdownContent` (and from any other call site that wants math normalization). `ai-elements/streamdown.tsx` reverts to the registry shape, and a future regenerate doesn't wipe anything.
- Alternative: if the maintainers consider this fork-of-the-registry intentional, update `CLAUDE.md` to carve out `streamdown.tsx` as a custom-extended component, so the rule and reality match.

willem-bd (commented):
Adding to @willem-bd's review. Confirmed the escaped-backslash, double-normalization, per-line `inInlineCode`, and unwired `parseIncompleteMarkdown` findings have either been addressed in the PR head or are still real as stated. Four net-new items below — three correctness, one convention. None blocking; #2 (`%` comment flattening) is the highest-signal because it silently changes rendered math without a visible error.

Boooobby:
@willem-bd Thanks for the detailed follow-up review. I pushed an update addressing the four correctness items and the registry-clean concern:

1. **Display math with TeX `%` comments** — `compactDisplayMathBlocks` now detects unescaped `%` in display-math body lines and preserves those lines instead of flattening them, so KaTeX comment semantics do not silently swallow following content. Escaped `\%` still allows normal compaction.

2. **Multi-backtick code spans** — inline code tracking now consumes a full backtick run and only exits code mode when it sees a matching-length closing run. This keeps CommonMark code spans like `` ``\(literal\)`` `` protected from delimiter conversion.

3. **`parseIncompleteMarkdown` default** — `ClipboardSafeStreamdown` no longer overrides Streamdown’s default with `false`; it now passes props through and lets Streamdown keep its own default behavior. `MarkdownContent` still explicitly passes `parseIncompleteMarkdown={isLoading}` for the streaming chat path.

4. **`ai-elements/` registry-clean rule** — moved the custom math/nesting preprocessing out of `components/ai-elements/streamdown.tsx` into `src/core/streamdown/`:
   - `safe-children.ts` centralizes `capMarkdownNesting` + math normalization.
   - `components.tsx` exposes `SafeStreamdown`, `SafeMessageResponse`, and `SafeReasoningContent` for workspace/business callers.
   - `ai-elements/streamdown.tsx` is back to a thin Streamdown wrapper with clipboard fallback and error boundary only.

I also added unit coverage for TeX `%` comments, escaped `\%`, and multi-backtick code spans. Local validation is green: `pnpm format`, `pnpm lint`, `pnpm typecheck`, and `pnpm test` all pass.

Boooobby:
@WillemJiang Thanks for the reminder. I fixed the frontend formatting issue and pushed the update. The local frontend checks are now green: `pnpm format`, `pnpm lint`, `pnpm typecheck`, and `pnpm test` all pass.

WillemJiang (approved):

PR #3563, opened 2026-06-13
fix(skills): offload blocking filesystem IO in get_custom_skill_history

ly-wang19:
## Why

`make detect-blocking-io` (the `blocking-io-guard` SOP, #3503) flags
`GET /api/skills/custom/{skill_name}/history`: `get_custom_skill_history` runs all of its
filesystem work directly on the asyncio event loop — `get_or_new_skill_storage()`,
`custom_skill_exists()`, `get_skill_history_file().exists()`, and `read_history()` (which
reads the per-skill `.history` JSONL). Under concurrency a slow disk stalls every other
in-flight request on that worker. Same class as the offloads already merged for the
custom-agent router (#3457) and channels (#3529).

## What changed

No caller-visible change. `get_custom_skill_history` now performs its storage probes and
the history read inside a nested sync function run via `asyncio.to_thread`; a `None` return
signals the 404 (kept distinct from a legitimately empty history list). Status codes and the
response body are unchanged.

## Surface area

- [x] **Backend API** — `GET /api/skills/custom/{name}/history` (internal perf; request/response shape unchanged)

## Bug fix verification

Following the `blocking-io-guard` SOP:

- **Candidate:** `routers/skills.py:224` `get_custom_skill_history` — `FILE_METADATA`, `DIRECT_ASYNC` → **FIX+ANCHOR**.
- **Re-scan (Step 2):** after the fix `make detect-blocking-io` no longer lists `get_custom_skill_history` (the remaining `rollback_custom_skill` / `update_skill` findings are untouched — see Scope).
- **Anchor + teeth (Step 5):** `backend/tests/blocking_io/test_skills_router.py` drives the real handler against a real on-disk skill + history. Teeth verified: **RED** against pre-fix code (`BlockingError`), **GREEN** with the offload, under `make test-blocking-io`.

## Validation

In `backend/`:

- `python scripts/detect_blocking_io_static.py backend/app/gateway/routers/skills.py` → `get_custom_skill_history` finding gone.
- `uv run pytest tests/blocking_io/test_skills_router.py tests/test_skills_custom_router.py tests/test_skills_validation.py` → **31 passed** (behavior preserved + new anchor).

### Scope

Limited to the self-contained read handler. `rollback_custom_skill` and `update_skill` also
touch blocking IO, but each interleaves it with `await`s (the security scan / system-prompt
cache refresh) **and** does a config/file read-modify-write — offloading those reintroduces
the within-process write serialization the single-threaded loop gives for free, so they need
the `asyncio.Lock` treatment from #3552. Left as a separate fix unit to keep this PR small.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Ran the `blocking-io-guard` SOP — `make detect-blocking-io` to find the candidate, implemented the `asyncio.to_thread` offload, and wrote the runtime anchor with the red→green teeth check. I reviewed every line and verified locally with the detector, ruff, and the test runs above.


backend/tests/blocking_io/test_skills_router.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+
+
+async def test_get_custom_skill_history_does_not_block_event_loop(tmp_path: Path, monkeypatch) -> None:
+    skills_root = tmp_path / "skills"
+    config = _config(skills_root)
+    monkeypatch.setattr("deerflow.config.get_app_config", lambda: config)
+
+    def _seed() -> None:
+        # Real custom skill + history file on disk (seeding offloaded; not under test).
```
willem-bd: Nit: this monkeypatch is dead on the path under test.

The handler receives `config` via `Depends(get_config)` and passes it as `get_or_new_skill_storage(app_config=config)`. That function calls `_make_storage(app_config.skills, ...)` → `LocalSkillStorage(host_path=str(skills_config.get_skills_path()), ...)` with `host_path` explicitly provided. `LocalSkillStorage.__init__` only falls through to `get_app_config()` when `host_path is None`, which never happens here (and `_seed` also constructs storage with `app_config=config`).

Safe to delete — leaving it in suggests storage construction depends on the global config when it doesn't.

backend/tests/blocking_io/test_skills_router.py
```diff
@@ -0,0 +1,59 @@
...
+
+pytestmark = pytest.mark.asyncio
+
+
+def _config(skills_root: Path) -> SimpleNamespace:
+    return SimpleNamespace(
+        skills=SimpleNamespace(
+            get_skills_path=lambda: skills_root,
+            container_path="/mnt/skills",
+            use="deerflow.skills.storage.local_skill_storage:LocalSkillStorage",
```
willem-bd, line 35: Nit: `skill_evolution` is unused on this handler's call path.

It's only read by `scan_skill_content`, which is called from `update_custom_skill` / `rollback_custom_skill` — not from `get_custom_skill_history`. The anchor will pass identically with just `skills=...`.

If you want this as a copy-paste template for sibling anchors of the write handlers, fine to keep; otherwise drop it to keep the fixture minimal.

willem-bd (commented):
Reviewed the blocking-IO offload for `GET /api/skills/custom/{skill_name}/history`. The substantive change is clean and correct: the nested `_read_history` closure correctly captures the post-strip `skill_name`, the `None`-vs-`[]` return distinguishes 404 from an empty history list (so the response shape is unchanged), the anchor exercises real on-disk storage via `LocalSkillStorage`, and deferring `rollback_custom_skill` / `update_skill` (which would need the #3552 lock treatment) to a follow-up is reasonable scoping.

Two nits in the test setup are inline — both safe to delete without weakening the anchor.

WillemJiang:
@ly-wang19 Please fix the lint error and unit test error.

ly-wang19:
Thanks @WillemJiang — both are fixed in `e35e360b`.

**Root cause:** merging `main` left the offloaded reader calling `get_or_new_skill_storage(...)`, which isn't defined in this module — that's the `lint-backend` failure (ruff `F821 Undefined name`), and the same undefined name made the handler raise `NameError` at runtime, which is what broke `backend-unit-tests` and `backend-blocking-io`.

**Fix:** use `_get_user_skill_storage(config)` inside the worker thread — the same user-scoped accessor every other handler in this router uses. I also updated the regression test for the route's current signature (it became admin-only and now takes a `Request`), supplying `request.state.user` the way `tests/blocking_io/test_channel_runtime_config_store.py` does, and seeding the history through that same accessor so the seed and the read resolve to one storage root.

**Verified locally:**
- `make lint` — clean (`ruff check` + `ruff format --check`)
- `pytest tests/blocking_io` — 37 passed
- `pytest tests/ -k skill` — 614 passed, 1 skipped

The only failures left in a full local `pytest tests/` run are 17–18 in `test_client_live.py` / `test_crawl4ai_tools.py` / `test_fastcrw_tools.py`; those are live-network/crawler tests that fail identically with my commit stashed, so they're pre-existing here and unrelated to this change.

backend/tests/blocking_io/test_skills_router.py
ly-wang19: Good catch — the monkeypatch is gone: the test now builds a minimal `SimpleNamespace` config and lets the handler resolve storage through `_get_user_skill_storage(config)`, seeding the history through that same accessor.

backend/tests/blocking_io/test_skills_router.py
ly-wang19: Agreed — `skill_evolution` is not on this handler's path and has been dropped; the config stub is now just `get_skills_path` / `container_path` / `use`.

WillemJiang (approved):

PR #3565, opened 2026-06-13
fix(skills): offload blocking filesystem IO in update_skill and serialize writes

ly-wang19:
## Why

`make detect-blocking-io` (the `blocking-io-guard` SOP, #3503) flags
`PUT /api/skills/{skill_name}`: `update_skill` runs its config read-modify-write
directly on the asyncio event loop — `load_skills()`,
`ExtensionsConfig.resolve_config_path()`, `get_extensions_config()`, the
`open()` + `json.dump` write to `extensions_config.json`, and
`reload_extensions_config()` are all blocking filesystem IO. Under concurrency a
slow disk stalls every other in-flight request on that worker. Same class as the
offloads merged for the custom-agent router (#3457) and channels (#3529).

## What changed

No caller-visible change. The read-modify-write now runs inside a nested sync
function via `asyncio.to_thread` (a `"not_found"` return signals the 404),
guarded by a module-level `asyncio.Lock`.

The lock matters: offloading the RMW removes the implicit serialization the
single-threaded event loop gave it, so without it two concurrent `PUT`s could
interleave and clobber each other's toggle on the shared `extensions_config`
singleton (lost update). The lock restores that serialization — the same pattern
reviewed and accepted on #3552. Status codes and the response body are unchanged.

## Surface area

- [x] **Backend API** — `PUT /api/skills/{name}` (internal perf + concurrency-safety; request/response shape unchanged)

## Bug fix verification

Following the `blocking-io-guard` SOP:

- **Candidate:** `update_skill` config write — `FILE_OPEN`, `DIRECT_ASYNC` → **FIX+ANCHOR**.
- **Re-scan (Step 2):** after the fix `make detect-blocking-io` no longer lists `update_skill`.
- **Anchors + teeth (Step 5):** `backend/tests/blocking_io/test_skills_update_router.py`
  - `test_update_skill_does_not_block_event_loop` — drives the real handler; teeth verified **RED** (pre-fix `BlockingError`) → **GREEN** (offloaded), under `make test-blocking-io`.
  - `test_update_skill_serializes_concurrent_writes` — fires two concurrent calls and asserts a max in-flight RMW count of **1**; teeth verified **RED** (max 2) when the lock is removed → **GREEN** with it. Marked `@pytest.mark.allow_blocking_io` because the strict gate serializes worker threads, which would otherwise mask the overlap the test needs to observe.

## Validation

In `backend/`:

- `python scripts/detect_blocking_io_static.py backend/app/gateway/routers/skills.py` → `update_skill` finding gone.
- `uv run pytest tests/blocking_io/test_skills_update_router.py tests/test_skills_custom_router.py tests/test_skills_validation.py` → **32 passed** (incl. the existing `test_update_skill_refreshes_prompt_cache_before_return`; behavior preserved).

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Ran the `blocking-io-guard` SOP — `make detect-blocking-io` to find the candidate, implemented the `asyncio.to_thread` offload + `asyncio.Lock`, and wrote the two runtime anchors with their red→green teeth checks. I reviewed every line and verified locally with the detector, ruff, and the test runs above.


backend/app/gateway/routers/skills.py
```diff
@@ -310,34 +318,51 @@ async def get_skill(skill_name: str, config: AppConfig = Depends(get_config)) ->
...
+            if next((s for s in skills if s.name == skill_name), None) is None:
+                return "not_found"
+
+            config_path = ExtensionsConfig.resolve_config_path()
+            if config_path is None:
+                config_path = Path.cwd().parent / "extensions_config.json"
+                logger.info(f"No existing extensions config found. Creating new config at: {config_path}")
+
+            extensions_config = get_extensions_config()
+            extensions_config.skills[skill_name] = SkillStateConfig(enabled=request.enabled)
```
willem-bd: This mutates the **shared** `get_extensions_config()` singleton in place, from a worker thread. The `_skills_config_write_lock` only serializes `update_skill` against other `update_skill` calls — it does not block an event-loop reader (MCP/skill/agent-prompt loading) from touching `get_extensions_config()` concurrently. Before this PR this line ran on the loop with no interleaving; offloading it widened the race surface. Build the payload from a snapshot of `.skills` instead of mutating the shared singleton here, then let `reload_extensions_config()` swap it atomically.

willem-bd (commented):
Thanks @ly-wang19. I found one issue that should be addressed before this is ready.

[P2] Shared `extensions_config` singleton mutated in place from a worker thread — races with event-loop readers

- Location: `app/gateway/routers/skills.py` — `_apply_update`, run via `asyncio.to_thread`, line 337 (`extensions_config.skills[skill_name] = SkillStateConfig(...)`)
- Problem: `_apply_update` is offloaded to a worker thread. Inside it, `get_extensions_config()` returns the **cached shared singleton**, and `extensions_config.skills[skill_name] = ...` mutates that singleton's `.skills` dict in place. The module-level `_skills_config_write_lock` is an `asyncio.Lock` that serializes concurrent `update_skill` calls against each other, but it does **not** prevent the event loop (or another thread) from reading `get_extensions_config()` while the worker thread is mid-mutation.
- Evidence: Before this PR the mutation ran on the event loop with no `await` inside the block, so no other event-loop coroutine could interleave. Moving it to a worker thread frees the loop to run readers (MCP/skill/agent-prompt loading all call `get_extensions_config()`) concurrently with the mutation. `get_extensions_config()` is a cached singleton (`extensions_config.py:228-243`), so those readers observe the very object being mutated — a reader iterating `.skills` while a previously-absent skill is first enabled can raise `RuntimeError: dictionary changed size during iteration`; point reads can see a toggled state before the file is written/reloaded.
- Suggested fix: Don't mutate the shared singleton off-loop. Build the to-be-written payload from a **snapshot** — copy `extensions_config.skills`, set the toggle on the copy, serialize the copy — then write the file and let `reload_extensions_config()` (which atomically swaps the singleton pointer) refresh in-memory state. That keeps every cross-thread access read-only.
- Test: a regression test that iterates `get_extensions_config().skills` on the event loop while `update_skill` toggles a previously-absent skill, asserting no `RuntimeError` and that readers see a consistent snapshot.

The offload itself and the `_skills_config_write_lock` (update-vs-update serialization) are correct; the only gap is update-vs-reader on the shared singleton.

WillemJiang:
@ly-wang19 Thanks for your contribution. Please resolve the conflicts with the main branch.

backend/app/gateway/routers/skills.py
```diff
@@ -20,6 +21,13 @@
 
 router = APIRouter(prefix="/api", tags=["skills"])
 
+# Serializes the extensions_config read-modify-write in update_skill. Offloading
+# that RMW to a worker thread removes the implicit serialization the
+# single-threaded event loop provided, so concurrent PUT /skills/{name} calls
+# could otherwise interleave and clobber each other's toggle on the shared
+# extensions_config singleton.
+_skills_config_write_lock = asyncio.Lock()
```
ShenAC-SAC: IMPORTANT: This lock only serializes `update_skill` against itself, but `update_mcp_configuration` in `app/gateway/routers/mcp.py` performs the same `extensions_config.json` read-modify-write without taking it — and this PR is what makes that pair racy.

Why this matters: before this change both RMWs ran on the event loop with no `await` inside the read→write window, so they were implicitly serialized against each other. Once this RMW moves to a worker thread, the loop is free to run a concurrent `PUT /api/mcp/config` mid-window: the worker snapshots `mcp_servers`, the MCP handler writes a newly added server and reloads, then the worker writes its stale snapshot — the new MCP server is silently dropped from disk (the reverse ordering loses the skill toggle instead).

Requested action: make the lock shared by every `extensions_config.json` writer rather than module-local to this router — e.g. host it next to `reload_extensions_config()` in `deerflow.config.extensions_config` and acquire it in both routers. Offloading the MCP RMW can stay a follow-up, but having it acquire the shared lock closes the window now.

backend/app/gateway/routers/skills.py
```diff
@@ -310,34 +318,53 @@ async def get_skill(skill_name: str, config: AppConfig = Depends(get_config)) ->
 async def update_skill(skill_name: str, request: SkillUpdateRequest, config: AppConfig = Depends(get_config)) -> SkillResponse:
```
ShenAC-SAC: IMPORTANT: Before resolving the conflict flagged above — `update_skill` on `main` is no longer the function this diff rewrites, so the rebase is a re-implementation, not a mechanical merge.

`main` has since added admin gating (`require_admin_user`), user-scoped storage (`_get_user_skill_storage`), and a PUBLIC vs CUSTOM/LEGACY split: only PUBLIC skills go through the `extensions_config.json` RMW; CUSTOM/LEGACY toggles write per-user `_skill_states.json` via `UserScopedSkillStorage.set_skill_enabled_state` (plus a fallback RMW branch), and cache invalidation differs per branch. Taking either side wholesale loses something: keeping this PR's `_apply_update` drops the auth + per-user behavior; keeping `main`'s shape drops the offload.

Requested action: re-apply the offload per path on top of `main` — the (shared-)lock-guarded RMW for the PUBLIC branch, and `asyncio.to_thread` around `set_skill_enabled_state` and both `load_skills` calls for the rest. The test anchors need the same rework: they patch `get_or_new_skill_storage`, which the handler on `main` no longer calls (`_get_user_skill_storage`), so post-rebase they would go green against the wrong surface.

ShenAC-SAC (commented):
@ly-wang19 Thanks — the offload, the snapshot fix from the earlier thread, and the red→green anchors all look solid. Two points the existing threads don't cover, detailed inline:

1. **IMPORTANT** — the new lock doesn't cover `update_mcp_configuration`, which does the same `extensions_config.json` read-modify-write. Offloading this side is exactly what opens a cross-router lost-update window, so the lock needs to be shared across writers (inline at the lock declaration).
2. **IMPORTANT** — the conflict with `main` isn't mechanical: `update_skill` there has since gained admin gating and a PUBLIC vs CUSTOM/LEGACY split with per-user state writes, so the offload (and the tests' patch surface) needs re-applying per path rather than merging either side wholesale (inline at the handler).

backend/app/gateway/routers/skills.py
ly-wang19: Fixed. The lock moved to `deerflow.config.extensions_config` as `get_extensions_config_write_lock()` and both routers acquire it; the MCP router's module-local `_mcp_config_write_lock` is gone. Worth noting the MCP RMW was *already* offloaded (`asyncio.to_thread(_apply_mcp_config_update, ...)`), so the cross-router window you describe was already live rather than only being opened by this PR. Added `test_skill_and_mcp_config_writes_are_serialized`: it observes a max in-flight RMW of 2 when the two routers use separate locks, and 1 with the shared lock. One deviation from the literal suggestion — the lock is keyed by the running event loop instead of being a plain module-level singleton, because asyncio primitives bind to the first loop that awaits them, which made a module-level lock raise "bound to a different event loop" as soon as a second loop used it.

backend/app/gateway/routers/skills.py
ly-wang19: Done — reset onto `main` and re-applied the offload per path rather than merging either side. PUBLIC keeps the (now shared-)lock-guarded `extensions_config.json` RMW, offloaded via `asyncio.to_thread`; CUSTOM/LEGACY offloads `set_skill_enabled_state`; the non-user-scoped fallback reuses the same RMW helper (which also de-duplicates the two identical copies on `main`); and both `load_skills` calls plus storage construction are offloaded. Admin gating and the per-branch cache invalidation are left exactly as `main` has them. You were right about the test surface too: the anchors now patch `_get_user_skill_storage` and pass an admin `Request`, so they exercise the handler's real path instead of going green against `get_or_new_skill_storage`.

backend/app/gateway/routers/skills.py
ly-wang19: Still addressed after the re-apply: `_write_extensions_skill_state` builds the write payload from `dict(...)` snapshots of `mcp_servers`/`skills`, so the cached singleton is never mutated in place (and nothing is left applied if the write fails). `test_update_skill_writes_from_snapshot_without_mutating_singleton` guards it with a dict subclass whose `__setitem__` raises.

backend/app/gateway/routers/skills.py
```diff
@@ -20,6 +21,13 @@
 
 router = APIRouter(prefix="/api", tags=["skills"])
 
+# Serializes the extensions_config read-modify-write in update_skill. Offloading
+# that RMW to a worker thread removes the implicit serialization the
+# single-threaded event loop provided, so concurrent PUT /skills/{name} calls
+# could otherwise interleave and clobber each other's toggle on the shared
+# extensions_config singleton.
```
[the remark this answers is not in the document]
ly-wang19: Addressed by removal — that router-local lock and its comment are gone; the lock now lives in `deerflow.config.extensions_config`. Remaining endpoint references use the `/api` prefix.

backend/tests/blocking_io/test_skills_update_router.py
```diff
@@ -0,0 +1,134 @@
...
+read-modify-write, and the reload are blocking filesystem IO; they are offloaded
+via ``asyncio.to_thread`` and serialized with a module-level ``asyncio.Lock`` so
+concurrent ``PUT /skills/{name}`` calls cannot clobber each other's toggle on the
+shared ``extensions_config`` singleton.
+
+- ``test_update_skill_does_not_block_event_loop``: the strict Blockbuster gate
+  fails if the config write regresses back onto the loop (teeth: red pre-fix).
+- ``test_update_skill_serializes_concurrent_writes``: two concurrent calls observe
+  a max in-flight RMW count of 1 — red if the lock is removed.
+
```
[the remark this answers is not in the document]
ly-wang19: Addressed — the module docstring now enumerates all four tests, including the new cross-router serialization anchor.

ly-wang19:
Thanks @ShenAC-SAC @WillemJiang — conflicts resolved and both IMPORTANT points addressed in `0012b25d`. Per @ShenAC-SAC's note this was **re-applied on top of `main`, not merged**, since `update_skill` there has changed shape.

**1. Shared lock across writers.** The RMW lock now lives next to `reload_extensions_config()` as `get_extensions_config_write_lock()`, and both the skills router and the MCP router acquire it; the MCP router's module-local `_mcp_config_write_lock` is removed. Note the MCP RMW was *already* offloaded, so that cross-router window was already live rather than newly opened here. New anchor `test_skill_and_mcp_config_writes_are_serialized` observes a max in-flight RMW of **2 with separate locks and 1 with the shared lock**.

One deviation from the literal suggestion: the lock is keyed by the running event loop rather than being a plain module-level singleton. asyncio primitives bind to the first loop that awaits them, so a module-level `asyncio.Lock()` raised `bound to a different event loop` as soon as a second loop touched it. Happy to change the shape if you'd prefer something else.

**2. Offload re-applied per path.**
- PUBLIC → shared-lock-guarded `extensions_config.json` RMW, offloaded via `asyncio.to_thread`, payload built from a snapshot so the cached singleton is never mutated in place.
- CUSTOM/LEGACY → `set_skill_enabled_state` offloaded; the non-user-scoped fallback reuses the same RMW helper (which also de-duplicates the two identical copies currently on `main`).
- Both `load_skills` calls + storage construction offloaded.
- Admin gating and per-branch cache invalidation left exactly as `main` has them.

The test anchors were reworked accordingly — they patch `_get_user_skill_storage` and pass an admin `Request`, so they no longer go green against a surface the handler doesn't call.

**Verified locally:** `make lint` clean · `pytest tests/blocking_io` 48 passed · `pytest tests/ -k "skill or mcp"` 1104 passed, 2 skipped.

fancyboi999 (changes requested):
Thanks @ly-wang19. I found one issue that should be addressed before this is ready.

[P1] Keep serialization owned by the worker until the config RMW has finished

- Location: `backend/app/gateway/routers/skills.py:476-491` and `backend/app/gateway/routers/mcp.py:464-466` (the `async with get_extensions_config_write_lock()` blocks around `asyncio.to_thread`)
- Problem: the asyncio lock protects the awaiting task, not the worker-thread read-modify-write. If the request task is cancelled while `to_thread` is running, the context manager releases the lock immediately, but Python cannot stop the underlying thread. A second skill or MCP writer can then acquire the lock and operate on `extensions_config.json` concurrently with the first worker, reopening the lost-update window this PR is intended to close. The cancelled path can also commit a config change without reaching the subsequent prompt/MCP cache invalidation.
- Evidence: a current-head runtime probe blocked the first RMW inside its worker, cancelled the awaiting task, and then started a second writer through the same `get_extensions_config_write_lock()` path. The second writer entered before the first worker was released, and the observed maximum concurrent RMW count was `2`. The four new serialization tests pass because none cancels an in-flight writer.
- Suggested fix: put a shared process-wide `threading.Lock` around the complete synchronous RMW and acquire it inside the worker used by both routers, so the worker retains ownership until the file write and reload really finish even if its caller is cancelled. This also avoids the per-event-loop locks allowing concurrent writers from different loops. An async design is viable only if cancellation explicitly waits for the worker to finish before releasing the lock.
- Test: add a cancellation regression that pauses the first worker after its snapshot, cancels the route task, starts the other router's writer, and asserts the second RMW cannot enter until the first exits; then verify that both intended config changes survive and the required cache invalidation still occurs.

ly-wang19:
Thanks @fancyboi999 — correct on both counts, and the same point you raised on #3560. Fixed in `38710d8a`.

**What was wrong.** The `asyncio.Lock` was held around `await asyncio.to_thread(...)`, so it guarded the awaiting *task*, not the worker thread doing the RMW. Cancel the task and the context manager releases immediately while the thread keeps writing, letting the other router's writer straight in. Your second observation was also right and I'd underweighted it: keying the lock per event loop meant writers on different loops got *different* locks and never excluded each other at all.

**Fix.** `extensions_config_write_lock` is now a process-wide `threading.Lock`, acquired **inside** the worker that performs the RMW — `_write_extensions_skill_state` in the skills router and `_apply_mcp_config_update` in the MCP router. Ownership belongs to the thread doing the writing and is held until the write and reload actually finish, whatever happens to the caller. It also has no event-loop affinity, so the cross-loop hole is gone. Neither handler wraps the `await` in a lock any more.

**Cancellation regression** (`test_cancelled_writer_keeps_the_lock_until_its_worker_finishes`) follows your outline: pause the skills worker inside the lock (after its write), cancel the route task, then start the MCP writer and assert it cannot enter until the first worker is released; it also asserts the ordering `['skills-enter', 'skills-exit', 'mcp-enter']` and that the surviving writer still ran its cache invalidation. Against the previous design it fails exactly as your probe described:

```
AssertionError: MCP writer entered while the cancelled worker was still inside: ['skills-enter', 'mcp-enter']
```

**One thing your review surfaced that I'd have missed:** the two existing serialization tests passed by *replacing* the worker functions with tracking stubs. That was fine while the lock lived in the caller, but with the lock inside the worker those stubs bypassed the very thing under test — they'd have stayed green against a broken lock. Both now instrument *inside* the real workers (each module's `reload_extensions_config`, the last step under the lock) so the production lock is actually exercised.

**Not fixed, and I don't think it can be here:** a request cancelled mid-write still commits its config change without reaching the prompt/MCP cache invalidation. The lock stops the concurrent-writer corruption, but a cancellation between the file write and the invalidation is inherent to doing those as two steps. Happy to follow up separately if you want the invalidation moved inside the worker.

**Verified locally:** `make lint` clean · `pytest tests/blocking_io` 49 passed · `pytest tests/ -k "skill or mcp or config"` 1687 passed. (Remaining failures in a full run — `test_client_live`, `test_crawl4ai_tools`, `test_fastcrw_tools`, `test_checkpointer` — reproduce identically with my files reverted to `origin/main`.)

ly-wang19:
Rebased onto `main` — conflicts resolved.

`main` switched the `extensions_config.json` serialization to `ExtensionsConfig.to_file_dict()` (`model_dump(by_alias=True)`, which also emits `middlewares` and any other top-level keys), while this branch had moved the read-modify-write into the worker-owned, lock-guarded `_write_extensions_skill_state` helper.

I kept the offloaded helper and **adopted `to_file_dict()` inside it**, applied to a `model_copy(deep=True)` so the cached singleton is still not mutated in place. That was the non-mechanical part: keeping this branch's old manual `{"mcpServers", "skills"}` dict would have silently dropped `main`'s new top-level fields (e.g. `middlewares`) from the written file. The snapshot anchor now builds a real `ExtensionsConfig` and asserts both that the new skill exists only in the serialized copy (singleton untouched) and that unrelated top-level keys survive the write.

Verified locally: `make lint` clean · `pytest tests/blocking_io` 54 passed. (The failures in a full `pytest tests/` run — `test_client_live`, `test_crawl4ai_tools`, `test_fastcrw_tools`, `test_checkpointer`, `test_skillscan_native` — reproduce in isolation / with these files reverted to `origin/main`, i.e. they're environment/pre-existing, not from this change.)

backend/app/gateway/routers/skills.py
```diff
@@ -410,6 +410,38 @@ async def get_skill(skill_name: str, config: AppConfig = Depends(get_config)) ->
...
+        # singleton in place would publish the new state to readers before it is
+        # durable on disk, and leave it applied even if the write below fails.
+        # to_file_dict() serializes the full extensions_config.json shape (all
+        # top-level keys), so no field is dropped from the file.
+        extensions_config = get_extensions_config().model_copy(deep=True)
+        extensions_config.skills[skill_name] = SkillStateConfig(enabled=enabled)
+
+        config_data = extensions_config.to_file_dict()
+
+        with open(config_path, "w", encoding="utf-8") as f:
```
willem-bd, line 438: **Suggestion - the config write is non-atomic; a crash mid-write leaves extensions_config.json torn/empty.** open(path, "w") truncates before json.dump writes, so an interruption (process kill, OOM) - or a concurrent reload from a non-locked writer (e.g. deerflow.client, cross-process) - can leave it unparseable, dropping all MCP servers + skill states. The lock prevents lost updates between in-process writers but not a torn write. Pre-existing (old inline PUBLIC branch wrote the same way), not a regression - but this PR is a write-safety pass, so it's the natural complement. deerflow.client._atomic_write_json already implements the safe pattern (temp file + Path.replace); reusing it here and in _apply_mcp_config_update (identical open(...,"w") write) would close the gap.

backend/tests/blocking_io/test_skills_update_router.py
```diff
@@ -0,0 +1,273 @@
...
+            counters["active"] -= 1
+
+    _patch_config_infra(monkeypatch, tmp_path / "extensions_config.json", reload_hook=_tracking_reload)
+
+    await asyncio.gather(
+        update_skill("demo-skill", SkillUpdateRequest(enabled=False), _admin_request(), SimpleNamespace()),
+        update_skill("demo-skill", SkillUpdateRequest(enabled=True), _admin_request(), SimpleNamespace()),
+    )
+
+    # The shared asyncio.Lock must serialize the offloaded read-modify-write.
```
willem-bd, line 144: **Nit - stale lock type.** Says asyncio.Lock, but the lock under test is extensions_config_write_lock (a threading.Lock) - and this PR's thesis (test_cancelled_writer_keeps_the_lock_until_its_worker_finishes below) is that asyncio.Lock is the *wrong* choice. Worth correcting so the comment doesn't contradict the design it anchors.

willem-bd (commented):
Follow-up review on the current head (f01cf66). Prior concerns all look resolved: RMW builds from a snapshot (model_copy(deep=True)), the shared threading.Lock is acquired inside the worker (cancellation-safe - fancyboi999's P1), and it serializes both routers (ShenAC-SAC's point). Verified reload/get_extensions_config/to_file_dict don't take the lock, so no deadlock. Two minor notes inline, neither a blocker.

WillemJiang (approved):

PR #3570, opened 2026-06-14
fix(frontend): cap deeply nested list indentation to prevent render crash (#3393)

### LINKED ISSUE #3393: [bug] Uncaught RangeError: Maximum call stack size exceeded
### Before you start

### Problem summary

在frontend前端，当单个消息内容过大（也肯能是markdown格式原因？）导致界面崩溃

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

在frontend前端，当单个消息内容过大（也肯能是markdown格式原因？）导致界面崩溃。

Image

Image

### Expected behavior

应该不崩才对。

### Steps to reproduce

1. 对话产生大量内容
2. token统计，共使用527k用量，单个文本内容大概2万字

### Relevant logs

```shell
日志一切正常
```

Other

### Operating system

Linux

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

Eilen6316:
### Summary

Fixes #3393 — a large/deeply-structured assistant message crashes the chat route with an uncaught `RangeError: Maximum call stack size exceeded`.

The stack trace from the issue points straight at `marked`'s recursive **list** tokenizer, called inside Streamdown's lexing `useMemo`:

```
at RegExp.exec (<anonymous>)
at er.lheading
at e.blockTokens
at er.list        ← list tokenizer recursing per nesting level
at e.blockTokens
at e.lex          ← inside Streamdown's useMemo
```

`marked` recurses once per list nesting level (`list -> blockTokens -> list -> ...`). In the browser's tighter stack a deeply nested list overflows during render and throws the uncaught `RangeError`; on larger stacks the same input instead goes quadratic and exhausts the heap (which the render error boundary added in #3502 **cannot** catch).

#3502 already caps deeply nested blockquotes (`capBlockquoteNesting`) but left the list path — named in the same code comment — unguarded.

### What this changes

- Adds `capListNesting()` in `core/streamdown/preprocess.ts`, mirroring the existing blockquote guard: when (and only when) a line is indented past a pathological threshold, it clamps leading whitespace to **200 columns (~100 nesting levels)** — far beyond any legitimate content, while keeping `marked` safe. Fenced code blocks are left untouched so literal indentation isn't corrupted.
- Adds `capMarkdownNesting()` that composes the blockquote + list guards, and wires it into `ClipboardSafeStreamdown`.
- Normal content returns the same string reference (the hint regex short-circuits), so there is no behavior change for ordinary messages.

### Testing

- `pnpm test` — added 5 unit tests for `capListNesting` / `capMarkdownNesting` (normal content unchanged, pathological indentation capped, fenced code preserved, only pathological lines rewritten). All 13 tests in the file pass.
- `pnpm check` — lint + typecheck pass.
- Verified against `marked@16` directly: a 2000-level nested list goes from OOM to ~300ms after capping.

### DIFF
--- a/frontend/src/core/streamdown/preprocess.ts
+++ b/frontend/src/core/streamdown/preprocess.ts
@@ -18,6 +18,18 @@ const BLOCKQUOTE_PREFIX_RE = /^ {0,3}(?:[ \t]*>)+/;
 const CODE_FENCE_RE = /^ {0,3}(?:```|~~~)/;
 const INDENTED_CODE_RE = /^(?: {4}|\t)/;
 
+// marked's list tokenizer recurses once per nesting level too (list ->
+// blockTokens -> list -> ...). In the browser's tighter stack a deeply nested
+// list overflows during render and throws "Maximum call stack size exceeded"
+// from inside Streamdown's lexing useMemo (see issue #3393); on larger stacks
+// the same input instead goes quadratic and exhausts the heap. Each list level
+// requires at least ~2 columns of indentation, so capping leading whitespace at
+// 200 columns bounds the effective nesting near 100 levels — far beyond any
+// legitimate content while keeping marked safe. Anything indented past this is
+// pathological nesting, not prose or code.
+const MAX_LIST_INDENT = 200;
+const DEEP_INDENT_HINT_RE = new RegExp(`^[ \\t]{${MAX_LIST_INDENT + 1},}`, "m");
+
 export function capBlockquoteNesting(markdown: string): string {
   if (!DEEP_BLOCKQUOTE_HINT_RE.test(markdown)) {
     return markdown;
@@ -53,6 +65,39 @@ export function capBlockquoteNesting(markdown: string): string {
       return line;
     })
     .join("\n");
+}
+
+export function capListNesting(markdown: string): string {
+  if (!DEEP_INDENT_HINT_RE.test(markdown)) {
+    return markdown;
+  }
+
+  let insideFence = false;
+  return markdown
+    .split("\n")
+    .map((line) => {
+      if (CODE_FENCE_RE.test(line)) {
+        insideFence = !insideFence;
+        return line;
+      }
+      // Indentation inside fenced code is literal layout (ASCII art, pasted
+      // source); collapsing it would corrupt the rendered block.
+      if (insideFence) {
+        return line;
+      }
+      const whitespace = /^[ \t]*/.exec(line)![0];
+      if (whitespace.length <= MAX_LIST_INDENT) {
+        return line;
+      }
+      return " ".repeat(MAX_LIST_INDENT) + line.slice(whitespace.length);
+    })
+    .join("\n");
+}
+
+// Cap every runaway nesting construct that can take down a message render
+// before marked sees the content.
+export function capMarkdownNesting(markdown: string): string {
+  return capListNesting(capBlockquoteNesting(markdown));
 }
 
 export function preprocessStreamdownMarkdown(markdown: string): string {
--- a/frontend/src/core/streamdown/preprocess.ts
+++ b/frontend/src/core/streamdown/preprocess.ts
@@ -73,34 +73,16 @@ export function capListNesting(markdown: string): string {
   }
 
   let insideFence = false;
-  // A markdown indented code block opens only after a blank line, so a deeply
-  // indented line that follows blank/code lines is literal code, not list
-  // nesting. We track that context with `afterBlank` so we exempt genuine
-  // indented code without also exempting (and thus failing to cap) the
-  // pathological deep-indent list lines this function guards against.
-  let afterBlank = false;
-  let insideIndentedCode = false;
   return markdown
     .split("\n")
     .map((line) => {
       if (CODE_FENCE_RE.test(line)) {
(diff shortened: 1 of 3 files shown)


WillemJiang:
@Eilen6316, please fix the lint error.

frontend/src/core/streamdown/preprocess.ts
```diff
@@ -55,6 +67,39 @@ export function capBlockquoteNesting(markdown: string): string {
...
+  return markdown
+    .split("\n")
+    .map((line) => {
+      if (CODE_FENCE_RE.test(line)) {
+        insideFence = !insideFence;
+        return line;
+      }
+      // Indentation inside fenced code is literal layout (ASCII art, pasted
+      // source); collapsing it would corrupt the rendered block.
+      if (insideFence) {
```
fancyboi999, line 85: Parity nit (non-blocking): `capBlockquoteNesting` also exempts `INDENTED_CODE_RE` lines (`preprocess.ts:48`), but `capListNesting` only checks `insideFence` here. So a non-fenced indented code block (4-space) whose content line carries >200 leading spaces would be capped to 200 and mildly corrupted. Indented code can't drive the list recursion you're guarding against, so widening this to `if (insideFence || INDENTED_CODE_RE.test(line))` would mirror the blockquote cap and be strictly safer with no downside. Rare input, and your comment notes you treat >200 indent as always-pathological — so your call; just flagging the one spot the two caps diverge.

fancyboi999 (commented):
Traced this against #3502's `capBlockquoteNesting`, which it deliberately mirrors. The fix is sound and lands at the right layer:

- **Root cause** matches the stack trace in #3393 — `marked`'s list tokenizer recurses per nesting level and Streamdown lexes inside a `useMemo`, so a deeply nested list overflows the stack (or goes quadratic and exhausts the heap, which the #3502 render error boundary can't catch). Capping leading whitespace *before* `marked` sees the content is the same proven approach as the blockquote cap.
- **`capListNesting`** (`preprocess.ts:70-93`) is structurally faithful to the blockquote version: the `DEEP_INDENT_HINT_RE` fast-path (`:31`) keeps the per-line scan off the hot path for normal content, fence state is tracked so fenced ASCII art / pasted source is preserved (`:85`), and only lines with >200 leading whitespace are rewritten to exactly 200 columns (`:88-92`).
- **Wiring** (`streamdown.tsx:68`) is a clean swap to `capMarkdownNesting`, and the composition `capListNesting(capBlockquoteNesting(...))` (`:99`) keeps the two caps independent.
- **Tests** cover the paths that matter — normal nesting left untouched, pathological depth capped to 200, fenced content preserved, mixed pathological/plain lines, and the combined blockquote+list case. (Reviewed at the code level; I didn't run the vitest suite locally.)

One non-blocking parity note inline (`:85`).

Minor, not worth changing: the comment reasons in "columns" while the cap counts characters. For space-indented input — every realistic case here — those are identical. Tab-indented input expands ~4 columns/char, so the effective level-bound is looser for tabs, but still far below `marked`'s crash threshold, so it's harmless; just flagging that the "near 100 levels" math is space-specific.

Looks good to merge.

frontend/src/core/streamdown/preprocess.ts
Eilen6316: Good call! Fixed the condition to cover indented code blocks as you suggested, matches the blockquote cap logic now. Ready for recheck.

Eilen6316: Update after testing this more carefully — I ended up reverting this change (839b2af). Two problems surfaced:

1. **The literal guard no-ops the function.** `INDENTED_CODE_RE` is `/^(?: {4}|\t)/`, which matches *every* line with 4+ leading spaces — i.e. exactly the deep-indent lines `capListNesting` is meant to cap. `insideFence || INDENTED_CODE_RE.test(line)` therefore exempts all the pathological lines and the cap stops doing anything (it fails the existing `caps pathologically deep list indentation` test). The parity holds for `capBlockquoteNesting` only because blockquote markers take ≤3 leading spaces, so deep-quote lines never trip `INDENTED_CODE_RE` — that property doesn't carry over to list indentation.

2. **A context-aware exemption reopens the crash.** I tried the stricter reading (only treat a 4+-space line as indented code when it actually opens/continues an indented-code block, i.e. after a blank line). That exempts genuine indented code, but it also exempts *blank-separated* deeply-nested list items — and those still blow up marked. Verified against `marked@16`: a blank-line-separated nested list OOMs at depth ~1.5k, same failure class as the contiguous case. Line-by-line we can't distinguish "indented code" from "deeply nested list content" (both can sit 4+ spaces in after a blank line), so any indentation-based exemption is exploitable.

So I kept the original behavior: cap all deep indentation outside fenced code. The only cost is mild corruption of a >200-column indented-code line, which doesn't occur in real content and is strictly preferable to a render crash. Added a regression test (`caps deep indentation even after a blank line`) to lock the blank-line case. Thanks for the careful read — it was the right thing to dig into, just landed on the opposite conclusion.

WillemJiang (approved):

PR #3571, opened 2026-06-14
fix(backend): persist partial answer when a run is interrupted mid-stream (#3403)

### LINKED ISSUE #3403: [feat] 支持被取消的半截对话显示
### Before you start

### Problem / motivation

在发送问题之后，中途cancel。前端会保留这部分内容，如图一，但是后端并没有持久化下来。刷新之后或者重新对话之后，按后端的数据刷新之后，ai回答的半截会消失，如图二。体验上不好。

Image

Image

### Proposed solution

1. 支持run_events 存储这种还没有回答完成的对话
2. 前端历史会话恢复先查run_events。langGraph的checkpoint最多作为补充，这样后续可能更灵活

### Affected area(s)

### Alternatives considered

_No response_

### Additional context

_No response_

Eilen6316:
### Summary

Fixes #3403 — when a user cancels a streaming response midway, the partial AI answer shown in the UI disappears after a refresh.

**Root cause:** thread history is rebuilt from the `run_events` store, but `RunJournal` only writes an AI message on `on_llm_end`. A cancel stops streaming mid-generation, so `on_llm_end` never fires for the in-flight LLM call and its partial text — already streamed to the client — is never persisted. (The `>` blockquote framing in the issue thread aside, this is the data-consistency gap the maintainer pointed at.)

### What changes

- `run_agent` (`runtime/runs/worker.py`) accumulates streamed `messages`-mode AI chunks by message id during the stream (consecutive `AIMessageChunk`s for one response merge with `+`).
- On an **interrupt** cancel — but **not** a `rollback`, which intentionally discards the run — the worker hands the not-yet-completed partials to the new `RunJournal.record_interrupted_ai_messages()`, right before flushing the journal in the `finally` block.
- `record_interrupted_ai_messages()` writes each partial in the **same** `llm.ai.response` / `category="message"` shape that `on_llm_end` produces (so history rebuilds it identically), flagged `interrupted: true`. It skips ids already completed via `on_llm_end` (tracked in a new `_completed_message_ids` set) and dedups its own writes, so nothing is duplicated. Incomplete tool calls are dropped (their args may be truncated mid-JSON); user-visible text and reasoning content are kept.

No new storage layer is introduced — the reporter's suggested "store run_events for unfinished turns" is exactly the existing `run_events` mechanism; this just fills the partial in on interrupt. Normal (non-cancelled) runs are unaffected.

### Testing

- `tests/test_run_journal.py::TestInterruptedPartialMessages` — persists a partial; skips a message already completed via `on_llm_end`; skips empty chunks; keeps reasoning-only chunks; dedups repeated emits.
- `tests/test_worker_partial_persist.py` — chunk accumulation by id, merge of same-id chunks, and ignoring of non-`messages` modes / non-AI chunks / id-less chunks.
- All 11 new tests pass; `ruff check` and `ruff format --check` clean on the changed files.

> Note: a handful of pre-existing tests in `test_run_journal.py` / `test_run_worker_rollback.py` fail in my sandbox due to a `langgraph` version mismatch (`No module named 'langgraph.runtime'`); I confirmed they fail identically on `main` without this change, so they're unrelated to this PR.

### DIFF
--- a/backend/packages/harness/deerflow/runtime/journal.py
+++ b/backend/packages/harness/deerflow/runtime/journal.py
@@ -82,6 +82,12 @@ class RunJournal(BaseCallbackHandler):
         self._counted_external_source_ids: set[str] = set()
         self._counted_message_llm_run_ids: set[str] = set()
 
+        # Message ids already persisted by on_llm_end (a completed LLM call) and
+        # by record_interrupted_ai_messages (a cancelled-mid-stream partial), so
+        # an interrupt never re-writes a message that already reached the store.
+        self._completed_message_ids: set[str] = set()
+        self._partial_message_ids: set[str] = set()
+
         # Convenience fields
         self._last_ai_msg: str | None = None
         self._first_human_msg: str | None = None
@@ -261,6 +267,12 @@ class RunJournal(BaseCallbackHandler):
         for message in messages:
             caller = self._identify_caller(tags)
 
+            # Remember the id so an interrupt does not re-persist this message
+            # as a "partial" — on_llm_end means the call completed in full.
+            completed_id = getattr(message, "id", None)
+            if completed_id:
+                self._completed_message_ids.add(completed_id)
+
             # Latency
             rid = str(run_id)
             start = self._llm_start_times.pop(rid, None)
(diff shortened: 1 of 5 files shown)


fancyboi999 (commented):
Traced the full path on this — it's a clean, surgical fix for #3403 and the persistence shape is right:

- **`_accumulate_partial_ai_chunk`** (`worker.py:714`) only touches `messages`-mode `AIMessageChunk`s, keys by message id, and merges consecutive chunks with `+` — the correct way to rebuild the streamed partial. Good unit coverage in `test_worker_partial_persist.py` (by-id merge, separate ids, and the non-messages / non-AI / no-id / bare-chunk guards).
- **`record_interrupted_ai_messages`** (`journal.py:347`) writes each partial in the *exact* shape `on_llm_end` uses — `event_type="llm.ai.response"`, `category="message"`, `content=model_dump()`, plus `_record_message_summary` — so history rebuilds it identically. Dedup via `_completed_message_ids` (populated in `on_llm_end`) and `_partial_message_ids` correctly stops an interrupt from re-writing a message that already completed, and the empty content/reasoning skip avoids persisting tool-call-only chunks.
- **The cancel gate** (`worker.py:413`: `partial_ai_chunks and record.abort_event.is_set() and record.abort_action != "rollback"`) correctly skips rollback cancellations and is wrapped so a persistence failure can't break the cancel path. Dropping partial (truncated-JSON) tool calls is the right call — they're not resumable.

One non-blocking note inline about the hardcoded `caller`.

Heads-up — not a defect in this PR, but worth flagging: **#3572 also targets "Fixes #3403"** and changes the same `journal.py` + `worker.py`, so the two will conflict. The scope differs: this PR persists the partial to `run_events` only — which fixes the literal "answer vanishes on refresh" symptom — whereas #3572 also writes the partial into the LangGraph checkpoint (so it lands in the next-turn agent context) and synthesizes closure ToolMessages for interrupted tool calls. Whether #3403 wants the minimal run_events fix or the broader one is a scope call for the maintainer; flagging so they don't get merged on top of each other.

The core fix here is correct and merge-ready for the refresh symptom. (Reviewed at code level; I didn't run the suite locally.)

backend/packages/harness/deerflow/runtime/journal.py
```diff
@@ -332,6 +344,55 @@ def on_llm_end(
...
+                response_metadata=dict(getattr(chunk, "response_metadata", {}) or {}),
+            )
+
+            if message_id:
+                self._partial_message_ids.add(message_id)
+            self._put(
+                event_type="llm.ai.response",
+                category="message",
+                content=message.model_dump(),
+                metadata={"caller": "lead_agent", "interrupted": True},
```
fancyboi999, line 390: Non-blocking: `on_llm_end` derives the caller dynamically (`caller = self._identify_caller(tags)`, e.g. `journal.py:268`), but here it's hardcoded `caller="lead_agent"` — also passed to `_record_message_summary`. If an interrupted partial can belong to a subagent LLM call (i.e. subagent `AIMessageChunk`s reach the `messages`-mode accumulator), it'd be mis-attributed to the lead agent; if only lead-agent chunks ever reach it, this is fine. Either way worth resolving the divergence from `on_llm_end` (reuse `_identify_caller`, or a short comment on why lead_agent is always correct here).

WillemJiang:
The feature request is invalid.

PR #3572, opened 2026-06-14
feat(runtime): persist partial stream output on cancel

### LINKED ISSUE #3403: [feat] 支持被取消的半截对话显示

AnoobFeng:
Accumulate streamed AIMessageChunk data during runs and persist interrupted responses when a run is cancelled.

- write partial AI messages to run_events and checkpoint on interrupt
- synthesize ToolMessage closures for interrupted tool calls
- handle tool execution cancellation after on_llm_end without duplicating AIMessage
- scope closure synthesis to the current run to avoid modifying historical dangling tool calls
- skip partial persistence for rollback cancellation
- add unit and worker-level coverage for partial persistence paths

Fixes #3403

## Why

When a user stops a streaming AI response, the frontend keeps the already-rendered partial answer in React state, but the backend previously did not persist that partial output. After refresh or reopening the thread, the backend-hydrated state no longer contained the interrupted answer, so the response appeared to disappear.

This PR makes interrupted stream output durable and keeps the checkpoint/run_events views consistent enough for refresh, history loading, and the next agent turn.

## What changed

- Interrupted assistant responses are now persisted on cancel.
- Partial AI messages are written to both `run_events` and the LangGraph checkpoint.
- Open tool calls from interrupted responses are closed with synthetic `ToolMessage(status="error")` entries.
- If the AIMessage was already recorded by `on_llm_end` but its tool was still executing, only the missing synthetic ToolMessage is added.
- Synthetic tool closures are scoped to AI messages emitted by the current run, so older dangling tool calls or HITL pending state are not modified.
- Rollback cancellation keeps its previous behavior and does not persist partial output.
- No frontend UI changes are included; partial responses render as normal messages.

## Surface area

## Screenshots / Recording

N/A. This PR does not change frontend UI.

## Bug fix verification

  - `backend/tests/test_partial_persist.py`
  - `backend/tests/test_run_worker_partial_persist.py`

  - Not run against `main` locally. The new tests encode the previously missing persistence and tool-call closure behavior.

  - The behavior is primarily backend persistence state, so it is covered with unit tests for the persistence helper and worker-level tests for cancel handling.

## Validation

```bash
cd backend
uv run pytest tests/test_partial_persist.py tests/test_run_worker_partial_persist.py tests/test_run_worker_rollback.py tests/test_run_journal.py


## AI assistance

Tool(s) used: Claude Code
How you used it: Used Codex for code review, risk analysis, and PR description drafting. Implementation was reviewed against the issue and proposal, including partial persistence, checkpoint/run_events consistency, rollback behavior, and interrupted tool-call closure.

### CODE REVIEW

**Review by fancyboi999 — COMMENTED** [review]
Reviewed the architecture and integration points: the journal hooks (`journal.py:454 recorded_ai_message_ids` + `:463 record_partial_response`), the worker wiring (`worker.py:697 _feed_partial_accumulator` / `:715 _persist_partial_safely`, called in both cancel branches), the `partial_persist.py` public surface (`PartialMessageAccumulator`, `find_open_tool_calls` / `build_closure_tool_messages`, `append_messages_to_checkpoint:333`), the middleware refactor, and the design proposal. Honest scope: I read the integration + the module's public surface, but did not line-by-line audit the full 605-line module or the ~1,150 lines of new tests — so treat this as an architecture/scoping review, not a full correctness sign-off.

What's genuinely good here:
- Persisting the partial into **both** `run_events` and the LangGraph **checkpoint** (`append_messages_to_checkpoint`, with the `new_versions` channel-version bump so savers don't discard the write) means the partial survives a refresh *and* is in the next-turn agent context — a real step beyond run_events-only.
- Synthesizing closure ToolMessages for interrupted tool calls, and reusing `DanglingToolCallMiddleware`'s text via the new `INTERRUPTED_TOOL_MESSAGE_CONTENT` constant instead of duplicating the string, is a clean touch.
- `_persist_partial_safely` swallowing exceptions so the cancel path stays deterministic (run still reaches `interrupted`, bridge still emits `end`) is the right instinct.

The thing to resolve before either lands: **#3571 also targets "Fixes #3403"** and edits the same `journal.py` + `worker.py`, so the two conflict. #3571 is the minimal version (~235 lines, persist to `run_events` only) and covers the literal "vanishes on refresh" symptom; this PR is the comprehensive version (~2,300 lines + design doc: checkpoint + tool closures + token accounting). Both can't merge as-is — worth the maintainer settling scope first: does #3403 want the surgical fix with checkpoint/tool-closures as a follow-up, or this end-to-end approach? Coordinating now saves one of you a painful rebase.

One area I'd want extra eyes on, since it's the riskiest part: `append_messages_to_checkpoint`'s version bumping is saver-specific (`get_next_version`, with a lexicographic uuid6 fallback). A wrong `new_versions` silently drops the appended `messages` payload on read, so it's worth confirming in CI against the SQLite/Postgres savers, not just InMemorySaver.

Ambitious, well-structured work — the main blocker is the overlap with #3571, not the code itself.

**Comment by AnoobFeng** [comment]
@fancyboi999 
Thanks for your good code review, I agree with your assessment.

The conflict with #3571 is a critical blocker, so I’ll hold off on the rebase and sync with the maintainers first to align on the final scope.

Regarding the high-risk area you pointed out: I actually verified this against Postgres during development to ensure the behavior holds, but you're absolutely right that our automated test suite currently lacks coverage for real databases. >
Maybe I work on supplementing the integration tests with actual SQLite and Postgres savers to make sure this is fully bulletproof in CI and rule out any silent drop risks.

Appreciate the stellar feedback.

**Comment by WillemJiang** [comment]
@AnoobFeng please fix the lint error.

File: backend/packages/harness/deerflow/runtime/runs/partial_persist.py [hunk 1]
```diff
@@ -0,0 +1,609 @@
...
+                "id": itc.get("id"),
+                "name": itc.get("name") or "unknown",
+                "args": {},
+                "invalid": True,
+                "error": itc.get("error"),
+            }
+        )
+    return result
+
+
```
willem-bd, line 304: **Subagent-tagged chunks may leak into the lead-agent partial on cancel**

`is_middleware_chunk` only filters chunks whose tags start with `middleware:`. Per `RunJournal._identify_caller`, subagent invocations are tagged `subagent:<name>`, which this filter lets through. If subagents stream `AIMessageChunk`s under the same `messages` mode while the lead run is active, those chunks would be:

1. Fed into the lead run's `PartialMessageAccumulator` (keyed by message id).
2. On cancel, finalized as if they were a lead-agent partial response and persisted to both `run_events` (as `llm.ai.response` with `caller="lead_agent"`) and the LangGraph checkpoint's `messages` channel.

The user-visible result would be a subagent's mid-thought text surfacing inside the chat history of the lead conversation — exactly the kind of "ghost reply" the `middleware:*` filter is there to prevent (the docstring even calls out "half-generated thread titles" as the reason).

Two questions worth confirming before merge:

- Do subagent invocations actually stream through this same `messages` path on the lead run's bridge, or are they on a separate stream that this code never sees? If the latter, this is moot.
- If they do share the path, what's the right scope? Probably either:
  - Drop `subagent:*` tags too: `tag.startswith(("middleware:", "subagent:"))`, or
  - Invert the filter to **only** retain chunks tagged `lead_agent` (stricter, but mirrors how `_identify_caller` decides what counts as a lead message).

If you go with the inverted approach, the function name `is_middleware_chunk` should change to something like `is_non_lead_chunk` so the contract stays self-documenting.

A regression test would look like the existing `test_persists_text_only_partial`, but with one of the chunk metadatas set to `{"tags": ["subagent:researcher"]}` — assert that the resulting `run_events` contains zero `llm.ai.response` events for that id.

AnoobFeng: thanks for your code review, let me try review again.

AnoobFeng: I reviewed the current subagent path. the subagents run inside the task tool via their own `agent.astream(..., stream_mode="values")`, and their final output returns to the lead conversation through the task tool result. So I do not think subagent token chunks are part of the normal lead `messages` stream path today.

But the partial accumulator is consuming stream metadata based on the same tag convention used by `RunJournal._identify_caller()`, and the journal already treats `subagent:*` as a non-lead caller. The safer invariant here is that partial persistence should only persist lead-agent output into the user-visible run history/checkpoint.

I updated the filter to make that boundary explicit:

- renamed `is_middleware_chunk()` to `is_non_lead_chunk()`
- filter both `middleware:*` and `subagent:*`
- keep `lead_agent` and untagged chunks eligible, matching the journal's default-to-lead behavior
- added regression coverage ensuring `subagent:researcher` chunks are ignored and do not produce interrupted `llm.ai.response` events

AnoobFeng:
@WillemJiang  I fixed the lint error and pushed. Thanks for the notice

WillemJiang:
The feature request is invalid.

PR #3573, opened 2026-06-14
fix(config): coerce null object config sections to their defaults

ly-wang19:
## Why

#3434 made commented-out **list** sections (`models` / `tools` / `tool_groups`) parse as `[]` instead of crashing — but the same scenario still crashes for **object** sections. Commenting out every key under e.g. `memory:`, `summarization:`, `guardrails:`, `tool_output:` or `run_events:` makes PyYAML parse the value as `None`, and `AppConfig` then raises an opaque `Input should be a valid dictionary` for that section. This breaks the documented `cp config.example.yaml config.yaml` first-run flow (the exact flow #3434 fixed), since `config.example.yaml` ships those object sections and a user disabling one by commenting out its body hits the crash.

Repro on `main`: a config containing `memory:` (null) fails to load with a `ValidationError`.

## What changed

No change for valid configs. A present-but-null section now falls back to its default instead of crashing: a `model_validator(mode="before")` drops `None`-valued top-level sections so each field uses its default (list sections → `[]` via `default_factory`, object sections → their default config). This **generalizes** the previous list-only `field_validator` (now removed as subsumed) and the `database` special-case in `from_file` to every section. Required sections with no default (`sandbox`) still error when null — there is nothing to fall back to.

## Surface area

Internal config loading (`packages/harness/deerflow/config/app_config.py`) — no frontend, endpoint, agent, sandbox, skills, or dependency change; robustness fix for malformed/partial `config.yaml`.

## Bug fix verification

- Test: `backend/tests/test_app_config_reload.py::test_app_config_coerces_commented_out_object_sections`.
- Red on `main` / green on branch: **yes** — on `main`, `AppConfig.from_file` of a config with a null object section raises `ValidationError`; the new test reproduces that and passes on this branch. The existing `test_app_config_coerces_commented_out_list_sections` (#3434) still passes, confirming the generalization preserves list-section behavior.

## Validation

In `backend/`:

- `uv run pytest tests/test_app_config_reload.py` → **12 passed**.
- `uv run ruff check` + `uv run ruff format --check` on the changed files → clean.

## AI assistance

**Tool(s) used:** Claude Code

**How:** Used to spot that #3434's null-section handling was list-only, reproduce the object-section crash, generalize the coercion via a `model_validator`, and add the regression test. I reviewed every line and verified locally with the test + ruff runs above.

### DIFF
--- a/backend/packages/harness/deerflow/config/app_config.py
+++ b/backend/packages/harness/deerflow/config/app_config.py
@@ -7,7 +7,7 @@ from typing import Any, Self
 
 import yaml
 from dotenv import load_dotenv
-from pydantic import BaseModel, ConfigDict, Field, field_validator
+from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 from deerflow.config.acp_config import ACPAgentConfig, load_acp_config_from_dict
 from deerflow.config.agents_api_config import AgentsApiConfig, load_agents_api_config_from_dict
@@ -156,20 +156,28 @@ class AppConfig(BaseModel):
         ),
     )
 
-    @field_validator("models", "tools", "tool_groups", mode="before")
+    @model_validator(mode="before")
     @classmethod
-    def _coerce_null_list_sections(cls, value: Any) -> Any:
-        """Treat a present-but-empty config section as an empty list.
+    def _drop_null_config_sections(cls, data: Any) -> Any:
+        """Treat a present-but-null config section as absent so its default applies.
 
         Commenting out every entry under a top-level YAML key — e.g. ``models:``
-        with only comments beneath it, exactly as shipped in
-        ``config.example.yaml`` — makes PyYAML parse the value as ``None``.
-        Without this, the documented ``cp config.example.yaml config.yaml``
-        first-run flow crashes with an opaque ``Input should be a valid list``
-        pydantic error. Coercing ``None`` to ``[]`` keeps that flow working and
-        matches the field's own ``default_factory=list``.
+        (a list) or ``memory:`` (an object), with only comments beneath it as
+        shipped throughout ``config.example.yaml`` — makes PyYAML parse the value
+        as ``None``. Without this, the documented ``cp config.example.yaml
+        config.yaml`` first-run flow crashes with an opaque ``Input should be a
+        valid list`` / ``valid dictionary`` pydantic error for that section.
+
+        Dropping the ``None`` lets each field fall back to its default: list
+        sections become ``[]`` via ``default_factory=list`` and object sections
+        get their default config. This generalizes the earlier list-only
+        handling (and the ``database`` special-case in ``from_file``) to every
+        section. Required sections without a default (``sandbox``) intentionally
+        still error when null — there is nothing to fall back to.
         """
-        return [] if value is None else value
+        if isinstance(data, dict):
+            return {key: value for key, value in data.items() if value is not None}
+        return data
 
     @classmethod
     def resolve_config_path(cls, config_path: str | None = None) -> Path:
--- a/backend/packages/harness/deerflow/config/app_config.py
+++ b/backend/packages/harness/deerflow/config/app_config.py
@@ -171,9 +171,11 @@ class AppConfig(BaseModel):
         Dropping the ``None`` lets each field fall back to its default: list
         sections become ``[]`` via ``default_factory=list`` and object sections
         get their default config. This generalizes the earlier list-only
-        handling (and the ``database`` special-case in ``from_file``) to every
-        section. Required sections without a default (``sandbox``) intentionally
-        still error when null — there is nothing to fall back to.
+        handling to every section that defines a default. The ``database``
+        section is independent and still owned by ``_apply_database_defaults``
+        (in ``from_file``), which applies concrete defaults beyond null-coercion.
+        Required sections without a default (``sandbox``) intentionally still
+        error when null — there is nothing to fall back to.
         """
         if isinstance(data, dict):
             return {key: value for key, value in data.items() if value is not None}
--- a/backend/tests/test_app_config_reload.py
+++ b/backend/tests/test_app_config_reload.py
@@ -170,6 +170,41 @@ def test_app_config_coerces_commented_out_list_sections(tmp_path, monkeypatch):
     assert config.tool_groups == []
 
 
+def test_app_config_coerces_commented_out_object_sections(tmp_path, monkeypatch):
+    """Commenting out every entry under an object key makes PyYAML parse it as None.
+
+    Same documented ``cp config.example.yaml config.yaml`` flow as the list
+    sections: object sections (memory, summarization, ...) must fall back to
+    their defaults instead of raising ``Input should be a valid dictionary``.
+    """
+    config_path = tmp_path / "config.yaml"
+    extensions_path = tmp_path / "extensions_config.json"
+    _write_extensions_config(extensions_path)
+    config_path.write_text(
+        yaml.safe_dump(
+            {
+                "sandbox": {"use": "deerflow.sandbox.local:LocalSandboxProvider"},
+                "memory": None,
+                "summarization": None,
+                "guardrails": None,
+                "tool_output": None,
+                "run_events": None,
+            }
+        ),
+        encoding="utf-8",
+    )
+    monkeypatch.setenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH", str(extensions_path))
+
+    config = AppConfig.from_file(str(config_path))
+
+    # Each present-but-null object section falls back to its default config.
+    assert config.memory is not None
+    assert config.summarization is not None
+    assert config.guardrails is not None
+    assert config.tool_output is not None
+    assert config.run_events is not None
+
+
 def test_app_config_warns_when_no_models_configured(tmp_path, monkeypatch, caplog):
     config_path = tmp_path / "config.yaml"
     extensions_path = tmp_path / "extensions_config.json"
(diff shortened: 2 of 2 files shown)


fancyboi999 (commented):
Read the full change against the field definitions — it's a correct, well-scoped generalization of the #3434 list-only fix, and the broadening to all sections is safe:

- **`_drop_null_config_sections`** (`app_config.py:161`) swaps the old 3-field `field_validator` for a `model_validator(mode="before")` that strips every present-but-null top-level key, so each section falls back to its default. For the object sections in the bug (`memory` / `summarization` / `guardrails` / `tool_output` / `run_events`, all `default_factory=...`) a null value now yields the default config instead of the opaque `Input should be a valid dictionary`. Matches the documented `cp config.example.yaml config.yaml` flow that #3434 set out to keep working.
- I specifically checked the broadening doesn't regress the **Optional** sections: `checkpointer` (`:144`) and `stream_bridge` (`:151`) are the only `| None` top-level fields, and both are `Field(default=None)` — so dropping a null key resolves to `None`, identical to the previous `| None` parsing. No section silently changes value.
- **`sandbox`** (`:99`) has no default, so a null `sandbox:` still errors — the message just shifts from `Input should be a valid dictionary` to `Field required` (the docstring calls this out). Still fails closed, and the test wisely supplies a valid `sandbox` so it doesn't trip on this.
- `model_config = ConfigDict(extra="allow")` means dropped unknown-null keys are a non-issue — they were only ever kept as `None` extras anyway.

Test covers the object-section case end-to-end through `from_file`, asserting each null section becomes a non-None default.

One forward-looking, non-blocking note: this now bakes in an invariant that *no top-level field treats `null` as a meaningful value distinct from its default* — true today, since the only Optional fields default to `None`. If a future section is ever added as `X | None = Field(default=<non-None>)` where a user's explicit `null` should mean "disabled/None", this validator would silently override it with the non-None default. Might be worth a one-line comment on the validator capturing that assumption so it's not a surprise later.

Looks good to merge. (Reviewed at code level; I didn't run the suite locally.)

backend/packages/harness/deerflow/config/app_config.py
```diff
@@ -156,20 +156,28 @@ class AppConfig(BaseModel):
...
-        matches the field's own ``default_factory=list``.
+        (a list) or ``memory:`` (an object), with only comments beneath it as
+        shipped throughout ``config.example.yaml`` — makes PyYAML parse the value
+        as ``None``. Without this, the documented ``cp config.example.yaml
+        config.yaml`` first-run flow crashes with an opaque ``Input should be a
+        valid list`` / ``valid dictionary`` pydantic error for that section.
+
+        Dropping the ``None`` lets each field fall back to its default: list
+        sections become ``[]`` via ``default_factory=list`` and object sections
+        get their default config. This generalizes the earlier list-only
```
willem-bd, line 173: [P2] Docstring/PR claim "generalizes the database special-case" is inaccurate — `_apply_database_defaults` still owns it

- Location: `app_config.py:173` (docstring); helper at `from_file:227` / `_apply_database_defaults:283-292` (unchanged by this PR)
- Problem: The docstring (and PR description) say this validator "generalizes … the `database` special-case in `from_file`," implying the special-case is subsumed. It isn't — and a future reader could conclude `_apply_database_defaults` is now redundant and delete it.
- Evidence: `from_file` calls `_apply_database_defaults(config_data)` at `:227`, **before** `model_validate` at `:237`. That helper converts `database: None` → `{}` and applies `CONFIG_FILE_DATABASE_DEFAULTS` via `setdefault`. So by the time this `model_validator(mode="before")` runs, `database` is already a dict; the `if value is not None` filter never sees it, and the validator is a no-op for `database`. The two mechanisms overlap rather than one generalizing the other. If `CONFIG_FILE_DATABASE_DEFAULTS` differs from `DatabaseConfig`'s own field defaults, removing the helper would silently change persistence behavior.
- Suggested fix: Reword to say this validator is a general backstop for all sections while `database` keeps its dedicated `_apply_database_defaults` — or, if the defaults genuinely match, remove the helper and let the validator + `DatabaseConfig` defaults handle it (verify the defaults match first).
- Test: if the helper is removed, add a `database: null` round-trip asserting the resolved `DatabaseConfig` matches today's output.

backend/tests/test_app_config_reload.py
```diff
@@ -170,6 +170,41 @@ def test_app_config_coerces_commented_out_list_sections(tmp_path, monkeypatch):
...
+            }
+        ),
+        encoding="utf-8",
+    )
+    monkeypatch.setenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH", str(extensions_path))
+
+    config = AppConfig.from_file(str(config_path))
+
+    # Each present-but-null object section falls back to its default config.
+    assert config.memory is not None
```
willem-bd: [P2] New test doesn't cover the PR's "required section still errors" claim, and assertions are weaker than they look

- Location: `test_app_config_reload.py:173` (new test) and `:201-205` (assertions)
- Problem: The PR explicitly states a null required section (`sandbox`) still raises, but nothing tests that. The `assert config.<section> is not None` checks (×5) also only prove "something truthy," not that the resolved config equals the expected default.
- Evidence: Every `sandbox` in the suite is a valid `{"use": ...}` (lines 47/70/88/129/156/186/215); there is no `sandbox: null` case, and the only `pytest.raises(ValidationError)` (`:465`) is an unrelated reload/`title: False` test. The new test supplies a valid sandbox precisely to avoid tripping the required-field error, so the still-errors behavior is unverified. A default that silently became wrong-but-truthy would still pass the `is not None` checks.
- Suggested fix: Add `def test_app_config_null_required_section_still_errors` writing `sandbox: null` and asserting `with pytest.raises(ValidationError): AppConfig.from_file(...)`. Optionally strengthen to equality, e.g. `config.memory == MemoryConfig()` (or assert one distinguishing default field).
- Test: the two cases above.

willem-bd (commented):
Adding to @fancyboi999's review — the change is correct and the broadening is safe; two follow-ups worth tightening: (1) the docstring's "generalizes the database special-case" claim is inaccurate (`_apply_database_defaults` still owns `database`), and (2) the new test doesn't cover the PR's "required section still errors" claim and only asserts non-None. Details inline.

backend/packages/harness/deerflow/config/app_config.py
fancyboi999: Confirmed — and this one my original pass missed, so good catch. `_apply_database_defaults` does more than null-coercion: at `:296` it `setdefault`s `CONFIG_FILE_DATABASE_DEFAULTS = {"backend": "sqlite", "sqlite_dir": ".deer-flow/data"}` (`:43-46`) into the section. The new `_drop_null_config_sections` validator only removes the null key so the field falls back to its *model* default — it never fills nested per-key file defaults. So `database` is genuinely not subsumed, and deleting `_apply_database_defaults` on the strength of the docstring would silently drop the sqlite backend/dir defaults. Agree the wording should be narrowed to "covers the null-coercion part of the database handling" (or drop the database mention), so nobody reads it as "now redundant."

backend/tests/test_app_config_reload.py
fancyboi999: Confirmed on both halves. (1) Required-null isn't covered: every `sandbox` in the suite is a valid `{"use": ...}` (lines 47/70/88/129/156/186/215) — there's no `sandbox: null` case, so the PR's "required section still errors when null" claim rides on the docstring, not a test. A `sandbox: None` case asserting `pytest.raises(ValidationError)` would lock it in. (2) The five `is not None` assertions (`:201-205`) only prove "something truthy resolved" — they'd still pass if a section coerced to the wrong-but-truthy object. Asserting equality to the expected default (e.g. `config.memory == MemoryConfig()`) would actually pin "null → the documented default." My earlier review said the test "covers the object-section case" without flagging that the assertions are this loose — agree it's worth tightening.

ly-wang19:
Thanks @willem-bd @fancyboi999 — both follow-ups addressed in the latest push:

1. **Docstring** (`app_config.py`): removed the inaccurate "generalizes the database special-case" claim. It now states that `_apply_database_defaults` still owns `database` and applies concrete defaults beyond null-coercion — the two are complementary, not subsumed.
2. **Tests** (`test_app_config_reload.py`):
   - Strengthened `test_app_config_coerces_commented_out_object_sections` to assert each null section falls back to its **expected default config type** (e.g. `MemoryConfig`), not merely non-None.
   - Added `test_app_config_null_required_section_still_errors`, covering the `sandbox: null` → `ValidationError` case (the "required section without a default still errors" claim).

Both tests pass locally. 🙏

fancyboi999 (commented):
Both follow-ups check out — verified against the latest push.

- **Docstring (`app_config.py:173-178`):** the "generalizes the database special-case" line is gone; it now says the `database` section is independent, still owned by `_apply_database_defaults`, and applies concrete defaults beyond null-coercion. Accurate now — no more "this subsumes it" reading.
- **Tests (`test_app_config_reload.py`):** `test_app_config_coerces_commented_out_object_sections` now asserts each section resolves to its actual config type (`type(...).__name__ == "MemoryConfig"`, etc.) instead of just `is not None`, and `test_app_config_null_required_section_still_errors` covers `sandbox: null` → `ValidationError`. I ran both locally and they pass — the required-null claim is now pinned by a test rather than the docstring.

One unrelated heads-up I hit while running the full file (not from your change — pure branch staleness, non-blocking): this branch predates #3610, which bumped `SubagentsAppConfig.timeout_seconds` 900 → 1800 on main. So your tree still has the 900 default and `test_get_app_config_resets_singleton_configs_when_sections_removed` still asserts `== 900`, while main has both at 1800. Since this PR doesn't touch `subagents_config.py` or that assertion line, a sync with main resolves cleanly to 1800 on its own — just don't let a manual conflict resolution pin it back to 900, or that test goes red on main.

Nice — the introduction/database split and the test tightening both land correctly.

WillemJiang (approved):

PR #3574, opened 2026-06-14
fix(persistence): offload sqlite dir creation off the lifespan event loop

ly-wang19:
## What

`init_engine` (`persistence/engine.py`) runs on the FastAPI lifespan event loop, but created the SQLite data directory with a synchronous `os.makedirs` (a `stat` + `mkdir` syscall), blocking the loop during startup. This dispatches it via `asyncio.to_thread`.

It's the same class of bug — and the same fix — as #1912, which moved the checkpointer's `ensure_sqlite_parent_dir` off the lifespan loop (anchored by `tests/blocking_io/test_sqlite_lifespan.py`). The persistence engine's own `os.makedirs` was the remaining sibling.

Surfaced by `make detect-blocking-io` (DIRECT_ASYNC, `persistence/engine.py:100`).

## Disposition
- **FIX + ANCHOR** — `os.makedirs(...)` → `await asyncio.to_thread(os.makedirs, ...)`. Behavior-preserving: same directory, same `exist_ok=True`. The rest of `init_engine` already uses async SQLAlchemy (`async with engine.begin()` / `run_sync`).

## Teeth (red → green)
New `tests/blocking_io/test_persistence_engine_sqlite.py` drives the real `init_engine(backend="sqlite", ...)` under the strict Blockbuster gate with a `sqlite_dir` that does not exist yet (so `os.makedirs` actually runs); the async engine/session machinery is mocked so directory creation is the only host FS op under test.
- With the offload → **passes**.
- Reverted to the sync `os.makedirs` → **`BlockingError: Blocking call to os.stat`** (the makedirs stat trips the gate).

## Verification
- `pytest tests/blocking_io/` → **17 passed** (16 prior + 1 new)
- `ruff check` → clean

### DIFF
--- a/backend/packages/harness/deerflow/persistence/engine.py
+++ b/backend/packages/harness/deerflow/persistence/engine.py
@@ -10,6 +10,7 @@ None and fall back to in-memory implementations.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 
@@ -97,7 +98,11 @@ async def init_engine(
 
         from sqlalchemy import event
 
-        os.makedirs(sqlite_dir or ".", exist_ok=True)
+        # Offload the directory creation: ``init_engine`` runs on the FastAPI
+        # lifespan event loop, and a sync ``os.makedirs`` (a stat + mkdir
+        # syscall) blocks it during startup. Mirrors the #1912 fix for the
+        # checkpointer's ``ensure_sqlite_parent_dir``.
+        await asyncio.to_thread(os.makedirs, sqlite_dir or ".", exist_ok=True)
         _engine = create_async_engine(url, echo=echo, json_serializer=_json_serializer)
 
         # Enable WAL on every new connection. SQLite PRAGMA settings are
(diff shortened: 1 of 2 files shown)


fancyboi999 (commented):
Clean, minimal offload — verified it's behavior-preserving and correctly scoped:

- **`init_engine`** is `async` (`engine.py:58`), so `await asyncio.to_thread(os.makedirs, sqlite_dir or ".", exist_ok=True)` (`:105`) is valid; it forwards the same path and `exist_ok=True`, and any error (e.g. `PermissionError`) still propagates through `to_thread`, so it's a faithful swap for the old sync `os.makedirs`.
- It's the right *and complete* target: in the sqlite branch, the `os.makedirs` was the only direct synchronous filesystem syscall on the lifespan loop — `create_async_engine` is lazy (no connect at construction) and `Base.metadata.create_all` runs via `await conn.run_sync(...)` through the async driver, so neither blocks the loop. Matches what `make detect-blocking-io` flagged (`engine.py:100`) and the #1912 sibling fix for the checkpointer's `ensure_sqlite_parent_dir`.
- The **anchor test** is well built: it drives the real `init_engine(backend="sqlite", ...)` under the strict Blockbuster gate with a `sqlite_dir` that doesn't exist yet (so `os.makedirs` actually runs), mocks out only the engine/session machinery so the directory creation is the sole host FS op under test, and pre-imports `deerflow.persistence.models` so `init_engine`'s lazy import isn't itself a false-positive under the gate. A regression back onto the loop would raise `BlockingError`.

Looks good to merge. (Reviewed at code level; didn't run the suite locally.)

WillemJiang (approved):

PR #3575, opened 2026-06-14
feat(community): add Serper Google Images provider for image_search

18062706139fcz:
## Why

DeerFlow's `image_search` tool currently only ships DuckDuckGo (no key) and
InfoQuest providers. Users who already configure the Serper `web_search`
provider have a `SERPER_API_KEY` on hand but cannot reuse it to pull Google
Images results. Reference-image lookup quality matters because the harness uses
`image_search` results as reference inputs for downstream image generation, and
Google Images via Serper gives noticeably better coverage for people, products
and scene references than the default provider.

This PR adds a Serper-backed `image_search` provider so those users get a
higher-quality, opt-in option without any new dependency or key.

## What changed

- **New opt-in provider**: `deerflow.community.serper.tools:image_search_tool`.
  Configure it under `tools:` with `api_key: $SERPER_API_KEY`. It returns
  normalized `{title, image_url, thumbnail_url}` entries plus a `usage_hint`,
  and reuses the same `SERPER_API_KEY` as the Serper `web_search` tool.
- **Shared, hardened request layer**: `web_search` and `image_search` now share
  `_serper_post` / `_response_items`, with `max_results` bounded (capped at 10),
  query normalization (trim + 500-char cap), and structured JSON errors for
  missing key / HTTP error / network error / malformed payloads. A missing or
  null-typed `organic`/`images` field is treated as "no results" rather than a
  malformed payload.
- **Best-effort SSRF guard**: image URLs go through `_safe_public_url`, which
  rejects non-`http(s)` schemes, `localhost` (including trailing-dot FQDN forms
  like `localhost.`/`127.0.0.1.`), obfuscated decimal/hex/octal IPv4 literals,
  and private/non-global IP literals. `image_url` and `thumbnail_url` only
  cross-fall back when the other field is *absent*; a field that was present but
  filtered stays empty rather than collapsing onto its counterpart, so a dropped
  high-res URL never silently masquerades as the preview (and vice versa).
  Filtered entries are dropped (and do not consume the result limit); if
  everything is filtered the tool returns an explicit "No safe image URLs found"
  error instead of reporting fake successes. The guard documents that it cannot
  defend against DNS rebinding, so any consumer that downloads these URLs must
  re-validate the resolved IP at fetch time.
- **`doctor` security nudge**: a literal `api_key` value in `config.yaml` is now
  reported as a **warning** (was treated as OK), with a fix that points users to
  move the key into `.env` and reference it as `$VAR`. This applies to every
  key-based provider in the doctor checks, not just Serper. When a config
  `$VAR` reference is unset, the check falls through to the provider's default
  env var (the tool's runtime fallback) before reporting it as not set.
- **Docs & config**: documented the Serper `image_search` provider and
  `SERPER_API_KEY` in `config.example.yaml`, `backend/docs/CONFIGURATION.md`,
  and the EN/ZH harness tool docs; example config now discourages committing
  literal keys.

The default `image_search` provider stays DuckDuckGo and Serper is purely
opt-in. See "Behavior change" below for the one existing-behavior change.

## Behavior change

The `doctor` literal-key check is shared across all key-based providers, so the
stricter "literal `api_key` ⇒ warning" rule also applies to the existing
**Brave** provider that landed on `main` in the meantime. Previously a literal
inline `api_key` was reported as `ok`; it is now reported as `warn` with a fix
to move the key into `.env`. Health status only — no runtime/tool behavior
changes, and `$VAR` references and env-var-based config remain `ok`. The Brave
doctor test was updated accordingly.

## Surface area

- [x] **Agents / LangGraph** — adds a new tool provider usable by the harness
- [x] **Default behavior change** — `doctor` now warns (was `ok`) on literal
  inline `api_key` for key-based providers, including Brave; health-status only
- [x] **Docs / tests / CI only** — config/docs updates accompany the feature
- [ ] **Frontend UI**
- [ ] **Backend API**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies** — none added (reuses `httpx`)

## Bug fix verification

Not a bug fix. This is a new opt-in feature; see Validation for the tests added.

## Validation

```
cd backend && uv run pytest tests/test_serper_tools.py tests/test_doctor.py -q
# 147 passed
```

Tests cover: api-key fallback chain, `max_results` bounds, the SSRF filter
(including trailing-dot FQDN forms, obfuscated IPv4 literals, "all filtered",
and "filtered entries don't consume the limit"), the no-collapse semantics
between `image_url`/`thumbnail_url`, missing/null `organic`/`images` fields,
HTTP / network / timeout / malformed-payload errors, warn-once logging, and the
`doctor` literal-key warning path plus the `$VAR`-unset → default-env-var
fallback for both Serper and Brave.

## Review follow-ups

Addressed the review feedback in a follow-up commit:

- Blocked the trailing-dot hostname SSRF bypass (`localhost.`/`127.0.0.1.`).
- Kept a filtered `image_url`/`thumbnail_url` empty instead of collapsing it
  onto its counterpart.
- Evaluated the SSRF guard once per field instead of twice.
- Treated a null-typed `organic`/`images` field as "no results".
- Restored the default-env-var fallback in `doctor` when a `$VAR` is unset.

## AI assistance

**How you used it:** AI performed a structured code review across security,
architecture, feature-fit and test-coverage dimensions, implemented the review
findings (image URL filtering semantics, trailing-dot SSRF bypass, doctor
literal-key warning + `$VAR` fallback, null-field handling), added the
accompanying tests, resolved the rebase conflicts against the Brave provider,
and drafted this PR description. I reviewed the diff and validation output.

### DIFF
--- a/backend/packages/harness/deerflow/community/serper/tools.py
+++ b/backend/packages/harness/deerflow/community/serper/tools.py
@@ -8,7 +8,7 @@ API. An API key is required. Sign up at https://serper.dev to get one.
 import json
 import logging
 import os
-from ipaddress import ip_address
+from ipaddress import IPv4Address, ip_address
 from urllib.parse import urlparse
 
 import httpx
@@ -80,7 +80,7 @@ def _clean_query(query: str) -> str:
     return query
 
 
-def _decode_ipv4(host: str) -> object | None:
+def _decode_ipv4(host: str) -> IPv4Address | None:
     """Decode obfuscated IPv4 literals that ``ip_address`` rejects.
 
     Mirrors the permissive ``inet_aton`` parsing many HTTP clients use, so that
--- a/backend/packages/harness/deerflow/community/serper/tools.py
+++ b/backend/packages/harness/deerflow/community/serper/tools.py
@@ -65,7 +65,11 @@ def _unexpected_format_error(query: str) -> str:
 
 
 def _response_items(data: dict, field: str, query: str) -> tuple[list[dict] | None, str | None]:
-    items = data.get(field, [])
+    items = data.get(field)
+    # Treat a missing or null field as "no results" (some APIs return
+    # ``{"organic": null}`` to signal that) rather than a malformed payload.
+    if items is None:
+        return [], None
     if not isinstance(items, list):
         logger.error("Serper returned unexpected '%s' payload type: %s", field, type(items).__name__)
         return None, _unexpected_format_error(query)
@@ -122,6 +126,16 @@ def _decode_ipv4(host: str) -> IPv4Address | None:
     return ip_address(result)
 
 
+def _is_url_present(value: object) -> bool:
+    """Return ``True`` when *value* is a non-empty URL string.
+
+    Used to distinguish a field that was *absent* (eligible for cross-field
+    fallback) from one that was *present but filtered* by the SSRF guard (which
+    must stay empty rather than collapse onto its counterpart).
+    """
+    return isinstance(value, str) and bool(value.strip())
+
+
 def _safe_public_url(value: object) -> str:
     """Return ``value`` only if it is a safe, public http(s) URL, else "".
 
@@ -139,7 +153,12 @@ def _safe_public_url(value: object) -> str:
     if parsed.scheme not in {"http", "https"} or not parsed.netloc or not parsed.hostname:
         return ""
 
-    host = parsed.hostname.lower()
+    # Strip a single trailing dot (FQDN root label). ``localhost.`` and
+    # ``127.0.0.1.`` resolve to loopback on common resolvers but would
+    # otherwise slip past the localhost/IP checks below.
+    host = parsed.hostname.lower().rstrip(".")
+    if not host:
+        return ""
     if host == "localhost" or host.endswith(".localhost"):
         return ""
 
@@ -268,8 +287,18 @@ def image_search_tool(query: str, max_results: int = 5) -> str:
 
     normalized_results = []
     for r in images:
-        image_url = _safe_public_url(r.get("imageUrl")) or _safe_public_url(r.get("thumbnailUrl"))
-        thumbnail_url = _safe_public_url(r.get("thumbnailUrl")) or _safe_public_url(r.get("imageUrl"))
+        raw_image = r.get("imageUrl")
+        raw_thumb = r.get("thumbnailUrl")
+        # Evaluate the (non-trivial) SSRF guard once per field instead of twice.
+        safe_image = _safe_public_url(raw_image)
+        safe_thumb = _safe_public_url(raw_thumb)
+        # Cross-fall back only when the other field was *absent*. A field that
+        # was present but failed the SSRF filter is left empty rather than
+        # collapsed onto its counterpart, so a dropped high-res URL never
+        # silently masquerades as the preview (and vice versa), preserving the
+        # high-res/preview contract callers rely on.
+        image_url = safe_image or (safe_thumb if not _is_url_present(raw_image) else "")
+        thumbnail_url = safe_thumb or (safe_image if not _is_url_present(raw_thumb) else "")
         if not image_url and not thumbnail_url:
             continue
         normalized_results.append(
--- a/backend/packages/harness/deerflow/community/serper/tools.py
+++ b/backend/packages/harness/deerflow/community/serper/tools.py
@@ -80,14 +80,57 @@ def _clean_query(query: str) -> str:
     return query
 
 
+def _decode_ipv4(host: str) -> object | None:
+    """Decode obfuscated IPv4 literals that ``ip_address`` rejects.
+
+    Mirrors the permissive ``inet_aton`` parsing many HTTP clients use, so that
+    integer (``2130706433``), hex (``0x7f000001``) and octal (``0177.0.0.1``)
+    encodings of an address are recognized. Returns an ``IPv4Address`` when the
+    host decodes to one, otherwise ``None`` (e.g. real domains like
+    ``cafe.com`` fail to decode and are left for the caller to treat as a host).
+    """
+    parts = host.split(".")
+    if not 1 <= len(parts) <= 4:
+        return None
+
+    values: list[int] = []
+    for part in parts:
+        if not part:
+            return None
+        try:
+            if part.startswith(("0x", "0X")):
+                values.append(int(part, 16))
+            elif part.startswith("0") and len(part) > 1:
+                values.append(int(part, 8))
+            else:
+                values.append(int(part, 10))
+        except ValueError:
+            return None
+
+    *leading, last = values
+    for value in leading:
+        if not 0 <= value <= 0xFF:
+            return None
+    max_last = (1 << (8 * (4 - len(leading)))) - 1
+    if not 0 <= last <= max_last:
+        return None
+
+    result = 0
+    for value in leading:
+        result = (result << 8) | value
+    result = (result << (8 * (4 - len(leading)))) | last
+    return ip_address(result)
+
+
 def _safe_public_url(value: object) -> str:
     """Return ``value`` only if it is a safe, public http(s) URL, else "".
 
     This is a best-effort SSRF guard that rejects non-http(s) schemes,
-    ``localhost``, and private/non-global IP literals. It only inspects the URL
-    string and cannot catch public hostnames that resolve to internal IPs
-    (e.g. DNS rebinding); any consumer that actually downloads these URLs must
-    re-validate the resolved IP at fetch time.
+    ``localhost``, and private/non-global IP literals (including obfuscated
+    decimal/hex/octal encodings). It only inspects the URL string and cannot
+    catch public hostnames that resolve to internal IPs (e.g. DNS rebinding);
+    any consumer that actually downloads these URLs must re-validate the
+    resolved IP at fetch time.
     """
     if not isinstance(value, str):
         return ""
@@ -103,7 +146,9 @@ def _safe_public_url(value: object) -> str:
     try:
         ip = ip_address(host)
     except ValueError:
-        return url
+        ip = _decode_ipv4(host)
+        if ip is None:
+            return url
     return url if ip.is_global else ""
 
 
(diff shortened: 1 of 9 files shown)


backend/packages/harness/deerflow/community/serper/tools.py
```diff
@@ -16,43 +18,104 @@
...
-    if not api_key:
-        if not _api_key_warned:
-            _api_key_warned = True
-            logger.warning("Serper API key is not set. Set SERPER_API_KEY in your environment or provide api_key in config.yaml. Sign up at https://serper.dev")
-        return json.dumps(
-            {"error": "SERPER_API_KEY is not configured", "query": query},
-            ensure_ascii=False,
-        )
+    try:
+        ip = ip_address(host)
```
fancyboi999, line 166: Minor / non-blocking, and you've already framed this as best-effort: the IP-literal check only rejects *parseable* private IPs. Encoded hosts — decimal (`http://2130706433/` == `127.0.0.1`), hex, or octal — fail `ip_address(host)`, fall through the `except ValueError`, and are returned as allowed. Since some HTTP clients will happily resolve those, it's one more reason the real defense has to be the fetch-time re-validation your docstring already calls for. Worth confirming the downstream image-gen downloader actually re-validates the resolved IP before treating these URLs as trusted.

fancyboi999 (commented):
Nice addition — the refactor and the new provider are genuinely well done. Went through `serper/tools.py` end to end:

- **Shared helpers** (`_serper_post`, `_response_items`, `_coerce_max_results`, `_clean_query`, `_missing_key_error`) cleanly de-duplicate web/image search, and the defensive parsing is solid: `_response_items` type-checks the payload list and filters non-dicts, `_serper_post` rejects non-dict JSON, and error/response text is truncated to 500 chars. `_coerce_max_results` bounding at 10 and `_clean_query` truncating at 500 are good input hygiene, and the per-tool `_api_key_warned` set fixes the old global-bool "warn once across all tools" behavior.
- **The SSRF handling is thoughtful**: applying `_safe_public_url` to the image `imageUrl`/`thumbnailUrl` (surfaced as reference images and potentially downloaded) but deliberately *not* to `web_search` links (citations the model reads, not fetched) — with a comment explaining the distinction — is the right call, and the honest docstring about DNS-rebinding being out of scope is appreciated.
- Spot-checks pass: `__init__.py` exports `image_search_tool`, `_serper_post` sends `{"q": query, "num": max_results}`, and the `imageUrl`↔`thumbnailUrl` fallback + skip-if-neither-safe logic all hold up.

Two things:

1. **This PR currently conflicts with `main`** — `mergeable: false`, and I confirmed 6 real conflict hunks locally. The conflicting files are the shared registry/docs/config set that other recently-merged PRs also touched: `scripts/doctor.py`, `tests/test_doctor.py`, `config.example.yaml`, `backend/docs/CONFIGURATION.md`, and `frontend/.../{en,zh}/harness/tools.mdx`. It'll need a rebase + re-resolution before it can merge — the `doctor.py` / `test_doctor.py` hunks are the ones to re-check carefully afterward, since that's where provider registration lives.
2. One non-blocking note on the SSRF guard, inline.

Solid, opt-in, no new dependency or key — looks good once it's rebased. (Reviewed the provider logic at code level; didn't run the suite.)

WillemJiang:
@18062706139fcz, please resolve the conflicts of the main branch.

backend/packages/harness/deerflow/community/serper/tools.py
```diff
@@ -16,43 +18,149 @@
...
+    """
+    if not isinstance(value, str):
+        return ""
+    url = value.strip()
+    parsed = urlparse(url)
+    if parsed.scheme not in {"http", "https"} or not parsed.netloc or not parsed.hostname:
+        return ""
+
+    host = parsed.hostname.lower()
+    if host == "localhost" or host.endswith(".localhost"):
```
willem-bd, line 162: **SSRF guard misses trailing-dot hostnames.**

`http://localhost./x.jpg`, `http://127.0.0.1./x.jpg`, and `http://10.0.0.1./x.jpg` all bypass both branches here:

- `host == "localhost"` is `False` (host is `localhost.`)
- `host.endswith(".localhost")` is `False` (it ends with `.`)
- `ip_address('127.0.0.1.')` raises
- `_decode_ipv4('127.0.0.1.')` splits to 5 parts (last empty) and returns `None`

So `_safe_public_url` returns the URL as safe. `socket.gethostbyname('localhost.')` resolves to `127.0.0.1` on every common resolver, and httpx will happily fetch it — defeating the guard for the canonical loopback case the docstring claims it covers.

Fix: strip a single trailing `.` from `host` before running the localhost / `ip_address` / `_decode_ipv4` checks (or reject any host ending in `.`). Worth a regression test alongside the existing `test_localhost_is_filtered`.

backend/packages/harness/deerflow/community/serper/tools.py
```diff
@@ -93,3 +234,61 @@ def web_search_tool(query: str, max_results: int = 5) -> str:
...
+    images, error_json = _response_items(data, "images", query)
+    if error_json is not None:
+        return error_json
+    if not images:
+        return json.dumps({"error": "No images found", "query": query}, ensure_ascii=False)
+
+    normalized_results = []
+    for r in images:
+        image_url = _safe_public_url(r.get("imageUrl")) or _safe_public_url(r.get("thumbnailUrl"))
+        thumbnail_url = _safe_public_url(r.get("thumbnailUrl")) or _safe_public_url(r.get("imageUrl"))
```
willem-bd: **`image_url` and `thumbnail_url` silently collapse to the same URL when one fails the SSRF filter.**

For a Serper entry like `{"imageUrl":"http://10.0.0.1/full.jpg","thumbnailUrl":"https://example.com/t.jpg"}`, `_safe_public_url` drops the private `imageUrl`, both fields then resolve via the `or` fallback to the same thumbnail URL, and the model receives `image_url == thumbnail_url == 'https://example.com/t.jpg'` — but the docstring contract says one is the high-res reference and the other is its preview.

Downstream image-generation steps that pick the "better" of the two are now picking between identical URLs, and any retry strategy that falls back to the thumbnail will no-op.

Suggest: drop the field that didn't pass (leave it `""`), or surface a flag indicating the high-res URL was filtered.

scripts/doctor.py
```diff
@@ -478,43 +469,62 @@ def check_web_tool(config_path: Path, *, tool_name: str, label: str) -> CheckRes
...
+            },
         }
 
-        for tool_entry in tool_entries:
-            use = tool_entry.get("use", "")
+        def _configured_key_detail(tool: dict, default_var: str) -> tuple[Status, str] | None:
+            api_key = tool.get("api_key")
+            if isinstance(api_key, str) and api_key.strip():
+                key = api_key.strip()
+                if key.startswith("$"):
```
willem-bd, line 489: **Behavior regression: `api_key: $UNSET_VAR` now masks a working default env var.**

If the config has `api_key: $MY_BRAVE` but `MY_BRAVE` is unset, while `BRAVE_SEARCH_API_KEY` *is* set in `.env` (which works fine at runtime — tools fall back to the default var), `_configured_key_detail` returns `None` from this branch and never reaches the `os.environ.get(default_var)` fallback below. The caller then reports `BRAVE_SEARCH_API_KEY not set`, contradicting actual runtime behavior.

The pre-PR code (around old line 460–462) checked the default env var as a final fallback regardless of the `api_key` value. Suggest preserving that fallback: when the `$VAR` form fails to resolve, fall through to the default var check before declaring "not set".

backend/packages/harness/deerflow/community/serper/tools.py
```diff
@@ -93,3 +234,61 @@ def web_search_tool(query: str, max_results: int = 5) -> str:
...
+
+    images, error_json = _response_items(data, "images", query)
+    if error_json is not None:
+        return error_json
+    if not images:
+        return json.dumps({"error": "No images found", "query": query}, ensure_ascii=False)
+
+    normalized_results = []
+    for r in images:
+        image_url = _safe_public_url(r.get("imageUrl")) or _safe_public_url(r.get("thumbnailUrl"))
```
willem-bd: **Minor: `_safe_public_url` is invoked up to four times per result.**

For each entry the loop runs `_safe_public_url(r.get("imageUrl"))` and `_safe_public_url(r.get("thumbnailUrl"))` twice each (once on each line). With `max_results` up to 10, that's 40 urlparse + ip_address + `_decode_ipv4` evaluations per tool call when 20 suffice.

Compute once:
```python
safe_image = _safe_public_url(r.get("imageUrl"))
safe_thumb = _safe_public_url(r.get("thumbnailUrl"))
image_url = safe_image or safe_thumb
thumbnail_url = safe_thumb or safe_image
```
Not a hot path today, but the helper is non-trivial.

backend/packages/harness/deerflow/community/serper/tools.py
```diff
@@ -16,43 +18,149 @@
...
+def _unexpected_format_error(query: str) -> str:
+    return json.dumps(
+        {"error": "Serper returned an unexpected response format", "query": query},
+        ensure_ascii=False,
+    )
+
+
+def _response_items(data: dict, field: str, query: str) -> tuple[list[dict] | None, str | None]:
+    items = data.get(field, [])
+    if not isinstance(items, list):
```
willem-bd, line 73: **Stricter than the old code on null-typed result fields.**

If Serper returns `{"organic": null}` (or `{"images": null}`) — a legitimate JSON shape some APIs use to signal "no results" — `data.get(field, [])` returns `None`, `isinstance(None, list)` is `False`, and the tool surfaces `"Serper returned an unexpected response format"`.

The pre-PR `web_search` code at this site did `organic = data.get("organic", []); if not organic: "No results found"` — null was treated as empty. Either accept `None` as empty (`items = data.get(field) or []`) or document this as an intentional behavior change. Currently no test covers the null-field case, so it'd silently flip a user-visible error string.

willem-bd (commented):
Automated review (high-effort `/code-review` pass). Five findings inline; the trailing-dot SSRF bypass is the one I'd block on.

WillemJiang (approved):

PR #3577, opened 2026-06-14
fix(sandbox): create shell session before retrying on a fresh id

stphtt:
## Why

I hit this issue. Agent runs that used the AIO sandbox would intermittently get stuck in a crash loop - the same shell command retried over and over until the run died at the LangGraph recursion limit, with no useful output to the user. It was intermittent enough to be painful to pin down, so I dug into the sandbox execution path to find out what was actually failing.

The AIO sandbox serializes shell commands on a single persistent session. When that session is lost or corrupted, the container returns an `ErrorObservation` (`'ErrorObservation' object has no attribute 'exit_code'`) instead of real output, and `execute_command` is supposed to recover by retrying on a fresh session.

That recovery was the bug. It generated a UUID and passed it straight to `exec_command(id=fresh_id)` - but the sandbox image only auto-creates a session when `exec_command` is called with *no* id. An exec carrying an unknown id returns HTTP 404 "Session not found", so every recovery attempt 404'd too. A transient session lapse turned into an unrecoverable tool error, the agent kept retrying the same call, and the run spun until it hit the recursion limit - exactly the crash loop I was seeing.

## What changed

`execute_command`'s recovery path now calls `create_session(id=fresh_id)` before targeting that id on retry. `create_session` is idempotent (it returns the existing session if the id is already present), and the whole sequence stays under the existing serializing lock, so no new concurrency surface is introduced. No change to normal (non-error) command execution.

## Surface area

## Bug fix verification

- Test path: `backend/tests/test_aio_sandbox.py::TestErrorObservationRetry::test_retry_creates_fresh_session_before_targeting_it`
- Went red on `main` and green on this branch: **yes** — verified by temporarily reverting `aio_sandbox.py` to the pre-fix version while keeping the new test (test fails: `create_session` is never called, `assert len(created_ids) == 1` → `0 == 1`), then restoring the fix (passes). The test asserts the retry targets exactly the created session id rather than a fabricated, uncreated one.

## Validation

- `cd backend && PYTHONPATH=. uv run pytest tests/test_aio_sandbox.py -v` → **25 passed**
- `cd backend && uv run ruff check packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py tests/test_aio_sandbox.py` → All checks passed
- `cd backend && uv run ruff format --check packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py tests/test_aio_sandbox.py` → already formatted

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Code review of the change, and drafting this PR description. The fix and the diagnosis are my own.


backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
```diff
@@ -132,8 +132,12 @@ def execute_command(self, command: str) -> str:
                 output = result.data.output if result.data else ""
 
                 if output and _ERROR_OBSERVATION_SIGNATURE in output:
-                    logger.warning("ErrorObservation detected in sandbox output, retrying with a fresh session")
+                    logger.warning("ErrorObservation detected in sandbox output, retrying on a fresh session")
+                    # exec_command only auto-creates a session when called with
+                    # no id, so the recovery session must be created explicitly
+                    # before we target it on retry.
                     fresh_id = str(uuid.uuid4())
+                    self._client.shell.create_session(id=fresh_id)
```
fancyboi999, line 140: Small thing, non-blocking: each recovery now creates a real session that's used exactly once — the next `execute_command` runs with no id and goes back to the default session, so this `fresh_id` session is never reused or released. Pre-fix nothing was actually created (the exec 404'd first), so this adds a new, if minor, accumulation surface on the error path. The SDK exposes `shell.cleanup_session(session_id)` if you'd want to release it after the retry. Negligible for transient corruption; only worth it if a sandbox can keep hitting this.

fancyboi999 (commented):
Traced the diagnosis against the `agent_sandbox` SDK and it holds:

- `shell.exec_command(id=...)` only auto-creates a session when `id` is omitted — its own docstring says *"if not provided, one will be automatically created"* (`agent_sandbox/shell/client.py`). So the old recovery path handing a fabricated UUID straight to `exec_command` would 404 on the unknown id exactly as you describe, and the retry could never recover.
- `shell.create_session(id=...)` is idempotent — *"If id already exists, return the existing session"* — so creating then targeting the same id is safe.
- The new call stays inside the existing `with self._lock:`, so no new concurrency surface.

Ran the test both ways: green on the branch, and red when I revert just the `create_session` line (`assert len(created_ids) == 1` → `0 == 1`). It has teeth.

One non-blocking note inline.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
stphtt: Good call - took it. The recovery session is one-shot (the next `execute_command` runs with no id and goes back to the default session), so under persistent corruption it would leak a session per command.

WillemJiang (approved):

WillemJiang:
@stphtt  thanks for your contribution, keep it up :)

PR #3578, opened 2026-06-14
fix(channels): require bound identity for user-owned IM messages

### LINKED ISSUE #3538: [IM] Enforce connection ownership for user-owned channel messages
## Context

DeerFlow now supports user-owned IM channel bindings: a browser-authenticated DeerFlow user creates a one-time connect code, sends it to a provider bot, and future messages from that external identity should run as the DeerFlow owner. The intended owner path exists: provider adapters attach `connection_id` / `owner_user_id`, and `ChannelManager` places the owner into the run context while preserving the platform user as `channel_user_id`.

However, unresolved IM messages still flow through the manager. When no connection is found, `attach_connection_identity()` returns the inbound message unchanged, and `ChannelManager` falls back to the raw platform `user_id` in auth-enabled mode. Existing tests explicitly preserve this behavior for legacy unbound channel messages.

Relevant code paths:

- `backend/app/channels/connection_identity.py` returns unchanged messages when no persisted binding is found.
- `backend/app/channels/manager.py` falls back to `msg.user_id` when `owner_user_id` is absent.
- `backend/tests/test_channels.py::TestResolveRunParamsUserId::test_unbound_channel_messages_keep_platform_user_id_when_auth_is_enabled` documents the current behavior.
- `backend/docs/IM_CHANNEL_CONNECTIONS.md` describes the desired owner behavior for connected messages.

## Problem

In an auth-enabled deployment with user-owned channel connections enabled, an unbound external platform user can still message a configured bot and trigger DeerFlow runs under a synthetic platform-ID user bucket. That has several undesirable effects:

- It bypasses the browser account binding flow that the UI presents as the way to connect IM accounts.
- It creates shadow user identities and storage buckets that are not tied to a DeerFlow account.
- It makes audit trails confusing: connected messages use DeerFlow owner IDs, while unconnected messages use provider IDs.
- It relies on optional provider-level allowlists as the only practical guardrail.

This may be intentional legacy behavior for operator-owned bots, but it is risky as the default for user-owned channel mode.

## What to build

Define and enforce the policy for unbound IM messages when `channel_connections.enabled` is true. The likely shape is:

- In auth-disabled/local mode, keep the current default-user behavior.
- In auth-enabled user-owned mode, require a resolved channel connection before starting a run.
- If legacy open-bot behavior is still needed, put it behind an explicit compatibility setting and document the security tradeoff.

The final implementation should be consistent across Telegram, Slack, Discord, Feishu/Lark, DingTalk, WeChat, and WeCom.

## Implementation notes

- Add one central decision point rather than duplicating policy in every provider adapter.
- Consider a config such as `channel_connections.require_bound_identity`, defaulting to true when auth is enabled and user-owned connections are enabled.
- Provider adapters should still consume valid `/connect <code>` or Telegram `/start <code>` messages before rejecting ordinary unbound chat.
- Rejection should be friendly and actionable, e.g. “Connect this channel from DeerFlow Settings first.”
- Preserve `channel_user_id` / platform identity metadata for bound messages.
- Avoid breaking auth-disabled local deployments and any explicitly configured legacy operator bot mode.

## Acceptance criteria

- [ ] In auth-enabled user-owned mode, ordinary messages without a resolved `connection_id` do not create DeerFlow runs unless an explicit legacy compatibility setting is enabled.
- [ ] Valid connect-code messages still complete binding even before a user is connected.
- [ ] Bound messages run as the DeerFlow owner and preserve the raw provider user ID as channel metadata.
- [ ] Auth-disabled local mode keeps routing channel messages to the auth-disabled default user.
- [ ] The compatibility/default behavior is documented in `backend/docs/IM_CHANNEL_CONNECTIONS.md` and config examples.
- [ ] Tests cover bound messages, unbound auth-enabled messages, auth-disabled local behavior, and valid connect-code handling.

## Validation plan

- Add unit tests around `ChannelManager._resolve_run_params` or the new policy boundary.
- Add provider-level tests for at least one deep-link provider and one binding-code provider.
- Run the existing channel connection suites plus the new regression tests.

## Blocked by

None - can start immediately, but maintainers should confirm the desired legacy compatibility policy before implementation.

ggnnggez:
Closes #3538

# Summary

- Add a `require_bound_identity` channel-connections policy and pass it into `ChannelManager` when browser-owned IM connections are enabled.
- Reject unbound auth-enabled chat messages before thread/run creation, so platform user IDs cannot silently create DeerFlow-owned runs.
- Keep command/connect handling available while enforcing the bound-user requirement for ordinary chat messages.
- Document the new user-owned IM admission behavior.

# Tests

- `uv run pytest tests/test_channel_connections_config.py tests/test_channels.py -q`
- `uv run ruff check app/channels/manager.py app/channels/service.py packages/harness/deerflow/config/channel_connections_config.py tests/test_channel_connections_config.py tests/test_channels.py`

# Notes

This PR establishes the central admission invariant used by the later ownership and connect-flow fixes: a user-owned IM provider may only create agent runs after the external platform identity has been bound to a DeerFlow owner.

### DIFF
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -55,6 +55,7 @@ STREAM_UPDATE_MIN_INTERVAL_SECONDS = 0.35
 STREAM_MODES = ["messages-tuple", "values"]
 MESSAGE_STREAM_EVENTS = ("messages-tuple", "messages")
 THREAD_BUSY_MESSAGE = "This conversation is already processing another request. Please wait for it to finish and try again."
+BOUND_IDENTITY_REQUIRED_MESSAGE = "Connect this channel from DeerFlow Settings first, then send your message again."
 
 CHANNEL_CAPABILITIES = {
     "dingtalk": {"supports_streaming": False},
@@ -735,6 +736,7 @@ class ChannelManager:
         default_session: dict[str, Any] | None = None,
         channel_sessions: dict[str, Any] | None = None,
         connection_repo: Any | None = None,
+        require_bound_identity: bool = False,
     ) -> None:
         self.bus = bus
         self.store = store
@@ -745,6 +747,7 @@ class ChannelManager:
         self._default_session = _as_dict(default_session)
         self._channel_sessions = dict(channel_sessions or {})
         self._connection_repo = connection_repo
+        self._require_bound_identity = require_bound_identity
         self._client = None  # lazy init — langgraph_sdk async client
         self._channel_metadata_synced: set[str] = set()
         self._skill_storage: SkillStorage | None = None
@@ -950,6 +953,29 @@ class ChannelManager:
 
     # -- chat handling -----------------------------------------------------
 
+    def _requires_bound_identity(self, msg: InboundMessage) -> bool:
+        if not self._require_bound_identity:
+            return False
+        if _auth_disabled_owner_user_id():
+            return False
+        return not (bool(msg.connection_id) and bool(msg.owner_user_id))
+
+    async def _reject_unbound_channel_message(self, msg: InboundMessage) -> None:
+        logger.info(
+            "[Manager] rejecting unbound channel message: channel=%s, chat_id=%s",
+            msg.channel_name,
+            msg.chat_id,
+        )
+        outbound = OutboundMessage(
+            channel_name=msg.channel_name,
+            chat_id=msg.chat_id,
+            thread_id=await self._lookup_thread_id(msg) or "",
+            text=BOUND_IDENTITY_REQUIRED_MESSAGE,
+            thread_ts=msg.thread_ts,
+            metadata=_slim_metadata(msg.metadata),
+        )
+        await self.bus.publish_outbound(outbound)
+
     async def _lookup_thread_id(self, msg: InboundMessage) -> str | None:
         if msg.connection_id and self._connection_repo is not None:
             return await self._connection_repo.get_thread_id(
@@ -1012,6 +1038,10 @@ class ChannelManager:
         self._channel_metadata_synced.add(thread_id)
 
     async def _handle_chat(self, msg: InboundMessage, extra_context: dict[str, Any] | None = None) -> None:
+        if self._requires_bound_identity(msg):
+            await self._reject_unbound_channel_message(msg)
+            return
+
         client = self._get_client()
 
         # Look up existing DeerFlow thread.
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -56,6 +56,7 @@ STREAM_MODES = ["messages-tuple", "values"]
 MESSAGE_STREAM_EVENTS = ("messages-tuple", "messages")
 THREAD_BUSY_MESSAGE = "This conversation is already processing another request. Please wait for it to finish and try again."
 BOUND_IDENTITY_REQUIRED_MESSAGE = "Connect this channel from DeerFlow Settings, complete the in-channel connect step, then send your message again."
+WORKSPACE_OPTIONAL_CONNECTION_PROVIDERS = {"dingtalk", "discord", "wecom"}
 
 CHANNEL_CAPABILITIES = {
     "dingtalk": {"supports_streaming": False},
@@ -921,51 +922,80 @@ class ChannelManager:
 
     async def _handle_message(self, msg: InboundMessage) -> None:
         msg = _apply_effective_owner(msg)
-        if msg.msg_type != InboundMessageType.COMMAND and self._requires_bound_identity(msg):
-            await self._reject_unbound_channel_message(msg)
-            return
+        try:
+            if msg.msg_type != InboundMessageType.COMMAND and await self._requires_bound_identity(msg):
+                await self._reject_unbound_channel_message(msg)
+                return
 
-        async with self._semaphore:
-            try:
+            async with self._semaphore:
                 if msg.msg_type == InboundMessageType.COMMAND:
                     await self._handle_command(msg)
                 else:
                     await self._handle_chat(msg)
-            except InvalidChannelSessionConfigError as exc:
-                logger.warning(
-                    "Invalid channel session config for %s (chat=%s): %s",
-                    msg.channel_name,
-                    msg.chat_id,
-                    exc,
-                )
-                await self._send_error(msg, str(exc))
-            except SlashSkillCommandResolutionError as exc:
-                logger.warning(
-                    "Slash skill command resolution failed for %s (chat=%s): %s",
-                    msg.channel_name,
-                    msg.chat_id,
-                    exc,
-                )
-                await self._send_error(msg, str(exc))
-            except Exception:
-                logger.exception(
-                    "Error handling message from %s (chat=%s)",
-                    msg.channel_name,
-                    msg.chat_id,
-                )
-                await self._send_error(msg, "An internal error occurred. Please try again.")
+        except InvalidChannelSessionConfigError as exc:
+            logger.warning(
+                "Invalid channel session config for %s (chat=%s): %s",
+                msg.channel_name,
+                msg.chat_id,
+                exc,
+            )
+            await self._send_error(msg, str(exc))
+        except SlashSkillCommandResolutionError as exc:
+            logger.warning(
+                "Slash skill command resolution failed for %s (chat=%s): %s",
+                msg.channel_name,
+                msg.chat_id,
+                exc,
+            )
+            await self._send_error(msg, str(exc))
+        except Exception:
+            logger.exception(
+                "Error handling message from %s (chat=%s)",
+                msg.channel_name,
+                msg.chat_id,
+            )
+            await self._send_error(msg, "An internal error occurred. Please try again.")
 
     # -- chat handling -----------------------------------------------------
 
-    def _requires_bound_identity(self, msg: InboundMessage) -> bool:
+    def _bound_identity_workspace_candidates(self, msg: InboundMessage) -> list[str | None]:
+        workspace_candidates: list[str | None] = [msg.workspace_id or None]
+        if msg.channel_name in WORKSPACE_OPTIONAL_CONNECTION_PROVIDERS and workspace_candidates[0] is not None:
+            workspace_candidates.append(None)
+        return workspace_candidates
+
+    async def _has_verified_bound_identity(self, msg: InboundMessage) -> bool:
+        has_connection = bool(msg.connection_id)
+        has_owner = bool(msg.owner_user_id)
+        if not (has_connection and has_owner):
+            return False
+        if self._connection_repo is None:
+            return False
+
+        # The manager is the run-creation security boundary, so it does not
+        # trust mutable InboundMessage identity fields by themselves. Re-read
+        # the binding by provider identity and require it to match the asserted
+        # connection owner before creating DeerFlow threads or runs.
+        for workspace_id in self._bound_identity_workspace_candidates(msg):
+            connection = await self._connection_repo.find_connection_by_external_identity(
+                provider=msg.channel_name,
+                external_account_id=msg.user_id,
+                workspace_id=workspace_id,
+            )
+            if connection is None:
+                continue
+            if connection.get("id") == msg.connection_id and connection.get("owner_user_id") == msg.owner_user_id:
+                return True
+
+        return False
+
+    async def _requires_bound_identity(self, msg: InboundMessage) -> bool:
         if not self._require_bound_identity:
             return False
         if _auth_disabled_owner_user_id():
             return False
 
-        has_connection = bool(msg.connection_id)
-        has_owner = bool(msg.owner_user_id)
-        return not (has_connection and has_owner)
+        return not await self._has_verified_bound_identity(msg)
 
     async def _reject_unbound_channel_message(self, msg: InboundMessage) -> None:
         logger.info(
@@ -976,7 +1006,7 @@ class ChannelManager:
         outbound = OutboundMessage(
             channel_name=msg.channel_name,
             chat_id=msg.chat_id,
-            thread_id=await self._lookup_thread_id(msg) or "",
+            thread_id="",
             text=BOUND_IDENTITY_REQUIRED_MESSAGE,
             thread_ts=msg.thread_ts,
             metadata=_slim_metadata(msg.metadata),
@@ -1045,7 +1075,7 @@ class ChannelManager:
         self._channel_metadata_synced.add(thread_id)
 
     async def _handle_chat(self, msg: InboundMessage, extra_context: dict[str, Any] | None = None) -> None:
-        if self._requires_bound_identity(msg):
+        if await self._requires_bound_identity(msg):
             await self._reject_unbound_channel_message(msg)
             return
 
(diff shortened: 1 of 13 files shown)


fancyboi999 (commented):
Read through the whole admission path and the invariant holds end to end:

- Every run-creating path funnels through `_handle_chat` — direct chat, the streaming branch (`_handle_streaming_chat` is called from *inside* it, after the guard), and the command path (`_handle_command` → `_handle_chat` for bootstrap/slash-skill turns). Guarding at the top of `_handle_chat` rather than in `_handle_message` is the right call: non-run commands (connect/help) still get through while every `runs.wait`/`runs.stream` is gated.
- `connection_id`/`owner_user_id` aren't forgeable by the platform user — they're resolved server-side in `attach_connection_identity` via `repo.find_connection_by_external_identity(external_account_id=inbound.user_id)`, so an unbound sender has no record and the `not (connection_id and owner_user_id)` check trips. All seven channel adapters (slack/telegram/discord/feishu/dingtalk/wechat/wecom) attach it before dispatch, so no bound user is falsely rejected.
- Secure-by-default: `require_bound_identity` defaults to `True` on the config model and is only armed when `channel_connections.enabled`, with the legacy open-bot escape hatch documented.

`tests/test_channels.py::TestChannelManagerBoundIdentityPolicy` + the config test pass locally (8 passed) and cover the three states that matter: unbound rejected before thread/run creation, bound allowed, auth-disabled still routed to the default user.

Reads correct to me.

backend/app/channels/manager.py
```diff
@@ -1012,6 +1038,10 @@ async def _update_thread_channel_metadata(self, client, msg: InboundMessage, thr
         self._channel_metadata_synced.add(thread_id)
 
     async def _handle_chat(self, msg: InboundMessage, extra_context: dict[str, Any] | None = None) -> None:
+        if self._requires_bound_identity(msg):
```
fancyboi999: Confirmed this is the single choke point: `_handle_streaming_chat` is invoked from within `_handle_chat` further down, and `_handle_command` routes its bootstrap/slash-skill turns back through here too — so guarding here covers every run-creating path in one place while leaving non-run command handling untouched. Good placement.

willem-bd (commented):
## Code Review — `fix(channels): require bound identity for user-owned IM messages`

**+195 / −2 across 6 files** · closes #3538

### Overview

Establishes a security admission invariant for user-owned IM channels: **in auth-enabled deployments with `channel_connections` enabled, ordinary chat messages may only create DeerFlow threads/runs when the platform identity has been bound to a DeerFlow owner** (message carries both `connection_id` and `owner_user_id`).

- New `require_bound_identity: bool = True` on `ChannelConnectionsConfig`.
- `ChannelService`/`ChannelManager` thread the flag; `from_app_config` derives it from `channel_connections.enabled AND require_bound_identity`.
- `_handle_chat` gains an early gate (`_requires_bound_identity` → `_reject_unbound_channel_message`) that fires **before** thread lookup/creation.
- Commands (`/connect`, `/new`, …) are deliberately left alone — they route through `_handle_command`, bypassing `_handle_chat`.

### What's good

- **Correct placement.** The gate sits at the top of `_handle_chat`, ahead of `_lookup_thread_id`/`_create_thread`/`runs.*`. The "before thread/run creation" claim holds.
- **Sound boolean.** `not (bool(msg.connection_id) and bool(msg.owner_user_id))` requires **both** fields — a worker that sets only one still gets rejected, so there's no half-bound bypass. Both fields are populated only by the channel workers' connection-record lookup (which platform users can't spoof).
- **Auth-disabled exemption.** Reusing `_auth_disabled_owner_user_id()` keeps no-auth local mode working, consistent with the `_apply_effective_owner` call already in `_handle_message`.
- **No state created on rejection.** `_reject_unbound_channel_message` does one store **read** + a bus publish — no thread, run, or persisted mapping.
- **Constructor defaults are the safe direction.** Manager/service default to `False` (don't surprise direct instantiations); `from_app_config` computes the real value; the config model defaults to `True` (secure by default). Deliberately different and correct.
- **Test coverage is thorough.** The four manager tests pin the full matrix — (policy on/off) × (auth on/off) × (bound/unbound) — asserting both the positive path (`threads.create`/`runs.wait` called, `run_context` identities correct) and the negative (`assert_not_called()`). Plus config + service-forwarding tests.

### Suggestions / nits

1. **`config.example.yaml` is missing the new field.** `IM_CHANNEL_CONNECTIONS.md` documents `require_bound_identity`, but the canonical `config.example.yaml` commented `channel_connections:` block (~lines 1207–1208) doesn't surface it. Operators copying from the example won't discover the knob or its secure default. Worth adding a commented line there.

2. **`channel_connections` fetched twice in `from_app_config`.** The new code reads `getattr(app_config, "channel_connections", None)` into a local, then `_make_connection_repo(app_config)` reads it again internally. Minor DRY — could pass the already-resolved config to a helper or inline it.

3. **Long boolean expression (service.py).** The single-line `require_bound_identity = bool(connection_config is not None and getattr(...) and getattr(...))` is within the 240-col limit but reads as a wall; breaking it across lines would help reviewers. It does follow the existing `getattr(..., "enabled", False)` defensive pattern in this file, so it's consistent — purely readability.

4. **Rejection copy could hint at the IM-native connect step.** `BOUND_IDENTITY_REQUIRED_MESSAGE` says *"Connect this channel from DeerFlow Settings first…"*. For most providers the user then has to type `/connect <code>` **back in the channel**. Not wrong (Settings is the first stop), but a user already inside Slack/Telegram may not realize there's a second in-channel step. Optional, e.g. *"...then run `/connect` with the code from Settings"*.

5. **Unbound messages still occupy a semaphore slot.** The check runs inside `_handle_chat`, under `self._semaphore`. The reject path is fast so it's not a real DoS surface — but lifting the check above `async with self._semaphore:` in `_handle_message` would free the slot immediately. Low priority.

### Risks

- **Behavior change for existing auth-enabled + channel-connections deployments.** Deployments relying on unbound platform users falling through to a platform-ID user bucket will now see those messages rejected. That's the intended fix, and the legacy escape hatch (`require_bound_identity: false`) plus docs cover it — but worth calling out in release notes since it's a default-on gating change.
- **`require_bound_identity` is startup-only.** It inherits `channel_connections`'s `STARTUP_ONLY_FIELDS` registration, so toggling it needs a restart. Correct, just noting.

### Verdict

Solid, well-scoped security fix with excellent test coverage and correct gate placement. The core invariant is sound with no bypass path. Items above are all minor (docs/DRY/UX); none block merge. I'd ask for the `config.example.yaml` addition (#1) before merge since it affects operator discoverability of a security-relevant default.

ggnnggez:
@willem-bd 
 Thanks for the detailed review. Addressed the actionable items:

  - Added `require_bound_identity: true` to `config.example.yaml`, with a security warning
  explaining that disabling it restores legacy open-bot behavior for unbound IM users.
  - Refactored `from_app_config()` to reuse the resolved `channel_connections` config instead
  of fetching it again inside `_make_connection_repo()`.
  - Split the long `require_bound_identity` expression for readability.
  - Updated the unbound-message rejection copy to mention completing the in-channel connect
  step, so it covers both Telegram deep links and `/connect <code>` providers.
  - Moved the ordinary-chat unbound rejection before semaphore acquisition, while keeping the
  `_handle_chat()` guard for command paths that later create runs.
  - Added an extra manager-level verification step: before creating threads/runs, the manager
  now re-reads the channel connection by provider identity and requires the asserted
  `connection_id` / `owner_user_id` to match. This keeps the run-creation boundary from
  trusting mutable `InboundMessage` fields alone.

  Validation:
  - `uv run pytest tests/test_channel_connections_config.py tests/test_channels.py -q`
  - `uv run ruff check app/channels/manager.py app/channels/service.py packages/harness/
  deerflow/config/channel_connections_config.py tests/test_channel_connections_config.py
  tests/test_channels.py`

backend/app/channels/manager.py
ggnnggez: Thanks for review.

backend/app/channels/manager.py
```diff
@@ -55,6 +55,8 @@
 STREAM_MODES = ["messages-tuple", "values"]
 MESSAGE_STREAM_EVENTS = ("messages-tuple", "messages")
 THREAD_BUSY_MESSAGE = "This conversation is already processing another request. Please wait for it to finish and try again."
+BOUND_IDENTITY_REQUIRED_MESSAGE = "Connect this channel from DeerFlow Settings, complete the in-channel connect step, then send your message again."
+WORKSPACE_OPTIONAL_CONNECTION_PROVIDERS = {"dingtalk", "discord", "wecom"}
```
hetaoBackend: **Two sources of truth coupled only by convention.** This hardcoded set must stay identical to the providers that pass `fallback_without_workspace=True` to `attach_connection_identity` — today exactly `dingtalk.py:493`, `discord.py:457`, `wecom.py:338`. They match now, but nothing pins them together. If a new provider is added (or a provider's `fallback_without_workspace` flips) and this set isn't updated in lockstep, the failure is silent: `_has_verified_bound_identity` returns `False` and a *legitimately bound* user gets rejected with the connect prompt.

Could this reuse the same per-provider `fallback_without_workspace` decision (threaded through to the manager) instead of a separate hardcoded set? At minimum, a test asserting this set == the providers using `fallback_without_workspace=True` would fail CI on drift.

backend/app/channels/manager.py
```diff
@@ -918,38 +922,97 @@ def _log_task_error(task: asyncio.Task) -> None:
...
+            logger.exception(
+                "Error handling message from %s (chat=%s)",
+                msg.channel_name,
+                msg.chat_id,
+            )
+            await self._send_error(msg, "An internal error occurred. Please try again.")
 
     # -- chat handling -----------------------------------------------------
 
+    def _bound_identity_workspace_candidates(self, msg: InboundMessage) -> list[str | None]:
```
hetaoBackend: `attach_connection_identity` already overwrites `inbound.workspace_id` with the connection's stored value (`connection_identity.py:41`) before the message reaches the manager, and this method's lookup only runs when `connection_id`+`owner` are set (i.e. attach already matched). So is the second (`None`) candidate here ever actually reachable on the bound path? If it's purely defensive, a comment would help; if it's load-bearing, a test should exercise it.

backend/packages/harness/deerflow/config/channel_connections_config.py
```diff
@@ -42,6 +42,7 @@ class ChannelConnectionsConfig(BaseModel):
     """Top-level config for browser-connectable IM channels."""
 
     enabled: bool = False
+    require_bound_identity: bool = True
```
hetaoBackend, line 45: Adding a field to the config schema — per `backend/CLAUDE.md` ("When changing the config schema, bump `config_version` in `config.example.yaml`"), `config_version` should be bumped (currently still `12`). Otherwise the outdated-config warning won't fire and `make config-upgrade` won't surface the new field.

backend/app/channels/manager.py
```diff
@@ -1012,6 +1075,10 @@ async def _update_thread_channel_metadata(self, client, msg: InboundMessage, thr
         self._channel_metadata_synced.add(thread_id)
 
     async def _handle_chat(self, msg: InboundMessage, extra_context: dict[str, Any] | None = None) -> None:
+        if await self._requires_bound_identity(msg):
```
hetaoBackend: Redundant binding query on the hot path: for a normal bound chat, `_requires_bound_identity` already ran pre-semaphore in `_handle_message` (~L926), and this re-check issues a second `find_connection_by_external_identity` for every legitimate message. The pre-semaphore check is justified (reject before consuming a slot), but is the in-`_handle_chat` re-check needed for the direct path, or only for the `/bootstrap` / slash-routing callers? Could scope it so the happy path queries once.

backend/app/channels/manager.py
```diff
@@ -918,38 +922,97 @@ def _log_task_error(task: asyncio.Task) -> None:
...
+
+    async def _requires_bound_identity(self, msg: InboundMessage) -> bool:
+        if not self._require_bound_identity:
+            return False
+        if _auth_disabled_owner_user_id():
+            return False
+
+        return not await self._has_verified_bound_identity(msg)
+
+    async def _reject_unbound_channel_message(self, msg: InboundMessage) -> None:
```
hetaoBackend: Asymmetry vs `_send_error` / the command reply, which both set `connection_id` and `owner_user_id` on the outbound. The gate also rejects the *unverified-binding* case (`connection_id` present but owner mismatch); dropping `connection_id` here means outbound routing / per-connection credential selection can't see it. Intended?

hetaoBackend (commented):
Reviewed the diff against the merge-base; the new bound-identity tests pass locally (12/12). The admission gate is well-placed at all run-creation entry points, it re-reads the binding from the repo instead of trusting mutable `InboundMessage` fields (nice defense-in-depth), and it correctly no-ops in auth-disabled and legacy open-bot modes. It faithfully implements #3538. A few questions inline; two cross-cutting ones below.

**Behavior flip on upgrade (please confirm it's in release notes, not just the feature doc).** `require_bound_identity` defaults to `True` and `service.py` applies `getattr(..., "require_bound_identity", True)`. So an existing auth-enabled deployment that already has `channel_connections.enabled: true` and relied on unbound users chatting will start rejecting them after this upgrade, with no config change on their side. This is the intended secure default per the issue, but since it's a `startup-only` field it's a silent runtime change — operators running a legacy operator-bot need to know to set `require_bound_identity: false` *before* upgrading.

**Fail-closed message is misleading on infra failure.** If `require_bound_identity` is true but `_make_connection_repo` returns `None` (e.g. `get_session_factory()` is `None` because the DB is unavailable at startup), `_has_verified_bound_identity` returns `False`, so *every* message is rejected with "Connect this channel from DeerFlow Settings…". Failing closed is the right call, but the text blames the user for a server-side outage. Worth differentiating transient unavailability.

ggnnggez:
@hetaoBackend 
Thanks, addressed the review feedback in the follow-up commits:

  - Removed the separate manager-side `WORKSPACE_OPTIONAL_CONNECTION_PROVIDERS` source of
  truth. The manager now re-checks against the normalized `msg.workspace_id` written by
  `attach_connection_identity()`, so the second `None` workspace candidate is gone rather than
  kept as defensive behavior.
  - Renamed the identity helper from `_requires_bound_identity()` to
  `_get_bound_identity_rejection()` / rejection-oriented flow so the return value is clearer:
  `None` means the message may proceed, a `_BoundIdentityRejection` means it must be rejected.
  - Avoided the hot-path duplicate binding lookup. `_handle_message()` and `_handle_command()`
  perform the admission check once, then pass `bound_identity_checked=True` into
  `_handle_chat()` for already-checked paths. Direct `_handle_chat()` callers still fail
  closed by default.
  - Preserved outbound routing context for the unverified-binding reject path without trusting
  inbound identity fields. Rejection outbounds now use only server-side connection metadata
  re-read from the repository (`outbound_connection_id` / `outbound_owner_user_id`) when
  available; fully unbound rejects still send no connection metadata.
  - Bumped `config.example.yaml` to `config_version: 14` after rebasing onto a `main` that
  already had version `13`, so `make config-upgrade` and outdated-config warnings will surface
  `require_bound_identity`.

  Validation:
  - `cd backend && uv run pytest tests/test_channels.py::TestChannelManagerBoundIdentityPolicy
  -q`
  - `cd backend && uv run pytest tests/test_config_version.py -q`
  - `cd backend && uv run ruff check app/channels/manager.py tests/test_channels.py`

WillemJiang (approved):

PR #3579, opened 2026-06-14
fix(channels): scope IM files and helper commands to owner

### LINKED ISSUE #3539: [IM] Scope channel files and helper commands to the connection owner
## Context

Connected IM messages carry both the external provider identity and the DeerFlow owner identity. The run path already uses the owner correctly: `ChannelManager` builds owner headers for Gateway calls and puts the owner in run context, while preserving the raw platform user as `channel_user_id`.

The gap is that several file and helper-command paths still resolve user-scoped resources through the ambient `ContextVar` / default user. Channel workers run outside a normal browser request context, so these calls can resolve to `default` instead of the bound `owner_user_id`.

Relevant code paths:

- `backend/app/channels/manager.py::_resolve_attachments()` calls `get_effective_user_id()` before resolving `/mnt/user-data/outputs/...` artifacts.
- `backend/app/channels/manager.py::_ingest_inbound_files()` calls `ensure_uploads_dir(thread_id)`, which also uses `get_effective_user_id()`.
- `backend/app/channels/feishu.py::_receive_single_file()` resolves the upload directory via `get_effective_user_id()`.
- `backend/app/channels/manager.py::_fetch_gateway()` sends internal auth without the connection owner header for `/models` and `/memory` slash commands.
- `backend/app/gateway/routers/memory.py` reads memory by effective user ID, so `/memory` is owner-sensitive.

## Problem

A connected user can successfully run the agent as their DeerFlow account, but related IM file and helper-command behavior can silently point at the wrong user bucket:

- Inbound files from IM may be written under the default/internal user instead of the connected owner.
- Agent artifacts may be looked up in the wrong output directory, causing missing attachments in provider replies.
- `/memory` can display default/internal memory instead of the connected owner’s memory.
- Tests currently cover run-context ownership but do not fully cover owner-scoped file or helper-command behavior.

This is a consistency and data-isolation bug: every side effect of a bound IM message should use the same DeerFlow owner identity.

## What to build

Propagate the resolved IM owner identity through every user-scoped channel path, not only the run creation path.

The desired invariant is: if `InboundMessage.owner_user_id` is set, then all user-scoped reads/writes caused by that message use that owner unless a code path intentionally uses provider-scoped metadata.

## Implementation notes

- Prefer an explicit `owner_user_id` parameter over hidden ambient context for channel-manager helper functions.
- Alternatively, set the current user context around the entire bound-message handling block and reset it in a `finally`, but be careful with async tasks and provider callbacks.
- Update upload helpers only if needed; avoid making generic upload APIs depend on channel concepts.
- `_fetch_gateway()` should accept the inbound message or owner headers so `/memory` and future owner-scoped slash commands are correct.
- Keep `/models` behavior unchanged except for harmlessly passing owner headers.
- Provider-specific file receivers, especially Feishu, should receive the same owner identity used for the message.

## Acceptance criteria

- [ ] Inbound files received from a bound IM message are staged under the bound DeerFlow owner’s upload bucket.
- [ ] Outbound artifacts generated by a bound run are resolved from the bound DeerFlow owner’s output bucket and can be attached to provider replies.
- [ ] `/memory` invoked from a bound IM conversation returns the bound owner’s memory, not the default/internal user’s memory.
- [ ] Auth-disabled local mode still resolves files and commands under the auth-disabled default user.
- [ ] The implementation does not weaken existing path traversal protections for artifact delivery.
- [ ] Regression tests cover inbound file ingest, outbound artifact resolution, and owner-scoped `/memory` command behavior.

## Validation plan

- Add tests that create two user buckets and assert a bound IM message reads/writes only the owner bucket.
- Add a `/memory` command test with a bound owner and verify the internal owner header is sent.
- Run the existing channel file attachment and channel connection suites.

## Blocked by

None - can start immediately.

ggnnggez:
Closes #3539

# Summary

- Thread owner identity through inbound file staging so bound IM uploads land in the DeerFlow owner's upload bucket.
- Use owner headers for manager helper commands that call Gateway endpoints such as `/models` and `/memory`.
- Preserve raw platform identity separately from owner identity for platform-facing behavior.
- Update Feishu inbound file handling to accept the explicit storage owner.

# Tests

- `uv run pytest tests/test_channels.py -q`
- `uv run pytest tests/test_uploads_manager.py tests/test_channel_file_attachments.py tests/blocking_io/test_channels_ingest.py -q`
- `uv run pytest tests/test_feishu_parser.py -q`
- `uv run ruff check app/channels/manager.py app/channels/base.py app/channels/feishu.py packages/harness/deerflow/uploads/manager.py tests/test_channels.py`

# Notes

This PR completes the owner propagation path after #3538 blocks unbound runs. Bound IM users now use the same DeerFlow owner for files, commands, run context, and Gateway calls.

### DIFF
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -436,6 +436,7 @@ Bridges external messaging platforms (Feishu, Slack, Telegram, Discord, DingTalk
 - Per-agent per-user memory at `{base_dir}/users/{user_id}/agents/{agent_name}/memory.json`
 - Custom agent definitions (`SOUL.md` + `config.yaml`) are also per-user at `{base_dir}/users/{user_id}/agents/{agent_name}/`. The legacy shared layout `{base_dir}/agents/{agent_name}/` remains read-only fallback for unmigrated installations
 - `user_id` is resolved via `get_effective_user_id()` from `deerflow.runtime.user_context`
+- The `/api/memory*` endpoints resolve the owner through `_resolve_memory_user_id(request)`: trusted internal callers (IM channel workers carrying the `X-DeerFlow-Owner-User-Id` header, e.g. a bound `/memory` command) act for the connection owner; browser/API callers fall back to `get_effective_user_id()`. The header is only honored after `AuthMiddleware` validated the internal token, mirroring `get_trusted_internal_owner_user_id` used by the threads router
 - In no-auth mode, `user_id` defaults to `"default"` (constant `DEFAULT_USER_ID`)
 - Absolute `storage_path` in config opts out of per-user isolation
 - **Migration**: Run `PYTHONPATH=. python scripts/migrate_user_isolation.py` to move legacy `memory.json`, `threads/`, and `agents/` into per-user layout. Supports `--dry-run` (preview changes) and `--user-id USER_ID` (assign unowned legacy data to a user, defaults to `default`).
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -405,6 +405,12 @@ Bridges external messaging platforms (Feishu, Slack, Telegram, Discord, DingTalk
 10. For commands (`/new`, `/status`, `/models`, `/memory`, `/help`): handle locally or query Gateway API
 11. Outbound → channel callbacks → platform reply
 
+**Owner-scoped file storage**: inbound files, uploads, and output artifacts are staged under the DeerFlow owner's bucket so they land where the agent run reads/writes (`users/{user_id}/threads/{thread_id}/user-data/{uploads,outputs}`). `ChannelManager._handle_chat` resolves the storage owner once via `_channel_storage_user_id(msg)` (sanitized owner id, falling back to `safe(msg.user_id)` for unbound auth-enabled channels — mirroring `_resolve_run_params`'s run identity; `None` only when no identity is available) and threads it as the `user_id=` kwarg through the file pipeline:
+- `Channel.receive_file(msg, thread_id, user_id=...)` — owner-bound channels persist downloaded files under the owner's bucket instead of the default bucket
+- `_ingest_inbound_files(...)` and the underlying `ensure_uploads_dir` / `get_uploads_dir` — owner-scoped via the same kwarg
+- `_resolve_attachments` / `_prepare_artifact_delivery` — resolve output artifacts from the bound owner's bucket
+The cached value is reused for both the blocking (`runs.wait`) and streaming (`_handle_streaming_chat`) paths, so uploads and artifact delivery always target the same bucket even if a channel returns a rewritten `InboundMessage` from `receive_file`. The bucket id matches the memory bucket resolved by `_resolve_memory_user_id` (both normalize through `make_safe_user_id`).
+
 **Configuration** (`config.yaml` -> `channels`):
 - `langgraph_url` - LangGraph-compatible Gateway API base URL (default: `http://localhost:8001/api`)
 - `gateway_url` - Gateway API URL for auxiliary commands (default: `http://localhost:8001`)
@@ -623,7 +629,7 @@ Multi-file upload with automatic document conversion:
 - Supports: PDF, PPT, Excel, Word documents (converted via `markitdown`)
 - Rejects directory inputs before copying so uploads stay all-or-nothing
 - Reuses one conversion worker per request when called from an active event loop
-- Files stored in thread-isolated directories
+- Files stored in thread-isolated directories under the resolving user's bucket (`users/{user_id}/threads/{thread_id}/user-data/uploads`). For IM channels the owner is threaded explicitly via the `user_id=` kwarg (see IM Channels → Owner-scoped file storage); HTTP/embedded callers resolve it from `get_effective_user_id()`
 - Duplicate filenames in a single upload request are auto-renamed with `_N` suffixes so later files do not truncate earlier files
 - Agent receives uploaded file list via `UploadsMiddleware`
 
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -519,10 +519,24 @@ def _safe_user_id_for_run(raw_user_id: str) -> str:
 
 
 def _channel_storage_user_id(msg: InboundMessage) -> str | None:
+    """Resolve the storage bucket owner for an inbound message's files/artifacts.
+
+    Mirrors the run-identity policy in ``_resolve_run_params`` so uploads land in
+    and artifacts are read from the same bucket the agent run uses: prefer the
+    bound DeerFlow owner, otherwise fall back to the sanitized raw platform user
+    id. Without this fallback, an unbound auth-enabled channel would run under
+    ``safe(msg.user_id)`` but stage files under ``get_effective_user_id()`` (the
+    dispatcher task's unset contextvar → ``"default"``), so uploads would land in
+    ``users/default/...`` while the agent reads ``users/{safe_platform_user_id}/...``.
+    Returns ``None`` only when neither identity is available, leaving the caller to
+    fall back to the contextvar/default user.
+    """
     owner_user_id = _effective_owner_user_id(msg)
-    if not owner_user_id:
-        return None
-    return _safe_user_id_for_run(owner_user_id)
+    if owner_user_id:
+        return _safe_user_id_for_run(owner_user_id)
+    if msg.user_id:
+        return _safe_user_id_for_run(msg.user_id)
+    return None
 
 
 def _resolve_slash_skill_command(
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -519,17 +519,24 @@ def _safe_user_id_for_run(raw_user_id: str) -> str:
 
 
 def _channel_storage_user_id(msg: InboundMessage) -> str | None:
-    """Resolve the storage bucket owner for an inbound message's files/artifacts.
+    """Resolve the canonical DeerFlow user id for a channel-triggered message.
 
-    Mirrors the run-identity policy in ``_resolve_run_params`` so uploads land in
-    and artifacts are read from the same bucket the agent run uses: prefer the
-    bound DeerFlow owner, otherwise fall back to the sanitized raw platform user
-    id. Without this fallback, an unbound auth-enabled channel would run under
-    ``safe(msg.user_id)`` but stage files under ``get_effective_user_id()`` (the
-    dispatcher task's unset contextvar → ``"default"``), so uploads would land in
-    ``users/default/...`` while the agent reads ``users/{safe_platform_user_id}/...``.
-    Returns ``None`` only when neither identity is available, leaving the caller to
-    fall back to the contextvar/default user.
+    Single source of truth for both the agent **run identity**
+    (``_resolve_run_params`` → ``run_context["user_id"]``) and the **file/artifact
+    storage bucket** (``receive_file`` / ``_ingest_inbound_files`` /
+    ``_prepare_artifact_delivery``), so the bucket the agent reads/writes always
+    matches where channel files are staged. Prefer the bound DeerFlow owner,
+    otherwise fall back to the sanitized raw platform user id. Without that
+    fallback, an unbound auth-enabled channel would run under ``safe(msg.user_id)``
+    but stage files under ``get_effective_user_id()`` (the dispatcher task's unset
+    contextvar → ``"default"``), so uploads would land in ``users/default/...``
+    while the agent reads ``users/{safe_platform_user_id}/...``. Returns ``None``
+    only when neither identity is available, leaving the caller to fall back to the
+    contextvar/default user.
+
+    Distinct from :func:`_owner_headers`, which deliberately sends the *raw* owner
+    id (no sanitize, no platform fallback) over HTTP for gateway to re-resolve;
+    this helper is the in-process, sanitized, filesystem-facing identity.
     """
     owner_user_id = _effective_owner_user_id(msg)
     if owner_user_id:
@@ -845,11 +852,12 @@ class ChannelManager:
         # owns the connection. Preserve the raw platform user under
         # ``channel_user_id`` for platform-facing lookups and audits.
         run_context_identity: dict[str, Any] = {"thread_id": thread_id}
-        owner_user_id = _effective_owner_user_id(msg)
-        if owner_user_id:
-            run_context_identity["user_id"] = _safe_user_id_for_run(owner_user_id)
-        elif msg.user_id:
-            run_context_identity["user_id"] = _safe_user_id_for_run(msg.user_id)
+        # Single source of truth for the run identity: the same helper that scopes
+        # inbound files and outbound artifacts, so the bucket the agent reads/writes
+        # always matches where channel files are staged.
+        run_user_id = _channel_storage_user_id(msg)
+        if run_user_id:
+            run_context_identity["user_id"] = run_user_id
         if msg.user_id:
             run_context_identity["channel_user_id"] = msg.user_id
 
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -1213,7 +1213,10 @@ class ChannelManager:
             len(artifacts),
         )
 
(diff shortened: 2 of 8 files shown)


backend/app/gateway/routers/memory.py
```diff
@@ -18,6 +19,19 @@
...
+    """Resolve the memory owner for this request.
+
+    Honors the trusted internal owner header that channel workers attach when
+    acting for a connection owner, so an IM ``/memory`` command reads the bound
+    owner's memory instead of the synthetic internal user. The header is only
+    honored after ``AuthMiddleware`` validated the internal token (see
+    ``get_trusted_internal_owner_user_id``). Browser/API callers are never
+    internal, so this falls back to the normal contextvar-based effective user.
+    """
+    return get_trusted_internal_owner_user_id(request) or get_effective_user_id()
```
willem-bd: **Memory router uses the raw owner header value without sanitizing.**

`get_trusted_internal_owner_user_id(request)` returns the header verbatim, then it flows into `get_memory_data(user_id=...)` → `user_dir(user_id)` → `_validate_user_id(user_id)`, which raises `ValueError` for anything outside `^[A-Za-z0-9_\-]+$`. The `owner_user_id` column is `String(64)` with no charset constraint, and the rest of the channel pipeline already runs the same id through `_safe_user_id_for_run` (= `make_safe_user_id` / `prepare_user_dir_for_raw_id`).

For a UUID/`default` owner this is a no-op, but for any owner_user_id needing sanitization:
- `/memory` returns 500 → channel reports `"Failed to fetch memory information."`
- The owner's *files* live under `users/{safe_owner_id}/...` (via `_channel_storage_user_id`), but the *memory* path here would target `users/{raw_owner_id}/...` if validation ever loosened.

Suggest aligning by sanitizing here too, e.g.

```python
raw = get_trusted_internal_owner_user_id(request)
if raw:
    return make_safe_user_id(raw)
return get_effective_user_id()
```

or better: have `_resolve_memory_user_id` reuse whatever helper `_resolve_run_params` consolidates around so the run-identity, file-bucket, and memory-bucket are guaranteed to match for a given owner.

backend/app/channels/manager.py
```diff
@@ -1094,7 +1104,7 @@ async def _handle_chat(self, msg: InboundMessage, extra_context: dict[str, Any]
             len(artifacts),
         )
 
-        response_text, attachments = _prepare_artifact_delivery(thread_id, response_text, artifacts)
+        response_text, attachments = _prepare_artifact_delivery(thread_id, response_text, artifacts, user_id=_channel_storage_user_id(msg))
```
willem-bd: **Recomputes `_channel_storage_user_id(msg)` after caching it as `storage_user_id` at line 1025.**

Minor cleanup, but also a real consistency hazard: between line 1025 and here, `msg = await channel.receive_file(msg, thread_id, user_id=storage_user_id)` (line 1051) is allowed to return a *different* `InboundMessage`. If a future channel impl ever rewrites `owner_user_id` (e.g. resolving an alias inside `receive_file`), the cached `storage_user_id` and the re-derived value here would point at different buckets — uploads would have been written under one owner and outputs would be looked up under the other.

Reuse the cached value:

```python
response_text, attachments = _prepare_artifact_delivery(
    thread_id, response_text, artifacts, user_id=storage_user_id
)
```

The `_handle_streaming_chat` path at line 1211 has the same pattern — passing `storage_user_id` in from `_handle_chat` would let it reuse the same cached value as well.

backend/app/channels/manager.py
```diff
@@ -498,6 +498,13 @@ def _safe_user_id_for_run(raw_user_id: str) -> str:
         return make_safe_user_id(raw_user_id)
 
 
+def _channel_storage_user_id(msg: InboundMessage) -> str | None:
+    owner_user_id = _effective_owner_user_id(msg)
+    if not owner_user_id:
+        return None
+    return _safe_user_id_for_run(owner_user_id)
```
willem-bd: **Owner identity is now derived three different ways for a single inbound msg**, each producing a different shape:

- `_channel_storage_user_id(msg)` — sanitized owner, `None` when unbound (used for files/uploads/artifacts)
- inline `_safe_user_id_for_run(...)` in `_resolve_run_params` — sanitized owner, **fallback to safe(`msg.user_id`)** when unbound (used for the actual run identity)
- `_owner_headers(msg)` — *raw* owner, no fallback (used for HTTP headers to gateway)

The PR threads the kwarg through 6+ entry points (`receive_file`, `_ingest_inbound_files`, `_resolve_attachments`, `_prepare_artifact_delivery`, `ensure_uploads_dir`, `get_uploads_dir`) and each call-site picks one of the three. A future caller easily picks the wrong helper and silently writes to a bucket the agent never reads.

Worth consolidating to one canonical helper that returns the per-msg storage identity (with the same fallback policy as `_resolve_run_params`) and pinning that as the single source of truth for both file paths and memory routing — the alternative is more of these one-line `user_id=...` arguments accumulating, each with its own subtle resolution rule.

backend/CLAUDE.md
```diff
@@ -436,6 +436,7 @@ Bridges external messaging platforms (Feishu, Slack, Telegram, Discord, DingTalk
 - Per-agent per-user memory at `{base_dir}/users/{user_id}/agents/{agent_name}/memory.json`
 - Custom agent definitions (`SOUL.md` + `config.yaml`) are also per-user at `{base_dir}/users/{user_id}/agents/{agent_name}/`. The legacy shared layout `{base_dir}/agents/{agent_name}/` remains read-only fallback for unmigrated installations
 - `user_id` is resolved via `get_effective_user_id()` from `deerflow.runtime.user_context`
+- The `/api/memory*` endpoints resolve the owner through `_resolve_memory_user_id(request)`: trusted internal callers (IM channel workers carrying the `X-DeerFlow-Owner-User-Id` header, e.g. a bound `/memory` command) act for the connection owner; browser/API callers fall back to `get_effective_user_id()`. The header is only honored after `AuthMiddleware` validated the internal token, mirroring `get_trusted_internal_owner_user_id` used by the threads router
```
willem-bd, line 445: **CLAUDE.md update is partial.** The new line documents `_resolve_memory_user_id`, but the IM Channels and Uploads sections still describe pre-PR behaviour — there's no mention that:

- `Channel.receive_file(...)` now takes a `user_id` kwarg and that owner-bound channels persist downloaded files under the owner's bucket
- `_ingest_inbound_files` / `ensure_uploads_dir` / `get_uploads_dir` are now owner-scoped via the same kwarg
- `_resolve_attachments` / `_prepare_artifact_delivery` resolve outputs from the bound owner's bucket

`backend/CLAUDE.md` says

> CRITICAL: Always update README.md and CLAUDE.md after every code change
> ...
> Update `CLAUDE.md` for development changes (architecture, commands, workflows, internal systems)

A reader looking at the IM Channels section to understand where uploads land on disk still sees the default-bucket story, which is exactly the behaviour this PR is changing for bound users.

backend/app/channels/manager.py
```diff
@@ -1038,12 +1048,12 @@ async def _handle_chat(self, msg: InboundMessage, extra_context: dict[str, Any]
...
             channel = service.get_channel(msg.channel_name) if service else None
             logger.info("[Manager] preparing receive file context for %d attachments", len(msg.files))
-            msg = await channel.receive_file(msg, thread_id) if channel else msg
+            msg = await channel.receive_file(msg, thread_id, user_id=storage_user_id) if channel else msg
         if extra_context:
             run_context.update(extra_context)
 
         original_text = msg.text
-        uploaded = await _ingest_inbound_files(thread_id, msg)
+        uploaded = await _ingest_inbound_files(thread_id, msg, user_id=storage_user_id)
```
willem-bd, line 1187: **Unbound auth-enabled IM channels still cross buckets — the new kwarg leaves this case as it was, even though the threading machinery to fix it is now in place.**

In `_resolve_run_params`, an unbound msg falls back to `_safe_user_id_for_run(msg.user_id)` (i.e. `safe(platform_user_id)`) for the run identity, so the agent reads/writes uploads under `users/{safe_platform_user_id}/threads/{tid}/user-data/uploads/`.

But `_channel_storage_user_id(msg)` returns `None` for unbound msgs, so `_ingest_inbound_files(thread_id, msg, user_id=None)` falls through to `get_effective_user_id()` in the dispatcher task — where the contextvar is unset → `DEFAULT_USER_ID = "default"`. Uploads land in `users/default/.../uploads/` while the agent looks in `users/{safe_platform_user_id}/.../uploads/`. The same mismatch hits `_resolve_attachments` for output artifacts.

This is pre-existing in spirit, but the PR is precisely the place to align it — `_channel_storage_user_id` could mirror `_resolve_run_params`'s policy (fall back to `safe(msg.user_id)` instead of returning `None`), so bound and unbound channels both write where the agent reads. As-is, the new kwarg surface reads like the bug is fixed but only the bound case actually is.

backend/app/channels/manager.py
```diff
@@ -1198,7 +1208,7 @@ async def _handle_streaming_chat(
             response_text = _extract_response_text(result)
             pending_clarification = _has_current_turn_clarification(result)
             artifacts = _extract_artifacts(result)
-            response_text, attachments = _prepare_artifact_delivery(thread_id, response_text, artifacts)
+            response_text, attachments = _prepare_artifact_delivery(thread_id, response_text, artifacts, user_id=_channel_storage_user_id(msg))
```
willem-bd: Minor: `_channel_storage_user_id(msg)` is called inside the `finally` block, which means it does its (synchronous) `_safe_user_id_for_run` work — including a possible `prepare_user_dir_for_raw_id` filesystem touch — even on the streaming-error path. `_safe_user_id_for_run` has a broad `try/except`, so this won't mask the original error, but combined with the `_handle_chat` cache (line 1025) the cleanest fix is to pass `storage_user_id` in from `_handle_chat` and reuse it here, rather than rederiving in two places.

willem-bd (commented):
## Review summary

The owner-scoping plumbing through `receive_file`, `_ingest_inbound_files`, `_resolve_attachments`, `_prepare_artifact_delivery`, and `ensure_uploads_dir`/`get_uploads_dir` looks right for the bound case the PR targets, and the trusted-header gating on `_resolve_memory_user_id` is correctly anchored on `system_role == INTERNAL_SYSTEM_ROLE` so a browser caller cannot spoof another owner.

A few things worth a second look — most are cleanup or robustness, but a couple are real bucket-mismatch risks worth confirming or pushing into a follow-up:

1. **Memory router does not sanitize the raw owner header value** — diverges from `_channel_storage_user_id`, which sanitizes via `_safe_user_id_for_run`/`make_safe_user_id`. For owner ids needing sanitization, files and memory land in different buckets (and `/memory` 500s).
2. **`_handle_chat` recomputes `_channel_storage_user_id(msg)` after caching it** — line 1025 caches `storage_user_id`; line 1107 re-derives it. If a custom channel ever returns a new `InboundMessage` from `receive_file`, the two values can disagree and artifact delivery looks at a different bucket than where uploads were written.
3. **Three helpers express "who owns this msg"** (`_channel_storage_user_id`, `_safe_user_id_for_run`, `_owner_headers`), each yielding a different shape (safe / safe-with-platform-fallback / raw). Easy to wire a future call site to the wrong one.
4. **CLAUDE.md update is partial** — the IM Channels and Uploads sections still describe pre-PR behaviour. `backend/CLAUDE.md` says "Always update README.md and CLAUDE.md after every code change".
5. **Unbound auth-enabled IM channels still cross buckets** (pre-existing): the agent runs under `safe(platform_user_id)` from `_resolve_run_params`, but `_channel_storage_user_id(msg)` returns `None`, so `_ingest_inbound_files`/`_resolve_attachments` fall back to `get_effective_user_id()` in the dispatcher task → `"default"`. Files write to `users/default/...` while the agent reads from `users/{safe_platform_user_id}/...`. Worth aligning while the threading is being added; otherwise the new kwarg surface looks like the bug is fixed when only the bound case is.
6. **`_handle_streaming_chat` finally-block re-derivation** does sync filesystem-style work in an error path; minor.

Detailed comments inline.

backend/app/gateway/routers/memory.py
ggnnggez: Fixed in 52e87a9. `_resolve_memory_user_id` now runs the trusted owner header through `make_safe_user_id(...)` before routing — the same normalization the file pipeline applies via `_safe_user_id_for_run` / `prepare_user_dir_for_raw_id` (which itself returns `make_safe_user_id(raw)`), so the memory bucket matches the owner file/upload bucket and an unsafe raw id no longer 500s in `_validate_user_id`. Added `test_get_memory_sanitizes_unsafe_owner_header`.

backend/app/channels/manager.py
ggnnggez: Fixed in 696d2ed. `_handle_chat` now reuses the `storage_user_id` cached at the top of the method for `_prepare_artifact_delivery` instead of re-deriving `_channel_storage_user_id(msg)`.

backend/app/channels/manager.py
ggnnggez: Consolidated in c2bb176. After the unbound fix below, the inline resolution in `_resolve_run_params` and `_channel_storage_user_id` computed the identical sanitized-with-platform-fallback value, so `_resolve_run_params` now calls the same helper — it is the single source of truth for both `run_context["user_id"]` and the file/artifact bucket. `test_run_identity_matches_storage_bucket` pins run identity == storage identity so they cannot drift again.

`_owner_headers` is intentionally kept separate rather than merged: it sends the *raw* owner over HTTP for the gateway to re-resolve, and must return `None` for an unbound message rather than fall back to `safe(msg.user_id)` — that platform id is not a DeerFlow account, so forging an owner header would be wrong. Storage needs the opposite (a platform fallback, since files always need a home). Those mutually-exclusive fallback policies are why only #2/#3 merge; documented the contract on both helpers.

backend/CLAUDE.md
ggnnggez: Documented in fd8e9db. Added an "Owner-scoped file storage" subsection to IM Channels covering the `user_id=` kwarg threaded through `receive_file` / `_ingest_inbound_files` / `ensure_uploads_dir` / `get_uploads_dir` / `_resolve_attachments` / `_prepare_artifact_delivery`, and noted the bucket matches `_resolve_memory_user_id` (both normalize via `make_safe_user_id`). Also updated the File Upload section.

backend/app/channels/manager.py
ggnnggez: Aligned in 91a7fd1. `_channel_storage_user_id` now mirrors the `_resolve_run_params` policy: prefer the bound owner, else fall back to `safe(msg.user_id)`, returning `None` only when neither identity is available. Bound and unbound auth-enabled channels now stage files where the agent reads, instead of the unbound case landing in `users/default/...`. `test_channel_storage_user_id_falls_back_to_platform_user` covers bound / unbound / anonymous.

backend/app/channels/manager.py
ggnnggez: Fixed in c510b93. `storage_user_id` is now threaded into `_handle_streaming_chat` and reused in the finally block, so we no longer re-run `_safe_user_id_for_run` (and its possible filesystem touch) on the streaming-error path, and artifact delivery targets the same bucket as the uploads.

ggnnggez:
Thanks for the thorough review, @willem-bd — all six points addressed (replies inline). Branch was rebased onto the latest `main`. Summary:

| # | Item | Commit |
|---|------|--------|
| 1 | Memory router sanitizes the raw owner header | `52e87a9` |
| 2 | `_handle_chat` reuses the cached storage owner | `696d2ed` |
| 3 | Run identity + storage bucket unified into one resolver | `c2bb176` |
| 4 | CLAUDE.md owner-scoped storage docs | `fd8e9db` |
| 5 | Unbound auth-enabled channels stage files under the run bucket | `91a7fd1` |
| 6 | Streaming finally-block reuses the cached owner | `c510b93` |

On #3: `_owner_headers` is deliberately *not* merged with `_channel_storage_user_id` — they have mutually-exclusive unbound fallback policies (raw owner / `None` for the HTTP act-as header vs. sanitized platform fallback for the local bucket), detailed in the inline reply.

Tests: `test_channels.py`, `test_channel_file_attachments.py`, `test_memory_router.py`, `tests/blocking_io/test_channels_ingest.py` all green.

WillemJiang (approved):

PR #3580, opened 2026-06-14
fix(channels): make runtime provider state authoritative

### LINKED ISSUE #3541: [IM] Make runtime provider status transactional and instance-wide
## Context

Runtime channel credentials are instance-wide. An admin can configure or remove provider credentials through the browser, and the shared channel worker is started/stopped for the whole deployment. Per-user `channel_connections` rows then bind DeerFlow users to external identities on top of that shared provider worker.

The current implementation mixes instance-wide and current-user behavior in a few places:

- Runtime configure updates `request.app.state.channels_config` before confirming the provider worker starts successfully.
- Runtime provider disconnect removes/stops the shared worker, but only revokes the current admin user’s connections.
- Provider responses can still surface a connected row even when the shared provider is unavailable or removed.
- Frontend settings/sidebar treat `connection_status === "connected"` or a connected row as enough to show a connected badge.

Relevant code paths:

- `backend/app/gateway/routers/channel_connections.py::configure_channel_provider_runtime()`
- `backend/app/gateway/routers/channel_connections.py::disconnect_channel_provider_runtime()`
- `backend/app/gateway/routers/channel_connections.py::_provider_response()`
- `frontend/src/components/workspace/settings/channels-settings-page.tsx`
- `frontend/src/components/workspace/channels/workspace-channels-list.tsx`

## Problem

Users can see stale or misleading channel state:

- If runtime configure fails to start the worker, in-memory state can still contain the submitted credentials/config for the remainder of the process.
- If an admin disconnects a shared provider runtime, other users’ provider connection rows can remain `connected` even though the worker is gone.
- The UI can render a connected state from stale row data instead of showing that the provider is unavailable/unconfigured.

This causes confusing UX and can also hide operational failures. Runtime provider state should be transactional and should reflect the fact that the provider worker is shared by all users.

## What to build

Make runtime provider state updates transactional and make provider availability authoritative over stale per-user connection rows.

A good target model:

- Configure: validate credentials, start/restart provider, then commit state; rollback if start fails.
- Disconnect: stop/remove provider and update all affected users’ provider connection state or expose a provider-unavailable state that overrides connected rows.
- UI: show “unavailable”, “needs setup”, or “needs reconnect” when the shared provider cannot receive messages, even if a historical user connection row exists.

## Implementation notes

- Avoid mutating `app.state.channels_config` until provider startup succeeds, or keep a copy and restore it on failure.
- Add repository support to mark/revoke all active connections for a provider when runtime credentials are removed, if that is the chosen model.
- Alternatively, leave rows intact but expose provider availability separately and make the UI treat unavailable runtime state as not actively connected.
- Be explicit about the meaning of `revoked`, `not_connected`, `needs_reauth`, and provider-level `unavailable_reason`.
- Consider whether reconnect should reuse old conversation mappings or require new mappings.
- Keep auth-disabled local mode behavior understandable: a configured running channel may still appear connected without per-user bindings.

## Acceptance criteria

- [ ] Failed runtime configure does not leave submitted credentials/config in active in-memory provider state.
- [ ] Runtime provider disconnect consistently affects all users because the provider worker is shared.
- [ ] Provider API responses make provider-unavailable state clear even when historical connection rows exist.
- [ ] Settings and sidebar UI no longer show a provider as connected when its shared runtime is unconfigured or not running.
- [ ] Existing connected-user rows either transition to an intentional status or are rendered with an intentional “provider unavailable / reconnect” state.
- [ ] Tests cover failed configure rollback, multi-user runtime disconnect, provider response status, and frontend badge/button behavior.

## Validation plan

- Add backend tests with two users connected to the same provider, then remove runtime credentials as an admin.
- Add backend tests for `configure_channel_provider_runtime()` when `service.configure_channel()` returns false.
- Add frontend unit tests for connected row + unavailable provider combinations.
- Run the channel connection router tests and channel frontend tests.

## Blocked by

None - can start immediately.

ggnnggez:
Closes #3541

# Summary

- Treat runtime provider availability as authoritative when reporting connection status, even if a stale connected row still exists.
- Make runtime configure/disconnect flows transactional at the service boundary: worker start/stop succeeds before app state and runtime config are committed.
- Revoke active provider connection rows when a runtime provider is disconnected.
- Update frontend provider lists so connected status is hidden when the runtime provider is unavailable.

# Tests

- `uv run pytest tests/test_channel_connections_router.py -q`
- `uv run ruff check app/gateway/routers/channel_connections.py packages/harness/deerflow/persistence/channel_connections/sql.py tests/test_channel_connections_router.py`
- `pnpm --dir frontend lint`
- Pre-commit: `pnpm --dir frontend check` and backend lint

# Notes

This prevents the UI and backend from treating stale per-user bindings as active when the shared runtime provider is unavailable or has been removed.


willem-bd:
Reviewed at high effort. The core fix — making runtime provider state authoritative so a stale `connected` row no longer shows as connected when the provider is unavailable — is sound and well-tested at the happy-path level (the `_provider_response` override, the transactional configure rollback test, and the bulk-revoke test all check out). The frontend guards are also consistent (`provider-state.ts` is safe because `connectable` already encodes unavailability, and `getStatusLabel` short-circuits on `unavailable_reason`).

Three findings, ranked by severity:

## 1. Lost-update race on `app.state.channels_config` (regression)

`configure_channel_provider_runtime` (and `disconnect_channel_provider_runtime`) now do:

```python
candidate_channels_config = dict(channels_config)          # snapshot the shared cached dict
candidate_channels_config[provider] = runtime_config
...
started = await _restart_runtime_channel_if_available(...)  # multi-second await
...
request.app.state.channels_config = candidate_channels_config   # reassign AFTER the await
```

The **old** configure mutated the live dict in place *before* the await (`channels_config[provider] = runtime_config`), so two concurrent writes for *different* providers each set their own key on the shared object and both survived. The new snapshot → await → reassign pattern opens a lost-update window: two concurrent admin requests snapshot the same pre-update dict, and the second to finish reassigns its stale snapshot, clobbering the first.

**Scenario:** Admin A saves Slack credentials while Admin B concurrently saves Telegram credentials. Both read the same `app.state.channels_config`, each shallow-copies it, then both block on the channel-restart await. A finishes → `app.state = {slack:new, telegram:old}`. B finishes → `app.state = {telegram:new2}` (snapshotted before A's write landed, so Slack is gone). Since `_get_channels_config` returns the cached dict forever once set, subsequent `GET /providers` reads report Slack as unconfigured/unavailable even though the worker is running and `runtime-config.json` was persisted — it stays stale until process restart.

Suggested fix: re-read the live `app.state.channels_config` immediately before mutating/reassigning once the await succeeds (so the read-modify-write window contains no await), or serialize these mutations with a lock.

## 2. disconnect commits store + app.state before revoking DB rows

The new ordering is: stop worker → `store.set_provider_disconnected` (persisted to disk) → `app.state.channels_config = candidate` → `repo.disconnect_provider_connections(...)`. If the repo call raises, the store/cache say disconnected and the worker is stopped, but the DB rows stay `connected` and the endpoint returns 500.

**Scenario:** Admin disconnects Slack; the worker stops, `runtime-config.json` is written disconnected, `app.state` drops Slack, then `disconnect_provider_connections` raises (transient DB error). DB still has `status='connected'` rows. The new `_provider_response` masks this in the UI (shows `not_connected` because the provider is unavailable), but if an admin later re-configures Slack, those un-revoked `connected` rows silently go live again — the revocation the admin intended never actually happened. Worth wrapping the commit steps so a repo failure doesn't leave store/cache and DB divergent.

## 3. `_provider_response` masks a `revoked` row as `not_connected`

```python
if unavailable_reason is not None:
    connection_status = "not_connected"
elif connection:
    connection_status = connection["status"]
```

A legitimately revoked connection (`status='revoked'`) whose provider is *also* currently unavailable is now reported as `not_connected` instead of `revoked` — an undocumented contract change to `connection_status`, with no test asserting revoked is preserved when unavailable.

**Scenario:** user's Slack binding was revoked and the Slack runtime is not running (`unavailable_reason` set). Old code returned `revoked`; new code returns `not_connected`. Any consumer that distinguishes `revoked` (re-authorize / audit) from `not_connected` (fresh connect) on this field gets the wrong value. Currently masked because the frontend short-circuits to the unavailable label via `unavailable_reason` before reading status — but it's worth either documenting the new semantics or preserving `revoked`.

---

Finding #1 is the one I'd want addressed before merge; #2 and #3 are lower-severity robustness/contract gaps.

ggnnggez:
Thanks for the detailed review @willem-bd — all three findings are addressed in 4ea1bd22.

**#1 Lost-update race on `app.state.channels_config`** — `configure_channel_provider_runtime` and `disconnect_channel_provider_runtime` now re-read the live `app.state.channels_config` after the worker await and mutate only the affected provider key in place (no await between the read and the reassignment), so a concurrent mutation for a different provider is no longer clobbered by a stale pre-await snapshot.

**#2 disconnect commits store + app.state before revoking DB rows** — reordered so `repo.disconnect_provider_connections` runs *before* the store/cache commit. If the repo call raises, the store/cache stay consistent with the DB (provider remains configured), so a later re-configure can't silently reactivate un-revoked rows.

**#3 `_provider_response` masks `revoked` as `not_connected`** — when the provider is unavailable, only a stale `connected` row is downgraded to `not_connected`; other statuses such as `revoked` are now preserved.

Each fix has a dedicated regression test in `tests/test_channel_connections_router.py`:
- `test_configure_provider_runtime_does_not_clobber_concurrent_config_update`
- `test_disconnect_provider_runtime_keeps_state_consistent_when_revoke_fails`
- `test_get_providers_preserves_revoked_status_when_provider_unavailable`

WillemJiang (approved):

PR #3581, opened 2026-06-14
fix(channels): harden runtime credential management APIs

### LINKED ISSUE #3542: [IM] Harden runtime bot credential storage and channel management APIs
## Context

The browser-facing channel settings allow admins to enter runtime bot credentials for Telegram, Slack, Discord, Feishu/Lark, DingTalk, WeChat, and WeCom. These credentials are shared deployment-level secrets used to start provider workers. The code also has a future-looking `channel_credentials` table with encryption support for per-connection OAuth-style credentials, but the current runtime credential path is separate.

Current behavior observed in the IM audit:

- Runtime provider credentials are persisted in a local JSON file via `ChannelRuntimeConfigStore`.
- The file is chmodded to `0600` after replacement, but the values themselves are plaintext.
- WeChat QR login auth state can persist a raw `bot_token` to its state file.
- Browser responses mask password fields, and masked submissions preserve existing secrets, which is good and should remain.
- The legacy channel restart endpoint mutates shared channel worker state but does not apply the same admin guard used by runtime credential endpoints.

Relevant code paths:

- `backend/app/channels/runtime_config_store.py`
- `backend/app/channels/wechat.py::_save_auth_state()`
- `backend/app/gateway/routers/channel_connections.py::_credential_values()` and `_required_runtime_values()`
- `backend/app/gateway/routers/channels.py::restart_channel()`
- `backend/packages/harness/deerflow/persistence/channel_connections/sql.py::ChannelCredentialCipher`

## Problem

Runtime bot credentials are sensitive deployment secrets. Plaintext local storage may be acceptable for a local-only mode, but the current behavior is easy to misread as the same encrypted credential model used by the repository’s per-connection credential table. That creates security and documentation risk.

There is also an authorization consistency issue: runtime credential endpoints require admin privileges, but legacy mutating channel management endpoints can restart shared provider workers without an equivalent admin check.

## What to build

Harden the runtime credential lifecycle and align channel management permissions.

The desired result should be explicit and secure by default:

- Runtime credentials should be encrypted at rest, stored in a platform secret/keychain mechanism, or very clearly documented as local-development-only plaintext storage.
- WeChat QR-derived tokens should follow the same storage guarantees as other runtime credentials.
- Mutating channel management endpoints should require admin privileges.
- Docs should distinguish runtime bot credentials from per-user/per-connection credentials.

## Implementation notes

- Reuse or adapt existing encryption primitives only if key management is clear. A cipher without a stable configured key is not sufficient by itself.
- Consider environment-backed or OS-keychain-backed secret storage for local desktop deployments.
- If plaintext JSON remains as a supported fallback, write temp files with restrictive permissions before replace, keep final chmod, and document the fallback prominently.
- Keep existing masked-secret edit behavior: submitting `********` should preserve the stored secret.
- Decide whether read-only channel status endpoints need role restrictions or just mutating endpoints such as restart.
- Update docs/tests so future maintainers do not conflate runtime config storage with `channel_credentials` encryption.

## Acceptance criteria

- [ ] Runtime bot credentials are encrypted at rest or stored through a documented secret backend by default for non-local deployments.
- [ ] Plaintext local fallback, if retained, is explicitly documented with its threat model and file permission behavior.
- [ ] WeChat QR-derived `bot_token` storage follows the same protection model as other runtime credentials.
- [ ] Existing frontend masked-secret editing still preserves secrets without returning raw password values to the browser.
- [ ] Mutating legacy channel management endpoints require admin authorization.
- [ ] `backend/docs/IM_CHANNEL_CONNECTIONS.md` accurately describes runtime credentials vs per-connection credentials.
- [ ] Tests cover secret persistence, masked edit preservation, WeChat token storage behavior, and admin access checks.

## Validation plan

- Add unit tests for the runtime config store’s secure persistence mode and plaintext fallback, depending on the chosen design.
- Add router tests for non-admin access to mutating channel management endpoints.
- Add or update docs tests/snapshots if applicable.
- Manually verify that provider configuration still works through the settings UI with masked values.

## Blocked by

None - can start immediately, though the exact secret backend may need maintainer preference.

ggnnggez:
Closes #3542

# Summary

- Require admin privileges for the mutable channel runtime restart API while keeping read-only status available.
- Persist local runtime credential JSON and WeChat auth state with owner-only file permissions.
- Document runtime provider credentials, per-connection credentials, local plaintext fallback behavior, and admin-only mutation rules.
- Add router coverage for restart authorization and status read-only access.

# Tests

- `uv run pytest tests/test_channels_router.py tests/blocking_io/test_channel_runtime_config_store.py tests/test_wechat_channel.py tests/test_channel_connections_router.py -q`
- `uv run ruff check app/gateway/routers/channels.py app/channels/runtime_config_store.py app/channels/wechat.py tests/test_channels_router.py tests/blocking_io/test_channel_runtime_config_store.py tests/test_wechat_channel.py`
- Pre-commit backend lint

# Notes

This is a local hardening step. It does not introduce a new external secret manager, but it narrows API write access and prevents world/group-readable runtime secret files.


backend/app/channels/wechat.py
```diff
@@ -1383,6 +1383,7 @@ def _save_auth_state(
             try:
                 self._auth_path.parent.mkdir(parents=True, exist_ok=True)
                 self._auth_path.write_text(json.dumps(data, ensure_ascii=False, indent=2), encoding="utf-8")
+                self._auth_path.chmod(0o600)
```
willem-bd: **Race window between `write_text` and `chmod`.**

`Path.write_text(...)` opens the file with the user's umask (typically 0o644 on Linux/macOS), writes the JSON containing the iLink `bot_token`, and only then does the next line tighten it to 0o600. On a multi-user host, a concurrent local reader (cron job, sibling unprivileged user, monitoring agent) can open `wechat-auth.json` during that window and capture the secret.

The sibling change in `runtime_config_store.py` already does this safely: write to a `NamedTemporaryFile` (mkstemp gives 0o600 at create time), then `Path.replace`. Mirror that pattern here, or use `os.open(path, O_WRONLY|O_CREAT|O_TRUNC, 0o600)` so the mode is set atomically at creation rather than after the secret is on disk.

backend/app/channels/wechat.py
```diff
@@ -1383,6 +1383,7 @@ def _save_auth_state(
...
             except OSError:
                 logger.warning("[WeChat] failed to persist auth state to %s", self._auth_path)
         return data
```
willem-bd, line 1403: **Misleading log message after the new chmod.**

With the chmod inside the same `try/except OSError`, a chmod failure on a filesystem that doesn't support POSIX permissions (vfat/exFAT, certain NFS mounts, Windows shares) logs `[WeChat] failed to persist auth state to <path>` even though `write_text` succeeded and the JSON is on disk. Operators chase a missing file that's actually present.

Split the chmod into its own try/except so the persist-vs-harden errors don't collide:

```python
try:
    self._auth_path.parent.mkdir(parents=True, exist_ok=True)
    self._auth_path.write_text(...)
except OSError:
    logger.warning("[WeChat] failed to persist auth state to %s", self._auth_path)
else:
    try:
        self._auth_path.chmod(0o600)
    except OSError:
        logger.debug("[WeChat] unable to chmod auth state at %s", self._auth_path, exc_info=True)
```

backend/app/gateway/routers/channels.py
```diff
@@ -22,6 +22,17 @@ class ChannelRestartResponse(BaseModel):
     message: str
 
 
+async def _require_admin_user(request: Request) -> None:
```
willem-bd: **Third near-identical copy of `_require_admin_user`.**

The same 7-line helper now exists in three routers — only the 403 detail string differs:

- `routers/mcp.py:75` — "Admin privileges required to manage MCP configuration."
- `routers/channel_connections.py:138` — "Admin privileges required to manage channel runtime credentials."
- `routers/channels.py:25` (this PR) — "Admin privileges required to manage channel runtime workers."

A future change to the admin definition (e.g. allowing `INTERNAL_SYSTEM_ROLE` for trusted Gateway-internal calls, adding audit logging, switching to a permission-based check) has to land in three places, and drift between the copies will silently make some routes more or less restrictive than the others.

Extract a single helper into `app/gateway/deps.py` (or a dedicated `auth_helpers.py`):

```python
async def require_admin_user(request: Request, *, detail: str) -> None:
    user = getattr(request.state, "user", None)
    if user is None:
        user = await get_current_user_from_request(request)
    if getattr(user, "system_role", None) != "admin":
        raise HTTPException(status_code=403, detail=detail)
```

and have all three routers depend on it.

backend/tests/blocking_io/test_channel_runtime_config_store.py
```diff
@@ -104,3 +107,17 @@ async def test_disconnect_runtime_channel_does_not_block_event_loop(tmp_path) ->
...
+    store = await asyncio.to_thread(ChannelRuntimeConfigStore, path)
+
+    await asyncio.to_thread(
+        store.set_provider_config,
+        "slack",
+        {"enabled": True, "bot_token": "xoxb-ui", "app_token": "xapp-ui"},
+    )
+
+    mode = await asyncio.to_thread(lambda: path.stat().st_mode & 0o777)
+    assert mode == 0o600
```
willem-bd, line 121: **This test does not actually verify the new pre-write chmod.**

On POSIX, `tempfile.NamedTemporaryFile` uses `mkstemp` under the hood and creates the file at 0o600 from the start. `Path.replace` preserves the source's permissions, and the post-replace `self._path.chmod(0o600)` was already in the file before this PR. So `path.stat().st_mode & 0o777 == 0o600` holds even if the new pre-`json.dump` chmod (the actual code added by this diff) is removed — meaning a regression that drops it would not be caught here.

To actually exercise the new chmod, either:

- pre-create `path` at 0o644 and assert it is tightened to 0o600 after a save, or
- inject an `os.chmod` failure on the temp-file path and assert it is logged at debug while the post-replace chmod still secures the destination.

As written, the assertion is a property of `mkstemp`, not of this PR.

backend/tests/blocking_io/test_channel_runtime_config_store.py
```diff
@@ -66,6 +66,9 @@ class _FakeRepo:
     async def list_connections(self, owner_user_id):
         return []
 
+    async def disconnect_provider_connections(self, *, provider):
```
willem-bd: **`disconnect_provider_connections` stub is dead code.**

`grep -rn disconnect_provider_connections backend/` returns no hits — neither `ChannelConnectionRepository` nor any router exposes a method by this name. `disconnect_channel_provider_runtime` calls `repo.disconnect_connection(...)` (singular) per row in `repo.list_connections(...)`.

The stub is never invoked, so it neither documents an interface nor protects against a regression. Either remove it, or — if it's anticipating a planned refactor that consolidates the per-connection loop into one repo call — land that refactor in this PR so the method has a real caller.

willem-bd (commented):
Five findings on the credential-hardening pass. The diff is strictly an improvement, but a few sharp edges are worth tightening:

1. **WeChat: write-then-chmod race window** — bot_token is briefly readable at umask defaults before chmod restricts it; the runtime_config_store path uses the safer tempfile+rename pattern and wechat.py should mirror it.
2. **Three near-identical `_require_admin_user` helpers** (mcp.py, channel_connections.py, and now channels.py) — extract one shared dependency before adding a fourth.
3. **Misleading `failed to persist` log** when only the chmod fails on filesystems without POSIX perms.
4. **The new `test_runtime_config_store_file_is_owner_only` does not exercise the new chmod** — `tempfile.NamedTemporaryFile` already yields 0o600 on POSIX, so the test would pass even if the diff were dropped.
5. **Dead `_FakeRepo.disconnect_provider_connections` stub** — no production code path calls a method by that name.

Details inline.

backend/app/channels/wechat.py
ggnnggez: Fixed in `45ab711a`. `_save_auth_state` now writes through a 0o600 `NamedTemporaryFile` (mkstemp creates it owner-only) and `Path.replace`s it into place, mirroring `ChannelRuntimeConfigStore._save`, so the iLink `bot_token` is never on disk at umask defaults. Added `test_save_auth_state_tightens_preexisting_loose_file`, which seeds the destination at 0o644 and asserts the post-save file is 0o600 with no `.tmp` residue.

backend/app/gateway/routers/channels.py
ggnnggez: Fixed in `45ab711a`. Extracted `require_admin_user(request, *, detail)` into `app/gateway/deps.py`; `mcp.py`, `channel_connections.py`, and `channels.py` each pass their own `_ADMIN_REQUIRED_DETAIL` to the shared helper, and the three local copies are gone.

backend/tests/blocking_io/test_channel_runtime_config_store.py
ggnnggez: Strengthened in `45ab711a`. Added `test_runtime_config_store_chmod_failure_is_logged_not_fatal`, which patches the temp-file chmod to raise and asserts the debug log fires while the destination is still 0o600 — mutation-verified to fail if the pre-write chmod is dropped. Also added `test_runtime_config_store_overwrites_loose_existing_file` (0o644 -> 0o600). Kept the original owner-only assertion as a cheap property check.

backend/tests/blocking_io/test_channel_runtime_config_store.py
ggnnggez: Removed in `45ab711a` — with a correction to the reasoning: `disconnect_provider_connections` is live production code (`deerflow/persistence/channel_connections/sql.py:180`) and is called by the handler under test (`channel_connections.py:578`). It was nonetheless dead *in this test*: `_get_repository`'s `isinstance(repo, ChannelConnectionRepository)` gate rejects the plain fake, so the handler takes the `get_session_factory()`->None->503 path with `repo=None` and line 578 is never reached. Emptying `_FakeRepo` entirely left all three tests passing, so `list_connections` was dead too — I removed the whole unused `_FakeRepo` and its `app.state` assignment rather than just the one method.

backend/app/channels/wechat.py
ggnnggez: Fixed in `45ab711a`. The chmod moved into its own `else:` branch after the write succeeds; a chmod `OSError` now logs at debug (`unable to chmod auth state`) instead of the `failed to persist` warning. Covered by `test_save_auth_state_chmod_failure_is_logged_not_warned`, which injects a chmod failure and asserts the debug record is present and the persist warning is absent.

WillemJiang (approved):

PR #3582, opened 2026-06-14
fix(channels): make channel connect flow deterministic

### LINKED ISSUE #3543: [IM] Fix channel connect flow edge cases and stale UI state
## Context

The intended user flow is straightforward:

1. User opens DeerFlow channel settings.
2. User clicks Connect for a provider.
3. DeerFlow creates a short-lived binding code.
4. User sends the code to the provider bot, or follows a Telegram deep link.
5. DeerFlow records the provider identity as connected to the browser-authenticated DeerFlow user.
6. UI reflects the connected state.

The current implementation has several edge cases that can break or obscure this flow:

- Some provider adapters check legacy `allowed_users` before processing connect codes, so a valid code can be ignored if the external provider user is not pre-allowlisted.
- The frontend invalidates provider queries when a code is generated, but it does not poll for the later provider-side code consumption event.
- The data model allows the same external provider identity/workspace to be bound by multiple DeerFlow owners; event lookup returns the most recently updated connected row.
- The UI collapses connections by provider, which can hide multi-workspace/multi-account cases.

Relevant code paths:

- `backend/app/channels/telegram.py::_cmd_start()`
- `backend/app/channels/slack.py::_handle_message_event()`
- `backend/app/channels/dingtalk.py::_on_chatbot_message()`
- `backend/app/channels/wechat.py::_handle_update()`
- `backend/packages/harness/deerflow/persistence/channel_connections/model.py`
- `backend/packages/harness/deerflow/persistence/channel_connections/sql.py::find_connection_by_external_identity()`
- `frontend/src/components/workspace/channels/workspace-channels-list.tsx`
- `frontend/src/components/workspace/settings/channels-settings-page.tsx`

## Problem

A user can do everything the UI asks and still end up in a confusing state:

- The provider may silently ignore their valid `/connect <code>` because they are not in a legacy allowlist.
- The backend may bind the external identity, but the UI may continue showing stale state until manual refresh.
- If the same provider identity is bound to another DeerFlow user, messages may route to whichever connection was updated last.
- If a provider supports multiple workspaces/accounts, the single provider-level UI state may not represent the real connection set.

This makes channel setup feel unreliable and can create real routing ambiguity.

## What to build

Make the connect flow deterministic and observable from click to connected state.

The final behavior should answer these questions clearly:

- Are connect codes allowed to bypass legacy `allowed_users` checks, or is the allowlist intentionally part of binding policy?
- How does the frontend learn that a provider-side code was consumed?
- Can the same external identity be connected to multiple DeerFlow owners? If not, how is transfer/rebind handled?
- Does the UI support multiple connections per provider, or does the backend enforce one active connection per provider per user?

## Implementation notes

- Process valid connect-code messages before ordinary allowlist rejection, unless maintainers decide allowlists are meant to restrict who may bind.
- Add frontend polling after `connect` response, using `expires_in` as the stop condition. Poll both providers and connections, and show timeout/expiry feedback.
- Consider a repository-level uniqueness rule for active `(provider, external_account_id, workspace_id)` across owners, or explicitly revoke/transfer older bindings on successful bind.
- If multiple provider connections remain supported, update the settings UI to render the actual connection list instead of collapsing too aggressively by provider.
- Keep Telegram deep-link behavior smooth: opening the link should not leave a blank/stale state if the user returns to DeerFlow.

## Acceptance criteria

- [ ] A valid connect code can be consumed successfully under the intended allowlist policy, and the policy is documented.
- [ ] Frontend settings/sidebar update automatically after provider-side binding succeeds, without requiring manual refresh.
- [ ] Expired or unused connect codes stop polling and show actionable feedback.
- [ ] Duplicate external identity binding is handled deterministically: prevented, transferred, or explicitly supported with documented routing semantics.
- [ ] UI behavior for multiple provider workspaces/accounts is intentional and covered by tests.
- [ ] Tests cover allowlist-plus-connect-code, post-connect polling, duplicate external identity routing, and multi-connection provider display.

## Validation plan

- Add provider tests for at least Telegram and one `/connect <code>` provider with `allowed_users` configured.
- Add repository tests for duplicate external identity behavior.
- Add frontend tests for polling success and polling expiry.
- Run targeted backend channel connection tests and frontend channel unit tests.

## Blocked by

None - can start immediately.

ggnnggez:
Closes #3543

# Summary

- Let Slack and DingTalk consume valid `/connect <code>` messages before applying `allowed_users`, matching the intended first-bind flow.
- Make duplicate external identity binding deterministic by revoking other active owners for the same provider/workspace/account when a new owner connects.
- Preserve the existing concurrent insert retry behavior with `session.no_autoflush`.
- Add bounded frontend polling after connect-code creation so the settings UI observes successful binding without manual refresh.

# Tests

- `uv run pytest tests/test_channel_connections_repository.py tests/test_channels.py tests/test_dingtalk_channel.py -q`
- `uv run ruff check app/channels/slack.py app/channels/dingtalk.py packages/harness/deerflow/persistence/channel_connections/sql.py tests/test_channels.py tests/test_dingtalk_channel.py tests/test_channel_connections_repository.py`
- `pnpm --dir frontend check`
- Pre-commit frontend check and backend lint

# Notes

This makes connection ownership transfer explicit: the latest successful bind becomes active, and previous active owner rows for the same external identity are revoked.


willem-bd (commented):
Thanks @ggnnggez. I traced the diff against the touched code paths and have a few items to consider before this is ready. Marking severities P1/P2 by impact; the first two are the ones I'd most like to see addressed.

---

### [P1] The connect-code-before-`allowed_users` fix is incomplete

- **Location:** `backend/app/channels/telegram.py:466` and `backend/app/channels/wechat.py:606`
- **Problem:** This PR fixes the bind-bootstrap UX bug only in `slack.py` and `dingtalk.py`. Two other adapters still gate `_check_user`/`allowed_users` *before* connect-code dispatch, so a new operator-allowlisted user trying to bind via the browser flow on Telegram or WeChat is silently rejected — the same bug the PR is fixing.
- **Evidence:**
  - `telegram.py:466` — `_cmd_start` returns on `not self._check_user(update.effective_user.id)` *before* `_bind_connection_from_start_token` at line 470.
  - `wechat.py:606` — `_handle_update` returns on `not self._check_user(chat_id)` *before* `extract_connect_code` at line 622 and `_bind_connection_from_connect_code` at line 624.
  - Feishu and WeCom appear unaffected (no `_check_user` gate in their inbound paths).
- **Suggested fix:** Apply the same reorder to telegram and wechat. Better still, factor the connect-code dispatch into a small helper on `Channel` (or the channel manager) that runs *before* per-adapter authorization, so future adapters inherit the right ordering. The presence of identical fixes in two adapters is the signal that this is base-class behavior.
- **Test:** Add `test_connect_code_bypasses_allowed_users_filter` for telegram and wechat mirroring the slack/dingtalk tests added in this PR.

---

### [P1] `_revoke_other_active_owners` does not enforce the single-active-owner invariant under concurrency

- **Location:** `backend/packages/harness/deerflow/persistence/channel_connections/sql.py:131-146`
- **Problem:** The new helper uses `SELECT-then-UPDATE` in app code with no row lock or partial unique constraint covering active rows. Two concurrent connect-code consumes for the same `(provider, external_account_id, workspace_id)` from *different* owners can both observe \"no other active owner\" and both commit `connected` rows.
- **Evidence:** The existing schema constraint `uq_channel_connection_owner_provider_identity` in `model.py:41-46` is on `(owner_user_id, provider, external_account_id, workspace_id)` — it includes `owner_user_id`, so it does **not** prevent two owners from both holding `connected` rows for the same external identity. Under PostgreSQL `READ COMMITTED` (the default), T1 and T2 can each `SELECT` and miss the other's uncommitted row, then both `COMMIT`. `find_connection_by_external_identity` then resolves nondeterministically.
- **Suggested fix:** Move the invariant to the database layer with a partial unique index — e.g. `CREATE UNIQUE INDEX uq_channel_connection_active_identity ON channel_connections (provider, external_account_id, workspace_id) WHERE status != 'revoked'` (PostgreSQL syntax; for SQLite use a partial index or a `CHECK`-equivalent) — and use `INSERT ... ON CONFLICT DO UPDATE` (or `SELECT ... FOR UPDATE` on a sentinel row) so concurrent transfers serialize. Keep `_revoke_other_active_owners` as the bookkeeping step *after* the lock is held.
- **Test:** Concurrent-upsert test using two `asyncio.create_task` upsert_connection calls with overlapping `(provider, external_account_id, workspace_id)` and different owners; assert exactly one row ends `status=\"connected\"`.

---

### [P2] Connect-status poll loop has no cancellation or per-provider dedup

- **Location:** `frontend/src/core/channels/hooks.ts:14-62`
- **Problem:** `pollChannelConnectionUntilResolved` is a recursive `window.setTimeout` chain with no `AbortController`, no `useEffect` cleanup, no per-provider active-poll handle. Every successful `connect` mutation starts a fresh chain that keeps fetching against the shared `QueryClient` until `expires_in` elapses, regardless of whether the calling component unmounted, the user navigated away, or the user clicked Connect again.
- **Evidence:** The poll loop is started from `onSuccess` in `useConnectChannelProvider`; both call sites (`workspace-channels-list.tsx` and `channels-settings-page.tsx`) use the hook normally with no external cleanup. Repeated clicks → multiple parallel chains racing on `channelProviderQueryKey`/`channelConnectionsQueryKey`.
- **Suggested fix:** Thread an `AbortController` (or `AbortSignal`) through the poll loop. Store one active controller per `provider` in a `useRef<Map<ChannelProviderId, AbortController>>`. On a new `connect` for the same provider, abort the existing controller before starting the new one. Abort all controllers in a `useEffect` cleanup. Alternatively: drive polling via React Query's `refetchInterval` keyed on a `pendingConnect` state — that gives you cancellation via query invalidation for free.
- **Test:** RTL test that mounts the hook, triggers `connect`, unmounts, and asserts no further fetches fire.

---

### [P2] `[DingTalk] parsed message` INFO log now leaks blocked-user message content

- **Location:** `backend/app/channels/dingtalk.py:374-381`
- **Problem:** Moving the `allowed_users` gate past `_extract_text` means the parsed-message INFO log (`text=%r`, first 100 chars) now fires for senders who would have been rejected before. Operators relying on `allowed_users` as a noise / privacy filter will see those messages in INFO-level log shipping.
- **Evidence:** Base version returned at the `allowed_users` check before `_extract_text`. PR head emits `logger.info(\"[DingTalk] parsed message: ..., text=%r\", ..., text[:100])` at line 374, then runs `extract_connect_code`, then the new `allowed_users` gate at line 401.
- **Suggested fix:** Move the parsed-message INFO log to *after* the `allowed_users` gate at line 401, or drop it to DEBUG. The Slack diff doesn't have this issue because its only blocked-user log (`debug` level, no message content) sits past the new gate.
- **Test:** Extend `test_allowed_users_blocks_chat_message` to also assert the parsed-message INFO log is not emitted for blocked senders.

---

### [P2] Connect-status poll fires both `listChannelProviders` and `listChannelConnections` per tick

- **Location:** `frontend/src/core/channels/hooks.ts:21-32`
- **Problem:** Each 2-second tick runs `Promise.all([listChannelProviders(), listChannelConnections()])`, but either response alone is sufficient to detect `connected` — the gateway computes `ChannelProviderResponse.connection_status` from the same per-user connection row that `listChannelConnections()` returns. Doubles backend load for every polling user.
- **Evidence:** `expires_in ≈ 600 s` and `CONNECT_POLL_INTERVAL_MS = 2000` ⇒ up to ~300 ticks × 2 endpoints = ~600 backend requests for one uncompleted bind. With multiple concurrent users, the duplication multiplies.
- **Suggested fix:** Poll only `listChannelConnections()`, then `queryClient.invalidateQueries({ queryKey: channelProviderQueryKey })` once after the connection flips to `connected` (so the providers UI refetches exactly once when the bind lands).
- **Test:** Existing tests still pass; add a fetch-count assertion that during the poll loop only the connections endpoint is hit.

---

### [P2] `allowed_users` is no longer a defense against connect-code leakage — worth documenting

- **Location:** `backend/app/channels/slack.py:321` / `backend/app/channels/dingtalk.py:407`
- **Problem:** This is the PR's documented intent (`# Check allowed users after connect-code handling so browser-initiated binding can bootstrap a new external identity`), and the safeguards are real: `secrets.token_urlsafe(16)` codes, 600 s TTL, one-time `consume_oauth_state`, codes shown only in browser `toast.success(result.instruction)` rather than echoed in any group chat. The trade-off worth surfacing: any sender who *possesses* a valid code can now consume it, not only the legitimate user. Pre-PR, even a leaked code couldn't be consumed by a non-allowlisted Slack/DingTalk user.
- **Suggested fix:** No code change required. Add a short paragraph to the PR description (and to `backend/CLAUDE.md`'s channel-connections section) explicitly stating that `allowed_users` is no longer a bind-time defense and that the security model relies on the code's confidentiality + 600 s TTL + one-time consumption. Operators should be told.
- **Test:** N/A.

---

### [P2] `Math.max(1, expiresInSeconds) * 1000` does not defend against NaN

- **Location:** `frontend/src/core/channels/hooks.ts:19`
- **Problem:** `result.expires_in` is typed as `number` in `core/channels/types.ts` but parsed via `response.json() as Promise<ChannelConnectResponse>` in `core/channels/api.ts` with no runtime validator. If a future backend or proxy ever drops/garbles the field, `expires_in` is `undefined` ⇒ `Math.max(1, undefined) === NaN` ⇒ `deadline === NaN` ⇒ `Date.now() >= NaN` is permanently false ⇒ the poll loop runs until the page is closed.
- **Suggested fix:** `const expires = Number.isFinite(expiresInSeconds) && expiresInSeconds > 0 ? Math.max(1, expiresInSeconds) : DEFAULT_EXPIRES_IN_S;` before computing the deadline. Or — preferred — add a zod (or comparable) validator at the API boundary so unchecked drift can't reach the hook.
- **Test:** Hook test where `connectChannelProvider` resolves with `expires_in: undefined` (or `NaN`); assert the poll loop terminates within a sane bound.

---

### [P2] Behavior change is operator-relevant; docs not updated

- **Location:** PR-wide
- **Problem:** `backend/CLAUDE.md:70` states `**CRITICAL: Always update README.md and CLAUDE.md after every code change**`, with the surrounding rule scoping `README.md` to user-facing changes and `CLAUDE.md` to development / internal-system changes. This PR ships two operator-relevant semantic changes — `allowed_users` no longer guards Slack/DingTalk bind, and `upsert_connection` now transfers ownership across users — that are not reflected in either.
- **Suggested fix:** Add a short note to `backend/CLAUDE.md`'s channel-connections section covering (a) the connect-code-vs-allowlist ordering and (b) the single-active-owner-per-external-identity transfer semantics. Optionally add a one-line entry to the changelog so operators see it without grepping.
- **Test:** N/A.

---

The connect-flow ordering change itself is the right idea — the suggestions above are about generalizing it to the other adapters, hardening the persistence-layer transfer against concurrent connects, and tightening the new frontend polling primitive.

ggnnggez:
Thanks for the thorough review, @willem-bd — all eight items are addressed, plus the base-class generalization you suggested under P1. Summary of the pushed changes:

### P1
- **Connect-code-before-`allowed_users` incomplete (Telegram/WeChat)** — `c9ac94ca`. Telegram consumes the `/start` deep-link token before the `_check_user` gate; WeChat consumes the `/connect` code before the gate and now defers inbound file extraction + context-token tracking past the gate so blocked senders don't trigger CDN downloads. Regression tests added for both (`test_start_token_bypasses_allowed_users_filter`, `test_connect_code_bypasses_allowed_users_filter`).
- **Generalized to a base-class helper** — `f8d16a49`. Lifted the ordering-sensitive dispatch into `Channel._pending_connect_code(text)` (guards the repo + extracts the code, documenting that adapters must call it *before* authorization), routed all six `/connect` adapters through it, and moved `_connection_repo` onto the base. This also fixes a latent inconsistency where Slack dispatched a bind even with no repo configured.
- **`_revoke_other_active_owners` concurrency** — `d9bdc75a`. Added a partial unique index `uq_channel_connection_active_identity` on `(provider, external_account_id, workspace_id) WHERE status != 'revoked'` (portable to SQLite ≥3.8 and PostgreSQL), so the DB guarantees at most one non-revoked row per identity. `upsert_connection` now revokes other owners *before* the connected row is flushed, wrapped in a bounded rollback-and-retry loop — a losing concurrent writer retries against the now-visible state instead of committing a duplicate. Added a DB-constraint test plus revoked-slot-reuse and concurrent-upsert tests.

### P2
- **Poll loop cancellation / dedup** — `1361f25c`. Extracted a framework-agnostic, cancellable poller (`connect-poll.ts`); `useConnectChannelProvider` tracks one handle per provider in a ref Map (a new connect cancels the prior poll) and a `useEffect` cleanup cancels all on unmount.
- **Poll double-fetch per tick** — `1361f25c`. The poller hits only `listChannelConnections()` and invalidates the providers query once when the bind resolves.
- **`Math.max(1, expiresInSeconds)` NaN** — `1361f25c`. `expires_in` is guarded with `Number.isFinite(...) && > 0`, falling back to a default window, so a non-finite value can no longer produce an unbounded loop. Unit tests cover resolve-and-stop, cancellation, and non-finite expiry.
- **DingTalk INFO log leaks blocked-sender content** — `df1df9a6`. The parsed-message log moved to after the `allowed_users` gate; `test_non_allowed_user_message_content_not_logged` asserts blocked senders' text never reaches INFO.
- **`allowed_users` is no longer a bind-time defense / docs** — `df1df9a6` + `77785da7`. Documented in both `backend/CLAUDE.md` and the operator-facing `backend/docs/IM_CHANNEL_CONNECTIONS.md`: connect codes are consumed before the allowlist (so the security model rests on code confidentiality — 128-bit, 600s TTL, one-time consume), and the single-active-owner last-bind-wins transfer semantics.

### A few notes for the record
1. The concurrent-upsert test is an end-to-end invariant guard; the *deterministic* regression for the fix is the DB-constraint test (two `connected` rows for one identity → `IntegrityError`), since SQLite WAL serializes writers. The real race protection on PostgreSQL is the partial unique index.
2. `cancel()` stops scheduling and discards the in-flight result but does not abort the in-flight request — I kept it simple given the 2s interval; happy to thread an `AbortSignal` if you'd prefer.
3. Migration caveat: the schema is created via `create_all` (no Alembic), so an existing deployment that already accumulated duplicate active rows from the old behavior would fail to build the unique index on startup. Since this transfer behavior ships in this PR there's no deployed data depending on it, but worth flagging for operators.

WillemJiang (approved):

PR #3584, opened 2026-06-14
fix(channels): add operational guardrails

### LINKED ISSUE #3544: [IM] Add inbound dedupe, rate limits, and safer channel observability
## Context

The IM channel layer integrates several providers with different delivery semantics: long polling, socket mode, bot SDK callbacks, and local runtime workers. Provider delivery can retry, replay, rate-limit, or fail in provider-specific ways. The current implementation already handles many happy paths, but the audit found several resilience and observability gaps that should be addressed together.

Examples from the current codebase:

- Inbound provider events generally do not have a shared dedupe/idempotency layer.
- Binding code creation is cryptographically strong and single-use, but there is no apparent per-user/provider rate limit or creation cap.
- Slack has an `event_delivery="http"` mode that can start without a Socket Mode app token, but there is no corresponding Gateway webhook receiver/signature verification path.
- Slack outbound retries use generic exponential retry behavior rather than provider-aware `Retry-After` handling.
- Normal info logs include message text snippets in several channel paths, which can capture PII or `/connect` codes.
- Discord file upload opens a local file handle without an explicit close path.
- Legacy `ChannelStore` serializes keys with colon-delimited strings, which can parse incorrectly if provider IDs ever contain separator characters.

Relevant code paths:

- `backend/app/channels/message_bus.py`
- `backend/app/channels/manager.py`
- `backend/app/channels/slack.py`
- `backend/app/channels/discord.py`
- `backend/app/channels/store.py`
- `backend/packages/harness/deerflow/persistence/channel_connections/sql.py`

## Problem

These are not all the same bug, but they share the same operational failure mode: under retries, abuse, provider rate limits, or unusual IDs, IM behavior can become duplicated, noisy, misleading, or resource-leaky.

Concrete risks:

- Duplicate provider deliveries can trigger duplicate DeerFlow runs.
- A logged-in user can create many pending binding states without a clear cap/cleanup strategy.
- Slack HTTP Events mode can appear configured/running while never receiving inbound events.
- Provider rate-limit responses may be retried inefficiently or fail noisily.
- Logs can retain user message content and binding codes.
- Long-lived workers can leak file descriptors during repeated Discord attachment sends.
- Legacy channel mapping inspection can misrepresent conversation IDs containing separators.

## What to build

Add a small set of channel-layer guardrails that make IM integrations safer under non-happy-path conditions.

The work can be implemented incrementally, but the completed slice should provide visible operational improvements:

- Idempotency/dedupe for inbound events where providers can retry/replay.
- Rate limiting or caps for connect-code creation.
- Explicit support or removal for unsupported delivery modes.
- Provider-aware outbound retry handling.
- Redacted logs for message content and binding codes.
- Resource cleanup for file send paths.
- Safer legacy store serialization/parsing.

## Implementation notes

- Dedupe can start with provider/message IDs and a bounded TTL cache or persisted idempotency table, depending on durability needs.
- Connect-code caps can use the existing OAuth state count helper, plus opportunistic cleanup of expired states on creation.
- Slack HTTP Events should either be fully implemented with signed request verification and event dispatch into the same channel path, or disabled so it cannot start in a misleading partial mode.
- Slack retry behavior should honor `Retry-After` for 429s and keep existing transient retry behavior for other errors.
- Log redaction should happen before normal info logs, especially for `/connect <code>` and user message text.
- File send paths should use context managers or provider APIs that take ownership of file objects safely.
- Legacy `ChannelStore` should use structured keys or escaped segments instead of ambiguous colon splitting.

## Acceptance criteria

- [ ] Inbound provider events that include stable provider message IDs are deduped so retries do not create duplicate runs.
- [ ] Connect-code creation is rate-limited or capped per user/provider, with expired-state cleanup that does not rely only on consumption.
- [ ] Slack HTTP Events delivery mode is either fully implemented with signed webhook verification and dispatch, or removed/disabled with clear configuration errors.
- [ ] Slack outbound retry behavior respects provider rate-limit responses such as `Retry-After`.
- [ ] Info-level logs no longer include raw user message text or binding codes by default.
- [ ] File upload/send paths close local file handles reliably.
- [ ] Legacy channel conversation storage safely handles provider IDs that contain separator characters.
- [ ] Tests cover dedupe/idempotency, connect-code caps, Slack delivery-mode behavior, log redaction, file-handle cleanup, and legacy key parsing.

## Validation plan

- Add provider-independent unit tests for dedupe and connect-code cap behavior.
- Add Slack tests for HTTP delivery-mode configuration and 429 retry handling.
- Add logging tests that assert binding codes are redacted.
- Add a Discord file-send test that verifies file handles are closed.
- Add `ChannelStore` tests with IDs containing separator-like characters.

## Blocked by

None - can start immediately.

ggnnggez:
Closes #3544

# Summary

- Add a bounded inbound dedupe cache in `ChannelManager` keyed by provider/workspace/chat/stable message ID.
- Add missing stable `message_id` metadata for Slack, Telegram, WeChat, and WeCom inbound messages.
- Cap pending connect-code issuance per owner/provider and clear expired OAuth state before counting active pending codes.
- Remove message text snippets from manager and provider parsing logs; log text length and metadata instead.
- Reject unsupported Slack `event_delivery="http"` mode clearly instead of starting a partially functional adapter.

# Tests

- `uv run pytest tests/test_channels.py tests/test_channel_connections_router.py tests/test_channel_connections_repository.py tests/test_slack_channel_connections.py -q`
- `uv run ruff check app/channels/manager.py app/channels/slack.py app/channels/telegram.py app/channels/wechat.py app/channels/wecom.py app/channels/feishu.py app/channels/dingtalk.py app/gateway/routers/channel_connections.py packages/harness/deerflow/persistence/channel_connections/sql.py tests/test_channels.py tests/test_channel_connections_router.py tests/test_channel_connections_repository.py tests/test_slack_channel_connections.py`
- Pre-commit backend lint

# Notes

The inbound dedupe intentionally only uses stable provider message IDs from metadata. It does not guess based on `thread_ts`, because threaded platforms can legitimately reuse thread/root IDs for multiple user messages.

### DIFF
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -62,6 +62,9 @@ MESSAGE_STREAM_EVENTS = ("messages-tuple", "messages")
 THREAD_BUSY_MESSAGE = "This conversation is already processing another request. Please wait for it to finish and try again."
 BOUND_IDENTITY_REQUIRED_MESSAGE = "Connect this channel from DeerFlow Settings, complete the in-channel connect step, then send your message again."
 BOUND_IDENTITY_UNAVAILABLE_MESSAGE = "Channel connection verification is temporarily unavailable. Please try again later or contact the DeerFlow operator."
+INBOUND_DEDUPE_TTL_SECONDS = 10 * 60
+INBOUND_DEDUPE_MAX_ENTRIES = 4096
+INBOUND_DEDUPE_METADATA_KEYS = ("event_id", "message_id", "msg_id", "client_msg_id", "client_id")
 
 CHANNEL_CAPABILITIES = {
     "dingtalk": {"supports_streaming": False},
@@ -774,6 +777,7 @@ class ChannelManager:
         self._semaphore: asyncio.Semaphore | None = None
         self._running = False
         self._task: asyncio.Task | None = None
+        self._recent_inbound_events: dict[tuple[str, str, str, str], float] = {}
 
     @staticmethod
     def _channel_supports_streaming(channel_name: str) -> bool:
@@ -920,14 +924,73 @@ class ChannelManager:
                 break
 
             logger.info(
-                "[Manager] received inbound: channel=%s, chat_id=%s, type=%s, text=%r",
+                "[Manager] received inbound: channel=%s, chat_id=%s, type=%s, text_len=%d, files=%d",
                 msg.channel_name,
                 msg.chat_id,
                 msg.msg_type.value,
-                msg.text[:100] if msg.text else "",
+                len(msg.text or ""),
+                len(msg.files),
             )
+            if self._is_duplicate_inbound(msg):
+                continue
             task = asyncio.create_task(self._handle_message(msg))
             task.add_done_callback(self._log_task_error)
+
+    @staticmethod
+    def _inbound_dedupe_key(msg: InboundMessage) -> tuple[str, str, str, str] | None:
+        metadata = msg.metadata or {}
+        message_id = None
+        for key in INBOUND_DEDUPE_METADATA_KEYS:
+            value = metadata.get(key)
+            if value:
+                message_id = str(value)
+                break
+        if message_id is None:
+            raw_message = metadata.get("raw_message")
+            if isinstance(raw_message, Mapping):
+                for key in INBOUND_DEDUPE_METADATA_KEYS:
+                    value = raw_message.get(key)
+                    if value:
+                        message_id = str(value)
+                        break
+        if message_id is None:
+            return None
+
+        workspace_id = (
+            msg.workspace_id
+            or metadata.get("workspace_id")
+            or metadata.get("team_id")
+            or metadata.get("guild_id")
+            or metadata.get("aibotid")
+            or ""
+        )
+        return (msg.channel_name, str(workspace_id), msg.chat_id, message_id)
+
+    def _is_duplicate_inbound(self, msg: InboundMessage) -> bool:
+        key = self._inbound_dedupe_key(msg)
+        if key is None:
+            return False
+
+        now = time.monotonic()
+        for seen_key, seen_at in list(self._recent_inbound_events.items()):
+            if now - seen_at > INBOUND_DEDUPE_TTL_SECONDS:
+                self._recent_inbound_events.pop(seen_key, None)
+        if len(self._recent_inbound_events) > INBOUND_DEDUPE_MAX_ENTRIES:
+            overflow = len(self._recent_inbound_events) - INBOUND_DEDUPE_MAX_ENTRIES
+            for seen_key in list(self._recent_inbound_events)[:overflow]:
+                self._recent_inbound_events.pop(seen_key, None)
+
+        if key in self._recent_inbound_events:
+            logger.info(
+                "[Manager] duplicate inbound ignored: channel=%s, chat_id=%s, message_id=%s",
+                msg.channel_name,
+                msg.chat_id,
+                key[-1],
+            )
+            return True
+
+        self._recent_inbound_events[key] = now
+        return False
 
     @staticmethod
     def _log_task_error(task: asyncio.Task) -> None:
@@ -1169,7 +1232,7 @@ class ChannelManager:
             )
             return
 
-        logger.info("[Manager] invoking runs.wait(thread_id=%s, text=%r)", thread_id, msg.text[:100])
+        logger.info("[Manager] invoking runs.wait(thread_id=%s, text_len=%d)", thread_id, len(msg.text or ""))
         run_kwargs: dict[str, Any] = {
             "input": {"messages": [human_message]},
             "config": run_config,
@@ -1236,7 +1299,7 @@ class ChannelManager:
         run_context: dict[str, Any],
         human_message: dict[str, Any],
     ) -> None:
-        logger.info("[Manager] invoking runs.stream(thread_id=%s, text=%r)", thread_id, msg.text[:100])
+        logger.info("[Manager] invoking runs.stream(thread_id=%s, text_len=%d)", thread_id, len(msg.text or ""))
 
         last_values: dict[str, Any] | list | None = None
         streamed_buffers: dict[str, str] = {}
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -956,14 +956,7 @@ class ChannelManager:
         if message_id is None:
             return None
 
-        workspace_id = (
-            msg.workspace_id
-            or metadata.get("workspace_id")
-            or metadata.get("team_id")
-            or metadata.get("guild_id")
-            or metadata.get("aibotid")
-            or ""
-        )
+        workspace_id = msg.workspace_id or metadata.get("workspace_id") or metadata.get("team_id") or metadata.get("guild_id") or metadata.get("aibotid") or ""
         return (msg.channel_name, str(workspace_id), msg.chat_id, message_id)
 
     def _is_duplicate_inbound(self, msg: InboundMessage) -> bool:
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -7,6 +7,7 @@ import logging
 import mimetypes
 import re
 import time
+from collections import OrderedDict
 from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import dataclass
 from pathlib import Path
@@ -64,7 +65,10 @@ BOUND_IDENTITY_REQUIRED_MESSAGE = "Connect this channel from DeerFlow Settings, 
 BOUND_IDENTITY_UNAVAILABLE_MESSAGE = "Channel connection verification is temporarily unavailable. Please try again later or contact the DeerFlow operator."
 INBOUND_DEDUPE_TTL_SECONDS = 10 * 60
 INBOUND_DEDUPE_MAX_ENTRIES = 4096
-INBOUND_DEDUPE_METADATA_KEYS = ("event_id", "message_id", "msg_id", "client_msg_id", "client_id")
+# Only server-stable provider message ids: client-generated ids (client_msg_id,
+# client_id) are not guaranteed identical across a provider's own redelivery, so
+# keying dedupe on them would miss exactly the retries we want to absorb.
+INBOUND_DEDUPE_METADATA_KEYS = ("event_id", "message_id", "msg_id")
 
 CHANNEL_CAPABILITIES = {
     "dingtalk": {"supports_streaming": False},
@@ -777,7 +781,10 @@ class ChannelManager:
         self._semaphore: asyncio.Semaphore | None = None
         self._running = False
         self._task: asyncio.Task | None = None
-        self._recent_inbound_events: dict[tuple[str, str, str, str], float] = {}
+        # Insertion order == chronological (keys are never re-inserted), so an
+        # OrderedDict lets us evict expired/overflow entries from the front in
+        # O(k) instead of scanning all entries on every inbound message.
+        self._recent_inbound_events: OrderedDict[tuple[str, str, str, str], float] = OrderedDict()
 
     @staticmethod
     def _channel_supports_streaming(channel_name: str) -> bool:
@@ -923,6 +930,14 @@ class ChannelManager:
             except asyncio.CancelledError:
                 break
 
+            # Dedupe before logging "received" so a provider retrying an event N
+            # times does not log N accepts; duplicates are logged once as ignored.
+            # Note: this manager-level dedupe only guards the agent run / final
+            # answer. Provider adapters may emit ack side-effects (a "Working on
+            # it…" reply, an "eyes" reaction) before publish_inbound, so those are
+            # intentionally not deduped here.
+            if self._is_duplicate_inbound(msg):
+                continue
             logger.info(
                 "[Manager] received inbound: channel=%s, chat_id=%s, type=%s, text_len=%d, files=%d",
                 msg.channel_name,
@@ -931,8 +946,6 @@ class ChannelManager:
                 len(msg.text or ""),
                 len(msg.files),
             )
-            if self._is_duplicate_inbound(msg):
-                continue
             task = asyncio.create_task(self._handle_message(msg))
             task.add_done_callback(self._log_task_error)
 
@@ -970,13 +983,16 @@ class ChannelManager:
             return False
 
         now = time.monotonic()
-        for seen_key, seen_at in list(self._recent_inbound_events.items()):
-            if now - seen_at > INBOUND_DEDUPE_TTL_SECONDS:
-                self._recent_inbound_events.pop(seen_key, None)
-        if len(self._recent_inbound_events) > INBOUND_DEDUPE_MAX_ENTRIES:
-            overflow = len(self._recent_inbound_events) - INBOUND_DEDUPE_MAX_ENTRIES
-            for seen_key in list(self._recent_inbound_events)[:overflow]:
-                self._recent_inbound_events.pop(seen_key, None)
+        # Entries are in chronological insertion order, so expired ones cluster at
+        # the front: pop from the front until we hit a still-live entry.
+        while self._recent_inbound_events:
+            _, oldest_at = next(iter(self._recent_inbound_events.items()))
+            if now - oldest_at > INBOUND_DEDUPE_TTL_SECONDS:
+                self._recent_inbound_events.popitem(last=False)
+            else:
+                break
+        while len(self._recent_inbound_events) > INBOUND_DEDUPE_MAX_ENTRIES:
+            self._recent_inbound_events.popitem(last=False)
 
         if key in self._recent_inbound_events:
             logger.info(
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -956,7 +956,12 @@ class ChannelManager:
         if message_id is None:
             return None
 
-        workspace_id = msg.workspace_id or metadata.get("workspace_id") or metadata.get("team_id") or metadata.get("guild_id") or metadata.get("aibotid") or ""
+        # Fail closed: without a workspace/team/guild identifier we cannot tell two
+        # workspaces apart (e.g. Slack channel ids are not globally unique), so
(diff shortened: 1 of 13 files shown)


backend/app/gateway/routers/channel_connections.py
```diff
@@ -332,12 +333,26 @@ async def _create_state(
     owner_user_id: str,
     provider: str,
 ) -> str:
+    now = datetime.now(UTC)
```
willem-bd, line 336: **Race condition: connect-code cap is leaky under concurrency.**

`_create_state` issues `delete_expired_oauth_states`, `count_oauth_states`, and `create_oauth_state` as three independent transactions with no row-level lock or unique-constraint guard. Two concurrent `POST /api/channels/{provider}/connect` calls from the same user holding 4 active codes can both observe `pending_count == 4`, both pass the `>= 5` check, and both insert — final state is 6 active codes for that user/provider.

The new `test_connect_binding_code_caps_pending_states_per_provider` only fires 6 sequential POSTs through the sync `TestClient`, so it can't exercise this path.

**Suggested fix:** wrap delete + count + insert in a single transaction with `SELECT ... FOR UPDATE` (or a per-(owner_user_id, provider) advisory lock), or rely on a unique partial index over `(owner_user_id, provider) WHERE consumed_at IS NULL` and catch IntegrityError to enforce the cap atomically.

backend/app/gateway/routers/channel_connections.py
```diff
@@ -332,12 +333,26 @@ async def _create_state(
...
+    await repo.delete_expired_oauth_states(now=now)
```
willem-bd: **Efficiency: DB DELETE on every connect POST.**

`delete_expired_oauth_states` issues `DELETE FROM channel_oauth_states WHERE expires_at < now` on every `/connect` request, across all users and providers, even when no codes have actually expired since the last call. Under bursty connect traffic this produces unnecessary write contention and WAL churn for what is conceptually a janitor task.

Consider either a periodic background sweep, or a fast `SELECT 1 ... LIMIT 1` precheck before issuing the DELETE so the write only fires when there's something to clean up.

backend/app/channels/manager.py
```diff
@@ -898,15 +902,67 @@ async def _dispatch_loop(self) -> None:
...
+
+        workspace_id = msg.workspace_id or metadata.get("workspace_id") or metadata.get("team_id") or metadata.get("guild_id") or metadata.get("aibotid") or ""
+        return (msg.channel_name, str(workspace_id), msg.chat_id, message_id)
+
+    def _is_duplicate_inbound(self, msg: InboundMessage) -> bool:
+        key = self._inbound_dedupe_key(msg)
+        if key is None:
+            return False
+
+        now = time.monotonic()
```
willem-bd, line 985: **Efficiency: TTL pruning is O(N) per inbound.**

This runs on every inbound message that has a dedupe key:

```python
for seen_key, seen_at in list(self._recent_inbound_events.items()):
    if now - seen_at > INBOUND_DEDUPE_TTL_SECONDS:
        self._recent_inbound_events.pop(seen_key, None)
```

With `INBOUND_DEDUPE_MAX_ENTRIES = 4096`, that's a full dict copy + 4096 comparisons per dispatch tick on the hot path, even though insertion order already matches chronological order (we never re-insert the same key).

**Suggested fix:** use `collections.OrderedDict` and pop from the front while the front entry is older than TTL — O(k) where k is the number of expired entries. Same pattern handles the overflow eviction without the second `list(...)[:overflow]` copy.

backend/app/channels/manager.py
```diff
@@ -898,15 +902,67 @@ async def _dispatch_loop(self) -> None:
                 break
 
             logger.info(
```
willem-bd, line 941: **Observability: `received inbound` log fires before the dedupe check.**

When a provider retries an event 3 times within the dedupe window, the dispatcher logs `[Manager] received inbound: channel=slack, ...` 3 times AND `[Manager] duplicate inbound ignored ...` 2 times. The logs misleadingly imply 3 distinct messages were accepted; the dedupe outcome is only visible from the second log line.

This partly defeats the redaction/observability goal of the PR — operators investigating duplicate runs still see N inbound events at INFO level.

**Suggested fix:** move the `received inbound` log to *after* the dedupe check (or downgrade it to DEBUG), and include workspace/team identity on the `duplicate inbound ignored` line so operators can correlate.

willem-bd (commented):
## Code review 

Thanks for tackling this — the dedupe cache, connect-code cap, log redaction, and Slack HTTP-mode rejection all move in the right direction. Eight findings from a careful pass, ranked by severity:

**Correctness / behavior**
1. **Connect-code cap is racy** — `_create_state` does delete-expired / count / insert across three separate DB sessions with no locking, so concurrent POSTs from one user can each see count<5 and all insert (cap leaks).
2. **WeChat `thread_ts` priority changed silently** — old: `client_id → msg_id`; new: `message_id → msg_id → client_id`. For raw_messages with `client_id` and `msg_id` only, `thread_ts` switches values across the upgrade.
3. **`_inbound_dedupe_key` workspace_id falls back to `""`** — if a provider ever delivers without team/guild/aibotid, two real workspaces collapse to the same key and dedupe each other's messages.

**Cleanup / efficiency**
4. **TTL pruning is O(N) per inbound** — iterating the full 4096-entry dict on every dispatch tick instead of evicting from the front (OrderedDict.popitem(last=False)).
5. **Pre-dedupe `received inbound` log fires before the dedupe check** — Slack retries produce 3 "received" + 2 "duplicate ignored" lines; logs imply 3 distinct accepts.
6. **`team_id or event.get("team")` is redundant** in slack.py — the caller already resolved that fallback at the function entry.
7. **`delete_expired_oauth_states` runs on every connect POST** — DB DELETE per user action; should be periodic or guarded by a `SELECT ... LIMIT 1` precheck.

**Tests**
8. **`tempfile.mkdtemp()` without cleanup** in the new dedupe test — match the surrounding `tmp_path` convention.

Details inline on each line.

backend/app/channels/manager.py
```diff
@@ -898,15 +902,67 @@ async def _dispatch_loop(self) -> None:
...
+            if isinstance(raw_message, Mapping):
+                for key in INBOUND_DEDUPE_METADATA_KEYS:
+                    value = raw_message.get(key)
+                    if value:
+                        message_id = str(value)
+                        break
+        if message_id is None:
+            return None
+
+        workspace_id = msg.workspace_id or metadata.get("workspace_id") or metadata.get("team_id") or metadata.get("guild_id") or metadata.get("aibotid") or ""
```
willem-bd: **Potential cross-workspace collision when no workspace identifier is present.**

```python
workspace_id = msg.workspace_id or metadata.get("workspace_id") or metadata.get("team_id") or metadata.get("guild_id") or metadata.get("aibotid") or ""
```

The final `or ""` fallback means: if a provider delivery ever lands without any of those identifiers, the dedupe key collapses to `(channel_name, "", chat_id, message_id)`. Slack channel ids are not globally unique across workspaces, so two such workspaces could in principle dedupe each other's messages.

In practice Slack always sends `team_id`, but legacy Slack apps + some Enterprise Grid scenarios have been observed to omit it. Safer to **fail closed** (return `None` from `_inbound_dedupe_key` when no workspace identifier is found) so we don't dedupe when we can't be sure.

backend/app/channels/slack.py
```diff
@@ -354,6 +347,11 @@ def _handle_message_event(self, event: dict, *, team_id: str | None = None) -> N
             text=text,
             msg_type=msg_type,
             thread_ts=thread_ts,
+            metadata={
+                "team_id": team_id or event.get("team"),
```
willem-bd: **Cleanup: redundant fallback.**

`team_id` was already resolved at the caller:

```python
self._handle_message_event(
    event,
    team_id=req.payload.get("team_id") or req.payload.get("team") or event.get("team"),
)
```

so `team_id or event.get("team")` here can never observe a case the caller didn't already cover. Just `team_id` is enough; the duplicated fallback obscures that the upstream value is already final.

backend/app/channels/wechat.py
```diff
@@ -612,7 +612,8 @@ async def _handle_update(self, raw_message: Any) -> None:
             return
 
         context_token = str(raw_message.get("context_token") or "").strip()
-        thread_ts = context_token or str(raw_message.get("client_id") or raw_message.get("msg_id") or "").strip() or None
+        message_id = str(raw_message.get("message_id") or raw_message.get("msg_id") or raw_message.get("client_id") or "").strip()
```
willem-bd: **Behavior change: thread_ts priority quietly reordered.**

Before: `context_token > client_id > msg_id`.
After: `context_token > message_id > msg_id > client_id`.

For a WeChat raw_message lacking the (newly preferred) `message_id` but with both `client_id=A` and `msg_id=B`, this PR flips the resolved `thread_ts` from `A` to `B`.

The persisted thread mapping is keyed on `topic_id=None` for WeChat so this won't break thread lookups, but `thread_ts` is still surfaced via `InboundMessage.thread_ts` and used to key `_context_tokens_by_thread` (line 621). An in-flight `context_token` registered under the old key becomes unreachable across an upgrade-and-resume.

Worth a deliberate decision: either keep the old priority for backwards compatibility, or migrate the `_context_tokens_by_thread` cache key on startup.

backend/tests/test_channels.py
```diff
@@ -712,6 +712,42 @@ async def capture_outbound(msg):
 
         _run(go())
 
+    def test_dispatch_loop_dedupes_stable_provider_message_id(self):
+        from app.channels.manager import ChannelManager
+
```
willem-bd, line 805: **Test cleanup: leaked tempdir.**

```python
store = ChannelStore(path=Path(tempfile.mkdtemp()) / "store.json")
```

This creates a `/tmp/tmpXXXXXX` directory that is never deleted; over many test runs they accumulate. The rest of this file uses the `tmp_path` pytest fixture or context managers — match that convention here.

ShenAC-SAC (commented):
@ggnnggez Thanks for tackling #3544 — the dedupe cache, connect-code cap, log redaction, and Slack HTTP-mode rejection are all good moves. Building on @willem-bd's review (the 8 findings there all look valid to me), here are some **additional** issues from a second pass that aren't covered yet, ranked by severity.

## Correctness

### 1. Dedupe key is recorded on *receipt*, not after successful handling → a transient failure turns into permanent message loss

`ChannelManager._is_duplicate_inbound` stores the key (`self._recent_inbound_events[key] = now`) and `_dispatch_loop` then schedules `_handle_message` as a background task. But `_handle_message` swallows every exception in its `except Exception` block (a DB blip in `find_connection_by_external_identity`, a Gateway 503 from `runs.wait`, etc.) and replies *"An internal error occurred. Please try again."*

Because the key is already stored with the 10-minute TTL, a provider redelivery of the **same** `message_id` within that window — exactly the retry this PR is meant to absorb — is silently dropped, so the user never gets a response. A recoverable transient error becomes a 10-minute black hole, and the "please try again" reply is misleading since a redelivery of the same message is precisely what gets eaten.

**Suggestion:** record the key only after the message is successfully accepted (post `_handle_chat` / `_handle_command`), or `pop()` it in the failure branches so retries can recover.

## Behavior / layering

### 2. The dedupe sits below the provider's inbound ack side-effects, so retries still produce duplicate "Working on it…" replies and reactions

In `slack.py._handle_message_event` the adapter calls `_add_reaction(..., "eyes")` and `_send_running_reply(...)` **before** `publish_inbound`; the manager-level dedupe only runs later in `_dispatch_loop`. (Same pre-publish pattern in dingtalk/feishu/telegram.) So a redelivery is correctly deduped for the **agent run / final answer**, but the user still sees N "Working on it…" messages and N reactions — one per retry. The "ignore retries" goal is only half-met at the layer where the dedupe lives.

**Suggestion:** either move the dedupe signal to the provider ingress (before the running reply), or document that running-replies/reactions are intentionally not deduped.

## Test coverage (passing today, but wouldn't catch the regression)

### 3. `test_slack_http_events_mode_is_rejected` — two of three assertions are vacuous

The config supplies `bot_token` + `event_delivery="http"` but no `app_token`, and the HTTP check in `start()` runs **before** the `if not bot_token or not app_token: return` guard. So even if the HTTP-mode guard were deleted entirely, control would fall through to the missing-token early return and `_running is False` / `_web_client is None` would still hold. Only the log-substring assertion actually pins HTTP-mode rejection. Adding an `app_token` to the config so the missing-token path can't fire would give the state assertions teeth.

### 4. `test_dispatch_loop_dedupes_stable_provider_message_id` has no negative control

It only publishes the *same* inbound twice and asserts `runs.wait.call_count == 1`. It never sends a message with a *different* `message_id` and asserts it **is** processed. So a regression that over-dedupes (e.g. `_inbound_dedupe_key` returning a constant, or `_is_duplicate_inbound` returning `True` for everything after the first) — which would silently drop real distinct user messages in prod — still passes. (Minor: it also reuses one object reference for both publishes and leans on a `sleep(0.05)` race for its teeth.)

### 5. `test_count_oauth_states_active_only_and_delete_expired` doesn't pin which row survives

After `delete_expired_oauth_states` it asserts the return is `1` and the total is `1`, but not that the **active** row survived and the **expired** one is gone. An inverted predicate (`expires_at > now`) would delete the active row, still return `1`, and leave the total at `1` — test still green. The sibling `test_consume_oauth_state_deletes_expired_states` already asserts on `state_hash`; worth doing the same here.

## Minor

### 6. Client-generated ids in `INBOUND_DEDUPE_METADATA_KEYS`

`client_msg_id` / `client_id` are client-side identifiers and aren't guaranteed stable across a provider's **server-side** redelivery (reachable via WeChat's `message_id = … or raw_message.get("client_id")` fallback). If a retry carries a fresh client id the dedupe misses. Prefer keying only on server-stable ids (`ts` / `msg_id` / `event_id`).

---

Two hypotheses I checked and **cleared**, so they don't get re-chased:
- Feishu and DingTalk already set `message_id` in inbound metadata on `main`, and `INBOUND_DEDUPE_METADATA_KEYS` includes `message_id`, so they **are** deduped — not a coverage gap.
- `_initialize_operator_web_client` is still called on the Socket Mode path — not dead code after the HTTP-mode removal.

The DB/repository layer otherwise looks clean: the naive/aware datetime comparison, the `>= 5` cap boundary, and the `count_oauth_states` signature change all check out against the existing callers.

ggnnggez:
Thanks both — really thorough passes. All 14 findings are addressed; the branch is rebased onto latest `main` and pushed. Breakdown below (commits: `0eb3fe8d` convergence, `7c4b546c` correctness, `ca5b708c` cleanup/tests).

### @willem-bd

1. **Connect-code cap race** — fixed. Added `ChannelConnectionRepository.create_oauth_state_within_cap`, which runs delete-expired + count + insert in a **single transaction serialized per `(owner_user_id, provider)`**: PostgreSQL via `pg_advisory_xact_lock`, SQLite via the write lock the leading DELETE takes. The router now calls it. New test `test_create_oauth_state_within_cap_does_not_leak_under_concurrency` fires 8 concurrent issuances and asserts exactly 3 commit.
2. **WeChat `thread_ts` priority change** — reverted. The old `client_id → msg_id` priority is restored; dedupe now relies only on a `message_id` added to inbound metadata, so there's no silent behavior change across upgrade.
3. **`_inbound_dedupe_key` `""` workspace fallback** — fixed, fails closed. Returns `None` when no workspace/team/guild/aibotid identifier is present, so we never dedupe when we can't tell workspaces apart.
4. **O(N) TTL pruning** — fixed. `_recent_inbound_events` is now an `OrderedDict`; expired and overflow entries are evicted from the front with `popitem(last=False)` (O(k)).
5. **`received inbound` logged before dedupe** — fixed. The log now fires after the dedupe check, so a provider retrying N times no longer logs N accepts.
6. **Redundant `team_id or event.get("team")`** — fixed. The caller already resolves the fallback; both occurrences are now just `team_id`.
7. **`delete_expired_oauth_states` on every POST** — fixed. The in-transaction prune is now scoped to the requesting `(owner, provider)` (the rows that actually affect this cap) instead of a global DELETE; stale rows for other owners are still pruned globally by `consume_oauth_state`.
8. **Leaked `tempfile.mkdtemp()`** — fixed, switched to the `tmp_path` fixture.

### @ShenAC-SAC

1. **Dedupe key recorded on receipt → message loss on transient failure** — fixed. The key is still recorded on receipt (so retries arriving *during* handling are deduped), but `_handle_message` now releases it in the unexpected-exception branch, so a redelivery of the same `message_id` after a transient failure (DB blip, Gateway 503) is reprocessed instead of black-holed. New test `test_dispatch_loop_releases_dedupe_key_when_handling_fails`.
2. **Dedupe below provider ingress side-effects** — documented as intentional. Added a comment at the dedupe site clarifying that manager-level dedupe guards the agent run / final answer only; provider ack side-effects ("Working on it…", reactions) emitted before `publish_inbound` are intentionally not deduped.
3. **`test_slack_http_events_mode_is_rejected` vacuous assertions** — fixed. The config now supplies `app_token`, so the missing-token early return can't fire before the HTTP-mode guard; the `_running`/`_web_client` assertions now have teeth.
4. **No negative control in the dedupe test** — fixed. The test now also publishes a *different* `message_id` and asserts it **is** processed (`runs.wait` called twice), catching over-dedupe regressions; it also uses distinct objects per publish.
5. **`test_count_oauth_states...` doesn't pin which row survives** — fixed. It now asserts the surviving row is the **active** one by `state_hash`, so an inverted expiry predicate would fail.
6. **Client-generated ids in the dedupe key set** — fixed. `INBOUND_DEDUPE_METADATA_KEYS` now keeps only server-stable ids (`event_id`, `message_id`, `msg_id`); `client_msg_id`/`client_id` are dropped. Every provider already emits a server-stable `message_id`, so no provider loses dedupe.

### One cross-PR note (DingTalk log redaction)

To keep this PR conflict-free with #3582 (which restructures the same DingTalk handler), the DingTalk INFO-log redaction was **handed off to #3582** — this PR no longer touches `dingtalk.py`, and #3582 now logs `text_len` there. The two PRs can merge to `main` in either order without a conflict.

WillemJiang (approved):

PR #3585, opened 2026-06-14
feat: add fastCRW provider

us:
## What
Adds **fastCRW** as a web scrape/search provider alongside the existing Firecrawl integration — additive only, Firecrawl untouched.

## Why

[fastCRW](https://fastcrw.com) is a fully open-source (~8 MB single Rust binary, AGPL) web scraping engine that outperforms Firecrawl on Firecrawl's own benchmark dataset: **truth-recall 63.74 % vs 56.04 %**, faster median latency (**p50 ~1.9 s vs ~2.3 s**), ~6 MB RAM at rest.

**Runs 100 % locally with no cloud dependency.** Anti-bot/stealth handling, BYO-proxy + rotation, and JS rendering (Cloudflare challenge bypass, SPA rendering, HTTP → headless → proxy fallback ladder) all ship in the open core. Firecrawl's OSS self-host gates its stealth engine (`fire-engine`) behind a cloud-only flag, so a self-hosted Firecrawl instance can't reach protected or JS-heavy sites; fastCRW's can — one binary, no external services required.

**Search** is built on top of SearXNG, not as an alternative to it. SearXNG is the metasearch aggregator underneath; fastCRW adds a quality layer on top: query expansion (multi-variant rewrite), content-aware reranking (re-scoring by fetched content instead of SearXNG's content-blind ordering), and category routing (research queries fan out to arxiv / semantic scholar / Google Scholar, code queries to GitHub). You get SearXNG's breadth plus a measurable accuracy layer — all open-source (AGPL) and self-hostable with configurable engines.

Flat pricing: 1 credit = 1 page, no 4× stealth surcharge, no billed-on-failure. Free tier at https://fastcrw.com/dashboard; self-host base URL supported via `CRW_API_URL`.

The integration diff is small because fastCRW exposes a Firecrawl-compatible API — that compatibility is why this is an additive two-file change, not the reason to use it. I maintain the integration and can provide free credits to evaluate.


WillemJiang:
@us, thanks for your contribution.  Please click the CLA assistance button to sign the CLA before we start to review your PR.

us:
accepted :)

willem-bd (commented):
## Code Review: fastCRW provider

Solid, additive integration — `fastcrw/tools.py` follows the Firecrawl provider pattern closely, the `base_url`/self-host override is a nice addition, and the doctor/wizard/config/docs wiring all match the established shape. Verified the package structure (no `__init__.py`, matching `firecrawl/`), the client resolution order (config → env → default), and that Firecrawl is untouched.

Three issues worth addressing:

### 1. Env-var name mismatch between PR body and code

The PR description says *"self-host base URL supported via `CRW_BASE_URL`"*, but the code reads `os.getenv("CRW_API_URL", DEFAULT_BASE_URL)` (and the in-file comment also says `CRW_API_URL`). A user following the PR body would set the wrong variable and silently hit the cloud default (`https://fastcrw.com/api`). Pick one — `CRW_API_URL` to match the code, or `CRW_BASE_URL` to match the description — and align the PR body, the code comment, and any docs. The API-key var (`CRW_API_KEY`) is consistent everywhere. ✅

This is the one I'd want fixed before merge — it's a real user-confusion / silent-cloud-default footgun for self-hosters.

### 2. No-op test decorator

Both tests carry `@patch.dict("os.environ", {}, clear=False)` — an empty dict with `clear=False` sets nothing and clears nothing, so it's a literal no-op. It *looks* like env isolation but provides none. If the intent is to prevent `CRW_API_KEY` / `CRW_API_URL` leaking in from the host environment, use `@patch.dict(os.environ, {}, clear=True)` or explicitly delete those keys. Currently harmless because the mocked config always supplies the values so env vars are never read — but the decorator is misleading and should either be fixed or removed.

### 3. Test coverage gaps

Only the happy path (config-supplied key, non-empty results) is covered. Missing:
- **Exception path** — when `client.search` / `client.scrape` raises → the `except Exception` returning `f"Error: {e}"`.
- **"No content found" branch** in `web_fetch_tool` (empty markdown).
- **Env-var fallback** — `CRW_API_KEY` / `CRW_API_URL` when `get_tool_config` returns `None`.
- **Default `max_results=5`** when config is `None`.

These are the branches most likely to regress. The sibling Firecrawl tests have the same gaps, so this is consistent — but for a new provider, covering at least the exception and env-fallback paths would be worthwhile.

### Minor (not blocking)

- ~90% duplication with `firecrawl/tools.py`. Consistent with how every community provider is self-contained, but a shared `_firecrawl_compatible_client(base_url)` helper would prevent future drift. Optional follow-up.
- The "Only one web_fetch provider can be active at a time" note is only on the fastCRW entry, not on Firecrawl/Jina/Exa — slight asymmetry.

### Verdict

Approve with minor changes. **#1 before merge**; **#2 and #3** can be quick follow-ups.

us:
Thanks for the careful review, this is really helpful.

All three addressed:

1. **Env-var mismatch (the merge-blocker)** — aligned the PR description to `CRW_API_URL`, which is what the code reads (`os.getenv("CRW_API_URL", DEFAULT_BASE_URL)`) and what the in-file comment already documented. So the body, the comment, and the code now agree; the silent-cloud-default footgun for self-hosters is gone.

2. **No-op test decorator** — switched the `@patch.dict("os.environ", {}, clear=False)` decorators to `clear=True` so the env isolation is real (previously a no-op, as you noted).

3. **Coverage gaps** — added tests for the exception path on both tools (`web_search_tool` and `web_fetch_tool` returning `Error: {e}`), the "No content found" branch in `web_fetch_tool`, and the env-var fallback (`CRW_API_KEY` / `CRW_API_URL` when `get_tool_config` returns `None`) together with the default `max_results=5`.

On the minor points: agreed a shared `_firecrawl_compatible_client(base_url)` helper would be the right way to avoid future drift across the firecrawl/fastcrw providers; happy to do that as a follow-up so this PR stays additive and scoped. I left the "one web_fetch provider active at a time" note as-is to avoid touching the other providers' docs in this PR, but can align all of them if you'd prefer.

WillemJiang (approved):

PR #3587, opened 2026-06-15
Join active runs when opening thread

Jholly2008:
## Why

Opening an existing thread while a background run is already pending or running can leave the chat view looking idle until the page is refreshed. This happens when another client creates the run, so the current chat view loads persisted history but never attaches to the active run stream.

This fixes the user-facing gap where an existing chat can be actively running on the backend, but the frontend does not show the streaming/running state.

## What changed

- Existing chat views now inspect the thread's run list and join the latest `pending` or `running` run.
- Auto-join is best effort, avoids duplicate joins for the same run, and defers to the SDK's existing reconnect metadata when present.
- E2E mocks can now provide per-thread run lists.
- Added regression coverage for joining a run created by another client.

## Surface area

## Screenshots / Recording

Not attached. This is a stream attachment behavior change rather than a visual layout change, and it is covered by the E2E regression test.

## Bug fix verification

- Test path that reproduces the bug: `frontend/tests/e2e/thread-history.spec.ts` (`existing thread joins an active run created by another client`).
- Did it go red on `main` and green on this branch? Yes. Before the hook change, the new E2E timed out waiting for the active run stream to be joined; after the change it passes.
- Unit helper tests also covered selecting the latest active run.

## Validation

- `cd frontend && pnpm exec vitest run tests/unit/core/threads/message-merge.test.ts`
- `cd frontend && pnpm exec prettier --check src/core/threads/hooks.ts tests/unit/core/threads/message-merge.test.ts tests/e2e/thread-history.spec.ts tests/e2e/utils/mock-api.ts`
- `cd frontend && pnpm exec tsc --noEmit`
- `cd frontend && env SKIP_ENV_VALIDATION=1 DEER_FLOW_AUTH_DISABLED=1 pnpm build`
- `cd frontend && pnpm exec eslint src/core/threads/hooks.ts tests/e2e/thread-history.spec.ts tests/e2e/utils/mock-api.ts tests/unit/core/threads/message-merge.test.ts --ext .ts,.tsx`
- Production-start E2E smoke: `cd frontend && DEER_FLOW_AUTH_DISABLED=1 SKIP_ENV_VALIDATION=1 pnpm start` plus `pnpm exec playwright test tests/e2e/thread-history.spec.ts --project=chromium --reporter=line --timeout=120000 --grep=joins`

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Codex investigated the run creation/stream attachment path, implemented the frontend hook change, added regression tests, and ran the validation commands above.


frontend/src/core/threads/hooks.ts
```diff
@@ -717,6 +761,51 @@ export function useThreadStream({
...
+      try {
+        if (window.sessionStorage.getItem(`lg:stream:${threadId}`)) {
+          return;
+        }
+      } catch {
+        // Ignore storage access failures; joining the active run is best effort.
+      }
+    }
+
+    joinedActiveRunIdsRef.current.add(activeRunId);
```
fancyboi999, line 1826: `activeRunId` goes into `joinedActiveRunIdsRef` before the join, and the `.catch` below only logs — so a non-inactive failure (a transient network blip, say) permanently marks this run as joined and the effect won't retry even when `runs` re-fires, leaving the view idle-looking until a manual refresh. The inactive-run path is already handled cleanly upstream (the `joinStream` wrapper calls `clearReconnectRun` and returns), so those correctly shouldn't retry. For the genuinely-transient case, would it be worth deleting `activeRunId` from the set in the `.catch` so a later `runs` update can have another go? Minor given it's best-effort, but right now a one-off blip is indistinguishable from a permanent give-up.

fancyboi999 (commented):
Read through the join logic and it hangs together well:

- `findLatestActiveRun` (`hooks.ts:148`) filters to `pending`/`running` and picks the newest by `updated_at ?? created_at`. Ran the new unit tests locally — 2 passed, including the case where a `success` run with a *newer* timestamp is correctly skipped in favor of an older `running` one.
- The join effect (`hooks.ts:761`) is guarded three ways before it attaches: the permanent `joinedActiveRunIdsRef`, the in-flight `activeRunJoinInFlightRef`, and the `lg:stream:${threadId}` sessionStorage check. That last one is the nice part — it's the same reconnect key the SDK uses (`reconnectOnMount: true` at `hooks.ts:574`, and `clearReconnectRun` at `api-client.ts:65` already manages it), so when the SDK is auto-reconnecting on mount this effect stands down instead of double-attaching. In the cross-client case the run was created elsewhere, so this client's key is unset and the manual join fires — exactly the gap you're closing.
- The refs reset on `threadId` change (`hooks.ts:843`), so switching threads doesn't carry stale join state.

One non-blocking thought inline. Reads correct to me otherwise.

frontend/src/core/threads/hooks.ts
Jholly2008: Good point. Updated the catch path to remove the run id from joinedActiveRunIdsRef when joinStream throws, so later runs updates can retry transient failures. The inactive-run path still returns normally and keeps the run marked as handled.

WillemJiang:
@Jholly2008, could you take a look at the failed E2E test?

Jholly2008:
@WillemJiang Thanks for the heads-up.

I rebased this PR onto the latest `main`, resolved the merge conflicts, and force-pushed the updated branch. The PR is mergeable now.

I also re-ran the relevant local validation:

- `corepack pnpm exec rstest run tests/unit/core/threads/message-merge.test.ts`
- `corepack pnpm exec prettier --check src/core/threads/hooks.ts tests/unit/core/threads/message-merge.test.ts tests/e2e/thread-history.spec.ts tests/e2e/utils/mock-api.ts`
- `corepack pnpm exec eslint src/core/threads/hooks.ts tests/e2e/thread-history.spec.ts tests/e2e/utils/mock-api.ts tests/unit/core/threads/message-merge.test.ts --ext .ts,.tsx`
- `corepack pnpm exec tsc --noEmit`
- `SKIP_ENV_VALIDATION=1 DEER_FLOW_AUTH_DISABLED=1 corepack pnpm build`
- `tests/e2e/thread-history.spec.ts` with `--grep=joins` locally using system Chrome

The latest GitHub Actions jobs are currently showing `action_required`, so they likely need maintainer approval to run for this fork PR.

The earlier inline comment about retrying transient `joinStream` failures has also been addressed; please feel free to resolve that thread if it looks good.

WillemJiang:
@Jholly2008, the E2E test is still failing.

Jholly2008:
@WillemJiang Thanks for following up. I reproduced and fixed the Replay E2E failure caused by a duplicate stream join.

The PR is rebased onto the latest `main`. The failing Replay E2E now passes, along with `pnpm check`, the production build, and 52 related unit tests.

The new Actions runs show `action_required`; could you please approve them?

frontend/src/core/threads/hooks.ts
```diff
@@ -1303,6 +1435,8 @@ export function useThreadStream({
     pendingUsageBaselineMessageIdsRef.current = new Set();
     setPendingSupersededRunIds(new Set());
     setPendingSupersededMessageIds(new Set());
+    joinedActiveRunIdsRef.current = new Set();
```
willem-bd: Medium: this reset can cause a duplicate `joinStream` for the same run when re-visiting a thread with an active run.

This effect runs on `[threadId]` changes (L1442), but it is defined AFTER the auto-join effect above (deps `[isMock, runs, thread, threadId]`, L1382). React runs effects in definition order, so on a `threadId` change the auto-join effect runs first — it can join the new thread's active run and populate `joinedActiveRunIdsRef` / `activeRunJoinInFlightRef` (L1366/L1369) — and then this reset fires on the same commit and clears both. The dedup guards that `shouldAutoJoinActiveRun` relies on are then empty for the new thread.

Any subsequent re-run of the auto-join effect will rejoin the same run. That re-run is reachable on re-visit: `useThreadRuns` has no `staleTime` and default `refetchOnMount`, so navigating back to a previously-visited thread returns cached `runs` immediately (auto-join fires, join recorded), then a background refetch resolves; for a `running` run whose `updated_at` has advanced, the refetched `runs` is structurally different, so TanStack structural sharing returns a new array reference, which re-triggers the auto-join effect — now with the guards cleared by this reset — yielding a second `joinStream` for the still-active run. (A `thread` identity change during streaming would trigger the same re-run.)

The E2E test (`thread-history.spec.ts`) navigates to the thread once — a first visit with no cached `runs` — so it doesn't exercise this re-visit path and won't catch the double-join.

Fix: clear these refs BEFORE the auto-join effect runs. Move the two resets into the earlier threadId-normalization effect (~L1071, which already clears `localStreamRef` and runs before the auto-join), so stale state from the previous thread is cleared before the new thread's join is recorded. Alternatively, gate this reset so it doesn't blow away guards that were just set for the current thread on this same commit.

willem-bd (commented):
Fresh review. One medium finding: the threadId reset effect clears `joinedActiveRunIdsRef`/`activeRunJoinInFlightRef`, but it is defined after the auto-join effect, so on a `threadId` change it wipes the dedup guards the auto-join just set on the same commit. Re-visiting a thread with an active run (cached `runs` + background refetch) can then trigger a second `joinStream` for the same run. The E2E test only covers a first-visit, so it misses this. Details inline.

LittleChenLiya:
Two explicit merge blockers remain:

1. The branch conflicts with `main` (`mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`). Please rebase and revalidate the automatic run-join behavior.
2. `lint-backend` is failing. The log shows multiple Python and Markdown files failing the Ruff formatting check. These failures may not have been introduced by this PR, but the current revision does not yet have complete green CI evidence.

After resolving the conflicts, please rerun the frontend unit/E2E suites and the complete lint checks on the new merge base. If the backend lint failure is already present on the latest `main`, please provide a direct comparison so it is not mistaken for a regression introduced by this change.

frontend/src/core/threads/hooks.ts
LittleChenLiya: 复核了当前 head `2d2d273b`，这个顺序问题仍然存在：自动 join effect 仍在 1336–1382，而 `[threadId]` reset 仍在 1427–1442，因此同一次 commit 中新 run 的两个 guard 仍会先写入、后清空。建议按上面的方案把这两项 reset 移到前面的 threadId 归一化 effect，并增加“访问 A → B → 返回 A，缓存 runs 先返回后 refetch”的回归用例；当前首次访问用例不能覆盖它。

Jholly2008:
@LittleChenLiya Rebased onto the latest `main`, fixed the duplicate active-run join issue, and added the requested A → B → A regression test. Full frontend checks, backend lint, build, and related E2E tests pass locally. Could you please approve the new workflow runs?

PR #3591, opened 2026-06-15
feat(suggest_agent): make AI follow-up suggestions optional 

chetan655:
### Description
Make automatic AI follow-up suggestions configurable from the backend.

Currently, `suggest_agent` runs at the end of every AI message stream to fetch follow-up questions. This PR introduces a new toggle `(suggestions.enabled)` in `AppConfig` to allow users to completely disable this feature if they wish to save tokens or prefer a cleaner UI without suggestions.

### Changes Made
Config: Added `SuggestionsConfig` to the main `AppConfig` and exposed `suggestions.enabled` (default `true`) in `config.example.yaml`.
Router: Updated `generate_suggestions` in `suggestions.py` to immediately return an empty array if suggestions are disabled, bypassing the LLM invocation.

### Tests:
Updated existing parsing tests to run with `enabled=True`.
Added a new unit test to ensure the endpoint correctly bypasses the create_chat_model call and returns early when `enabled=False`.
Bumped `config_version` to `13`.

### DIFF
--- a/backend/app/gateway/routers/suggestions.py
+++ b/backend/app/gateway/routers/suggestions.py
@@ -135,6 +135,8 @@ async def generate_suggestions(
     request: Request,
     config: AppConfig = Depends(get_config),
 ) -> SuggestionsResponse:
+    if not config.suggestions.enabled:
+        return SuggestionsResponse(suggestions=[])
     if not body.messages:
         return SuggestionsResponse(suggestions=[])
 
--- a/backend/packages/harness/deerflow/config/app_config.py
+++ b/backend/packages/harness/deerflow/config/app_config.py
@@ -28,6 +28,7 @@ from deerflow.config.skill_evolution_config import SkillEvolutionConfig
 from deerflow.config.skills_config import SkillsConfig
 from deerflow.config.stream_bridge_config import StreamBridgeConfig, load_stream_bridge_config_from_dict
 from deerflow.config.subagents_config import SubagentsAppConfig, load_subagents_config_from_dict
+from deerflow.config.suggestions_config import SuggestionsConfig
 from deerflow.config.summarization_config import SummarizationConfig, load_summarization_config_from_dict
 from deerflow.config.title_config import TitleConfig, load_title_config_from_dict
 from deerflow.config.token_usage_config import TokenUsageConfig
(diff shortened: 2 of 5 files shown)


backend/app/gateway/routers/suggestions.py
```diff
@@ -135,6 +135,8 @@ async def generate_suggestions(
     request: Request,
     config: AppConfig = Depends(get_config),
 ) -> SuggestionsResponse:
+    if not config.suggestions.enabled:
```
fancyboi999, line 138: Good placement — ahead of the `body.messages` check and the model call, so disabled = zero LLM work. One non-blocking thought: this is backend-only, so when `enabled: false` the frontend still POSTs here and just gets `suggestions: []` back. That fully satisfies "optional" server-side, but if you want to skip the wasted round-trip entirely, a follow-up could expose the flag to the client (or short-circuit the fetch) so it doesn't call at all when disabled. Fine to leave for later.

fancyboi999 (commented):
Traced the wiring end to end and it holds:

- The `suggestions:` YAML key actually reaches the new check. `AppConfig` is built with `cls.model_validate(config_data)` (`app_config.py:234`), and since `suggestions` is a plain pydantic field (`Field(default_factory=SuggestionsConfig)`), pydantic parses the nested dict automatically. Ran it to be sure: `model_validate({...'suggestions': {'enabled': False}})` → `config.suggestions.enabled == False`; omitted → defaults `True`. So no `load_suggestions_config_from_dict` is needed (the router reads straight off the `AppConfig` instance via `Depends(get_config)`, unlike the global-singleton configs that use those loaders) — the absence is correct, not an oversight.
- The early return sits at the top of `generate_suggestions` (`suggestions.py:138`), before any model construction, so disabling truly skips the LLM call. The new `test_generate_suggestions_returns_empty_when_disabled` asserts exactly that (`ainvoke.assert_not_called()`). Whole file is green locally — 17 passed.
- `config_version` 12→13 plus the `config.example.yaml` block is the documented procedure (`backend/CLAUDE.md:231`); `_check_config_version` only warns on drift, no per-version migration required. Default `enabled=True` keeps existing behavior.

One non-blocking note inline. Looks good to me otherwise.

chetan655:
Thanks for the detailed review. I agree that stopping the network round-trip entirely is the best end-state. Since that involves exposing the backend config to the frontend, I'd like to keep this PR tightly scoped to just the backend toggle. I have opened #3593  to formally track the frontend optimization so it isn't forgotten. This PR is ready to merge whenever you have a chance

hetaoBackend (approved):

PR #3592, opened 2026-06-15
feat(memory): add guaranteed injection for correction facts with graceful fallback

sontianye:
## Why

`format_memory_for_injection()` ranks facts by confidence and greedily fills a fixed token budget. When the budget is under pressure from numerous knowledge/context facts, high-signal corrections — the very facts the memory system exists to surface — are silently evicted.

This PR introduces **guaranteed injection**: a configurable set of fact categories that draw from a reserved token budget, ensuring critical corrections (e.g. *"don't use `pip`, use `uv`"*) survive regardless of budget contention.

### New configuration

| Field | Type | Default | Description |
|-------|------|---------|-------------|
| `guaranteed_categories` | `list[str]` | `["correction"]` | Categories that bypass the regular budget |
| `guaranteed_token_budget` | `int` | `500` | Reserved token allowance for guaranteed facts (additional to `max_injection_tokens`) |

### Implementation

- **Fact partitioning** — valid facts are split into *guaranteed* and *regular* groups, each ranked by confidence, each drawing from its own budget.
- **Header-agnostic line selection** — extracted `_select_fact_lines()` that counts only inter-line `\n` separators; the caller owns header and inter-section `\n\n` accounting, eliminating prior double-counting.
- **Single `"Facts:"` header** — guaranteed and regular lines merge under one header regardless of which groups contributed.
- **Truncation ceiling extension** — the post-assembly safety truncation now uses `max_tokens + guaranteed_actual_usage` so guaranteed facts are not silently clipped.
- **Graceful fallback** — any unexpected exception in the guaranteed path falls back to the original confidence-only single-pass ranking, logged with `exc_info`.
- **Backward compatible** — `guaranteed_categories=None` or `[]` preserves the original single-budget behaviour exactly.

### Files touched

| File | Change |
|------|--------|
| `memory_config.py` | +2 fields on `MemoryConfig` |
| `agents/memory/prompt.py` | Refactored fact selection with guaranteed path + fallback |
| `agents/lead_agent/prompt.py` | Pass-through of new config fields |
| `tests/test_memory_prompt_injection.py` | +9 tests covering injection, isolation, fallback, single-header |
| `tests/test_lead_agent_prompt.py` | Updated mock signature for new kwargs |

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [x] **Agents / LangGraph** — prompt injection path
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [x] **Default behavior change** — correction facts are now guaranteed by default; opt out via `guaranteed_categories: []`

## Validation

```
cd backend
uvx ruff check .                                          # ✓
uvx ruff format --check .                                 # ✓
uv run pytest tests/test_memory_prompt_injection.py -v    # 20 passed
uv run pytest tests/test_lead_agent_prompt.py -v          # 17 passed
uv run pytest tests/ -k "memory or injection or lead_agent"  # 329 passed, 1 skipped
```

## AI assistance

**Tool(s) used:** Qoder

**How you used it:** Interactive pair-programming for design, implementation, and iterative code review.


backend/packages/harness/deerflow/agents/memory/prompt.py
```diff
@@ -374,66 +479,121 @@ def format_memory_for_injection(memory_data: dict[str, Any], max_tokens: int = 2
...
+                    use_tiktoken=use_tiktoken,
+                )
+
+            # ── Phase 2: select regular lines ────────────────────────────
+            # Regular facts compete for *max_tokens* (the main budget).
+            # Subtract everything already accounted for:
+            #   base sections + inter-section separator + header + guaranteed lines
+            regular_lines: list[str] = []
+            if regular:
+                used_before_regular = base_tokens + header_cost + guaranteed_line_tokens
```
fancyboi999: Non-blocking semantics nit: the docstring frames the guaranteed budget as *"an additional allowance on top of max_tokens"*, but Phase 2 subtracts `guaranteed_line_tokens` from the regular budget here — so in the common case (budget to spare) the total still lands at ≤ `max_tokens`, with guaranteed facts *displacing* regular ones rather than being purely additive. It only becomes truly additive (total > `max_tokens`) once `base + header + guaranteed` already exceeds `max_tokens`. The eviction-protection the feature promises is intact either way (guaranteed are picked first) — it's just that 'additional allowance on top of max_tokens' slightly oversells the common case. Might be worth a word in the docstring so a future reader doesn't expect every prompt to grow by up to `guaranteed_token_budget`.

fancyboi999 (commented):
Read through the budgeting logic and ran the suite — solid feature, and the core guarantee holds.

- Verified the guarantee end to end: `test_guaranteed_correction_injected_when_budget_tight` is a real test — three *higher*-confidence regular facts that would otherwise eat the 200-token budget, plus a lower-confidence (0.7) correction, and the correction still lands. Whole file green locally (37 passed). The mechanism is sound: guaranteed facts are selected first (Phase 1, own budget) and placed at the *front* of the single `Facts:` block, and the safety-truncation ceiling is raised to `max_tokens + guaranteed_line_tokens` (`prompt.py:587`) — since truncation cuts from the tail, regular lines go first and the front-loaded guaranteed ones survive. Nicely arranged.
- The guaranteed path is wrapped in try/except with `_fallback_format_facts` (confidence-only, single budget), and `test_fallback_on_ranking_error` covers it — so a bug in the partition logic degrades to the original behaviour rather than dropping memory entirely. Good call.
- `_select_fact_lines` now keeps scanning instead of `break`-ing on the first fact that doesn't fit (`prompt.py:374`) — a behaviour change from the original, which stopped at the first over-budget fact. It packs the budget better, but it also means a shorter lower-confidence fact can now slip in ahead of a skipped longer higher-confidence one. Looks intentional; flagging so it's a conscious call.

One note on the budget semantics inline. Reads correct otherwise.

sontianye:
Thanks for the thorough read, @fancyboi999. Pushed `9123b61` addressing each point:

**1. `_select_fact_lines` keep-scanning vs `break` (review body)**

You're right to flag this — on reflection, packing the budget better is an
orthogonal optimization that doesn't belong in a PR scoped to guaranteed
injection. Restored the original `break`-on-overflow behavior so the caller's
confidence-ordered ranking is preserved strictly. Added
`test_strict_confidence_order_when_high_confidence_fact_overflows` to lock the
invariant in as a regression test, and made the contract explicit in the
function docstring.

**2. Docstring framing of `guaranteed_token_budget` (inline comment)**

Fair point — "additional allowance on top of max_tokens" oversells the common
case. Rewrote the docstring on `format_memory_for_injection` and the design
note block to distinguish:

- *Displacement* (common): total output stays within `max_tokens`; guaranteed
  facts simply win the eviction order.
- *Additive* (only when guaranteed alone would overflow): the safety-truncation
  ceiling is raised to `max_tokens + guaranteed_actual_usage`.

Also synced the `MemoryConfig.guaranteed_token_budget` Field description so
users see the same wording at the configuration surface.

**3. Bonus: inter-group `\n` separator (self-found)**

While re-reading the budgeting, noticed Phase 2 wasn't accounting for the
single `\n` that joins the regular block to the guaranteed block when both
are present — a 1-token undercount in the regular budget. Fixed with an
explicit `inter_group_newline_tokens` term and a comment explaining the
deduction.

Full memory suite: **330 passed, 1 skipped**. Ruff clean.

xg-gh-25:
The guaranteed injection mechanism is well-motivated: correction facts like "use `uv` not `pip`" are exactly the high-signal memories that should never be evicted under budget pressure. Silently losing them defeats the core value of a memory system.

Design strengths:

1. **Dual-budget architecture** — separating guaranteed (corrections) from regular (knowledge/context) facts with independent token allocations prevents one category from crowding out the other
2. **Graceful fallback** — any exception in the guaranteed path falls back to the original single-pass ranking, so this never breaks existing behavior
3. **Zero schema change** — using existing metadata for categorization keeps the blast radius minimal

The token accounting fix in the latest commit (inter-group `\n` separator) is exactly the kind of detail that matters at scale — 1-token undercount compounds across hundreds of calls.

From production experience: guaranteed injection becomes essential once you cross ~100 accumulated facts, where budget contention starts evicting critical rules. One refinement to consider: **confidence thresholds for guaranteed categories**. If a correction fact has confidence <0.6, should it still bypass the regular budget? You might want `guaranteed_categories` to also specify a minimum confidence to prevent low-signal corrections from monopolizing the reserved budget.

The strict confidence-order test (`test_strict_confidence_order_when_high_confidence_fact_overflows`) is a solid regression guard for the restore-break-on-overflow behavior.

---
*This connects to memory prioritization and reliable fact injection in agent systems. Built by [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-MEM](https://github.com/xg-gh-25/SwarmAI/discussions/3)*

backend/packages/harness/deerflow/agents/memory/prompt.py
```diff
@@ -374,66 +490,128 @@ def format_memory_for_injection(memory_data: dict[str, Any], max_tokens: int = 2
...
     # Use accurate token counting with tiktoken (or the char-based estimate
     # when use_tiktoken is False).
     token_count = _count_tokens(result, use_tiktoken=use_tiktoken)
-    if token_count > max_tokens:
+    if token_count > effective_limit:
         # Truncate to fit within token limit
         # Estimate characters to remove based on token ratio
         char_per_token = len(result) / token_count
-        target_chars = int(max_tokens * char_per_token * 0.95)  # 95% to leave margin
+        target_chars = int(effective_limit * char_per_token * 0.95)  # 95% to leave margin
```
willem-bd: **[P1] Safety prefix-truncation silently discards guaranteed-category facts — defeats the PR's core invariant.**

Guaranteed facts are appended *after* user/history sections, but the final safety truncation keeps only the prefix of `result`. When preceding sections are large enough to push the assembled output past `effective_limit`, the truncation cuts off the trailing `Facts:` block — including the very guaranteed-category facts the PR exists to protect.

**Empirical reproduction against PR head:**

```python
from deerflow.agents.memory.prompt import format_memory_for_injection

memory = {
    'user': {'workContext': {'summary': 'X' * 4000}},  # large
    'facts': [
        {'content': 'CRITICAL: never use pip', 'category': 'correction',
         'confidence': 1.0, 'sourceError': 'pip is deprecated'},
        {'content': 'B', 'category': 'knowledge', 'confidence': 0.5},
    ]
}
out = format_memory_for_injection(
    memory, max_tokens=200,
    guaranteed_categories=['correction'], guaranteed_token_budget=500,
    use_tiktoken=False,
)

assert 'never use pip' in out  # FAILS
assert 'pip is deprecated' in out  # FAILS
# out ends with: 'XXXX...XXXXX\n...'  (Facts block truncated away)
```

The issue is structural: `result = "\n\n".join(sections)` places `Facts:` last, and `result[:target_chars]` always favors the beginning. `effective_limit = max_tokens + guaranteed_line_tokens` only widens the ceiling — it doesn't change which end gets cut.

**Fix options (any one works):**
1. Emit the `Facts:` section *before* `User Context:` / `History:` so the prefix-truncation favors guaranteed facts.
2. Make the truncation structure-aware: keep the entire `Facts:` section verbatim (or just its guaranteed prefix), truncate inside `User Context:` / `History:` instead.
3. Budget the preceding sections (rather than only facts) so they cannot grow past `max_tokens - guaranteed_line_tokens - header_cost` in the first place.

Option 2 preserves the existing visual ordering; option 1 is the simplest one-line change. Adding a regression test that mirrors the snippet above would lock this.

backend/packages/harness/deerflow/agents/memory/prompt.py
```diff
@@ -374,66 +490,128 @@ def format_memory_for_injection(memory_data: dict[str, Any], max_tokens: int = 2
...
+                        regular,
+                        token_budget=regular_line_budget,
+                        use_tiktoken=use_tiktoken,
+                    )
+
+            # ── Emit a single "Facts:" section ───────────────────────────
+            all_fact_lines = guaranteed_lines + regular_lines
+            if all_fact_lines:
+                section_text = sep + facts_header + "\n".join(all_fact_lines)
+                sections.append(section_text)
```
willem-bd, line 613: **[P2] Double inter-section separator between `User Context:` and `Facts:`.**

The facts section embeds its own leading separator via `sep = "\n\n" if has_preceding else ""`, then `section_text = sep + facts_header + ...` gets appended into `sections`. Later, `result = "\n\n".join(sections)` inserts *another* `\n\n` between every adjacent pair. Net result with preceding sections: four newlines, not two.

**Empirical reproduction against PR head:**

```python
memory = {
    'user': {'workContext': {'summary': 'Python developer'}},
    'facts': [
        {'content': 'fact A', 'category': 'knowledge', 'confidence': 0.9},
        {'content': 'fact B', 'category': 'correction', 'confidence': 0.8,
         'sourceError': 'avoid X'},
    ]
}
out = format_memory_for_injection(memory, max_tokens=2000)
# out == 'User Context:\n- Work: Python developer\n\n\n\nFacts:\n- ...'
#                                            ^^^^^^^ four newlines, not two
```

**Why this matters beyond cosmetics:**
- `header_cost = _count_tokens(sep + facts_header, ...)` budgets exactly *one* separator. The output emits two, so the running token total under-counts the real cost by 1–2 tiktokens per memory injection. Compounds with the [P1] truncation issue: under-counting makes overflow more likely, and overflow drops guaranteed facts.
- `_fallback_format_facts` has the **identical** bug (`return sep + header + "\n".join(lines)` then the caller appends it into `sections`).

**Fix:** keep `sections` separator-free and emit only `facts_header + "\n".join(all_fact_lines)`; let the final `"\n\n".join(sections)` insert the single separator. Same fix for `_fallback_format_facts`. Drop the `sep` accounting from `header_cost`.

No existing test catches this — `test_single_facts_header_when_both_guaranteed_and_regular` only counts `"Facts:"` occurrences. A regression test for the exact separator count between `User Context:` and `Facts:` would lock both this and the fallback.

backend/packages/harness/deerflow/agents/memory/prompt.py
```diff
@@ -325,14 +426,29 @@ def format_memory_for_injection(memory_data: dict[str, Any], max_tokens: int = 2
...
 
     Returns:
         Formatted memory string for system prompt injection.
     """
     if not memory_data:
         return ""
 
-    sections = []
+    # Normalise guaranteed_categories so callers can pass None/[] safely.
+    effective_guaranteed: frozenset[str] = frozenset(c.strip() for c in guaranteed_categories if isinstance(c, str) and c.strip()) if guaranteed_categories else frozenset()
```
willem-bd, line 462: **[P2] Passing a bare string for `guaranteed_categories` silently disables the guarantee.**

```python
effective_guaranteed: frozenset[str] = frozenset(
    c.strip() for c in guaranteed_categories
    if isinstance(c, str) and c.strip()
) if guaranteed_categories else frozenset()
```

A truthy non-list iterable falls through. `guaranteed_categories="correction"` iterates as individual chars and produces `frozenset({'c','o','r','e','t','i','n'})` — no real category name matches. The function returns, no warning, and the guarantee is silently off.

**Empirical reproduction against PR head** (tight budget so eviction is observable):

```python
memory = {
    'facts': [
        {'content': 'long fact A', 'category': 'knowledge', 'confidence': 0.9},
        {'content': 'long fact B', 'category': 'knowledge', 'confidence': 0.85},
        {'content': 'CRITICAL', 'category': 'correction', 'confidence': 0.8},
    ]
}
list_form = format_memory_for_injection(memory, max_tokens=100,
    guaranteed_categories=['correction'], use_tiktoken=False)
# Facts:\n- [correction | 0.80] CRITICAL\n- [knowledge | 0.90] long fact A\n- ...

str_form = format_memory_for_injection(memory, max_tokens=100,
    guaranteed_categories='correction', use_tiktoken=False)
# Facts:\n- [knowledge | 0.90] long fact A\n- [knowledge | 0.85] long fact B\n- ...
# correction last, NOT prioritized — guarantee silently off
```

The config path is Pydantic-validated as `list[str]`, so this is unreachable from `config.yaml`. But `format_memory_for_injection` is a **public helper**: anyone embedding the harness or wiring up a custom config layer can mistype it, get a green test, and ship a silently-broken guarantee.

**Fix:** explicit type check at the top of the normalization:

```python
if isinstance(guaranteed_categories, str):
    raise TypeError(
        "guaranteed_categories must be a list/iterable of strings, not a bare str"
    )
```

Or: normalize a bare `str` to `[guaranteed_categories]`. Either is fine; silent disable is the bad option.

backend/packages/harness/deerflow/agents/memory/prompt.py
```diff
@@ -374,66 +490,128 @@ def format_memory_for_injection(memory_data: dict[str, Any], max_tokens: int = 2
...
-            line_tokens = _count_tokens(line_text, use_tiktoken=use_tiktoken)
 
-            if running_tokens + line_tokens <= max_tokens:
-                fact_lines.append(line)
-                running_tokens += line_tokens
+        try:
+            # Collect valid facts and partition into guaranteed vs regular.
+            valid_facts = [f for f in facts_data if isinstance(f, dict) and isinstance(f.get("content"), str) and f.get("content", "").strip()]
+
+            def _confidence_key(fact: dict[str, Any]) -> float:
```
willem-bd, line 549: **[P2] Category-less / malformed facts get silently promoted into the guaranteed budget when `guaranteed_categories=['context']`.**

```python
if effective_guaranteed:
    guaranteed = sorted(
        [f for f in valid_facts
         if (str(f.get("category", "")).strip() or "context") in effective_guaranteed],
        ...
    )
```

The `(str(...).strip() or "context")` normalization treats missing or empty `category` fields as `"context"`. So if an operator configures `guaranteed_categories=["context"]` intending to guarantee only explicit context facts, **any malformed or legacy fact missing its category field** is promoted into the guaranteed budget — consuming `guaranteed_token_budget` and potentially displacing intentionally-categorized context facts.

Failure path: legacy fact `{"content": "...", "confidence": 0.9}` (no category) is treated as `"context"`, ranked into the guaranteed pool, and pushes a real context fact out at the budget edge.

**Fix options:**
1. Distinguish missing/empty categories from explicit `"context"` in the guaranteed-partition predicate (use the raw field, not the normalized one).
2. Document the behavior explicitly in `MemoryConfig.guaranteed_categories.description` and add a unit test pinning the current semantics.

Option 1 is safer — "category" is fact metadata, not a default-eligible signal for guaranteed promotion.

backend/packages/harness/deerflow/agents/memory/prompt.py
```diff
@@ -316,7 +316,108 @@ def _coerce_confidence(value: Any, default: float = 0.0) -> float:
...
+    max_tokens: int,
+    use_tiktoken: bool,
+) -> str | None:
+    """Confidence-only ranking used when the primary path raises an exception.
+
+    Returns the formatted ``"Facts:\\n..."`` section string, or ``None`` if
+    no facts survive.  *preceding_section_cost* is the tokens already consumed
+    by user-context / history sections (used to derive the remaining budget).
+    """
+    valid_facts = [f for f in facts_data if isinstance(f, dict) and isinstance(f.get("content"), str) and f.get("content", "").strip()]
```
willem-bd: **[P2] Simplification: `_fallback_format_facts` re-filters and re-sorts `facts_data` from scratch.**

The main path at L518 builds:
```python
valid_facts = [f for f in facts_data
               if isinstance(f, dict) and isinstance(f.get("content"), str)
               and f.get("content", "").strip()]
```

If an exception fires anywhere downstream (Phase 1 / Phase 2 / emission), `_fallback_format_facts(facts_data, ...)` is called and **redoes the same validation and the same confidence sort** from raw `facts_data`. So:

1. Maintenance coupling: any future change to the valid-facts filter or sort key must be applied in two places.
2. Wasted work on the hot prompt-injection path during fallback (proportional to `len(facts_data)`).

**Fix:** lift `valid_facts` construction outside the `try`, and pass it pre-built into `_fallback_format_facts`. Drop the duplicate filter from the fallback:

```python
valid_facts = [f for f in facts_data if ...]
try:
    ranked = sorted(valid_facts, key=_confidence_key, reverse=True)
    ...
except Exception:
    logger.warning(...)
    fallback = _fallback_format_facts(
        valid_facts,  # <-- already validated
        preceding_section_cost=base_tokens,
        max_tokens=max_tokens,
        use_tiktoken=use_tiktoken,
    )
    ...
```

This also makes the fallback faster and the contract of `_fallback_format_facts` cleaner ("format pre-validated facts").

willem-bd (commented):
**Code review (high effort) — 8 findings**

Reviewed in full (not just net-new over @fancyboi999's existing review). Two of these are correctness bugs I empirically reproduced against the PR head, and they both directly attack the core invariant of this feature ("guaranteed-category facts cannot be evicted under budget pressure"). The remaining six are smaller — API surface gaps, footguns, conventions, and one altitude observation about write-vs-read layer.

**Reproductions** (run against `pr-3592` head with the in-tree `format_memory_for_injection`):

1. **Double separator** between `User Context:` and `Facts:` — empirical output shows `\n\n\n\n` where the budget only accounted for `\n\n`.
2. **Guaranteed correction dropped by safety truncation** — with a large `user.workContext.summary` + tight `max_tokens=200` + a high-confidence correction fact, the final result keeps only the prefix of `User Context` and drops the entire `Facts:` block (correction included).

Details inline below for the five findings in the diff. Three more findings target files outside the diff and are inlined here in the summary:

---

### [P2] API surface gap — `MemoryConfigResponse` omits the new fields

**File:** `backend/app/gateway/routers/memory.py:115` (and `backend/packages/harness/deerflow/client.py:1143`)

`MemoryConfigResponse` only declares `max_injection_tokens` and `token_counting` (plus older fields). `GET /api/memory/config` and `GET /api/memory/status` therefore never expose the new `guaranteed_categories` / `guaranteed_token_budget` even when they're set in `config.yaml`.

Same gap in `DeerFlowClient.get_memory_config` — it manually builds the same dict shape and inherits the omission.

**Why it matters:**
- Operator who configures `memory.guaranteed_categories: ["correction", "goal"]` cannot verify it took effect through the API.
- Frontend / dashboard tooling that reads `/api/memory/config` shows incomplete state.
- Embedded SDK consumers via `DeerFlowClient` get a config view that doesn't round-trip with the on-disk schema.

DeerFlow has a `TestGatewayConformance` test suite (per `backend/CLAUDE.md`) that validates client/Gateway schema parity — once `MemoryConfigResponse` is updated to add the fields, the client conformance test will fail until `DeerFlowClient.get_memory_config` is also updated.

**Fix:** add both fields to `MemoryConfigResponse` (lines 115–125), wire them into the two constructors (~lines 344–352 and 373–381), and add them to `DeerFlowClient.get_memory_config` (`client.py:1143–1146`).

---

### [P2] Conventions — new memory config fields not documented in `config.example.yaml`, `config_version` not bumped

**File:** `config.example.yaml:1128` (memory section), `config.example.yaml:18` (`config_version: 14`)

`backend/CLAUDE.md` states: "When changing the config schema, bump `config_version` in `config.example.yaml`."

This PR adds two new fields to `MemoryConfig` but:
- `config.example.yaml`'s memory section still only mentions `max_injection_tokens` and `token_counting`.
- `config_version` is still 14.

**Consequences:**
- Existing deployments with custom `config.yaml` will not get the outdated-config warning from `AppConfig.from_file()`.
- New users copying the example have no documented way to discover or tune `guaranteed_categories` / `guaranteed_token_budget`. Behavior depends entirely on hidden Pydantic defaults.
- `make config-upgrade` (per `backend/CLAUDE.md`) will not auto-merge missing fields into a user's `config.yaml`.

**Fix:** add a documented `guaranteed_categories` + `guaranteed_token_budget` stanza under the `memory:` block in `config.example.yaml`, then bump `config_version` from 14 to 15.

---

### Altitude — guarantee operates only at read-time; write-time pruning can still evict corrections

**File:** `backend/packages/harness/deerflow/agents/memory/updater.py` (not in this PR's diff)

This PR enforces the guarantee at the read/inject layer. But `MemoryConfig.max_facts` (default 100) is enforced at the **write** layer in `updater.py` via confidence-only sort+truncate. So:

1. User memory accumulates 100 high-confidence knowledge facts (confidence ≥ 0.9).
2. User explicitly corrects the agent: "don't use `pip`, use `uv`." Updater extracts this as a `correction` fact with confidence 0.6 (corrections often have lower confidence — said once, vs knowledge facts reinforced over many sessions).
3. Updater applies `max_facts` pruning sorted by confidence. The new correction is below the 100-cap floor and **gets dropped before being persisted**.
4. `format_memory_for_injection` never sees the correction. The read-time guarantee has nothing to protect.

The user-facing promise ("corrections like *use uv not pip* should never be evicted under budget pressure") is only half-kept by a read-layer fix.

**Suggested deeper fix (follow-up PR, not blocking this one):** mirror the guaranteed-category logic at write-time in `updater.py` so guaranteed categories are protected from `max_facts` pruning at both layers, and the promise holds end-to-end.

---

The in-diff findings follow inline. The two correctness bugs (P1 + P2 separator) are blockers IMO; the rest are non-blocking.

sontianye:
Hi @willem-bd, thanks for the thorough review. I've pushed a new commit addressing all 8 findings. A quick mapping:

**P1 — safety truncation silently discards guaranteed facts**
Adopted Option 2 (structure-aware truncation) rather than just reordering. The `Facts:\n...` block is now treated as a **protected suffix**: only the preceding user-context / history sections are eligible for truncation, and the safety-truncation ceiling is raised by `guaranteed_line_tokens` when guaranteed facts overflow `max_tokens`. This keeps the visual order intact while guaranteeing that corrections can never be clipped.

**P2 findings**
- **Double separator**: removed the embedded `sep` from the facts section; `"\n\n".join(sections)` is now the single source of truth for section spacing.
- **Bare string `guaranteed_categories`**: now raises `TypeError("guaranteed_categories must be an iterable of strings, not a bare str")` instead of silently iterating characters.
- **Category-less fact promotion**: partition predicate now checks the *raw* `category` field, so missing-category legacy facts fall through to the regular path instead of default-promoting into `"context"`.
- **Fallback duplication / wasted validation**: `valid_facts` is pre-filtered before the `try`, and `_fallback_format_facts` receives that same list. The helper now returns `(section_text, fact_lines)` so the structure-aware truncation also protects fallback facts without string parsing.
- **API schema**: `MemoryConfigResponse` and `DeerFlowClient.get_memory_config` now expose `guaranteed_categories` and `guaranteed_token_budget`.
- **Config example**: documented both fields and bumped `config_version` from 12 to 13.

**Not in this commit**
The write-time protection in `updater.py` (corrections could be evicted by the confidence-based `max_facts` truncation) is a great follow-up. I'd like to tackle that as a separate PR to keep this one focused on the injection-time guarantee.

Regression tests are in `backend/tests/test_memory_prompt_injection.py`; the memory/router/lead-agent prompt suites all pass.

WillemJiang (approved):

WillemJiang (approved):

PR #3599, opened 2026-06-15
feat(suggest_agent): stop frontend from fetching when suggestions disabled

### LINKED ISSUE #3593: [feat] Stop frontend from fetching suggestions when disabled in backend
### Before you start

### Problem / motivation

In PR #3591 , the AI follow-up suggestions were made optional. If disabled in the backend configuration, the backend gracefully bypasses the LLM and returns an empty array.

However, the frontend currently still makes a fetch POST request to /api/threads/{thread_id}/suggestions at the end of every message because it doesn't know the feature is disabled. This results in a wasted network round-trip.

### Proposed solution

We should expose the suggestions toggle state to the frontend so it can completely short-circuit the network request

### Affected area(s)

### Alternatives considered

_No response_

### Additional context

_No response_

chetan655:
Closes #3593 

### Description
This PR is the frontend follow-up to PR #3591. It prevents the `InputBox` component from making a redundant `POST /suggestions` network request when the AI follow-up feature is disabled in the backend configuration.

### Changes Made
1. **Backend**: Added a lightweight `GET /api/suggestions/config` endpoint in the suggestions router that exposes the `suggestions.enabled` AppConfig boolean.
2. **Frontend**: Created `core/suggestions/api.ts` to cleanly fetch the new endpoint (mirroring the MCP config fetch pattern).
3. **Frontend**: Wired up `useQuery` with `staleTime: Infinity` in `input-box.tsx` so the configuration is fetched exactly once and cached infinitely.
4. **Frontend**: Added a short-circuit guard in the `generateSuggestions` effect to drop the request entirely if the feature is disabled.

### Testing
- Tested locally with `suggestions.enabled: false`. Verified via Network Tab that the `POST` request is successfully bypassed.
- Tested locally with `suggestions.enabled: true`. Verified that suggestions are generated and rendered correctly.


frontend/src/components/workspace/input-box.tsx
```diff
@@ -540,6 +547,11 @@ export function InputBox({
       return;
     }
 
+    if (suggestionsConfig && !suggestionsConfig.enabled) {
```
fancyboi999: Minor, non-blocking: this guard is `suggestionsConfig && !...enabled`, so while the config query is still in flight (`suggestionsConfig === undefined`) it's falsy and the effect falls through to the POST. On the very first message after mount — before the one-shot query resolves — a single `/suggestions` request can still fire even when the feature is disabled. Harmless (the backend early-returns `[]`), but it's exactly the round-trip this PR is trying to drop, so if you want it airtight you could also skip while the config is still loading (e.g. treat `undefined` as 'not yet known, hold off').

fancyboi999 (commented):
Followed the wiring through and it matches the direction from #3593 cleanly:

- `GET /api/suggestions/config` (`suggestions.py:126`) returns `config.suggestions.enabled` straight off the `AppConfig` instance — same `Depends(get_config)` the POST uses. (I'd separately confirmed on #3591 that `suggestions.enabled` parses correctly out of config, so this reads the right value.)
- `loadSuggestionsConfig` falls back to `{ enabled: true }` on any non-ok/error (`core/suggestions/api.ts`), so an older backend without this route keeps suggestions on — good backward-compat default.
- `useQuery` with `staleTime: Infinity` fetches once and caches; the effect's dep array now includes `suggestionsConfig` so the guard re-evaluates once it loads; and the backend POST early-return from #3591 is left in place, so a stale client still can't force the LLM call. Defence in depth intact.

One small timing note inline. Looks correct otherwise.

chetan655:
Great catch! You're completely right about that timing gap on the initial mount.

I've pushed a small fix to address it: by updating the guard to if `(suggestionsConfig === undefined || !suggestionsConfig.enabled)`, the effect now properly holds off on making the POST request while the config query is still in flight, ensuring the short-circuit is 100% airtight

frontend/src/components/workspace/input-box.tsx
```diff
@@ -540,6 +547,11 @@ export function InputBox({
       return;
     }
 
+    if (!suggestionsConfig?.enabled) {
+      setFollowups([]);
+      return;
+    }
```
hetaoBackend, line 551: ⚠️ **Possible silent regression while this config query is still loading.**

The AI message id is claimed earlier in this same effect — `lastGeneratedForAiIdRef.current = lastAiId` at line 534 — **before** this config gate. So if `suggestionsConfig` is still `undefined` (the `GET /api/suggestions/config` query hasn't resolved yet) at the moment an assistant reply finishes streaming, `!undefined?.enabled` is `true` → we `setFollowups([])` and return *without* fetching. When the config later resolves and the effect re-runs, it bails at the falling-edge guard (`if (!wasStreaming || streaming) return;`) **and** the id is already consumed, so it never re-arms → suggestions are lost for that turn even when `enabled: true`.

In practice this is low-probability (the config endpoint is a trivial boolean that almost always resolves before the first LLM reply, and the mock path already returns at line 523), but the failure is silent and the previous commit's `if (suggestionsConfig && !suggestionsConfig.enabled)` was actually safer here.

Suggested fix: gate on a resolved value and don't claim the id before deciding — e.g. add `if (suggestionsConfig === undefined) return;` *above* line 534, then keep `if (!suggestionsConfig.enabled) { setFollowups([]); return; }`. Note the fix must still let the turn fetch once config resolves (it interacts with the `wasStreamingRef` falling-edge guard), so it's a bit more than a one-line change.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -576,7 +588,14 @@ export function InputBox({
...
 
     return () => controller.abort();
-  }, [context.model_name, disabled, isMock, status, threadId]);
+  }, [
+    context.model_name,
+    disabled,
+    isMock,
+    status,
+    threadId,
+    suggestionsConfig,
```
hetaoBackend: Minor: this effect only reads `suggestionsConfig?.enabled`, so depending on the whole `suggestionsConfig` object is looser than needed. `suggestionsConfig?.enabled` is the precise dep — it works either way thanks to `staleTime: Infinity`, but it matches the value actually consumed at line 550.

hetaoBackend (commented):
Reviewed the diff (3 files, +43/-1) against the spec (#3593) and the repo standards (`backend/CLAUDE.md`, `frontend/CLAUDE.md`, `CONTRIBUTING.md`). The approach is right and works in the common case. Findings are inline; summary below.

**Logic (worth fixing):** a low-probability but *silent* regression — suggestions can be dropped for a turn if the config query is still in flight when an AI reply finishes streaming. See the inline comment on the new guard.

**Standards / cleanup:**
- Add a unit test for the new endpoint — `backend/CLAUDE.md` states TDD is MANDATORY ("Every new feature or bug fix MUST be accompanied by unit tests. No exceptions."). A frontend test for the short-circuit would be good too.
- Drop the unused `request` param in `get_suggestions_config`.
- Tighten the effect dependency to `suggestionsConfig?.enabled`.
- Update the Routers table in `backend/CLAUDE.md` with the new `GET /api/suggestions/config` (Documentation Update Policy).
- Optional consistency: other domains (`models`, `mcp`, `channels`) pair `core/<x>/api.ts` with a `hooks.ts` exposing a `use…` hook; consider `core/suggestions/hooks.ts` `useSuggestionsConfig` instead of inlining `useQuery` in `input-box.tsx`.

The unauthenticated endpoint is acceptable (it exposes only a global boolean, like the other config GETs `models`/`memory`), but worth a one-line note in the PR description since the sibling route is permission-gated. Spec check: the backend path resolves correctly to `/api/suggestions/config`, and the `{ enabled: true }` fallback on error is a sensible default. Nice work overall.

backend/app/gateway/routers/suggestions.py
```diff
@@ -122,6 +126,14 @@ def _format_conversation(messages: list[SuggestionMessage]) -> str:
     return "\n".join(parts).strip()
 
 
+@router.get("/suggestions/config", response_model=SuggestionsConfigResponse, summary="Get Suggestions Configuration", description="Returns the global configuration for follow-up suggestions.")
+async def get_suggestions_config(
+    request: Request,
```
hetaoBackend: `request: Request` is unused in this handler. The sibling config-getter pattern — `models.py::list_models(config: AppConfig = Depends(get_config))` — takes only `config`. `request` is needed by `generate_suggestions` below because its `@require_permission(...)` decorator reads it, but there's no such decorator here, so this param can be dropped.

backend/app/gateway/routers/suggestions.py
```diff
@@ -122,6 +126,14 @@ def _format_conversation(messages: list[SuggestionMessage]) -> str:
     return "\n".join(parts).strip()
 
 
+@router.get("/suggestions/config", response_model=SuggestionsConfigResponse, summary="Get Suggestions Configuration", description="Returns the global configuration for follow-up suggestions.")
```
hetaoBackend: Two repo-standard follow-ups for this new endpoint:

- **Tests** — `backend/CLAUDE.md`: "TDD is MANDATORY … Every new feature or bug fix MUST be accompanied by unit tests. No exceptions." `tests/test_suggestions_router.py` covers `generate_suggestions` but nothing exercises `get_suggestions_config` for `enabled: true`/`false`. Please add coverage.
- **Docs** — per the Documentation Update Policy, the Routers table in `backend/CLAUDE.md` currently lists only `POST /` for Suggestions; add the new `GET /api/suggestions/config`.

chetan655:
@hetaoBackend  Thanks for the incredibly thorough review.

I've just pushed a force update that addresses all of your feedback:

1. Frontend Timing Bug: Added the if `(suggestionsConfig === undefined) return`; guard before claiming the `lastAiId` so we safely bail out while the config is in-flight, and tightened the dependency array to precisely `suggestionsConfig?.enabled`.
2. Frontend Test: Added a new Playwright e2e test in `chat.spec.ts` that explicitly verifies the frontend short-circuits and never makes the `POST` request when the config resolves to `disabled`.
3. Frontend Refactor: Extracted the React Query logic into a clean `useSuggestionsConfig()` hook inside `core/suggestions/hooks.ts` to match the repo's consistency standards.
4. Backend Cleanup: `Dropped` the unused `request: Request` parameter from `get_suggestions_config`.
5. Backend Tests: Added the mandatory unit test for `get_suggestions_config` returning True/False in `test_suggestions_router.py`.
6. Documentation: Updated the Routers table in `backend/CLAUDE.md` to properly document the new `GET /api/suggestions/config` route.

hetaoBackend:
Nice, focused PR — the short-circuit works and it's well covered (backend unit test + e2e). A few things I'd suggest tightening:

**Worth fixing**
1. **Multi-line the new route decorator** (`suggestions.py:129`). It's a single ~178-char line, while every other endpoint in this same file (and `memory.py`) splits `path`/`response_model`/`summary`/`description` across lines. ruff won't wrap it at the 240-col limit, so it stays inconsistent.
2. **Add an explicit return type / shared type for `loadSuggestionsConfig`** (`api.ts:4`). Siblings annotate (`loadModels(): Promise<ModelsResponse>`); here the shape only appears inline on the success path and the `{ enabled: true }` fallback is inferred. A small exported type mirroring the backend `SuggestionsConfigResponse` keeps both ends in sync.
3. **Typo in test comment** (`test_suggestions_router.py:205`): "Test then disabled" → "when".

**Worth a conscious decision (not necessarily a change)**
4. `GET /suggestions/config` has no `@require_permission`, unlike `generate_suggestions`. That's consistent with the other global `GET /config` endpoints (`memory.py`, `mcp.py`), so it's defensible — just confirming the unauthenticated global `enabled` boolean is intended.
5. The `{ enabled: true }` fallback (`api.ts:6-8`) is right for an older backend (404 → feature exists), but on a *transient* error from a new backend it re-enables the very POST this PR removes. Consider falling back to `true` only on 404, or noting that the fail-open is deliberate.
6. `staleTime: Infinity` caches the toggle for the whole session, so flipping `suggestions.enabled` (hot-reloaded backend-side) won't reach an open tab until reload. Probably fine — just flagging.

Overall LGTM once the decorator formatting + the small type/typo nits are in.

chetan655:
@hetaoBackend  Thanks again for the review.

### Fixes applied:

1. Multi-lined the `get_suggestions_config` decorator to match the style of the rest of the file.
2. Exported `SuggestionsConfigResponse` in `api.ts` and strongly typed `loadSuggestionsConfig`.
3. Updated the fallback in `loadSuggestionsConfig`: it now strictly fails-open on a `404 (for backwards compatibility)`, but throws a standard Error for transient 500s from new backends so React Query can properly catch and retry it.

### Regarding the conscious decisions:

4.  Yes, leaving it unauthenticated without `@require_permission` is intended, matching the pattern of other global config endpoints like `mcp.py` and `memory.py`.
5.  Caching it for the session with `staleTime: Infinity` feels correct here. A hard reload is a reasonable expectation if someone toggles global features behind the scenes.

WillemJiang (approved):

PR #3600, opened 2026-06-15
fix(mcp): make stdio MCP-produced files resolvable via virtual sandbox paths (#3597)

### LINKED ISSUE #3597: [bug] playwright/mcp 与deerflow路径不共享，playwrigh写入的文件deerflow读取不到
### Before you start

### Problem summary

playwright/mcp 与deerflow路径不共享，导致文件无法读取

### Affected area(s)

MCP

### What happened?

接入playwright/mcp 后，playwright的部分方法会保存文件到文件夹中，但是由于deerflow运行在沙盒中，加上用户隔离机制，playwright写入文件后返回的路径deerflow找不到。即便为playwright指定--ouput-dir为指定路径也没有作用，用户隔离的情况下，每个个用户的目录是不一样的

### Expected behavior

提供路径共享或其他解决方案

### Steps to reproduce

无

### Relevant logs

```shell
无
```

Local (make dev)

### Operating system

Linux

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

18062706139fcz:
## What

Stdio MCP servers (e.g. Playwright) save files to host paths that the sandbox/artifact API cannot resolve, since it only serves paths under `/mnt/user-data`. As a result, files produced by these servers are not readable by the rest of DeerFlow, and passing `--output-dir` does not help because that directory is outside the sandbox mounts.

This PR makes such files readable **without copying them** — it pins where stdio servers write and then deterministically remaps the returned references to virtual paths:

- **Pin the write location.** For stdio transports, the pooled subprocess `cwd` and its temp dir (`TMPDIR`/`TMP`/`TEMP`) are set inside the thread's mounted user-data tree (`workspace/` and `workspace/.mcp/tmp/`). Tools that write relative paths or to the OS temp dir then land in a servable location. Operator-provided `env` is merged (`setdefault`), not replaced.
- **Deterministic host→virtual remap (no copy).** `ResourceLink` URIs and local-file references in free text are translated from their real host path to the matching `/mnt/user-data/...` virtual path (`workspace/...` or `outputs/...`, wherever the file actually lives). Translation only happens after `.resolve()` confirms the file exists and is inside this thread's user-data root (`relative_to(user_data_root)`), which is symlink-escape-safe. The file itself is never copied, sanitized, de-duplicated, or size-checked; its relative layout is preserved verbatim.
- **Conservative text fallback.** Some servers (e.g. Playwright `browser_take_screenshot`) report saved files only as free text. Path-like tokens are rewritten only when they resolve to an existing file inside the tree; bare filenames (e.g. `page-2026-06-16T10-21-46-864Z.yml` with no `--output-dir`) are rewritten only when correlated to exactly one file created/modified by that same tool call. Non-matching tokens are left untouched.
- Remote URIs (http/https/data) and files outside the thread's user-data tree are left untouched.
- Pooled MCP sessions are scoped by `user_id:thread_id` instead of `thread_id` alone, matching the per-(user_id, thread_id) filesystem isolation.

The cwd/temp pinning and workspace snapshots apply to **stdio transports only**; SSE/HTTP servers skip the filesystem work entirely. The synchronous filesystem operations (dir prep, snapshot diff, per-token path resolution) are offloaded via `asyncio.to_thread` so they never block the event loop, and the post-call snapshot diff is skipped when the result has no text content to rewrite.

## Testing

- New `backend/tests/test_mcp_file_migration.py` covering URI/path parsing, deterministic host→virtual translation (including the reviewer's relative `temp/...` case and `--output-dir` cases), security-boundary rejection (outside tree, symlink escape), free-text rewriting, and bare-filename correlation.
- Updated session-pool scope and cwd/temp-env pinning assertions.
- `uv run pytest tests/test_mcp_file_migration.py tests/test_mcp_session_pool.py` — all passing.
- Static blocking-IO detector reports no event-loop blocking risk in the changed file.

Closes #3597


backend/packages/harness/deerflow/mcp/tools.py
```diff
@@ -3,22 +3,137 @@
...
+    """
+    src = _local_path_from_uri(uri)
+    if src is None:
+        return None
+
+    try:
+        real = src.resolve()
+    except OSError:
+        return None
+    if not real.is_file():
```
fancyboi999, line 106: Security consideration, and it's a bit sharper precisely because this PR is hardening multi-user isolation: `_migrate_local_file_to_outputs` accepts *any* absolute `file://`/bare path the MCP server returns, and as long as it's a regular file under 100 MB it gets `shutil.copy2`'d into the requesting user's `/mnt/user-data/outputs/` and handed back as a readable virtual path. There's no check that the source lives under an expected MCP-output root. So a shared/operator-configured MCP server — or a tricked one (e.g. Playwright induced to save to an attacker-influenced path) — could return `file:///etc/passwd`, `/app/.env`, or *another user's* sandbox file as a `ResourceLink`, and it'd be copied into this user's outputs. `test_sanitizes_unsafe_chars` (`_safe_output_name('/etc/passwd') == 'passwd'`) shows the source isn't rejected, only the destination *name* is sanitized. Worth confining migration to files under a sanctioned root (the server's configured working/output dir, or a per-(user,thread) temp area) and refusing sources outside it.

fancyboi999 (commented):
Nice — this takes the post-hoc-migration route (copy the MCP-produced file into the thread's sandbox outputs and rewrite the URI) rather than templating an output-dir, which sidesteps having to relaunch the server per session. Ran the suite: `test_mcp_file_migration` + session-pool + sync-wrapper all green (57 passed). Two things I checked specifically:

- **The scope change is safe.** Moving the pool key from `thread_id` to `user_id:thread_id` (`tools.py`) had me worried about orphaned cleanup, but `close_scope()` has no production caller — only tests — and the one prod teardown is `close_all_sync()` (`cache.py`), which is scope-agnostic. And `resolve_runtime_user_id()` always returns a `str` (falls back to `DEFAULT_USER_ID`), so the key is never malformed and matches the per-(user_id, thread_id) filesystem isolation. No leak, good call.
- **One security concern inline** worth weighing before merge — the migration will copy *any* absolute local path the MCP server names, with no confinement to an expected output root.

backend/packages/harness/deerflow/mcp/tools.py
18062706139fcz: Great catch, thanks for flagging this. You're right that the original revision sanitized only the destination name , not the source — so a server returning file:///etc/passwd , /app/.env , or another user's sandbox file would have been copied into the requesting user's outputs.

I've addressed this in 0bbd9fb ("restrict file migration to trusted source roots"). Migration is now confined to a source-root allowlist:

- The resolved source path ( src.resolve() , so symlink escapes are caught too) must live under one of: the OS temp dir (Playwright's default --output-dir location), the current (user_id, thread_id) sandbox user-data tree, or an operator-configured root via DEERFLOW_MCP_MIGRATION_SOURCE_ROOTS .
- Anything outside those roots is refused and logged — no copy, original URI left untouched.
This covers all three examples you raised:

- file:///etc/passwd / /app/.env → not under any trusted root → refused.
- Another user's sandbox file → only the current user/thread user-data dir is added to the allowlist, so cross-user paths don't match → refused.
Added tests for the escape-refusal path, trusted-root acceptance, the env-configured root, and URL-encoded file:// inputs. PTAL when you get a chance.

firefishu:
> ## 什么
> Stdio MCP 服务器（例如 Playwright）会将文件保存到沙盒/artifact API 无法解析的主机路径，因为该 API 只处理特定路径`/mnt/user-data`。因此，这些服务器生成的文件无法被 DeerFlow 的其他部分读取，即使传递参数也`--output-dir`无济于事，因为该目录位于沙盒挂载点之外。
> 
> 此 PR 使此类文件可读：
> 
> * 结果引用的本地文件`ResourceLink`被复制到线程的沙盒输出目录中，它们的 URI 被重写为`/mnt/user-data/outputs/...`，以便它们可以通过现有的沙盒/工件路径解析。
> * 远程 URI（http/https/data）和无法访问的文件保持不变。
> * 文件名经过清理和去重处理；过大的文件将被跳过。
> * 池化的 MCP 会话现在按(user_id, thread_id)`user_id:thread_id`而不是`thread_id`单独进行范围划分，与每个 (user_id, thread_id) 文件系统隔离相匹配。
> 
> ## 测试
> * 新增了`backend/tests/test_mcp_file_migration.py`路径解析、文件名清理、迁移流程和`ResourceLink`URI 重写功能。
> * 更新了会话池作用域断言。
> * `uv run pytest tests/test_mcp_session_pool.py tests/test_mcp_file_migration.py tests/test_mcp_sync_wrapper.py`全部通过。
> 
> 关闭 #3597

@18062706139fcz 
palywrite mcp工具以文本形式返回写入的路径信息，我实测不会被拦截到

18062706139fcz:
**Follow-up fixes and regression verification:**

**Fixes:**
- Playwright MCP actually returns the screenshot path as a markdown/text path inside `TextContent`, not as a `ResourceLink`. The conversion now scans local paths in `TextContent` / embedded text and migrates them to `/mnt/user-data/outputs/...`.
- The stdio MCP session now uses the per-user/per-thread sandbox workspace as the subprocess `cwd`, ensuring Playwright's relative output paths (e.g. `.playwright-mcp/page-*.png`, `./homepage.png`) land inside that thread's user-data tree.
- When the same source path appears multiple times in a single response text, it is migrated only once, avoiding inconsistent `_1` / `_2` paths.

**Regression verification:**
- Launched a real `@playwright/mcp@latest` and invoked `browser_navigate` + `browser_take_screenshot` through the full DeerFlow wrapper path.
- Default screenshot filename case: Playwright returns the text path `.playwright-mcp/page-*.png`; verified the returned text is rewritten to `/mnt/user-data/outputs/page-*.png` and a non-empty png is produced under outputs.
- Explicit filename case: Playwright returns the text path `./homepage.png`; verified the returned text is rewritten to `/mnt/user-data/outputs/homepage.png` and a non-empty png is produced under outputs.
- Automated regression: `pytest tests/test_mcp_file_migration.py`, 52 passed.
- Static check: `ruff check packages/harness/deerflow/mcp/tools.py tests/test_mcp_file_migration.py`, passed.

firefishu:
> **Follow-up fixes and regression verification:**
> 
> **Fixes:**
> 
> * Playwright MCP actually returns the screenshot path as a markdown/text path inside `TextContent`, not as a `ResourceLink`. The conversion now scans local paths in `TextContent` / embedded text and migrates them to `/mnt/user-data/outputs/...`.
> * The stdio MCP session now uses the per-user/per-thread sandbox workspace as the subprocess `cwd`, ensuring Playwright's relative output paths (e.g. `.playwright-mcp/page-*.png`, `./homepage.png`) land inside that thread's user-data tree.
> * When the same source path appears multiple times in a single response text, it is migrated only once, avoiding inconsistent `_1` / `_2` paths.
> 
> **Regression verification:**
> 
> * Launched a real `@playwright/mcp@latest` and invoked `browser_navigate` + `browser_take_screenshot` through the full DeerFlow wrapper path.
> * Default screenshot filename case: Playwright returns the text path `.playwright-mcp/page-*.png`; verified the returned text is rewritten to `/mnt/user-data/outputs/page-*.png` and a non-empty png is produced under outputs.
> * Explicit filename case: Playwright returns the text path `./homepage.png`; verified the returned text is rewritten to `/mnt/user-data/outputs/homepage.png` and a non-empty png is produced under outputs.
> * Automated regression: `pytest tests/test_mcp_file_migration.py`, 52 passed.
> * Static check: `ruff check packages/harness/deerflow/mcp/tools.py tests/test_mcp_file_migration.py`, passed.

@18062706139fcz 
 temp/page-2026-06-16T10-21-46-864Z.yml 实测这种路径没有支持，修改一下正则表达式就可以了。但是感觉这种从文本中提取路径并不是一个完善的解决方案，如果未指定output-dir就是会是page-2026-06-16T10-21-46-864Z.yml 这种路径，是无法保证是否是文件路径的

18062706139fcz:
@firefishu 

问题
Playwright 这类 stdio MCP 工具产出的文件（截图、yaml 等），写到了 sandbox/artifact API 解析不了的路径，导致前端链接打不开。原 PR 靠正则从文本里处理路径，大概率覆盖不全这么多 case。

解决思路
不再走 copy 逻辑，改成三步：
1. 固定写入位置 ：给 stdio 子进程注入 cwd 和 TMPDIR/TMP/TEMP ，都指向当前 thread 的 workspace，文件必然落在 mounted 目录里。
2. 确定性翻译 ：把落在 user-data 树内的路径翻译成 /mnt/user-data/... ，树外的原样保留，不暴露任意文件。
3. 文本兜底 ：自由文本里的路径只在"树内真实存在文件"时才替换；裸文件名靠"本次调用新增文件唯一匹配"判断，不盲猜。

reviewer 的 case 已修复，单测 83 passed，并用真实 stdio 子进程验证文件确实落进 workspace 并正确翻译。

可以再看看。

willem-bd:
Thanks for this — the pin-and-remap design is sound and the test coverage is thorough. I traced the serving path (`artifacts.py` → `resolve_virtual_path`) and confirmed files written to the pinned cwd are genuinely servable as `/mnt/user-data/workspace/...`, and the security boundary in `_local_uri_to_virtual_path` (`.resolve()` + `relative_to(user_data_root)`, symlink-escape-safe) is correct. Three blockers before merge:

## 1. PR description does not match the implementation

The "What" section describes a **copy-based** approach that doesn't exist in the code:

| PR body claims | Actual implementation |
|---|---|
| "files … are **copied** into the thread's sandbox outputs directory" | No copy — pure deterministic host→virtual path remap (`_local_uri_to_virtual_path`). |
| "Filenames are sanitized and de-duplicated" | No sanitization, no de-dup; relative path preserved verbatim. |
| "oversized files are skipped" | No size check anywhere. |
| "rewritten to `/mnt/user-data/outputs/...`" | Rewritten to `/mnt/user-data/workspace/...` (cwd) or `/outputs/` — wherever the file actually lives. |

The new test file's docstring ("translating MCP-produced local files into virtual sandbox paths") matches the **implementation**, not the body. The pin-and-remap design is genuinely cleaner than copy+sanitize, but reviewers reading the body will be misled — please rewrite the PR body before merge.

## 2. Blocking-IO on the event loop (project has a hard CI gate for this)

`call_with_persistent_session` is an `async` coroutine, but it now does a lot of **synchronous filesystem work inline**:

- `paths.ensure_thread_dirs(...)` → 4× `mkdir` + `chmod`
- `tmp_dir.mkdir()` + `tmp_dir.chmod(0o777)`
- `_snapshot_workspace_files(source_base_dir)` **before** the call → full `rglob("*")` + `stat()` per file
- `_changed_workspace_files(...)` **after** → another full `rglob` + `stat` per file
- `_convert_call_tool_result` → `_local_uri_to_virtual_path` per text token → `resolve()` + `is_file()` + `relative_to()`

This repo maintains a blocking-IO runtime gate (`tests/blocking_io/`, `make test-blocking-io`, `.github/workflows/backend-blocking-io-tests.yml`, the `blocking-io-guard` skill) and offloads comparable FS work via `asyncio.to_thread` (skills loading, sqlite, uploads middleware, JSONL event store). The new code should follow the same pattern — at minimum the snapshot diff (two recursive walks) and the per-token path resolution should be wrapped in `asyncio.to_thread`. As written, a workspace with many files stalls the event loop on every MCP call.

## 3. Performance: two full recursive walks per tool call + applies to all transports

`_snapshot_workspace_files` (before) + `_changed_workspace_files` (after) run on **every** MCP tool invocation, and the only consumer is the bare-filename correlation in `_rewrite_unique_bare_filenames`. For a stateful Playwright session calling `navigate` repeatedly while accumulating screenshots, that's 2× `rglob` + stat-every-file per call. Consider deferring/skipping the `after` walk when the result has no text content, and/or scoping the walk to known output subdirs instead of the whole `workspace/`.

Separately, the cwd/env/`TMPDIR` pinning is meaningless for SSE/HTTP servers, but `ensure_thread_dirs`, the `mkdir`/`chmod`, and **both** snapshot walks run unconditionally for every server type. Gate the whole block on `connection.get("transport") == "stdio"`.

---

(I also noted some non-blocker items — colon-join in `scope_key`, operator `cwd` overridden unconditionally, and confirming the frontend renders `/mnt/user-data/workspace/...` URLs — happy to detail if useful.)

WillemJiang:
@18062706139fcz Please fix the lint error and address the review comments.

willem-bd (changes requested):
## Overview

Stdio MCP servers (notably Playwright) were writing files to host paths the sandbox/artifact API can't serve. The fix avoids copying by pinning the subprocess `cwd` and `TMPDIR`/`TMP`/`TEMP` inside the thread's mounted user-data tree, then deterministically remapping returned host paths to `/mnt/user-data/...` via `Path.resolve()` + `relative_to(user_data_root)` (symlink-escape-safe). Free-text rewriting is conservative — it only rewrites tokens that resolve to real files inside the tree — and bare filenames are correlated against a workspace snapshot diff. Session pool key widened from `thread_id` to `user_id:thread_id`. Synchronous filesystem work is offloaded via `asyncio.to_thread`.

The core design is the right shape, the security boundary is correct, and the test coverage is excellent (575 lines of new tests + 240 lines of session-pool tests). I have a small set of requested changes before merging, plus a couple of follow-ups that don't have to block this PR.

## Requested changes

### 1. `tmp_dir.chmod(0o777)` is overly permissive

`backend/packages/harness/deerflow/mcp/tools.py:167-168`

```python
tmp_dir.mkdir(parents=True, exist_ok=True)
tmp_dir.chmod(0o777)
```

That makes a directory inside the user's tree (`users/{user_id}/threads/{thread_id}/user-data/workspace/.mcp/tmp/`) world-readable and world-writable. On a multi-tenant host or a shared CI runner, any local user can read/write another user's MCP temp files. The PR description doesn't motivate the choice — the stdio subprocess is presumably the same UID as the gateway, so `0o700` is sufficient.

Please either drop the `chmod` and rely on inherited umask/ownership, or set `0o700`. If `0o777` is genuinely required for a non-root sandbox container with a different UID, please add a comment explaining why so the next reviewer doesn't strip it.

### 2. `cwd` should `setdefault`, not unconditionally overwrite

`backend/packages/harness/deerflow/mcp/tools.py:409`

The env handling correctly respects operator config:

```python
session_env.setdefault("TMPDIR", str(tmp_dir))
```

…but `cwd` does not:

```python
session_connection["cwd"] = source_base_dir
```

So an operator who explicitly configures `cwd:` for a stdio MCP server (e.g. to resolve a local config file or sqlite DB) silently has it overwritten. Inconsistent with the env handling. Suggest:

```python
session_connection.setdefault("cwd", source_base_dir)
```

…and a regression test mirroring `test_session_pool_tool_does_not_override_explicit_tmpdir` for `cwd`.

While here, please also stringify `cwd` (`str(source_base_dir)`) — both are accepted by `asyncio.create_subprocess_*`, but downstream code that copies `connection` for logging or cache keys may stringify a `Path` differently than a `str`, and the env values are already strings. Easy consistency win.

### 3. Update `backend/CLAUDE.md`

Per the top-of-file policy:

> **CRITICAL: Always update README.md and CLAUDE.md after every code change** … Update `CLAUDE.md` for development changes (architecture, commands, workflows, internal systems)

The MCP System section currently lists only "Lazy initialization", "Cache invalidation", "Transports", "OAuth", "Runtime updates" — none of the new behavior is documented. Please add a short paragraph covering at least:

- The cwd/temp pinning contract for stdio transports (and that SSE/HTTP skip it entirely).
- The host→virtual translation behavior, scoped to files already inside the thread's tree (`/mnt/user-data/...`), no copying.
- The session-pool scope key change from `thread_id` to `user_id:thread_id` (relevant for anyone tracing session reuse).

### 4. Add at least one debug log on rewrite

When a path is translated, there's no log entry. Debugging "why doesn't my Playwright screenshot show up?" is harder without a `logger.debug("MCP path rewrite: %s -> %s", source, virtual)` in `_local_uri_to_virtual_path`. Same for the bare-filename case — operators cannot tell whether the rewrite fired and which candidate was chosen. One debug log on each rewrite path (and one on "regex matched but resolved outside tree" / "matched but no unique snapshot candidate") would pay for itself the first time someone files a bug here.

## Follow-ups (do not need to block this PR)

### Recursive workspace snapshot on every stdio MCP call

`backend/packages/harness/deerflow/mcp/tools.py:135-145` — `_snapshot_workspace_files` `rglob("*")`s the entire thread workspace, twice per call (the after-walk is gated on text content, but the before-walk runs unconditionally in `_prepare_stdio_workspace`). On a research thread that has accumulated a checked-out repo, `node_modules`, browser caches, or downloaded artifacts, every stdio MCP invocation now pays an O(N) walk — and Playwright sessions can be chatty.

Possible mitigations: scope the walk to known output prefixes (`.mcp/tmp/`, configured `--output-dir`, operator-provided hint), and only fall back to a full walk when nothing is found there; or maintain the snapshot per session and update it incrementally. Even an early-out by file count would protect against pathological workspaces.

### `_snapshot_workspace_files` swallows mid-walk OSError silently

`backend/packages/harness/deerflow/mcp/tools.py:144` — if `rglob` raises mid-iteration (permission denied on a subdir), the partial snapshot is returned, and the after-walk diff will look like every unwalked file is "new" → false-positive bare-filename rewrites. Consider logging the OSError and either propagating to disable rewriting for this call, or marking the snapshot as incomplete so the caller can downgrade.

### Symlinks in the snapshot are inefficient

`is_file()` follows symlinks, so a dangling-symlink workspace adds the symlink to the snapshot, the diff flags it as "changed" each call, and `_local_uri_to_virtual_path` then rejects it. Net behavior is correct but wastes work. Consider `path.is_file() and not path.is_symlink()` at snapshot time.

### Regex assumes POSIX paths

`backend/packages/harness/deerflow/mcp/tools.py:47` — both branches require `/`. A native-Windows MCP server returning `C:\...` won't have its references rewritten. Probably out of scope (DeerFlow's deployments are Linux/macOS), but a brief comment stating the assumption would help anyone porting later.

## Backward compatibility

The unconditional `cwd` overwrite (issue #2 above) is the main backcompat concern: stdio MCP servers that previously resolved config files or sqlite DBs relative to the gateway's cwd will break. Switching to `setdefault` solves this and is the cleanest fix. Worth a note in the CHANGELOG/release notes either way.

The session-pool scope-key change means existing pooled sessions are not adopted across deploy — cold-start cost only, acceptable.

## Test coverage

Excellent for the rewriting logic (URI parsing, virtual translation, free-text rewriting, snapshots, bare-filename correlation, security boundaries). Good for the session-pool scope-key change. Missing only:

- A test for `cwd` being preserved when operator supplied one (please add as part of fix #2 above).
- The partial-snapshot edge case (issue #7 / follow-up territory).

## Recommendation

**Request changes** on the four items above (`chmod`, `cwd setdefault`, CLAUDE.md, debug log). The follow-ups are tracked for a later PR. Core design is sound and ready to merge once the four small changes land.

18062706139fcz:
Addressed the latest requested changes in `b01c3fb5`:

- Changed the stdio MCP temp directory permission from `0o777` to `0o700`.
- Preserved operator-provided `cwd` for stdio MCP servers and kept injected/default `cwd` values as strings.
- Added debug logs for host→virtual path rewrites, outside-tree skips, and bare-filename rewrite decisions.
- Updated `backend/CLAUDE.md` with the stdio cwd/temp pinning contract, host→virtual translation behavior, and `user_id:thread_id` session scope.
- Added regression coverage for explicit `cwd` preservation and temp-dir permissions.

Validation:

```bash
cd backend && uv run pytest tests/test_mcp_session_pool.py tests/test_mcp_file_migration.py tests/test_mcp_sync_wrapper.py -q
# 98 passed, 1 warning

cd backend && uvx ruff check packages/harness/deerflow/mcp/tools.py tests/test_mcp_session_pool.py tests/test_mcp_file_migration.py tests/test_mcp_sync_wrapper.py
# All checks passed!

cd backend && uvx ruff format --check packages/harness/deerflow/mcp/tools.py tests/test_mcp_session_pool.py tests/test_mcp_file_migration.py tests/test_mcp_sync_wrapper.py
# 4 files already formatted

# 16 passed, 1 warning
```

willem-bd (approved):

WillemJiang (approved):

Issue #3222, opened 2026-05-25
[Design] Model-bound compactable tool payload policy for long-running agents


kibabsquirrel:
# [Design] Model-bound compactable tool payload policy for long-running agents

Status: draft / design topic

Intent: define a practical first boundary for large tool payload governance in DeerFlow: what representation of tool payloads should be sent to models in long-running agent runs.

Longer term, this may grow into a broader RFC for large tool payload governance across persistence, replay, artifact references, frontend/trace representation, and task-state convergence. For now, I'm intentionally scoping this topic to the first practical layer: model-bound projections.

<details>
<summary>中文摘要</summary>

本文提议先定义一个 **model-bound compactable tool payload policy**：只讨论发送给模型的工具载荷视图，不在第一阶段处理 checkpoint、trace、frontend、ArtifactRef 或 task-state convergence。

核心动机是：`write_file.args.content` 代表大型工具输入，`web_search` / `web_fetch` 代表大型工具输出。两者都可能在后续 model request、summary prompt、subagent request 或 provider serialization 中造成 context/token 压力。

第一阶段建议只覆盖：

- lead-agent model request；
- summary-bound prompt；
- subagent model request；
- provider raw metadata / serialization consistency。

其他 lifecycle surface，例如 generation-time artifact writing、checkpoint persistence、frontend/trace representation、task convergence，都是相关问题，但应作为后续独立设计处理。

</details>

---

## Summary

Large tool payloads can enter model-bound context without an explicit budget or representation policy.

The motivating input-side case is completed `write_file(content=...)` replay: after a large file write has completed, the original large `content` argument may remain in historical tool-call arguments and be sent again in later model requests.

The corresponding output-side pressure appears in large tool results, especially `web_search` / `web_fetch` outputs in research and report-generation flows.

This design topic proposes a focused first boundary:

> Define a model-bound compactable tool payload policy for large tool inputs, outputs, and diagnostic observations.

The first phase should cover only model-bound projections:

- lead-agent model requests;
- summary-bound prompts;
- subagent model requests;
- provider serialization / raw tool-call metadata consistency.

It should not decide persistence, replay, frontend event, trace, ArtifactRef, or task-state semantics in the first phase.

---

## Problem

Today, different tool payloads can enter model-bound context with inconsistent handling:

- large tool inputs, such as `write_file.args.content`;
- large tool outputs, such as `web_search` / `web_fetch` results;
- diagnostic outputs, such as long `bash` output, logs, or error observations;
[...]

Issue #3164, opened 2026-05-22
Tool-returned Error strings bypass centralized tool error handling and lose structured failure semantics


tomsen02:
## Problem

I am opening this as a design question rather than a bug report, because part of the current behavior may be intentional for agent self-recovery.

DeerFlow has centralized tool exception handling via `ToolErrorHandlingMiddleware`. It converts uncaught tool exceptions into `ToolMessage(status="error")`, allowing the conversation to continue without aborting the whole run.

However, many built-in/community tools catch failures internally and return plain strings such as `"Error: ..."`. These returned strings are then wrapped by LangChain as normal tool outputs, usually as `ToolMessage(status="success")`. As a result, these failures do not reach `ToolErrorHandlingMiddleware`, and downstream components cannot reliably distinguish recoverable task-level misses from provider/config/auth/transient/system failures.

## Code Observations

For example, in the Jina `web_fetch` path:

- `deerflow/community/jina_ai/jina_client.py`
  - non-200 responses return strings like `Error: Jina API returned status ...`
- `deerflow/community/jina_ai/tools.py`
  - if crawl returns an `Error:` string, `web_fetch` returns that string directly

Similar patterns also exist in several sandbox tools, for example:

- `read_file`
- `write_file`
- `bash`
- `ls`
- `glob`
- `grep`
- `str_replace`

Many of these tools return `"Error: ..."` strings for failures instead of exposing structured error information or letting system/provider failures reach the centralized tool error handling layer.

## Why This Matters

Returning failure text to the model is reasonable for some recoverable task-level cases, such as:

- file not found
- no search results
- string to replace not found

The model can read those results and try another path, query, or strategy.

The concern is that the same `"Error: ..."` channel can also carry more systemic failures, such as:

- invalid API key / 401
- provider configuration problems
- timeout / rate limit
- permission errors
- unexpected internal errors

When all of these are represented as plain successful tool outputs with error text, DeerFlow loses structured failure semantics. This makes it harder to:

- distinguish recoverable misses from auth/config/transient failures
- apply consistent retry/backoff for transient tool failures
- stop early on repeated auth/config failures and tell the user what to fix
- display tool failures clearly in the frontend
- record tool failure counts or error types in run metadata
- make loop/progress detection reason about repeated failures or stagnation from tool results

## Observed Local Behavior

In a local run using Jina as the `web_fetch` provider, an invalid or placeholder `JINA_API_KEY` caused repeated `web_fetch` results like:

```text
[...]

PR #3601, opened 2026-06-15
feat(middleware): add structured tool result meta and tool-progress state machine

whhe:
## Summary

Refs #3177 — implements the structured tool-result semantics and per-tool
stagnation guard proposed in that RFC as a first iteration.
`ToolProgressMiddleware` tracks per-(thread_id, tool_name) consecutive
"no-new-info" results with three escalation paths driven by
`deerflow_tool_meta.recoverable_by_model`: recoverable errors (`no_results`,
`not_found`, `permission`, Jaccard-duplicate success) stay in WARNED so the
model can retry with different parameters; non-recoverable non-stop errors
(`rate_limited`, `transient`) escalate WARNED → BLOCKED after
`warn_escalation_count` more problems; non-recoverable stop errors (`auth`,
`config`, `internal`) go directly to BLOCKED. This addresses the runaway-loop
failure mode without blocking legitimate tool reuse with fresh parameters.

Note: the RFC also envisions topic/subtask-level progress state
(not_started → searching → evidence_found → enough_evidence → summarized),
which requires a topic-boundary recognition mechanism not in scope for this
PR — that layer warrants a separate design discussion in #3177.

Also introduces `deerflow_tool_meta` — a structured per-result signal stamped
on every `ToolMessage` by `ToolErrorHandlingMiddleware` — as the signal source
for the state machine. This classification layer (`status`, `error_type`,
`recoverable_by_model`, `recommended_next_action`) implements the structured
tool-result semantics proposed in Refs #3164 and discussed in the #3177
comments as a prerequisite for reliable state-machine transitions.
`deerflow_tool_meta` also provides result-quality signals that a future payload
compaction policy (Refs #3222) can consume to guide per-result compaction
decisions.

Note: Refs #3205 introduces a similar `deerflow_tool_error` key for error
paths. If both PRs merge, the two keys coexist on different field names;
`ToolProgressMiddleware` reads only `deerflow_tool_meta`. Maintainers may want
to coordinate between the two approaches before merging.

## Changes

- **`tool_result_meta.py`** (new): `ToolResultMeta` dataclass with `status`,
  `error_type`, `recoverable_by_model`, `recommended_next_action`, and `source`
  fields (`retryable` removed — no consumer, overlaps semantically with
  `recoverable_by_model`); `normalize_tool_result` classifies every
  `ToolMessage`; `stamp_exception_meta` always overwrites tool-return stamps
  (exception classification is authoritative); `_extract_json_error_text`
  isolates the JSON `error` field — JSON dicts without an `error` key resolve
  to `unknown` instead of classifying from incidental field values (e.g.
  `{"user_id": 401}` no longer triggers auth→stop); non-string `error` fields
  (dict/list) are serialized via `json.dumps` rather than `str()` to avoid repr
  collisions with keyword rules; numeric HTTP codes (`401`, `403`, `404`, `500`)
  anchored to word boundaries to prevent substring hits on `"500ms"` or
  `"4010 rows"`; `"no results found"` / `"no content found"` /
  `"no images found"` added to `_PARTIAL_MARKERS` so tools returning these with
  `status="success"` are still detected by stagnation tracking; `partial_success`
  is now gated only on `_PARTIAL_MARKERS` (the `<80`-char length gate removed —
  terse mutation confirmations like `"OK"` are full successes); `_SEMANTIC_ZERO_ERROR_STRINGS`
  frozenset suppresses false positives from tools that signal success via
  `{"error":"none"}` / `{"error":"null"}` / `{"error":"ok"}`; `"no api key
  configured"` classifies as `config` (not `auth`); `_NUMERIC_KW_RE`
  pre-compiled at module load from `_ERROR_RULES`; `content.lower()` extracted
  once before the `_PARTIAL_MARKERS` check so it is computed at most once per
  call (was up to 7× inside the generator, one per marker)
- **`tool_error_handling_middleware.py`**: wire `stamp_exception_meta` into
  `_build_error_message` and `normalize_tool_result` into both
  `wrap_tool_call` / `awrap_tool_call` success paths so every result carries
  `deerflow_tool_meta` regardless of path; `isinstance`-based ordering guard at
  end of `_build_runtime_middlewares` raises `RuntimeError` at build time if
  `ToolProgressMiddleware` appears after `ToolErrorHandlingMiddleware`;
  `logger.warning` emitted when `deerflow_tool_meta` is missing for a
  non-exempt tool to surface misordering at runtime
- **`tool_progress_middleware.py`** (new): `ToolProgressMiddleware` with
  per-(thread_id, tool_name) `ToolPhaseState`; Jaccard-based near-duplicate
  detection (word_set capped at 8192 chars, computed lazily only for
  `status="success"` results); hint injection as `HumanMessage` before the next
  model call; three-path escalation (WARNED terminal for
  `recoverable_by_model=True`, WARNED→BLOCKED for non-recoverable non-stop,
  immediate BLOCKED for auth/config/internal with `consecutive_problems`
  correctly incremented on all paths); LRU-bounded thread store
  (`max_tracked_threads` default 100) with `_get_block_reason` on the read path
  no longer calling `move_to_end` (prevented blocked threads from being
  LRU-evicted); `_reset_run_states` (formerly `_reset_blocked_states`) called in `before_agent`
          resets all tracked tools unconditionally at run start — BLOCKED, WARNED, and ACTIVE —
          clearing `consecutive_problems`, `block_reason`, and `recent_word_sets`; ACTIVE tools
          with sub-threshold problem counters or cached word-sets from a prior run no longer
          bleed into the next run (commit d3d3920f);
  `_queue_assessment` guards against phantom `_pending` entries for threads
  evicted from `_phase_states`; `task` added to default `exempt_tools`
  (delegation primitive, not a search/query tool); structured state-transition
  logging (ACTIVE/WARNED/BLOCKED transitions, intercepts, hint injection);
  `_MAX_PENDING_PER_RUN = 3` per run; module docstring documents division of
  labor with `LoopDetectionMiddleware` (position 19): ToolProgressMiddleware is
  a result-quality guard (per-tool BLOCK, fires after execution),
  LoopDetectionMiddleware is a call-pattern guard (whole-turn hard-stop, fires
  before execution) — they operate on independent state and cannot double-stop;
  `threading.Lock` comment explains why `asyncio.Lock` is not used (critical
  sections are short dict ops; `asyncio.Lock` would not protect the sync
  `wrap_tool_call` path used by subagent executor threads)
- **`tool_progress_config.py`** (new): `ToolProgressConfig` Pydantic model
  exposing all thresholds (`stagnation_threshold`, `warn_escalation_count`,
  `jaccard_similarity_threshold`, `min_word_count_for_similarity`,
  `exempt_tools`, `max_tracked_threads`); disabled by default (`enabled: false`);
  `task` added to default `exempt_tools` alongside `ask_clarification`,
  `write_todos`, `present_files`
- **`app_config.py`**: add `tool_progress: ToolProgressConfig` field
- **`config.example.yaml`**: document `tool_progress` section with defaults
  and three-path state machine description; `config_version` bumped to 16
  (upstream at 15; `tool_progress` section is additive)
- **`backend/CLAUDE.md`**: middleware chain table updated — entry 8
  (ToolProgressMiddleware) gains a division-of-labor paragraph explaining
  coexistence with LoopDetectionMiddleware; entry 9 (ToolErrorHandlingMiddleware)
  corrects the `deerflow_tool_meta` field list (removes stale `retryable`,
  adds `recoverable_by_model` and `source`)
- **`_build_runtime_middlewares`**: insert `ToolProgressMiddleware` at lower
  index than `ToolErrorHandlingMiddleware` (outer wrapper position), after the
  already-merged `ToolOutputBudgetMiddleware`; `ReadBeforeWriteMiddleware` placed
  before `ToolProgressMiddleware` as the outermost write gate so blocked writes
  return immediately without entering the ToolProgress/ToolErrorHandling handler chain
- **`read_before_write_middleware.py`**: call `normalize_tool_result` on blocked
  `ToolMessage`s so they carry `deerflow_tool_meta` (`recoverable_by_model=True`)
  even though blocked writes bypass `ToolErrorHandlingMiddleware`; fixes a merge
  conflict resolution that had accidentally placed `ReadBeforeWriteMiddleware`
  after `ToolErrorHandlingMiddleware` (inner), reversing the original intent
- **Tests**: `test_tool_result_meta.py` (classification paths,
  `stamp_exception_meta`, JSON error extraction, semantic-zero strings (test_normalize_json_semantic_zero_error_string_not_treated_as_error now uses json.dumps to produce valid JSON so the parser actually reaches _SEMANTIC_ZERO_ERROR_STRINGS, commit d3d3920f),
  word-boundary negative cases for `500ms`/`4010`/`401A`,
  JSON-without-error-key resolves to `unknown`, dict error field via
  `json.dumps`, `no results found` success → `partial_success`, terse-success
  not partial); `test_tool_progress_middleware.py` (async paths, Jaccard
  detection, LRU eviction, hint injection, recoverable-stays-WARNED,
  BLOCKED/WARNED/ACTIVE run-scoped reset (test_before_agent_resets_active_state_consecutive_problems_and_word_sets locks the ACTIVE-phase case), `consecutive_problems` correct on
  immediate-block, warning log for missing meta,
  `test_tool_progress_and_loop_detection_coexist_without_interfering` — drives
  both middlewares to WARNED simultaneously, verifies independent state, no
  cross-contamination, independent hint queues); `test_tool_error_handling_middleware.py`
  (middleware ordering guard raises `RuntimeError` for wrong order,
  `isinstance`-based detection, meta stamping on exception)


backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,405 @@
...
+    ) -> ToolMessage | Command:
+        tool_name = str(request.tool_call.get("name", ""))
+        if not tool_name or tool_name in self._exempt_tools:
+            return handler(request)
+        runtime = getattr(request, "runtime", None)
+        if runtime is None:
+            return handler(request)
+        block_reason = self._get_block_reason(runtime, tool_name)
+        if block_reason:
+            return self._make_blocked_message(request, tool_name, block_reason)
```
fancyboi999, line 504: Worth weighing: `blocked` is terminal per `(thread_id, tool_name)`. Once a tool is blocked, every later call short-circuits here and `_update_state_from_result` never runs for it again — so there's no path back to `active`, despite the module docstring's "any problem-free call resets consecutive_problems=0 and reverts to ACTIVE" (that reset only fires while the tool still executes, i.e. active/warned). And since the key is the tool *name*, not the call arguments, a later genuinely-different invocation of the same tool (search with a fresh query, fetch of a new URL) is also blocked for the rest of the thread. For a pure runaway-loop guard that's defensible — stopping the loop is the point — but it can over-fire on tools the model legitimately reuses with new inputs. Might be worth a recovery path (decay after N model turns, or reset when the argument set differs enough) or at least tightening the docstring to say BLOCKED is terminal for the thread. Non-blocking, and it's off by default — just flagging.

fancyboi999 (commented):
Went through this carefully — the state machine, the middleware wiring, and the locking all hold up, and the 69 tests pass locally. Things I verified specifically:

- **Middleware ordering is correct** (the part most likely to silently break a feature like this). `ToolProgressMiddleware` is appended *before* `ToolErrorHandlingMiddleware` → lower index, and LangChain composes "first in list = outermost" (`langchain/agents/factory.py:575`, `middleware/types.py:741`). So ToolError (inner) stamps `deerflow_tool_meta` on the way out and ToolProgress (outer) reads the already-stamped result in `_update_state_from_result`. The `test_tool_progress_middleware_is_outer_relative_to_error_handling` invariant locks it — easy to get backwards, good that it's pinned.
- **The short-circuit is real**: when phase is `blocked`, `wrap_tool_call` returns `_make_blocked_message` without invoking the handler, so the tool genuinely doesn't run (`tool_progress_middleware.py:344`/`361`).
- **State store is properly locked**: `_phase_states` mutations (`move_to_end`/`popitem`/`__setitem__`) all run under `self._lock` via `_get_block_reason` and `_update_state_from_result`, so the across-threads-shared instance is safe.
- Disabled by default (`tool_progress.enabled: false`) — no behavior change unless opted in.

One design note inline. Your own #3205 `deerflow_tool_error` vs `deerflow_tool_meta` coexistence flag is a fair thing for maintainers to settle before both land. Reads solid otherwise.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
whhe: The latest commit addresses this directly by splitting the escalation logic on `recoverable_by_model`:

**`recoverable_by_model=True` errors** (`no_results`, `not_found`, `permission`, Jaccard-duplicate success): `WARNED` is now **terminal** — the tool never reaches `BLOCKED`. Every subsequent call still executes, so a search with a fresh query or a fetch of a new URL will run normally. The stagnation hint is re-injected on each problem call to keep nudging the model toward a strategy change.

**`recoverable_by_model=False, action≠stop`** (`rate_limited`, `transient`): still escalates `WARNED` → `BLOCKED` after `warn_escalation_count` more calls. The tool itself is unavailable (rate-limited, network failure), not just the parameters, so blocking is appropriate here — a different query won't help.

**`recoverable_by_model=False, action=stop`** (`auth`, `config`, `internal`): unchanged, still immediately `BLOCKED`.

The module docstring was updated to document all three paths, and the misleading "reverts to ACTIVE" note no longer applies to the recoverable path.

backend/packages/harness/deerflow/agents/middlewares/tool_result_meta.py
```diff
@@ -0,0 +1,176 @@
...
+        json_error = _extract_json_error_text(content)
+        attrs = _classify_error_text(json_error if json_error is not None else content)
+        meta = _make_meta(status="error", source="tool_return", **attrs)
+    elif content.startswith(_ERROR_PREFIX):
+        attrs = _classify_error_text(content[len(_ERROR_PREFIX) :])
+        meta = _make_meta(status="error", source="tool_return", **attrs)
+    elif (json_error := _extract_json_error_text(content)) is not None:
+        attrs = _classify_error_text(json_error)
+        meta = _make_meta(status="error", source="tool_return", **attrs)
+    elif any(m in content.lower() for m in _PARTIAL_MARKERS) or (0 < len(content) < _MIN_SUBSTANTIAL_CONTENT):
```
willem-bd: **Short successful results become `partial_success`.** Several built-in tools return short strings on success: `write_file`/`str_replace` return literal `"OK"` (~2 chars) and `view_image` returns `"Successfully read image"` (22 chars) — none are in `exempt_tools`. With `tool_progress.enabled`, every such success is stamped `status="partial_success"` here, which `ToolProgressMiddleware` counts as a "problem". After `stagnation_threshold` (default 3) writes/edits in a thread the tool flips to WARNED and a spurious `[PROGRESS HINT] ... returned incomplete results ... rephrasing your search query` `HumanMessage` is injected — and "rewrite your search query" is nonsensical for file tools.

The `< 80` length gate treats terseness as incompleteness. Consider gating `partial_success` behind the explicit `_PARTIAL_MARKERS` only (not length), or adding the mutation/confirmation tools (`write_file`, `str_replace`, `view_image`, `task`) to the exempt set like `ask_clarification`/`write_todos`/`present_files` already are.

backend/packages/harness/deerflow/agents/middlewares/tool_result_meta.py
```diff
@@ -0,0 +1,176 @@
...
+    (
+        ["no results found", "no content found", "no images found", "no results"],
+        {"error_type": "no_results", "retryable": False, "recoverable_by_model": True, "recommended_next_action": "rewrite_query"},
+    ),
+    (
+        ["not found", "no such file", "does not exist", "404"],
+        {"error_type": "not_found", "retryable": False, "recoverable_by_model": True, "recommended_next_action": "rewrite_query"},
+    ),
+    (
+        ["unexpected error", "internal error", "500"],
```
willem-bd, line 73: **Bare-token substring matching can trigger a false immediate BLOCK.** `_classify_error_text` matches `"500"` / `"401"` / `"403"` / `"disabled"` etc. as substrings of arbitrary error text. Several resolve to `recoverable_by_model=False, recommended_next_action="stop"` (internal, auth, config), and `_assess_and_transition`'s immediate-block branch hard-BLOCKs the tool on the **first** occurrence for those.

So an error whose text merely contains `"500"` (e.g. `"error 500 in logs"`, `"took 500ms"`) or `"disabled"` permanently blocks the tool for the (thread, tool) until the thread is evicted — even though the error was neither auth nor internal. Anchoring the numeric codes (word boundaries / `http 500`) or requiring whole-token matches would avoid the incidental collisions.

willem-bd (commented):
Thanks for this — the state-machine design and test coverage (concurrency / LRU / semantic-zero cases) are solid. I verified the framework symbols, `ModelRequest.override`, and the "first-in-list = outermost" composition claim all hold.

The highest-impact notes are about the **classification heuristics in `tool_result_meta.py`**: the `< 80`-char length gate and the bare-token substring matching misclassify legitimate results, and since `ToolProgressMiddleware` consumes them they can spuriously WARN or hard-block tools that are working correctly (e.g. `write_file`/`str_replace` return `"OK"` → `partial_success` → guaranteed WARNED after 3 writes; an error mentioning `"500"` → immediate permanent BLOCK).

Findings (most severe first):
1. Short successful results (`"OK"`, `"Successfully read image"`) → `partial_success` → spurious WARNED/hints — `tool_result_meta.py:157`
2. Substring match on `"500"`/`"disabled"`/`"401"` → false immediate BLOCK — `tool_result_meta.py:64`
3. `task` delegation tool not in `exempt_tools` — `tool_progress_config.py:37`
4. `_get_block_reason` LRU bump on read pins blocked threads, evicts active ones — `tool_progress_middleware.py:245`
5. BLOCKED persists across runs but WARN hints are dropped by `_clear_stale_pending` — `tool_progress_middleware.py:387`
6. `config_version` not bumped despite the schema addition — `config.example.yaml`
7. `word_set(content)` computed unconditionally (O(n) on every result) — `tool_progress_middleware.py:343`
8. `recent_word_sets` retains 3 full frozensets of large content — `tool_progress_middleware.py:348`
9. `retryable`/`source` stamped but unread — `tool_result_meta.py:28`
10. Outer/inner ordering invariant enforced only by comment + test — `tool_error_handling_middleware.py:192`

Ruled out after checking: the `threading.Lock` use is fine (await-free critical sections, separate instances per agent, matches the existing `LoopDetectionMiddleware` pattern); `stamp_exception_meta` can't escape the try/except (its input is a guaranteed `str`); and a tool can't both return a stamped message and raise, so the meta-overwrite concern doesn't apply.

backend/packages/harness/deerflow/config/tool_progress_config.py
```diff
@@ -0,0 +1,45 @@
...
+        default=0.8,
+        ge=0.0,
+        le=1.0,
+        description="Word-set Jaccard similarity threshold for near-duplicate result detection",
+    )
+    min_word_count_for_similarity: int = Field(
+        default=10,
+        description="Minimum unique word count to apply Jaccard check; shorter content skips near-duplicate detection entirely",
+    )
+    exempt_tools: set[str] = Field(
```
willem-bd, line 37: **`task` (subagent delegation) isn't in the default `exempt_tools`.** `task` is the lead agent's primary delegation primitive and returns the subagent's summary. Dispatching several similar research subtasks can produce Jaccard-near-duplicate summaries (≥0.8 over ≥10 words), so after `stagnation_threshold` the `task` tool gets WARNED with "returning duplicate results" and the hint is re-injected on every subsequent dispatch. A terse subagent answer (<80 chars) also trips the `partial_success` path. Like `ask_clarification`/`write_todos`/`present_files`, `task` is control-flow tooling where stagnation heuristics don't really apply — worth adding to the default exempt set.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,489 @@
...
+    def _set_state(self, thread_id: str, tool_name: str, state: ToolPhaseState) -> None:
+        self._phase_states[thread_id][tool_name] = state
+
+    def _get_block_reason(self, runtime: Runtime, tool_name: str) -> str | None:
+        thread_id = self._thread_id(runtime)
+        with self._lock:
+            thread_tools = self._phase_states.get(thread_id)
+            if thread_tools is None:
+                return None
+            self._phase_states.move_to_end(thread_id)
```
willem-bd: **`move_to_end` on the read path pins blocked threads, defeating LRU.** `_get_block_reason` runs before the handler on *every* `wrap_tool_call` and calls `move_to_end(thread_id)` whenever the thread exists. A thread with a BLOCKED tool keeps being polled (the model keeps retrying), so it's refreshed to most-recently-used on every attempt and can never be evicted. Under sustained load the blocked threads occupy the LRU slots and `max_tracked_threads` ends up evicting *active/healthy* threads instead — silently resetting their `consecutive_problems` mid-run. Consider bumping recency only on the write path (in `_get_state`), not on this read-only check.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,489 @@
...
+            queue = self._pending[key]
+            if len(queue) < _MAX_PENDING_PER_RUN:
+                queue.append(text)
+
+    def _drain_pending(self, runtime: Runtime) -> list[str]:
+        key = self._pending_key(runtime)
+        with self._lock:
+            return self._pending.pop(key, [])
+
+    def _clear_stale_pending(self, runtime: Runtime) -> None:
```
willem-bd, line 438: **Asymmetric persistence: BLOCKED survives across runs but WARN hints are dropped.** `_clear_stale_pending` (run in `before_agent`) deletes prior-run pending hints, while `_phase_states` — including a BLOCKED reached in a previous run — is never cleared. So if run R1 queues a WARN hint but ends without a model call (hit max iterations, or a `ClarificationMiddleware` `Command(goto=END)` fired), the hint is discarded on R2's `before_agent`, yet the tool stays hard-blocked in R2 from R1's history. The warning that exists *to let the model change strategy* is lost while the block decision it was meant to avert persists. Worth either re-queuing an undrained hint at run start, or scoping BLOCKED to the run.

config.example.yaml
```diff
@@ -744,6 +744,30 @@ loop_detection:
   #     warn: 150
   #     hard_limit: 300
 
+# ============================================================================
+# Tool Progress State Machine Configuration (RFC #3177)
```
willem-bd, line 938: **`config_version` not bumped.** `backend/CLAUDE.md` (Configuration System) states: *"When changing the config schema, bump `config_version` in `config.example.yaml`."* This PR adds the `tool_progress` section here and a `ToolProgressConfig` field on `AppConfig`, but `config_version` is unchanged (no `config_version` token appears in the diff). Existing users won't get the new section surfaced by the staleness comparison / `make config-upgrade`.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,489 @@
...
+
+        # Immediately block on unrecoverable stop signals (auth, config, internal).
+        if not meta.recoverable_by_model and meta.recommended_next_action == "stop":
+            return replace(
+                state,
+                phase="blocked",
+                block_reason=_block_reason(meta),
+            ), None
+
+        ws = word_set(content)
```
willem-bd: **`word_set(content)` is computed for every result, but only the success branch reads it.** This is an O(n) `re.findall` + `frozenset` build over the entire content, paid on every tool result — including errors and `partial_success` where `is_problem` is short-circuited by the `or` left operand and the word set is never used. For large fetched content this is avoidable work on the post-tool hot path. Lazy option: `ws = word_set(content) if meta.status == "success" else None`.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,489 @@
...
+        is_problem = meta.status in ("error", "partial_success") or (meta.status == "success" and is_near_duplicate(ws, state.recent_word_sets, self._jaccard_threshold, self._min_words))
+
+        if not is_problem:
+            # Good result: reset consecutive count, return to active.
+            new_recent = (*state.recent_word_sets, ws)[-3:]
```
willem-bd, line 394: **`recent_word_sets` retains up to 3 full frozensets of each result per (thread, tool).** For large results (~10KB → thousands of strings each), three frozensets are held in `ToolPhaseState` for the thread's lifetime (× `max_tracked_threads` × tools). The retained data only feeds the Jaccard check; a bounded/hashable representation (a hash + length, or a capped sample) would bound the retention while preserving near-duplicate detection.

backend/packages/harness/deerflow/agents/middlewares/tool_result_meta.py
```diff
@@ -0,0 +1,176 @@
...
+_ERROR_PREFIX = "Error:"
+_PARTIAL_MARKERS = ("partial results", "limited results", "truncated", "results may be incomplete")
+_MIN_SUBSTANTIAL_CONTENT = 80
+
+
+@dataclass(frozen=True, slots=True)
+class ToolResultMeta:
+    status: Literal["success", "error", "partial_success"]
+    error_type: str | None
+    retryable: bool
```
willem-bd: **`retryable` (and `source`) are stamped on every result but read by no production code in this PR.** `ToolProgressMiddleware` consumes `status`/`error_type`/`recoverable_by_model`/`recommended_next_action`, but not `retryable` or `source`. Two carried fields with no consumer risk rotting silently — if they're forward-looking (e.g. for the compaction policy in #3222), worth gating behind that consumer or a TODO rather than stamping now.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -183,7 +188,19 @@ def _build_runtime_middlewares(
     from deerflow.agents.middlewares.sandbox_audit_middleware import SandboxAuditMiddleware
 
     middlewares.append(SandboxAuditMiddleware())
+
+    # ToolProgressMiddleware must be outer (lower index) so its wrap_tool_call handler
```
willem-bd, line 231: **The outer-vs-inner ordering invariant is enforced only by this comment + an index-comparison test, not by construction.** If a future middleware is inserted between these two `append()` calls, `ToolProgressMiddleware` becomes inner: its `_update_state_from_result` runs before `normalize_tool_result` stamps `TOOL_META_KEY`, `_parse_tool_meta` returns `None`, the `if meta is None: return result` early-exit fires, and the whole state machine silently no-ops — with no log or assertion that the meta it expected was missing. A runtime warn-log when meta is absent on a non-exempt, non-passthrough result would make a future misordering loud instead of silent.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
whhe: Fixed in commit 2eaabd5f. We scoped both BLOCKED and WARNED states to a single run: `_reset_blocked_states` is called in `before_agent` and now resets tools in either `blocked` or `warned` phase back to `active`, clearing `consecutive_problems` and `recent_word_sets`. This means a WARNED tool with accumulated counts from a prior run won't silently reach BLOCK in the next run before the model has seen any hint.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
whhe: Fixed in commit 2eaabd5f. Rather than a hash-based approach, we capped the content fed to `word_set()` at `_MAX_CONTENT_FOR_WORDSET = 8192` chars before the `re.findall` call. This bounds both the per-call CPU cost and the size of each stored frozenset to a predictable maximum, while keeping the Jaccard logic unchanged. The cap is applied at classification time, so no changes to `ToolPhaseState` storage format were needed.

backend/packages/harness/deerflow/agents/middlewares/tool_result_meta.py
whhe: Fixed in commit 2eaabd5f. Removed `retryable` from `ToolResultMeta` entirely — it had no consumer and overlapped semantically with `recoverable_by_model`. Kept `source` as it has diagnostic value (distinguishes exception-stamped vs tool-return vs content-analysis classifications). All `_ERROR_RULES` entries and `_make_meta` signature updated accordingly.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
whhe: Fixed in commit 2eaabd5f. Added both layers: (1) a `logger.warning` in `_update_state_from_result` when `deerflow_tool_meta` is missing for a non-exempt tool — surfaces misordering at runtime; (2) an `isinstance`-based `RuntimeError` guard at the end of `_build_runtime_middlewares` that fires at build time if `ToolProgressMiddleware` ends up at a higher index than `ToolErrorHandlingMiddleware`. The guard uses `isinstance` (not `type().__name__`) so subclasses are also covered. Both code paths have dedicated tests.

backend/packages/harness/deerflow/agents/middlewares/tool_result_meta.py
```diff
@@ -0,0 +1,209 @@
...
+                is_json_dict = False
+            attrs = {**_UNKNOWN_ERROR} if is_json_dict else _classify_error_text(content)
+        meta = _make_meta(status="error", source="tool_return", **attrs)
+    elif content.startswith(_ERROR_PREFIX):
+        attrs = _classify_error_text(content[len(_ERROR_PREFIX) :])
+        meta = _make_meta(status="error", source="tool_return", **attrs)
+    elif (json_error := _extract_json_error_text(content)) is not None:
+        attrs = _classify_error_text(json_error)
+        meta = _make_meta(status="error", source="tool_return", **attrs)
+    elif any(m in content.lower() for m in _PARTIAL_MARKERS):
```
willem-bd: **Performance: `content.lower()` is re-evaluated for every marker on each success result.** (perf, medium)

`any(m in content.lower() for m in _PARTIAL_MARKERS)` sits inside a generator, so `content.lower()` runs once per marker — up to 8× — and `any()` only short-circuits early when a marker actually matches, which is rare for normal success bodies. This branch is reached for **every successful tool result**, including large web-search / page payloads, so a ~50 KB result allocates ~400 KB of transient strings per call.

Suggested fix:
```python
lowered = content.lower()
elif any(m in lowered for m in _PARTIAL_MARKERS):
```
Optionally also gate the whole heuristic block on `len(content) < some_bound`, since multi-KB success bodies almost never contain `"partial results"`.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,543 @@
...
+        "Tool has not produced new information after multiple attempts — summarize and move on.",
+    )
+
+
+# ---------------------------------------------------------------------------
+# Middleware
+
+
+class ToolProgressMiddleware(AgentMiddleware[AgentState]):
+    """State-machine-based tool stagnation guard (RFC #3177)."""
```
willem-bd, line 195: **Architecture: overlap with the existing `LoopDetectionMiddleware` is unaddressed.** (architecture, medium)

The lead-agent chain already has `LoopDetectionMiddleware` (item 18 in `backend/CLAUDE.md`) that "detects repeated tool-call loops" and force-stops the run by stripping `tool_calls`. This PR adds a second, finer-grained stagnation detector with its own WARN/BLOCK semantics. Neither the PR body, this docstring, nor the `CLAUDE.md` middleware table explains the intended division of labor:

- Could both fire on the same pathological loop, producing a double hard-stop or contradictory signals (e.g. `LoopDetection` stripping `tool_calls` while `ToolProgress` is injecting a WARN hint)?
- Which one owns the terminal decision when they disagree?

Recommend adding a one-paragraph note to `backend/CLAUDE.md`'s middleware section (and the PR description) on how they coexist, plus at least one test exercising both enabled together. This is the main merge blocker for me — two parallel loop/stagnation systems hardening around each other is hard to untangle later.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,543 @@
...
+                "tool_progress: %s/%s call intercepted (blocked): %s",
+                self._thread_id(runtime),
+                tool_name,
+                block_reason,
+            )
+            return self._make_blocked_message(request, tool_name, block_reason)
+        return self._update_state_from_result(handler(request), tool_name, runtime)
+
+    @override
+    async def awrap_tool_call(
```
willem-bd, line 508: **Concurrency: blocking `threading.Lock` acquired on the async path.** (concurrency, low)

`awrap_tool_call` → `_update_state_from_result` → `with self._lock` (line 294) acquires a blocking `threading.Lock` while running on the event loop. The critical sections are short (in-memory dict ops), so event-loop impact is negligible today. The risk is if the sync `wrap_tool_call` path (subagent executor thread pools hold the same lock) ever does heavier work under the lock — the event loop would stall until it releases.

Informational rather than blocking, but worth either a one-line comment noting the lock is held only around dict updates, or a revisit to `asyncio.Lock` / `to_thread` if the state machine grows heavier work under the lock.

willem-bd (commented):
## Review — structured tool-result meta + tool-progress state machine

Strong, well-engineered PR. Default-off gating materially de-risks the rollout, the state-machine accounting is correct on all three escalation paths, and the test coverage is excellent (106 new tests). Posting the three findings below inline.

- **Item 1 (perf, medium)** — `content.lower()` recomputed per marker on every success result.
- **Item 2 (architecture, medium)** — overlap with the existing `LoopDetectionMiddleware` is unaddressed.
- **Item 4 (concurrency, low)** — blocking `threading.Lock` acquired on the async path.

(Skipping the `deerflow_tool_meta` vs `deerflow_tool_error` coordination point — already flagged in the PR body; and the minor nits — `set()` redundancy, comment wording.)

Overall: the code-level findings are minor; the main merge blockers are the two architecture items, especially #2.

willem-bd (commented):
Two findings from a precision-focused review (medium effort). Both verified against the diff; the second is a test-quality issue that masks the very behavior it claims to lock in.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,569 @@
...
+        Jaccard recent_word_sets is also cleared so stale content windows from a prior run
+        do not cause false near-duplicate detections on the first success of the new run.
+        """
+        thread_id = self._thread_id(runtime)
+        with self._lock:
+            thread_tools = self._phase_states.get(thread_id)
+            if thread_tools is None:
+                return
+            for tool_name, tool_state in list(thread_tools.items()):
+                if tool_state.phase in ("blocked", "warned"):
```
willem-bd: **ACTIVE-phase state leaks across runs.**

This filter skips tools whose phase is still `"active"`, but an ACTIVE tool can end a run with non-zero `consecutive_problems` and non-empty `recent_word_sets`:

- `_assess_and_transition` increments `consecutive_problems` on every problem result and only flips phase at `stagnation_threshold`; the `else` branch (`new_state = replace(state, consecutive_problems=new_count)`) keeps the phase as-is. So one or two problems below threshold ⇒ phase stays ACTIVE, counter > 0.
- `recent_word_sets` is appended on every successful result and is never cleared on the problem path.

Concrete R2 failure (defaults: `stagnation_threshold=3`):

1. R1: tool returns 2 successes (word_sets cached) and 1 error. End state: `phase=active, consecutive_problems=1, recent_word_sets=(ws1, ws2)`.
2. R1 ends; this `_reset_blocked_states` call in `before_agent` skips the tool because phase is ACTIVE.
3. R2 first call: a clean success that shares vocabulary with R1's cached word_sets gets Jaccard-flagged as a near-duplicate ⇒ `consecutive_problems=2`. A single second mild problem then trips WARNED in R2 with the model never having seen the R1 context that produced the counter.

Worst case: R1 leaves `consecutive_problems = stagnation_threshold - 1`, so a single (real or false-positive) problem on the first R2 call trips WARNED immediately on stale state.

Fix: drop the `phase in ("blocked", "warned")` filter and reset all tracked tools at the run boundary — or at minimum, always clear `consecutive_problems` and `recent_word_sets` regardless of phase. (The method name `_reset_blocked_states` is now misleading too — it's really a per-run reset.)

backend/tests/test_tool_result_meta.py
```diff
@@ -0,0 +1,468 @@
...
+    ["none", "None", "NONE", "null", "Null", "false", "False", "no", "ok", "success", "n/a", ""],
+)
+def test_normalize_json_semantic_zero_error_string_not_treated_as_error(error_value: str):
+    """M2 regression: error field containing a conventional 'no-error' string must not trigger misclassification.
+
+    Tools sometimes return {"error": "none", "results": [...]} on success.
+    The string "none" is truthy in Python, so without this guard the message
+    would have been classified as error (unknown), inflating stagnation counters.
+    """
+    content = f'{{"error": {error_value!r}, "results": ["item1", "item2", "item3"]}}'
```
willem-bd: **This test does not actually exercise `_SEMANTIC_ZERO_ERROR_STRINGS`.**

`{error_value!r}` emits Python's single-quoted repr, producing invalid JSON:

```python
>>> f'{{"error": {"none"!r}, "results": [...]}}'
'{"error": \'none\', "results": [...]}'
```

`json.loads` rejects this with `JSONDecodeError`, so `_extract_json_error_text` returns `None` at the `except` clause — **before** reaching the `_SEMANTIC_ZERO_ERROR_STRINGS` check on lines 109–110. In `normalize_tool_message`, with `status="success"` and the JSON-parse path failing, the content falls through `_ERROR_PREFIX` → JSON → `_PARTIAL_MARKERS` to the final `success` branch and the assertion passes trivially.

Verification: deleting the entire `_SEMANTIC_ZERO_ERROR_STRINGS` guard does **not** make any of these 12 parametrized cases fail — they pass because the JSON is malformed, not because the guard works.

Fix: build the content with `json.dumps` so the parser actually sees valid JSON:

```python
import json
content = json.dumps({"error": error_value, "results": ["item1", "item2", "item3"]})
```

This way each case enters `_extract_json_error_text`, hits the semantic-zero branch, and returns `None` — which is the behavior the M2 regression test is supposed to lock down.

WillemJiang:
@whhe Please take a look at the latest review comments.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
whhe: Fixed in commit d3d3920f. Renamed the method to `_reset_run_states` and dropped the `phase in ("blocked", "warned")` filter — all tracked `(thread, tool)` pairs are now reset unconditionally in `before_agent`. An ACTIVE tool that exits R1 with non-zero `consecutive_problems` or non-empty `recent_word_sets` starts R2 fully clean.

Added `test_before_agent_resets_active_state_consecutive_problems_and_word_sets` to lock the behavior: drives the tool to `phase=active, consecutive_problems=1, recent_word_sets≠()` in R1 (2 successes + 1 sub-threshold error with `stagnation_threshold=3`), then asserts both fields are zero/empty after `before_agent` fires for R2. Without the fix, the old phase filter would leave those fields unchanged and the assertions would fail.

backend/tests/test_tool_result_meta.py
whhe: Fixed in commit d3d3920f. Replaced the `{error_value!r}` f-string with `json.dumps({"error": error_value, "results": [...]})` so each parametrized case produces valid JSON. The parser now reaches `_extract_json_error_text`, the `_SEMANTIC_ZERO_ERROR_STRINGS` check fires for non-empty strings, and removing the guard would actually cause the 11 non-empty cases to fail.

The `""` (empty-string) case still exits at `if not error: return None` before reaching the semantic set — this is now documented in the test docstring so future readers don't have to trace the implementation to understand why.

backend/packages/harness/deerflow/agents/middlewares/tool_progress_middleware.py
```diff
@@ -0,0 +1,567 @@
...
+            return self._pending.pop(key, [])
+
+    def _clear_stale_pending(self, runtime: Runtime) -> None:
+        thread_id, current_run = self._pending_key(runtime)
+        with self._lock:
+            for key in list(self._pending):
+                if key[0] == thread_id and key[1] != current_run:
+                    del self._pending[key]
+
+    def _reset_run_states(self, runtime: Runtime) -> None:
```
willem-bd, line 445: **Cross-run reset diverges from `LoopDetectionMiddleware` (design consistency).** `_reset_run_states` clears `consecutive_problems`, `block_reason`, and `recent_word_sets` for every tracked tool at each `before_agent`. The docstring rationale (the model has no memory of the prior hint, so stale counters shouldn't trip WARNED on the first call) is sound, and it cleanly resolves the earlier asymmetry where BLOCKED persisted across runs but WARN hints were dropped.

The trade-off worth a conscious call: non-recoverable errors (`rate_limited`, `transient`) can now only escalate WARNED→BLOCKED *within a single run*. If a tool is rate-limited on every call across several consecutive user turns (each a new run), the counter resets at each `before_agent`, so the tool never reaches BLOCKED across turns — it only ever gets WARNED. `LoopDetectionMiddleware` deliberately takes the opposite stance: it *retains* `_history` across runs and only clears *other-run* pending warnings (`loop_detection_middleware.py:525`), so its detection accumulates over the thread's lifetime.

Neither policy is obviously wrong — per-run reset says "each turn is a fresh attempt," LoopDetection says "accumulated pattern across the thread." But having the two stagnation middlewares use opposite cross-run scoping is a consistency rough edge future readers will trip on. Options: (a) keep per-run reset and add a one-line note in the docstring contrasting it with LoopDetection's thread-scoped retention so the divergence is intentional-on-record; (b) retain `consecutive_problems` across runs and reset only the hint/word-set state (closer to LoopDetection), so cross-turn stagnation can still escalate. Non-blocking — off by default.

whhe: Thanks for the careful analysis. Went with option (a).

Expanded the `_reset_run_states` docstring to make the divergence intentional-on-record: result-quality errors (`rate_limited`, `transient`) are time-bound failures whose root cause may resolve between turns, so per-run reset avoids false-positive BLOCKED calls on retries that would now succeed. `LoopDetectionMiddleware` retains history across runs because call-pattern loops are time-invariant — the model keeps issuing the same calls regardless of when the run started. The two middlewares guard different failure modes, and their cross-run scoping policies intentionally differ as a consequence.

WillemJiang:
@whhe  Could you resolve the conflicts with the main branch.

WillemJiang (approved):

PR #3605, opened 2026-06-15
fix(runtime): propagate interrupt through SSE values events for LangGraph SDK

### LINKED ISSUE #3595: [bug] __interrupt__不能正常使用
### Before you start

### Problem summary

__interrupt__不能正常使用

### Affected area(s)

### What happened?

  Bug 1: serialization.py — __interrupt__ 被错误过滤
  
  文件: backend/packages/harness/deerflow/runtime/serialization.py:62

  问题: serialize_channel_values() 把 __interrupt__ key 过滤掉了，和 __pregel_* 一视同仁。但前端 SDK (@langchain/langgraph-sdk) 通过 stream.values.__interrupt__ 检测中断。key 被过滤后 SDK 永远看不到中断事件，凭证弹窗/审核卡片弹不出来。  
  修复: 删除 or key == "__interrupt__" 条件，只过滤 __pregel_*。  
  Docstring 也要更新（第 57-58 行，现在还说“Internal keys like __pregel_* and __interrupt__ are removed”）。

  ---
  Bug 2: worker.py — __interrupt__ SSE mode 前端 SDK 不识别
  
  文件: backend/packages/harness/deerflow/runtime/runs/worker.py:353-357
  
  问题: LangGraph 的 astream() 在触发 interrupt() 时 yield ("__interrupt__", data)。worker 把它当独立 SSE event 发出去。但 @langchain/langgraph-sdk 只处理 values event 里的 __interrupt__ 字段，不认识 __interrupt__ 这个 SSE event
  type。结果：SDK 的 state 里 interrupt 字段始终为空。

  修复: 检测到 mode == "__interrupt__" 时，重新包装为 values event：

  if mode == "__interrupt__":
    t await bridge.publish(run_id,1"values",
          serialize({"__interrupt__": chunk}, mode="values"))

Docker (make docker-start)

### Operating system

Linux

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

ech0hol:
Fixes #3595

## Why

`__interrupt__` (used by LangGraph's `interrupt()` for human-in-the-loop flows like credential prompts) never reached the frontend SDK. Two independent bugs conspired to block it at different stages of the serialization pipeline. Users with graphs that rely on `interrupt()` / `interrupt_before` / `interrupt_after` saw no interrupt UI — the SDK's `stream.values.__interrupt__` field was always empty.

## What changed

- `__interrupt__` now survives the channel-value serializer and is correctly delivered as a `values` SSE event, matching what `@langchain/langgraph-sdk` expects. Interrupt-based flows (credential popups, confirmation dialogs) now work end-to-end.

## Surface area

## Bug fix verification

- **Test path:** `backend/tests/test_interrupt_serialization.py` — 4 new regression tests covering single-mode, multi-mode, standalone-event-gating, and unit-level channel-value preservation.
- **Red on `main`, green on this branch:**—  the new tests fail on `main` because `serialize_channel_values` strips `__interrupt__` and multi-mode streams emit it as a standalone SSE event type.

## Validation

```bash
# Backend (container)
docker exec deer-flow-gateway bash -c 'cd /app/backend && uv run ruff check packages/harness/deerflow/runtime/serialization.py packages/harness/deerflow/runtime/runs/worker.py tests/test_serialization.py tests/test_interrupt_serialization.py && uv run ruff format --check packages/harness/deerflow/runtime/serialization.py packages/harness/deerflow/runtime/runs/worker.py tests/test_serialization.py tests/test_interrupt_serialization.py && uv run pytest tests/test_serialization.py tests/test_interrupt_serialization.py -v'

### CODE REVIEW

File: backend/packages/harness/deerflow/runtime/runs/worker.py [hunk 1]
```diff
@@ -332,6 +332,13 @@ async def run_agent(
                 if mode is None:
                     continue
 
+                # Repackage __interrupt__ as a "values" event so the LangGraph SDK
+                # can detect it — the SDK only looks for __interrupt__ inside
+                # values events, not as a standalone SSE event type (issue #3595).
+                if mode == "__interrupt__":
```
fancyboi999: Under the pinned `langgraph==1.1.9`, `graph.astream` never yields a `("__interrupt__", chunk)` item for `interrupt()` — the interrupt rides inside a `values` (or `updates`) chunk in every `stream_mode` combination, so this branch never executes (I verified by running the graph). Bug 1 already delivers the interrupt as a `values` event for the modes production uses (`values` is always present). Fine to keep this as a forward-defensive guard, but worth a one-line comment saying so — otherwise it reads as the load-bearing half of the fix, and it isn't.

backend/tests/test_interrupt_serialization.py
```diff
@@ -0,0 +1,115 @@
...
+            continue
+        if mode == "__interrupt__":
+            data = serialize({"__interrupt__": chunk}, mode="values")
+            await collector.publish("test", "values", data)
+            continue
+        sse_event = _lg_mode_to_sse_event(mode)
+        data = serialize(chunk, mode=mode)
+        await collector.publish("test", sse_event, data)
+    standalone = [e for e in collector.events if e[0] == "__interrupt__"]
+    assert len(standalone) == 0, "Bug #2: __interrupt__ leaked as standalone event"
```
fancyboi999: This is the only one of the four new tests that passes on `main` in my run — it's vacuously green: no langgraph version emits a standalone `__interrupt__` event for `interrupt()`, so `standalone` is always empty whether or not the fix is applied. Combined with `test_multimode_interrupt_as_values` (which goes green via the `values` branch, not the `mode == "__interrupt__"` branch), the `worker.py` change isn't exercised by any test. The three serializer-level assertions are the ones doing the real work and they're solid.

fancyboi999 (commented):
I triaged #3595, so I traced the whole chain again here — this time verifying the runtime behavior end-to-end against the pinned `langgraph==1.1.9` and the bundled `@langchain/langgraph-sdk`. **The fix is correct and works; Bug 1 is what carries it.** A heads-up on Bug 2 below, with the ground truth I collected.

### Bug 1 (`serialization.py`) — the real fix, confirmed
`serialize(chunk, mode="values")` routes to `serialize_channel_values` (`serialization.py:126-127`), which on `main` strips `__interrupt__` alongside `__pregel_*`. Dropping the `or key == "__interrupt__"` clause lets it survive. I ran your test file against `main`: `test_values_mode_includes_interrupt`, `test_multimode_interrupt_as_values`, and `test_serialize_channel_values_keeps_interrupt` all go **red on `main`** / green on this branch — every one of them catches the serializer stripping the key.

### SDK side checks out
A `values` event carrying `__interrupt__` is detected at `react/stream.lgp.js:417` (`Array.isArray(values.__interrupt__)`). Because the interrupt value is a tuple of `Interrupt` objects, it serializes to a JSON array, so that check passes. The SDK also *merges* rather than replaces interrupt-bearing values events — `ui/manager.js:228`: `setStreamValues(prev => ({ ...prev, ...interruptData }))` — so emitting a partial `{"__interrupt__": [...]}` doesn't clobber prior state. The delivery shape is exactly right.

### The Bug 2 change (`worker.py`) is inert under the pinned langgraph
The premise (mine in #3595 too, so this corrects my own earlier triage) was that `graph.astream` yields a standalone `("__interrupt__", chunk)` item that gets mis-emitted as an `__interrupt__` SSE event. I checked what `langgraph 1.1.9` actually yields for `interrupt()`, running the exact graph from your tests across every relevant `stream_mode`:

```
["values","messages"]  -> ("values", {..., "__interrupt__": (Interrupt(...),)})   # rides INSIDE values
["messages","values"]  -> same
["updates"]            -> ("updates", {"__interrupt__": (...)})
"values"               -> {..., "__interrupt__": (...)}
["messages"]           -> (nothing emitted)
```

In **no** configuration does it yield a bare `("__interrupt__", ...)` tuple, so `mode == "__interrupt__"` is never true in `run_agent`. Production always includes `"values"` (`channels/manager.py:55` `STREAM_MODES = ["messages-tuple", "values"]`, default `["values"]`), and the interrupt arrives inside that `values` chunk — handled entirely by Bug 1. The `worker.py` branch is harmless, but it does nothing today.

This is visible in the tests too: in my run against `main`, `test_no_standalone_interrupt_event` is the **one test that passes** — it's vacuously green, because nothing ever emits a standalone `__interrupt__` event to leak. And `test_multimode_interrupt_as_values` goes green via the `values` branch (its `assert event == "values"` holds because langgraph genuinely delivers it as `values`), not via the `mode == "__interrupt__"` branch, which never executes there either. So neither test actually exercises the `worker.py` change.

### Recommendation
Ship it — Bug 1 fixes the user-facing problem and is well covered. For Bug 2 it's your call: either drop the `worker.py` branch (and the two tests modeling the path that doesn't occur) to keep the diff to the one change that matters, or keep it as an explicit forward guard with a short comment saying it's defensive against a future langgraph that emits a standalone `__interrupt__`. I'd lean toward the smaller diff.

Verification method, honestly: I ran `langgraph 1.1.9`'s `astream` for `interrupt()` directly and ran your test file against `main` (3 red / 1 green as above). I did not run the full backend suite or a live browser flow.

ech0hol:
@fancyboi999 
Thanks for the thorough review — I independently verified your finding on Bug 2.

I wrote a test harness that runs `graph.astream` with `interrupt()` across all `stream_mode` combinations (`"values"`, `"messages"`, `"updates"`, and their multi-mode variants). The result confirms your observation: **langgraph 1.1.9 never yields a bare `("__interrupt__", chunk)` tuple**. The interrupt always arrives as a field inside a `values` or `updates` chunk.

| stream_mode | interrupt delivery |
|---|---|
| `"values"` | `{"__interrupt__": (Interrupt(...),), ...}` inside values chunk |
| `"updates"` | `{"__interrupt__": (...)}` inside updates chunk |
| `["values", "messages"]` | mode=`values`, chunk contains `__interrupt__` |
| `["values", "messages", "updates"]` | both `values` and `updates` carry `__interrupt__` |
| `"messages"` (single or in list) | nothing emitted |

No combination produces a standalone `__interrupt__` SSE event type.

Based on this, I've removed the `worker.py` change and the two tests that exercised the unreachable path (`test_multimode_interrupt_as_values`, `test_no_standalone_interrupt_event`). The PR now only contains the Bug 1 fix:

- `serialization.py`: drop `or key == "__interrupt__"` filter
- `test_serialization.py`: updated assertions
- `test_interrupt_serialization.py`: 2 tests covering the real fix

willem-bd (commented):
Thanks @snaplap. I found one issue that should be addressed before this is ready — it's separate from (and refines) the SDK-shape confirmation in the earlier review. The serializer key now survives (presence detection works), but the per-element payload does not match the Platform API the PR says it targets.

[P1] Interrupt objects serialize to `str()` reprs, not structured dicts

- Location: `backend/packages/harness/deerflow/runtime/serialization.py` — the `result[key] = serialize_lc_object(value)` call for the `__interrupt__` key, falling into the `str(obj)` last-resort branch of `serialize_lc_object`.
- Problem: `langgraph.types.Interrupt` is a plain class — it has no `model_dump()` / `dict()`, and it is not a `tuple`/`dict`/`list` — so `serialize_lc_object` falls through every branch to the `return str(obj)` fallback. `__interrupt__` is therefore delivered to the SDK as `["Interrupt(value='...', id='...')"]`, i.e. a list of **strings**, not the structured `[{value, ...}]` objects the LangGraph Platform API emits. The SDK's `stream.interrupt` getter (`@langchain/langgraph-sdk` `react/stream.lgp.js:419`) returns `values.__interrupt__[0]`, so `stream.interrupt` becomes that string; any UI reading `stream.interrupt.value` / `.resumable` gets `undefined`. The interrupt is now *detectable* (the prior review's `Array.isArray(values.__interrupt__)` check passes — that's why it looked correct), but the human-readable prompt value this PR targets (credential question, etc.) is buried inside an opaque repr string, so interrupt-based flows still don't expose their value end-to-end. The same malformed payload also flows through `serialize_channel_values_for_api` into the REST thread/run responses.
- Evidence: verified empirically against the pinned `langgraph==1.1.9`. A live `interrupt({'question': 'Enter API key', 'field': 'password'})` yields `__interrupt__: (Interrupt(...),)`; running that tuple through `serialize_lc_object` returns `["Interrupt(value={'question': 'Enter API key', 'field': 'password'}, id='324cc96b...')"]` with element type `str`. `Interrupt` exposes `value`, `id`, `interrupt_id`, `from_ns` but has no serializer hook (`hasattr(obj, "model_dump")` is `False`).
- Suggested fix: serialize `Interrupt` to a dict before the `str()` fallback. Minimal targeted form:
  ```python
  from langgraph.types import Interrupt
  ...
  if isinstance(obj, Interrupt):
      return serialize_lc_object({"value": obj.value, "id": getattr(obj, "id", None)})
  ```
  A more general option (covers similar plain LangGraph objects) is to add, before the last-resort `str()`:
  ```python
  if hasattr(obj, "__dict__") and obj.__dict__:
      return serialize_lc_object(vars(obj))
  ```
- Test: drive a real `interrupt()` graph and assert the structured shape (see P2).

[P2] New tests assert key presence, not payload shape

- Location: `backend/tests/test_interrupt_serialization.py` — `test_serialize_channel_values_keeps_interrupt` and `test_values_mode_includes_interrupt`.
- Problem: `test_serialize_channel_values_keeps_interrupt` feeds plain dicts (`[{"value": "test", "resumable": True}]`) instead of real `Interrupt` objects, so it never exercises the `str()`-fallback path that production hits. `test_values_mode_includes_interrupt` only asserts `"__interrupt__" in e[1]`. Both stay green while the payload is malformed (P1), so the "red on `main`, green on branch" claim only proves the key survives — not that the value is usable.
- Suggested fix / Test: add one assertion against a real interrupt graph that inspects the element, e.g.:
  ```python
  assert isinstance(result["__interrupt__"][0], dict)
  assert result["__interrupt__"][0]["value"] == "Please provide API credentials"
  ```
  This would go red on the current branch and confirm the P1 fix.

fancyboi999:
@snaplap good call dropping the `worker.py` branch and those two tests — that path really is unreachable on 1.1.9, so the PR is tighter without it.

On @willem-bd's **P1**: it's correct, and it's a gap in my earlier review. I confirmed the key *survives* and is array-shaped (so `Array.isArray` passes and detection works), but I stopped there and didn't check the element shape — which is exactly where it breaks. Reproduced against the pinned `langgraph==1.1.9`:

```
serialize_lc_object((Interrupt(value={'question': 'Enter API key'}, id='...'),))
# -> ["Interrupt(value={'question': 'Enter API key'}, id='...')"]   # element type: str
```

`Interrupt` is a `__slots__` class with no `model_dump`/`dict` and it isn't tuple/dict/list, so `serialize_lc_object` falls all the way to `return str(obj)`. The SDK's `get interrupt()` (`react/stream.lgp.js:420`) returns `values.__interrupt__[0]`, so `stream.interrupt` becomes that repr string and `stream.interrupt.value` is `undefined`. Presence is fixed, payload isn't usable. P2 is right too — the current tests feed hand-built dicts / only assert presence, so they stay green over the malformed payload.

One refinement on the fix, from the same repro: the **general** alternative (`if hasattr(obj, "__dict__") and obj.__dict__: return serialize_lc_object(vars(obj))`) won't catch this one — `Interrupt` has no `__dict__` (it's `__slots__ = ('value', 'id')`), so `hasattr(obj, "__dict__")` is `False` and `vars(obj)` actually raises `TypeError: vars() argument must have __dict__ attribute`. The **targeted** form is the one that works:

```python
from langgraph.types import Interrupt
...
if isinstance(obj, Interrupt):
    return serialize_lc_object({"value": obj.value, "id": getattr(obj, "id", None)})
```

Routing `value` back through `serialize_lc_object` matters since `value` is itself a dict in the credential case. Two notes: on 1.1.9 `Interrupt` exposes `value`/`id`/`interrupt_id`/`from_ns` but **no `resumable`**, so a UI reading `stream.interrupt.resumable` stays `undefined` regardless of this fix — if the UI needs it, it'll have to be synthesized separately. And the malformed shape also reaches REST through `serialize_channel_values_for_api`, as @willem-bd noted, so the P2 assertion should drive a real `interrupt()` graph rather than a constructed dict.

Agree this should land before merge. Good catch on the element shape.

ech0hol:
@willem-bd @fancyboi999 Thanks both for identifying the two gaps — P1 (Interrupt serializes to `str` via fallback) and P2 (tests use hand-built dicts, never exercising the real `str()` path). Latest push fixes both.

P1: Added `isinstance(obj, Interrupt)` in `serialize_lc_object` before the `str()` fallback. `Interrupt` is a `__slots__` class so `vars()` / `__dict__` won't work — went with the targeted form you both suggested.

Verified the output change:

```python
# before: ["Interrupt(value=..., id=...)"] → str elements
# after: [{"value": ..., "id": "..."}] → dict elements
```

P2: Both tests now use real `Interrupt` objects and assert the payload is a dict with the correct value. They fail without the `isinstance` branch and pass with it, so the fix is actually load-bearing.

fancyboi999 (commented):
Confirmed — the targeted `isinstance(obj, Interrupt)` form is the right call (a generic `vars(obj)` / `__dict__` path wouldn't work: `Interrupt.__slots__ == ('value', 'id')`, no `__dict__`), and both fixes hold up end to end. I pulled the branch and ran your regression tests red/green against the pinned `langgraph`:

- **Base (no fix)**: both tests fail — `__interrupt__` is stripped (`assert '__interrupt__' in result`).
- **Full PR**: both pass.
- **P1 isolation** (kept the `__interrupt__`-preserve change, removed *only* the `isinstance` branch): the dict-shape assertions fail with the str fallback — e.g. `assert isinstance(result["__interrupt__"][0], dict)` → `isinstance("Interrupt(value={'question': 'Enter API key'}, id='test-interrupt-id')", dict)`. So the `isinstance` branch is genuinely load-bearing, exactly as you said — the tests aren't vacuous.

Placement is right too: `Interrupt` has no `model_dump`/`dict`, so it falls through to the new branch before the `str()` last resort, and the recursive `serialize_lc_object({...})` keeps `value` JSON-safe.

One forward-looking, non-blocking note: the payload is an explicit `{value, id}` allowlist. The pinned `langgraph` `Interrupt` only carries those two, so it's complete today — but if a future `langgraph` adds an SDK-relevant attribute (e.g. a `resumable`/`ns`-type field), it won't auto-propagate through this branch. Worth a quick revisit if/when the SDK starts reading more off the interrupt; not needed for this fix.

Looks correct to me — recommend merge.

WillemJiang (approved):

PR #3610, opened 2026-06-16
fix(subagents): raise general-purpose max_turns to 150 and default timeout to 30min

hetaoBackend:
## Problem

Deep-research runs fail out of the box: parallel subtasks (子任务) die with
`GraphRecursionError: Recursion limit of 100 reached without hitting a stop condition`.

The failing subtasks are `general-purpose` subagents, whose `max_turns` defaults to
**100** (`general_purpose.py`). A thorough research subtask doing many search→fetch
rounds exhausts 100 turns before producing a final answer, and the whole subtask is
lost. This is a *separate* limit from the lead agent's `recursion_limit` (also 100) —
raising the lead's limit does **not** fix it.

## Change

- `general-purpose` `max_turns`: **100 → 150** — the direct fix for the observed error.
- Default subagent `timeout_seconds`: **900 → 1800** (15 → 30 min). The two limits are
  coupled: at the observed pace, 150 turns can exceed the old 15-min timeout, so a
  turns-only bump would just shift the failure from recursion to timeout. For built-in
  agents the per-agent timeout is always taken from the global `subagents.timeout_seconds`
  (`registry.py`), so the global default is the only lever.

The lead agent's `recursion_limit` (100) is intentionally left unchanged — the failures
are subagent-only; the lead orchestrates + synthesizes without deep looping.

## Not the whole story

Raising the limit is the pragmatic fix. The structural fix is finer task decomposition:
the failing brief asked a single subtask to cover 5 sections / a dozen sub-questions and
cite every source. Even 150 turns may not be enough for briefs that large.

## Tests

- New regression `test_builtin_defaults_have_research_headroom` asserts the resolved
  out-of-box `general-purpose` config has `max_turns=150` / `timeout_seconds=1800`.
- Updated default-value assertions in `test_subagent_timeout_config.py` and
  `test_app_config_reload.py`.
- `145 passed` across subagent-config / reload / task-tool / gateway / prompt-security
  suites; `ruff` clean.

### DIFF
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -311,7 +311,7 @@ Proxied through nginx: `/api/langgraph/*` → Gateway LangGraph-compatible runti
 
 **Built-in Agents**: `general-purpose` (all tools except `task`) and `bash` (command specialist)
 **Execution**: Dual thread pool - `_scheduler_pool` (3 workers) + `_execution_pool` (3 workers)
-**Concurrency**: `MAX_CONCURRENT_SUBAGENTS = 3` enforced by `SubagentLimitMiddleware` (truncates excess tool calls in `after_model`), 15-minute timeout
+**Concurrency**: `MAX_CONCURRENT_SUBAGENTS = 3` enforced by `SubagentLimitMiddleware` (truncates excess tool calls in `after_model`); default subagent timeout `subagents.timeout_seconds=1800` (30 min) and built-in `general-purpose` `max_turns=150` (raised from 100/15-min so deep-research subtasks stop hitting `GraphRecursionError` out of the box)
 **Flow**: `task()` tool → `SubagentExecutor` → background thread → poll 5s → SSE events → result
 **Events**: `task_started`, `task_running`, `task_completed`/`task_failed`/`task_timed_out`
 **Deferred MCP tools** (if `tool_search.enabled`): `SubagentExecutor._build_initial_state` assembles deferral after policy filtering via the shared `assemble_deferred_tools` (fail-closed), appends the `tool_search` tool, injects the `<available-deferred-tools>` section into the subagent's `SystemMessage`, and threads the setup to `_create_agent`, which attaches `DeferredToolFilterMiddleware` through `build_subagent_runtime_middlewares(deferred_setup=...)`. Subagents thus withhold full MCP schemas until promotion, same as the lead agent; each task run gets a fresh `ThreadState` so promotion is isolated per run
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -42,6 +42,11 @@ DEFAULT_GATEWAY_URL = "http://localhost:8001"
 DEFAULT_ASSISTANT_ID = "lead_agent"
 CUSTOM_AGENT_NAME_PATTERN = re.compile(r"^[A-Za-z0-9-]+$")
 
+# Lead-agent recursion budget (LangGraph super-steps for the lead graph only).
+# This is independent of subagent depth: a `task()` dispatch runs the whole
+# subagent inside ONE lead tools-node step, and subagents enforce their own
+# limit via `subagents.max_turns` (see SubagentExecutor). Do not conflate this
+# 100 with the general-purpose subagent's max_turns.
 DEFAULT_RUN_CONFIG: dict[str, Any] = {"recursion_limit": 100}
 DEFAULT_RUN_CONTEXT: dict[str, Any] = {
     "thinking_enabled": True,
--- a/backend/app/gateway/services.py
+++ b/backend/app/gateway/services.py
@@ -224,6 +224,11 @@ def build_run_config(
     the LangGraph Platform-compatible HTTP API and the IM channel path behave
     identically.
     """
+    # Lead-agent recursion budget (LangGraph super-steps for the lead graph
+    # only). Independent of subagent depth: a `task()` dispatch runs the whole
+    # subagent inside ONE lead tools-node step, and subagents enforce their own
+    # limit via `subagents.max_turns`. Do not conflate this 100 with the
+    # general-purpose subagent's max_turns.
     config: dict[str, Any] = {"recursion_limit": 100}
     if request_config:
         # LangGraph >= 0.6.0 introduced ``context`` as the preferred way to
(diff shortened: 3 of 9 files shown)


fancyboi999 (commented):
Nice diagnosis — I traced it through and it holds up. The `general-purpose` failures really are subagent-scoped, and `max_turns` is the right lever.

### Why this actually fixes the `GraphRecursionError` (verified)
`SubagentExecutor` sets the subagent graph's LangGraph budget straight from `max_turns` — `executor.py:521`: `"recursion_limit": self.config.max_turns`. So `general-purpose` `max_turns` 100 → 150 raises that subagent's `recursion_limit` to 150, which is exactly the "Recursion limit of 100 reached" the brief was hitting. And you're right that the lead's `recursion_limit=100` is a separate budget — a `task()` dispatch runs the whole subagent inside one lead tools-node step, so raising the lead wouldn't have helped. The clarifying comments you added in `manager.py` / `services.py` are a good touch; that exact conflation is how this kind of bug gets misfiled.

### Resolution checks out for the out-of-box path
`general-purpose` doesn't set its own `timeout_seconds`, so `get_subagent_config` resolves:
- `max_turns` → 150 from the builtin def — global `subagents.max_turns` defaults to `None`, so the `is_builtin` override branch at `registry.py:97` is skipped.
- `timeout_seconds` → 1800 via the builtin global-override branch (`registry.py:88`: `is_builtin and 1800 != the builtin's own 900`).

So `test_builtin_defaults_have_research_headroom` asserts the right resolved pair, and the coupling argument (150 turns can outrun the old 15-min cap, so bump to 30) is reasonable.

### Two minor, non-blocking notes
1. `SubagentConfig` in `subagents/config.py` still carries `timeout_seconds: int = 900` plus a docstring reading "default: 900 = 15 minutes" (`:24` / `:35`). Out of this diff and functionally harmless — for builtins the global override wins, so the resolved value is 1800 — but it's now the one spot that still advertises 900/15 min as the default. A one-line docstring touch-up in a follow-up would keep the layered default from misleading a reader.
2. Scope nuance: the global timeout override is builtin-only (the `is_builtin` guard at `registry.py:88`). So the new 1800 lands on built-in subagents; a custom agent that omits `timeout_seconds` still resolves to its own config default, not 1800. Probably intended (custom agents own their limits) — just flagging since the body and `config.example.yaml` phrase it as "all subagents".

**Verdict: looks correct, recommend merge.** Verification was code-level: I traced the resolver branches and confirmed `executor.py:521` ties `recursion_limit` to `max_turns`. I didn't run the full suite or reproduce a live deep-research run.

willem-bd:
A couple of small follow-ups on this PR:

### 1. Pin `bash.max_turns` in the new defaults test

`config.example.yaml` now advertises `Built-in defaults: general-purpose=150, bash=60`, but `test_builtin_defaults_have_research_headroom` only pins `general-purpose`. If someone later tweaks `bash`'s `max_turns`, the comment in `config.example.yaml` goes stale silently. Consider adding a one-line assertion (in the same test or a sibling) for `bash.max_turns == 60` so the doc and the constant move together. Low cost, prevents drift.

### 2. Misleading test name in `test_subagent_timeout_config.py`

`test_default_timeout_preserved_when_no_config` (around line 325) calls `_reset_subagents_config(timeout_seconds=900)` — i.e. it deliberately uses the **old** default to exercise propagation, then asserts `max_turns == 150` (the new built-in). The behavior under test is fine, but the function name now reads confusingly: the "default timeout preserved" being tested is `900`, which is no longer the package default. Either:

- rename to something like `test_explicit_global_timeout_propagates_to_general_purpose`, or
- switch the value to `1800` for symmetry with the new default.

Minor, but it'll save a future reader a double-take.

hetaoBackend:
Addressed in `ebad96e8`.

**@fancyboi999**
- *Note 1 (stale 900 in `SubagentConfig`):* clarified the `config.py` docstring and the `subagents.timeout_seconds` field description to spell out the layering — the bare `900` is only a fallback; for built-in agents the effective cap is the global `subagents.timeout_seconds` (default 1800). Left the field value at `900` on purpose: changing it is a no-op for builtins (the registry override wins) and would break `test_subagent_skills_config.py:111`, which pins the base default.
- *Note 2 (builtin-only scope):* you're right — the global override is gated by `is_builtin` at `registry.py:88`, so a custom agent that omits `timeout_seconds` resolves to its own `900`, not `1800`. That's intended (custom agents own their limits), so I fixed the wording rather than the behavior: `config.example.yaml` and the field description now say "built-in subagents" and note custom agents keep their own `timeout_seconds`.

**@willem-bd**
- *Pin `bash.max_turns`:* added `assert get_subagent_config("bash").max_turns == 60` to `test_builtin_defaults_have_research_headroom`, so the `config.example.yaml` "bash=60" comment can't drift silently.
- *Misleading test name:* renamed `test_default_timeout_preserved_when_no_config` → `test_explicit_global_timeout_propagates_to_general_purpose`. Kept the explicit `900` (rather than switching to `1800`) because it deliberately exercises a *non-default* global value propagating to the builtin — that propagation case is distinct from the new defaults test.

`106 passed` across subagent-config / skills-config / reload suites; `ruff` clean.

WillemJiang (approved):

fancyboi999:
Thanks for the quick turnaround. The docstring rewrite in `subagents/config.py` is exactly right — "this 900 only applies when no differing global value exists" captures the layering precisely, and the `built-in subagents` wording on the field description + `config.example.yaml` closes the scope point cleanly. Both were non-blocking and this puts them to rest.

One small correction on the reasoning, just to keep the record straight — not asking for any change:

> changing it ... would break `test_subagent_skills_config.py:111`, which pins the base default

That assertion is on `CustomSubagentConfig` (the config.yaml model in `config/subagents_config.py:65`), which carries its own independent `Field(default=900)` — a different type from the `SubagentConfig` dataclass (`subagents/config.py:35`) my note was about. They aren't linked. I flipped the `SubagentConfig` dataclass default to `1800` locally and ran `test_subagent_skills_config.py` + `test_subagent_timeout_config.py`: 94 passed, nothing broke. So that test wasn't actually the constraint.

The decision to leave it at `900` is still the right one, for the first reason you gave: it's a no-op for built-ins (the registry's global override wins), so the dataclass value is a pure fallback and bumping it would just be churn. Keeping it plus the clarified docstring is the cleaner call.

PR #3611, opened 2026-06-16
feat(subagents): attribute subagent spans to parent thread's Langfuse…

### LINKED ISSUE #3604: [feat] Attribute subagent LLM/tool spans to the same Langfuse session as the lead agent
### Before you start

### Problem / motivation

Subagent LLM and tool calls appear in Langfuse as standalone top-level traces (e.g. model-level `PatchedChatDeepSeek` entries) — they are not missing, but incorrectly grouped and attributed.

Since the subagent execution path skips `inject_langfuse_metadata(...)`, these traces do not inherit the parent thread's `session_id` and `user_id`. Instead, they generate isolated session identifiers and often fall back to the `default` user. While searchable in the full trace list, they cannot be accessed directly from the parent lead agent session view.

**Asymmetry vs. lead agent path**
| Aspect | Lead agent | Subagent |
|---|---|---|
| Tracing callbacks at graph root | Attached | **Not attached** — only `SubagentTokenCollector` in run config callbacks |
| Session/user metadata injection | Applied via `inject_langfuse_metadata(...)` | **Skipped entirely** |
| Model-level tracing handlers | Disabled (`attach_tracing=False`) | **Enabled by default** — traces originate at the model layer |

**Practical impact**
1. No unified session view to review all subagent activity for a given thread.
2. DeerFlow internal token accounting remains accurate, but Langfuse cost aggregation breaks for subagent traffic due to mismatched session and user labels.

Image

Image

### Proposed solution

From the user's / operator's perspective: opening a thread in the Langfuse
Sessions page shows **all** traces for that conversation — the lead agent
trace **and** every subagent trace — grouped under the same session card,
carrying the same `user_id`, with each subagent trace clearly named
(`subagent:general-purpose`, `subagent:bash`, etc.) so they're
distinguishable from the lead trace.

This issue does **not** change the trace topology: subagent traces remain
**separate top-level traces** in the same session. They are not nested as
child spans under the lead trace. That nesting is a separate, harder
follow-up (Plan B below).

### Affected area(s)

### B. Nested parent-child span attachment

Make subagent spans appear as child observations **under** the lead trace,
producing a single tree per conversation.

Pros: cleanest Langfuse UI (one trace = one conversation); nested cost
breakdown; subagent tool/LLM calls visually nest under the `task` tool
call that triggered them.

Cons: Langfuse v4 strings parent-child via OTEL contextvar. Subagents run
on `_isolated_subagent_loop` (separate thread + event loop); Python's
`copy_context()` copies ContextVar snapshots, but the LangChain
`CallbackHandler` for Langfuse decides root-ness via
`on_chain_start(parent_run_id=None)`, which fires fresh on the subagent
side. To force nesting we'd have to explicitly capture the parent trace id
+ observation id at `task()` tool invocation time and re-attach them on
the subagent side via Langfuse's OTEL `Context.attach` API.

### C. Fully robust nested link across concurrency

Same as B but with hardened concurrency: wrap
`_submit_to_isolated_loop_in_context` to explicitly propagate the OTEL
`Context` object, test under parallel subagent fan-out, and handle the
Langfuse SDK's internal OTEL state.

### Why A first

- A is a **strict prerequisite** for B: you need root-level callbacks and
  metadata injection regardless of whether you also do nested attachment.
- A alone solves ~90% of the operator pain (session-grouped visibility,
  correct user attribution, distinguishable subagent trace names).
- B and C can be staged as follow-ups once A is in production and we have
  real signal on whether nested tree view is actually needed.

### Additional context

- Token usage is **not** affected by this issue. Subagent token accounting
  flows through `SubagentTokenCollector` → `RunJournal` (#2838) and is
  unrelated to Langfuse trace visibility. Don't conflate the two.

#3376

heart-scalpel:
Fixes #3604

## Why

Subagent LLM/tool spans *were* landing in Langfuse — they were not invisible. What was missing was grouping and attribution: the subagent execution path did not call `inject_langfuse_metadata(...)` and built its model with `attach_tracing=True`, so every subagent run produced an isolated top-level trace with a fresh `session_id` and the default user. They were findable in the unfiltered trace list (the `subagent:<name>` tag narrowed things down) but could not be pivoted to from the parent thread's Sessions view, and Langfuse cost attribution for subagent traffic did not line up with the parent conversation — even though DeerFlow's internal `SubagentTokenCollector` accounting was already correct.

This extends the lead-agent tracing wiring to the subagent path. 

## What changed

From the operator's perspective: opening a thread in the Langfuse **Sessions** page now shows all traces for that conversation — the lead agent trace and every subagent trace — under the same session card, carrying the same `user_id`, with each subagent trace clearly named (`subagent:general-purpose`, `subagent:bash`, …) so it stays distinguishable from the lead trace.

Trace topology is unchanged: subagent traces remain separate top-level traces in the same session. Nesting them as child spans under the lead trace is a harder follow-up (Plan B) and intentionally out of scope.

Three concrete code changes implement this:

1. **`subagents/executor.py`** — In `_aexecute`, `build_tracing_callbacks()` is appended to `run_config["callbacks"]` (alongside the existing `SubagentTokenCollector`, not replacing it) and `inject_langfuse_metadata(...)` is called with the parent `thread_id`, the constructor-supplied `user_id`, and a normalized `subagent:<name>` trace name. `_create_agent` now builds the model with `attach_tracing=False` — the same pairing the lead agent uses, so model-level tracing does not double-count with the graph-root callbacks.
2. **`tools/builtins/task_tool.py`** — `user_id` is resolved via `resolve_runtime_user_id(runtime)` at the parent tool layer (before the background thread starts) and threaded through `SubagentExecutor.__init__`. This matters because `_current_user` is a `ContextVar` and the subagent runs in `_execution_pool`, where the contextvar is not guaranteed to be propagated.
3. **`backend/CLAUDE.md`** — Tracing System section now lists `subagents/executor.py::_aexecute` as a third trace-attribute injection point alongside `runtime/runs/worker.py::run_agent` and `client.py::DeerFlowClient.stream`, and the field-mapping table documents the subagent-specific sources (user_id via `resolve_runtime_user_id`, trace name `subagent:<name>` with `_` → `-` normalization).

Name normalization is inline (lowercase, `_` → `-`) because `runtime/runs/naming.py::resolve_root_run_name` only handles the lead-agent path; there is no shared helper to call.

## Surface area

Runtime behavior change is limited to tracing attribution. No user-facing API or default setting changes shape — the Langfuse session card just gains the subagent traces that already existed as isolated top-level traces. Operators who have not enabled Langfuse are unaffected (`inject_langfuse_metadata` no-ops when Langfuse is not in the enabled providers, and `build_tracing_callbacks` returns an empty list).

## Test verification

New behavior is pinned by a dedicated test class added before the implementation was considered done (TDD):

- **Test path that locks in the new wiring:** `backend/tests/test_subagent_executor.py::TestSubagentTracingWiring`
  - `test_aexecute_appends_tracing_callbacks_to_run_config` — `build_tracing_callbacks()` output is appended (not replacing) to `run_config["callbacks"]`, and `SubagentTokenCollector` survives the merge.
  - `test_aexecute_injects_langfuse_session_user_and_trace_name` — when Langfuse is enabled, `run_config["metadata"]` carries `langfuse_session_id == parent_thread_id`, `langfuse_user_id` from the constructor, `langfuse_trace_name == "subagent:general-purpose"` (underscore→hyphen normalization), and a `model:<name>` tag.
  - `test_aexecute_skips_langfuse_metadata_when_disabled` — LangSmith-only deployments stay inert.
  - `test_user_id_defaults_when_not_supplied`, `test_trace_name_falls_back_when_config_name_empty`, `test_environment_tag_emitted_from_deer_flow_env` — pin the metadata field sources so the wiring cannot regress silently.
- The pre-existing `test_create_agent_threads_explicit_app_config_to_model_and_middlewares` was extended to assert `attach_tracing=False`, locking in the pairing with graph-root callbacks.

## Validation

```
cd backend
# Lint
uv run ruff check \
  packages/harness/deerflow/subagents/executor.py \
  packages/harness/deerflow/tools/builtins/task_tool.py \
  tests/test_subagent_executor.py
# All checks passed!

# Targeted: new + updated tests
PYTHONPATH=. uv run pytest tests/test_subagent_executor.py::TestSubagentTracingWiring -v
# 6 passed

# Full executor suite (no regression)
PYTHONPATH=. uv run pytest tests/test_subagent_executor.py
# 55 passed

# All tracing-related suites stay aligned
PYTHONPATH=. uv run pytest \
  tests/test_subagent_executor.py \
  tests/test_tracing_factory.py \
  tests/test_tracing_metadata.py \
  tests/test_worker_langfuse_metadata.py \
  tests/test_client_langfuse_metadata.py \
  tests/test_lead_agent_model_resolution.py
# 94 passed

# Harness/app boundary still holds (executor.py now imports deerflow.tracing)
PYTHONPATH=. uv run pytest tests/test_harness_boundary.py tests/test_reload_boundary.py
# 11 passed
```

End-to-end Langfuse UI verification is pending — the unit tests pin the wiring (`run_config["callbacks"]` contents and `run_config["metadata"]` keys) but do not exercise the Langfuse handler against a live Langfuse instance. Reviewers with a Langfuse project configured should confirm in Sessions view that a thread with subagent delegations shows the subagent traces under the parent session card.

### Manual E2E Validation
Verified on a live Langfuse-enabled instance:
- Subagent traces grouped under parent session card
- Parent `session_id` / `user_id` correctly inherited
- No duplicate / double-counted traces

Screenshots attached below.
截屏2026-06-17 01 10 40

截屏2026-06-17 01 11 47

截屏2026-06-17 01 40 30

截屏2026-06-17 01 12 00

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:**

- Initial Plan A implementation (staged diff: `build_tracing_callbacks` append, `inject_langfuse_metadata` call, `attach_tracing=False`, `user_id` plumbing through `task_tool.py`) was reviewed by Claude Code against the lead-agent pattern, and the review flagged the stale `build_run_config` comment reference, the inconsistent `run_config["callbacks"]` merge shape, the misleading `user_id` docstring, the missing TDD coverage, and the stale `CLAUDE.md` Tracing section.
- Polish (comments, docstring, callback merge shape) and the entire `TestSubagentTracingWiring` test class were written by Claude Code. The test class was iterated against the real executor until all six cases passed; one case (`model:` tag emission) caught a real setup issue (`self.model_name` was `None` because `_create_agent` was mocked) and was fixed by passing `parent_model` through `_make_executor` so `__init__` resolves the model name eagerly.
- I (the author) reviewed every diff line, ran the test/lint matrix above, and take responsibility for the change.

### DIFF
--- a/backend/packages/harness/deerflow/subagents/executor.py
+++ b/backend/packages/harness/deerflow/subagents/executor.py
@@ -3,6 +3,7 @@
 import asyncio
 import atexit
 import logging
+import os
 import threading
 import uuid
 from collections.abc import Callable, Coroutine
@@ -27,6 +28,7 @@ from deerflow.skills.tool_policy import filter_tools_by_skill_allowed_tools
 from deerflow.skills.types import Skill
 from deerflow.subagents.config import SubagentConfig, resolve_subagent_model_name
 from deerflow.subagents.token_collector import SubagentTokenCollector
+from deerflow.tracing import build_tracing_callbacks, inject_langfuse_metadata
 
 if TYPE_CHECKING:
     # Imported lazily at runtime inside _build_initial_state: importing
@@ -286,6 +288,7 @@ class SubagentExecutor:
         thread_data: ThreadDataState | None = None,
         thread_id: str | None = None,
         trace_id: str | None = None,
+        user_id: str | None = None,
     ):
         """Initialize the executor.
 
@@ -300,6 +303,8 @@ class SubagentExecutor:
             thread_data: Thread data from parent agent.
             thread_id: Thread ID for sandbox operations.
             trace_id: Trace ID from parent for distributed tracing.
+            user_id: User ID captured from the parent tool's runtime context.
+                When None, the tracing layer falls back to DEFAULT_USER_ID.
         """
         self.config = config
         self.app_config = app_config
@@ -316,6 +321,7 @@ class SubagentExecutor:
         self.thread_id = thread_id
         # Generate trace_id if not provided (for top-level calls)
         self.trace_id = trace_id or str(uuid.uuid4())[:8]
+        self.user_id = user_id
 
         self._base_tools = _filter_tools(
             tools,
@@ -336,7 +342,7 @@ class SubagentExecutor:
         app_config = self.app_config or get_app_config()
         if self.model_name is None:
             self.model_name = resolve_subagent_model_name(self.config, self.parent_model, app_config=app_config)
-        model = create_chat_model(name=self.model_name, thinking_enabled=False, app_config=app_config)
+        model = create_chat_model(name=self.model_name, thinking_enabled=False, app_config=app_config, attach_tracing=False)
 
         from deerflow.agents.middlewares.tool_error_handling_middleware import build_subagent_runtime_middlewares
 
@@ -522,6 +528,36 @@ class SubagentExecutor:
                 "callbacks": [collector],
                 "tags": [collector_caller],
             }
+
+            # Inject tracing callbacks at the graph level so a single subagent run
+            # produces one trace with all node / LLM / tool calls as child spans.
+            # This mirrors the lead agent pattern: graph-level tracing paired with
+            # attach_tracing=False on the model avoids double-counted traces.
+            tracing_callbacks = build_tracing_callbacks()
+            if tracing_callbacks:
+                existing_callbacks = list(run_config.get("callbacks") or [])
+                run_config["callbacks"] = [*existing_callbacks, *tracing_callbacks]
+
+            # Normalize subagent name for tracing so it matches the lead-agent
+            # naming shape (lowercase, hyphens only). Inline because there is no
+            # shared helper — runtime/runs/naming.py only handles lead-agent runs.
+            if self.config.name:
+                normalized_name = self.config.name.strip().lower().replace("_", "-")
+                assistant_id = f"subagent:{normalized_name}"
+            else:
+                assistant_id = "subagent"
+
+            # Inject Langfuse trace-attribute metadata so the subagent trace
+            # links to the parent thread and carries the correct session/user IDs.
+            inject_langfuse_metadata(
+                run_config,
+                thread_id=self.thread_id,
+                user_id=self.user_id,
+                assistant_id=assistant_id,
+                model_name=self.model_name,
+                environment=os.environ.get("DEER_FLOW_ENV") or os.environ.get("ENVIRONMENT"),
+            )
+
             context: dict[str, Any] = {}
             if self.thread_id:
                 run_config["configurable"] = {"thread_id": self.thread_id}
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -481,16 +481,16 @@ LangSmith and Langfuse are both supported. The wiring lives in two layers:
 - `factory.py::build_tracing_callbacks()` — returns the LangChain `CallbackHandler` list for the providers currently enabled via env vars (`LANGSMITH_TRACING`, `LANGFUSE_TRACING`, etc.). The handlers are attached at the **graph invocation root** for in-graph runs (`make_lead_agent` and `DeerFlowClient.stream` both append them to `config["callbacks"]` before invoking the graph) so a single run produces one trace with all node / LLM / tool calls as child spans. Standalone callers — anything that invokes a model outside such a graph (e.g. `MemoryUpdater`) — keep `create_chat_model`'s default `attach_tracing=True`, which falls back to model-level callback attachment.
 - `metadata.py::build_langfuse_trace_metadata()` — builds the Langfuse-reserved trace attributes for `RunnableConfig.metadata`. The Langfuse v4 `langchain.CallbackHandler` lifts these onto the root trace (see its `_parse_langfuse_trace_attributes`), but only when it sees `on_chain_start(parent_run_id=None)` — which is why the callbacks have to live at the graph root, not the model.
 
-**Trace-attribute injection points**: both `runtime/runs/worker.py::run_agent` (gateway path) and `client.py::DeerFlowClient.stream` (embedded path) merge the metadata into `config["metadata"]` right before constructing the graph. Caller-supplied keys win via `setdefault`, so an external `session_id` override is preserved. Field mapping:
+**Trace-attribute injection points**: both `runtime/runs/worker.py::run_agent` (gateway path) and `client.py::DeerFlowClient.stream` (embedded path) merge the metadata into `config["metadata"]` right before constructing the graph. `subagents/executor.py::_aexecute` does the same for every subagent run so subagent traces group under the parent thread's session card (carrying the parent `thread_id` → `langfuse_session_id`, the user_id captured at `task_tool` → `langfuse_user_id`, and a `subagent:<normalized-name>` trace name). Caller-supplied keys win via `setdefault`, so an external `session_id` override is preserved. Field mapping:
 
 | Langfuse field         | Source                                       |
 |-----------------------|----------------------------------------------|
 | `langfuse_session_id` | LangGraph `thread_id`                         |
-| `langfuse_user_id`    | `get_effective_user_id()` (`default` in no-auth) |
-| `langfuse_trace_name` | `RunRecord.assistant_id` / client `agent_name` (defaults to `lead-agent`) |
+| `langfuse_user_id`    | `get_effective_user_id()` (`default` in no-auth); for subagents, captured from `runtime.context` at `task_tool` time via `resolve_runtime_user_id()` |
+| `langfuse_trace_name` | `RunRecord.assistant_id` / client `agent_name` (defaults to `lead-agent`); for subagents, `subagent:<name>` (lowercased, `_` → `-`) |
 | `langfuse_tags`       | `env:<DEER_FLOW_ENV>` + `model:<model_name>`  |
 
-Returns `{}` when Langfuse is not in the enabled providers — LangSmith-only deployments are unaffected. Set `DEER_FLOW_ENV` (or `ENVIRONMENT`) to tag traces by deployment environment. Tests live in `tests/test_tracing_factory.py`, `tests/test_tracing_metadata.py`, `tests/test_worker_langfuse_metadata.py`, and `tests/test_client_langfuse_metadata.py`.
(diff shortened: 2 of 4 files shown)


backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -522,6 +528,36 @@ async def _aexecute(self, task: str, result_holder: SubagentResult | None = None
...
+            tracing_callbacks = build_tracing_callbacks()
+            if tracing_callbacks:
+                existing_callbacks = list(run_config.get("callbacks") or [])
+                run_config["callbacks"] = [*existing_callbacks, *tracing_callbacks]
+
+            # Normalize subagent name for tracing so it matches the lead-agent
+            # naming shape (lowercase, hyphens only). Inline because there is no
+            # shared helper — runtime/runs/naming.py only handles lead-agent runs.
+            if self.config.name:
+                normalized_name = self.config.name.strip().lower().replace("_", "-")
```
fancyboi999, line 545: Non-blocking / future-proofing: this normalizes case and `_`->`-` only, so a subagent name with spaces or other punctuation would pass through verbatim into `langfuse_trace_name` (e.g. `Code Reviewer` -> `subagent:code reviewer`). Harmless today since the registered subagent names are already kebab/snake and Langfuse accepts arbitrary trace names — just flagging in case you'd rather fully slugify here later.

fancyboi999 (commented):
Read the full diff and traced the wiring against the existing tracing layer — this looks correct and lands cleanly. What I verified:

- **Signature match (the main risk for this kind of change):** `inject_langfuse_metadata(config, *, thread_id, user_id, assistant_id, model_name, environment)` (`tracing/metadata.py:73-82`) matches the call in `executor.py` exactly — no kwarg drift.
- **No-op when Langfuse is off is real:** `inject_langfuse_metadata` returns early when `build_langfuse_trace_metadata()` yields `{}` (`metadata.py:99-100`), so LangSmith-only / no-provider deployments are unaffected — consistent with `test_aexecute_skips_langfuse_metadata_when_disabled`. And the `setdefault` merge keeps any caller-supplied `langfuse_*` override, same as the worker/client paths.
- **Pattern parity:** graph-root `build_tracing_callbacks()` + `attach_tracing=False` on the model mirrors the lead-agent wiring documented in `backend/CLAUDE.md`, so there's no double-counted trace. The append (`[*existing_callbacks, *tracing_callbacks]`) preserves the `SubagentTokenCollector`, which `test_aexecute_appends_tracing_callbacks_to_run_config` pins (asserts >=2 callbacks).
- **user_id capture:** `resolve_runtime_user_id` (`runtime/user_context.py:112`) and `DEFAULT_USER_ID = "default"` (`:97`) exist as the tests assume; the default-fallback assertion is accurate, and `resolve_runtime_user_id(None)` is safe (signature takes `object | None`), so calling it unconditionally in `task_tool` won't blow up when there's no runtime.
- **Mapping:** `thread_id -> langfuse_session_id` matches the field table you updated in `CLAUDE.md`.
- **Tests** are genuine regression anchors (session_id == parent thread_id, user_id == captured, `_`->`-` trace-name normalization, env tag, empty-name fallback, disabled no-op) rather than vacuous mocks.

Scope is tight — all four files serve the attribution goal and the doc is updated to match. Looks correct to me; good to merge once CI is green. One optional non-blocking nit inline.

fancyboi999:
Please fix the lint check error @heart-scalpel

backend/packages/harness/deerflow/subagents/executor.py
heart-scalpel: Initially, I intended to align the logic with the main agent by adding regex checks to filter out special characters. However, subagent configurations are predefined in `config.yaml`, making the naming rules transparent to users. Since these names are defined by users and presented back to them, I’ll leave the current logic unchanged for now. Thank you very much for your thorough review and valuable suggestions.

heart-scalpel:
@fancyboi999 Thank you very much for your thorough review and valuable suggestions.

fancyboi999:
Thanks! And your call on the subagent-name normalization is the right one — since the names come from `config.yaml` and are shown back to the operator verbatim, slugifying them further would just obscure them. That note was non-blocking anyway, so no change needed there.

For the failing **Lint Check**: it isn't a `ruff check` rule — that step passes clean. It's the `ruff format --check .` half of `make lint`, and it flags a single file: `backend/tests/test_subagent_executor.py`. A handful of long one-liners exceed the format width and ruff wants to wrap them (the inline `lambda`s in the storage stub ~L78, the long `AssertionError(...)` message ~L266, and the `monkeypatch.setattr(...)` calls ~L355). Just run `make format` in `backend/` (or `uvx ruff format tests/test_subagent_executor.py`) and commit the result — CI should go green after that.

heart-scalpel:
Bash(uvx ruff format --check tests/test_subagent_executor.py && uvx ruff check
      tests/test_subagent_executor.py)
  ⎿  1 file already formatted
     All checks passed!
  
  @fancyboi999 
I’ve finished the work and committed the changes with git amend.

fancyboi999:
Confirmed on my end. I pulled the current PR head (`cb02d2a2`) and ran the repo's lint config (`backend/ruff.toml`, line-length 240) on `tests/test_subagent_executor.py` with ruff 0.15.17 (what `uvx ruff` resolves to in CI): `ruff format --check` → *1 file already formatted*, `ruff check` → *All checks passed!*. So the formatting is good.

Two heads-ups so the red ❌ makes sense:

- The **Lint Check** workflow is currently sitting in `action_required` — GitHub holds CI runs from first-time contributors until a maintainer approves them, so it hasn't re-run since the original failure. It needs a maintainer to approve/re-run the workflow before it goes green; nothing left for you to format.
- The PR still shows just the original commit `cb02d2a2`, so if your `git commit --amend` was meant to push a change it may not have force-pushed (`git push --force-with-lease`). As it stands the file already passes lint, so you're fine either way — just flagging in case you expected to see a new commit.

heart-scalpel:
截屏2026-06-17 10 39 16
I've pushed the changes here.  The checks need to be manually re-triggered.

heart-scalpel:
> Confirmed on my end. I pulled the current PR head (`cb02d2a2`) and ran the repo's lint config (`backend/ruff.toml`, line-length 240) on `tests/test_subagent_executor.py` with ruff 0.15.17 (what `uvx ruff` resolves to in CI): `ruff format --check` → _1 file already formatted_, `ruff check` → _All checks passed!_. So the formatting is good.
> 
> Two heads-ups so the red ❌ makes sense:
> 
> * The **Lint Check** workflow is currently sitting in `action_required` — GitHub holds CI runs from first-time contributors until a maintainer approves them, so it hasn't re-run since the original failure. It needs a maintainer to approve/re-run the workflow before it goes green; nothing left for you to format.
> * The PR still shows just the original commit `cb02d2a2`, so if your `git commit --amend` was meant to push a change it may not have force-pushed (`git push --force-with-lease`). As it stands the file already passes lint, so you're fine either way — just flagging in case you expected to see a new commit.

It was my fault for not communicating clearly. I’ll lay everything out in detail next time. Thank you for your thoughtful advice and help — it’s truly appreciated.

WillemJiang (approved):

PR #3618, opened 2026-06-17
fix(sandbox): make setup-sandbox.sh script executable

zhfeng:
## Why

The `setup-sandbox.sh` script was not executable, requiring users to run it with `bash scripts/setup-sandbox.sh` explicitly instead of being able to run it directly as `./scripts/setup-sandbox.sh`.

## What changed

Changed file permissions of `scripts/setup-sandbox.sh` from 644 to 755 (added executable bit).

## Surface area

## Validation

Verified the file now has executable permissions (755).

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Automated the git workflow for creating branch, commit, and PR.

- [x] I've read and understand every line of this change and take responsibility for it.


WillemJiang (approved):

fancyboi999 (commented):
Confirmed — and it's a bit more than a convenience tweak: it actually unbreaks `make setup-sandbox` on Linux/macOS.

What I checked:

- `Makefile:92` runs the target as `@$(RUN_WITH_GIT_BASH) ./scripts/setup-sandbox.sh`, and `RUN_WITH_GIT_BASH` is **empty on non-Windows** (`Makefile:16`; it's only the git-bash shim on Windows, `:13`). So on Linux/macOS the recipe invokes the script *directly* (`./scripts/setup-sandbox.sh`), which needs the exec bit — at `100644` that fails with `permission denied` (verified the mode behavior locally: a `644` script run as `./x.sh` → `permission denied`, `755` → runs). So `make setup-sandbox` was broken on a fresh Unix checkout before this; the `755` flip fixes that, not just the manual `./` invocation mentioned in the description.
- Consistency: every other script under `scripts/` is already `100755` (`check.sh`, `serve.sh`, `docker.sh`, `config-upgrade.sh`, `deploy.sh`, …) — `setup-sandbox.sh` was the lone `100644`, so this brings it in line.
- Shebang is present (`#!/usr/bin/env bash`), so the exec bit is meaningful; the change is mode-only (no content diff), and the `bash scripts/setup-sandbox.sh` path keeps working (exec bit is irrelevant when invoked via `bash`).

Looks correct — recommend merge.

PR #3623, opened 2026-06-17
fix(sandbox): stop flagging string-literal path fragments as unsafe absolute paths

fancyboi999:
Relates to #3598

## Why

Investigating #3598 surfaced a false positive in the host-bash path guard. The guard (`validate_local_bash_command_paths`) scans the **raw command string** for absolute paths, so `/segment` sequences sitting inside string literals, f-strings, and templates get rejected as "unsafe absolute paths" — e.g. `python3 -c "print(f'/端口{port}')"`, or a REST template like `/devices/{id}/port`.

Worse, whether a fragment trips the guard depends on the character right **before** the slash: the regex has a `(?<![:\w])` look-behind, so a word char (including any CJK char) suppresses the match. That means the same literal can pass or fail unpredictably across runs — the model then loops rephrasing a command that was never touching the host filesystem at all, wasting LLM turns and wall-clock time.

This is a long-standing bug, not a recent regression — the regex and validator are byte-identical across the commit range in the issue, and it reproduces on `main`.

## What changed

The absolute-path guard now exempts matches that carry **non-ASCII characters or format braces** (`{` / `}`). Real host filesystem paths a command would open contain neither, so such a fragment is text, not a path argument.

- Now **allowed** (were wrongly blocked): `python3 -c "print(f'/端口{port}')"`, `echo '... /端口 ...'`, `/devices/{id}/port`
- Still **blocked**: `cat /etc/passwd`, `--input=/etc/secret`, and even a plain ASCII host path written inside a code string such as `open('/etc/passwd')`

The guard is explicitly best-effort and *not a security boundary* (per its own docstring). So this trades a vanishingly rare case — a real host path that itself contains non-ASCII chars or braces — for eliminating the false positives. Plain ASCII host paths are unaffected.

## Surface area

- [x] **Sandbox** — sandboxed execution (host-bash absolute-path validation)

## Bug fix verification

- Test path: `backend/tests/test_sandbox_tools_security.py::test_validate_local_bash_command_paths_allows_non_path_string_literals` (plus `test_validate_local_bash_command_paths_still_blocks_ascii_host_path_in_code` pinning the boundary).
- Red on `main` / green on this branch? **yes** — the literal cases go red on `main` (e.g. `Unsafe absolute paths in command: /devices/{id}/port`) and green here; the ASCII-host-path case stays green (still blocked) both before and after.

## Validation

```
```
- `tests/test_sandbox_tools_security.py`: **110 passed** — all pre-existing security cases (host paths, `..` traversal, cd escapes, file://, bare root) intact.
- Full-repo `ruff check .` + `ruff format --check .`: clean.
- Real tool-chain check (full `bash_tool` path → validate → virtual-path replace → real sandbox exec): `echo "健康检查 /端口 状态"` and `python3 -c "print(f'/端口{55}')"` now **execute** (real output), while `cat /etc/hostname` is **still rejected**.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Reproduced the false positive from the issue, compared two fix strategies (token-level vs shape-based exemption) with a behavior matrix over real commands, wrote the failing tests first (TDD), then added the minimal exemption. I reviewed every line.

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/tools.py
+++ b/backend/packages/harness/deerflow/sandbox/tools.py
@@ -24,6 +24,11 @@ from deerflow.sandbox.security import LOCAL_HOST_BASH_DISABLED_MESSAGE, is_host_
 from deerflow.tools.types import Runtime
 
 _ABSOLUTE_PATH_PATTERN = re.compile(r"(?<![:\w])(?<!:/)/(?:[^\s\"'`;&|<>()]+)")
+# A ``{...}`` block holding a single identifier-like placeholder (e.g. ``{id}``
+# in a REST template or ``{port}`` in an f-string). Bash brace expansion such as
+# ``{passwd,shadow}`` or ``{,.bak}`` does NOT match (commas/dots/empty inner).
+_IDENTIFIER_BRACE_BLOCK_PATTERN = re.compile(r"\{([^{}]*)\}")
+_IDENTIFIER_PATTERN = re.compile(r"[A-Za-z_][A-Za-z0-9_]*")
 _FILE_URL_PATTERN = re.compile(r"\bfile://\S+", re.IGNORECASE)
 _URL_WITH_SCHEME_PATTERN = re.compile(r"^[a-z][a-z0-9+.-]*://", re.IGNORECASE)
 _URL_IN_COMMAND_PATTERN = re.compile(r"\b[a-z][a-z0-9+.-]*://[^\s\"'`;&|<>()]+", re.IGNORECASE)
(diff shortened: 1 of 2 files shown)


willem-bd (commented):
Thanks @fancyboi999. One finding from a focused look at the new exemption:

[P1] Brace-only exemption opens a bash brace-expansion bypass for real host paths

- Location: `backend/packages/harness/deerflow/sandbox/tools.py:941` (`_is_non_path_literal_fragment` returning `True` whenever the fragment contains `{` or `}`)
- Problem: The new `{`/`}` shape exemption fires on every match that contains a brace, but bash brace expansion turns those exact matches into real, plain-ASCII host filesystem reads at runtime. With this change, these all pass `validate_local_bash_command_paths` whereas `main` rejects them:
  - `cat /etc/{passwd,shadow}` -> `cat /etc/passwd /etc/shadow`
  - `cat /etc/passwd{,.bak}` -> `cat /etc/passwd /etc/passwd.bak`
  - `cat /{etc,var}/passwd` -> `cat /etc/passwd /var/passwd`
  - `bash -c \"cat /etc/{passwd,shadow}\"`
- Evidence: Confirmed by running `validate_local_bash_command_paths` on PR head vs `main` against the four commands above (PR head: ALLOWED; `main`: BLOCKED with `Unsafe absolute paths in command: /etc/{passwd,shadow}`). The exemption check is `any(ord(ch) > 127 for ch in fragment) or "{" in fragment or "}" in fragment`, with no further structural test, and the regex `_ABSOLUTE_PATH_PATTERN` does not exclude `{}` so it captures the brace block as part of the path token. The PR description states "Still blocked: `cat /etc/passwd` ... Plain ASCII host paths are unaffected," but plain-ASCII host paths combined with brace expansion are now allowed — that invariant doesn't hold.
- Suggested fix: Tighten the brace branch so it only exempts template/format-style braces and not bash brace expansion. Two cheap options:
  1. Only exempt fragments where every `{...}` block matches a single "identifier-like" template placeholder (no commas, no `/`, no nested slashes), e.g. require the content between matching `{` and `}` to satisfy something like `re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*", inner)` — this keeps `/devices/{id}/port` exempted but rejects `/etc/{passwd,shadow}` and `/{etc,var}/passwd`.
  2. Or, before exempting, refuse the shortcut if the fragment contains `,` inside `{...}` (the brace-expansion signal) — same effect for the cases above, slightly looser.
  Either keeps the false-positive fix for the f-string / REST-template cases that motivated the PR.
- Test: Extend `test_validate_local_bash_command_paths_still_blocks_ascii_host_path_in_code` (or add a sibling) with parametrized brace-expansion cases — `cat /etc/{passwd,shadow}`, `cat /etc/passwd{,.bak}`, `cat /{etc,var}/passwd` — asserting `PermissionError` is still raised. The current happy-path tests only cover quoted f-strings / REST templates and don't exercise this branch.

Side note (sub-threshold, not blocking): the docstring rightly calls this a best-effort guard rather than a security boundary, so this is about preserving the stated guard behavior, not relying on it for isolation.

fancyboi999:
Good catch — thanks, and you're right. I reproduced it independently: ran `validate_local_bash_command_paths` (with `thread_data` set) on the PR head vs `main`:

- **PR head**: `cat /etc/{passwd,shadow}`, `cat /etc/passwd{,.bak}`, `cat /{etc,var}/passwd` → all **ALLOWED**; `cat /etc/passwd` still BLOCKED; `/devices/{id}/port` ALLOWED (the intended false-positive fix).
- **main**: all three brace cases BLOCKED (`Unsafe absolute paths in command: /etc/{passwd,shadow}` etc.).

So the blanket `{`/`}` exemption lets bash brace-expansion reconstitute plain host paths past the guard — the "plain ASCII host paths unaffected" line in the description doesn't hold once a `{...}` block is present. My miss.

Going with your option 1: only exempt when every `{...}` block is a single identifier-like placeholder (`re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*", inner)`), so `/devices/{id}/port` and the f-string / REST-template cases stay exempt while `/etc/{passwd,shadow}`, `/etc/passwd{,.bak}`, `/{etc,var}/passwd` go back to BLOCKED. I'll also add parametrized brace-expansion cases to the path-validation tests (the current ones only exercise the quoted-template happy path). Pushing the tightened check + tests.

fancyboi999:
Pushed `9d0150f` with option 1.

`_is_non_path_literal_fragment` now only exempts a braced fragment when every `{...}` block is a single identifier-like placeholder (`re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*", inner)`):

- `/devices/{id}/port`, f-string `/{port}`, and the CJK fragments stay exempt (the false positives this PR targets)
- `cat /etc/{passwd,shadow}`, `cat /etc/passwd{,.bak}`, `cat /{etc,var}/passwd` (and the `bash -c "…"` wrapper) go back to BLOCKED, matching `main`

While re-testing I hit the same bypass family one layer over: `${VAR}` expansion. `cat /home/${USER}/.ssh/id_rsa` was exempted because `USER` is identifier-shaped, but bash still expands it to a real host path at runtime. So I also disqualify any fragment containing `${`. Stray/empty/nested braces (`{}`, `/{a{b}}`) are rejected too — each `{`/`}` has to belong to one balanced single-placeholder block.

Tests: parametrized `..._blocks_brace_expansion_host_paths` covering the three brace cases + the `bash -c` wrapper + `${USER}`, asserting `PermissionError`. Confirmed red on the previous head (all four brace cases were ALLOWED before the tighten), green now; full `test_sandbox_tools_security.py` is 115 passed, `ruff check` + `ruff format --check` clean on both files.

WillemJiang (approved):

PR #3627, opened 2026-06-17
feat(frontend): implement (thought for x second) thinking duration in…

chetan655:
### Description
This PR implements a comprehensive, model-agnostic timer that measures the Total Turn Time of the agent's actions (including both tool executions and text generation), providing users with instant and continuous visual feedback from the moment they click "Send".

### Key Features & Implementation:

1. **Full-Turn Tracking**: The timer state (`turnStartTime`) is now hoisted to the `MessageList` level, allowing it to capture the exact timestamp when `thread.isLoading` becomes true.
2. **Immediate Visual Feedback**: Added a standalone `<Reasoning>` block at the bottom of the active turn that renders immediately while tools are executing, giving users a clear "Thinking..." indicator before any text is streamed.
3. **Seamless State Handoff**: Once the agent finishes tool execution and begins generating the final text response, the standalone timer unmounts and seamlessly passes the `turnStartTime` to the `MessageListItem` bubble, maintaining a continuous, uninterrupted timer.
4. **Model-Agnostic Architecture**: Transitioned the trigger mechanism to rely on Time-to-First-Token (TTFT) streaming logic, ensuring flawless compatibility with all models.
5. **UI Polish**: Cleaned up the layout by removing redundant `StreamingIndicator` components from the bottom of the chat list, centralizing the loading state into the new Thinking... block.

`Follow-up Note`: Because the backend does not currently persist the duration of the generation inside the message metadata, the exact timer value will reset to "Thought for a few seconds" upon refreshing the page. A follow-up backend issue should be opened to persist this value.

### Testing:

 * Verified timer starts immediately upon user submission and persists visually during tool calls.
 * Verified timer seamlessly transitions into the final text bubble without resetting.
 * Verified past messages render their final state correctly.
 * All pnpm test and pnpm test:e2e suites pass locally.

### Screen-shot

image


fancyboi999 (commented):
Nice idea — a live turn timer is a real UX improvement. I read the three files and traced the render paths; one thing should be fixed before merge, plus a few smaller notes.

**Blocker — every historical / non-reasoning assistant message now renders a stray "Thought for a few seconds".**
`MessageContent_` now renders `<Reasoning>` for *every* non-human message, unconditionally (`message-list-item.tsx:333`). For a historical or plain assistant message — not streaming, no `reasoningContent`, `duration === undefined` — `defaultGetThinkingMessage` falls through to `return <span>Thought for a few seconds</span>` (`reasoning.tsx:168`). So after this change, every past assistant turn (including ones the model never reasoned on) shows a Brain icon + "Thought for a few seconds". `hasContent={!!reasoningContent}` only hides the chevron — it doesn't suppress the trigger text. Before this PR, `MessageContent_` rendered no Reasoning block in the message body at all, so this is net-new for all assistant messages. Suggested gate: `{!isHuman && (reasoningContent || isLoading) && (<Reasoning>…)}`.

**Minor — live vs final rounding mismatch.** `LiveTimer` rounds with `Math.floor` (`reasoning.tsx:138`) while the final duration uses `Math.ceil` (`reasoning.tsx:84`). The shown seconds can jump +1 the instant streaming ends (live "5s" → final "6s"). Worth matching one or the other.

**Minor — fragile `useState` initializer (`reasoning.tsx:70`).** `useState(startTimeProp ?? (() => (isStreaming ? Date.now() : null)))` mixes a plain value and a lazy-initializer function in one expression. It happens to work (number → value; null → function → lazy init), but it's easy to misread/break later — a plain ternary or an explicit lazy initializer would read clearer.

**Design call for the maintainers (not a code bug).** The metric is *Total Turn Time* (from the Send click, including tool execution), but it's surfaced through the Reasoning / "thinking" block, which conventionally represents model reasoning. A turn that's mostly tool execution would read "Thought for 30 seconds" under a Brain icon, which can mislead. Might be worth a distinct label rather than overloading the reasoning block — your call.

Verified by code trace (the three files + render paths); I didn't run the frontend, so a quick visual check of the historical-message case would confirm the blocker.

frontend/src/components/workspace/messages/message-list-item.tsx
```diff
@@ -324,6 +329,14 @@ function MessageContent_({
   return (
     <AIElementMessageContent className={className}>
       {filesList}
+      {!isHuman && (
+        <Reasoning isStreaming={isLoading} startTimeProp={turnStartTime}>
```
fancyboi999, line 338: This renders `<Reasoning>` for *every* non-human message, unconditionally. For historical / no-`reasoningContent` / not-streaming messages (`duration === undefined`), the trigger ends up showing `Thought for a few seconds` (see `reasoning.tsx:168`) — a stray thinking indicator on past turns that never reasoned. `hasContent` only controls the chevron, not the trigger text.

Suggest gating so only the active turn or messages with actual reasoning render it:

```tsx
{!isHuman && (reasoningContent || isLoading) && (
  <Reasoning isStreaming={isLoading} startTimeProp={turnStartTime}>
    ...
  </Reasoning>
)}
```

frontend/src/components/ai-elements/reasoning.tsx
```diff
@@ -115,10 +123,44 @@ export const Reasoning = memo(
...
+    startTime?: number | null,
+  ) => ReactNode;
+  hasContent?: boolean;
+};
+
+const LiveTimer = ({ startTime }: { startTime: number }) => {
+  const [elapsed, setElapsed] = useState(0);
+
+  useEffect(() => {
+    const calculateElapsed = () => Math.floor((Date.now() - startTime) / 1000);
```
fancyboi999: `LiveTimer` rounds elapsed with `Math.floor` here, but the final duration is computed with `Math.ceil` (line 84). The displayed seconds can tick up by 1 the moment streaming ends (live "5s" → final "6s"). Consider matching them so the number doesn't jump on handoff.

chetan655:
@fancyboi999 
Thank you for the incredibly thorough review 

I've pushed updates to address all your notes:

1. **Blocker (Historical Messages)**:  Instead of just `(reasoningContent || isLoading)`, I actually implemented a local `wasLoading` state trick in `MessageListItem`. This is the best of both worlds: It keeps the timer permanently visible for the current session, but it instantly hides on page refresh for historical messages that don't have reasoning content, completely avoiding the stray "Thought for a few seconds" bug.
2. **LiveTimer Rounding**:  Switched the `LiveTimer` to use `Math.ceil` so it perfectly matches the final duration logic without jumping.
3. **useState Initializer**: Wrapped the fallback in an explicit lazy-initializer arrow function `() =>` so it evaluates properly and is easier to read.

Regarding the Design Call: That is a very fair point about "Thought for 30s" capturing both tool execution and reasoning. However, since apps like Grok conventionally wrap the entire backend TTFT (including initial routing and tool selection) into a single "Thinking..." metric for simplicity, I think it works great for now as a unified "Total Turn Time" metric. If we want to get granular later and split out "Tool Execution Time" vs "Pure LLM Reasoning Time", we can definitely iterate on that in a future PR!

WillemJiang:
@chetan655, please resolve the conflict in the main branch.

fancyboi999:
Thanks @chetan655 — went through the `wasLoading` approach on the current head (`message-list-item.tsx`) and it does resolve the regression I flagged, cleanly:

- A historical assistant message on refresh mounts with `isLoading=false`, so `wasLoading` initializes `false` and the effect never latches it; with no `reasoningContent`, the gate `!isHuman && (!!reasoningContent || wasLoading)` is `false` → `<Reasoning>` isn't rendered → no stray "Thought for a few seconds".
- A message that streamed in the live session keeps the timer (`wasLoading` latched `true`), and one that actually had reasoning still renders via `reasoningContent`.

That's arguably better than the `(reasoningContent || isLoading)` gate I'd suggested — mine would have dropped the timer the instant `isLoading` flipped false post-turn, whereas `wasLoading` keeps it visible for the session. The latch effect reads correctly too (idempotent, never un-latches). `Math.ceil` and the lazy initializer both look good.

On the Total-Turn-Time design call: agreed, a single unified metric is a reasonable v1; splitting tool-exec vs pure-reasoning can be a later iteration.

(Verified by reading the gate/latch logic, not by running the UI.) Blocker-wise this looks resolved to me — only thing left is the merge conflict @WillemJiang flagged.

chetan655:
@WillemJiang Done 🚀

WillemJiang (approved):

PR #3629, opened 2026-06-17
fix: preserve sandbox reducer in middleware state

### LINKED ISSUE #3628: [bug] Subagent 并发 sandbox 工具调用时可能触发 INVALID_CONCURRENT_GRAPH_UPDATE
### Before you start

### Problem summary

在 subagent 执行过程中，如果同一个 graph step 内有多个 sandbox 工具并发懒初始化 sandbox，LangGraph报错

### Affected area(s)

## 问题描述

在 subagent 执行过程中，如果同一个 graph step 内有多个 sandbox 工具并发懒初始化 sandbox，LangGraph 可能报错：

```text
langgraph.errors.InvalidUpdateError: At key 'sandbox': Can receive only one value per step. Use an Annotated key to handle multiple values.
```
实际日志示例：
```txt
Subagent general-purpose async execution failed
langgraph.errors.InvalidUpdateError: At key 'sandbox': Can receive only one value per step. Use an Annotated key to handle multiple values.
```

### Expected behavior

同一步内多个工具写入相同的 sandbox_id 时，应通过 merge_sandbox 幂等合并，不应失败。
如果写入不同 sandbox_id，则继续由 merge_sandbox 抛出冲突错误。

### Steps to reproduce

-

### Relevant logs

```shell
-
```

Docker (make docker-start)

### Operating system

Linux

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

## 根因分析

`SandboxMiddleware.wrap_tool_call/awrap_tool_call` 会在懒初始化 sandbox 后返回：

```python
Command(update={"sandbox": {"sandbox_id": sandbox_id}, "messages": [...]})
```

当同一个 graph step 内多个 sandbox 工具都懒初始化并写入相同的 `sandbox_id` 时，理论上应该由 `merge_sandbox` 幂等合并。

但是 `SandboxMiddlewareState` 自己声明了 `sandbox` 字段，且没有 `Annotated[..., merge_sandbox]`。LangGraph 合并 middleware state schema 时，这个局部 schema 可能覆盖 `ThreadState.sandbox` 上的 reducer，导致 `sandbox` 退化成 LastValue channel，于是同一步多个写入触发：

```text
Can receive only one value per step
```

## 建议修复

让 `SandboxMiddlewareState.sandbox` 保留和 `ThreadState.sandbox` 一致的 reducer：

```python
from typing import Annotated, NotRequired, override

from deerflow.agents.thread_state import SandboxState, ThreadDataState, merge_sandbox

class SandboxMiddlewareState(AgentState):
    sandbox: Annotated[NotRequired[SandboxState | None], merge_sandbox]
    thread_data: NotRequired[ThreadDataState | None]
```

Huixin615:
Fixes #3628

## Why

Subagent sandbox tools can lazily initialize the same sandbox within the same LangGraph step. When multiple tool calls emit the same `sandbox` state update concurrently, LangGraph should merge those idempotent writes through `merge_sandbox` instead of treating `sandbox` as a last-value channel and raising `INVALID_CONCURRENT_GRAPH_UPDATE`.

## What changed

`SandboxMiddlewareState.sandbox` now keeps the same `merge_sandbox` reducer as `ThreadState.sandbox`, so concurrent writes of the same `sandbox_id` merge safely while conflicting sandbox ids still fail closed.
SandboxStateField alias
Added a regression test to ensure the middleware-local state schema preserves the sandbox reducer.

## Surface area

## Screenshots / Recording

N/A. Backend-only bug fix.

## Bug fix verification

  - `backend/tests/test_sandbox_middleware.py::test_sandbox_middleware_state_matches_thread_state_sandbox_field`
  - Existing reducer behavior is covered by `backend/tests/test_thread_state_reducers.py`
- Did it go red on `main` and green on this branch? No. I added schema-level regression coverage for the reducer wiring because reproducing the full concurrent LangGraph subagent step in a small unit test would require heavier graph orchestration.
- The test verifies the exact regression point: `SandboxMiddlewareState.sandbox` must carry `merge_sandbox` metadata.

## Validation

```shell
cd backend && uv run pytest tests/test_sandbox_middleware.py tests/test_thread_state_reducers.py
cd backend && uv run ruff check packages/harness/deerflow/sandbox/middleware.py tests/test_sandbox_middleware.py tests/test_thread_state_reducers.py
cd backend && uv run ruff format --check packages/harness/deerflow/sandbox/middleware.py tests/test_sandbox_middleware.py tests/test_thread_state_reducers.py
git diff --check

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/middleware.py
+++ b/backend/packages/harness/deerflow/sandbox/middleware.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 from collections.abc import Awaitable, Callable
 from dataclasses import replace as dc_replace
-from typing import Annotated, NotRequired, override
+from typing import NotRequired, override
 
 from langchain.agents import AgentState
 from langchain.agents.middleware import AgentMiddleware
@@ -11,7 +11,7 @@ from langgraph.prebuilt.tool_node import ToolCallRequest
 from langgraph.runtime import Runtime
 from langgraph.types import Command
 
-from deerflow.agents.thread_state import SandboxState, ThreadDataState, merge_sandbox
+from deerflow.agents.thread_state import SandboxStateField, ThreadDataState
 from deerflow.sandbox import get_sandbox_provider
 
 logger = logging.getLogger(__name__)
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 class SandboxMiddlewareState(AgentState):
     """Compatible with the `ThreadState` schema."""
 
-    sandbox: Annotated[NotRequired[SandboxState | None], merge_sandbox]
+    sandbox: SandboxStateField
     thread_data: NotRequired[ThreadDataState | None]
 
 
--- a/backend/packages/harness/deerflow/sandbox/middleware.py
+++ b/backend/packages/harness/deerflow/sandbox/middleware.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 from collections.abc import Awaitable, Callable
 from dataclasses import replace as dc_replace
-from typing import NotRequired, override
+from typing import Annotated, NotRequired, override
 
 from langchain.agents import AgentState
 from langchain.agents.middleware import AgentMiddleware
@@ -11,7 +11,7 @@ from langgraph.prebuilt.tool_node import ToolCallRequest
 from langgraph.runtime import Runtime
 from langgraph.types import Command
 
-from deerflow.agents.thread_state import SandboxState, ThreadDataState
+from deerflow.agents.thread_state import SandboxState, ThreadDataState, merge_sandbox
 from deerflow.sandbox import get_sandbox_provider
 
 logger = logging.getLogger(__name__)
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 class SandboxMiddlewareState(AgentState):
     """Compatible with the `ThreadState` schema."""
 
-    sandbox: NotRequired[SandboxState | None]
+    sandbox: Annotated[NotRequired[SandboxState | None], merge_sandbox]
     thread_data: NotRequired[ThreadDataState | None]
 
 
--- a/backend/packages/harness/deerflow/agents/thread_state.py
+++ b/backend/packages/harness/deerflow/agents/thread_state.py
@@ -39,6 +39,9 @@ def merge_sandbox(existing: SandboxState | None, new: SandboxState | None) -> Sa
     raise ValueError(f"Conflicting sandbox state updates: {existing_id!r} != {new_id!r}")
 
 
+SandboxStateField = Annotated[NotRequired[SandboxState | None], merge_sandbox]
+
+
 def merge_artifacts(existing: list[str] | None, new: list[str] | None) -> list[str]:
     """Reducer for artifacts list - merges and deduplicates artifacts."""
     if existing is None:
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: backend/packages/harness/deerflow/sandbox/middleware.py [hunk 1]
```diff
@@ -20,7 +20,7 @@
 class SandboxMiddlewareState(AgentState):
     """Compatible with the `ThreadState` schema."""
 
-    sandbox: NotRequired[SandboxState | None]
+    sandbox: Annotated[NotRequired[SandboxState | None], merge_sandbox]
```
fancyboi999: This now mirrors `ThreadState.sandbox` exactly, which is the whole point. Since the class docstring promises "Compatible with the `ThreadState` schema", it might be worth a tiny guard so the two can't silently drift — e.g. assert both fields resolve to the same reducer rather than just asserting `merge_sandbox` is present:

```python
assert (
    get_type_hints(SandboxMiddlewareState, include_extras=True)["sandbox"].__metadata__
    == get_type_hints(ThreadState, include_extras=True)["sandbox"].__metadata__
)
```

That way a future swap of ThreadState's reducer can't quietly leave the middleware behind. Purely optional — the current test already prevents the reducer from being dropped.

fancyboi999 (commented):
Read through the change against `origin/main`. Solid, surgical fix that matches the existing pattern.

What I checked:
- `merge_sandbox` (thread_state.py) is idempotent for the same `sandbox_id` and fails closed on a conflicting id — exactly the semantics #3628 needs. `ThreadState.sandbox` already carries this same `Annotated[..., merge_sandbox]` binding (thread_state.py:109), so this PR just aligns the middleware-local schema to the canonical one.
- `SandboxMiddleware` sets `state_schema = SandboxMiddlewareState` (middleware.py:38), so LangGraph actually consumes the reducer from the schema metadata — the annotation isn't dead.
- The new test asserts `merge_sandbox in hints["sandbox"].__metadata__`, which is RED without this change, and it mirrors the existing `test_sandbox_field_is_wired_to_merge_sandbox` in test_thread_state_reducers.py. The end-to-end concurrency behavior is already covered by the `merge_sandbox` reducer tests there, so the structural assertion here is enough to lock the wiring.

One forward-looking nit inline; not a blocker. Looks correct to me.

willem-bd:
Correct, minimal fix. One suggestion:

**Drift hazard from duplicate annotation.** The same `Annotated[NotRequired[SandboxState | None], merge_sandbox]` now exists in two places (`ThreadState.sandbox` and `SandboxMiddlewareState.sandbox`). If someone later changes the reducer or the value type, they must remember both. Consider extracting a shared alias in `thread_state.py`:

```python
SandboxStateField = Annotated[NotRequired[SandboxState | None], merge_sandbox]
```

and using it in both schemas. As a lighter alternative, strengthen the test to assert `SandboxMiddlewareState.sandbox`'s metadata matches `ThreadState.sandbox`'s metadata exactly, so future divergence is caught.

Huixin615:
> Correct, minimal fix. One suggestion:
> 
> **Drift hazard from duplicate annotation.** The same `Annotated[NotRequired[SandboxState | None], merge_sandbox]` now exists in two places (`ThreadState.sandbox` and `SandboxMiddlewareState.sandbox`). If someone later changes the reducer or the value type, they must remember both. Consider extracting a shared alias in `thread_state.py`:
> 
> ```python
> SandboxStateField = Annotated[NotRequired[SandboxState | None], merge_sandbox]
> ```
> 
> and using it in both schemas. As a lighter alternative, strengthen the test to assert `SandboxMiddlewareState.sandbox`'s metadata matches `ThreadState.sandbox`'s metadata exactly, so future divergence is caught.

Thanks for your suggestions, I have updated them.

WillemJiang (approved):

PR #3535, opened 2026-06-12
fix(history): strip base64 image data from REST endpoint responses (#3496)

hata33:
## Why

  When a thread contains images viewed via `view_image`, the REST endpoints that return checkpoint channel values include multi-megabyte base64
  payloads in the JSON response, causing severe frontend lag or freezes. After uploading and viewing multiple images in a conversation,
  refreshing the page or re-entering the thread triggers `/history` which returns all persisted base64 data, freezing the UI for seconds.

  Reported in #3496.

  ## Root cause

  `ViewImageMiddleware.before_model` (line 120-125) injects a `HumanMessage` with `image_url: data:image/...;base64,...` content blocks, marked
  `hide_from_ui: True`. These messages are persisted in the LangGraph checkpoint. All six REST endpoints returning
  `serialize_channel_values(channel_values)` sent these base64 payloads verbatim — the serialization layer had no awareness of `hide_from_ui` or
  `data:` scheme image URLs.

  While the frontend's `isHiddenFromUIMessage()` filters these from display, the network response is still huge because filtering happens
  client-side after the full download.

  ## What changed

  - Added `strip_data_url_image_blocks()` in `serialization.py` — removes `data:`-scheme `image_url` content blocks from `hide_from_ui`
  messages. Text blocks, `https://` image URLs, and non-hidden messages are left untouched so message ordering and count are preserved.
  - Added `serialize_channel_values_for_api()` — convenience wrapper combining `serialize_channel_values` + `strip_data_url_image_blocks`, for
  use in all REST endpoints that return channel values to the frontend.
  - Replaced `serialize_channel_values()` with `serialize_channel_values_for_api()` at all six call sites across three routers:
    - `GET /{thread_id}` (`threads.py`)
    - `GET /{thread_id}/state` (`threads.py`)
    - `POST /{thread_id}/state` (`threads.py`)
    - `POST /{thread_id}/history` (`threads.py`)
    - `POST /api/runs/wait` (`runs.py`)
    - `POST /{thread_id}/runs/wait` (`thread_runs.py`)

  Externally observable behavior is unchanged — same response shapes, same message ordering. The only difference is that `data:`-scheme
  `image_url` blocks are no longer included in `hide_from_ui` messages, dramatically reducing response sizes for threads with viewed images.

  SSE streaming (`worker.py`) is unaffected — it still uses `serialize_channel_values()` directly so base64 image data continues to reach the
  model during active runs.

  ## Surface area

[...]

PR #3631, opened 2026-06-17
fix(serialization): strip base64 image data from streamed values events

he-yufeng:
## Why

`#3535` stopped the REST endpoints from leaking base64 image data to the frontend: `ViewImageMiddleware` stores full base64 image payloads inside `hide_from_ui` human messages, and the wait/history/state endpoints now run those through `serialize_channel_values_for_api`, which drops the `data:` image blocks.

The SSE streaming path was missed. With the default stream mode (`values`), the worker publishes full-state snapshots via `serialize(chunk, mode="values")`, and that branch still went through `serialize_channel_values` (no stripping). So the exact payload `#3535` removed from the REST responses was still streamed to the frontend over SSE on every `values` event.

## What changed

`serialize(..., mode="values")` now routes through `serialize_channel_values_for_api` instead of `serialize_channel_values`, so streamed state snapshots drop base64 `data:` image blocks from `hide_from_ui` messages, matching the REST endpoints. Message count and ordering are preserved; non-hidden messages and `https://` image URLs are left untouched. `serialize()` is only used by the gateway SSE publisher, so persisted/internal serialization is unaffected.

## Surface area

- [x] **Backend API** — SSE event / response shape (`values` stream events)
- [ ] **Default behavior change** — no opt-in change; this removes data that was never meant to reach the client

## Bug fix verification

Before: a run whose state contains a `hide_from_ui` human message with a `data:image/...;base64,...` block streams that base64 to the client in every `values` SSE event.

After: the `values` event keeps the message (count/order intact) but the base64 `data:` image block is gone; `https://` image URLs and non-hidden messages are still sent.

## Validation

- Added `test_serialize_values_mode_strips_base64_from_hidden_messages` in `backend/tests/test_serialization.py`, alongside the existing `serialize_channel_values_for_api` / `strip_data_url_image_blocks` tests from #3535.
- `ruff check` and `ruff format --check` pass on the changed files.
- I reproduced the leak and confirmed the fix with a standalone load of `serialization.py` (it only depends on `typing`): on `main`, `serialize(state, mode="values")` keeps the base64; with the change it is stripped while `__pregel_*` keys, non-hidden messages, and `https` URLs are preserved. The full `make test` suite needs the backend workspace env (langgraph etc.), so I rely on CI to run the pytest test end to end.

## AI assistance

**Tool:** Claude Code.
**How:** It helped trace the streaming serialization path and spot that it was the unfixed sibling of #3535, and drafted the one-line fix and the test. I reviewed the serialization module, confirmed `serialize()` is used only by the SSE publisher (not persistence), and verified the behavior with the standalone reproduction above.

### DIFF
--- a/backend/packages/harness/deerflow/runtime/serialization.py
+++ b/backend/packages/harness/deerflow/runtime/serialization.py
@@ -133,11 +133,14 @@ def serialize(obj: Any, *, mode: str = "") -> Any:
     """Serialize LangChain objects with mode-specific handling.
 
     * ``messages`` — obj is ``(message_chunk, metadata_dict)``
-    * ``values`` — obj is the full state dict; ``__pregel_*`` keys stripped
+    * ``values`` — obj is the full state dict; ``__pregel_*`` keys stripped and
+      base64 ``data:`` image blocks dropped from hide_from_ui messages
     * everything else — recursive ``model_dump()`` / ``dict()`` fallback
     """
     if mode == "messages":
         return serialize_messages_tuple(obj)
     if mode == "values":
-        return serialize_channel_values(obj) if isinstance(obj, dict) else serialize_lc_object(obj)
+        # ``values`` snapshots stream the full state to the frontend, so they
+        # must drop base64 image payloads the same way the REST endpoints do.
+        return serialize_channel_values_for_api(obj) if isinstance(obj, dict) else serialize_lc_object(obj)
     return serialize_lc_object(obj)
--- a/backend/tests/test_serialization.py
+++ b/backend/tests/test_serialization.py
@@ -328,3 +328,32 @@ def test_serialize_channel_values_for_api_no_messages():
 
     result = serialize_channel_values_for_api({"title": "empty"})
     assert result == {"title": "empty"}
+
+
+def test_serialize_values_mode_strips_base64_from_hidden_messages():
+    """The SSE stream emits ``values`` snapshots of the full state, so it must
+    strip base64 image data from hide_from_ui messages just like the REST
+    endpoints do — otherwise the same payload leaks over the stream."""
+    import json
+
+    from deerflow.runtime.serialization import serialize
+
+    state = {
+        "messages": [
+            _make_msg(
+                [
+                    {"type": "text", "text": "context"},
+                    {
+                        "type": "image_url",
+                        "image_url": {"url": "data:image/png;base64,iVBOR..."},
+                    },
+                ],
+                hide_from_ui=True,
+            ),
+        ],
+    }
+    result = serialize(state, mode="values")
+    # the hidden message survives (count/order preserved) but the data: block is gone
+    assert len(result["messages"]) == 1
+    assert "data:image/png;base64" not in json.dumps(result)
+    assert result["messages"][0]["content"] == [{"type": "text", "text": "context"}]


fancyboi999 (commented):
Read the change against `origin/main`. Clean, correctly scoped — the `values` branch was indeed the unfixed sibling of #3535's REST-path fix.

What I checked:
- `serialize()` is only reached from the gateway SSE publisher (worker.py:318 single-mode and :337 multi-mode); the `client.py:559` hit is just a docstring. So routing `values` through `serialize_channel_values_for_api` changes the streamed payload only — persisted/internal serialization is untouched, matching your Surface-area note.
- `serialize_channel_values_for_api` wraps `serialize_channel_values` and then strips `data:` image blocks from `hide_from_ui` messages only; count/order, non-hidden messages and `https://` URLs are preserved (verified against the function body).
- Because it reuses `serialize_channel_values`, the `__interrupt__` key stays preserved — the deliberate carve-out from #3595. So this doesn't regress interrupt detection on `values` chunks, which is the easy thing to break here.
- Both the single-mode and multi-mode publish paths funnel through the same `serialize()`, so every `values` event is covered, not just one path.
- The new test is RED on main (the old branch keeps the base64). Honest call-out that full `make test` needs the backend env and you leaned on CI for the e2e pytest run — reasonable.

Looks correct to me.

backend/packages/harness/deerflow/runtime/serialization.py
```diff
@@ -133,11 +133,14 @@ def serialize(obj: Any, *, mode: str = "") -> Any:
...
+      base64 ``data:`` image blocks dropped from hide_from_ui messages
     * everything else — recursive ``model_dump()`` / ``dict()`` fallback
     """
     if mode == "messages":
         return serialize_messages_tuple(obj)
     if mode == "values":
-        return serialize_channel_values(obj) if isinstance(obj, dict) else serialize_lc_object(obj)
+        # ``values`` snapshots stream the full state to the frontend, so they
+        # must drop base64 image payloads the same way the REST endpoints do.
+        return serialize_channel_values_for_api(obj) if isinstance(obj, dict) else serialize_lc_object(obj)
```
fancyboi999, line 145: This one line is the whole fix and it's the right lever: `serialize_channel_values_for_api` is a strict superset of `serialize_channel_values` (same `__pregel_*` stripping, same `__interrupt__` carve-out from #3595, plus the base64 strip), so both `values` callers in worker.py inherit the REST behavior without touching the call sites. Nothing to change here.

willem-bd, line 145: LGTM — this is the right call. `serialize(values=)` is now the SSE sibling of the REST endpoints fixed in #3535, and dropping the base64 also shrinks the streamed payload as a bonus.

Verified that `serialize()` is only invoked from the gateway SSE publisher (`worker.py:318`/`337`), so persisted/internal serialization is unaffected by this routing change.

backend/packages/harness/deerflow/runtime/serialization.py
```diff
@@ -133,11 +133,14 @@ def serialize(obj: Any, *, mode: str = "") -> Any:
     """Serialize LangChain objects with mode-specific handling.
 
     * ``messages`` — obj is ``(message_chunk, metadata_dict)``
-    * ``values`` — obj is the full state dict; ``__pregel_*`` keys stripped
+    * ``values`` — obj is the full state dict; ``__pregel_*`` keys stripped and
+      base64 ``data:`` image blocks dropped from hide_from_ui messages
     * everything else — recursive ``model_dump()`` / ``dict()`` fallback
     """
     if mode == "messages":
         return serialize_messages_tuple(obj)
```
willem-bd, line 141: Follow-up worth tracking (not a blocker for this PR): this fix scopes to the `values` mode, mirroring what #3535 covered. But the frontend can also stream `messages`-mode deltas via `serialize_messages_tuple`, and if `ViewImageMiddleware`-injected `hide_from_ui` messages ever surface on that stream, the same base64 could still leak over SSE. Worth confirming whether that path is reachable so the two stream modes are treated consistently — a separate issue is fine.

willem-bd (commented):
Code review — verified by tracing the code path, not just reading the description.

**Verdict: approve.** Minimal, precisely-scoped, and correct. `serialize_channel_values_for_api` = `serialize_channel_values` + base64 stripping, so `__pregel_*` removal and `__interrupt__` preservation (#3595) are inherited unchanged; the only new behavior is dropping `data:` image blocks from `hide_from_ui` messages, and message count/order are preserved. Test coverage is solid (`test_serialize_values_mode_strips_base64_from_hidden_messages` hits the real `serialize(state, mode="values")` entry point).

Key risk question answered: `serialize()` is **SSE-only** — its only runtime callers are the two `bridge.publish(...)` sites in `worker.py:318`/`337` (the `client.py:559` hit is a docstring). Persistence/checkpointer use LangGraph's own serializers, so persisted/internal state is unaffected.

Two non-blocking follow-ups noted inline below.

WillemJiang (approved):

PR #3637, opened 2026-06-18
feat: support regenerating latest answer

### LINKED ISSUE #3634: [feat] regenerate the answer in deerflow disucssion flow
### Discussed in https://github.com/bytedance/deer-flow/discussions/3596

<div type='discussions-op-text'>

<sup>Originally posted by **Kakar0tt0** June 15, 2026</sup>
image
主流大模型都有对某个对话进行重新生成的需求，deerflow有没有这个接口能力</div>

Huixin615:
Fixes #3634

## Why

Users can currently copy assistant responses, but there is no way to regenerate the latest answer in an existing conversation. Retrying by simply resending the last prompt is unsafe because the model would see the old assistant answer in context.

This PR adds a regenerate flow that reruns the latest completed assistant turn from the correct checkpoint while preserving existing run records for auditability, feedback, and token accounting.

## What changed

- Added a backend prepare endpoint for regenerating the latest assistant answer:
  - `POST /api/threads/{thread_id}/runs/regenerate/prepare`
  - validates that the target message is the latest visible assistant message
  - finds the previous visible human message
  - resolves the base checkpoint before that human message
  - returns clean input, checkpoint, metadata, and the source run id

- Completed checkpoint support in the standard runs API:
  - `checkpoint` / `checkpoint_id` are now applied to run config
  - checkpoint ownership is validated before execution

- Added a regenerate button beside the assistant copy action in the chat UI.

- Added frontend history filtering so regenerated runs supersede the previous run without deleting old data.

- Preserved compatibility for local/dev environments where `run_events` may not be persisted by falling back to the latest successful run summary when event rows are missing.

## Surface area

## Screenshots / Recording

Verified locally in the browser:

- The latest assistant answer shows a `Regenerate` / `重新生成` action next to copy.
- Clicking regenerate creates a new successful run.
- The old answer is hidden from the main chat view.
- The regenerated answer is displayed after reload.
- Token usage updates to include the regenerated run.
image
image

## Bug fix verification

This is a feature PR.

## Validation

Backend:

```bash
cd backend
uv run pytest tests/test_gateway_services.py tests/test_thread_regenerate_prepare.py tests/test_thread_run_messages_pagination.py
Result:
63 passed
Frontend:
cd frontend
pnpm exec tsc --noEmit
pnpm exec eslint . --ext .ts,.tsx
pnpm exec vitest run tests/unit/core/threads/message-merge.test.ts
Result:
typecheck passed
eslint passed
35 tests passed

### CODE REVIEW

File: backend/app/gateway/routers/thread_runs.py [hunk 1]
```diff
@@ -131,11 +144,215 @@ def _record_to_response(record: RunRecord) -> RunResponse:
...
+        logger.exception("Failed to list checkpoints for regenerate thread %s", thread_id)
+        raise HTTPException(status_code=500, detail="Failed to inspect checkpoint history") from exc
+
+    previous_checkpoint = None
+    for checkpoint_tuple in reversed(checkpoints):
+        messages = _checkpoint_messages(checkpoint_tuple)
+        message_ids = {_message_id(message) for message in messages}
+        if human_message_id in message_ids:
+            if previous_checkpoint is None:
+                raise HTTPException(status_code=409, detail="Could not find checkpoint before target human message")
```
fancyboi999: This 409 fires when the target human message lives in the *earliest* checkpoint (`previous_checkpoint` is still None). The happy-path test sidesteps it by seeding an empty `ckpt-base` before the human checkpoint — but is that guaranteed in production for the **first turn** of a thread? If LangGraph/worker doesn't leave an addressable checkpoint *before* the first human message lands, regenerating the very first answer would hit "Could not find checkpoint before target human message" instead of regenerating from an empty base. Worth a test for the `previous is None` case to confirm it either regenerates cleanly or fails intentionally — right now that path is untested.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -131,11 +144,215 @@ def _record_to_response(record: RunRecord) -> RunResponse:
...
+    if fallback_record is not None:
+        return fallback_record.run_id
+    raise HTTPException(status_code=409, detail="Could not find source run for assistant message")
+
+
+async def _find_base_checkpoint_before_human(thread_id: str, human_message_id: str, request: Request) -> Any:
+    checkpointer = get_checkpointer(request)
+    base_config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
+    try:
+        checkpoints = [item async for item in checkpointer.alist(base_config, limit=200)]
```
fancyboi999: `limit=200` here (and the `list_messages(..., limit=200)` at :253) caps how far back the scan goes. Regenerating the *latest* answer is almost always within that window, so this is minor — but on a very long thread the target human/checkpoint could fall outside it and surface as a 409 / a fallback run id rather than a clear "thread too long" signal. Fine to leave as-is; flagging it so it's a conscious choice.

fancyboi999 (commented):
Read the backend path end to end against `main` (head b4f774365). This is a well-built feature — the security and data-preservation design in particular hold up.

What I verified:
- **Checkpoint ownership is sound.** The prepare endpoint carries `@require_permission("runs", "create", owner_check=True, require_existing=True)` (thread_runs.py:241), and `apply_checkpoint_to_run_config` (services.py) resolves the checkpoint via `aget_tuple` using the *request* `thread_id`, not the client-supplied one — it 400s on a `thread_id` mismatch and 404s when the checkpoint isn't in that thread's namespace. So a caller can't pull another thread's checkpoint even knowing its id; thread isolation gives you checkpoint isolation.
- **Latest-only guard is correct.** `_prepare_regenerate_payload` rejects non-latest and non-visible assistant messages (409s), and `_clean_human_message_for_regenerate` strips the injected `<uploaded_files>` / `hide_from_ui` / `ORIGINAL_USER_CONTENT_KEY` so the rerun sees the real prompt, not the augmented one — that's the actual point of regenerate vs. a plain resend.
- **No data loss.** The frontend supersedes via `supersededRunIds` filtering (hooks.ts), not deletion — old runs stay for audit/feedback/token accounting, matching the description. message-merge.test.ts covers 10 merge/dedup/optimistic cases, which is thorough.

Two boundary notes inline — neither blocks; the first is worth a quick confirm.

Overall this looks correct and mergeable to me.

Huixin615:
@fancyboi999 

Thanks for the careful review.

I addressed both boundary notes in the latest push:

- Added explicit test coverage for the `previous_checkpoint is None` case. The behavior is intentionally a 409 because without an addressable checkpoint before the target human message we cannot safely construct a LangGraph branch from an arbitrary synthetic empty checkpoint.
- Made that 409 message clearer: `Could not find an addressable checkpoint before the target user message`.
- Lifted the checkpoint/event scan window into `REGENERATE_HISTORY_SCAN_LIMIT`.
- Added a test for the recent checkpoint scan limit path and updated the error to include `(limit=200)`, so long-thread failures are clearer and the limit is now an explicit choice.

Validation:
- `uv run pytest tests/test_thread_regenerate_prepare.py tests/test_gateway_services.py tests/test_thread_run_messages_pagination.py`
- `uv run ruff check app/gateway/routers/thread_runs.py tests/test_thread_regenerate_prepare.py`

Could you please another review when you have time？ Thank you very much！

fancyboi999 (commented):
Re-reviewed the new push (3acaacda1) — both notes are handled cleanly, thanks for the fast turnaround.

What I checked in the delta:
- **Scan window is now explicit.** `REGENERATE_HISTORY_SCAN_LIMIT = 200` (thread_runs.py:32) replaces both magic `200`s (`_find_target_run_id` :254, `_find_base_checkpoint_before_human` :276), and both 409 messages now read clearly — the earliest-checkpoint one is `"Could not find an addressable checkpoint before the target user message"` and the out-of-window one carries `(limit=200)`.
- **The boundary I flagged is now pinned by tests.** `test_prepare_regenerate_payload_requires_addressable_checkpoint_before_human` builds a thread with a single `[human, ai]` checkpoint and no earlier base, and asserts the 409 + the new message + `alist_limits == [200]` — exactly the `previous_checkpoint is None` path that was untested before. `test_prepare_regenerate_payload_reports_recent_checkpoint_scan_limit` covers the 201-checkpoint out-of-window case. Both are real red anchors, not vacuous.

On the design call: returning a deliberate 409 for the first-turn / no-addressable-base case — rather than synthesizing an empty checkpoint to branch from — is the safe choice, and I agree with it. The only product implication is that the very first answer in a thread can't be regenerated when there's no earlier checkpoint; that's a reasonable trade-off to leave for later, not something this PR needs to solve.

From my side this looks correct and ready. Nice work.

Huixin615:
> Re-reviewed the new push ([3acaacd](https://github.com/bytedance/deer-flow/commit/3acaacda1142ce7c6a67bdddcbefffa5da616c93)) — both notes are handled cleanly, thanks for the fast turnaround.
> 
> What I checked in the delta:
> 
> * **Scan window is now explicit.** `REGENERATE_HISTORY_SCAN_LIMIT = 200` (thread_runs.py:32) replaces both magic `200`s (`_find_target_run_id` :254, `_find_base_checkpoint_before_human` :276), and both 409 messages now read clearly — the earliest-checkpoint one is `"Could not find an addressable checkpoint before the target user message"` and the out-of-window one carries `(limit=200)`.
> * **The boundary I flagged is now pinned by tests.** `test_prepare_regenerate_payload_requires_addressable_checkpoint_before_human` builds a thread with a single `[human, ai]` checkpoint and no earlier base, and asserts the 409 + the new message + `alist_limits == [200]` — exactly the `previous_checkpoint is None` path that was untested before. `test_prepare_regenerate_payload_reports_recent_checkpoint_scan_limit` covers the 201-checkpoint out-of-window case. Both are real red anchors, not vacuous.
> 
> On the design call: returning a deliberate 409 for the first-turn / no-addressable-base case — rather than synthesizing an empty checkpoint to branch from — is the safe choice, and I agree with it. The only product implication is that the very first answer in a thread can't be regenerated when there's no earlier checkpoint; that's a reasonable trade-off to leave for later, not something this PR needs to solve.
> 
> From my side this looks correct and ready. Nice work.

Thanks for your review!

willem-bd (commented):
## Code review — regenerate latest answer

Overall the design is sound: the `prepare` endpoint resolves a base checkpoint before the last human turn and the runs API forks from it via `configurable.checkpoint_id`. The end-to-end checkpoint path checks out (`useStream` → `client.runs.stream` → `body.checkpoint` → `apply_checkpoint_to_run_config`), and the per-turn run model means superseding the old run doesn't drop earlier turns. `_clean_human_message_for_regenerate` also correctly operates on a copy of `additional_kwargs`, so it doesn't mutate the live checkpoint message.

The main risks are in the **failure paths** — a regenerate that doesn't reach a clean success can hide the original answer. Detailed scenarios are in the inline comments; findings are ranked most-severe first.

1. **Failed/errored regenerate run still hides the original answer** — `getSupersededRunIds` keys off `regenerate_from_run_id` without checking the new run's status. (`hooks.ts:161`)
2. **State leak when `thread.submit` fails after prepare** — the pending superseded sets aren't cleared in the catch, so the old answer is hidden with no replacement until a thread switch. (`hooks.ts:1196`)
3. **First-turn regenerate likely 409s** — `_find_base_checkpoint_before_human` needs an addressable checkpoint predating the human message, which usually doesn't exist for the first turn. (`thread_runs.py:289`)
4. **200-entry scan cap silently 409s on long threads** — both regenerate scans are capped; no pagination. (`thread_runs.py:276`)
5. **`_find_target_run_id` fallback can attribute the wrong run** — returns newest success run without verifying it produced the target message. (`thread_runs.py:265`)
6. **Docs** — the new `POST /{thread_id}/runs/regenerate/prepare` endpoint and the new checkpoint run-config capability aren't in `backend/CLAUDE.md`'s router table, contrary to its "Always update CLAUDE.md after every code change" rule.

#1 and #2 are the ones I'd want addressed before merge — they turn any regenerate that doesn't cleanly succeed into lost/hidden answers. Happy to help iterate on fixes.

frontend/src/core/threads/hooks.ts
```diff
@@ -121,6 +132,55 @@ function dedupeMessagesByIdentity(messages: Message[]): Message[] {
...
+
+export function getSupersededRunIds(
+  runs: Run[] | undefined,
+  pendingSupersededRunIds?: ReadonlySet<string>,
+) {
+  const ids = new Set(pendingSupersededRunIds ?? []);
+  for (const run of runs ?? []) {
+    const metadata = run.metadata;
+    if (metadata && typeof metadata === "object") {
+      const fromRunId = Reflect.get(metadata, "regenerate_from_run_id");
```
willem-bd, line 164: This supersede is status-agnostic: a failed, errored, or interrupted regenerate run still carries `regenerate_from_run_id`, so `run-old` gets filtered out of `buildVisibleHistoryMessages` and replaced by the failed run's partial/empty output. This survives reload since the filter is metadata-driven.

**Scenario:** user clicks Regenerate → new run is created with `metadata.regenerate_from_run_id = run-old` but then errors (model failure, `GraphRecursionError`, tool error, interrupt) → `getSupersededRunIds` adds `run-old` unconditionally (the loop never inspects `run.status`) → original answer hidden.

Suggest only superseding when the newer run reached a success/terminal state, e.g. skip error/interrupted runs, or only honor `regenerate_from_run_id` on the newest *successful* run.

frontend/src/core/threads/hooks.ts
```diff
@@ -1011,6 +1105,104 @@ export function useThreadStream({
...
+        void queryClient.invalidateQueries({ queryKey: ["threads", "search"] });
+        void queryClient.invalidateQueries({
+          queryKey: INFINITE_THREADS_QUERY_KEY_PREFIX,
+        });
+        void queryClient.invalidateQueries({
+          queryKey: threadTokenUsageQueryKey(threadId),
+        });
+      } catch (error) {
+        setLiveMessagesThreadId(null);
+        if (!submitted) {
```
willem-bd: On a submit failure after a successful prepare, `pendingSupersededRunIds`/`pendingSupersededMessageIds` were already set (so the old AI message is hidden from `persistedMessages`) but this catch never clears them. Result: the user sees a gap (question, no answer) that persists until they switch threads.

**Scenario:** prepare succeeds → both pending sets populated, `submitted = true` → `thread.submit` throws (409 multitask reject from another client, 500, or stream error) → catch only resets `liveMessagesThreadId` and shows no toast (because `submitted` is true) → `pendingSupersededMessageIds` never cleared → original AI message stays filtered.

`sendMessage`'s catch clears its optimistic state for comparison; suggest clearing both pending sets here (and/or on stream `onError`) when `submitted` is true.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -131,11 +145,224 @@ def _record_to_response(record: RunRecord) -> RunResponse:
...
+
+    previous_checkpoint = None
+    for checkpoint_tuple in reversed(checkpoints):
+        messages = _checkpoint_messages(checkpoint_tuple)
+        message_ids = {_message_id(message) for message in messages}
+        if human_message_id in message_ids:
+            if previous_checkpoint is None:
+                raise HTTPException(
+                    status_code=409,
+                    detail="Could not find an addressable checkpoint before the target user message",
```
willem-bd, line 337: This requires an addressable checkpoint (one with a `checkpoint_id`) that predates the human message. For the first Q&A turn the checkpoints are typically `C1=[human-1]` (initial) and `C2=[human-1, ai-1]` — iterating reversed, `C1` already contains `human-1` while `previous_checkpoint` is still `None` → 409. Regenerating the very first answer in a conversation would fail unless there's an empty seed checkpoint with a `checkpoint_id` before the first run.

If a pre-human checkpoint genuinely doesn't exist, consider forking with no `checkpoint_id` (fresh run from the cleaned input) instead of erroring — that keeps the feature working on the first turn.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -131,11 +145,224 @@ def _record_to_response(record: RunRecord) -> RunResponse:
...
+    if fallback_record is not None:
+        return fallback_record.run_id
+    raise HTTPException(status_code=409, detail="Could not find source run for assistant message")
+
+
+async def _find_base_checkpoint_before_human(thread_id: str, human_message_id: str, request: Request) -> Any:
+    checkpointer = get_checkpointer(request)
+    base_config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
+    try:
+        checkpoints = [item async for item in checkpointer.alist(base_config, limit=REGENERATE_HISTORY_SCAN_LIMIT)]
```
willem-bd, line 324: Both `_find_base_checkpoint_before_human` (here) and `_find_target_run_id` (`list_messages`) are capped at `REGENERATE_HISTORY_SCAN_LIMIT` (200). Threads with heavy tool use / summarization can exceed 200 checkpoints since the target turn, so regenerate silently 409s ("Could not locate target user message…") instead of paginating backward.

Worth at least raising the limit or paginating, and logging when the cap — not a missing message — is the reason for the 409.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -131,11 +145,224 @@ def _record_to_response(record: RunRecord) -> RunResponse:
...
+        if row.get("event_type") not in {"ai_message", "llm.ai.response"}:
+            continue
+        if _event_message_id(row) == message_id:
+            run_id = row.get("run_id")
+            if isinstance(run_id, str) and run_id:
+                return run_id
+    run_mgr = get_run_manager(request)
+    user_id = await get_current_user(request)
+    records = await run_mgr.list_by_thread(thread_id, user_id=user_id, limit=10)
+    fallback_record = next((record for record in records if record.status == RunStatus.success), None)
```
willem-bd: When run_events aren't persisted (local/dev), this returns the newest `RunStatus.success` record (or `records[0]`) as `target_run_id` without confirming it produced the target assistant message. That id becomes `metadata.regenerate_from_run_id`, so the frontend hides whatever run this points at — which may not be the run that generated the message being regenerated.

Huixin615:
@willem-bd 
Thanks for the detailed failure-path review. I addressed the merge-blocking cases and tightened the fallback behavior in the latest push.

Changes:
- Persistent supersede filtering now only honors `regenerate_from_run_id` from successful regenerate runs, so error/interrupted runs no longer hide the original answer after reload.
- The regenerate submit path now rolls back the pending superseded run/message ids if `thread.submit` fails after prepare, and stream `onError` clears pending supersede state as well.
- The backend run-store fallback now verifies `RunRecord.last_ai_message` matches the target assistant message before returning a source run id. If it cannot verify ownership, it returns 409 instead of attributing to the newest success run.
- Added warning logs for the checkpoint/event scan cap path.
- Updated `backend/CLAUDE.md` for the regenerate prepare endpoint and checkpoint/checkpoint_id run config support.

Validation:
- `uv run pytest tests/test_thread_regenerate_prepare.py tests/test_gateway_services.py tests/test_thread_run_messages_pagination.py`
- `uv run ruff check app/gateway/routers/thread_runs.py tests/test_thread_regenerate_prepare.py`
- `pnpm exec tsc --noEmit`
- `pnpm exec eslint . --ext .ts,.tsx`
- `pnpm exec vitest run tests/unit/core/threads/message-merge.test.ts`
- `git diff --check`

Could you please take another look when you have time? Thanks a lot!

Huixin615:
Sorry for forget to run the format, I have fixed it.

WillemJiang:
> Sorry for forget to run the format, I have fixed it.

@Huixin615 You can set up the pre-commit hook by running make install from the repository root to avoid this kind of situation.

BTW, please drop me an email willem.jiang AT Gmail, to make a connection.

Huixin615:
> > Sorry for forget to run the format, I have fixed it.
> 
> @Huixin615 You can set up the pre-commit hook by running make install from the repository root to avoid this kind of situation.
> 
> BTW, please drop me an email willem.jiang AT Gmail, to make a connection.

Thanks a lot for the suggestion! I've already sent you an email and look forward to connecting with you!

WillemJiang (approved):

PR #3639, opened 2026-06-18
fix(frontend): resolve stale subagent running state after stop

### LINKED ISSUE #3638: [bug] 停止包含子 Agent 的会话后，子任务卡片仍显示运行中，并可能在后续对话中重新变为运行中
### Before you start

### Problem summary

当主会话触发 task 工具创建子 Agent 后，如果用户在子 Agent 尚未返回结果前点击停止会话，前端会保留一个没有对应 ToolMessage 的 task tool call。当前 UI 会把这个子任务继续渲染为运行中。刷新页面后仍然如此；如果继续发送下一条消息，旧的子任务卡片还可能因为新一轮 streaming 再次显示为运行中。

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

当我开启一轮新对话的时候，使用ultra模式。
比如输入：
请开启两个子agent，为我干两件事：1.xxx 2.xxx
此时子agent启动，然后我将对话取消掉，此时子任务卡片依旧显示：“子任务运行中”
实际预期应该显示：“子任务失败” 或者 “子任务中止”

### Expected behavior

停止会话后，未完成且没有 ToolMessage 的子任务不应继续显示为运行中。
刷新页面后，该子任务也不应恢复为运行中。
后续发送新消息时，历史轮次里已停止的子任务不应因为新的全局 loading 状态重新变成运行中。

### Steps to reproduce

1. 启动deer-flow前后端
2. 打开工作区页面
3. 打开ultra模式
4. 发送一个可以触发子agent的任务
5. 等待出现子任务卡片，然后取消会话
6. 观察卡片状态
7. 刷新页面，再次观察卡片状态

### Relevant logs

```shell
无
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

问题定位：
前端会根据消息历史里的 task tool call 创建子任务卡片，初始状态是 in_progress。只有当后续出现对应的 ToolMessage 时，子任务状态才会被更新为成功或失败。
用户停止会话时，task 工具可能被取消，导致历史中只留下 task tool call，但没有对应的 ToolMessage。于是前端重新渲染或刷新页面时，会把这个子任务再次恢复为运行中。
另外，当前判断还依赖全局 thread.isLoading。当同一会话里发送下一条消息时，新的 streaming 会让整个 thread 进入 loading 状态，历史里的已停止子任务也可能被误判为仍在运行。

最小修复路径：
前端根据当前轮次的对话是否loading以及是否有subtask对应的toolResult去匹配，如果匹配到了，状态可以是：运行中，否则是：失败。相当于前端直接根据后端接口的内容，可以直接处理。

本地已经解决并且验证。

AnoobFeng:
Subtask cards were recreated from task tool calls as in_progress whenever the thread was loading. If a user stopped a run before the task tool returned a ToolMessage, the card could remain running after reload, and could flip back to running when a later user turn started streaming.

Derive pending subtask state from the current assistant turn instead of the global thread loading flag. A task now stays in progress only while its own turn is loading or while a matching ToolMessage exists for result parsing; otherwise it is marked failed.

Add unit coverage for task ToolMessage matching and for the later-turn loading regression.

Fixes #3638

## Why

Subtask cards are built from `task` tool calls in the message history. Before this change, each discovered `task` call was recreated as `in_progress` unless a matching ToolMessage later updated it.

When a user stops a run before the `task` tool returns, the history can contain the task tool call without a matching ToolMessage. That left the subtask card stuck in "Running subtask" after stop or reload.

There was also a second edge case: if a later user turn started streaming, the global `thread.isLoading` flag could make an earlier stopped subtask appear to be running again.

## What changed

- Added helpers to detect whether a subtask tool call has a matching ToolMessage and derive its pending UI status.
- Scoped subtask loading state to the current assistant message group instead of the whole thread.
- Mark unfinished subtasks as `failed` when their own assistant turn is no longer loading and no ToolMessage exists.
- Added unit tests for ToolMessage matching and the later-turn loading regression.

## Bug fix verification

Reproduced from code path:

1. A `task` tool call creates a subtask card.
2. User stops the run before a ToolMessage is produced.
3. The message history contains the task tool call without a matching ToolMessage.
4. The old logic recreated the subtask as `in_progress`.
5. The new logic marks it `failed` once that assistant turn is no longer loading.
6. Starting a later user turn no longer revives the earlier stopped subtask.

## Validation

- `pnpm vitest run tests/unit/core/tasks/subtask-result.test.ts`
- `pnpm lint`
- `pnpm format`
- `pnpm format:write`
- `pnpm check`

## AI assistance
use: claude code

### DIFF
--- a/frontend/src/components/workspace/messages/message-list.tsx
+++ b/frontend/src/components/workspace/messages/message-list.tsx
@@ -30,7 +30,7 @@ import { useRehypeSplitWordsIntoSpans } from "@/core/rehype";
 import type { Subtask } from "@/core/tasks";
 import { useUpdateSubtask } from "@/core/tasks/context";
 import {
-  getPendingSubtaskStatus,
+  derivePendingSubtaskStatus,
   parseSubtaskResult,
 } from "@/core/tasks/subtask-result";
 import type { AgentThreadState } from "@/core/threads";
@@ -183,6 +183,7 @@ export function MessageList({
   const updateSubtask = useUpdateSubtask();
   const messages = thread.messages;
   const groupedMessages = getMessageGroups(messages);
+  const lastGroupIndex = groupedMessages.length - 1;
   const turnUsageMessagesByGroupIndex =
     getAssistantTurnUsageMessages(groupedMessages);
   const tokenDebugSteps = useMemo(
@@ -279,10 +280,7 @@ export function MessageList({
         {groupedMessages.map((group, groupIndex) => {
           const turnUsageMessages = turnUsageMessagesByGroupIndex[groupIndex];
           const groupIsLoading =
-            thread.isLoading &&
-            !groupedMessages
-              .slice(groupIndex + 1)
-              .some((nextGroup) => nextGroup.type === "human");
+            thread.isLoading && groupIndex === lastGroupIndex;
 
           if (group.type === "human" || group.type === "assistant") {
             return (
@@ -367,13 +365,17 @@ export function MessageList({
               if (message.type === "ai") {
                 for (const toolCall of message.tool_calls ?? []) {
                   if (toolCall.name === "task") {
-                    const status = getPendingSubtaskStatus(
-                      toolCall.id,
+                    const taskId = toolCall.id;
+                    if (!taskId) {
+                      continue;
+                    }
+                    const status = derivePendingSubtaskStatus(
+                      taskId,
                       group.messages,
                       groupIsLoading,
                     );
                     const task: Subtask = {
-                      id: toolCall.id!,
+                      id: taskId,
                       subagent_type: toolCall.args.subagent_type,
                       description: toolCall.args.description,
                       prompt: toolCall.args.prompt,
@@ -430,14 +432,14 @@ export function MessageList({
               } else if (message.id) {
                 subagentDebugMessageIds.push(message.id);
               }
-              const taskIds = message.tool_calls
-                ?.filter((toolCall) => toolCall.name === "task")
-                .map((toolCall) => toolCall.id);
+              const taskIds = message.tool_calls?.flatMap((toolCall) =>
+                toolCall.name === "task" && toolCall.id ? [toolCall.id] : [],
+              );
               for (const taskId of taskIds ?? []) {
                 results.push(
                   <SubtaskCard
                     key={"task-group-" + taskId}
-                    taskId={taskId!}
+                    taskId={taskId}
                     isLoading={groupIsLoading}
                   />,
                 );
--- a/frontend/src/components/workspace/messages/message-list.tsx
+++ b/frontend/src/components/workspace/messages/message-list.tsx
@@ -29,7 +29,10 @@ import {
 import { useRehypeSplitWordsIntoSpans } from "@/core/rehype";
 import type { Subtask } from "@/core/tasks";
 import { useUpdateSubtask } from "@/core/tasks/context";
-import { parseSubtaskResult } from "@/core/tasks/subtask-result";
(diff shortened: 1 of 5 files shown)


WillemJiang:
@AnoobFeng, please check that the Git commit author email is set in your GitHub profile's email settings to address the CLA assistant's complaint.

AnoobFeng:
@WillemJiang Thanks! I’ve updated the commit author email to my GitHub-provided noreply address, and the CLA check is now passing.

willem-bd:
Nice fix — correct root cause and good helper extraction. A few suggested improvements:

- **`updateSubtask` re-render dependency.** `useUpdateSubtask` only calls `setTasks` when `task.latestMessage` is set, so the running→failed flip relies on `MessageList` re-rendering for an unrelated reason. Consider calling `setTasks` on terminal status transitions (`failed`/`completed`), or at least add a comment about the implicit parent-re-render dependency.

- **Tighten the "current turn" heuristic.** `groupedMessages.slice(groupIndex + 1).some(g => g.type === "human")` treats every group in the latest turn as loading. If a turn has a subagent group followed by another assistant group, an earlier subagent missing its `ToolMessage` will still be shown as `in_progress`. Rare, but tightenable.

- **Rename `getPendingSubtaskStatus`.** It only ever returns `in_progress` or `failed` — `derivePendingSubtaskStatus` or `subtaskStatusBeforeToolResult` would read more accurately.

- **Guard the `toolCall.id!` non-null assertion.** The helper now defensively handles `undefined` ids, but the `id: toolCall.id!` in the `Subtask` literal would still crash. Worth a matching guard at the call site.

- **Add an integration test.** The unit tests cover the helpers; an RTL/Playwright test for the full "stop → reload → card shows failed" flow would harden against regressions in `MessageList`/`SubtaskCard`.

- **Minor perf nit.** `slice(groupIndex + 1).some(...)` runs per group (O(N²)). Precompute the last human group index once: `groupIsLoading = thread.isLoading && groupIndex >= lastHumanIndex`.

AnoobFeng:
@willem-bd 
Thanks for the thorough review — all 6 points addressed:

- **(1) `useUpdateSubtask`**: terminal transitions now also notify React via a ref set during render + an effect that calls `setTasks` after render, avoiding setState-in-render. A guard keeps existing terminal status stable so the next render's pending tool-call update can't demote it back to `in_progress` and trigger a refresh loop.
- **(2 + 6)**: collapsed into one change — precomputed `lastGroupIndex` before the map and replaced `slice(...).some(...)` with `groupIndex === lastGroupIndex`. Chose `===` over `>= lastHumanIndex` because once another group has appeared after a subagent group, that subagent has logically finished and shouldn't be pulled back to `in_progress` by the heuristic.
- **(3)**: renamed `getPendingSubtaskStatus` → `derivePendingSubtaskStatus` to reflect that it only returns `in_progress` / `failed`; updated imports and unit tests.
- **(4)**: narrowed `toolCall.id` at both `MessageList` call sites — task calls without an id are skipped; all `id!` non-null assertions are gone.
- **(5)**: added `tests/e2e/subtask-card.spec.ts` — mocks a historical thread whose task tool call has no matching ToolMessage, reloads, and asserts `Subtask failed` is shown instead of `Running subtask`. Exercises the full `MessageList → useUpdateSubtask → SubtaskCard` path.

WillemJiang (approved):

PR #3641, opened 2026-06-18
fix(smoke-test): add auth-aware frontend checks with login support

### LINKED ISSUE #3632: [bug] smok-test does not support the user auth case
### Before you start

### Problem summary

When running the smoke-test skill, the fronend test doesn't login the system to do some basic test

### Affected area(s)

Skills

### What happened?

The frontend route check complains about getting 401 when it tries to acess the /workspace

### Expected behavior

the frontend route can login with pre setup the account

### Steps to reproduce

/smoke-test docker

### Relevant logs

```shell
- Phase 5 frontend route check: ⚠️ 1 flags
  - /workspace → redirects to /login (auth is enabled — same reason the LangGraph API returns the acceptable 401).
```

Local (make dev)

### Operating system

macOS

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

ech0hol:
Fixes #3632

## Why

`frontend_check.sh` performs anonymous `curl -L` requests against all `/workspace/*` routes. When `DEER_FLOW_AUTH_DISABLED != 1` (auth enabled — the default), protected workspace routes redirect anonymous requests to `/login`, causing two symptoms:

1. **False alarm** — `check_final_url "Workspace redirect"` expects the effective URL to match `/workspace/chats/` but gets `/login` (the flag reported in #3632).
2. **Silent false pass** — `/workspace/chats/new`, `/workspace/chats`, and `/workspace/agents` are only checked with `check_status`. Since `curl -L` follows the 302 → `/login` → 200 chain, these routes show as "passed" while only verifying the login page rendered.

When auth is on, the frontend check can't verify the actual `/workspace/*` pages. This is asymmetric with `health_check.sh`, which already treats `200|401` as acceptable for the LangGraph API (`health_check.sh:107`); `frontend_check.sh` never received that treatment.

## What changed

`frontend_check.sh` is now auth-aware:

- **New `authenticate()` function** — detects whether auth is enabled by probing a protected endpoint. When enabled, it registers or logs in a smoke-test user and passes the session cookie to all subsequent `curl` calls so the real pages are verified rather than the login form.
- **New `check_final_url` assertions** — added effective URL checks for `/workspace/chats/new`, `/workspace/chats`, and `/workspace/agents` so the final URL stays under `/workspace/`, closing the silent-pass blind spot.
- **`SKILL.md` and `SOP.md`** — updated to document the automatic auth handling and the configurable `SMOKE_TEST_EMAIL` / `SMOKE_TEST_PASSWORD` env vars.

Auth flow: `setup-status` → `register` (201 with auto-login) → `login` → `/me` verification. `register` is attempted before `login` to avoid counting a failed login attempt toward rate-limiting when the user already exists.

## Surface area

## Bug fix verification

- **Test path**: Run `bash .agent/skills/smoke-test/scripts/frontend_check.sh` with auth enabled (`DEER_FLOW_AUTH_DISABLED != 1` or unset).
- **Before**: `check_final_url "Workspace redirect"` fails (effective URL is `/login`), and workspace routes silently pass against the login page.
- **After**: Script registers/logs in a smoke-test user and verifies real `/workspace/*` pages.

```
==========================================
  Frontend Page Smoke Check
==========================================

🔐 Auth is enabled — setting up smoke test session...
✓ Logged in as smoke-test@deerflow.dev

1. Checking entry pages...
✓ Landing page (http://localhost:2026/) -> 200
✓ Workspace redirect (http://localhost:2026/workspace) -> 200
✓ Workspace redirect redirect target -> http://localhost:2026/workspace/chats/new

2. Checking key workspace routes...
✓ New chat page (http://localhost:2026/workspace/chats/new) -> 200
✓ Workspace redirect (http://localhost:2026/workspace) -> 200
✓ Workspace redirect redirect target -> http://localhost:2026/workspace/chats/new

2. Checking key workspace routes...
✓ New chat page (http://localhost:2026/workspace/chats/new) -> 200

2. Checking key workspace routes...
✓ New chat page (http://localhost:2026/workspace/chats/new) -> 200
2. Checking key workspace routes...
✓ New chat page (http://localhost:2026/workspace/chats/new) -> 200
✓ New chat page redirect target -> http://localhost:2026/workspace/chats/new
✓ New chat page (http://localhost:2026/workspace/chats/new) -> 200
✓ New chat page redirect target -> http://localhost:2026/workspace/chats/new
✓ Chats list page (http://localhost:2026/workspace/chats) -> 200
✓ New chat page redirect target -> http://localhost:2026/workspace/chats/new
✓ Chats list page (http://localhost:2026/workspace/chats) -> 200
✓ Chats list page redirect target -> http://localhost:2026/workspace/chats
✓ Chats list page (http://localhost:2026/workspace/chats) -> 200
✓ Chats list page redirect target -> http://localhost:2026/workspace/chats
✓ Chats list page redirect target -> http://localhost:2026/workspace/chats
✓ Agents gallery page (http://localhost:2026/workspace/agents) -> 200

3. Checking docs route (optional)...
✓ Docs page (http://localhost:2026/en/docs) -> 200

==========================================
  Frontend Smoke Check Summary
==========================================


✅ Frontend smoke checks passed!
```

## Validation

- [x] Ran `bash .agent/skills/smoke-test/scripts/frontend_check.sh` with auth enabled — all routes pass, effective URLs stay under `/workspace/`.

### DIFF
--- a/.agent/skills/smoke-test/SKILL.md
+++ b/.agent/skills/smoke-test/SKILL.md
@@ -97,14 +97,14 @@ smoke-test/
 2. **Check frontend service** - Visit `http://localhost:2026` and verify that the page loads
 3. **Check API Gateway** - Verify the `http://localhost:2026/health` endpoint
 4. **Check LangGraph-compatible API** - Verify the `/api/langgraph/*` route exposed by Gateway
-5. **Frontend route smoke check** - Run `bash .agent/skills/smoke-test/scripts/frontend_check.sh` to verify key routes under `/workspace`
+5. **Frontend route smoke check** - Run `bash .agent/skills/smoke-test/scripts/frontend_check.sh` to verify key routes under `/workspace`. The script auto-detects whether authentication is enabled and, if so, registers / logs in a smoke-test user so the real pages are verified rather than the login redirect.
 
 **Docker mode health check** (when using Docker):
 1. **Check container status** - Run `docker ps` and confirm that all containers are running
 2. **Check frontend service** - Visit `http://localhost:2026` and verify that the page loads
 3. **Check API Gateway** - Verify the `http://localhost:2026/health` endpoint
 4. **Check LangGraph-compatible API** - Verify the `/api/langgraph/*` route exposed by Gateway
-5. **Frontend route smoke check** - Run `bash .agent/skills/smoke-test/scripts/frontend_check.sh` to verify key routes under `/workspace`
+5. **Frontend route smoke check** - Run `bash .agent/skills/smoke-test/scripts/frontend_check.sh` to verify key routes under `/workspace`. The script auto-detects whether authentication is enabled and, if so, registers / logs in a smoke-test user so the real pages are verified rather than the login redirect.
 
 ### Optional Functional Verification
 
@@ -136,6 +136,7 @@ smoke-test/
 The following warnings can appear during smoke testing and do not block a successful result:
 - Feishu/Lark SSL errors in Gateway logs (certificate verification failure) can be ignored if that channel is not enabled
 - Warnings in Gateway logs about missing methods in the custom checkpointer, such as `adelete_for_runs` or `aprune`, do not affect the core functionality
+- The `frontend_check.sh` script automatically handles authentication. When auth is enabled it registers / logs in a smoke-test user (`smoke-test@deerflow.dev` by default) to verify the real `/workspace/*` pages. The registration may produce a log entry from the auth provider, which is expected and harmless.
 
 ## Key Tools
 
--- a/.agent/skills/smoke-test/references/SOP.md
+++ b/.agent/skills/smoke-test/references/SOP.md
@@ -361,6 +361,22 @@ curl http://localhost:2026/health
 
 ---
 
+#### 5.1.5 Frontend Route Smoke Check
+
+**Objective**: Verify key `/workspace/*` frontend routes render correctly.
+
+**Steps**:
+1. Run `bash .agent/skills/smoke-test/scripts/frontend_check.sh`.
+2. The script auto-detects whether authentication (`DEER_FLOW_AUTH_DISABLED`) is enabled.
+3. When auth is on, the script registers / logs in a smoke-test user (`smoke-test@deerflow.dev` by default) and passes the session cookie so the real `/workspace/*` pages are verified — not the login redirect.
+4. When auth is off, the routes are checked anonymously as before.
+
+**Customisation**:
+- `SMOKE_TEST_EMAIL` — email for the test account (default: `smoke-test@deerflow.dev`).
+- `SMOKE_TEST_PASSWORD` — password for the test account (default: `SmokeTest123!`).
+
+---
+
 ### 5.2 Docker Mode Health Check (When Using Docker)
 
 #### 5.2.1 Check Container Status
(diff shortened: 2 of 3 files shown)


willem-bd:
## `initialize` branch ignores its HTTP status and trusts the cookie unconditionally

In `authenticate()`, the `needs_setup=true` branch discards the response code and prints success even if `initialize` failed:

```bash
curl -s -X POST "${BASE_URL}/api/v1/auth/initialize" \
    -H "Content-Type: application/json" \
    -d "{\"email\":\"${SMOKE_TEST_EMAIL}\",\"password\":\"${SMOKE_TEST_PASSWORD}\"}" \
    -c "$COOKIE_JAR" > /dev/null
echo "✓ Admin initialized & logged in"
CURL_AUTH_OPTS="-b $COOKIE_JAR"
return 0
```

The other two auth paths guard their outcome:
- `register` checks `register_code == "201"`
- `login` verifies the session via `/api/v1/auth/me` (200)

But `initialize` does neither. `/initialize` returns `409` if the system is already set up and `400` on a bad body (`auth.py:initialize_admin`), so a setup race (the `needs_setup` flag flipping between the probe and the POST) or any other failure is masked — the script prints "✓ Admin initialized & logged in" and proceeds with a possibly-empty cookie jar. The problem only surfaces later as a confusing `check_final_url` failure ("redirect target -> /login") rather than a clear "initialize failed" message.

**Suggested fix** — capture the status and verify, mirroring the other branches:

```bash
local init_code
init_code=$(curl -s -o /dev/null -w "%{http_code}" \
    -X POST "${BASE_URL}/api/v1/auth/initialize" \
    -H "Content-Type: application/json" \
    -d "{\"email\":\"${SMOKE_TEST_EMAIL}\",\"password\":\"${SMOKE_TEST_PASSWORD}\"}" \
    -c "$COOKIE_JAR" 2>/dev/null)
if [ "$init_code" = "201" ]; then
    echo "✓ Admin initialized & logged in"
    CURL_AUTH_OPTS="-b $COOKIE_JAR"
    return 0
fi
echo "✗ Initialize failed (HTTP $init_code)"
return 1
```

(Bonus consistency: verifying the session via `/api/v1/auth/me` once at the end of `authenticate()` would cover all three branches uniformly — `register(201)` and `initialize` currently don't confirm the cookie is usable, only `login` does.)

ech0hol:
@willem-bd Thanks for the catch on the initialize branch! That was indeed an oversight — ignoring the HTTP status code while the other two branches both guard their outcomes.

Fixed in the latest push. While applying your suggestion I also noticed the same pattern in the login branch (no status check either), so I fixed that too. Here's a summary of the changes:

**Addressing your review:**
- `initialize`: now captures HTTP status and only proceeds on 201, returns 1 with a clear error message on any other code

**Related fixes found along the way:**
- `login`: added status code validation (same blind spot as initialize — only register was checking its response code)
- Added a reachability check before auth detection — `curl` against an unreachable service returns `000`, which would be misinterpreted as "auth disabled"
- Clear cookie jar before login after a failed register attempt, to avoid stale cookies from the register response
- Fallback for when `/setup-status` is unreachable or returns an unexpected format (skip initialize, try register/login directly)
- Login now uses `--data-urlencode` instead of raw `-d` to handle special characters in credentials correctly

All three auth branches (initialize / register / login) now consistently verify their outcome, and the unified `/auth/me` check at the end of `authenticate()` covers all of them uniformly as you suggested.

WillemJiang (approved):

PR #3644, opened 2026-06-18
Avoid blank previews for unsupported artifact files

vantanco:
## Summary
- Avoid rendering unsupported binary/document artifacts in an iframe.
- Show a clear fallback with file metadata and a download action for files the browser cannot preview.
- Keep inline previews for browser-supported media such as PDFs, images, audio, and video.

## Testing
- `pnpm exec eslint src/components/workspace/artifacts/artifact-file-detail.tsx src/core/utils/files.tsx`
- `pnpm exec prettier --check src/components/workspace/artifacts/artifact-file-detail.tsx src/core/utils/files.tsx`
- `pnpm exec tsc --noEmit`

### DIFF
--- a/frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
+++ b/frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
@@ -20,6 +20,7 @@ import {
   ArtifactTitle,
 } from "@/components/ai-elements/artifact";
 import { ClipboardSafeStreamdown } from "@/components/ai-elements/streamdown";
+import { Button } from "@/components/ui/button";
 import { Select, SelectItem } from "@/components/ui/select";
 import {
   SelectContent,
@@ -43,7 +44,13 @@ import { useI18n } from "@/core/i18n/hooks";
 import { findToolCallResult } from "@/core/messages/utils";
 import { installSkill } from "@/core/skills/api";
 import { streamdownPlugins } from "@/core/streamdown";
-import { checkCodeFile, getFileName } from "@/core/utils/files";
+import {
+  canBrowserPreviewFile,
+  checkCodeFile,
+  getFileExtensionDisplayName,
+  getFileIcon,
+  getFileName,
+} from "@/core/utils/files";
 import { env } from "@/env";
 import { cn } from "@/lib/utils";
 
@@ -92,6 +99,9 @@ export function ArtifactFileDetail({
     }
     return checkCodeFile(filepath);
   }, [filepath, isWriteFile, isSkillFile]);
+  const canPreviewInBrowser = useMemo(() => {
+    return canBrowserPreviewFile(filepath);
+  }, [filepath]);
   const isSupportPreview = useMemo(() => {
     return language === "html" || language === "markdown";
   }, [language]);
@@ -303,14 +313,66 @@ export function ArtifactFileDetail({
             readonly
           />
         )}
-        {!isCodeFile && (
+        {!isCodeFile && canPreviewInBrowser && (
           <iframe
             className="size-full"
             src={urlOfArtifact({ filepath, threadId, isMock })}
           />
         )}
+        {!isCodeFile && !canPreviewInBrowser && (
+          <ArtifactDownloadFallback
+            filepath={filepath}
+            threadId={threadId}
+            isMock={isMock}
+          />
+        )}
       </ArtifactContent>
     </Artifact>
+  );
+}
+
+function ArtifactDownloadFallback({
+  filepath,
+  threadId,
+  isMock,
+}: {
+  filepath: string;
+  threadId: string;
+  isMock?: boolean;
+}) {
+  const filename = getFileName(filepath);
+  const fileType = getFileExtensionDisplayName(filepath);
+
+  return (
+    <div className="flex size-full items-center justify-center p-6">
+      <div className="flex max-w-sm flex-col items-center gap-4 text-center">
+        <div className="text-muted-foreground">
+          {getFileIcon(filepath, "size-12")}
+        </div>
+        <div className="space-y-1">
+          <div className="font-medium break-all">{filename}</div>
+          <div className="text-muted-foreground text-sm">{fileType} file</div>
+        </div>
+        <p className="text-muted-foreground text-sm">
+          This file type cannot be previewed in the browser.
+        </p>
+        <Button asChild>
+          <a
+            href={urlOfArtifact({
+              filepath,
+              threadId,
+              download: true,
+              isMock,
+            })}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            <DownloadIcon className="size-4" />
+            Download
+          </a>
+        </Button>
+      </div>
+    </div>
   );
 }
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
```diff
@@ -303,17 +313,69 @@ export function ArtifactFileDetail({
...
+      <div className="flex max-w-sm flex-col items-center gap-4 text-center">
+        <div className="text-muted-foreground">
+          {getFileIcon(filepath, "size-12")}
+        </div>
+        <div className="space-y-1">
+          <div className="font-medium break-all">{filename}</div>
+          <div className="text-muted-foreground text-sm">{fileType} file</div>
+        </div>
+        <p className="text-muted-foreground text-sm">
+          This file type cannot be previewed in the browser.
```
willem-bd, line 357: **i18n regression:** this message (and the literal `Download` a few lines below) are hardcoded English, whereas every other label in this file goes through `useI18n()` / `t.*` — e.g. `t.common.download` on the header Download action, which has a zh-CN translation `下载`. Under a non-English locale this whole fallback card stays untranslated.

Fix: pull `t` from `useI18n()` in the parent and pass the localized strings down, adding the new keys to `core/i18n/locales/{en-US,zh-CN}.ts`.

frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
```diff
@@ -303,17 +313,69 @@ export function ArtifactFileDetail({
...
+  filepath,
+  threadId,
+  isMock,
+}: {
+  filepath: string;
+  threadId: string;
+  isMock?: boolean;
+}) {
+  const filename = getFileName(filepath);
+  const fileType = getFileExtensionDisplayName(filepath);
```
willem-bd, line 344: **Extensionless / malformed filenames:** `getFileExtensionDisplayName` falls back to `extension.toUpperCase()`, so:
- `LICENSE` (no extension) → returns `LICENSE` → title `LICENSE` + subtitle `LICENSE file` (name shown twice).
- `notes.` (trailing dot) → returns `` (empty) → subtitle ` file` (blank type).

Both are reachable for any extensionless artifact routed here. Consider guarding for the no-extension case (e.g. return `Unknown`).

frontend/src/core/utils/files.tsx
```diff
@@ -143,6 +143,29 @@ const extensionMap: Record<string, string> = {
   v: "v",
 };
 
+const browserPreviewExtensions = new Set([
```
willem-bd, line 146: **Third parallel media classifier:** this set overlaps with `extensionMap` and the image/audio/video switches in `getFileIcon` (further down this file), and they've already drifted in this PR — `getFileIcon` treats `svg`/`heic`/`tiff` as images and `wma`/`aiff`/`ape` as audio (none present here), while this set adds `apng`/`avif`/`pdf`/`webm` (absent from `getFileIcon`).

Concrete effect: a `.heic` or `.tiff` artifact renders an `ImageIcon` yet hits the "cannot be previewed" fallback. A single `getFileType()` returning `image|audio|video|pdf|document` — used by both the icon and the preview decision — would keep the two consistent.

frontend/src/core/utils/files.tsx
```diff
@@ -170,6 +193,10 @@ export function checkCodeFile(
   };
 }
 
+export function canBrowserPreviewFile(filepath: string) {
+  return browserPreviewExtensions.has(getFileExtension(filepath));
+}
+
 export function getFileExtensionDisplayName(filepath: string) {
   const fileName = getFileName(filepath);
   const extension = fileName.split(".").pop()!.toLocaleLowerCase();
```
willem-bd, line 202: **Duplicated extension extraction:** this line re-derives the extension via `getFileName(filepath).split(".").pop().toLocaleLowerCase()` instead of calling the existing `getFileExtension(filepath)`. Two extraction paths in one module — if extraction logic changes (case normalization, no-extension handling) they can silently diverge. Could be `const extension = getFileExtension(filepath);`.

frontend/src/core/utils/files.tsx
```diff
@@ -170,6 +193,10 @@ export function checkCodeFile(
   };
 }
 
+export function canBrowserPreviewFile(filepath: string) {
```
willem-bd, line 196: **Altitude:** the client now guesses browser capability from a hardcoded extension list, but the backend (`artifacts.py`) already computes the MIME via `mimetypes.guess_type()` and controls inline-vs-attachment `Content-Disposition` per request (and already forces download for active content: HTML/XHTML/SVG). The client guess and the server disposition are decided independently and can disagree. If feasible, driving preview-vs-download from the server's actual content-type/disposition (a HEAD, or a metadata field) would be more robust than maintaining this whitelist.

frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
```diff
@@ -92,6 +99,9 @@ export function ArtifactFileDetail({
     }
     return checkCodeFile(filepath);
   }, [filepath, isWriteFile, isSkillFile]);
+  const canPreviewInBrowser = useMemo(() => {
```
willem-bd, line 102: **Needless memo:** `canBrowserPreviewFile` is a `Set.has` + one `split` — cheaper than the `useMemo` bookkeeping. A plain `const canPreviewInBrowser = canBrowserPreviewFile(filepath);` suffices and removes the deps-array lint surface.

frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
```diff
@@ -303,17 +313,69 @@ export function ArtifactFileDetail({
             readonly
           />
         )}
-        {!isCodeFile && (
+        {!isCodeFile && canPreviewInBrowser && (
```
willem-bd, line 316: **Minor:** the two branches below repeat the `!isCodeFile` guard. They collapse to `{!isCodeFile && (canPreviewInBrowser ? <iframe …/> : <ArtifactDownloadFallback …/>)}` — one guard to keep in sync instead of two.

willem-bd (commented):
High-effort multi-angle review. The change is sound — and it actually *fixes* the SVG/HTML blank-preview case: the backend (`artifacts.py`) already forces those active-content types to download, so the old `<iframe>` was the bug, not a feature. Items to consider, most important first:

**Correctness**
1. `ArtifactDownloadFallback` hardcodes English copy while the rest of the file routes through `useI18n()` / `t.*` — stays untranslated under zh-CN.

**Minor / cleanup**
2. Extensionless filenames render the name twice (`LICENSE` / `LICENSE file`); trailing-dot names show an empty type (` file`).
3. `browserPreviewExtensions` is a third parallel media classifier and has already drifted from `getFileIcon` (e.g. `.heic`/`.tiff` show an image icon yet "cannot be previewed").
4. `getFileExtensionDisplayName` re-derives the extension instead of calling `getFileExtension`.
5. The client whitelist duplicates browser-capability knowledge the server already computes (MIME / disposition) — they can disagree.
6. `canPreviewInBrowser` memoizes a `Set.has` (needless).
7. The two `!isCodeFile && …` branches can collapse to one ternary.

Details inline. Happy to help with the fixes.

WillemJiang (approved):

PR #3646, opened 2026-06-19
fix(frontend): improve mobile workspace layout

### LINKED ISSUE #3642: [feat] web没做移动端适配
### Before you start

### Problem / motivation

web没做移动端适配，欢迎文字超出排名，左侧也没有导航栏了

Image

### Proposed solution

加入移动端审批

### Affected area(s)

Frontend (UI / Next.js)

### Alternatives considered

_No response_

### Additional context

_No response_

Huixin615:
Fixes #3642

## Why

The workspace welcome page was not properly adapted for mobile screens. Welcome text could overflow horizontally, input controls became crowded, and users had no visible way to open the workspace navigation.

## What changed

- Allow welcome descriptions and headings to wrap within narrow viewports.
- Add a mobile sidebar trigger to workspace and chat headers.
- Adjust welcome input positioning, spacing, model controls, and suggestions for small screens.
- Compact agent chat header actions on mobile.
- Add Playwright coverage for viewport overflow and mobile sidebar navigation.

## Surface area

- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**

## Screenshots / Recording

Before: see #3642.

After:
- Mobile welcome layout with wrapped text and responsive controls.
- Mobile workspace navigation drawer.

image
image

## Bug fix verification

- Test path: `frontend/tests/e2e/sidebar.spec.ts`
- Red on `main`, green on this branch: yes
- On `main@6044e5c5`, the mobile test fails because no visible
  `[data-sidebar='trigger']` exists. The same spec passes 3/3 on this branch.

## Validation

- `pnpm format:write`
- `pnpm format`
- `pnpm check`
- `pnpm test` — 34 files, 315 tests passed
- `sidebar.spec.ts` — 3 Playwright tests passed
- Manual verification at 390px, 768px, and 1440px viewport widths
- `git diff --check`

### DIFF
--- a/frontend/src/components/workspace/welcome.tsx
+++ b/frontend/src/components/workspace/welcome.tsx
@@ -10,6 +10,14 @@ import { AuroraText } from "../ui/aurora-text";
 
 let waved = false;
 
+function WelcomeDescription({ children }: { children: string }) {
+  return (
+    <p className="max-w-full text-wrap break-words whitespace-pre-line">
+      {children}
+    </p>
+  );
+}
+
 export function Welcome({
   className,
   mode,
@@ -32,15 +40,15 @@ export function Welcome({
   return (
     <div
       className={cn(
-        "mx-auto flex w-full flex-col items-center justify-center gap-2 px-8 py-4 text-center",
+        "mx-auto flex w-full max-w-full flex-col items-center justify-center gap-2 px-4 py-4 text-center sm:px-8",
         className,
       )}
     >
-      <div className="text-2xl font-bold">
+      <div className="max-w-full text-2xl font-bold">
         {searchParams.get("mode") === "skill" ? (
           `✨ ${t.welcome.createYourOwnSkill} ✨`
         ) : (
-          <div className="flex items-center gap-2">
+          <div className="flex max-w-full flex-wrap items-center justify-center gap-2">
             <div className={cn("inline-block", !waved ? "animate-wave" : "")}>
               {isUltra ? "🚀" : "👋"}
             </div>
@@ -49,24 +57,14 @@ export function Welcome({
         )}
       </div>
       {searchParams.get("mode") === "skill" ? (
-        <div className="text-muted-foreground text-sm">
-          {t.welcome.createYourOwnSkillDescription.includes("\n") ? (
-            <pre className="font-sans whitespace-pre">
-              {t.welcome.createYourOwnSkillDescription}
-            </pre>
-          ) : (
-            <p>{t.welcome.createYourOwnSkillDescription}</p>
-          )}
+        <div className="text-muted-foreground max-w-full text-sm">
+          <WelcomeDescription>
+            {t.welcome.createYourOwnSkillDescription}
+          </WelcomeDescription>
         </div>
       ) : (
-        <div className="text-muted-foreground text-sm">
-          {t.welcome.description.includes("\n") ? (
-            <pre className="font-sans whitespace-pre">
-              {t.welcome.description}
-            </pre>
-          ) : (
-            <p>{t.welcome.description}</p>
-          )}
+        <div className="text-muted-foreground max-w-full text-sm">
+          <WelcomeDescription>{t.welcome.description}</WelcomeDescription>
         </div>
       )}
     </div>
--- a/frontend/src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx
+++ b/frontend/src/app/workspace/agents/[agent_name]/chats/[thread_id]/page.tsx
@@ -6,6 +6,7 @@ import { useCallback, useEffect, useState } from "react";
 
 import type { PromptInputMessage } from "@/components/ai-elements/prompt-input";
 import { Button } from "@/components/ui/button";
+import { SidebarTrigger } from "@/components/ui/sidebar";
 import { AgentWelcome } from "@/components/workspace/agent-welcome";
 import { ArtifactTrigger } from "@/components/workspace/artifacts";
 import { ChatBox, useThreadChat } from "@/components/workspace/chats";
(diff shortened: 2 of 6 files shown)


frontend/src/components/workspace/welcome.tsx
```diff
@@ -10,6 +10,14 @@ import { AuroraText } from "../ui/aurora-text";
 
 let waved = false;
 
+function WelcomeDescription({ children }: { children: string }) {
+  return (
+    <p className="max-w-full text-wrap break-words whitespace-pre-line">
```
fancyboi999, line 15: Non-blocking, just flagging this is a deliberate behavior change: the old code rendered descriptions containing `\n` in `<pre className="whitespace-pre">` (no wrapping, all whitespace preserved), and this switches every description to `whitespace-pre-line`. That's exactly what fixes the overflow (it now wraps), but it also collapses runs of consecutive spaces — newlines are still preserved, but any intentional multi-space alignment inside a translation string would change. Almost certainly fine for the welcome copy; just worth a glance that no `t.welcome.description` / `createYourOwnSkillDescription` locale relies on aligned spacing.

fancyboi999 (commented):
Reviewed the full diff and traced the mobile-nav mechanism end-to-end on `main` @ `6044e5c5` — clean fix for #3642, and I checked the pieces actually line up rather than just looking plausible:

- **Breakpoints leave no dead zone.** The new `SidebarTrigger className="md:hidden"` is visible below 768px, and the drawer switches to a `Sheet` at the same threshold — `useIsMobile()` is `window.innerWidth < 768` (`hooks/use-mobile.ts:13`), matching Tailwind's `md`. So "trigger visible" and "sidebar is a Sheet" always agree; there's no width where the trigger shows but nothing opens (or vice-versa).
- **The trigger can't throw.** All three headers that gained `SidebarTrigger` render under `SidebarProvider` (`app/workspace/layout.tsx` → `WorkspaceContent`, `workspace-content.tsx:32`, which also mounts `<WorkspaceSidebar />`), so `useSidebar()` resolves. On mobile that sidebar is the `Sheet` with `data-mobile="true" data-sidebar="sidebar"` (`components/ui/sidebar.tsx:183-204`) — exactly what the new test targets.
- **The new e2e test is a real regression anchor.** `data-slot="suggestions-list"` exists (`ai-elements/suggestion.tsx:23`) and `mockLangGraphAPI` is already imported. On `main` none of these headers render a `SidebarTrigger`, so at 390px `[data-sidebar='trigger']:visible` matches nothing → the sidebar half of the test fails without this change. It anchors.

One thing to close out (Bug-fix verification): you marked *"Red on `main`: no — not executed against `main`."* From inspection it would be red (no trigger in those headers on main), but since the template asks for it, worth actually confirming — e.g. `git stash` the `src/` changes, keep the spec, run `pnpm exec playwright test sidebar.spec.ts` to watch it fail, then unstash. Cheap, and it turns "should be red" into "is red".

Everything else checks out — scope is all mobile-layout, it reuses the existing `SidebarTrigger`/`sm:`/`md:` conventions, and the welcome overflow fix (`whitespace-pre-line` + `break-words` + `flex-wrap`) is the right call. Looks correct to me; recommend merge once red-on-main is confirmed. One non-blocking note inline.

Huixin615:
> Reviewed the full diff and traced the mobile-nav mechanism end-to-end on `main` @ `6044e5c5` — clean fix for #3642, and I checked the pieces actually line up rather than just looking plausible:
> 
> * **Breakpoints leave no dead zone.** The new `SidebarTrigger className="md:hidden"` is visible below 768px, and the drawer switches to a `Sheet` at the same threshold — `useIsMobile()` is `window.innerWidth < 768` (`hooks/use-mobile.ts:13`), matching Tailwind's `md`. So "trigger visible" and "sidebar is a Sheet" always agree; there's no width where the trigger shows but nothing opens (or vice-versa).
> * **The trigger can't throw.** All three headers that gained `SidebarTrigger` render under `SidebarProvider` (`app/workspace/layout.tsx` → `WorkspaceContent`, `workspace-content.tsx:32`, which also mounts `<WorkspaceSidebar />`), so `useSidebar()` resolves. On mobile that sidebar is the `Sheet` with `data-mobile="true" data-sidebar="sidebar"` (`components/ui/sidebar.tsx:183-204`) — exactly what the new test targets.
> * **The new e2e test is a real regression anchor.** `data-slot="suggestions-list"` exists (`ai-elements/suggestion.tsx:23`) and `mockLangGraphAPI` is already imported. On `main` none of these headers render a `SidebarTrigger`, so at 390px `[data-sidebar='trigger']:visible` matches nothing → the sidebar half of the test fails without this change. It anchors.
> 
> One thing to close out (Bug-fix verification): you marked _"Red on `main`: no — not executed against `main`."_ From inspection it would be red (no trigger in those headers on main), but since the template asks for it, worth actually confirming — e.g. `git stash` the `src/` changes, keep the spec, run `pnpm exec playwright test sidebar.spec.ts` to watch it fail, then unstash. Cheap, and it turns "should be red" into "is red".
> 
> Everything else checks out — scope is all mobile-layout, it reuses the existing `SidebarTrigger`/`sm:`/`md:` conventions, and the welcome overflow fix (`whitespace-pre-line` + `break-words` + `flex-wrap`) is the right call. Looks correct to me; recommend merge once red-on-main is confirmed. One non-blocking note inline.

Confirmed. I checked both `en-US` and `zh-CN` welcome descriptions. None relies on consecutive spaces for alignment; they only use normal prose and explicit newlines. `whitespace-pre-line` therefore preserves the intended line breaks without changing meaningful formatting.

WillemJiang (approved):

PR #3648, opened 2026-06-19
perf(sandbox): cache LocalSandbox path-rewrite regexes per instance (#3647)

### LINKED ISSUE #3647: perf(sandbox): LocalSandbox recompiles path-rewrite regexes on every tool call
### Affected area(s)

Backend (sandbox / tools)

### Problem

`LocalSandbox` rebuilds and re-`re.compile()`s its path-rewrite regex on **every** tool call, even though the inputs (`self.path_mappings`) are fixed for the sandbox's entire lifetime.

`self.path_mappings` is assigned once in `__init__` (`local_sandbox.py:81`) and is never mutated anywhere in the codebase. Yet these three hot-path helpers re-sort the mappings, `re.escape` each one, join, and `re.compile` from scratch on every invocation:

| Method | Lines | Runs on |
| --- | --- | --- |
| `_resolve_paths_in_command` | 220–250 | every `bash` (`execute_command`, line 313) |
| `_resolve_paths_in_content` | 252–283 | every `write_file` (line 419) |
| `_reverse_resolve_paths_in_output` | 185–218 | every `bash` output + every `read_file` (lines 356 / 380) |

Each call does `sorted(self.path_mappings, ...)` → per-mapping `re.escape(...)` → `"|".join(...)` → `re.compile(...)`. `re` is also imported inside each function instead of at module level.

A single agent turn issues many `bash`/`read_file`/`write_file` calls, so this regex compilation is repeated work on the agent's hottest path.

### Proposed fix

Cache the compiled patterns (and the sorted mappings) on the instance with `functools.cached_property`, since they're derived from immutable `path_mappings`. Behavior is unchanged; the patterns are simply compiled once per sandbox instead of once per tool call. Also hoist `import re` to module scope.

This is a small, self-contained, behavior-preserving change. I'll open a PR.

Eilen6316:
### Summary

Fixes #3647. `LocalSandbox` re-sorted `path_mappings` and re-`re.compile()`d its path-rewrite regex on **every** tool call, even though those patterns are derived solely from `self.path_mappings`, which is set once in `__init__` and never mutated.

Affected hot-path methods:
- `_resolve_paths_in_command` — every `bash` (`execute_command`)
- `_resolve_paths_in_content` — every `write_file`
- `_reverse_resolve_paths_in_output` — every `bash` output + every `read_file`

### What changes

- Compile the patterns once per sandbox via `functools.cached_property` (`_command_pattern`, `_content_pattern`, `_reverse_output_patterns`) and reuse them across calls.
- Hoist `import re` from inside three methods to module scope.
- **Behavior is identical** — the regex sources, sort order, and segment-boundary lookaheads are unchanged; only the per-call sort + `re.escape` + `re.compile` is removed. Safe because `path_mappings` is immutable for the sandbox's lifetime (assigned once at `__init__`, never reassigned anywhere in the tree).

Net: a typical agent turn fires many `bash`/`read_file`/`write_file` calls; this removes the repeated regex compilation from each of them.

### Testing

- New `tests/test_local_sandbox_path_regex_cache.py`: cached-pattern identity is stable across accesses; command/content paths still resolve to local; `/mnt/skills` does not match inside `/mnt/skills-extra` (segment boundary); reverse-resolve maps local paths back to container paths; empty mappings pass through unchanged.
- `ruff check` + `ruff format --check` clean on the changed files.

> Note: in my sandbox a few unrelated suite files can't be collected due to missing optional deps (`agent_sandbox`, `langchain`); the same failures reproduce on `main` without this change. The path-resolution behavior is covered by the new test and the unchanged `test_local_sandbox_virtual_path_contract.py` cases that don't require those deps.

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
+++ b/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
@@ -113,8 +113,23 @@ class LocalSandbox(Sandbox):
     @cached_property
     def _reverse_output_patterns(self) -> list[re.Pattern[str]]:
         """Compiled matchers for local paths in command output (longest local path first)."""
-        sorted_mappings = sorted(self.path_mappings, key=lambda m: len(m.local_path), reverse=True)
-        return [re.compile(re.escape(str(Path(m.local_path).resolve())) + r"(?:[/\\][^\s\"';&|<>()]*)?") for m in sorted_mappings]
+        return [re.compile(re.escape(self._resolved_local_paths[m]) + r"(?:[/\\][^\s\"';&|<>()]*)?") for m in self._mappings_by_local_specificity]
+
+    @cached_property
+    def _resolved_local_paths(self) -> dict[PathMapping, str]:
+        """Filesystem-resolved local root per mapping. ``Path.resolve()`` hits the
+        disk, and the mounted directories don't move, so resolve once and reuse."""
+        return {m: str(Path(m.local_path).resolve()) for m in self.path_mappings}
+
+    @cached_property
+    def _mappings_by_container_specificity(self) -> list[PathMapping]:
+        """Mappings ordered most-specific-container-first (for forward resolution)."""
+        return sorted(self.path_mappings, key=lambda m: len(m.container_path.rstrip("/") or "/"), reverse=True)
+
+    @cached_property
+    def _mappings_by_local_specificity(self) -> list[PathMapping]:
+        """Mappings ordered longest-local-path-first (for reverse resolution)."""
+        return sorted(self.path_mappings, key=lambda m: len(m.local_path), reverse=True)
 
     def _is_read_only_path(self, resolved_path: str) -> bool:
         """Check if a resolved path is under a read-only mount.
@@ -129,7 +144,7 @@ class LocalSandbox(Sandbox):
         best_prefix_len = -1
 
         for mapping in self.path_mappings:
-            local_resolved = str(Path(mapping.local_path).resolve())
+            local_resolved = self._resolved_local_paths[mapping]
             if resolved == local_resolved or resolved.startswith(local_resolved + os.sep):
                 prefix_len = len(local_resolved)
                 if prefix_len > best_prefix_len:
@@ -144,7 +159,7 @@ class LocalSandbox(Sandbox):
     def _find_path_mapping(self, path: str) -> tuple[PathMapping, str] | None:
         path_str = str(path)
 
-        for mapping in sorted(self.path_mappings, key=lambda m: len(m.container_path.rstrip("/") or "/"), reverse=True):
+        for mapping in self._mappings_by_container_specificity:
             container_path = mapping.container_path.rstrip("/") or "/"
             if container_path == "/":
                 if path_str.startswith("/"):
@@ -174,7 +189,7 @@ class LocalSandbox(Sandbox):
             return ResolvedPath(path_str, None)
 
         mapping, relative = mapping_match
-        local_root = Path(mapping.local_path).resolve()
+        local_root = Path(self._resolved_local_paths[mapping])
         resolved_path = (local_root / relative).resolve() if relative else local_root
 
         try:
@@ -204,8 +219,8 @@ class LocalSandbox(Sandbox):
         path_str = str(Path(normalized_path).resolve())
 
         # Try each mapping (longest local path first for more specific matches)
-        for mapping in sorted(self.path_mappings, key=lambda m: len(m.local_path), reverse=True):
-            local_path_resolved = str(Path(mapping.local_path).resolve())
+        for mapping in self._mappings_by_local_specificity:
+            local_path_resolved = self._resolved_local_paths[mapping]
             if path_str == local_path_resolved or path_str.startswith(local_path_resolved + "/"):
                 # Replace the local path prefix with container path
                 relative = path_str[len(local_path_resolved) :].lstrip("/")
--- a/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
+++ b/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
@@ -2,9 +2,11 @@ import errno
 import logging
 import ntpath
 import os
+import re
 import shutil
 import subprocess
 from dataclasses import dataclass
+from functools import cached_property
 from pathlib import Path
 from typing import NamedTuple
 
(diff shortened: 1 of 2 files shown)


willem-bd (commented):
## Code review summary

Focused, well-scoped perf fix. The premise — `path_mappings` is set once in `__init__` and never mutated — checks out: I grepped the tree and there are zero assignments or list mutations of `self.path_mappings` outside `__init__`, no subclasses, and the provider always builds a fresh `LocalSandbox` instead of mutating an existing one. The regex sources, sort order, and segment-boundary lookaheads are byte-identical to the originals; behavior matches. New tests pass plus the 66 existing sandbox/provider tests.

No correctness bugs found. Two non-blocking observations below — both about *altitude* of the fix rather than regressions:

1. The same immutability premise generalizes to the per-call sorts in `_find_path_mapping`, `_reverse_resolve_path`, and `_is_read_only_path`, which still fire on the same hot path the regex caching is shaving.
2. `replace_match` in `_reverse_resolve_paths_in_output` is now defined inside the per-pattern loop without capturing anything loop-local — easy to hoist while the function is already being touched.

Neither is a blocker; happy to land as-is and address in a follow-up if you'd rather keep this PR tight.

backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
```diff
@@ -83,6 +85,37 @@ def __init__(self, id: str, path_mappings: list[PathMapping] | None = None):
...
+        if not mappings:
+            return None
+        patterns = [re.escape(m.container_path) + r"(?=/|$|[^\w./-])(?:/[^\s\"';&|<>()]*)?" for m in mappings]
+        return re.compile("|".join(f"({p})" for p in patterns))
+
+    @cached_property
+    def _reverse_output_patterns(self) -> list[re.Pattern[str]]:
+        """Compiled matchers for local paths in command output (longest local path first)."""
+        sorted_mappings = sorted(self.path_mappings, key=lambda m: len(m.local_path), reverse=True)
+        return [re.compile(re.escape(str(Path(m.local_path).resolve())) + r"(?:[/\\][^\s\"';&|<>()]*)?") for m in sorted_mappings]
```
willem-bd: **Altitude — adjacent per-call sorts left in place.** The PR's premise (`path_mappings` is immutable, so derived structures can be cached) generalizes cleanly to three sites this PR doesn't touch:

- `_find_path_mapping` (line 147) — sorted on every `_resolve_path` call, which fires from the regex callbacks this PR introduces, so a long bash command with N container-path matches still does N sorts.
- `_reverse_resolve_path` (line 207) — sorted on every match in `_reverse_resolve_paths_in_output`, so the per-output sort the PR removed is partially re-introduced inside `replace_match`.
- `_is_read_only_path` (line 131) — iterates `self.path_mappings` per `write_file`.

A cleaner shape would be two cached views at the same level as the regexes — e.g. `_mappings_sorted_by_container_desc` and `_mappings_sorted_by_local_desc` — used by all four sites. That lands the optimization at the right depth instead of as a regex-only special case. Not a blocker for this PR, but worth a follow-up so the next reader doesn't see "some sorts cached, some not" and have to reason about why.

backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
```diff
@@ -192,22 +225,10 @@ def _reverse_resolve_paths_in_output(self, output: str) -> str:
...
+        # correct prefix matching) and reused across calls.
         result = output
-        for mapping in sorted_mappings:
-            # Escape the local path for use in regex
-            escaped_local = re.escape(str(Path(mapping.local_path).resolve()))
-            # Match the local path followed by optional path components with either separator
-            pattern = re.compile(escaped_local + r"(?:[/\\][^\s\"';&|<>()]*)?")
+        for pattern in self._reverse_output_patterns:
 
             def replace_match(match: re.Match) -> str:
```
willem-bd, line 248: **Minor cleanup — `replace_match` redefined inside the loop.** This closure doesn't capture `pattern` or any other loop-local, so the per-iteration `def` is pure leftover from the pre-PR shape (when the regex was also rebuilt each iteration). Since you've already restructured this method (removed the in-loop `re.compile` and the early `if not sorted_mappings`), hoisting `replace_match` above the loop — or making it a tiny `self._reverse_resolve_match` method — finishes the cleanup and avoids constructing a fresh function object per pattern on the same hot path.

backend/tests/test_local_sandbox_path_regex_cache.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_empty_mappings_yield_no_pattern(tmp_path):
+    sb = LocalSandbox(id="empty", path_mappings=[])
+    assert sb._command_pattern is None
+    assert sb._content_pattern is None
+    assert sb._reverse_output_patterns == []
+    # No mappings -> command/content pass through unchanged.
+    assert sb._resolve_paths_in_command("echo hello") == "echo hello"
+    assert sb._resolve_paths_in_content("plain text") == "plain text"
```
fancyboi999, line 44: Suggestion (non-blocking): this is the only `_resolve_paths_in_content` call in the suite, and it's the empty-mappings passthrough — so nothing pins the content-specific boundary `(?=/|$|[^\w./-])` that distinguishes `_content_pattern` from `_command_pattern`. A future refactor that accidentally gave content the shell boundary would still pass every test here.

A positive case on the fixture sandbox would lock it in: pick a container path followed by a char that's a *content* boundary but **not** a shell one — e.g. a comma. `_resolve_paths_in_content("see /mnt/skills,then")` should rewrite the `/mnt/skills` prefix (since `,` satisfies `[^\w./-]`), whereas `_resolve_paths_in_command` on the same string would leave it untouched (`,` isn't in `[\s\"';&|<>()]`). That asserts content really uses its own lookahead, not the command one.

fancyboi999 (commented):
Following up from the #3647 triage — I re-verified the refactor against `main`, focusing on the spots where a "behavior-preserving" regex move can quietly not be:

- **The two forward patterns keep their distinct boundaries byte-for-byte.** `_command_pattern` keeps the shell-aware `(?=/|$|[\s\"';&|<>()])`, `_content_pattern` keeps the text `(?=/|$|[^\w./-])` (and the forward-slash normalization stays in `_resolve_paths_in_content`'s `replace_match`). They're separate cached properties, so the two boundaries can't collapse into one — that was the main trap and it's avoided.
- **`_reverse_output_patterns` preserves the sequential semantics.** The original compiled one pattern per mapping and applied them longest-local-path-first via successive `result = pattern.sub(...)`; the cached version is the same per-mapping list in the same sort order, run by the same sequential loop. Reverse-resolve is order-sensitive (each `sub` runs over the previous result), so keeping both the list and the order is exactly what makes it identical — and it does.

Agree with @willem-bd's two notes (the adjacent per-call sorts in `_find_path_mapping` / the read-only check, and the now-redundant per-iteration `replace_match` def). Both non-blocking — the closure doesn't capture the loop variable, so there's no late-binding bug, just leftover shape from when the regex was rebuilt each iteration.

One small test suggestion inline: the suite pins the command boundary and the empty-mappings passthrough, but there's no positive `_resolve_paths_in_content` assertion, so the content-specific boundary isn't anchored.

Correct, minimal, well-scoped. Looks good to merge.

Eilen6316:
Expanded the scope slightly in a follow-up commit: besides the regex compilation, `_find_path_mapping`, `_is_read_only_path`, `_resolve_path_with_mapping`, and `_reverse_resolve_path` also re-sorted `path_mappings` and re-ran `Path(local_path).resolve()` (a filesystem syscall) on every call. Since `path_mappings` is immutable for the sandbox's lifetime, those are now cached per instance too (`_resolved_local_paths`, `_mappings_by_container_specificity`, `_mappings_by_local_specificity`), and the reverse-output pattern builder reuses the same resolved-path cache. Behavior is unchanged; added tests cover forward/reverse resolution and read-only detection.

WillemJiang (approved):

PR #3651, opened 2026-06-19
fix(gateway): attach thread-message feedback by real event_type (#3650)

### LINKED ISSUE #3650: bug(gateway): thread messages feedback never attached (event_type mismatch)
### Affected area(s)

Backend API (gateway / endpoints)

### Problem

`GET /api/threads/{thread_id}/messages` (`thread_runs.list_thread_messages`) attaches user feedback to the last AI message of each run by matching:

```python
if msg.get("event_type") == "ai_message":
    last_ai_per_run[msg["run_id"]] = i
```

But AI messages are persisted by `RunJournal` with `event_type="llm.ai.response"` (`runtime/journal.py:295`), and the event store returns that value verbatim (`DbRunEventStore._row_to_dict` does no remapping). No production code path ever writes `event_type="ai_message"`.

Consequences:
- `last_ai_per_run` is always empty, so **feedback is never attached** — every message in the response gets `feedback: null`. Thumbs up/down state is not restored when the thread view reloads.
- `feedback_repo.list_by_thread_grouped(...)` runs on **every** request to this endpoint and its result is never used — a wasted DB query per call.

There is no test covering feedback attachment on this thread-level endpoint (only the per-run `/runs/{run_id}/messages` endpoint is tested), which is why it went unnoticed.

### Proposed fix

Match the real event type (`"llm.ai.response"`), and only run the grouped-feedback query when there is at least one AI message to attach it to. Add a regression test. PR incoming.

Eilen6316:
### Summary

Fixes #3650. `GET /api/threads/{thread_id}/messages` never attached user feedback, because it looked for `event_type == "ai_message"` while AI messages are stored by `RunJournal` as `"llm.ai.response"` (returned verbatim by the event store). No production code writes `"ai_message"`, so the match never hit:

- feedback was never attached — every message came back with `feedback: null` (thumbs up/down not restored on thread reload);
- `feedback_repo.list_by_thread_grouped(...)` ran on every request and its result was discarded.

### What changes

- Match the real `"llm.ai.response"` event type when locating each run's last AI message.
- Only run the grouped-feedback query when there is at least one AI message to attach it to (threads with no completed AI turn no longer pay for an unused query).

### Testing

- New `tests/test_thread_messages_feedback.py`: feedback is attached to the last AI message per run (and only there); the query is skipped when the thread has no AI message.
- `ruff check` / `ruff format --check` clean; `py_compile` OK.

> Note: the gateway TestClient suite can't execute in my sandbox (no `fastapi`/`dotenv` installed there); the test is written against the existing `_router_auth_helpers.make_authed_test_app` harness used by the sibling `test_thread_run_messages_pagination.py` and will run in CI.

### DIFF
--- a/backend/app/gateway/routers/thread_runs.py
+++ b/backend/app/gateway/routers/thread_runs.py
@@ -349,18 +349,23 @@ async def list_thread_messages(
     event_store = get_run_event_store(request)
     messages = await event_store.list_messages(thread_id, limit=limit, before_seq=before_seq, after_seq=after_seq)
 
-    # Attach feedback to the last AI message of each run
-    feedback_repo = get_feedback_repo(request)
-    user_id = await get_current_user(request)
-    feedback_map = await feedback_repo.list_by_thread_grouped(thread_id, user_id=user_id)
-
-    # Find the last ai_message per run_id
+    # Find the last AI message per run_id. AI messages are persisted by
+    # RunJournal with event_type "llm.ai.response" (see runtime/journal.py);
+    # the event store returns that value verbatim, so match on it here.
     last_ai_per_run: dict[str, int] = {}  # run_id -> index in messages list
     for i, msg in enumerate(messages):
-        if msg.get("event_type") == "ai_message":
+        if msg.get("event_type") == "llm.ai.response":
             last_ai_per_run[msg["run_id"]] = i
 
-    # Attach feedback field
+    # Attach feedback to the last AI message of each run. Only query when there
+    # is an AI message to attach it to — threads with no completed AI turn yet
+    # would otherwise pay for a grouped feedback lookup whose result is unused.
+    feedback_map: dict[str, dict] = {}
+    if last_ai_per_run:
+        feedback_repo = get_feedback_repo(request)
+        user_id = await get_current_user(request)
+        feedback_map = await feedback_repo.list_by_thread_grouped(thread_id, user_id=user_id)
+
     last_ai_indices = set(last_ai_per_run.values())
     for i, msg in enumerate(messages):
         if i in last_ai_indices:
--- a/backend/tests/test_thread_messages_feedback.py
+++ b/backend/tests/test_thread_messages_feedback.py
@@ -1,0 +1,76 @@
+"""Regression for the thread-messages feedback attachment.
+
+GET /api/threads/{thread_id}/messages attaches user feedback to the last AI
+message of each run. AI messages are stored by RunJournal with event_type
+"llm.ai.response"; the endpoint previously matched the non-existent
+"ai_message", so feedback was never attached and the grouped-feedback query
+ran on every request for nothing.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, MagicMock
+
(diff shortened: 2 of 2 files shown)


backend/app/gateway/routers/thread_runs.py
```diff
@@ -349,18 +349,23 @@ async def list_thread_messages(
...
-    feedback_map = await feedback_repo.list_by_thread_grouped(thread_id, user_id=user_id)
-
-    # Find the last ai_message per run_id
+    # Find the last AI message per run_id. AI messages are persisted by
+    # RunJournal with event_type "llm.ai.response" (see runtime/journal.py);
+    # the event store returns that value verbatim, so match on it here.
     last_ai_per_run: dict[str, int] = {}  # run_id -> index in messages list
     for i, msg in enumerate(messages):
-        if msg.get("event_type") == "ai_message":
+        if msg.get("event_type") == "llm.ai.response":
```
fancyboi999, line 357: Non-blocking: `"llm.ai.response"` is now a magic string shared by the writer (`runtime/journal.py:295`) and this reader, with nothing tying them together — which is exactly how `"ai_message"` drifted here in the first place. A single shared constant (exported next to where the journal emits it, imported here) would make a future rename one line and keep this class of bug from recurring. Same drift-hazard shape @willem-bd flagged on #3648.

fancyboi999 (commented):
Confirmed the bug on `main` @ `6044e5c5`, and the fix lines up.

- **The match was dead.** `"ai_message"` appears exactly once in the whole tree — the comparison this PR changes (`thread_runs.py:360` on main). Nothing ever *writes* it. AI messages are persisted as `"llm.ai.response"` (`runtime/journal.py:295`), the event store returns `event_type` verbatim (no remap in the run-event row mapping), and the seed + test suite treat `"llm.ai.response"` as the canonical AI type (`tests/seed_runs_router.py:37`, `test_run_journal.py`). So `last_ai_per_run` was always empty → `feedback` came back `null` on every message, and `list_by_thread_grouped(...)` ran on every call for a result that was then discarded. Real bug, accurate root cause.
- **The fix is right.** Matching `"llm.ai.response"` and keeping the last index per `run_id` lands feedback on each run's final AI message (the loop overwrites, so the highest-seq AI row wins) — the intended "last AI message of the run." Gating the query behind `if last_ai_per_run:` is behavior-preserving; it only skips a query whose result wouldn't have been used anyway.
- **On coverage:** this event_type-based attachment is unique to the thread-level endpoint — `list_run_messages` (`:386`) doesn't attach feedback at all — so there was no working sibling to catch the drift, and no test until now. The new `test_thread_messages_feedback.py` is a genuine regression anchor (feedback only on the run's last AI row, and `assert_not_awaited()` on the no-AI-message path).

One non-blocking suggestion inline: the bug was fundamentally a duplicated magic string, so a shared constant would keep the writer and reader from drifting again.

Honest scope: verified by reading the code + grepping writers/readers; I didn't run the gateway suite (you note it needs fastapi/dotenv absent from your sandbox, and it runs in CI). Looks correct — recommend merge.

WillemJiang (approved):

PR #3652, opened 2026-06-19
Add rvLLM as an OpenAI-compatible inference stack (RvllmChatModel)

m0at:
## What

Adds **rvLLM** as an additional inference stack alongside the existing vLLM provider.

[rvLLM](https://github.com/m0at/rvllm) is a native Rust inference server (`rvllm-server`) that exposes an OpenAI-compatible API and serves greedy models with server-side **MTP speculative decoding** — a small drafter proposes K tokens, the target verifies them in one forward pass, and a greedy accept/reject keeps the emitted tokens identical to plain greedy decoding. Speculation changes throughput only, never the output (an "identity guarantee").

Because `rvllm-server` is greedy by default, `RvllmChatModel` defaults to deterministic decoding (`temperature=0`) so multi-step agent trajectories are reproducible across runs. An explicit `temperature` (including a non-zero value) always wins. The endpoint is otherwise plain OpenAI-compatible, so tool calls, streaming, and usage accounting are inherited from `ChatOpenAI` unchanged.

## Changes

- `deerflow/models/rvllm_provider.py` — `RvllmChatModel`, a thin `ChatOpenAI` subclass that defaults to greedy decoding and reports `_llm_type = "rvllm-openai-compatible"`.
- `tests/test_rvllm_provider.py` — covers the greedy default, explicit-temperature override, request-payload temperature, and `_llm_type`.
- `README.md`, `config.example.yaml`, `.env.example` — document the provider right after the vLLM entry, mirroring its format (`RVLLM_API_KEY`, `base_url`, `use: deerflow.models.rvllm_provider:RvllmChatModel`).

## Usage

```yaml
models:
  - name: gemma-4-e4b-it-rvllm
    display_name: Gemma 4 E4B (rvLLM)
    use: deerflow.models.rvllm_provider:RvllmChatModel
    model: gemma-4-e4b-it
    api_key: $RVLLM_API_KEY
    base_url: http://localhost:18086/v1
    temperature: 0
```

## Tests

`uv run pytest tests/test_rvllm_provider.py` → 4 passed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,7 @@ INFOQUEST_API_KEY=your-infoquest-api-key
 # MINIMAX_API_KEY=your-minimax-api-key  # OpenAI-compatible, see https://platform.minimax.io
 # STEPFUN_API_KEY=your-stepfun-api-key  # OpenAI-compatible, see https://platform.stepfun.com
 # VLLM_API_KEY=your-vllm-api-key  # OpenAI-compatible
+# RVLLM_API_KEY=your-rvllm-api-key  # OpenAI-compatible, native Rust rvllm-server (https://github.com/m0at/rvllm)
 # FEISHU_APP_ID=your-feishu-app-id
 # FEISHU_APP_SECRET=your-feishu-app-secret
 
--- a/README.md
+++ b/README.md
@@ -157,6 +157,14 @@ That prompt is intended for coding agents. It tells the agent to clone the repo 
          extra_body:
            chat_template_kwargs:
              enable_thinking: true
+
+     - name: gemma-4-e4b-it-rvllm
+       display_name: Gemma 4 E4B (rvLLM)
+       use: deerflow.models.rvllm_provider:RvllmChatModel
+       model: gemma-4-e4b-it
+       api_key: $RVLLM_API_KEY
+       base_url: http://localhost:18086/v1
+       temperature: 0
    ```
 
    OpenRouter and similar OpenAI-compatible gateways should be configured with `langchain_openai:ChatOpenAI` plus `base_url`. If you prefer a provider-specific environment variable name, point `api_key` at that variable explicitly (for example `api_key: $OPENROUTER_API_KEY`).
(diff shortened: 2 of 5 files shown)


fancyboi999 (commented):
Reviewed the integration end-to-end on `main` @ `6044e5c5`. Whether to add another inference backend is a maintainer call; on the code itself it's a correct, minimal, convention-consistent wrapper — and, the part worth verifying for any new provider, it's wired in properly:

- **No registration needed / nothing missing.** `use:` is resolved by dynamic import (`models/factory.py:110`, `resolve_class(model_config.use, BaseChatModel)`); there's no provider allowlist, so `use: deerflow.models.rvllm_provider:RvllmChatModel` is sufficient — same as `VllmChatModel`/`ClaudeChatModel`, which aren't registered anywhere either. `RvllmChatModel(ChatOpenAI)` passes the `BaseChatModel` check.
- **Usage accounting really is inherited.** The narrow `_enable_stream_usage_by_default` only fires for the exact `langchain_openai:ChatOpenAI` path (`factory.py:34`), which a custom subclass would miss — but the broader guard at `factory.py:187-194` sets `stream_usage=True` for any `model_class` whose `model_fields` include `stream_usage`, which every `ChatOpenAI` subclass (including this one) does. So token usage is enabled for rvLLM exactly like vLLM; the "inherited unchanged" claim holds.
- **Convention fit.** `arbitrary_types_allowed` + `_llm_type` + `ChatOpenAI` subclass mirrors `vllm_provider.py:159-165` (and the claude/codex providers). The only behavioral delta from plain `ChatOpenAI` is `kwargs.setdefault("temperature", 0.0)`; `_get_request_payload` carries it through, and the tests pin default-0 / explicit-wins / payload-level-0 (the last is the one that actually proves temperature=0 isn't dropped).
- Docs (`README`, `config.example.yaml`, `.env.example`) are consistent.

Honest scope: static review + tracing the factory wiring; I didn't stand up an `rvllm-server`. The code is correct and self-contained — the adoption decision is yours.

WillemJiang:
@m0at, thanks for your contribution. Please click the CLA assistance button to sign the CLA before we begin reviewing your PR.

PR #3654, opened 2026-06-19
perf(persistence): cache Base.to_dict column reflection per class (#3653)

### LINKED ISSUE #3653: perf(persistence): cache Base.to_dict column reflection per class
### Affected area(s)

Backend (persistence)

### Problem

`Base.to_dict()` and `Base.__repr__()` (`persistence/base.py`) call `sqlalchemy.inspect(type(self)).mapper.column_attrs` on **every** invocation:

```python
return {c.key: getattr(self, c.key) for c in sa_inspect(type(self)).mapper.column_attrs if c.key not in exclude}
```

`to_dict()` runs once per row when serializing ORM results — e.g. `DbRunEventStore._row_to_dict` calls it for every event/message row, so a messages or events page serializes N rows = N mapper-reflection passes. The set of mapped columns is fixed at class-definition time, so re-running the reflection per row is pure overhead.

### Proposed fix

Cache the mapped column keys per ORM class (`functools.cache` on a small helper keyed by class) and iterate the cached tuple in `to_dict()`/`__repr__()`. Behavior is identical. PR incoming.

Eilen6316:
### Summary

Fixes #3653. `Base.to_dict()` / `Base.__repr__()` re-ran `sqlalchemy.inspect(type(self)).mapper.column_attrs` on every call. `to_dict()` is called once per row when serializing ORM results (e.g. `DbRunEventStore._row_to_dict` for every event/message row), so a page of N rows did N mapper-reflection passes for a column set that is fixed at class-definition time.

### What changes

- Add a `functools.cache`d `_column_keys(cls)` helper returning the mapped column keys in mapper order.
- `to_dict()` / `__repr__()` iterate the cached tuple. Behavior is identical (same columns, same order, `exclude` honored).

### Testing

- New `tests/test_base_to_dict.py`: `to_dict()` returns all columns, `exclude` works (incl. empty set), keys are cached per class (same tuple object), `__repr__` lists columns.
- `ruff check` / `ruff format --check` clean; `py_compile` OK.

> Note: `sqlalchemy` isn't installed in my sandbox so the test can't execute locally; it's a plain ORM-model unit test and will run in CI.

### DIFF
--- a/backend/packages/harness/deerflow/persistence/base.py
+++ b/backend/packages/harness/deerflow/persistence/base.py
@@ -9,8 +9,21 @@ LangGraph's checkpointer tables are NOT managed by this Base.
 
 from __future__ import annotations
 
+from functools import cache
+
 from sqlalchemy import inspect as sa_inspect
 from sqlalchemy.orm import DeclarativeBase
+
+
+@cache
+def _column_keys(cls: type) -> tuple[str, ...]:
+    """Mapped column keys for an ORM class, in mapper order.
+
+    ``to_dict``/``__repr__`` run per row (e.g. once per event when serializing a
+    messages page), so the SQLAlchemy mapper reflection is cached per class —
+    the mapping is fixed at class-definition time, so this never goes stale.
+    """
+    return tuple(c.key for c in sa_inspect(cls).mapper.column_attrs)
 
 
 class Base(DeclarativeBase):
@@ -24,7 +37,7 @@ class Base(DeclarativeBase):
     def to_dict(self, *, exclude: set[str] | None = None) -> dict:
         """Convert ORM instance to plain dict.
 
-        Uses SQLAlchemy's inspect() to iterate mapped column attributes.
+        Uses cached mapped-column keys (see :func:`_column_keys`).
 
         Args:
             exclude: Optional set of column keys to omit.
@@ -32,9 +45,11 @@ class Base(DeclarativeBase):
         Returns:
             Dict of {column_key: value} for all mapped columns.
         """
-        exclude = exclude or set()
-        return {c.key: getattr(self, c.key) for c in sa_inspect(type(self)).mapper.column_attrs if c.key not in exclude}
+        keys = _column_keys(type(self))
+        if exclude:
+            return {k: getattr(self, k) for k in keys if k not in exclude}
+        return {k: getattr(self, k) for k in keys}
 
     def __repr__(self) -> str:
-        cols = ", ".join(f"{c.key}={getattr(self, c.key)!r}" for c in sa_inspect(type(self)).mapper.column_attrs)
+        cols = ", ".join(f"{k}={getattr(self, k)!r}" for k in _column_keys(type(self)))
         return f"{type(self).__name__}({cols})"
--- a/backend/tests/test_base_to_dict.py
+++ b/backend/tests/test_base_to_dict.py
@@ -1,0 +1,46 @@
+"""Base.to_dict()/__repr__ caching + behavior.
+
+These run once per row when serializing ORM rows (e.g. every event in a
+messages page), so the mapped-column reflection is cached per class. Behavior
+must stay identical.
+"""
+
+from __future__ import annotations
+
+from sqlalchemy import Integer, String
+from sqlalchemy.orm import Mapped, mapped_column
+
+from deerflow.persistence.base import Base, _column_keys
+
+
+class _Widget(Base):
+    __tablename__ = "_widget_to_dict_test"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True)
+    name: Mapped[str] = mapped_column(String(32))
+    color: Mapped[str] = mapped_column(String(16))
+
+
+def test_to_dict_returns_all_columns():
+    w = _Widget(id=1, name="gear", color="red")
+    assert w.to_dict() == {"id": 1, "name": "gear", "color": "red"}
+
+
+def test_to_dict_exclude():
+    w = _Widget(id=2, name="cog", color="blue")
+    assert w.to_dict(exclude={"color"}) == {"id": 2, "name": "cog"}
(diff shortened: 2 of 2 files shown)


fancyboi999 (commented):
Confirmed on `main` @ `6044e5c5` — clean, behavior-preserving cache. Verified:

- **Equivalence.** `_column_keys(cls)` returns `tuple(c.key for c in sa_inspect(cls).mapper.column_attrs)` — the same keys, same mapper order, as the old per-call comprehension; `to_dict`/`__repr__` just iterate the cached tuple. The `exclude=set()` path is preserved: the old code did `exclude or set()` + `not in`, the new `if exclude:` falls through to the no-filter branch for an empty set — same result, and `test_to_dict_exclude` pins exactly that.
- **Cache safety.** Keyed by the ORM class, and the mapper's column set is fixed at class-definition time, so it never goes stale — same immutability rationale as the sandbox-regex cache in #3648. `@cache` is unbounded but the key space is the finite set of ORM model classes, so no growth concern; reflection still happens lazily on first use (after the mapper is configured), just once instead of per row.
- **Hot path.** This is the per-row cost in `DbRunEventStore._row_to_dict`, so an N-row messages/events page now does one reflection instead of N — matches the issue.

Tests cover all-columns, exclude (incl. the empty-set case), per-class cache identity (`is`), and repr. Looks correct — recommend merge.

WillemJiang (approved):

willem-bd (commented):
Code review pass (recall-biased, automated). This is a clean, behavior-preserving optimization — verified that `to_dict()`/`__repr__()` produce identical keys, values, and insertion order for every realistic `exclude` input, and that `__repr__` does route through the cached helper. No active bugs; nothing blocks merge. Three minor / defensive findings below — all theoretical today (no reachable trigger in the current codebase).

1. **`@cache` staleness** (base.py:24) — the docstring's "never goes stale" is now a load-bearing invariant the cache can't enforce.
2. **`@cache` strong-ref retention** (base.py:18) — unbounded cache pins ORM classes alive for process lifetime.
3. **Test couples to private `_column_keys`** (test_base_to_dict.py:38) — importing a `_`-prefixed symbol + asserting `is` locks the caching implementation.

Optional zero-risk hardening: `functools.cache` → `functools.lru_cache(maxsize=128)`, and assert `to_dict()` output stability instead of `_column_keys` identity. None of this blocks merge.

backend/packages/harness/deerflow/persistence/base.py
```diff
@@ -9,10 +9,23 @@
...
 from sqlalchemy.orm import DeclarativeBase
 
 
+@cache
+def _column_keys(cls: type) -> tuple[str, ...]:
+    """Mapped column keys for an ORM class, in mapper order.
+
+    ``to_dict``/``__repr__`` run per row (e.g. once per event when serializing a
+    messages page), so the SQLAlchemy mapper reflection is cached per class —
+    the mapping is fixed at class-definition time, so this never goes stale.
```
willem-bd, line 24: **`@cache` staleness (defensive — not reachable today).** The docstring's "never goes stale" is now a load-bearing invariant the cache can't enforce. If a model's mapped columns were ever extended after the first `to_dict()`/`__repr__()` call (e.g. `mapper.add_property()` / `column_property` / polymorphic), the cached tuple would silently drop the new columns for the process lifetime with no invalidation path — the old per-call `sa_inspect(...)` always reflected the current mapper.

Not reachable in the current codebase: all 8 models declare columns statically, and SQLAlchemy raises `InvalidRequestError` on `add_property` against an already-configured mapper. Flagging because `deerflow-harness` is extensible. To close the gap: a one-line note documenting the assumption, or `@cache` keyed on `(cls, id(cls.__mapper__))`.

backend/packages/harness/deerflow/persistence/base.py
```diff
@@ -9,10 +9,23 @@
 
 from __future__ import annotations
 
+from functools import cache
+
 from sqlalchemy import inspect as sa_inspect
 from sqlalchemy.orm import DeclarativeBase
 
 
+@cache
```
willem-bd, line 18: **Unbounded `@cache` holds a strong ref to `cls` for process lifetime (defensive).** Every ORM class that ever hits `to_dict()`/`__repr__()` is pinned alive along with its mapper + Table + MetaData. All production models are module-level singletons, so nothing leaks today — but since `deerflow-harness` is a publishable/extensible framework, any future plugin or test fixture that creates an ephemeral `Base` subclass (e.g. `class _Tmp(Base)` inside a function) would leak.

A `WeakKeyDictionary` keyed on `cls`, or `functools.lru_cache(maxsize=128)`, gives identical behavior without the retention.

backend/tests/test_base_to_dict.py
```diff
@@ -0,0 +1,46 @@
...
+def test_to_dict_exclude():
+    w = _Widget(id=2, name="cog", color="blue")
+    assert w.to_dict(exclude={"color"}) == {"id": 2, "name": "cog"}
+    # Empty exclude behaves like no exclude.
+    assert w.to_dict(exclude=set()) == {"id": 2, "name": "cog", "color": "blue"}
+
+
+def test_column_keys_are_cached_per_class():
+    # Same tuple object returned across calls -> reflection ran once.
+    assert _column_keys(_Widget) is _column_keys(_Widget)
```
willem-bd, line 38: **Couples the test to the caching implementation.** Importing the `_`-prefixed `_column_keys` and asserting identity (`is`) locks the caching mechanism: a later refactor to a `Base` classmethod, `@cached_property`, or a different cache backend (e.g. to address the staleness/retention notes) breaks this test even though `to_dict()`/`__repr__()` behavior is unchanged.

Consider asserting only the public contract — e.g. that `to_dict()` output is identical across calls, or that the mapper reflection isn't re-run (mock on `sa_inspect`) — and dropping the import of the private symbol.

PR #3657, opened 2026-06-19
perf(sandbox): speed up should_ignore_name in glob/grep walks (#3655)

### LINKED ISSUE #3655: perf(sandbox): speed up should_ignore_name in glob/grep walks
### Affected area(s)

Backend (sandbox / tools — glob/grep search)

### Problem

`should_ignore_name` (`sandbox/search.py`) is called once per directory entry during the recursive `glob`/`grep` tree walks (`find_glob_matches`, `find_grep_matches` filter both dirs and files through it). Each call loops over the ~57-entry `IGNORE_PATTERNS` list invoking `fnmatch.fnmatch` per pattern:

```python
def should_ignore_name(name: str) -> bool:
    for pattern in IGNORE_PATTERNS:
        if fnmatch.fnmatch(name, pattern):
            return True
    return False
```

On a large tree that's ~50 `fnmatch` calls × every file and directory. The vast majority of `IGNORE_PATTERNS` are literal names (`.git`, `node_modules`, `__pycache__`, …); only ~10 are globs (`*.egg-info`, `*.swp`, `*.log`, `*~`, …).

### Proposed fix

Precompute (module load) the literal names into a `frozenset` (O(1) membership) and the glob patterns into a single combined compiled regex via `fnmatch.translate`. `should_ignore_name` then does one `os.path.normcase` + a set lookup + (at most) one regex match. `os.path.normcase` preserves `fnmatch`'s case behavior (case-sensitive on POSIX, folded on Windows), so behavior is identical. PR includes an equivalence test against the original per-pattern loop.

Eilen6316:
### Summary

Fixes #3655. `should_ignore_name` runs once per directory entry during `glob`/`grep` tree walks and looped over the ~57-entry `IGNORE_PATTERNS` calling `fnmatch.fnmatch` per pattern — ~50 fnmatch calls × every file and directory on a large tree.

### What changes

- Precompute at module load: literal ignore names into a `frozenset` (O(1) membership), and the ~10 glob patterns into a single combined compiled regex via `fnmatch.translate`.
- `should_ignore_name` now does one `os.path.normcase` + a set lookup + at most one regex match.
- `os.path.normcase` preserves `fnmatch`'s case behavior (case-sensitive on POSIX, folded on Windows), so results are identical.

### Testing

- New `tests/test_should_ignore_name.py` includes an **equivalence oracle** asserting the new implementation matches the original per-pattern `fnmatch` loop across a broad sample (exact ignores, glob ignores, and look-alikes that must be kept like `node_modules_x`, `x.git`, `myenv`), plus `should_ignore_path` segment checks. All pass locally.
- `ruff check` / `ruff format --check` clean.

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/search.py
+++ b/backend/packages/harness/deerflow/sandbox/search.py
@@ -67,11 +67,22 @@ class GrepMatch:
     line: str
 
 
+# ``should_ignore_name`` runs once per directory entry during glob/grep tree
+# walks, so we avoid ~50 ``fnmatch`` calls per name. Most ignore patterns are
+# literal names (O(1) set lookup after normcase); the few glob patterns are
+# pre-translated into a single combined regex. ``os.path.normcase`` keeps the
+# same case behavior ``fnmatch`` applies (case-sensitive on POSIX, folded on
+# Windows).
+_EXACT_IGNORE_NAMES = frozenset(os.path.normcase(p) for p in IGNORE_PATTERNS if not any(c in p for c in "*?["))
+_GLOB_IGNORE_PATTERNS = [p for p in IGNORE_PATTERNS if any(c in p for c in "*?[")]
+_GLOB_IGNORE_RE = re.compile("|".join(fnmatch.translate(os.path.normcase(p)) for p in _GLOB_IGNORE_PATTERNS)) if _GLOB_IGNORE_PATTERNS else None
+
+
 def should_ignore_name(name: str) -> bool:
-    for pattern in IGNORE_PATTERNS:
-        if fnmatch.fnmatch(name, pattern):
-            return True
-    return False
+    normalized = os.path.normcase(name)
+    if normalized in _EXACT_IGNORE_NAMES:
+        return True
+    return _GLOB_IGNORE_RE is not None and _GLOB_IGNORE_RE.match(normalized) is not None
 
 
 def should_ignore_path(path: str) -> bool:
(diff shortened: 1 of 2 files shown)


fancyboi999 (commented):
Confirmed on `main` @ `6044e5c5` — behavior-identical and a real hot-path win. This is the PR for #3655, and the equivalence holds where it's easy to get subtly wrong:

- **Double-normcase mirrors `fnmatch.fnmatch` exactly.** `fnmatch.fnmatch(name, pat)` internally compares `normcase(name)` against `normcase(pat)`. The new code mirrors both sides: literals are `normcase`'d at precompute (`_EXACT_IGNORE_NAMES`), glob patterns are `normcase`'d *before* `fnmatch.translate` (`_GLOB_IGNORE_RE`), and the lookup `normcase`'s the name once. So Windows case-folding and POSIX case-sensitivity are both preserved.
- **The literal/glob split is the right predicate.** `any(c in p for c in "*?[")` keys on exactly fnmatch's special characters, so every non-glob pattern goes to the O(1) set and only the ~10 real globs hit the regex. A literal routed to the set matches iff `normcase(name) == normcase(literal)` — same as `fnmatch` on a literal pattern.
- **The combined regex is sound.** `fnmatch.translate` emits `(?s:…)\Z`; joined with `|` and used with `re.match` (start-anchored), each branch is anchored at both ends, so a hit means the name fully matches one glob — equivalent to `any(fnmatch(name, g))`.
- **`should_ignore_path` is untouched** and still delegates per-segment, so it inherits the identical behavior (covered by `test_should_ignore_path_segments`).

The `_reference` oracle test (parity against the original loop across exact / glob / look-alike samples like `node_modules_x`, `x.git`, `myenv`) is the right safety net. One caveat, not a blocker: on Linux CI `os.path.normcase` is a no-op, so the Windows case-folding branch is preserved by construction but never actually exercised by the suite — worth a one-line comment so nobody assumes CI covers it.

Verified by reading the code + reasoning the fnmatch equivalence; I didn't run the suite (you note it passes locally + ruff clean). Looks correct — recommend merge.

WillemJiang (approved):

PR #3658, opened 2026-06-19
fix(gateway): attribute token usage to actual models

### LINKED ISSUE #3645: [bug] token-usage by_model attributes subagent/middleware tokens to the lead agent's model
## Problem summary

The `/api/threads/{id}/token-usage` aggregation reports a `by_model` breakdown that attributes **the entire run's token total — including subagent and middleware tokens — to the lead agent's single `model_name`**. Subagent and middleware models never appear, and the lead model is overstated. This also makes `by_model` internally inconsistent with `by_caller`.

Found by code inspection on latest `main` (`6044e5c5`).

## Affected area(s)

Backend API (gateway / endpoints / SSE), Agents / LangGraph

## What happened?

Token accounting is bucketed **by caller** (`lead_agent` / `subagent` / `middleware`) but **never by model**. The journal only accumulates caller buckets:

`backend/packages/harness/deerflow/runtime/journal.py` (`on_llm_end` and `record_external_llm_usage_records`):

```python
if caller.startswith("subagent:"):
    self._subagent_tokens += total_tk
elif caller.startswith("middleware:"):
    self._middleware_tokens += total_tk
else:
    self._lead_agent_tokens += total_tk
```

Only a single run-level `model_name` is persisted on the `RunRow`. The aggregation then groups by that one column and sums the **full** `total_tokens` per run:

`backend/packages/harness/deerflow/persistence/run/sql.py::aggregate_tokens_by_thread`:

```python
model_name = func.coalesce(RunRow.model_name, "unknown")
stmt = (
    select(
        model_name.label("model"),
        ...
        func.coalesce(func.sum(RunRow.total_tokens), 0).label("total_tokens"),
        func.coalesce(func.sum(RunRow.subagent_tokens), 0).label("subagent"),
        func.coalesce(func.sum(RunRow.middleware_tokens), 0).label("middleware"),
    )
    .where(_thread, _completed)
    .group_by(model_name)
)
...
by_model[r.model] = {"tokens": r.total_tokens, "runs": r.runs}
```

(The in-memory store in `runtime/runs/store/memory.py::aggregate_tokens_by_thread` has the same behavior.)

But subagents and middleware can run a **different** model than the lead agent:

- Subagents: `backend/packages/harness/deerflow/subagents/config.py` exposes `model: str = "inherit"`, and `resolve_subagent_model_name()` returns `config.model` when it isn't `"inherit"`.
- The title and summarization middlewares can be configured with their own `model_name` (`config.yaml` → `title.model_name`, `memory.model_name`, `summarization`).

So when a subagent (or the title/summary middleware) uses a different model, **its tokens are attributed to the lead agent's model** in `by_model`, the subagent/middleware model never shows up, and `by_model` contradicts `by_caller` (which correctly shows non-zero `subagent`/`middleware`).

## Expected behavior

`by_model` should attribute tokens to the model that actually produced them. `sum(by_model[*].tokens)` should equal `total_tokens`, and per-model totals should be consistent with `by_caller` (e.g. a subagent on a different model should contribute to that model's bucket, not the lead model's).

## Steps to reproduce

1. Configure a subagent (or `title`/`summarization`/`memory` middleware) to use a model different from the lead agent's model in `config.yaml`.
2. Enable token tracking (`run_events.track_token_usage: true`).
3. Run a conversation in a thread that dispatches a subagent (e.g. via the `task` tool) so subagent tokens are non-zero.
4. `GET /api/threads/{thread_id}/token-usage`.
5. Observe: `by_caller.subagent > 0`, but `by_model` only lists the lead agent's model, and that model's `tokens` includes the subagent's tokens. The subagent's actual model is absent.

## Relevant logs

```text
N/A — logic/accounting bug; no error is raised. Reproduced by inspecting the
/token-usage response (by_model vs by_caller) rather than a crash/log line.
```

## Root cause

The journal records token usage along the **caller** dimension only; it never records a `(caller, model)` or per-model dimension, and only one `model_name` is persisted per run. There is therefore no stored data from which a correct per-model breakdown can be derived. Fixing `by_model` requires recording token usage per model (e.g. capturing each LLM call's actual model, including subagent/middleware calls) rather than grouping by the run's single `model_name`.

## How are you running DeerFlow?

Local (make dev)

## Additional context

- `by_caller` and the headline totals (`total_tokens`, input/output) appear correct, with solid dedup (subagents run with a fresh callback config and report via `record_external_llm_usage_records`, deduped by `source_run_id`; the journal's own `on_llm_end` dedups by `run_id`). The issue is specifically `by_model`.
- Related prior work: #2734 (token accounting consistency / subagent coverage, closed) added the `by_caller` breakdown but did not address per-model attribution.

AnoobFeng:
Capture per-call model names from LLM response metadata for lead, middleware, and subagent calls.

Persist a per-run token_usage_by_model breakdown and aggregate by that map in both SQL and memory stores, with legacy fallback to the run-level model_name for older rows.

Add regression coverage for by_model totals, caller consistency, active progress snapshots, store parity, and SubagentTokenCollector model propagation.

Fixes #3645
Closes #3645

## Why

`GET /api/threads/{thread_id}/token-usage` previously built `by_model` from the run-level `RunRow.model_name`, which only represents the lead agent model. When subagents or middleware used different models, their tokens were still attributed to the lead model, so `by_model` overstated the lead model and omitted the actual subagent/middleware models.

The per-call model name is already available on LLM responses via `response_metadata`, but it was not being persisted or aggregated.

## What changed

- Capture per-call model names from LLM response metadata for lead, middleware, and subagent calls.
- Add a per-run `token_usage_by_model` breakdown.
- Persist that breakdown on run completion and progress snapshots.
- Aggregate `by_model` from `token_usage_by_model` in both SQL and memory stores.
- Preserve legacy fallback behavior for older rows without `token_usage_by_model`, using `RunRow.model_name + total_tokens`.
- Add regression coverage for:
  - `sum(by_model[*].tokens) == total_tokens`
  - subagent and middleware tokens landing on their actual model buckets
  - `by_caller` consistency
  - SQL and memory store parity
  - active progress snapshots
  - `SubagentTokenCollector` propagating model names into external usage records

## Database Migration Note

This PR adds a new `runs.token_usage_by_model` JSON column, but intentionally does not include a database migration yet.

I would like maintainer guidance on the preferred migration path for this project:

- add an Alembic migration in this PR
- provide manual SQL migration instructions
- handle the column/backfill through application startup or another compatibility path

The code includes a legacy fallback for rows where `token_usage_by_model` is empty or null, but existing SQL deployments still need the new column added before this code can query or write it.

## Bug fix verification

Test paths that reproduce/pin the bug:

- `backend/tests/test_token_usage_by_model.py`
- `backend/tests/test_subagent_token_collector.py`

These tests cover the regression where subagent/middleware tokens were incorrectly attributed to the lead model. The expected fixed behavior is that each actual model appears in `by_model` with its own token total.

I did not run the tests against `main` in this checkout, but the new tests encode the previously broken behavior and pass on this branch.

## Validation

Ran from `backend/`:

```shell
uv run pytest tests/test_subagent_token_collector.py tests/test_token_usage_by_model.py
uvx ruff check .
uvx ruff format --check .
```

## AI assistance

**Tool(s) used:** Claude Code


fancyboi999 (commented):
This implements the #3645 fix I triaged — capturing `response_metadata.model_name` per call and bucketing by it. Read the whole change on `main` @ `6044e5c5`; the accounting is correct end-to-end and the key invariant holds by construction. Verified:

- **`sum(by_model[*].tokens) == total_tokens` can't drift.** `_record_model_usage` is called from the exact two dedup-gated spots that add to the totals — inside `on_llm_end` under `total_tk > 0 and rid not in _counted_llm_run_ids`, and inside `record_external_llm_usage_records` under the `source_id` dedup. So every token added to a total is added to exactly one model bucket, never twice, never zero.
- **Model capture mirrors the existing extractor.** Lead/middleware (`journal.py`) and `SubagentTokenCollector` both read `response_metadata.get("model_name") or response_metadata.get("model")` — the same source `converters.py` already uses — falling back to an `"unknown"` bucket. Propagation is end-to-end: collector record → `record_external_llm_usage_records` → `_record_model_usage` → `get_completion_data` → store.
- **Persistence is wired on both paths.** `get_completion_data()` now includes `token_usage_by_model`; it flows through `**completion` at completion (`worker.py:411`) and through the progress snapshot (`_schedule_progress_flush` builds the snapshot from `get_completion_data()`, so `update_run_progress(**snapshot)` carries it) — so active runs get the breakdown too, matching your snapshot test.
- **Store parity + legacy fallback.** SQL and memory `aggregate_tokens_by_thread` reduce identically, and both fall back to `model_name + total_tokens` when the JSON is empty, so old rows keep the legacy lead-only behavior instead of vanishing. Headline totals / `by_caller` still come from their own columns, unaffected.

**On your migration question** — concrete answer with the repo's current behavior:

- Schema is bootstrapped by `Base.metadata.create_all` (`persistence/engine.py:151`/`161`), and `migrations/versions/` holds only `.gitkeep` — there are **no Alembic revisions in the repo yet**, so this would be the first one.
- `create_all` only creates *missing tables*; it does **not** add a column to a table that already exists. So fresh deployments get `runs.token_usage_by_model` automatically, but **existing SQL deployments won't** — and the write path (`update_run_completion`/`update_run_progress`) plus the new `select(... RunRow.token_usage_by_model)` will error with "no such column" there. The legacy fallback only covers the column being *present but NULL* (e.g. after a manual `ALTER ADD COLUMN` without backfill); it does not cover the column being *absent*.
- So existing deployments do need the column added. With no migration precedent, it's genuinely a maintainer call between (a) shipping the first Alembic migration here (sets the pattern) or (b) documenting a one-line `ALTER TABLE runs ADD COLUMN token_usage_by_model ...` for upgraders. Either works; what shouldn't ship is the column with no path for existing DBs.

Two minor, non-blocking notes inline.

Verified by reading the code + tracing the wiring; I didn't run the suite (you note the two new test files pass on the branch). Modulo the migration decision, the implementation looks correct — nice job closing this out.

backend/packages/harness/deerflow/persistence/run/sql.py
```diff
@@ -329,14 +341,28 @@ async def aggregate_tokens_by_thread(self, thread_id: str, *, include_active: bo
...
+            middleware += r.middleware_tokens
+
+            # ``or {}`` covers rows written before ``token_usage_by_model``
+            # existed (the column is NULL on a manual ALTER ADD COLUMN without
+            # backfill); fresh rows always carry the journal-produced dict.
+            usage_by_model = r.token_usage_by_model or {}
+            if usage_by_model:
+                for model, usage in usage_by_model.items():
+                    entry = by_model.setdefault(model, {"tokens": 0, "runs": 0})
+                    entry["tokens"] += usage["total_tokens"]
```
fancyboi999: Two small non-blocking things here:

- `usage["total_tokens"]` assumes each per-model dict always carries that key. It does for journal-written rows (`_record_model_usage` always sets all three counters), so it's safe today — but `usage.get("total_tokens", 0)` would harden it against a hand-edited / partially-written JSON blob without changing behavior.
- `entry["runs"] += 1` per model means a run that used N distinct models now contributes 1 to each of N model buckets, so `sum(by_model[*].runs)` can exceed `total_runs` (it used to be equal, since each run had a single `model_name`). Likely fine — just flagging in case anything downstream assumes the per-model run counts sum to `total_runs`.

AnoobFeng:
@fancyboi999 
Thanks for the detailed review.

A few follow-ups:

1. I pushed a small hardening update for the per-model reducer. SQL and memory aggregation now use `usage.get("total_tokens", 0)` when reading each per-model usage blob, so a partially written or manually edited JSON object will not break aggregation.

2. On `by_model[*].runs`, my intended semantics are “number of runs in which this model appeared”, not “exclusive run count for this model”. So if one run uses a lead model, a subagent model, and a middleware model, that run can increment multiple model buckets and `sum(by_model[*].runs)` may exceed `total_runs`.

3. I also checked the current frontend usage. It only maps `total_input_tokens`, `total_output_tokens`, and `total_tokens` for display, and does not consume `by_model[*].runs` today. So I think the current behavior is okay as long as the API meaning is clear.

4. On the migration path, both approaches you suggested sound good. I’ll wait for maintainer guidance first, then add a follow-up commit accordingly.

fancyboi999:
Thanks — verified the hardening landed: both stores now read `usage.get("total_tokens", 0)` (`sql.py:359`, `memory.py:117`), and the follow-up commit is exactly that 2-line change with nothing else, so the per-model reduce won't trip on a partial / hand-edited JSON blob. 👍

On `by_model[*].runs`: agreed, "number of runs in which this model appeared" is a sensible meaning. I checked the frontend too — `by_model` only shows up in `frontend/src/core/threads/types.ts:42` and isn't read by any component today, so nothing relies on `sum(runs) == total_runs`. Optional, non-blocking: a short comment on the response shape making the non-exclusive semantics explicit would be cheap insurance for whoever wires up a per-model view later and might otherwise expect `runs` to sum to `total_runs`.

On the migration: agreed, waiting for maintainer direction is the right call. Whichever way they go (first Alembic revision vs. a documented `ALTER`), the only must-have is that existing SQL deployments get the column before this code reads/writes it.

Nice work — the accounting itself is solid.

AnoobFeng:
Thanks, that makes sense. I added a short response-schema description for `by_model[*].runs` to make the non-exclusive semantics explicit for future consumers.

WillemJiang (approved):

PR #3660, opened 2026-06-19
fix(gateway): use correct event_type for thread message feedback

### LINKED ISSUE #3650: bug(gateway): thread messages feedback never attached (event_type mismatch)

itxaiohanglover:
Fixes #3650

## Why

`GET /api/threads/{thread_id}/messages` (`list_thread_messages`) attaches user feedback to the last AI message of each run by matching `event_type == "ai_message"`. But AI messages are persisted by `RunJournal` with `event_type="llm.ai.response"` (`runtime/journal.py`), and the event store returns that value verbatim. No production code path ever writes `event_type="ai_message"`.

Consequences:
- `last_ai_per_run` is always empty, so **feedback is never attached** — every message gets `feedback: null`. Thumbs up/down state is not restored when the thread view reloads.
- `feedback_repo.list_by_thread_grouped(...)` ran on **every** request and its result was never used — a wasted DB query per call.

## What changed

- Match the real event type (`"llm.ai.response"`) instead of the never-written `"ai_message"` when finding the last AI message per run in `list_thread_messages`. Feedback is now correctly attached to the last AI message of each run.
- Only run the grouped-feedback DB query when there is at least one AI message to attach it to, avoiding a wasted query on threads/runs with no AI messages yet.

## Surface area

## Bug fix verification

There was no existing test covering feedback attachment on this thread-level endpoint (only the per-run `/runs/{run_id}/messages` endpoint was tested), which is why the mismatch went unnoticed. The fix is a one-line event-type correction plus a guard around the feedback query. A dedicated regression test for feedback attachment on `GET /api/threads/{thread_id}/messages` is a natural follow-up.

## Validation

- Reviewed `runtime/journal.py` to confirm AI messages are persisted with `event_type="llm.ai.response"`.
- Confirmed `DbRunEventStore._row_to_dict` returns the stored `event_type` verbatim (no remapping).
- Confirmed existing tests (`test_thread_run_messages_pagination.py`) cover the per-run messages endpoint and are unaffected — they test `/{thread_id}/runs/{run_id}/messages`, not the thread-level `/{thread_id}/messages` endpoint changed here.

## AI assistance

**Tool(s) used:** Qoder

**How you used it:** Investigated the issue, located the mismatched string in `thread_runs.py`, confirmed the persisted value in `journal.py`, and applied the fix.


fancyboi999 (commented):
Took this through the full producer→store→consumer path and it holds up — the fix is correct and matches every claim in the description. Verified on current head:

- **Producer**: AI responses are persisted with `event_type="llm.ai.response"` (`runtime/journal.py:295`) — a bare string literal, no remapping.
- **`"ai_message"` is genuinely dead**: it's compared only on the line you're changing and written nowhere in `app/` or `packages/`, so `last_ai_per_run` was always empty and feedback never attached. Confirmed.
- **Verbatim pass-through**: `DbRunEventStore._row_to_dict` (`runtime/events/store/db.py:31-50`) massages `metadata`/`created_at`/`content` but leaves `event_type` untouched; the memory store keeps events as-is. So both backends surface `"llm.ai.response"`.
- **"last AI per run" is correct**: `list_messages` returns seq-ascending in both stores, so iterating and overwriting `last_ai_per_run[run_id] = i` lands on the highest-seq AI message of each run.
- **Guard is auth-safe**: moving `get_current_user` inside `if last_ai_indices:` doesn't change auth behavior — `get_current_user` (`deps.py:413`) is the non-raising `str | None` adapter (delegates to `get_optional_user_from_request`), not the 401-raising `get_current_user_from_request`. And `feedback_map` is only read inside the `i in last_ai_indices` branch, so the `{}` path is never dereferenced — no use-before-assign, no behavior change for the populated case.

Looks correct and mergeable to me. Two non-blocking suggestions:

1. **Add the regression test in this PR rather than deferring it.** The reason this slipped is precisely that thread-level feedback attachment has zero coverage — `test_thread_run_messages_pagination.py` only exercises the per-run `/{thread_id}/runs/{run_id}/messages` endpoint. A small test (thread with an AI message + feedback → `GET /{thread_id}/messages` returns it on the last AI message; thread with no AI messages → `feedback: null` and no grouped query) would both prove the fix and pin the contract so it can't silently regress again.
2. See the inline note on the matched string — worth pinning the event-type contract so producer and consumer can't drift a second time.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -352,19 +352,24 @@ async def list_thread_messages(
...
-    user_id = await get_current_user(request)
-    feedback_map = await feedback_repo.list_by_thread_grouped(thread_id, user_id=user_id)
-
-    # Find the last ai_message per run_id
+    # Find the last AI message per run_id. RunJournal persists these with
+    # event_type="llm.ai.response" (see runtime/journal.py), not "ai_message".
     last_ai_per_run: dict[str, int] = {}  # run_id -> index in messages list
     for i, msg in enumerate(messages):
-        if msg.get("event_type") == "ai_message":
+        if msg.get("event_type") == "llm.ai.response":
```
fancyboi999: This is the heart of the bug, and it's a string-drift between two copies of the same literal: the producer hard-codes `"llm.ai.response"` (`runtime/journal.py:295`) and this consumer now hard-codes it again. The original `"ai_message"` bug was the same failure mode — nothing keeps the two in sync, so the next time someone renames the journal event type, feedback attachment breaks silently in exactly this way (and stays invisible, since it's untested).

Worth extracting a single source of truth — e.g. an `EventType` enum or module-level constant referenced by both `journal.py` and here — so producer and consumer can't diverge. Not a blocker for this fix, but it closes the whole class rather than this one instance.

itxaiohanglover: Great point — implemented in follow-up commits. Extracted `AI_RESPONSE_EVENT_TYPE = "llm.ai.response"` as a module-level constant in `runtime/journal.py` and imported it in `thread_runs.py`. Both producer and consumer now reference the same constant, closing the string-drift class of bugs.

WillemJiang:
the issue is fixed in https://github.com/bytedance/deer-flow/pull/3651

PR #3663, opened 2026-06-20
feat: persist AI turn duration in backend and UI

chetan655:
### **Summary:**
This PR is a follow-up to #3627  which introduced the frontend UI for the AI execution timer. 

Previously, the timer was purely a frontend calculation that disappeared upon refreshing the page because the backend API did not persist or serve the execution times of historical runs. This PR introduces a full-stack feature to calculate historical execution times on the backend and serve them directly to the UI, bridging the gap between live streaming and static history.

### **Changes Included:**

**Backend (`backend/app/gateway/routers/`)**
*   **`threads.py` & `thread_runs.py`**: Fetched historical execution times for runs from the database and injected the timer duration directly into the message data payload at the API boundary.

**Frontend (`frontend/src/`)**
*   **`components/workspace/messages/message-list-item.tsx`**: Updated the message UI to read the duration directly from the backend data, and built a client-side cache to seamlessly preserve the timer when switching between live streaming and static history states.
*   **`core/threads/hooks.ts`**: Ensured the frontend deduplication logic properly preserves the backend metadata when merging data with the raw LangGraph stream.

### **Testing:**
*   Verified that live streaming timers still function properly.
*   Verified that timers accurately persist upon refreshing the page.


backend/app/gateway/routers/thread_runs.py
```diff
@@ -378,6 +393,12 @@ async def list_thread_messages(
                 if fb
                 else None
             )
+
+            # Inject turn_duration into additional_kwargs
+            if run_id in run_durations:
```
fancyboi999: This injection never fires on this endpoint. It's gated by `i in last_ai_indices`, and `last_ai_per_run` is built a few lines up from `msg.get("event_type") == "ai_message"` (:378) — but AI responses are persisted as `llm.ai.response`, never `ai_message` (the exact mismatch #3660 is open to fix). So `last_ai_indices` is empty here and `turn_duration` is never attached on `list_thread_messages`.

Two follow-ons: (1) this block textually conflicts with #3660, which rewrites the same loop; (2) even standalone, the thread-level duration won't show until it matches `llm.ai.response`. Easiest path is to rebase on #3660 and reuse its corrected `last_ai_per_run` instead of carrying a second copy here.

frontend/src/components/workspace/messages/message-list-item.tsx
```diff
@@ -210,6 +210,8 @@ function MessageImage({
   );
 }
 
+const clientTurnDurations = new Map<string, number>();
```
fancyboi999, line 213: This module-level `Map` is process-global and never cleared, so it accumulates an entry per AI-message id for the lifetime of the tab and is shared across every `MessageContent_` instance and across thread switches — a slow leak and a surprising shared-state coupling for what's meant to be a per-message cache. If the goal is just to bridge live→history for the current view, consider keying this off React state / a ref scoped to the thread (cleared on thread change), or a bounded LRU, rather than an unbounded module singleton.

fancyboi999 (commented):
Nice follow-up to #3627 — persisting the timer so it survives refresh is the right call. I traced all three backend injection sites and the frontend cache; the direction is good, but there's a real dead path and meaningful duplication I'd sort before merge.

The core issue: **"compute run duration + inject into the AI message" is implemented three times, and the three have already drifted apart.**
- `list_thread_messages` (thread_runs.py) keys the injection off the last-AI-per-run set built from `event_type == "ai_message"` (:378). That string is never produced — `RunJournal` persists AI responses as `llm.ai.response` — which is exactly the mismatch #3660 is open to fix. So `last_ai_indices` is empty and **`turn_duration` is never injected on this endpoint** (inline). This block also textually overlaps #3660, so they'll conflict.
- `list_run_messages` detects the target via `content.get("type") == "ai"` (correct shape) — works.
- `get_thread_history` (threads.py) builds a separate `msg_to_run` map from an extra `event_store.list_messages(..., limit=1000)` and injects into every AI message — also works, but via a third mechanism.

So the same AI message gets a duration from two endpoints and nothing from the thread-level one — a user-visible inconsistency that falls straight out of the duplication. I'd extract one helper (`compute_run_durations(runs) -> {run_id: seconds}`) plus one agreed "which message carries it" rule, and call it from all three (rebasing the thread-level one onto #3660's `llm.ai.response`).

Other things worth addressing:
- **Fail-silent + magic cap.** All three sites wrap injection in `except Exception: pass` with no log, so any future timestamp/shape drift silently yields no durations. And `get_thread_history`'s `limit=1000` second full message-fetch silently drops durations for older messages on long threads. At least log in the except, and reconsider the extra full fetch.
- **Semantics:** `updated_at - created_at` is the run row's lifetime (creation → last mutation), not strictly turn end — `manager.py` rewrites `updated_at` on several status transitions, so if the row is touched after the turn completes the number can overshoot. Probably fine as "total turn time," just worth stating the assumption.
- **Frontend:** the module-level `clientTurnDurations` Map (inline) is never cleared — it grows for the whole session and is shared across instances/thread switches. And the reasoning-only branch swaps `className={className}` for a hardcoded `"w-full max-w-full"` (:329), dropping the caller's className — looks like unrelated layout scope-creep in a persistence PR.
- **No backend tests** for any of the three new paths; the dead `ai_message` path above is exactly what one endpoint-level test ("the AI message in the response carries `turn_duration`") would have caught.

Good feature, fixable issues — but I'd consolidate the three injectors and fix the dead thread-level path before this merges.

chetan655:
@fancyboi999 Thanks for the review. I've just pushed an update that addresses all of your feedback:

**1. Consolidated Extraction & Injection Logic**
Extracted a single `compute_run_durations(runs) -> dict[str, int]` helper to handle the timestamp math. Also standardized the injection rule across all 3 endpoints (`list_thread_messages`, `list_run_messages`, `get_thread_history`) to look for the raw LangGraph `content.get("type") == "ai"` instead of relying on the buggy `event_type` column. This completely bypasses the #3660 mismatch since it targets the standardized `content` payload directly.

**2. Removed Fail-Silent & Documented Overshoot**
Replaced the bare `except Exception: pass` blocks with `logger.warning` with `exc_info=True`. Also added a comment inside the `compute_run_durations` helper documenting the semantics that `updated_at - created_at` represents the row's total lifetime and can slightly overshoot. Added a `FIXME` for the `limit=1000` cap in `threads.py` as well.

**3. Frontend Memory Leak & Scope Creep Fixed**
Added a `useEffect` cleanup hook in `message-list-item.tsx` to explicitly clear the `clientTurnDurations` Map whenever the `threadId` changes so it no longer leaks globally across the session. Also reverted the unrelated `className={className}` layout change.

**4. Added Backend Tests**
Added a new endpoint-level backend test (`test_list_run_messages_injects_turn_duration` in `test_thread_run_messages_pagination.py`) to explicitly verify that the injection correctly fires and attaches `turn_duration` to the AI messages in the response. I also updated the `_make_app` test helper to always provide a mock `run_manager` so the duration calculations don't cause 503 errors on the existing pagination tests.

*(Note: During the previous commits, I accidentally deleted the `return {"data": data, "has_more": has_more}` statement at the bottom of `list_run_messages`, which was causing some pagination tests to fail with a 503/ResponseValidationError. That is now fixed, and all tests are passing cleanly!)*

fancyboi999 (approved):
Good iteration — the core issues from my previous review are resolved. Approving with a few remaining notes.

**What's fixed**
- `compute_run_durations()` helper extracted and shared across all three endpoints ✅
- Dead path in `list_thread_messages` fixed: now targets `content.get("type") == "ai"` (the LangGraph payload shape) instead of the never-populated `event_type == "ai_message"` column ✅
- `except Exception: pass` replaced with `logger.warning(..., exc_info=True)` ✅
- `updated_at - created_at` overshoot semantics documented ✅
- `limit=1000` FIXME'd ✅
- Frontend map now cleared on thread switch via `useEffect` cleanup ✅
- Unrelated `className` layout change reverted ✅
- Backend test for `list_run_messages` added ✅

**Three remaining things worth addressing (non-blocking but worth knowing)**

1. **`clientTurnDurations.clear()` fires on every component unmount, not just thread-switch.**

   The `useEffect(() => () => clientTurnDurations.clear(), [threadId])` cleanup runs on component unmount too — not just when `threadId` changes. If any `MessageContent_` instance unmounts for any reason (e.g. list re-render, scroll virtualization), it wipes the entire shared map, clearing cached durations for all other rendered messages. Local `cachedDuration` state softens the impact (the value survives while the component stays mounted), but any remount will reinitialize to undefined.

   A safer fix: key the map as `${threadId}:${messageId}` so you can clear by thread prefix on switch without wiping unrelated live entries.

2. **`list_thread_messages` and `list_run_messages` inject `turn_duration` at different levels.**

   `list_thread_messages` writes to `msg["additional_kwargs"]` (outer RunMessage envelope), while `list_run_messages` writes to `content["additional_kwargs"]` (inner LangGraph Message). The current frontend only calls `list_run_messages` for history (via `buildRunMessagesUrl`), so no user-visible bug today — but any external caller of `list_thread_messages` would find `turn_duration` at a different path than expected.

3. **Only one of three injection paths has a backend test.**

   The new `test_list_run_messages_injects_turn_duration` test is a good pattern — worth replicating for `list_thread_messages` (which now unconditionally calls `run_mgr.list_by_thread` on every request, a new dependency) and for the `get_thread_history` injection path.

Overall a clear improvement. The dead-path fix and consolidation are the most important parts and both land correctly.

chetan655:
Thanks! Pushed fixes for all three:

1. **Frontend map:** Re-keyed as `${threadId}:${messageId}`, cleanup now deletes by prefix instead of `.clear()` — unmounts no longer wipe other messages' cached durations.
2. **Injection level:** Aligned `list_thread_messages` to inject into `content["additional_kwargs"]` to match `list_run_messages`.
3. **Test coverage:** Added `test_list_thread_messages_injects_turn_duration`. 
4. **Live Timer UX Sync:** Refactored the live reasoning timer in `reasoning.tsx` to use `Math.floor` instead of `Math.ceil`. Previously, the moment the timer started (e.g., at 0.01s), it would instantly display `(1s)` which felt unnatural. `Math.floor` makes it behave like a natural stopwatch. Crucially, this also ensures the frontend calculation exactly matches the backend's `int()` truncation, which fixes a visual glitch where the timer would briefly jump backward by 1 second when the final server response arrived.

fancyboi999 (commented):
Re-checked the new push against the three notes — all addressed, and I re-ran the backend tests to confirm. My earlier approval stands.

- **Frontend map (note 1):** Re-keyed to `${threadId}:${message.id}` and the cleanup now deletes by `${threadId}:` prefix instead of `.clear()` (`message-list-item.tsx:237/244/252` + the `[threadId]` cleanup at `:259-267`). The cross-thread / cross-instance wipe is gone — an unmount no longer nukes other threads' cached durations.
- **Injection level (note 2):** `list_thread_messages` now writes to `content["additional_kwargs"]` (`thread_runs.py:689-690`), matching `list_run_messages` (`:732-733`). The two endpoints agree on where `turn_duration` lands now.
- **Test coverage (note 3):** `test_list_thread_messages_injects_turn_duration` added. I ran the file locally: `15 passed` for the whole `test_thread_run_messages_pagination.py` suite, and the two `*_injects_turn_duration` tests pass specifically. The new one correctly asserts the AI message carries `turn_duration` in `content.additional_kwargs` while the human message doesn't.
- **Bonus `Math.floor` (reasoning.tsx):** Good catch and on-topic for a persistence PR — `Math.floor` matches the backend's `int((updated - created).total_seconds())` truncation in `compute_run_durations`, so the live timer no longer jumps back a second when the persisted value arrives. Both `int()` (toward zero) and `Math.floor` agree for the non-negative durations here.

One tiny residual on note 1, non-blocking and fine to leave: the prefix-cleanup still removes *all* `${threadId}:*` entries when a single `MessageContent_` for that thread unmounts, so same-thread siblings lose their cached entry on a lone unmount (list virtualization, etc.). In practice the local `cachedDuration` state plus the backend injection re-populate it, so it's not user-visible — just noting it since it's the same class as the original concern, much reduced.

Nice iteration — this reads as ready to me.

WillemJiang (approved):

chetan655:
Great point about the cleanup still being tied to the message unmount. Since it's protected by the local `cachedDuration` state, leaving it as-is works for now, but that pattern is definitely worth keeping in mind for future thread-level scoping.

PR #3673, opened 2026-06-20
fix(frontend): reset new chat on client-side navigation

AnoobFeng:
Drive the chat reset effect with Next.js's reactive pathname instead of the render-time window.location-derived flag.

During App Router transitions, window.location may still point to the previous thread until commit, leaving chat state stale until another UI interaction triggers a render. Preserve the stale 'new' param guard so created thread UUIDs are not overwritten.

## Why

After the first message creates a thread, DeerFlow updates the URL with `history.replaceState()` to avoid remounting the active chat. This can leave `useParams().thread_id` as `"new"` while the browser URL points to the created thread.

When **New Chat** was clicked, React could render before `window.location.pathname` was updated. As a result, `isNewPath` remained `false` and the reset effect did not run.

From the user's perspective, clicking **New Chat** appeared to do nothing: the page continued showing the previous conversation. A second click or another UI interaction triggered a re-render and completed the reset.

## What changed

- Drive the chat synchronization effect with Next.js's reactive pathname.
- Determine the new-chat destination from pathname inside the effect.
- Add a Playwright regression test covering the history-only URL transition used after thread creation.
- Verify that one New Chat click clears the previous conversation without requiring a second click or unrelated UI interaction.

## Bug fix verification

    - `Thread history › new chat resets immediately after a history-only thread URL update`

- Did it go red on main and green on this branch? yes
- Red result on the old implementation:
    - The URL changed to `/workspace/chats/new`.
    - `Hello from DeerFlow!` remained visible.
    - The assertion expected 0 matching messages but received 1.

- Green result on this branch:
    - The URL changed to `/workspace/chats/new`.
    - The previous message disappeared after one click.
    - No second click or unrelated UI interaction was required.

## Validation

Commands actually run from `frontend/`:
```bash
pnpm format
pnpm lint
pnpm typecheck
BETTER_AUTH_SECRET=local-dev-secret pnpm build
pnpm test
pnpm test:e2e -- --reporter=line
```

Results:
- Prettier: passed.
- ESLint: passed with 0 errors and 2 pre-existing warnings in unrelated files.
- TypeScript: passed.
- Production build: passed with one pre-existing Turbopack NFT tracing warning.
- Vitest: 34 test files passed, 328 tests passed.
- Playwright E2E: 49 tests passed.
- Regression test:
    - Red on the old implementation.
    - Green on the fixed implementation.

## AI assistance

**Tool(s) used:** Codex

### DIFF
--- a/frontend/src/components/workspace/chats/use-thread-chat.ts
+++ b/frontend/src/components/workspace/chats/use-thread-chat.ts
@@ -27,6 +27,9 @@ export function resetThreadChatAfterDelete(detail: ThreadChatResetDetail) {
 export function useThreadChat() {
   const { thread_id: threadIdFromPath } = useParams<{ thread_id: string }>();
   const pathname = usePathname();
+  // Render-time values use the committed browser URL. The sync effect below
+  // intentionally watches the reactive pathname so client navigation still
+  // schedules a reset when window.location is stale during render.
   const actualPathname =
     typeof window === "undefined" ? pathname : window.location.pathname;
   const isNewPath = actualPathname.endsWith("/new");
@@ -65,11 +68,9 @@ export function useThreadChat() {
       return;
     }
     newThreadIdRef.current = null;
-    // Guard: after history.replaceState updates the URL from /chats/new to
-    // /chats/{UUID}, Next.js useParams may still return the stale "new" value
-    // because replaceState does not trigger router updates.  Avoid propagating
-    // this invalid thread ID to downstream hooks (e.g. useStream), which would
-    // cause a 422 from LangGraph Server.
+    // Native history updates the canonical pathname but preserves the route
+    // tree, so useParams may still return the stale "new" value. Avoid passing
+    // it to downstream hooks (e.g. useStream), which would cause a 422.
     if (threadIdFromPath === "new") {
       return;
     }
--- a/frontend/src/components/workspace/chats/use-thread-chat.ts
+++ b/frontend/src/components/workspace/chats/use-thread-chat.ts
@@ -57,7 +57,7 @@ export function useThreadChat() {
   }, []);
 
   useEffect(() => {
-    if (isNewPath) {
+    if (pathname.endsWith("/new")) {
       const nextThreadId = newThreadIdRef.current ?? uuid();
       newThreadIdRef.current = nextThreadId;
       setIsNewThreadState(true);
@@ -75,7 +75,7 @@ export function useThreadChat() {
     }
     setIsNewThreadState(false);
     setThreadIdState(threadIdFromPath);
-  }, [isNewPath, threadIdFromPath]);
+  }, [pathname, threadIdFromPath]);
 
   useEffect(() => {
     const handleReset = (event: Event) => {
--- a/frontend/tests/e2e/thread-history.spec.ts
+++ b/frontend/tests/e2e/thread-history.spec.ts
@@ -202,6 +202,43 @@ test.describe("Thread history", () => {
     await expect(page.getByPlaceholder(/how can i assist you/i)).toBeVisible();
   });
 
+  test("new chat resets immediately after a history-only thread URL update", async ({
+    page,
+  }) => {
+    mockLangGraphAPI(page);
+
+    await page.goto("/workspace/chats/new");
+    const textarea = page.getByPlaceholder(/how can i assist you/i);
+    await expect(textarea).toBeVisible({ timeout: 15_000 });
+    await textarea.fill("Message that must disappear in the next new chat");
+    await textarea.press("Enter");
+    await expect(page.getByText("Hello from DeerFlow!")).toBeVisible({
+      timeout: 15_000,
+    });
+
+    // A newly created chat changes the URL with history.replaceState so the
+    // active stream is not remounted. Reproduce that history-only transition:
+    // the canonical pathname becomes the UUID while useParams can stay "new".
+    await page.evaluate((threadId) => {
+      history.replaceState(null, "", `/workspace/chats/${threadId}`);
+    }, MOCK_THREAD_ID);
+
+    const newChatLink = page.locator(
+      "[data-sidebar='sidebar'] a[href='/workspace/chats/new']",
+    );
+    await expect(page).toHaveURL(
+      new RegExp(`/workspace/chats/${MOCK_THREAD_ID}$`),
+    );
+    await expect(newChatLink).toHaveAttribute("data-active", "false");
+
+    // One click must reset the chat without a second click or unrelated UI
+    // interaction forcing another render.
+    await newChatLink.click();
+    await expect(page).toHaveURL(/\/workspace\/chats\/new$/);
+    await expect(page.getByText("Hello from DeerFlow!")).toHaveCount(0);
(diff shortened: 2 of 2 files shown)


frontend/src/components/workspace/chats/use-thread-chat.ts
```diff
@@ -57,7 +57,7 @@ export function useThreadChat() {
   }, []);
 
   useEffect(() => {
-    if (isNewPath) {
+    if (pathname.endsWith("/new")) {
```
fancyboi999, line 63: Consistency note (non-blocking): the effect now reads the reactive `pathname` here, but the top-level `isNewPath` (line 31) and its render-time ref-init (line 37) still derive from the non-reactive `window.location.pathname`. So there are now two sources of truth for "is this the new path" in the same hook. It works — this effect is self-sufficient (`newThreadIdRef.current ?? uuid()` covers the case where the line-37 init didn't run because the window value was stale) — but for clarity you might move line 31/37 onto the same `pathname` too, or drop a comment explaining the intentional split. Purely a readability/maintenance point, not a correctness issue.

fancyboi999 (commented):
Traced the mechanism in `use-thread-chat.ts` and the new e2e test — the fix is correct. Confirming review (+ one consistency note inline), since review hands are short here.

What I verified by reading the source:

- **Root cause is real.** The reset effect keyed off `isNewPath`, which is derived at render time from `window.location.pathname` (`use-thread-chat.ts:29-31`), a non-reactive source. After thread creation the URL is changed with `history.replaceState` (no router update), and on a New Chat click React can render before `window.location.pathname` flips — so `isNewPath` stays `false`, the `[isNewPath, ...]` effect dep doesn't change, and the reset never fires until some later render. That matches the "first click does nothing" symptom.
- **The fix addresses it at the right layer.** Switching the effect to `pathname.endsWith("/new")` with `usePathname()` (reactive) as the dep means the effect re-runs on the client-side navigation itself, not on a stale render-time read. `usePathname()` excludes query/hash and thread IDs are UUIDs (never `/new`), so `.endsWith("/new")` won't false-match.
- **No regression for thread-to-thread switches:** `threadIdFromPath` is still a dep, and the stale-`"new"` guard (the `history.replaceState` -> 422 protection) is preserved unchanged.

One note left inline. Net: looks correct to me. I verified the render-vs-reactive mechanism by source-tracing rather than re-running your Playwright suite, but your documented red-on-main / green-on-branch result is consistent with the code path.

frontend/src/components/workspace/chats/use-thread-chat.ts
```diff
@@ -75,7 +75,7 @@ export function useThreadChat() {
     }
     setIsNewThreadState(false);
     setThreadIdState(threadIdFromPath);
-  }, [isNewPath, threadIdFromPath]);
+  }, [pathname, threadIdFromPath]);
```
willem-bd, line 79: Building on @fancyboi999's review — one path I'd like to double-check, because I suspect the reset effect doesn't actually re-run in the exact scenario this e2e test exercises.

After `onStart`, the URL is moved with `history.replaceState` (`page.tsx`), which updates `window.location` but not the Next.js router — so `usePathname()` is still `/workspace/chats/new` and `useParams().thread_id` is still `"new"`. The **New Chat** link then targets that same `/workspace/chats/new`, so across the click `pathname` and `threadIdFromPath` are both unchanged. That makes the new dep array `[pathname, threadIdFromPath]` here referentially identical on the next render, and I don't think the effect re-runs — the opposite of "it re-runs on the client-side navigation."

If that's right, the message still disappears only via the unchanged return-value branch (`isNewPath` from `window.location` flipping `isNewThread`→`true`), and the effect body that mints a fresh id / re-syncs `threadIdState` never executes — leaving `isNewThread=true` while the returned `threadId`/`displayThreadId` still hold the previous thread's UUID.

Would you mind confirming by checking the hook's `threadId` right after the single New Chat click? A fresh UUID means the effect ran (and @fancyboi999's read holds for this path); the previous thread's UUID means it didn't, and the reset is coming entirely from the return value. I couldn't fully reconcile the red-on-main / green-on-branch result with this model, so there may be a Next.js timing detail I'm missing — happy to be wrong.

willem-bd (commented):
Thanks @AnoobFeng. Building on @fancyboi999's review, I have one verification question about whether the reset effect actually re-runs in the `history.replaceState` → New Chat path the new e2e test covers. Details inline.

frontend/src/components/workspace/chats/use-thread-chat.ts
fancyboi999: Good question — I didn't want to hand-wave it, so I instrumented the hook (counters inside each branch of this effect + a snapshot of the returned `threadId`/`isNewThread`) and ran your exact e2e steps against a production build, on both the `isNewPath` (main) and `pathname` (this PR) versions. The data says the effect *does* re-run on this branch — and the Next.js detail you suspected you were missing is the crux.

Measured right across the single New Chat click:

| | main (`isNewPath` dep) | this PR (`pathname` dep) |
|---|---|---|
| pre-click `usePathname()` | `/workspace/chats/<uuid>` | `/workspace/chats/<uuid>` |
| effect runs during click | **0** | **2** (both new-path branch) |
| returned `threadId` after click | **unchanged** (same uuid) | **fresh uuid** (minted in effect) |
| `Hello from DeerFlow!` after click | **still visible (1)** | **gone (0)** |

The key is the pre-click row: after `history.replaceState(..., /chats/<uuid>)`, `usePathname()` already reads `/workspace/chats/<uuid>`, **not** `/workspace/chats/new`. Since Next 14.1 (we're on 16.2.6) Next instruments `history.pushState`/`replaceState` so the App Router picks them up and `usePathname()` reflects the new URL — even though `useParams().thread_id` stays `"new"` (route-segment params only update on a real navigation, which is the original bug). So across the click `pathname` genuinely changes `/chats/<uuid>` → `/new`; the `[pathname, threadIdFromPath]` dep does change, and the effect re-runs (effectRuns=2, new-path branch, returned `threadId` becomes a fresh uuid ≠ the pre-click one).

Your litmus test answers itself with that: fresh uuid ⇒ the effect ran, so the reset comes from the effect, not the return-value branch. And it lines up with red-on-main: there the dep is window-derived `isNewPath`, which stays `false` through the click (effectRuns=0, returned `threadId` unchanged, message still visible) — exactly the "first click does nothing" symptom.

Where your model is exactly right is on a Next without that History-API instrumentation (pre-14.1): there `usePathname()` would stay `/new` after `replaceState`, the dep wouldn't change, and the reset would have to come from the return value — leaving the `isNewThread=true` / stale-`threadId` mismatch you described. On this repo's Next 16 it doesn't reproduce, but it's a real fragility: the fix leans on the replaceState→usePathname behavior. Happy to share the instrumented patch if you want to re-run it.

AnoobFeng:
Thanks both for the thorough reviews and verification! I added a comment clarifying that this relies on Next.js's native History API integration: `replaceState()` updates the canonical pathname observed by `usePathname()`, while the preserved route tree may leave `useParams()` stale. The regression E2E also captures this behavior for future Next.js upgrades.

WillemJiang (approved):

PR #3674, opened 2026-06-20
fix(channels): let UI runtime channel config win over config.yaml

he-yufeng:
### Problem

`merge_runtime_channel_configs` merges the persisted runtime channel config into the config.yaml-derived `channels_config`, but the merge order is reversed, so the yaml value wins on any shared key:

```python
existing = channels_config.get(provider)   # from config.yaml
merged = dict(runtime_config)               # from the UI runtime store
if isinstance(existing, dict):
    merged.update(existing)                 # yaml overwrites the UI value
channels_config[provider] = merged
```

`ChannelRuntimeConfigStore` is documented as the store for "channel credentials entered from the UI", existing so a deployment "get[s] durable runtime configuration without needing a config.yaml edit". So when a provider is present in both places, the value the user just saved in the UI should take effect. Today, if that provider also has a `config.yaml` entry, the yaml value silently overwrites the UI one — e.g. a `bot_token` set in the UI is ignored in favor of the stale yaml token whenever both exist.

### Fix

Use the yaml entry as the base and let the runtime (UI) values override it. yaml-only keys are still preserved, runtime-only keys are still added, and shared keys now resolve to the UI value:

```python
existing = channels_config.get(provider)
merged = dict(existing) if isinstance(existing, dict) else {}
merged.update(runtime_config)
channels_config[provider] = merged
```

### Test

Added `backend/tests/test_runtime_channel_config_merge.py` (pure, no event loop / network). It asserts the UI value wins on a shared key while a yaml-only key is preserved; it fails on `main` with `assert 'yaml_token' == 'user_token'` and passes with this change. The existing channel-connection router/config suites still pass (41 tests), so nothing relied on the previous precedence.

If the yaml-wins behavior was intentional (e.g. treating config.yaml as the authoritative source operators can't override from the UI), let me know and I'll close this — but it seems to contradict the runtime store's stated purpose.

### DIFF
--- a/backend/app/channels/runtime_config_store.py
+++ b/backend/app/channels/runtime_config_store.py
@@ -125,9 +125,8 @@ def merge_runtime_channel_configs(
             channels_config.pop(provider, None)
             continue
         existing = channels_config.get(provider)
-        merged = dict(runtime_config)
-        if isinstance(existing, dict):
-            merged.update(existing)
+        merged = dict(existing) if isinstance(existing, dict) else {}
+        merged.update(runtime_config)
         channels_config[provider] = merged
 
 
--- a/backend/tests/test_channels.py
+++ b/backend/tests/test_channels.py
@@ -4528,7 +4528,8 @@ class TestChannelService:
         service = ChannelService.from_app_config(app_config)
 
         assert service._config["telegram"]["bot_token"] == "telegram-token"
-        assert service._config["slack"]["app_token"] == "xapp"
+        # The runtime (UI-entered) value must win over the yaml value.
+        assert service._config["slack"]["app_token"] == "xapp-ui"
         assert service._config["discord"]["bot_token"] == "discord-bot-token"
 
     def test_from_app_config_loads_persisted_runtime_channel_config(self, monkeypatch, tmp_path):
--- a/backend/tests/test_runtime_channel_config_merge.py
+++ b/backend/tests/test_runtime_channel_config_merge.py
@@ -1,0 +1,43 @@
+"""Precedence tests for merge_runtime_channel_configs (pure, no event loop)."""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+
+from app.channels.runtime_config_store import merge_runtime_channel_configs
+
+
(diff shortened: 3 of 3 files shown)


fancyboi999 (commented):
Read the diff, the function, both call sites, and the new test — this is a correct, well-scoped fix. Sharing a confirming review since review hands are short here.

What I verified:

- **The bug is real.** `runtime_config_store.py:127-130` builds `merged = dict(runtime_config)` then `merged.update(existing)`, so the config.yaml `existing` value clobbers the UI/runtime value on any shared key. Your fix flips the base/override order so runtime wins.
- **Runtime-wins is the correct precedence**, not the other way around. The store's own docstring (`runtime_config_store.py:17-22`) states it holds "channel credentials entered from the UI" so a deployment gets "durable runtime configuration without needing a config.yaml edit" — i.e. the UI value the user just saved must take effect. And the function name/intent is "merge persisted *runtime* provider config into channels_config", so the thing being merged in should override the base.
- **Both call sites agree.** `service.py:65` (`_merge_channel_connection_runtime_config`) and `channel_connections.py:181` (`_load_channels_config`) both pass the yaml-derived dict as `channels_config` and layer runtime on top — neither wants yaml to win. So the flip is consistent everywhere, no caller relied on the old precedence.
- **Key-preservation semantics hold:** yaml-only keys survive (base copied first), runtime-only keys are added (`update`), shared keys resolve to the UI value. The `_runtime_channel_disconnected` pop path and the non-dict-`existing` branch are unchanged.
- **Test genuinely pins it.** I ran your assertion against current main and it fails (`yaml_token` != `user_token`); applying the 3-line fix turns it green. Real red→green, not vacuous.

The "if yaml-wins was intentional" question you raised in the description: I don't think it was — nothing in the store's purpose or either caller supports yaml being authoritative over a value the user explicitly set in the UI. Looks correct to me and safe to merge.

WillemJiang:
@he-yufeng thanks for the contribution. Please fix the unit test errors.

WillemJiang (approved):

PR #3675, opened 2026-06-20
feat(community): add GroundRoute web search + fetch engine

jp0xz:
## What
Adds GroundRoute as an opt-in community search engine providing `web_search` and `web_fetch` tools, mirroring the existing community engines (tavily, serper, exa, brave, firecrawl).

GroundRoute is a meta search layer: one API in front of six engines (Serper, Brave, Exa, Tavily, Firecrawl, Perplexity). It routes each query to the cheapest engine that clears a quality bar and fails over if one is down, so high-volume research runs keep working when a single provider is rate-limited or unavailable.

## Why
DeerFlow's research loop fans out into many searches per task. Teams running it at volume hit single-engine cost, rate-limit, and outage pain. GroundRoute is a drop-in multi-engine option with automatic failover, fully opt-in and selected via config like any other engine.

## Changes
- `backend/packages/harness/deerflow/community/groundroute/{tools.py,__init__.py}`: `web_search_tool` + `web_fetch_tool` (httpx only, no new required dependency; httpx is already core).
- `scripts/wizard/providers.py` + `config.example.yaml`: registration.
- Harness pyproject: `groundroute = []` optional extra so `uv add 'deerflow-harness[groundroute]'` matches the documented pattern (no-op, httpx is core).
- Docs: `backend/docs/CONFIGURATION.md` + `frontend/src/content/en,zh/harness/tools.mdx`.
- Tests: `backend/tests/test_groundroute_tools.py`.

## Notes
- Fully opt-in. No behavior change and no new required dependencies for existing users.
- API key via `GROUNDROUTE_API_KEY` or the tool `api_key` config.

### DIFF
--- a/backend/packages/harness/deerflow/community/groundroute/tools.py
+++ b/backend/packages/harness/deerflow/community/groundroute/tools.py
@@ -87,10 +87,7 @@ def web_search_tool(query: str, max_results: int = 5) -> str:
     if not api_key:
         if not _api_key_warned:
             _api_key_warned = True
-            logger.warning(
-                "GroundRoute API key is not set. Set GROUNDROUTE_API_KEY in your environment "
-                "or provide api_key in config.yaml. Get a free key at https://groundroute.ai/keys"
-            )
+            logger.warning("GroundRoute API key is not set. Set GROUNDROUTE_API_KEY in your environment or provide api_key in config.yaml. Get a free key at https://groundroute.ai/keys")
         return json.dumps(
             {"error": "GROUNDROUTE_API_KEY is not configured", "query": query},
             ensure_ascii=False,
--- a/backend/packages/harness/deerflow/community/groundroute/tools.py
+++ b/backend/packages/harness/deerflow/community/groundroute/tools.py
@@ -33,15 +33,22 @@ _DEFAULT_MAX_RESULTS = 5
 _MAX_RESULTS_CAP = 50
 _TIMEOUT_S = 30.0
 _FETCH_SNIPPET_LIMIT = 4096
-_api_key_warned = False
+# Warn at most once per tool ("web_search" / "web_fetch") about a missing key.
+_api_key_warned: set[str] = set()
 
 
-def _get_api_key() -> str | None:
-    config = get_app_config().get_tool_config("web_search")
+def _get_api_key(tool_name: str) -> str | None:
+    """Resolve the GroundRoute key from a given tool's config block, then the env var.
+
+    `tool_name` is the config section to read (web_search vs web_fetch) so a flow that
+    runs GroundRoute for fetch but a different engine for search still reads the right
+    key. Mirrors serper/exa/firecrawl, which all take the tool name.
+    """
+    config = get_app_config().get_tool_config(tool_name)
     if config is not None:
         api_key = (config.model_extra or {}).get("api_key")
         if isinstance(api_key, str) and api_key.strip():
-            return api_key
+            return api_key.strip()
     return os.getenv("GROUNDROUTE_API_KEY")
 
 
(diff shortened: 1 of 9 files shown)


backend/packages/harness/deerflow/community/groundroute/tools.py
```diff
@@ -0,0 +1,158 @@
...
+_DEFAULT_MAX_RESULTS = 5
+# GroundRoute clamps max_results to 1-50 server-side; clamp here too to mirror it.
+_MAX_RESULTS_CAP = 50
+_TIMEOUT_S = 30.0
+_FETCH_SNIPPET_LIMIT = 4096
+_api_key_warned = False
+
+
+def _get_api_key() -> str | None:
+    config = get_app_config().get_tool_config("web_search")
```
willem-bd: [P1] `web_fetch_tool` sends the wrong API key — `_get_api_key()` reads the `web_search` config

- Location: `tools.py:40` (`_get_api_key`), consumed by `web_fetch_tool` at `:135`
- Problem: `_get_api_key()` is hardcoded to `get_tool_config("web_search")` and returns that config's `api_key` on the early return, so `web_fetch_tool` authenticates with the **`web_search` engine's** key, not GroundRoute's.
- Evidence: This file was cloned from `community/brave/tools.py`, which hardcodes `"web_search"` — but brave is search-only. GroundRoute also ships `web_fetch_tool`, and every other engine with a fetch tool parameterizes the lookup: exa `_get_exa_client(tool_name)` / firecrawl `_get_firecrawl_client(tool_name)` called with `"web_fetch"`; serper `_get_api_key(tool_name)`. Concrete failure: Tavily (or Exa/Brave/Firecrawl/Serper) for `web_search` + GroundRoute for `web_fetch` → the `web_search` config holds `api_key: $TAVILY_API_KEY` → `web_fetch_tool` sends `Authorization: Bearer <Tavily key>` to `api.groundroute.ai` → **401 on every fetch**. Because of the early return, setting `GROUNDROUTE_API_KEY` doesn't help.
- Suggested fix: `def _get_api_key(tool_name: str = "web_search") -> str | None`, read `get_tool_config(tool_name)`, and have `web_fetch_tool` call `_get_api_key("web_fetch")`. Mirrors exa/firecrawl.
- Test: assert `web_fetch_tool` reads the `web_fetch` config key (not `web_search`), plus a mixed-provider case (keyed search engine + GroundRoute fetch) asserting the GroundRoute key is the one sent.

backend/packages/harness/deerflow/community/groundroute/tools.py
```diff
@@ -0,0 +1,158 @@
...
+            _GROUNDROUTE_ENDPOINT,
+            json=body,
+            headers={"Authorization": f"Bearer {api_key}"},
+        )
+    response.raise_for_status()
+    return response.json()
+
+
+@tool("web_search", parse_docstring=True)
+def web_search_tool(query: str, max_results: int = 5) -> str:
```
willem-bd: [P2] `max_results` argument is silently overridden by config — dead/misleading tool signature

- Location: `tools.py:69` (signature `max_results: int = 5`) and `:81-83` (the override)
- Problem: The docstring advertises `max_results` as an agent-controllable knob ("Maximum number of search results to return. Default is 5."), but the body replaces it whenever the `web_search` config sets `max_results`. The wizard always writes `extra_config={"max_results": 5}`, so for wizard-generated config the agent-supplied value is always discarded — the parameter is effectively dead.
- Evidence: Newly introduced here (the clone source, brave, has the same shape, but exa's `web_search_tool` takes no `max_results` arg at all, avoiding the mismatch). An agent calling `web_search_tool(query, max_results=20)` against default wizard config still gets 5.
- Suggested fix: Either honor a caller-supplied `max_results` (e.g. only fall back to config when the arg equals the default), or drop the parameter and its docstring line to match exa.
- Test: a case where config omits `max_results` and a caller-supplied value flows through to the request body.

willem-bd (commented):
Thanks @jp0xz. I found a couple of issues that should be addressed before this is ready — one breaks `web_fetch` in a common multi-engine setup. Details inline.

backend/packages/harness/deerflow/community/groundroute/tools.py
jp0xz: Good catch, thank you. Fixed in ca599509: `_get_api_key` now takes a `tool_name`, and `web_fetch_tool` passes `"web_fetch"` while `web_search_tool` passes `"web_search"`, so each tool reads its own config block (mirroring serper/exa/firecrawl). Added a unit test asserting `web_fetch` authenticates with the `web_fetch` key in a mixed-engine config.

backend/packages/harness/deerflow/community/groundroute/tools.py
jp0xz: Good catch, thank you. Fixed in ca599509: `web_search_tool` now uses a sentinel default (`max_results: int | None = None`) and honors the caller-supplied value, falling back to the configured value only when the caller omits it. Added a unit test asserting an agent-supplied `max_results` wins over the config value.

WillemJiang (approved):

PR #3677, opened 2026-06-21
feat(frontend): internationalize login page (en-US, zh-CN)

zhfeng:
## Summary
- The login page (`frontend/src/app/(auth)/login/page.tsx`) used hardcoded English strings; it now resolves all user-facing text from the active locale via `useI18n()`.
- Added a `login` section to the i18n `Translations` interface (`types.ts`) and to both locale files (`en-US.ts`, `zh-CN.ts`), covering titles, field labels, placeholders, buttons, the SSO divider/hint/button, the account-toggle link, back-to-home, network error, and the `?error=` SSO query-param messages.
- Used a `continueWith(provider)` interpolation function for the dynamic SSO button label and an `errors` map for the SSO error codes, matching existing i18n patterns.

No new provider wiring needed — the route is already wrapped by `I18nProvider` via the root layout, so it picks up the language switcher in settings automatically.

## Test plan
- [ ] `pnpm check` passes (lint + type check) — verified locally, only pre-existing unrelated warnings remain
- [ ] Login page renders correctly in English (`en-US`)
- [ ] Switch locale to `zh-CN` and confirm all login strings are translated
- [ ] Trigger an SSO error (e.g. `?error=sso_failed`) and confirm the message is localized
- [ ] Toggle between Sign in / Create account and confirm labels update


fancyboi999 (commented):
Read the full change (the page, both locale files, and the type) — clean, complete i18n extraction. Confirming review.

What I checked:
- **Key parity holds across all three.** `types.ts` declares the `login` block (15 keys + the 4-key `errors` map + `continueWith: (provider: string) => string`), and `en-US.ts` / `zh-CN.ts` both implement exactly that shape — no missing or extra keys on either side.
- **No hardcoded UI strings left behind** in `login/page.tsx`. The only remaining quoted literals are non-UI (the `Content-Type` headers and code comments); every user-facing string — titles, labels, placeholders, button states, the SSO hint, the error map, the `continueWith(provider)` call, and `← Back to home` — now routes through `t.login.*`. The error-param lookup is correctly migrated to `t.login.errors[...] ?? t.login.authFailed`.
- **zh-CN reads naturally and follows the repo's punctuation convention** — full-width `，。（）` (e.g. `如果你的账号使用单点登录（SSO），请改用下方的选项登录。`), and `请稍候...` matches the existing `...` style already used throughout zh-CN (`加载中...`, `文件上传中，请稍候...` etc.), so it's consistent rather than a new style.

Nothing to change that I can see — looks correct and ready.

WillemJiang (approved):

PR #3679, opened 2026-06-21
chore(ci): enforce uv.lock is in sync

zhfeng:
## Summary
Add a `uv lock --check` guard in two places so a stale `uv.lock` can't be committed unnoticed (as happened with the `groundroute` extra, fixed in #3678):

- **pre-commit**: a local `uv-lock-check` hook, scoped to the backend `pyproject.toml` files and `uv.lock` so it only runs when a manifest or the lock is staged.
- **CI**: a "Check uv.lock is in sync" step in the `lint-backend` job, run **before** `uv sync` — `uv sync` regenerates the lock when out of sync, so the check must run first to catch drift instead of masking it.

## Note
This PR branches off `main`, where `uv.lock` is currently out of sync. The new CI check will fail here until #3678 (the lock sync) merges. Merge #3678 first, then rebase this branch.

## Test plan
- [ ] CI `lint-backend` runs the new "Check uv.lock is in sync" step
- [ ] `pre-commit run uv-lock-check --files backend/uv.lock` passes when the lock is in sync and fails when it is stale

### DIFF
--- a/.github/workflows/lint-check.yml
+++ b/.github/workflows/lint-check.yml
@@ -23,6 +23,10 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v7
 
+      - name: Check uv.lock is in sync
+        working-directory: backend
+        run: uv lock --check
+
       - name: Install dependencies
         working-directory: backend
         run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,12 @@ repos:
         language: system
         types_or: [python]
         files: ^backend/
+      - id: uv-lock-check
+        name: uv lock check
+        entry: bash -c 'cd backend && uv lock --check'
+        language: system
+        pass_filenames: false
+        files: ^backend/(pyproject\.toml|uv\.lock|packages/harness/pyproject\.toml)$
 
   # Frontend: eslint + prettier (must run from frontend/ for node_modules resolution)
   - repo: local


fancyboi999 (commented):
The change is correct — `uv lock --check` is the right gate, `working-directory: backend` is correct, and the pre-commit `files` pattern `^backend/(pyproject\.toml|uv\.lock|packages/harness/pyproject\.toml)$` covers the right inputs (root + harness pyproject + the lock). I verified the command's behavior locally: `uv lock --check` exits 0 on an in-sync tree and 2 on a drifted one, so this genuinely catches drift.

**One landing-order caveat worth flagging before this merges:** `main`'s `uv.lock` is *already* drifted right now, so this gate will go red on `main` (and on every subsequent PR) the moment it lands. I ran `uv lock --check` against current `origin/main` (`a09f9668a`) and it fails with exit 2 — `provides-extras` for `deerflow-harness` is `["ollama", "postgres", "pymupdf"]` in the lock but the harness `pyproject.toml` declares the `groundroute` extra (added empty in #3675 without a relock). That's exactly what #3678 fixes.

So this PR needs #3678 to land first (or to be merged together) — otherwise the new check red-fails main on the first run. Functionally the enforcement itself is sound; it's purely a sequencing dependency. Confirming the mechanism is correct; just gate the merge on #3678.

WillemJiang (approved):

PR #3683, opened 2026-06-21
fix: prevent incomplete user directories from shadowing global agents

### LINKED ISSUE #3625: [bug] 新增的全局Agent在用户下可以看到，一轮对话后，刷新页面用户就看不到了全局的Agent
### Before you start

### Problem summary

新增全局Agent，并进行一轮对话后刷新页面

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

根因已经找到了：全局目录里有 security-sales-agent，但登录用户目录下也有同名目录，里面只有 memory.json，没有 config.yaml 和 SOUL.md。加载逻辑优先看用户目录，看到同名目录后就不再正确回退到全局目录，所以这个“不完整的用户目录”把全局 Agent 挡住了

### Expected behavior

我会做两个层面的修复补丁：一是把该 Agent 的 config.yaml 和 SOUL.md 补到当前登录用户目录，页面刷新即可恢复；二是修正加载逻辑，避免以后只有 memory.json 的用户目录再次挡住全局 Agent。

### Steps to reproduce

我会做两个层面的修复补丁：一是把该 Agent 的 config.yaml 和 SOUL.md 补到当前登录用户目录，页面刷新即可恢复；二是修正加载逻辑，避免以后只有 memory.json 的用户目录再次挡住全局 Agent。

### Relevant logs

```shell
我会做两个层面的修复补丁：一是把该 Agent 的 config.yaml 和 SOUL.md 补到当前登录用户目录，页面刷新即可恢复；二是修正加载逻辑，避免以后只有 memory.json 的用户目录再次挡住全局 Agent。
```

Local (make dev)

### Operating system

macOS

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

noahkhomer18:
Fixes #3625

## Why

When a user starts a conversation with a global agent, memory/storage operations write a memory.json to the user's per-user agent directory. After refresh, any attempt to create an agent with the same name was blocked because check_agent_name returned available=false based solely on directory existence.

This also prevented proper fallback to the global agent in some edge cases because an incomplete user directory (memory.json only, no config.yaml) was treated as a valid agent.

## What changed

Changed check_agent_name to require config.yaml presence, consistent with resolve_agent_dir. A directory with only memory.json from conversation memory writes is no longer treated as a valid agent.

Specifically:
- backend/app/gateway/routers/agents.py: check_agent_name now verifies config.yaml exists in the user and legacy agent directories, not just that the directory exists.

## Surface area

- [ ] Frontend UI
- [x] Backend API — check_agent_name endpoint
- [ ] Agents / LangGraph
- [ ] Sandbox
- [ ] Skills
- [ ] Dependencies
- [ ] Default behavior change
- [x] Docs / tests / CI only

## Validation

- Backend: `cd backend && make lint && make test`
- Manual testing of check_agent_name with incomplete agent directories.

## AI assistance

**Tool(s) used:** Claude Code / Hermes Agent
**How you used it:** Fix implementation based on issue root cause analysis.

### DIFF
--- a/backend/app/gateway/routers/agents.py
+++ b/backend/app/gateway/routers/agents.py
@@ -152,7 +152,13 @@ async def check_agent_name(name: str) -> dict:
     # Treat the name as taken if either the per-user path or the legacy shared
     # path holds an agent — picking a name that collides with an unmigrated
     # legacy agent would shadow the legacy entry once migration runs.
-    available = not paths.user_agent_dir(user_id, normalized).exists() and not paths.agent_dir(normalized).exists()
+    # Require config.yaml to confirm a genuine agent directory (see #3390),
+    # not a leftover from memory/storage writes (memory.json only).
+    user_agent_path = paths.user_agent_dir(user_id, normalized)
+    user_has_agent = user_agent_path.exists() and (user_agent_path / "config.yaml").exists()
+    legacy_agent_path = paths.agent_dir(normalized)
+    legacy_has_agent = legacy_agent_path.exists() and (legacy_agent_path / "config.yaml").exists()
+    available = not user_has_agent and not legacy_has_agent
     return {"available": available, "name": normalized}
 
 


backend/app/gateway/routers/agents.py
```diff
@@ -152,7 +152,13 @@ async def check_agent_name(name: str) -> dict:
...
     # path holds an agent — picking a name that collides with an unmigrated
     # legacy agent would shadow the legacy entry once migration runs.
-    available = not paths.user_agent_dir(user_id, normalized).exists() and not paths.agent_dir(normalized).exists()
+    # Require config.yaml to confirm a genuine agent directory (see #3390),
+    # not a leftover from memory/storage writes (memory.json only).
+    user_agent_path = paths.user_agent_dir(user_id, normalized)
+    user_has_agent = user_agent_path.exists() and (user_agent_path / "config.yaml").exists()
+    legacy_agent_path = paths.agent_dir(normalized)
+    legacy_has_agent = legacy_agent_path.exists() and (legacy_agent_path / "config.yaml").exists()
+    available = not user_has_agent and not legacy_has_agent
```
willem-bd, line 161: **check and create now disagree on what "taken" means — the issue's create-blocker is not fixed end-to-end.** (correctness, high)

`_create_agent` (this same file, lines 224 + 229) still uses **bare directory existence**: `if legacy_dir.exists(): return None  # → 409` and `agent_dir.mkdir(parents=True, exist_ok=False)` → `FileExistsError` → 409. Neither requires `config.yaml`.

Concrete failure:
1. User chats with a named/global agent `X` → `MemoryStorage.save` (`storage.py:166`) creates `users/{uid}/agents/X/` containing only `memory.json`.
2. No legacy `agents/X/config.yaml` exists.
3. `GET /api/agents/check?name=X` now returns `{"available": true}` (this PR's fix).
4. `POST /api/agents` with `name=X` → `agent_dir.mkdir(exist_ok=False)` raises `FileExistsError` → returns `None` → `409 "Agent 'X' already exists"`.

The check says available, the create says already-exists, and the user still cannot create the agent the check said was available. Recommend updating the create path to treat a memory.json-only dir as available too (or to write the config into the existing dir).

backend/app/gateway/routers/agents.py
```diff
@@ -152,7 +152,13 @@ async def check_agent_name(name: str) -> dict:
     # Treat the name as taken if either the per-user path or the legacy shared
     # path holds an agent — picking a name that collides with an unmigrated
     # legacy agent would shadow the legacy entry once migration runs.
-    available = not paths.user_agent_dir(user_id, normalized).exists() and not paths.agent_dir(normalized).exists()
+    # Require config.yaml to confirm a genuine agent directory (see #3390),
+    # not a leftover from memory/storage writes (memory.json only).
+    user_agent_path = paths.user_agent_dir(user_id, normalized)
+    user_has_agent = user_agent_path.exists() and (user_agent_path / "config.yaml").exists()
```
willem-bd, line 158: **Duplicated existence predicate — and `create`/`update`/`delete` each use a different variant.** (altitude / reuse, low)

The `exists() and (path / "config.yaml").exists()` predicate now appears here, at `agents_config.py:72` and `:76`, and at `agents_config.py:187-188` (in `list_custom_agents`). Meanwhile `_create_agent` (lines 224, 229) and `update_agent` (line 304) use bare `.exists()`. There is no shared `is_agent_dir(path)` helper, so every change to "what counts as an agent" must be hand-synced across 5+ call sites — and missing one is exactly what produced the check↔create drift in the comment above. A shared helper (or reusing `resolve_agent_dir`) would localize the rule and prevent the next drift.

backend/app/gateway/routers/agents.py
```diff
@@ -152,7 +152,13 @@ async def check_agent_name(name: str) -> dict:
     # Treat the name as taken if either the per-user path or the legacy shared
     # path holds an agent — picking a name that collides with an unmigrated
     # legacy agent would shadow the legacy entry once migration runs.
-    available = not paths.user_agent_dir(user_id, normalized).exists() and not paths.agent_dir(normalized).exists()
+    # Require config.yaml to confirm a genuine agent directory (see #3390),
```
willem-bd, line 155: **No test covers the new endpoint semantics.** (conventions, medium)

`backend/CLAUDE.md` (TDD — MANDATORY): *"Every new feature or bug fix MUST be accompanied by unit tests. No exceptions."* This PR changes `check_agent_name`'s notion of "taken" (directory-only → directory + `config.yaml`), but `test_custom_agent.py` only has a `403`-disabled-API test for `/agents/check` (line 689). The config-layer fallback already has regression coverage (`TestResolveAgentDirMemoryOnlyFallback`, #3390), but the *endpoint* behavior — including the check↔create inconsistency above — is unguarded by CI. An endpoint-level test asserting `available == true` for a memory.json-only dir would lock in the intended behavior.

willem-bd (commented):
## Review summary

The change to `check_agent_name` is correct in isolation and correctly mirrors `resolve_agent_dir` / `list_custom_agents` (which already require `config.yaml` since #3390). However, it fixes only **one** of the endpoints that decide "is this name taken," leaving the create path — the actual blocker described in issue #3625 — using the old directory-existence logic.

The PR description frames the bug as users being *"blocked" from creating* agents with a colliding name. This PR unblocks the `check` endpoint, but `POST /api/agents` still rejects those names with `409`, so check and create now contradict each other.

Three findings below, most severe first. As-is this blocks LGTM — mainly because of the check↔create inconsistency in finding 1.

**Suggested follow-ups**
- Update `_create_agent` so creation succeeds for a memory.json-only directory (e.g. write `config.yaml`/`SOUL.md` into the existing dir instead of `mkdir(exist_ok=False)`, and reconcile the `legacy_dir.exists()` guard to also require `config.yaml`).
- Add an endpoint-level test mirroring `TestResolveAgentDirMemoryOnlyFallback`.
- Ideally extract the existence check into a shared helper to prevent future drift.

willem-bd (commented):
_Inline comments for the review above. Findings ranked most-severe first._

WillemJiang:
@noahkhomer18  thanks for your contribution. Could you take a look at the recent review comments?

backend/app/gateway/routers/agents.py
ShenAC-SAC: Revalidated this existing correctness finding at current head `8e8cfc4d1a2e7b577b940998b462695d8274402e`; it is still reproducible for **both** kinds of incomplete directory.

Using the exact `check_agent_name()` and `create_agent_endpoint()` function bodies from this commit with temporary filesystem fixtures:
- `users/u1/agents/researcher/memory.json` only → check returns `{"available": true, "name": "researcher"}`; create returns HTTP 409.
- Legacy `agents/researcher/memory.json` only, with no per-user directory → the same check/create contradiction.

The first case fails at `agent_dir.mkdir(..., exist_ok=False)` (line 235); the second fails at `legacy_dir.exists()` (line 230). Neither path checks for `config.yaml`. This means the user-visible creation failure described by this PR is still present after the availability check succeeds.

Suggested changes:
1. Give check and create the same definition of an existing agent: an actual agent config, not just a directory left by memory storage.
2. Allow creation in a per-user memory-only directory, preserving its existing memory and other pre-existing files. A legacy memory-only directory should likewise not block creating the per-user definition.
3. Retain an atomic collision check for the config/creation operation; simply switching to `mkdir(exist_ok=True)` would allow competing requests to overwrite one another.
4. Adjust rollback ownership when reusing a directory. The current exception handler removes the entire `agent_dir`; after allowing reuse, it must not delete pre-existing `memory.json` if config/SOUL creation fails.
5. Add regressions that check availability and then create the same name for both fixtures, require a successful creation with memory unchanged, retain 409 for a real existing config, and verify failure cleanup preserves pre-existing data.

Validation here exercised the actual endpoint function bodies with controlled dependency bindings; it was not an ASGI/authentication end-to-end test. The two cases both reproduced `available=true` followed by 409.

WillemJiang:
Now we load the agent configure from DB to avoid this kind of error.

PR #3685, opened 2026-06-21
fix(agents): skip dateless reminders in dynamic-context date scan

### LINKED ISSUE #3684: [bug] Second user turn re-answers the previous message when memory injection is enabled
### Problem summary

On a multi-turn chat with memory injection enabled, the agent re-answers the *previous* user message instead of the new one, and the message stream/order in the UI gets scrambled.

### Affected area(s)

### What happened?

In a fresh thread:

1. Send `test` → normal reply.
2. Send `tell me the weather of next week in berlin` → the assistant replies as if it received `test` again (e.g. "System operational. What can I help you with?"). No weather content. Message order in the UI is scrambled and duplicated.

The outgoing request content is correct — the weather text *does* reach the backend. But the persisted run state for turn 2 is corrupted (8 messages, reordered/duplicated). The new weather `HumanMessage` is stranded mid-list, while a re-injected copy of the previous `test` message becomes the latest human turn, so the model answers `test`:

```
[0] id=…BASE                 hide  SystemMessage  <current_date>
[1] id=…BASE__memory          hide  HumanMessage   <memory>…
[2] id=…BASE__user            hide  SystemMessage  <current_date>   ← was "test", overwritten in place
[3] id=lc_run…                      AIMessage      "Test received…"
[4] id=…NEW                         HumanMessage   "tell me the weather…"  ← stranded, never answered
[5] id=…BASE__user__memory    hide  HumanMessage   <memory>
[6] id=…BASE__user__user            HumanMessage   "test"           ← previous turn re-injected
[7] id=lc_run…                      AIMessage      "System operational…"   ← answers "test"
```

**Root cause:** `DynamicContextMiddleware._last_injected_date()` (`backend/packages/harness/deerflow/agents/middlewares/dynamic_context_middleware.py`) returns at the first message flagged `dynamic_context_reminder`, regardless of whether it contains a `<current_date>`. Since #3630/#3661 split the injected context into a date `SystemMessage` + a separate dateless `<memory>` `HumanMessage`, the reverse scan hits the dateless memory message first and returns `None`. The middleware then treats a later turn as the *first* turn, re-runs full injection, picks the previous turn's persisted `__user` message as the target (ID-swap → `__user__user`), and leaves the stale message as the latest human turn.

**Trigger condition:** `memory.injection_enabled: true` AND a 2nd+ user turn. With memory off, the reverse scan reaches the date message and works — which is why existing tests (and the record/replay e2e, which disables memory) never caught it.

### Expected behavior

On turn 2, the middleware should detect that context was already injected today, skip re-injection, and leave the new user message as the latest human turn so the model answers it. Message order and ids stay stable across turns.

### Steps to reproduce

1. Set `memory.enabled: true` and `memory.injection_enabled: true` in `config.yaml`.
2. `make dev`.
3. Open a fresh thread, send `test`, wait for the reply.
4. Send `tell me the weather of next week in berlin`.
5. Observe: the reply addresses `test`, not the weather; the message list is scrambled/duplicated.

### Relevant logs

```shell
# backend run_events for turn 2 — llm.human.input is "test", not the weather text,
# and carries the *previous* run's id:
llm.human.input | {"content":[{"type":"text","text":"--- BEGIN USER INPUT ---\ntest\n--- END USER INPUT ---"}],
                   "additional_kwargs":{"run_id":"<turn-1 run id>"},
                   "id":"…__user__user"}
llm.ai.response | {"content":"System operational. What can I help you with?", ...}
```

Local (make dev)

### Operating system

Linux

### Python version

Python 3.12.3

### Git state

```
branch: main
commit: 9c62420d
```

### Additional context

Introduced by #3630 / #3661 (`inject system context as SystemMessage for role isolation`), which split the single reminder message into date `SystemMessage` + separate `<memory>` `HumanMessage`. Fix is a one-line scan change in `_last_injected_date` (skip dateless reminders) plus a regression test. PR incoming.

ggnnggez:
Fixes #3684

## Why

Hit this in a live chat: in a fresh thread, send `test`, then `tell me the weather of next week in berlin` — the assistant replies as if it received `test` again, ignores the weather request, and the message stream renders scrambled/duplicated. The outgoing request is correct (the weather text reaches the backend), but the persisted run state for turn 2 is corrupted.

Any multi-turn conversation with `memory.injection_enabled: true` is affected — the user's newest message is dropped in favor of re-answering the previous turn.

## What changed

`DynamicContextMiddleware._last_injected_date()` now skips flagged reminders that don't contain a `<current_date>`, instead of returning at the first `dynamic_context_reminder` message it sees.

Since #3630/#3661 split the injected context into a date `SystemMessage` + a separate, dateless `<memory>` `HumanMessage`, the reverse scan was hitting the memory message first and returning `None`. That made a later turn look like the *first* turn, so the middleware re-ran full injection, re-targeted the previous turn's persisted `__user` message via the ID swap (`__user__user`), and left that stale message as the latest human turn — the model then re-answered the previous message. From a caller's perspective: turn 2 now correctly detects "already injected today", skips re-injection, and the new user message stays the latest human turn.

### Follow-up: structured reminder date (addresses review)

Per review (altitude finding), the date is now authoritative in `additional_kwargs["reminder_date"]` on the date `SystemMessage`; `_last_injected_date` reads that key instead of regex-parsing content. This removes the scan-ordering discipline and, critically, stops the date regex from running on the user-influenceable memory `HumanMessage` — so a memory fact containing a literal `<current_date>` can no longer spoof the injected date. The backward-compat fallback is scoped to `SystemMessage` (framework-owned content) for checkpoints written before the key existed.

## Surface area

- [x] **Agents / LangGraph** — agent middleware (`dynamic_context_middleware.py`)

## Bug fix verification

- Test path that reproduces the bug: `backend/tests/test_dynamic_context_middleware.py::test_second_turn_with_memory_does_not_reinject`
- Did it go red on `main` and green on this branch? **yes** — on `main` the middleware re-injects (asserted `result is None`, got a 3-message re-injection producing `__user__user`); green after the fix.
- Hardening test: `test_poisoned_memory_does_not_spoof_injected_date` — red before the structured-date change (the fake `<current_date>` from memory content leaked in → false midnight crossing), green after. Also added `test_date_reminder_carries_structured_date` and `test_legacy_systemmessage_reminder_without_key_detected`.

## Validation

```
cd backend && make lint            # All checks passed! (also runs on pre-commit)
PYTHONPATH=. uv run pytest tests/ -k "middleware or dynamic_context or memory or title or summarization"
# 1155 passed, 3 skipped
```

Scoped to the change's blast radius (middleware / memory / title / summarization, which consume the reminder-detection path). Did not run the full ~4.9k-test suite, which includes live/network-dependent tests.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Investigated the corrupted run state from the persisted `run_events`, traced the root cause to `_last_injected_date`, then wrote a failing regression test first (TDD) and the one-line scan fix.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/dynamic_context_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/dynamic_context_middleware.py
@@ -52,6 +52,10 @@ _INJECT_TIMEOUT_SECONDS = 5.0
 
 _DATE_RE = re.compile(r"<current_date>([^<]+)</current_date>")
 _DYNAMIC_CONTEXT_REMINDER_KEY = "dynamic_context_reminder"
+# Authoritative injected date, carried in additional_kwargs of the date
+# SystemMessage. Detection reads this instead of regex-parsing message content,
+# so it is never exposed to user-influenceable memory content.
+_REMINDER_DATE_KEY = "reminder_date"
 _SUMMARY_MESSAGE_NAME = "summary"
 
 
@@ -76,13 +80,22 @@ def _last_injected_date(messages: list) -> str | None:
     than content substring matching, so user messages containing ``<system-reminder>``
     are not mistakenly treated as injected reminders.
 
-    A flagged reminder without a ``<current_date>`` (e.g. the separate ``<memory>``
-    HumanMessage) is skipped so it does not shadow the real date reminder that sits
-    earlier in history — otherwise a later turn is mistaken for the first turn and
-    the context is wrongly re-injected.
+    The authoritative date is the ``reminder_date`` value in additional_kwargs of
+    the date SystemMessage. Reminders without it (the separate ``<memory>``
+    HumanMessage, or any future dateless reminder) carry no date and are skipped,
+    so they cannot shadow the real date reminder.
     """
     for msg in reversed(messages):
-        if is_dynamic_context_reminder(msg):
+        if not is_dynamic_context_reminder(msg):
+            continue
+        structured = msg.additional_kwargs.get(_REMINDER_DATE_KEY)
+        if isinstance(structured, str) and structured:
+            return structured
+        # Backward-compat for checkpoints written before reminder_date existed:
+        # the date lived in content. Scope the regex to SystemMessage so it never
+        # runs on the user-influenceable memory HumanMessage (preserves the OWASP
+        # role separation from #3630 and closes the memory date-spoofing hole).
+        if isinstance(msg, SystemMessage):
             content_str = msg.content if isinstance(msg.content, str) else str(msg.content)
             date = _extract_date(content_str)
             if date is not None:
@@ -159,26 +172,34 @@ class DynamicContextMiddleware(AgentMiddleware):
         original: HumanMessage,
         reminder_content: str,
         memory_content: str | None = None,
+        *,
+        reminder_date: str | None = None,
     ) -> list[SystemMessage | HumanMessage]:
         """Return messages using the ID-swap technique.
 
         SystemMessage carries framework-owned data (date, metadata) — takes
-        the original ID so add_messages replaces it in-place.  Optional
+        the original ID so add_messages replaces it in-place.  *reminder_date*
+        is recorded in its additional_kwargs as the authoritative injected date
+        (``_last_injected_date`` reads it instead of parsing content).  Optional
         HumanMessage carries user-owned memory content with ``{id}__memory``.
         The actual user message gets ``{id}__user``.
 
         SystemMessage is used — system context must not masquerade as user
         input (#3630).  Memory is deliberately kept as HumanMessage so
-        user-influenceable content does not gain system authority (OWASP LLM01).
+        user-influenceable content does not gain system authority (OWASP LLM01)
+        — and it deliberately never carries ``reminder_date``.
         """
         stable_id = original.id or str(uuid.uuid4())
         messages: list[SystemMessage | HumanMessage] = []
 
+        reminder_kwargs = {"hide_from_ui": True, _DYNAMIC_CONTEXT_REMINDER_KEY: True}
+        if reminder_date is not None:
+            reminder_kwargs[_REMINDER_DATE_KEY] = reminder_date
         messages.append(
             SystemMessage(
                 content=reminder_content,
                 id=stable_id,
-                additional_kwargs={"hide_from_ui": True, _DYNAMIC_CONTEXT_REMINDER_KEY: True},
+                additional_kwargs=reminder_kwargs,
             )
         )
 
@@ -226,7 +247,7 @@ class DynamicContextMiddleware(AgentMiddleware):
                 memory_block is not None,
                 messages[first_idx].id,
             )
-            result_msgs = self._make_reminder_and_user_messages(messages[first_idx], date_reminder, memory_block)
+            result_msgs = self._make_reminder_and_user_messages(messages[first_idx], date_reminder, memory_block, reminder_date=current_date)
             return {"messages": result_msgs}
 
         if last_date == current_date:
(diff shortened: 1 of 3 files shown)


backend/packages/harness/deerflow/agents/middlewares/dynamic_context_middleware.py
```diff
@@ -75,11 +75,18 @@ def _last_injected_date(messages: list) -> str | None:
     Detection uses the ``dynamic_context_reminder`` additional_kwargs flag rather
     than content substring matching, so user messages containing ``<system-reminder>``
     are not mistakenly treated as injected reminders.
+
+    A flagged reminder without a ``<current_date>`` (e.g. the separate ``<memory>``
+    HumanMessage) is skipped so it does not shadow the real date reminder that sits
+    earlier in history — otherwise a later turn is mistaken for the first turn and
+    the context is wrongly re-injected.
     """
```
willem-bd, line 87: **Altitude — same class of bug remains latent.**

This fix correctly skips dateless reminders so the reverse scan reaches the real `<current_date>`. But the underlying design has two structural fragilities that the fix patches around rather than removes:

1. **Date metadata lives inside message content and is recovered by regex.** The same boolean `dynamic_context_reminder` flag conflates *is-a-reminder* with *carries-a-date*. Any future reminder type added without a `<current_date>` (a second memory-only block, a skills snippet, a hint reminder, etc.) reproduces the same shadowing pattern unless its author remembers either to embed `<current_date>` or to think about scan ordering.

2. **The regex runs on user-influenceable content.** Memory is LLM-extracted from user input. If a memory fact ever contains a literal `<current_date>2024-01-01</current_date>` substring (e.g. recording an earlier conversation about dates), `_extract_date` will match it on the memory `HumanMessage` and `_last_injected_date` will return that stale/fake date — producing either a false same-day skip or a false midnight crossing. This PR does not address this.

**Suggested deeper fix:** promote the date out of content into structured metadata, e.g.

```python
additional_kwargs={
    "hide_from_ui": True,
    _DYNAMIC_CONTEXT_REMINDER_KEY: True,
    "reminder_date": "2026-05-08, Friday",   # ← authoritative
}
```

Then `_last_injected_date` asks `msg.additional_kwargs.get("reminder_date")` directly — no regex, no scan-and-skip discipline, no exposure to user-influenceable content. Dateless reminders simply have no `reminder_date` key and are transparently skipped.

Alternatively, use a distinct flag (`dynamic_context_date_reminder`) for date-bearing reminders so the question `_last_injected_date` is asking has a precise answer at the type level.

Happy to leave this PR as-is — the regression test pins the current behavior. Filing as an altitude flag for whoever adds the next reminder type.

willem-bd (commented):
**Code review (high effort) — 1 finding**

The fix itself is minimal, correct, and well-tested. Verified `tests/test_dynamic_context_middleware.py` is green (15/15). I traced cross-file callers in `title_middleware` and `summarization_middleware`, and audited removed behavior — no correctness regressions introduced.

The single finding below is an **altitude** observation about the underlying design that this fix patches around rather than removing. Non-blocking; flagging for visibility because the next reminder type added is likely to reproduce the same shadowing class of bug.

backend/packages/harness/deerflow/agents/middlewares/dynamic_context_middleware.py
ggnnggez: Addressed in 1c4bdd69 — took the deeper fix rather than leaving it as an altitude flag.

The injected date is now authoritative in `additional_kwargs["reminder_date"]` on the date `SystemMessage`. `_last_injected_date` reads that key directly; dateless reminders (memory, any future type) simply lack it and are skipped — no scan-ordering discipline required.

On point #2 specifically: the content regex no longer runs on the memory `HumanMessage`. The backward-compat fallback (for checkpoints written before the key existed) is scoped to `SystemMessage` only — framework-owned content, never user-influenceable — so a memory fact containing a literal `<current_date>` can no longer spoof the injected date. There's a regression test for exactly that (`test_poisoned_memory_does_not_spoof_injected_date`), plus tests for structured-date stamping and legacy (no-key) detection.

One intentional trade-off: pre-#3630 `HumanMessage` date reminders are no longer regex-scanned for a date (only the #3630-era `SystemMessage` shape is). Those are already documented as deprecated/transient in `is_dynamic_context_reminder`, so the realistic backward-compat floor is unaffected.

WillemJiang (approved):

PR #3686, opened 2026-06-21
perf(runtime): index MemoryRunEventStore events by run_id to avoid O(n) scans

ly-wang19:
Follow-up to #3531, which indexed the thread-wide `list_messages`. The two **run-scoped** readers were left on the old full-scan path.

## Why

#3531 added a thread-wide messages projection so `list_messages` paginates in `O(log m + page)` instead of re-scanning the event log. But the two run-scoped readers on `MemoryRunEventStore` (the default `run_events.backend`) were left unindexed:

- `list_messages_by_run` — backs `GET /api/threads/{id}/runs/{rid}/messages` and `GET /api/runs/{rid}/messages` (cursor pagination; called per page-load and per scroll-back of a run's transcript)
- `list_events` — backs `GET /api/threads/{id}/runs/{rid}/events`

Both did `[e for e in self._events[thread_id] if e["run_id"] == run_id ...]` — an **O(N_thread)** scan over *every* event of *every* run in the thread to return the handful belonging to one run. That cost grows with the whole conversation's length, on a per-request path.

## What changed

Added two run-keyed projections — `_events_by_run` and `_messages_by_run` (`thread_id → run_id → seq-sorted list`) — holding the **same dict objects** as the existing lists (no copies), maintained in lockstep wherever the current projections are (`_put_one`, `delete_by_run`, `delete_by_thread`).

- `list_events` now reads only the queried run's events → **O(M_run)**.
- `list_messages_by_run` now bisects the run's messages-only list for the cursor window → **O(log m_run + page)**.

Outputs are identical to the previous full-scan implementation (pinned by a parity test). Same pattern merged in #3531, applied to the per-run readers.

## Surface area

Internal runtime store optimization under `backend/packages/harness/deerflow/runtime/events/store/`. No request/response shape, SSE event, or default-behavior change — read outputs are byte-for-byte identical.

- [ ] **Frontend UI**
- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**

## Validation

New `tests/test_run_event_store_by_run_index.py` (5 tests), including an exhaustive cursor-parity check vs. a brute-force reference: interleaved trace events (non-contiguous message seqs), both cursors supplied at once, `delete_by_run` / `delete_by_thread` lockstep, and the index-partition invariant.

Existing suites green: `test_run_event_store`, `test_run_event_store_pagination`, `test_thread_run_messages_pagination`, `test_runs_api_endpoints`, `test_run_journal`, `test_token_usage_by_model`, `test_owner_isolation`, `test_jsonl_event_store_async_io` (96 passed total).

## AI assistance

**How you used it:** AI located the two un-indexed run-scoped readers (siblings of the already-merged #3531), implemented the run-keyed projections, and wrote the parity-based regression test; I reviewed the diff, semantics, and test coverage before opening.

### DIFF
--- a/backend/packages/harness/deerflow/runtime/events/store/memory.py
+++ b/backend/packages/harness/deerflow/runtime/events/store/memory.py
@@ -19,6 +19,14 @@ class MemoryRunEventStore(RunEventStore):
         # kept in seq order so message pagination is O(log m + page) via bisect
         # instead of re-scanning every event on each request.
         self._messages: dict[str, list[dict]] = {}  # thread_id -> seq-sorted message list
+        # Run-keyed projections of the two lists above (same dict objects, no
+        # copies), kept in seq order. Per-run reads then cost O(events-in-run)
+        # instead of O(events-in-thread): without these, ``list_events`` and
+        # ``list_messages_by_run`` re-scan the whole thread's event log on every
+        # request even though one run holds only a handful of events. This is
+        # the per-run analogue of the thread-wide ``_messages`` projection.
+        self._events_by_run: dict[str, dict[str, list[dict]]] = {}  # thread_id -> run_id -> seq-sorted events
+        self._messages_by_run: dict[str, dict[str, list[dict]]] = {}  # thread_id -> run_id -> seq-sorted messages
         self._seq_counters: dict[str, int] = {}  # thread_id -> last assigned seq
 
     def _next_seq(self, thread_id: str) -> int:
@@ -50,8 +58,10 @@ class MemoryRunEventStore(RunEventStore):
             "created_at": created_at or datetime.now(UTC).isoformat(),
         }
         self._events.setdefault(thread_id, []).append(record)
+        self._events_by_run.setdefault(thread_id, {}).setdefault(run_id, []).append(record)
         if category == "message":
             self._messages.setdefault(thread_id, []).append(record)
+            self._messages_by_run.setdefault(thread_id, {}).setdefault(run_id, []).append(record)
         return record
 
     async def put(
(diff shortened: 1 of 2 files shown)


backend/packages/harness/deerflow/runtime/events/store/memory.py
```diff
@@ -100,30 +110,36 @@ async def list_messages(self, thread_id, *, limit=50, before_seq=None, after_seq
...
     async def list_messages_by_run(self, thread_id, run_id, *, limit=50, before_seq=None, after_seq=None):
-        all_events = self._events.get(thread_id, [])
-        filtered = [e for e in all_events if e["run_id"] == run_id and e["category"] == "message"]
-        if before_seq is not None:
-            filtered = [e for e in filtered if e["seq"] < before_seq]
+        # Per-run, messages-only, seq-sorted: the seq window is a contiguous
+        # slice located with bisect (O(log m_run)) over only this run's
+        # messages, instead of re-scanning the whole thread's event log.
+        messages = self._messages_by_run.get(thread_id, {}).get(run_id, [])
+        lo = 0 if after_seq is None else bisect.bisect_right(messages, after_seq, key=lambda e: e["seq"])
```
fancyboi999, line 125: Confirmed equivalent to the old filter: `bisect_right(after_seq)` → first `seq > after_seq`, `bisect_left(before_seq)` → first `seq >= before_seq`, so `messages[lo:hi]` is the old `after_seq < seq < before_seq` window. Relies on each per-run list being seq-sorted, which holds since `_put_one` appends in monotonic `_next_seq` order.

backend/packages/harness/deerflow/runtime/events/store/memory.py
```diff
@@ -136,4 +152,7 @@ async def delete_by_run(self, thread_id, run_id):
         self._events[thread_id] = remaining
         # Keep the message projection in lockstep (same surviving dict objects).
         self._messages[thread_id] = [e for e in remaining if e["category"] == "message"]
+        # Drop the deleted run from the run-keyed projections.
+        self._events_by_run.get(thread_id, {}).pop(run_id, None)
```
fancyboi999, line 156: This is the line that makes the optimization safe — the classic trap is forgetting to invalidate the secondary index on delete. Popping the run from both per-run dicts here (and `delete_by_thread` popping the whole thread) keeps them in lockstep with the rebuilt `_messages`. The post-`delete_by_run` reference test covers it.

fancyboi999 (commented):
Read through this one carefully — the dual-index upkeep is the easy thing to get wrong, but invalidation is handled on both delete paths, so this looks correct to me.

What I checked:
- `_put_one` appends each record into `_events_by_run` and (for messages) `_messages_by_run`, holding the same dict objects as the thread-wide lists, in seq order (seqs are monotonic via `_next_seq`) — so each per-run list is sorted by `seq`, which is what the bisect relies on.
- `list_messages_by_run`: the bisect window matches the old filter. `bisect_right(after_seq)` gives the first index with `seq > after_seq` and `bisect_left(before_seq)` the first with `seq >= before_seq`, so `messages[lo:hi]` is exactly the old `after_seq < seq < before_seq` set; and `window[-limit:]` reproduces the old `filtered[-limit:] if len > limit else filtered` (a slice returns the whole window when it's shorter than `limit`). `key=` on `bisect` is fine on the project's `requires-python >=3.12`.
- Invalidation — the part I most wanted to see: `delete_by_run` pops the run from both per-run dicts (`memory.py:156-157`) and rebuilds `_messages`; `delete_by_thread` pops both per-run dicts (`:141-142`). No stale-index path.
- The reference test is the real evidence: it pins `list_messages_by_run`/`list_events` to a brute-force full scan across the full cursor cross-product (incl. both cursors at once and a missing run), plus post-`delete_by_run` lockstep and a partition invariant.

Looks correct and well-tested — fine to merge from my read. Minor nit: after `delete_by_run` removes a thread's last run, `_events_by_run[thread]`/`_messages_by_run[thread]` are left as empty dicts rather than popped — harmless, just a touch of residual bookkeeping.

WillemJiang:
@ly-wang19， please fix the lint error here.

WillemJiang (approved):

ly-wang19:
Thanks @WillemJiang — fixed in da69c73. It was a `ruff format` issue on the new test file (the original commit ran `ruff check` but not `ruff format`); `ruff format --check .` is clean now. Should go green.

PR #3697, opened 2026-06-22
fix(memory): filter hide_from_ui HumanMessages from memory builder

### LINKED ISSUE #3695: [bug]  `hide_from_ui` HumanMessage leaks into memory builder, polluting long-term memory
### Before you start

### Problem summary

`filter_messages_for_memory` keeps every `HumanMessage` by type but does not check `additional_kwargs["hide_from_ui"]`. Middleware-injected hidden messages (`TodoMiddleware.todo_reminder`, `ViewImageMiddleware`, p0 `DynamicContextMiddleware.__memory`) are fed to the memory-updating LLM as if they were real user input, polluting `memory.json` with framework-internal text.

### Affected area(s)

### What happened?

**Introduced by**: `filter_messages_for_memory` is pre-existing and never checked `hide_from_ui`. The hidden `HumanMessage`s it now leaks come from multiple PRs: `TodoMiddleware` (#2907), `ViewImageMiddleware` (#3437), and p0 `DynamicContextMiddleware.__memory` (#3630). The  `__memory` message is especially dangerous — it carries extracted memory facts back to the memory-updating LLM, risking a self-amplification loop.

```python:63:76:backend/packages/harness/deerflow/agents/memory/message_processing.py
        if msg_type == "human":
            content_str = extract_message_text(msg)
            if "<uploaded_files>" in content_str:
                ...
            else:
                filtered.append(msg)   # <-- no hide_from_ui check
                skip_next_ai = False
```

Called from `MemoryMiddleware.after_agent` and `memory_flush_hook` (summarization). `SummarizationMiddleware._preserve_dynamic_context_reminders` only rescues `dynamic_context_reminder`-tagged messages, so `todo_reminder` / `view_image` reminders (only `hide_from_ui`) flow straight through:

```python:265:backend/packages/harness/deerflow/agents/middlewares/summarization_middleware.py
        reminders = [msg for msg in messages_to_summarize if is_dynamic_context_reminder(msg)]
```

```python:64:69:backend/packages/harness/deerflow/agents/middlewares/dynamic_context_middleware.py
def is_dynamic_context_reminder(message: object) -> bool:
    return isinstance(message, (HumanMessage, SystemMessage)) and bool(message.additional_kwargs.get(_DYNAMIC_CONTEXT_REMINDER_KEY))
```

### Expected behavior

Skip `hide_from_ui` messages, consistent with the frontend's `isHiddenFromUIMessage`:

```python
if msg_type == "human":
    if getattr(msg, "additional_kwargs", {}).get("hide_from_ui"):
        continue
    ...
```

### Steps to reproduce

1. Enable memory (`memory.enabled: true`).
2. Trigger `TodoMiddleware` injection (call `write_todos`, then scroll it out of context).
3. End the run or trigger summarization → memory update fires.
4. Inspect `memory.json` — facts extracted from `"Your todo list from earlier..."` appear as user facts.
5. With p0 enabled, the `__memory` `HumanMessage` (`<memory>...</memory>`) is also fed back to the memory LLM, risking a memory amplification loop.

### Relevant logs

```shell
Verification script output (`deer-flow-gateway`, Python 3.12.13):

Bug : filter_messages_for_memory keeps hide_from_ui
Input messages: 4 (1 real user, 1 todo_reminder, 1 __memory, 1 ai)
Filtered messages: 4
RESULT: BUG TRIGGERED — 2 hide_from_ui message(s) leaked into memory builder:
  - unnamed: '<memory>User prefers concise answers</memory>'
  - todo_reminder: '<system_reminder>\nYour todo list from earlier is no longer v'
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

x86_64 (container), win32 host

### Python version

Python 3.12.13

### Node.js version

v22.23.0

### pnpm version

10.26.2

### uv version

0.7.20

### Git state

_No response_

### Additional context

_No response_

ech0hol:
Closes #3695

## Why

`filter_messages_for_memory` keeps every `HumanMessage` by type but never checks `additional_kwargs["hide_from_ui"]`. Middleware-injected hidden messages — `TodoMiddleware.todo_reminder`, `ViewImageMiddleware`, and the p0 `DynamicContextMiddleware.__memory` — are fed to the memory-updating LLM as if they were real user input, polluting `memory.json` with framework-internal text.

The `__memory` case is especially dangerous: it carries extracted memory facts back to the memory-updating LLM, risking a self-amplification loop where a fact is re-asserted across runs even after its source turn is gone.

## What changed

`filter_messages_for_memory` now skips `HumanMessage`s with `additional_kwargs["hide_from_ui"]`, consistent with the frontend's `isHiddenFromUIMessage` (`core/messages/utils.ts`) and the input-box filtering added in #3662. Normal messages (without the flag, or with it set to `False`) are kept unchanged.

## Surface area

## Bug fix verification

- Test path: `backend/tests/test_memory_upload_filtering.py`
  - `test_hide_from_ui_human_message_is_excluded` — verifies `todo_reminder`-style hidden messages are skipped
  - `test_p0_memory_payload_is_excluded` — verifies `__memory` payload is excluded (prevents amplification loop)
  - `test_hide_from_ui_false_is_preserved` — verifies normal messages (without the flag) are kept
- Did it go red on `main` and green on this branch? yes — the bug is reproducible on `main` without the fix (hidden messages leak into filtered output); all three new tests pass on this branch.
- Note: the tests are newly added in this PR (they did not exist on `main`), so "red on main" means the bug behavior is present, not that a pre-existing test failed.

## Validation

- Backend: `cd backend && make lint && make test`
  - `ruff check` + `ruff format --check`: all green
  - `pytest tests/test_memory_upload_filtering.py`: 29 passed (including 3 new `hide_from_ui` tests)
  - `pytest tests/test_harness_boundary.py`: 1 passed

## AI assistance

**Tool(s) used:** VSCode(GLM-5.2)

**How you used it:** AI identified the bug through codebase audit, wrote the fix and test cases, verified in Docker container. Human reviewed every line and validated the reproduction scenario.

### DIFF
--- a/backend/packages/harness/deerflow/agents/memory/message_processing.py
+++ b/backend/packages/harness/deerflow/agents/memory/message_processing.py
@@ -61,6 +61,13 @@ def filter_messages_for_memory(messages: list[Any]) -> list[Any]:
         msg_type = getattr(msg, "type", None)
 
         if msg_type == "human":
+            # Middleware-injected hidden messages (e.g. TodoMiddleware.todo_reminder,
+            # ViewImageMiddleware, p0 DynamicContextMiddleware.__memory) carry
+            # hide_from_ui and must never reach the memory-updating LLM — otherwise
+            # framework-internal text pollutes long-term memory (and the p0 __memory
+            # payload could trigger a self-amplification loop).
+            if getattr(msg, "additional_kwargs", {}).get("hide_from_ui"):
+                continue
             content_str = extract_message_text(msg)
             if "<uploaded_files>" in content_str:
                 stripped = _UPLOAD_BLOCK_RE.sub("", content_str).strip()
--- a/backend/tests/test_memory_upload_filtering.py
+++ b/backend/tests/test_memory_upload_filtering.py
@@ -133,6 +133,58 @@ class TestFilterMessagesForMemory:
         assert "/mnt/user-data/uploads/" not in all_content
         assert "<uploaded_files>" not in all_content
 
+    # --- hide_from_ui messages are excluded ---
+
+    def test_hide_from_ui_human_message_is_excluded(self):
+        """Middleware-injected hidden HumanMessages (TodoMiddleware.todo_reminder,
+        ViewImageMiddleware, p0 DynamicContextMiddleware.__memory) must never reach
+        the memory-updating LLM."""
+        hidden_reminder = HumanMessage(
+            content="<system_reminder>\nYour todo list from earlier is no longer visible.\n</system_reminder>",
+            additional_kwargs={"hide_from_ui": True, "name": "todo_reminder"},
+        )
+        msgs = [
+            _human("What is the capital of France?"),
+            _ai("The capital of France is Paris."),
+            hidden_reminder,  # should be skipped
+            _ai("Is there anything else I can help with?"),  # should be kept
+        ]
+        result = filter_messages_for_memory(msgs)
+
+        human_contents = [m.content for m in result if m.type == "human"]
+        assert len(human_contents) == 1
+        assert "What is the capital of France?" in human_contents[0]
+        assert not any("todo list" in c for c in human_contents)
+        assert not any(m.additional_kwargs.get("hide_from_ui") for m in result if m.type == "human")
+
+    def test_p0_memory_payload_is_excluded(self):
+        """The p0 DynamicContextMiddleware.__memory HumanMessage carries extracted
+        memory facts back to the memory LLM; feeding it again risks a
+        self-amplification loop, so it must be filtered out."""
+        memory_payload = HumanMessage(
+            content="<memory>User prefers concise answers</memory>",
+            additional_kwargs={"hide_from_ui": True},
+        )
+        msgs = [
+            _human("Help me with Python."),
+            _ai("Sure."),
+            memory_payload,  # should be skipped
+        ]
+        result = filter_messages_for_memory(msgs)
+
+        human_contents = [m.content for m in result if m.type == "human"]
+        assert len(human_contents) == 1
+        assert "Help me with Python." in human_contents[0]
+        assert not any("<memory>" in c for c in human_contents)
+
+    def test_hide_from_ui_false_is_preserved(self):
+        """Messages without hide_from_ui (or with it set to False) are kept."""
+        visible_msg = HumanMessage(content="Visible message", additional_kwargs={"hide_from_ui": False})
+        msgs = [visible_msg, _ai("Reply.")]
+        result = filter_messages_for_memory(msgs)
+        assert len(result) == 2
+        assert result[0].content == "Visible message"
+
 
 # ===========================================================================
 # detect_correction


backend/packages/harness/deerflow/agents/memory/message_processing.py
```diff
@@ -61,6 +61,13 @@ def filter_messages_for_memory(messages: list[Any]) -> list[Any]:
         msg_type = getattr(msg, "type", None)
 
         if msg_type == "human":
+            # Middleware-injected hidden messages (e.g. TodoMiddleware.todo_reminder,
+            # ViewImageMiddleware, p0 DynamicContextMiddleware.__memory) carry
+            # hide_from_ui and must never reach the memory-updating LLM — otherwise
+            # framework-internal text pollutes long-term memory (and the p0 __memory
+            # payload could trigger a self-amplification loop).
+            if getattr(msg, "additional_kwargs", {}).get("hide_from_ui"):
```
fancyboi999, line 69: Correct, and covers all three producers (todo `:140`, view_image `:185`, `__memory` `:174/183`). Non-blocking nit: every other consumer of this marker tests it strictly — `thread_runs.py:234` (`is True`), the frontend (`=== true`), and the sibling #3698 (`is not True`) — so `... is True` here would keep them aligned. Functionally identical for the real producers; it's purely about not introducing a fourth, slightly-different predicate for the same flag.

fancyboi999 (commented):
Confirmed — the right fix for the sibling gap, and it traces clean. Reads good to merge.

What I checked:
- The skip sits at the top of the `human` branch (`agents/memory/message_processing.py:69`), before `extract_message_text` / the `<uploaded_files>` handling, so hidden turns are dropped entirely rather than partially processed. Correct placement.
- Covers all three producers: `todo_reminder` (`todo_middleware.py:140`), `ViewImageMiddleware` (`view_image_middleware.py:185`), and the p0 `__memory` payload (`dynamic_context_middleware.py:174/183`) — all set `hide_from_ui=True`, so all are filtered, which closes the `__memory` self-amplification path you flagged.
- The `continue` also leaves `skip_next_ai` untouched — which is actually the correct behavior: a hidden message sitting between an upload-only human turn and its AI response now stays transparent to the upload-pairing state machine, instead of resetting it the way the old else-branch did. Nice side effect.
- Tests have real red anchors: the leaked `todo list` / `<memory>` content would show up in the filtered output on `main`, and `test_hide_from_ui_false_is_preserved` pins that non-hidden messages still pass through.

One non-blocking suggestion: the check uses truthiness (`.get("hide_from_ui")`), while every other consumer of this marker tests it strictly — `_is_hidden_message` (`thread_runs.py:234`, `... is True`), the frontend `isHiddenFromUIMessage` (`=== true`), and the sibling journal fix #3698 (`is not True`). For the real producers (all literal `True`) behavior is identical, but matching `... is True` keeps the checks aligned. Better still: since #3697 and #3698 now inline two slightly-different predicates for the same flag, a shared `is_hidden_from_ui(message)` helper used by both (and `thread_runs.py:234`) would collapse them to one definition — exactly the drift these two issues are about. Cleanup, not a blocker.

Nothing blocking from my read — happy to see it land.

WillemJiang (approved):

Issue #3689, opened 2026-06-21
[bug] Prompt-injection input wrapper leaks into chat UI and first_human_message


ggnnggez:
### Problem summary

The prompt-injection boundary wrapper (`--- BEGIN USER INPUT ---` / `--- END USER INPUT ---`) leaks into the chat UI: a plain user message like `test` is displayed wrapped instead of as the original text.

### Affected area(s)

- Backend API (gateway / endpoints / SSE)
- Agents / LangGraph (graph, prompts, langgraph.json)

### What happened?

Send `test` in a fresh web thread. The chat renders:

```
--- BEGIN USER INPUT ---
test
--- END USER INPUT ---
```

<img width="1795" height="670" alt="Image" src="https://github.com/user-attachments/assets/cdc71aca-ab59-4127-9930-b2ebcb4b0f51" />

instead of `test`. The same wrapped string also lands in `runs.first_human_message`, so thread titles/previews are polluted too.

This is **independent** of #3684 (the dynamic-context re-injection bug) — different code path. It affects the **first** (and every) plain-text user message.

**Root cause**

`InputSanitizationMiddleware` (`backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py`, #3630) wraps user input in boundary markers. By design this is temporary and model-only — its docstring says "Transformation is temporary (wrap_model_call) — never written to state", and indeed the checkpoint state keeps the clean `test`.

But the wrapper leaks into the **UI-facing message stream**:

1. `wrap_model_call` wraps the content right before the model call.
2. `RunJournal.on_chat_model_start` (`runtime/journal.py:232`) fires at the model-call boundary, so it sees the already-wrapped model-facing message and persists it (`m.model_dump()`) into the `llm.human.input` event — and into `first_human_message`.
3. `GET /api/threads/{id}/messages` (`list_thread_messages`) returns event-store content verbatim; it never calls the unwrap helper `get_original_user_content_text` (that helper is only used on the regenerate path, `thread_runs.py:272`).
4. The frontend renders the wrapped text.

The fallback that should preserve the pre-wrap text (`original_user_content` in `additional_kwargs`) is only stamped for **uploads** (`uploads_middleware.py:270`, and it returns early at line 261 when there are no files) and **IM channels** (`channels/manager.py:487`). A plain web text message is never stamped, so even where unwrap is attempted it falls back to the wrapped content.

### Expected behavior

[...]

PR #3698, opened 2026-06-22
fix(runtime): skip hidden human messages in journal

### LINKED ISSUE #3694: [bug] Journal captures `hide_from_ui` HumanMessage as `first_human_message`
### Before you start

### Problem summary

`RunJournal.on_chat_model_start` captures the first `HumanMessage` via reverse scan, but its filter only excludes `name="summary"` — it does not check `additional_kwargs["hide_from_ui"]`. Hidden `HumanMessage`s injected *after* the real user message (e.g. `TodoMiddleware.todo_reminder`, `ViewImageMiddleware`) get captured as `first_human_message`, polluting the `llm.human.input` event and the runs API `first_human_message` field.

### Affected area(s)

### What happened?

**Introduced by**: not a single PR — the Journal filter (`m.name != "summary"`) is pre-existing and was never updated when `hide_from_ui` markers were added by multiple PRs: `TodoMiddleware` (#2907), `ViewImageMiddleware` (#3437), and  `DynamicContextMiddleware.__memory` (#3630). The downstream filter was not kept in sync with the marker.

```python:225:240:backend/packages/harness/deerflow/runtime/journal.py
        if not self._first_human_msg and messages:
            for batch in reversed(messages):
                for m in reversed(batch):
                    if isinstance(m, HumanMessage) and m.name != "summary":
                        ...
                        self.set_first_human_message(m.text)
```

Reverse scan + no `hide_from_ui` check → the last `HumanMessage` wins. `before_model` middlewares append their hidden messages *after* the user message, so they are hit first:

```python:138:140:backend/packages/harness/deerflow/agents/middlewares/todo_middleware.py
        reminder = HumanMessage(
            name="todo_reminder",
            additional_kwargs={"hide_from_ui": True},
```

```python:185:backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
        human_msg = HumanMessage(content=image_content, additional_kwargs={"hide_from_ui": True})
```

### Expected behavior

Skip `hide_from_ui` messages so only real user input is captured:

```python
if isinstance(m, HumanMessage) and m.name != "summary" and not m.additional_kwargs.get("hide_from_ui"):
```

### Steps to reproduce

1. Have the model call `write_todos` once, then continue until the call scrolls out of context — `TodoMiddleware` injects `todo_reminder` after the latest user message.
2. Inspect the run via the runs API.
3. `first_human_message` / `llm.human.input` content is `"Your todo list from earlier..."` instead of the user's actual input.

### Relevant logs

```shell
Verification script output (`deer-flow-gateway`, Python 3.12.13):

    Bug 3: Journal first_human_message capture
    Expected first_human_message: 'What is the weather today?'
    Actual   first_human_message: '<system_reminder>\nYour todo list from earlier is no longer v'...
    RESULT: BUG TRIGGERED — Journal captured todo_reminder as first_human_message
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

x86_64 (container), win32 host

### Python version

_No response_

### Node.js version

 v22.23.0

### pnpm version

10.26.2

### uv version

0.7.20

### Git state

_No response_

### Additional context

_No response_

Huixin615:
Fixes #3694

## Why

`RunJournal.on_chat_model_start` scans prompt messages in reverse to capture the user input associated with a run.

The existing filter excluded `HumanMessage` objects named `summary`, but did not exclude messages marked with `additional_kwargs["hide_from_ui"]`. Internal messages appended by middleware, such as `TodoMiddleware` reminders and `ViewImageMiddleware` context, could therefore be captured before the real user message.

This polluted the persisted `first_human_message` value and the `llm.human.input` journal event with framework-internal content.

## What changed

- Exclude `HumanMessage` objects marked with `hide_from_ui=True` when selecting the run's first human message.
- Preserve the existing reverse scan so the latest visible user input in the initial model prompt remains the selected message.
- Add regression coverage for:
  - A hidden middleware message appended after real user input.
  - Prompts containing only hidden human messages.
  - A visible user message arriving after an internal-only model prompt.
- Verify hidden messages do not create `llm.human.input` events or increment the human message summary count.

## Surface area

## Screenshots / Recording

Not applicable. This change affects backend journal message selection and has no UI changes.

## Bug fix verification

- Test path that reproduces the bug: `backend/tests/test_run_journal.py`
- Did it go red on `main` and green on this branch? Yes. Before the fix, the journal selected the trailing hidden `todo_reminder` as `first_human_message`; the regression tests pass with the new visibility filter.
- Additional scenarios verify that internal-only prompts create no human input event and do not prevent a later visible user message from being captured.

## Validation

Executed from `backend/`:

```bash
make lint
make test
Results:
Ruff lint passed.
Ruff format check passed for 587 files.
The complete backend test suite completed successfully with 4,907 tests collected.
git diff --check passed.
The final diff contains only:backend/packages/harness/deerflow/runtime/journal.py
backend/tests/test_run_journal.py

### CODE REVIEW

File: backend/packages/harness/deerflow/runtime/journal.py [hunk 1]
```diff
@@ -225,7 +225,7 @@ def on_chat_model_start(
         if not self._first_human_msg and messages:
             for batch in reversed(messages):
                 for m in reversed(batch):
-                    if isinstance(m, HumanMessage) and m.name != "summary":
+                    if isinstance(m, HumanMessage) and m.name != "summary" and m.additional_kwargs.get("hide_from_ui") is not True:
```
fancyboi999, line 228: Correct, and it matches the existing convention rather than inventing a new check — this is the exact negation of `_is_hidden_message` (`app/gateway/routers/thread_runs.py:234`, `... is True`), so the journal now honors `hide_from_ui` the same way the display path does. Covers all three producers (todo `:140`, view_image `:185`, dynamic-context `__memory` `:174/183`), which all set the flag to literal `True`.

backend/tests/test_run_journal.py
```diff
@@ -854,6 +854,76 @@ async def test_skips_summary_named_human_messages(self, journal_setup):
...
+
+        j, store = journal_setup
+        hidden_message = HumanMessage(
+            content="Internal context",
+            additional_kwargs={"hide_from_ui": True},
+        )
+        j.on_chat_model_start({}, [[hidden_message]], run_id=uuid4(), tags=["lead_agent"])
+        await j.flush()
+
+        assert j._first_human_msg is None
```
fancyboi999, line 896: Good edge to pin: an all-hidden prompt leaves `_first_human_msg` as `None` (no wrong capture, no spurious `llm.human.input` event), and the following test proves a later visible turn is still captured since `_first_human_msg` stays unset. This is the case I most wanted to see covered.

fancyboi999 (commented):
Thanks for the quick turnaround. Reviewed the diff and traced it against the producers — this is correct and well-scoped; reads good to merge from my side.

What I checked:
- The filter change is right and, importantly, consistent with the existing backend convention: `hide_from_ui ... is not True` is the exact negation of `_is_hidden_message` at `app/gateway/routers/thread_runs.py:234` (`additional_kwargs.get("hide_from_ui") is True`), so the journal now honors the marker the same way the display path already does.
- It covers all three producers that set the marker: `TodoMiddleware` (`todo_middleware.py:140`), `ViewImageMiddleware` (`view_image_middleware.py:185`), and the p0 `__memory` turn (`dynamic_context_middleware.py:174/183`) — all set `hide_from_ui=True`, so all three are now skipped.
- The reverse scan still selects the latest *visible* human turn in the prompt, which is the run's actual input — the pre-bug intent is preserved, only the hidden trailers are stepped over.
- Regression coverage is solid with real red anchors: the trailing-hidden case (picks `todo_reminder` on `main`), the all-hidden prompt → `first_human_msg` stays `None` with no `llm.human.input` event, and the hidden-only-then-visible case proving a skipped internal prompt doesn't block a later real capture. Pinning the `llm.human.input` event counts (not just `_first_human_msg`) is the right call.

Two optional follow-ups, neither a blocker here:
- #3695 is the same marker gap in `filter_messages_for_memory`. When that lands, a shared `is_hidden_from_ui(message)` predicate used by both consumers (and `thread_runs.py:234`) would keep them from drifting apart a third time.
- This same capture (lines 228-234) is where #3689/#3692 lives — the prompt-injection wrapper leaking into `first_human_message`. Separate issue, just flagging it's the same spot for whoever picks it up.

Nothing blocking from my read — happy to see it merge.

rayhpeng:
LGTM — clean, minimal, well-targeted fix. ✅

The root cause analysis is spot on: the reverse scan in `on_chat_model_start` was meant to grab the latest visible user message in the initial prompt, but `before_model` middlewares (`TodoMiddleware`, `ViewImageMiddleware`, `DynamicContextMiddleware`) append hidden `HumanMessage`s *after* the real input, so they were getting captured first.

A few things I like:

- **The `is not True` check is the right call.** The issue suggested truthiness (`not ...get("hide_from_ui")`), but using strict identity matches the existing consumer convention in `app/channels/manager.py` (`additional_kwargs.get("hide_from_ui") is True`). Since every producer sets a boolean `True` (the `TodoMiddleware` test even asserts `is True`), behavior is equivalent, and aligning with the established pattern is better for consistency.
- **Test coverage is thorough** — hidden message after real input, hidden-only prompt, and a visible message arriving after a hidden-only prompt, all verifying both the `llm.human.input` event and the message count.
- Single capture point, no other paths missed.

One non-blocking note: `is not True` means a future producer setting a truthy-but-not-`True` value (e.g. a `"true"` string from a serialization round-trip) would slip through. That's effectively zero risk here since the contract is strictly boolean `True` and these messages are constructed in-process before the LLM call, but worth keeping in mind if the marker ever crosses a serialization boundary.

Nice fix — approving.

rayhpeng (approved):

PR #3700, opened 2026-06-22
perf(runtime): resolve SSE resume offset in O(1) via the event-id seq

ly-wang19:
## Why

`MemoryStreamBridge._resolve_start_offset` translates a reconnecting client's `Last-Event-ID` into a buffer offset by **linearly scanning the retained event buffer** (up to `queue_maxsize=256` entries). This runs on **every SSE subscribe / reconnect** that carries a `Last-Event-ID` — and flaky/mobile clients reconnect repeatedly mid-run.

The scan is unnecessary: event ids are `"{ts}-{seq}"` where `seq` (from `_next_id`) is a per-run, monotonically increasing counter that **equals the event's absolute offset**. So the offset is pure arithmetic.

## What changed

Parse `seq` from `last_event_id`, index into the buffer at `seq - start_offset`, and **verify the id at that position matches exactly**. On any mismatch — evicted (`seq < start_offset`), out-of-range, foreign (right seq, wrong `ts`), or malformed id — it falls back to replay-from-earliest with the same warning, **byte-for-byte identical to the previous linear scan**.

**O(buffer ≤ 256) → O(1)** per reconnect. Behavior is unchanged (verified by a parity test against the old scan).

> Scope note: the retained buffer is capped at 256, so this is a bounded scan today — the change removes per-reconnect linear work and, just as much, makes the resume math intention-revealing (the id already encodes the offset). Same "the data already encodes what you're scanning for" principle as the recently-merged event-store indexing (#3531, #3499).

## Surface area

Internal runtime optimization under `backend/packages/harness/deerflow/runtime/stream_bridge/`. No API, SSE event, or resume-behavior change.

- [ ] **Frontend UI**
- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**

## Validation

`tests/test_stream_bridge.py` — 20 passed. Existing resume tests (`test_subscribe_replays_after_last_event_id`, `test_slow_subscriber_does_not_skip_after_buffer_trim`) stay green; added:
- `test_parse_event_seq` — id parsing incl. malformed forms;
- `test_resolve_start_offset_matches_linear_scan` — **parity vs the old scan** across retained / evicted / foreign (same seq, wrong ts) / malformed / `None` ids;
- `test_subscribe_with_unknown_last_event_id_replays_from_earliest` — end-to-end fallback.

## AI assistance

**How you used it:** AI spotted the linear `Last-Event-ID` scan, replaced it with an O(1) seq-indexed lookup (keeping an exact-match verify + identical fallback), and wrote the parity/edge-case tests; I reviewed the semantics and ran the suite before opening.

### DIFF
--- a/backend/packages/harness/deerflow/runtime/stream_bridge/memory.py
+++ b/backend/packages/harness/deerflow/runtime/stream_bridge/memory.py
@@ -48,13 +48,36 @@ class MemoryStreamBridge(StreamBridge):
         seq = self._counters[run_id] - 1
         return f"{ts}-{seq}"
 
+    @staticmethod
+    def _parse_event_seq(event_id: str) -> int | None:
+        """Extract the per-run sequence number from a ``{ts}-{seq}`` event id.
+
+        ``seq`` (assigned by :meth:`_next_id`) increases by one per published
+        event, so it equals the event's absolute offset within the run. Returns
+        ``None`` for ids that do not match the expected format.
+        """
+        _, sep, seq_text = event_id.rpartition("-")
+        if not sep:
+            return None
+        try:
+            return int(seq_text)
+        except ValueError:
+            return None
+
     def _resolve_start_offset(self, stream: _RunStream, last_event_id: str | None) -> int:
         if last_event_id is None:
             return stream.start_offset
 
-        for index, entry in enumerate(stream.events):
-            if entry.id == last_event_id:
-                return stream.start_offset + index + 1
+        # Event ids embed a per-run, monotonically increasing ``seq`` that equals
+        # the event's absolute offset, so locate the event by arithmetic in O(1)
+        # rather than scanning the retained buffer. The id is verified at the
+        # computed index, so a stale/evicted/foreign/malformed id still falls back
+        # to replay-from-earliest — identical to the previous linear scan.
+        seq = self._parse_event_seq(last_event_id)
+        if seq is not None:
+            local_index = seq - stream.start_offset
+            if 0 <= local_index < len(stream.events) and stream.events[local_index].id == last_event_id:
+                return stream.start_offset + local_index + 1
 
         if stream.events:
             logger.warning(
--- a/backend/tests/test_stream_bridge.py
+++ b/backend/tests/test_stream_bridge.py
@@ -334,3 +334,73 @@ async def test_make_stream_bridge_defaults():
     """make_stream_bridge() with no config yields a MemoryStreamBridge."""
     async with make_stream_bridge() as bridge:
         assert isinstance(bridge, MemoryStreamBridge)
+
+
+# ---------------------------------------------------------------------------
+# _resolve_start_offset: O(1) seq-indexed resolution (parity with linear scan)
+# ---------------------------------------------------------------------------
+
+
+def _linear_resolve(stream, last_event_id):
+    """The original linear-scan resolver, kept as a parity reference."""
+    if last_event_id is None:
+        return stream.start_offset
+    for index, entry in enumerate(stream.events):
+        if entry.id == last_event_id:
+            return stream.start_offset + index + 1
+    return stream.start_offset
+
+
+@pytest.mark.parametrize(
+    "event_id,expected",
+    [
+        ("1718000000000-0", 0),
+        ("1718000000000-42", 42),
+        ("garbage", None),  # no separator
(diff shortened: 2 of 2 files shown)


backend/packages/harness/deerflow/runtime/stream_bridge/memory.py
```diff
@@ -48,13 +48,36 @@ def _next_id(self, run_id: str) -> str:
...
-                return stream.start_offset + index + 1
+        # Event ids embed a per-run, monotonically increasing ``seq`` that equals
+        # the event's absolute offset, so locate the event by arithmetic in O(1)
+        # rather than scanning the retained buffer. The id is verified at the
+        # computed index, so a stale/evicted/foreign/malformed id still falls back
+        # to replay-from-earliest — identical to the previous linear scan.
+        seq = self._parse_event_seq(last_event_id)
+        if seq is not None:
+            local_index = seq - stream.start_offset
+            if 0 <= local_index < len(stream.events) and stream.events[local_index].id == last_event_id:
```
fancyboi999, line 79: This verify-at-position is the crux of the safety. Arithmetically `start_offset + local_index + 1 == seq + 1`, identical to the old scan's return for a retained id; and because you re-check `events[local_index].id == last_event_id`, a foreign id (right seq, wrong ts) or an evicted/out-of-range one fails here and falls through to the same replay-from-earliest fallback — so behavior is unchanged, not just faster. The `seq == offset` premise it relies on is structurally guaranteed by one `_next_id` per `append` (`publish`, `:70`) plus the `start_offset += overflow` trim (`:73-76`).

backend/packages/harness/deerflow/runtime/stream_bridge/memory.py
```diff
@@ -48,13 +48,36 @@ def _next_id(self, run_id: str) -> str:
...
 
+    @staticmethod
+    def _parse_event_seq(event_id: str) -> int | None:
+        """Extract the per-run sequence number from a ``{ts}-{seq}`` event id.
+
+        ``seq`` (assigned by :meth:`_next_id`) increases by one per published
+        event, so it equals the event's absolute offset within the run. Returns
+        ``None`` for ids that do not match the expected format.
+        """
+        _, sep, seq_text = event_id.rpartition("-")
```
fancyboi999, line 59: `rpartition("-")` is the right call — it splits on the last `-`, so the parse stays correct regardless of the `ts` shape, and the no-separator / non-int / empty cases all return `None`. `test_parse_event_seq` covers those forms.

fancyboi999 (commented):
Confirmed — traced the seq↔offset relationship and the fallback parity; this is correct and safe to merge.

What I checked:
- The premise holds structurally, not by luck: `_next_id` is called exactly once per `events.append` (`memory.py:70`), incrementing a 0-based per-run counter (`:45-50`), and front-trimming bumps `start_offset` in lockstep (`:73-76`). So a retained event with `seq=S` is always at `events[S - start_offset]`, and the new `start_offset + local_index + 1` reduces to `S + 1` — byte-identical to the old scan's return.
- The id-verify at the computed index is what makes it safe: evicted (`seq < start_offset`), out-of-range, foreign (right seq / wrong ts), and malformed ids all fail the `0 <= local_index < len` + `events[local_index].id == last_event_id` guard and fall through to the same replay-from-earliest path. I walked each class against the old scan and they match.
- `_parse_event_seq` uses `rpartition("-")`, so the `{ts}-{seq}` parse is robust regardless of `ts`, and returns `None` for no-separator / non-int / empty (covered by `test_parse_event_seq`).
- The parity test (`test_resolve_start_offset_matches_linear_scan`) pins behavioral equivalence across None / garbage / non-int / evicted / foreign / retained, including the buffer-trim case (`start_offset == 6`), and keeps the old scan as the reference. Scope is clean — only `memory.py` + tests, no API or resume-behavior change.

One forward-looking note (non-blocking): the optimization rests on the "one `_next_id` per append" invariant. If a future change ever appends to `stream.events` without `_next_id` (or vice-versa), `seq` would drift from the offset and the id-verify would silently downgrade precise resumes to replay-from-earliest — still correct, but a silent perf regression. A one-line comment at `_next_id`/`publish` noting that coupling would help it not get broken inadvertently.

WillemJiang (approved):

PR #3701, opened 2026-06-22
perf(models): index assistant-payload signature matching (O(P×A) → O(P+A))

ly-wang19:
## Why

`restore_assistant_payloads` re-attaches provider fields (e.g. `reasoning_content`) onto serialized assistant payloads. When the serialized payload length differs from the original message list (dropped/reordered messages), it falls back to signature matching — and for **each** assistant payload it **rescanned every original AIMessage**:

```python
matches = [index for index, ai_msg in enumerate(ai_messages)
           if index not in used_ai_indexes and _ai_signature(ai_msg) == payload_key]
if len(matches) == 1:
    ...
```

That's **O(P × A)** (P = assistant payloads, A = AIMessages, both grow with conversation length), recomputed on the relevant provider requests (vLLM/MiniMax/DeepSeek/StepFun reasoning replay, OpenAI tool-call signatures).

## What changed

The rule — restore only when **exactly one *unused*** AIMessage shares the payload's signature — is equivalent to *"that signature's unused-index bucket has size 1"*. So build a `signature → unused indexes` map once (O(A)) and look it up per payload in **O(1)**, removing each index from its bucket as it's consumed. The positional fallback (`_next_unused_index_at_or_after`) is unchanged.

**O(P × A) → O(P + A)** for the signature path. Behavior is identical.

## Surface area

Internal optimization under `backend/packages/harness/deerflow/models/`. No API or behavior change.

- [ ] **Frontend UI**
- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**

## Validation

`tests/test_assistant_payload_replay.py` + `tests/test_patched_*` — 56 passed. All existing replay tests (unique content/tool-call/structured signature, ambiguous→positional, forward-scan, no-wrap-to-earlier) stay green; added:
- `test_restore_assistant_payloads_parity_with_reference` — **2000 fuzzed scenarios** (mixed AI/Human, colliding/empty/tool-call signatures, divergent lengths) asserting the new matcher restores byte-for-byte the same fields as a kept copy of the original O(P×A) implementation.

## AI assistance

**How you used it:** AI identified the per-payload rescan, proved the "exactly one unused match" rule equals a size-1 signature bucket, implemented the O(1) indexed lookup (positional fallback untouched), and wrote the fuzz parity test; I reviewed the equivalence argument and ran the suite before opening.

### DIFF
--- a/backend/packages/harness/deerflow/models/assistant_payload_replay.py
+++ b/backend/packages/harness/deerflow/models/assistant_payload_replay.py
@@ -9,6 +9,7 @@ decide which fields to restore.
 from __future__ import annotations
 
 import json
+from collections import defaultdict
 from collections.abc import Callable, Sequence
 from typing import Any
 
@@ -33,10 +34,23 @@ def restore_assistant_payloads(
     assistant_payloads = [m for m in payload_messages if m.get("role") == "assistant"]
     used_ai_indexes: set[int] = set()
 
+    # Index unused AI messages by signature so the unique-match lookup is O(1)
+    # per payload instead of rescanning every AI message (O(P*A) -> O(P+A)).
+    signature_by_index: list[tuple[str, str] | None] = [_ai_signature(m) for m in ai_messages]
+    unused_by_signature: dict[tuple[str, str], set[int]] = defaultdict(set)
+    for index, signature in enumerate(signature_by_index):
+        if signature is not None:
+            unused_by_signature[signature].add(index)
+
     for ordinal, payload_msg in enumerate(assistant_payloads):
-        ai_msg = _match_ai_message(payload_msg, ai_messages, used_ai_indexes, ordinal)
-        if ai_msg is not None:
-            restore(payload_msg, ai_msg)
+        ai_index = _match_ai_index(payload_msg, ai_messages, used_ai_indexes, unused_by_signature, ordinal)
+        if ai_index is None:
+            continue
+        used_ai_indexes.add(ai_index)
+        signature = signature_by_index[ai_index]
+        if signature is not None:
+            unused_by_signature[signature].discard(ai_index)
+        restore(payload_msg, ai_messages[ai_index])
 
 
 def restore_additional_kwargs_field(payload_msg: dict[str, Any], orig_msg: AIMessage, field_name: str) -> None:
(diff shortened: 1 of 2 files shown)


fancyboi999 (commented):
Confirmed the equivalence and traced the bucket maintenance — correct and well-scoped, good to merge.

What I checked:
- The core claim holds: "exactly one *unused* AIMessage shares the signature" == "that signature's unused bucket has size 1". `unused_by_signature[sig]` is built from all AI messages up front and stays in lockstep with `used_ai_indexes` because every consumption path discards the consumed index from its bucket. So `len(candidates) == 1` is exactly the old `len(matches) == 1`, and `next(iter(candidates))` returns the same index the old `matches[0]` did.
- The refactor is the right shape: `_match_ai_message` → `_match_ai_index` is now **pure** (returns an index, mutates nothing), and the caller does `used_ai_indexes.add` + `unused_by_signature.discard` together. That's what keeps the two structures from drifting — a matcher that mutated only one would be the classic dual-index bug.
- Edge classes line up with the old code: signature-`None` AI messages are in no bucket (and never matched the old filter either, since the path is gated on `payload_key is not None`); an emptied bucket is a size-0 set (not `None`), so it falls through to the positional fallback, same as old `matches == []`. The `_next_unused_index_at_or_after` fallback is untouched.
- The 2000-case fuzz parity test against the kept O(P×A) reference is the real evidence — colliding/empty/tool-call signatures, divergent lengths, and the `len(payload)==len(original)` fast path, asserting byte-identical restored fields.

backend/packages/harness/deerflow/models/assistant_payload_replay.py
```diff
@@ -33,10 +34,23 @@ def restore_assistant_payloads(
     assistant_payloads = [m for m in payload_messages if m.get("role") == "assistant"]
     used_ai_indexes: set[int] = set()
 
+    # Index unused AI messages by signature so the unique-match lookup is O(1)
+    # per payload instead of rescanning every AI message (O(P*A) -> O(P+A)).
+    signature_by_index: list[tuple[str, str] | None] = [_ai_signature(m) for m in ai_messages]
```
fancyboi999, line 39: Built once over all AI messages — the O(A) setup that replaces the per-payload O(A) rescan. The parallel `signature_by_index` list lets the consumption site look up a consumed index's signature in O(1) to keep the bucket in sync.

backend/packages/harness/deerflow/models/assistant_payload_replay.py
```diff
@@ -33,10 +34,23 @@ def restore_assistant_payloads(
...
-        ai_msg = _match_ai_message(payload_msg, ai_messages, used_ai_indexes, ordinal)
-        if ai_msg is not None:
-            restore(payload_msg, ai_msg)
+        ai_index = _match_ai_index(payload_msg, ai_messages, used_ai_indexes, unused_by_signature, ordinal)
+        if ai_index is None:
+            continue
+        used_ai_indexes.add(ai_index)
+        signature = signature_by_index[ai_index]
+        if signature is not None:
+            unused_by_signature[signature].discard(ai_index)
```
fancyboi999, line 52: This is the line that keeps `unused_by_signature` equivalent to the old `index not in used_ai_indexes` filter — every consumed index (signature-matched *or* positional-fallback) is discarded from its bucket here, so a later `len(candidates) == 1` sees exactly the still-unused matches. Matcher staying pure + caller owning both mutations is what makes it safe.

PR #3702, opened 2026-06-22
fix(claude): coalesce non-consecutive system messages before Anthropic request

NTNguyen13:
## Why

Running DeerFlow against a Claude model (e.g. Opus 4.8) fails with `LLM request failed: Received multiple non-consecutive system messages.` once a turn has a `SystemMessage` outside the leading block. Anthropic accepts system content only as a single leading block, so `langchain_anthropic._format_messages` raises as soon as it meets a second, non-consecutive system message. Several middleware paths produce that shape — summarization removing messages around an injected reminder, the midnight date-update reminder, or skill content — and it reproduces reliably with the `skill-creator` workflow (activate a skill, run a few tool calls; the next model call throws). It's Anthropic-specific: OpenAI-style providers tolerate mid-list system messages, so it only surfaces on Claude models.

## What changed

`ClaudeChatModel` now folds all system messages into a single leading
`SystemMessage` before the request is formatted — but only when they aren't already a front-anchored block. Valid requests are returned unchanged (same list object), so there's no behavior drift for the common case. Claude requests now succeed regardless of where middleware placed system content.

## Surface area

- [x] **Agents / LangGraph** — model-call path: system-message formatting in the Claude provider

## Bug fix verification

- Test: `backend/tests/test_claude_provider_system_coalesce.py::test_non_consecutive_system_messages_do_not_raise`
- Red on `main`, green on this branch: **yes**. On `main`, `_get_request_payload` with a `SystemMessage` stranded after human/AI turns raises `ValueError: Received multiple non-consecutive system messages.`; with the fix it returns a payload whose single leading `system` block carries both instructions.

## Validation

Run in the backend dev container:
- `ruff check` + `ruff format` on the changed files — clean
- `python -m pytest tests/test_claude_provider_system_coalesce.py tests/test_claude_provider_oauth_billing.py` — **19 passed**
- Live end-to-end: a real Claude Opus 4.8 request containing non-consecutive system messages now succeeds and honors both system instructions (previously errored).

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Claude Code root-caused the failure to `langchain_anthropic._format_messages`, implemented the coalescing fix in `ClaudeChatModel`, and wrote the regression test. A human reviewed the diff and ran the validation above.

### DIFF
--- a/backend/packages/harness/deerflow/models/claude_provider.py
+++ b/backend/packages/harness/deerflow/models/claude_provider.py
@@ -26,7 +26,7 @@ from typing import Any
 
 import anthropic
 from langchain_anthropic import ChatAnthropic
-from langchain_core.messages import BaseMessage
+from langchain_core.messages import BaseMessage, SystemMessage
 from pydantic import PrivateAttr
 
 logger = logging.getLogger(__name__)
@@ -39,6 +39,39 @@ THINKING_BUDGET_RATIO = 0.8
 # Override with ANTHROPIC_BILLING_HEADER env var if the hardcoded version drifts.
 _DEFAULT_BILLING_HEADER = "x-anthropic-billing-header: cc_version=2.1.85.351; cc_entrypoint=cli; cch=6c6d5;"
 OAUTH_BILLING_HEADER = os.environ.get("ANTHROPIC_BILLING_HEADER", _DEFAULT_BILLING_HEADER)
+
+
+def _coalesce_system_messages(messages: list[BaseMessage]) -> list[BaseMessage]:
+    """Merge non-leading system messages into one leading SystemMessage.
+
+    Anthropic accepts system content only as a single leading block. Several
+    DeerFlow middleware paths (skill activation, summarization message removal,
+    midnight date reminders) can leave a SystemMessage stranded mid-conversation,
+    which makes langchain_anthropic raise "Received multiple non-consecutive
+    system messages." Returns the input unchanged when the system messages
+    already form a front-anchored block (the common case), so normal requests
+    stay byte-for-byte identical.
+    """
+    system_idxs = [i for i, m in enumerate(messages) if isinstance(m, SystemMessage)]
+    if system_idxs == list(range(len(system_idxs))):
+        return messages
+
+    blocks: list[Any] = []
+    others: list[BaseMessage] = []
+    for message in messages:
+        if isinstance(message, SystemMessage):
+            content = message.content
+            if isinstance(content, str):
+                if content:
+                    blocks.append({"type": "text", "text": content})
+            elif isinstance(content, list):
+                blocks.extend(b if isinstance(b, dict) else {"type": "text", "text": str(b)} for b in content)
+        else:
+            others.append(message)
+
+    if not blocks:
+        return others
+    return [SystemMessage(content=blocks), *others]
 
 
 class ClaudeChatModel(ChatAnthropic):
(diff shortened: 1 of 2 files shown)


fancyboi999 (commented):
Confirmed the Anthropic constraint and traced the fix — correct and well-scoped, good to merge.

What I checked:
- The premise is right: Anthropic accepts system content only as the top-level `system` parameter (a single leading block), never as a `system`-role entry interleaved in `messages`. `langchain_anthropic._format_messages` collects leading `SystemMessage`s into that param and raises once it meets a non-leading one — exactly the error here. Folding all system content into one leading block is the faithful fix.
- The common case is untouched: `_coalesce_system_messages` returns the **same list object** when the system messages already form a front-anchored block (`system_idxs == list(range(len(system_idxs)))`), and `_get_request_payload` passes the original `input_` through in that case (`input_ if coalesced is messages else coalesced`). So a normal one-leading-system request sees zero behavior or cache change — important since `ClaudeChatModel` does prompt caching.
- Structured `cache_control` blocks survive: list-content system messages are extended verbatim, so an ephemeral-cache block keeps its marker (`test_coalesce_preserves_list_content_blocks`). Empty-string system content is dropped, which is harmless.
- Tests are solid: the red-on-main anchor (`test_non_consecutive_system_messages_do_not_raise`), the front-anchored no-op, A/B/C + non-system order preservation, and the identity-return cases.

One thing worth naming (not a blocker): coalescing changes the *position* of a mid-conversation system message — a date/skill reminder injected after several turns now appears in the leading block. That's inherent to Anthropic's single-leading-block constraint (no faithful position-preserving representation for a `SystemMessage` on the standard path), and the role stays system-level, so for the cited producers it's the right tradeoff. Just flagging it as a conscious choice.

backend/packages/harness/deerflow/models/claude_provider.py
```diff
@@ -41,6 +41,39 @@
...
+    Anthropic accepts system content only as a single leading block. Several
+    DeerFlow middleware paths (skill activation, summarization message removal,
+    midnight date reminders) can leave a SystemMessage stranded mid-conversation,
+    which makes langchain_anthropic raise "Received multiple non-consecutive
+    system messages." Returns the input unchanged when the system messages
+    already form a front-anchored block (the common case), so normal requests
+    stay byte-for-byte identical.
+    """
+    system_idxs = [i for i, m in enumerate(messages) if isinstance(m, SystemMessage)]
+    if system_idxs == list(range(len(system_idxs))):
```
fancyboi999, line 56: This is the key no-drift guarantee: a front-anchored (or absent) system block makes `system_idxs == list(range(len(system_idxs)))` true, so the helper returns the *same object* and `_get_request_payload` passes the original `input_` through. The common single-leading-system case is byte-for-byte unchanged, so prompt caching isn't disturbed.

backend/packages/harness/deerflow/models/claude_provider.py
```diff
@@ -139,7 +172,12 @@ def _get_request_payload(
         **kwargs: Any,
     ) -> dict:
         """Override to inject prompt caching, thinking budget, and OAuth billing."""
-        payload = super()._get_request_payload(input_, stop=stop, **kwargs)
+        # Anthropic requires all system content as one leading block; coalesce
+        # defensively so stray mid-conversation SystemMessages don't blow up
+        # _format_messages with "multiple non-consecutive system messages".
+        messages = self._convert_input(input_).to_messages()
+        coalesced = _coalesce_system_messages(messages)
+        payload = super()._get_request_payload(input_ if coalesced is messages else coalesced, stop=stop, **kwargs)
```
fancyboi999, line 180: The `input_ if coalesced is messages else coalesced` identity check is what keeps the valid path zero-cost — only stray-system requests pay the rebuild. `_convert_input` does run twice when coalescing (here + inside super), but only on the error path, so it's negligible.

WillemJiang:
#3711 addressed the issue more generally.

PR #3703, opened 2026-06-22
test(frontend): migrate unit tests to rstest

chenjiahan:
## Why

This PR migrates the frontend unit test runner from Vitest to [Rstest](https://rstest.rs/).

Rstest is also an open-source project from ByteDance. Using it here keeps the frontend testing stack closer to the ByteDance open-source ecosystem and internal engineering practices.

This migration also improves test performance in the current repository (15% faster):

| Runner | Mean Time | Std Dev |
| --- | ---: | ---: |
| Vitest | 1.799s | +/- 0.157s |
| Rstest | 1.530s | +/- 0.086s |

## What changed

- Frontend unit tests now run with Rstest instead of Vitest.
- Test APIs were migrated from `vitest` / `vi.*` to `@rstest/core` / `rs.*`.
- Added `rstest.config.ts` and removed the old Vitest config/dependency.

## Surface area

- [x] **Dependencies** — adds `@rstest/core` and `@rsbuild/plugin-react`; removes `vitest`

## Screenshots / Recording

N/A

## Validation

```bash
cd frontend && pnpm typecheck
cd frontend && pnpm test
```

The migrated suite passes with 34 test files and 328 tests.

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Migrated the test runner, updated config and test imports, validated the suite, and measured performance with hyperfine.


frontend/rstest.config.ts
```diff
@@ -0,0 +1,19 @@
...
+  plugins: [pluginReact()],
+  resolve: {
+    alias: {
+      "@": resolve(__dirname, "src"),
+    },
+  },
+  output: {
+    // Streamdown imports KaTeX CSS as a side effect. Bundle these packages so
+    // Rsbuild processes that CSS import instead of Node trying to load it.
+    bundleDependencies: ["streamdown", "katex"],
```
fancyboi999, line 16: This is the one non-obvious config addition, and it's necessary rather than drift: under Rstest's bundler model `node_modules` are external by default, so streamdown's side-effect KaTeX CSS import would be handed to Node (which can't load `.css`). Bundling `streamdown`+`katex` routes it through Rsbuild's CSS handling. The streamdown tests (mermaid/plugins/preprocess) are in the suite, so it's load-bearing. The `@`→`src` alias and `include` match the old `vitest.config.ts` exactly.

frontend/tests/unit/core/tasks/subtask-result.test.ts
```diff
@@ -24,13 +24,9 @@ interface ContractFile {
...
 
-// The frontend package is ESM (`"type": "module"`), so `__dirname` is not
-// defined. Resolve the cross-language fixture relative to this module URL.
-const CONTRACT_PATH = fileURLToPath(
-  new URL(
-    "../../../../../contracts/subagent_status_contract.json",
-    import.meta.url,
-  ),
+const CONTRACT_PATH = resolve(
+  __dirname,
```
fancyboi999, line 28: This is the only non-mechanical change in the PR. `__dirname` works here because Rstest bundles via Rspack (which provides it), whereas Vitest ran the ESM package where it was undefined (the removed comment). Path resolution is otherwise identical, and the passing suite confirms it resolves to the contract fixture — just flagging it as the single spot whose correctness depends on the runner, not a pure rename.

fancyboi999 (commented):
Read the whole diff — this is a clean, complete, faithful migration. Good to merge from my side.

What I checked:
- **API mapping is 1:1 with exact counts.** Every `vi.*` call maps to its `rs.*` equivalent with identical occurrence counts across the 34 files: `fn` ×110, `stubGlobal` ×16, `mock` ×10, `advanceTimersByTimeAsync` ×8, `unstubAllGlobals` ×6, `resetModules` ×5, `mocked` ×4, `spyOn` ×3, `doMock` ×3, `restoreAllMocks` ×2, `doUnmock` ×2, `useFakeTimers`/`useRealTimers` ×1 each. No `vi.` or `from "vitest"` survives on any added line — the migration is complete, not half-done.
- **No tests dropped, skipped, or added.** No `it(`/`test(` declaration lines change, and no `.skip` / `.only` / `.todo` is introduced anywhere. The "34 files / 328 tests" claim lines up with the 34 migrated test files (the +1 `@rstest/core` import is `rstest.config.ts` itself).
- **Config parity holds.** `rstest.config.ts` keeps the same `@`→`src` alias and the byte-identical `include: ["tests/unit/**/*.test.ts"]` from the old `vitest.config.ts`. Neither config sets `globals`, a setup file, or an environment, so the explicit per-file imports stay correct. The two additions are necessary rstest glue, not behavior drift: `pluginReact()` for the JSX/TS transform Rsbuild needs, and `bundleDependencies: ["streamdown", "katex"]` so Rsbuild processes streamdown's side-effect KaTeX CSS import — which is load-bearing, since the streamdown tests are in the suite.

One thing worth naming (not a blocker): `subtask-result.test.ts` is the only file that changes module semantics rather than just swapping a name — it moves from `fileURLToPath(new URL(..., import.meta.url))` to `resolve(__dirname, ...)`. That's correct for the new runner (Vitest ran the ESM package where `__dirname` was undefined; Rstest bundles via Rspack, which provides it), and the contract-path resolution is otherwise unchanged. It's the one spot whose correctness rests on the runner rather than a pure rename — the passing suite confirms it resolves.

Reads correct and complete — happy to see it land.

WillemJiang:
@chenjiahan  please fix the lint error of the frontend.

chenjiahan:
Thanks for the review! The lint errors have been fixed.

WillemJiang (approved):

PR #3706, opened 2026-06-22
feat(persistence): wire alembic migrations, bootstrap schema on startup

### LINKED ISSUE #3682: Schema migrations never wired up: create_all can't add columns to existing DBs
## Description

After upgrading to a version that includes #3658 (`e7a03e52` `fix(gateway): attribute token usage to actual models`), **any pre-existing SQLite/Postgres database** returns 500 on every endpoint that touches the `runs` table:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such column: runs.token_usage_by_model
```

Affected endpoints (anything that queries `runs`):

- `POST /api/threads/{id}/runs/stream` → 500
- `GET  /api/threads/{id}/token-usage` → 500

A brand-new empty database is unaffected, so this is easy to miss in CI / clean environments.

## Root cause

This is not a one-off omission in #3658 — it exposes a **project-level gap: the schema migration mechanism was never actually wired up.**

1. #3658 added a new column to the ORM model (`persistence/run/model.py`):
   ```python
   token_usage_by_model: Mapped[dict] = mapped_column(JSON, default=dict)
   ```
   The `runs` queries in `sql.py` then started doing `SELECT runs.token_usage_by_model`, **but the PR shipped no migration script.**

2. Tables are created solely via `Base.metadata.create_all` (`persistence/engine.py:155-156`). `create_all` **only creates missing tables; it never adds columns to tables that already exist.** So the new column can never appear on an existing `runs` table.

3. The project ships an alembic skeleton, but the whole chain is broken:
   - `migrations/versions/` contains only a 0-byte `.gitkeep` — **there is not a single revision script.**
   - **Nothing anywhere runs `alembic upgrade`** (no Makefile / scripts / CI workflow / docs); the only `run_migrations` references live in `migrations/env.py` boilerplate, which nobody calls.
   - Consequently no database has an `alembic_version` table (`sqlite> SELECT * FROM alembic_version;` → `no such table`); alembic has never managed the schema.
   - `engine.py:142` even comments `# Auto-create tables (dev convenience). Production should use Alembic.` — but "Production should use Alembic" was never implemented.

Conclusion: any change that adds/alters a column on an existing table will break every pre-existing database unless users `ALTER` it by hand. This is structural and will recur on every future schema change.

## Steps to reproduce

1. Start once on a version *before* #3658 so `create_all` builds the `runs` table.
2. Upgrade to a version that includes #3658.
3. Open any existing thread or start a new run → 500 `no such column: runs.token_usage_by_model`.

## Workaround (user side)

```sql
-- SQLite
ALTER TABLE runs ADD COLUMN token_usage_by_model JSON;
```

Postgres equivalent: `ALTER TABLE runs ADD COLUMN token_usage_by_model JSONB;` (nullable is fine — old rows are handled by the legacy fallback in the code).

## Suggested proper fix

1. **Add an alembic migration for this column** under `migrations/versions/` (`op.add_column("runs", ...)`) and stamp existing databases.
2. **Wire migrations into the run flow**: execute `alembic upgrade head` in `make dev` / the Docker entrypoint / the deploy scripts, instead of relying on `create_all`.
3. **Establish the convention**: every ORM model change must ship a migration; keep `create_all` only for fast-starting brand-new empty databases, or hand schema ownership entirely to alembic.
4. Optional: add a CI regression that runs the key endpoints against a database built on the *old* schema + latest code, so this isn't only ever validated on an empty DB.

## Environment

- backend SQLite backend (`.deer-flow/data/deerflow.db`)
- Triggering commit: `e7a03e52` (#3658)

AnoobFeng:
Closes #3682. Pre-#3658 DBs lack the `runs.token_usage_by_model` column because alembic was never wired up — startup only ran `create_all`, which never ALTERs existing tables.

Adds a hybrid bootstrap in FastAPI lifespan (replaces bare `create_all`):
- empty DB     → create_all + stamp head
- legacy DB    → stamp 0001_baseline + upgrade head
- versioned DB → upgrade head

Concurrency: Postgres `pg_advisory_lock` (cross-process); SQLite per-engine `asyncio.Lock` + 30s `PRAGMA busy_timeout` on both prod and alembic engines. Column revisions use `safe_add_column` / `safe_drop_column` idempotent helpers as fallback.

Other bits:
- 0001 baseline (chain root) + 0002 add `runs.token_usage_by_model`
- `include_object` filter so alembic ignores LangGraph checkpointer tables
- `make migrate-rev MSG="..."` for authoring new revisions (no migrate/stamp targets — startup is the only execution path)
- Tests: three-branch decision, concurrency, #3682 regression, env filter, blocking-IO gate anchor
- CLAUDE.md: new "Schema migrations" section

Closes #3682

### 问题描述
升级到含 PR #3658 (`fix(gateway): attribute token usage to actual models`, commit `e7a03e52`) 的版本后, 任何**已存在**的 SQLite / Postgres 数据库的 `runs` 查询接口都会返回 500

### 根因
不是 #3658 加错了列, 而是项目从未真正启用迁移机制:
- 启动只跑 `Base.metadata.create_all` —— 它只 CREATE 缺失的**表**, 不会 ALTER 已存在的表
- alembic 骨架虽然存在, 但 `versions/` 仅有 0 字节 `.gitkeep`, 全链路从未启用

任何"加列"性质的 ORM 变更对老用户都是问题.

### 本 PR 的范围
同时干两件事:
1. 修 #3682 这条具体症状
2. 把"加列就崩老库"这类结构性短板补齐, 让后续 ORM 变更自动应用

### 启动期自动 bootstrap (面向用户)

Gateway 启动时根据 DB 当前形态走一条决策表, 全程无需手工 `alembic` 命令, 也不需要先停服做迁移:

| DB 状态 | 判定依据 | 动作 |
|---|---|---|
| **empty** | 无任何 DeerFlow 表 | `create_all` + `stamp head` |
| **legacy** | 有 DeerFlow 表, 无 `alembic_version` | `stamp 0001_baseline` + `upgrade head` |
| **versioned** | 已有 `alembic_version` 行 | `upgrade head` (已 head 时为 no-op) |

启动日志清晰打印走的是哪条分支 (`bootstrap: branch=empty/legacy/versioned -> ...`), 可直接在 `docker logs` 观察.

### 多实例并发安全

| Backend | 机制 | 强度 |
|---|---|---|
| **Postgres** | `pg_advisory_lock` (session-level) | 真·跨进程互斥, 会话断开自动释放 |
| **SQLite** | 进程内 `asyncio.Lock` + 30s `PRAGMA busy_timeout` (生产 engine 和 alembic-spawned engine 都设) | 单进程严格串行, 跨进程尽力而为 |

兜底: 每条列变更 revision 用幂等 helper (`safe_add_column` / `safe_drop_column`), 即使锁失效也不重复改.

### 后续开发约定

- 任何新增 / 修改 / 删除 ORM 字段, 必须配一个 alembic revision
- 开发命令: `make migrate-rev MSG="add foo column to runs"` (基于当前模型 autogenerate, review 后把裸 `op.add_column` 换成幂等 helper)
- LangGraph 自管的 checkpointer 表 (`checkpoints` / `checkpoint_blobs` / `checkpoint_writes` / `checkpoint_migrations`) 通过 `include_object` 过滤器排除在 alembic 视野外, 未来 autogenerate 不会误生成 `drop_table`
- `CLAUDE.md` 新增 "Schema migrations" 小节

### 其他细节说明
legacy 分支只覆盖"已有完整 baseline 表集合, 但缺后续 revision 添加的列"这种 DB. 它**不会**为以下情况额外补建表

### 自动化测试

```bash
```

### 仅跑本 PR 新增的迁移相关测试 

```bash
cd backend && PYTHONPATH=. uv run pytest -v \
  tests/test_persistence_bootstrap.py \
  tests/test_persistence_bootstrap_concurrency.py \
  tests/test_persistence_bootstrap_url.py \
  tests/test_persistence_migrations_env.py \
  tests/test_persistence_autogen_script.py
```

### Blocking-IO 运行时

```bash
```

(锚点: `tests/blocking_io/test_persistence_bootstrap.py` 保证 `bootstrap_schema` 内所有 alembic 同步调用都过 `asyncio.to_thread`, 不阻塞 lifespan eventloop.)

### 测试覆盖矩阵

| 测试文件 | 覆盖项 |
|---|---|
| [`test_persistence_bootstrap.py`](backend/tests/test_persistence_bootstrap.py) | 三分支决策 (empty / legacy-w/o-col / legacy-w/-col / versioned) + fresh vs legacy-upgraded 的列 nullability 一致性 |
| [`test_persistence_bootstrap_concurrency.py`](backend/tests/test_persistence_bootstrap_concurrency.py) | 2/5 并发 caller、cancel mid-flight、慢 upgrade 下的收敛性 |
| [`test_persistence_bootstrap_regression.py`](backend/tests/test_persistence_bootstrap_regression.py) | **#3682 直接回归**: 手工构造老 schema → bootstrap → SELECT `runs.token_usage_by_model` 成功 |
| [`test_persistence_migrations_env.py`](backend/tests/test_persistence_migrations_env.py) | `include_object` 过滤器: LangGraph 表跳过, DeerFlow 表保留 |
| [`test_persistence_bootstrap_url.py`](backend/tests/test_persistence_bootstrap_url.py) | URL 渲染: 密码不被掩码 + ConfigParser `%` 转义 |
| [`test_persistence_autogen_script.py`](backend/tests/test_persistence_autogen_script.py) | `scripts/_autogen_revision.py` 行为锁定: temp DB 走完整 migration history, 不走 `create_all` |
| [`blocking_io/test_persistence_bootstrap.py`](backend/tests/blocking_io/test_persistence_bootstrap.py) | Blockbuster gate 锚点 |

> 三种 DB 形态(empty / legacy / versioned)和并发收敛已在上述测试里完整覆盖, 不需要另起 Docker / 手工 SQL 复现.

## AI assistance

**Tool(s) used:** Claude Code


backend/packages/harness/deerflow/persistence/bootstrap.py
```diff
@@ -0,0 +1,343 @@
...
+
+# Per-engine SQLite bootstrap locks. Per-engine (not module-global) so each
+# engine instance pairs with a lock bound to the event loop that uses that
+# engine -- necessary because ``asyncio.Lock`` binds to the first loop it sees,
+# and pytest gives each async test its own loop. Production uses one engine
+# per process so this dict collapses to a single entry in practice.
+_SQLITE_LOCKS: dict[int, asyncio.Lock] = {}
+
+
+def _get_sqlite_local_lock(engine: AsyncEngine) -> asyncio.Lock:
```
willem-bd, line 160: **`_SQLITE_LOCKS` keyed by `id(engine)` can return a Lock bound to a dead event loop.**

CPython reuses `id()` after GC. A pytest suite that constructs engine A in loop 1, disposes it, then constructs engine B in loop 2 can hit the same address — `_get_sqlite_local_lock(B)` returns the cached Lock that's still bound to loop 1, and `async with` raises `RuntimeError: ... is bound to a different event loop`. Production single-engine is fine; the bootstrap concurrency tests in this PR happen to dispose each engine before the next test starts, but anything that holds two engines simultaneously across loops will surface this.

A `weakref.WeakKeyDictionary` keyed on the engine object both fixes this and removes the leak (see comment further down).

backend/packages/harness/deerflow/persistence/bootstrap.py
```diff
@@ -0,0 +1,343 @@
...
+# releases the prior lock).
+_PG_LOCK_KEY = 0x0DEE_12F1_0BEE_3682
+
+
+# Per-engine SQLite bootstrap locks. Per-engine (not module-global) so each
+# engine instance pairs with a lock bound to the event loop that uses that
+# engine -- necessary because ``asyncio.Lock`` binds to the first loop it sees,
+# and pytest gives each async test its own loop. Production uses one engine
+# per process so this dict collapses to a single entry in practice.
+_SQLITE_LOCKS: dict[int, asyncio.Lock] = {}
```
willem-bd: **This dict accumulates one entry per engine and never shrinks.**

Disposed engines are never popped; the lock dict grows without bound. Single-engine production is fine, but test runs that build many engines (and any future per-thread / per-tenant engine pool) leak forever. `WeakKeyDictionary` keyed on the engine would auto-clean — and as a bonus dodges the `id()` reuse hazard above.

backend/packages/harness/deerflow/persistence/bootstrap.py
```diff
@@ -0,0 +1,343 @@
...
+@asynccontextmanager
+async def _postgres_lock(engine: AsyncEngine):
+    """Hold a Postgres session-level advisory lock for the body of the block.
+
+    Session-level (not transaction-level) so the lock outlives implicit
+    transactions opened by alembic during ``stamp`` / ``upgrade``. The lock
+    is released explicitly on the way out and -- as a safety net -- when the
+    backing session disconnects (process crash, kill -9).
+    """
+    async with engine.connect() as conn:
```
willem-bd, line 347: **The advisory-lock connection sits idle in a transaction during stamp/upgrade — managed Postgres can kill it.**

`engine.connect()` implicitly opens a transaction. While `asyncio.to_thread(_stamp/_upgrade, ...)` is running alembic on a separate connection, this one is idle-in-transaction. Several managed Postgres providers (RDS, Cloud SQL, Supabase) set `idle_in_transaction_session_timeout` by default (e.g. 1–10 min). When it fires Postgres terminates the lock-holding session and `pg_advisory_lock` is silently released — a second Gateway process can then acquire the lock and run `upgrade head` concurrently with the first's in-flight DDL.

The idempotent helpers in `_helpers.py` cover `add_column` / `drop_column`, but not the `create_table` / `create_index` ops a future revision will use. Two fixes that don't disturb the rest of the design:

```python
async with engine.connect() as conn:
    # Disable idle-in-transaction kill for THIS session only.
    await conn.execute(text("SET LOCAL idle_in_transaction_session_timeout = 0"))
    await conn.execute(text("SELECT pg_advisory_lock(:k)"), {"k": _PG_LOCK_KEY})
```

or use `AUTOCOMMIT` isolation on the lock connection so it never sits in a transaction at all.

willem-bd (commented):
## Code review (high effort)

Nice piece of plumbing — the hybrid bootstrap + advisory-lock + idempotent-helpers triangle is well thought out, and the docstrings in `bootstrap.py` / the new revisions read like documentation rather than code comments. The test matrix (three branches × concurrency × URL pitfalls × blocking-IO anchor × #3682 regression) is comprehensive.

The findings below are all secondary — none of them block the #3682 fix. They cluster on the `_SQLITE_LOCKS` cache, the Postgres advisory-lock connection's transactional state, and a couple of small docstring / coverage gaps.

### Findings

1. **`_SQLITE_LOCKS` keyed by `id(engine)`** (`bootstrap.py:114`). `id()` is reused after GC; in pytest each test may run in a fresh event loop with a freshly-allocated engine that happens to land on the same address. The cached `asyncio.Lock` is bound to the prior loop and `async with` raises `RuntimeError: ... bound to a different event loop`. Switching to `WeakKeyDictionary` keyed on the engine object fixes this and the leak below in one move.

2. **Postgres advisory-lock session can be killed mid-upgrade** (`bootstrap.py:251`). `pg_advisory_lock` is session-scoped, but the `engine.connect()` context implicitly opens a transaction. Managed Postgres providers commonly set `idle_in_transaction_session_timeout` (the connection is idle while `to_thread(_upgrade, ...)` runs alembic in another connection); when that fires Postgres terminates the lock-holding session and the lock is silently released. A second Gateway process then acquires the lock and runs `upgrade head` concurrently with the first's in-flight DDL. The idempotent helpers cover column ops but not future `create_table` / `create_index` revisions. Two options: set `idle_in_transaction_session_timeout = 0` on the lock connection, or keep it busy with periodic `SELECT 1`.

3. **`_SQLITE_LOCKS` grows unbounded** (`bootstrap.py:111`). Each engine adds an entry that's never removed. Production single-engine is fine, but test suites that build many engines leak entries. `WeakKeyDictionary` keyed on the engine object would auto-clean and also solves finding #1.

4. **Docstring points at a file that doesn't exist** (`tests/blocking_io/test_persistence_engine_sqlite.py:17`). The comment names `test_persistence_bootstrap_alembic.py`; the actual anchor is `tests/blocking_io/test_persistence_bootstrap.py`. Small but it's a breadcrumb someone will follow during a future regression.

5. **`safe_add_column` silently accepts drift** (`_helpers.py:41`). It only compares the column NAME — a manually-applied `ALTER TABLE runs ADD COLUMN token_usage_by_model JSON` (nullable, no default) from the #3682 workaround is treated as "already present" and skipped, leaving a NULLABLE column where `create_all` would produce `NOT NULL DEFAULT '{}'`. INSERTs that rely on the server-side default will fail on legacy-upgraded DBs but succeed on fresh ones — exactly the cross-deployment drift this PR set out to eliminate. At minimum a `logger.warning` when name matches but type/nullability differ would let operators notice.

6. **Schema-parity test is narrow** (`test_persistence_bootstrap.py:215`). Today only `token_usage_by_model.nullable` is asserted equal between the create_all path and the alembic-upgrade path. Server defaults, index names, FK ondelete clauses — all room for `0001_baseline.upgrade()` to drift from `Base.metadata.create_all`. A reflected-MetaData comparison across every table would pin the contract this PR is built on.

backend/packages/harness/deerflow/persistence/migrations/_helpers.py
```diff
@@ -0,0 +1,56 @@
...
+
+    - Missing table => nothing to add to. Skip silently because bootstrap only
+      supports legacy DBs that already have the baseline table set.
+    - Column already exists => no-op.
+    """
+    insp = _inspector()
+    if table not in insp.get_table_names():
+        return
+    existing = {c["name"] for c in insp.get_columns(table)}
+    if column.name in existing:
```
willem-bd, line 173: **Skipping on name match alone silently accepts schema drift.**

A user who applied the #3682 workaround manually ran `ALTER TABLE runs ADD COLUMN token_usage_by_model JSON` — nullable, no default. `safe_add_column` here finds the column by name and no-ops, leaving:

- legacy-upgraded DB: `token_usage_by_model JSON NULL` (no server default)
- fresh DB via `create_all`: `token_usage_by_model JSON NOT NULL DEFAULT '{}'`

That's exactly the cross-deployment drift this PR set out to eliminate, just one layer deeper. INSERT statements that rely on the server-side default succeed on fresh DBs and fail on legacy-upgraded ones with `NOT NULL constraint failed`.

At minimum: when the name matches but type/nullability/server_default differ from the desired `sa.Column`, emit a `logger.warning` so operators see the drift in startup logs. A stronger version would `ALTER COLUMN` to repair, but the warning alone unblocks diagnosis.

backend/tests/blocking_io/test_persistence_engine_sqlite.py
```diff
@@ -11,6 +11,12 @@
 actually runs. The async engine/session machinery is mocked out so the only host
 filesystem operation under test is the directory creation; if it regresses to run
 directly on the event loop, Blockbuster raises `BlockingError` and this fails.
+
+We also stub ``bootstrap_schema`` so the alembic stamp/upgrade path -- which has
+its own ``asyncio.to_thread`` regression anchor in
+``test_persistence_bootstrap_alembic.py`` -- does not turn this test into a
```
willem-bd: **Stale docstring breadcrumb.** This points at `test_persistence_bootstrap_alembic.py`, which doesn't exist — the actual anchor is `tests/blocking_io/test_persistence_bootstrap.py` (added in this PR). Worth fixing now while the rename is fresh; a future regression hunter following this comment will think the offload coverage is missing and either duplicate or weaken it.

backend/tests/test_persistence_bootstrap.py
```diff
@@ -0,0 +1,277 @@
...
+
+        # Legacy DB -> stamp baseline + 0002 upgrade
+        await _seed_legacy_without_column(upgraded)
+        await bootstrap_schema(upgraded, backend="sqlite")
+        upgraded_col = await _runs_column_meta(upgraded, "token_usage_by_model")
+
+        # Pin the contract: the column must have the same nullability after
+        # either bootstrap path. If 0002 ever drifts from the model's
+        # ``Mapped[dict]`` (i.e. ``nullable=False``), this fires.
+        assert fresh_col["nullable"] == upgraded_col["nullable"], f"nullability drift: fresh={fresh_col['nullable']} upgraded={upgraded_col['nullable']}"
```
willem-bd, line 423: **Parity test only checks `nullable` on one column.**

The whole point of `0001_baseline.upgrade()` existing as a stamp target + chain root (per the revision's own docstring) is that it must stay byte-equivalent to `Base.metadata.create_all` for round-trip fixtures. Today only `token_usage_by_model.nullable` is pinned between the two paths. Server defaults, partial-index `where` clauses, FK `ondelete`, index naming — all of those can drift between `create_all` and the autogen-frozen `0001_baseline` DDL without firing this test.

It's worth adding a reflected-MetaData comparison across every table the revision creates, e.g.:

```python
fresh_md = sa.MetaData(); fresh_md.reflect(bind=fresh_sync_conn)
upgraded_md = sa.MetaData(); upgraded_md.reflect(bind=upgraded_sync_conn)
for name, t in fresh_md.tables.items():
    if name == "alembic_version": continue
    assert name in upgraded_md.tables
    # compare columns: name, type.python_type, nullable, server_default
```

This is the property the PR's design rests on; without it, a future autogen revision can quietly drift the baseline.

willem-bd (commented):
补充 @willem-bd 的 review,再提两点(都不阻塞 #3682 的修复):

[P1] legacy 分支不会补建缺失的 baseline 表

- Location: `backend/packages/harness/deerflow/persistence/bootstrap.py:333-334`
- Problem: legacy 分支执行 `stamp 0001_baseline` + `upgrade head`。但 alembic 只运行 stamp 位置**之后**的 revision,因此 `0001_baseline.upgrade()`(其中 `op.create_table` 建立全部 baseline 表)永远不会被执行。旧的 `Base.metadata.create_all` 是幂等的,每次启动都会补建**任何**缺失的表;新逻辑对 legacy DB 只通过 baseline 之后的 revision 加列,从不加表。
- Evidence: 任何"已有部分 DeerFlow 表、但没有 `alembic_version`"的 DB 都会落到 legacy 分支。`channel_*` 表是在统一持久层(#1930)之后才加入的,从一个早期版本跳过若干 release 升级上来的用户,其 DB 里可能没有 `channel_connections` / `channel_credentials` 等表。bootstrap 会正常完成并打到 head,启动日志一切正常,但首次访问 channel 接口时 500 报 `no such table: channel_connections`。
- Suggested fix: legacy 分支在 stamp 之前额外跑一次 `Base.metadata.create_all`(对已存在的表是 no-op,只补建缺失表);或对 legacy DB 改为 `stamp base` + `upgrade head`,让 0001 的建表 DDL 真正执行(需确认 baseline 与 `create_all` 产物一致)。
- Test: 构造一个只含 `runs`、缺失 `channel_*` 的 legacy DB,跑 `bootstrap_schema`,断言 `channel_connections` 等表已创建。

[P2] 0002 的 `server_default` 造成 fresh DB 与升级 DB 的列定义漂移

- Location: `backend/packages/harness/deerflow/persistence/migrations/versions/0002_runs_token_usage.py:68`
- Problem: 模型定义为 `mapped_column(JSON, default=dict)`,**没有** `server_default`,所以 empty 分支的 `create_all` 产出的是 `token_usage_by_model JSON NOT NULL`(无 DB 级默认值)。但 0002 迁移加了 `server_default=sa.text("'{}'")`,于是经 legacy 升级上来的 DB 会带上 `... NOT NULL DEFAULT '{}'`,与 fresh DB 不一致。
- Evidence: 上一条 review 的 #6 已经指出 parity 测试只校验 nullability、`server defaults` 是未覆盖的漂移面 —— 这里进一步指出该漂移在 0002 里**已经发生**(不是潜在风险):同一份 ORM 模型,两条 bootstrap 路径产出的列 DEFAULT 子句不同。后果:裸 SQL 插入在两种 DB 上行为不同;未来对 fresh DB 跑 `alembic revision --autogenerate` 会冒出一个多余的 `alter_column` 试图加/去 server_default。
- Suggested fix: 二选一 —— 要么给模型也声明同样的 `server_default`(两路一致),要么从迁移里去掉 `server_default`,改用先 `ADD COLUMN ... NULL` + 回填 `'{}'` + 再 `ALTER ... SET NOT NULL` 的方式(仍保留 `nullable=False`)。
- Test: 扩展 `test_token_usage_column_parity_between_fresh_and_upgraded`,除了 nullability,再断言两路产出的 `server_default` 一致。

AnoobFeng:
感谢code review，最近我会着手逐一优化处理

AnoobFeng:
感谢前面详细的review，针对这几点，做出了以下改动：

### 1. 修正文档中的失效测试路径

将 blocking-IO 测试说明中不存在的 `test_persistence_bootstrap_alembic.py` 修正为实际的 `test_persistence_bootstrap.py`，并同步更新 legacy bootstrap 的相关文档描述。

### 2. 修复 SQLite bootstrap lock 的生命周期问题

将 `_SQLITE_LOCKS` 从以 `id(engine)` 为 key 的普通字典改为以 engine 对象为 key 的 `WeakKeyDictionary`：

- 避免 engine GC 后 `id` 被复用，错误取得旧 event loop 上的 lock。
- engine 回收后缓存项可自动清理，避免长期增长。
- 新增覆盖 engine 隔离、GC 清理及跨 event-loop 场景的回归测试。

### 3. 增加已有列的 schema drift 告警

`safe_add_column` 遇到同名列时不再直接静默返回，而是比较：

- `nullable`
- `server_default`

发现不一致时记录 warning，帮助定位手工执行 `ALTER TABLE` 后遗留的 schema drift；当前仍保持幂等，不自动修改已有列。

### 4. 保护 Postgres advisory-lock 连接

在获取 `pg_advisory_lock` 前执行：

```sql
SET LOCAL idle_in_transaction_session_timeout = 0
```

避免托管 Postgres 因 lock connection 长时间处于 idle-in-transaction 状态而终止 session、意外释放 advisory lock。

新增测试验证 SQL 执行顺序，以及异常退出时仍会尝试释放锁。

### 5. 补建 legacy DB 缺失的 baseline 表

legacy 分支现在会在 stamp baseline 前补建缺失的 baseline 表，覆盖跨多个版本升级、缺少 `channel_*` 等表的旧数据库。

为避免提前创建未来 revision 管理的新表：

- backfill 仅允许创建 `_BASELINE_TABLE_NAMES` 中的表。
- 新增测试校验该集合与 `0001_baseline` 的真实建表结果一致。
- 注入 phantom future table，验证 backfill 不会越界创建它。

### 6. 统一 `token_usage_by_model` 的 server default

在 ORM model、`0001_baseline` 和 `0002_runs_token_usage` 中统一声明：

```python
server_default=text("'{}'")
```

### 7. 扩展 schema parity 测试

新增 create-all 与 Alembic `base -> head` 的全表对照，当前覆盖：

- 表集合
- 每张表的列集合
- 列的 `nullable`
- 列的 `server_default`

同时对 default 表达式进行归一化，避免 SQLite/Postgres 表达形式差异造成误报。

## 本轮未覆盖的边界

- `safe_add_column` 目前只告警，不自动修复已有 drift。
- drift 检测暂不比较 column type，避免 JSON/JSONB 等方言反射差异带来误报。
- schema parity 暂不比较 type、index、partial-index 条件和 FK 元数据；这些需要独立的跨方言方案。

## 验证

相关 persistence、migration、concurrency 和 blocking-IO 测试共 `39 passed`，Ruff 检查通过。

AnoobFeng:
hello 打扰了。
我注意到单元测试失败了，是新加的 test_create_all_and_alembic_upgrade_produce_same_schema drift 测试 —— 它在 create_all 这一侧多检测到一张 _widget_to_dict_test 表。

排查下来是测试间污染，不是真正的 schema drift：tests/test_base_to_dict.py 在模块顶层定义了 class _Widget(Base):，pytest collection 阶段 import 模块时，SQLAlchemy 的 DeclarativeBase 元类会把这张表自动注册到共享的 Base.metadata，整个进程都带着。同样的隐患在 tests/test_persistence_scaffold.py 里也有一处（方法内的 class _Tmp(Base):，只是当前跑序下还没暴露）。

是否在这个 PR 里一并修掉？有以下两个方案，想听一下你的偏好：

- 测试侧隔离：测试装置改用独立的 DeclarativeBase，需要测 Base 的方法时显式调用 Base.to_dict(w) / Base.__repr__(w)，再加一个小的 tripwire 测试，断言 Base.metadata 里不会有 _ 开头的测试装置泄漏进来。只动测试代码，PR scope 保持收敛。

- 生产侧抽 Mixin：在 base.py 把 to_dict / __repr__ 抽成 mixin，Base 和测试专用的 DeclarativeBase 都复用这个 mixin。架构上更干净，但改到了本 PR 范围之外的生产代码。
我倾向方案一，改动小，只动单测部分代码

WillemJiang:
@AnoobFeng, please drop me an email (willem.jiang) AT gmail.

backend/packages/harness/deerflow/persistence/migrations/_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+        text = text.split("::", 1)[0].strip()
+    return text or None
+
+
+def _check_column_drift(table: str, desired: sa.Column, actual: dict) -> None:
+    """Warn if an existing column's attributes diverge from the desired model.
+
+    Compares ``nullable`` and ``server_default``. Type comparison is skipped
+    because reflected type repr varies enough across dialects (e.g. ``JSON``
+    vs ``JSONB``) that it would produce too many false positives to act on.
```
willem-bd: [P2] Drift check skips column TYPE, so a manual workaround with the wrong type survives silently

- Problem: `_check_column_drift` compares only `nullable` and `server_default`. A user who applied the #3682 workaround as `ALTER TABLE runs ADD COLUMN token_usage_by_model TEXT NOT NULL DEFAULT '{}'` (right nullability/default, wrong type) gets zero warning — `safe_add_column` sees the name match and returns. The DB now has `TEXT` where the model expects `JSON`, which matters on Postgres (`json` operators don't apply to `text`).
- Evidence: lines 76-78 explicitly skip type comparison citing JSON/JSONB false positives. The author's follow-up already acknowledged this as a deliberate trade-off ("drift 检测暂不比较 column type, 避免 JSON/JSONB 等方言反射差异带来误报").
- Suggested fix: not necessarily auto-comparing, but at minimum emit the warning with the reflected type string included (`actual_type=<repr>`) so an operator triaging a JSON-vs-TEXT drift sees the type in the log line even if we don't assert on it. Cheap, no false-positive risk because there's no equality check.
- Test: extend `test_legacy_with_manual_workaround_column_warns_on_drift` to seed the column as `TEXT` and assert the warning message contains the reflected type.

backend/scripts/_autogen_revision.py
```diff
@@ -0,0 +1,77 @@
...
+MIGRATIONS_DIR = BACKEND_DIR / "packages/harness/deerflow/persistence/migrations"
+
+
+def _alembic_config(url: str) -> Config:
+    cfg = Config()
+    cfg.set_main_option("script_location", str(MIGRATIONS_DIR))
+    # ConfigParser treats ``%`` as interpolation syntax. Escape literal URL
+    # percents so encoded paths/passwords survive round-tripping.
+    cfg.set_main_option("sqlalchemy.url", url.replace("%", "%%"))
+    return cfg
```
willem-bd, line 54: [P2] Duplicates the `%`-escape / URL-rendering pattern from `bootstrap._alembic_safe_url`

- Problem: `_alembic_config` here does `url.replace("%", "%%")` to dodge the same ConfigParser interpolation pitfall that `bootstrap._alembic_safe_url` handles. Same logic in two places — if a third URL/ConfigParser corner case shows up (BOM, trailing whitespace, another interpolation token), it has to be fixed twice, and `tests/test_persistence_bootstrap_url.py` only exercises the bootstrap copy.
- Suggested fix: import `_alembic_safe_url` from `deerflow.persistence.bootstrap` and use it here. The autogen script accepts a raw URL string rather than an engine, so either: (a) wrap the URL in a tiny `SimpleNamespace(url=make_url(url))` shim and call `_alembic_safe_url`, or (b) factor the `.replace("%", "%%")` line into a `_escape_url_for_alembic(str) -> str` helper and use it from both call sites.
- Test: existing `test_alembic_config_accepts_url_with_percent_and_round_trips` would then cover both call paths.

backend/CLAUDE.md
```diff
@@ -95,6 +95,7 @@ make test               # Run all backend tests
 make test-blocking-io   # Run strict Blockbuster runtime gate on tests/blocking_io/
 make lint               # Lint with ruff
 make format             # Format code with ruff
+make migrate-rev MSG="..."  # Autogenerate a new alembic revision (see Schema Migrations section)
```
willem-bd, line 98: [P2] New developer-facing command added, but no `README.md` update — repo policy asks for one

- Problem: `backend/CLAUDE.md` itself states (lines 70-75, unchanged by this PR): *"**CRITICAL: Always update README.md and CLAUDE.md after every code change** … Update `README.md` for user-facing changes (features, setup, usage instructions) … Update `CLAUDE.md` for development changes"*. This PR adds a new `make migrate-rev MSG="..."` usage instruction (this line) and changes the operator-visible startup behavior (auto-migrate on lifespan), but updates only `backend/CLAUDE.md`. Neither `README.md` nor `backend/README.md` is touched.
- Evidence: borderline because production bootstrap is transparent (no manual `alembic` command needed), but `make migrate-rev` IS a developer usage instruction and would naturally live in a backend README "developer commands" section if one exists.
- Suggested fix: either add a short paragraph to the relevant `README.md` covering the new auto-bootstrap behavior + `make migrate-rev`, or — if maintainers consider it CLAUDE.md-only — relax the policy wording in CLAUDE.md to match actual practice.

willem-bd (commented):
Building on @willem-bd's two earlier reviews and the author's follow-up rev — the hybrid bootstrap + advisory-lock + idempotent-helpers triangle now reads very cleanly, and the test matrix (three branches × concurrency × URL pitfalls × Blockbuster anchor × #3682 regression × table-name pinning × create_all-vs-alembic parity) is genuinely thorough. Three secondary findings below — none block the #3682 fix.

WillemJiang:
@AnoobFeng Please take a look at the unit test error.

AnoobFeng:
Thanks for the code review — three follow-ups landed:

1. Type drift in _check_column_drift — Went a step beyond the "just echo the type in the payload" minimum. Added _type_equivalent with an explicit dialect-synonym allowlist (currently {JSON, JSONB} only) so the TEXT NOT NULL DEFAULT '{}' workaround you flagged now actually triggers a "type drift" entry, while JSON↔JSONB stays quiet on Postgres. Type info is still echoed in every warning payload for triage. New test_legacy_with_wrong_type_workaround_warns_on_type_drift pins your exact scenario, plus 4 unit tests pin the allowlist behavior.

Kept the allowlist intentionally minimal — only the one JSON/JSONB pair we actually hit today. Considered TypeEngine.as_generic() for principled family detection but deferred until the first non-JSON column ships via safe_add_column; adding pairs preemptively risks reopening the silent-drift hole.

2. %-escape duplication — Extracted _escape_url_for_alembic into bootstrap.py; _autogen_revision._alembic_config imports and uses it. Added a direct unit test so the shared rule has its own coverage anchor.

3. README sync — backend/README.md gains make migrate-rev MSG="..." in the Commands table plus a new Schema Migrations section covering startup auto-bootstrap and the autogen workflow, pointing at backend/CLAUDE.md for the full design. Left the root README untouched since the bootstrap is transparent to end users.

CI red, unrelated to your review: The create_all-vs-alembic parity test from the previous commit went red on full-suite runs (passed in isolation). Root cause: tests/test_base_to_dict.py declares a test-only _Widget(Base) at module level, leaking _widget_to_dict_test into Base.metadata during pytest collection — create_all then renders the phantom table while alembic does not.

Fixed at the source rather than allowlisting on the parity test side: _Widget now carries metadata = MetaData() so it scopes to its own metadata and never touches the application Base.metadata. Two-line change. 44 passed in the persistence + base_to_dict subset locally.

WillemJiang (approved):

WillemJiang (approved):

PR #3709, opened 2026-06-22
fix(middleware): fix positional fallback consuming unrelated todo when same-content list is exhausted

he-yufeng:
When `next_todos` contains the same content string twice (e.g. two `"A"` entries), the first match pops from `previous_by_content["A"]`, leaving an empty list. The second `"A"` sees that empty list, treats it as falsy, and sets `previous_match = None`. The positional fallback then fires unconditionally — consuming `previous_todos[index]` even when it holds a completely different entry (e.g. `"B"`). That marks `"B"` as matched, so the final loop never emits a `todo_remove` action for it.

**Trigger case:**
```python
previous_todos = [{'content': 'A', 'status': 'pending'}, {'content': 'B', 'status': 'pending'}]
next_todos     = [{'content': 'A', 'status': 'in_progress'}, {'content': 'A', 'status': 'completed'}]
# Expected: [todo_start for A, todo_complete for second A, todo_remove for B]
# Actual:   [todo_start for A, todo_update for A]  — B silently disappears
```

**Fix:** guard the positional fallback with `content not in previous_by_content` so it only fires when the content was genuinely absent from previous (never in the map at all), not when the same-content list happened to be exhausted by earlier iterations.

**Regression test** added to `backend/tests/test_token_usage_middleware.py` covering this exact scenario — all 9 tests pass.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/token_usage_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/token_usage_middleware.py
@@ -97,7 +97,7 @@ def _build_todo_actions(previous_todos: list[Todo], next_todos: list[Todo]) -> l
                 previous_index, previous_match = content_matches.pop(0)
                 matched_previous_indices.add(previous_index)
 
-        if previous_match is None and index < len(previous_todos) and index not in matched_previous_indices:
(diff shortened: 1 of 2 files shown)


WillemJiang:
@he-yufeng, please fix the lint error.

WillemJiang (approved):

PR #3711, opened 2026-06-22
fix(agents): coalesce SystemMessages before LLM request

### LINKED ISSUE #3707: [bug] LLM request failed: Error code: 400 - {'error': {'message': 'System message must be at the beginning.', 'type': 'BadRequestError', 'param': None, 'code': 400}}
### Before you start

### Problem summary

System message must be at the beginning

### Affected area(s)

### What happened?

最新的代码，按照拉取代码，用生产模式和开发模式部署，然后开始对话的时候会出现：LLM request failed: Error code: 400 - {'error': {'message': 'System message must be at the beginning.', 'type': 'BadRequestError', 'param': None, 'code': 400}}

模型配置如下
```
  # OpenRouter models use the same ChatOpenAI + base_url pattern as other OpenAI-compatible gateways.
   - name: Qwen3.5-122B-A10B
     display_name: Qwen3.5-122B-A10B
     use: langchain_openai:ChatOpenAI
     model: Qwen3.5-122B-A10B
     api_key: no_case
     base_url: http://127.0.0.1:8000/v1
     request_timeout: 600.0
     max_retries: 2
     max_tokens: 8192
     temperature: 0.7
```
一直报错LLM request failed: Error code: 400 - {'error': {'message': 'System message must be at the beginning.', 'type': 'BadRequestError', 'param': None, 'code': 400}}

### Expected behavior

正常对话问答才对

### Steps to reproduce

make up     # 构建镜像并启动全部生产服务
make down   # 停止并移除容器

### Relevant logs

```shell
2026-06-22 09:09:54 - httpx - INFO - HTTP Request: POST http://127.0.0.1:8000/v1/chat/completions "HTTP/1.1 400 Bad Request"
2026-06-22 09:09:54 - deerflow.agents.middlewares.llm_error_handling_middleware - WARNING - LLM call failed after 1 attempt(s): Error code: 400 - {'error': {'message': 'System message must be at the beginning.', 'type': 'BadRequestError', 'param': None, 'code': 400}}
Traceback (most recent call last):
  File "/app/backend/packages/harness/deerflow/agents/middlewares/llm_error_handling_middleware.py", line 365, in awrap_model_call
    response = await handler(request)
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain/agents/factory.py", line 376, in inner_handler
    inner_result = await inner(req, handler)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain/agents/factory.py", line 386, in composed
    outer_result = await outer(request, inner_handler)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py", line 286, in awrap_model_call
    return await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain/agents/factory.py", line 376, in inner_handler
    inner_result = await inner(req, handler)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py", line 593, in awrap_model_call
    return await handler(self._augment_request(request))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain/agents/factory.py", line 1330, in _execute_model_async
    output = await model_.ainvoke(messages)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_core/runnables/base.py", line 5766, in ainvoke
    return await self.bound.ainvoke(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 495, in ainvoke
    llm_result = await self.agenerate_prompt(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 1763, in agenerate_prompt
    return await self.agenerate(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 1721, in agenerate
    raise exceptions[0]
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 2013, in _agenerate_with_cache
    async for chunk in self._astream(messages, stop=stop, **kwargs):
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_openai/chat_models/base.py", line 3361, in _astream
    async for chunk in super()._astream(*args, **kwargs):
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_openai/chat_models/base.py", line 1863, in _astream
    _handle_openai_bad_request(e)
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_openai/chat_models/base.py", line 1834, in _astream
    response = await self.async_client.create(**payload)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/openai/resources/chat/completions/completions.py", line 2714, in create
    return await self._post(
           ^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/openai/_base_client.py", line 1913, in post
    return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/backend/.venv/lib/python3.12/site-packages/openai/_base_client.py", line 1698, in request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': 'System message must be at the beginning.', 'type': 'BadRequestError', 'param': None, 'code': 400}}
2026-06-22 09:09:54 - deerflow.agents.memory.queue - INFO - Memory update queued for thread c19ca4b0-c248-447b-9b7b-2606984ef3e3, queue size: 1
2026-06-22 09:09:54 - deerflow.runtime.runs.manager - INFO - Run 60c1b46b-0027-4538-8dae-df7a50b6f7e3 -> error
INFO:     172.19.0.4:54476 - "POST /api/threads/c19ca4b0-c248-447b-9b7b-2606984ef3e3/history HTTP
```

Docker (make docker-start)

### Operating system

Linux

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

ech0hol:
Closes #3707

## Why

Strict OpenAI-compatible backends (vLLM, SGLang, Qwen) and Anthropic reject non-leading `SystemMessage`s with `"System message must be at the beginning"` or `"Received multiple non-consecutive system messages"`. The official OpenAI API tolerates mid-conversation system messages, so the bug only surfaces when users point DeerFlow at a strict backend — any such user hits a 400 on the very first turn. There is no config-only workaround: date injection is unconditional on the lead agent.

## Root cause

Traced end-to-end in container (`deer-flow-gateway`, Python 3.12.13) by inspecting both `state["messages"]` (persisted) and the request payload sent to the LLM:

1. **`DynamicContextMiddleware`** (PR #3630) uses the ID-swap technique to replace the first/last `HumanMessage` with a triplet `[SystemMessage(reminder), HumanMessage(__memory), HumanMessage(__user)]`. The reminder intentionally stays a `SystemMessage` so framework-owned data (current date, metadata) does not masquerade as user input (OWASP LLM01). Verified via `add_messages` reducer: the `SystemMessage(id=X)` replaces the original `HumanMessage(id=X)` in-place, and `__memory`/`__user` (new ids) are appended immediately after — order is correct, no orphaning on the state side.
2. **`create_agent`** then prepends the static `system_prompt` as yet another `SystemMessage` at request time. Confirmed in `langchain.agents.create_agent` source (`_execute_model_sync` / `_execute_model_async`):
   ```python
   messages = request.messages
   if request.system_message:
       messages = [request.system_message, *messages]
   ```
   So `system_prompt` is **not** in `state["messages"]` — it is injected per-request.
3. The LLM receives two or more `SystemMessage`s — only the first is at the beginning → strict backends reject.

**Note:** ID-swap does **not** cause message reordering. I verified this in-container with a reproduction script: `__memory` and `__user` always follow their corresponding `SystemMessage` immediately. The only bug is the multiple-`SystemMessage` issue.

PR #3702 already fixes this for the Claude provider only (`claude_provider._coalesce_system_messages`), but the main agent on `ChatOpenAI` (OpenAI/Qwen/vLLM) is uncovered.

## What changed

- **New `SystemMessageCoalescingMiddleware`** (`system_message_coalescing_middleware.py`) merges every `SystemMessage` in the request payload into a single leading `SystemMessage` before the request reaches the provider. Non-system messages keep their original order; the merged `SystemMessage` preserves the id of the first system message (typically the static `system_prompt`).
- **Registered on the lead agent** in `build_lead_middlewares()` (`agent.py`), after `DeferredToolFilterMiddleware`.

### Design decisions

| Decision | Rationale |
|----------|-----------|
| `wrap_model_call` (not `before_agent`) | Runs on the final request payload — *after* `create_agent` prepends `system_prompt` — and never touches `state["messages"]`. Checkpoint structure is unchanged. |
| Fast path returns `None` | ≤1 `SystemMessage` → same request object returned, zero overhead, prefix-cache unaffected. |
| Merge, not downgrade to `HumanMessage` | Keeps `SystemMessage` semantics (OWASP LLM01: framework context must not masquerade as user input). |
| Provider-agnostic | One middleware covers every backend; mirrors #3702's per-request coalescing but at a shared layer so we stop patching per-provider. |
| Preserve first system message id | Downstream consumers keying off the leading system id are unaffected. |

### Trade-off: prefix-cache invalidation on date change

⚠️ Merging `system_prompt + date_reminder` into a single `SystemMessage` means the leading system content changes whenever the date changes (daily, or on midnight crossing). This breaks prefix-cache hits on the system block for OpenAI-compatible backends that rely on a static `system_prompt` prefix.

This is an accepted trade-off:

- Without this fix, strict backends (vLLM, SGLang, Qwen) **fail on the very first turn** — there is no working cache to preserve.
- `DynamicContextMiddleware`'s original design (SystemMessage injected mid-conversation) already broke the static-prefix assumption on strict backends; the coalescing middleware makes the request valid, at the cost of cache churn on the system block.
- The fast path (≤1 `SystemMessage`) returns the same request object with zero mutation, so backends/turns that never trigger coalescing keep full prefix-cache fidelity.
- This mirrors the trade-off already accepted and approved by a maintainer in #3702 (Claude provider), which uses the same coalesce strategy.

### Transform effect

```
# Before (sent to LLM) — 2+ SystemMessages, strict backends reject:
[0] SystemMessage(system_prompt)                  ← create_agent prepend
[1] SystemMessage(id=msg-1, date_reminder)        ← DynamicContextMiddleware
[2] HumanMessage(id=msg-1__memory)
[3] HumanMessage(id=msg-1__user)
[4] AIMessage(...)
[5] SystemMessage(id=msg-2, date_update)          ← midnight crossing
[6] HumanMessage(id=msg-2__user)

# After — single leading SystemMessage:
[0] SystemMessage(system_prompt + date_reminder + date_update)  ← merged
[1] HumanMessage(id=msg-1__memory)
[2] HumanMessage(id=msg-1__user)
[3] AIMessage(...)
[4] HumanMessage(id=msg-2__user)
```

## Surface area

## Bug fix verification

- Test path: `backend/tests/test_system_message_coalescing_middleware.py`
  - `TestCoalesceHelper` (9 tests) — pure `_coalesce_system_messages` helper: no/one/multiple system messages, content merging with `\n\n` separator, id preservation, non-system message ordering, empty list.
  - `TestWrapModelCall` (4 tests) — sync path: passthrough on fast path, `request.override` on coalesce path, no-system passthrough, handler result return.
  - `TestAwrapModelCall` (2 tests) — async path: passthrough + coalesce.
  - `TestRealisticScenario` (2 tests) — end-to-end simulating the exact #3707 trigger:
    - `test_first_turn_with_dynamic_context_injection` — first-turn ID-swap + `create_agent` prepend → 2 SystemMessages coalesced to 1, user-owned memory stays `HumanMessage` (OWASP LLM01 preserved).
    - `test_midnight_crossing_scenario` — 3 SystemMessages (prompt + reminder + date_update) coalesced to 1, all non-system messages keep original order.
- Did it go red on `main` and green on this branch? yes — on `main` the realistic-scenario tests fail (request keeps 2+ `SystemMessage`s); on this branch all 17 tests pass.
- Note: the tests are newly added in this PR (they did not exist on `main`), so "red on main" means the bug behavior is present, not that a pre-existing test failed.

## Validation

All checks ran in the `deer-flow-gateway` Docker container (Python 3.12.13, Linux x86_64).

### Lint / format

```
$ make lint
uvx ruff check .
All checks passed!
uvx ruff format --check .
589 files already formatted
```

### Test results

**Middleware + downstream consumers (267 passed):**
```
$ pytest tests/test_system_message_coalescing_middleware.py \
          tests/test_dynamic_context_middleware.py \
          tests/test_dangling_tool_call_middleware.py \
          tests/test_loop_detection_middleware.py \
          tests/test_input_sanitization_middleware.py \
          tests/test_clarification_middleware.py \
          tests/test_safety_finish_reason_middleware.py \
          tests/test_harness_boundary.py \
          tests/test_memory_upload_filtering.py -v
...
267 passed, 1 warning in 16.03s
```

Covers: the new middleware itself, every middleware whose `wrap_model_call` could interact with message ordering, the memory builder (state-scanning consumer), and the harness boundary.

**Broader regression suite (298 passed):**
```
$ pytest tests/test_create_deerflow_agent.py \
          tests/test_subagent_executor.py \
          tests/test_lead_agent_prompt.py \
          tests/test_model_factory.py \
          tests/test_claude_provider_prompt_caching.py \
          tests/test_token_budget_middleware.py \
          tests/test_tool_output_budget_middleware.py \
          tests/test_title_middleware_core_logic.py -q
...
298 passed, 1 warning in 14.71s
```

Covers: agent factory (middleware registration), subagent path (already self-coalesces `system_prompt` + skills — confirms no double-coalescing conflict), lead agent prompt, model factory, Claude provider prompt caching (confirms #3702's per-provider path becomes a no-op once the generic middleware runs), token/title/tool-output budget middleware.

**Total: 565 passed, 0 failed, 0 regressions.**

## AI assistance

**Tool(s) used:** VSCode(GLM-5.2)

**How you used it:** AI traced the root cause by inspecting `create_agent` source and verifying `add_messages` reducer behavior in-container (confirmed ID-swap does not reorder messages — only the multiple-SystemMessage issue is the bug). AI wrote the middleware, tests, and validated in Docker. Human reviewed every line and confirmed the root-cause analysis against `create_agent` source.

### DIFF
--- a/backend/packages/harness/deerflow/agents/lead_agent/agent.py
+++ b/backend/packages/harness/deerflow/agents/lead_agent/agent.py
@@ -348,6 +348,13 @@ def build_middlewares(
 
         middlewares.append(DeferredToolFilterMiddleware(deferred_setup.deferred_names, deferred_setup.catalog_hash))
 
+    # Coalesce every SystemMessage into a single leading one before the request
+    # reaches the provider. Strict backends (vLLM, SGLang, Qwen, Anthropic)
+    # reject non-leading SystemMessages. See system_message_coalescing_middleware.py.
+    from deerflow.agents.middlewares.system_message_coalescing_middleware import SystemMessageCoalescingMiddleware
+
+    middlewares.append(SystemMessageCoalescingMiddleware())
+
     # Add SubagentLimitMiddleware to truncate excess parallel task calls
     subagent_enabled = cfg.get("subagent_enabled", False)
     if subagent_enabled:
--- a/backend/packages/harness/deerflow/agents/middlewares/system_message_coalescing_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/system_message_coalescing_middleware.py
@@ -1,0 +1,150 @@
+"""Middleware to coalesce multiple SystemMessages into a single leading one.
+
+Strict OpenAI-compatible backends (vLLM, SGLang, Qwen) and Anthropic reject
+non-leading SystemMessages with errors like "System message must be at the
+beginning" or "Received multiple non-consecutive system messages". The
+official OpenAI API tolerates mid-conversation system messages, so the issue
+only surfaces on strict backends.
+
+DeerFlow's lead agent accumulates multiple SystemMessages because
+DynamicContextMiddleware uses the ID-swap technique to replace the first or
+last HumanMessage with a triplet whose first element is a SystemMessage
+reminder (framework-owned date/metadata must not masquerade as user input,
+per OWASP LLM01). On midnight crossings a second SystemMessage (date update)
+is injected. create_agent holds the static system_prompt in the separate
+``request.system_message`` field and only flattens it into the message list
+inside the model-call handler (``[request.system_message, *messages]``).
+
+This middleware runs in wrap_model_call — before the handler flattens the two
+— and merges ``request.system_message`` plus every SystemMessage found in
+``request.messages`` into a single leading SystemMessage emitted via the
+``system_message`` field. It only touches the request payload; the persistent
+conversation state (checkpoint) is unchanged, so middleware that scans history
+by marker (e.g. is_dynamic_context_reminder) keeps working.
+
+Note: Mirrors the per-request coalescing already done for Claude in
+claude_provider._coalesce_system_messages but at a provider-agnostic layer so
+every backend benefits from a single fix instead of per-provider patches.
+"""
+
+from collections.abc import Awaitable, Callable
+from typing import override
+
+from langchain.agents import AgentState
+from langchain.agents.middleware import AgentMiddleware
+from langchain.agents.middleware.types import ModelCallResult, ModelRequest, ModelResponse
+from langchain_core.messages import SystemMessage
+
+from deerflow.agents.middlewares.dynamic_context_middleware import is_dynamic_context_reminder
+
+
+def _flatten_content(content) -> str:
+    """Convert message content to a plain string, handling both str and list types.
+
+    langchain messages support list-type content for multimodal (e.g.
+    ``[{"type": "text", "text": "..."}]``). SystemMessages in DeerFlow are always
+    plain strings, but this helper ensures robustness for any content shape.
+    """
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts = []
+        for item in content:
+            if isinstance(item, str):
+                parts.append(item)
+            elif isinstance(item, dict) and "text" in item:
+                parts.append(item["text"])
+            else:
+                parts.append(str(item))
+        return "\n".join(parts)
+    return str(content)
+
+
+def _coalesce_request(request: ModelRequest) -> ModelRequest | None:
+    """Merge ``request.system_message`` and in-``messages`` SystemMessages into one.
+
+    On langchain >= 1.2.15 the static system prompt lives in the separate
+    ``request.system_message`` field, not in ``request.messages``. The model-call
+    handler flattens them at the very last moment (``[system_message, *messages]``),
+    so a middleware that only scans ``messages`` cannot see the prompt and ends up
+    a no-op. This helper inspects both sources, merges every SystemMessage into a
+    single entry, and emits the result via ``system_message`` so the handler still
+    prepends it correctly.
+
+    Returns None when no SystemMessages live inside ``messages`` — in that case
+    ``system_message`` (if set) is already the sole leading system block and the
+    request can pass through with zero mutation, preserving prefix-cache hits.
+    """
+    in_msg_systems = [m for m in request.messages if isinstance(m, SystemMessage)]
+    if not in_msg_systems:
+        return None
+
+    # Merge system_message (if any) + all in-messages SystemMessages.
+    parts: list[SystemMessage] = []
+    if request.system_message is not None:
+        parts.append(request.system_message)
+    parts.extend(in_msg_systems)
+
+    # Deduplicate dynamic_context_reminder SystemMessages: only keep the last
+    # one (most recent date), drop earlier reminders. On midnight crossings
+    # the merged content would otherwise contain two adjacent contradictory
+    # <current_date> blocks with no temporal anchor — the intervening turns
+    # that originally separated them are gone after coalescing. The model
+    # should see only the latest date, not a stale one it must guess to
+    # ignore.
+    reminder_indices = [i for i, p in enumerate(parts) if is_dynamic_context_reminder(p)]
+    if len(reminder_indices) > 1:
+        keep_last = reminder_indices[-1]
+        parts = [p for i, p in enumerate(parts) if i not in reminder_indices[:-1] or i == keep_last]
+
+    # Preserve the id of the first SystemMessage (typically the static
+    # system_prompt) so downstream consumers that key off the leading system
+    # message id are unaffected. Merge additional_kwargs from all parts so
+    # markers like hide_from_ui / dynamic_context_reminder from reminders are
+    # retained on the coalesced block.
+    first = parts[0]
+    merged_kwargs: dict = {}
+    for p in parts:
+        merged_kwargs.update(p.additional_kwargs or {})
+    merged = SystemMessage(
+        content="\n\n".join(_flatten_content(p.content) for p in parts),
+        id=first.id,
+        additional_kwargs=merged_kwargs,
+    )
+
+    non_system = [m for m in request.messages if not isinstance(m, SystemMessage)]
+    return request.override(system_message=merged, messages=non_system)
+
+
+class SystemMessageCoalescingMiddleware(AgentMiddleware[AgentState]):
+    """Merge all SystemMessages into a single leading SystemMessage.
+
+    Uses wrap_model_call (not before_agent) so the merge runs on the final
+    request payload — where ``system_message`` and ``messages`` are still
+    separate fields — and never touches the persisted state["messages"]. This
+    keeps the checkpoint structure intact for every consumer that scans history
+    (memory builder, journal, summarization, dynamic-context detection).
+    """
+
+    @staticmethod
+    def _maybe_coalesce(request: ModelRequest) -> ModelRequest:
+        coalesced = _coalesce_request(request)
+        if coalesced is None:
+            return request
+        return coalesced
+
+    @override
+    def wrap_model_call(
+        self,
+        request: ModelRequest,
+        handler: Callable[[ModelRequest], ModelResponse],
+    ) -> ModelCallResult:
+        return handler(self._maybe_coalesce(request))
+
+    @override
+    async def awrap_model_call(
+        self,
+        request: ModelRequest,
+        handler: Callable[[ModelRequest], Awaitable[ModelResponse]],
+    ) -> ModelCallResult:
+        return await handler(self._maybe_coalesce(request))
--- a/backend/CLAUDE.md
+++ b/backend/CLAUDE.md
@@ -30,7 +30,7 @@ deer-flow/
 │   │       └── deerflow/
 │   │           ├── agents/            # LangGraph agent system
 │   │           │   ├── lead_agent/    # Main agent (factory + system prompt)
-│   │           │   ├── middlewares/   # 10 middleware components
+│   │           │   ├── middlewares/   # 11 middleware components
 │   │           │   ├── memory/        # Memory extraction, queue, prompts
 │   │           │   └── thread_state.py # ThreadState schema
 │   │           ├── sandbox/           # Sandbox execution system
@@ -218,9 +218,10 @@ Lead-agent middlewares are assembled in strict append order across `packages/har
 14. **MemoryMiddleware** - Queues conversations for async memory update (filters to user + final AI responses)
 15. **ViewImageMiddleware** - Injects base64 image data before LLM call (conditional on vision support)
 16. **DeferredToolFilterMiddleware** - Hides deferred (MCP) tool schemas from the bound model using a build-time deferred-name set + catalog hash, reading per-thread promotions from `ThreadState.promoted` (hash-scoped, no ContextVar); a tool becomes bound on subsequent turns after `tool_search` returns its schema (optional, if `tool_search.enabled`)
-17. **SubagentLimitMiddleware** - Truncates excess `task` tool calls from model response to enforce `MAX_CONCURRENT_SUBAGENTS` limit (optional, if `subagent_enabled`)
-18. **LoopDetectionMiddleware** - Detects repeated tool-call loops; hard-stop responses clear both structured `tool_calls` and raw provider tool-call metadata before forcing a final text answer
-19. **ClarificationMiddleware** - Intercepts `ask_clarification` tool calls, interrupts via `Command(goto=END)` (must be last)
+17. **SystemMessageCoalescingMiddleware** - Merges `request.system_message` and every in-`request.messages` SystemMessage into a single leading SystemMessage in `wrap_model_call`; provider-agnostic fix for strict backends (vLLM/SGLang/Qwen/Anthropic) that reject non-leading system messages. Touches the per-request payload only — checkpoint state is unchanged so `is_dynamic_context_reminder` history scanners keep working. On midnight crossings, earlier `dynamic_context_reminder`-tagged SystemMessages are dropped and only the latest (most recent date) survives, avoiding two contradictory `<current_date>` blocks appearing adjacent with no temporal anchor.
+18. **SubagentLimitMiddleware** - Truncates excess `task` tool calls from model response to enforce `MAX_CONCURRENT_SUBAGENTS` limit (optional, if `subagent_enabled`)
+19. **LoopDetectionMiddleware** - Detects repeated tool-call loops; hard-stop responses clear both structured `tool_calls` and raw provider tool-call metadata before forcing a final text answer
+20. **ClarificationMiddleware** - Intercepts `ask_clarification` tool calls, interrupts via `Command(goto=END)` (must be last)
 
 ### Configuration System
 
--- a/backend/tests/test_system_message_coalescing_middleware.py
+++ b/backend/tests/test_system_message_coalescing_middleware.py
@@ -1,0 +1,681 @@
+"""Tests for SystemMessageCoalescingMiddleware.
+
+Verifies that ``request.system_message`` and in-``messages`` SystemMessages are
+merged into a single leading SystemMessage before the request reaches the LLM,
+fixing the "System message must be at the beginning" error on strict
+OpenAI-compatible backends (vLLM, SGLang, Qwen) and Anthropic.
+
+On langchain >= 1.2.15 the static system prompt lives in the separate
+``request.system_message`` field, not in ``request.messages``. The model-call
+handler flattens them at the very last moment (``[system_message, *messages]``),
+so tests must build requests with the same split to exercise the real code path.
+"""
+
+from unittest.mock import MagicMock
+
+import pytest
+from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, SystemMessage
+
+from deerflow.agents.middlewares.system_message_coalescing_middleware import (
+    SystemMessageCoalescingMiddleware,
+    _coalesce_request,
+    _flatten_content,
+)
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_request(system_message: SystemMessage | None, messages: list[BaseMessage]):
+    """Build a minimal ModelRequest stand-in matching langchain 1.2.15 shape.
+
+    ``system_message`` and ``messages`` are separate fields — the handler
+    flattens them into ``[system_message, *messages]`` only at call time.
+    """
+    request = MagicMock()
+    request.system_message = system_message
+    request.messages = list(messages)
+    request.override = lambda **updates: _override_request(request, updates)
+    return request
+
+
+def _override_request(request, updates):
+    """Mimic ModelRequest.override(): return a copy with fields replaced."""
+    new = MagicMock()
+    new.system_message = updates.get("system_message", request.system_message)
+    new.messages = updates.get("messages", request.messages)
+    new.override = lambda **kw: _override_request(new, kw)
+    return new
+
+
+def _capture_handler():
+    """Return (captured_requests, handler) that records what was sent."""
+    captured = []
+
+    def handler(req):
+        captured.append(req)
+        return "response"
+
+    return captured, handler
+
+
+def _final_payload(request) -> list[BaseMessage]:
+    """Simulate what the LLM receives: [system_message, *messages] (if set)."""
+    if request.system_message is not None:
+        return [request.system_message, *request.messages]
+    return list(request.messages)
+
+
+# ===========================================================================
+# _coalesce_request (pure helper)
+# ===========================================================================
+
+
+class TestCoalesceRequest:
+    def test_no_system_anywhere_returns_none(self):
+        """Zero SystemMessages → passthrough."""
+        request = _make_request(system_message=None, messages=[HumanMessage(content="hi")])
+        assert _coalesce_request(request) is None
+
+    def test_only_system_message_field_returns_none(self):
+        """system_message set, no SystemMessages in messages → passthrough.
+
+        The single system block is already in the right place; coalescing would
+        just create a new object with the same content (zero drift).
+        """
+        request = _make_request(
+            system_message=SystemMessage(content="prompt"),
+            messages=[HumanMessage(content="hi")],
+        )
+        assert _coalesce_request(request) is None
+
+    def test_system_message_plus_one_in_msg_system_coalesces(self):
+        """system_message + 1 in-messages SystemMessage → merge into system_message."""
+        prompt = SystemMessage(content="You are DeerFlow.", id="sys-1")
+        reminder = SystemMessage(content="<system-reminder>date</system-reminder>", id="msg-1")
+        user = HumanMessage(content="Hello", id="msg-1__user")
+        request = _make_request(system_message=prompt, messages=[reminder, user])
+
+        result = _coalesce_request(request)
+        assert result is not None
+        assert result.system_message is not None
+        assert "You are DeerFlow." in result.system_message.content
+        assert "<system-reminder>date</system-reminder>" in result.system_message.content
+        # messages no longer contain any SystemMessage
+        assert not any(isinstance(m, SystemMessage) for m in result.messages)
+        # user message preserved
+        assert any(m.content == "Hello" for m in result.messages)
+
+    def test_system_message_plus_two_in_msg_systems_coalesces(self):
+        """system_message + 2 non-reminder SystemMessages → merge all (no dedup)."""
+        prompt = SystemMessage(content="prompt")
+        reminder = SystemMessage(content="<system-reminder>day1</system-reminder>")
+        date_update = SystemMessage(content="<system-reminder>day2</system-reminder>")
+        user = HumanMessage(content="next")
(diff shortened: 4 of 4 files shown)


backend/packages/harness/deerflow/agents/middlewares/system_message_coalescing_middleware.py
```diff
@@ -0,0 +1,94 @@
...
+    Uses wrap_model_call (not before_agent) so the merge runs on the final
+    request payload — after create_agent prepends system_prompt — and never
+    touches the persisted state["messages"]. This keeps the checkpoint
+    structure intact for every consumer that scans history (memory builder,
+    journal, summarization, dynamic-context detection).
+    """
+
+    @staticmethod
+    def _maybe_coalesce(request: ModelRequest) -> ModelRequest:
+        coalesced = _coalesce_system_messages(request.messages)
```
fancyboi999: Crux of the issue: on `langchain>=1.2.15`, `request.messages` *excludes* the system prompt — it's `request.system_message` (`types.py:97`), prepended only at invocation (`factory.py:1278-1281`). So this passes only the conversation-history SystemMessages (the DynamicContext reminder), never the static prompt, and the merge can't unify them. Coalesce `request.system_message` + the in-`messages` systems and emit via `request.override(system_message=…, messages=<non-system>)`.

backend/packages/harness/deerflow/agents/middlewares/system_message_coalescing_middleware.py
```diff
@@ -0,0 +1,94 @@
...
+def _coalesce_system_messages(messages: list[BaseMessage]) -> list[BaseMessage] | None:
+    """Merge every SystemMessage in *messages* into one leading entry.
+
+    Returns None when the list already has zero or one SystemMessage so
+    callers can short-circuit and avoid mutating the request payload —
+    keeping zero behavior drift on the common path and preserving
+    prefix-cache hits when there is nothing to coalesce.
+    """
+    system_indices = [i for i, m in enumerate(messages) if isinstance(m, SystemMessage)]
+    if len(system_indices) <= 1:
```
fancyboi999: In the real first-turn flow `request.messages` has exactly one SystemMessage (the reminder), so this short-circuits to a no-op — yet the final payload still has two (the model node prepends `request.system_message` ahead of it). The condition should be "no SystemMessages in `messages` AND `request.system_message is None`", and the count must include `request.system_message`.

fancyboi999 (commented):
Thanks for taking #3707 with a provider-agnostic middleware — that's the right layer. But I don't think it lands the fix on `langchain>=1.2.15` (the pinned version): in the real flow this middleware is a no-op, and the tests pass against a request shape `create_agent` never produces. Walking it through:

**The static system prompt isn't in `request.messages`.** On langchain 1.2.15, `ModelRequest.messages` *excludes* the system prompt — it lives in a separate `request.system_message` field (`langchain/agents/middleware/types.py:97`: `messages: list[AnyMessage]  # excluding system message`). `create_agent(system_prompt=...)` (used by the lead agent at `lead_agent/agent.py:510`/`:538`) becomes `system_message`, and the agent node builds the request with `system_message=system_message, messages=state["messages"]` (`factory.py:1298-1303`). The prompt is prepended only at invocation: `messages = request.messages; if request.system_message: messages = [request.system_message, *messages]` (`factory.py:1278-1281`).

**So in the real #3707 flow this is a no-op.** `DynamicContextMiddleware` injects its reminder `SystemMessage` into `state["messages"]`, so on the first turn `request.messages` holds exactly one `SystemMessage` (the reminder); the static prompt is in `request.system_message`, not here. `_coalesce_system_messages(request.messages)` sees `len(system_indices) == 1` → returns `None` → passthrough. The model node then prepends `request.system_message`, yielding `[system_prompt, reminder, …]` — the same two non-leading SystemMessages #3707 reports. On a midnight crossing (two reminders in `messages`) it merges those two, but the framework still prepends `system_message` ahead of the result → still two. Either way the strict backend still 400s.

**The tests are green against a shape that can't occur.** `test_first_turn_with_dynamic_context_injection` builds `request.messages = [SystemMessage("You are DeerFlow…"), SystemMessage(reminder), …]` — the system prompt placed *inside* `messages`. On 1.2.15 it's never in `messages`, so these pass without exercising the real path.

**Fix direction:** coalesce `request.system_message` together with the in-`messages` SystemMessages and emit via `system_message`, not `messages`:
```python
in_msg_systems = [m for m in request.messages if isinstance(m, SystemMessage)]
if not in_msg_systems and request.system_message is None:
    return request
parts = ([request.system_message.content] if request.system_message else []) + [m.content for m in in_msg_systems]
merged = SystemMessage(content="\n\n".join(str(p) for p in parts),
                       id=(request.system_message.id if request.system_message else in_msg_systems[0].id))
non_system = [m for m in request.messages if not isinstance(m, SystemMessage)]
return request.override(system_message=merged, messages=non_system)
```
Then the single system the model node prepends is the fully-merged one. A test that builds the request the way `create_agent` does (`system_message=<prompt>` + a reminder `SystemMessage` in `messages`) and asserts the final `[request.system_message, *request.messages]` has exactly one leading SystemMessage would have caught this.

For reference, the existing Claude fix (`claude_provider._coalesce_system_messages`, #3702) works because it runs in `_get_request_payload`, where langchain has already flattened `system_message + messages` into one list — it sees the prompt. This middleware runs one layer up, where the two are still split.

ech0hol:
@fancyboi999 Thanks for the detailed walkthrough — you're right, the initial version was a no-op on `langchain>=1.2.15` because it only scanned `request.messages` and missed `request.system_message` entirely.

Updated in the latest commit. Changes:

1. **`_coalesce_request` now inspects both sources.** It merges `request.system_message` + all in-`messages` SystemMessages into a single `SystemMessage` emitted via the `system_message` field, and strips SystemMessages from `messages`. This matches your suggested fix direction.

2. **Passthrough condition fixed.** Returns `None` (no mutation) only when there are zero SystemMessages in `messages` — in that case `system_message` is already the sole leading system block. When `system_message` is set but `messages` has no SystemMessages, the request passes through untouched, preserving prefix-cache hits.

3. **Tests rebuilt against the real request shape.** All test requests now use `_make_request(system_message=..., messages=...)` matching the langchain 1.2.15 split, and the realistic scenario tests assert on `_final_payload(request)` — the `[system_message, *messages]` list the LLM actually receives — confirming exactly one leading SystemMessage.

The key insight I missed: `wrap_model_call` runs *before* the handler flattens `system_message` into `messages`, so at this layer the two are still separate fields and both must be inspected. The existing Claude fix (#3702) works because it runs in `_get_request_payload`, where flattening has already happened. This middleware sits one layer up, so it needs to handle the split explicitly.

Appreciate the precise code references (`types.py:97`, `factory.py:1278-1281`) — made the fix straightforward.

fancyboi999:
Verified the new commit (`ce5d3f2a4`) — this resolves it. `_coalesce_request` now pulls `request.system_message` into `parts` alongside the in-`messages` SystemMessages and emits the merge through the `system_message` field, so when the model node flattens `[request.system_message, *request.messages]` the result carries exactly one leading SystemMessage. Walked the two real shapes:

- First turn — `system_message=<prompt>`, `messages=[SystemMessage(reminder), memory, user]` → merges `[prompt, reminder]` into `system_message`, `messages=[memory, user]` → flattened `[merged, memory, user]` = one leading system. ✓
- Midnight crossing (two reminders in `messages`) → both fold into the single `system_message` alongside the prompt. ✓

Passthrough is right too: `None` only when `messages` has no SystemMessages, so a request whose sole system is the already-leading `system_message` is left untouched (prefix cache preserved). And `id` carries the first part's id so leading-system-id consumers are unaffected. This closes #3707 for the strict OpenAI-compatible backends (vLLM/SGLang/Qwen), not just Claude.

willem-bd (commented):
## Review summary

The architecture is right — a provider-agnostic coalescing layer in `wrap_model_call` that leaves checkpointed state untouched — but **as written this does not fix #3707**, because the static `system_prompt` is not in `request.messages`.

**Root issue (see inline):** `create_agent` stores the prompt in the **separate `request.system_message` field**, not in `request.messages`. The prepend into the message list happens *inside* `_execute_model_sync` (the handler this middleware calls) — i.e. *after* the middleware has already run. So `_coalesce_system_messages(request.messages)` only ever sees the DynamicContext reminder(s), never the prompt.

Runtime consequence:
- **First turn:** `request.messages` has one reminder `SystemMessage` → `len <= 1` → no-op → `_execute_model_sync` still prepends `system_prompt` → LLM gets `[SystemMessage(prompt), SystemMessage(reminder), …]` → 2 system messages → strict backend still 400s.
- **Midnight crossing:** two reminders coalesce to one → prompt still prepended → still 2.

Evidence (`langchain/agents/factory.py`, installed in `backend/.venv`):
- `model_node`: `ModelRequest(..., system_message=system_message, messages=state['messages'])` then `wrap_model_call_handler(request, _execute_model_sync)`.
- `_execute_model_sync`: `messages = request.messages; if request.system_message: messages = [request.system_message, *messages]`.
- Lead agent passes the prompt this way: `agent.py:510` / `:538` → `create_agent(..., system_prompt=apply_prompt_template(...), ...)`.

**The tests don't catch it** because they build the request by hand and put `system_prompt` as the first element of `request.messages` (e.g. `test_first_turn_with_dynamic_context_injection`), with a docstring claiming this is what create_agent passes to wrap_model_call after prepending — which the source above contradicts. So the helper is validated on an input shape that never occurs in production, and the 565-passing suite never drives a real strict backend through the real `create_agent` request flow.

**Suggested direction:** merge `request.system_message` *together with* the `SystemMessage`s in `request.messages` into one, write the merged result back into `request.system_message`, and strip the `SystemMessage`s from `request.messages`. Then add a test that builds the request the way `create_agent` does (prompt in `system_message`, not `messages`) and asserts against a strict-backend stub.

**Minor nits (non-blocking):** `additional_kwargs` taken only from the first system message (`:56`); `str(content)` assumes string content (`:55`).

**What's good:** sound design rationale (provider-agnostic, keep `SystemMessage` semantics / OWASP LLM01, preserve first id, fast-path `None`); clean docs; no risk to subagents (`executor.py:367-373` passes `system_prompt=None`).

ech0hol:
@WillemJiang Thanks for the thorough review. I want to clarify the timeline — the core issue you identified was actually fixed in the **first** force push, before this review was posted. The second force push addresses the non-blocking nits and adds the strict-backend stub tests you suggested.

**First push: core fix already in place**

The v1 rewrite already handles the `request.system_message` vs `request.messages` split. `_coalesce_request` merges `request.system_message` (if set) together with every `SystemMessage` in `request.messages`, writes the merged result back via `request.override(system_message=merged, messages=non_system)`, and strips `SystemMessage`s from `messages`. This is the fix you described in your review — it was already there.

The tests also already use the real request shape (`system_message` as a separate field, not stuffed into `messages`). The `TestRealisticScenario` class simulates the exact #3707 trigger: system prompt in `system_message`, DynamicContext reminder in `messages`.

**Second push: non-blocking nits + strict-backend stub**

This push fixes the two nits you flagged and implements the suggested direction:

Nit 1: `additional_kwargs` only from first SystemMessage → fixed. Now merges all parts' kwargs via `dict.update()` so markers like `hide_from_ui` and `dynamic_context_reminder` from reminders are retained. When two parts share a key, the later part's value wins. New tests: `test_merged_kwargs_combine_all_parts`, `test_merged_kwargs_later_parts_override`.

Nit 2: `str(content)` assumes string content → fixed. New `_flatten_content` helper handles langchain's list-type content (`[{"type": "text", "text": "..."}]`) by extracting text from dicts instead of producing `"['item1', 'item2']"`. New test class: `TestFlattenContent` (5 tests).

Strict-backend stub assertions → implemented per your suggestion. `TestStrictBackendStub` builds requests the way `create_agent` does and asserts against a stub that simulates Qwen/SGLang/vLLM rejection behavior (accepts exactly 0 or 1 `SystemMessage` at position 0, rejects otherwise):

| Test | Scenario | Expected |
|---|---|---|
| `test_first_turn_without_middleware_rejects` | 2 SystemMessages, no middleware | `StrictBackendError` (reproduces #3707) |
| `test_first_turn_with_middleware_accepts` | same input, middleware applied | `"ok"` |
| `test_midnight_crossing_without_middleware_rejects` | 3 SystemMessages, no middleware | `StrictBackendError` |
| `test_midnight_crossing_with_middleware_accepts` | same input, middleware applied | `"ok"` |
| `test_async_path_with_middleware_accepts` | async path + strict stub | `"ok"` |
| `test_clean_request_no_middleware_needed` | 1 SystemMessage, no middleware needed | `"ok"` |

Test count: 31 total (17 from first push + 14 from second push). All pass. `ruff check` + `ruff format --check` clean.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -348,6 +348,13 @@ def build_middlewares(
 
         middlewares.append(DeferredToolFilterMiddleware(deferred_setup.deferred_names, deferred_setup.catalog_hash))
 
+    # Coalesce every SystemMessage into a single leading one before the request
+    # reaches the provider. Strict backends (vLLM, SGLang, Qwen, Anthropic)
+    # reject non-leading SystemMessages. See system_message_coalescing_middleware.py.
+    from deerflow.agents.middlewares.system_message_coalescing_middleware import SystemMessageCoalescingMiddleware
```
willem-bd, line 354: **CLAUDE.md not updated for the new chain entry.**

`backend/CLAUDE.md` → `### Middleware Chain` (lines 202–224) enumerates the lead-agent middleware chain as numbered items 1–19, and the file's own *Documentation Update Policy* says:

> **CRITICAL: Always update README.md and CLAUDE.md after every code change** — *Update `CLAUDE.md` for development changes (architecture, commands, workflows, internal systems)*.

This PR appends `SystemMessageCoalescingMiddleware` to `build_middlewares` (registered after `DeferredToolFilterMiddleware`, before `SubagentLimitMiddleware`) but the canonical list in CLAUDE.md is not updated. A maintainer reading the docs will get a stale picture of the chain.

Suggest inserting a new entry between current items 16 (`DeferredToolFilterMiddleware`) and 17 (`SubagentLimitMiddleware`), e.g.:

> **SystemMessageCoalescingMiddleware** — Merges `request.system_message` and every in-`request.messages` `SystemMessage` into a single leading `SystemMessage` in `wrap_model_call`; provider-agnostic fix for strict backends (vLLM/SGLang/Qwen/Anthropic) that reject non-leading system messages. Touches the per-request payload only — checkpoint state is unchanged so `is_dynamic_context_reminder` history scanners keep working.

backend/packages/harness/deerflow/agents/middlewares/system_message_coalescing_middleware.py
```diff
@@ -0,0 +1,136 @@
...
+    # message id are unaffected. Merge additional_kwargs from all parts so
+    # markers like hide_from_ui / dynamic_context_reminder from reminders are
+    # retained on the coalesced block.
+    first = parts[0]
+    merged_kwargs: dict = {}
+    for p in parts:
+        merged_kwargs.update(p.additional_kwargs or {})
+    merged = SystemMessage(
+        content="\n\n".join(_flatten_content(p.content) for p in parts),
+        id=first.id,
```
willem-bd, line 111: **Midnight-crossing reminders are coalesced in encounter order, producing two adjacent contradictory `<current_date>` blocks.**

`_coalesce_request` joins all `SystemMessage` bodies with `\n\n` and separates them from the intervening `HumanMessage`/`AIMessage` turns. On the midnight-crossing payload:

```
[SystemMessage(<reminder day1>), Human, AI, SystemMessage(<reminder day2>), Human]
```

The merged leading system block becomes:

```
<system_prompt>

<system-reminder><current_date>day1</current_date></system-reminder>

<system-reminder><current_date>day2</current_date></system-reminder>
```

followed by the bare turns. The model now sees **two contradictory `<current_date>` reminders adjacent at the top** with no temporal anchor — the intermediate Human/AI turns that originally separated them are gone. Most LLMs will pick the last value seen, but the behaviour is fragile and not free: it depends on the model's recency bias rather than an explicit signal.

The PR description's *Trade-off* section only discusses prefix-cache invalidation; this temporal-ordering side-effect is worth calling out, and ideally fixing — e.g. by detecting reminder blocks (already marked with `additional_kwargs['dynamic_context_reminder']=True`) and keeping only the last `<current_date>` payload during coalescing, or by appending a normalized "current date is X" tail derived from the latest reminder rather than concatenating raw bodies in encounter order.

The existing test `test_midnight_crossing_single_system_message_in_final_payload` asserts both `day1` and `day2` survive in the merged content, so this is a behaviour worth a deliberate test, not an incidental one.

willem-bd (commented):
Two follow-ups from a code review — the diff is clean and the `_flatten_content` helper already handles list-content `SystemMessage`s correctly (good defense-in-depth). Inline comments below.

ech0hol:
@WillemJiang Thanks for the thorough review. Pushed v3 (`d60e717`) that addresses both inline comments.

**1. CLAUDE.md updated** → fixed

Inserted `SystemMessageCoalescingMiddleware` as item 17 between `DeferredToolFilterMiddleware` (16) and `SubagentLimitMiddleware` (18), renumbering subsequent entries to 18–20. Also updated the middleware count from 10 to 11 in the directory tree. Description includes the midnight-crossing dedup behavior.

**2. Midnight-crossing contradictory `<current_date>` blocks** → fixed

You're right — concatenating raw bodies in encounter order leaves two adjacent date blocks with no temporal anchor, relying on the model's recency bias. Not great.

v3 adds dedup logic in `_coalesce_request`: when multiple SystemMessages carry `dynamic_context_reminder=True`, only the last one (most recent date) survives; earlier reminders are dropped. Non-reminder SystemMessages are never deduplicated.

```python
reminder_indices = [i for i, p in enumerate(parts) if is_dynamic_context_reminder(p)]
if len(reminder_indices) > 1:
    keep_last = reminder_indices[-1]
    parts = [p for i, p in enumerate(parts) if i not in reminder_indices[:-1] or i == keep_last]
```

This uses the same `is_dynamic_context_reminder` check that the summarization and memory builder already rely on, so it's consistent with existing marker-based filtering throughout the codebase.

Updated test `test_midnight_crossing_single_system_message_in_final_payload` now asserts `day2` is present and `day1` is absent. New dedicated tests:
- `test_reminder_dedup_keeps_only_last` — multiple reminders, only last survives
- `test_reminder_dedup_does_not_affect_non_reminder_systems` — non-reminder SystemMessages untouched
- `test_single_reminder_not_deduplicated` — single reminder kept as-is

The CLAUDE.md entry also documents this behavior: *"On midnight crossings, earlier `dynamic_context_reminder`-tagged SystemMessages are dropped and only the latest (most recent date) survives, avoiding two contradictory `<current_date>` blocks appearing adjacent with no temporal anchor."*

WillemJiang (approved):

PR #3722, opened 2026-06-23
Lock sandbox provider lifecycle

### LINKED ISSUE #3721: [harness][sandbox] get_sandbox_provider() check-then-create can double-initialize the provider across threads
## Summary

`get_sandbox_provider()` lazily initializes the process-wide default sandbox provider with an unsynchronized check-then-create. When it is called from two OS threads concurrently during cold start (the main event-loop thread and the Feishu channel thread, which runs its own thread + event loop), both can observe `_default_sandbox_provider is None` and each construct a provider, with the second assignment overwriting the first.

With the default `LocalSandboxProvider` this is benign (the discarded instance is just garbage). With `AioSandboxProvider` it is more serious: the overwritten instance's `__init__` has already started a background idle-checker thread and registered `atexit`/signal handlers, so the discarded instance leaks an orphan thread that no one can later `shutdown()`.

This is a latent issue under the default configuration; filing it mainly so the double-checked-locking gap is closed before more providers with side-effectful constructors are added. Priority is the maintainers' call.

## Where

```python
# backend/packages/harness/deerflow/sandbox/sandbox_provider.py:57-62
global _default_sandbox_provider
if _default_sandbox_provider is None:          # :58  check
    config = get_app_config()                  # :59  (may do file I/O / stat on first call)
    cls = resolve_class(config.sandbox.use, SandboxProvider)  # :60 (may import a module)
    _default_sandbox_provider = cls(**kwargs)   # :61-62  construct + assign
return _default_sandbox_provider
```

There is no lock here. Contrast with the sibling `get_memory_storage()`, which already uses the correct double-checked pattern:

```python
# backend/packages/harness/deerflow/agents/memory/storage.py:193-202
_storage_lock = threading.Lock()
def get_memory_storage() -> MemoryStorage:
    ...
    with _storage_lock:
        ...
```

## Why two threads can race here

`get_sandbox_provider()` is reached from the main event loop (e.g. `sandbox/middleware.py`, `sandbox/tools.py`, `uploads.py`) **and** from the Feishu channel, which deliberately runs on its own OS thread with its own event loop:

```python
# backend/app/channels/feishu.py
self._thread = threading.Thread(...)   # :131
loop = asyncio.new_event_loop()        # :152
...
sandbox_provider = get_sandbox_provider()  # :382  (runs on the Feishu thread)
```

Inside the check-then-create gap there are GIL yield points — `get_app_config()` (`:59`, may `open()`/`stat()` the config on first use) and `resolve_class()` -> `importlib.import_module` (`:60`, holds the import lock). So when both threads first reach `get_sandbox_provider()` during cold start, both can pass the `is None` check and each construct a provider.

## Interleaving (Aio path, where the consequence is real)

```
time   main event-loop thread                  Feishu channel thread (own loop)
----   ------------------------------------     ------------------------------------
t0     get_sandbox_provider()                   get_sandbox_provider()
t1     :58 _default is None? -> True            :58 _default is None? -> True   (both pass)
t2     :59 get_app_config()/import (GIL yield) -->
t3                                              :61 cls() -> AioProvider #2
                                                   -> __init__ starts idle-checker thread #2,
                                                      atexit.register, signal handlers
t4     :61 cls() -> AioProvider #1
          -> __init__ starts idle-checker thread #1
t5     :62 _default = #1                        :62 _default = #2   (overwrites #1)
----   #1 is now unreachable, but idle-checker thread #1 keeps running = orphan thread ----
```

Relevant `AioSandboxProvider.__init__` side effects:

```python
# backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
atexit.register(self.shutdown)        # :113
self._register_signal_handlers()      # :114
self._reconcile_orphans()             # :117
self._start_idle_checker()            # :121  -> threading.Thread(...).start()  :310-315
```

## Impact

- Default `LocalSandboxProvider` (`config.example.yaml:560`): benign — the discarded wrapper has no background resources.
- `AioSandboxProvider` (commented out by default, `config.example.yaml:586`): the overwritten instance leaks a running `_idle_checker_thread` that nothing holds a reference to (so its normal `shutdown()` path is unreachable), plus duplicate `atexit`/signal-handler registration.

## Fix

Add a `threading.Lock` double-checked guard, mirroring `get_memory_storage()`:

```python
_provider_lock = threading.Lock()

def get_sandbox_provider(**kwargs) -> SandboxProvider:
    global _default_sandbox_provider
    if _default_sandbox_provider is None:
        with _provider_lock:
            if _default_sandbox_provider is None:
                config = get_app_config()
                cls = resolve_class(config.sandbox.use, SandboxProvider)
                _default_sandbox_provider = cls(**kwargs)
    return _default_sandbox_provider
```

## Environment

- File: `backend/packages/harness/deerflow/sandbox/sandbox_provider.py` (line numbers verified against current `main` at audit time).
- Reachable when a Feishu channel is enabled and the first sandbox-provider acquisition races between the main thread and the Feishu thread during process startup. Severe consequence requires `sandbox.use = AioSandboxProvider` (not the default).

Found by @LittleS321, @hemnd, @Czile-create and Claude Code via formal concurrency analysis.

Bortlesboat:
## Summary
- add a shared lifecycle lock around the sandbox provider singleton
- guard creation, reset, shutdown, and explicit test injection with the same lock
- add a concurrent-access regression test that previously created multiple providers

Closes #3721

## Root cause
`get_sandbox_provider()` used an unsynchronized check-then-create flow. Concurrent cold-start callers could all observe `_default_sandbox_provider is None`, construct separate provider instances, and then overwrite the global. For providers with constructor side effects, the losing instances can leave orphaned background resources.

## Validation
- `uv run pytest tests/test_sandbox_provider_lifecycle.py` -> passed
- `uv run ruff check tests/test_sandbox_provider_lifecycle.py packages/harness/deerflow/sandbox/sandbox_provider.py` -> passed
- `uv run pytest tests/test_sandbox_provider_lifecycle.py tests/test_local_sandbox_encoding.py tests/test_sandbox_tools_security.py --basetemp ./.pytest-tmp` -> 128 passed, 1 upstream dependency warning

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/sandbox_provider.py
+++ b/backend/packages/harness/deerflow/sandbox/sandbox_provider.py
@@ -1,4 +1,5 @@
 import asyncio
+import threading
 from abc import ABC, abstractmethod
 
 from deerflow.config import get_app_config
@@ -55,6 +56,7 @@ class SandboxProvider(ABC):
 
 
 _default_sandbox_provider: SandboxProvider | None = None
+_provider_lock = threading.RLock()
 
 
 def get_sandbox_provider(**kwargs) -> SandboxProvider:
@@ -68,9 +70,11 @@ def get_sandbox_provider(**kwargs) -> SandboxProvider:
     """
     global _default_sandbox_provider
     if _default_sandbox_provider is None:
-        config = get_app_config()
-        cls = resolve_class(config.sandbox.use, SandboxProvider)
-        _default_sandbox_provider = cls(**kwargs)
+        with _provider_lock:
+            if _default_sandbox_provider is None:
+                config = get_app_config()
+                cls = resolve_class(config.sandbox.use, SandboxProvider)
+                _default_sandbox_provider = cls(**kwargs)
     return _default_sandbox_provider
 
 
@@ -90,8 +94,9 @@ def reset_sandbox_provider() -> None:
     Use `shutdown_sandbox_provider()` for proper cleanup.
     """
     global _default_sandbox_provider
-    if _default_sandbox_provider is not None:
-        _default_sandbox_provider.reset()
+    with _provider_lock:
+        if _default_sandbox_provider is not None:
+            _default_sandbox_provider.reset()
         _default_sandbox_provider = None
 
 
@@ -103,10 +108,11 @@ def shutdown_sandbox_provider() -> None:
     is shutting down or when you need to completely reset the sandbox system.
     """
     global _default_sandbox_provider
-    if _default_sandbox_provider is not None:
-        if hasattr(_default_sandbox_provider, "shutdown"):
-            _default_sandbox_provider.shutdown()
-        _default_sandbox_provider = None
+    with _provider_lock:
+        if _default_sandbox_provider is not None:
+            if hasattr(_default_sandbox_provider, "shutdown"):
+                _default_sandbox_provider.shutdown()
+            _default_sandbox_provider = None
 
 
 def set_sandbox_provider(provider: SandboxProvider) -> None:
@@ -118,4 +124,5 @@ def set_sandbox_provider(provider: SandboxProvider) -> None:
         provider: The SandboxProvider instance to use.
     """
     global _default_sandbox_provider
-    _default_sandbox_provider = provider
+    with _provider_lock:
+        _default_sandbox_provider = provider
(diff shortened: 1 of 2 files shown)


Bortlesboat:
Closing as superseded by merged #3730. Current main now synchronizes all singleton lifecycle access, constructs and shuts down plugin callbacks outside the lock to avoid reentrant or slow-callback deadlocks, reconciles concurrent cold-start winners, and cleans up losing instances. That covers #3721 more completely than this draft.

PR #3729, opened 2026-06-23
fix: align sandbox artifact mounts with channel user

### LINKED ISSUE #3724: Bug: Artifact file delivery fails due to user_id mismatch between sandbox mount and artifact resolution
## Description

When an agent generates an artifact file (e.g., .md) inside the sandbox and attempts to deliver it via an IM channel (Feishu), the file delivery fails with:

```
[Manager] artifact not found on disk: /mnt/user-data/outputs/xxx.md -> .../users/{safe_user_id}/.../outputs/xxx.md
```

The file actually exists under `users/default/...`.

## Root Cause

Two code paths resolve user_id from different sources:

1. **Sandbox mount creation** (`aio_sandbox_provider.py:313`): uses `get_effective_user_id()` which returns `"default"` when ContextVar is unset (background/dispatcher task context). So sandbox writes to `users/default/...`.

2. **Artifact delivery** (`manager.py:596-604`): uses `user_id` from `_channel_storage_user_id(msg)` which returns the sanitized platform user ID (e.g. `ou_xxxx...`). So manager looks for the file under `users/{safe_platform_user_id}/...`.

The sandbox writes to one path but the manager reads from another.

## Environment
- DeerFlow version: latest (2026-06-22)
- Channel: Feishu (likely affects all IM channels with auth enabled)
- Backend: local container sandbox (non-provisioner mode)

## Suggested Fix

Pass the user_id from run_context to the sandbox provider instead of relying on get_effective_user_id(). Or add a fallback in _resolve_attachments() to check the default user path.

Huixin615:
Closes #3724

## Why

IM channel artifact delivery can fail when an agent writes output files inside the sandbox and the manager later tries to deliver them as attachments.

The bug happens because the sandbox write side and artifact read side can resolve different user buckets for unbound auth-enabled channel messages:

- artifact read / run identity uses `_channel_storage_user_id(msg)`, which falls back to the sanitized platform user id
- sandbox mounts used `get_effective_user_id()`, which falls back to `default` when the ContextVar is unset

That makes the sandbox write artifacts under `users/default/...`, while delivery looks under `users/{safe_platform_user_id}/...`.

## What changed

- Sandbox acquisition now accepts an explicit `user_id`.
- Sandbox middleware and lazy sandbox tools pass the resolved runtime user into sandbox acquisition.
- AIO sandbox mounts, lock paths, deterministic sandbox ids, and per-thread caches are scoped by `(user_id, thread_id)`.
- Local sandbox path mappings and caches are also scoped by `(user_id, thread_id)`.
- Feishu file sync and upload-router sandbox sync pass the same effective user used for storage.
- Remote provisioner sandbox creation now prefers the explicit `user_id`.
- No read-side fallback to `users/default` was added, preserving user isolation.

## Surface area

## Screenshots / Recording

N/A. Backend sandbox/file-delivery fix only.

## Bug fix verification

- `backend/tests/test_aio_sandbox_provider.py`
- `backend/tests/test_local_sandbox_virtual_path_contract.py`
- `backend/tests/test_sandbox_middleware.py`
- `backend/tests/test_feishu_parser.py`
- `backend/tests/test_uploads_router.py`

Yes. The new/updated tests pin that sandbox acquisition and mounts use the resolved runtime/channel user instead of falling back to the ambient ContextVar/default bucket.

## Validation

```bash
cd backend
uv run pytest tests/test_sandbox_middleware.py tests/test_aio_sandbox_provider.py tests/test_local_sandbox_virtual_path_contract.py tests/test_uploads_router.py tests/test_feishu_parser.py tests/test_channels.py tests/test_sandbox_orphan_reconciliation.py
uv run ruff check app/channels packages/harness/deerflow tests/test_sandbox_middleware.py tests/test_aio_sandbox_provider.py tests/test_local_sandbox_virtual_path_contract.py tests/test_uploads_router.py tests/test_feishu_parser.py tests/test_channels.py tests/test_sandbox_orphan_reconciliation.py
uv run ruff format --check app/channels packages/harness/deerflow tests/test_sandbox_middleware.py tests/test_aio_sandbox_provider.py tests/test_local_sandbox_virtual_path_contract.py tests/test_uploads_router.py tests/test_feishu_parser.py tests/test_channels.py tests/test_sandbox_orphan_reconciliation.py

Results:
364 passed
ruff check: All checks passed!
ruff format --check: 265 files already formatted

### DIFF
--- a/backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
+++ b/backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
@@ -133,8 +133,8 @@ class AioSandboxProvider(SandboxProvider):
         self._lock = threading.Lock()
         self._sandboxes: dict[str, AioSandbox] = {}  # sandbox_id -> AioSandbox instance
         self._sandbox_infos: dict[str, SandboxInfo] = {}  # sandbox_id -> SandboxInfo (for destroy)
-        self._thread_sandboxes: dict[str, str] = {}  # thread_id -> sandbox_id
-        self._thread_locks: dict[str, threading.Lock] = {}  # thread_id -> in-process lock
+        self._thread_sandboxes: dict[tuple[str, str], str] = {}  # (user_id, thread_id) -> sandbox_id
+        self._thread_locks: dict[tuple[str, str], threading.Lock] = {}  # (user_id, thread_id) -> in-process lock
         self._last_activity: dict[str, float] = {}  # sandbox_id -> last activity timestamp
         # Warm pool: released sandboxes whose containers are still running.
         # Maps sandbox_id -> (SandboxInfo, release_timestamp).
@@ -276,22 +276,30 @@ class AioSandboxProvider(SandboxProvider):
     # ── Deterministic ID ─────────────────────────────────────────────────
 
     @staticmethod
-    def _deterministic_sandbox_id(thread_id: str) -> str:
-        """Generate a deterministic sandbox ID from a thread ID.
+    def _effective_acquire_user_id(user_id: str | None) -> str:
+        return user_id or get_effective_user_id()
 
-        Ensures all processes derive the same sandbox_id for a given thread,
-        enabling cross-process sandbox discovery without shared memory.
+    @staticmethod
+    def _thread_key(thread_id: str, user_id: str) -> tuple[str, str]:
+        return (user_id, thread_id)
+
+    @staticmethod
+    def _deterministic_sandbox_id(thread_id: str, user_id: str) -> str:
+        """Generate a deterministic sandbox ID from user/thread scope.
+
+        Includes user_id so a previously-created default-bucket sandbox cannot be
+        reused for an auth/channel run that should mount a user-scoped bucket.
         """
-        return hashlib.sha256(thread_id.encode()).hexdigest()[:8]
+        return hashlib.sha256(f"{user_id}:{thread_id}".encode()).hexdigest()[:8]
 
     # ── Mount helpers ────────────────────────────────────────────────────
 
-    def _get_extra_mounts(self, thread_id: str | None) -> list[tuple[str, str, bool]]:
+    def _get_extra_mounts(self, thread_id: str | None, *, user_id: str | None = None) -> list[tuple[str, str, bool]]:
         """Collect all extra mounts for a sandbox (thread-specific + skills)."""
         mounts: list[tuple[str, str, bool]] = []
 
         if thread_id:
-            mounts.extend(self._get_thread_mounts(thread_id))
+            mounts.extend(self._get_thread_mounts(thread_id, user_id=user_id))
             logger.info(f"Adding thread mounts for thread {thread_id}: {mounts}")
 
         skills_mount = self._get_skills_mount()
@@ -302,7 +310,7 @@ class AioSandboxProvider(SandboxProvider):
         return mounts
 
     @staticmethod
-    def _get_thread_mounts(thread_id: str) -> list[tuple[str, str, bool]]:
+    def _get_thread_mounts(thread_id: str, *, user_id: str | None = None) -> list[tuple[str, str, bool]]:
         """Get volume mounts for a thread's data directories.
 
         Creates directories if they don't exist (lazy initialization).
@@ -310,16 +318,16 @@ class AioSandboxProvider(SandboxProvider):
         mounted Docker socket (DooD), the host Docker daemon can resolve the paths.
         """
         paths = get_paths()
-        user_id = get_effective_user_id()
-        paths.ensure_thread_dirs(thread_id, user_id=user_id)
+        effective_user_id = AioSandboxProvider._effective_acquire_user_id(user_id)
+        paths.ensure_thread_dirs(thread_id, user_id=effective_user_id)
 
         return [
-            (paths.host_sandbox_work_dir(thread_id, user_id=user_id), f"{VIRTUAL_PATH_PREFIX}/workspace", False),
-            (paths.host_sandbox_uploads_dir(thread_id, user_id=user_id), f"{VIRTUAL_PATH_PREFIX}/uploads", False),
-            (paths.host_sandbox_outputs_dir(thread_id, user_id=user_id), f"{VIRTUAL_PATH_PREFIX}/outputs", False),
+            (paths.host_sandbox_work_dir(thread_id, user_id=effective_user_id), f"{VIRTUAL_PATH_PREFIX}/workspace", False),
+            (paths.host_sandbox_uploads_dir(thread_id, user_id=effective_user_id), f"{VIRTUAL_PATH_PREFIX}/uploads", False),
+            (paths.host_sandbox_outputs_dir(thread_id, user_id=effective_user_id), f"{VIRTUAL_PATH_PREFIX}/outputs", False),
             # ACP workspace: read-only inside the sandbox (lead agent reads results;
             # the ACP subprocess writes from the host side, not from within the container).
-            (paths.host_acp_workspace_dir(thread_id, user_id=user_id), "/mnt/acp-workspace", True),
+            (paths.host_acp_workspace_dir(thread_id, user_id=effective_user_id), "/mnt/acp-workspace", True),
         ]
 
     @staticmethod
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

**Comment by Huixin615** [comment]
@fancyboi999 哈喽 辛苦帮忙review一下

**Review by fancyboi999 — COMMENTED** [review]
Reviewed — this correctly closes #3724, and it takes the safe path.

The fix is write-side alignment: the resolved channel user_id is threaded explicitly through the sandbox chain — `acquire(thread_id, user_id=...)` and `_get_thread_mounts(thread_id, *, user_id=...)`, so the outputs/work/uploads mounts are built under the same bucket the manager reads from. That replaces the old behavior where the mount silently resolved the bucket via `get_effective_user_id()` and fell back to `"default"` on the channel-dispatcher path while the artifact read used `_channel_storage_user_id(msg)` → `users/{safe_platform_id}`.

The important call: you did **not** add a read-side fallback to `users/default`. That's the right decision — masking the miss on the read side would risk handing back another user's artifact from the shared `default` bucket in a multi-user auth deployment. Fixing the write identity instead of papering over the read is the correct fix. `resolve_runtime_user_id` keeping `user_id or get_effective_user_id()` also preserves the ContextVar path for callers that don't thread an explicit id (e.g. authenticated HTTP GETs), so no regression there.

Re-keying the deterministic sandbox id / caches by `(user_id, thread_id)` is what prevents a stale `default`-bucket sandbox from being reused for a user-scoped run. One consequence worth noting (not a defect): in-flight sandboxes created on the old id scheme won't be rediscovered after deploy and will be recreated under the new key — expected churn for a correctness fix.

Tests look like real anchors — `test_get_thread_mounts_uses_explicit_user_id` fails on main (the old signature didn't take `user_id`), so it's pinning behavior, not a tautology.

Looks good to me.

File: backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py [hunk 1]
```diff
@@ -276,22 +276,30 @@ def _reconcile_orphans(self) -> None:
...
+    def _thread_key(thread_id: str, user_id: str) -> tuple[str, str]:
+        return (user_id, thread_id)
+
+    @staticmethod
+    def _deterministic_sandbox_id(thread_id: str, user_id: str) -> str:
+        """Generate a deterministic sandbox ID from user/thread scope.
+
+        Includes user_id so a previously-created default-bucket sandbox cannot be
+        reused for an auth/channel run that should mount a user-scoped bucket.
         """
```
willem-bd, line 292: **[P2] Upgrade migration — pre-existing AIO containers are silently orphaned at deploy time.**

Before this PR the deterministic id was `sha256(thread_id)[:8]`, producing container name `deer-flow-sandbox-<8 hex>`. After this PR the hash includes `user_id`: `sha256(f"{user_id}:{thread_id}")[:8]`. The two hashes differ for every active thread.

On the first request after deploy in any active thread:
1. `_reuse_in_process_sandbox` misses (in-memory cache wiped by restart).
2. `_reclaim_warm_pool_sandbox` misses (warm pool empty post-restart).
3. `_backend.discover(sandbox_id)` looks up `deer-flow-sandbox-<new hash>` — **the existing container has the old hash and is not found**.
4. `_create_sandbox` creates a fresh container alongside the orphan.

The orphan container keeps consuming CPU/memory. Warm-pool eviction only fires on `release()`, and an orphan never enters the warm pool (this provider has no handle on it). So nothing cleans the orphan up automatically — operators have to `docker rm` them manually.

For a deployment with N concurrently-active threads, this is a one-time **2N containers** spike at upgrade. Probably tolerable, but worth knowing.

**Suggested mitigations (pick one):**

1. **Document the migration:** add a note in the PR body / release notes telling operators to stop the gateway, `docker ps --filter name=deer-flow-sandbox- -q | xargs docker rm -f`, then start the new version. Cheapest fix.
2. **Adoption at startup:** scan existing `deer-flow-sandbox-*` containers, read their thread_id/user_id from labels or mounts (the labels are set in `local_backend.create`), and adopt them into `_thread_sandboxes` so discovery hits.
3. **Compat probe in `_discover_or_create_with_lock`:** as a fallback, also try `sha256(thread_id)[:8]` for the old format; if a match is found, log a one-shot warning and adopt.

Option 1 is acceptable for an internal-ish deployment surface. Option 2/3 is what you'd want for a SaaS rollout.

backend/packages/harness/deerflow/sandbox/local/local_sandbox_provider.py
```diff
@@ -185,19 +185,46 @@ def _setup_path_mappings(self) -> list[PathMapping]:
...
+
+    @staticmethod
+    def _thread_key(thread_id: str, user_id: str) -> tuple[str, str]:
+        return (user_id, thread_id)
+
+    @staticmethod
+    def _sandbox_id_for_thread(thread_id: str, user_id: str) -> str:
+        if user_id == "default":
+            return f"local:{thread_id}"
+        return f"local:{user_id}:{thread_id}"
```
willem-bd, line 199: **Two-format encoding is back-compat-only and structurally fragile.**

```python
if user_id == "default":
    return f"local:{thread_id}"
return f"local:{user_id}:{thread_id}"
```

This produces two distinct sandbox-id shapes:
- `local:<thread_id>` — when user_id is `"default"` (legacy)
- `local:<user_id>:<thread_id>` — otherwise (new)

`_key_from_sandbox_id` then has to round-trip both, and uses `split(':', 1)` to disambiguate. That's safe **today** because user_id and thread_id are both validated by `[A-Za-z0-9_-]+`, so neither can contain `:`. But:

1. **The encoding couples to a charset constraint that lives in a different file.** A future change in `config/paths.py` that loosens the user_id or thread_id regex (to permit `:` — say, for compatibility with some external identity scheme) will silently mis-parse here, with no test failure until someone hits the specific input.

2. **The `"default"` sentinel is a magic string in the encoding.** A user whose effective id is literally `"default"` (the unauth fallback) shares the `local:<thread_id>` namespace with the legacy pre-PR sandboxes for the same thread. There's no namespace separator between them; the encoding relies on the assumption that no two writers ever produce the same `local:<thread_id>` while meaning different things.

**Suggested cleanup (non-blocking):** always include the user_id segment in the per-thread format:

```python
return f"local:{user_id}:{thread_id}"   # uniform format for ALL per-thread acquires
```

Keep the bare `"local"` id only for the no-thread generic singleton. Then `_key_from_sandbox_id` becomes a single-shape parser, and the encoding is independent of validation regex changes.

The rest of the codebase already handles this — `is_local_sandbox` only checks `startswith('local:')`, and there's no caller that depends on the exact `local:<thread_id>` shape after this PR.

willem-bd (commented):
**Code review (high effort) — 3 findings**

Reviewed in full, net-new over @fancyboi999's existing review. The core fix (threading `user_id` through `acquire()` / `acquire_async()` / mounts / locks / deterministic ids / caches) is correct and consistent — verified by walking each call site and matching write-side `effective_user_id` to read-side `_channel_storage_user_id`. The user_id/thread_id charset (`[A-Za-z0-9_-]+`) makes the new sandbox-id encoding parser-safe for current inputs.

Findings below are about **upgrade migration**, **encoding fragility**, and one **type-annotation drift**. None block merge IMO; the first one is worth a deployment note.

---

### Type annotation drift — `release()` local var (consequence of the diff, body is unchanged so it doesn't fit inline)

**File:** `backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py:930`

The PR re-keys `_thread_sandboxes` to `dict[tuple[str, str], str]` (line 136). The local-var annotation in `release()` was not updated:

```python
thread_ids_to_remove: list[str] = []   # line 930 — annotation is now wrong
...
thread_ids_to_remove = [tid for tid, sid in self._thread_sandboxes.items() if sid == sandbox_id]
```

The comprehension iterates the dict's keys (which are now tuples), so `thread_ids_to_remove` actually holds `list[tuple[str, str]]`. Code is functionally correct (the `del self._thread_sandboxes[tid]` line happily deletes by tuple key), but the annotation is misleading and the variable name is stale — these aren't "thread_ids" anymore but `(user_id, thread_id)` keys.

**Fix:**

```python
thread_keys_to_remove: list[tuple[str, str]] = []
...
thread_keys_to_remove = [key for key, sid in self._thread_sandboxes.items() if sid == sandbox_id]
for key in thread_keys_to_remove:
    del self._thread_sandboxes[key]
```

Cleanup-only. Strict mypy would flag this; ruff's default config doesn't.

---

**Also verified (no findings):**
- Async chain propagates `user_id` consistently (`acquire_async` → `_acquire_internal_async` → `_discover_or_create_with_lock_async` → `_recheck_cached_sandbox` → `_reuse_in_process_sandbox` / `_reclaim_warm_pool_sandbox` / `_create_sandbox_async`).
- `is_local_sandbox()` still accepts the new `local:<user>:<thread>` format via its `startswith('local:')` check.
- `feishu.py:415` check `sandbox_id != 'local'` still works — thread-scoped acquires now always return `local:<user>:<thread>`, never `local`.
- `_uses_thread_data_mounts(provider)` decision and mount user_id are now consistent — same `effective_user_id` drives both the upload directory and the AIO mount.
- Local-sandbox path-mapping `_build_thread_path_mappings` now honors the passed `user_id`, eliminating the #3724 root cause (ContextVar fallback to `default`).
- Remote provisioner POST body prefers explicit `user_id` over the contextvar fallback.
- Test coverage pins the multi-user isolation invariant (`test_same_thread_different_users_are_isolated`).

WillemJiang (approved):

PR #3730, opened 2026-06-23
fix(sandbox): synchronize sandbox provider singleton lifecycle (+ concurrency regression tests)

### LINKED ISSUE #3721: [harness][sandbox] get_sandbox_provider() check-then-create can double-initialize the provider across threads

Czile-create:
## Summary

Adds a single module-level lock around the **entire** sandbox provider singleton lifecycle, plus a concurrent regression test.

- `get_sandbox_provider()` used an unsynchronized check-then-create. Two OS threads — notably the main event loop and the Feishu channel thread, which runs its own loop and calls `get_sandbox_provider()` from `app/channels/feishu.py` — could both observe `_default_sandbox_provider is None`, construct separate providers, and overwrite the global.
- `reset_sandbox_provider()`, `shutdown_sandbox_provider()` and `set_sandbox_provider()` also touched the global without a lock, so a reset/shutdown racing an in-flight create could clear it mid-construction or tear down an instance another thread was about to return.
- For `AioSandboxProvider`, a losing instance is unrecoverable: its `__init__` starts an idle-checker thread and registers atexit/signal handlers, and the only path that joins that thread (`shutdown()`) is reachable only through the reference the winning assignment just overwrote — so the loser leaks an orphan background thread.

Closes #3721.

## Changes

- Add a module-level `threading.Lock` (`_provider_lock`) guarding **all four** lifecycle entry points: double-checked locking in `get_sandbox_provider()`, and `with _provider_lock:` around `reset` / `shutdown` / `set`. This mirrors the existing double-checked pattern in `agents/memory/storage.py` (`get_memory_storage`).
- Add `tests/test_sandbox_provider_lifecycle.py`:
  - 8 threads racing on a cold start, aligned with a `threading.Barrier` so the check-then-create race fires deterministically rather than probabilistically; asserts exactly one provider instance is created.
  - a `reset` racing concurrent `get`s, asserting every returned value is a fully constructed provider (never `None` / half-built).

## Notes on the design

- A plain `Lock` (not `RLock`) is sufficient: the calls made while holding the lock — `get_app_config()`, `resolve_class()`, the provider `__init__`, and `provider.reset()` / `shutdown()` — do not re-enter the locked module functions (providers use their own instance `self._lock`). Keeping a plain `Lock` makes the contract strict and would surface any future accidental re-entry instead of hiding it.
- `shutdown_sandbox_provider()` calls `provider.shutdown()` (which `join`s the idle-checker with a timeout) while holding the lock; that critical section can block briefly, but the idle-checker never acquires `_provider_lock` and `shutdown()` is idempotent, so there is no deadlock.

## Relationship to #3722

Thanks to #3722, which targets the same issue — that PR is what prompted a closer look. This PR is offered as an **alternative implementation** so maintainers can pick whichever they prefer; the differences are:

- **Plain `Lock` instead of `RLock`.** There is no re-entrant call path while the lock is held (see the note above), so `RLock` isn't needed; a plain `Lock` is stricter and fails loudly on any future accidental re-entry.
- **`reset_sandbox_provider()` keeps the original behavior** of clearing the global only inside the `if _default_sandbox_provider is not None:` branch (in #3722 the `= None` is unconditional, which is harmless but diverges from the original contract).
- **Deterministic regression tests.** A `threading.Barrier` aligns the racing threads so the bug is triggered reliably (verified: the unlocked version creates one provider per thread under the same barrier), plus an extra reset-vs-get case.

Happy to consolidate into whichever direction the maintainers prefer.

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/sandbox_provider.py
+++ b/backend/packages/harness/deerflow/sandbox/sandbox_provider.py
@@ -56,12 +56,20 @@ class SandboxProvider(ABC):
 
 
 _default_sandbox_provider: SandboxProvider | None = None
-# Guards every access to `_default_sandbox_provider`. The singleton is reachable
-# from more than one OS thread (e.g. the main event loop and the Feishu channel
-# thread, which runs its own loop), so a bare check-then-create can double
-# initialize the provider, and an unsynchronized reset/shutdown racing an
-# in-flight create can tear it down or clear it mid-construction. Hold this lock
-# across the whole lifecycle (get / reset / shutdown / set) to keep it consistent.
+# Guards every read and write of `_default_sandbox_provider`. The singleton is
+# reachable from more than one OS thread (e.g. the main event loop and the Feishu
+# channel thread, which runs its own loop), so a bare check-then-create can double
+# initialize the provider, and an unsynchronized reset/shutdown racing a get can
+# hand a caller `None` or a torn instance. Every access to the global below takes
+# this lock, including the read+return in `get_sandbox_provider()`.
+#
+# The lock guards only the reference swap. Provider callbacks (`__init__`,
+# `reset()`, `shutdown()`) and the dynamic import in `resolve_class()` run
+# *outside* the lock: they are plugin-supplied (`config.sandbox.use` resolves to
+# an arbitrary class) and may be slow or, worse, re-enter these lifecycle
+# functions. Holding a non-reentrant `threading.Lock` across them would
+# self-deadlock such a provider and would block every concurrent `get()` during a
+# slow teardown. Keeping callbacks off the lock avoids both.
 _provider_lock = threading.Lock()
 
 
@@ -75,13 +83,33 @@ def get_sandbox_provider(**kwargs) -> SandboxProvider:
         A sandbox provider instance.
     """
     global _default_sandbox_provider
-    if _default_sandbox_provider is None:
-        with _provider_lock:
-            if _default_sandbox_provider is None:
-                config = get_app_config()
-                cls = resolve_class(config.sandbox.use, SandboxProvider)
-                _default_sandbox_provider = cls(**kwargs)
-    return _default_sandbox_provider
+    # Fast path: a single locked read so a concurrent reset/shutdown can't null
+    # the global between the check and the return.
+    with _provider_lock:
+        if _default_sandbox_provider is not None:
+            return _default_sandbox_provider
+
+    # Cold start. Resolve + construct outside the lock: the import and the
+    # provider constructor are plugin code and must not run under a non-reentrant
+    # lock. The construction may race another caller; we reconcile under the lock.
+    config = get_app_config()
+    cls = resolve_class(config.sandbox.use, SandboxProvider)
+    provider = cls(**kwargs)
+
+    with _provider_lock:
+        if _default_sandbox_provider is None:
+            _default_sandbox_provider = provider
+            return provider
+        # We lost the install race: another thread got there first. `winner` is
+        # read under the same lock, so it is always a live instance, never None.
+        winner = _default_sandbox_provider
+
+    # Discard the instance we just built (outside the lock). For providers with
+    # side-effectful constructors (e.g. AioSandboxProvider starts an idle-checker
+    # thread), this tears down the orphan so it does not leak — issue #3721.
+    if hasattr(provider, "shutdown"):
+        provider.shutdown()
+    return winner
 
 
 def reset_sandbox_provider() -> None:
@@ -100,10 +128,13 @@ def reset_sandbox_provider() -> None:
     Use `shutdown_sandbox_provider()` for proper cleanup.
     """
     global _default_sandbox_provider
+    # Detach the reference under the lock, then run the provider's `reset()`
+    # callback outside it (see the `_provider_lock` note).
     with _provider_lock:
-        if _default_sandbox_provider is not None:
-            _default_sandbox_provider.reset()
-            _default_sandbox_provider = None
+        provider = _default_sandbox_provider
+        _default_sandbox_provider = None
+    if provider is not None:
+        provider.reset()
 
 
 def shutdown_sandbox_provider() -> None:
@@ -114,17 +145,22 @@ def shutdown_sandbox_provider() -> None:
     is shutting down or when you need to completely reset the sandbox system.
     """
     global _default_sandbox_provider
+    # Detach the reference under the lock, then run the (potentially slow)
+    # `shutdown()` callback outside it (see the `_provider_lock` note).
     with _provider_lock:
-        if _default_sandbox_provider is not None:
-            if hasattr(_default_sandbox_provider, "shutdown"):
-                _default_sandbox_provider.shutdown()
-            _default_sandbox_provider = None
+        provider = _default_sandbox_provider
+        _default_sandbox_provider = None
+    if provider is not None and hasattr(provider, "shutdown"):
+        provider.shutdown()
 
 
 def set_sandbox_provider(provider: SandboxProvider) -> None:
     """Set a custom sandbox provider instance.
 
     This allows injecting a custom or mock provider for testing purposes.
+
+    Note: any previously installed provider is replaced but not shut down; the
+    caller owns the lifecycle of the instance it is overwriting.
 
     Args:
         provider: The SandboxProvider instance to use.
--- a/backend/packages/harness/deerflow/sandbox/sandbox_provider.py
+++ b/backend/packages/harness/deerflow/sandbox/sandbox_provider.py
@@ -1,4 +1,5 @@
 import asyncio
+import threading
 from abc import ABC, abstractmethod
 
 from deerflow.config import get_app_config
@@ -55,6 +56,13 @@ class SandboxProvider(ABC):
 
 
 _default_sandbox_provider: SandboxProvider | None = None
+# Guards every access to `_default_sandbox_provider`. The singleton is reachable
+# from more than one OS thread (e.g. the main event loop and the Feishu channel
+# thread, which runs its own loop), so a bare check-then-create can double
+# initialize the provider, and an unsynchronized reset/shutdown racing an
+# in-flight create can tear it down or clear it mid-construction. Hold this lock
+# across the whole lifecycle (get / reset / shutdown / set) to keep it consistent.
+_provider_lock = threading.Lock()
 
 
 def get_sandbox_provider(**kwargs) -> SandboxProvider:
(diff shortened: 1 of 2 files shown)


backend/packages/harness/deerflow/sandbox/sandbox_provider.py
```diff
@@ -68,9 +76,11 @@ def get_sandbox_provider(**kwargs) -> SandboxProvider:
...
     if _default_sandbox_provider is None:
-        config = get_app_config()
-        cls = resolve_class(config.sandbox.use, SandboxProvider)
-        _default_sandbox_provider = cls(**kwargs)
+        with _provider_lock:
+            if _default_sandbox_provider is None:
+                config = get_app_config()
+                cls = resolve_class(config.sandbox.use, SandboxProvider)
+                _default_sandbox_provider = cls(**kwargs)
     return _default_sandbox_provider
```
willem-bd: **`get`'s fast path reads the singleton without the lock — can return `None` or a torn-down instance.**

The outer `if _default_sandbox_provider is None:` (L78) and this `return` (L84) both read the global *outside* `_provider_lock`. So a concurrent `reset_sandbox_provider`/`shutdown_sandbox_provider`/`set_sandbox_provider` can null it (or call `provider.reset()` clearing its caches) between the check and the return, handing the caller `None` or a partially-reset instance — despite the `-> SandboxProvider` annotation. The `get_sandbox_provider().get(sandbox_id)` call sites (`tools.py`, `tool_output_budget_middleware.py`) would then raise `AttributeError`.

The PR's own comment (L59-64) claims the lock is held "across the whole lifecycle (get / reset / shutdown / set)" to guard against "reset/shutdown racing an in-flight create", but `get`'s read path isn't actually covered. Latent today — `reset`/`shutdown`/`set` have no in-tree runtime callers (sandbox is startup-only) — but they're documented public shutdown APIs of a publishable harness, and the get-vs-reset/shutdown race is exactly what the comment says it prevents.

Fix: capture the return value under the lock (e.g. assign into a local inside the `with`, then return the local), or hold the lock across the read+return.

backend/packages/harness/deerflow/sandbox/sandbox_provider.py
```diff
@@ -68,9 +76,11 @@ def get_sandbox_provider(**kwargs) -> SandboxProvider:
...
     global _default_sandbox_provider
     if _default_sandbox_provider is None:
-        config = get_app_config()
-        cls = resolve_class(config.sandbox.use, SandboxProvider)
-        _default_sandbox_provider = cls(**kwargs)
+        with _provider_lock:
+            if _default_sandbox_provider is None:
+                config = get_app_config()
+                cls = resolve_class(config.sandbox.use, SandboxProvider)
+                _default_sandbox_provider = cls(**kwargs)
```
willem-bd: **Non-reentrant lock held across arbitrary plugin callbacks → deadlock hazard for custom providers.**

This line runs under `_provider_lock` (acquired at L79), and `resolve_class()` just above it is a dynamic `importlib` import. `reset`/`shutdown` hold the same lock across `provider.reset()`/`.shutdown()`. `threading.Lock` is non-reentrant, and the provider is a plugin point (`config.sandbox.use` → any class). A custom provider whose `__init__`/`reset`/`shutdown` re-enters `get`/`reset`/`shutdown`/`set_sandbox_provider`, or whose import/ctor takes a lock held in reverse order elsewhere, self-deadlocks on the cold-start path (runs every process start) — every sandbox acquire then hangs.

The built-in `LocalSandboxProvider` is safe (its `reset`/`shutdown` only touch an instance lock, no re-entry), so this only affects custom providers.

Fix: resolve + construct into a local, then assign under the lock with a re-check; for reset/shutdown, capture the old reference, null the global under the lock, and call `.reset()`/`.shutdown()` on it *outside* the lock (this also stops blocking every concurrent `get()` during slow teardown).

backend/tests/test_sandbox_provider_lifecycle.py
```diff
@@ -0,0 +1,117 @@
...
+        thread.join()
+
+    try:
+        assert len({id(provider) for provider in providers}) == 1
+        assert SlowSandboxProvider.instances_created == 1
+    finally:
+        sandbox_provider.reset_sandbox_provider()
+
+
+def test_reset_racing_get_does_not_tear_down_returned_instance(monkeypatch):
```
willem-bd: **This test races a no-op reset — it never exercises reset-of-a-live-provider.**

Line 87 calls `reset_sandbox_provider()` *before* the barrier, clearing the global to `None`. So when the resetter thread calls `reset_sandbox_provider()` again it hits `if _default_sandbox_provider is not None:` → `False` → no-op. All getters therefore just construct fresh providers and `all(isinstance(p, SlowSandboxProvider) …)` passes trivially, without ever racing a reset of a *live* singleton against `get`.

That means it wouldn't catch the get-vs-reset/shutdown defect (finding 1) — it asserts getters never get `None`/torn instances, but the scenario that could produce them is never set up. To make it meaningful, populate the singleton first (`get_sandbox_provider()` before the barrier) so the resetter actually tears down a live instance while getters read it.

willem-bd (commented):
Reviewed the singleton lifecycle locking. The double-checked-locking fix for double-init is correct and production-relevant — `get_sandbox_provider` is reached from the main event loop and the Feishu channel thread (separate loop), so the old unlocked check-then-create could genuinely double-initialize.

Two gaps in the locking and one test that doesn't exercise its stated scenario:

1. `get`'s fast path (unlocked outer check + unlocked `return`) reads the singleton without the lock, so a concurrent `reset`/`shutdown` can null or reset it mid-read → `None`/torn instance returned.
2. The non-reentrant `_provider_lock` is held across arbitrary plugin callbacks (`resolve_class` import + `cls(**kwargs)` in `get`; `provider.reset`/`.shutdown`), so a custom provider that re-enters self-deadlocks.
3. `test_reset_racing_get_does_not_tear_down_returned_instance` resets to `None` before the barrier, so the racing reset is a no-op.

None of these affect the default provider today. The cleaner shape is to protect only the reference swap and run provider callbacks outside the lock.

WillemJiang (approved):

PR #3732, opened 2026-06-23
fix: preserve LangGraph resume command

### LINKED ISSUE #3720: Follow up of interrupt
### Before you start

### Problem / motivation

Command(resume) 未传递

  文件 1: backend/app/gateway/services.py:408

──# 当前：只取 body.input，忽略 body.command──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  graph_input = normalize_input(body.input)
  # 应改为：
  if body.command and body.command.get("resume") is not None:
      from langgraph.types import Command
      graph_input = Command(resume=body.command["resume"])
  else:
      graph_input = normalize_input(body.input)

  文件 2: backend/app/gateway/services.py:420-430

  # 当前 run_agent 调用缺少 command 参数：
  task = asyncio.create_task(
      run_agent(
          bridge, run_mgr, record, ctx=run_ctx,
          agent_factory=agent_factory, graph_input=graph_input,
          config=config, stream_modes=stream_modes,
          ...
      )
  )

  # 应加上 command=body.command

  文件 3: backend/packages/harness/deerflow/runtime/runs/worker.py:131

  # 当前签名没有 command 参数：
  async def run_agent(
      bridge, run_mgr, record, ctx, agent_factory,
      graph_input: dict, config, stream_modes, ...

  # 应加上
      command: dict[str, Any] | None = None,

  文件 4: worker.py — astream() 调用前（约 310 行前）

  # 在 astream(graph_input, ...) 之前加上：
  from langgraph.types import Command as LGCommand
  if command is not None and isinstance(command, dict):
      resume_value = command.get("resume")
      if resume_value is not None:
          graph_input = LGCommand(resume=resume_value)

  影响: langgraph.types.interrupt() 能暂停 graph，但 resume 后 interrupt() 永远返回 None。所有依赖 interrupt() 返回值的人机交互流程（凭证收集、审核确认、多步表单等）全部失效。

### Proposed solution

11

### Affected area(s)

### Alternatives considered

_No response_

### Additional context

_No response_

Huixin615:
Closes #3720

## Why

Resume requests sent through the LangGraph-compatible runs API already accept `command`, but `start_run()` only used `body.input` when constructing the graph input.

As a result, requests like `{"command": {"resume": ...}}` silently dropped the resume value, so resumed graphs could restart from checkpoint without delivering the value back to `interrupt()`.

## What changed

- `start_run()` now converts `body.command.resume` into `langgraph.types.Command(resume=...)` before launching the graph.
- Normal `input` handling is unchanged when no resume command is provided.
- Added regression coverage for both resume-command and normal-input paths.

## Surface area

## Screenshots / Recording

N/A. Backend-only bug fix.

## Bug fix verification

- Test path that reproduces the bug: `backend/tests/test_gateway_services.py::test_start_run_translates_resume_command_to_langgraph_command`
- The regression test verifies that `command.resume` reaches `run_agent` as a real LangGraph `Command` instead of being dropped.

## Validation

```bash
cd backend && uv run pytest tests/test_gateway_services.py
cd backend && uv run ruff check app tests/test_gateway_services.py
cd backend && uv run ruff format --check app tests/test_gateway_services.py

### DIFF
--- a/backend/app/gateway/services.py
+++ b/backend/app/gateway/services.py
@@ -18,6 +18,7 @@ from typing import Any
 from fastapi import HTTPException, Request
 from langchain_core.messages import BaseMessage
 from langchain_core.messages.utils import convert_to_messages
+from langgraph.types import Command
 
 from app.gateway.deps import get_checkpointer, get_run_context, get_run_manager, get_stream_bridge
 from app.gateway.internal_auth import INTERNAL_SYSTEM_ROLE, get_trusted_internal_owner_user_id
@@ -456,7 +457,11 @@ async def start_run(
             logger.warning("Failed to upsert thread_meta for %s (non-fatal)", sanitize_log_param(thread_id))
 
         agent_factory = resolve_agent_factory(body.assistant_id)
-        graph_input = normalize_input(body.input)
+        command = getattr(body, "command", None)
+        if command and command.get("resume") is not None:
+            graph_input = Command(resume=command["resume"])
+        else:
+            graph_input = normalize_input(body.input)
         config = build_run_config(thread_id, body.config, body.metadata, assistant_id=body.assistant_id)
         await apply_checkpoint_to_run_config(config, body=body, thread_id=thread_id, request=request)
 
--- a/backend/tests/test_gateway_services.py
+++ b/backend/tests/test_gateway_services.py
@@ -612,6 +612,89 @@ def test_inject_authenticated_user_context_skips_internal_role():
     assert config["context"]["user_id"] == "channel-user-7"
 
 
+async def _capture_start_run_graph_input(body):
+    from types import SimpleNamespace
+    from unittest.mock import patch
+
+    from langgraph.checkpoint.memory import InMemorySaver
+    from langgraph.store.memory import InMemoryStore
+
+    from app.gateway.services import start_run
+    from deerflow.persistence.thread_meta.memory import MemoryThreadMetaStore
+    from deerflow.runtime import RunManager
+    from deerflow.runtime.runs.store.memory import MemoryRunStore
+
+    run_manager = RunManager(store=MemoryRunStore())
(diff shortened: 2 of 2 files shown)


Huixin615:
@fancyboi999 哈喽 辛苦帮忙review

fancyboi999 (commented):
Reviewed — this correctly closes #3720. The translation is exactly the right shape: `start_run` builds `Command(resume=...)` from `body.command["resume"]` and otherwise falls through to `normalize_input(body.input)` unchanged. Using `is not None` rather than a truthy check is the right call — it preserves falsy resume payloads (`False`, `0`, `""`) instead of dropping them back onto the input path.

I checked the one non-obvious risk — that `graph_input` is now sometimes a `Command` (not a `dict`) flowing into the worker — and it's safe: `run_agent` hands `graph_input` straight to `agent.astream(...)` (`worker.py:312`/`:322`), which accepts a `Command` natively, and the only dict-assuming consumer, `_extract_human_message` (`worker.py:653`, `graph_input.get("messages")`), is currently uncalled, so there's no `.get()` AttributeError on the resume path. `apply_checkpoint_to_run_config` still runs afterward, so resume-from-checkpoint and the resume value coexist correctly.

The tests are real anchors — `test_start_run_translates_resume_command_to_langgraph_command` fails on main (graph_input would be the `{}` from `normalize_input(None)`), and the no-command test guards the unchanged path.

Non-blocking: this handles `resume` only; a `command` carrying `goto`/`update` falls through to the input path. That's the right scope for #3720 (resume is the interrupt/HITL case) — just flagging in case you want a follow-up for the other `Command` fields later.

Looks good to me.

WillemJiang (approved):

PR #3733, opened 2026-06-23
fix(frontend): make recent chat rows fully clickable

zwj110610:
## Why
Clicking a recent chat in the sidebar only worked when the pointer was over the chat title/link area. Blank space within the same visible row did not navigate, which made the interaction feel inconsistent: the row looked like one selectable item, but only part of it was actually clickable.

This PR makes the recent chat row behave like a standard sidebar item: the main row area is the navigation target, while the overflow menu remains a separate action.

## What changed
- Recent chat rows now use `SidebarMenuButton asChild` directly with the chat `Link`, so the row's main area is the actual clickable navigation target.
- The overflow menu action is kept as a sibling action within the same sidebar item, instead of being nested inside the link wrapper.
- The E2E coverage now clicks blank space inside the row padding to verify that the whole row navigation target works.

## Surface area

## Screenshots / Recording

N/A. This is an interaction target fix rather than a visual layout change. The behavior is covered by E2E.

## Bug fix verification

The E2E test clicks blank space inside a recent chat row, instead of clicking the title text directly. Before the fix, this click target does not navigate because it is outside the row's navigation target. With this change, the same click navigates to the selected chat.

- Yes. The scenario is encoded as an E2E regression test and passes on this branch.

## Validation

cd backend && make lint: passed
cd backend && make test: passed, 4918 passed, 19 skipped
cd frontend && pnpm format: passed
cd frontend && pnpm lint: passed with 4 existing warnings
cd frontend && pnpm typecheck: passed
cd frontend && BETTER_AUTH_SECRET=local-dev-secret pnpm build: passed with existing Turbopack NFT warning
cd frontend && make test: passed, 341 passed
cd frontend && CI=1 make test-e2e: passed, 55 passed
git diff --check: passed

## AI assistance

**Tool(s) used:** codex

**How you used it:** Use AI for testing and code review

### DIFF
--- a/frontend/src/components/workspace/recent-chat-list.tsx
+++ b/frontend/src/components/workspace/recent-chat-list.tsx
@@ -242,7 +242,7 @@ export function RecentChatList() {
                   >
                     <SidebarMenuButton isActive={isActive} asChild>
                       <Link
-                        className="text-muted-foreground min-w-0 gap-1.5 whitespace-nowrap group-hover/side-menu-item:overflow-hidden"
+                        className="text-muted-foreground min-w-0 whitespace-nowrap group-hover/side-menu-item:overflow-hidden"
                         href={pathOfThread(thread)}
                       >
                         <ThreadChannelIcon source={channelSource} />
@@ -266,7 +266,7 @@ export function RecentChatList() {
                         <DropdownMenuTrigger asChild>
                           <SidebarMenuAction
                             showOnHover
-                            className="bg-background/50 hover:bg-background"
+                            className="bg-background/50 hover:bg-background after:left-0!"
                           >
                             <MoreHorizontal />
                             <span className="sr-only">{t.common.more}</span>
--- a/frontend/src/components/workspace/recent-chat-list.tsx
+++ b/frontend/src/components/workspace/recent-chat-list.tsx
@@ -241,95 +241,91 @@ export function RecentChatList() {
                     className="group/side-menu-item"
                   >
                     <SidebarMenuButton isActive={isActive} asChild>
-                      <div>
-                        <Link
-                          className="text-muted-foreground flex min-w-0 items-center gap-1.5 pr-7 whitespace-nowrap group-hover/side-menu-item:overflow-hidden"
-                          href={pathOfThread(thread)}
-                        >
-                          <ThreadChannelIcon source={channelSource} />
-                          <span className="min-w-0 truncate">
-                            {titleOfThread(thread)}
-                          </span>
-                          {channelSource && (
-                            <span
-                              className="bg-muted text-muted-foreground ml-auto inline-flex h-5 max-w-14 shrink-0 items-center rounded-md px-1.5 text-[10px] font-medium"
-                              title={`${channelSource.label} channel`}
-                            >
-                              <span className="truncate">
-                                {channelSource.label}
-                              </span>
+                      <Link
+                        className="text-muted-foreground min-w-0 gap-1.5 whitespace-nowrap group-hover/side-menu-item:overflow-hidden"
+                        href={pathOfThread(thread)}
+                      >
+                        <ThreadChannelIcon source={channelSource} />
+                        <span className="min-w-0 truncate">
+                          {titleOfThread(thread)}
+                        </span>
+                        {channelSource && (
+                          <span
+                            className="bg-muted text-muted-foreground ml-auto inline-flex h-5 max-w-14 shrink-0 items-center rounded-md px-1.5 text-[10px] font-medium"
+                            title={`${channelSource.label} channel`}
+                          >
+                            <span className="truncate">
+                              {channelSource.label}
                             </span>
-                          )}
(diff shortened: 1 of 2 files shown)


zwj110610:
While using the app, I noticed that switching to a recent chat only worked when clicking directly on the chat title text. Clicking other blank areas within the same visible row did not navigate to the chat, which can make the UI feel unresponsive or laggy.

This PR fixes the issue by making the recent chat row's main area the actual navigation target, while keeping the overflow menu as a separate action.

frontend/src/components/workspace/recent-chat-list.tsx
```diff
@@ -241,95 +241,91 @@ export function RecentChatList() {
...
-                              align={"start"}
-                            >
+                          </span>
+                        )}
+                      </Link>
+                    </SidebarMenuButton>
+                    {env.NEXT_PUBLIC_STATIC_WEBSITE_ONLY !== "true" && (
+                      <DropdownMenu>
+                        <DropdownMenuTrigger asChild>
+                          <SidebarMenuAction
```
willem-bd, line 267: ⚠️ Mobile interaction — now that the row `<Link>` is full-width (`w-full`), it overlaps the overflow action's expanded mobile hit area. `SidebarMenuAction` carries `after:absolute after:-inset-2 md:after:hidden` (`sidebar.tsx:566`), an invisible ~8px-expanded tap target shown only below the `md` breakpoint. Before this change the right side of the row was dead padding, so that was harmless; now on mobile the action (absolutely positioned, paints on top of the static `<a>`) covers the rightmost ~32px, so a tap there opens the '⋯' menu instead of navigating to the chat.

Worth verifying on a narrow viewport — e.g. leave a non-clickable right gutter or scope down the expanded hit area.

willem-bd (commented):
**Code review pass** (recall-focused). Overall this is a clean refactor to the canonical shadcn pattern (`SidebarMenuButton asChild` over the `Link`, with the overflow `DropdownMenu` as a sibling action) and a net a11y improvement — the whole row is now one focusable `<a>`.

Verified non-issues: layout is preserved (the button supplies `flex w-full items-center gap-2`), `pr-8` auto-applies via `group-has-data-[sidebar=menu-action]/menu-item:pr-8`, the title still truncates (keeps its own `min-w-0 truncate`), the icon isn't force-resized (`ThreadChannelIcon` renders a `<span>` wrapper so `[&>svg]:size-4` never fires on it), and the E2E click target is valid (the `<a>` is now the full-width menu-button, so `box.x + 4` hits its left padding and navigates).

Three inline notes below — one mobile interaction worth a look, two minor cleanups. No correctness blockers.

frontend/src/components/workspace/recent-chat-list.tsx
```diff
@@ -241,95 +241,91 @@ export function RecentChatList() {
...
-                          {channelSource && (
-                            <span
-                              className="bg-muted text-muted-foreground ml-auto inline-flex h-5 max-w-14 shrink-0 items-center rounded-md px-1.5 text-[10px] font-medium"
-                              title={`${channelSource.label} channel`}
-                            >
-                              <span className="truncate">
-                                {channelSource.label}
-                              </span>
+                      <Link
+                        className="text-muted-foreground min-w-0 gap-1.5 whitespace-nowrap group-hover/side-menu-item:overflow-hidden"
```
willem-bd: Minor: `gap-1.5` here is now shadowed by `SidebarMenuButton`'s own `gap-2`. With `asChild`, Radix `Slot` **concatenates** classes (it doesn't run `twMerge`), so both `gap-1.5` and `gap-2` land on the same `<a>`; equal specificity means the one later in Tailwind's generated stylesheet wins, so the effective gap silently becomes `0.5rem` instead of the intended `0.375rem` (~33% wider).

Since the button now owns the flex/gap, this `gap-1.5` can be dropped.

frontend/tests/e2e/thread-history.spec.ts
```diff
@@ -53,6 +53,32 @@ test.describe("Thread history", () => {
...
+      .first();
+    await expect(firstThreadItem).toBeVisible({ timeout: 15_000 });
+
+    const box = await firstThreadItem.boundingBox();
+    expect(box).not.toBeNull();
+    if (!box) {
+      return;
+    }
+
+    await page.mouse.click(box.x + 4, box.y + box.height / 2);
```
willem-bd: Robustness nit: `page.mouse.click` with a captured coordinate skips Playwright's actionability auto-wait, and `SidebarMenuButton` has `transition-[width,height,padding]`. On a slow CI runner where the sidebar is still settling, `boundingBox()` could return a stale coordinate, the click could miss, and it would surface as a `waitForURL` timeout rather than a clean assertion failure.

`firstThreadItem.getByRole("link").click({ position: { x: 4, y: box.height / 2 } })` auto-waits while still targeting the blank left padding.

WillemJiang (approved):

Issue #3489, opened 2026-06-11
[UI Polish][Phase 1] Fix critical mobile workspace and accessibility blockers


hetaoBackend:
## Phase

Phase 1: Critical usability

## Priority

P0/P1

## Background

The UI audit found several issues that currently affect basic usability, especially on mobile workspace screens and keyboard accessibility. This phase should be handled before broader visual polish.

Source audit: `docs/ui-polish-audit-2026-06-10.md`
Local issue draft: `docs/ui-polish-issues-2026-06-11.md`

Suggested labels: `ui-polish`, `phase-1`, `mobile`, `accessibility`, `workspace`

### 1. Make keyboard focus visible globally

- Update light/dark `--ring` tokens in `frontend/src/styles/globals.css`.
- Verify focus-visible states for sidebar trigger, command palette, chat composer, artifact toolbar, settings tabs, dropdowns, and dialogs.
- Do not directly edit generated component directories.

### 2. Add a reliable mobile workspace navigation entry

- Show a persistent mobile menu/sidebar trigger in the workspace top header.
- Use a drawer/sheet pattern for mobile sidebar navigation.
- Ensure collapsed sidebar state still has a discoverable trigger.

### 3. Replace mobile artifact split panes with a drawer/sheet

- Do not render chat and artifacts as desktop split panes on narrow screens.
- Avoid auto-opening artifacts in mobile demo mode.
- Prevent hidden/offscreen artifact controls from remaining focusable.
- Rework artifact header actions for narrow widths.

### 4. Clean up static demo and backend-unavailable states

- In static website/demo mode, use explicit demo data or clear unavailable states.
- Cover loading/empty/error/ready for channels, agents, and skills.
- Make disabled composer state propagate to all child controls.

### 5. Fix landing/header mobile overflow and hero composition

- Create a compact mobile marketing header variant.
- Remove transform scaling from hero text/buttons.
- Avoid `h-screen` locking that hides the next section completely.
- Ensure the header brand is not the page H1.

## Suggested files

- `frontend/src/styles/globals.css`
- `frontend/src/components/workspace/workspace-container.tsx`
- `frontend/src/components/workspace/sidebar/workspace-sidebar.tsx`
- `frontend/src/components/workspace/sidebar/workspace-header.tsx`
- `frontend/src/components/workspace/chats/chat-box.tsx`
- `frontend/src/components/workspace/artifacts/artifact-panel.tsx`
[...]

PR #3740, opened 2026-06-23
Fix(frontend): Fix mobile workspace and accessibility blockers

18062706139fcz:
## Summary

- Fix mobile landing overflow and keep the landing page to a single real `h1`.
- Move mobile artifact viewing into a drawer so chat/artifact content does not force horizontal scrolling.
- Restore visible global focus ring tokens and add Playwright regression coverage for the mobile/layout cases from #3489.
- Keep the PR scoped to one frontend fix commit and six effective frontend source/test files.

## Verification

- `pnpm check` passed on the rewritten single-commit PR branch.
- `pnpm test` passed on the rewritten single-commit PR branch: 33 files / 312 tests.
- `SKIP_ENV_VALIDATION=1 DEER_FLOW_AUTH_DISABLED=1 pnpm exec next build --webpack` passed on the rewritten single-commit PR branch: 79 pages generated.
- On a temporary `origin/main` merge-state verification branch, `pnpm check` passed with only four pre-existing warnings.
- On that same merge-state verification branch, `pnpm test` passed: 35 files / 341 tests.
- On that same merge-state verification branch, `SKIP_ENV_VALIDATION=1 DEER_FLOW_AUTH_DISABLED=1 pnpm exec next build --webpack` passed: 81 pages generated.
- On that same merge-state verification branch, full Chromium E2E passed: `pnpm exec playwright test --config=/private/tmp/deer-flow-playwright-local.config.ts --project=chromium` -> 58/58 tests.

## UI Evidence

- Checked landing screenshots at 320px, 375px, and 390px widths.
- Checked mobile workspace sidebar drawer at 375px.
- Checked mobile artifacts drawer at 375px.
- Screenshots were captured locally and intentionally not committed so the PR diff stays limited to effective source/test changes.

## Screenshot

- Landing, 320px:  image
- Landing, 375px: image
- Landing, 390px: 
image
- Workspace sidebar drawer, 375px: image
- Workspace artifact drawer, 375px: image


fancyboi999:
Heads-up on the PR composition: the diff here carries a lot more than the #3489 UI polish. Three of the four commits are the custom-postgres-schema feature — `dc77c5c` / `742579f` / `c8cf787` (`feat(persistence): support custom postgres schema`) — which is already up as its own PR in #3442. So #3740's diff includes ~10 backend persistence files (`postgres_schema.py`, the checkpointer/store providers, `database_config.py`, the pg tests, …) on top of the single frontend commit `a9bb980`.

That's almost certainly because this branch was cut from the #3442 branch rather than from `main` — which also explains the `mergeable: dirty` state. Rebasing #3740 onto current `main` should drop the persistence commits and leave just the mobile/a11y changes, so each PR can be reviewed on its own and the persistence work doesn't land under a UI-polish title with less scrutiny than it's getting in #3442.

Happy to look at the UI polish once it's down to just the #3489 changes.

WillemJiang:
@18062706139fcz, please take a look at the lint error.

WillemJiang:
@18062706139fcz  please fix the e2e test.

frontend/src/components/landing/header.tsx
```diff
@@ -21,20 +21,21 @@ export async function Header({ className, homeURL, locale }: HeaderProps) {
...
           target={isExternalHome ? "_blank" : "_self"}
           rel={isExternalHome ? "noopener noreferrer" : undefined}
+          className="font-serif text-xl whitespace-nowrap"
         >
-          <h1 className="font-serif text-xl">DeerFlow</h1>
+          DeerFlow
         </a>
       </div>
-      <nav className="mr-8 ml-auto flex items-center gap-8 text-sm font-medium">
+      <nav className="ml-auto hidden items-center gap-5 text-sm font-medium sm:flex md:mr-8 md:gap-8">
```
willem-bd, line 40: **Mobile nav has no replacement.** The nav is now `hidden ... sm:flex md:mr-8 md:gap-8`, so below 640px users see only the brand and the Star button — there's no hamburger or mobile menu, and the landing page exposes no other entry point to `/docs` or `/blog/posts`. This is a regression for the same mobile cohort the PR is fixing. Consider adding a `Sheet`/hamburger trigger below `sm:`.

frontend/src/components/workspace/chats/chat-box.tsx
```diff
@@ -100,6 +109,70 @@ const ChatBox: React.FC<{ children: React.ReactNode; threadId: string }> = ({
...
+              files={thread.values.artifacts ?? []}
+              threadId={threadId}
+            />
+          </main>
+        </div>
+      )}
+    </div>
+  );
+
+  if (isMobile) {
```
willem-bd, line 198: **Hydration flash on mobile.** `useIsMobile()` returns `false` during SSR and the first client commit (it only flips after `useEffect` runs), so phones receive the desktop `ResizablePanelGroup` HTML, hydrate it, then immediately unmount it and remount the `Sheet` tree. That causes a visible layout flash + reflow and discards any panel state the server-rendered tree set up.

A CSS-only swap (`hidden md:flex` on the desktop tree, `flex md:hidden` on the mobile tree) or a `useSyncExternalStore`-backed matchMedia hook would render the right tree from the first paint and avoid the swap entirely.

frontend/src/components/landing/hero.tsx
```diff
@@ -89,6 +90,47 @@ export function Hero({ className }: { className?: string }) {
   );
 }
 
+function HeroWordRotate({
```
willem-bd, line 93: **Reuse: `HeroWordRotate` duplicates `WordRotate`.** `@/components/ui/word-rotate.tsx` already has the same setInterval state, `AnimatePresence mode="popLayout"`, motion props, and `AuroraText` with the identical `["#efefbb","#e9c665","#e3a812"]` colors. It already accepts `className` and `motionProps`.

And because this PR removed the only call site of `WordRotate`, that file is now dead code shipped in the bundle. Either:
- reuse `WordRotate` here (parameterize the wrapper element / pass `className`), or
- delete `frontend/src/components/ui/word-rotate.tsx`.

(Note: `ui/` is registry-generated per `frontend/CLAUDE.md`, so deleting needs care — but leaving it unimported is the worst of both worlds.)

frontend/src/components/workspace/chats/chat-box.tsx
```diff
@@ -100,6 +109,70 @@ const ChatBox: React.FC<{ children: React.ReactNode; threadId: string }> = ({
     }
   }, [artifactPanelOpen]);
 
+  const artifactContent = selectedArtifact ? (
```
willem-bd: **Wasted work in the streaming hot path.** `artifactContent` now builds the full conditional JSX (Button, ConversationEmptyState, ArtifactFileList, the wrapper divs, etc.) on **every** `ChatBox` render, regardless of whether the panel is open. `ChatBox` re-renders on every streamed token via `useThread()`, so this runs hundreds of times per turn while the user can never see the result.

Either wrap in `useMemo` (deps: `selectedArtifact`, `thread.values.artifacts`, `threadId`, `setArtifactsOpen`) or move the JSX back inside the branches that actually render it.

frontend/src/components/landing/hero.tsx
```diff
@@ -89,6 +90,47 @@ export function Hero({ className }: { className?: string }) {
...
+    }, duration);
+
+    return () => clearInterval(interval);
+  }, [words, duration]);
+
+  return (
+    <div className="relative max-w-full min-w-0 overflow-hidden py-2">
+      <AnimatePresence mode="popLayout">
+        <motion.div
+          key={words[index]}
```
willem-bd: **`key` should be the slot, not the content.** `key={words[index]}` collides silently if `HERO_WORDS` ever gets a duplicate string — `AnimatePresence` would treat two consecutive renders with the same key as the same element and skip the exit/enter transition, with no obvious symptom in code review.

`key={index}` (or a stable id alongside the word) matches the semantics here: "this is the same animation slot, with new content".

willem-bd (commented):
Automated review — `/code-review` (high effort, recall-biased).

No correctness-fatal bugs in this diff. Summary of findings (severity-ranked):

1. **Mobile UX regression** — landing nav (docs/blog) is hidden below `sm:` with no hamburger replacement, so mobile users lose access to those routes from the landing header.
2. **Hydration/layout flash on mobile** — `useIsMobile()` returns `false` during SSR + first client render, so the desktop `ResizablePanelGroup` tree is rendered then swapped for the `Sheet` tree on mount. A CSS-only `hidden md:flex / flex md:hidden` swap would avoid this.
3. **Reuse / dead code** — `HeroWordRotate` duplicates `@/components/ui/word-rotate.tsx` (same setInterval + AnimatePresence/popLayout + AuroraText with identical colors). The PR's only call site of `WordRotate` was removed, leaving `WordRotate` as dead code.
4. **Efficiency** — `artifactContent` JSX is built unconditionally on every `ChatBox` render (which is in the streaming hot path).
5. **Fragility** — `key={words[index]}` collides silently if `HERO_WORDS` ever gets a duplicate entry.
6. **Visual cohesion at 320px** — the flex-wrap row breaks `"SuperAgent"` onto its own line for longer rotating words; the overflow test only checks `scrollWidth`.
7. **Test locale coupling** — `getByRole("button", { name: /artifacts/i })` only matches the en-US label `"Artifacts"`; zh-CN renders `"文件"`.
8. **Test self-blindness** — the focus-ring test reads `--ring` once before adding `.dark`, but never asserts light mode first; if the workspace boots with `.dark` already applied, both reads return the dark token.
9. **Test coverage shape** — three viewports in one `for`-loop hide which viewport regressed.
10. **Section padding contract** — `px-4` was added to the header but not to `main`, so children that don't self-pad will misalign with the title on mobile.

The focus-ring CSS change (`--ring: transparent` → colored oklch token) is a clean accessibility win.

frontend/src/components/landing/hero.tsx
```diff
@@ -36,27 +54,14 @@ export function Hero({ className }: { className?: string }) {
...
-              "Do Anything",
-              "Learn Anything",
-            ]}
-          />{" "}
-          <div>with DeerFlow</div>
+      <div className="container-md relative z-10 mx-auto flex min-h-[92svh] flex-col items-center justify-center px-4 pt-20 pb-14">
+        <h1 className="text-center text-5xl leading-tight font-bold break-words md:text-6xl">
+          DeerFlow
         </h1>
+        <div className="mt-3 flex w-full max-w-full min-w-0 flex-wrap items-center justify-center gap-x-2 gap-y-1 text-center text-2xl font-semibold md:text-4xl">
```
willem-bd: **flex-wrap orphans `"SuperAgent"` at 320px.** With longer rotating words (`"Generate Webpages"`, `"Generate Podcasts"`), the row exceeds 320px and `flex-wrap` pushes `"SuperAgent"` onto its own line, leaving an orphan label below the rotating word. The mobile-overflow test in `landing.spec.ts` only asserts `scrollWidth <= width`, so this passes CI while looking broken.

Options: drop `flex-wrap` and let the rotator shrink (`min-w-0` already present), or render `"SuperAgent"` inside the same `motion.div` so they always wrap together.

frontend/tests/e2e/ui-polish-mobile.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+        {
+          thread_id: MOCK_THREAD_ID,
+          title: "Thread with artifact",
+          artifacts: ["reports/mobile-summary.md"],
+        },
+      ],
+    });
+
+    await page.goto(`/workspace/chats/${MOCK_THREAD_ID}`);
+    await page.getByRole("button", { name: /artifacts/i }).click();
```
willem-bd: **Locale-coupled selector.** `name: /artifacts/i` only matches the en-US value of `t.common.artifacts` (`"Artifacts"`); zh-CN renders `"文件"` (see `src/core/i18n/locales/zh-CN.ts`), so if the test runner ever picks up zh-CN this fails to find the button.

Either force the locale in test setup (`process.env.NEXT_PUBLIC_LOCALE = "en-US"`) or query by a stable hook — e.g. `data-testid="artifact-trigger"` on `ArtifactTrigger`, or `getByRole("button")` filtered by the `FilesIcon` via a more stable selector.

frontend/tests/e2e/ui-polish-mobile.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      .toBeLessThanOrEqual(375);
+  });
+
+  test("global focus ring tokens are visible in light and dark themes", async ({
+    page,
+  }) => {
+    mockLangGraphAPI(page);
+    await page.goto("/workspace/chats/new");
+
+    const lightRing = await page.evaluate(() =>
```
willem-bd: **Test passes without proving the light variant.** The first read is called `lightRing` but the test never asserts `documentElement` is in light mode first. If a previous test or the workspace defaults `<html>` to `.dark`, the initial `getComputedStyle(...).getPropertyValue("--ring")` already returns the **dark** oklch token. Both reads then check the same non-transparent value and the test passes trivially — light-mode regression slips through.

Fix: `await page.evaluate(() => document.documentElement.classList.remove("dark"));` before the first read, and ideally assert each value matches its expected oklch token (or at least that the two reads differ).

frontend/tests/e2e/landing.spec.ts
```diff
@@ -6,17 +6,30 @@ test.describe("Landing page", () => {
...
+    await expect(page.locator("h1")).toHaveCount(1);
+    await expect(page.locator("h1")).toContainText("DeerFlow");
 
     // "Get Started" call-to-action button in hero
     await expect(
       page.getByRole("link", { name: /get started/i }),
     ).toBeVisible();
   });
 
+  test("does not overflow at common mobile widths", async ({ page }) => {
```
willem-bd: **One test, three viewports — lossy failure mode.** A `for`-loop over `[320, 375, 390]` reports a single test result; if 320 regresses but 375/390 are fine, Playwright only tells you the test failed — not which viewport. Retry reruns all three.

Split into three named tests (or `test.describe.each([320,375,390])`) so each viewport gets its own pass/fail signal and the suite parallelizes.

frontend/src/components/landing/section.tsx
```diff
@@ -12,18 +12,20 @@ export function Section({
...
+      className={cn("mx-auto flex w-full min-w-0 flex-col py-16", className)}
+    >
+      <header className="flex flex-col items-center justify-between px-4">
+        <div className="mb-4 max-w-full bg-linear-to-r from-white via-gray-200 to-gray-400 bg-clip-text text-center text-3xl font-bold break-words text-transparent sm:text-4xl md:text-5xl">
           {title}
         </div>
         {subtitle && (
-          <div className="text-muted-foreground text-center text-xl">
+          <div className="text-muted-foreground max-w-full text-center text-base break-words sm:text-lg md:text-xl">
             {subtitle}
```
willem-bd, line 24: **Padding contract split between header and main.** `px-4` is added to the inner `<header>` but not to the inner `<main className="mt-4 w-full min-w-0">`. Sections whose children don't self-pad will render with the title inset 16px but the body flush to the viewport edge on mobile — a hidden contract that the next contributor adding a section is likely to trip over.

Moving `px-4` to the outer `<section>` (and dropping it from the header) makes the gutter uniform and removes the per-section responsibility.

WillemJiang:
@18062706139fcz  Please fix the lint error and resolve the conflicts with the main branch.

WillemJiang (approved):

PR #3741, opened 2026-06-23
fix: allow unauthenticated access to mock case study pages

singlaamitesh:
The workspace layout's auth guard redirects unauthenticated visitors to /login before the chat page can read the `mock=true` query parameter, making public case study demos unreachable.

Security fix: scope the bypass to only specific demo thread IDs via a server-side allowlist, and only when the URL matches `/workspace/chats/`. This prevents `?mock=true` from being used to bypass auth on other workspace routes (agents, settings, root).

Fixes #3693

### DIFF
--- a/frontend/src/app/workspace/layout.tsx
+++ b/frontend/src/app/workspace/layout.tsx
@@ -11,7 +11,17 @@ export const dynamic = "force-dynamic";
 
 export default async function WorkspaceLayout({
   children,
-}: Readonly<{ children: React.ReactNode }>) {
+  searchParams,
+}: Readonly<{
+  children: React.ReactNode;
+  searchParams: Promise<{ mock?: string }>;
+}>) {
+  // Allow unauthenticated access for public case study demos
+  const params = await searchParams;
+  if (params?.mock === "true") {
+    return <WorkspaceContent>{children}</WorkspaceContent>;
+  }
+
   const result = await getServerSideUser();
 
   switch (result.tag) {


frontend/src/app/workspace/layout.tsx
```diff
@@ -11,7 +11,17 @@ export const dynamic = "force-dynamic";
...
-}: Readonly<{ children: React.ReactNode }>) {
+  searchParams,
+}: Readonly<{
+  children: React.ReactNode;
+  searchParams: Promise<{ mock?: string }>;
+}>) {
+  // Allow unauthenticated access for public case study demos
+  const params = await searchParams;
+  if (params?.mock === "true") {
+    return <WorkspaceContent>{children}</WorkspaceContent>;
```
fancyboi999, line 22: **P1 (security):** `?mock=true` is client-controlled, and this check lives in the layout that guards *all* of `/workspace/*`. Any visitor can bypass `getServerSideUser()` on any workspace URL (e.g. `/workspace/chats/<thread_id>?mock=true`), not just case-study demos. Scope the bypass to the actual demo pages — specific routes, a server-side allowlist of demo thread ids, or a dedicated public route outside this layout — rather than a blanket skip keyed on a query param.

fancyboi999 (commented):
## Review

**P1 (security — auth bypass).** This gates the bypass on a **client-controlled query param at the whole-`/workspace` layout level**, which turns `?mock=true` into a blanket auth-skip for every route under `/workspace`. `WorkspaceLayout` is the auth gate for the entire `/workspace/*` subtree — it's the thing that runs `getServerSideUser()` before rendering. Returning early on `params.mock === "true"` means anyone can append `?mock=true` to **any** workspace URL (e.g. `/workspace/chats/<thread_id>?mock=true`) and render the authenticated workspace shell without logging in — far broader than the "public case study demos" intent in #3693.

Even if the backend APIs still enforce per-user auth (limiting actual data exfiltration), this defeats the intended front-end auth boundary, is trivially triggerable, and applies to every workspace page rather than just the demo/case-study ones.

**Suggested direction:** scope the public-demo path explicitly instead of blanket-skipping the workspace auth guard — restrict to the specific case-study routes, gate on a server-side allowlist of demo thread ids, or move the public demo to a dedicated route that doesn't live under this auth-gated layout. A client-supplied `?mock=true` shouldn't be sufficient to skip authentication.

PR #3727, opened 2026-06-23
feat(setup): ask custom OpenAI-compatible users about thinking support

wuhongqiang888:
## Summary

When configuring the **Other OpenAI-compatible** provider in `make setup`, the wizard now asks whether the model supports thinking/reasoning output. If the user confirms, the generated config.yaml includes `supports_thinking: true` and `when_thinking_enabled`/`when_thinking_disabled` blocks.

Built-in providers (OpenAI, Anthropic, DeepSeek, etc.) are not affected - they already have pre-configured thinking support.

## Changes
- `scripts/wizard/steps/llm.py`: Add `ask_yes_no` prompt for thinking support after base_url entry; add `supports_thinking` field to `LLMStepResult`
- `scripts/setup_wizard.py`: Merge `OPENAI_COMPAT_THINKING_CONFIG` into extra model config when user confirms thinking support

Fixes #3162

PR #3743, opened 2026-06-23
feat(setup): ask custom OpenAI-compatible users about thinking support

singlaamitesh:
When configuring the **Other OpenAI-compatible** provider in `make setup`, the wizard now asks whether the model supports thinking/reasoning output. If the user confirms, the generated config.yaml includes `supports_thinking`, `when_thinking_enabled`, and `when_thinking_disabled` blocks.

Built-in providers (OpenAI, Anthropic, DeepSeek, etc.) are not affected — they already have pre-configured thinking support.

### DIFF
--- a/scripts/wizard/steps/llm.py
+++ b/scripts/wizard/steps/llm.py
@@ -4,11 +4,16 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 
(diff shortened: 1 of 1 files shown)


WillemJiang:
duplicated PR of https://github.com/bytedance/deer-flow/pull/3727

PR #3744, opened 2026-06-23
docs(atlascloud): expand official Atlas model list

singlaamitesh:
Add the full curated Atlas Cloud chat model id list, grouped by provider (Anthropic, OpenAI, Google, DeepSeek, Qwen, Kimi, GLM, MiniMax, xAI, ByteDance, KAT), as a comment block in the existing Atlas Cloud example section. Users can now drop any model id into the `model:` field without leaving the file.

Follow-up to #3704.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -496,6 +496,150 @@ models:
   #     extra_body:
   #       thinking:
   #         type: disabled
+  #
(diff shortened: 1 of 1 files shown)


WillemJiang:
Please don't list all the supported models there.

PR #3745, opened 2026-06-23
fix(middleware): fix positional fallback consuming unrelated todo when same-content list is exhausted

singlaamitesh:
When `next_todos` contains the same content string twice, the first match pops from `previous_by_content[content]`, leaving an empty list. The second match sees the empty list (falsy), and the positional fallback would unconditionally consume `previous_todos[index]` even when it holds a completely different entry.

Fix: use `is not None` instead of truthiness check to distinguish "content was never in previous" (None) from "all content matches consumed" (empty list). Change the positional fallback to `elif` so it only fires when the key was never in `previous_by_content`.


WillemJiang:
duplicated PR of #3709

PR #3751, opened 2026-06-24
feat(skills): add stock-support-finder skill

Charles-Liang1229:
Adds a new public skill that detects support and resistance levels from daily OHLC price data using swing-point detection and clustering. Includes MA50, RSI(14) entry condition checks and stop-loss suggestions based on nearest support.

### DIFF
--- a/skills/public/stock-support-finder/scripts/find_support.py
+++ b/skills/public/stock-support-finder/scripts/find_support.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python3
+"""
+Stock Support & Resistance Level Finder
+Fetches daily OHLC data, detects swing lows/highs, clusters them into
+meaningful support/resistance zones, and outputs a trading report.
+"""
+
+import argparse
+import sys
+from datetime import datetime, timedelta
+
+try:
+    import yfinance as yf
+    import numpy as np
+except ImportError:
+    print("Missing dependencies. Run: pip install yfinance numpy")
+    sys.exit(1)
+
+
+def fetch_data(ticker: str, months: int = 6):
+    end = datetime.today()
+    start = end - timedelta(days=months * 30)
+    df = yf.download(ticker, start=start.strftime("%Y-%m-%d"), end=end.strftime("%Y-%m-%d"),
+                     auto_adjust=True, progress=False)
+    if df.empty:
+        print(f"❌ No data found for '{ticker}'. Check the ticker symbol.")
+        sys.exit(1)
+    # Flatten MultiIndex columns if present
+    if isinstance(df.columns, type(df.columns)) and hasattr(df.columns, 'levels'):
+        df.columns = df.columns.get_level_values(0)
+    return df
+
+
+def find_swing_lows(closes, window: int = 5):
+    """Local minima: lower than `window` bars on each side."""
+    lows = []
+    arr = list(closes)
+    for i in range(window, len(arr) - window):
+        segment = arr[i - window: i + window + 1]
+        if arr[i] == min(segment):
+            lows.append((i, arr[i]))
+    return lows
+
+
+def find_swing_highs(closes, window: int = 5):
+    """Local maxima: higher than `window` bars on each side."""
+    highs = []
+    arr = list(closes)
+    for i in range(window, len(arr) - window):
+        segment = arr[i - window: i + window + 1]
+        if arr[i] == max(segment):
+            highs.append((i, arr[i]))
+    return highs
+
+
+def cluster_levels(points, tolerance_pct: float = 0.015):
+    """
+    Group price points within `tolerance_pct` of each other into one zone.
+    Returns list of (avg_price, touch_count) sorted by touch count desc.
+    """
+    if not points:
+        return []
+    prices = sorted([p for _, p in points])
+    clusters = []
+    current = [prices[0]]
+    for p in prices[1:]:
+        if (p - current[0]) / current[0] <= tolerance_pct:
+            current.append(p)
+        else:
+            clusters.append(current)
+            current = [p]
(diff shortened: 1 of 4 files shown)


skills/public/stock-support-finder/scripts/find_support.py
```diff
@@ -0,0 +1,233 @@
...
+    print("─" * 58)
+    cond_ma = ma50 and current_price > ma50
+    cond_rsi = 35 <= rsi <= 65
+    all_pass = cond_ma and cond_rsi
+    print(f"  价格在 MA50 上方 : {'✅' if cond_ma else '❌'}")
+    print(f"  RSI 在 35–65    : {'✅' if cond_rsi else '❌'} (当前 {rsi})")
+    print(f"  MACD 金叉       : ⚠️  请在 Moomoo 图表上手动确认")
+    print()
+    if all_pass:
+        nearest_support = supports[0][0] if supports else None
```
fancyboi999: **P2:** `supports` comes back sorted by touch-count (strongest first, from `cluster_levels`), so `supports[0]` is the *most-touched* support, not the nearest one to price — yet it's used as `nearest_support` for the stop and R:R. Because the supports filter allows `p < current_price * 1.03`, this can even be a level *above* the current price, making `stop = nearest_support * 0.99` land above price (negative risk → `rr` forced to 0, and a printed stop above entry). Use the nearest support below price here, e.g. `min((p for p, _ in supports if p < current_price), key=lambda p: current_price - p, default=None)`. Same issue for `target = resistances[0][0]` on L209 — you want the nearest resistance *above* price.

skills/public/stock-support-finder/scripts/find_support.py
```diff
@@ -0,0 +1,233 @@
...
+    print("─" * 58)
+    print(f"  当前价格  : ${current_price}")
+    print(f"  MA50      : ${ma50 or 'N/A'}  ({price_vs_ma50})")
+    print(f"  MA200     : ${ma200 or 'N/A'}  ({price_vs_ma200})")
+    print(f"  RSI(14)   : {rsi}  —  {rsi_label}")
+    print()
+
+    # Detect swing points
+    swing_lows = find_swing_lows(closes, args.window)
+    swing_highs = find_swing_highs(closes, args.window)
```
fancyboi999: **P2:** swings are detected on `closes`, but `highs_raw` / `lows_raw` (L121–122) are fetched and never used. For a support/resistance finder the levels usually come from intraday wicks — support from `Low`, resistance from `High`. Either pass `lows_raw` to `find_swing_lows` and `highs_raw` to `find_swing_highs`, or remove the two unused fetches; as written they're dead code and the detection misses wick-based levels.

fancyboi999 (commented):
Nice self-contained skill, and it follows the existing `skills/public/<name>/{SKILL.md, scripts/}` layout. Two correctness issues in `find_support.py` are worth fixing before merge — both are in the analysis logic, not style.

**P2 — the "nearest" support/resistance used for the trade plan is actually the *strongest* level, not the nearest.** `cluster_levels()` returns clusters sorted by touch-count desc (`sorted(result, key=lambda x: x[1], reverse=True)`), so `supports[0]` / `resistances[0]` are the most-touched zones, not the closest to price. The entry block then uses them as `nearest_support` / "最近阻力" for the stop, target and R:R. Concrete failure mode: the supports filter admits `p < current_price * 1.03`, so the strongest "support" can sit up to 3% *above* the current price → `stop = nearest_support * 0.99` can land **above** the current price, printing a nonsensical stop-loss for a long and forcing `rr = 0` (risk goes negative). Even when it is below price, choosing the strongest-but-far level over a nearer one mis-states risk and the R:R verdict.

**P2 — swing detection runs on `Close`, while the `High`/`Low` you fetch are dead.** `highs_raw` / `lows_raw` (L121–122) are never read; swings come from `closes` (L157–158). Support/resistance are conventionally anchored on intraday wicks (Low for support, High for resistance), and you already pull them. Either feed `lows_raw` / `highs_raw` into `find_swing_lows` / `find_swing_highs`, or drop the two unused fetches. (Note: the backend lint job runs in `backend/` only, so `skills/` isn't linted — flagging this as a real gap, not something CI will catch.)

Inline details below. The disclaimer and rule-based framing are good. Happy to take another look once these are addressed.

fancyboi999:
Thanks for putting this together, @Charles-Liang1229. It's clean and self-contained, and the layout follows the existing `skills/public/<name>/{SKILL.md, scripts/}` convention.

Following up on my earlier review: stepping back from the line-level fixes, there's a more basic fit question I should have raised first. `skills/public/` ships to every DeerFlow user as part of the shared catalog, so it's meant for general-purpose capabilities. This skill is tailored to a personal trading setup: it emails a daily report to specific personal addresses, references manual steps in Moomoo, and hardcodes a personal risk/position-sizing ruleset (2% risk per trade, 25% single-stock cap). It's a genuinely useful personal tool, but narrower than what the public catalog is for, so I don't think it fits under `skills/public/`.

You don't have to lose the work, though. DeerFlow has a first-class path for exactly this: put it under `skills/custom/stock-support-finder/` (that directory is gitignored and auto-discovered on the next `load_skills()`), or use "Create Your Own Skill" in the UI. No PR needed and it stays private to your setup. See the "Custom skills" section in the harness customization docs.

One heads-up regardless of where it lands: the `EMAIL_TO` list near the top of `daily_report.py` hardcodes two personal email addresses. This branch is public, so they're already exposed in the PR diff and commit history. Worth moving them to an env var or CLI flag, and note that closing this PR won't remove them from the history here.

Given the scope, I'll close this for now. If you'd like to carve out a general, non-personal version of the analysis pieces, I'm glad to help think that through.

Issue #3739, opened 2026-06-23
RFC: GitHub as a Webhook-Driven Channel


zhfeng:
# RFC: GitHub as a Webhook-Driven Channel

## 1. Problem

DeerFlow integrates with IM platforms (Feishu, Slack, Telegram, Discord, DingTalk) through a uniform `Channel` abstraction on a `MessageBus`. Each implementation owns its long-poll / WebSocket listener, translates the platform's wire format into `InboundMessage`, and renders `OutboundMessage` back to the platform.

GitHub does not fit this shape. There is no long-poll API; events arrive as HTTP POST webhooks. A repository can have many distinct "conversations" (issue threads, PR threads, review threads) that are not 1:1 with a chat. Multiple agents may need to react to the same delivery — e.g. a coder that implements issues, a reviewer that comments on PRs. And the cost model is different: every reply is a permanent, public, often-emailed comment, so `(No response from agent)` placeholders are not acceptable.

Without channel support, the only way to involve a DeerFlow agent in a GitHub repo today is manual: paste an issue into the UI, copy the response back. We want the agent loop to close inside the platform.

## 2. Goals

- A repo can register one or more DeerFlow agents to react to specific GitHub events (`issues`, `issue_comment`, `pull_request`, review events).
- One webhook delivery can fan out to multiple bound agents, each on its own LangGraph thread.
- An agent that has nothing useful to add to a particular event stays silent — no placeholder reply.
- An agent never re-fires on its own activity, but external bots (Copilot, CodeRabbit, Dependabot) reach it normally.
- The webhook route stays cheap so GitHub's 10s delivery timeout is never at risk.
- The integration sits alongside the other IM channels on the existing `MessageBus`, not as a separate runtime.

## 3. Non-Goals

- Line-level GitHub reviews (`POST /pulls/{n}/reviews`) — first cut posts plain comments via `gh`.
- A general-purpose GitHub API client. The channel only handles inbound webhooks; outbound posting is delegated to `gh` in the agent's sandbox.
- Replacing existing IM channels. GitHub is added; nothing else changes.
- Auto-replying to bot-authored PRs the agent itself opened (covered by the self-event gate, see §6).

### 4.1 Dependency direction

```
app.gateway.routers.github_webhooks   (HTTP entry, HMAC verify)
  └── app.gateway.github.dispatcher   (target extraction, fan-out, gates)
        ├── app.gateway.github.identity   (repo/number/thread_id)
[...]

PR #3754, opened 2026-06-24
feat(channels): add GitHub as a webhook-driven channel

zhfeng:
## Summary

Adds GitHub as a webhook-driven `Channel` on the existing `MessageBus`. A custom agent declares `github:` bindings in its `config.yaml`; the gateway fans each verified webhook delivery out to every bound agent on its own deterministic LangGraph thread, and the agent uses the `gh` CLI in its sandbox to read repo state and post replies. The channel itself is log-only — the LLM decides when to post.

Closes the design discussion in [RFC #3739](https://github.com/bytedance/deer-flow/issues/3739) (kept in sync with the implementation).

## What ships

**Inbound** — `POST /api/webhooks/github`
- HMAC verification (`X-Hub-Signature-256`), exempt from auth + CSRF middleware (HMAC is the auth).
- `fanout_event` extracts `(repo, number)`, looks up bound agents via an mtime-cached registry, applies a per-agent self-event gate and a per-binding trigger filter, then publishes one `InboundMessage` per surviving agent onto the bus.
- Returns 200 to GitHub immediately so the 10s delivery timeout is never at risk; the bus consumer runs the agent off-thread.

**Outbound** — `GitHubChannel.send()` is log-only
- The agent posts mid-run via `gh issue comment` / `gh pr comment` / `gh pr create`. The channel does not auto-post the final assistant message, so silent runs are silent (no `(No response from agent)` on PRs).
- Per-run installation token flows through `run_context["github_token"]` as a refresh-capable provider closure — the bash tool calls it on every invocation, so a 90-min coder run gets a fresh token at minute ~55 instead of 401-ing on the final `git push`.

**Per-channel run policy** — `app.channels.run_policy.ChannelRunPolicy`
- One table that drives `is_interactive`, `default_recursion_limit`, and `credentials_provider`. The GitHub entry is registered as an import side-effect of `app.gateway.github.run_policy`. Adding the next webhook channel is a one-row registration, no edits to `ChannelManager`.

**Config** — additive only
- `AgentConfig.github` is an optional new top-level block. Existing custom agents load and run unchanged.
- `GitHubAgentConfig` carries `installation_id`, `bot_login`, `recursion_limit`, and `bindings` (each binding has `repo` + opt-in `triggers` map).

**Sandbox** — `Sandbox.execute_command(*, extra_env=...)`
- Per-call env overlay (keyword-only, defaults to None — no behavior change for existing callers).
- `LocalSandbox` passes through `subprocess.run(env=...)`; `AioSandbox` wraps in `( export …; cmd )` so the export does not leak onto its persistent shell session.

## Test plan

- 396 tests across `tests/test_github_*.py`, `tests/test_channels.py`, `tests/test_update_agent_tool.py` cover the dispatcher fan-out, trigger filter (action whitelist, `require_mention` per event type, `allow_authors`, `mention_login` override, boundary-aware match), self-event gate (including the `agent.name` fallback semantics), registry mtime cache (warm/cold, invalidation on add/edit/delete), token plumbing (provider callable, refresh, missing-installation degradation), per-installation lock concurrency, `_create_thread` race recovery, and the channel-run-policy table.
- Full test suite: 5169 passed, 18 skipped.

## Code review applied

Ten findings from a high-effort code review pass were fixed before this PR. Highlights:

1. `update_agent_tool` preserves every non-managed `AgentConfig` field (was silently stripping the `github:` block on description/model/tool_groups/skills changes).
2. `AioSandbox` wraps `extra_env` exports in a subshell to prevent token leaks across calls on its persistent shell session.
3. Self-event gate makes `agent.name` a true fallback (only added to the self-identity set when no `bot_login` / `mention_login` is configured), so a real GitHub user named `reviewer` is not silently filtered.
4. `_create_thread` recovers from the race where two webhooks for the same `(repo, number)` both fall through to `threads.create(thread_id=<same uuid5>)`.
5. `_comment_body` reads `payload.review.body` for `pull_request_review` events (was returning empty string).
6. `@`-mention match is boundary-aware on `[A-Za-z0-9-]`, so `@deerflow` does not match `@deerflow-bot`.
7. Installation token refresh via provider callable in `run_context["github_token"]` — long runs past the 1h TTL get a fresh token.
8. Registry uses an mtime-keyed cache; `fanout_event` runs the scan via `asyncio.to_thread` so disk I/O never blocks the event loop.
9. `mint_installation_token` uses a per-installation `asyncio.Lock` and a lock-free cache-hit fast path, so a slow mint for installation A never blocks lookups for installation B.
10. Three `if msg.channel_name == "github"` branches in `ChannelManager` collapsed into a `CHANNEL_RUN_POLICY` lookup driven by `app.channels.run_policy.ChannelRunPolicy`.

## Out of scope (future)

- Line-level GitHub reviews (`POST /pulls/{n}/reviews`).
- `pull_request.synchronize` / `reopened` events — easy to add per binding once needed.
- Discussions (`discussion`, `discussion_comment`) — different data model, separate extraction work.
- Trusted-bot allowlist for the self-event gate.


backend/app/gateway/routers/github_webhooks.py
```diff
@@ -0,0 +1,244 @@
...
+    - Returns ``{"ok": True, ...}`` so GitHub marks the delivery successful
+      and does not retry.
+
+    If ``GITHUB_WEBHOOK_SECRET`` is unset the request is accepted without
+    verification, with a WARNING in the log. Set the env var in production.
+    """
+    body = await request.body()
+
+    secret = _get_webhook_secret()
+    if secret is None:
```
fancyboi999, line 208: **P2 (security):** fail-open here means an unset `GITHUB_WEBHOOK_SECRET` accepts **unverified** deliveries with only a WARNING. This route is auth/CSRF-exempt and side-effectful — `fanout_event` runs bound agents on attacker-controlled payload text and posts as the App — so one missing env var = an unauthenticated endpoint that can trigger agent runs (with their tools) and comment as your bot, plus a prompt-injection vector via the forged issue/PR/comment body.

Recommend fail-**closed**: when `_get_webhook_secret()` returns `None`, reject the delivery (503/401) or skip mounting the route, instead of accepting it. The safe default for a privileged, internet-exposed surface is inert-until-secured, not open-with-a-log-line.

fancyboi999 (commented):
Focused review of the **webhook entry path** (the security-critical surface) — I didn't go through all 39 files, flagging that so the scope is clear. The HMAC verification itself is done right: constant-time `hmac.compare_digest`, verify-then-parse ordering, empty-string-secret treated as unset, 401 on mismatch, 200 on unknown events to avoid GitHub retry storms. The dispatcher's self-event / bot-loop gate also reads carefully.

One security concern worth a decision before this lands (inline):

**P2 (security) — the webhook fails _open_ when `GITHUB_WEBHOOK_SECRET` is unset.** This route is auth- and CSRF-exempt by design, internet-facing (it has to be), and side-effectful: a delivery fans out to bound agents → runs the lead agent on attacker-controllable payload text (issue / PR / comment bodies) → posts back as your GitHub App. With the secret unset, that entire surface is reachable by anyone who can `POST /api/webhooks/github`, gated only by a WARNING log. A single missed env var turns it into an unauthenticated agent-trigger + comment-as-bot + prompt-injection surface.

I'd make it fail _closed_: when no secret is configured, reject the delivery (or don't register the route at all) so the channel stays inert until it's secured, rather than relying on the operator noticing the warning. "Set it in production" is a lot of trust to place on operator diligence for a privileged, internet-exposed endpoint.

backend/app/channels/manager.py
```diff
@@ -1202,10 +1292,37 @@ async def _create_thread(self, client, msg: InboundMessage) -> str:
...
+        if isinstance(preferred_thread_id, str) and preferred_thread_id:
+            create_kwargs["thread_id"] = preferred_thread_id
         if owner_headers:
-            thread = await client.threads.create(metadata=metadata, headers=owner_headers)
-        else:
-            thread = await client.threads.create(metadata=metadata)
+            create_kwargs["headers"] = owner_headers
+        try:
+            thread = await client.threads.create(**create_kwargs)
+        except Exception as exc:
```
willem-bd: **Race-recovery catches bare `Exception` — any failure becomes a "race," not just duplicate-id conflicts.**

When the gateway has a transient DB outage and the very first webhook for a new `(repo, number)` hits this block, `client.threads.create(thread_id=...)` raises (HTTPStatusError, network timeout, 5xx). The except branch treats it as a race, writes `preferred_thread_id` into the store via `_store_thread_id`, and returns the id — but the thread never actually got created.

Every subsequent webhook for that PR then finds the mapping in `_lookup_thread_id`, calls `runs.create` on a non-existent thread, and 404s permanently. There is no code path that retries the create because the store now lies.

**Suggested fix:** narrow the except to specifically the duplicate-id condition (e.g. inspect the HTTP status / error body for 409/422 / `thread_id already exists`), and add a defensive `await client.threads.get(preferred_thread_id)` before returning the id, so a non-race failure surfaces as a real error rather than silently corrupting the store.

backend/app/gateway/github/dispatcher.py
```diff
@@ -0,0 +1,237 @@
...
+            logger.info(
+                "github_fanout: agent=%s skipped (reason=self_event, sender=%s)",
+                b.agent.name,
+                sender_login,
+            )
+            skipped.append({"agent": b.agent.name, "reason": "self_event"})
+            continue
+
+        # 5. Trigger filter.
+        fire, reason = event_should_fire(event, payload, triggers, b.agent.name)
```
willem-bd: **Mention fallback uses `agent.name`, never `bot_login` — asymmetric with the self-event gate three lines up.**

The self-event gate at line 156 correctly prefers `b.github.bot_login`. This call passes `b.agent.name` as the `default_mention_login`, so an operator who sets `github.bot_login: deerflow-bot` and a binding `triggers: { issue_comment: {} }` (no per-trigger `mention_login` override — they reasonably expect `bot_login` to be the agent's @-handle) will see every `@deerflow-bot review please` rejected with `mention required for @<agent-dir-name>`.

**Suggested fix:**
```python
fire, reason = event_should_fire(event, payload, triggers, b.github.bot_login or b.agent.name)
```
Matches the precedence used by `_is_self_event` so both gates agree on who this agent is.

backend/app/gateway/routers/github_webhooks.py
willem-bd, line 208: **Unset/empty `GITHUB_WEBHOOK_SECRET` accepts every delivery without verification — combined with the auth+CSRF exemption this is a SSRF-ish public attack surface.**

The route is exempt from both auth and CSRF (see `auth_middleware._PUBLIC_PATH_PREFIXES` and `csrf_middleware.should_check_csrf`), so when the secret env var is missing — a fresh staging box, a Helm chart misconfig, a stray `.env.local` typo — there is nothing left guarding it. An attacker who can reach the URL POSTs a forged `pull_request_review_comment` referencing a repo with bound agents. `fanout_event` mints an installation token and runs the agent with the (attacker-controlled) comment body as the user prompt, including any prompt-injection payload they choose. The single WARNING log line is easy to miss.

**Suggested fix:** fail-closed by default — refuse to start the route (or `raise HTTPException(503)` here) when the secret is unset, gated by an explicit opt-in like `DEER_FLOW_ALLOW_UNVERIFIED_GITHUB_WEBHOOKS=1` for the dev/loopback case.

backend/app/channels/manager.py
```diff
@@ -887,8 +900,85 @@ def _resolve_run_params(self, msg: InboundMessage, thread_id: str) -> tuple[str,
...
+        policy = CHANNEL_RUN_POLICY.get(msg.channel_name)
+        if policy is not None and policy.default_recursion_limit is not None:
+            # Per-message override (via msg.metadata) wins over channel
+            # default; both must clear the user/channel-config base
+            # (max(...)) so an explicit higher value in session config
+            # is never lowered.
+            channel_meta = (msg.metadata or {}).get(msg.channel_name, {})
+            override = channel_meta.get("recursion_limit") if isinstance(channel_meta, dict) else None
+            limit = override if isinstance(override, int) and override > 0 else policy.default_recursion_limit
+            run_config["recursion_limit"] = max(run_config.get("recursion_limit", 100), limit)
```
willem-bd: **Per-agent `recursion_limit` can only raise the ceiling, never lower it.**

`max(run_config.get("recursion_limit", 100), limit)` silently clamps any per-agent value below 100 up to 100. A safety-conscious operator who sets `github.recursion_limit: 50` on a review-only agent expects the run to halt at 50 super-steps; instead the agent gets 100. This contradicts the documented intent in `GitHubAgentConfig.recursion_limit` ("a review-only agent might be happy at 100, a multi-file refactor agent might need 500+").

**Suggested fix:** when a per-message override is present, honor it directly instead of `max()`-ing against the previous value:
```python
if isinstance(override, int) and override > 0:
    run_config["recursion_limit"] = override
else:
    run_config["recursion_limit"] = max(run_config.get("recursion_limit", 100), policy.default_recursion_limit)
```
Or, if the floor is intentional, document it on the field's docstring so operators don't expect a setting that silently doesn't apply.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
```diff
@@ -122,12 +122,29 @@ def execute_command(self, command: str) -> str:
...
             try:
+                if extra_env:
+                    # shlex.quote makes the values safe to splice into an
+                    # export statement even if they contain quotes, spaces,
+                    # or shell metacharacters. Wrapping the whole thing in
+                    # ``(...)`` runs it in a subshell so the export does
+                    # NOT persist on this AioSandbox's long-lived shell
+                    # session — critical for not leaking GH_TOKEN across
+                    # runs that share a sandbox via warm-pool reuse.
+                    exports = "".join(f"export {k}={shlex.quote(v)}; " for k, v in extra_env.items())
```
willem-bd: **`extra_env` keys are spliced raw into the export statement — shell-metachar in a key becomes command injection.**

Only values are `shlex.quote`-d; keys go straight in:
```python
exports = "".join(f"export {k}={shlex.quote(v)}; " for k, v in extra_env.items())
command = f"( {exports}{command} )"
```

The abstract `Sandbox.execute_command(extra_env: dict[str, str] | None = None)` contract says nothing about key shape. Today the only call site passes `{"GH_TOKEN": ..., "GITHUB_TOKEN": ...}`, so this is latent — but a future channel or test fixture passing `{"X;rm -rf /mnt/user-data;Y": "v"}` produces `export X;rm -rf /mnt/user-data;Y='v'; <cmd>` running inside the persistent shell session.

**Suggested fix:** validate keys against `^[A-Za-z_][A-Za-z0-9_]*$` (the POSIX env-var name rule) before splicing, and raise on violation. Same check applies to `local_sandbox.py` for defense in depth, even though `subprocess.run(env=...)` is safer there.

backend/app/gateway/github/app_auth.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+_token_cache: dict[int, _CachedToken] = {}
+# Per-installation locks so a cold mint for installation A does not block
+# concurrent lookups (cache hits OR independent cold mints) for any other
+# installation. A single process-wide lock would serialize the entire
+# fleet behind one slow GitHub /access_tokens roundtrip; this map gives
+# each installation its own ~hundreds-of-ms HTTPS critical section while
+# letting unrelated installations proceed concurrently.
+_install_locks: dict[int, asyncio.Lock] = {}
```
willem-bd, line 65: **`_install_locks` and `_token_cache` grow monotonically — no eviction when installations are uninstalled.**

Fine for a single-tenant install, but the module comment says "intentionally simple" without mentioning the unbounded-size assumption. A multi-tenant gateway serving thousands of installations over its lifetime accumulates one `_CachedToken` + one `asyncio.Lock` per installation forever (small per entry — not catastrophic, but never released).

**Suggested fix (pick one):**
1. Bounded LRU dict (cap ~10k installs, evict the coldest); or
2. Evict entries whose `expires_at + grace` is in the past on every mint call; or
3. Add a note to the module docstring that operators of large multi-tenant deployments should restart periodically / size accordingly.

Option 2 is the cheapest — fold an eviction sweep into `_request_new_installation_token` so cache-miss paths pay the cleanup cost.

backend/app/channels/manager.py
```diff
@@ -887,8 +900,85 @@ def _resolve_run_params(self, msg: InboundMessage, thread_id: str) -> tuple[str,
...
+                # Credential failures must NOT drop the delivery — the
+                # provider's own logging records the cause; we keep the
+                # run going (read-only is better than no response).
+                logger.warning(
+                    "[Manager] channel=%s credentials_provider raised; run proceeds without injected credentials",
+                    msg.channel_name,
+                    exc_info=True,
+                )
+
+    async def _inject_github_credentials(self, msg: InboundMessage, run_context: dict[str, Any]) -> None:
```
willem-bd: **Backward-compat shims exist only because tests reach in directly — production routes through `_apply_channel_policy`.**

Both `_inject_github_credentials` and `_mark_non_interactive` (line 970) say in their own docstrings that they exist to keep existing tests green. The production dispatch path calls `_apply_channel_policy`, which derives non-interactivity from `CHANNEL_RUN_POLICY[channel].is_interactive` and invokes the same provider closure.

This is the altitude smell called out in CLAUDE.md: special-cased methods layered on shared infrastructure. As the next webhook channel ships, either the shim pattern gets duplicated, or these methods hang as dead code on `ChannelManager` forever.

**Suggested fix:** migrate `tests/test_channels.py` / `tests/test_gateway_services.py` to call `_apply_channel_policy` (the real entry point), then delete the two shims. The tests get exercised against the real code path; the manager's public surface shrinks back to one channel-policy hook.

willem-bd (commented):
Code review (high effort) — 7 findings surfaced across correctness, security, altitude, and code-shape angles. Ranked most-severe first; details inline.

1. **`_create_thread` race-recovery catches bare `Exception`** — any DB/network failure on a webhook with a deterministic `preferred_thread_id` is misclassified as a race and the store is poisoned with a mapping to a thread that never existed.
2. **Mention fallback ignores `bot_login`** — `event_should_fire(...)` is called with `b.agent.name` as the default mention login, never `b.github.bot_login`. Asymmetric with the self-event gate, so `@deerflow-bot` is silently ignored when only `bot_login` is set.
3. **Webhook secret-unset path accepts every delivery** — combined with the auth+CSRF exemption, an attacker reaching `/api/webhooks/github` on a misconfigured gateway can forge payloads and trigger agent runs with attacker-chosen prompts.
4. **Per-agent `recursion_limit` can only raise, never lower** — `max(run_config[...], limit)` clamps below 100 to 100, silently breaking a `recursion_limit: 50` safety setting.
5. **`extra_env` keys spliced unvalidated** into the export statement — currently only the hardcoded `GH_TOKEN`/`GITHUB_TOKEN` reach this path so it's latent, but the public `Sandbox.execute_command(extra_env=...)` contract accepts arbitrary keys and a metachar in a key becomes command injection on the persistent shell session.
6. **`_install_locks` and `_token_cache` grow unbounded** — fine for a single tenant, but a multi-tenant deployment serving thousands of installations has no eviction path. Either add a cap or document the growth.
7. **`_inject_github_credentials` / `_mark_non_interactive` shims exist only to keep tests green** — the production path routes through `_apply_channel_policy`. The shims either repeat for every new webhook channel or hang as dead code; CLAUDE.md's altitude rule applies.

The trigger filter, prompt builders, mtime-cached registry, and per-installation lock design are otherwise solid — most findings are defense-in-depth tightening, not show-stoppers. #3 (unset-secret accepts unverified deliveries) is the only one I'd block the merge on.

backend/app/gateway/routers/github_webhooks.py
zhfeng: Fixed in 47d659c on this branch — the route is now fail-closed.

- `app/gateway/app.py` only mounts the router when `github_webhooks.is_route_enabled()` returns True.
- `is_route_enabled()` returns True iff `GITHUB_WEBHOOK_SECRET` is set OR `DEER_FLOW_ALLOW_UNVERIFIED_GITHUB_WEBHOOKS=1` (explicit dev/loopback opt-in).
- Default deployment with neither env var → the URL responds 404. No handler exists to be reached, so a misconfig cannot serve forged deliveries even by accident.
- Defense in depth: the runtime handler also returns 503 if the secret was rotated/cleared without a process restart, so a mounted-but-unprotected state never exists.

Thanks for catching this — it was the exact privileged-endpoint-with-side-effects shape that needed inert-until-secured, not log-and-hope.

zhfeng: Fixed in 47d659c — fail-closed. `app/gateway/app.py` consults `github_webhooks.is_route_enabled()` before mounting the router. The route is mounted only when `GITHUB_WEBHOOK_SECRET` is set, OR when the explicit dev opt-in `DEER_FLOW_ALLOW_UNVERIFIED_GITHUB_WEBHOOKS=1` is set — matching the env-var name shape you suggested. Default deployment with neither env var → 404 (no handler exists to reach). The unverified-accept code path logs a loud `UNVERIFIED delivery ... dev/loopback mode ONLY` warning so operators do not mistake it for production-safe.

Departure from the suggestion: I also kept a defense-in-depth 503 inside the handler for the case where the secret is rotated/cleared *after* startup without a process restart — so a mounted-but-unprotected state never exists even briefly. The PR description has been updated as BREAKING CHANGE for deployments that relied on the previous unverified-accept behaviour.

backend/app/channels/manager.py
zhfeng: Fixed in 8df6145. The except clause now catches only `langgraph_sdk.errors.ConflictError` (HTTP 409) — the only failure shape that legitimately means "another writer just created this thread." Non-conflict failures (network error, DB outage, 5xx from the underlying thread_store) propagate cleanly so the delivery fails fast and the store stays clean.

Also added the defensive `client.threads.get(preferred_thread_id)` verify step you suggested: if ConflictError fires but the follow-up `threads.get` also fails (stale conflict signal, inconsistent store), the verify exception propagates rather than caching a mapping to a thread that may not exist. Regression test `test_create_thread_non_conflict_failure_propagates_and_does_not_poison_store` pins the store-not-poisoned invariant for the exact 500-on-create shape you described.

backend/app/gateway/github/dispatcher.py
zhfeng: Fixed in 0a04e11 — using exactly the precedence you suggested:
```python
default_mention_login = b.github.bot_login or b.agent.name
fire, reason = event_should_fire(event, payload, triggers, default_mention_login)
```
The two gates (`_is_self_event` and `event_should_fire`) now agree on identity: `bot_login` first, `agent.name` fallback when `bot_login` is unset. Regression test `test_require_mention_uses_bot_login_when_trigger_omits_mention_login` pins that `@bot_login` fires the agent and `@<agent-dir-name>` is rejected — the exact misbehaviour reported. The fallback path (no `bot_login` set → `agent.name`) is also pinned by `test_require_mention_falls_back_to_agent_name_when_no_bot_login`.

backend/app/channels/manager.py
zhfeng: Fixed in e5c386f — using the branch shape you suggested:
```python
if isinstance(override, int) and override > 0:
    run_config["recursion_limit"] = override
else:
    run_config["recursion_limit"] = max(run_config.get("recursion_limit", 100), policy.default_recursion_limit)
```
Per-agent override now wins verbatim — `github.recursion_limit: 50` halts at 50 super-steps as configured, no clamping. The `max(...)` floor still applies when no override is present, so a user/channel-config base of 400 isn't lowered to the channel default of 250.

Also updated the `GitHubAgentConfig.recursion_limit` docstring to call out that any positive integer is honored verbatim, including below 100 — so operators don't expect a clamp that no longer exists. Parametrised regression test `test_github_channel_per_agent_recursion_limit_honors_value_below_100` pins values 1, 25, 50, 99, 100, 249, 250, 251, 1024 (50 being the exact value from your example) are all honored verbatim.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
zhfeng: Fixed in b14a4f1. Validation lives in the abstract layer (`deerflow/sandbox/sandbox.py::_validate_extra_env`) using exactly the `^[A-Za-z_][A-Za-z0-9_]*$` rule you specified, and is called from both implementations:

- `aio_sandbox.py` — at the injection site, before splicing keys into the `export` statement.
- `local_sandbox.py` — defense in depth before passing through to `subprocess.run(env=...)`. You're right that the subprocess path doesn't go through a shell, but enforcing the same contract on both implementations means a future implementation that does go through a shell doesn't have to re-derive the rule.

The abstract `Sandbox.execute_command` docstring also documents the contract now ("Keys must be valid POSIX environment-variable names; ValueError raised on violation").

15 parametrised regression cases in `test_extra_env_rejects_invalid_keys` cover the exact attack shapes (`;`, backtick, `$(...)`, `|`, `>`, `<`, space, tab, newline, leading digit, empty, non-str). End-to-end tests on both implementations confirm `subprocess.run` / `shell.exec_command` is NOT called when a key is invalid.

backend/app/gateway/github/app_auth.py
zhfeng: Deferred for this round — agreed it's a real concern at scale, but DeerFlow's current shape is single-tenant / low-installation-count and the operational cost of a periodic restart is below the cost of adding the eviction path now. Concrete reasoning:

- Per-entry footprint is ~200 bytes (`_CachedToken` + `asyncio.Lock`); even at 10k distinct installations the total is ~4 MB.
- The cache exists to avoid GitHub `/access_tokens` rate-limiting, not to act as a long-term registry — its purpose is bounded by the *active* installation set, not the historical one.
- Option 2 (opportunistic eviction on the cache-miss path) is the right fix when a multi-tenant deployment actually hits the wall; it's a ~10-line change with one regression test, and can land as a separate PR without touching the surface this PR is reviewing.

Will open a follow-up issue with your option 2 sketch attached so the work is queued explicitly rather than relying on someone re-discovering this comment thread.

backend/app/channels/manager.py
zhfeng: Fixed in dc7bdd4 — both shims deleted, tests migrated to `_apply_channel_policy` (the real production entry point).

- tests/test_github_token_plumbing.py — two redundant tests (token-installed, slack-noop) folded into the existing test_apply_channel_policy_installs_token_for_github and test_apply_channel_policy_is_noop_for_unregistered_channels. Three unique cases (JSON-serializability pin, mint-failure degrade, missing-installation_id) rewritten to call _apply_channel_policy. The two _mark_non_interactive tests are deleted — their assertions are now subsumed by the unified-policy tests (github path asserts disable_clarification is True, slack path's noop check covers the negative).
- tests/test_gateway_services.py — one docstring reference updated.
- app/channels/manager.py — both shims gone. The manager's public surface shrinks back to one channel-policy hook. Adding the next webhook channel (GitLab, Bitbucket, …) stays a one-row CHANNEL_RUN_POLICY entry, not a new manager method.

Net diff: 23 insertions, 114 deletions. No behaviour change in production.

zhfeng:
## F6 (token-cache eviction, `app/gateway/github/app_auth.py`) — deferred, with follow-up plan

Re: @willem-bd's review comment on `app/gateway/github/app_auth.py:65` — flagged here for visibility rather than buried in the inline thread.

**Decision:** not landing in this PR. Sketch attached so the work is queued explicitly and the next person doesn't have to re-discover the thread.

### Why defer

- Per-entry footprint is ~200 bytes (`_CachedToken` + `asyncio.Lock`). Even 10k distinct installations is ~4 MB — well below the noise floor for the gateway process.
- DeerFlow's current deployment shape is single-tenant / low-installation-count. The unbounded-growth failure mode requires a multi-tenant deployment with very long uptime and high installation churn, which is not where we are today.
- This PR is already large (F1–F5, F7 across six commits). Adding the eviction path expands the review surface without addressing a current problem.

### What to land later (option 2 from the review)

Opportunistic eviction on the cache-miss path in `_request_new_installation_token`:

```python
async def _request_new_installation_token(installation_id: int) -> _CachedToken:
    # Drop entries that are already past their leeway window — they would never
    # be returned by mint_installation_token() anyway. Bounded work per miss.
    now = time.time()
    expired = [
        iid for iid, tok in _token_cache.items()
        if tok.expires_at + _EXPIRY_LEEWAY <= now
    ]
    for iid in expired:
        _token_cache.pop(iid, None)
        _install_locks.pop(iid, None)
    ...
```

Plus a regression test (seed N expired entries + 1 live entry, mint a new id, assert the N expired entries are gone and the live one survives) and a docstring update so the eviction contract is documented next to the cache.

This bounds the cache to roughly the *active* installation set without pulling in an LRU dependency or a background sweep loop. ~10-line change, isolated to one file.

### Out of scope (not planned)

- LRU cap (option 1) — only matters if a deployment ever sees more *concurrently active* installations than its memory budget allows. Not a shape we see today.
- Background timer / periodic sweep — opportunistic eviction on cache-miss is sufficient.

Happy to take this as a separate PR once this one lands, or hand the sketch off if someone else wants to pick it up.

backend/app/gateway/github/identity.py
```diff
@@ -0,0 +1,74 @@
...
+            the GitHub side, so we don't need to distinguish here).
+
+    Returns:
+        Stringified UUID5 under :data:`GITHUB_THREAD_NAMESPACE`.
+    """
+    if not isinstance(repo, str) or "/" not in repo:
+        raise ValueError(f"Expected repo as 'owner/name', got {repo!r}")
+    if not isinstance(issue_or_pr_number, int):
+        raise ValueError(f"Expected issue_or_pr_number as int, got {type(issue_or_pr_number).__name__}")
+    return str(uuid.uuid5(GITHUB_THREAD_NAMESPACE, f"{repo}#{issue_or_pr_number}"))
```
willem-bd: **Deterministic thread id collides between agents bound to the same PR/issue.**

The UUID5 seed is `f"{repo}#{issue_or_pr_number}"`, so a repo with a coder + reviewer agent both bound to `owner/repo#7` produces the **same** thread id for both fan-out messages. `ChannelStore` keys by `(channel_name, chat_id, topic_id) = ("github", repo, str(number))`, which has the same collision. `ChannelManager._dispatch_loop` spawns one task per inbound message (`asyncio.create_task(self._handle_message(msg))`), so both runs race to invoke `client.runs.wait`/`stream` against the **same** LangGraph thread — either one busy-rejects the other or both interleave on a shared checkpoint and the agents see each other's conversation history.

This contradicts the headline feature in the PR description ("each on its own deterministic thread") and the new CLAUDE.md row. Include `b.agent.name` (or `b.user_id`) in the UUID5 seed, and key `ChannelStore` accordingly (e.g. extend `topic_id` to `f"{number}:{agent_name}"`).

backend/packages/harness/deerflow/tools/builtins/update_agent_tool.py
```diff
@@ -192,6 +192,21 @@ def _err(message: str) -> Command:
...
+    # added to AgentConfig is automatically carried through). Without this,
+    # operators who hand-author a ``github:`` block on a custom agent would
+    # silently lose it the next time the agent self-updates its description
+    # or tool_groups via update_agent.
+    #
+    # ``exclude_unset=True`` is recursive in Pydantic v2, so a sub-field the
+    # user didn't write (and that defaulted to a Pydantic default) doesn't
+    # get materialized into the YAML — the file round-trips visually intact.
+    _MANAGED_FIELDS = {"name", "description", "model", "tool_groups", "skills"}
+    preserved = existing_cfg.model_dump(exclude_unset=True, exclude=_MANAGED_FIELDS)
```
willem-bd: **This fix doesn't cover the HTTP `PATCH /api/agents/{name}` endpoint, which still drops `github:`.**

The harness `update_agent_tool` (this file) now preserves every non-managed `AgentConfig` field — good. But `app/gateway/routers/agents.py::update_agent` (around L317–340 on `main`) builds `updated = {"name", "description", "model", "tool_groups", "skills"}` and writes that with `yaml.dump(updated, …)`, with no equivalent `model_dump(exclude_unset=True, exclude=_MANAGED_FIELDS)` step. An operator who hand-authors a `github:` binding and later edits the agent description/model/skills from the Web UI silently loses the binding; the next webhook silently no-ops because `build_github_agent_registry()` no longer finds the agent.

Fix should be applied symmetrically in `routers/agents.py` (and ideally factored into a shared helper so the next added field doesn't need to be remembered in two places).

backend/app/gateway/github/dispatcher.py
```diff
@@ -0,0 +1,245 @@
...
+
+    sender_login = (payload.get("sender") or {}).get("login")
+
+    for b in bound:
+        # 3. Get this agent's triggers for the repo.
+        triggers: dict[str, GitHubTriggerConfig] = {}
+        for binding in b.github.bindings:
+            if binding.repo == repo:
+                triggers = binding.triggers
+                break
```
willem-bd: **Trigger lookup breaks on the *first* binding for a repo, but the registry indexes events across *all* bindings.**

`registry.py:96` indexes `(repo, event)` once per declared trigger across **every** binding, so an agent with two bindings on the same repo (one per event, a perfectly legal shape) ends up registered for events sourced from binding[1] while this loop returns binding[0]'s triggers.

Example:
```yaml
github:
  bindings:
    - repo: foo/bar
      triggers: {pull_request: {actions: [opened]}}
    - repo: foo/bar
      triggers: {issue_comment: {require_mention: true}}
```
An `issue_comment` webhook resolves the agent via the registry, but then `event_should_fire` receives the `pull_request`-only triggers from `binding[0]`, falls through `_resolved_trigger`, and returns `"disabled"` — the agent never fires for the event it explicitly subscribed to.

Either find the binding whose `triggers` actually contains `event` (`next((b for b in b.github.bindings if b.repo == repo and event in b.triggers), …)`), or flatten/merge triggers in `registry._rebuild` so the dispatcher has a single trigger map per `(repo, event, agent)`.

backend/app/gateway/routers/github_webhooks.py
```diff
@@ -0,0 +1,298 @@
...
+                    "error": "channel_service_not_available",
+                    "hint": "add channels.github.enabled: true to config.yaml",
+                }
+        except Exception:
+            logger.exception(
+                "github_webhook: fanout failed (delivery=%s event=%s)",
+                x_github_delivery,
+                x_github_event,
+            )
+            dispatch_result = {"error": "fanout failed"}
```
willem-bd: **Fanout failures are swallowed into a 200 OK, so GitHub never retries.**

The `except Exception` catches everything — including transient registry filesystem errors, bus publish errors, asyncio cancellation, etc. — logs, sets `dispatch_result = {"error": "fanout failed"}`, and falls through to the `return {"ok": True, …}` below. GitHub treats the delivery as successful and removes it from its retry queue, so a transient registry/bus error permanently drops a real webhook.

Returning 5xx here lets GitHub redeliver (it retries on 5xx but not on 200). The fail-closed startup-time `is_route_enabled()` check is the right move for *configuration* errors; this catch-all converts *runtime* errors into permanent silent drops. Suggest narrowing the catch and re-raising as `HTTPException(status_code=503)` (or letting the exception propagate) so GitHub's retry path is preserved for transient failures.

backend/app/gateway/github/registry.py
```diff
@@ -0,0 +1,172 @@
...
+    users_dir: Path = paths.base_dir / "users"
+    if not users_dir.exists():
+        return [DEFAULT_USER_ID]
+
+    found: list[str] = []
+    for entry in sorted(users_dir.iterdir()):
+        if entry.is_dir() and (entry / "agents").exists():
+            found.append(entry.name)
+    if DEFAULT_USER_ID not in found:
+        found.append(DEFAULT_USER_ID)
```
willem-bd, line 99: **Legacy `base_dir/agents/` shared layout is documented but never scanned, regressing unmigrated installs.**

The module docstring (L3-4) explicitly states the scan walks "every user's custom-agent directory under `{base_dir}/users/` **plus the legacy shared layout at `{base_dir}/agents/`**". `_discover_user_ids` only walks `users/`; `_gather_agent_signature` only calls `paths.user_agents_dir(user_id)`. `paths.agents_dir` is never touched.

Per the existing CLAUDE.md (`Memory System` row): "The legacy shared layout `{base_dir}/agents/{agent_name}/` remains read-only fallback for unmigrated installations". `load_agent_config(name)` still falls back to `paths.agent_dir(name)`, so an unmigrated install that adds a `github:` block to a shared agent loads fine on every other code path but produces **zero matches** here. The webhook silently fans out to nothing.

Add a pass over `paths.agents_dir.iterdir()` synthesizing `(DEFAULT_USER_ID, name)` tuples for any entry not already present in the per-user discovery — or update the docstring to remove the claim and document the migration requirement.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
```diff
@@ -122,12 +122,42 @@ def execute_command(self, command: str) -> str:
...
+                    # ``(...)`` runs it in a subshell so the export does
+                    # NOT persist on this AioSandbox's long-lived shell
+                    # session — critical for not leaking GH_TOKEN across
+                    # runs that share a sandbox via warm-pool reuse.
+                    #
+                    # Keys are NOT shlex.quote-d because they were validated
+                    # by ``_validate_extra_env`` above to be POSIX env-var
+                    # names (no shell metachars by construction).
+                    exports = "".join(f"export {k}={shlex.quote(v)}; " for k, v in extra_env.items())
+                    command = f"( {exports}{command} )"
```
willem-bd: **Wrapping the whole command in `( … )` discards `cd` and any other state the agent depends on persisting across bash calls.**

The AioSandbox's documented contract (and what the existing code base relies on) is a **persistent** shell session — `cd /workspace/repo` in one `bash` call is expected to affect the cwd of the next call. Wrapping in `(…)` makes that subshell exit immediately after the command, so the parent shell's cwd is unchanged. With `extra_env` set on every GitHub run, the agent doing `cd repo && git pull` followed by `git status` runs `git status` from the *original* directory.

Low-disruption fix: scope only the export, not the command:
```python
exports = "".join(f"export {k}={shlex.quote(v)}; " for k, v in extra_env.items())
unsets = " ".join(f"unset {k};" for k in extra_env)
command = f"{exports}{command}; __rc=$?; {unsets} (exit $__rc)"
```
or, if the AIO client supports per-exec env, pass it through that path instead of splicing strings.

The self-tests should grow a coverage case asserting `cd` survives a bash call when `extra_env` is set.

backend/app/gateway/routers/github_webhooks.py
```diff
@@ -0,0 +1,298 @@
...
+        handled = True
+        # Publish inbound messages onto the channel bus so the
+        # ChannelManager picks them up and routes them to the right
+        # custom agents. No direct agent-run calls here.
+        try:
+            from app.channels.service import get_channel_service
+
+            service = get_channel_service()
+            if service is not None:
+                dispatch_result = await fanout_event(
```
willem-bd, line 323: **Webhook fires agents regardless of `channels.github.enabled`.**

CLAUDE.md added in this PR documents `channels.github` as the place to set `enabled: true`, but this route unconditionally grabs `get_channel_service()` and calls `fanout_event(service.bus, …)`. If the service started without `channels.github.enabled=true`, the `GitHubChannel` outbound subscription is missing (which doesn't matter because the channel is log-only) — but the **bus consumer** in `ChannelManager._dispatch_loop` still picks up the inbound and runs the agent, which then posts to GitHub via `gh` regardless.

Net effect: an operator who flips `enabled: false` to disable the integration *without also unsetting `GITHUB_WEBHOOK_SECRET`* keeps running agents on every delivery, contradicting the documented kill-switch.

Gate the dispatch on `app_config.channels.github.enabled` (read once at request time or stashed on app state) before calling `fanout_event`; on disabled, return 200 with `dispatch_result={"skipped": "channel_disabled"}` so GitHub doesn't retry.

backend/app/gateway/github/dispatcher.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # 5. Trigger filter.
+        # default_mention_login mirrors the precedence used by _is_self_event
+        # three lines up: prefer the configured ``github.bot_login`` (the App's
+        # @-handle) and fall back to the agent's directory name. An operator
+        # who sets ``github.bot_login: deerflow-bot`` reasonably expects every
+        # ``@deerflow-bot`` mention to gate on that handle even when a binding
+        # leaves ``mention_login`` unset on a trigger — having the two gates
+        # disagree on identity would silently drop those mentions.
+        default_mention_login = b.github.bot_login or b.agent.name
```
willem-bd: **`channels.github.default_mention_login` from `config.yaml` is documented but never read.**

The new CLAUDE.md row claims the global `channels.github` block carries the "default `mention_login` for `require_mention` triggers". But `default_mention_login` here is computed strictly as `b.github.bot_login or b.agent.name` — `channels.github.default_mention_login` is never threaded into the dispatcher and never consulted.

An operator who sets `channels.github.default_mention_login: deerflow-bot` (per the docs) and has an agent named `coder` with `require_mention: true` and no per-trigger `mention_login`/`github.bot_login` override will silently require `@coder` mentions instead of `@deerflow-bot`, surprising both the operator and the user typing the mention.

Either thread the channel config into `fanout_event` (or read it through `get_app_config()` at the top of the function) and use it as the final fallback in the precedence chain, or drop `default_mention_login` from the documented `channels.github` block.

willem-bd (commented):
Reviewed at medium effort. The PR is well-tested and the architectural separation (`ChannelRunPolicy` table, harness/app boundary preserved, fail-closed route, narrow `ConflictError` recovery) is clean. Eight surviving findings below, most-severe first — they are all things a maintainer would likely want to act on before this lands.

1. **Deterministic thread id collides between coder + reviewer on the same PR** — `resolve_thread_id` keys only on `(repo, number)`, breaking the headline "each on its own deterministic thread" claim.
2. **HTTP `update_agent` endpoint silently strips the `github:` block** — the harness `update_agent_tool` fix from this PR isn't mirrored in `app/gateway/routers/agents.py`.
3. **First-binding trigger lookup** misses events when an agent has multiple bindings for the same repo.
4. **Fanout failures return 200**, so GitHub never retries a transient bus/registry error.
5. **Registry skips the legacy `base_dir/agents/` shared layout** even though the module docstring says it walks both.
6. **`AioSandbox` wrap in `( … )` drops `cd` persistence** for any command carrying `extra_env` — every GitHub run.
7. **Webhook fires agents regardless of `channels.github.enabled`** — the kill-switch documented in CLAUDE.md is bypassed by the route.
8. **`channels.github.default_mention_login` from `config.yaml` is never read** — the dispatcher falls back to `b.github.bot_login or b.agent.name` only.

Details inline.

backend/app/gateway/github/identity.py
zhfeng: Fixed in ee8ad787 by moving to **per-agent threads** (`(repo, PR/issue number, agent_name)`).

The shared-thread design relied on a per-agent self-event gate plus distinct `mention_login` to keep coder and reviewer from cross-firing, but it still had two failure modes that aren't really avoidable on a single thread:

1. **Dual-mention silent drop.** Both `ChannelManager.runs.wait` and `runs.stream` submit with `multitask_strategy="reject"`. A single comment that mentions `@coder @reviewer` (or a `@coder` mention while the reviewer is still running on the same PR) produces two near-simultaneous publishes to the same thread, and the loser is rejected by langgraph. The channel is log-only on outbound, so the user just sees silence — nothing surfaces back to the PR. Worth being explicit: the existing tests assert per-agent fan-out and the self-event gate, but neither covers the dual-mention race, because the bus fan-out happens before the manager's reject window.

2. **Failure isolation.** A bug or sandbox poison in one agent's run would corrupt the shared checkpoint for the other on the same PR.

Design now:
- `resolve_thread_id(repo, number, agent_name)` seeds UUID5 with `f"{repo}#{number}:{agent_name}"` — coder and reviewer on `owner/repo#7` land on different threads (`test_thread_id_differs_per_agent`).
- ChannelStore row uses `topic_id = f"{number}:{agent_name}"` so each agent's mapping is independent and `_lookup_thread_id` never collides them.
- Both agents proceed in parallel; coordination happens through GitHub (PR comments, review threads), which is the source of truth humans see anyway.
- Test added: `test_coder_and_reviewer_on_same_pr_get_distinct_threads` — verifies two bindings on the same `(repo, event)` produce two `InboundMessage`s with distinct `preferred_thread_id`s and distinct `topic_id`s.
- CLAUDE.md rows rewritten with the mechanism and the `multitask_strategy="reject"` rationale.

Thanks for catching this — the shared-thread version did promise more atomicity than it could actually deliver.

backend/packages/harness/deerflow/tools/builtins/update_agent_tool.py
zhfeng: Fixed in 830555e6.

You were right that the fix needed to live in both surfaces — and you correctly called out factoring it into a shared helper. Done:

- `deerflow/config/agents_config.py` now exports `MANAGED_AGENT_CONFIG_FIELDS` and `preserve_non_managed_fields(existing_cfg) -> dict[str, object]`.
- Both surfaces use it. `update_agent_tool` was rewritten to call the helper; `routers/agents.py::update_agent` now `setdefault`s the result over its managed-fields dict before `yaml.dump`. Any future top-level field on `AgentConfig` is carried through both paths automatically; no two-place edit is needed.
- Test added: `test_update_agent_preserves_hand_authored_github_block` in `tests/test_custom_agent.py` — creates an agent through the API, hand-authors a `github:` block into the on-disk `config.yaml`, PATCHes only `description`, and asserts the `github:` block survives verbatim.

While reviewing this with the maintainer, we noticed a second issue on the same `update_agent` tool that R2 happens to live on. Surfacing it here for visibility:

**The lead-agent prompt instructs the model to call `update_agent` when "the user" asks to change its own skills/tool_groups/SOUL.md.** On the GitHub channel, "the user" is whatever external commenter posted the triggering `@<bot>` mention. Anyone who can comment on a configured repo clears the trigger gate (`require_mention` only checks the @-mention text, not the author's `author_association`). Exposing `update_agent` on webhook runs gave that commenter a path to durably mutate the bot's `tool_groups` / SOUL.md / `model` for every subsequent run.

Same commit fixes that:
- `channels/manager.py` plumbs `msg.channel_name` onto `run_context`.
- `lead_agent/agent.py` drops `update_agent` from `extra_tools` when `run_context["channel_name"]` is in a new `_WEBHOOK_CHANNELS = frozenset({"github"})`.
- `update_agent_tool.py` mirrors the same set as defence in depth: it returns an error and does not touch the filesystem when invoked on a webhook channel, in case some future factory re-attaches the tool without going through `_make_lead_agent`.
- Tests in `test_lead_agent_skills.py`, `test_update_agent_tool.py`, and `test_channels.py` lock all three gates down.

Thanks again — R2 was the seam that exposed the bigger issue.

backend/app/gateway/github/dispatcher.py
zhfeng: Fixed in 05b6a40a. Took the stricter path you implied — fail loudly at config load instead of letting the dispatcher try to merge ambiguous triggers from multiple same-repo bindings:

**1. `GitHubAgentConfig` rejects duplicate `repo` values across bindings.** New `model_validator` on `GitHubAgentConfig` raises `ValueError: Agent github.bindings has duplicate repos [...]. Each repo must appear at most once — merge their \`triggers:\` maps into a single binding.` Lists every offender so operators can locate them in one pass.

**2. Registry pre-resolves each `(repo, event)` to exactly one trigger per agent.** `_Registry` is now `dict[(repo, event), list[GitHubAgentMatch]]` where each match embeds the trigger already merged with `DEFAULT_TRIGGERS` field defaults. Dispatcher's binding-loop is gone entirely — no more "first binding wins" failure mode, and the secondary double-registration risk (one repo declared twice in different bindings) is foreclosed at the validator.

**3. `event_should_fire` takes the pre-resolved trigger directly** instead of a `{event: trigger}` dict + internal lookup. No behavioral change beyond moving the lookup to build time.

Also collapsed the `GitHubBoundAgent` wrapper into the flat `GitHubAgentMatch` (`user_id`, `agent: AgentConfig`, `trigger: GitHubTriggerConfig`) — `AgentConfig` already carries `.github`, so the wrapper just produced `.agent.agent.name` accessors with no information added.

**Tests:** 4 new validator regression tests in `test_github_agents_config.py`; `test_github_triggers.py` rewritten to call `event_should_fire` with a pre-resolved trigger via a small `_resolve()` helper; `test_github_registry.py` updated for the flat match shape. 204 GitHub-related tests pass.

backend/app/gateway/routers/github_webhooks.py
zhfeng: Fixed in 8d4ef66f. Took the broad version of your suggestion: any exception inside `fanout_event` propagates as `HTTPException(status_code=503)` with the delivery id in the detail, so GitHub's retry path picks it up. The `except Exception → 200` swallow is gone.

**Scope of the 503:**

| Case | Status | Why |
|---|---|---|
| `fanout_event` raised (registry scan, bus publish, prompt build) | **503** | Transient resolves on redelivery; this is the win. |
| `channels.github` not enabled (channel service is `None`) | 200 | Permanent config state — 5× retry buys nothing. |
| Unknown event (not in `_KNOWN_EVENTS`) | 200, `handled=false` | Same reasoning. |
| Malformed payload / missing event header | 400 (unchanged) | Client-side bug, not retryable. |
| Signature verification failure | 401 (unchanged) | Authenticity, not retryable. |

Startup-time `is_route_enabled` still handles fail-closed *configuration* errors (route absent → 404). 503 is reserved for runtime failures GitHub can retry past.

**One thing worth naming explicitly**: agent-runtime errors are *not* redelivered. Once `bus.publish_inbound(msg)` returns, we 200 GitHub. Failures inside the actual agent run (LLM call 500s, `gh` command fails, sandbox crashes) are owned by `ChannelManager` and surface via run state / logs, not by webhook retry. Same contract Feishu/Slack/Telegram have. So the 503 catches the "we never even got the message onto the bus" class of failure, which is the class GitHub redelivery can usefully fix.

**Failure-mode tradeoff**: for a *permanent* bug in fanout (e.g. `build_prompt` crashing on a particular comment-body shape) GitHub will hit us ~5 times over ~8 hours before giving up. End state matches the old swallow-and-200 path (delivery effectively dropped), but with 5× the log volume and red rows in the "Recent Deliveries" panel. I think that's net good — silent drops were worse than loud ones — but it's a real cost. If it becomes a problem in practice we can narrow the catch to transient-flavored exception types (`OSError`, `asyncio.CancelledError`, `asyncio.QueueFull`) and let programmer-error exceptions stay as 500. Holding off on that until we see evidence it matters.

**Tests:**
- Flipped `test_dispatch_failure_does_not_500` (which had codified the buggy 200 behaviour) to `test_dispatch_failure_returns_503_so_github_retries`.
- Added `test_dispatch_failure_503_lets_github_redeliver_successfully` — same delivery sent twice, first 503 (raises), second 200 (resolves) — pins down that the 503 is not sticky and a real redelivery does land on the success path. Guards against any future "cache failed delivery" regression.

40 webhook tests pass; full 205-test GitHub sweep green.

backend/app/gateway/github/registry.py
zhfeng: Fixed in e7323ae.

You're right on all three points — the docstring claims both layouts, `load_agent_config(name)` does fall back to `paths.agent_dir(name)` via `resolve_agent_dir`, and CLAUDE.md L457 keeps the legacy layout as a read-only fallback for unmigrated installs. The webhook path was the only one quietly refusing the legacy shape, so an unmigrated install with a `github:` block silently fanned out to nothing.

Went with **option (a)** — restore the legacy scan rather than require migration. The rest of the codebase (`load_agent_config`, `list_custom_agents`, memory, ACP) all honour the legacy layout, so making GitHub the lone exception would just be another silent failure mode in a code path nobody can observe from the outside.

Implementation: extended `_gather_agent_signature` with a second pass over `paths.agents_dir`, bucketing legacy entries under `DEFAULT_USER_ID` (the same user_id `load_agent_config` resolves them under at run-time). Per-user entries shadow legacy entries of the same name via a `seen` set, matching `list_custom_agents`' precedence — so once `migrate_user_isolation.py` runs, the legacy row silently disappears rather than producing duplicate trigger sets. The mtime-keyed cache picks up legacy edits via the same signature mechanism.

Regression tests added in `tests/test_github_registry.py`:
- `test_registry_indexes_legacy_shared_agent` — legacy agent indexes under DEFAULT_USER_ID.
- `test_registry_per_user_entry_shadows_legacy_with_same_name` — per-user shadow precedence verified by routing each layout to a different repo and asserting only the per-user one resolves.
- `test_registry_legacy_cache_invalidates_on_legacy_config_edit` — mtime-bump on a legacy `config.yaml` triggers a rebuild.

111/111 GitHub tests pass.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
zhfeng: Fixed in 6411f72.

You're right — `( ... )` was too broad. The subshell exits the moment the user command finishes, so `cd` (and any other persistent shell-state side effect the agent relied on) was silently discarded. With `extra_env` set on every GitHub run, an agent doing `cd repo && git pull` followed by `git status` ran `git status` from the original cwd.

Went with the unset-based scope you suggested:

```sh
export K=V; <cmd>; __rc=$?; unset K; (exit $__rc)
```

- Exports run top-level so `cd` and other shell state persist across exec_command calls — matches the AIO contract.
- `$__rc` captures the command's exit code before the cleanup statements clobber `$?`.
- The `unset` only fails to run if `<cmd>` killed the shell itself (`exit`, `kill $$`). In that case the session is dead anyway and the warm pool reissues a fresh one, so there's nothing to leak into. Trade-off favors the fix.
- `(exit $__rc)` propagates the original exit code to `$?` so callers inspecting it see the real result.

Skipped the per-exec env path on the AIO client since the existing `shell.exec_command(...)` signature has no env kwarg, and splicing into the command keeps the change minimal.

Tests:
- Renamed `test_aio_sandbox_extra_env_wraps_in_subshell` → `test_aio_sandbox_extra_env_does_not_wrap_in_subshell`. Old assertions pinned the `(...)` shape; new ones pin the unset-based shape (top-level export, `__rc=$?` capture, `unset K;`, trailing `(exit $__rc)`, no leading `(`).
- Added `test_aio_sandbox_extra_env_preserves_cd_across_calls` — the regression you asked for. Asserts the `cd` is at top level (not inside a subshell wrapper) and positioned before the cleanup so the cwd change reaches the live session.

97/97 sandbox + token-plumbing tests pass.

backend/app/gateway/routers/github_webhooks.py
zhfeng: Fixed in ed7351d.

Right — the `channels.github.enabled` flag claimed kill-switch authority but had none. The route used to ask only "is the channel service running at all?", and once that passed, `fanout_event` published inbound onto the bus regardless. The `ChannelManager._dispatch_loop` consumer is permanently subscribed (it's not per-channel), so the agent ran and posted via `gh` from the sandbox even with `enabled: false`.

Added `ChannelService.is_channel_enabled(name)` reading the live `_config` dict (the same one `configure_channel` updates for UI-driven flips, so the helper tracks runtime changes without a config.yaml reload). The webhook route gates on it between the existing `service is None` branch and `fanout_event` — on disabled, log INFO and return 200 with `dispatch={"skipped": "channel_disabled"}`. Permanent state, so GitHub doesn't retry. The route itself stays mounted because `GITHUB_WEBHOOK_SECRET` is the route-existence gate and we want the secret and kill-switch to be independent (you can run a CI deploy with the secret set but the integration off, or vice-versa).

Tests:
- `test_channel_disabled_skips_fanout` — disabled service → 200, `dispatch={"skipped": "channel_disabled"}`, `fanout_event.await_count == 0` (belt-and-braces patched-fake check that fan-out is genuinely never invoked, in case a future regression bypasses the gate).
- `test_channel_enabled_dispatches_normally` — positive counterpart pinning the enabled branch.
- `test_is_channel_enabled_reflects_live_config` (in `tests/test_channels.py::TestChannelService`) — pins the helper across enabled, disabled, missing entry, non-dict garbage, and `configure_channel` runtime flip.

The existing autouse `_stub_channel_service` fixture in `test_github_webhooks.py` grew an `is_channel_enabled` method defaulting to True so the 40 pre-existing tests stay green.

CLAUDE.md's "non-retryable conditions" parenthetical updated to spell out the kill-switch path explicitly (was just "channel service is `None`"; now also covers `enabled: false`).

161/161 GitHub + token-plumbing tests pass; 26/26 ChannelService unit tests pass.

backend/app/gateway/github/dispatcher.py
zhfeng: Fixed in c4c1715d.

You're right — the operator default was advertised in CLAUDE.md but never read. The dispatcher computed `default_mention_login = github.bot_login or agent.name` and skipped `channels.github` entirely, so an operator who set `default_mention_login: deerflow-bot` still saw `require_mention` gate on `@coder` (the agent directory name). Silent contract drift between docs and code.

Threaded the operator default through `fanout_event` rather than have the dispatcher reach into `get_app_config()` itself — keeps the dispatcher pure and testable without the singleton dep. New precedence chain (most specific → fallback):

    trigger.mention_login
      → github.bot_login
      → channels.github.default_mention_login    (live, via get_channel_config)
      → agent.name                                (last-resort fallback)

The router pulls the value from `ChannelService.get_channel_config("github")` (a new shallow-copy accessor on the runtime-authoritative `_config` dict, so UI-driven flips via `configure_channel` take effect without a restart) and passes it through. Whitespace-only operator defaults are stripped and treated as unset, so a stray `default_mention_login: ""` falls through to `agent.name` rather than silently gating on the empty string.

Tests pin all 4 precedence rungs at dispatcher level (`test_operator_default_mention_login_used_when_agent_omits_bot_login`, `test_agent_bot_login_outranks_operator_default`, `test_trigger_mention_login_outranks_operator_default`, `test_operator_default_blank_string_treated_as_none`) plus the router-level threading (`test_operator_default_mention_login_is_threaded_to_fanout` / `..._absent_passes_none`). CLAUDE.md extended with the full 4-step chain so future readers don't have to reverse-engineer it from the dispatcher. 328 passed.

zhfeng:
@willem-bd I squash all the commits because of the AGENT.md and CLAUDE.md updated in the upstream. So please take a review again, thanks!

zhfeng:
@WillemJiang please take a look, thanks!

WillemJiang (approved):

PR #3759, opened 2026-06-24
fix(scripts): make stop reaps repo-owned nginx on macOS

### LINKED ISSUE #3758: [bug] make stop fails to reap repo-owned nginx on macOS — argv[0] rewrite breaks _is_repo_nginx_pid in scripts/serve.sh
## Summary

On macOS, `_is_repo_nginx_pid()` in `scripts/serve.sh` rejects every nginx
process — including the ones the repo itself just started. As a consequence
`_kill_repo_nginx()` is a no-op on macOS, `make stop` leaves nginx running on
port 2026, and the next `make start` aborts:

```
✗ Nginx cannot start because port 2026 is already in use.
  If it belongs to this worktree, run 'make stop'; otherwise free the port manually.
```

The advice is misleading: `make stop` is precisely the command that should
have killed it.

Reproduced against upstream `main` @ commit `debb0fd` (current HEAD at the
time of this report), unmodified.

## Affected code

`scripts/serve.sh:200-209`:

```bash
_is_repo_nginx_pid() {
    local pid=$1
    local command
    local args

    command=$(ps -p "$pid" -o comm= 2>/dev/null) || return 1
    case "$command" in
        nginx|*/nginx) ;;
        *) return 1 ;;
    esac
    ...
}
```

This is consumed by `_kill_repo_nginx()` (lines 223-246), which is called by
`stop_all()` and therefore by `make stop`, `make restart`, and the implicit
stop run at the start of every `make start`.

## Why it's broken on macOS

Standard nginx behavior: master and worker processes rewrite their own
`argv[0]` after fork, becoming `nginx: master process …` and
`nginx: worker process` respectively.

On macOS, `ps -p PID -o comm=` returns that rewritten string rather than the
executable basename:

```
$ ps -p <master> -o comm=
nginx: master process /opt/homebrew/bin/nginx -c /…/nginx.local.conf -p /…
$ ps -p <worker> -o comm=
nginx: worker process
```

Neither matches `nginx` nor `*/nginx`, so the case statement falls through to
`*) return 1 ;;` for every nginx the repo started.

Linux is unaffected, because the kernel's `comm` field is the executable
basename (capped at 16 bytes) and is not influenced by `argv[0]` rewrites.

## Reproduction (no deer-flow installation needed — pure upstream repro)

```bash
# 1. Get a clean upstream copy
git clone --depth=1 https://github.com/bytedance/deer-flow.git /tmp/deer-flow-upstream
cd /tmp/deer-flow-upstream && git rev-parse HEAD     # debb0fd at the time of writing

# 2. Spin up an unrelated, standalone nginx so this test depends on nothing else
mkdir -p /tmp/repro/logs
cat > /tmp/repro/nginx.conf <<'CONF'
worker_processes 1;
daemon off;
error_log /tmp/repro/logs/error.log;
pid /tmp/repro/logs/nginx.pid;
events { worker_connections 64; }
http { server { listen 19999; location / { return 200 "ok"; } } }
CONF
nginx -c /tmp/repro/nginx.conf -p /tmp/repro &
sleep 1
pgrep -f 'nginx.*repro/nginx.conf'   # → master pid; worker is its child

# 3. Source upstream's helpers verbatim and ask them to kill that nginx
cat > /tmp/repro/probe.sh <<'BASH'
#!/usr/bin/env bash
SERVE=/tmp/deer-flow-upstream/scripts/serve.sh
DEERFLOW_ROOTS="/tmp/repro"
REPO_ROOT="/tmp/repro"
_is_deerflow_pid() { return 1; }
eval "$(awk 'NR>=200 && NR<=246' "$SERVE")"   # _is_repo_nginx_pid + _kill_repo_nginx
kill() { echo "[would kill -9] $*"; }         # observe instead of really killing
echo "=== upstream _kill_repo_nginx invoked ==="
_kill_repo_nginx
echo "=== returned ==="
BASH
bash /tmp/repro/probe.sh
```

### Actual output (upstream `main` `debb0fd`, macOS):

```
=== upstream _kill_repo_nginx invoked ===
=== returned ===
```

`_kill_repo_nginx` walked `pgrep -f nginx`, asked `_is_repo_nginx_pid` about
each PID, was told "no" for every one, and emitted **zero** kill calls. The
nginx is still running afterwards:

```
$ ps -ax -o pid,command | grep '[r]epro/nginx.conf'
72488 nginx: master process /opt/homebrew/bin/nginx -c /tmp/repro/nginx.conf -p /tmp/repro
```

### After the proposed one-line fix

Patch the case to also accept `nginx:*` (the macOS form), then re-run the
exact same probe script:

```
=== patched _kill_repo_nginx invoked ===
[would kill -9] -9 72488
=== returned ===
```

The helper now correctly identifies the nginx master and signals it.

## Real-world steps to reproduce in a normal dev workflow

```bash
make start-daemon          # services come up
make stop                  # claims success
lsof -nP -iTCP:2026 -sTCP:LISTEN
# → repo's nginx master+worker are still listening on 2026
make start                 # ✗ Nginx cannot start because port 2026 is already in use
```

The Ctrl+C path on a foreground `make start` masks the bug because SIGINT
goes to the whole foreground process group, killing nginx without going
through `_kill_repo_nginx`. The bug surfaces whenever nginx has to be killed
through the helper: daemon mode, terminal closed, SIGTERM, killed by
timeout/CI, etc.

## Proposed fix

Accept the macOS form (`nginx: master process …` / `nginx: worker process`)
in addition to the Linux form. Safety is preserved by the existing downstream
checks (the `args` case and `_is_deerflow_pid`), which still require the
process to be tied to this worktree — so a Homebrew system-managed nginx or
any other unrelated nginx on the machine will *not* be killed.

```diff
--- a/scripts/serve.sh
+++ b/scripts/serve.sh
@@ -200,9 +200,13 @@ _is_repo_nginx_pid() {
     local pid=$1
     local command
     local args

     command=$(ps -p "$pid" -o comm= 2>/dev/null) || return 1
+    # nginx rewrites argv[0] for its master/worker processes
+    # ("nginx: master process …" / "nginx: worker process"). On macOS,
+    # `ps -o comm=` returns that rewritten string rather than the executable
+    # basename, so the previous patterns missed every repo-owned nginx and
+    # made `make stop` a no-op for nginx on macOS. Accept the rewritten form;
+    # the args/_is_deerflow_pid checks below still gate ownership.
     case "$command" in
-        nginx|*/nginx) ;;
+        nginx|*/nginx|nginx:*) ;;
         *) return 1 ;;
     esac
```

Happy to send a PR if a maintainer confirms the direction.

## Environment

- macOS (any recent version on Apple Silicon or Intel)
- Homebrew nginx 1.31.x (`/opt/homebrew/bin/nginx`)
- bash (system 3.2 reproduces; Homebrew bash 5.x reproduces)
- deer-flow `main` @ `debb0fd`

xiawiie:
## Summary

- `scripts/serve.sh`: accept the macOS-rewritten `argv[0]` form (`nginx:*`) in `_is_repo_nginx_pid` so `make stop` actually kills repo-owned nginx on macOS.

Fixes #3758.

## Why

On macOS, `_is_repo_nginx_pid()` rejected every nginx process — including the ones the repo itself just started. As a consequence `_kill_repo_nginx()` was a no-op on macOS, `make stop` left nginx running on port 2026, and the next `make start` aborted with:

```
✗ Nginx cannot start because port 2026 is already in use.
```

nginx master/worker processes rewrite their own `argv[0]` after fork (`nginx: master process …` / `nginx: worker process`). On macOS, `ps -p PID -o comm=` returns that rewritten string rather than the executable basename, so the previous `case nginx|*/nginx)` never matched. Linux is unaffected because the kernel's `comm` field is the executable basename and is not influenced by `argv[0]` rewrites.

## Change

```diff
 _is_repo_nginx_pid() {
     ...
     command=$(ps -p "$pid" -o comm= 2>/dev/null) || return 1
+    # On macOS nginx rewrites argv[0] (e.g. "nginx: master process ..." or
+    # "nginx: worker process"), and `ps -o comm=` returns that rewritten string
+    # rather than the executable basename. Accept the macOS form as well so
+    # `make stop` actually reaps repo-owned nginx processes here.
     case "$command" in
-        nginx|*/nginx) ;;
+        nginx|*/nginx|nginx:*) ;;
         *) return 1 ;;
     esac
```

The downstream `args` case and `_is_deerflow_pid` fallback still gate ownership, so a Homebrew system nginx or any other unrelated nginx on the machine is not at risk.

## Test plan

End-to-end repro + verification (full repro details, including a self-contained probe script that sources the helper verbatim, are in #3758).

- [x] **Before patch (upstream `main` `debb0fd`, macOS):** spawn an unrelated nginx via the conf at `/tmp/repro/nginx.conf`, then call upstream's `_kill_repo_nginx` with `kill` overridden to print instead of execute. The function emitted **zero** kill calls and the nginx remained alive. `_is_repo_nginx_pid` returned non-zero for both the master (`nginx: master process …`) and worker (`nginx: worker process`).
- [x] **After patch (this branch):** the same probe against the same nginx now prints `[would kill -9] -9 <master_pid>` and `_is_repo_nginx_pid` matches both master and worker.
- [x] Real workflow (`make start-daemon` → `make stop` → `make start`) on macOS no longer leaves nginx running on port 2026 between cycles.
- [x] No-op on Linux (the new `nginx:*` arm is additive; Linux `ps -o comm=` continues to return `nginx`).

## Risk

- Same kill-target ownership invariants as before (the `args`/`_is_deerflow_pid` checks immediately following the case statement are unchanged).
- The Ctrl+C path on a foreground `make start` was already working because SIGINT goes to the whole process group; this change only affects code paths that go through the helper (daemon mode, terminal closed, SIGTERM, CI without TTY signal forwarding).

### DIFF
--- a/scripts/serve.sh
+++ b/scripts/serve.sh
@@ -203,8 +203,12 @@ _is_repo_nginx_pid() {
     local args
 
     command=$(ps -p "$pid" -o comm= 2>/dev/null) || return 1
+    # On macOS nginx rewrites argv[0] (e.g. "nginx: master process ..." or
+    # "nginx: worker process"), and `ps -o comm=` returns that rewritten string
+    # rather than the executable basename. Accept the macOS form as well so
+    # `make stop` actually reaps repo-owned nginx processes here.
     case "$command" in
-        nginx|*/nginx) ;;
+        nginx|*/nginx|nginx:*) ;;
         *) return 1 ;;
     esac
 


WillemJiang:
@xiawiie, thanks for your contribution. Please sign the CLA by clicking the CLA assistant button before we start to review the code. 
To address the complaint of the CLA, you need to add the git commit author email into your GitHub profile email settings.

WillemJiang:
The issue was addressed in #3828

xiawiie:
Why I found this first, but you merge his first? I didn't truly understand it

Issue #3540, opened 2026-06-12
[RFC] Hermes-like DeerFlow TUI backed by DeerFlowClient


hetaoBackend:
## Before you start

- [x] I searched existing issues and this is not a duplicate.

Related issues:

- #3292 discusses a broader CLI engine with per-session SQLite isolation.
- #1665 asks about skills CLI access.

This RFC has been reframed as a **TUI-first** proposal. The goal is not merely a command set, but a Hermes-like terminal workbench backed by the existing embedded `DeerFlowClient`.

References:

- Hermes TUI docs: https://hermes-agent.nousresearch.com/docs/user-guide/tui
- Hermes CLI/TUI overview: https://hermes-agent.nousresearch.com/docs/user-guide/cli

## Problem / motivation

DeerFlow has a capable embedded Python client and a full Gateway/Web UI, but no first-class terminal workbench. A plain command set would help scripts, but it would not cover the interactive developer experience we want: a persistent, keyboard-driven TUI where users can chat, switch threads, inspect tools, manage files, view live execution, trigger slash commands, and keep context without opening the browser.

Hermes Agent's TUI is a useful reference point: the terminal UI is backed by the same runtime as the classic CLI, shares sessions and slash commands, supports modal overlays, non-blocking input, live session switching, status indicators, and a TTY-aware fallback path.

DeerFlow needs the same product shape:

1. A terminal-native primary interface for users who live in the shell.
2. The same agent runtime, memory, skills, MCP tools, sandbox behavior, and thread persistence as the web app and embedded client.
3. A richer interactive surface than one-shot commands can provide.
4. Scriptable headless commands for automation, but not as the center of the product.

## Proposed solution

Build a first-party `deerflow` TUI backed by `DeerFlowClient`.

Default interactive path:

```bash
deerflow
deerflow --tui
deerflow chat
deerflow --continue
deerflow --resume THREAD_ID
```

Headless/script path:

```bash
deerflow chat --print "summarize this repo"
deerflow models list --json
deerflow threads list --json
```

The v1 TUI should run in embedded mode only. It should not require Gateway, frontend, nginx, or Docker services to be running, though it must honor the same `config.yaml`, `extensions_config.json`, runtime paths, sandbox configuration, tracing configuration, skills, memory, and checkpointer settings used by the rest of DeerFlow.

Remote HTTP/Gateway mode can be added later as a transport option. The first implementation should ship a dependable local terminal workbench.

## Reference behavior from Hermes

[...]

PR #3760, opened 2026-06-24
feat(tui): Hermes-like terminal workbench (`deerflow`) backed by DeerFlowClient

hetaoBackend:
Implements the **v1** of #3540 (the TUI-first RFC). Referencing rather than auto-closing, since the RFC scopes follow-ups beyond v1 (see *Out of scope* below).

## Why

#3540 proposed a Hermes-like terminal workbench: a keyboard-driven TUI for people who live in the shell, backed by the **same embedded harness** as the web app — no Gateway / frontend / nginx / Docker required. There was no first-class terminal surface for DeerFlow; this is the v1, and it also delivers the RFC's key architectural decision (terminal sessions show up in the Web UI **without** running the Gateway).

The design follows the RFC and learns UX/architecture patterns (pure view-state reducer, command registry, compact tool cards) from `hetaoBackend/tao-pi`, but is implemented Python-native with Textual so it drives `DeerFlowClient` in-process.

## What changed

- **New `deerflow` console script** (added to the harness `pyproject.toml`). `textual` is an optional `[tui]` extra — the script degrades to headless help when it's absent, so the core harness install stays lean.
- **`deerflow` launches a Textual TUI**: header (model / thread / cwd / skills), live streaming transcript, compact tool cards (`⚙ Read path ✓` with dimmed result previews), status line (run state + spinner, model, thread title, token usage, `esc interrupt`), a `/`-triggered slash-command palette, `/model` + `/threads` modal pickers, `↑/↓` input history, `Ctrl+C` / `Esc` interrupt, and TTY-aware fallback.
- **Slash surface**: `/help /new /threads /switch /model /skills /tools /mcp /memory /uploads /usage /config /quit`, plus `/<skill-name> task` to activate any enabled skill (existing slash-skill semantics preserved).
- **Headless modes**: `deerflow --print "…"`, `deerflow --json "…"` (NDJSON `StreamEvent`s), stdin, `--continue`, `--resume THREAD`, `--tui` / `--cli`, `DEER_FLOW_TUI=1`.
- **Web UI visibility** (the RFC's core decision): on the first turn of a thread the TUI writes a `threads_meta` row — owned by the local default user (`"default"`) — into the **same** DB the Gateway reads (via the harness-only `deerflow.persistence.engine.init_engine_from_config`), and syncs the generated title. So terminal sessions appear in the Web UI sidebar **without the Gateway process running**. Best-effort: a no-op on the `memory` backend. Proven end-to-end against a real SQLite file.

It's a UI shell over the existing embedded harness — agent behaviour, persistence, streaming semantics and tools are **not** forked.

## Surface area

- [x] **Agents / LangGraph** — consumes the agent runtime via `DeerFlowClient` (no graph/prompt change)
- [x] **Dependencies** — `textual>=0.80` as an optional `deerflow-harness[tui]` extra (also in the backend dev group) — buys the terminal UI; not pulled into the core install
- [x] **Docs / tests** — `backend/docs/TUI.md`, CLAUDE.md/README updates, 107 new tests
- New harness subpackage `packages/harness/deerflow/tui/` + the `deerflow` console entry point (no existing behaviour changes; nothing is opt-out)

## Architecture

Every layer except the Textual app is pure (no Textual import) and unit-tested directly:

```
cli.py          launch-mode planning (pure) + headless print/json + `deerflow` entry
view_state.py   ViewState + reduce(state, action)  — the testable heart
runtime.py      StreamEvent -> reducer actions (pure translate + threaded driver)
message_format / command_registry / input_history / render / theme   — pure
app.py          Textual App: runs the sync DeerFlowClient.stream() on a worker
                thread, marshals actions back to the UI thread, renders ViewState
session.py / persistence.py   client + checkpointer + the threads_meta writer
widgets/composer.py           CJK-aware composer input
```

`DeerFlowClient.stream()` is a synchronous generator, so runs execute on a Textual worker thread and each action is marshalled back via `call_from_thread`.

## Screenshots

Conversation surface and the slash-command palette (committed SVGs under `docs/tui/`):

- [`docs/tui/tui-preview.svg`](https://github.com/bytedance/deer-flow/raw/feat/deerflow-tui/docs/tui/tui-preview.svg)
- [`docs/tui/tui-palette.svg`](https://github.com/bytedance/deer-flow/raw/feat/deerflow-tui/docs/tui/tui-palette.svg)

## Testing

`cd backend && PYTHONPATH=. uv run pytest tests/ -k tui` → **107 passed**. Pure layers via plain pytest; the app, slash palette and modal overlays via Textual's pilot harness with a fake in-process session; `test_tui_persistence.py` proves the `threads_meta` read/write round-trip. `ruff` clean; respects the harness→app import boundary (`test_harness_boundary.py`).

Several fixes landed during hands-on self-testing and are covered by regression tests: streamed assistant-text de-dup, tool-card de-dup / empty-card suppression, no duplication of prior answers on long threads, `Esc` interrupt, `Tab` keeping composer focus, and a CJK cursor-drift fix (a Textual `Input` `_cursor_offset` off-by-one after wide chars, fixed via a small subclass) — verified in iTerm2.

## Out of scope (RFC follow-ups)

- Hydrating a resumed thread's prior messages into the transcript (resume continues server-side but starts the UI fresh).
- Markdown rendering of finalized assistant messages.
- Hard cancellation of the underlying model stream on interrupt (the UI frees immediately; the harness has no stream-cancel hook yet).
- Remote HTTP/Gateway transport.


hetaoBackend:
Addressed the `github-code-quality` findings in 18dd1cee:
- `runtime.py`: replaced the bare `...` body on the `_ClientLike` Protocol method with a docstring (matches the harness convention for Protocol stubs, e.g. `SafetyTerminationDetector`).
- `test_tui_cli_main.py`: dropped the unnecessary `lambda: _FakeSession()` wrappers — `monkeypatch.setattr(cli, "_make_session", _FakeSession)` has identical behavior.

`ruff check` + `ruff format --check` clean; 109 TUI tests pass.

fancyboi999 (commented):
Pulled this onto a clean worktree and ran it end to end. It's in really good shape.

What I checked:

- **Ran the suite** — `PYTHONPATH=. uv run pytest tests/ -k tui` came back `112 passed, 1 skipped`. Keeping the pure layers (`view_state` / `runtime` / `render` / `command_registry` / `message_format`) Textual-free and unit-testing them with synthetic `StreamEvent`s is the right call: the reducer becomes the testable heart and `app.py` stays a thin shell.
- **Drove the real UI** through the pilot harness with a fake in-process session (no live model). Typed a turn, watched the streamed assistant text land and snap to the finalized render, with header / status / composer composing cleanly and token usage updating. The "stream stays plain text, snaps to Markdown when complete" choice is a nice anti-reflow detail.
- **Entry point** — `deerflow` with no TTY prints the headless guidance instead of hanging, and the `--print` / `--json` modes give it a clean scripting surface.
- **Skimmed for safety** — nothing alarming under `tui/`: no shelling out, no `eval`/`exec`, no raw network (it all goes through `DeerFlowClient`), and the single `os.environ` read is just launch-mode planning. `persistence.py` is a clean way to earn Web UI visibility without standing up the Gateway: best-effort `threads_meta` write into the shared DB, silent no-op on the `memory` backend (or if `get_app_config()` isn't resolvable), and all DB work pinned to one long-lived loop since the async engine binds to its creating loop.

I didn't read every line of `app.py` / `view_state.py`, but the risk surface plus the test coverage hit the places most likely to hide bugs. The "embedded over `DeerFlowClient`, doesn't fork agent behavior" framing is exactly what a workbench like this should be. Nice work.

hetaoBackend:
Thanks for the thorough pass, @fancyboi999 — appreciate you pulling it onto a clean worktree and driving it end to end.

A couple of notes on what you called out:
- The pure-layer split was deliberate so the reducer stays the testable heart and `app.py` is a thin shell — glad that landed.
- On `persistence.py`: the one-long-lived-loop detail is load-bearing (the async SQLAlchemy engine binds to its creating loop), and it degrades to a silent no-op on the `memory` backend so it never blocks a conversation.

For the next reviewer, the intentional **out-of-scope follow-ups** (also in the PR description) are:
- Hydrating a resumed thread's prior messages into the transcript (resume continues server-side today but starts the UI fresh).
- Hard cancellation of the underlying model stream on interrupt (the UI frees immediately; the harness has no stream-cancel hook yet).
- Markdown styling polish and a remote HTTP/Gateway transport.

(Re the `1 skipped` you saw — none of the `test_tui_*` tests skip here; that's a pre-existing test elsewhere that `-k tui` happens to match and skips in some environments. All TUI tests pass.)

Thanks again for the careful review!

backend/packages/harness/deerflow/tui/app.py
```diff
@@ -0,0 +1,627 @@
...
+        elif name == "tools":
+            self._dispatch(SystemMessage("Tools are listed in the agent's runtime; use /mcp for MCP servers."))
+        elif name == "uploads":
+            self._show_uploads()
+        elif name == "artifacts":
+            self._dispatch(SystemMessage("Artifacts appear inline as the agent writes them.", tone="info"))
+        elif name == "details":
+            self._dispatch(SystemMessage("Verbose activity is always shown in this build.", tone="info"))
+        else:
+            self._dispatch(SystemMessage(f"/{name} is not available yet.", tone="info"))
```
willem-bd, line 390: **`/resume` is dead-ended.** The command is registered in `command_registry.py:41` ("Resume a thread by id or title"), listed in the `/help` text on `app.py:37`, and `tests/test_tui_command_registry.py:73` even asserts it resolves to a `builtin` — but this dispatch has no `elif name == "resume":` branch, so submitting `/resume thread-abc` falls through to the final `else` and tells the user `/resume is not available yet.` Either wire it to `_open_thread_switcher` / a thread lookup, or remove it from the registry.

backend/packages/harness/deerflow/tui/session.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@dataclass
+class Session:
+    client: DeerFlowClient
+    writer: ThreadMetaWriter | None = None
+
+    def resolve_thread(self, plan: LaunchPlan) -> str | None:
+        """Resolve the thread id to run against, honoring --resume / --continue."""
+        if plan.thread_id:
```
willem-bd, line 29: **`--resume <title>` is silently broken.** `cli.py:64` and `command_registry.py:41` both promise resume "by id or title", but here the raw `plan.thread_id` is forwarded straight into `client.chat(thread_id=...)`. If the user passes a title (e.g. `deerflow --resume "Bug triage"`), the literal string becomes a brand-new checkpointer namespace and the prior conversation is not loaded — the user sees a blank thread and assumes resume is broken. Either resolve titles via `list_threads()` first, or drop "by id or title" from the docs.

backend/packages/harness/deerflow/tui/runtime.py
```diff
@@ -0,0 +1,123 @@
...
+    if message_type == "ai":
+        text = _extract_text(data.get("content"))
+        if text:
+            actions.append(AssistantDelta(id=str(data.get("id", "")), text=text))
+        for tool_call in data.get("tool_calls") or []:
+            if not isinstance(tool_call, dict):
+                continue
+            actions.append(
+                ToolStarted(
+                    tool_call_id=str(tool_call.get("id", "")),
```
willem-bd: **`str(tool_call.get("id", ""))` is wrong for `id: None`.** When a provider stream chunk arrives as `{"id": None, "name": "..."}` (real for some providers' first chunk before the id is assigned), `.get("id", "")` returns `None` (not the default — the default is only used when the key is *missing*), and `str(None) == "None"`. That truthy `"None"` then passes the `if not action.tool_call_id` guard in `view_state._apply_tool_started`, so the row sticks with `tool_call_id = "None"`. A second unrelated null-id call collapses into the same row and its result attaches to the wrong card.

Suggest: `str(tool_call.get("id") or "")` (or `tool_call.get("id") or ""` without `str`). Same pattern on line 83 / 85 for `tool_call_id` and `name`.

backend/packages/harness/deerflow/tui/session.py
```diff
@@ -0,0 +1,50 @@
...
+def open_session() -> Session:
+    """Build an embedded session backed by the configured checkpointer + thread store."""
+    from deerflow.client import DeerFlowClient
+    from deerflow.runtime.checkpointer.provider import get_checkpointer
+
+    from .persistence import build_persistence
+
+    checkpointer = get_checkpointer()
+    client = DeerFlowClient(checkpointer=checkpointer)
+    _loop, writer = build_persistence()
```
willem-bd: **Persistence is initialised even for headless one-shots.** `_make_session()` is called from `_run_print` / `_run_json` too, so `deerflow --print "hi"` spins up a daemon event-loop thread, runs `init_engine_from_config`, builds an `async_sessionmaker`, and (for Postgres) opens a connection pool — all to be discarded after one chat. The writer is never used in headless mode (the `app.py` worker is the only caller of `writer.ensure_created` / `writer.set_title`).

Suggest gating `build_persistence()` on whether the caller will actually use it — e.g. only construct it in `run_tui`, or do it lazily on first `ensure_created`.

backend/packages/harness/deerflow/tui/app.py
```diff
@@ -0,0 +1,627 @@
...
+        self.query_one("#status", Static).update(render_status(self.state, model=self._model, thread_label=self._thread_label(), spinner=spinner))
+
+
+def run_tui(plan) -> int:
+    """Construct the embedded session and run the app. Returns a process exit code."""
+    from .session import open_session
+
+    session = open_session()
+    app = DeerFlowTUI(session, plan)
+    app.run()
```
willem-bd: **`_LoopThread` + engine are never closed.** `app.run()` returns when the user `/quit`s or hits Ctrl+C, and `run_tui` returns 0 — but nothing calls `_LoopThread.close()` or disposes the SQLAlchemy engine singleton from `deerflow.persistence.engine`. The DB connection pool and the daemon event loop linger until interpreter teardown, the SQLite `connect` event listener stays hooked, and in tests/embedded callers that invoke `run_tui` repeatedly within one process the leak accumulates.

Suggest a `Session.close()` that calls `writer._loop.close()` (and ideally awaits `engine.dispose()`), and a `try / finally` around `app.run()` here.

backend/packages/harness/deerflow/tui/cli.py
```diff
@@ -0,0 +1,232 @@
...
+            return LaunchPlan(mode="headless-help", reason="--json needs a MESSAGE argument or piped stdin.")
+        return LaunchPlan(mode="json", message=message, read_stdin=message is None, thread_id=resume, continue_recent=continue_recent)
+
+    if args.cli:
+        if positional:
+            return LaunchPlan(mode="print", message=positional, thread_id=resume, continue_recent=continue_recent)
+        return LaunchPlan(
+            mode="headless-help",
+            reason='--cli needs a message. Try: deerflow --print "your question".',
+        )
```
willem-bd, line 114: **`--cli --continue` is rejected.** The `--cli` arm checks only `positional` and otherwise returns `headless-help` with reason "--cli needs a message". But `--continue` + piped stdin is the natural way to drop into a follow-up turn on the most recent thread; `--print` honours that combination (via `read_stdin=message is None` and `continue_recent`), and `--cli` should mirror it. Right now `deerflow --cli --continue` errors out, and `deerflow --cli --continue <<<'hello'` also errors out before ever reading stdin.

backend/packages/harness/deerflow/tui/app.py
```diff
@@ -0,0 +1,627 @@
...
+        latest_title: str | None = None
+        for action in stream_actions(self.session.client, text, thread_id=thread_id, **kwargs):
+            if self._cancelled:
+                break
+            if isinstance(action, ThreadTitle):
+                latest_title = action.title
+            self.call_from_thread(self._on_action, action)
+
+        if writer is not None and latest_title:
+            writer.set_title(thread_id, latest_title)
```
willem-bd, line 543: **Partial titles from cancelled runs leak into `threads_meta`.** The worker loop breaks early on `self._cancelled`, but the post-loop block still runs `writer.set_title(thread_id, latest_title)` if any `ThreadTitle` action was seen — including the title-middleware's first, possibly-truncated emission. So a user who hits Esc mid-stream can have their thread display name persisted as e.g. `"Resear"`.

Suggest guarding with `if writer is not None and latest_title and not self._cancelled` (or only honouring titles from a final/completed values event).

backend/packages/harness/deerflow/tui/app.py
```diff
@@ -0,0 +1,627 @@
...
+        composer = self.query_one("#composer", Input)
+        composer.value = value
+        composer.cursor_position = len(value)
+
+    def _open_palette(self, items: list) -> None:
+        if not items:
+            self._close_palette()
+            return
+        self._palette_items = items
+        self._palette_index = min(self._palette_index, len(items) - 1)
```
willem-bd, line 278: **Palette highlight follows the wrong item after filter narrowing.** `_palette_index` is only clipped to `len(items) - 1`, never reset to 0 when the filter set changes. Repro: type `/m` (palette = `[memory, mcp, model]`), press Down twice (`_palette_index = 2`, `model` selected), then type `e` (palette = `[memory, mcp]`). `min(2, 1) = 1`, so `mcp` is highlighted — but the user was hovering `model`, which no longer exists in the candidate set.

Suggest resetting `_palette_index = 0` in `on_input_changed` when the candidate list changes (or use a stable command-name cursor: remember the highlighted command's name and re-pick its new index, falling back to 0).

backend/packages/harness/deerflow/tui/app.py
```diff
@@ -0,0 +1,627 @@
...
+        super().__init__()
+        self.session = session
+        self.plan = plan
+        self.state = initial_state()
+        self._conv_thread_id: str | None = None
+        self._model = ""
+        self._skill_names: list[str] = []
+        self._skills = 0
+        self._tools = 0
+        self._spinner_idx = 0
```
willem-bd, line 165: **`_tools` is never populated.** `__init__` sets `self._tools = 0`, `_load_session_info` doesn't touch it, and nothing else writes it — yet `_refresh_header` passes it to `render_header`, which formats `"{tools} tools"` when nonzero. So the header is hard-wired to never show a tool count.

Either populate it (e.g. from `client.list_models()` / a tool-discovery call) or drop the `tools=` argument from `render_header` so the API doesn't imply a count the code never computes.

backend/packages/harness/deerflow/tui/view_state.py
```diff
@@ -0,0 +1,303 @@
...
+    full re-sends / cumulative snapshots vs. genuine incremental deltas:
+
+    * new text == accumulated, or starts with it  -> cumulative/re-send: replace
+    * accumulated starts with new text            -> stale/shorter re-send: keep
+    * otherwise                                   -> a real delta: append
+    """
+
+    rows = list(state.rows)
+    for i, row in enumerate(rows):
+        if isinstance(row, AssistantRow) and row.id == action.id and not row.error:
```
willem-bd, line 224: **Assistant-delta merge skips error rows by `not row.error`.** Today this is benign because `AssistantError` rows are appended with no `id` (so the match also fails). But the moment a future change sets `id` on error rows (e.g. to retry-replace them), the `not row.error` guard will silently *bypass* the matching row and append a duplicate. Tightening to match by `(id, not error)` AND removing the error row first would make the invariant explicit and future-proof — worth at least a comment if not a code change.

willem-bd (commented):
## Code review (high effort, recall-biased)

Reviewed the TUI subpackage end-to-end. 10 findings below, ranked roughly most-severe → least. The biggest ones are user-visible mismatches between what `/help` and `--help` advertise vs. what the code actually does (`/resume`, `--resume <title>`), plus a few state-leak / resource-lifecycle issues around the persistence wiring.

Findings:

1. `app.py` — `/resume` is in the registry + `/help` but `_handle_builtin` has no case for it (dead-ends with "not available yet").
2. `session.py` — `--resume <title>` is documented but only id is honoured; a title silently spawns a fresh thread.
3. `runtime.py` — `str(tool_call.get("id", ""))` becomes `"None"` when `id` is present-but-`None`, letting unrelated null-id tool calls collide.
4. `session.py` — `open_session()` unconditionally spins up the DB loop + engine for `--print` / `--json` headless runs.
5. `app.py` — `run_tui` never closes the `_LoopThread` or disposes the engine (lingers until interpreter exit).
6. `cli.py` — `--cli --continue` is rejected even though `--print` honours the same combination.
7. `app.py` — partial titles from cancelled runs are still written to `threads_meta`.
8. `app.py` — palette filter narrowing doesn't reset `_palette_index`, highlights an unrelated command after a Down + keystroke.
9. `app.py` — header always shows `0 tools` because `_tools` is never populated.
10. `view_state.py` — assistant-delta merge skips error rows; a future change that sets `id` on errors would silently create duplicate rows.

Nice work overall on the layering — `view_state` / `runtime` / `render` being pure and unit-tested separately made this review tractable.

hetaoBackend:
Thanks @willem-bd — genuinely excellent review, all 10 are valid. Fixed in cb3ff22c, with a regression test for each (126 TUI tests pass; `ruff check` + `ruff format --check` clean):

1. **`/resume` dead-end** — added the dispatch branch: `/resume <id-or-title>` switches threads, bare `/resume` opens the switcher.
2. **`--resume <title>`** — `Session.resolve_ref()` resolves id-or-title via `list_threads()` (used by both `--resume` and `/resume`); an unknown ref still falls back to a literal id.
3. **`str(get("id", ""))` → `"None"`** — great catch, this actually defeated the empty-id skip. Coerced via a None-safe helper (`"" if v is None else str(v)`) for tool-call id/name and tool-message id/name.
4. **Persistence for headless** — `open_session(persistence=False)` for `--print` / `--json`; no loop/engine/pool stood up.
5. **`_LoopThread` / engine leak** — `Session.close()` disposes the engine (`close_engine()`) and stops the loop, called from a `try/finally` around `app.run()`.
6. **`--cli --continue`** — now runs headless (mirrors `--print`: a piped message or `--continue` is enough).
7. **Partial title on cancel** — guarded `set_title` with `not self._cancelled`.
8. **Palette highlight** — `_palette_index` resets to 0 when the candidate set changes.
9. **`_tools` never populated** — dropped the `tools=` arg from `render_header`.
10. **`not row.error` guard** — added the explaining comment (error rows have no id so never match; it's belt-and-suspenders for a future change that gives them one).

Thanks again for the careful pass!

WillemJiang (approved):

PR #3764, opened 2026-06-24
fix: block unresolved suggestion template placeholders

### LINKED ISSUE #3756: Suggestion template [主题] placeholder leaks into submitted messages
### Describe the bug

When using suggestion templates (Research, Writing, Collecting, Learning buttons), the `[主题]` placeholder text can leak into the actual message sent to the backend if the user doesn't replace the selected text in time, or if the selection is lost before submission.

### Steps to reproduce

1. Open DeerFlow and navigate to the chat page
2. Click the "Research" (研究) suggestion button
3. The input box is filled with the template: `深入浅出的研究一下[主题]，并总结发现。`
4. The code attempts to auto-select `[主题]` via `setSelectionRange`, but this happens inside a 500ms `setTimeout`
5. If the user starts typing before the selection takes effect (or the selection is lost due to IME input, clicking elsewhere, or React rendering timing), the placeholder text remains in the input
6. Submit the message — the message bubble shows the raw `[主题]` placeholder text

### Expected behavior

The placeholder `[主题]` should always be either replaced by the user's input or stripped before the message is sent to the backend. The user should never see raw placeholder text in their submitted message.

### Actual behavior

The raw `[主题]` (or `[来源]`) placeholder text appears in the submitted message and is sent to the LLM backend as part of the user prompt, leading to confusing or degraded responses.

### Root cause

In `frontend/src/components/workspace/input-box.tsx` (lines ~1233-1252), the `handleSuggestionClick` function relies on a fragile mechanism to select the placeholder text:

```ts
const handleSuggestionClick = useCallback(
  (prompt: string | undefined) => {
    if (!prompt) return;
    textInput.setInput(prompt);
    setTimeout(() => {
      const textarea = document.querySelector<HTMLTextAreaElement>(
        "textarea[name='message']",
      );
      if (textarea) {
        const selStart = prompt.indexOf("[");
        const selEnd = prompt.indexOf("]");
        if (selStart !== -1 && selEnd !== -1) {
          textarea.setSelectionRange(selStart, selEnd + 1);
          textarea.focus();
        }
      }
    }, 500);
  },
  [textInput],
);
```

This approach has several problems:

1. **Race condition**: The 500ms `setTimeout` creates a window where the user can type before the selection takes effect
2. **DOM query instead of ref**: Uses `document.querySelector` instead of the existing `textareaRef` (defined at line ~206)
3. **IME interference**: Chinese IME input can cause `setSelectionRange` to be ignored
4. **No fallback in `handleSubmit`**: The submit handler (lines ~350-394) passes the message text verbatim to the backend with no placeholder validation or stripping

The template definitions in `frontend/src/core/i18n/locales/zh-CN.ts` (lines ~114-161) all use `[主题]` or `[来源]` as placeholders.

### Suggested fix

1. Replace `document.querySelector` with the existing `textareaRef` for reliable textarea access
2. Replace the hardcoded 500ms `setTimeout` with `requestAnimationFrame` or a state-driven approach to ensure the selection happens after React renders
3. Add a safety net in `handleSubmit`: strip any remaining placeholder patterns (e.g., `/\[(?:主题|topic|来源|source)\]/g`) before sending the message to the backend
4. If the placeholder is not replaced, show a validation hint asking the user to fill in the topic

### Environment

- OS: Windows 10 (build 26200)
- Browser: Chrome (latest)
- DeerFlow version: main branch (commit b66e325)

### Additional context

This is a pure frontend issue with a small, well-scoped fix area. The backend is unaffected except for receiving the malformed prompt.

Huixin615:
Closes #3756

## Why

Suggestion templates can prefill prompts with placeholders like `[topic]`, `[主题]`, or `[来源]`. If the user submits before replacing the selected placeholder, the raw placeholder can be sent to the backend and shown in the conversation, producing confusing LLM prompts.

## What changed

- Block chat submission when a known suggestion-template placeholder is still unresolved.
- Refocus the textarea and re-select the unresolved placeholder so the user can replace it.
- Replace the suggestion prefill selection logic with `textareaRef` + `requestAnimationFrame` instead of `document.querySelector` + a fixed timeout.
- Add localized validation text.
- Add an E2E regression test for blocking unresolved placeholders and allowing submission after replacement.

## Surface area

## Bug fix verification

- Test path that reproduces the bug: `frontend/tests/e2e/chat.spec.ts`

## Validation

- `cd frontend && pnpm lint`
- `cd frontend && pnpm exec playwright test --config /private/tmp/deer-flow-playwright-noweb.config.ts tests/e2e/chat.spec.ts`

Note: `pnpm test:e2e -- chat.spec.ts` initially hit local Playwright webServer/browser sandbox limits in this environment, so I verified the same chat spec against a running dev server with a temporary no-webserver Playwright config.


Huixin615:
@fancyboi999 辛苦帮忙review

MeloMei:
Nice work on this — the E2E test is genuinely impressive, especially the full lifecycle coverage (block → verify no backend call → verify selection → type replacement → verify correct payload). I had the same bug in my sights (#3771) so I know this code well, and here are a few observations from comparing our approaches:

**Things this PR does better than mine:**

1. **`Promise.reject` in handleSubmit** — I initially thought this was risky, but tracing the caller in `prompt-input.tsx`, the `.catch()` handler intentionally skips `clearSubmittedState()`, which is correct behavior when submission is blocked. My early `return` would have incorrectly cleared attached files. Good call.

2. **No `g` flag on the regex** — cleaner than my approach. I used `/gi` with an explicit `lastIndex = 0` reset, which works but treats the symptom. Dropping `g` entirely is the right fix since neither `.test()` nor a single `.exec()` needs global matching.

3. **`requestAnimationFrame` in handleSubmit** — safer than my synchronous textarea access, which could theoretically race with a React re-render triggered by the toast.

**A few suggestions:**

1. **Toast severity** — `toast.error()` (red) feels heavy for a validation hint. `toast.warning()` (amber) better communicates "please fix this" vs "something broke." Minor UX polish.

2. **SuggestionList coupling** — passing `textareaRef` directly couples `SuggestionList` to the parent's DOM ref and selection logic. An alternative is a callback prop (`onSelectPlaceholder`) that `InputBox` defines — `SuggestionList` just calls it, staying agnostic of DOM details. Inverts control nicely and makes the child trivially testable.

3. **Regex duplication** — the pattern is defined once as a module constant here, but the inline `findSuggestionTemplatePlaceholder` function lives inside `input-box.tsx`. If this check is ever needed elsewhere (e.g., in a future follow-up suggestion feature), extracting it to `core/suggestions/placeholders.ts` with unit tests would be a clean factoring. I built exactly that in #3771 — happy to port it as a follow-up if useful.

Happy to contribute the extracted utility + unit test coverage as a follow-up PR if the maintainers are open to it.

fancyboi999 (approved):
Reviewed now that #3771 is closed and this is the consolidated fix for #3756 — it's correct and well-targeted.

The subtle part is right: returning `Promise.reject` from `handleSubmit` on a blocked submit routes into `PromptInput`'s existing handler (`ai-elements/prompt-input.tsx:816-822`), where the `.catch()` deliberately *skips* `clearSubmittedState()` — so a blocked submit preserves the user's text and attached files. An early `return` would have fallen into the sync branch (`:825`) and cleared them. Easy to get wrong, and you got it right.

Also verified:
- `findSuggestionTemplatePlaceholder` uses a non-global regex with `.exec()` — no `lastIndex` statefulness, and `start`/`end` drive the re-selection cleanly.
- Swapping the `setTimeout(500)` + `document.querySelector` in `handleSuggestionClick` for `requestAnimationFrame` + the passed `textareaRef` removes the original race (the #3756 root cause) instead of just widening the window.
- The `suggestionPlaceholderRequired` key is added across `en-US` / `zh-CN` / `types`.

One note on method: I verified the submit-boundary behaviour by reading the caller chain rather than running the Playwright e2e locally. The lifecycle test you added (block → no backend call → reselect → replace → correct payload) covers exactly the path that matters.

@MeloMei's offer to extract the regex into a unit-tested `placeholders.ts` is a reasonable follow-up if the maintainers want it reused, but nothing here is blocking. No issues from me.

frontend/src/components/workspace/input-box.tsx
WillemJiang, line 753: @Huixin615, could you take a look at this warningg message?

Huixin615: > [@Huixin615](https://github.com/Huixin615), could you take a look at this warningg message?

Thanks for your review! I have fixed it.

WillemJiang (approved):

PR #3769, opened 2026-06-24
feat(agents): feature-gate the agents UI behind the agents_api flag (#3757)

### LINKED ISSUE #3757: Frontend sends redundant 403 requests when agents_api is disabled
### Describe the bug

When the `agents_api` feature is disabled in `config.yaml` (`agents_api.enabled: false`), the frontend unconditionally sends multiple requests to the agents API endpoints, all returning `403 Forbidden`. This results in unnecessary network traffic, console noise, and degraded performance on the agents page.

### Steps to reproduce

1. Ensure `agents_api.enabled: false` in `config.yaml`
2. Navigate to the agents page at `/workspace/agents`
3. Click "New Agent" (新建智能体), enter a valid name, and click "Continue" (继续)
4. Open the browser's Network tab and observe the requests

### Expected behavior

The frontend should check whether the `agents_api` feature is enabled before making any requests to `/api/agents` endpoints. When the feature is disabled:
- The agents page should display a "Feature not enabled" message instead of attempting API calls
- No requests should be sent to `/api/agents` or `/api/agents/check`
- The browser console should be free of 403 errors

### Actual behavior

The following redundant requests are observed:

```
GET  403  /api/agents                           (repeated 4 times)
GET  403  /api/agents/check?name=code-reviewer  (1 time)
```

The frontend makes these requests unconditionally on page navigation and component mount, without first checking whether the feature is enabled. Each visit to the agents page or name input triggers a fresh batch of 403 responses.

### Root cause

The frontend agents-related hooks and components do not have a pre-check for the `agents_api` configuration flag. They issue API calls regardless of whether the backend has the feature enabled, relying on the 403 response as the only signal that the feature is unavailable.

### Suggested fix

1. Add a feature flag check in the agents hooks: before making requests to `/api/agents`, query the configuration endpoint (e.g., `/api/models` or a dedicated config endpoint) to check if `agents_api` is enabled
2. Alternatively, cache the first 403 response and short-circuit subsequent requests within the same session
3. When `agents_api` is disabled, the agents page should render a user-friendly message (e.g., "Agents feature is not enabled. Please enable it in config.yaml.") instead of silently failing
4. Consider adding a `useFeatureFlag` hook or extending the existing configuration state to include feature availability

### Environment

- OS: Windows 10 (build 26200)
- Browser: Chrome (latest)
- DeerFlow version: main branch (commit b66e325)
- Config: `agents_api.enabled: false` in `config.yaml`

### Additional context

This is a pure frontend issue. The backend correctly returns 403 when the feature is disabled. The problem is that the frontend doesn't respect this signal and keeps sending duplicate requests. A small, well-scoped fix in the frontend hooks layer.

ggnnggez:
## Summary

Closes #3757.

When `agents_api.enabled: false`, the frontend used to fire redundant `GET /api/agents`, `/api/agents/check`, and `/api/agents/{name}` requests that all returned `403`, producing console noise and wasted traffic on the agents page.

This PR makes the frontend feature-aware instead of reacting to 403s:

- **Backend** — new read-only `GET /api/features` returning `{"agents_api": {"enabled": <bool>}}`, read through the hot-reloaded `get_config()` dependency (same pattern as `/api/models` exposing `token_usage.enabled`). No new restart-required state.
- **Frontend hook** — `useAgentsApiEnabled()` (TanStack Query, `staleTime: 0` + `refetchOnMount`), **fail-open**: while loading or on error the feature is treated as enabled, so an outage of `/api/features` can never hide a working feature.
- **Page-level gate** — a new `app/workspace/agents/layout.tsx` gates the whole `/workspace/agents` segment. When disabled it renders a "feature not enabled" panel and **no agent child component mounts**, so `/api/agents*` is never called — including on direct-URL access. This eliminates the 403s at the source rather than caching the first one.
- **Sidebar** — the Agents button is greyed out and shows a "Feature not enabled" tooltip on hover when disabled.

### Auto-enable on config change

Because `get_config()` hot-reloads `config.yaml` and the feature query refetches on mount, flipping `agents_api.enabled: true` and revisiting/refreshing the page enables the feature with **no rebuild**.

### No backend config leaked to users

The disabled message shown to end users does **not** expose backend configuration (no `agents_api.enabled` / `config.yaml`); it simply says the feature is not enabled and to contact an administrator (en + zh-CN).

## Screenshots

**1. Sidebar — the Agents button is greyed out with a "Feature not enabled" tooltip**

image

**2. Direct URL access to `/workspace/agents` — the page shows the feature-unavailable message**

image

## Testing

- Backend: `test_features_router.py` (enabled/disabled response shape).
- Frontend unit: `features.test.ts` (enabled / disabled / throw-on-error → fail-open).
- E2E: `agents-feature-disabled.spec.ts` asserts the contact-admin message renders, **zero `/api/agents*` requests** fire on direct navigation, and no `config.yaml`/`agents_api` text is leaked; `sidebar.spec.ts` asserts the greyed button (`aria-disabled`) and the hover tooltip.

### DIFF
--- a/frontend/src/components/workspace/workspace-nav-chat-list.tsx
+++ b/frontend/src/components/workspace/workspace-nav-chat-list.tsx
@@ -10,6 +10,11 @@ import {
   SidebarMenuButton,
   SidebarMenuItem,
 } from "@/components/ui/sidebar";
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from "@/components/ui/tooltip";
 import { useAgentsApiEnabled } from "@/core/agents";
 import { useI18n } from "@/core/i18n/hooks";
 
@@ -40,15 +45,26 @@ export function WorkspaceNavChatList() {
               </Link>
             </SidebarMenuButton>
           ) : (
-            <SidebarMenuButton
-              className="text-muted-foreground/50 cursor-not-allowed"
-              tooltip={t.sidebar.agentsDisabledTooltip}
-              aria-disabled
-              onClick={(e) => e.preventDefault()}
-            >
-              <BotIcon />
-              <span>{t.sidebar.agents}</span>
-            </SidebarMenuButton>
+            // Disabled: aria-disabled drives the sidebar CVA to suppress
+            // pointer events on the button, so wrap it in a hoverable span
+            // that still surfaces the "feature not enabled" tooltip.
+            <Tooltip>
+              <TooltipTrigger asChild>
+                <span className="block w-full">
+                  <SidebarMenuButton
+                    className="text-muted-foreground/50 cursor-not-allowed"
+                    aria-disabled
+                    tabIndex={-1}
+                  >
+                    <BotIcon />
+                    <span>{t.sidebar.agents}</span>
+                  </SidebarMenuButton>
+                </span>
+              </TooltipTrigger>
+              <TooltipContent side="right">
+                {t.sidebar.agentsDisabledTooltip}
+              </TooltipContent>
+            </Tooltip>
           )}
         </SidebarMenuItem>
       </SidebarMenu>
--- a/frontend/src/components/workspace/workspace-nav-chat-list.tsx
+++ b/frontend/src/components/workspace/workspace-nav-chat-list.tsx
@@ -50,9 +50,11 @@ export function WorkspaceNavChatList() {
             // that still surfaces the "feature not enabled" tooltip.
             <Tooltip>
               <TooltipTrigger asChild>
-                <span className="block w-full">
+                {/* cursor-not-allowed lives on the span (the element that
+                    still receives pointer events), not the inert button. */}
+                <span className="block w-full cursor-not-allowed">
                   <SidebarMenuButton
-                    className="text-muted-foreground/50 cursor-not-allowed"
+                    className="text-muted-foreground/50"
                     aria-disabled
                     tabIndex={-1}
                   >
--- a/frontend/src/components/workspace/workspace-nav-chat-list.tsx
+++ b/frontend/src/components/workspace/workspace-nav-chat-list.tsx
@@ -10,11 +10,13 @@ import {
   SidebarMenuButton,
   SidebarMenuItem,
 } from "@/components/ui/sidebar";
+import { useAgentsApiEnabled } from "@/core/agents";
 import { useI18n } from "@/core/i18n/hooks";
 
 export function WorkspaceNavChatList() {
   const { t } = useI18n();
   const pathname = usePathname();
+  const { enabled: agentsEnabled } = useAgentsApiEnabled();
   return (
     <SidebarGroup className="pt-1">
       <SidebarMenu>
@@ -27,15 +29,27 @@ export function WorkspaceNavChatList() {
           </SidebarMenuButton>
         </SidebarMenuItem>
         <SidebarMenuItem>
-          <SidebarMenuButton
-            isActive={pathname.startsWith("/workspace/agents")}
-            asChild
-          >
-            <Link className="text-muted-foreground" href="/workspace/agents">
+          {agentsEnabled ? (
+            <SidebarMenuButton
+              isActive={pathname.startsWith("/workspace/agents")}
+              asChild
+            >
+              <Link className="text-muted-foreground" href="/workspace/agents">
+                <BotIcon />
+                <span>{t.sidebar.agents}</span>
+              </Link>
+            </SidebarMenuButton>
+          ) : (
+            <SidebarMenuButton
+              className="text-muted-foreground/50 cursor-not-allowed"
+              tooltip={t.sidebar.agentsDisabledTooltip}
+              aria-disabled
+              onClick={(e) => e.preventDefault()}
+            >
               <BotIcon />
               <span>{t.sidebar.agents}</span>
-            </Link>
-          </SidebarMenuButton>
+            </SidebarMenuButton>
+          )}
         </SidebarMenuItem>
       </SidebarMenu>
     </SidebarGroup>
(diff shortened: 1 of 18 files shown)


frontend/src/core/agents/hooks.ts
```diff
@@ -3,12 +3,27 @@ import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
...
+    queryKey: ["features", "agents_api"],
+    queryFn: () => fetchAgentsApiEnabled(),
+    // Re-check on every mount so flipping config.yaml + revisiting the
+    // agents section auto-enables the feature without a rebuild.
+    staleTime: 0,
+    refetchOnMount: true,
+    retry: false,
+  });
+  // Fail-open: while loading or on error, do not hide a working feature.
+  return { enabled: data ?? true, isLoading: isPending };
```
willem-bd: **Fail-open re-introduces the 403 storm when `/api/features` is down _and_ the feature is actually disabled.**

The PR body explicitly chooses fail-open so "an outage of `/api/features` can never hide a working feature" — correct for the enabled-feature case. But the symmetric case isn't called out: when `agents_api.enabled` is genuinely `false` and `/api/features` returns 5xx or the network drops, `data` is `undefined` → `enabled = true` → the layout mounts the agents page → `useAgents` / `getAgent` / `checkAgentName` fire and each 403s, i.e. the exact console-noise/wasted-traffic problem (#3757) this PR eliminates comes back for the duration of the features-endpoint outage. With `retry: false`, even a single transient blip triggers it.

Caching the last-known value (or treating a prior `false` as sticky) would close the gap. Fine as-is if the tradeoff is intentional — just flagging it for an explicit decision.

willem-bd (commented):
Review of #3769 — posting two non-blocking design notes from review as inline comments.

Overall this is a well-built PR. The core guarantee — _no `/api/agents*` request fires when the feature is disabled, including on direct-URL navigation_ — is implemented correctly: during the query's `isPending` state the layout renders a loading placeholder rather than children, so the agents page never mounts before the flag is known, and the e2e test pins this with a request listener.

Verified during review:
- `agents_api.enabled` exists on `AgentsApiConfig` (default `false`).
- `get_config()` hot-reloads, and `AppConfig._apply_singleton_configs` calls `load_agents_api_config_from_dict(...)`, so `/api/features` (reads `config.agents_api.enabled`) and the agents router's `_require_agents_api_enabled` (reads `get_agents_api_config()`) stay consistent across reloads — no split-brain.
- The unit test's `@/core/api/fetcher` mock matches `api.ts`'s actual import (`import { fetch } from "@/core/api/fetcher"`), and `getBackendBaseURL` is mocked to `""` so the assertion `toHaveBeenCalledWith("/api/features")` holds.
- `/api/features` inherits the app's auth/CSRF middleware (same mount path as `/api/models`), so it introduces no new unauthenticated config leak; `agents_api` is not in `STARTUP_ONLY_FIELDS`, so the "no rebuild" auto-enable claim holds.

No correctness bugs. Two design considerations inline.

frontend/src/components/workspace/workspace-nav-chat-list.tsx
```diff
@@ -27,15 +34,40 @@ export function WorkspaceNavChatList() {
...
+            // pointer events on the button, so wrap it in a hoverable span
+            // that still surfaces the "feature not enabled" tooltip.
+            <Tooltip>
+              <TooltipTrigger asChild>
+                {/* cursor-not-allowed lives on the span (the element that
+                    still receives pointer events), not the inert button. */}
+                <span className="block w-full cursor-not-allowed">
+                  <SidebarMenuButton
+                    className="text-muted-foreground/50"
+                    aria-disabled
```
willem-bd, line 61: **Disabled-state tooltip is hover-only; keyboard/screen-reader users get no explanation.**

`tabIndex={-1}` removes the disabled Agents entry from the tab order, so keyboard and screen-reader users never focus it, and Radix Tooltip fires on hover/focus — so the "Feature not enabled" explanation (`t.sidebar.agentsDisabledTooltip`) is only reachable with a mouse (the test's `agentsButton.hover({ force: true })` confirms pointer events are otherwise suppressed). The greyed button is visible, but the reason is not announced to non-mouse users.

Consider moving the tooltip trigger to a focusable element, or adding `aria-describedby` / visually-hidden text so the disabled reason is announced.

WillemJiang:
@ggnnggez, please change the update of backend/CLAUDE.md to the backend/Agent.md

ggnnggez:
Thanks for the thorough review @willem-bd — both design notes addressed. Also rebased onto latest `main`.

**1. Fail-open re-introducing the 403 storm** (`frontend/src/core/agents/hooks.ts`) — fixed in e64bcc2d.
The flag is now **sticky**: the last definitive `/api/features` answer is persisted (`localStorage`) and used as the fallback before failing open. So when `agents_api` is genuinely `false` and `/api/features` is down (5xx / network drop, `retry: false`), the last-known `false` holds and the agents UI doesn't remount — the 403 storm stays closed. It still fails open **only** when no value has ever been observed, preserving "an outage can never hide a working feature."
- Resolution logic: `live ?? cached ?? true`.
- The cached value is read in a mount effect (not during render), so the first client render matches the server and there's no hydration mismatch on the non-loading-gated sidebar.
- Coverage: unit tests for the resolve/persistence logic + an e2e that sees `disabled`, then 500s `/api/features` on reload and asserts zero `/api/agents` requests.

**2. Disabled tooltip hover-only (a11y)** (`frontend/src/components/workspace/workspace-nav-chat-list.tsx`) — fixed in 77ee2ceb.
Removed `tabIndex={-1}` so the disabled Agents entry stays in the tab order (keyboard-focusable), and wired `aria-describedby` to a visually-hidden reason element so screen-reader users hear "Feature not enabled" instead of it being reachable only by mouse hover. The mouse hover tooltip and the existing `getByRole("button", { name: "Agents" })` / `aria-disabled` contract are preserved.
- Coverage: extended the sidebar e2e to assert the `aria-describedby` wiring + focusability.

WillemJiang (approved):

PR #3775, opened 2026-06-25
feat(skills): add deferred skill discovery via describe_skill tool

sontianye:
## Why

All enabled skills' full metadata (name, description, location, mutability) is injected into the system prompt unconditionally. With 18 built-in skills this costs ~900 tokens per turn; with 50-100+ custom skills it becomes the largest dead-weight in the system prompt, directly undermining the prefix-cache optimization that the static system prompt design depends on.

Meanwhile, `tool_search` already solves the identical problem for MCP tools with a proven deferred pattern: names in prompt, schema on demand. Skills lack the same mechanism.

## What changed

- **New `describe_skill` tool** — closure over a `SkillCatalog`, returns structured metadata (description, allowed-tools, location) via `Command`+`ToolMessage`. Symmetric with `build_tool_search_tool`: same query syntax (`select:`, `+required-prefix`, free-text regex), same return shape.

- **Compact `<skill_index>` in system prompt** — when `skills.deferred_discovery` is enabled, the prompt renders skill names only (~200 tokens) instead of full metadata (~900 tokens). The LLM calls `describe_skill(name)` to fetch details before deciding whether to `read_file` the full SKILL.md.

- **New config flag `skills.deferred_discovery`** — independent from `tool_search.enabled`, defaults to `false`. Zero behavior change unless explicitly opted in.

| Config | Default | Effect |
|---|---|---|
| `skills.deferred_discovery` | `false` | `false` → legacy `<available_skills>` full metadata (current behavior) |
| | | `true` → `<skill_index>` names-only + `describe_skill` tool |

- **Backward compatible** — `get_skills_prompt_section` gains an optional `skill_names` parameter. Existing callers that don't pass it get identical behavior.

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [x] **Agents / LangGraph** — agent node wiring, prompt change
- [ ] **Sandbox**
- [x] **Skills** — new modules under `skills/` (catalog.py, describe.py), config field
- [ ] **Dependencies**
- [ ] **Default behavior change** — no (gated by opt-in config flag, defaults to `false`)

## Validation

```
cd backend && uv run python -m pytest tests/ -k "skill or prompt or lead_agent or config" -x
# 984 passed, 1 skipped, 0 failures

cd backend && uv run python -m pytest tests/ --ignore=tests/blocking_io -q
# 5086 passed, 19 skipped, 0 failures
```


willem-bd (commented):
Thanks @sontianye. I found one issue with the default-behavior path.

### [P2] `describe_skill` instruction baked into the shared prompt breaks the legacy (default) path

**Location:** `backend/packages/harness/deerflow/agents/lead_agent/prompt.py`, line ~563 (`<critical_reminders>` → "Skill First").

**Problem:** The `<critical_reminders>` block is part of the shared system prompt rendered for *every* agent build. This PR rewrites the "Skill First" line to:

> For complex tasks, call `describe_skill(name)` to check if a matching skill exists, then `read_file` to load it.

But `describe_skill` is only registered when `skills.deferred_discovery=true` (the tool-append in `agent.py` is gated on `skill_setup.describe_skill_tool`, which is `None` when the flag is off). Since `deferred_discovery` **defaults to `false`**, the default-config agent now ships a prompt instructing the model to call a tool that does not exist. The model will either emit a `describe_skill` tool call that fails, or skip skill discovery because the gating instruction never resolves. The previous wording ("Always load the relevant skill before starting complex tasks") was tool-agnostic and worked correctly with the legacy `<available_skills>` block that still renders full metadata.

**Evidence:** `grep -n "describe_skill" prompt.py` shows the only occurrence in the shared template is the hardcoded line 563; there is no conditional around it. Tool registration in `agent.py` (`if skill_setup.describe_skill_tool: final_tools.append(...)`) correctly withholds the tool in legacy mode, so the prompt and the toolset disagree.

**Suggested fix:** Gate the instruction on the deferred path. For example, keep the legacy wording for the default path and only emit the `describe_skill` instruction when `skill_names is not None` (or pass a `deferred_skills` boolean into `apply_prompt_template` alongside `skill_names` and switch the line via an f-string/conditional, mirroring how `deferred_tools_section` is already threaded through).

**Test:** Add an assertion in `test_get_skills_prompt_section_legacy_path_when_skill_names_none` that the rendered prompt does **not** mention `describe_skill`, and a mirror assertion in the deferred-path test that it does.

willem-bd (commented):
Thanks @sontianye. I found one issue with the default-behavior path.

### [P2] `describe_skill` instruction baked into the shared prompt breaks the legacy (default) path

**Location:** `backend/packages/harness/deerflow/agents/lead_agent/prompt.py`, line ~563 (`<critical_reminders>` → "Skill First").

**Problem:** The `<critical_reminders>` block is part of the shared system prompt rendered for *every* agent build. This PR rewrites the "Skill First" line to:

> For complex tasks, call `describe_skill(name)` to check if a matching skill exists, then `read_file` to load it.

But `describe_skill` is only registered when `skills.deferred_discovery=true` (the tool-append in `agent.py` is gated on `skill_setup.describe_skill_tool`, which is `None` when the flag is off). Since `deferred_discovery` **defaults to `false`**, the default-config agent now ships a prompt instructing the model to call a tool that does not exist. The model will either emit a `describe_skill` tool call that fails, or skip skill discovery because the gating instruction never resolves. The previous wording ("Always load the relevant skill before starting complex tasks") was tool-agnostic and worked correctly with the legacy `<available_skills>` block that still renders full metadata.

**Evidence:** `grep -n "describe_skill" prompt.py` shows the only occurrence in the shared template is the hardcoded line 563; there is no conditional around it. Tool registration in `agent.py` (`if skill_setup.describe_skill_tool: final_tools.append(...)`) correctly withholds the tool in legacy mode, so the prompt and the toolset disagree.

**Suggested fix:** Gate the instruction on the deferred path. For example, keep the legacy wording for the default path and only emit the `describe_skill` instruction when `skill_names is not None` (or pass a `deferred_skills` boolean into `apply_prompt_template` alongside `skill_names` and switch the line via an f-string/conditional, mirroring how `deferred_tools_section` is already threaded through).

**Test:** Add an assertion in `test_get_skills_prompt_section_legacy_path_when_skill_names_none` that the rendered prompt does **not** mention `describe_skill`, and a mirror assertion in the deferred-path test that it does.

sontianye:
Thanks for catching this @willem-bd. Fixed in 1e4a1250 — extracted the Skill First line into a `{skill_first_reminder}` placeholder, conditionally rendered based on `skill_names is not None`. Added two `apply_prompt_template` tests asserting mutual exclusion between the legacy and deferred wordings.

willem-bd (commented):
**Correction — withdrawing my earlier review above.** I misread the prompt assembly and the finding does not hold.

The "Skill First" instruction is **already gated** on `skill_names is not None`:

```python
_skill_first = (
    "- Skill First: For complex tasks, call describe_skill(name) …"   # deferred path only
    if skill_names is not None
    else "- Skill First: Always load the relevant skill …"            # legacy/default path
)
```

and this PR *removes* the previously-hardcoded "Skill First" line from the shared `<critical_reminders>` template. So in the default config (`skills.deferred_discovery=false`, `skill_names is None`) the prompt keeps the tool-agnostic wording and never references `describe_skill` — the prompt and the (withheld) toolset agree. My earlier claim that "there is no conditional around it" was wrong; the change is a clean net improvement. Apologies for the noise, @sontianye.

willem-bd (commented):
To be precise: the same incorrect finding appears **twice** above (posted twice by mistake). Both copies are withdrawn — the gating shown in my correction applies to both. Sorry for the extra noise, @sontianye.

sontianye:
Friendly ping — happy to address any feedback if needed.

backend/packages/harness/deerflow/skills/describe.py
```diff
@@ -0,0 +1,179 @@
...
+
+    names = ", ".join(sorted(skill_names))
+    evolution = f"\n{skill_evolution_section}" if skill_evolution_section else ""
+
+    return f"""<skill_system>
+You have access to skills that provide optimized workflows for specific tasks.
+
+**Skill Discovery:**
+1. Check <skill_index> for a skill name that matches your task
+2. Call describe_skill(name) to fetch its description and capabilities
```
willem-bd, line 167: **Prompt/tool parameter-name mismatch — strict function-calling models will hard-fail the deferred-discovery flow.**

The prompt tells the LLM to `Call describe_skill(name)`, but the tool's declared parameter is `query` (line 63). The same mismatch shows up in `agents/lead_agent/prompt.py:851` (the `skill_first_reminder`).

A strict tool-calling model dutifully emits `{"name": "describe_skill", "arguments": {"name": "podcast-generation"}}`. Pydantic rejects it — verified live:

```
ValidationError: 1 validation error for describe_skill
query
  Field required [type=missing, input_value={'name': 'podcast-generation', 'tool_call_id': 't1'}]
```

So the LLM gets a tool error instead of skill metadata, and the entire deferred-discovery flow degrades on exactly the models the prefix-cache optimization is aimed at. The existing tests only assert the literal string `describe_skill(name)` is present in the prompt — they don't catch this.

Suggest one of:
- rename the tool parameter to `name` (and update the docstring's `query forms:` accordingly), **or**
- change the prompt to reference `describe_skill(query=…)` and teach the query syntax (`select:`, `+`, free-text) that the tool actually accepts.

Also worth adding a regression test that actually invokes the tool with the argument name written in the prompt, so future prompt edits can't silently drift again.

backend/packages/harness/deerflow/agents/lead_agent/prompt.py
```diff
@@ -818,6 +846,14 @@ def apply_prompt_template(
     custom_mounts_section = _build_custom_mounts_section(app_config=app_config)
     acp_and_mounts_section = "\n".join(section for section in (acp_section, custom_mounts_section) if section)
 
+    # Gate the "Skill First" instruction on the deferred discovery path:
+    # legacy mode uses tool-agnostic wording; deferred mode references describe_skill.
+    skill_first_reminder = (
```
willem-bd, line 948: Same `describe_skill(name)` vs. tool-parameter `query` mismatch as `skills/describe.py:166` — see that thread. The reminder here is the more prominent of the two: it lands directly in `<critical_reminders>`, so it's the first place the LLM will look when deciding how to invoke the tool.

backend/packages/harness/deerflow/skills/catalog.py
```diff
@@ -0,0 +1,102 @@
...
+    try:
+        return re.compile(pattern, re.IGNORECASE)
+    except re.error:
+        return re.compile(re.escape(pattern), re.IGNORECASE)
+
+
+# NOTE: frozen=True without slots=True keeps __dict__, which is what lets the
+# @cached_property fields below cache (they write to instance.__dict__, bypassing
+# the frozen __setattr__). Do NOT add slots=True or hash/names break at runtime.
+@dataclass(frozen=True)
```
willem-bd, line 41: **`@dataclass(frozen=True)` implies hashability, but real (non-empty) catalogs are unhashable.**

`Skill` is a plain mutable `@dataclass` (`skills/types.py:19`), so a tuple of `Skill` instances is unhashable, and hashing any populated `SkillCatalog` raises. Verified live:

```
>>> hash(SkillCatalog((real_skill,)))
TypeError: unhashable type: 'Skill'
```

The two tests that check this — `test_frozen_catalog_is_hashable` and `test_setup_frozen` — both use empty inputs (`SkillCatalog(())`, `build_skill_search_setup([], enabled=True)`), so the tuple has nothing to hash and the assertion passes trivially. That's a misleading contract: `SkillSearchSetup` inherits the same issue.

No current caller hashes these objects, so this isn't a live bug — but the `frozen=True` + "must be hashable" test comments say otherwise. Either:
- make `Skill` `frozen=True` (probably best — it already isn't mutated after load), **or**
- drop the `frozen=True` claim on `SkillCatalog` / `SkillSearchSetup` and delete the misleading hashability tests.

Whichever direction, the tests should exercise a populated catalog so the invariant is actually pinned.

backend/packages/harness/deerflow/skills/catalog.py
```diff
@@ -0,0 +1,102 @@
...
+        Returns at most ``MAX_RESULTS`` skills, ranked by relevance.
+        """
+        query = query.strip()
+        if not query:
+            return []
+
+        # ── Exact selection ────────────────────────────────────────────
+        if query.startswith("select:"):
+            wanted = {n.strip() for n in query[7:].split(",")}
+            return [s for s in self.skills if s.name in wanted][:MAX_RESULTS]
```
willem-bd: **`select:` is silently truncated to `MAX_RESULTS=5` — that defeats its purpose.**

The other three paths (`+prefix`, free-text, empty) rationally cap ranked search results at 5. But `select:` is an *exact request by name*, not a ranked search: the LLM is saying "I already know I want these specific skills." With a typical 15-20 skill fleet the model may issue `select:a,b,c,d,e,f,g,h` in one call and only get 5 back — in insertion order, with no truncation marker, so the model doesn't know it needs to re-issue for the rest.

Suggest either dropping the `[:MAX_RESULTS]` slice on the `select:` branch, or emitting a trailing marker (e.g. `"...and 3 more (re-issue select: for remaining)"`) so the model can recover. The tool docstring should mention whatever cap ends up sticking, too.

willem-bd (commented):
Thanks for this — the deferred-tools pattern maps cleanly onto skills and the diff is well-structured. A few things I'd like to see addressed before we merge; details left inline. Two cross-cutting notes I couldn't attach to a diff line:

1. **`backend/AGENTS.md` isn't updated.** The repo convention (`backend/AGENTS.md`) says "**CRITICAL: Always update README.md and AGENTS.md after every code change** … Update `AGENTS.md` for development changes (architecture, commands, workflows, internal systems)." This PR adds a new config field (`skills.deferred_discovery`), a new tool (`describe_skill`), and two new modules (`skills/catalog.py`, `skills/describe.py`) — exactly the kind of change the *Skills System* / *Tool System* / *Config Schema* sections are meant to capture. Please extend those sections so the next contributor learns that `<skill_index>` can replace `<available_skills>` without spelunking through the code.

2. **Some findings are inline on `agent.py` but the parity gap actually lives in `client.py`** (which isn't in the diff, so GitHub won't let me anchor there directly). See the `agent.py` comment for detail.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -520,17 +535,27 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
...
             ),
             state_schema=ThreadState,
         )
 
     # Custom agents can update their own SOUL.md / config via update_agent.
     # The default agent (no agent_name) does not see this tool.
     extra_tools = [update_agent] if agent_name else []
+    # Build skill search setup from policy-filtered skills (same list used for
+    # tool-policy filtering), so describe_skill only exposes allowed skills.
+    skill_setup = build_skill_search_setup(
```
willem-bd, line 574: **Parity gap: `DeerFlowClient` (embedded path) never wires deferred skill discovery.**

This wiring — `build_skill_search_setup(...)` + appending `describe_skill` + threading `skill_names` into `apply_prompt_template` — is only added to the LangGraph agent-factory paths (bootstrap + default). The embedded client at `client.py:242-265` calls the same `apply_prompt_template` / `assemble_deferred_tools` sequence but doesn't get the new hookup, so with `skills.deferred_discovery: true`:

- LangGraph path → `<skill_index>` (compact) + `describe_skill` tool ✅
- Embedded `DeerFlowClient.chat` / `stream` path → legacy `<available_skills>` full metadata, no `describe_skill` tool ❌

The config flag becomes a silent no-op on one of the two supported entry points — exactly the drift the harness/app split is meant to prevent. `assemble_deferred_tools` was refactored earlier precisely so tool_search would behave identically across both call sites; the same treatment is needed here.

Suggest either:
- factoring the skill-search wiring into a small helper that both `agent.py` and `client.py` call (mirrors `assemble_deferred_tools`), **or**
- inlining the same three lines into `_ensure_agent()` and adding a parity test that asserts the two prompt outputs contain `<skill_index>` in both entry points when `deferred_discovery` is enabled.

backend/packages/harness/deerflow/skills/describe.py
sontianye: Fixed in `d2aa0ef1`: renamed the tool parameter from `query` to `name`, matching the prompt wording. Added a schema regression test (`test_describe_skill_parameter_name_matches_prompt`) that pins `"name"` in the JSON schema and asserts `"query"` does not exist, so future drift is caught at the test layer rather than at runtime.

backend/packages/harness/deerflow/agents/lead_agent/prompt.py
sontianye: Same root cause as the `describe.py` thread — fixed together in `d2aa0ef1`. Both the tool parameter and all prompt invocations now consistently use `name`.

backend/packages/harness/deerflow/skills/catalog.py
sontianye: Fixed in `d2aa0ef1`: made `Skill` `@dataclass(frozen=True)` and changed `allowed_tools: list[str] | None` → `tuple[str, ...] | None` (lists are unhashable). Updated `skill_storage.py` to use `dataclasses.replace()` instead of direct field mutation. Updated the parser, all test helpers, and `test_frozen_catalog_is_hashable` to use a populated catalog so the invariant is actually exercised rather than passing trivially on an empty input.

backend/packages/harness/deerflow/skills/catalog.py
sontianye: Fixed in `d2aa0ef1`: removed `[:MAX_RESULTS]` from the `select:` branch only. The tool docstring now distinguishes "no cap" (`select:`) from "up to 5" (keyword/prefix search). The test was renamed `test_select_returns_all_requested` and asserts `len(result) == len(sample_skills)` to make the contract explicit.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
sontianye: Fixed in `d2aa0ef1`: `DeerFlowClient._ensure_agent` now calls `build_skill_search_setup(...)` and appends `describe_skill_tool` + threads `skill_names` into `apply_prompt_template`, exactly mirroring the LangGraph path. Two parity tests (`test_deferred_skill_discovery_wired_when_enabled` / `_disabled`) assert the symmetry holds for both entry points.

sontianye:
Thanks @willem-bd for the detailed review! All four inline issues and both cross-cutting notes are addressed in `d2aa0ef1`:

**AGENTS.md** — `backend/AGENTS.md` is a single-line file (`@./CLAUDE.md`), delegating all documentation to `CLAUDE.md`. The *Skills System* and *Config Schema* sections in `CLAUDE.md` were updated in `d2aa0ef1` to document `skills.deferred_discovery`, the `describe_skill` tool, and the new `catalog.py` / `describe.py` modules.

**Parity gap (`client.py`)** — `DeerFlowClient._ensure_agent` now wires `build_skill_search_setup` and threads `skill_names` into `apply_prompt_template`, matching the LangGraph path. Two parity tests assert the symmetry for both paths.

Details for each inline finding are in the respective threads.

WillemJiang:
@sontianye Thanks for your contribution. Please resolve the conflicts with the main branch.

sontianye:
Ready for merge. All review feedback addressed, plus discovered and fixed a bug during audit: `describe.py` was re-applying `[:MAX_RESULTS]` which truncated `select:` queries. Added regression test. 153 tests passing.

WillemJiang (approved):

PR #3786, opened 2026-06-25
fix: allow valid heredoc commands in sandbox audit

### LINKED ISSUE #3774: [bug] SandboxAuditMiddleware incorrectly blocks valid heredoc commands due to shlex.ValueError
## Bug Description

`SandboxAuditMiddleware._classify_single_command` uses `shlex.split()` to tokenize commands before pattern matching. When a command contains a heredoc (e.g. `python3 << 'PYEOF' ... PYEOF`), `shlex.split()` raises `ValueError` due to unclosed quotes — and the exception handler immediately returns `"block"`, treating all heredoc commands as high-risk.

This causes **false positives** that block legitimate Python analysis commands.

## Steps to Reproduce

Upload an image to a DeerFlow thread and ask the agent to analyze it. The agent will attempt something like:

\`\`\`bash
python3 << 'PYEOF'
from PIL import Image
img = Image.open('/mnt/user-data/uploads/image.jpg')
print(img.size)
PYEOF
\`\`\`

This gets blocked with: `Command blocked: ...`

## Root Cause

In `sandbox_audit_middleware.py`:

\`\`\`python
try:
    tokens = shlex.split(command)
    joined = " ".join(tokens)
    for pattern in _HIGH_RISK_PATTERNS:
        if pattern.search(joined):
            return "block"
except ValueError:
    # shlex.split fails on unclosed quotes — treat as suspicious
    return "block"   # ← heredocs always hit this path
\`\`\`

Heredoc syntax is valid bash but `shlex` cannot parse multiline strings, so it always raises `ValueError` and is always blocked.

## Fix

Change the `except ValueError` handler to fall through to the existing `_MEDIUM_RISK_PATTERNS` check on the `normalized` string (already computed before the try block), instead of immediately returning `"block"`:

\`\`\`python
except ValueError:
    # shlex.split fails on heredocs / multiline strings — fall through to pattern match
    pass
\`\`\`

## Impact

- Any multiline Python/bash heredoc command is completely blocked
- Image analysis, data processing, and file inspection tasks silently fail
- Agent reports "I cannot see this image" even when the file is accessible

Huixin615:
Fixes #3774

## Why

`SandboxAuditMiddleware` was blocking valid multiline heredoc commands such as Python image/file analysis snippets.

These commands are valid bash, but `shlex.split()` can raise `ValueError` on heredoc or multiline input. The previous handler treated any `ValueError` as suspicious and returned `block`, which caused legitimate image analysis and file inspection commands to fail.

The issue and maintainer review confirmed this was over-conservative: high-risk patterns are already scanned on the normalized raw command before the `shlex.split()` path, so falling through on `ValueError` keeps the primary high-risk guard in place.

## What changed

- Valid heredoc / multiline commands that `shlex` cannot parse are no longer automatically blocked.
- The middleware still scans the raw normalized command for high-risk patterns before attempting `shlex.split()`.
- Added regression tests for:
  - a benign Python heredoc command, which now passes classification
  - a heredoc command whose shell portion contains a high-risk pattern, which still blocks

## Surface area

## Screenshots / Recording

N/A. This is a backend middleware classification fix with regression tests.

## Bug fix verification

- `backend/tests/test_sandbox_audit_middleware.py::TestClassifyCommand::test_benign_heredoc_classified_as_pass`
- `backend/tests/test_sandbox_audit_middleware.py::TestClassifyCommand::test_heredoc_with_high_risk_pattern_still_blocks`

## Validation

From `backend/`:

```bash
uv run pytest tests/test_sandbox_audit_middleware.py
uv run ruff check packages/harness/deerflow/agents/middlewares/sandbox_audit_middleware.py tests/test_sandbox_audit_middleware.py
uv run ruff format --check packages/harness/deerflow/agents/middlewares/sandbox_audit_middleware.py tests/test_sandbox_audit_middleware.py
Additional git validation from repo root:
git diff --check
Results:
149 passed
ruff check: all checks passed
ruff format --check: 2 files already formatted
git diff --check: no whitespace errors

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/sandbox_audit_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/sandbox_audit_middleware.py
@@ -150,8 +150,9 @@ def _classify_single_command(command: str) -> str:
             if pattern.search(joined):
                 return "block"
     except ValueError:
-        # shlex.split fails on unclosed quotes — treat as suspicious
-        return "block"
+        # Heredocs and other multiline shell forms may be valid bash but
+        # unparseable by shlex. Raw high-risk patterns were already checked.
+        pass
 
     for pattern in _MEDIUM_RISK_PATTERNS:
         if pattern.search(normalized):
--- a/backend/tests/test_sandbox_audit_middleware.py
+++ b/backend/tests/test_sandbox_audit_middleware.py
@@ -183,19 +183,12 @@ class TestClassifyCommand:
     def test_safe_classified_as_pass(self, cmd):
         assert _classify_command(cmd) == "pass", f"Expected 'pass' for: {cmd!r}"
 
-    def test_benign_heredoc_classified_as_pass(self):
-        cmd = """python3 << 'PYEOF'
-from PIL import Image
-img = Image.open('/mnt/user-data/uploads/image.jpg')
-print(img.size)
-PYEOF"""
+    def test_unparseable_heredoc_classified_as_pass(self):
+        cmd = "python3 << 'EOF'\necho it's fine\nEOF"
         assert _classify_command(cmd) == "pass"
 
-    def test_heredoc_with_high_risk_pattern_still_blocks(self):
-        cmd = """python3 << 'PYEOF'
-print('inspect first')
-PYEOF
-cat /etc/shadow"""
+    def test_unparseable_heredoc_with_high_risk_pattern_still_blocks(self):
+        cmd = "python3 << 'EOF'\necho it's fine\ncat /etc/shadow\nEOF"
         assert _classify_command(cmd) == "block"
 
     # --- Compound commands: sub-command splitting ---
--- a/backend/tests/test_sandbox_audit_middleware.py
+++ b/backend/tests/test_sandbox_audit_middleware.py
@@ -183,6 +183,21 @@ class TestClassifyCommand:
     def test_safe_classified_as_pass(self, cmd):
         assert _classify_command(cmd) == "pass", f"Expected 'pass' for: {cmd!r}"
 
+    def test_benign_heredoc_classified_as_pass(self):
+        cmd = """python3 << 'PYEOF'
+from PIL import Image
+img = Image.open('/mnt/user-data/uploads/image.jpg')
+print(img.size)
+PYEOF"""
+        assert _classify_command(cmd) == "pass"
+
+    def test_heredoc_with_high_risk_pattern_still_blocks(self):
+        cmd = """python3 << 'PYEOF'
+print('inspect first')
+PYEOF
+cat /etc/shadow"""
+        assert _classify_command(cmd) == "block"
+
     # --- Compound commands: sub-command splitting ---
 
     @pytest.mark.parametrize(

### CODE REVIEW

**Comment by Huixin615** [comment]
@fancyboi999 辛苦review一下

**Review by fancyboi999 — COMMENTED** [review]
Thanks for taking this — I traced #3774, and the production change is exactly right. Falling through on `ValueError` instead of returning `block` is safe: high-risk patterns are scanned against the whole normalized command *before* this branch ever matters — once in `_classify_command` Pass 1 (`sandbox_audit_middleware.py:175-178`) and again at the top of `_classify_single_command` (`:139-142`) — so dropping the shlex-failure block can't let a high-risk command slip through. I confirmed it end-to-end: a heredoc that genuinely breaks `shlex` (`python3 << 'EOF'` / `echo it's fine` / `EOF`, which raises `ValueError: No closing quotation`) classifies as `block` on `main` and `pass` on this branch. The fix works.

One thing to fix before merge — **the two new regression tests don't actually exercise the branch you changed.**

Both `test_benign_heredoc_classified_as_pass` and `test_heredoc_with_high_risk_pattern_still_blocks` use heredocs whose quotes are *balanced*, so `shlex.split()` parses them without raising — the `except ValueError` block is never entered. I verified by reverting only `sandbox_audit_middleware.py` to `main` and re-running: both still pass (`2 passed`). They'd stay green even with the fix reverted, so they don't anchor it.

- `test_benign_heredoc_classified_as_pass`: the apostrophe-free body parses cleanly via `shlex`; it classifies as `pass` with or without this change.
- `test_heredoc_with_high_risk_pattern_still_blocks`: this blocks via the Pass-1 scan on `cat /etc/shadow`, before `shlex` is ever reached — it would block even without a heredoc, so it doesn't prove the high-risk guard survives the new fall-through path either.

To make the benign test a real regression guard, use a command that actually trips `shlex` — a heredoc body with a lone apostrophe:

```python
def test_unparseable_heredoc_classified_as_pass(self):
    cmd = "python3 << 'EOF'\necho it's fine\nEOF"
    assert _classify_command(cmd) == "pass"
```

That's `block` on `main` and `pass` here, so it fails if the fix ever regresses. For the high-risk counterpart, pair an *unparseable* heredoc with a high-risk token (e.g. a trailing `cat /etc/shadow`) so it proves the guard still fires through the new `except` path rather than only via Pass 1.

Logic ships; just want the tests to lock it in.

Huixin615:
@fancyboi999 Thanks for your review! I have updated it, could you please take a review again when you have time? Thanks!

fancyboi999 (approved):
Re-reviewed — the updated tests now anchor the fix. Thanks for the quick turnaround.

`test_unparseable_heredoc_classified_as_pass` uses `python3 << 'EOF'` / `echo it's fine` / `EOF`, whose lone apostrophe makes `shlex.split()` raise `ValueError: No closing quotation` — so it genuinely drives the changed `except` branch. I verified by reverting only `sandbox_audit_middleware.py` to `main`: that test now fails there (`block` != `pass`) and passes on this branch, so it's a real regression guard rather than one that's green either way.

`test_unparseable_heredoc_with_high_risk_pattern_still_blocks` still blocks (via the Pass-1 high-risk scan on `cat /etc/shadow`), confirming the fall-through doesn't weaken the high-risk guard. Full file: 149 passed on this branch.

Production change is correct and the tests now lock it in. No further issues from me.

WillemJiang (approved):

PR #3790, opened 2026-06-25
fix(models): normalize api_base->base_url for ChatOpenAI + warn on unknown config keys

### LINKED ISSUE #3789: [Bug] ChatOpenAI models silently break when config uses `api_base` instead of `base_url`
# [Bug] ChatOpenAI models silently break when config uses `api_base` instead of `base_url`

## Describe the bug

A model configured with `use: langchain_openai:ChatOpenAI` and an `api_base:` key (instead of `base_url:`) fails at **request time** with an opaque error, even though config validation passes and the model builds successfully:

```
LLM request failed: AsyncCompletions.create() got an unexpected keyword argument 'api_base'
```

This is easy to hit because `config.example.yaml` itself uses `api_base:` for *other* model classes — e.g. the Volcengine/Doubao example (`api_base: https://ark.cn-beijing.volces.com/api/v3`) and the Moonshot/Kimi example (`api_base: https://api.moonshot.cn/v1`). A user adapting one of those examples to an OpenAI-compatible gateway naturally copies `api_base`, but `langchain_openai:ChatOpenAI` only accepts `base_url` (with `openai_api_base` as a legacy alias).

## Root cause

`ModelConfig` is declared with `model_config = ConfigDict(extra="allow")`, so unknown keys like `api_base` are accepted at config-load time and flow into `model_settings_from_config`. `factory.py` then does `model_class(**kwargs, **model_settings_from_config)`. LangChain's `ChatOpenAI` does **not** reject the unknown kwarg — it emits a `UserWarning` ("api_base is not default parameter ... transferred to model_kwargs") and stores it in `model_kwargs`, which is subsequently spread into every `Completions.create()` call. The OpenAI SDK then rejects it at request time. Two consequences:

1. The error surfaces far from its cause (request time, not config load), with a message that doesn't mention the config key.
2. The intended endpoint override is silently dropped (`base_url` stays unset), so the model quietly targets the default OpenAI endpoint instead of the configured gateway.

## To Reproduce

```yaml
models:
  - name: my-gateway-model
    use: langchain_openai:ChatOpenAI
    model: claude-haiku-4-5
    api_base: http://host.docker.internal:4001/v1   # <-- should be base_url
    api_key: $MY_KEY
```
Build the model and send any message → request-time crash above.

## Expected behavior

Either the model should work (treat `api_base` as the obvious alias of `base_url`, which is what the user means), or DeerFlow should surface a clear, actionable error/warning at build time pointing at the config key — not a cryptic OpenAI-SDK error at request time.

## Environment
- DeerFlow 2.0 (current `main`)
- `langchain_openai:ChatOpenAI` model class, OpenAI-compatible gateway (LiteLLM → AWS Bedrock)

## Proposed fix
I have a fix ready (normalize `api_base`→`base_url` for the ChatOpenAI family + a build-time warning for unrecognized config keys, with tests). Happy to open a PR — see below.

praxstack:
Fixes #3789

## Why

**Trigger:** Configuring a `langchain_openai:ChatOpenAI` model with an `api_base:` key (instead of `base_url:`) crashes at request time with `AsyncCompletions.create() got an unexpected keyword argument 'api_base'`, even though config load and model construction both succeed.

**Pain:** This is easy to hit because `config.example.yaml` itself uses `api_base:` for other model classes (the Volcengine/Doubao and Moonshot/Kimi examples), so users copying those onto an OpenAI-compatible gateway naturally reach for `api_base`. The failure surfaces far from its cause and the error never mentions the config key. Worse, the endpoint override is silently dropped (`base_url` stays unset), so the model quietly targets the default OpenAI endpoint instead of the configured gateway.

Root cause: `ModelConfig` is `extra="allow"`, so `api_base` flows into `model_settings_from_config` and is splatted into the constructor. `ChatOpenAI` doesn't reject it — it transfers it into `model_kwargs` (with a `UserWarning`), which is then spread into every `Completions.create()` call and rejected by the OpenAI SDK at request time.

## What changed

Two small helpers in `models/factory.py`, mirroring the existing `_enable_stream_usage_by_default` / `_apply_stream_chunk_timeout_default` pattern:

- **`_normalize_openai_base_url`** — for the OpenAI-compatible family (`langchain_openai:ChatOpenAI` and the in-repo `PatchedChatOpenAI`), renames `api_base` → `base_url`. If an endpoint key (`base_url` or `openai_api_base`) is already present, the alias is dropped with a warning. Runs before the `stream_usage` / `stream_chunk_timeout` heuristics so they see the canonical key. Non-OpenAI classes (e.g. `PatchedChatDeepSeek`, which legitimately takes `api_base`) are left untouched.
- **`_warn_unknown_model_settings`** — **scoped to the same OpenAI-compatible family** (where the `model_kwargs` divert-and-crash actually happens and the field/alias set is accurate), logs an actionable warning at model-build time for config keys that aren't recognized fields/aliases of the resolved model class. Other providers route extra kwargs differently and are intentionally left alone (no false positives).
- The three OpenAI-compatible helpers now share a single `_OPENAI_COMPAT_USE_PATHS` tuple instead of disagreeing on the literal class string.

No behavior change for any currently-working config: a model that already uses `base_url` is unaffected, and the warning is silent on clean configs (verified against the real `ChatOpenAI` field/alias schema) and on non-OpenAI providers.

## Surface area

- [x] **Agents / LangGraph** — model factory (`backend/packages/harness/deerflow/models/factory.py`)
- [x] **Docs / tests / CI only** — adds a note to `backend/docs/CONFIGURATION.md` and tests; no change to existing public behavior

## Bug fix verification

- Test path: `backend/tests/test_model_factory.py` — `test_api_base_normalized_to_base_url_for_chatopenai`, `test_api_base_normalized_for_patched_chatopenai`, `test_no_unknown_key_warning_for_non_openai_class`.
- Red on `main` / green on this branch: the normalization tests fail on `main` (the key is not renamed → `base_url` absent / `api_base` forwarded). They pass on this branch.

## Validation

```
cd backend && make lint          # ruff check: All checks passed; ruff format: clean
cd backend && make test          # full suite green; test_model_factory.py: 52 passed (43 baseline + 9 new), 0 regressions
                                 # broader model/config/resolution/tracing suites: 77 passed, 0 regressions
```
9 new tests cover: `api_base`→`base_url` normalization for ChatOpenAI and PatchedChatOpenAI, both-set precedence (for both `base_url` and the `openai_api_base` field name), non-OpenAI class left untouched, no-op when no endpoint key is set, the unknown-key warning firing on a typo, staying silent on a clean config, and staying silent for a non-OpenAI provider (`ChatAnthropic` + `frequency_penalty`).

## AI assistance

**Tool(s) used:** Hermes (Claude-based agent)

**How you used it:** The bug was hit in real use (an OpenAI-compatible gateway → AWS Bedrock). The agent reproduced it, root-caused it to the `extra="allow"` + `model_kwargs` transfer path, implemented the fix mirroring the file's existing helper pattern, and wrote the tests TDD-style. The change went through two rounds of adversarial code review (which caught and fixed a false-positive in the warning helper for non-OpenAI providers). I reviewed every line and take responsibility for it.

### DIFF
--- a/backend/docs/CONFIGURATION.md
+++ b/backend/docs/CONFIGURATION.md
@@ -81,6 +81,8 @@ models:
 
 For OpenAI-compatible gateways (for example Novita or OpenRouter), keep using `langchain_openai:ChatOpenAI` and set `base_url`:
 
+> **Note:** for `langchain_openai:ChatOpenAI` the endpoint override key is `base_url` (not `api_base`). If you write `api_base` it is automatically normalized to `base_url`, and unrecognized keys are logged with a warning at model-build time. Some other model classes (e.g. `PatchedChatDeepSeek`) do use `api_base` — match the key to the class you configured.
+
 ```yaml
 models:
   - name: novita-deepseek-v3.2
--- a/backend/packages/harness/deerflow/models/factory.py
+++ b/backend/packages/harness/deerflow/models/factory.py
@@ -31,6 +31,14 @@ def _vllm_disable_chat_template_kwargs(chat_template_kwargs: dict) -> dict:
     return disable_kwargs
 
 
+# OpenAI-compatible model classes whose constructor takes ``base_url`` (not ``api_base``)
+# and to which the OpenAI-specific defaults below apply.
+_OPENAI_COMPAT_USE_PATHS = (
+    "langchain_openai:ChatOpenAI",
+    "deerflow.models.patched_openai:PatchedChatOpenAI",
+)
+
+
 def _enable_stream_usage_by_default(model_use_path: str, model_settings_from_config: dict) -> None:
     """Enable stream usage for OpenAI-compatible models unless explicitly configured.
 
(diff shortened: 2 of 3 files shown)


itxaiohanglover:
Excellent fix! Normalizing api_base to base_url for ChatOpenAI is important — the config.example.yaml inconsistency was a real user trap. The warning on unknown config keys will catch future similar issues. Docs update is thorough.

praxstack:
Hi @WillemJiang — gentle review request on this small fix. @itxaiohanglover already kindly looked it over ("Excellent fix") — thank you! 🙏

Status:
- **Scope:** normalize `api_base` → `base_url` for `ChatOpenAI` + warn on unknown config keys (the `config.example.yaml` inconsistency was a real user trap). Docs updated.
- **Freshly rebased onto latest `main`** so it's a clean single commit with no drift.
- **CLA signed; `make test` is green locally** (`backend/tests/test_model_factory.py` — 52 passed). The remaining CI checks are gated on the fork-PR **approve-and-run** gate (Lint / Unit / Frontend / E2E / Backend IO show `action_required`, 0 runs).

Could you approve the workflow run + take a review pass when you have a moment? Happy to adjust anything.

WillemJiang (approved):

PR #3798, opened 2026-06-25
fix: update subtask cards on parent completion (#3779)

### LINKED ISSUE #3779: 子任务卡片执行过程展示不完整、无持久化，无法回溯
### Before you start

### Problem summary

子任务卡片执行过程展示不完整、无持久化，无法回溯

### Affected area(s)

### What happened?

子任务卡片在展开状态下，只显示最新一条 AI 消息的工具调用信息，之前的步骤被覆盖。例如一个子任务依次执行了 `web_search` → `read_file` → `web_search` → 生成报告，页面上只能看到最后一步的状态，前面几步一闪而过，无法回溯。

用户只能看到：
- 任务描述（description）
- 当前最新一条工具调用的简要说明（如 "正在搜索..."）
- 最终结果

无法看到子智能体中间经历了哪些步骤、调用了哪些工具、每步的产出是什么。

### Expected behavior

期望可以看到子任务的执行步骤。

### Steps to reproduce

make docker-start

### Relevant logs

```shell
无
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

C1-BA-B1-F3:
Fixes #3779

## Problem
Subtask cards only displayed the latest AI message's tool call info. When a subtask executed multiple steps (e.g.  →  →  → generate report), previous steps were overwritten and the user could only see the last step's status.

## Root Cause
The  type stored only a single  field. Each  SSE event from the backend overwrote the previous message, so intermediate execution history was lost.

## Fix
Three targeted changes in the frontend:

1. **** — Added  to the  interface to accumulate all intermediate messages.

2. **** — Updated  to append each new  to  instead of letting it be overwritten by re-renders. The spread operator is now guarded to exclude  from the incoming partial update.

3. **** — Replaced the single-step rendering with a loop over , showing all tool calls from every intermediate step. Each tool call gets its own , with a spinner icon only on the latest step.

4. **** — Initialized  when creating new subtask objects.

### DIFF
--- a/backend/app/gateway/routers/skills.py
+++ b/backend/app/gateway/routers/skills.py
@@ -325,6 +325,7 @@ async def update_skill(skill_name: str, request: SkillUpdateRequest, config: App
         extensions_config.skills[skill_name] = SkillStateConfig(enabled=request.enabled)
 
         config_data = {
+            **{k: v for k, v in extensions_config.model_dump(by_alias=True, exclude_none=True).items() if k not in ("mcpServers", "skills")},
             "mcpServers": {name: server.model_dump() for name, server in extensions_config.mcp_servers.items()},
             "skills": {name: {"enabled": skill_config.enabled} for name, skill_config in extensions_config.skills.items()},
         }
--- a/backend/app/gateway/routers/skills.py
+++ b/backend/app/gateway/routers/skills.py
@@ -325,7 +325,6 @@ async def update_skill(skill_name: str, request: SkillUpdateRequest, config: App
         extensions_config.skills[skill_name] = SkillStateConfig(enabled=request.enabled)
 
         config_data = {
-            **{k: v for k, v in extensions_config.model_dump(by_alias=True, exclude_none=True).items() if k not in ("mcpServers", "skills")},
             "mcpServers": {name: server.model_dump() for name, server in extensions_config.mcp_servers.items()},
             "skills": {name: {"enabled": skill_config.enabled} for name, skill_config in extensions_config.skills.items()},
         }
--- a/backend/packages/harness/deerflow/client.py
+++ b/backend/packages/harness/deerflow/client.py
@@ -1039,6 +1039,7 @@ class DeerFlowClient:
         extensions_config.skills[name] = SkillStateConfig(enabled=enabled)
 
         config_data = {
+            **{k: v for k, v in extensions_config.model_dump(by_alias=True, exclude_none=True).items() if k not in ("mcpServers", "skills")},
             "mcpServers": {n: s.model_dump() for n, s in extensions_config.mcp_servers.items()},
             "skills": {n: {"enabled": sc.enabled} for n, sc in extensions_config.skills.items()},
         }
--- a/backend/packages/harness/deerflow/client.py
+++ b/backend/packages/harness/deerflow/client.py
@@ -1039,7 +1039,6 @@ class DeerFlowClient:
         extensions_config.skills[name] = SkillStateConfig(enabled=enabled)
 
         config_data = {
-            **{k: v for k, v in extensions_config.model_dump(by_alias=True, exclude_none=True).items() if k not in ("mcpServers", "skills")},
             "mcpServers": {n: s.model_dump() for n, s in extensions_config.mcp_servers.items()},
             "skills": {n: {"enabled": sc.enabled} for n, sc in extensions_config.skills.items()},
         }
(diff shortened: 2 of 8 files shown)


willem-bd:
Thanks for the subtask-card work. The history accumulation in `context.tsx` and the multi-step rendering in `subtask-card.tsx` look right. One scope issue to address before merge.

[P2] Unrelated backend changes bundled with the subtask feature
- Location: `backend/app/gateway/routers/skills.py`, `backend/packages/harness/deerflow/client.py`, `backend/packages/harness/deerflow/runtime/journal.py`
- Problem: The `update_skill` config-write change and the journal change are unrelated to subtask-card rendering and ship under this PR's title with no tests.
- Evidence: `skills.py` and `client.py` now spread `extensions_config.model_dump(by_alias=True, exclude_none=True)` (minus `mcpServers`/`skills`) into `config_data`, which changes the persisted `extensions_config.json` shape — extra top-level keys that were previously dropped on a skill toggle are now preserved. `journal.py` skips `middleware:` callers in the first-human-message detection. Neither touches subtask rendering.
- Suggested fix: Split the skills config-preservation fix into its own PR with a round-trip test that toggling a skill preserves unrelated extension-config keys; split the journal change separately.
- Test: assert non-`mcpServers`/`skills` keys survive an `update_skill` enable/disable round-trip.

xiawiie:
Followed up on the points from #3798 (review) and tried a draft on a branch off your latest commit (60f102a6). Two commits, four files, no backend changes:

https://github.com/xiawiie/deer-flow/tree/pr-3798-improvements

**1. fix(tasks): dedupe task_running messages on stream resume** (`205b9829`)

- `useStream` runs with `reconnectOnMount: true` + `streamResumable: true`, so the same `task_running` event is re-delivered after a reconnect / mid-run page reload. Each replay pushed another copy of the same message onto `messageHistory`.
- Resolves message identity in three steps before appending:
  1. `AIMessage.id` — stable across re-emits when present.
  2. backend `message_index` — 1-based monotonic, already emitted by `task_tool.py:382` but discarded by the current narrowing in `hooks.ts:815`.
  3. Reference equality — covers the rare same-session id-less repeat.
- Pipes `message_index` through `onCustomEvent` → `updateSubtask` → context, persisting it as `seenMessageIndices` on the task so dedup survives React re-renders.
- Also widens `Subtask.messageHistory` to optional (`?:`) to match the existing runtime guards (`previous?.messageHistory ?? []` in `context.tsx`; `task.messageHistory &&` in `subtask-card.tsx`).

**2. refactor(subtask-card): hoist messageHistory snapshot before map** (`a7ae1e83`)

- Reading `task.messageHistory.length - 1` inside the map callback relied on TS narrowing carrying through the `&&` chain. With `messageHistory` optional that no longer holds in strict mode. Hoists into a local `const` inside an IIFE; behaviour unchanged.

Same scope as your PR (frontend only, same four files plus an extra dedup field in `types.ts`).

What I did and did not verify:

- Read the LangGraph SDK source to confirm `streamResumable: true` makes `onDisconnect` default to `"continue"` and that custom events go through the same dispatcher on resume — the duplication is a static-analysis conclusion from that plus the unconditional append in `context.tsx:88`, not a recording of it happening live.
- `pnpm typecheck` passes for `src/`. `pnpm lint` passes (the two unused `_ignored*` warnings exist in your original PR too).
- `pnpm test` was not run — the local checkout has `@rstest/core` missing, unrelated to this change.

Feel free to cherry-pick or ignore. Happy to rebase if you'd like this on top of any follow-up commits.

C1-BA-B1-F3:
Split the backend config/journal changes out into #3807. This PR is now frontend-only (`message-list.tsx`, `subtask-card.tsx`, `context.tsx`, `types.ts`), so the original scope issue is resolved. Ready for another look.

C1-BA-B1-F3:
Thanks @xiawiie, I cherry-picked both frontend-only follow-up commits onto this PR:

- `070df1b8` fix(tasks): dedupe task_running messages on stream resume
- `28099982` refactor(subtask-card): hoist messageHistory snapshot before map

The branch now matches your suggested frontend-only improvements, with no backend scope added back. Checks are still clean, so this should be ready for maintainer review again.

C1-BA-B1-F3:
Hi @WillemJiang, could you please re-review this PR when you have a chance? The backend scope issue has been split into #3807, and this PR now contains only the frontend subtask-card changes. The follow-up frontend improvements suggested by @xiawiie were cherry-picked, and checks are clean.

xiawiie:
thanks for your help, wish you all well bro

ggnnggez:
addressed in #3845

PR #3800, opened 2026-06-25
fix(gateway): keep create_thread idempotent when the insert loses a race

he-yufeng:
## Why

While reading the thread gateway I noticed `POST /api/threads` does a check-then-insert: it calls `thread_store.get()` to honour idempotency, and if the row is absent it calls `thread_store.create()`. Those two steps aren't atomic. If two requests create the same `thread_id` at the same time (e.g. a channel adapter or orchestration layer retrying on a timeout with a client-chosen id), one commits first and the second's INSERT hits the duplicate primary key. The SQL-backed store raises, the handler treats it as a generic write failure, and the losing request gets a 500 — even though the endpoint's own docstring promises the call is idempotent ("returns the existing record when `thread_id` already exists").

## What changed

On a failed insert the handler now re-reads the row. If it's present (the competing request won the race), it returns that record, so concurrent creates of the same thread are idempotent from the caller's side instead of one of them 500-ing. A genuine write failure where the row is still absent keeps returning 500 as before. This mirrors how `channel_connections.upsert_connection` already treats a concurrent `IntegrityError` as "someone else committed it" rather than a hard error.

## Surface area

- [x] **Backend API** — `POST /api/threads` error path under `backend/app/gateway`

## Screenshots / Recording

Not applicable — backend only.

## Bug fix verification

- Test path that reproduces the bug: `backend/tests/test_threads_router.py::test_create_thread_returns_existing_when_insert_loses_race`
- Did it go red on `main` and green on this branch? **yes** — on `main` the endpoint returns 500 (the `IntegrityError` is caught as a generic failure); on this branch it returns the existing thread (200). The test drives the endpoint with a store whose first `create()` commits the row and then raises, simulating the lost race.

## Validation

Run from `backend/`:

- `uv run pytest tests/test_threads_router.py` — 25 passed (includes the new regression test)
- `uv run ruff check app/gateway/routers/threads.py tests/test_threads_router.py` — clean
- `uv run ruff format --check app/gateway/routers/threads.py tests/test_threads_router.py` — clean

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Used it to help locate the check-then-create window and draft the regression test; I reviewed the data flow, confirmed the SQL backend raises on the duplicate key while the memory backend overwrites, and decided on the re-read-on-conflict approach myself.


backend/app/gateway/routers/threads.py
```diff
@@ -289,6 +289,22 @@ async def create_thread(body: ThreadCreateRequest, request: Request) -> ThreadRe
             metadata=body.metadata,
         )
     except Exception:
+        # The idempotency ``get`` above and this insert are not atomic: a
+        # concurrent request for the same thread_id can commit in between,
+        # and the store then rejects ours on the duplicate primary key.
+        # Honour the documented idempotency contract by returning the
+        # now-existing record instead of surfacing the conflict as a 500.
+        existing_record = await thread_store.get(thread_id, **thread_owner_kwargs)
```
willem-bd: **This re-read runs under a broad `except Exception`, so any non-race failure that coincides with an existing row is silently swallowed as 200 — with no log.**

The comment scopes the fix to the duplicate-key `IntegrityError`, but the `except Exception:` above was already broad and the new re-read runs for *every* exception type. Failure scenario: `thread_store.create()` raises an `OperationalError` (connection dropped after the INSERT committed), or a different constraint/CHECK violation, while a pre-existing row for that `thread_id` is present (a prior legitimate create, or a client retry that already got a row). The handler now returns that row as 200 and, unlike before, **never logs the exception** — `logger.exception` only runs in the fall-through 500 branch, so genuine failures become invisible to ops.

The codebase's own canonical pattern catches `IntegrityError` specifically (`backend/app/gateway/auth/repositories/sqlite.py:74`). Narrowing here (catch `IntegrityError` only) or at minimum logging the swallowed exception at warning/debug would preserve observability. Minor aside: the PR says this mirrors `channel_connections.upsert_connection`, but that `IntegrityError`-as-"someone else committed" handling actually lives in `auth/repositories/sqlite.py`, and it catches `IntegrityError`, not `Exception`.

backend/app/gateway/routers/threads.py
```diff
@@ -289,6 +289,22 @@ async def create_thread(body: ThreadCreateRequest, request: Request) -> ThreadRe
...
+        if existing_record is None and thread_owner_user_id:
+            existing_record = await thread_store.get(thread_id, user_id=None)
```
willem-bd: **The error-path re-read skips the `update_owner` reconciliation the happy-path idempotency block performs — so ownership of the same thread diverges by timing.**

The initial idempotency block (~lines 267-273) claims an unscoped record for a trusted internal owner via `update_owner` when `unscoped_record.user_id != thread_owner_user_id`. This error-path block only does the unscoped `get` and returns directly — no `update_owner`.

Failure scenario: a channel adapter (trusted internal owner B) and an ownerless request race to create thread `X`; the ownerless request commits first, so B's INSERT loses and hits this error path, which returns the row still owned by `None`. Had B's request arrived slightly earlier and found the row via the happy path, the thread would have been re-assigned to B. The same logical "create `X` as owner B" therefore yields different ownership (and thus different access isolation) depending on insert ordering. If the omission is intentional it's worth a comment; otherwise the two paths should reconcile ownership identically. (Extracting the duplicated `ThreadResponse(...)` construction into a `_record_to_response(...)` helper would keep the two idempotency paths trivially identical and prevent this kind of drift.)

willem-bd (commented):
Review — the race fix is correct and the regression test is well-constructed. Two notes inline: the re-read runs under a broad `except Exception` (masking/logging concern) and the error path skips the owner reconciliation the happy-path idempotency block performs.

WillemJiang:
@he-yufeng Do you mind taking a look at the review comments?

he-yufeng:
Thanks @WillemJiang, and thanks to the reviewer for catching both of these; they were real gaps. Pushed 6c01960b:

1. **Broad `except` swallowing non-race failures** — the recovery is now scoped to `sqlalchemy.exc.IntegrityError` (the duplicate-key race the SQL store actually raises). Any other exception falls through to the existing `logger.exception` + 500, so a non-race failure that happens to coincide with an existing row is no longer returned as a 200. The memory store overwrites on duplicate rather than raising, so it never reaches this path (noted in a comment).

2. **Recovery skipped the `update_owner` reconciliation** — I pulled the fast-path idempotency resolution into a shared `_resolve_existing_thread` helper and call it from both the fast path and the recovery, so the recovery now claims a legacy unscoped (`user_id=None`) row for a trusted internal owner exactly as the fast path does. Ownership no longer depends on which path resolved the record.

Added two regression tests: one drives the insert-race interleaving so the recovery path (not the fast path) is the one that claims the unscoped row for a trusted owner, and one asserts a non-`IntegrityError` failure surfaces as a 500 instead of being swallowed. Full `test_threads_router.py` suite is green.

WillemJiang (approved):

PR #3805, opened 2026-06-26
fix: add heading anchors to markdown artifact previews

### LINKED ISSUE #3803: 产物预览区域，md文档，点击章节无法跳转到对用章节
### Before you start

### Problem summary

产物预览区域，md文档，点击章节无法跳转到对用章节

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

UI页面上，markdown文档产物预览区域，点击章节目录，无法跳转到对应章节

### Expected behavior

UI页面上，markdown文档产物预览区域，点击章节目录，可以跳转到对应章节

### Steps to reproduce

make docker-start

### Relevant logs

```shell
null
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

Huixin615:
Fixes #3803

## Why

Markdown artifact previews did not generate `id` attributes for headings, so table-of-contents links like `[概述](#概述)` rendered as normal anchors but could not scroll to the target section.

This affected users viewing generated `.md` artifacts in the preview panel: clicking a chapter link did nothing because the corresponding heading had no anchor target.

## What changed

- Markdown artifact previews now generate GitHub-style heading anchors via `rehype-slug`.
- Existing same-page artifact links continue to use the current `ArtifactLink` behavior.
- Added regression coverage for Chinese heading anchors and markdown artifact preview anchor navigation.

## Surface area

- [x] **Dependencies** — adds `rehype-slug` to generate heading ids for markdown preview content
- [x] **Default behavior change** — markdown artifact headings now include generated `id` attributes by default

## Screenshots / Recording

image
image
✅

## Bug fix verification

  - `frontend/tests/unit/core/streamdown-plugins.test.ts`
  - `frontend/tests/e2e/artifact-preview.spec.ts`
  - Not verified by switching branches locally.
  - The unit regression asserts that markdown rendered through the artifact Streamdown plugin path generates `id="概述"` for `## 概述`; this is the missing behavior on `main`.
  - A focused unit test was added for the renderer/plugin behavior.
  - An E2E regression was also added for clicking a markdown artifact TOC link and checking the target heading scrolls into view.

## Validation

Ran:

- `cd frontend && pnpm lint`
  - Passed with existing unrelated warnings.
- `cd frontend && pnpm test`
  - Passed: 354/354.
- `git diff --check`


Huixin615:
@fancyboi999 辛苦帮忙review

willem-bd (commented):
Thanks @Huixin615. I found one issue that should be addressed before this is ready — the fix is correct for artifacts but the slug plugin lands on the shared config, which widens the blast radius.

frontend/src/core/streamdown/plugins.ts
```diff
@@ -19,6 +20,7 @@ export const streamdownPlugins = {
   ] as StreamdownProps["remarkPlugins"],
   rehypePlugins: [
     rehypeRaw,
+    rehypeSlug,
```
willem-bd: [P2] `rehype-slug` is added to the shared plugin config rather than just artifact previews, which produces duplicate heading `id`s across messages.

- **Problem:** `streamdownPlugins` is the shared config consumed by every chat message (`markdown-content.tsx`), reasoning content (`reasoning.tsx`, via `reasoningPlugins` which derives from this array), `memory-settings-page.tsx`, and `about-settings-page.tsx` — not just the artifact preview. The PR title/description scope this to "markdown artifact previews", but the change is global. `rehype-slug` v6 keeps a single module-level `GitHubSlugger` and calls `slugs.reset()` at the start of each transform, so uniqueness is only guaranteed within one render. Two assistant messages that both contain a heading with the same text (e.g. `## Summary`) render two `id="summary"` elements into the same DOM — invalid HTML, and any in-page `#summary` link could scroll to the first match instead of the heading in the message the user clicked.
- **Evidence:** `streamdownPlugins` is imported in `frontend/src/components/workspace/messages/markdown-content.tsx:12`; `reasoningPlugins` derives its `rehypePlugins` from `streamdownPlugins.rehypePlugins` (`frontend/src/core/streamdown/plugins.ts`).
- **Suggested fix:** scope `rehypeSlug` to the artifact render path (`frontend/src/components/workspace/artifacts/artifact-file-detail.tsx`, which already spreads `{...streamdownPlugins}`) — the only place TOC anchor navigation is needed — instead of adding it to the shared config.
- **Test:** a unit test that renders two separate messages each containing the same heading text and asserts the resulting ids differ (or that ids are namespaced per render).

Huixin615:
@willem-bd Thanks for your suggestions! I have updated the PR. Please take another look when you have time, thanks a lot!

WillemJiang (approved):

PR #3812, opened 2026-06-26
fix(mcp): route tools by source server, not name prefix

### LINKED ISSUE #3811: bug: MCP tools mis-routed when one server name is a prefix of another
## Describe the bug

## Root cause

## Expected

AnnaSuSu:
Fixes #3811

## Why

Configuring two MCP servers where one name is a prefix of another (e.g. `web` and `web_scraper`, or `github` and `github_enterprise`) silently mis-routes the longer-named server's tools. `get_mcp_tools()` re-derives each tool's source server by scanning `servers_config` for a name that prefixes the (already prefixed) tool name and taking the **first** match, so `"web_scraper_search".startswith("web_")` matches `web` first. The tool is then pooled under `web` and invoked with the wrong stripped name (`scraper_search`), so the call fails with "tool not found"; for mixed transports a stdio tool from the longer-named server can also lose session pooling. It is order-dependent on config insertion order, so it can look intermittent.

## What changed

MCP tools are now routed to the server that actually produced them — the per-server grouping `tools_by_server` (index `i` ↔ the `i`-th server in `servers_config`) — instead of by ambiguous name-prefix inference. No config change; servers whose names share a prefix now work correctly. A prefix guard is kept so unprefixed tools still fall through unwrapped (behavior unchanged for them).

## What changed (caller's perspective)

A workspace with MCP servers whose names overlap by prefix now invokes each server's tools correctly, instead of the longer-named server's tools failing with "tool not found".

## Surface area

- [x] **Agents / LangGraph** — MCP tool loading/routing (`backend/packages/harness/deerflow/mcp/tools.py`)

## Bug fix verification

- Test path: `backend/tests/test_mcp_session_pool.py::test_mcp_tools_routed_to_source_server_with_prefix_overlap`
- Did it go red on `main` and green on this branch? **yes** — red on `main` with `AssertionError: tool mis-routed to 'web', expected 'web_scraper'`, green on this branch.

## Validation

- `cd backend && make lint` — ruff check + format clean.
- `cd backend && make test` — **5179 passed, 19 skipped**. One unrelated timing test (`test_gateway_lifespan_shutdown::test_shutdown_is_bounded_when_channel_stop_hangs`) flaked under the parallel run; it passes in isolation on both this branch and `main`.
- MCP suites (`test_mcp_session_pool.py`, `test_mcp_sync_wrapper.py`, `test_mcp_custom_interceptors.py`): 54 passed.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Claude Code located the bug, wrote the fix and the regression test. The contributor reviewed the diff and verified the red/green test and the full backend suite locally.


itxaiohanglover:
Good fix! The prefix collision is a classic bug — first-match prefix is wrong when one server name is a prefix of another. Routing by source server directly instead of re-deriving from name is the right approach.

AnnaSuSu:
Rebased onto latest main and resolved the conflict with #3843 (per-server tool_call_timeout): routing now goes through the source-server grouping, and tool_call_timeout is still passed into the stdio session-pool wrapper.

MCP suites are green locally (test_mcp_session_pool, test_mcp_client_config, test_mcp_custom_interceptors: 63 passed) and ruff is clean. Ready for another look — thanks!

WillemJiang (approved):

PR #3822, opened 2026-06-26
fix(gateway): fix oversized upload replacements deleting existing files

fallintoplace:
## Summary
- stage upload bytes in a temp file inside the uploads directory
- atomically replace the destination only after all size checks pass
- add a regression test proving a 413 replacement keeps the original file bytes intact

## Root cause
The streaming upload path opened and truncated the destination before upload size validation finished. When the replacement later failed with 413, cleanup deleted the path that had been holding the original valid file.

## Impact
Failed replacement uploads no longer destroy an existing upload with the same name. Successful replacements keep the current overwrite behavior.

## Validation
- `uv run pytest tests/test_uploads_router.py tests/test_uploads_manager.py`
- `uv run ruff check app/gateway/routers/uploads.py packages/harness/deerflow/uploads/manager.py tests/test_uploads_router.py`


willem-bd (commented):
Thanks @fallintoplace. The atomic temp-file → `os.replace` approach is the right fix, and the regression test cleanly proves the original bytes survive the 413. Two non-blocking hardening observations for a follow-up if you're interested — neither blocks this PR.

**1. Transient `.part` file is visible during upload**
`tempfile.mkstemp(prefix=".upload-", suffix=".part", dir=uploads_dir)` places the staging file inside the uploads directory for the duration of the upload. `list_files_in_dir` (and any `ls` in a thread-data-mount/local sandbox, where the uploads dir is mounted directly at `/mnt/user-data/uploads`) doesn't filter dotfiles, so a concurrent read can observe the partial `.upload-*.part`. Low impact in practice — uploads finish before the agent turn, the name is dot-prefixed, and it's cleaned on every normal failure path — but filtering `*.part` (or `.upload-*`) out of `list_files_in_dir` would close the window entirely.

**2. Orphaned temp on hard crash**
If the Gateway process is killed mid-upload, the `.upload-*.part` file is orphaned and nothing sweeps stale `.part` files at startup. This is strictly better than the pre-fix behavior (which truncated/corrupted the destination), so it's not a regression — just noting that a startup sweep of stale `.upload-*.part` files would prevent accumulation across crashes over time.

WillemJiang (approved):

PR #3823, opened 2026-06-26
fix(skills): handle failed settings API responses

fallintoplace:
## Summary
- reject non-OK skills list and enable/disable responses instead of treating error JSON as success
- show the backend error message when a skill toggle is rejected in Settings
- add unit coverage for the skills API helper and a Playwright regression for rejected toggles

## Root cause
The frontend skills helpers parsed JSON without consistently checking `response.ok`. A 4xx or 5xx response could resolve as if the request succeeded, so the settings mutation path treated failed enable/disable requests like successful ones.

## Impact
Skill enable/disable failures now stay on the error path, surface a visible failure to the user, and keep the UI from pretending the change succeeded.

## Validation
- `pnpm exec eslint src/core/skills/api.ts src/components/workspace/settings/skill-settings-page.tsx tests/unit/core/skills/api.test.ts tests/e2e/skills-settings.spec.ts`
- `pnpm test tests/unit/core/skills/api.test.ts`
- `pnpm exec playwright test tests/e2e/skills-settings.spec.ts`


WillemJiang:
@fallintoplace Thanks for your contribution. Please resolve the conflicts with main.

WillemJiang:
@fallintoplace Please resolve the conflicts of the main branch, so we can move on the review process.

fallintoplace:
@WillemJiang I will push the change in a few hours.

fallintoplace:
@WillemJiang Can you please check again?

frontend/tests/e2e/skills-settings.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  await sidebar.getByRole("button", { name: /Settings and more/ }).click();
+  await page.getByRole("menuitem", { name: "Settings" }).click();
+  const dialog = page.getByRole("dialog", { name: "Settings" });
+  await expect(dialog).toBeVisible();
+  await dialog.getByRole("button", { name: "Skills" }).click();
+  return dialog;
+}
+
+test.describe("Skill settings", () => {
+  test("shows a failure and keeps the toggle state when enabling a skill is rejected", async ({
```
willem-bd: Minor: the test name says "enabling a skill is rejected", but the mock skill starts with `enabled: true` and clicking a checked Switch fires `onCheckedChange(false)` -- so the request being rejected here is actually a disable (PUT `{ enabled: false }`), not an enable. Consider rewording to "disabling" (or seeding the skill with `enabled: false` and clicking to enable) so the description matches what the test exercises. Behavior assertions are correct either way.

willem-bd (commented):
Reviewed the consumer-side error surfacing for skill toggles. Switching `mutate` -> `mutateAsync` with a `.catch` toast correctly exposes failed toggles that were previously swallowed silently. Because the `Switch` is controlled by `checked={skill.enabled}` (cache-bound) and `useEnableSkill` only invalidates the skills query on success, the toggle stays visually stable on rejection, which the new e2e test asserts. Unit coverage for both happy and rejection paths in `api.test.ts` looks solid. One minor note on a test description below.

WillemJiang (approved):

PR #3824, opened 2026-06-26
fix(sandbox): push read_file ranges into sandbox reads

fallintoplace:
## Summary
- push bounded and open-ended `start_line` and `end_line` arguments into sandbox reads
- stream only the requested window in `LocalSandbox` and forward native range arguments through the AIO client
- keep E2B and BoxLite compatible with the widened interface and add provider-level regressions

## Root cause
`read_file` previously asked the sandbox for the complete file and applied line selection afterward. A small excerpt from a large text file could therefore allocate the entire file before truncation, and widening the interface required every sandbox implementation to accept optional range arguments consistently.

## Impact
Local and AIO ranged reads now avoid fetching lines outside the requested 1-indexed, inclusive window. All current sandbox providers support bounded and open-ended calls without unexpected keyword errors; E2B and BoxLite apply ranges after their underlying full-file reads because those client paths do not expose equivalent ranged-read APIs.

## Validation
- `uv run ruff check packages/harness/deerflow/sandbox/sandbox.py packages/harness/deerflow/sandbox/local/local_sandbox.py packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py packages/harness/deerflow/community/boxlite/box.py packages/harness/deerflow/sandbox/tools.py tests/test_read_file_tool_binary.py tests/test_read_file_line_range.py tests/test_aio_sandbox.py tests/test_local_sandbox_provider_mounts.py tests/test_e2b_sandbox_provider.py tests/test_boxlite_provider.py`
- `uv run pytest tests/test_read_file_tool_binary.py tests/test_read_file_line_range.py tests/test_aio_sandbox.py tests/test_local_sandbox_provider_mounts.py tests/test_e2b_sandbox_provider.py tests/test_boxlite_provider.py`


WillemJiang:
@fallintoplace  Thanks for your contribution. Please take a look at the lint error.

fallintoplace:
@WillemJiang Please check again, I have amended the PR.

WillemJiang:
@fallintoplace please take a look at the review comments and resolve the conflict with the main branch.

backend/packages/harness/deerflow/sandbox/sandbox.py
```diff
@@ -44,11 +44,18 @@ def execute_command(
         pass
 
     @abstractmethod
-    def read_file(self, path: str) -> str:
+    def read_file(
```
willem-bd, line 94: Heads-up for merge: widening the abstract `read_file` contract means every `Sandbox` subclass must accept `start_line`/`end_line`, because `read_file_tool` now calls `sandbox.read_file(path, start_line=..., end_line=...)` unconditionally on the bounded-read branch (no sandbox-type guard).

This branch updates Local, AIO, and E2B — but `boxlite/box.py:152` (added to `main` in #3940, after this branch's base) still has `def read_file(self, path: str)`. Once this PR merges, `BoxliteBox` will raise `TypeError: read_file() got an unexpected keyword argument 'start_line'` on every ranged read; it's caught by the generic `except Exception` and returned as an error string, so the ranged-read feature is silently broken for any BoxLite deployment. Full reads keep working via the kwargs-free `read_current_file_content` path.

Cheapest fix: add `start_line`/`end_line` (defaulting `None`) to `BoxliteBox.read_file` and either forward or slice `r.stdout`, plus a one-line unit test on a stub. (Not anchored to `box.py` because that file isn't in this branch yet.)

backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py
```diff
@@ -202,13 +202,23 @@ def ping(self) -> bool:
...
         resolved = self._resolve_path(path)
         try:
             content = self._client.files.read(resolved)
-            if isinstance(content, bytes):
-                return content.decode("utf-8", errors="replace")
-            return content if content is not None else ""
+            if start_line is None and end_line is None:
+                return content.decode("utf-8", errors="replace") if isinstance(content, bytes) else content or ""
+            text = content.decode("utf-8", errors="replace") if isinstance(content, bytes) else content or ""
+            lines = text.splitlines()
```
willem-bd, line 220: The new e2b ranged-read branch ships without test coverage. The other three touched paths all get range tests (`test_local_sandbox_provider_mounts.py`, `test_aio_sandbox.py::TestReadFile`, `test_read_file_tool_binary.py`), but `test_e2b_sandbox_provider.py` has no `read_file` / `start_line` / `end_line` coverage. `backend/AGENTS.md` is explicit: "Every new feature or bug fix MUST be accompanied by unit tests. No exceptions."

Worth pinning a few observable edge cases:
- the `isinstance(content, bytes)` decode decision is duplicated across the no-range and range paths (easy to diverge on a later refactor);
- `start_line=None` is silently coerced to `1` via `start_line or 1`;
- a `start_line` past EOF returns `""`, indistinguishable from an empty file.

(Side note: unlike Local/AIO, this path still calls `self._client.files.read(resolved)` for the whole file before slicing — fine if the e2b client has no ranged API, just noting the "pushdown" memory win doesn't extend here.)

WillemJiang:
@fallintoplace, please fix the lint error and add the implementation of BoxliteBox at same time.

fallintoplace:
@WillemJiang Can you please check again?

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -2128,29 +2128,42 @@ def read_file_tool(
...
+        start_line: Optional starting line number (1-indexed, inclusive). Omit to start at the first line.
+        end_line: Optional ending line number (1-indexed, inclusive). Omit to read through the last line.
     """
     try:
         # Block access to disabled skill files
         if _is_disabled_skill_path(path, user_id=resolve_runtime_user_id(runtime)):
             skill_name = _extract_skill_name_from_skills_path(path) or "unknown"
             return f"Error: Skill '{skill_name}' is disabled. Access to its files is blocked. Enable the skill in settings before using it."
+        effective_start = max(start_line or 1, 1)
+        if end_line is not None and end_line < 1:
```
willem-bd, line 2142: Minor asymmetry: this rejects `end_line < 1` but the `start_line < 1` case is only coerced inside `effective_start` for the comparison — the original `start_line` is still forwarded to the sandbox unchanged.

The four providers then diverge on a negative `start_line`:
- `LocalSandbox` (`local_sandbox.py:694`): `max(start_line or 1, 1)` — clamps to 1.
- `AioSandbox` (`aio_sandbox.py:301`): `max(start_line - 1, 0)` — clamps to 0.
- `E2BSandbox` (`e2b_sandbox.py:221`) and `BoxliteBox` (`box.py:226`): `start = start_line or 1`. Because `-5 or 1` evaluates to `-5` (any non-zero int is truthy), `lines[start - 1:end]` becomes `lines[-6:end]` — i.e. the *tail* of the file rather than line 1 onward. LocalSandbox/AIO happen to be safe because they re-clamp, but E2B/BoxLite silently return the wrong window.

Cheapest fix is to mirror the `end_line` check here: `if start_line is not None and start_line < 1: return "(start_line must be >= 1)"`. That matches the 1-indexed contract in the docstring above and lets the provider implementations stay as-is. (Noticed while re-checking the prior review thread — the old `max(start_line, 1)` post-slice in tools.py used to clamp this, but the pushdown moved the clamp into the providers and only two of four picked it up.)

willem-bd (commented):
Second-look review after the boxlite signature widening and the new e2b/boxlite provider tests — both of my prior concerns are resolved (boxlite now carries the kwargs; e2b has bounded-range coverage). One remaining gap worth a small follow-up: `start_line < 1` is not validated, and the four providers diverge on how they handle a negative `start_line`.

WillemJiang (approved):

PR #3826, opened 2026-06-27
fix(frontend): preserve messages across context summarization (#3825)

### LINKED ISSUE #3825: fix(frontend): messages lost after context summarization due to stale history race in onUpdateEvent
## Bug Description

After context summarization (`SummarizationMiddleware`) triggers during a long multi-round conversation, the frontend loses previously displayed messages:
- Previous round's assistant output disappears
- Current round's user input disappears

## Root Cause

In `frontend/src/core/threads/hooks.ts`, the `onUpdateEvent` handler (~L755) has a race condition:

```typescript
// L755-756 in onUpdateEvent
appendMessages(_movedMessages);  // async React setState — doesn't take effect until next render
messagesRef.current = [];
```

Then in the same render cycle, `mergeMessages` is called:

```typescript
const mergedMessages = mergeMessages(
  visibleHistory,      // ← STALE: appendMessages hasn't updated this yet
  persistedMessages,
  visibleOptimisticMessages,
);
```

**Sequence:**
1. Backend summarization: `RemoveMessage(ALL)` → add summary (hidden) + preserved messages
2. Frontend `onUpdateEvent`: computes `_movedMessages`, calls `appendMessages(_movedMessages)` to archive
3. `appendMessages` uses React's `setMessages` — **async, not effective until next render**
4. Same render: `mergeMessages` uses **stale** `visibleHistory` (missing `_movedMessages`)
5. Result: old messages not in `persistedMessages` (removed by `REMOVE_ALL`) AND not in `visibleHistory` (state update pending) → **lost**

## Impact

- Any long-running conversation that triggers context summarization
- Messages flash and disappear; user sees a nearly empty conversation
- Reproducible on latest `main` (7a6c4a99)

## Suggested Fix

Use a synchronous snapshot ref instead of relying on async `visibleHistory`:

```typescript
const historyRef = useRef<Message[]>([]);
const historyMergeRef = useRef<Message[] | null>(null);

useEffect(() => {
  historyRef.current = visibleHistory;
}, [visibleHistory]);

// In onUpdateEvent:
historyMergeRef.current = [...historyRef.current, ..._movedMessages];
appendMessages(_movedMessages);
messagesRef.current = [];

// In merge call:
const effectiveHistory = historyMergeRef.current ?? visibleHistory;
if (historyMergeRef.current) historyMergeRef.current = null;
const mergedMessages = mergeMessages(effectiveHistory, persistedMessages, visibleOptimisticMessages);
```

This ensures the merge uses the complete history without waiting for React's async state update.

fancyboi999:
## Summary

Fixes #3825. After `SummarizationMiddleware` runs during a long conversation, the merged conversation view could drop already-displayed messages (previous assistant output, current user input), leaving a nearly-empty thread.

## Root cause (first-principles)

The displayed conversation is computed in `frontend/src/core/threads/hooks.ts` by:

```ts
mergeMessages(visibleHistory, persistedMessages, optimistic)
```

These two inputs come from **two independent state systems**:

- `persistedMessages` ← `thread.messages`, owned by the LangGraph SDK `useStream` and exposed through an **external store** (`useSyncExternalStore`).
- `visibleHistory` ← `appendedMessages`, owned by `useThreadHistory` as **React `useState`**.

On summarization the backend emits `RemoveMessage(ALL)` + a hidden summary + the retained tail. `onUpdateEvent` rescues the about-to-be-removed messages by calling `appendMessages(...)` — an **async** `setAppendedMessages`.

The defect is that the merge's correctness depends on the external-store shrink and the React-state archival landing in the **same render**. They are scheduled independently, so there exists a render where `persistedMessages` has already shrunk but `visibleHistory` has not yet absorbed the rescued messages. In that render the rescued messages are present in **neither** merge input and are dropped.

The original issue located the right spot but framed it as "stale within the same render cycle as `onUpdateEvent`" (`onUpdateEvent` is a stream callback, not a render). The precise framing is **cross-store update ordering** between the SDK external store and the history React state.

## Fix

Bridge the async gap with a **synchronous buffer** (`pendingArchivedMessagesRef`) so the merge no longer depends on how the two channels interleave:

- Written the moment `onUpdateEvent` derives the moved messages (synchronous), and read by the merge on **every** render via `resolvePreservedHistory`.
- `resolvePreservedHistory` only appends messages still **missing** from history — once history absorbs a rescued message its live copy stays authoritative (the buffered copy is an older snapshot), and ordering is preserved.
- Drained via `pruneConfirmedArchivedMessages` once history confirms absorption, so the buffer is a transient bridge rather than a second long-lived source of truth.
- Tagged with the thread it was captured from; the merge overlays it **only when that matches the viewed `threadId`** (the same prop `visibleHistory` is gated on), so it can never leak into another thread or the new-chat screen. This is a read-only check at merge time — no render-phase ref mutation, so it is concurrent-mode safe.

The moved-message derivation and the merge overlay are extracted into pure, unit-tested helpers: `computeSummarizationMovedMessages`, `resolvePreservedHistory`, `pruneConfirmedArchivedMessages`.

> Note: I deliberately did **not** adopt the issue's suggested `historyMergeRef` snapshot. It is consumed on first read (`= null` after one merge), so any additional render before `setAppendedMessages` commits re-reads the stale `visibleHistory` and the bug resurfaces; it also keeps a `historyRef` synced one commit behind via `useEffect`. The buffer here survives multiple intermediate renders and self-heals.

## Goal contract

- **End state:** rescued history messages never disappear from the merged view, regardless of how the two state channels interleave.
- **Evidence:** a regression test that fails on `main` and passes here (below).
- **Constraints:** no permanent second source of truth; `mergeMessages` semantics unchanged; backend/SDK untouched; surgical diff.

## Verification

```
pnpm test       → 367 passed, 0 failed, 0 skipped (36 files)
pnpm typecheck  → clean
pnpm lint       → 0 errors, 3 warnings (all pre-existing, in files this PR does not touch)
pnpm build      → exit 0
```

New deterministic regression coverage in `tests/unit/core/threads/message-merge.test.ts`, including the full rescue pipeline (`computeSummarizationMovedMessages` → buffer → `resolvePreservedHistory` → `mergeMessages`) asserting no loss while history state lags, and that a live history copy wins over a stale buffered duplicate.

An independent reviewer (clean context, did not write the code) ran the suite, a 5-way "vaccine test", and adversarial edge-case analysis (multiple summarizations, thread-switch mid-summarization, React 19 strict-mode double-render idempotency, regenerate/superseded interaction). Verdict: fix is sound; the one rejected suggestion (a superseded message stuck in the buffer) was confirmed **unreachable** because regenerate is only enabled on the latest assistant turn (`enableRegenerateForTurn = group.id === latestAssistantGroupId`) and moved messages are never the latest turn.

## Fail-loud: what is NOT covered

- **Hook runtime-wiring binding** (the single line where the hook calls `resolvePreservedHistory` with the buffer ref) and the real React/SDK render timing are **not** covered by an automated test. This project intentionally keeps unit tests as pure-function tests (no `jsdom`/`@testing-library`) and puts browser-level checks in Playwright E2E; adding a hook-render harness is a separate infra decision. This binding is verified by code review.
- **No live end-to-end summarization reproduction.** Triggering `SummarizationMiddleware` requires a very long real conversation; the E2E mock delivers the stream in one batch and so cannot deterministically reproduce the render-timing race (it would pass on buggy code too), which is why the deterministic guard lives at the pure-function/algorithm level.
- **No before/after screenshots.** The change has no visible effect in normal operation — it only prevents loss in the hard-to-trigger summarization window — so there is nothing meaningful to capture visually.

## Scope

`frontend/src/core/threads/hooks.ts` (+1 file of tests). No backend, SDK, or doc/architecture changes (this is a bug fix, not an architecture change); `CHANGELOG.md` is curated at release time, not per-PR.

`hooks.ts` was already ~1818 lines before this change and the three new pure helpers push it further over the 1000-line guideline. They are kept here (rather than split into a sibling module) because their shared private utilities — `messageIdentity`, `dedupeMessagesByIdentity`, `isNonEmptyString` — also back the existing `mergeMessages`/`buildVisibleHistoryMessages` in this file; a clean extraction means moving that whole utility cluster, which is a separate refactor and out of scope for a bug fix.

### DIFF
--- a/frontend/src/core/threads/hooks.ts
+++ b/frontend/src/core/threads/hooks.ts
@@ -365,6 +365,107 @@ export function mergeMessages(
   });
 }
 
+/**
+ * Derive the live turns that context summarization is about to drop and that
+ * therefore must be re-archived into history.
+ *
+ * Summarization emits `RemoveMessage(ALL)` + a hidden summary + the retained
+ * tail. Everything in the current live thread before the first retained visible
+ * message is being removed; we keep those (minus the summary control messages
+ * already tracked) so the UI can still show the full conversation (#3825).
+ */
+export function computeSummarizationMovedMessages(
+  currentMessages: Message[],
+  summarizationMessages: Message[],
+  summarizedMessageIds: ReadonlySet<string>,
+): Message[] {
+  const firstRetainedVisibleIdentity = summarizationMessages
+    .filter((message) => message.type !== "remove")
+    .filter((message) => !isHiddenFromUIMessage(message))
+    .map(messageIdentity)
+    .find(isNonEmptyString);
+
+  const moved: Message[] = [];
+  for (const message of currentMessages) {
+    if (
+      firstRetainedVisibleIdentity &&
+      messageIdentity(message) === firstRetainedVisibleIdentity
+    ) {
+      break;
+    }
+    if (!summarizedMessageIds.has(message.id ?? "")) {
+      moved.push(message);
+    }
+  }
+  return moved;
+}
+
+/**
+ * Overlay the messages rescued from context summarization on top of the
+ * (possibly stale) visible history so the merged view never drops them.
+ *
+ * Background (#3825): after summarization the backend removes every live
+ * message (`RemoveMessage(ALL)`) and `onUpdateEvent` re-archives the removed
+ * messages into history through an async `setState`. The live thread messages
+ * are owned by the LangGraph SDK external store while the archived history is
+ * React state, so a render can observe the post-summary (shrunk) thread before
+ * the archive `setState` commits — leaving the rescued messages in neither
+ * merge input. Reading them from a synchronous buffer here keeps the merge
+ * correct at every render regardless of how the two state channels interleave.
+ *
+ * The rescued messages are the oldest live turns, so they follow whatever the
+ * already-loaded history holds. Only messages still missing from history are
+ * appended: once history absorbs a rescued message, its live copy stays
+ * authoritative (the buffered copy is an older snapshot and must never overwrite
+ * it), and ordering is preserved.
+ */
+export function resolvePreservedHistory(
+  visibleHistory: Message[],
+  pendingArchivedMessages: Message[],
+): Message[] {
+  if (pendingArchivedMessages.length === 0) {
+    return visibleHistory;
+  }
+  const presentIdentities = new Set(
+    visibleHistory.map(messageIdentity).filter(isNonEmptyString),
+  );
+  const missing = pendingArchivedMessages.filter((message) => {
+    const identity = messageIdentity(message);
+    // Identity-less messages are intentionally skipped: without a stable
+    // identity they cannot be matched against history to drain or dedupe, so
+    // overlaying them would risk a permanent duplicate. They are still archived
+    // through appendMessages and surface via the normal history path instead.
+    return identity !== undefined && !presentIdentities.has(identity);
+  });
+  if (missing.length === 0) {
+    return visibleHistory;
+  }
+  return [...visibleHistory, ...missing];
+}
+
+/**
+ * Drop the archive-buffer entries that the canonical history state has already
+ * absorbed. This keeps the buffer a transient bridge across the async gap
+ * rather than a second long-lived source of truth — otherwise a stale copy
+ * could resurrect a message that history later filtered out (e.g. a superseded
+ * or regenerated run).
+ */
+export function pruneConfirmedArchivedMessages(
+  pendingArchivedMessages: Message[],
+  visibleHistory: Message[],
+): Message[] {
+  if (pendingArchivedMessages.length === 0) {
+    return pendingArchivedMessages;
+  }
+  const confirmedIdentities = new Set(
+    visibleHistory.map(messageIdentity).filter(isNonEmptyString),
+  );
+  return pendingArchivedMessages.filter((message) => {
+    const identity = messageIdentity(message);
+    return !identity || !confirmedIdentities.has(identity);
+  });
+}
+
 function getMessagesAfterBaseline(
   messages: Message[],
   baselineMessageIds: ReadonlySet<string>,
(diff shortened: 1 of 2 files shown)


itxaiohanglover:
Excellent fix! The computeSummarizationMovedMessages function cleanly derives which messages are being dropped by summarization. The overlay approach ensures the UI never loses messages even with the async setState race. The root cause analysis is thorough.

WillemJiang:
@fancyboi999, please take a look at the lint error.

frontend/src/core/threads/hooks.ts
```diff
@@ -896,6 +987,15 @@ export function useThreadStream({
   const latestMessageCountsRef = useRef({ humanMessageCount });
   const sendInFlightRef = useRef(false);
   const messagesRef = useRef<Message[]>([]);
+  // Synchronous bridge for messages rescued from context summarization. The
+  // archived-history `setState` (via appendMessages) lands on a different
+  // schedule than the live thread external store, so the merge reads this buffer
+  // to avoid dropping rescued messages in the render window before history
+  // catches up (#3825).
+  const pendingArchivedMessagesRef = useRef<Message[]>([]);
```
willem-bd, line 1000: The rescue buffer isn't cleared in `onError`. Every other per-stream mutable ref is reset there (`optimisticMessages`, `pendingSupersededRunIds`, `pendingSupersededMessageIds`, `pendingUsageBaselineMessageIdsRef`), but `pendingArchivedMessagesRef` is the lone exception.

`onError` also doesn't invalidate the run-messages/history queries, so the `[visibleHistory]` prune effect has no trigger to run afterward. The buffer then keeps being overlaid by `resolvePreservedHistory` on every subsequent render of this thread until a thread switch or a manual history reload.

In the common case the buffered messages are legitimately archived history, so the visible impact is minor — the real defect is the loss of the "transient bridge, never resurrect" contract: a rolled-back summarization, or a later superseding run, would leave the buffer holding messages that no longer match server state and they couldn't be pruned without a thread switch.

One-line fix: add `pendingArchivedMessagesRef.current = [];` to `onError` alongside its siblings.

frontend/src/core/threads/hooks.ts
```diff
@@ -365,6 +365,103 @@ export function mergeMessages(
...
+ * or regenerated run).
+ */
+export function pruneConfirmedArchivedMessages(
+  pendingArchivedMessages: Message[],
+  visibleHistory: Message[],
+): Message[] {
+  if (pendingArchivedMessages.length === 0) {
+    return pendingArchivedMessages;
+  }
+  const confirmedIdentities = new Set(
```
willem-bd, line 460: nit (optional, not blocking): `pruneConfirmedArchivedMessages`'s filter body (`return !identity || !confirmedIdentities.has(identity)`) is byte-for-byte `getMessagesAfterBaseline` a few lines below, and both this function and `resolvePreservedHistory` independently rebuild `new Set(messages.map(messageIdentity).filter(isNonEmptyString))`. A shared identity-set helper — or having this delegate to `getMessagesAfterBaseline` — would remove the duplication. Flagging only for awareness; the pattern already recurs elsewhere in the file, so feel free to dismiss to keep the diff surgical.

willem-bd (commented):
Reviewed via an automated precision pass (8 finder angles + 1-vote verification).

The fix is sound: the pure helpers (`computeSummarizationMovedMessages`, `resolvePreservedHistory`, `pruneConfirmedArchivedMessages`) are correct and well-tested, the refactor of the inlined moved-messages computation is behaviorally faithful, and the render-phase prop-tracking guard is a sanctioned React pattern.

Most of the dramatic-sounding candidates did not survive verification: the render-phase ref mutation is an accepted React pattern (converges on the prop value, no-op for same-thread re-renders); the superseded-run resurrection is unreachable (the backend hard-rejects regenerating anything but the latest visible AI in the checkpoint, so a rescued/oldest message can never be a regenerate target); the merge-overlap tension doesn't fire in normal operation (the suffix scan breaks on the first appended rescued message, whose identity is absent from the retained tail); and the identity-less leak needs live messages with no id/tool_call_id, which summarization doesn't emit.

One actionable lifecycle gap below, plus one optional cleanup nit.

WillemJiang (approved):

PR #3827, opened 2026-06-27
feat(tracing): add Langfuse session metadata to memory and suggestions

AnoobFeng:
## Summary

PR #2944 added Langfuse Session/User propagation for the main agent graph,
embedded client, and graph-inherited middleware traces.

This PR extends the same attribution behavior to standalone `memory_agent`
and `suggest_agent` model calls, so their independent Langfuse traces are
grouped under the correct thread session and user.

## Root Cause

`memory_agent` and `suggest_agent` call the model directly via
`model.invoke(...)` / `model.ainvoke(...)`.

Unlike the main agent graph, these calls do not start from the graph-root
LangChain callback path where Langfuse promotes reserved metadata such as
`langfuse_session_id` and `langfuse_user_id` onto the root trace.

As a result, these traces could be created without reliable top-level
Session/User attribution.

## Changes

- Reuse the existing Langfuse metadata injection flow for standalone calls.
- Add a Langfuse propagation context for standalone model-level traces.
- Attribute `memory_agent` traces with the queued `thread_id` and `user_id`.
- Attribute `suggest_agent` traces with the route `thread_id` and effective user.
- Add tests covering metadata propagation and no-op behavior when Langfuse is disabled.
- Update backend tracing documentation.

## Compatibility

When Langfuse is not enabled, this change is inert:

- metadata injection remains a no-op
- the propagation context returns a no-op context manager
- Langfuse propagation is not imported or called


backend/packages/harness/deerflow/tracing/metadata.py
```diff
@@ -103,3 +104,41 @@ def inject_langfuse_metadata(
...
+        thread_id=thread_id,
+        user_id=user_id,
+        assistant_id=assistant_id,
+        model_name=model_name,
+        environment=environment,
+    )
+    if not langfuse_metadata:
+        return nullcontext()
+
+    from langfuse import propagate_attributes
```
fancyboi999: **P2 (compat / runtime crash):** `propagate_attributes` was only added to the `langfuse` SDK in **v3.9.0** — it's absent in v3.4.1–v3.8.x (I checked the tags: present in v3.9.0, gone in v3.8.0). But `backend/packages/harness/pyproject.toml` pins `langfuse>=3.4.1`, so on an install in the 3.4.1–3.8.x range with Langfuse enabled this `from langfuse import propagate_attributes` raises `ImportError`, which propagates out of `langfuse_trace_attribute_context()` and crashes the `memory_agent` / `suggest_agent` call. The file's other langfuse usage (`Langfuse`, `langfuse.langchain.CallbackHandler`) works on 3.4.1, so this is the first call that breaks the declared floor.

Fix: bump the floor to `langfuse>=3.9.0` in `pyproject.toml`, or degrade gracefully like the disabled path:
```python
try:
    from langfuse import propagate_attributes
except ImportError:
    return nullcontext()
```

fancyboi999 (commented):
Thanks for extending Langfuse attribution to the standalone `memory_agent` / `suggest_agent` calls. The inert-when-disabled shape (lazy import + `nullcontext`) is clean, the `thread_id`/`user_id`/`model_name` scopes and the metadata keys check out, and the CLA is signed.

One thing to fix before merge:

**P2 — langfuse version floor.** The new standalone path imports `propagate_attributes`, which only exists since langfuse **v3.9.0**, but `backend/packages/harness/pyproject.toml` still pins `langfuse>=3.4.1`. An install in the 3.4.1–3.8.x range (within your declared floor) with Langfuse enabled will hit an `ImportError` and crash memory/suggestion generation. CI won't catch it because a fresh resolve picks up the latest langfuse. Either raise the floor to `>=3.9.0` or guard the import — details inline.

Rest looks good to me. (1 P2 to address.)

AnoobFeng:
Thanks for the review! Went with the graceful-degrade option — wrapped the import in try/except and fall back to nullcontext() if propagate_attributes is missing. Bumping the floor could affect downstream users pinned to older Langfuse, so this keeps 3.4.1–3.8.x installs working with no regression vs. pre-PR behavior. Added a test for the ImportError branch.

PR #3829, opened 2026-06-27
fix(runtime): keep prompt guardrail wrapper out of UI events

### LINKED ISSUE #3689: [bug] Prompt-injection input wrapper leaks into chat UI and first_human_message

ly-wang19:
## Summary
- stamp pre-sanitization user text on model-facing HumanMessages before adding prompt guardrail wrappers
- persist `llm.human.input` events and `first_human_message` from the user-facing original text
- keep file metadata while stripping the internal original-content marker from stored UI message events

Closes #3689

## Tests
- `cd backend && uv run pytest tests/test_input_sanitization_middleware.py tests/test_run_journal.py`
- `cd backend && uv run pytest tests/test_utils_messages.py tests/test_thread_run_messages_pagination.py tests/test_thread_messages_feedback.py`
- `cd backend && uv run ruff check packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py packages/harness/deerflow/runtime/journal.py tests/test_input_sanitization_middleware.py tests/test_run_journal.py`
- `cd backend && uv run ruff format --check packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py packages/harness/deerflow/runtime/journal.py tests/test_input_sanitization_middleware.py tests/test_run_journal.py`

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
@@ -31,6 +31,8 @@ from langchain.agents.middleware.types import (
 from langchain_core.messages import HumanMessage
 from langgraph.errors import GraphBubbleUp
 
+from deerflow.utils.messages import ORIGINAL_USER_CONTENT_KEY, message_content_to_text
+
 logger = logging.getLogger(__name__)
 
 _SUMMARY_MESSAGE_NAME = "summary"
@@ -233,11 +235,14 @@ class InputSanitizationMiddleware(AgentMiddleware[AgentState]):
             else:
                 new_content = processed
 
+            additional_kwargs = dict(msg.additional_kwargs or {})
+            additional_kwargs.setdefault(ORIGINAL_USER_CONTENT_KEY, message_content_to_text(content))
+
             messages[i] = HumanMessage(
                 content=new_content,
                 id=msg.id,
                 name=msg.name,
-                additional_kwargs=msg.additional_kwargs,
+                additional_kwargs=additional_kwargs,
             )
             logger.debug(
                 "InputSanitizationMiddleware: original=%r -> processed=%r",
--- a/backend/packages/harness/deerflow/runtime/journal.py
+++ b/backend/packages/harness/deerflow/runtime/journal.py
@@ -29,7 +29,7 @@ from langchain_core.callbacks import BaseCallbackHandler
 from langchain_core.messages import AIMessage, AnyMessage, BaseMessage, HumanMessage, ToolMessage
 from langgraph.types import Command
 
-from deerflow.utils.messages import message_to_text
+from deerflow.utils.messages import ORIGINAL_USER_CONTENT_KEY, get_original_user_content_text, message_to_text
 
 if TYPE_CHECKING:
     from deerflow.runtime.events.store.base import RunEventStore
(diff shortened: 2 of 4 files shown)


ly-wang19:
Closing this older/conflicting version because #3847 supersedes it with the same prompt-wrapper display fix plus the thread message read-path regression coverage. Keeping #3847 as the merge-ready PR for #3689.

PR #3831, opened 2026-06-27
fix(frontend): keep subtask running steps visible

ly-wang19:
## Summary
- Accumulate `task_running` messages for each subtask instead of replacing the previous running message.
- Render every accumulated running tool-call step in the expanded subtask card while keeping the collapsed summary on the latest message.
- Use the backend `message_index` for stable ordering and dedupe, with message-id fallback for unindexed events.

## Why
Addresses the live-display portion of #3779. The backend already emits each subagent AI message as a separate `task_running` event, but the frontend only kept `latestMessage`, so earlier steps flashed and disappeared.

This does not claim to solve refresh/replay persistence; the issue thread calls that out as a larger backend/storage design.

## Testing
- `frontend/node_modules/.bin/rstest run tests/unit/core/tasks/steps.test.ts tests/unit/core/tasks/subtask-result.test.ts`
- `frontend/node_modules/.bin/rstest run`
- `frontend/node_modules/.bin/tsc --noEmit`
- `frontend/node_modules/.bin/eslint . --ext .ts,.tsx` (passes with 3 existing warnings outside this change)
- `git diff --check`
- `BETTER_AUTH_SECRET=codex-frontend-build-secret-at-least-32-chars SKIP_ENV_VALIDATION=1 frontend/node_modules/.bin/next build` (passes; existing Turbopack NFT tracing warning)
- `PLAYWRIGHT_TEST_BASE_URL=http://localhost:3000 frontend/node_modules/.bin/playwright test --config=playwright.config.ts tests/e2e/subtask-card.spec.ts --project=chromium` against a manual `next start` server

### DIFF
--- a/frontend/src/components/workspace/messages/subtask-card.tsx
+++ b/frontend/src/components/workspace/messages/subtask-card.tsx
@@ -42,6 +42,13 @@ export function SubtaskCard({
   const [collapsed, setCollapsed] = useState(true);
   const rehypePlugins = useRehypeSplitWordsIntoSpans(isLoading);
   const task = useSubtask(taskId)!;
+  const runningSteps =
+    task.steps
+      ?.map((step) => step.message)
+      .filter((message) => hasToolCalls(message)) ??
+    (task.latestMessage && hasToolCalls(task.latestMessage)
+      ? [task.latestMessage]
+      : []);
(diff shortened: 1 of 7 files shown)


WillemJiang:
It was addressed in #3845

PR #3835, opened 2026-06-27
fix(mcp): route tools by authoritative grouping, not name-prefix matching

### LINKED ISSUE #3811: bug: MCP tools mis-routed when one server name is a prefix of another

ly-wang19:
## Summary

When two MCP server names overlap by prefix (e.g. `web` and `web_scraper`), `get_mcp_tools()` re-derived each tool's source server by scanning `servers_config` for the first name that is a prefix of the tool name and taking the first match. A tool from `web_scraper` named `web_scraper_search` matched `web` first (length-4 strip → `scraper_search`), so the call was routed to the wrong server with the wrong stripped name and failed at the MCP layer. The failure was also order-dependent on `servers_config` insertion order, so it could look intermittent.

Fix: drop the prefix-matching loop and use the authoritative `tools_by_server` grouping (`asyncio.gather` preserves `servers_config` order), which is built directly from `client.get_tools(server_name=X)` calls. Same wrap behavior, correct attribution.

Closes #3811.

## Root cause

`get_mcp_tools()` already loads each server's tools via `client.get_tools(server_name=X)` into `tools_by_server` (index `i` corresponds to the i-th server in `servers_config`). That grouping is authoritative. But the existing wrap loop then **discarded** it — flattening `tools_by_server` and re-deriving the source server via ambiguous `tool.name.startswith(f"{name}_")` matching:

```python
for tool in tools:
    tool_server: str | None = None
    for name in servers_config:
        if tool.name.startswith(f"{name}_"):
            tool_server = name
            break
```

With `web` and `web_scraper` both configured, `web_scraper_search` matches `web` first. The downstream `_make_session_pool_tool(tool, "web", ...)` then strips only `web_` (length 4) from the name when invoking the MCP, leaving `scraper_search`, which the `web` server does not expose → `tool not found`. For mixed transports a stdio tool from the longer-named server can also lose session pooling.

## Change

- `backend/packages/harness/deerflow/mcp/tools.py`: replace the prefix-matching loop with `for tool_server, server_tools in zip(servers_config, tools_by_server):` — same wrap decision per tool, but routed via the authoritative grouping.
- `backend/tests/test_mcp_session_pool.py`: regression test `test_overlapping_server_name_prefixes_route_to_correct_server` spies on `_make_session_pool_tool` and asserts each tool is paired with the server that produced it. Verified **red** on the old code (`web_scraper_search` routed to `web`) and **green** on the fix.
- `backend/tests/test_mcp_sync_wrapper.py`: one-line tightening — `test_mcp_tool_sync_wrapper_generation` now uses `transport: http` for its mock server so it no longer relies on the buggy prefix-mismatch skip path to avoid the session-pool wrap.

## Tests

```
cd backend && PYTHONPATH=. uv run pytest tests/test_mcp_session_pool.py tests/test_mcp_sync_wrapper.py
# 44 passed
```

`ruff format` and `ruff check` clean. No harness→app boundary violations.

## AI assistance disclosure

This change was authored with AI assistance (Claude Code). The bug analysis, fix design, red-green regression test, and PR description were produced collaboratively. All code was reviewed and tested by the author before pushing.

### DIFF
--- a/backend/packages/harness/deerflow/mcp/tools.py
+++ b/backend/packages/harness/deerflow/mcp/tools.py
@@ -635,22 +635,20 @@ async def get_mcp_tools() -> list[BaseTool]:
         # Only pool stdio sessions. HTTP/SSE transports use anyio TaskGroups
         # internally which cannot be closed from a different async task, so
         # pooling them causes RuntimeError on cleanup (see #3203).
+        # Route each tool to the server that produced it via the authoritative
+        # tools_by_server grouping (gather preserves servers_config order).
+        # Re-deriving the server by `tool.name.startswith(f"{name}_")` is
+        # ambiguous when one server name is a prefix of another (e.g. `web` and
+        # `web_scraper`): the first prefix match wins and the tool is routed to
+        # the wrong server (see #3811).
         wrapped_tools: list[BaseTool] = []
-        for tool in tools:
-            tool_server: str | None = None
-            for name in servers_config:
-                if tool.name.startswith(f"{name}_"):
-                    tool_server = name
-                    break
-
-            if tool_server is not None:
+        for tool_server, server_tools in zip(servers_config, tools_by_server):
+            for tool in server_tools:
                 transport = servers_config[tool_server].get("transport", "stdio")
                 if transport == "stdio":
                     wrapped_tools.append(_make_session_pool_tool(tool, tool_server, servers_config[tool_server], tool_interceptors))
                 else:
                     wrapped_tools.append(tool)
-            else:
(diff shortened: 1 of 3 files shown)


ly-wang19:
Closing in favor of #3841 — same fix (zip servers with their tools in lockstep rather than re-deriving the server from each tool-name prefix), narrower scope. This branch also bundles unrelated changes to feishu/channel manager/frontend threads that I'll move into separate PRs if I pursue them. Apologies for the churn.

PR #3837, opened 2026-06-27
feat(guardrails): persist security interventions as run events

### LINKED ISSUE #3836: [feat]: Persist guardrail interventions as run events
### Before you start

## Background

DeerFlow already provides `RunJournal.record_middleware()` so middleware can
persist meaningful state changes into `RunEventStore`. On `origin/main`, this
mechanism already has two middleware callers:

- [`SkillActivationMiddleware`](https://github.com/bytedance/deer-flow/blob/69cf4f4d86ceef60a7b35a21581106b2573bbbfd/backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py#L201-L222)
  records `middleware:skill_activation` when a slash command activates a skill;
- [`SafetyFinishReasonMiddleware`](https://github.com/bytedance/deer-flow/blob/69cf4f4d86ceef60a7b35a21581106b2573bbbfd/backend/packages/harness/deerflow/agents/middlewares/safety_finish_reason_middleware.py#L230-L262)
  records `middleware:safety_termination` when a provider safety finish reason
  suppresses requested tool calls.

The contract documented directly on
[`RunJournal.record_middleware()`](https://github.com/bytedance/deer-flow/blob/69cf4f4d86ceef60a7b35a21581106b2573bbbfd/backend/packages/harness/deerflow/runtime/journal.py#L505-L524)
states that it is intended for middleware that performs a meaningful state
change, while pure-observation middleware should not call it:

```python
Called by middleware implementations when they perform a meaningful
state change (e.g., title generation, summarization, HITL approval).
Pure-observation middleware should not call this.
```

Guardrail changes whether a requested tool call is executed, so it represents
a meaningful execution state change rather than pure observation.

`GuardrailMiddleware` is also a security-related middleware. It evaluates tool calls before execution through a `GuardrailProvider`:

- when the provider returns `allow=False`, the tool is not executed and the middleware returns an error `ToolMessage`;
- when the provider raises and `fail_closed=True`, the tool is blocked;
- when the provider raises and `fail_closed=False`, the tool is allowed to continue.

Today these security decisions are only written to the Python logger. They are not persisted into the run event stream. After a run completes, it is hard to answer from structured run events:

- whether a tool call was blocked by guardrail or failed during execution;
- whether the guardrail provider failed;
- whether a provider failure resulted in fail-open or fail-closed behavior;
- which policy / reason code caused the denial.

## Problem

Guardrail deny / fail-open / fail-closed outcomes change the actual agent execution path and are security-relevant interventions. Logger-only visibility is not enough:

- logs may not be retained with run history;
- logs are harder to query by `thread_id` / `run_id`;
- frontend/API/external consumers cannot read these events through `RunEventStore`;
- the persisted run history has no structured event representing this
  execution-path decision.

## Relationship to External Tracing

DeerFlow currently supports LangSmith and Langfuse through graph-level
LangChain callbacks. Those integrations trace model/tool execution, but
`GuardrailMiddleware` does not currently emit a custom trace/span or structured
decision metadata to either provider.

This proposal does not claim that those products lack that capability.
`middleware:guardrail` instead makes the decision part of DeerFlow's own
run-event stream, available through `RunEventStore` even when external tracing
is disabled. It also:

- orders the decision with the other events in the same run using the existing
  sequence and `thread_id` / `run_id` correlation;
- makes it queryable through DeerFlow's run-events API and `event_types`
  filtering;
- gives frontend, API, and external consumers one DeerFlow-native event shape;
- follows the retention and persistence behavior of the configured run-event
  backend instead of requiring a tracing provider.

Exporting the same decision to tracing providers can remain a separate
integration. See the
[DeerFlow tracing callback factory](https://github.com/bytedance/deer-flow/blob/69cf4f4d86ceef60a7b35a21581106b2573bbbfd/backend/packages/harness/deerflow/tracing/factory.py#L12-L52).

## Proposed Behavior

Reuse the existing `RunJournal.record_middleware()` path in `GuardrailMiddleware` and persist `middleware:guardrail` run events only for security-relevant branches:

1. Provider returns `allow=False`
   - `action`: `deny_tool_call`
   - `provider_error`: `false`

2. Provider raises and `fail_closed=True`
   - `action`: `deny_tool_call`
   - `provider_error`: `true`
   - `fail_closed`: `true`

3. Provider raises and `fail_closed=False`
   - `action`: `allow_tool_call_after_provider_error`
   - `provider_error`: `true`
   - `fail_closed`: `false`

Normal `allow=True` decisions should not be recorded by default to avoid run event noise.

Both an explicit policy denial and fail-closed handling after a provider error use `action="deny_tool_call"` because they take the same execution action: preventing the tool from running. Their causes remain distinguishable through `provider_error`, `fail_closed`, `reason_codes`, and `policy_id`, allowing callers to query all denials first and then filter by cause.

## Proposed Event Payload

The `changes` payload can include:

```json
{
  "tool_name": "bash",
  "tool_call_id": "tool_call_1",
  "agent_id": "agent_id",
  "is_subagent": false,
  "user_role": "user",
  "allow": false,
  "policy_id": "test.deny.v1",
  "reason_codes": ["oap.denied"],
  "reason_messages": ["all tools blocked"],
  "fail_closed": true,
  "provider_error": false
}
```

This example mirrors the test fixture and current field sources:

- [`agent_id`](https://github.com/bytedance/deer-flow/blob/69cf4f4d86ceef60a7b35a21581106b2573bbbfd/backend/packages/harness/deerflow/guardrails/middleware.py#L31-L54)
  is `GuardrailMiddleware.passport`, configured as an OAP passport path or
  hosted agent ID; it is not the DeerFlow agent name;
- [`user_role`](https://github.com/bytedance/deer-flow/blob/69cf4f4d86ceef60a7b35a21581106b2573bbbfd/backend/app/gateway/services.py#L169-L190)
  comes from the authenticated user's
  [`system_role`](https://github.com/bytedance/deer-flow/blob/69cf4f4d86ceef60a7b35a21581106b2573bbbfd/backend/app/gateway/auth/models.py#L18-L25),
  whose current values are `user` and `admin`;
- `is_subagent` comes from runtime context and is `false` on the native
  lead-agent Journal path shown here.

To avoid expanding sensitive data persistence, the event should not record:

- `tool_input`
- raw tool arguments
- `user_id`
- `oauth_provider`
- `oauth_id`

`agent_id` / `is_subagent` / `user_role` help explain the guardrail decision.
`user_id` is not duplicated in middleware content: the event remains correlated
through its existing `thread_id` and `run_id`, and continues to use DeerFlow's
existing ownership and access-control paths. This proposal does not change
identity semantics across run-event backends.

`reason_messages` should be bounded, for example 500 characters per message, so provider messages cannot accidentally persist sensitive input verbatim.

## Applicability Boundary

This proposal follows the optional-Journal pattern already used by both
`SkillActivationMiddleware` and `SafetyFinishReasonMiddleware`: an event is
persisted only when the execution context already provides
`runtime.context["__run_journal"]`. Both existing callers skip recording when
the Journal is absent and treat recording failure as a non-fatal debug log.

- Frontend/API and Message Channel runs ultimately enter Gateway `run_agent()`, so they are covered.
- Embedded `DeerFlowClient`, TUI, and direct `agent.invoke()` execution do not currently create a `RunJournal`; guardrail enforcement still runs, but this event is not persisted.
- Subagents use an independent execution context and do not inherit the parent run's `__run_journal`; this proposal does not claim persistence for guardrail decisions made inside subagents.

`is_subagent` remains part of the payload schema because it accurately preserves existing `GuardrailRequest` attribution and supports custom runtimes that provide a Journal, as well as a future subagent Journal bridge. The field does not imply that DeerFlow's native subagent path currently propagates the Journal.

This proposal does not add a cross-thread Journal bridge or change those existing runtime boundaries.

## Failure Semantics

Run event recording should be best-effort and must not change guardrail behavior:

- skip silently when `__run_journal` is absent;
- log only at debug level if `record_middleware()` raises;
- audit persistence failure must not convert allow to deny or deny to allow;
- `GraphBubbleUp` should continue to be re-raised and must not be swallowed by recording logic.

## Expected Benefits

- guardrail security interventions become queryable through `RunEventStore`;
- run history can explain why a tool call did not execute;
- provider error fail-open / fail-closed behavior becomes visible;
- behavior aligns with `SafetyFinishReasonMiddleware`;
- no new audit subsystem or `RunEventStore` schema change is required.

## Acceptance Criteria

- The three security-relevant branches emit the actions described above when a
  run-scoped Journal is available.
- A normal `allow=True` decision emits no guardrail event.
- The event contains attribution and decision fields but excludes tool input,
  raw arguments, and user identity fields.
- Journal absence or persistence failure does not change the guardrail result.
- Sync and async tool-call paths have equivalent behavior.

## Non-Goals

- Do not record all allow decisions;
- do not persist tool input / raw args;
- do not create a new audit subsystem;
- do not change guardrail allow/deny behavior;
- do not make run event persistence fail-closed.
- do not add Journal lifecycle or propagation for Embedded/TUI or subagent execution.

### Affected area(s)

### Alternatives considered

Considered introducing a separate audit subsystem or a new RunEventStore schema. Both were rejected because DeerFlow already provides RunJournal.record_middleware(), and reusing it keeps the change small and consistent with SafetyFinishReasonMiddleware.

### Additional context

_No response_

Miracle778:
Closes #3836 

## Summary

This PR persists security-relevant `GuardrailMiddleware` outcomes as run events. When guardrail intervenes, it writes a `middleware:guardrail` event through the existing `RunJournal.record_middleware()` path.

The recorded scope is intentionally narrow:

- provider deny;
- provider error + fail-closed;
- provider error + fail-open;
- normal allow decisions are not recorded.

The implementation follows the optional-Journal pattern already used by
`SkillActivationMiddleware` and `SafetyFinishReasonMiddleware`: read the
run-scoped journal from `runtime.context["__run_journal"]`, skip when it is
absent, and treat recording failures as non-fatal debug-only errors.

## Motivation

`GuardrailMiddleware` changes the actual execution path of tool calls, but these decisions were previously only written to the logger. After a run completes, structured run events could not explain whether a tool call was denied by guardrail or failed during execution, nor whether a provider error resulted in fail-open or fail-closed behavior.

By persisting these interventions as run events, operators and developers can use `RunEventStore` to review:

- which tool call was denied by guardrail;
- which policy / reason code was involved;
- whether the provider failed;
- whether provider failure allowed or blocked the tool call.

Explicit policy denial and fail-closed handling after a provider error both use `action="deny_tool_call"` to represent the same final execution action. `provider_error`, `fail_closed`, `reason_codes`, and `policy_id` distinguish the causes.

## Surface Area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [x] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**

### `GuardrailMiddleware`

Adds a private helper:

```python
_record_guardrail_event(...)
```

The helper:

- reads `__run_journal` from `ToolCallRequest.runtime.context`;
- builds a bounded `changes` payload;
- calls `journal.record_middleware(tag="guardrail", ...)`;
- catches recording failures and logs them at debug level without changing execution behavior.

The `changes` payload includes:

- `tool_name`
- `tool_call_id`
- `agent_id`
- `is_subagent`
- `user_role`
- `allow`
- `policy_id`
- `reason_codes`
- `reason_messages`
- `fail_closed`
- `provider_error`

It does not record:

- `tool_input`
- raw tool arguments
- `user_id`
- `oauth_provider`
- `oauth_id`

Each reason message is truncated to 500 characters.

### Applicability Boundary

Recording follows the existing `SkillActivationMiddleware` and
`SafetyFinishReasonMiddleware` boundary: an event is written only when
`runtime.context["__run_journal"]` already exists.

- Frontend/API and Message Channel runs execute through Gateway `run_agent()` and are covered.
- Embedded `DeerFlowClient`, TUI, and direct `agent.invoke()` execution do not create a `RunJournal` and are not recorded.
- Subagents do not inherit the parent run's `__run_journal`; this PR does not add a cross-thread Journal bridge or claim persistence for decisions made inside subagents.

`is_subagent` remains an attribution field so the event accurately preserves `GuardrailRequest` context and remains compatible with custom runtimes or a future Journal bridge. Its presence does not imply native subagent Journal propagation.

This is not a Guardrail-specific limitation. Lead-agent and subagent executions
currently have independent runtime contexts, so any middleware that relies on
the lead run's Journal has the same boundary. Defining Journal ownership,
lifecycle, cross-thread propagation, and event ordering for subagents is a
shared runtime concern that should be designed by the project maintainers
rather than solved locally inside `GuardrailMiddleware`. It is explicitly out
of scope for this PR.

### Tests

Extends `test_guardrail_middleware.py` to cover:

- deny records `middleware:guardrail`;
- provider error + fail-closed records and denies;
- provider error + fail-open records and allows the handler;
- normal allow does not record;
- payload excludes tool args and user identity fields;
- `record_middleware()` failure does not affect denial behavior;
- sync and async paths remain aligned.

## Verification

Ran:

```bash
uv run pytest tests/test_guardrail_middleware.py tests/test_sandbox_audit_middleware.py -q
```

Result:

**189 passed.**

## Manual Runtime Validation

The implementation is also intended to be exercised against a locally running
DeerFlow instance with persistent run-event storage enabled.

Validation procedure:

1. Start the DeerFlow Gateway and frontend using the project-supported setup,
   with a persistent run-event backend configured.
2. Configure a guardrail provider to deny a known tool call such as `bash`.
3. Trigger that tool call from the DeerFlow UI.
4. Confirm that the tool is not executed and that the run contains a
   `middleware:guardrail` event.
5. Inspect the persisted event and verify `action`, `tool_name`, `is_subagent`,
   `provider_error`, and `fail_closed`.

config.yaml:
```
run_events:
  backend: db
  max_trace_content: 10240
  track_token_usage: true
guardrails:
  enabled: true
  fail_closed: true
  provider:
    use: deerflow.guardrails.builtin:AllowlistProvider
    config:
      denied_tools:
      - bash
```

Example query when using the DB-backed event store:

```sql
sqlite3 -json backend/.deer-flow/data/deerflow.db "
SELECT run_id, seq, event_type, content
FROM run_events
WHERE event_type = 'middleware:guardrail'
ORDER BY id DESC
LIMIT 10;"
```

### Runtime Behavior

image

### Persisted Run Event
image

The observed values and screenshots will be added after the local runtime
validation and before this PR is submitted.

## Security and Privacy Notes

This PR does not persist tool input or raw args, avoiding storage of sensitive
content that the guardrail may have blocked. It also does not duplicate
`user_id` in middleware content; the event continues to use existing
`thread_id` / `run_id` correlation and DeerFlow's ownership/access-control
paths. This PR does not change identity semantics across run-event backends.

Recording is best-effort and does not change guardrail allow/deny behavior.

## Non-Goals

- Do not record every allow decision;
- do not add a new `RunJournal` API;
- do not change the `RunEventStore` schema;
- do not create a new audit subsystem;
- do not change guardrail provider decision semantics.
- do not add Journal lifecycle or propagation for Embedded/TUI or subagent execution.

## AI Assistance

**Tool(s) used:** Claude Code and Codex

**How they were used:** The implementation was produced from a reviewed design
with Claude Code, then reviewed and refined with Codex. Tests and repository
behavior were inspected as part of that review.

- [x] I've read and understand every line of this change and take
  responsibility for it; it is not unreviewed AI output.

### DIFF
--- a/backend/packages/harness/deerflow/guardrails/middleware.py
+++ b/backend/packages/harness/deerflow/guardrails/middleware.py
@@ -16,6 +16,8 @@ from deerflow.guardrails.provider import GuardrailDecision, GuardrailProvider, G
 
 logger = logging.getLogger(__name__)
 
+_REASON_MESSAGE_LIMIT = 500
+
 
 class GuardrailMiddleware(AgentMiddleware[AgentState]):
     """Evaluate tool calls against a GuardrailProvider before execution.
@@ -63,6 +65,57 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
             status="error",
         )
 
+    def _record_guardrail_event(
+        self,
+        request: ToolCallRequest,
+        guardrail_request: GuardrailRequest,
+        decision: GuardrailDecision,
+        *,
+        action: str,
+        provider_error: bool,
+    ) -> None:
+        """Persist a security-relevant guardrail decision to RunJournal.
+
+        This mirrors SafetyFinishReasonMiddleware: audit persistence is
+        best-effort and must never change tool execution behavior.
+        """
+        runtime = getattr(request, "runtime", None)
+        context = getattr(runtime, "context", None) if runtime is not None else None
+        if not isinstance(context, dict):
+            return
+
+        journal = context.get("__run_journal")
+        if journal is None:
+            return
+
+        reason_codes = [reason.code for reason in decision.reasons if reason.code]
+        reason_messages = [reason.message[:_REASON_MESSAGE_LIMIT] for reason in decision.reasons if reason.message]
+
+        changes = {
+            "tool_name": guardrail_request.tool_name,
+            "tool_call_id": guardrail_request.tool_call_id,
+            "agent_id": guardrail_request.agent_id,
+            "is_subagent": guardrail_request.is_subagent,
+            "user_role": guardrail_request.user_role,
+            "allow": decision.allow,
+            "policy_id": decision.policy_id,
+            "reason_codes": reason_codes,
+            "reason_messages": reason_messages,
+            "fail_closed": self.fail_closed,
+            "provider_error": provider_error,
+        }
+
+        try:
+            journal.record_middleware(
+                tag="guardrail",
+                name=type(self).__name__,
+                hook="wrap_tool_call",
+                action=action,
+                changes=changes,
+            )
+        except Exception:  # noqa: BLE001
+            logger.debug("Failed to record middleware:guardrail event", exc_info=True)
+
     @override
     def wrap_tool_call(
         self,
@@ -79,10 +132,33 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
             logger.exception("Guardrail provider error (sync)")
             if self.fail_closed:
                 decision = GuardrailDecision(allow=False, reasons=[GuardrailReason(code="oap.evaluator_error", message="guardrail provider error (fail-closed)")])
+                self._record_guardrail_event(
+                    request,
+                    gr,
+                    decision,
+                    action="deny_tool_call",
+                    provider_error=True,
+                )
+                return self._build_denied_message(request, decision)
             else:
+                decision = GuardrailDecision(allow=True, reasons=[GuardrailReason(code="oap.evaluator_error", message="guardrail provider error (fail-open)")])
+                self._record_guardrail_event(
+                    request,
+                    gr,
+                    decision,
+                    action="allow_tool_call_after_provider_error",
+                    provider_error=True,
+                )
                 return handler(request)
         if not decision.allow:
             logger.warning("Guardrail denied: tool=%s policy=%s code=%s", gr.tool_name, decision.policy_id, decision.reasons[0].code if decision.reasons else "unknown")
+            self._record_guardrail_event(
+                request,
+                gr,
+                decision,
+                action="deny_tool_call",
+                provider_error=False,
+            )
             return self._build_denied_message(request, decision)
         return handler(request)
 
@@ -102,9 +178,32 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
             logger.exception("Guardrail provider error (async)")
             if self.fail_closed:
                 decision = GuardrailDecision(allow=False, reasons=[GuardrailReason(code="oap.evaluator_error", message="guardrail provider error (fail-closed)")])
+                self._record_guardrail_event(
+                    request,
+                    gr,
+                    decision,
+                    action="deny_tool_call",
+                    provider_error=True,
+                )
+                return self._build_denied_message(request, decision)
             else:
+                decision = GuardrailDecision(allow=True, reasons=[GuardrailReason(code="oap.evaluator_error", message="guardrail provider error (fail-open)")])
+                self._record_guardrail_event(
+                    request,
+                    gr,
+                    decision,
+                    action="allow_tool_call_after_provider_error",
+                    provider_error=True,
+                )
                 return await handler(request)
         if not decision.allow:
             logger.warning("Guardrail denied: tool=%s policy=%s code=%s", gr.tool_name, decision.policy_id, decision.reasons[0].code if decision.reasons else "unknown")
+            self._record_guardrail_event(
+                request,
+                gr,
+                decision,
+                action="deny_tool_call",
+                provider_error=False,
+            )
             return self._build_denied_message(request, decision)
         return await handler(request)
--- a/backend/packages/harness/deerflow/guardrails/middleware.py
+++ b/backend/packages/harness/deerflow/guardrails/middleware.py
@@ -33,11 +33,13 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
         self.fail_closed = fail_closed
         self.passport = passport
 
-    def _build_request(self, request: ToolCallRequest) -> GuardrailRequest:
+    @staticmethod
+    def _resolve_context(request: ToolCallRequest) -> dict:
         runtime = getattr(request, "runtime", None)
         context = getattr(runtime, "context", None) if runtime is not None else None
-        context = context if isinstance(context, dict) else {}
+        return context if isinstance(context, dict) else {}
 
+    def _build_request(self, request: ToolCallRequest, context: dict) -> GuardrailRequest:
         return GuardrailRequest(
             tool_name=str(request.tool_call.get("name", "")),
             tool_input=request.tool_call.get("args", {}),
@@ -67,7 +69,7 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
 
     def _record_guardrail_event(
         self,
-        request: ToolCallRequest,
+        context: dict,
         guardrail_request: GuardrailRequest,
         decision: GuardrailDecision,
         *,
@@ -81,11 +83,6 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
         behavior. Runtimes without ``__run_journal`` (including embedded and
         subagent execution) skip persistence.
         """
-        runtime = getattr(request, "runtime", None)
-        context = getattr(runtime, "context", None) if runtime is not None else None
-        if not isinstance(context, dict):
-            return
-
         journal = context.get("__run_journal")
         if journal is None:
             return
@@ -97,6 +94,8 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
             "tool_name": guardrail_request.tool_name,
             "tool_call_id": guardrail_request.tool_call_id,
             "agent_id": guardrail_request.agent_id,
+            # Native subagents do not currently inherit __run_journal; custom
+            # runtimes may still provide one with subagent attribution.
             "is_subagent": guardrail_request.is_subagent,
             "user_role": guardrail_request.user_role,
             "allow": decision.allow,
@@ -124,7 +123,8 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
         request: ToolCallRequest,
         handler: Callable[[ToolCallRequest], ToolMessage | Command],
     ) -> ToolMessage | Command:
-        gr = self._build_request(request)
+        context = self._resolve_context(request)
+        gr = self._build_request(request, context)
         try:
             decision = self.provider.evaluate(gr)
         except GraphBubbleUp:
@@ -135,7 +135,7 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
             if self.fail_closed:
                 decision = GuardrailDecision(allow=False, reasons=[GuardrailReason(code="oap.evaluator_error", message="guardrail provider error (fail-closed)")])
                 self._record_guardrail_event(
-                    request,
+                    context,
                     gr,
                     decision,
                     action="deny_tool_call",
@@ -145,7 +145,7 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
             else:
                 decision = GuardrailDecision(allow=True, reasons=[GuardrailReason(code="oap.evaluator_error", message="guardrail provider error (fail-open)")])
                 self._record_guardrail_event(
-                    request,
+                    context,
                     gr,
                     decision,
                     action="allow_tool_call_after_provider_error",
@@ -155,7 +155,7 @@ class GuardrailMiddleware(AgentMiddleware[AgentState]):
         if not decision.allow:
             logger.warning("Guardrail denied: tool=%s policy=%s code=%s", gr.tool_name, decision.policy_id, decision.reasons[0].code if decision.reasons else "unknown")
             self._record_guardrail_event(
-                request,
+                context,
                 gr,
                 decision,
                 action="deny_tool_call",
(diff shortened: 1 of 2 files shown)


backend/packages/harness/deerflow/guardrails/middleware.py
```diff
@@ -63,6 +65,59 @@ def _build_denied_message(self, request: ToolCallRequest, decision: GuardrailDec
...
+            return
+
+        reason_codes = [reason.code for reason in decision.reasons if reason.code]
+        reason_messages = [reason.message[:_REASON_MESSAGE_LIMIT] for reason in decision.reasons if reason.message]
+
+        changes = {
+            "tool_name": guardrail_request.tool_name,
+            "tool_call_id": guardrail_request.tool_call_id,
+            "agent_id": guardrail_request.agent_id,
+            "is_subagent": guardrail_request.is_subagent,
```
willem-bd, line 99: **`is_subagent` is structurally always `False` in persisted records.** `__run_journal` is set only in `runtime/runs/worker.py:224` (the lead-agent worker); the subagent executor builds a fresh context without it, so `_record_guardrail_event` short-circuits (`journal is None → return`) whenever `is_subagent` would be `True`. An operator querying `WHERE is_subagent = true` to audit subagent guardrail denials gets zero rows even when subagent tool calls were blocked. The PR body acknowledges this and keeps the field for future/custom runtimes — so this is a *confirm the trade-off is acceptable* point, not a defect. If the lead-only boundary is intentional for now, consider a one-line comment on the field so the misleading-query outcome isn't a surprise to future readers.

backend/packages/harness/deerflow/guardrails/middleware.py
```diff
@@ -63,6 +65,59 @@ def _build_denied_message(self, request: ToolCallRequest, decision: GuardrailDec
...
+        provider_error: bool,
+    ) -> None:
+        """Persist a security-relevant guardrail decision to RunJournal.
+
+        This follows the optional-Journal pattern used by existing middleware:
+        audit persistence is best-effort and must never change tool execution
+        behavior. Runtimes without ``__run_journal`` (including embedded and
+        subagent execution) skip persistence.
+        """
+        runtime = getattr(request, "runtime", None)
```
willem-bd: **Minor: context is read twice.** `_record_guardrail_event` re-does the `getattr(request, "runtime", None)` → `getattr(runtime, "context", None)` chain that `_build_request` (line 37, run just before to produce `gr`) already performed. `runtime.context` is a stable dict for the call's lifetime, so this isn't a bug, but the journal could be resolved once and threaded through (or via a shared `_resolve_context` helper) instead of re-reading. Matches the sibling-middleware pattern, so low priority.

backend/packages/harness/deerflow/guardrails/middleware.py
```diff
@@ -79,10 +134,33 @@ def wrap_tool_call(
             logger.exception("Guardrail provider error (sync)")
             if self.fail_closed:
                 decision = GuardrailDecision(allow=False, reasons=[GuardrailReason(code="oap.evaluator_error", message="guardrail provider error (fail-closed)")])
+                self._record_guardrail_event(
```
willem-bd, line 137: **Minor: sync/async recording blocks are copy-pasted.** The three `_record_guardrail_event` call sites in `wrap_tool_call` (here, and at 147/157) are near-identical to the three in `awrap_tool_call` (183/193/203), differing only in `handler(request)` vs `await handler(request)`. The two audit trails must stay in sync; a shared `_handle_decision(..., is_async)` helper would prevent drift if action strings or `changes` fields evolve. Cleanup, not a bug today — `SafetyFinishReasonMiddleware` routes both hooks through a single `_apply`.

willem-bd (commented):
Reviewing for precision — no correctness bugs. The `record_middleware(tag, *, name, hook, action, changes)` call matches the journal signature, all referenced `GuardrailRequest`/`GuardrailDecision` fields exist, recording sits in the right paths (deny / fail-closed / fail-open; allows not recorded), and the sync `record_middleware` call from `awrap_tool_call` is safe — `_put` just appends to an in-memory buffer and schedules an async flush, so there's no blocking I/O on the event loop. Verified-not-bugs: the `reason.message=None` deref can't happen (`if r.message` filters before slicing); the fail-open `else` still returns the handler result unchanged. Three minor notes inline — one substantive (`is_subagent`), two cleanups.

backend/packages/harness/deerflow/guardrails/middleware.py
Miracle778: Confirmed: the lead-only boundary is intentional for this PR. I added an inline comment above `is_subagent` clarifying that native subagents do not inherit `__run_journal`, while custom runtimes may still provide a Journal with subagent attribution. Subagent Journal propagation remains outside this change.

backend/packages/harness/deerflow/guardrails/middleware.py
Miracle778: Resolved in `a1cffbad`: each hook now resolves `runtime.context` once via `_resolve_context()` and passes the same dict to `_build_request()` and `_record_guardrail_event()`.

backend/packages/harness/deerflow/guardrails/middleware.py
Miracle778: I evaluated extracting a shared decision handler, but the sync and async hooks still require different provider and handler invocation paths. The extraction made the control flow larger without removing those core differences.

The event payload construction is already centralized in `_record_guardrail_event()`, and the tests cover equivalent deny and provider-error behavior across both paths. I kept the explicit sync/async branches to avoid broadening this PR with a larger control-flow refactor.

WillemJiang (approved):

WillemJiang (approved):

PR #3665, opened 2026-06-20
feat(guardrails): expose authenticated runtime context in GuardrailRequest

Miracle778:
Closes #3664

## Why

当前 `GuardrailProvider` 已经可以返回结构化的 `allow/deny` 决策，包括 `reasons`、`policy_id` 和 `metadata`。但传给 provider 的 `GuardrailRequest` 缺少足够的已认证运行时上下文。

现在 provider 主要能看到 tool name 和 tool input，但无法稳定知道：

- 是哪个已认证 DeerFlow 用户触发了这次工具调用；
- 这个用户在 DeerFlow 内部是什么角色；
- 如果未来接入 OAuth / SSO，这个用户对应哪个外部 identity provider / subject；
- 这次决策属于哪个 `thread_id`、`run_id` 或 `tool_call_id`。

这会让多用户部署、role-aware tool policy、provider 侧审计日志等 custom provider 场景变得更难实现。

这个 PR 保持 Guardrail 的职责边界不变：不新增 policy engine、RBAC 系统、governance 子系统，也不改变默认行为。它只把 DeerFlow 已经掌握的 authenticated runtime context 传入 `GuardrailRequest`，让 custom provider 在需要时可以消费这些上下文。

## What changed

- `GuardrailRequest` 新增以下 optional runtime context 字段：
  - `user_id`
  - `user_role`
  - `oauth_provider`
  - `oauth_id`
  - `run_id`
  - `tool_call_id`
- `thread_id` 现在会从 `ToolCallRequest.runtime.context` 填充；该字段之前已经存在，但 `GuardrailMiddleware` 没有实际填充。
- Gateway 从 `request.state.user` 注入 authenticated user context：
  - `user_id`
  - `user_role`
  - `oauth_provider`
  - `oauth_id`
- client-supplied context 不能伪造或覆盖服务端认证态用户上下文。
- `GuardrailMiddleware` 从 `ToolCallRequest.runtime.context` 读取新增字段，并从 `request.tool_call` 读取 tool call id。
- `backend/docs/GUARDRAILS.md` 新增 Runtime Attribution 文档，并展示 custom provider 如何把这些字段归一化成 provider-defined policy context。
- 测试覆盖 Gateway 注入、防伪造覆盖、middleware request 构造、缺失 context 的行为，以及现有 provider 的兼容性。
-  `docs/superpowers` 中的 design spec 和 implementation plan 也同步更新，记录这次字段来源、兼容性和测试覆盖。

## Surface area

[...]

PR #3294, opened 2026-05-28
fix(mcp): add auth interceptor with channel user_id and keep header propagation to mcp tools

zhongli-sz:
## 背景 / Background

在非 Web 鉴权链路（如 飞书 channel 消息）中，`user_id` 未正确进入运行时上下文，导致 interceptor 侧拿到默认值（default）；同时 interceptor 注入的 header 在 MCP 工具调用时未透传，导致下游工具无法读取到鉴权/上下文信息。

In non-web auth flows (e.g., feishu channel messages), `user_id`  not be correctly propagated into runtime context, causing interceptor-side fallback/default values. Also, headers injected by interceptors were not forwarded during MCP tool calls, so downstream tools could not consume auth/context headers.

## 本次改动 / What Changed

1. 在 `ChannelManager` 构建上下文时注入 `user_id`（来自 `msg.user_id`），避免 channel 场景丢失用户身份。  
   Inject `user_id` from `msg.user_id` when building run context in `ChannelManager`, ensuring channel flows keep caller identity.

2. 在网关上下文合并逻辑中，确保 `user_id` 写入 runtime context（不仅限于 configurable 字段）。  
   Ensure `user_id` is merged into runtime context in gateway override merge logic (not only configurable fields).

3. 在 `inject_authenticated_user_context` 中跳过 `internal` 系统角色，避免内部调用覆盖/污染用户上下文。  
   Skip `internal` system-role users in `inject_authenticated_user_context` to avoid overriding/polluting user context for internal calls.

4. 在 MCP 工具调用时透传 interceptor header：将 `request.headers` 放入 `meta.headers` 传给 `session.call_tool(...)`。  
   Forward interceptor headers in MCP tool calls by passing `request.headers` via `meta.headers` to `session.call_tool(...)`.

## 影响 / Impact

- 修复 interceptor 获取到默认 `user_id` 的问题，用户身份在 channel 链路中可正确透传。  
- 修复 interceptor 注入 header 在 MCP 工具调用中被忽略的问题。  
- 降低鉴权上下文丢失风险，提升跨链路行为一致性。  

- Fixes default/fallback `user_id` issue in channel flows.  
- Fixes dropped interceptor headers during MCP tool invocation.  
- Reduces auth-context loss and improves cross-path consistency.

## 变更文件 / Files Changed

- `backend/app/channels/manager.py`
- `backend/app/gateway/services.py`
- `backend/packages/harness/deerflow/mcp/tools.py`

## 测试建议 / Test Plan

[...]

PR #3839, opened 2026-06-28
fix(guardrails): propagate internal owner attribution to guardrail context

### LINKED ISSUE #3672: feat(guardrails): propagate bound-owner role/oauth into guardrail context for IM/internal-auth runs
## Context

Follow-up from PR #3665 (exposes authenticated runtime context in `GuardrailRequest`) and @willem-bd's review there. #3665 covers the web (lead-agent) path and, in a follow-up commit, subagent tool calls. The IM / internal-auth path is still uncovered.

## Problem

For IM / internal-auth runs (Slack / Discord / Telegram), guardrail tool-call evaluation sees `user_role = None`, so role-aware policy (which the #3665 docs promote, e.g. "allow `bash` only for `admin`") cannot apply to channel-delivered work. `user_id` works (user-based policy is fine); only role/oauth attribution is missing.

## Root cause (verified)

- On the internal-auth path `request.state.user` is the synthetic sentinel `SimpleNamespace(id=DEFAULT_USER_ID, system_role="internal")` (`app/gateway/internal_auth.py:43`) — it does **not** carry the bound owner's identity at all.
- The real owner is delivered out-of-band via a trusted internal header and read with `get_trusted_internal_owner_user_id(request)` (`internal_auth.py:48-`), set by the channel layer (`app/channels/manager.py:512-515`).
- `inject_authenticated_user_context` (`app/gateway/services.py:181`) early-returns for `INTERNAL_SYSTEM_ROLE`, and `user_role`/`oauth_*` aren't in `_CONTEXT_CONFIGURABLE_KEYS`, so nothing populates the owner's role/oauth for these runs.

So unlike the web path, the owner's role/oauth is **not on hand** at the inject point — it requires resolving the owner from `owner_user_id`.

## Dependency

Blocked on #3665. The `GuardrailRequest.user_role`/`oauth_*` fields, the `inject` role/oauth stamping, and `GuardrailMiddleware` reading them do **not** exist on `main` yet — they live in #3665. This follow-up should be implemented on top of #3665 (after it merges, or as a commit on its branch), not as a standalone PR off `main`.

## Proposed direction (once #3665 lands)

- For internal-auth runs, resolve the bound owner via `get_trusted_internal_owner_user_id(request)`, look up that owner's `UserRow` (`system_role`, `oauth_provider`, `oauth_id`), and stamp them into the run context so `GuardrailMiddleware` exposes them in `GuardrailRequest` — mirroring what `inject_authenticated_user_context` does for web users.
- Keep the web path unchanged; default behavior (no guardrail provider) unchanged. One owner-record lookup per IM run (cache if it shows up as hot).

## Non-goals

- No new policy engine / RBAC — same boundary as #3665.
- No change to request routing or auth behavior beyond owner-role attribution for guardrail context.

## Related

- #3665, #3664

sqsge:
Fixes #3672

## Summary

- Resolve the trusted internal owner user for internal-auth channel runs.
- Stamp the owner role/oauth attribution into runtime context for `GuardrailRequest`.
- Preserve existing channel `user_id` storage/thread ownership semantics.
- Strip untrusted role/oauth context on internal-auth runs when no owner user is resolved.

## Why

#3665 added `GuardrailRequest.user_role` and `oauth_*` attribution for authenticated web runs, but IM/internal-auth runs still saw only the synthetic internal user. Role-aware guardrail policies therefore could not distinguish owner roles for Slack/Discord/Telegram/Feishu-triggered work.

The fix resolves the bound owner from the trusted internal owner header at the Gateway service boundary and uses the server-side owner record for attribution. Client-supplied role/oauth context is not trusted.

## Scope

This only affects internal-auth run attribution for guardrails. It does not add RBAC, change policy evaluation, alter request authorization, or change memory/thread metadata routers.

## Failure behavior

If owner lookup fails, the run continues with role/oauth attribution unset. This preserves existing default behavior and avoids treating attribution lookup as an authorization gate.

## Validation

- `uv run pytest tests/test_gateway_services.py -q`
- `uv run pytest tests/test_guardrail_middleware.py::TestGuardrailRequestAttribution tests/test_setup_agent_e2e_user_isolation.py::TestConfigAssembly -q`
- `uv run ruff check app/gateway/services.py tests/test_gateway_services.py`
- `uv run ruff format --check app/gateway/services.py tests/test_gateway_services.py`
- `git diff --check`

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Codex helped trace the internal-auth run context path, implement the minimal Gateway attribution fix, and add regression coverage. I reviewed the diff and validation results.


backend/app/gateway/services.py
```diff
@@ -180,6 +200,20 @@ def inject_authenticated_user_context(config: dict[str, Any], request: Request)
...
+            return
+        if internal_owner_user is None:
+            runtime_context.pop("user_role", None)
+            runtime_context.pop("oauth_provider", None)
+            runtime_context.pop("oauth_id", None)
+            return
+        owner_user_id = getattr(internal_owner_user, "id", None)
+        if owner_user_id is not None:
+            runtime_context.setdefault("user_id", str(owner_user_id))
+        runtime_context["user_role"] = getattr(internal_owner_user, "system_role", None)
```
willem-bd, line 307: **This change makes `backend/docs/GUARDRAILS.md` stale — please update it in the same PR.**

GUARDRAILS.md still documents the exact limitation this PR removes:

- **Line 272 ("Known limitation — IM / internal-auth runs")**: *"`inject_authenticated_user_context` early-returns for the internal system role … so on those runs `user_role`/`oauth_provider`/`oauth_id` stay `None` … role-based policy cannot be applied to channel-delivered work in this release … Threaded alongside the owner `system_role` in a follow-up if channel-originated role policy is needed."* — After this PR the internal branch no longer early-returns; it stamps `user_role`/`oauth_provider`/`oauth_id` from the resolved owner (this line and the two below). The whole paragraph is now false, and it even describes this PR as the unshipped "follow-up."
- **Line 270**: *"…writes `user_id`/`user_role`/`oauth_provider`/`oauth_id` from `request.state.user`"* — for the internal role these now come from the DB-resolved owner (`internal_owner_user`), not `request.state.user` (which is the synthetic internal user with `oauth_*` = None). A maintainer tracing attribution from that sentence will reach the wrong conclusion.

The repo's cross-cutting convention (AGENTS.md) asks for docs to be updated in the same change set as the code. Please refresh both the "Known limitation" paragraph and the line-270 wording to describe the new internal-owner attribution path.

backend/app/gateway/services.py
```diff
@@ -180,6 +200,20 @@ def inject_authenticated_user_context(config: dict[str, Any], request: Request)
...
+        if not isinstance(runtime_context, dict):
+            return
+        if internal_owner_user is None:
+            runtime_context.pop("user_role", None)
+            runtime_context.pop("oauth_provider", None)
+            runtime_context.pop("oauth_id", None)
+            return
+        owner_user_id = getattr(internal_owner_user, "id", None)
+        if owner_user_id is not None:
+            runtime_context.setdefault("user_id", str(owner_user_id))
```
willem-bd: **Authoritative role/oauth are paired with a still-client-controllable `user_id`.**

Here `user_id` is written with `setdefault`, so a client-supplied `user_id` (placed into `config["context"]` earlier by `build_run_config` copying `body.config["context"]`, or by `merge_run_context_overrides` from `body.context`) is kept — while `user_role`/`oauth_provider`/`oauth_id` on the lines just below are *hard-assigned* from the server-resolved owner.

Result: the `GuardrailRequest` can carry a client-chosen `user_id` alongside the authoritative role/oauth of a different (resolved-owner) principal. *Scenario:* an internal-token caller sets `body.config["context"]["user_id"]` to an arbitrary id while the header owner resolves to, say, an admin — guardrail decisions keyed on `user_id` are then evaluated against a spoofable id that no longer matches the role it's paired with.

I understand keeping the channel-supplied `user_id` is deliberate (PR #3294), so this may be acceptable — but this PR is what introduces the new *mismatch* between a spoofable `user_id` and server-authoritative role/oauth, and the surrounding code already treats leaked internal tokens as in-scope ("a leaked internal token must not grant cross-user thread access"). Please make an explicit call: either stamp `user_id` authoritatively from the resolved owner on the attribution path, or document why the `user_id`/role split is intended so a future reader doesn't treat it as a bug.

willem-bd (commented):
Thanks — the server-side owner resolution and the strip-on-no-owner path are the right shape, and the regression tests are thorough. Two things I'd like addressed before merge, left inline.

Note on the first comment: it's really about `backend/docs/GUARDRAILS.md:270-272`, which this PR makes stale — but since that file isn't part of the diff, GitHub won't let me anchor a comment there, so I've attached it to the code line that causes the staleness.

WillemJiang:
@sqsge, thanks for your contribution. Please take a look at the review comments.

sqsge:
Thanks for the review. I pushed `7b547903` to address both inline comments.

What changed:

- Made the trusted internal-owner attribution path stamp `context.user_id` authoritatively from the server-resolved owner instead of preserving a client-supplied value with `setdefault`. This keeps `user_id`, `user_role`, `oauth_provider`, and `oauth_id` aligned to the same principal when an owner is resolved.
- Updated `backend/docs/GUARDRAILS.md` so the runtime attribution section now documents the current web-auth and trusted IM/internal-auth behavior, including the no-resolved-owner fallback where client-supplied role/oauth attribution is stripped.
- Tightened the regression test by passing `context.user_id = "spoofed-client"` and asserting the guardrail context is stamped back to the resolved owner (`"owner-1"`).

Validation:

- `pytest tests/test_gateway_services.py -k "inject_authenticated_user_context or start_run_stamps_internal_owner_guardrail_attribution or start_run_uses_internal_owner_header_for_persistence" -v` -> 5 passed
- `pytest tests/test_gateway_services.py -v` -> 52 passed
- `pytest tests/ -v` -> 5188 passed, 19 skipped
- `ruff check .` -> passed
- `ruff format --check .` -> passed

backend/app/gateway/services.py
sqsge: Addressed in `7b547903`.

I updated `backend/docs/GUARDRAILS.md` to remove the stale internal-auth limitation and document the current attribution behavior:

- web-authenticated runs stamp from `request.state.user`;
- trusted IM/internal-auth runs resolve the owner user server-side and stamp `user_id`/`user_role`/`oauth_provider`/`oauth_id` from that owner;
- if no owner is resolved, client-supplied role/oauth attribution is stripped instead of treated as authoritative.

This should bring the guardrail docs back in sync with the code path changed by this PR.

backend/app/gateway/services.py
sqsge: Addressed in `7b547903`.

I changed the resolved-owner attribution path from `runtime_context.setdefault("user_id", ...)` to an authoritative assignment from the server-resolved owner. That keeps `user_id`, `user_role`, `oauth_provider`, and `oauth_id` aligned to the same principal when the trusted owner is available.

I also tightened the regression coverage so the request explicitly supplies `context.user_id = "spoofed-client"`, and the test asserts the final guardrail context is stamped back to the resolved owner (`"owner-1"`).

WillemJiang:
@sqsge, please resolve the conflicts with the main branch.

sqsge:
Conflicts with `main` have been resolved.

What I did:

- Merged latest `origin/main` (`f1225944`) into this branch.
- Resolved the conflicts in `backend/app/gateway/services.py` and `backend/tests/test_gateway_services.py`.
- Kept the new scheduler/internal-caller context handling from `main` while preserving the trusted internal-owner attribution fix from this PR.

Validation:

- `pytest tests/test_gateway_services.py -q` -> 65 passed
- `ruff check .` -> passed
- `ruff format --check .` -> passed
- `pytest tests/ -q` -> 6445 passed, 26 skipped

Note: the full backend suite needs DNS/network access for the web-fetch URL safety tests; the sandboxed run failed those DNS checks, and the same suite passed after rerunning with network access.

WillemJiang (approved):

PR #3841, opened 2026-06-28
fix(mcp): route tools by server index, not name prefix

### LINKED ISSUE #3811: bug: MCP tools mis-routed when one server name is a prefix of another

ly-wang19:
## What Problem This Solves

Closes #3811.

When two MCP servers are configured and one server's name is a prefix of another (e.g. `web` and `web_scraper`, or `github` and `github_enterprise`), tools from the longer-named server are routed to the wrong server. With `tool_name_prefix=True`, a tool from `web_scraper` is named `web_scraper_search`; the wrapping loop re-derives the source server by scanning `servers_config` for the first name that is a prefix of the tool name, so `"web_scraper_search".startswith("web_")` matches `web` first. The tool is then pooled under `web` and invoked with the wrongly-stripped original name (`scraper_search`), causing every call to fail with `"tool not found"`. For mixed transports, a stdio tool from the longer-named server can also lose session pooling.

## Why This Change Was Made

`get_mcp_tools()` already had the authoritative per-server grouping in hand — `tools_by_server[i]` aligns with the i-th server in `servers_config` because `asyncio.gather` preserves input order — but discarded it by flattening and re-deriving the server via ambiguous name-prefix matching.

The fix walks servers and their tools in lockstep (`zip(servers_config, tools_by_server)`) and wraps each tool with the server that actually produced it. This removes the prefix-match heuristic entirely.

## User Impact

- MCP servers whose names share a prefix (e.g. `web` + `web_scraper`, `github` + `github_enterprise`) now route tools correctly. No more silent "tool not found" failures for the longer-named server.
- Users running a single MCP server, or servers with no name overlap, see no behavior change.
- Config insertion order no longer affects routing.

## Evidence

New regression test `tests/test_mcp_tools_routing.py::test_prefix_collision_routes_each_tool_to_its_own_server`:

- Mocks `MultiServerMCPClient.get_tools(server_name=...)` to return `web_search` for `web` and `web_scraper_search` for `web_scraper`.
- Captures every `_make_session_pool_tool(tool, server_name, ...)` call.
- Asserts `web_scraper_search` is routed to `web_scraper`, not `web`.

Verified the test catches the bug — it FAILS against the pre-fix code:

```
E       AssertionError: assert ('web_scraper_search', 'web_scraper') in [('web_search', 'web'), ('web_scraper_search', 'web')]
tests/test_mcp_tools_routing.py:79: AssertionError
```

And PASSES with the fix.

Full local MCP test suite (no regressions):

```
uv run pytest tests/test_mcp_session_pool.py tests/test_mcp_custom_interceptors.py tests/test_mcp_tools_routing.py tests/test_mcp_client_config.py
======================== 60 passed, 1 warning in 1.48s =========================
```

### DIFF
--- a/backend/packages/harness/deerflow/mcp/tools.py
+++ b/backend/packages/harness/deerflow/mcp/tools.py
@@ -627,6 +627,8 @@ async def get_mcp_tools() -> list[BaseTool]:
 
         # Get tools from each server independently so one broken MCP server does
         # not prevent healthy servers from contributing their tools.
+        # ``tools_by_server[i]`` aligns with the i-th server in
+        # ``servers_config`` because ``asyncio.gather`` preserves input order.
         tools_by_server = await asyncio.gather(*(load_server_tools(name) for name in servers_config))
         tools = [tool for server_tools in tools_by_server for tool in server_tools]
         logger.info(f"Successfully loaded {len(tools)} tool(s) from MCP servers")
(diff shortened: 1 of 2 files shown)


WillemJiang:
@ly-wang19  thanks for your contribution. Please resolve the conflicts with the main branch and fix the unit test errors.

WillemJiang:
The issue is fixed in #3812

PR #3842, opened 2026-06-28
fix(setup): fail-safe to init_admin on rate-limited setup-status

### LINKED ISSUE #2999: First-boot /setup can redirect to /login because setup-status is rate limited
## Summary
On a fresh instance with no admin account, the gateway logs instruct the operator to visit `/setup`, but the browser can land on the normal login form instead of the "Create admin account" form.

The setup flow depends on repeated calls to `/api/v1/auth/setup-status`, while the backend allows only one setup-status request per client IP per 60 seconds. SSR auth, `/setup`, and `/login` can all call this endpoint during the same first-boot navigation. In dev, React Strict Mode can also run the client effect twice. A `429` response is then treated by the frontend as if setup is not required, causing a redirect to `/login`.

## Reproduction
Validated on latest `upstream/main` at `c0233cae`.

Environment:
- macOS
- Node.js `25.8.0`
- pnpm `10.6.5`
- uv `0.10.8`
- nginx `1.29.6`

Steps:
1. Start a fresh instance with no admin account:
   ```bash
   ./scripts/serve.sh --dev --skip-install
   ```
2. Gateway logs show:
   ```text
   First boot detected — no admin account exists.
   Visit /setup to complete admin account creation.
   ```
3. Open `http://localhost:2026/setup` after a recent setup-status check, or navigate through `/workspace` / `/login` first.
4. Observe the normal "Sign in to your account" page instead of the admin initialization form.

Observed API behavior:
```text
GET /api/v1/auth/setup-status -> 200 {"needs_setup":true}
GET /api/v1/auth/setup-status -> 429 {"detail":"Setup status check is rate limited"}
```

Gateway logs during the browser check showed repeated `429 Too Many Requests` for `/api/v1/auth/setup-status`.

## Relevant Code
- `frontend/src/core/auth/server.ts` calls `/setup-status` during SSR for unauthenticated requests.
- `frontend/src/app/(auth)/setup/page.tsx` calls `/setup-status` again and redirects to `/login` when the parsed JSON does not contain `needs_setup: true`.
- `frontend/src/app/(auth)/login/page.tsx` also calls `/setup-status` and redirects to `/setup` when it sees `needs_setup`.
- `backend/app/gateway/routers/auth.py` limits `/setup-status` to one request per client IP per 60 seconds.

## Expected Behavior
First-boot setup should be reliably reachable. A rate-limited or failed setup-status check should not be interpreted as "the system is already initialized", and the setup flow should avoid consuming multiple cooldown slots for one browser navigation.

ly-wang19:
## Problem

Closes #2999.

The `/setup` page probes `/api/v1/auth/setup-status` on first boot to decide whether to render the init-admin form. On any non-ok response — most importantly `429 Too Many Requests` — the page parses the body as `{ needs_setup: undefined }`, treats the falsy value as \"system already initialized\", and redirects the operator to `/login`.

Combined with the backend rate-limit, this makes first-boot admin creation unreachable until the cooldown window expires: SSR `/setup`, the setup page itself, and (in React Strict Mode) the doubled client effect can each consume a probe slot.

## Why This Change Was Made

- **First-boot setup must be reliably reachable.** The cost of an unnecessary init-admin form is one extra page load; the cost of an incorrect `/login` redirect is silent first-boot breakage that requires waiting out the rate-limit window. The fail-safe direction is unambiguous.
- **No backend change required.** The backend already caches the stable \"initialized\" result and dedupes inflight requests; the remaining gap is purely on the frontend's handling of non-ok responses.
- **Testable decision logic.** Extracted the mapping into a pure helper (`decideSetupModeFromSetupStatus`) so the fail-safe table can be unit-tested independently of the page effect — mirrors the existing `gateway-offline-banner-helpers` pattern.

## User Impact

- Operators on first boot reach the init-admin form reliably, even when the setup-status endpoint returns 429 or the network is flaky.
- No change for users on an already-initialized system: a successful `200 { needs_setup: false }` still redirects to `/login` as before.

## Change

- `frontend/src/app/(auth)/setup/setup-status-helpers.ts` — new pure helper `decideSetupModeFromSetupStatus(ok, needsSetup)` returning `\"init_admin\" | \"redirect_login\"`. Returns `\"redirect_login\"` only when `ok && needsSetup === false`; otherwise `\"init_admin\"`.
- `frontend/src/app/(auth)/setup/page.tsx` — probe effect now:
  - Guards `r.ok` before reading the body.
  - Maps the outcome through the helper.
  - `.catch` now fails safe to `init_admin` instead of redirecting to `/login` (network error → still reachable).

## Evidence

```
$ pnpm test -- run tests/unit/app/setup/setup-status-helpers.test.ts
tests: 374
failedTests: 0
passedTests: 374
```

New regression tests in `tests/unit/app/setup/setup-status-helpers.test.ts`:

- `redirects to /login when the system is already set up` — preserves existing behavior on `ok && needsSetup === false`.
- `stays on init_admin when needs_setup is true` — preserves existing behavior on `ok && needsSetup === true`.
- **`fails safe to init_admin on a 429 rate-limited response`** — the regression case from #2999: `ok=false, needsSetup=undefined` → `\"init_admin\"`, not `\"redirect_login\"`.
- `fails safe to init_admin on any other non-ok response` — covers 5xx, network, etc.
- `fails safe to init_admin when the body lacks needs_setup` — covers 2xx with malformed body.

### Type / lint

```
$ pnpm typecheck  # clean
$ pnpm lint       # 0 errors (3 pre-existing warnings in unrelated files)
```

### DIFF
--- a/frontend/src/app/(auth)/setup/page.tsx
+++ b/frontend/src/app/(auth)/setup/page.tsx
@@ -11,6 +11,8 @@ import { getCsrfHeaders } from "@/core/api/fetcher";
 import { useAuth } from "@/core/auth/AuthProvider";
 import { parseAuthError } from "@/core/auth/types";
 
+import { decideSetupModeFromSetupStatus } from "./setup-status-helpers";
+
 type SetupMode = "loading" | "init_admin" | "change_password";
 
 export default function SetupPage() {
@@ -35,20 +37,33 @@ export default function SetupPage() {
     if (isAuthenticated && user?.needs_setup) {
       setMode("change_password");
     } else if (!isAuthenticated) {
-      // Check if the system has no users yet
+      // Check if the system has no users yet. Fail-safe to init_admin on
+      // any non-ok response (e.g. 429 rate-limited) so the operator can
+      // still reach first-boot setup — see #2999.
       void fetch("/api/v1/auth/setup-status")
-        .then((r) => r.json())
-        .then((data: { needs_setup?: boolean }) => {
+        .then(async (r) => {
+          let needsSetup: boolean | undefined;
+          if (r.ok) {
+            try {
+              const data = (await r.json()) as { needs_setup?: boolean };
+              needsSetup = data.needs_setup;
+            } catch {
+              needsSetup = undefined;
+            }
+          }
           if (cancelled) return;
-          if (data.needs_setup) {
-            setMode("init_admin");
-          } else {
-            // System already set up and user is not logged in — go to login
+          if (
+            decideSetupModeFromSetupStatus(r.ok, needsSetup) === "redirect_login"
+          ) {
             router.push("/login");
+          } else {
+            setMode("init_admin");
           }
         })
         .catch(() => {
-          if (!cancelled) router.push("/login");
+          // Network error: fail-safe to init_admin so the operator can
+          // still reach the setup form when the gateway is flaky — see #2999.
+          if (!cancelled) setMode("init_admin");
         });
     } else {
       // Authenticated but needs_setup is false — already set up
--- a/frontend/src/app/(auth)/setup/setup-status-helpers.ts
+++ b/frontend/src/app/(auth)/setup/setup-status-helpers.ts
@@ -1,0 +1,30 @@
+/**
+ * Decision helper for the /setup page's setup-status probe.
+ *
+ * Extracted to a pure module so the fail-safe behaviour can be unit-tested
+ * without rendering the page. See #2999: a rate-limited (429) or otherwise
+ * non-ok response used to be parsed as ``{ needs_setup: undefined }``, which
+ * is falsy and therefore sent the operator to /login instead of the
+ * init-admin form — breaking first-boot setup until the rate-limit window
+ * expired.
+ */
+
+export type SetupStatusDecision = "init_admin" | "redirect_login";
+
+/**
+ * Map a setup-status probe outcome to the page-level action.
+ *
+ * Fail-safe to ``init_admin`` whenever the response is non-ok or the body
+ * does not explicitly say setup is done. Showing the init-admin form
+ * unnecessarily is one extra page load; redirecting the operator away from
+ * first-boot setup silently breaks admin creation.
+ */
+export function decideSetupModeFromSetupStatus(
+  ok: boolean,
+  needsSetup: boolean | undefined,
+): SetupStatusDecision {
+  if (ok && needsSetup === false) {
+    return "redirect_login";
+  }
+  return "init_admin";
+}
--- a/frontend/tests/unit/app/setup/setup-status-helpers.test.ts
+++ b/frontend/tests/unit/app/setup/setup-status-helpers.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, it } from "@rstest/core";
+
+import { decideSetupModeFromSetupStatus } from "@/app/(auth)/setup/setup-status-helpers";
+
+describe("decideSetupModeFromSetupStatus", () => {
+  it("redirects to /login when the system is already set up", () => {
+    expect(decideSetupModeFromSetupStatus(true, false)).toBe("redirect_login");
+  });
+
+  it("stays on init_admin when needs_setup is true", () => {
+    expect(decideSetupModeFromSetupStatus(true, true)).toBe("init_admin");
+  });
+
+  it("fails safe to init_admin on a 429 rate-limited response", () => {
+    // Regression for #2999: 429 used to be parsed as needs_setup=undefined
+    // (falsy) and redirect the operator to /login, breaking first-boot.
+    expect(decideSetupModeFromSetupStatus(false, undefined)).toBe("init_admin");
+  });
+
+  it("fails safe to init_admin on any other non-ok response", () => {
+    expect(decideSetupModeFromSetupStatus(false, false)).toBe("init_admin");
+    expect(decideSetupModeFromSetupStatus(false, true)).toBe("init_admin");
+  });
+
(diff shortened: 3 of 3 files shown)


frontend/src/app/(auth)/setup/setup-status-helpers.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Map a setup-status probe outcome to the page-level action.
+ *
+ * Fail-safe to ``init_admin`` whenever the response is non-ok or the body
+ * does not explicitly say setup is done. Showing the init-admin form
+ * unnecessarily is one extra page load; redirecting the operator away from
+ * first-boot setup silently breaks admin creation.
+ */
+export function decideSetupModeFromSetupStatus(
+  ok: boolean,
```
willem-bd, line 23: The `ok` parameter is dead at the only production call site. `fetchSetupStatus` (`core/auth/setup.ts`) throws on any non-ok response — including 429 — so a 429 routes to the `.catch` in `page.tsx`, never reaching this helper (called only from `.then` with a hardcoded `true` at `page.tsx:51`).

Consequence: the test `fails safe to init_admin on a 429 rate-limited response` calls `decideSetupModeFromSetupStatus(false, undefined)` — a branch production can never reach — while the actual 429 fix (`.catch → setMode("init_admin")` at `page.tsx:61`) has **no unit test**. A future change reverting `.catch` to `router.replace("/login")` would reintroduce #2999 with every helper test still green.

Suggested fix: drop the `ok` parameter (reduce to `needsSetup === false ? "redirect_login" : "init_admin"` and delete the three `ok=false` tests), or have `fetchSetupStatus` return a discriminated `{ ok, data }` result the helper consumes — and add a page-effect test asserting the `.catch` fail-safe.

(Side note: the PR description's "Guards `r.ok` before reading the body" is slightly off — the guard already lives in `fetchSetupStatus`; the page hardcodes `true`.)

willem-bd (commented):
Review of PR #3842 — the fail-safe direction is correct and the core fix is sound: on a 429, `fetchSetupStatus` throws → `.catch` now shows the init form, and the backend `initialize` endpoint guards actual admin creation (`system_already_initialized`). The findings below are about test-confidence and fix depth, not a correctness break.

1. **Dead `ok` param / misaimed 429 test** (inline on `setup-status-helpers.ts:23`) — `ok` is hardcoded `true` at the only call site (429 throws → `.catch`, never the helper), so the "429" regression test exercises an unreachable branch and the real `.catch` fix is untested.
2. **`/login` sibling 429 gap** (inline on `page.tsx:61`) — the fix is page-local; `/login` (the other `fetchSetupStatus` caller) misreads a 429 as "no admin setup needed". The deeper fix belongs in the shared fetcher.
3. **Test path drops the `(auth)` route group** (inline on the test file) — deviates from AGENTS.md's "mirror the `src/` layout" rule.

frontend/src/app/(auth)/setup/page.tsx
```diff
@@ -39,19 +41,24 @@ export default function SetupPage() {
...
+          ) {
             // System already set up and user is not logged in — go to login
             router.replace("/login");
+          } else {
+            setMode("init_admin");
           }
         })
         .catch(() => {
-          if (!cancelled) router.replace("/login");
+          if (!cancelled) setMode("init_admin");
```
willem-bd, line 61: This `.catch → setMode("init_admin")` is the correct fix for #2999 (429 throws in `fetchSetupStatus` → lands here). One altitude gap: this is page-local, but `/login` is a second caller of the shared `fetchSetupStatus` (`login/page.tsx:106`) and isn't fixed here.

On a 429, `/login`'s `.catch` does `setSetupStatus(null)` (`login/page.tsx:116`), which derives `regularSignupAllowed = canCreateRegularAccount({checked:true, status:null}) → true` (signup UI shown) and `systemNeedsAdminSetup = setupStatus?.needs_setup === true → false` (admin-setup nudge hidden) — i.e. `/login` interprets a rate-limited response as "no admin setup needed". Same root cause (`fetchSetupStatus` throws on non-ok, so each caller hand-rolls its `.catch`); `/setup` fixed, `/login` not.

Worth a follow-up: move the decision into the shared layer — colocate the helper in `core/auth/setup.ts` next to `canCreateRegularAccount` (which both pages already import) and have `fetchSetupStatus` expose a non-throwing discriminated result so the fail-safe is defined and tested once. (Trigger is narrow — first boot + an external 429 limiter, since the backend `setup-status` itself avoids 429 via cache/inflight dedup — but it's the exact scenario #2999 describes.)

frontend/tests/unit/app/setup/setup-status-helpers.test.ts
```diff
@@ -0,0 +1,29 @@
+import { describe, expect, it } from "@rstest/core";
```
willem-bd, line 1: Minor: the test path drops the `(auth)` route-group segment. `frontend/AGENTS.md` says unit tests "mirror the `src/` layout" (e.g. `tests/unit/core/api/stream-mode.test.ts` ↔ `src/core/api/stream-mode.ts`). The source is `src/app/(auth)/setup/setup-status-helpers.ts`, so the strict mirror is `tests/unit/app/(auth)/setup/setup-status-helpers.test.ts`. This is the first test under `tests/unit/app/`, so there's no precedent for eliding route groups — `find tests/unit/app/'(auth)'` won't locate it.

If dropping the parens is intentional for app-router route groups, a one-line note in AGENTS.md would make it explicit.

WillemJiang:
@ly-wang19  Please take a look at the review comments.

WillemJiang:
The issue has been addressed by #4371

Issue #3203, opened 2026-05-25
MCP工具调用完毕后系统报错：RuntimeError: athrow(): asynchronous generator is already running


htty326:
我使用fastmcp开发了工具，并在基于deerflow开发的智能体上通过配置extensions_config.json的方式使用MCP工具接入。在使用v2.0-m1-rc1代码时，智能体运行正常。但是使用fix-3127的代码时，工具调用完毕会抛出如下异常：
2026-05-25 02:50:10 - asyncio - ERROR - an error occurred during closing of asynchronous generator <async_generator object _create_streamable_http_session at 0x7fb30c18e110>
asyncgen: <async_generator object _create_streamable_http_session at 0x7fb30c18e110>
  + Exception Group Traceback (most recent call last):
  |   File "/code/QAgent_dev/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 799, in __aexit__
  |     raise BaseExceptionGroup(
  |         "unhandled errors in a TaskGroup", self._exceptions
  |     ) from None
  | BaseExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "/code/QAgent_dev/.venv/lib/python3.13/site-packages/langchain_mcp_adapters/sessions.py", line 360, in _create_streamable_http_session
    |     yield session
    | GeneratorExit
    +------------------------------------

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/code/QAgent_dev/.venv/lib/python3.13/site-packages/langchain_mcp_adapters/sessions.py", line 358, in _create_streamable_http_session
    ClientSession(read, write, **(session_kwargs or {})) as session,
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/code/QAgent_dev/.venv/lib/python3.13/site-packages/mcp/shared/session.py", line 238, in __aexit__
    return await self._task_group.__aexit__(exc_type, exc_val, exc_tb)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/code/QAgent_dev/.venv/lib/python3.13/site-packages/anyio/_backends/_asyncio.py", line 805, in __aexit__
[...]

PR #3843, opened 2026-06-28
feat(mcp): add per-server tool_call_timeout for MCP tool calls

### LINKED ISSUE #3840: feat: MCP tool call timeout support
## Problem

MCP tool calls have no timeout mechanism. When an MCP server becomes unresponsive (e.g., rate-limited HTTP API, hung subprocess), the agent run blocks indefinitely with no way to recover.

**Reproduction**: Configure an MCP server that calls an external API which rate-limits or hangs. The agent will wait forever on `session.call_tool()` with no timeout.

## Proposed Solution

Add a per-server `timeout` field to `McpServerConfig` in `extensions_config.json`. When set, wrap `session.call_tool()` with `asyncio.wait_for()` to enforce the timeout.

### Changes (3 files)

1. **`backend/packages/harness/deerflow/config/extensions_config.py`** — Add `timeout: float | None` field to `McpServerConfig`

2. **`backend/packages/harness/deerflow/mcp/client.py`** — Pass `timeout` through `build_server_params()` so it reaches the server config dict

3. **`backend/packages/harness/deerflow/mcp/tools.py`** — Accept `tool_call_timeout` parameter in `_make_session_pool_tool()` and wrap both `session.call_tool()` paths with `asyncio.wait_for(coro, timeout=...)`

### Config example

```json
{
  "mcpServers": {
    "biomcp": {
      "enabled": true,
      "type": "stdio",
      "command": "biomcp",
      "args": ["serve"],
      "timeout": 60
    }
  }
}
```

### Design notes

- `timeout: null` (default) = no timeout, preserving existing behavior
- The timeout wraps individual `call_tool` invocations, not the entire session
- `asyncio.wait_for` raises `asyncio.TimeoutError` which the existing error handling middleware converts to a user-facing error message

Happy to submit a PR if this approach looks good.

bjtolo:
## Summary

- Add `tool_call_timeout` field to `McpServerConfig` for per-server MCP tool call timeout
- Uses MCP SDK's built-in `read_timeout_seconds` on `ClientSession.call_tool` (avoids cross-task cancellation with session pool, ref #3379, #3203)
- Field named `tool_call_timeout` (not `timeout`) to avoid collision with langchain-mcp-adapters' existing `timeout` field on HTTP/SSE connections

Closes #3840

### DIFF
--- a/backend/packages/harness/deerflow/config/extensions_config.py
+++ b/backend/packages/harness/deerflow/config/extensions_config.py
@@ -47,7 +47,7 @@ class McpServerConfig(BaseModel):
     description: str = Field(default="", description="Human-readable description of what this MCP server provides")
     tool_call_timeout: float | None = Field(
         default=None,
-        description="Timeout in seconds for individual MCP tool calls. Uses the SDK's read_timeout_seconds to avoid cross-task cancellation. None means no timeout.",
+        description="Timeout in seconds for individual stdio MCP tool calls. HTTP/SSE servers use transport-level timeouts. None means no timeout.",
     )
     model_config = ConfigDict(extra="allow")
 
--- a/backend/packages/harness/deerflow/config/extensions_config.py
+++ b/backend/packages/harness/deerflow/config/extensions_config.py
@@ -45,6 +45,10 @@ class McpServerConfig(BaseModel):
     headers: dict[str, str] = Field(default_factory=dict, description="HTTP headers to send (for sse or http type)")
     oauth: McpOAuthConfig | None = Field(default=None, description="OAuth configuration (for sse or http type)")
     description: str = Field(default="", description="Human-readable description of what this MCP server provides")
+    tool_call_timeout: float | None = Field(
+        default=None,
+        description="Timeout in seconds for individual MCP tool calls. Uses the SDK's read_timeout_seconds to avoid cross-task cancellation. None means no timeout.",
+    )
     model_config = ConfigDict(extra="allow")
 
     @model_validator(mode="before")
--- a/backend/packages/harness/deerflow/mcp/client.py
+++ b/backend/packages/harness/deerflow/mcp/client.py
@@ -21,6 +21,9 @@ def build_server_params(server_name: str, config: McpServerConfig) -> dict[str, 
     transport_type = config.type or "stdio"
     params: dict[str, Any] = {"transport": transport_type}
 
+    if config.tool_call_timeout is not None:
+        params["tool_call_timeout"] = config.tool_call_timeout
+
     if transport_type == "stdio":
         if not config.command:
             raise ValueError(f"MCP server '{server_name}' with stdio transport requires 'command' field")
--- a/backend/packages/harness/deerflow/mcp/client.py
+++ b/backend/packages/harness/deerflow/mcp/client.py
@@ -21,9 +21,6 @@ def build_server_params(server_name: str, config: McpServerConfig) -> dict[str, 
     transport_type = config.type or "stdio"
     params: dict[str, Any] = {"transport": transport_type}
 
-    if config.tool_call_timeout is not None:
-        params["tool_call_timeout"] = config.tool_call_timeout
-
     if transport_type == "stdio":
         if not config.command:
             raise ValueError(f"MCP server '{server_name}' with stdio transport requires 'command' field")
--- a/backend/packages/harness/deerflow/mcp/tools.py
+++ b/backend/packages/harness/deerflow/mcp/tools.py
@@ -666,6 +666,13 @@ async def get_mcp_tools() -> list[BaseTool]:
                     _timeout = server_cfg.tool_call_timeout if server_cfg else None
                     wrapped_tools.append(_make_session_pool_tool(tool, tool_server, servers_config[tool_server], tool_interceptors, tool_call_timeout=_timeout))
                 else:
+                    server_cfg = extensions_config.mcp_servers.get(tool_server)
+                    if server_cfg and server_cfg.tool_call_timeout is not None:
+                        logger.warning(
+                            "Ignoring tool_call_timeout for MCP server '%s' because transport '%s' is not stdio; configure HTTP/SSE transport-level timeouts instead.",
+                            tool_server,
+                            transport,
+                        )
                     wrapped_tools.append(tool)
             else:
                 wrapped_tools.append(tool)
--- a/backend/packages/harness/deerflow/mcp/tools.py
+++ b/backend/packages/harness/deerflow/mcp/tools.py
@@ -480,23 +480,28 @@ def _make_session_pool_tool(
             session_connection["env"] = session_env
         session = await pool.get_session(server_name, scope_key, session_connection)
 
+        # Build common call_tool kwargs once — only add keys when needed so
+        # existing call-sites that assert on exact arguments are not affected.
+        call_kwargs: dict[str, Any] = {}
+        if tool_call_timeout:
+            call_kwargs["read_timeout_seconds"] = timedelta(seconds=tool_call_timeout)
+
         if tool_interceptors:
             from langchain_mcp_adapters.interceptors import MCPToolCallRequest
 
             async def base_handler(request: MCPToolCallRequest) -> Any:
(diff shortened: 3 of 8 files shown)


backend/packages/harness/deerflow/mcp/client.py
```diff
@@ -21,6 +21,9 @@ def build_server_params(server_name: str, config: McpServerConfig) -> dict[str,
     transport_type = config.type or "stdio"
     params: dict[str, Any] = {"transport": transport_type}
 
+    if config.tool_call_timeout is not None:
+        params["tool_call_timeout"] = config.tool_call_timeout
```
fancyboi999: **P1 root cause.** `params` is the connection dict passed to `MultiServerMCPClient` and then to `langchain_mcp_adapters` `create_session`, which forwards `**params` (minus `transport`) into `_create_stdio_session` / `_create_sse_session`. Those signatures have no `tool_call_timeout`, so session creation raises `TypeError` and the server's tools fail to load (confirmed against the installed adapters). Don't put the timeout in this dict — carry it separately and read it where you build the wrapper.

backend/packages/harness/deerflow/mcp/tools.py
```diff
@@ -646,7 +657,8 @@ async def load_server_tools(server_name: str) -> list[BaseTool]:
             if tool_server is not None:
                 transport = servers_config[tool_server].get("transport", "stdio")
                 if transport == "stdio":
-                    wrapped_tools.append(_make_session_pool_tool(tool, tool_server, servers_config[tool_server], tool_interceptors))
+                    _timeout = servers_config[tool_server].get("tool_call_timeout")
+                    wrapped_tools.append(_make_session_pool_tool(tool, tool_server, servers_config[tool_server], tool_interceptors, tool_call_timeout=_timeout))
```
fancyboi999, line 667: Same bug on the execution path, so fixing only discovery isn't enough: this passes `servers_config[tool_server]` (now containing `tool_call_timeout`) into `_make_session_pool_tool`, where `session_connection = dict(connection)` carries it into `pool.get_session(...)` -> `create_session()` in `session_pool.py`. Once the timeout is out of the connection dict, source `_timeout` from the `McpServerConfig`/extensions config rather than `servers_config[tool_server]`.

fancyboi999 (commented):
**P1 (breaks the feature it adds):** routing `tool_call_timeout` through the per-server connection dict makes any **stdio** MCP server that sets it fail to load its tools.

The dict `build_server_params` returns is consumed by `langchain_mcp_adapters.sessions.create_session` in two places: tool discovery (`MultiServerMCPClient.get_tools(server_name=...)` inside `get_mcp_tools`) and the session pool (`session_pool.py` -> `create_session(connection)`). `create_session` does `params = {k: v for k, v in connection.items() if k != "transport"}` and then `_create_stdio_session(**params)` (same for the SSE/HTTP builders), none of which accept `tool_call_timeout`, so it raises:

```
TypeError: _create_stdio_session() got an unexpected keyword argument 'tool_call_timeout'
```

Reproduced against the installed adapters with the exact dict shape your `build_server_params` now produces:

```python
from langchain_mcp_adapters.sessions import create_session
conn = {"transport": "stdio", "tool_call_timeout": 60.0, "command": "echo", "args": ["hi"]}
async with create_session(conn) as s:
    ...   # -> TypeError above
```

Discovery runs first, so the stdio server's tools never load at all — the servers this feature targets (the issue's `biomcp` example is stdio) are exactly the ones it breaks. `ConfigDict(extra="allow")` doesn't help; the key still reaches `create_session`.

The `read_timeout_seconds` call-site changes are right. The only problem is that the timeout value rides in the connection dict langchain sees. Keep it out of that dict: read it from the `McpServerConfig` (or a separate `{server: timeout}` map) where you build the wrapper, and drop the `build_server_params` change.

There are also no tests here. A regression that builds tools for an stdio server with `tool_call_timeout` set would have caught the `TypeError`, and is worth adding.

bjtolo:
## P1 Fix: `tool_call_timeout` removed from connection dict

**Root cause**: `build_server_params()` was putting `tool_call_timeout` into the connection dict that `langchain-mcp-adapters`'s `MultiServerMCPClient` passes to `create_session()`. For stdio transports, this dict becomes kwargs to `_create_stdio_session()`, which doesn't accept `tool_call_timeout` → `TypeError`.

**Fix** (2 files):
1. **`client.py`**: Removed the `tool_call_timeout` passthrough from `build_server_params()`. The connection dict now only contains transport-recognized keys (`transport`, `command`, `args`, `env`, `url`, `headers`).

2. **`tools.py`**: Changed the call-site to read `tool_call_timeout` directly from `ExtensionsConfig.mcp_servers[server_name]` (the Pydantic model) instead of from `servers_config[server].get("tool_call_timeout")` (the connection dict). This is the same `extensions_config` object already loaded at the top of `get_mcp_tools()`.

The timeout value never enters the connection dict — it's read from the config model where the wrapper is built and passed directly to `_make_session_pool_tool()`, which applies it via `session.call_tool(read_timeout_seconds=timedelta(...))`.

WillemJiang:
@bjtolo, to address the CLA assistant's complaint, please make sure your GitHub profile email settings use the git author email for your commits.

backend/packages/harness/deerflow/mcp/tools.py
```diff
@@ -646,7 +662,9 @@ async def load_server_tools(server_name: str) -> list[BaseTool]:
             if tool_server is not None:
                 transport = servers_config[tool_server].get("transport", "stdio")
                 if transport == "stdio":
-                    wrapped_tools.append(_make_session_pool_tool(tool, tool_server, servers_config[tool_server], tool_interceptors))
+                    server_cfg = extensions_config.mcp_servers.get(tool_server)
+                    _timeout = server_cfg.tool_call_timeout if server_cfg else None
+                    wrapped_tools.append(_make_session_pool_tool(tool, tool_server, servers_config[tool_server], tool_interceptors, tool_call_timeout=_timeout))
```
willem-bd, line 667: `tool_call_timeout` is silently a no-op for `sse`/`http` servers. It's only threaded through `_make_session_pool_tool`, which is only called on the `transport == "stdio"` branch; the `else` branch below appends the raw tool untouched. A user who sets `tool_call_timeout` on an http/sse server therefore gets no timeout and no warning.

The field description ("Timeout in seconds for individual MCP tool calls … None means no timeout.") doesn't mention a stdio-only restriction, so this is a UX trap. The PR body explains the intent (http/sse should use langchain's connection-level `timeout`), but that reasoning lives only in the PR — not in the field docs or any validation. Consider either logging a warning when `tool_call_timeout` is set on a non-stdio server, or stating the stdio-only scope in the field `description`.

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -45,6 +45,10 @@ class McpServerConfig(BaseModel):
     headers: dict[str, str] = Field(default_factory=dict, description="HTTP headers to send (for sse or http type)")
     oauth: McpOAuthConfig | None = Field(default=None, description="OAuth configuration (for sse or http type)")
     description: str = Field(default="", description="Human-readable description of what this MCP server provides")
+    tool_call_timeout: float | None = Field(
```
willem-bd, line 48: This is a new operator-facing config field, but `grep` finds no reference to `tool_call_timeout` in `docs/`, the READMEs, or any `extensions_config` example. Per the repo's cross-cutting convention in `AGENTS.md` ("keep docs in sync with code: update `README.md` for user-facing changes … in the same change set"), a new MCP config knob should be documented where the other `McpServerConfig` fields are surfaced to users. Low severity — worth a follow-up doc line, ideally noting the stdio-only scope from the comment above.

willem-bd (commented):
Reviewed the per-server `tool_call_timeout` change. The mechanism is sound — I confirmed `ClientSession.call_tool` accepts `read_timeout_seconds: timedelta | None` in the installed MCP SDK, that `build_server_params` whitelists fields so the new field can't leak into the langchain connection dict, and that `extensions_config.mcp_servers.get(...)` resolves the config correctly. No correctness defects survived verification, and the regression tests cover the leak path directly. Two behavior/docs notes inline.

WillemJiang (approved):

PR #3845, opened 2026-06-28
feat(subagents): persist and display subagent step history (#3779)

### LINKED ISSUE #3779: 子任务卡片执行过程展示不完整、无持久化，无法回溯

ggnnggez:
## Summary

Full-stack implementation of subagent (subtask) **step history** for #3779. The subtask card now shows a subagent's full step timeline — its assistant reasoning turns interleaved with the tools it ran — **both live and after a page reload**. Previously the card kept only `latestMessage`, so earlier steps flashed by and nothing survived a refresh.

## What & why

**Backend**
- `subagents/executor.py` captures both assistant turns (`AIMessage`) **and** tool outputs (`ToolMessage`) during subagent execution.
- New pure `subagents/step_events.py` (`capture_step_message` / `build_subagent_step` / `subagent_run_event`), shared by the live stream and persistence, so both agree on what a "step" is.
- `runtime/runs/worker.py` persists `task_*` custom events to the `RunEventStore` as `subagent.start` / `subagent.step` / `subagent.end` (`category="subagent"` keeps them out of the thread feed; `list_events` returns them for fetch-on-expand backfill).
- `list_events` gains `task_id` filtering + an `after_seq` forward cursor (all three stores + abstract base + the `/events` route) so the card pages through **one** subagent's steps without the run-wide `limit` truncating the tail. No schema migration — the DB filter rides the existing run-scoped index via `event_metadata["task_id"].as_string()`.

**Frontend**
- `core/tasks/steps.ts` pure model: `messageToStep` (live), `eventsToSteps` (reload), `mergeSteps` (dedup by `message_index`), `stepsForDisplay` (what the card renders).
- The subtask card accumulates live steps from `task_running` and **backfills on expand** for historical runs via `fetchSubtaskSteps` (task-scoped, paged with `after_seq`). It renders AI reasoning turns + tool steps; blank-text AI turns are dropped, and the trailing final-answer AI turn is dropped when completed (already shown as `task.result`).
- `run_id` is carried onto history content messages in `buildVisibleHistoryMessages` so the card can resolve the events endpoint.

## Relationship to #3831

#3831 also targets #3779 but is **frontend-only and explicitly defers persistence** ("does not claim to solve refresh/replay persistence; the issue thread calls that out as a larger backend/storage design"). This PR implements exactly that persistence/reload part — plus capturing tool **outputs**, not just tool-call requests — so it's a superset and overlaps the same files.

## Testing

- **Backend**: `task_id`/`after_seq` filtering + cursor pagination across memory/db/jsonl stores, `/events` route forwarding, `subagent_run_event`/`capture_step_message`, worker persistence — all green.
- **Frontend**: `stepsForDisplay`, `fetchSubtaskSteps` pagination, `eventsToSteps`/`mergeSteps`, message-merge — `pnpm test` 375 passed; `pnpm check` (eslint + tsc) clean.
- Docs updated in `backend/AGENTS.md` and `frontend/AGENTS.md`.

Closes #3779


backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -628,28 +629,13 @@ async def _aexecute(self, task: str, result_holder: SubagentResult | None = None
...
-                        else:
-                            # id-less messages can't be keyed; fall back to a full-dict compare
-                            is_duplicate = message_dict in ai_messages
-
-                        if not is_duplicate:
-                            ai_messages.append(message_dict)
-                            if message_id:
-                                seen_message_ids.add(message_id)
-                            logger.info(f"[trace={self.trace_id}] Subagent {self.config.name} captured AI message #{len(ai_messages)}")
+                if messages and capture_step_message(messages[-1], ai_messages, seen_message_ids):
```
willem-bd: **A multi-tool-call turn loses all but the last tool output.**

This loop runs on `stream_mode="values"` (full-state re-yield) and captures only `messages[-1]` per chunk. When the model emits several tool calls in one turn, LangGraph's `ToolNode` appends all their `ToolMessage`s in a single super-step, so only the last one is ever at `[-1]`. The earlier tool outputs are permanently missed — in both the live `task_running` stream and the persisted `subagent.step` history.

_Scenario:_ a subagent runs 3 tool calls in one turn → the card shows only the 3rd tool's result, silently dropping the other two. This is a regression introduced by capturing `ToolMessage`s (pre-PR only `AIMessage`s were captured, and each super-step yields exactly one, so `[-1]` was safe), and it undercuts the PR's stated goal of showing "what each step produced".

_Fix:_ iterate over the newly-appended tail (messages beyond the last-seen count) instead of only `messages[-1]`, capturing each new `AIMessage`/`ToolMessage`.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -118,6 +118,36 @@ def _agent_factory_supports_app_config(agent_factory: Any) -> bool:
...
+    # package __init__ (executor → agents → tools → task_tool), which imports
+    # back from deerflow.subagents and deadlocks at gateway startup. Deferring it
+    # to call time (after all modules are loaded) breaks that cycle.
+    from deerflow.subagents.step_events import subagent_run_event
+
+    record = subagent_run_event(chunk)
+    if record is None:
+        return
+    try:
+        await event_store.put(thread_id=thread_id, run_id=run_id, **record)
```
willem-bd: **Persisting each step with `put()` violates the store's documented low-frequency contract.**

`DbRunEventStore.put()`'s own docstring: _"Write a single event — low-frequency path only… Currently the only caller is `worker.run_agent` for the initial `human_message` event (once per run). For high-throughput writes use `put_batch`, which acquires the lock once for the whole batch."_

`_persist_subagent_event` now calls `put()` once per `task_running` step on the streaming loop. On Postgres each `put()` opens a dedicated transaction and takes a per-thread `pg_advisory_xact_lock`.

_Scenario:_ a deep subagent (`general-purpose max_turns=150`) generates hundreds of individually-locked transactions on the hot stream path, serializing against the run's own message-batch writer — lock contention and added stream latency exactly where the API says to batch. Consider buffering steps and flushing via `put_batch`, or a dedicated low-lock append path for `category="subagent"`.

backend/packages/harness/deerflow/subagents/step_events.py
```diff
@@ -0,0 +1,191 @@
...
+        "message_index": message_index,
+        "kind": kind,
+        "text": text,
+        "truncated": truncated,
+    }
+
+    if kind == "tool":
+        step["tool_name"] = message.get("name")
+    else:
+        step["tool_calls"] = [{"name": call.get("name"), "args": call.get("args")} for call in (message.get("tool_calls") or [])]
```
willem-bd: **`tool_calls` args are never truncated, so `subagent.step` rows are unbounded.**

`build_subagent_step` caps only `text` (to `SUBAGENT_STEP_MAX_CHARS`); this AI branch copies `tool_calls: [{name, args}]` verbatim. `DbRunEventStore._truncate_trace` byte-caps only `category == "trace"`, not `"subagent"`, so nothing else bounds the row.

_Scenario:_ a subagent calls `write_file`/`edit_file`/`bash` with a large argument payload (full file contents, a heredoc) → the persisted run-event row and the streamed frame carry the full args untruncated, defeating the row-bounding the `text` cap was added to provide. Consider capping serialized `args` (or total step size) the same way `text` is capped.

frontend/src/components/workspace/messages/subtask-card.tsx
```diff
@@ -32,16 +36,50 @@ import { MarkdownContent } from "./markdown-content";
...
+      return;
+    }
+    if (!threadId || !runId) {
+      return;
+    }
+    backfilledRef.current = true;
+    fetchSubtaskSteps(threadId, runId, taskId)
+      .then((steps) => {
+        if (steps.length > 0) {
+          updateSubtask({ id: taskId, steps });
```
willem-bd, line 75: **The backfill `.then(updateSubtask)` can clobber newer state via a stale closure.**

`useUpdateSubtask` mutates the `tasks` object in place then `setTasks({ ...tasks })`, and `updateSubtask` closes over the `tasks` snapshot from when this effect fired.

_Scenario:_ on reload of a still-active run, expanding a subtask that has 0 steps yet starts `fetchSubtaskSteps`; while it is in flight, SSE updates mutate later state copies; when the fetch resolves, the captured `updateSubtask` writes `setTasks({ ...staleTasks })`, discarding live steps/status that arrived during the window and dropping any sibling subtask (concurrent subagents, up to 3) added after the effect fired. Narrow trigger and it self-heals on the next SSE tick, but the card visibly reverts.

_Fix:_ use the functional-update / ref-to-latest form so the resolved fetch merges into current state rather than the captured snapshot.

backend/packages/harness/deerflow/subagents/step_events.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def truncate_step_text(text: str, max_chars: int) -> tuple[str, bool]:
+    """Return ``(text, truncated)``, clipping to ``max_chars`` when longer."""
+    if max_chars >= 0 and len(text) > max_chars:
+        return text[:max_chars], True
+    return text, False
+
+
+def _content_to_text(content: Any) -> str:
```
willem-bd: **`_content_to_text` reimplements an existing helper.**

`deerflow.utils.messages.message_content_to_text` already does str-passthrough + join of list text-blocks and is import-safe from the subagents package. This private copy is duplicated logic that will drift (e.g. when a new content-block shape is added). Prefer calling the existing helper. (Minor: the frontend `contentToText` in `steps.ts` similarly overlaps `extractTextFromMessage`, though their signatures differ more.)

willem-bd (commented):
Automated review of the subagent step-history persistence. Five findings inline — correctness first, then a store-contract/perf issue, an unbounded-row issue, a frontend state race, and one reuse nit. The paging/filtering machinery (`after_seq` cursor, `task_id` SQL filter, `category`-based feed exclusion) all checked out clean.

backend/packages/harness/deerflow/subagents/executor.py
ggnnggez: Fixed in c8c45778. Replaced the `messages[-1]`-only capture with `capture_new_step_messages`, which walks the newly-appended tail of each `values`-mode chunk, so every `ToolMessage` from a multi-tool-call super-step is captured (both the live `task_running` stream and the persisted history derive from the same `ai_messages` list). It still re-checks the trailing message on no-growth chunks so the existing id-less in-place-replacement dedup case is preserved. Added an end-to-end `_aexecute` regression capturing 3 tool outputs from one super-step.

backend/packages/harness/deerflow/runtime/runs/worker.py
ggnnggez: Fixed in c8c45778. Replaced the per-step `put()` with `_SubagentEventBuffer`, which accumulates recognized subagent events and persists them via `put_batch` (one lock per batch). It flushes on a terminal `subagent.end` (so a completed subagent's history is durable promptly), at `FLUSH_THRESHOLD=25`, and in the worker `finally` so buffered steps survive abort/exception paths. Best-effort semantics preserved: missing store / unknown chunk are no-ops and flush failures are logged, never propagated into the stream loop.

backend/packages/harness/deerflow/subagents/step_events.py
ggnnggez: Fixed in c8c45778. Added `_bounded_tool_call`: each tool call's `args` is JSON-serialized and, when it exceeds `SUBAGENT_STEP_MAX_CHARS` (the same cap used for `text`), replaced with a truncated preview string flagged `args_truncated`; small args stay structured. This bounds the persisted `subagent.step` row and streamed frame for large `write_file`/`bash` payloads. (Confirmed the frontend subtask card renders `text`/`tool_name` but not step `args`, so no UI change is needed.)

frontend/src/components/workspace/messages/subtask-card.tsx
ggnnggez: Fixed in 1beb8e34. `useUpdateSubtask` no longer closes over a `tasks` snapshot — the provider keeps a `tasksRef` mirroring the latest state, and `updateSubtask` reads/writes through it (ref-to-latest), so a late-resolving `fetchSubtaskSteps().then(...)` merges into current state instead of writing a stale map. Concurrently-arrived SSE steps/status and sibling subtasks are preserved. The pure per-subtask transition (`computeNextSubtask`, incl. the merge-not-replace of step deltas and the stable-terminal guard) is extracted to `core/tasks/subtask-update.ts` with unit tests, including one for the "backfill fires at 0 steps, resolves after SSE added 1..2" case.

backend/packages/harness/deerflow/subagents/step_events.py
ggnnggez: Fixed in 1beb8e34. Removed the private `_content_to_text` and now call `deerflow.utils.messages.message_content_to_text`. One behavioral guard: the shared helper `str()`-ifies non-str/list content, so a tool-call-only turn with `content=None` would render as the literal "None" — passing `message.get("content") or ""` keeps it as "". (Left the frontend `contentToText`/`extractTextFromMessage` overlap alone as noted, since their signatures differ more.)

ggnnggez:
@willem-bd thanks for the thorough review — all five findings are addressed across two commits (c8c45778, 1beb8e34), each with tests and replies inline:

1. **Multi-tool step loss** (`executor.py`) — capture now walks the newly-appended tail of each `values`-mode chunk (via `capture_new_step_messages`) instead of only `messages[-1]`, so every `ToolMessage` from a multi-tool-call super-step survives in both the live stream and persisted history. Kept the trailing re-check so the id-less in-place-replacement dedup case still works.
2. **Store low-frequency contract** (`worker.py`) — replaced per-step `put()` with `_SubagentEventBuffer`, which batches via `put_batch` (flush on terminal `subagent.end`, at a size threshold, and in the worker `finally`).
3. **Unbounded `tool_calls` args** (`step_events.py`) — each call's serialized `args` is now capped at `SUBAGENT_STEP_MAX_CHARS` and flagged `args_truncated`, matching the `text` cap.
4. **Frontend stale-closure race** (`subtask-card`/`context`) — `useUpdateSubtask` now reads/writes through a `tasksRef` mirroring the latest state, so a late-resolving backfill merges into current state instead of clobbering concurrent SSE steps / sibling subtasks. Pure transition extracted to `computeNextSubtask` with unit tests.
5. **Reuse nit** — `_content_to_text` dropped in favor of the shared `message_content_to_text` (guarding `None` content so a tool-call-only turn stays `""`).

Backend + frontend test suites and lint are green. Thanks again!

WillemJiang (approved):

PR #3846, opened 2026-06-28
fix(mcp): route tools by authoritative server, not name-prefix match

### LINKED ISSUE #3811: bug: MCP tools mis-routed when one server name is a prefix of another

ly-wang19:
## What Problem This Solves

Closes #3811.

`get_mcp_tools()` (`backend/packages/harness/deerflow/mcp/tools.py`) flattened the per-server grouping from `asyncio.gather` and then re-derived each tool's source server by scanning `servers_config` for the first name whose `name_` prefix matched the tool name. With `tool_name_prefix=True`, when one server name is a prefix of another (e.g. `web` and `web_scraper`), a tool from the longer-named server is misrouted:

- `web_scraper_fetch`.startswith(`web_`) → True, so the tool matches `web` first
- It is pooled under `web` (wrong session)
- `_make_session_pool_tool` strips `web_` → `original_name = "scraper_fetch"` (wrong)
- The call hits the `web` session with a tool name that does not exist there → "tool not found"

For mixed transports, a stdio tool from the longer-named server can also lose session pooling when misrouted to an HTTP/SSE server.

The behavior was also order-dependent on config insertion order, so it could look intermittent.

## Why This Change Was Made

`tools_by_server` is the authoritative pairing: `asyncio.gather` preserves order, so index `i` corresponds to the `i`-th server in `servers_config`. Iterating `(server_name, server_tools)` pairs directly removes the ambiguous name-prefix inference and the order-dependence in one step.

The `_make_session_pool_tool` helper already strips the prefix `f"{server_name}_"` from the tool name (no-op if the tool is misnamed), so once each tool is paired with its producing server, both the session lookup and the original-name stripping are correct.

The "no matching server" fallthrough branch is no longer reachable (every tool is paired with its producing server) and is removed.

## User Impact

Operators with MCP servers whose names share a prefix (e.g. `github` and `github_enterprise`, `web` and `web_scraper`) will see tools routed to the correct server. Previously this was order-dependent and silently broken for the longer-named server's tools. No config change is required.

## Evidence

Local tests on top of `main`:

```
backend/.venv/bin/pytest tests/test_mcp_session_pool.py tests/test_mcp_custom_interceptors.py
# 47 passed
```

New regression test: `test_get_mcp_tools_routes_tools_by_authoritative_server_when_names_overlap`:

- Two stdio servers `web` and `web_scraper` (shorter first, to make a prefix-match regression visible).
- Asserts both tools are returned with their original prefixed names (`web_search`, `web_scraper_fetch`).
- Invokes the longer-named tool and asserts `mock_session.call_tool` is called once with `("fetch", {"url": ...})` — i.e. pooled under `web_scraper` with the correct stripped name, not under `web` as `scraper_fetch`.

On the unpatched code, the regression test fails because `call_tool` is invoked with `("scraper_fetch", ...)` against the wrong session.

### DIFF
--- a/backend/packages/harness/deerflow/mcp/tools.py
+++ b/backend/packages/harness/deerflow/mcp/tools.py
@@ -635,22 +635,24 @@ async def get_mcp_tools() -> list[BaseTool]:
         # Only pool stdio sessions. HTTP/SSE transports use anyio TaskGroups
         # internally which cannot be closed from a different async task, so
         # pooling them causes RuntimeError on cleanup (see #3203).
+        #
+        # Iterate (server_name, server_tools) pairs from `tools_by_server` to
+        # preserve the authoritative per-server grouping from `asyncio.gather`.
+        # Re-deriving the source server from the tool-name prefix is ambiguous
+        # when one server name is a prefix of another (e.g. `web` and
+        # `web_scraper`): `web_scraper_fetch` matches `web_` first, so the tool
+        # is pooled under the wrong server and invoked with the wrong stripped
+        # name (see #3811).
         wrapped_tools: list[BaseTool] = []
-        for tool in tools:
-            tool_server: str | None = None
-            for name in servers_config:
-                if tool.name.startswith(f"{name}_"):
-                    tool_server = name
-                    break
-
-            if tool_server is not None:
-                transport = servers_config[tool_server].get("transport", "stdio")
+        for server_name, server_tools in zip(servers_config, tools_by_server):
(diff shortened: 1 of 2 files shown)


ly-wang19:
Closing as a duplicate of #3841, which I opened earlier today with the same fix (walk servers and tools in lockstep via `zip(servers_config, tools_by_server)` instead of inferring the server from the tool-name prefix) and a slightly cleaner local-variable extraction. Apologies for the noise.

PR #3847, opened 2026-06-28
fix(guardrail): keep prompt-injection wrapper out of chat UI and titles (#3689)

### LINKED ISSUE #3689: [bug] Prompt-injection input wrapper leaks into chat UI and first_human_message

ly-wang19:
## What Problem This Solves

Closes #3689.

`InputSanitizationMiddleware` (issue #3630) wraps clean user input in boundary markers (`--- BEGIN USER INPUT ---` / `--- END USER INPUT ---`) as a structured-prompt defense. The wrap is supposed to be model-facing only, but the wrapper leaked into:

- The chat UI — a plain `test` is rendered as `--- BEGIN USER INPUT ---\ntest\n--- END USER INPUT ---`
- `runs.first_human_message` — so thread titles / previews are polluted
- The `llm.human.input` event row persisted by `RunJournal.on_chat_model_start`

The root cause is that the middleware replaces the user message's content at the LLM-call boundary, but the journal observes the wrapped message when capturing `first_human_message` and `GET /api/threads/{id}/messages` returns event-store content verbatim. Unlike `UploadsMiddleware` (which stamps the pre-transform text in `additional_kwargs[ORIGINAL_USER_CONTENT_KEY]`) and IM channels, a plain web text message was never stamped.

## Why This Change Was Made

The fix follows the same pattern already established by `UploadsMiddleware` and the IM channel path: stamp the pre-wrap user text in `ORIGINAL_USER_CONTENT_KEY`, then read it back at the surfaces that display human text.

Three coordinated changes:

1. **`InputSanitizationMiddleware._process_request`** — stamp `ORIGINAL_USER_CONTENT_KEY` with `message_content_to_text(content)` *before* constructing the wrapped message, using `setdefault` so an earlier stamp from `UploadsMiddleware` (or any other upstream middleware) wins. This is the same shape as `uploads_middleware.py:270`.

2. **`RunJournal.on_chat_model_start`** — extract `first_human_message` via `get_original_user_content_text(m.content, m.additional_kwargs)` instead of `m.text`, so thread titles/previews stay clean.

3. **`list_thread_messages`** — for `llm.human.input` event rows whose inner message carries `ORIGINAL_USER_CONTENT_KEY`, replace the inner content with the original text before returning so the chat UI renders user text, not boundary markers.

The persisted `llm.human.input` event still carries the wrapped content for parity with the model-facing stream; only the read path (and the title extraction) unwraps. This keeps the boundary defense intact for the model while ensuring UI surfaces show the user's actual text.

## User Impact

- Sending a plain text message in a fresh web thread no longer renders the boundary markers in the chat UI. The original text (`test`) is displayed instead of `--- BEGIN USER INPUT ---\ntest\n--- END USER INPUT ---`.
- Thread titles / previews (`first_human_message`) reflect the original text.
- The model still receives the protective wrapper; no change to guardrail behavior.
- Messages without `ORIGINAL_USER_CONTENT_KEY` (e.g. before this fix lands, or messages that bypass the middleware) pass through unchanged — fully backward compatible.

## Evidence

```
$ .venv/bin/python -m pytest \
    tests/test_input_sanitization_middleware.py \
    tests/test_run_journal.py \
    tests/test_thread_messages_feedback.py \
    tests/test_thread_run_messages_pagination.py \
    tests/test_thread_regenerate_prepare.py \
    tests/test_uploads_middleware_core_logic.py \
    tests/test_slash_skills.py \
    tests/test_dynamic_context_middleware.py
======================= 257 passed, 2 warnings in 2.96s ========================
```

Regression coverage added:

- `test_input_sanitization_middleware.py::TestOriginalUserContentStamp` (5 tests):
  - `test_stamps_original_text_when_wrapping` — basic wrap stamps `ORIGINAL_USER_CONTENT_KEY`
  - `test_stamps_pre_wrap_text_when_tags_escaped` — pre-escape, pre-wrap text preserved verbatim
  - `test_stamps_joined_text_for_list_content` — multimodal text blocks joined with newlines
  - `test_does_not_overwrite_existing_original_stamp` — earlier middleware stamp wins via `setdefault`
  - `test_skips_stamp_when_message_not_wrapped` — image-only / empty messages bypass stamping
- `test_run_journal.py::test_first_human_message_unwraps_prompt_injection_boundary` — `first_human_message` is `"test"`, not the wrapped text; persisted event retains wrapped content for parity
- `test_thread_run_messages_pagination.py`:
  - `test_list_thread_messages_unwraps_prompt_injection_boundary_markers` — UI sees original text, AI messages untouched
  - `test_list_thread_messages_leaves_unstamped_human_messages_untouched` — no regression for unstamped messages

### Tested surface

- `backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py` — wrap path stamps `ORIGINAL_USER_CONTENT_KEY` before constructing the wrapped HumanMessage
- `backend/packages/harness/deerflow/runtime/journal.py` — `first_human_message` extraction reads via `get_original_user_content_text`
- `backend/app/gateway/routers/thread_runs.py` — `list_thread_messages` unwraps human input events before returning

### DIFF
--- a/backend/app/gateway/routers/thread_runs.py
+++ b/backend/app/gateway/routers/thread_runs.py
@@ -25,7 +25,12 @@ from app.gateway.deps import get_checkpointer, get_current_user, get_feedback_re
 from app.gateway.pagination import trim_run_message_page
 from app.gateway.services import sse_consumer, start_run, wait_for_run_completion
 from deerflow.runtime import RunRecord, RunStatus, serialize_channel_values_for_api
-from deerflow.utils.messages import ORIGINAL_USER_CONTENT_KEY, get_original_user_content_text, message_to_text
+from deerflow.utils.messages import (
+    ORIGINAL_USER_CONTENT_KEY,
+    get_original_user_content_text,
+    message_to_text,
+    restore_original_user_content_blocks,
+)
 
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/api/threads", tags=["runs"])
@@ -247,13 +252,17 @@ def _checkpoint_response(checkpoint_tuple: Any) -> dict[str, Any]:
 
 def _clean_human_message_for_regenerate(message: Any) -> dict[str, Any]:
     additional_kwargs = _message_additional_kwargs(message)
-    content = get_original_user_content_text(_message_content(message), additional_kwargs)
+    raw_content = _message_content(message)
+    original_text = get_original_user_content_text(raw_content, additional_kwargs)
     additional_kwargs.pop(ORIGINAL_USER_CONTENT_KEY, None)
     additional_kwargs.pop("hide_from_ui", None)
 
     clean_message: dict[str, Any] = {
         "type": "human",
-        "content": [{"type": "text", "text": content}],
+        # Preserve non-text blocks (images/files) for multimodal messages by
+        # restoring through the merge helper instead of collapsing to a lone
+        # text block. See restore_original_user_content_blocks.
+        "content": restore_original_user_content_blocks(raw_content, original_text),
         "additional_kwargs": additional_kwargs,
     }
     message_id = _message_id(message)
@@ -639,7 +648,14 @@ async def list_thread_messages(
         original_content = additional_kwargs.get(ORIGINAL_USER_CONTENT_KEY)
         if not isinstance(original_content, str) or not original_content.strip():
             continue
-        inner["content"] = [{"type": "text", "text": original_content}]
+        # Preserve non-text blocks (images, files) for multimodal messages by
+        # restoring through the merge helper instead of collapsing to a lone
+        # text block. Without this an uploaded image disappears from history
+        # on reload — see InputSanitizationMiddleware._rebuild_content for the
+        # mirror-shape on the write path. (#3689 review feedback.)
+        inner["content"] = restore_original_user_content_blocks(
+            inner.get("content"), original_content
+        )
 
     # Resolve the caller once; it is needed both to scope the feedback query
     # below and to list the thread's runs for turn-duration injection.
--- a/backend/app/gateway/routers/thread_runs.py
+++ b/backend/app/gateway/routers/thread_runs.py
@@ -624,6 +624,23 @@ async def list_thread_messages(
     event_store = get_run_event_store(request)
     messages = await event_store.list_messages(thread_id, limit=limit, before_seq=before_seq, after_seq=after_seq)
 
+    # Unwrap prompt-injection boundary wrappers stamped by InputSanitizationMiddleware
+    # so the UI sees the original user text, not the model-facing wrapper. Only
+    # human-input events are affected; AI messages and tool flows pass through.
+    for row in messages:
+        if row.get("event_type") != "llm.human.input":
+            continue
+        inner = row.get("content")
+        if not isinstance(inner, dict) or inner.get("type") != "human":
+            continue
+        additional_kwargs = inner.get("additional_kwargs")
+        if not isinstance(additional_kwargs, dict):
+            continue
+        original_content = additional_kwargs.get(ORIGINAL_USER_CONTENT_KEY)
+        if not isinstance(original_content, str) or not original_content.strip():
+            continue
+        inner["content"] = [{"type": "text", "text": original_content}]
+
     # Resolve the caller once; it is needed both to scope the feedback query
     # below and to list the thread's runs for turn-duration injection.
     user_id = await get_current_user(request)
--- a/backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
@@ -31,6 +31,8 @@ from langchain.agents.middleware.types import (
 from langchain_core.messages import HumanMessage
 from langgraph.errors import GraphBubbleUp
 
+from deerflow.utils.messages import ORIGINAL_USER_CONTENT_KEY, message_content_to_text
+
 logger = logging.getLogger(__name__)
 
 _SUMMARY_MESSAGE_NAME = "summary"
@@ -233,11 +235,22 @@ class InputSanitizationMiddleware(AgentMiddleware[AgentState]):
             else:
                 new_content = processed
 
+            # Stamp the pre-wrap text so downstream surfaces (journal's
+            # first_human_message, GET /threads/{id}/messages) can unwrap
+            # the boundary markers via get_original_user_content_text.
+            # setdefault preserves an existing stamp from an earlier middleware
+            # (e.g. UploadsMiddleware) so the earliest pre-transform text wins.
+            new_additional_kwargs = dict(msg.additional_kwargs or {})
+            new_additional_kwargs.setdefault(
+                ORIGINAL_USER_CONTENT_KEY,
+                message_content_to_text(content),
+            )
+
             messages[i] = HumanMessage(
                 content=new_content,
                 id=msg.id,
                 name=msg.name,
-                additional_kwargs=msg.additional_kwargs,
+                additional_kwargs=new_additional_kwargs,
             )
             logger.debug(
                 "InputSanitizationMiddleware: original=%r -> processed=%r",
--- a/backend/packages/harness/deerflow/runtime/journal.py
+++ b/backend/packages/harness/deerflow/runtime/journal.py
@@ -29,7 +29,7 @@ from langchain_core.callbacks import BaseCallbackHandler
 from langchain_core.messages import AIMessage, AnyMessage, BaseMessage, HumanMessage, ToolMessage
 from langgraph.types import Command
 
-from deerflow.utils.messages import message_to_text
+from deerflow.utils.messages import get_original_user_content_text, message_to_text
 
 if TYPE_CHECKING:
     from deerflow.runtime.events.store.base import RunEventStore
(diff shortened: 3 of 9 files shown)


backend/app/gateway/routers/thread_runs.py
```diff
@@ -624,6 +624,23 @@ async def list_thread_messages(
...
+        inner = row.get("content")
+        if not isinstance(inner, dict) or inner.get("type") != "human":
+            continue
+        additional_kwargs = inner.get("additional_kwargs")
+        if not isinstance(additional_kwargs, dict):
+            continue
+        original_content = additional_kwargs.get(ORIGINAL_USER_CONTENT_KEY)
+        if not isinstance(original_content, str) or not original_content.strip():
+            continue
+        inner["content"] = [{"type": "text", "text": original_content}]
```
fancyboi999: **P2 (multimodal regression).** This replaces the whole inner content with one text block, but `ORIGINAL_USER_CONTENT_KEY` only holds the joined text — `message_content_to_text` drops `image_url`/file blocks. The middleware's `_rebuild_content` keeps those blocks in the persisted `llm.human.input` event, so a `[text, image_url]` message comes back as `[text]` here and the image vanishes from the chat UI on reload (same for `UploadsMiddleware`-stamped multimodal messages). When `inner["content"]` is a list, restore only the text block(s) and keep the non-text blocks (mirror `_rebuild_content`); do the wholesale replace only when the content is a plain string.

fancyboi999 (commented):
Nice approach overall — stamping `ORIGINAL_USER_CONTENT_KEY` and unwrapping at the display surfaces is the right shape, and the middleware change preserves image blocks via `_rebuild_content`. One real bug on the read path though.

**P2 — multimodal user messages lose their non-text blocks (images/files) in the messages API.** `ORIGINAL_USER_CONTENT_KEY` only holds the joined text (`message_content_to_text` in `utils/messages.py` extracts `item["text"]` and drops `image_url`/other blocks). But the unwrap in `list_thread_messages` replaces the **entire** inner content with a single text block. For a multimodal message the middleware's `_rebuild_content` keeps the image blocks in the persisted `llm.human.input` event, so this read path discards them. By your own root-cause analysis the chat UI renders this event content verbatim, so the user's image disappears from history on reload. It also hits any `UploadsMiddleware`-stamped multimodal message, since the unwrap block keys only on `ORIGINAL_USER_CONTENT_KEY`.

Reproduced the read-path block against a `[text, image_url]` human event:

```
before: [{type:text, text:"--- BEGIN USER INPUT ---\nlook\n--- END USER INPUT ---"}, {type:image_url, ...}]
after:  [{type:text, text:"look"}]   # image block gone
```

The read-path tests only cover string content, so this isn't caught. Fix direction: when `inner["content"]` is a list, restore only the text block(s) and keep the non-text blocks (the same preserve-interleaved-blocks logic `_rebuild_content` already uses); do the wholesale replace only when the content is a plain string. A multimodal case in `test_list_thread_messages_*` would lock it in.

ly-wang19:
@fancyboi999 Thanks for catching the multimodal regression — fix pushed in \`eef38ef6\`.

**Root cause on the read path:** the wholesale \`inner["content"] = [{"type": "text", "text": original_content}]\` discarded any non-text blocks the middleware's \`_rebuild_content\` had preserved alongside the merged text block.

**Fix:** added \`restore_original_user_content_blocks(content, original_text)\` in [\`deerflow/utils/messages.py\`](https://github.com/bytedance/deer-flow/blob/eef38ef613e205cb74889b9ad7bf14892c11db3a/backend/packages/harness/deerflow/utils/messages.py) that mirrors the write-path merge shape:

- String/non-list content → single \`[{text: original_text}]\` (unchanged shape for plain-text messages).
- List content → replace the first text block's text with \`original_text\`, drop subsequent text blocks (already merged into the first by \`_rebuild_content\`), preserve all non-text blocks (image_url, files) in place.
- No text block in the persisted list → prepend the original text.

Applied to both unwrap surfaces — \`list_thread_messages\` and \`_clean_human_message_for_regenerate\` — since the regenerate path had the same wholesale-collapse shape and would lose image blocks for any \`UploadsMiddleware\`-stamped multimodal message.

**Tests added:**
- \`test_utils_messages.py\`: 8 helper tests (string, non-list, first-text replace, drop-subsequent-text, inter-block non-text, no-text-block prepend, empty list, no-mutation contract).
- \`test_thread_run_messages_pagination.py\`: \`test_list_thread_messages_unwraps_multimodal_message_preserving_non_text_blocks\` — your reproduced \`[text, image_url]\` case, now asserts \`[{text: "look"}, {image_url: ...}]\`.
- \`test_thread_regenerate_prepare.py\`: \`test_prepare_regenerate_payload_preserves_image_block_for_multimodal_message\` — same shape on the regenerate path.

All 225 tests in the touched files pass (\`tests/test_utils_messages.py tests/test_thread_run_messages_pagination.py tests/test_thread_regenerate_prepare.py tests/test_input_sanitization_middleware.py tests/test_run_journal.py tests/test_uploads_middleware_core_logic.py\`).

Reproduced the read-path case from your comment against the new helper:

\`\`\`py
>>> restore_original_user_content_blocks(
...     [{"type": "text", "text": "--- BEGIN USER INPUT ---\\nlook\\n--- END USER INPUT ---"},
...      {"type": "image_url", "image_url": {"url": "data:..."}}],
...     "look",
... )
[{'type': 'text', 'text': 'look'}, {'type': 'image_url', 'image_url': {'url': 'data:...'}}]
\`\`\`

Image block stays in place.

backend/app/gateway/routers/thread_runs.py
fancyboi999: Verified on `eef38ef6` — this resolves the P2. `restore_original_user_content_blocks` (`utils/messages.py:77`) restores the first text block from `ORIGINAL_USER_CONTENT_KEY` while keeping non-text blocks in place, and it's wired through the **persisted** content at both unwrap surfaces — the read path here (`inner.get("content")`, `thread_runs.py:656`) and the regenerate path (`raw_content`, `:265`). Since both pass the `_rebuild_content`-preserved content instead of collapsing to a lone text block, the reproduced `[text, image_url]` case now comes back as `[{text}, {image_url}]`. Nice catch extending it to the regenerate path as well.

WillemJiang:
@ly-wang19, please resolve the conflicts with main branch and address the lint error.

WillemJiang:
The issue is addressed in https://github.com/bytedance/deer-flow/pull/4065

PR #3198, opened 2026-05-24
fix(frontend): show filename for presented artifacts missing from thr…

Huixin615:
## Summary

Fixes #3192. The artifact preview header shows an empty filename after an AI run completes, even though the preview body renders correctly.

## Root cause

The artifact preview header is driven by a Radix UI `<Select>` whose `<SelectValue>` renders the label of the `<SelectItem>` matching the current `value`. The option list was built solely from `artifacts` in `ArtifactsContext`, which is only synced from `thread.values.artifacts` in `chat-box.tsx`:

> `setArtifacts(thread.values.artifacts);`

Artifacts surfaced via the message-layer `present_files` tool call are **never** written back to `thread.values.artifacts`. So when such a file becomes the selected artifact, its filepath has no matching `<SelectItem>`; Radix silently renders an empty trigger and the header filename appears blank, even though the preview body still loads correctly via the `/artifacts/...` route.

The existing TODO comment in `chat-box.tsx` already acknowledges that "artifacts auto discovering is not work now". This PR addresses the user-visible symptom without depending on that future work.

## Fix

In `artifact-file-detail.tsx`, compute `artifactOptions` as a defensive union of `artifacts` and the current `filepath`: if `filepath` is missing from `artifacts`, prepend it so a matching `<SelectItem>` always exists for `<SelectValue>` to render.

```tsx
const artifactOptions = useMemo(() => {
  const list = artifacts ?? [];
  if (list.includes(filepath)) return list;
  return [filepath, ...list];
}, [artifacts, filepath]);
```

Then render `artifactOptions` (instead of `artifacts`) inside `<SelectContent>`.

This keeps the header label in sync with the active file without:

- changing `ArtifactsContext` semantics,
- coupling the UI to a future auto-discovery mechanism,
- altering behavior for the common case where the artifact is already in `thread.values.artifacts`.

### New e2e coverage

Added a case in `tests/e2e/artifact-preview.spec.ts`:

`"shows the title for a presented artifact missing from thread artifacts"`

It mocks a thread where:

- `thread.values.artifacts = []`
- the only artifact reference comes from a `present_files` tool call

and asserts both the header filename and the preview content render after the user clicks the file in the message list. The test **fails on `main`** and **passes with this fix**.

### Full validation

All checks run locally per `CONTRIBUTING.md`:

| Check | Command | Result |
|---|---|---|
| Format | `pnpm format:write` | clean |
| Lint | `pnpm lint` | 0 errors |
[...]

Issue #3189, opened 2026-05-24
`write_file` 写入长报告时 LLM 流式超时/断连，输出到页面正常


sfsflm:
### 背景

执行调研/分析任务时，Agent 将完整报告通过 `write_file` 工具一次性写入文件。报告内容较长，LLM 需要在单次流式响应中生成整个 `content` 参数。

### 现象

写入文件时报错，两种错误交替出现：

**错误 1（在config.yaml中配置 `stream_chunk_timeout` 前）**
```
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_openai/chat_models/base.py", line 3361, in _astream
    async for chunk in super()._astream(*args, **kwargs):
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_openai/chat_models/base.py", line 1838, in _astream
    async for chunk in _astream_with_chunk_timeout(
  File "/app/backend/.venv/lib/python3.12/site-packages/langchain_openai/chat_models/_client_utils.py", line 665, in _astream_with_chunk_timeout
    raise StreamChunkTimeoutError(
langchain_openai.chat_models._client_utils.StreamChunkTimeoutError: No streaming chunk received for 120.0s (model=mimo-v2.5, chunks_received=58). The connection may be alive at the TCP layer but is not producing content. Tune or disable via the `stream_chunk_timeout` constructor kwarg (set to None or 0 to disable) or the `LANGCHAIN_OPENAI_STREAM_CHUNK_TIMEOUT_S` env var. See also `http_socket_options` for the kernel-level TCP timeout that catches dead TCP peers.
```

**错误 2（配置 `stream_chunk_timeout: 300` 后）**
```
2026-05-24 02:45:11 - deerflow.agents.middlewares.llm_error_handling_middleware - WARNING - Transient LLM error on attempt 1/3; retrying in 1000ms: peer closed connection without sending complete message body (incomplete chunked read)
```

### 复现条件

- **场景**：Agent 执行 `write_file(content="完整报告")`，单次工具调用 content 参数很大时
- **正常场景**：同样内容让 LLM 直接输出到页面（对话式输出），正常完成
- **多环境验证**：本地电脑（家庭宽带）和公司服务器（数据中心）均出现过
- **多模型验证**：不同 LLM 提供方均出现过

### 已尝试的修复

- 配置 `stream_chunk_timeout: 300`：无效，只是把报错从"客户端超时"变成"服务器断连"
[...]

PR #3854, opened 2026-06-28
fix(frontend): retain presented artifacts in header dropdown

### LINKED ISSUE #3853: [bug] Presented artifacts disappear from the header dropdown after switching files
### Before you start

### Problem summary

A presented artifact that is missing from thread.values.artifacts disappears from the artifact header dropdown after switching to a regular artifact.

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

PR #3198 added the currently selected filepath to the artifact header dropdown when that filepath is missing from thread.values.artifacts.

However, the fallback is derived only from the current filepath:

return [filepath, ...list];

After opening a presented file such as A.md, the dropdown initially contains A.md and the regular thread artifacts. If the user switches to a regular artifact such as B.md, filepath becomes B.md. Because B.md already exists in thread.values.artifacts, A.md is no longer included and disappears from the dropdown.

The file is not deleted and remains accessible from its original message card, but the user cannot return to it through the artifact header dropdown.

### Expected behavior

Presented artifacts opened during the current thread session should remain available in the artifact header dropdown after switching to another artifact.

The dropdown should merge previously opened presented filepaths with thread.values.artifacts, without changing artifact auto-discovery behavior.

### Steps to reproduce

1. Open a thread containing a presented file A.md that is not included in thread.values.artifacts.
2. Ensure the thread also contains a regular artifact B.md.
3. Click the A.md file card in the message list.
4. Open the artifact header dropdown and observe that both A.md and B.md are available.
5. Select B.md from the dropdown.
6. Open the dropdown again.
7. Observe that A.md has disappeared and cannot be selected without returning to the message list.

### Relevant logs

```shell
N/A — this is a client-side state/navigation issue and does not produce an error or relevant log output.
```

Local (make dev)

### Operating system

macOS

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

Huixin615:
Closes #3853

## Why

PR #3198 made the currently selected presented filepath visible in the artifact header dropdown when it was missing from `thread.values.artifacts`.

However, the fallback was derived only from the current filepath. After opening a presented file and then switching to a regular thread artifact, the presented file disappeared from the dropdown. Users had to return to the original message card to reopen it.

The root cause was that `artifactOptions` had no session memory. Adding presented paths directly to the artifacts context would not be reliable because `ChatBox` replaces that context with `thread.values.artifacts` during thread updates.

## What changed

- Presented filepaths opened in the current artifact-detail session now remain available in the header dropdown.
- Previously opened presented paths are merged with thread artifacts and deduplicated.
- The accumulator stays local to `ArtifactFileDetail`, outside the context list overwritten by thread synchronization.
- Temporary `write-file:` preview paths are not accumulated.
- Artifact auto-discovery and backend thread state remain unchanged.
- The existing Playwright regression test now covers the complete navigation path:
  1. Open presented file A.
  2. Switch to regular artifact B.
  3. Confirm A remains in the dropdown.
  4. Switch back to A successfully.

## Surface area

## Screenshots / Recording

Before: after switching from a presented file to a regular artifact, the presented file disappears from the header dropdown.

After: previously opened presented files remain available and can be selected again.

image

## Bug fix verification

- Test path that reproduces the bug: `frontend/tests/e2e/artifact-preview.spec.ts`
- Did it go red on `main` and green on this branch? **yes**
- On the original implementation, the test failed because the `presented-report.md` option could not be found after switching to `report.md`.
- With this change, the same A → B → A navigation test passes.

## Validation

Executed from `frontend/`:

- `pnpm format` — passed
- `pnpm lint` — passed with 3 pre-existing warnings in unrelated message-list files
- `pnpm typecheck` — passed
- `pnpm test` — passed, 373 tests
- `SKIP_ENV_VALIDATION=1 DEER_FLOW_AUTH_DISABLED=1 pnpm build` — passed
- `pnpm exec playwright test tests/e2e/artifact-preview.spec.ts --reporter=line` — passed, 6 tests

The production build emitted the existing Turbopack NFT tracing warning for `next.config.js`; the build completed successfully.

### DIFF
--- a/backend/packages/harness/deerflow/agents/lead_agent/prompt.py
+++ b/backend/packages/harness/deerflow/agents/lead_agent/prompt.py
@@ -565,12 +565,11 @@ combined with a FastAPI gateway for REST API access [citation:FastAPI](https://f
 - Output Files: Final deliverables must be in `/mnt/user-data/outputs`
 - File Editing Workflow: When revising an existing file, prefer
   `str_replace` over `write_file` — it sends only the diff and avoids
-  re-emitting the whole file (mirrors Claude Code's Edit and Codex's
-  apply_patch). When writing long new content from scratch, split it
-  into sections: the first `write_file` call creates the file, then use
-  `write_file` with append=True to extend it section by section. This
-  keeps each tool call small and avoids mid-stream chunk-gap timeouts
-  on oversized single-shot writes. (See issue #3189.)  
+  re-emitting the whole file. When writing long new content from scratch,
+  split it into sections: the first `write_file` call creates the file,
+  then use `write_file` with append=True to extend it section by section.
+  This keeps each tool call small and avoids mid-stream chunk-gap timeouts
+  on oversized single-shot writes.
 - Clarity: Be direct and helpful, avoid unnecessary meta-commentary
 - Including Images and Mermaid: Images and Mermaid diagrams are always welcomed in the Markdown format, and you're encouraged to use `![Image Description](image_path)\n\n` or "```mermaid" to display images in response or Markdown files
 - Multi-task: Better utilize parallel tool calling to call multiple tools at one time for better performance
--- a/backend/packages/harness/deerflow/agents/lead_agent/prompt.py
+++ b/backend/packages/harness/deerflow/agents/lead_agent/prompt.py
@@ -565,11 +565,12 @@ combined with a FastAPI gateway for REST API access [citation:FastAPI](https://f
 - Output Files: Final deliverables must be in `/mnt/user-data/outputs`
 - File Editing Workflow: When revising an existing file, prefer
   `str_replace` over `write_file` — it sends only the diff and avoids
-  re-emitting the whole file. When writing long new content from scratch,
-  split it into sections: the first `write_file` call creates the file,
-  then use `write_file` with append=True to extend it section by section.
-  This keeps each tool call small and avoids mid-stream chunk-gap timeouts
-  on oversized single-shot writes.
+  re-emitting the whole file (mirrors Claude Code's Edit and Codex's
+  apply_patch). When writing long new content from scratch, split it
+  into sections: the first `write_file` call creates the file, then use
+  `write_file` with append=True to extend it section by section. This
+  keeps each tool call small and avoids mid-stream chunk-gap timeouts
+  on oversized single-shot writes. (See issue #3189.)  
 - Clarity: Be direct and helpful, avoid unnecessary meta-commentary
 - Including Images and Mermaid: Images and Mermaid diagrams are always welcomed in the Markdown format, and you're encouraged to use `![Image Description](image_path)\n\n` or "```mermaid" to display images in response or Markdown files
 - Multi-task: Better utilize parallel tool calling to call multiple tools at one time for better performance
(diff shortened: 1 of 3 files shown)


Huixin615:
顺手带了个之前遗留的的提示词优化

Huixin615:
@fancyboi999 辛苦review一下

backend/packages/harness/deerflow/agents/lead_agent/prompt.py
```diff
@@ -565,12 +565,11 @@ def _build_subagent_section(max_concurrent: int, *, app_config: AppConfig | None
 - Output Files: Final deliverables must be in `/mnt/user-data/outputs`
 - File Editing Workflow: When revising an existing file, prefer
   `str_replace` over `write_file` — it sends only the diff and avoids
-  re-emitting the whole file (mirrors Claude Code's Edit and Codex's
-  apply_patch). When writing long new content from scratch, split it
-  into sections: the first `write_file` call creates the file, then use
-  `write_file` with append=True to extend it section by section. This
-  keeps each tool call small and avoids mid-stream chunk-gap timeouts
-  on oversized single-shot writes. (See issue #3189.)  
+  re-emitting the whole file. When writing long new content from scratch,
```
fancyboi999: This prompt edit is unrelated to the #3853 dropdown fix — recommend moving it to its own PR. Mixing a backend agent-prompt change into a frontend bugfix makes both harder to review and to revert independently, and dropping the `(See issue #3189.)` reference loses the rationale for the append-by-section guidance. Keeping this PR surgical to the dropdown change would be cleaner.

fancyboi999 (commented):
Thanks for taking this. The dropdown fix itself is correct — I traced the #3853 path (open presented A -> switch to regular B -> A stays in the dropdown via the local `openedPresentedFilepaths` accumulator, then A is selectable again). The thread-scoping I was worried about also holds: `ChatBox` calls `deselect()` on thread change, so `ArtifactFileDetail` unmounts and the accumulator resets per thread rather than leaking presented paths across threads. `write-file:` paths are correctly excluded, and the dedup against `thread.values.artifacts` avoids doubles.

One thing to change: this PR also bundles an unrelated `lead_agent/prompt.py` edit (the "顺手带了个提示词优化"). I'd split that into its own PR — it's a separate concern from the #3853 frontend fix (and it's an agent-prompt behavior change riding inside a `fix(frontend)` PR), so mixing them makes both harder to review and to revert independently. Separately, that edit drops the `(See issue #3189.)` reference, which is the traceability for the append-by-section guidance — worth keeping, or justifying on its own in the dedicated PR.

With the prompt change pulled out, the frontend fix looks good to me.

Huixin615:
这个修改没问题 属于冗余的东西 没有任何影响 顺便带上去

WillemJiang:
> 这个修改没问题 属于冗余的东西 没有任何影响 顺便带上去

As we have 2.0.x branch this time, please keep the PR simple and clear for maintenance.

Huixin615:
> > 这个修改没问题 属于冗余的东西 没有任何影响 顺便带上去
> 
> As we have 2.0.x branch this time, please keep the PR simple and clear for maintenance.

@WillemJiang Thanks for the suggestion. I have already retracted the change.

WillemJiang (approved):

PR #3855, opened 2026-06-28
fix(gateway): require admin for global skills management endpoints

DengY11:
## Summary

The skills router (`backend/app/gateway/routers/skills.py`) exposed its management endpoints with no authorization check (only `Depends(get_config)`), even though the MCP router already guards the equivalent global `extensions_config` mutations with `require_admin_user` (added in the #3425 security hardening). Skills storage is global and shared across users (there is no per-user path), it lives in the same `extensions_config.json` as MCP servers, and custom skill `SKILL.md` content is injected into every user's agent system prompt. As a result any authenticated non-admin user could mutate or read global skills that affect all tenants.

## Impact

A normal authenticated user (auth enabled, production) could call these endpoints without being an admin:

- `POST /api/skills/install` installs from an arbitrary `thread_id` and writes into the global custom skills tree.
- `PUT /api/skills/custom/{skill_name}` rewrites a global skill, which is then injected into every user's agent system prompt — i.e. cross-tenant prompt injection / agent hijacking.
- `DELETE /api/skills/custom/{skill_name}` and `POST /api/skills/custom/{skill_name}/rollback` let any user tamper with global skills.
- `GET /api/skills/custom`, `GET /api/skills/custom/{skill_name}`, and `GET /api/skills/custom/{skill_name}/history` expose the raw `SKILL.md` bodies and edit history of global custom skills.
- `PUT /api/skills/{skill_name}` (the enable toggle) writes the shared `extensions_config.json` and calls `refresh_skills_system_prompt_cache_async()`, so a non-admin could enable or disable any skill globally for every tenant. There is no per-user skill state backing this toggle, so despite being surfaced in the settings UI it is a global mutation, not a per-user preference.

This is the same class of risk that motivated the MCP hardening in #3425 ("the HTTP API is an untrusted boundary"), but the skills endpoints were missed at the time.

## Fix

Add `require_admin_user` to every endpoint that mutates global state or reads raw custom-skill content/history, mirroring the MCP router. The shared read path used internally by `update`/`rollback` was extracted into a non-auth helper (`_read_custom_skill_response`) so internal reuse does not double-check auth.

Only the read-only `GET /api/skills` and `GET /api/skills/{skill_name}` stay open to normal users: they return just name/description/enabled and back the user-facing settings UI.

## Frontend follow-up (known, not in this PR)

Two of the now-guarded endpoints are currently wired into the normal-user UI, so after this change a non-admin user will receive a 403 there:

- `PUT /api/skills/{skill_name}` — the toggle in Settings → Skills (`frontend/src/components/workspace/settings/skill-settings-page.tsx`).
- `POST /api/skills/install` — the "Install skill" action on `.skill` artifacts (`frontend/src/components/workspace/artifacts/`).

The guard itself is the correct security posture (both are global writes). The remaining UX work is to gate/hide these controls for non-admins and handle the 403 gracefully — mirroring how the MCP Tools settings page already degrades on 403 (`frontend/src/components/workspace/settings/tool-settings-page.tsx`). I'm happy to fold that frontend change into this PR or do it as a follow-up, whichever the maintainers prefer. Flagging it explicitly here so the backend/description and the UX impact are consistent (thanks @willem-bd for catching the description/code mismatch).

## Tests

- New `backend/tests/test_skills_router_authz.py`: a non-admin user receives 403 on every guarded endpoint (including the enable toggle); basic listing stays open to normal users; an admin can still toggle.
- Updated `backend/tests/test_skills_custom_router.py` to authenticate as an admin (the existing positive-path cases now assert that admins can still manage skills).
- `pytest tests/ -k skill` → 350 passed, 1 skipped. `ruff check` is clean.


fancyboi999 (commented):
The core hardening is correct and well-scoped: guarding the custom-skill mutations, install, and the raw-content/history reads with `require_admin_user` mirrors the #3425 MCP pattern, and pulling out `_read_custom_skill_response` so `update`/`rollback` don't double-check auth is the right call. Renaming the pydantic params to `body` to free up `request: Request` is correct too.

**P2 — the endpoint you deliberately left open, `PUT /skills/{skill_name}` (the enable toggle), is also a global mutation, not a per-user preference.** `update_skill` (skills.py:323) writes the shared `extensions_config.json`:

```python
extensions_config.skills[skill_name] = SkillStateConfig(enabled=request.enabled)   # :338
...
with open(config_path, "w", ...) as f:   # the same global config skills/MCP share
```

So a non-admin can enable or disable any skill globally for every tenant — turn off a skill everyone relies on, or turn on one that was intentionally disabled (which changes every user's injected skill set). That's the same "non-admin mutates global config affecting all tenants" class this PR is closing, so leaving it open is a gap in the PR's own threat model. The "surfaced as a per-user preference" justification doesn't match the code: as your summary notes, there is no per-user path, so this write lands in the shared file.

I'd guard `update_skill` as well. If a normal-user toggle is genuinely intended, it needs to be backed by per-user state rather than the global config — as written, "treated as a normal-user action" silently makes it a cross-tenant write. The read-only `GET /skills` and `GET /skills/{skill_name}` (name/description/enabled only) are fine to leave open.

backend/app/gateway/routers/skills.py
```diff
@@ -307,7 +320,12 @@ async def get_skill(skill_name: str, config: AppConfig = Depends(get_config)) ->
     summary="Update Skill",
     description="Update a skill's enabled status by modifying the extensions_config.json file.",
 )
-async def update_skill(skill_name: str, request: SkillUpdateRequest, config: AppConfig = Depends(get_config)) -> SkillResponse:
+async def update_skill(skill_name: str, body: SkillUpdateRequest, request: Request, config: AppConfig = Depends(get_config)) -> SkillResponse:
+    # Enabling/disabling a skill writes the shared extensions_config.json and
+    # refreshes the system prompt for every tenant, so it is a global mutation
+    # (there is no per-user skill state). Guard it as admin-only like the other
+    # global config writes, matching the MCP router.
+    await require_admin_user(request, detail=_ADMIN_REQUIRED_DETAIL)
```
willem-bd, line 328: **`PUT /api/skills/{name}` is now admin-only, but the PR description says it's "left open" — and this endpoint backs the normal-user settings toggle.**

The description states this endpoint *"is surfaced as a per-user preference in the settings dialog and is treated as a normal-user action, so it is left open to preserve that UX."* The code does the opposite: it adds `require_admin_user` here.

The frontend exposes this toggle to every user (`frontend/src/components/workspace/settings/skill-settings-page.tsx:103-109` → `enableSkill` → `PUT /api/skills/{name}`) with no role gating in the UI. So any **non-admin** user opening Settings → Skills and flipping the switch now gets a **403**, the toggle silently fails (no error handling in the UI), and the user can't enable/disable any skill.

The guard is security-defensible (the write *is* global — `extensions_config.json` + `refresh_skills_system_prompt_cache_async()`), but the description/code contradiction means a reviewer approving on the writeup won't catch this. Recommend either: (a) drop this guard to match the stated scope, or (b) keep it and update the description + add frontend handling (hide/disable the switch for non-admins, surface the 403).

backend/app/gateway/routers/skills.py
```diff
@@ -106,9 +108,10 @@ async def list_skills(config: AppConfig = Depends(get_config)) -> SkillsListResp
     summary="Install Skill",
     description="Install a skill from a .skill file (ZIP archive) located in the thread's user-data directory.",
 )
-async def install_skill(request: SkillInstallRequest, config: AppConfig = Depends(get_config)) -> SkillInstallResponse:
+async def install_skill(request: Request, body: SkillInstallRequest, config: AppConfig = Depends(get_config)) -> SkillInstallResponse:
+    await require_admin_user(request, detail=_ADMIN_REQUIRED_DETAIL)
```
willem-bd, line 112: **`POST /api/skills/install` is called from the normal-user artifact UI → non-admins hit 403 on an undocumented flow.**

`installSkill` is invoked from `frontend/src/components/workspace/artifacts/artifact-file-detail.tsx:158` and `artifact-file-list.tsx:55` when *any* user clicks "install skill" on a `.skill` artifact in a chat thread. With this guard, non-admins get a 403 and a generic failure toast.

Gating is defensible (install writes the global custom-skills tree), but unlike the custom-skill CRUD endpoints — which have **no** frontend caller and are clearly admin tooling — this one is wired into the general chat UX. Worth either documenting it as an intentional admin-only flow or gating the UI button behind an admin check, so the breakage is explicit rather than a silent 403 for end users.

willem-bd (commented):
Security review — the authz fix is sound and well-tested. The guards fire before each `try/except` (403s aren't swallowed into 500s), the `request`/`body` parameter split mirrors the MCP router pattern, and auth-disabled dev mode is unaffected (`get_auth_disabled_user()` returns `system_role="admin"`).

Two scope/UX notes inline below — both are behavior changes a maintainer should sign off on deliberately, not bugs in the guard logic.

DengY11:
Thanks @willem-bd — both points addressed.

I updated the PR description. `PUT /api/skills/{skill_name}` is no longer described as "left open" — it writes the shared `extensions_config.json` and refreshes the system prompt for every tenant, so it's a global mutation (there's no per-user skill state) and is now correctly documented as admin-only alongside the other guarded endpoints.

**Frontend follow-up (now in this PR)**: I added the matching UI changes so non-admins no longer hit a silent 403 (commit 34487f67):

- `core/skills/api.ts`: new `SkillRequestError` (`status` + `isAdminRequired`); `loadSkills`/`enableSkill` throw it on non-ok responses, `installSkill` throws on 403.
- `core/skills/hooks.ts`: `useSkills` no longer retries on `SkillRequestError`.
- Settings → Skills (`skill-settings-page.tsx`): shows an "admin required" message on 403 and disables the enable toggle for non-admins — mirroring the MCP tools page's 403 degradation.
- `.skill` artifacts (`artifact-file-detail.tsx` / `artifact-file-list.tsx`): the "Install skill" action only renders for admins, with an admin-required toast if a 403 still occurs.
- i18n: added `settings.skills.adminRequired` / `installAdminRequired` (en + zh).

DengY11:
Hi @WillemJiang  i had fixed the frontend Lint-check problem now, please run check again

WillemJiang (approved):

PR #3858, opened 2026-06-29
feat(runtime): implement goal continuations

hetaoBackend:
## Summary
- Adds thread-scoped goal state, Gateway/embedded/TUI/IM `/goal` management, and worker-side goal evaluation/continuation handling.
- Updates the frontend `/goal` command to display the active goal and start an agent run immediately when setting a goal.
- Adds regression coverage and docs for goal state reducers, evaluator runtime, worker continuation, Gateway/client/channel/TUI paths, and frontend e2e behavior.

## Validation
- `PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run pytest tests/test_goal_runtime.py tests/test_goal_worker.py tests/test_threads_router.py tests/test_client.py tests/test_channels.py tests/test_tui_command_registry.py tests/test_thread_state_reducers.py tests/test_slash_skills.py -v` — 455 passed, 6 warnings.
- `uv run ruff check . && uv run ruff format --check .` — passed; 646 files already formatted.
- `./node_modules/.bin/tsc --noEmit && ./node_modules/.bin/eslint . --ext .ts,.tsx` — passed with 0 errors and 3 pre-existing warnings.
- `./node_modules/.bin/playwright test tests/e2e/chat.spec.ts --reporter=list --workers=1` — 14 passed.

## Notes
- The PR is draft because the branch contains a broad backend/frontend goal feature surface and should get maintainer review before marking ready.


backend/app/gateway/routers/threads.py
```diff
@@ -205,6 +224,36 @@ def _derive_thread_status(checkpoint_tuple) -> str:
...
+    record = await thread_store.get(thread_id, **thread_owner_kwargs)
+    if record is None and thread_owner_user_id:
+        unscoped_record = await thread_store.get(thread_id, user_id=None)
+        if unscoped_record is not None:
+            if unscoped_record.get("user_id") != thread_owner_user_id:
+                await thread_store.update_owner(thread_id, thread_owner_user_id, user_id=None)
+            record = await thread_store.get(thread_id, **thread_owner_kwargs)
+    if record is None:
+        try:
+            await thread_store.create(thread_id, metadata={})
```
willem-bd: [P1] `_ensure_thread_for_goal` creates the thread without owner scoping

- Problem: `thread_store.create(thread_id, metadata={})` omits `**thread_owner_kwargs`, so the new `thread_meta` row is scoped to the contextvar user (`"default"` for internal channel calls) rather than the trusted owner in `X-DeerFlow-Owner-User-Id`.
- Evidence: the helper computes `thread_owner_kwargs` and uses it on the read path (`thread_store.get(..., **thread_owner_kwargs)`), but drops it on this create. The sibling `POST /threads` create endpoint passes `**thread_owner_kwargs` to `thread_store.create`. The goal route does not set `require_existing=True`, so `check_access` returns `True` for a missing row and this path is reachable as the first touch on a fresh thread UUID — exactly the `/chats/new` flow the route docstring describes. No later code re-scopes the row.
- Impact: the thread can be missed by owner-scoped `/threads/search` (and other owner-scoped reads) until the first run rewrites it; for the unbound-auth channel path the unscoped row can leak into other owners' listings.
- Suggested fix: `await thread_store.create(thread_id, metadata={}, **thread_owner_kwargs)`.
- Test: assert that a `PUT /api/threads/{id}/goal` on a not-yet-created thread, with a trusted owner header, creates a `thread_meta` row whose `user_id` equals the owner and appears in that owner's `/threads/search`.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -333,6 +353,21 @@ async def run_agent(
...
+                bridge=bridge,
+                checkpointer=checkpointer,
+                thread_id=thread_id,
+                run_id=run_id,
+                model_name=record.model_name,
+                app_config=ctx.app_config,
+            )
+            if continuation_input is None:
+                break
+            await _stream_once(continuation_input)
```
willem-bd: [P2] A Stop requested during goal evaluation still runs one more continuation turn

- Problem: the `while` loop only checks `record.abort_event` at the top (line 358). Between iterations, `_prepare_goal_continuation_input` awaits `evaluate_goal_completion` -> `model.ainvoke` (a slow LLM call) and has no access to `abort_event`, so an abort requested during evaluation is not honored before `await _stream_once(continuation_input)` runs.
- Evidence: `_stream_once` only checks `abort_event` *inside* the `async for chunk in agent.astream(...)` loop, i.e. after the graph has already been kicked off — so the extra turn is not cheaply short-circuited.
- Suggested fix: thread `record` (or `abort_event`) into `_prepare_goal_continuation_input` and re-check it after the evaluator returns, before calling `_stream_once`; or check `record.abort_event.is_set()` between lines 367 and 369 and `break`.
- Test: start a continuation, set `abort_event` during the evaluator's `model.ainvoke`, and assert no `_stream_once(continuation_input)` follows.

backend/app/channels/manager.py
```diff
@@ -1606,6 +1611,62 @@ async def _handle_command(self, msg: InboundMessage) -> None:
...
+                    timeout=10,
+                    headers=headers,
+                    json={"objective": command.objective},
+                )
+                response.raise_for_status()
+                goal = (response.json() or {}).get("goal")
+        except Exception:
+            logger.exception("Failed to set goal through gateway")
+            return "Failed to set goal."
+        return f"Goal set: {goal.get('objective') if goal else command.objective}"
```
willem-bd: [P2] IM `/goal <text>` sets the goal but never starts a run (breaks documented parity with the web UI)

- Problem: the `set` branch PUTs the goal and returns `"Goal set: ..."` without launching an agent turn. The web `InputBox` additionally submits the objective as the next user task so the agent starts immediately. Since goal evaluation is post-run by construction, an IM user gets silence after `/goal finish X` — the evaluator never fires until a separate follow-up message.
- Evidence: README states "The same command is available from the TUI and supported IM channels," and the continuation logic keys off a completed run's messages. Sibling commands that need a run (`/bootstrap`, skill `route_to_chat`) already transform to CHAT and call `_handle_chat`; `/goal set` does not.
- Suggested fix: after a successful PUT in the set branch, follow the `/bootstrap` precedent — replace the inbound message with the objective and `await self._handle_chat(...)` (or otherwise kick off a run).
- Test: assert that an IM `/goal <objective>` results in both a persisted goal and a started run.

backend/packages/harness/deerflow/runtime/goal.py
```diff
@@ -0,0 +1,504 @@
...
+        created_at=timestamp,
+        updated_at=timestamp,
+        continuation_count=0,
+        max_continuations=capped_max,
+        no_progress_count=0,
+        max_no_progress_continuations=max(0, int(max_no_progress_continuations)),
+    )
+
+
+def _strip_think_blocks(text: str) -> str:
```
willem-bd: [P2] `_strip_think_blocks` / `_strip_markdown_code_fence` / `_extract_response_text` duplicate `app/gateway/routers/suggestions.py`

- Problem: these helpers (and the `_THINK_BLOCK_RE` / `_OPEN_THINK_RE` regexes at lines 48-49) are byte-for-byte copies of the ones in `suggestions.py`. They exist precisely because reasoning models (e.g. MiniMax-M3) corrupt JSON parsing, so the two copies will drift the next time a reasoning-model quirk is patched in one place.
- Evidence: the harness->app boundary (`tests/test_harness_boundary.py`) forbids `deerflow.*` from importing `app.*`, which is why they were copied rather than imported.
- Suggested fix: move the canonical helpers into `deerflow.utils` (e.g. `deerflow/utils/llm_text.py`) and import them from both `suggestions.py` and `goal.py`.
- Test: existing think-strip tests should cover the shared helper; add a harness-boundary assertion if one does not already guard the util location.

backend/packages/harness/deerflow/client.py
```diff
@@ -410,6 +415,98 @@ def flush_pending_str_parts() -> None:
...
+                "source": "input",
+                "writes": None,
+                "parents": {},
+                "created_at": now_iso(),
+            },
+            {},
+        )
+
+    @staticmethod
+    def _write_goal_checkpoint(checkpointer, thread_id: str, goal: GoalState | None) -> None:
```
willem-bd: [P2] `_ensure_thread_checkpoint` / `_write_goal_checkpoint` hand-roll the checkpoint RMW that `runtime/goal.py` already encapsulates

- Problem: these static methods manually rebuild the checkpoint put logic (`empty_checkpoint()`, `step=-1` metadata, `channel_versions` + `get_next_version` bump, `writes={as_node: {...}}`) that `goal.py::ensure_thread_checkpoint` / `write_thread_goal` already implement (and `goal.py` already handles the sync path via `_call_checkpointer_method(async_name, sync_name, ...)`).
- Evidence: this module already depends on `goal.py` (it imports `build_goal_state` from it at line 47), so the fork is unnecessary.
- Suggested fix: drop both staticmethods and delegate to `goal.py`'s helpers (e.g. wrap the async helpers via the existing sync-aware dispatch) so the version-bump / `aput` logic has exactly one owner.
- Test: the existing `TestGatewayConformance` / client goal tests should continue to pass through the delegated path.

backend/app/channels/manager.py
```diff
@@ -1606,6 +1611,62 @@ async def _handle_command(self, msg: InboundMessage) -> None:
...
+    async def _handle_goal_command(self, msg: InboundMessage, args: str) -> str:
+        command = parse_goal_command(args)
+        thread_id = await self._lookup_thread_id(msg)
+        headers = _owner_headers(msg) or create_internal_auth_headers()
+
+        if command.kind == "status":
+            if not thread_id:
+                return "No active goal."
+            try:
+                async with httpx.AsyncClient() as http:
```
willem-bd: [P2] Three near-identical `async with httpx.AsyncClient()` blocks in `_handle_goal_command`

- Problem: the status/clear/set branches each open a fresh `AsyncClient`, rebuild the same URL (`/api/threads/{quote(thread_id, safe='')}/goal`), the same headers/timeout, call `raise_for_status`, and wrap a try/except that differs only in the one-line error string.
- Suggested fix: extract one `_goal_request(method, thread_id, *, json=None)` helper (or reuse the existing `_fetch_gateway` with the goal subpath), leaving each branch 2-3 lines that differ only in method and reply text.
- Test: existing `TestHandleGoalCommand` coverage should pass unchanged through the helper.

backend/packages/harness/deerflow/runtime/goal.py
```diff
@@ -0,0 +1,504 @@
...
+        "Do not assume files, commands, tests, or external state changed unless the conversation explicitly shows it.\n"
+        "If the visible evidence is too weak to prove progress, fail closed with blocker missing_evidence.\n"
+        "Use blocker needs_user_input when the assistant is waiting on the user, run_failed when the turn failed, "
+        "external_wait when work is waiting on an outside system, goal_not_met_yet when useful autonomous work can continue, "
+        "and none only when satisfied is true.\n"
+        'Output exactly one JSON object: {"satisfied": boolean, "blocker": string, "reason": string, "evidence_summary": string}.'
+    )
+    user_content = f"Active goal:\n{goal['objective']}\n\nVisible conversation evidence:\n{conversation}\n\nIs the active goal fully satisfied?"
+
+    model = create_chat_model(
```
willem-bd: [P2] `evaluate_goal_completion` rebuilds the evaluator model on every call

- Problem: `create_chat_model(...)` is invoked inside `evaluate_goal_completion`, and the continuation loop calls it up to 8 times per run with constant `model_name` / `app_config`. Each call re-resolves provider config and re-instantiates the client.
- Suggested fix: construct the evaluator model once outside the loop (e.g. in `_prepare_goal_continuation_input` / `run_agent`) and pass the instance into `evaluate_goal_completion`.
- Test: existing goal-evaluator tests should pass with the injected model; add an assertion that `create_chat_model` is called once per goal run, not once per continuation.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -365,15 +330,87 @@ export function InputBox({
...
-        return;
-      }
-      if (!message.text.trim() && message.files.length === 0) {
-        return;
+  const handleGoalCommand = useCallback(
+    async (command: GoalCommand): Promise<boolean> => {
+      try {
+        let goal: GoalState | null = null;
+        if (command.kind === "status") {
+          const response = await fetch(
```
willem-bd, line 422: [P2] `/goal` requests use bare `fetch()` with no `AbortController`

- Problem: the GET/PUT/DELETE goal fetches in `handleGoalCommand` pass no `signal`. Switching threads, retrying, or unmounting the composer while a request is in flight leaves it orphaned, and its `.then` still runs `onGoalChange` / local-state updates against the (possibly unmounted or now-different-thread) hook.
- Suggested fix: tie an `AbortController` to the current `threadId` / component lifetime and abort in-flight goal requests on thread change or unmount (consistent with how stream requests are typically cancelled).
- Test: unit test that switching threads while a `/goal` GET is pending aborts the request and does not update the new thread's goal state.

willem-bd (commented):
Thanks @hetaoBackend. The goal-state modeling (real `ThreadState.goal` channel + `merge_goal` reducer) and the channel-agnostic continuation loop in the shared worker are solid. I found one likely production issue (owner scoping on thread creation) plus a few correctness and maintainability items worth addressing before this leaves draft.

hetaoBackend:
@willamJiang CI is green now. Could you please help review this again when you have a chance? Thanks!

WillemJiang:
@hetaoBackend, there is a conflict with the main branch. Please resolve it.

backend/app/gateway/routers/threads.py
hetaoBackend: Fixed. `_ensure_thread_for_goal` now passes `**thread_owner_kwargs` to `thread_store.create(...)`, and `test_goal_thread_creation_uses_internal_owner_header` covers the trusted-owner creation/search path.

backend/packages/harness/deerflow/runtime/runs/worker.py
hetaoBackend: Fixed. `_prepare_goal_continuation_input` now accepts `abort_event` and checks it before and after evaluation; the run loop also re-checks before streaming a continuation. Covered by the abort-during-evaluation worker test.

backend/app/channels/manager.py
hetaoBackend: Fixed. IM `/goal <objective>` now persists the goal and then dispatches the objective as a CHAT message through `_handle_chat`, creating a thread first when needed. Covered by the IM goal set tests.

backend/packages/harness/deerflow/runtime/goal.py
hetaoBackend: Fixed. The shared LLM response text helpers now live in `deerflow.utils.llm_text`, and both `goal.py` and `suggestions.py` import from that module.

backend/packages/harness/deerflow/client.py
hetaoBackend: Fixed. `DeerFlowClient` now delegates goal reads/writes to the runtime goal helpers (`read_thread_goal` / `write_thread_goal`) instead of hand-rolling checkpoint RMW logic; the client goal round-trip/conformance coverage still passes.

backend/app/channels/manager.py
hetaoBackend: Fixed. `_goal_request` centralizes the IM `/goal` gateway request setup for GET/DELETE/PUT, leaving each command branch with only method-specific behavior.

backend/packages/harness/deerflow/runtime/goal.py
hetaoBackend: Fixed. The evaluator model is now cached per run via `_get_goal_evaluator_model` and passed through `evaluator_model_factory`, so continuation evaluations reuse one model. Covered by the worker model-cache test.

frontend/src/components/workspace/input-box.tsx
hetaoBackend: Fixed. `/goal` fetches now use per-thread `AbortController`s, abort on thread change/unmount, and guard stale completions with request identity checks. Covered by the goal request lifecycle unit test.

frontend/src/components/workspace/input-box-helpers.ts
```diff
@@ -0,0 +1,226 @@
...
+      skill,
+      index,
+      name: skill.name.toLowerCase(),
+    }))
+    .filter(({ skill, name }) => {
+      if (!skill.enabled) {
+        return false;
+      }
+      return !normalizedQuery || name.includes(normalizedQuery);
+    })
```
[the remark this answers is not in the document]
hetaoBackend: Fixed. Skill suggestions now filter enabled skills whose names collide with builtin commands, preserving builtin precedence; added collision coverage in `input-box-helpers.test.ts`.

backend/tests/test_gateway_lifespan_shutdown.py
```diff
@@ -35,6 +35,9 @@ async def hang_forever() -> None:
     app = FastAPI()
     startup_config = SimpleNamespace(log_level="INFO", memory=SimpleNamespace(token_counting="char"))
 
+    startup_config = MagicMock()
+    startup_config.log_level = "INFO"
+    startup_config.memory.token_counting = "char"
```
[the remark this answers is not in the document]
hetaoBackend: Fixed. Removed the overwritten `SimpleNamespace` assignment so the shutdown test fixture now builds only the `MagicMock` startup config it actually uses.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -818,21 +947,25 @@ export function InputBox({
                       ? "bg-accent text-accent-foreground"
                       : "text-popover-foreground hover:bg-accent/70 hover:text-accent-foreground",
                   )}
-                  key={skill.name}
-                  onClick={() => applySkillSuggestion(skill)}
+                  key={suggestion.name}
```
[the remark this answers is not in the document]
hetaoBackend: Fixed. Suggestion list keys now use `${suggestion.kind}:${suggestion.name}`, avoiding duplicate keys when builtin commands and skills share a name.

hetaoBackend:
Merged latest `upstream/main` into `codex/goal-evaluator-continuation` and resolved the main-branch conflict. I also addressed the outstanding review feedback and marked the fixed review threads as resolved. Local verification passed for the targeted backend/frontend tests, lint checks, ruff, and `git diff --check`.

WillemJiang (approved):

PR #3860, opened 2026-06-29
feat(memory): add staleness review to prune silently-outdated facts

sontianye:
## Why

Memory facts created long ago may become outdated without any future conversation explicitly contradicting them — "Silent Staleness". For example, a user mentions "I use Vue.js" and six months later switches to React without ever discussing their tech stack again. The Vue.js fact persists at confidence 0.95, occupying injection budget and fact quota forever. The current system only supports passive contradiction-based removal; this PR adds proactive staleness review.

## What changed

- Memory now has a staleness review mechanism that surfaces aged facts to the LLM during the normal memory-update call (same LLM invocation, no extra API call)
- The LLM semantically judges each candidate as KEEP or REMOVE — stable attributes stay; outdated ones go
- Five new config fields with validated ranges: `staleness_review_enabled`, `staleness_age_days`, `staleness_min_candidates`, `staleness_max_removals_per_cycle`, `staleness_protected_categories`
- Safety cap limits max removals per cycle; when exceeded, lowest-confidence entries are kept
- Correction facts (category=correction) are protected by default

## Surface area

- [ ] **Frontend UI**
- [x] **Backend API** — new fields in `GET /memory/config` and `GET /memory/status`
- [x] **Agents / LangGraph** — prompt injection + update pipeline in `updater.py`
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [ ] **Default behavior change** — staleness review is enabled by default but only triggers when aged facts exist
- [x] **Docs / tests / CI only** — `config.example.yaml` + 32 new tests

## Validation

```
```
- ruff check + format: all clean
- 306 tests pass (32 new + 274 existing)


fancyboi999 (commented):
Reviewed the full staleness pipeline (config → `_select_stale_candidates` → prompt section → `_apply_updates` → normalization). One P2 to fix before merge:

**`_apply_updates` doesn't enforce the staleness safeguards the PR advertises.** It removes any id the LLM returns in `staleFactsToRemove` (L745→L757) with only the per-cycle count cap — no intersection against the actual candidate set. The age filter and the `correction`-category protection are enforced *only* inside `_select_stale_candidates`, which shapes the prompt; but the full `current_memory` (every fact id, including protected `correction` facts and fresh facts) is serialized into the prompt JSON, so any LLM slip that emits a protected or non-aged id here silently and irreversibly deletes that fact — defeating the stated guarantee "Correction facts are protected by default" and the goal of deterministic guardrails that don't rely on the model. One-line fix + a suggested regression test inline.

Note on test coverage: `test_protected_category_excluded` only exercises `_select_stale_candidates` (the prompt layer); none of the `TestApplyUpdatesStaleness` cases assert that a protected/non-candidate id in `staleFactsToRemove` is refused at apply time — so the advertised protection is currently unverified where it actually matters.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -631,11 +734,37 @@ def _apply_updates(
...
-        # Remove facts
+        # Remove facts (contradiction-based)
         facts_to_remove = set(update_data.get("factsToRemove", []))
         if facts_to_remove:
             current_memory["facts"] = [f for f in current_memory.get("facts", []) if f.get("id") not in facts_to_remove]
 
+        # ── Staleness review removals ──
+        stale_removals = update_data.get("staleFactsToRemove", [])
+        if isinstance(stale_removals, list) and stale_removals:
+            stale_ids_to_remove = {entry["id"] for entry in stale_removals if isinstance(entry, dict) and "id" in entry}
```
fancyboi999, line 775: **P2 (data-loss / safeguard bypass)**: `stale_ids_to_remove` is built straight from the LLM's `staleFactsToRemove` and applied at L757 with only the count cap — there is no apply-time check that each id is an actual staleness candidate. The age filter and the documented `correction`-protection live only in `_select_stale_candidates` (prompt-surfacing); but the full `current_memory` is in the prompt JSON, so an LLM that emits a protected or fresh fact's id here will get it permanently deleted, bypassing both guards.

Suggest gating the removal set against the candidate ids, e.g.:
```python
candidate_ids = {f["id"] for f in _select_stale_candidates(current_memory, config)}
stale_ids_to_remove &= candidate_ids
```
and add a regression test where `staleFactsToRemove` names a `correction` id and a fresh-fact id and asserts both survive `_apply_updates` (the current protection test only covers `_select_stale_candidates`).

sontianye:
Good catch — the prompt-layer protection was not being enforced at the apply layer.

Fixed in `99f8027f`: `stale_ids_to_remove` is now intersected with `_select_stale_candidates(current_memory, config)` before the safety cap runs. The guardrail is unconditional (not gated on `staleness_review_enabled`), so the apply-layer protection is independent of both model behavior and the feature flag.

Three regression tests added at the apply layer:
- `test_protected_category_fact_refused_at_apply` — correction-category id in staleFactsToRemove is rejected
- `test_non_aged_fact_refused_at_apply` — fresh (non-aged) fact id is rejected
- `test_guardrail_runs_when_staleness_review_disabled` — guardrail still enforces even when the feature flag is off

xg-gh-25:
**Draft for bytedance/deer-flow #3860 (T-MEM confidence: 8)**

This staleness-review mechanism for pruning outdated facts is solid architectural thinking. The silently-outdated category (facts that remain structurally intact but semantically obsolete) is especially subtle — it's the memory equivalent of technical debt that passes all tests but fails in production.

A few observations from similar memory-pruning systems:

1. **Temporal decay isn't enough** — you need semantic drift detection. A fact like "user prefers Python 3.8" becomes stale not because it's old, but because Python 3.13 exists and their codebase migrated. Age-based heuristics miss this.

2. **Pruning trigger points matter** — are you reviewing on read (lazy), on write (eager), or on schedule (batch)? Each has tradeoffs. Lazy pruning keeps hot facts fresh but lets cold facts rot. Batch review catches drift but adds latency.

3. **False-positive risk** — aggressive pruning can delete facts that are dormant but still valid. Consider a "do not prune" flag for critical invariants (like security policies or user boundaries).

The PR shows a `staleness_threshold` param — curious if you're planning confidence-weighted pruning (facts with high certainty decay slower) or uniform decay. The former is more robust but harder to tune.

How are you planning to surface pruned facts for audit? Memory systems that silently delete create debugging nightmares. A pruning log (even append-only) helps recover from over-aggressive cleanups.

---
*Part of systematic memory architecture research via [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-MEM](https://github.com/xg-gh-25/SwarmAI/discussions/12)*

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -98,6 +98,40 @@ class MemoryConfig(BaseModel):
             "safety-truncation ceiling is raised accordingly."
         ),
     )
+    # ── Staleness review ────────────────────────────────────────────────
+    staleness_review_enabled: bool = Field(
```
willem-bd, line 102: Docs sync: `backend/AGENTS.md` (and the README memory section) should be updated in this change set. The Documentation Update Policy in `backend/AGENTS.md` is explicit — *"CRITICAL: Always update README.md and AGENTS.md after every code change"* / *"Update `AGENTS.md` for development changes (architecture, commands, workflows, internal systems)."*

The Memory System → Configuration block in `backend/AGENTS.md` enumerates the memory config fields verbatim (`enabled, storage_path, debounce_seconds, model_name, max_facts, fact_confidence_threshold, injection_enabled, max_injection_tokens`), and the overview bullet repeats the same list. This PR adds five new fields — `staleness_review_enabled`, `staleness_age_days`, `staleness_min_candidates`, `staleness_max_removals_per_cycle`, `staleness_protected_categories` — plus a new sub-mechanism (prompt surfacing + apply-layer guardrail), but none of the changed files is `AGENTS.md` or `README.md`, so that field list and the updater-workflow description are now stale. A short paragraph in the Workflow section noting the staleness pass would cover the architecture side.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -376,6 +397,76 @@ def _fact_content_key(content: Any) -> str | None:
...
+    """Parse an ISO-8601 datetime string from a fact's createdAt field.
+
+    Handles both ``Z``-suffixed and ``+00:00`` offset formats. Returns
+    ``None`` on any parse failure so callers can safely skip malformed facts.
+    """
+    if not raw:
+        return None
+    try:
+        # Handle Z suffix (e.g. "2025-06-01T12:00:00Z")
+        normalized = raw.replace("Z", "+00:00")
```
willem-bd: Minor cleanup: the manual `raw.replace("Z", "+00:00")` is unnecessary on the project's Python floor (`requires-python = ">=3.12"`). `datetime.fromisoformat` has parsed a trailing `Z` natively since 3.11, so this can be just `datetime.fromisoformat(raw)`. The replace is also global rather than suffix-anchored, which would mangle a `Z` appearing anywhere else in the string (not a live concern since only `createdAt` is parsed, but worth not keeping). The `tzinfo`-naive → UTC fallback below should stay. Behavior is unchanged today, so low priority.

backend/packages/harness/deerflow/config/memory_config.py
sontianye: Done in `e86df716`: added a staleness workflow step (step 5) to the Memory System → Workflow section in `backend/AGENTS.md` describing the pass, the candidate selection, the LLM judgment, and the apply-layer guardrail. Also added all five new fields to the Configuration block and updated the one-liner in the Config Schema section.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Done in `e86df716`: dropped the `raw.replace("Z", "+00:00")` and call `datetime.fromisoformat(raw)` directly. Also trimmed the docstring accordingly.

WillemJiang (approved):

PR #3863, opened 2026-06-29
fix: preserve read_file for skill progressive loading #3862

### LINKED ISSUE #3862: [bug] skill allowed-tools can remove read_file and break progressive skill loading
### Summary

When a skill declares `allowed-tools`, DeerFlow filters the final tool list down to the union of those allowed tools. This can remove `read_file`, which breaks the documented progressive skill loading flow because the lead agent is instructed to call `read_file` on `/mnt/skills/.../SKILL.md` when a user query matches a skill.

### Reproduction

1. Create or enable a custom skill whose `SKILL.md` declares only business tools, for example:

```yaml
---
name: data-query
description: Query business data.
allowed-tools:
  - dataagent_query
---
```

2. Start an agent with this skill enabled.
3. Ask a question that should match the skill.
4. The system prompt lists the skill and tells the model to read the skill file, but the final bound tool list no longer contains `read_file`.
5. If the model calls `read_file`, the run fails with an error similar to:

```text
read_file is not a valid tool
```

### Expected behavior

`allowed-tools` should still restrict business/action tools, but framework tools required for skill loading, at least `read_file`, should remain available so progressive skill loading continues to work.

### Actual behavior

`filter_tools_by_skill_allowed_tools()` only keeps tools explicitly listed by skills. If `read_file` is not repeated in every skill's `allowed-tools`, it is filtered out.

### Notes

This seems related to the skill design described in the slash skill RFC: non-slash, model-driven progressive loading via `read_file` should continue to work. The current filtering behavior can break that path for custom skills that correctly limit their business tools.

kongxiangxin:
Closes https://github.com/bytedance/deer-flow/issues/3862.

## Summary
- preserve `read_file` when skill `allowed-tools` filtering is active
- keep explicit skill tool allowlists restrictive for non-framework tools
- add regression coverage for progressive skill loading

## Test
- `uv run --group dev pytest tests/test_lead_agent_skills.py -q`

## Context
Skills prompt the model to progressively load matching skills by calling `read_file` on the skill's `SKILL.md`. Custom skills that declare `allowed-tools` can currently filter `read_file` out of the final tool list, causing progressive loading to fail with `read_file is not a valid tool`.

### DIFF
--- a/backend/packages/harness/deerflow/agents/lead_agent/agent.py
+++ b/backend/packages/harness/deerflow/agents/lead_agent/agent.py
@@ -43,7 +43,7 @@ from deerflow.agents.thread_state import ThreadState
 from deerflow.config.agents_config import load_agent_config, validate_agent_name
 from deerflow.config.app_config import AppConfig, get_app_config
 from deerflow.models import create_chat_model
-from deerflow.skills.tool_policy import filter_tools_by_skill_allowed_tools
+from deerflow.skills.tool_policy import SKILL_LOADING_TOOL_NAMES, filter_tools_by_skill_allowed_tools
 from deerflow.skills.types import Skill
 from deerflow.tracing import build_tracing_callbacks
 
@@ -504,7 +504,7 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
         # Keep the bootstrap skill set intentionally narrow so agent creation
         # remains deterministic before the custom agent's own config exists.
         raw_tools = get_available_tools(model_name=model_name, subagent_enabled=subagent_enabled, app_config=resolved_app_config) + [setup_agent]
-        filtered = filter_tools_by_skill_allowed_tools(raw_tools, skills_for_tool_policy)
+        filtered = filter_tools_by_skill_allowed_tools(raw_tools, skills_for_tool_policy, always_allowed_tool_names=SKILL_LOADING_TOOL_NAMES)
         final_tools, setup = assemble_deferred_tools(filtered, enabled=resolved_app_config.tool_search.enabled)
         return create_agent(
             model=create_chat_model(name=model_name, thinking_enabled=thinking_enabled, app_config=resolved_app_config, attach_tracing=False),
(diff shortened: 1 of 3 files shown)


ShenAC-SAC:
@kongxiangxin Thanks for the fix. I agree with the direction here: #3862 is a real lead-agent progressive-loading bug, because the prompt asks the model to read the matching skill's `SKILL.md` with `read_file`, while an `allowed-tools` allowlist can currently filter `read_file` out.

One blocker I found as a supplement to the automated review: `filter_tools_by_skill_allowed_tools()` is shared by both the lead agent and subagents. Making `read_file` always allowed by default fixes the lead-agent progressive-loading path, but it also changes the subagent allowlist contract. Subagents already load and inject their skill content before agent creation, so a subagent skill that only allows `bash` now also receives `read_file` even though it did not opt into that tool.

I rechecked this against the latest `upstream/main` and a temporary merge of this PR:

- `uv run --group dev pytest tests/test_lead_agent_skills.py -q` passes.
- `uv run --group dev pytest tests/test_subagent_executor.py::TestSkillAllowedTools -q` fails in the two mixed `allowed-tools` cases because `read_file` is now present alongside `bash`.
- The PR's `backend-unit-tests` check is failing for the same reason.

Could you either scope the `read_file` always-allowed exception to the lead-agent progressive-loading path, or explicitly update the subagent policy/tests/docs if the intended contract is that `read_file` is always an infrastructure tool there too?

kongxiangxin:
Addressed this by scoping the `read_file` infrastructure exception to the lead-agent skill progressive-loading path only. The shared skill allowed-tools policy now defaults to no extra always-allowed tools, so subagents preserve the declared allowed-tools contract.

Updated commit: `d2f9b654 fix(skills): preserve read_file for lead skill loading (#3862)`

Verification:
- `uv run --group dev pytest tests/test_lead_agent_skills.py tests/test_subagent_executor.py::TestSkillAllowedTools -q`
- `uv run --group dev pytest -q`

WillemJiang (approved):

PR #3866, opened 2026-06-29
feat(community): add Brave image search community tool

18062706139fcz:
## Summary
- add a Brave Image Search community provider using the official Brave images endpoint
- normalize image, thumbnail, source, and dimension fields with public URL filtering for image references
- wire Brave image search into doctor checks, config examples, package exports, and English/Chinese docs

## Tests
- cd backend && uv run pytest tests/test_brave_tools.py tests/test_doctor.py -q
- cd backend && uv run ruff check packages/harness/deerflow/community/brave/__init__.py packages/harness/deerflow/community/brave/tools.py tests/test_brave_tools.py tests/test_doctor.py ../scripts/doctor.py
- cd backend && uv run ruff format --check packages/harness/deerflow/community/brave/__init__.py packages/harness/deerflow/community/brave/tools.py tests/test_brave_tools.py tests/test_doctor.py ../scripts/doctor.py


backend/packages/harness/deerflow/community/brave/tools.py
```diff
@@ -48,55 +61,167 @@ def _coerce_max_results(value: object, *, default: int = _DEFAULT_MAX_RESULTS) -
...
+    This is a best-effort SSRF guard that rejects non-http(s) schemes,
+    ``localhost``, and private/non-global IP literals (including obfuscated
+    decimal/hex/octal encodings). It only inspects the URL string and cannot
+    catch public hostnames that resolve to internal IPs; any consumer that
+    actually downloads these URLs must re-validate the resolved IP at fetch time.
+    """
+    if not isinstance(value, str):
+        return ""
+    url = value.strip()
+    parsed = urlparse(url)
```
willem-bd: **`urlparse()` is uncaught and crashes the whole tool call on a malformed IPv6 URL.**

If Brave returns any item whose `properties.url`, `thumbnail.src`, or `url` is a malformed bracketed-IPv6 literal such as `http://[::1/i.jpg` (unclosed bracket), this line raises `ValueError: Invalid IPv6 URL`. There is no try/except, so the exception propagates out of the per-item normalization loop and out of `image_search_tool`, turning the entire search into a crash instead of skipping one bad URL and returning the rest. Reproducible on Python 3.12.

```suggestion
    try:
        parsed = urlparse(url)
    except ValueError:
        return ""
```

backend/packages/harness/deerflow/community/brave/tools.py
```diff
@@ -48,55 +61,167 @@ def _coerce_max_results(value: object, *, default: int = _DEFAULT_MAX_RESULTS) -
...
+    if host == "localhost" or host.endswith(".localhost"):
+        return ""
+
+    try:
+        ip = ip_address(host)
+    except ValueError:
+        ip = _decode_ipv4(host)
+        if ip is None:
+            return url
+    return url if ip.is_global else ""
```
willem-bd, line 197: **NAT64-style IPv6 literals embedding a loopback/private IPv4 pass the SSRF guard as "safe".**

`http://[64:ff9b::127.0.0.1]/i.jpg` (or `::10.0.0.1`) parses via `ip_address()` and `ip.is_global` returns `True`, so the URL is returned as safe. This is the same loopback/private bypass class the guard explicitly decodes obfuscated IPv4 literals to block, but reached via the IPv6 path. Consider also rejecting IPv4-mapped/embedded addresses (e.g. check `ip.ipv4_mapped` and the NAT64 `64:ff9b::/96` prefix, or reject any non-global embedded v4).

backend/packages/harness/deerflow/community/brave/tools.py
```diff
@@ -117,3 +242,93 @@ def web_search_tool(query: str, max_results: int = 5) -> str:
...
+            continue
+
+        normalized_results.append(
+            {
+                "title": item.get("title", ""),
+                "image_url": image_url,
+                "thumbnail_url": thumbnail_url,
+                "source_url": safe_source,
+                "source": item.get("source", ""),
+                "width": properties.get("width"),
```
willem-bd: **Reported `width`/`height` describe the dropped original image even when only the thumbnail URL survives.**

For an item with an unsafe original (`properties.url = http://127.0.0.1/...`, `width: 1920, height: 1080`) but a safe thumbnail, `image_url` becomes `""` and `thumbnail_url` becomes the safe thumbnail, yet `width`/`height` are still taken from `properties` (the dropped original) rather than the `thumbnail` dict's own dimensions. The returned URL and its reported dimensions end up mismatched. Minor data-correctness issue.

willem-bd (commented):
Automated review (precision-focused). Three findings below — one crash, one SSRF-guard gap, one minor metadata mismatch.

WillemJiang:
@18062706139fcz Thanks for your contribution. Please take a look at the review comments.

WillemJiang (approved):

Issue #3322, opened 2026-05-30
MCP servers lack per-user credential isolation — all users share the same API keys


tniuli:
### Problem
DeerFlow's MCP server configuration only supports global, shared credentials . All authenticated users access MCP-backed tools (GitHub, PostgreSQL, etc.) using the same API key/token configured in extensions_config.json :

This means:

1. No data isolation — User A can access User B's private GitHub repos (or vice versa) because they all use the same PAT.
2. No individual permission scoping — Fine-grained GitHub permissions (repo-level, org-level) cannot be enforced per user.
3. Single point of failure — If the shared token is revoked or expires, all users lose access simultaneously.
4. Auditability gap — API calls made through MCP tools are attributed to a single service account, not to the individual user who initiated the action.
### Current State
Component Per-user support Notes Model API keys ( config.yaml ) ❌ Global $ENV_VAR resolved at process start MCP env (stdio servers) ❌ Global Bound at process spawn, immutable per-session MCP headers (HTTP/SSE servers) ❌ Global Static values from config MCP oauth (built-in) ❌ Global OAuthTokenManager caches one token per server OceanEngine interceptor ✅ Per-user Only integration with per-user token injection via mcpInterceptors

The OceanEngine integration ( deerflow.community.oceanengine.mcp_interceptor ) demonstrates that the infrastructure for per-user MCP auth already exists:

- resolve_runtime_user_id(runtime) resolves the current user from LangGraph runtime context
- mcpInterceptors in extensions_config.json supports pluggable interceptors
- MCPToolCallRequest carries runtime (with context["user_id"] ) into the interceptor chain
However, this pattern is hardcoded and non-reusable — each new MCP server that needs per-user auth requires a custom interceptor, a custom database table, and custom OAuth endpoints.

### Impact on Common Use Cases
- GitHub MCP — Users cannot access their own private repos or create issues/PRs under their own identity
- Google Drive / Notion MCP — Users cannot access their own documents
- Slack MCP — Users cannot post as themselves
- Any OAuth-protected API — Per-user authorization is impossible with the current architecture
### Proposed Direction
A general-purpose per-user MCP credential framework could include:

1. Unified credential store — A single database table (e.g., mcp_user_credentials ) to persist per-user tokens for any MCP server, instead of one table per integration
2. Generic interceptor builder — A factory function that generates per-user interceptors from a declarative config:
   ```
   {
     "mcpServers": {
       "github": {
         "type": "stdio",
         "userCredentials": {
[...]

PR #3871, opened 2026-06-29
feat(skills): request-scoped secrets for skills (closes #3861)

### LINKED ISSUE #3861: 支持调用方将请求级上下文(如用户身份、动态凭据)传递给 skill
## 背景与场景

DeerFlow 作为 AI 能力平台被业务系统集成时，经常需要 skill 代表**当前发起请求的终端用户**去调用外部系统——例如以该用户的身份查询 ERP 数据、访问其有权限的业务接口。

这类场景的共同特征是：
- 调用方（业务系统）在发起请求时持有该用户的凭据（access token、session 令牌等）
- 凭据是**短时效的**，随每次请求动态变化，无法预先配置在 skill 里
- skill 必须拿到这个凭据才能以正确的用户身份执行操作，而不是用一个写死的全局服务账号

## 当前的困境

目前没有可靠的方式让调用方把这类「本次请求的用户上下文」传进 skill。已知变通方案都有明显缺陷：

| 方案 | 问题 |
|---|---|
| 写进 prompt 让 LLM 传给 skill | 依赖模型行为，不可靠；凭据会暴露在对话记录和 trace 中 |
| 预配置在  | 只能是静态全局值，无法区分不同用户，也无法应对过期更新 |
| 每次上传一个上下文文件 | 引入额外延迟；上传文件对模型可见，有信息泄露风险 |

## 期望的能力

希望 DeerFlow 提供一种机制，让调用方在发起请求时携带的上下文（如用户 token）能够**安全、可靠地**传递给 skill 执行环境：

- 每次请求独立，支持短时效凭据
- 凭据不进 prompt 和 trace
- 不依赖 LLM 中转

## 参考

MCP tool 侧已有类似能力： 允许在工具调用时从请求上下文中读取凭据并注入（见 #3322）。sandbox skill 目前缺少对等的机制。如果能以类似的方式支持 skill 场景，即可满足上述需求。

fancyboi999:
## What & why

Closes #3861. Lets a caller pass **per-request, short-lived end-user credentials** (e.g. an ERP token, a per-user cloud API key) to a skill's sandbox scripts **without the value entering the prompt, tool arguments, the executed command string, traces, the checkpoint, or the run record** — the capability the issue asks for, and the skill-side counterpart of the still-open #3322 (MCP per-user credentials).

The design follows LangGraph's "out-of-band runtime dependency injection" pattern (the same `runtime.context` DeerFlow already uses) and Anthropic Agent Skills' "credentials live in the execution environment, not the prompt" model.

## How it works

```
caller run request
  └─ config.context.secrets = { "ERP_TOKEN": "<short-lived>" }   ← out-of-band, never a message
       └─ build_run_config → runtime.context["secrets"]  (not mirrored into configurable)
            └─ /skill-name activates → SkillActivationMiddleware resolves the skill's
               declared `required-secrets` against context.secrets → per-run injection set
                 └─ bash tool → Sandbox.execute_command(command, env=…)
                      ├─ LocalSandbox: subprocess env (platform creds scrubbed first)
                      └─ AioSandbox:  bash.exec(env=…) on a fresh session
                           └─ skill script reads $ERP_TOKEN  (value never in any message)
```

- **Declare** — a skill lists what it needs in `SKILL.md` frontmatter: `required-secrets:` (string list or `{name, optional}`). `name` is both the `context.secrets` key and the env var exposed to the script.
- **Carry** — the caller sends values in the run request's `context.secrets` (out-of-band). Reuses the existing `context` passthrough; never mirrored into `configurable`.
- **Bind** — only the slash-activated skill's declared secrets are injected, scoped to that run.
- **Inject** — `Sandbox.execute_command` gains an `env` parameter (Local + AIO). `env=None` is fully backward-compatible.
- **Scrub** — `execute_command` no longer leaks the Gateway's `os.environ` to skill subprocesses: secret-looking names (`*KEY*`/`*SECRET*`/`*TOKEN*`/`*PASSWORD*`/`*CREDENTIAL*`/`*DSN*` + a connection-string denylist like `DATABASE_URL`/`REDIS_URL`/`GH_PAT`) are dropped. A skill can only receive what the caller explicitly supplies; a caller-supplied value wins over a scrubbed host value of the same name (the per-user-key-overrides-shared-key case).

## Leak surfaces — all sealed (verified)

| Surface | How |
|---|---|
| prompt | value travels via `runtime.context`, never a message |
| trace | `tracing/metadata.py` never copies `context`; `redact_secret_context_keys` available defensively |
| checkpoint | secrets live on `runtime.context`, not graph state |
| audit | the activation journal records secret **names** only |
| bash stdout | `mask_secret_values` redacts injected values from tool output before it re-enters context |
| run record + run API | `start_run` persists `redact_config_secrets(body.config)`, so `runs.kwargs_json` and `RunResponse.kwargs` never carry the secret |
| inherited env | platform creds scrubbed from the sandbox subprocess |

## Verification (evidence, not assertion)

- **Unit/integration tests** — new `tests/test_skill_request_scoped_secrets.py` (parser, carrier, activation binding, env scrub, the leak surfaces, a real-subprocess end-to-end) + updated `tests/test_gateway_services.py` and `tests/test_local_sandbox_encoding.py`.
- **Full backend suite (CI-equivalent): green**, all regression tests pass; blocking-IO gate passes.
- **Real gateway end-to-end (real LLM + skill activation + bash):** a request with `context.secrets` ran a skill whose script used the injected secret; the value reached the sandbox subprocess and appeared in **none** of the surfaces above (verified by scanning the run response, the run API, the trace events, and the full per-thread data dir).
- **Real third-party cloud call:** a skill called an external cloud API using a request-scoped API key supplied via `context.secrets`. The call succeeded (`HTTP 200`); a same-named key that happened to exist in the Gateway env was scrubbed and never used; the caller's key leaked nowhere. (Credential redacted from this report.)

## Notes for reviewers

- This PR also fixes a **pre-existing latent bug** surfaced during end-to-end testing: `InputSanitizationMiddleware` wraps user input in `--- BEGIN USER INPUT ---` markers, but only preserved the original text for upload / IM-channel messages — so slash-skill activation silently failed for plain text messages. The sanitizer now preserves the original text in `ORIGINAL_USER_CONTENT_KEY` (additive; sanitization behaviour unchanged), which `get_original_user_content_text` already expected. Slash activation now works for all messages.
- `build_run_config` now sets `configurable.thread_id` on the `context` path too — the checkpointer requires it, and the prior behaviour produced a `KeyError: 'thread_id'` for context-driven runs.

## Scope / non-goals

- Only `/slash`-activated skills receive secrets (autonomously invoked enabled skills do not).
- No persistence/vaulting — secrets are request-lived only.
- The MCP per-user-credential gap (#3322) is a sibling and is not addressed here, though `context.secrets` is designed to be reusable for it.

## Not verified

- No live Langfuse/LangSmith backend was exercised; trace-cleanliness was verified against DeerFlow's metadata builder and the in-graph callback payloads, not a live tracing UI.
- The AIO/Docker `bash.exec(env=…)` path is covered by unit tests against the SDK signature/return shape; it was not run against a live Docker sandbox (local sandbox was used for the live end-to-end).


yong326:
@fancyboi999 Code conflicts, please help resolve

willem-bd (commented):
Review findings from an automated pass. Focus: correctness bugs surfaced by the new `env` kwarg path, the secrets-carrier lifecycle, and output redaction. See inline comments.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -1431,7 +1459,7 @@ def bash_tool(runtime: Runtime, description: str, command: str) -> str:
             max_chars = sandbox_cfg.bash_output_max_chars if sandbox_cfg else 20000
         except Exception:
             max_chars = 20000
-        return _truncate_bash_output(sandbox.execute_command(command), max_chars)
+        return _truncate_bash_output(mask_secret_values(sandbox.execute_command(command, env=injected_env), injected_env), max_chars)
```
willem-bd, line 1472: **E2BSandbox.execute_command signature was not updated — this line breaks bash for every E2B user.**

`bash_tool` now unconditionally passes `env=injected_env` here (and at line 1449 on the local branch). The abstract `Sandbox` (`sandbox.py`), `LocalSandbox`, and `AioSandbox` all accept the new `env`/`timeout` kwargs, but `community/e2b_sandbox/e2b_sandbox.py:121` is still `def execute_command(self, command: str)`. Passing any keyword argument named `env` (including `env=None`) raises `TypeError: execute_command() got an unexpected keyword argument 'env'`, which is then swallowed by the outer `except Exception` and returned to the model as `"Error: Unexpected error executing command: ..."`.

Result: every bash tool call fails on E2B deployments after this PR, even when the caller sends no secrets. Please update `E2BSandbox.execute_command` to accept `env=None, timeout=None` (and, ideally, add a test covering the new signature).

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
```diff
@@ -242,11 +245,51 @@ def _prepare_model_request(self, request: ModelRequest, *, hook: str) -> ModelRe
...
+        injected: dict[str, str] = {}
+        missing: list[str] = []
+        for req in activation.required_secrets:
+            if req.name in request_secrets:
+                injected[req.name] = request_secrets[req.name]
+            elif not req.optional:
+                missing.append(req.name)
+
+        if injected:
+            context[ACTIVE_SECRETS_CONTEXT_KEY] = injected
```
willem-bd, line 294: **`ACTIVE_SECRETS_CONTEXT_KEY` is never cleared, so a later skill in the same run inherits an earlier skill's secrets.**

The write here is gated by `if injected:`, and the whole function short-circuits earlier when `activation.required_secrets` is empty. So:

- Turn 1: `/skill-A` (declares `A_TOKEN`, caller supplied it) → `context[__active_skill_secrets] = {A_TOKEN: v}`.
- Turn 2: `/skill-B` where either `required-secrets` is empty, or the caller didn't supply `B_TOKEN` → the write is skipped and `{A_TOKEN: v}` is still sitting on `context`.

`bash_tool` in skill-B's turn then reads `read_active_secrets(context)` and injects `A_TOKEN` into skill-B's subprocess (and into `mask_secret_values`'s redaction set for skill-B's output). This violates the PR's stated contract that a skill only receives what the caller supplied **and** that skill declared.

Suggested fix: unconditionally reset the key on every activation before the short-circuits — e.g. `context[ACTIVE_SECRETS_CONTEXT_KEY] = {}` (or `pop`) at the top of `_apply_skill_secrets`, and only overwrite with `injected` if non-empty.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
```diff
@@ -154,6 +172,34 @@ def execute_command(self, command: str) -> str:
...
+        so injected request-scoped secrets are scoped to this command and never
+        persist across calls. Secret values travel in the structured ``env`` field,
+        never in the command string.
+        """
+        with self._lock:
+            try:
+                result = self._client.bash.exec(
+                    command=command,
+                    env=env,
+                    hard_timeout=float(self._DEFAULT_NO_CHANGE_TIMEOUT),
```
willem-bd: **`hard_timeout` vs `no_change_timeout` have different semantics — same constant, different behavior on the env path.**

The legacy path (this same file, `execute_command`) uses `shell.exec_command(no_change_timeout=self._DEFAULT_NO_CHANGE_TIMEOUT)` — an **idle** timeout: kill after 600 s with no output. Here, `bash.exec(hard_timeout=…)` is a **total wall-clock** kill. Same value, different meaning.

Consequence: a skill script that legitimately runs longer than 10 minutes wall-clock while emitting periodic progress lines (ETL jobs, `while … do echo …; sleep 30; done`) succeeds without secrets but is killed at exactly 600 s the moment the skill declares a required-secret and routes through `_execute_with_env`. The regression appears only when secrets are injected — surprising and hard to diagnose.

Suggested fix: if `bash.exec` exposes an idle/no-change timeout, use that; otherwise rename the constant (or introduce a distinct `_DEFAULT_HARD_TIMEOUT`) so the intent is explicit at the call site.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -1309,6 +1310,27 @@ def ensure_thread_directories_exist(runtime: Runtime | None) -> None:
...
+    flow into the tool result — and thus into the prompt and the trace. This is
+    the skill-specific fifth leak surface (the bash tool returns subprocess stdout,
+    unlike MCP tools). Replace each non-empty secret value with a redaction marker.
+    Longest values first so a value that is a substring of another is not partially
+    revealed.
+    """
+    if not injected_env or not output:
+        return output
+    for value in sorted((v for v in injected_env.values() if v), key=len, reverse=True):
+        output = output.replace(value, _SECRET_REDACTION)
```
willem-bd, line 1340: **`mask_secret_values` has no minimum-length floor, so short/common secret values shred unrelated bytes of tool output.**

`str.replace(value, '[redacted]')` matches anywhere, without boundaries. With `context.secrets = {ID: "42"}` (or any short version identifier, region code, tenant id, PIN) declared as a required-secret, every occurrence of `42` in stdout — exit codes, timestamps, sizes, JSON, paths — becomes `[redacted]`, silently corrupting the tool result the LLM reads back and blowing up the truncation-math on the caller path.

The sort-longest-first only helps when secrets contain each other; a single short value has no defence.

Suggested mitigations: (a) skip values below a threshold (say ≤ 6–8 chars) from the mask set — a redacted 3-char value is more likely noise than genuine leak protection; and/or (b) warn at parse time when `context.secrets` has a suspiciously short value.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
```diff
@@ -154,6 +172,34 @@ def execute_command(self, command: str) -> str:
...
+        The persistent-shell ``shell.exec_command`` API has no env parameter, so
+        injected commands use the ``bash.exec`` API which accepts per-command env.
+        Each call lets the sandbox auto-create a fresh session (no ``session_id``),
+        so injected request-scoped secrets are scoped to this command and never
+        persist across calls. Secret values travel in the structured ``env`` field,
+        never in the command string.
+        """
+        with self._lock:
+            try:
+                result = self._client.bash.exec(
```
willem-bd, line 220: **`bash.exec` is called with no `session_id`, so consecutive env-bearing bash calls in the same skill activation share no cwd / env exports / venv state.**

A typical skill run issues sequential bash calls like:

```
cd /mnt/user-data/workspace && source .venv/bin/activate && pip install foo
python run.py
```

Without secrets, both go through `shell.exec_command` on the persistent session and the second call sees the first call's cwd/PATH. With secrets injected, both go through `_execute_with_env` and each gets a **fresh** auto-created session: the second call starts in `$HOME`, sees no venv, no `foo`, and fails with `No such file` or `command not found`.

I understand the design tension (a reused session_id would keep secrets scoped-to-thread rather than scoped-to-call). Two options worth considering:

1. Use a stable per-thread `session_id` for env-bearing calls and rely on the fact that `bash.exec`'s `env` overrides the session env per command (verify with the SDK — if that guarantee holds, secrets don't persist across calls).
2. If session reuse isn't safe, at least pass `exec_dir` (e.g. the thread's workspace) so relative paths behave the same way as on the legacy path.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
```diff
@@ -154,6 +172,34 @@ def execute_command(self, command: str) -> str:
                 logger.error(f"Failed to execute command in sandbox: {e}")
                 return f"Error: {e}"
 
+    def _execute_with_env(self, command: str, env: dict[str, str]) -> str:
```
willem-bd, line 185: **Env-bearing path has no `ErrorObservation` retry.**

The legacy `execute_command` detects `_ERROR_OBSERVATION_SIGNATURE` in the persistent-shell output and retries on a fresh session — the standard recovery path for AIO's shared-container corruption mode under concurrency. `_execute_with_env` doesn't. A concurrent-caller race that corrupts the AIO shell output on the env path will return the corrupted string verbatim as tool output.

Even if the freshness of the auto-created session makes the corruption less likely, the two branches should probably share the same recovery contract.

starhobby:
When will 2.1.0 plan to launch?

WillemJiang:
> When will 2.1.0 plan to launch?

We will release it this month.

backend/packages/harness/deerflow/sandbox/tools.py
fancyboi999: Fixed. `E2BSandbox.execute_command` now accepts `env=None, timeout=None` and routes them to `commands.run(envs=..., timeout=...)` (the e2b SDK's `commands.run` supports `envs`). `env=None` passes no `envs` kwarg so existing non-secret callers are byte-for-byte unchanged. Regression tests added in `test_e2b_sandbox_provider.py` cover both the forwarded case and the `None`-compat case.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
fancyboi999: Fixed. `_apply_skill_secrets` now does `context.pop(ACTIVE_SECRETS_CONTEXT_KEY, None)` unconditionally before any short-circuit, so a later activation in the same run can never inherit an earlier skill's injection set. Two regression tests in `test_skill_request_scoped_secrets.py::TestActivationBindsSecrets` cover both reported paths: next skill declares no secrets → prior value cleared; next skill declares but caller omits → set empty (not stale).

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
fancyboi999: Good catch on the semantic mismatch. Fixed by introducing a dedicated `_DEFAULT_HARD_TIMEOUT` constant for the env path with a docstring noting that `bash.exec` exposes no idle/no-change timeout (unlike `shell.exec_command`'s `no_change_timeout`), so the two call sites are now named independently and can evolve without silently altering each other. Numeric value kept at 600 to stay broadly aligned with the legacy idle budget; a test pins the env-path call to `_DEFAULT_HARD_TIMEOUT`.

backend/packages/harness/deerflow/sandbox/tools.py
fancyboi999: Fixed. `mask_secret_values` now skips values shorter than `_MIN_MASK_LENGTH` (8 chars). A short declared value (region code, numeric id, PIN) would otherwise shred unrelated bytes of tool output — the redaction of a 3-char token is more likely noise than genuine leak protection. The secret is still injected into the subprocess; only the output mask skips short values. Test added covering both the short-value-skipped and long-value-redacted cases.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
fancyboi999: Intentionally keeping the fresh-session choice and documenting why on `_execute_with_env`.

The core tension is real, but a shared `session_id` for env-bearing calls would reintroduce the leak surface this PR exists to close: if `bash.exec`'s `env` merges into the session env rather than being strictly per-command (the SDK does not contractually guarantee per-command isolation), a request-scoped secret would ride the session into later commands in the same skill — exactly the scoped-to-call invariant the design preserves. Without that SDK guarantee, reusing a session_id is a regression risk I'm not willing to take in a security-focused change.

The fresh-session model matches `LocalSandbox`, where each `execute_command` is already a fresh subprocess (cwd is recovered via the command-prefix; sourced venv / exports never cross calls either). So AIO-with-secrets is now consistent with the local backend, not an outlier.

Practical mitigation for the cwd/venv case, now stated in the docstring: skills that need setup must fold it into a single command (e.g. `cd /mnt/user-data/workspace && source .venv/bin/activate && python run.py`). That's the same pattern LocalSandbox skills already need.

If a later SDK version documents per-command env isolation on a reused session, `exec_dir` (for cwd) or a stable session_id become safe upgrades — happy to revisit then.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox.py
fancyboi999: Fixed. `_execute_with_env` now checks for `_ERROR_OBSERVATION_SIGNATURE` and retries on another fresh session, sharing the legacy persistent-shell recovery contract. Refactored the single-shot into a `_run_bash_exec` helper so the retry path is the same env-bearing call (secrets stay out of the command string and out of any shell session). Test added verifying the signature triggers a second call and the clean output wins.

fancyboi999:
## Review fixes — push `926c73e5`

Addressed the review from @willem-bd. CI is green across the board (backend-unit-tests, backend-blocking-io, lint-backend/frontend, Layer 1/2 golden).

### Resolved (5)

- **E2B bash breakage** (`tools.py`) — `E2BSandbox.execute_command` now accepts `env`/`timeout` and routes them to `commands.run(envs=, timeout=)`. The bash tool passes `env=` unconditionally, so the prior command-only signature raised `TypeError` on every e2b bash call. `env=None` is fully backward-compatible (no `envs` kwarg sent). Regression tests added.
- **Stale active secrets** (`skill_activation_middleware.py`) — `_apply_skill_secrets` now clears the active-secret set before each activation, so a later skill in the same run can never inherit an earlier skill's injection set. Two regression tests cover both reported paths (next skill declares nothing; caller omits the required value).
- **hard_timeout vs no_change_timeout** (`aio_sandbox.py`) — dedicated `_DEFAULT_HARD_TIMEOUT` for the env path with a docstring noting `bash.exec` exposes no idle/no-change timeout. Kept at 600 to stay aligned with the legacy idle budget; named independently so the two call sites evolve without silently altering each other.
- **Short-value mask corruption** (`tools.py`) — `mask_secret_values` skips values below `_MIN_MASK_LENGTH` (8 chars) so a short declared secret (region code, numeric id) cannot shred exit codes / timestamps / sizes in tool output. The secret is still injected into the subprocess; only the output mask skips short values.
- **ErrorObservation retry** (`aio_sandbox.py`) — the env path now detects `_ERROR_OBSERVATION_SIGNATURE` and retries on another fresh session, sharing the legacy persistent-shell recovery contract. Refactored into a `_run_bash_exec` helper so the retry stays on the env-bearing path (secrets never enter a shell session).

### Left open (1) — intentional

- **`session_id` reuse on the env path** (`aio_sandbox.py`) — I kept the fresh-session choice. A shared `session_id` could let request-scoped secrets ride the session env into later commands, and the SDK does not contractually guarantee per-command `env` isolation, so reuse is a leak-surface regression I'm not willing to take in a security-focused change. Fresh session matches the `LocalSandbox` model (each `execute_command` is already a fresh subprocess there). The trade-off (consecutive env-bearing calls do not share cwd/venv/exports) is documented on `_execute_with_env`; skills fold setup into a single command. Happy to revisit if a future SDK version documents per-command env isolation on a reused session. Full rationale is in the thread.

WillemJiang (approved):

PR #3874, opened 2026-06-29
fix: generate fallback title for interrupted first-turn runs

### LINKED ISSUE #3859: [bug] 中断对话时自动标题未生成
### Before you start

### Problem summary

中断对话时自动标题未生成

### Affected area(s)

### What happened?

在对话过程中，如果用户在模型流式输出期间主动中断对话（点击停止按钮），该对话不会自动生成标题。

### Expected behavior

即使对话被中断，系统也可以自动生成对话标题。

### Steps to reproduce

**复现步骤：**
1. 新建一个对话
2. 发送一条消息
3. 在模型开始输出回复后，点击停止按钮中断对话
4. 查看对话列表，显示“新对话”
5. 刷新页面，标题显示“Untitled”

**实际行为：**
中断后的对话没有标题，显示为 "Untitled" 或"新对话"。

### Relevant logs

```shell
null
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

xiawiie:
Fixes #3859

## Why

When a user interrupts the very first streaming response on a brand-new
thread, `TitleMiddleware.after_model` never runs, so the thread keeps
its default "Untitled" / "新对话" display name even after refresh
(reported in #3859, reproducible on `main`).

## What changed

- `runtime/runs/worker.run_agent` now invokes a new
  `_ensure_interrupted_title` helper in the `finally` block whenever a
  run ends with status `interrupted`. The helper:
  1. reads the latest checkpoint via the existing checkpointer,
  2. returns early if a title is already set (LLM titler beat us, or a
     prior interrupt already wrote one),
  3. asks `TitleMiddleware._generate_title_result(...,
     allow_partial_exchange=True)` for a local fallback title built
     from whatever messages are in `channel_values`,
  4. writes a new checkpoint with `metadata.source = "update"` and a
     bumped `step` so checkpoint history stays well-formed.
- The existing `thread_store.update_display_name` sync block in the
  same `finally` picks up the new title and writes it to
  `threads_meta.display_name`, so list/search APIs see the title.
- `TitleMiddleware`:
  - `_should_generate_title(..., allow_partial_exchange=True)` accepts
    a lone first-turn human message (covers the case where cancellation
    lands before any AI chunk is checkpointed).
  - `_message_type` / `_message_content` helpers tolerate the
    dict-form messages stored in checkpoint `channel_values` (the
    middleware itself sees `BaseMessage` objects, but the worker reads
    raw checkpoint state).
- Tests:
  - `tests/test_runtime_lifecycle_e2e.py::test_cancel_interrupt_generates_missing_title_from_checkpoint`
    end-to-end: starts a run on a title-enabled config, cancels with
    `action=interrupt` before the fake agent writes a title, asserts
    `GET /api/threads/{id}` and `POST /api/threads/search` both return
    the fallback title.
  - Four new unit tests in
    `tests/test_title_middleware_core_logic.py` cover
    `allow_partial_exchange`, existing-title precedence, and dict-form
    messages on both the predicate and the sync fallback paths.

## Surface area

- [ ] **Frontend UI**
- [x] **Backend API** — title now reliably populated on interrupted
  first-turn runs (`GET /api/threads/{id}`, `POST /api/threads/search`).
- [x] **Agents / LangGraph** — extends `TitleMiddleware`; adds a
  fallback path in `runtime/runs/worker.py`.
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [ ] **Default behavior change** — previously interrupted first-turn
  threads stayed "Untitled"; now they get a fallback title derived
  from the user message. No new config flag — gated by the existing
  `title.enabled` config.

## Bug fix verification

  `backend/tests/test_runtime_lifecycle_e2e.py::test_cancel_interrupt_generates_missing_title_from_checkpoint`
- Did it go red on `main` and green on this branch? **yes** — the
  `_ensure_interrupted_title` call site does not exist on `main`, so
  the assertion `thread.values.title == "Run lifecycle E2E prompt"`
  fails. With this PR the test passes.

## Validation

```
cd backend && PYTHONPATH=. uv run pytest tests/ -x --ignore=tests/blocking_io -q
# 4397 passed, 18 skipped in 122.96s
```

Targeted reruns: `tests/test_runtime_lifecycle_e2e.py`,
`tests/test_title_middleware_core_logic.py`,
`tests/test_title_generation.py` — all green.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** AI drafted the initial fix and E2E test; on this
follow-up commit I directed it to harden the `allow_partial_exchange`
predicate, make checkpoint-`step` arithmetic defensive, and add the
four unit tests above. I've read every line of the change and own it.


fancyboi999 (commented):
The shape is right — gating on `interrupted` in the finally hook, the `allow_partial_exchange` first-turn relaxation, and the dict-form `_message_type`/`_message_content` helpers (which also fix a latent `m.type` AttributeError on checkpoint-stored messages) all match the root cause. One P2 before merge:

**The fallback-title checkpoint write is dropped on the Sqlite/Postgres backends** — `put(..., new_versions={})` with no `channel_versions["title"]` bump means the title doesn't survive a read-back from the DB savers, so the fix only takes effect on the in-memory checkpointer. Inline has the details + a fix mirroring the restore path already in this file. (FYI the branch also currently conflicts with `main`.)

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -551,6 +557,45 @@ def _new_checkpoint_marker() -> dict[str, str]:
...
+    metadata["source"] = "update"
+    prev_step = metadata.get("step")
+    metadata["step"] = (prev_step + 1) if isinstance(prev_step, int) else 1
+    metadata["writes"] = {"runtime_interrupt_title": {"title": title}}
+
+    tuple_config = getattr(ckpt_tuple, "config", {}) or {}
+    tuple_configurable = tuple_config.get("configurable", {}) if isinstance(tuple_config, dict) else {}
+    checkpoint_ns = tuple_configurable.get("checkpoint_ns", "") if isinstance(tuple_configurable, dict) else ""
+    write_config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": checkpoint_ns}}
+    await _call_checkpointer_method(checkpointer, "aput", "put", write_config, checkpoint, metadata, {})
```
fancyboi999: **P2 — the fallback title isn't durable on the Sqlite/Postgres checkpointers.** This sets `channel_values["title"]` but passes `new_versions={}` and never bumps `channel_versions["title"]`. deer-flow supports `SqliteSaver`/`PostgresSaver` (`runtime/checkpointer/provider.py:68-91`), and LangGraph's DB savers store channel values in a separate version-keyed blob table: `put` strips inline `channel_values` and only writes blobs for channels listed in `new_versions`. So with `{}` the title blob is never persisted and `channel_versions["title"]` is unchanged — the title-sync block's fresh `aget_tuple` (`worker.py:425`) then reconstructs `channel_values` without it, so `display_name` stays "Untitled" after refresh, exactly as #3859 reports. It only works on the default `InMemorySaver` (which the e2e test uses).

The restore path in this same file already does it right (`worker.py:509-520` passes `new_versions=dict(channel_versions)`). Mirror that — bump the title version and declare it:
```python
channel_versions = dict(checkpoint.get("channel_versions", {}) or {})
get_next = getattr(checkpointer, "get_next_version", None)
cur = channel_versions.get("title")
next_version = get_next(cur, None) if callable(get_next) else (cur + 1 if isinstance(cur, int) else 1)
channel_versions["title"] = next_version
checkpoint["channel_versions"] = channel_versions
await _call_checkpointer_method(checkpointer, "aput", "put", write_config, checkpoint, metadata, {"title": next_version})
```
Adding a SqliteSaver-backed case to the e2e test would catch this — the current InMemory test passes despite the gap.

xiawiie: got it, I will do it better later, thanks for your exact comment

xiawiie:
Pushed `0a3614ee` (force-pushed after rebase onto latest `main` — branch is now `MERGEABLE`, no conflicts). Two follow-up commits address your P2 plus a second-pass review I did on my own:

**P2 fix (`05253957`)**
Now mirrors `_rollback_to_pre_run_checkpoint`: bumps `channel_versions["title"]` via `checkpointer.get_next_version` (with int/string fallbacks when unavailable) and declares `{"title": next_version}` in `aput`. The fallback title now survives a fresh `aget_tuple` on DB-backed savers — exactly the read-back path you flagged.

**Hardening + tests (`0a3614ee`)**
On the second pass I noticed a few more edge cases worth pinning, so I extracted the version bump into a `_bump_channel_version(checkpointer, current)` helper with an explicit invariant: *the returned version MUST differ from the prior*. It defends against three saver fault modes that would otherwise silently regress the fix:
- `get_next_version` not present (custom backends)
- `get_next_version` raising
- `get_next_version` returning an unchanged value
…and supports int / float / numeric-string / UUID-shaped string / None / bool seeds.

I also coerced `state.get("messages")` to `[]` on `None` in `TitleMiddleware._should_generate_title` and `_build_title_prompt` — without that, a partially-initialized checkpoint (rare but possible on early-interrupt) raised `TypeError: 'NoneType' object is not iterable` from the next() generator. Repro-verified by reverting the coercion.

**Test coverage (20 new tests, TDD red-green verified)**
- `_bump_channel_version` × 8 — every version type × every saver fault mode
- `_ensure_interrupted_title` × 5 boundary cases (title disabled, empty messages, `messages=None`, `aput` error propagation contract, idempotency across repeated calls on a real `InMemorySaver`)
- `test_ensure_interrupted_title_preserves_non_title_channel_versions` — pins that `new_versions` only contains `"title"` and other channels are untouched
- `test_ensure_interrupted_title_round_trip_with_real_sqlite_checkpointer` — disk-backed `AsyncSqliteSaver`, write → close → re-open → read; confirms the fallback title survives a connection boundary. (Per your suggestion.)
- `test_worker_finally_block_swallows_helper_exceptions` — even if the helper crashes, the worker still runs `threads_meta.update_status` and awaits `publish_end`, so the SSE stream closes cleanly.
- `TitleMiddleware` × 4 — `messages=None` on both predicate and prompt builder; OpenAI-style `role: user`/`role: assistant` dict normalization; partial-exchange × dict-form message.

**Red-green verification I ran locally** (since assertions that never go red are dead weight):
- Temporarily reverted `new_versions={"title": v}` → 4 new `_ensure_interrupted_title_*` tests went red as expected. Restored → all green.
- Temporarily reverted the `messages or []` coercion → 2 new `*_handles_none_messages_channel` tests went red with the exact `TypeError`. Restored → all green.

**Suite status**
`PYTHONPATH=. uv run pytest tests/ -x --ignore=tests/blocking_io -q` → **5215 passed, 18 skipped**.
`ruff check` + `ruff format --check` → clean on every touched file.

One thing I noticed while validating: in the version of `langgraph-checkpoint-sqlite` pinned in `pyproject.toml` (3.0.3), the sqlite saver actually uses a single-row layout (the full checkpoint is JSON-serialized into the `checkpoints` table), so it does happen to ignore `new_versions` and persist anyway. The fix is still strictly correct — `PostgresSaver` and the LangGraph v4 sqlite blob layout both honor `new_versions`, and a future `langgraph-checkpoint-sqlite` bump shouldn't quietly regress this — but I wanted to flag the observation in case it changes how you weight the priority. The mock tests are the real regression anchor either way.

Happy to iterate further if you spot anything else.

backend/packages/harness/deerflow/runtime/runs/worker.py
xiawiie: Done — `05253957` mirrors the snippet you posted (uses `get_next_version` when the saver exposes it, falls back to int/string bumps otherwise), and `0a3614ee` extracts the bump into `_bump_channel_version` with an explicit invariant (returned version ≠ prior) and tests for every saver fault mode I could think of (no `get_next_version` / raising / stuck on identity). I also added a disk-backed `AsyncSqliteSaver` round-trip case (write → close → re-open → read) as you suggested.

Full write-up + suite status in the top-level comment below. Thanks again — this was a great catch.

xiawiie:
Pushed `c763ed93` — one more consistency improvement. Want to flag upfront that this is **not a fix for a user-visible bug**, more a "the write should look like every other write in this file" cleanup with future-compatibility benefit.

**What I noticed**

`_ensure_interrupted_title` built `write_config` as `{"thread_id": ..., "checkpoint_ns": ...}` only — no `checkpoint_id`. LangGraph savers (`AsyncSqliteSaver`/`PostgresSaver`) read `configurable.checkpoint_id` from that config as the *parent* on insert (see `langgraph/checkpoint/sqlite/aio.py` `aput`), so with nothing there they write `parent_checkpoint_id = NULL`. Disk-backed repro:

```
[seed]    checkpoint_id = 1f173dbc...
[helper]  wrote title  = "Why is the sky blue?"
[result]  new cp parent = None   ← orphan in the graph
```

**Honest scoping of the impact**

What this does NOT break today:
- ✅ Frontend never reads `parent_checkpoint_id` (grepped — 0 usages in `frontend/src/`)
- ✅ Existing `regenerate` flow (`POST /runs/regenerate/prepare`) walks history by message-id matching, not by parent-pointer chain — verified by repro, regenerate correctly picks the title-bump checkpoint as base because its `channel_values["messages"] == [first_human]`
- ✅ Title sync to `threads_meta.display_name` is unaffected (that path reads `channel_values["title"]`, not parent links)

What it WOULD affect:
- If/when LangGraph time-travel (`runs.resume_from` / branching from a historical checkpoint) lands in deer-flow, an orphan title-bump checkpoint has no backward edge
- Any future history-visualization UI built on `BaseCheckpointSaver.alist()` would render the title-bump as a sibling of the seed rather than its descendant

**Why I'm submitting it anyway**

Every other checkpoint write in `runtime/runs/worker.py` (rollback path + middleware-driven writes) threads the prior `checkpoint_id` through. Keeping this one consistent is cheap (10-line diff) and prevents a "why is this one different?" review question later. The mock + real-saver tests give a regression anchor either way.

**Three new regression tests** (all against real `AsyncSqliteSaver`, disk-backed, fresh connections so the on-disk read path is exercised):

- `test_ensure_interrupted_title_links_new_checkpoint_to_its_parent` — asserts `latest.parent_config["configurable"]["checkpoint_id"]` equals the seeded id. TDD red-green verified: reverting the fix flips this red with `AssertionError: title-bump checkpoint must have a parent_config`.
- `test_ensure_interrupted_title_appears_in_history_with_audit_marker` — pins the audit contract. The title-bump intentionally *does* appear in `alist()` (audited writes belong in the saver), but `metadata.source == "update"` and `metadata.writes["runtime_interrupt_title"]` are unambiguous so any history UI can identify and group it.
- `test_ensure_interrupted_title_survives_immediate_next_turn` — cancel-then-immediate-user-follow-up scenario. Next agent turn appends `(user, ai)` without touching the title channel; fresh saver afterwards must still read back the fallback title. Pins the channel-version-blob invariant from the earlier `new_versions={"title": v}` fix (independent of the parent-id fix in this commit).

**Verification**
- `PYTHONPATH=. uv run pytest tests/ -x --ignore=tests/blocking_io -q` → **5222 passed, 15 skipped**.
- `ruff check` + `ruff format --check` clean.

If you'd rather keep PR scope tight and treat this as a separate follow-up (or drop entirely as YAGNI), happy to revert `c763ed93` and leave the title-bump as it was. The original `bc34d020` head is fully usable as-is.

xiawiie:
Two follow-up commits trimming the PR after a self-review pass:

**`90385de4`** — reverts `c763ed93` (the `parent_checkpoint_id` change). After grepping, that change wasn't fixing a user-visible bug: frontend doesn't read `parent_checkpoint_id`, the existing `regenerate` flow walks by message-id rather than parent-pointer, and there's no time-travel feature in the project today. It was a consistency improvement masquerading as a bug fix. If parent linkage matters later, it belongs in a separate PR with the use case it serves.

**`7bf304b0`** — trims 107 lines of over-engineered test coverage:
- `_bump_channel_version`: 8 tests → 2 (happy path + saver-error fallback). Dropped float / bool / numeric-string / UUID-string / missing-get-next-version / stuck-get-next-version — those were speculative scaffolding for savers we don't ship.
- `_ensure_interrupted_title`: dropped `returns_none_when_title_disabled`, `returns_none_with_no_user_message`, `returns_none_when_no_checkpoint` — defensive guards already exercised by the e2e test and the `handles_none_messages_channel` anchor.

Remaining coverage still pins every red-green-verified invariant: channel_versions bump, string-version bump, idempotency, sqlite round-trip, non-title channel preservation, aput-error contract, worker finally swallowing, partial-exchange + dict-form messages, dict-reminder filtering.

**Suite**: 5209 passed, 15 skipped. `ruff check` + `ruff format --check` clean.

PR is back to focused scope (the P2 fix + dict-reminder + partial-exchange) with proportionate test coverage.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -409,6 +409,12 @@ async def run_agent(
             except Exception:
                 logger.warning("Failed to persist run completion for %s (non-fatal)", run_id, exc_info=True)
 
+        if checkpointer is not None and record.status == RunStatus.interrupted:
```
Huixin615, line 416: [P1] Keep the run active until interrupted-title finalization completes
The run is already terminal before this helper executes, so create_or_reject() can admit another run on the same thread while _ensure_interrupted_title() is between aget_tuple() and aput(). The old worker then writes its copied snapshot after the new run has advanced the thread, making the stale checkpoint latest and potentially losing the new messages.
Please keep interrupted-title cleanup in an active/finalizing state, or serialize and re-read the latest checkpoint before applying a title-only update. A regression test should block the title write, attempt a second run, and verify that no newer messages can be overwritten.

backend/tests/test_runtime_lifecycle_e2e.py
```diff
@@ -581,6 +604,51 @@ def test_cancel_interrupt_stops_running_background_run(isolated_app):
...
+        thread_id = _create_thread(client, csrf_token)
+
+        created = client.post(
+            f"/api/threads/{thread_id}/runs",
+            json=_run_body(),
+            headers={"X-CSRF-Token": csrf_token},
+        )
+        assert created.status_code == 200, created.text
+        run_id = created.json()["run_id"]
+        assert controller.checkpoint_written.wait(5), "fake agent never wrote checkpoint"
```
Huixin615, line 667: [P2] Cover the real stop path and cancellation before a useful checkpoint exists
This test waits for checkpoint_written, whose state already contains the human message and complete AI response, then uses cancel?wait=true followed by explicit GET requests. The real frontend stop path aborts the stream and triggers cancellation without waiting, so this does not cover the reported UI behavior or cancellation before a useful checkpoint exists.
Please add coverage for an early cancellation with no usable checkpoint, using the existing graph_input user message as fallback, and verify that the thread/search title is refreshed after the real stop flow.

Huixin615 (dismissed):
The DB-saver durability issue from the previous review is fixed and the targeted tests pass. However, interrupted-title finalization can race with a new same-thread run and overwrite a newer checkpoint. The E2E test also does not cover the actual frontend stop path or cancellation before a useful checkpoint exists. Please address these cases before merge.

Alexlee19880625:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，您的邮件我已经收到！我会尽快回复或者阅读您的邮件！

Huixin615:
@xiawiie 感谢提交了pr 我对最新的代码进行了复现以及review 请看下评论 尝试进行复现并解决 有问题随时喊我

xiawiie:
Updated in `dfb2f234`.

What changed:
- P1: kept interrupted runs active during title finalization with a `finalizing` state, so same-thread runs are rejected until cleanup finishes.
- P1: hardened `_ensure_interrupted_title()` to re-read the latest checkpoint before writing and apply only a title-channel update, avoiding stale snapshot overwrites.
- P2: added an E2E case for the real frontend stop path: `POST /runs/{run_id}/stream?action=interrupt`, cancelled before a useful checkpoint exists, using `graph_input` as the fallback title source.
- Conflict: merged latest `origin/main` and resolved the `TitleMiddleware` conflict by preserving dict-message / partial-exchange support while keeping main's default local-title fallback and opt-in LLM title path.

Verification:
- `ruff check` / `ruff format --check` on touched backend files
- targeted runtime/title suite: `152 passed`
- backend suite: `5225 passed, 16 skipped`

The PR is now mergeable locally/GitHub-side; CI is queued on the new head.

Huixin615:
@xiawiie Thanks — the default reject path is now protected by finalizing. However, the race still exists for multitask_strategy="interrupt": create_or_reject() admits the replacement run before the old run enters title finalization, and the second aget_tuple() is not atomic with the following aput().  
I reproduced this on dfb2f234: after the replacement writes a checkpoint containing a second message, the old title aput() makes the stale first-message snapshot latest again. Please cover the interrupt/rollback replacement path as well, or prevent interrupted-title recovery from racing with an already-admitted replacement run.

The graph_input fallback is now covered, thanks. However, this test still does not exercise the frontend stop path used by the pinned @langchain/langgraph-sdk@1.6.0. useStream.stop() aborts the active SSE and then calls client.runs.cancel(), which posts to /runs/{run_id}/cancel with wait=false; it does not post to /stream?action=interrupt and drain an end event.
Please cover the actual abort + fire-and-forget /cancel path and verify that the current thread/sidebar cache is refreshed after title finalization.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -571,10 +584,12 @@ async def create_or_reject(
             if multitask_strategy not in _supported_strategies:
                 raise UnsupportedStrategyError(f"Multitask strategy '{multitask_strategy}' is not yet supported. Supported strategies: {', '.join(_supported_strategies)}")
 
-            inflight = [r for r in self._thread_records_locked(thread_id) if r.status in (RunStatus.pending, RunStatus.running)]
+            inflight = [r for r in self._thread_records_locked(thread_id) if r.status in (RunStatus.pending, RunStatus.running) or r.finalizing]
 
             if multitask_strategy == "reject" and inflight:
                 raise ConflictError(f"Thread {thread_id} already has an active run")
+            if any(r.finalizing for r in inflight):
```
willem-bd: This `any(r.finalizing ...)` reject fires for **every** strategy, not just `reject`. I confirmed at the manager level that with a `finalizing=True` record, `interrupt`, `rollback`, and `reject` all raise `ConflictError` — whereas before, `interrupt`/`rollback` would cancel the inflight run and proceed.

Since the default cancel is `wait=false` (the endpoint returns 202 immediately while the worker's `finally` title-write is still running), a client that stops a run and immediately resubmits with `interrupt` can get a spurious 409 during that window — a change to the "interrupt always supersedes" contract. `wait=true` closes the window (it awaits `record.task`), so this only bites the async-stop path.

The identity re-check inside `_ensure_interrupted_title` already guards against a newer run advancing the checkpoint mid-write, so consider limiting this hard reject to the `reject` strategy rather than all three.

willem-bd (commented):
Reviewed on the PR branch — the touched suites (`test_run_worker_rollback.py`, `test_title_middleware_core_logic.py`, and the two new E2E tests) all pass and the core mechanism is sound. Two things worth a look, posted inline.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -548,6 +559,162 @@ def _new_checkpoint_marker() -> dict[str, str]:
...
+        if existing_title:
+            return existing_title
+
+        result = middleware._generate_title_result(_title_generation_state(channel_values, graph_input), allow_partial_exchange=True)
+        title = result.get("title") if isinstance(result, dict) else None
+        if not title:
+            return None
+
+        base_identity = _checkpoint_identity(ckpt_tuple, checkpoint)
+        latest_tuple = await _call_checkpointer_method(checkpointer, "aget_tuple", "get_tuple", ckpt_config)
```
willem-bd, line 674: The `graph_input` fallback is effectively dead when no checkpoint exists. When `aget_tuple` returns `None`, `checkpoint`/`latest_checkpoint` each fall back to a fresh `empty_checkpoint()`, and two `empty_checkpoint()` calls generate **different** `id`s — so `latest_identity != base_identity` always holds, the loop `continue`s all 3 attempts, and `aput` is never called.

I reproduced this in isolation (title generation forced to succeed, `aget_tuple` → `None`): result was `title=None` with `aput` never awaited. This contradicts the documented behavior ("worker 会使用本次 `graph_input` 中的首条用户消息生成本地 fallback title").

It's masked in the normal flow because thread creation always seeds an `empty_checkpoint` first (`routers/threads.py`), so `aget_tuple` is never `None` there — but the `graph_input`-only path the docs describe never actually executes, and it would silently no-op for any thread lacking that seed checkpoint (e.g. stateless runs). Either treat a `None`/absent-checkpoint base as "no prior identity to match" so the first write proceeds, or drop the unreachable `graph_input` fallback and the doc claim.

xiawiie:
@Huixin615
Thanks again for the detailed follow-up review. I pushed a new update in `e2c5b2a`.

What changed:
- Serialized same-thread replacement execution around interrupted-title cleanup: a replacement worker now waits for older same-thread finalization before entering the graph/checkpoint path.
- Prevented stale interrupted-title recovery from writing once a later same-thread run has already been admitted, so an old title-only checkpoint cannot become the latest checkpoint over newer messages.
- Added regression coverage for the `multitask_strategy="interrupt"` admitted-replacement path and for replacement workers waiting on prior finalization.
- Updated the frontend stop path so `thread.stop()` refreshes the current thread, metadata, token usage, and sidebar/search caches even when the SDK stops via abort + fire-and-forget `/cancel?wait=false` without an `onFinish` event.
- Added coverage for fire-and-forget cancel before a useful checkpoint exists, including thread/search title refresh from the original graph input fallback.

Validation:
- `uv run pytest -q` -> 5251 passed, 16 skipped.
- `uv run pytest tests/test_run_worker_rollback.py tests/test_runtime_lifecycle_e2e.py tests/test_worker_langfuse_metadata.py -q` -> 57 passed.
- `CI=true pnpm test` -> 377 passed.
- `CI=true pnpm test tests/unit/core/threads/infinite.test.ts` -> 19 passed.
- `CI=true pnpm check` -> passed; only existing warnings in unrelated message-list files remain.
- `pnpm format` -> passed.
- `SKIP_ENV_VALIDATION=1 pnpm build` -> passed.
- Started the current frontend locally on `http://localhost:3300` and ran `tests/e2e/chat.spec.ts` against it -> 12 passed.
- `ruff check`, `ruff format --check`, and `git diff --check` -> passed.

This should cover both the interrupt/rollback replacement race and the actual frontend stop/cache refresh behavior called out in the review.

@willem-bd I will do the last part

backend/packages/harness/deerflow/runtime/runs/worker.py
xiawiie: I will do it right now

xiawiie: Addressed in `b598ec82`. `_ensure_interrupted_title()` now treats an absent checkpoint tuple as having no comparable base identity instead of comparing two fresh `empty_checkpoint()` ids. If both reads are absent, the helper writes the title from `graph_input`; if a real checkpoint appears between reads, it retries instead of writing stale state. Added a regression test that forces `aget_tuple -> None` and verifies `aput` is awaited with the graph-input title.

backend/packages/harness/deerflow/runtime/runs/manager.py
xiawiie: Addressed in `b598ec82`. `finalizing` now remains part of the inflight set for `reject`, but no longer creates a hard ConflictError for `interrupt`/`rollback`. I also added a regression test for the async-stop window: a finalizing interrupted run admits interrupt/rollback replacements without re-cancelling the cleanup task, while `reject` still returns ConflictError.

xiawiie:
@willem-bd 

I pushed `b598ec82` with the follow-up fixes for the latest review comments.

What changed:
- Kept `finalizing` runs in the inflight set for `multitask_strategy="reject"`, so reject still returns `ConflictError` while cleanup is active.
- Restored the `interrupt` / `rollback` superseding contract: replacement runs are now admitted even if the prior run is already in interrupted-title finalization.
- Avoided re-cancelling an already-finalizing interrupted run. This preserves the old worker's `finally` cleanup path (`_ensure_interrupted_title`, thread metadata sync, `publish_end`, and `set_finalizing(False)`) while the replacement run is admitted.
- Fixed the no-checkpoint `graph_input` fallback in `_ensure_interrupted_title()`: absent checkpoint tuples are treated as having no stable base identity, instead of comparing two fresh `empty_checkpoint()` ids. If a real checkpoint appears between reads, the helper retries rather than writing stale state.

Coverage added:
- Regression test for the async stop window: a finalizing interrupted run admits `interrupt` / `rollback` replacements without cancelling the cleanup task again, while `reject` still conflicts.
- Regression test for `aget_tuple -> None`: graph input is used to generate the fallback title and `aput` is called with the title update.
- Existing admitted-replacement checkpoint race test remains green, covering the no-overwrite behavior after a newer checkpoint is written.

Verification:
- `uv run pytest tests/test_run_worker_rollback.py tests/test_runtime_lifecycle_e2e.py tests/test_worker_langfuse_metadata.py -q` -> 59 passed
- `uv run pytest -q --ignore=tests/test_client_live.py` -> 5234 passed, 16 skipped
- `uv run pytest tests/test_runtime_lifecycle_e2e.py -q` -> 7 passed
- `CI=true pnpm test tests/unit/core/threads/infinite.test.ts` -> 19 passed
- Full frontend Playwright E2E against a live Next dev server -> 61 passed
- `uv run ruff check ...` -> passed
- `uv run ruff format --check ...` -> passed
- `CI=true pnpm check` -> 0 errors, 3 existing warnings in unrelated message-list files
- `git diff --check` -> passed

This keeps the original issue behavior intact: interrupted first-turn runs can still get a fallback title, fire-and-forget frontend stop paths refresh the relevant thread/sidebar/search caches, and replacement runs no longer risk making a stale title checkpoint the latest thread state.

Huixin615:
@xiawiie Thanks for you commit. Some details  you could take a look.  Or you could give me commit permissions and I'll help you fix it.

[P2] Preserve title recovery when a replacement run is serialized
The replacement worker now waits until the prior run finishes finalization, so skipping title recovery solely because has_later_run() is true drops the only opportunity to title the thread. If the old checkpoint already contains the first human message, the replacement adds a second human turn and normal TitleMiddleware will no longer generate a title.
I reproduced this on b598ec82: after admitting an interrupt replacement, the old checkpoint remains untitled, and the resulting two-user-turn state is rejected by the normal title predicate. Please let the old run finish its title write before releasing the replacement, or explicitly transfer title responsibility to the replacement.

[P2] Wait for backend finalization before the only cache refresh
SDK thread.stop() resolves after aborting the stream and firing the cancel request; it does not await the wait=false backend finalization. This helper therefore may invalidate/refetch before the title checkpoint and display_name are committed, cache the stale “New Conversation” result, and never refresh again.
Please await/poll the run’s terminal finalization or schedule a subsequent refetch, and add a test where backend title finalization is delayed and the sidebar eventually receives the generated title.

xiawiie:
I see, your comment is really exact and accurate, I will try to fix it

xiawiie:
Thanks for the follow-up review. I pushed commit 50ca8e97 to address both points.

What changed:
- Backend title recovery now distinguishes an admitted pending replacement from a replacement that has actually started. A pending interrupt/rollback replacement no longer suppresses the old interrupted run's fallback title write, so the old run can finish title recovery before the replacement enters the graph. The stale-overwrite guard is still preserved once a newer run is running/finalizing and may have advanced the checkpoint.
- Added a regression test for the serialized replacement case: the old run is held in finalization, an interrupt replacement is admitted and remains pending, then the old run must still write the fallback title.
- Frontend stop handling now performs the immediate cache invalidation and schedules one follow-up refetch after the fire-and-forget SDK stop/cancel path. This catches delayed backend title finalization and refreshes the sidebar/search cache after the title is committed.
- Added frontend tests for delayed title finalization and for the case where only sidebar/search can be refreshed because the stopped thread id is not available.
- Updated backend/frontend AGENTS docs to match the refined finalization behavior.

Validation:
- backend: uv run pytest tests/test_run_worker_rollback.py tests/test_runtime_lifecycle_e2e.py tests/test_worker_langfuse_metadata.py -q -> 60 passed
- backend full non-live suite: uv run pytest -q --ignore=tests/test_client_live.py -> 5235 passed, 16 skipped
- backend lint/format on touched files: ruff check + ruff format --check -> passed
- frontend unit: CI=true pnpm test tests/unit/core/threads/infinite.test.ts -> 21 passed
- frontend check: CI=true pnpm check -> passed with existing warnings only
- frontend dev server on :3000 + Playwright smoke: pnpm test:e2e tests/e2e/chat.spec.ts tests/e2e/thread-history.spec.ts -> 25 passed
- git diff --check -> passed

WillemJiang (approved):

Issue #3857, opened 2026-06-28
Lead agent redundantly delegates and produces duplicate content in long tasks (parent)


ggnnggez:
## Background

Running long tasks locally recently surfaced several clear waste/quality problems with related root causes. Consolidating them into one parent issue.

**Case 1: a single task spawned 9 redundant subagents**
Asked the agent to analyze a paper (Ultra mode). The lead dispatched "research" in three batches of 3 subagents each, but the three batches overlapped heavily in topic (academic / industry / theory researched repeatedly). It burned ~9.9M tokens, ran ~2 hours, and one subagent even hit the 30-minute timeout with zero output. The final report contained duplicate sections.

**Case 2: the same section was written 5 times in the report**
Asked the agent to produce a global AI news weekly. The final report was assembled by the lead itself, but the "US" section was appended 5 times in a row, so the file contained 5 copies of the same section. This run never triggered summarization and had full context (model: GLM-5.2).

## Root cause analysis

The two problems sit at two different layers and should be looked at separately:

**1. Planning layer — redundant delegation (Case 1)**
The summary/context did not record "which subtasks were already delegated and their status." At each planning checkpoint the lead treated research as a brand-new task and dispatched another batch → redundant delegation.

**2. Output layer — duplicate output (Case 2)**
The lead assembled the report in an "append-only, never read back" way, and the writes were not idempotent. Every time the model regenerated the same section, it stacked another copy. Case 2 proves this happens even with full context and no lost history — so it is a separate problem from "summary dropping state."

**3. Common thread — lack of deterministic guardrails, over-reliance on LLM self-check**
From "delegation dedup," "write idempotency," and "loop termination" to "output checking," everything currently relies on the model noticing:
- Subagents only have a concurrency cap (at most 3 at once), no total cap, and no dedup of "this subtask is nearly identical to one already dispatched."
- The existing loop detection dedupes by tool-call signature, but repeated writes differ slightly in text each time and slip through.
- Whether the output is correct also relies on model self-check; in Case 1 the lead even wrote a script afterward to clean up duplicates — showing such a check could be done in a few lines of deterministic code.

**4. No backstop that actually engages**
The token budget is disabled; subagents have no summarization and a high max_turns. When the model's self-regulation fails, the only deterministic backstop that fires is the timeout — and only after the tokens are already burned.

## Task breakdown (issues list)

[...]

PR #3877, opened 2026-06-29
feat(subagents): system-maintained delegation ledger to stop redundant re-delegation

### LINKED ISSUE #3876: [feat] Delegation ledger: stop the lead re-delegating already-dispatched subtasks
Sub-issue of #3857 (planning-layer delegation ledger).

## Problem
The lead re-delegates overlapping subtasks because nothing durable records "what was already delegated + its status" — that record is lost to summarization, so at each planning checkpoint the lead dispatches a fresh batch (Case 1 in #3857: 9 subagents in 3 near-duplicate batches).

## Solution
A **system-maintained** ledger kept in state and **re-injected into context every model call**:

- **`ThreadState.delegations`** — `list[{task_id, description, subagent_type, status}]` with a `merge_delegations` reducer (upsert by `task_id`; a terminal status is never downgraded). It lives in state, so it survives summarization (which only compresses `messages`).
- **`DelegationLedgerMiddleware`** (lead-only, when `subagent_enabled`):
  - `after_model` — derives entries from the messages (a `task` tool-call = a dispatch; its `ToolMessage` carries the status via the existing `subagent_status` contract) and upserts them into `delegations`.
  - `wrap_model_call` — reads `delegations` and injects a hidden `<delegated_subtasks>` `<system-reminder>` **for that call only** (ephemeral → no duplicate snapshots pile up in history).
  - Registered **before** `SystemMessageCoalescingMiddleware` so the injected SystemMessage is folded into the leading one.

## Scope / notes
- Lead-only; the LLM never writes the ledger itself.
- Status has at most a one-call lag (dispatch recorded immediately; terminal status on the next call).
- Deliberately **decoupled from the "reserved region"** item in #3857 — re-injection from `ThreadState` already makes it summarization-proof.
- This is a **soft** layer: it lowers redundant delegation but doesn't hard-guarantee it; the deterministic intent-dedup guardrail is a separate #3857 item.

## Acceptance
- `delegations` recorded in `ThreadState` (verified via checkpoint); injected each call; **not** present in persisted message history; unit tests for reducer / extract / format / middleware / registration.

ggnnggez:
Closes #3876. Part of #3857 (item ①, planning-layer delegation ledger).

### Why
The lead re-delegates overlapping subtasks because no durable record of "what was already delegated + status" survives summarization. This adds a system-maintained ledger, re-injected into context each model call.

### What
- **`ThreadState.delegations`** + `merge_delegations` reducer — upsert by `task_id`; a terminal status is never downgraded. Lives in state → survives summarization (which only compresses `messages`).
- **`DelegationLedgerMiddleware`** (lead-only, when `subagent_enabled`):
  - `after_model` derives entries from messages (task tool-calls → dispatches; their `ToolMessage`s → status via the existing `subagent_status` contract) and upserts into `delegations`.
  - `wrap_model_call` injects a hidden `<delegated_subtasks>` `<system-reminder>` for that call only (ephemeral → no duplicate snapshots accumulate in history).
  - Registered **before** `SystemMessageCoalescingMiddleware` so the injected SystemMessage is folded into the leading one.
- Pure helpers `extract_delegations` / `format_delegation_block`.
- Docs: `backend/AGENTS.md` (middleware chain + ThreadState).

### Design notes
- System-maintained (the LLM never writes the ledger); status has at most a one-call lag.
- Decoupled from the "reserved region" item — re-injection from `ThreadState` already makes it summarization-proof.
- Soft layer: lowers redundant delegation but doesn't hard-guarantee it (the deterministic intent-dedup guardrail is a separate #3857 item).

### Testing
- 15 unit tests (reducer / extract / format / middleware / registration); `make format` / lint clean.
- Verified live: `delegations` recorded in the checkpoint, and **not** present in persisted message history.


willem-bd (commented):
Thanks @ggnnggez. Adding one net-new finding on the duplicated terminal-status set.

**[P2] `TERMINAL_STATUSES` duplicates `SUBAGENT_STATUS_VALUES` with no drift guard**
- Location: `backend/packages/harness/deerflow/agents/thread_state.py` (new `TERMINAL_STATUSES` frozenset)
- Problem: `TERMINAL_STATUSES = frozenset({"completed", "failed", "cancelled", "timed_out", "polling_timed_out"})` is a hand-maintained copy of `deerflow.subagents.status_contract.SUBAGENT_STATUS_VALUES`. The code comment says "mirrors ... so `merge_delegations` can guard against status downgrades without importing the middleware/subagents layer." But that duplication is the load-bearing input to the reducer's no-downgrade guard (`prev["status"] in TERMINAL_STATUSES and entry["status"] not in TERMINAL_STATUSES`), and nothing pins the two together. `test_subagent_status_contract.py` already pins `SUBAGENT_STATUS_VALUES` against the shared `contracts/subagent_status_contract.json` fixture; `TERMINAL_STATUSES` has no equivalent. If a new terminal status is added to the contract later (e.g. an `"aborted"` value), `merge_delegations` will silently stop protecting it: a finished subtask of the new type could be regressed back to `"in_progress"` by a later `after_model` re-derivation from a partially-summarized message list — exactly the regression the guard exists to prevent.
- Evidence: `TERMINAL_STATUSES` is the only place the guard reads; no test asserts it equals `SUBAGENT_STATUS_VALUES`; `extract_subagent_status`/`make_subagent_additional_kwargs` validate against the contract tuple, so the rest of the system would already accept the new value while the reducer silently mishandles it.
- Suggested fix: import the values rather than re-declaring them — the import-cost concern in the comment is real but solvable without duplication, e.g. `from deerflow.subagents.status_contract import SUBAGENT_STATUS_VALUES` and `TERMINAL_STATUSES = frozenset(SUBAGENT_STATUS_VALUES)` (the import is a leaf module with no heavy deps), or keep the literal but add a pinning test: `assert TERMINAL_STATUSES == frozenset(SUBAGENT_STATUS_VALUES)` in `test_delegation_ledger.py` (or extend the existing contract test).
- Test: `def test_terminal_statuses_match_contract(): from deerflow.subagents.status_contract import SUBAGENT_STATUS_VALUES; from deerflow.agents.thread_state import TERMINAL_STATUSES; assert TERMINAL_STATUSES == frozenset(SUBAGENT_STATUS_VALUES)`.

ggnnggez:
@willem-bd Thanks for review.

Addressed the `TERMINAL_STATUSES` review point in 15cb51ff.

- `thread_state.TERMINAL_STATUSES` is now `frozenset(SUBAGENT_STATUS_VALUES)` (single source of truth) instead of a hardcoded literal set, so it can't drift from the status contract.
- Added a pinning test `test_terminal_statuses_derived_from_status_contract` asserting the derivation and that the non-terminal `in_progress` stays excluded.

On the original "keep thread_state import-light" note: I verified the leaf `status_contract` only imports `typing`, so it resolves cleanly mid-load of `subagents/__init__` — no new cycle. All real entry points (`deerflow.agents`, `deerflow.agents.thread_state`, `deerflow.agents.lead_agent.agent`) import fine; the only failing order (importing the `subagents` package first) is a pre-existing cycle, byte-identical with the change stashed. The two sets are currently identical, so behavior is unchanged.

Verification: delegation-ledger + status-contract tests (35 passed), harness-boundary + thread_state reducers (20 passed), ruff format/check clean.

WillemJiang (approved):

PR #3878, opened 2026-06-30
fix(frontend) : prevent stream cancellation on concurrent message submit

chetan655:
### What does this PR do?
Fixes a bug where pressing "Enter" or submitting a new message while the AI is currently streaming a response would unexpectedly abort the stream.

### Why is this necessary?
Previously, the form's `onSubmit` handler in `InputBox` overloaded the submit action with a "Stop" command when the UI status was `"streaming"`. This meant that if a user rapidly sent a second message before the first one finished, it would cancel the ongoing LangGraph stream (resulting in no response and no thread title being generated) and discard the new message.

### How does it work?
1. **Decoupled the Stop action:** Moved the `onStop` invocation out of `handleSubmit` and bound it directly to the `onClick` event of the visual "Stop" button icon (`PromptInputSubmit`).
2. **Graceful rejection:** If a user submits via the `PromptInput` (e.g. hitting Enter) while streaming, `handleSubmit` now returns a rejected promise and displays a "Please wait" toast. Returning a rejected promise correctly informs the `PromptInput` component not to clear the user's typed text. 

### Testing
- [x] Send a message, and while it is streaming, type a second message and hit Enter. The stream should continue uninterrupted, the typed message is preserved, and a toast appears.
- [x] Click the visual "Stop" (square) button while a response is generating. The stream should correctly abort.


frontend/src/components/workspace/input-box.tsx
```diff
@@ -368,8 +368,8 @@ export function InputBox({
   const handleSubmit = useCallback(
     (message: PromptInputMessage) => {
       if (status === "streaming") {
-        onStop?.();
-        return;
+        toast.info("Please wait for the current response to finish.");
```
zhfeng: Just wonder if this message could be with i18n supported?

WillemJiang: @chetan655  Could you take a look at it?

chetan655:
@zhfeng thank you for the suggestion. now toast message is i18n supported. 
@WillemJiang could you please review again?

zhfeng (approved):

WillemJiang (approved):

PR #3881, opened 2026-06-30
feat(community): add Browserless web_capture screenshot tool

18062706139fcz:
## Summary
- Add a `web_capture` community tool that renders a page via Browserless `/screenshot` and presents the image through DeerFlow's artifact system (complements the existing Browserless `web_fetch` provider).
- Harden it against SSRF, misleading captures, and silent overwrites:
  - **SSRF guard** — reject URLs resolving to private / loopback / link-local (incl. the `169.254.169.254` cloud-metadata endpoint) / reserved / multicast / unspecified addresses; literal IPs and every resolved DNS record are screened. Operators can opt out with `allow_private_addresses: true` for intentional internal targets.
  - **Target-status warning** — Browserless returns HTTP 200 for the render even when the target page itself responded 4xx/5xx; surface that via the `X-Response-Code` header so an error/anti-bot page isn't mistaken for valid visual evidence.
  - **Filename dedupe** — an explicit `filename` that collides now gets a `-1`/`-2` suffix instead of silently overwriting a prior capture.
- **Docs** — comment out `token: $BROWSERLESS_TOKEN` in all tool examples (an unset `$VAR` makes `AppConfig` fail at startup) and document the new `allow_private_addresses` option across `config.example.yaml`, `CONFIGURATION.md`, and the en/zh tool docs.

## Test plan
- [x] `cd backend && uv run pytest tests/test_browserless_client.py tests/test_doctor.py` — 78 passed
- [x] `cd backend && uv run ruff check` on the changed Python files — clean
- [ ] Manual: configure `web_capture` against a self-hosted Browserless and confirm a screenshot artifact is written, a non-2xx target surfaces the warning, and a private URL is rejected unless `allow_private_addresses: true`.


backend/docs/CONFIGURATION.md
```diff
@@ -235,14 +235,37 @@ tools:
...
 - `bash` - Execute bash commands
 
+Browserless can be configured as an opt-in visual capture tool:
+
+```yaml
+tools:
+  - name: web_capture
+    group: web
+    use: deerflow.community.browserless.tools:web_capture_tool
+    base_url: http://localhost:3032
```
WillemJiang, line 254: @18062706139fcz, please provide a way for us to start the local service here.  I cannot find any instructions here.

18062706139fcz: Thanks for the review! This point is addressed in `c0554d3`:

**Local service startup** — Added local Browserless startup instructions to `backend/docs/CONFIGURATION.md`, right after the `web_capture` config block. In short:

```bash
# Browserless listens on 3000 inside the container; map it to 3032 to match the
# default base_url (http://localhost:3032). Recent Browserless images always
# require a token (a random one is generated if you don't pass TOKEN), so set it:
docker run -d --name browserless -p 3032:3000 -e "TOKEN=local-dev-token" ghcr.io/browserless/chromium

# Then export the same token so the tool sends it (and uncomment `token: $BROWSERLESS_TOKEN` in the config):
export BROWSERLESS_TOKEN=local-dev-token
```

Verify before enabling the tool:

```bash
curl -sS "http://localhost:3032/screenshot?token=local-dev-token" \
  -H "Content-Type: application/json" \
  -d '{"url": "https://example.com", "options": {"type": "png"}}' \
  -o /tmp/browserless-check.png  # writes a PNG on success
```

For Docker Compose, run Browserless as a service and point `base_url` at the service name (e.g. `http://browserless:3000`) instead of `localhost`.

WillemJiang (approved):

PR #3883, opened 2026-06-30
feat(sandbox): add E2B sandbox provider

chengjoey:
## Why
Adds ``deerflow.community.e2b_sandbox.E2BSandboxProvider`` with parity to AioSandboxProvider: metadata-keyed per-thread persistence, server- side idle timeout, warm-pool reclaim with liveness checks, /mnt/user-data bootstrap symlinks, dead-sandbox auto-rebuild, and release-time mirror of agent outputs back to the host artifact directory.

## What changed
add `e2b sandbox` support, use `e2b` by this configuration:
```
sandbox:
    use: deerflow.community.e2b_sandbox:E2BSandboxProvider
    api_key: $E2B_API_KEY            # required (or via E2B_API_KEY env var)
    template: code-interpreter-v1     # default: e2b code-interpreter template
    domain: e2b.dev                  # optional; for self-hosted e2b
    idle_timeout: 600                # forwarded to ``set_timeout``
    replicas: 3                      # max concurrent sandboxes
    mounts:                          # one-shot uploads on sandbox start
      - host_path: /data/skills
        container_path: /home/user/skills
        read_only: true
    environment:                     # forwarded as e2b ``envs`` on create
      OPENAI_API_KEY: $OPENAI_API_KEY
```

## Surface area

## Screenshots / Recording
powered by e2b

image
image

## AI assistance

**Tool(s) used:** 

**How you used it:** 

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,10 @@ INFOQUEST_API_KEY=your-infoquest-api-key
 # MINIMAX_API_KEY=your-minimax-api-key  # OpenAI-compatible, see https://platform.minimax.io
 # STEPFUN_API_KEY=your-stepfun-api-key  # OpenAI-compatible, see https://platform.stepfun.com
 # VLLM_API_KEY=your-vllm-api-key  # OpenAI-compatible
+
+# E2B cloud sandbox API key — required only when using E2BSandboxProvider.
+# Sign up at https://e2b.dev/dashboard
+# E2B_API_KEY=your-e2b-api-key
 # FEISHU_APP_ID=your-feishu-app-id
 # FEISHU_APP_SECRET=your-feishu-app-secret
 
--- a/backend/docs/CONFIGURATION.md
+++ b/backend/docs/CONFIGURATION.md
@@ -274,6 +274,45 @@ When using Docker development (`make docker-start`), DeerFlow starts the `provis
 
 See [Provisioner Setup Guide](../../docker/provisioner/README.md) for detailed configuration, prerequisites, and troubleshooting.
 
+**E2B Cloud Sandbox** (runs sandbox code in [E2B](https://e2b.dev) cloud micro-VMs):
+
+```yaml
+sandbox:
+   use: deerflow.community.e2b_sandbox:E2BSandboxProvider
+   api_key: $E2B_API_KEY            # required; or set the E2B_API_KEY env var
+   template: code-interpreter-v1     # e2b sandbox template id
+   # domain: e2b.dev                # optional; for self-hosted e2b deployments
+   home_dir: /home/user             # /mnt/user-data is remapped under this directory
+   idle_timeout: 600                # forwarded to e2b's server-side set_timeout()
+   replicas: 3                      # max concurrent sandboxes per gateway process
+   mounts:                          # one-shot upload of host files at sandbox start
+     - host_path: /path/on/host
+       container_path: /home/user/shared
+       read_only: false
+   environment:                     # forwarded to the sandbox at create time
+     OPENAI_API_KEY: $OPENAI_API_KEY
+```
+
+`e2b-code-interpreter` is bundled as a core dependency of `deerflow-harness`,
+so no extra install step is needed; just supply your API key and switch the
+provider in `config.yaml`.
+
+Notes specific to `E2BSandboxProvider`:
+
+- Each DeerFlow thread is bound to its e2b sandbox via metadata
+  (`deer_flow_user`, `deer_flow_thread`), so the same thread reuses the same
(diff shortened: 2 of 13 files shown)


willem-bd (commented):
Thanks for the thorough provider — metadata-keyed discovery, warm-pool liveness checks, and the path-traversal guards all look solid. One issue worth addressing before merge:

[P2] download_file enforces the 100 MB cap only after the whole file is loaded into gateway memory
- Location: backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py — `download_file` (the `client.files.read(resolved, format="bytes")` call and the subsequent `len(data) > _MAX_DOWNLOAD_SIZE` check), and indirectly `e2b_sandbox_provider.py::_sync_outputs_to_host` which calls `download_file` per changed file.
- Problem: the effective code path calls `files.read(format="bytes")`, which the e2b SDK implements as `return bytearray(r.content)` — i.e. it materializes the entire file in the gateway process before returning. The `_MAX_DOWNLOAD_SIZE` (100 MB) guard is then applied *after* the full file is already in memory, so the cap cannot actually prevent a large file from being loaded. The streaming/iterable branch that does enforce the cap incrementally is effectively dead code, because `format="bytes"` never returns an iterator. This is a regression relative to AioSandbox.download_file, which streams chunks and enforces the cap incrementally so it never holds more than ~100 MB.
- Evidence: e2b SDK `Filesystem.read(..., format="bytes")` returns `bytearray(r.content)` (full body, not a stream). In the documented hosted/multi-tenant use case for this provider, a workload can write a multi-GB file under `/mnt/user-data/outputs`; at release time `_sync_outputs_to_host` auto-pulls every changed output file through `download_file` (no explicit download tool call required), loading each fully into memory and risking an OOM that takes down the shared gateway for all tenants.
- Suggested fix: read with `format="stream"` (which the SDK returns as an `Iterator[bytes]`) and keep the existing incremental-size accumulation so the cap is enforced before exceeding it; alternatively `stat` the remote file size first and refuse before reading. Apply the same bound in `_sync_outputs_to_host`.
- Test: add a unit test where the fake `files.read` returns bytes larger than `_MAX_DOWNLOAD_SIZE` (or a stream whose cumulative size exceeds it) and assert `download_file` raises `OSError(EFBIG)` without buffering the whole payload.

chengjoey:
> Thanks for the thorough provider — metadata-keyed discovery, warm-pool liveness checks, and the path-traversal guards all look solid. One issue worth addressing before merge:
> 
> [P2] download_file enforces the 100 MB cap only after the whole file is loaded into gateway memory
> 
> * Location: backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py — `download_file` (the `client.files.read(resolved, format="bytes")` call and the subsequent `len(data) > _MAX_DOWNLOAD_SIZE` check), and indirectly `e2b_sandbox_provider.py::_sync_outputs_to_host` which calls `download_file` per changed file.
> * Problem: the effective code path calls `files.read(format="bytes")`, which the e2b SDK implements as `return bytearray(r.content)` — i.e. it materializes the entire file in the gateway process before returning. The `_MAX_DOWNLOAD_SIZE` (100 MB) guard is then applied _after_ the full file is already in memory, so the cap cannot actually prevent a large file from being loaded. The streaming/iterable branch that does enforce the cap incrementally is effectively dead code, because `format="bytes"` never returns an iterator. This is a regression relative to AioSandbox.download_file, which streams chunks and enforces the cap incrementally so it never holds more than ~100 MB.
> * Evidence: e2b SDK `Filesystem.read(..., format="bytes")` returns `bytearray(r.content)` (full body, not a stream). In the documented hosted/multi-tenant use case for this provider, a workload can write a multi-GB file under `/mnt/user-data/outputs`; at release time `_sync_outputs_to_host` auto-pulls every changed output file through `download_file` (no explicit download tool call required), loading each fully into memory and risking an OOM that takes down the shared gateway for all tenants.
> * Suggested fix: read with `format="stream"` (which the SDK returns as an `Iterator[bytes]`) and keep the existing incremental-size accumulation so the cap is enforced before exceeding it; alternatively `stat` the remote file size first and refuse before reading. Apply the same bound in `_sync_outputs_to_host`.
> * Test: add a unit test where the fake `files.read` returns bytes larger than `_MAX_DOWNLOAD_SIZE` (or a stream whose cumulative size exceeds it) and assert `download_file` raises `OSError(EFBIG)` without buffering the whole payload.

done

WillemJiang (approved):

WillemJiang:
@chengjoey Thanks for your contribution.

PR #3886, opened 2026-06-30
feat(scripts): add redacted community support bundle generator

18062706139fcz:
## Why

Community bug reports for local setup, configuration, sandbox, and runtime
problems often arrive without the basic facts needed to triage them: toolchain
versions, whether `config.yaml`/`extensions_config.json` exist and are valid,
git state, and `make doctor` results. Maintainers then spend a round-trip just
asking for this. At the same time, asking users to paste raw config or logs
risks leaking secrets (API keys, tokens, DB passwords) into public issues.

This PR adds a one-command **support bundle** generator that collects the
high-signal diagnostics maintainers actually need, **with secrets redacted by
default**, and produces a ready-to-paste issue summary. It also emits a
machine-readable `triage.json` so maintainers and AI-assisted triage tools can
start from stable signals instead of re-deriving them.

## What it does

`make support-bundle` (wraps `scripts/support_bundle.py --include-doctor`) writes:

- **`*-issue-summary.md`** — a short Markdown block to paste into the issue body
  (triage status, active signals, doctor/git/version summary, next steps).
- **`*-issue-draft.md`** — a full issue scaffold for AI-assisted filing. It
  pre-fills known facts (OS, versions, git state) but keeps `REQUIRED`
  placeholders for repro steps, expected behavior, and problem summary, and
  explicitly instructs assistants **not to invent** unknown facts.
- **An evidence zip** under `.deer-flow/support-bundles/` containing:
  - `triage.json` — stable, machine-readable status + signals
    (`config_missing`, `models_missing`, `config_error`,
    `extensions_config_error`, `node_version_too_old`, `doctor_failed`,
    `dirty_worktree`, …) plus reporter/maintainer next steps.
  - `environment.json`, `config-summary.json`, `extensions-summary.json`,
    `git.json`, and (with `--include-doctor`) `doctor.json` — all redacted.
  - `README.md` / `manifest.json` describing the bundle and its privacy stance.

## Privacy / redaction

Redaction is applied to every collected value, not just by key name:

- Secret-like config keys (`api_key`, `token`, `secret`, `password`,
  `authorization`, `cookie`, `credential`, `private_key`) → masked.
- **URL userinfo** (`https://user:pass@host`, `redis://:pass@host`) → masked.
- **URL query secrets** (`?api_key=…`, `?access_token=…`) → masked.
- **CLI-arg secrets** in MCP `args` — both `--api-key VALUE` and
  `--api-key=VALUE` forms → masked.
- **Custom auth headers** (any `*header*` dict) → values masked.
- Bearer tokens, `sk-…` keys, and home-directory paths → masked.

The bundle **never** includes `.env`, raw conversation messages, or the
contents of files in thread workspaces/uploads/outputs. The optional
`--thread-id <id>` flag adds **file manifests only** (name, size, mtime), and
`thread_id` is validated against path traversal (rejects `.`, `..`, and any
`..` segment).

## How to use

```bash
# Generate a redacted bundle (runs make doctor too)
make support-bundle

# Or directly, e.g. to include a specific thread's file manifest
cd backend && uv run python ../scripts/support_bundle.py --thread-id <thread-id> --include-doctor
```

Then paste the printed `*-issue-summary.md` into your GitHub issue. Attach the
zip only if a maintainer asks for the evidence bundle, or if the summary alone
is not enough. Maintainers / AI triage should read `triage.json` first.

The bug-report issue template gains a "Support bundle summary" field pointing
users at this flow.

## Tests

`backend/tests/test_support_bundle.py` (23 cases) covers redaction (key-based,
URL userinfo, query, inline/split CLI flags, custom headers, env/bearer),
triage status derivation (incl. extensions parse errors), `thread_id`
path-traversal rejection, file-manifest behavior, and end-to-end bundle
assembly.

## Test plan

- [x] `cd backend && uv run pytest tests/test_support_bundle.py` — 23 passed
- [x] `cd backend && uv run ruff check` / `ruff format --check` — clean
- [x] End-to-end run with a crafted config containing 7 secret vectors — zero
      leaks in the generated zip
- [x] `make support-bundle` runs and writes summary + draft + zip

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -154,6 +154,16 @@ body:
         commit: abcdef1
 
   - type: textarea
+    id: support_bundle
+    attributes:
+      label: Support bundle summary
+      description: For local setup, configuration, sandbox, or runtime issues, run `make support-bundle` and paste the generated `*-issue-summary.md` here. AI-assisted reports can start from `*-issue-draft.md`, but every REQUIRED placeholder must be replaced before filing. Attach the zip only if a maintainer asks for the evidence bundle or the summary is not enough.
+      placeholder: |
+        ## DeerFlow support bundle summary
+        - Triage status: ...
+        - Active signals: ...
+
+  - type: textarea
     id: additional
     attributes:
       label: Additional context
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -51,7 +51,7 @@ deer-flow/
 ├── docker/                         # docker-compose files, nginx config, provisioner
 ├── skills/                         # Agent skills: public/ (committed), custom/ (gitignored)
 ├── contracts/                      # Cross-component JSON contracts (e.g. subagent status)
-├── scripts/                        # Root orchestration scripts invoked by the Makefile (check, configure, doctor, serve, docker, deploy, setup_wizard)
+├── scripts/                        # Root orchestration scripts invoked by the Makefile (check, configure, doctor, support_bundle, serve, docker, deploy, setup_wizard)
 ├── tests/                          # Root-level tests (currently tests/skills/ — public skill tests)
 └── docs/                           # Cross-cutting docs, plans, and design notes
 ```
@@ -69,6 +69,7 @@ Gateway API. Config schema and resolution order are documented in
 ```bash
 make setup       # Interactive setup wizard (recommended for new users)
 make doctor      # Check configuration and system requirements
+make support-bundle  # Generate redacted troubleshooting summary, AI issue draft, and optional zip
 make config      # Generate local config files from the examples
 make check       # Check that required tools are installed
 make install     # Install all dependencies (frontend + backend + pre-commit hooks)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -338,6 +338,38 @@ Every pull request triggers the following CI workflows:
 - [Architecture Overview](backend/CLAUDE.md) - Technical architecture
 - [MCP Setup Guide](backend/docs/MCP_SERVER.md) - Model Context Protocol configuration
 
+## Troubleshooting Bundle
+
+For setup, configuration, sandbox, or runtime issues, generate a redacted support
+summary before filing:
+
+```bash
+make support-bundle
+```
+
+The command prints reporter next steps, writes a `*-issue-summary.md` file that
+you can paste into the issue, writes a `*-issue-draft.md` file for AI-assisted
+issue filing, and writes an optional evidence zip under
+`.deer-flow/support-bundles/`. The zip includes toolchain versions, sanitized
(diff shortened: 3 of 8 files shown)


willem-bd (commented):
Thanks for this — the structured-vs-raw approach (parse to dict, drop comments, never read `.env`/file contents) is a solid privacy foundation, and DeerFlow's documented secret keys are well covered. One redaction gap is worth tightening before users are told the bundle is safe to attach to public issues.

[P2] Key-name allowlist misses literal secrets under non-standard keys (esp. MCP `env` blocks)
- Location: `scripts/support_bundle.py` — `SECRET_KEY_RE` (~L744) and `redact_data` (~L812), specifically how nested `env` / value mappings are handled.
- Problem: redaction of structured config keys on a fixed 9-keyword allowlist (`api_key|token|secret|password|passwd|authorization|cookie|credential|private_key`). When a key name is outside that list, the value falls through to `redact_text`, whose value-level fallback only catches `sk-…`, `Bearer …`, URL userinfo (`user:pass@`), and explicit `?api_key=` / `KEY=value` forms. A literal secret stored under a non-matching key name — and not in one of those value shapes — is emitted verbatim into `config-summary.json` / `extensions-summary.json`, i.e. into the artifact the bundle (README/issue-summary) tells users is safe to paste/attach publicly.
- Evidence: `redact_data` recurses into an `env` dict (the `env` key itself doesn't match), then masks each value only if its own key matches `SECRET_KEY_RE`. Common community MCP/tool env names are not on the list, e.g. `SUPABASE_SERVICE_ROLE_KEY`, generic `*_KEY` without "api" (`GEMINI_KEY`, `R2_ACCESS_KEY`), `aws_access_key_id`, `pwd`, `auth`, `pat`, `encrypt_key`. For a value like a Supabase service-role JWT or a Google `AIza…` key (neither `sk-`/`Bearer`/URL nor `KEY=` form, since it's already a parsed scalar), nothing redacts it. The example uses `$VAR` references, but users frequently hardcode secrets directly in `extensions_config.json` `env`, and that is exactly the file this bundle collects.
- Suggested fix: redact `env` (and `headers`/`args`) values by default rather than by key name — e.g. treat all `env` values as secret unless they are a `$VAR`/`${VAR}` reference, and/or add value-entropy heuristics for high-entropy tokens. Broaden the keyword set (`key`, `pwd`, `auth`, `pat`, `access[_-]?key`, `encrypt`, `signing`). A denylist-of-known-safe is safer than an allowlist-of-secret for a bundle marketed as public-safe.
- Test: extend `test_support_bundle.py` with an MCP `env` containing a non-keyword secret (e.g. `SUPABASE_SERVICE_ROLE_KEY` / `R2_ACCESS_KEY`) and assert the literal value does not appear in the generated zip.

Impact is conditional on non-standard key naming, so this is a hardening item rather than a guaranteed leak — but for a feature whose stated purpose is safe public sharing, the allowlist failure mode is the one most likely to surprise a user.

WillemJiang (approved):

PR #3889, opened 2026-06-30
feat(skills): per-user custom skill isolation with sandbox mounting

### LINKED ISSUE #2905: [Feature] User创建了自己的skill时应该像agent能保存在该user目录下的skills 并支持用户目录skills挂载到sandbox
目前试用skill create功能，参考文档配置DEER_FLOW_HOST_SKILLS_PATH 把 deerflow内置的skills目录能挂载到 /mnt/skills目录，这个很棒。 但是用skill-creator来创建用户自己的skill时，它底层使用 tool 工具 filesystem写文件只能写到事先在extensions_config.json配置中允许该工具写入的目录，比如如下 /app/project路径 (当然该路径事先会在sandbox config时配置挂载进入sandbox)

**问题是**   skill可以成功创建到/app/project/目录下，但是该目录不在deerflow skills搜索的配置中 (DEER_FLOW_HOST_SKILLS_PATH ) 环境变量中，当重新启动一个chat的时候deerflow lead agent没办法感知到它。尽管它已经在/app/project/目录下

**期望**  skill创建能像agent一样 使用harness tool 创建到user目录下进行隔离，同时保证创建好的skill 能被deerflow agent感知(使用)  即能自动挂载到sandbox里供agent使用 (新启动一个chat能直接找到它)

extensions_config.json
{
  "mcp_servers": {
    "filesystem": {
      "name": "filesystem",
      "command": "npx",
      "args": ["-y", "@modelcontextprotocol/server-filesystem","/app/project","/tmp"],
      "env": {},
      "enabled": true
    },

sandbox docker config.yaml

  mounts:
    - host_path: /home/chenwen9/sandbox/temp/deer-flow/project
      container_path: /app/project
      read_only: false
    #- host_path: /home/chenwen9/sandbox/temp/deer-flow/skills
    #  container_path: /app/skills
    #  read_only: false

ech0hol:
Fixes #2905

## Summary

Custom skills were stored in a single global directory (`skills/custom/`) shared by all users. User A's custom skills were visible, editable, and deletable by User B — no isolation. Additionally, the skill-creator workflow guided agents to use sandbox `write_file` for skill creation, which landed files in `/mnt/user-data/outputs/` (per-thread output directory) instead of the persistent skills directory, so newly created skills were invisible to the agent in new chats.

This PR introduces **per-user custom skill isolation**: each user's custom skills live in `{base_dir}/users/{user_id}/skills/custom/`, are mounted into the sandbox at `/mnt/skills/custom/` per-thread, and are loaded into the agent's system prompt and tool policy automatically.

## Issue Requirements vs. Implementation

Issue #2905 asks for three things. Here's exactly how this PR addresses each:

### 1. "Skill creation should save to the user's directory (like agents)"

**Issue text**: *"期望 skill 创建能像 agent 一样 使用 harness tool 创建到 user 目录下进行隔离"*

**Implementation**:

- New `UserScopedSkillStorage` (`user_scoped_skill_storage.py`, 262 lines) extends `LocalSkillStorage`, redirecting all custom skill writes from `skills/custom/` to `users/{user_id}/skills/custom/`
- `skill_manage_tool.py` now calls `get_or_new_user_skill_storage(resolve_runtime_user_id(runtime))` instead of the global singleton — every `create`/`edit`/`patch`/`delete`/`write_file`/`remove_file` action is user-scoped
- `routers/skills.py` — all custom skill endpoints (`GET`, `PUT`, `DELETE`, `install`, `content`, `history`, `rollback`) use `get_or_new_user_skill_storage(get_effective_user_id())`
- `config/paths.py` — new `user_custom_skills_dir(user_id)` method: `{base_dir}/users/{user_id}/skills/custom/`
- `skills/storage/__init__.py` — new `get_or_new_user_skill_storage(user_id)` factory with LRU cache (max 64 entries, double-check lock, `OrderedDict.move_to_end` for true LRU eviction)

**Result**: User A creates `my-skill` → stored at `users/userA/skills/custom/my-skill/SKILL.md`. User B cannot see, edit, or delete it.

### 2. "Created skills should be perceivable by the agent in new chats"

**Issue text**: *"当重新启动一个chat的时候 deerflow lead agent 没办法感知到它"*

**Implementation** (three layers of agent awareness):

**a) System prompt injection** (`agents/lead_agent/prompt.py`):
- `get_skills_prompt_section(user_id)` uses `get_or_new_user_skill_storage(user_id)` to load enabled skills
- Skill names + signatures injected into system prompt
- Per-user cache: `_enabled_skills_by_config_cache` keyed by `(id(app_config), user_id or "default")`, invalidated via `refresh_user_skills_system_prompt_cache_async(user_id)` → `invalidate_user_skill_cache(user_id)` + `_get_cached_skills_prompt_section.cache_clear()`

**b) Tool policy loading** (`agents/lead_agent/agent.py`):
- `_make_lead_agent` extracts `user_id` from LangGraph config or falls back to `get_effective_user_id()`
- Passes `user_id` to `get_enabled_skills_for_config(app_config, user_id=user_id)` which filters enabled skills per-user

**c) Skill activation middleware** (`middlewares/skill_activation_middleware.py`):
- `SkillActivationMiddleware` accepts `user_id` parameter
- Slash-skill resolution (`/skill-name`) uses per-user storage to find matching custom skills

**Result**: User creates a skill in chat A → opens chat B → agent's system prompt includes the skill, tool policy allows calling it, and `/skill-name` slash invocation works.

### 3. "Skills should be mounted to sandbox automatically"

**Issue text**: *"即能自动挂载到 sandbox 里供 agent 使用 (新启动一个 chat 能直接找到它)"*

**Implementation**:

**a) Per-thread sandbox mount** (`sandbox/local/local_sandbox_provider.py`):
- `_build_thread_path_mappings(thread_id, user_id=...)` now includes a `PathMapping` for `/mnt/skills/custom` → `users/{user_id}/skills/custom/` (read-only)
- Three sandbox mount points:
  - `/mnt/skills/public` → global `skills/public/` (static, shared)
  - `/mnt/skills/legacy` → global `skills/custom/` (static, read-only — for pre-migration fallback)
  - `/mnt/skills/custom` → per-user `users/{user_id}/skills/custom/` (per-thread, read-only)

**b) Virtual sub-directory overlay** (`sandbox/local/local_sandbox.py`):
- `list_dir("/mnt/skills")` now correctly shows `public/`, `custom/`, `legacy/` even though these are independent PathMappings with host paths outside the parent directory — the overlay patches missing virtual children back in

**c) Sandbox tools resolve via PathMapping** (`sandbox/tools.py`):
- `ls_tool` and `read_file_tool` no longer pre-resolve `/mnt/skills/...` paths via `_resolve_skills_path()` (which used the contextvar `get_effective_user_id()`)
- Instead, paths are passed through to `LocalSandbox._resolve_path()` which uses the `PathMapping` bound at `acquire()` time with the correct `user_id` from `runtime.context`
- This fixes a bug where the contextvar `user_id` could be lost in background tasks/worker pools, causing `/mnt/skills/custom` to resolve to the wrong user's directory

**d) Disabled skill access blocking** (`sandbox/tools.py`):
- New `_is_disabled_skill_path()` checks whether a path belongs to a disabled skill
- `ls_tool` and `read_file_tool` return an error message instead of file contents when the skill is disabled

**e) Skill-creator SKILL.md updated** (`skills/public/skill-creator/SKILL.md`):
- Added "DeerFlow Environment" section instructing the agent to use `skill_manage` tool (not `write_file`) for skill creation
- Ensures skills are persisted to per-user storage rather than per-thread temporary directories

### LEGACY fallback with shadow-mount semantics

When a user has no per-user custom skills yet, global `skills/custom/` skills appear as `SkillCategory.LEGACY` (read-only — visible but not editable/deletable). Once the user creates their first custom skill, LEGACY skills disappear (the user's own directory shadows the global one). This preserves backward compatibility during migration without leaking mutable access.

- New `SkillCategory.LEGACY` enum value (`skills/types.py`)
- `UserScopedSkillStorage._iter_skill_files()` yields LEGACY skills from global custom root only when user has no custom skills
- `ensure_custom_skill_is_editable()` gives a helpful error for LEGACY skills: *"create your own version with the same name — it will shadow the shared one"*
- `public_skill_exists()` checks both public and global-custom fallback so LEGACY skills are recognised as read-only

### Custom skill toggle

Previously, custom skills were hardcoded `enabled=True` and `PUT /api/skills/{name}` returned 400 for custom skills. Now:
- `SkillStorage.load_skills()` delegates to `is_skill_enabled()` for all categories (including custom, default True)
- Router no longer rejects toggle requests for custom skills
- Frontend removes `skill.editable` toggle guard (`skill-settings-page.tsx`)
- Frontend `type.ts` adds `category` field; `messages/utils.ts` handles LEGACY category

### Identity propagation

- `auth_middleware.py` — `effective_user_id` propagation for IM message scenarios (owner_user_id)
- `internal_auth.py` — internal auth headers carry user_id for service-to-service calls
- `client.py` — `DeerFlowClient` passes `user_id` to agent invocation config

### Migration script

`migrate_user_isolation.py` — `migrate_skills()` moves global `skills/custom/` content to per-user directories based on skill history metadata. Idempotent — re-running after successful migration is a no-op.

## Files Changed (37 files, +2382 / -311)

| Layer | Files | Description |
|-------|-------|-------------|
| **Storage** | `user_scoped_skill_storage.py` (new), `storage/__init__.py`, `skill_storage.py`, `local_skill_storage.py`, `types.py` | `UserScopedSkillStorage` class, LRU factory, LEGACY category, `is_skill_enabled` for custom |
| **Config** | `config/paths.py`, `config/extensions_config.py` | `user_custom_skills_dir()`, `_validate_user_id()`, docstring update |
| **Sandbox** | `local_sandbox_provider.py`, `local_sandbox.py`, `sandbox/tools.py` | Per-thread custom skills mount, virtual sub-directory overlay, PathMapping-based resolution, disabled skill blocking |
| **Agent** | `agents/lead_agent/agent.py`, `agents/lead_agent/prompt.py` | `user_id` extraction from LangGraph config, per-user prompt injection + cache |
| **Middleware** | `middlewares/skill_activation_middleware.py` | `user_id` parameter for slash skill resolution |
| **Tools** | `tools/skill_manage_tool.py` | Per-user storage in all skill manage actions |
| **API** | `routers/skills.py`, `auth_middleware.py`, `internal_auth.py`, `client.py` | User-scoped endpoints, effective_user_id propagation |
| **Frontend** | `skill-settings-page.tsx`, `skills/api.ts`, `skills/type.ts`, `messages/utils.ts` | Toggle fix, category field, LEGACY handling |
| **Skill content** | `skill-creator/SKILL.md` | Force `skill_manage` tool usage over `write_file` |
| **Migration** | `migrate_user_isolation.py` | `migrate_skills()` for legacy global custom skills |
| **Tests** | 12 test files (+1491 lines) | Full coverage: storage isolation, sandbox mounts, router CRUD, agent prompt, migration, sandbox tools |

## Surface area

- [x] Backend API (`skills.py` router: toggle enabled for custom skills; `auth_middleware.py` + `internal_auth.py`: effective_user_id propagation)
- [x] Agents (`agent.py`: user_id extraction from LangGraph config → middleware chain; `prompt.py`: per-user skill storage for prompt injection)
- [x] Middlewares (`skill_activation_middleware.py`: user_id parameter for slash skill resolution)
- [x] Sandbox (`local_sandbox_provider.py`: per-thread custom skills mount; `local_sandbox.py`: virtual sub-directory overlay for LEGACY; `tools.py`: skills/ACP paths resolved by sandbox PathMapping instead of pre-resolved contextvar)
- [x] Skills (`storage/__init__.py`: `get_or_new_user_skill_storage` LRU factory; `user_scoped_skill_storage.py`: new class; `skill_storage.py`: `load_skills` custom enabled follows config; `types.py`: `SkillCategory.LEGACY`)
- [x] Tools (`skill_manage_tool.py`: per-user storage; `skill_activation_middleware.py`: per-user resolution)
- [x] Config (`paths.py`: `user_custom_skills_dir`, `_validate_user_id`, `make_safe_user_id`; `extensions_config.py`: `is_skill_enabled` docstring update)
- [x] Frontend (`skill-settings-page.tsx`: removed `skill.editable` disabling toggle; `api.ts`: `response.ok` check for `enableSkill`; `type.ts`: category field; `messages/utils.ts`: LEGACY category handling)
- [x] Docs/tests/CI only (`migrate_user_isolation.py`, 12 test files, `skill-creator/SKILL.md`)

## Validation

```bash
# Backend lint
docker exec deer-flow-gateway sh -c "cd /app/backend && make lint"  # ✅ ruff check + format clean

# Backend tests (branch-relevant)
docker exec deer-flow-gateway sh -c "cd /app/backend && python -m pytest tests/test_user_scoped_skill_storage.py tests/test_skills_custom_router.py tests/test_lead_agent_prompt.py tests/test_lead_agent_skills.py tests/test_sandbox_search_tools.py tests/test_sandbox_tools_security.py tests/test_local_sandbox_provider_mounts.py tests/test_skill_storage_lifecycle.py tests/test_skill_manage_tool.py tests/test_client.py tests/test_client_e2e.py tests/test_slash_skills.py tests/test_migration_user_isolation.py -v"  # ✅ all passing

# Frontend
pnpm format:write && pnpm lint && pnpm typecheck  # ✅ all clean
```

## Implementation Chain (end-to-end)

```
1. Identity propagation
   HTTP request → auth_middleware (JWT/session → User)
   → contextvar: effective_user_id
   → LangGraph config["configurable"]["user_id"]
   → resolved_user_id threaded through:
     ├── build_middlewares(user_id=resolved_user_id)
     ├── SkillActivationMiddleware(user_id=resolved_user_id)
     ├── apply_prompt_template(user_id=resolved_user_id)
     └── get_enabled_skills_for_config(app_config, user_id=resolved_user_id)

2. Skill creation (skill_manage_tool)
   skill_manage_tool → get_or_new_user_skill_storage(resolve_runtime_user_id(runtime))
   → write_custom_skill(name, SKILL.md, content)
   → users/{user_id}/skills/custom/{name}/SKILL.md
   → refresh_user_skills_system_prompt_cache_async(user_id)

3. Agent awareness (new chat)
   _make_lead_agent(user_id) →
     ├── get_enabled_skills_for_config(app_config, user_id)
     │   → get_or_new_user_skill_storage(user_id).load_skills(enabled_only=True)
     │   → tool policy: allowed_skills = [public + user custom]
     ├── apply_prompt_template(user_id)
     │   → get_skills_prompt_section(user_id)
     │   → inject skill names + signatures into system prompt
     └── SkillActivationMiddleware(user_id)
         → slash skill resolution from per-user storage

4. Sandbox mount (per-thread)
   LocalSandboxProvider.acquire(thread_id, user_id)
   → _build_thread_path_mappings(thread_id, user_id)
   → PathMapping: /mnt/skills/custom → users/{user_id}/skills/custom/ (read-only)
   → LocalSandbox._resolve_path("/mnt/skills/custom/my-skill/SKILL.md")
   → correct host path via PathMapping (not contextvar)

5. LEGACY fallback (pre-migration)
   UserScopedSkillStorage._iter_skill_files()
   → 1. PUBLIC from global root (always)
   → 2. CUSTOM from user root (if user has custom skills)
   → 3. LEGACY from global custom root (only when user has no custom skills)
   → mounted at /mnt/skills/legacy/ (read-only)
```

### When to run the migration script

The migration script (`scripts/migrate_user_isolation.py`) is **not** executed automatically on startup. It must be run **manually** by the operator **once** after upgrading to this version, **only for existing deployments** that have legacy data in the global layout.

**Fresh installations** do not need to run it — the script is idempotent and will be a no-op if no legacy directories exist.

### How to run

- Python 3.12+ with the project's dependencies installed (or a running Docker stack)
- The `deerflow` package must be importable — this is what `PYTHONPATH=.` ensures when running locally

### Step 1: Dry-run (preview)

Inspect what will be moved without making any changes:

| Platform | Command |
|----------|---------|
| **macOS / Linux** | `cd backend && PYTHONPATH=. python scripts/migrate_user_isolation.py --dry-run` |
| **Windows (CMD)** | `cd backend && set PYTHONPATH=. && python scripts/migrate_user_isolation.py --dry-run` |
| **Windows (PowerShell)** | `cd backend; $env:PYTHONPATH="."; python scripts/migrate_user_isolation.py --dry-run` |
| **Docker** | `docker exec deer-flow-gateway sh -c "cd /app/backend && PYTHONPATH=. python scripts/migrate_user_isolation.py --dry-run"` |

### Step 2: Run migration (single-user / no-auth)

Defaults to `--user-id=default`, suitable for single-user or no-auth setups:

| Platform | Command |
|----------|---------|
| **macOS / Linux** | `cd backend && PYTHONPATH=. python scripts/migrate_user_isolation.py` |
| **Windows (CMD)** | `cd backend && set PYTHONPATH=. && python scripts/migrate_user_isolation.py` |
| **Windows (PowerShell)** | `cd backend; $env:PYTHONPATH="."; python scripts/migrate_user_isolation.py` |
| **Docker** | `docker exec deer-flow-gateway sh -c "cd /app/backend && PYTHONPATH=. python scripts/migrate_user_isolation.py"` |

### Step 3: Run migration (multi-user)

Specify the operator account that should inherit legacy global data (agents, custom skills, memory):

| Platform | Command |
|----------|---------|
| **macOS / Linux** | `cd backend && PYTHONPATH=. python scripts/migrate_user_isolation.py --user-id admin` |
| **Windows (CMD)** | `cd backend && set PYTHONPATH=. && python scripts/migrate_user_isolation.py --user-id admin` |
| **Windows (PowerShell)** | `cd backend; $env:PYTHONPATH="."; python scripts/migrate_user_isolation.py --user-id admin` |
| **Docker** | `docker exec deer-flow-gateway sh -c "cd /app/backend && PYTHONPATH=. python scripts/migrate_user_isolation.py --user-id admin"` |

> **Note:** Thread directories are migrated based on the `threads_meta` DB table — each thread goes to its recorded owner. Threads without an owner fall back to `default`. Agents, custom skills, and memory are assigned to the `--user-id` argument since they have no per-record ownership metadata.

> **Conflicts:** If a destination already exists, the legacy copy is moved to `{base_dir}/migration-conflicts/` for manual review rather than overwritten.`

### What the script migrates

| Data type | Legacy path | New path | Notes |
|-----------|------------|----------|-------|
| Threads | `threads/{id}/` | `users/{owner_id}/threads/{id}/` | Owner lookup from `threads_meta` DB table |
| Memory | `memory.json` | `users/{user_id}/memory.json` | Single global file → per-user |
| Custom agents | `agents/{name}/` | `users/{user_id}/agents/{name}/` | All legacy agents → specified user |
| Custom skills | `skills/custom/{name}/` | `users/{user_id}/skills/custom/{name}/` | All legacy custom skills → specified user |
| Skill history | `skills/custom/.history/` | `users/{user_id}/skills/custom/.history/` | Operation logs moved together |

### Conflict handling

If a per-user destination already exists, the legacy copy is moved to `migration-conflicts/` instead of being overwritten. The operator must manually review and merge conflicting files.

### Safety guarantees

- **Idempotent**: re-running is a no-op after successful migration
- **Dry-run**: preview mode logs all actions without modifying files
- **No overwrites**: conflicts go to `migration-conflicts/`, original per-user data is preserved
- **PUBLIC skills untouched**: only `skills/custom/` is migrated; `skills/public/` remains global

## Backward Compatibility

- **No breaking changes**: existing public skills work unchanged
- **LEGACY fallback**: pre-migration global custom skills remain visible (read-only) until users create their own
- **Migration script**: idempotent, can be run multiple times safely
- **Config**: no new required environment variables; `user_custom_skills_dir` derives from existing `base_dir`

## AI Assistance

- **Tools used**: VSCode（GLM5.1/5.2）
- **How used**: Architecture design, implementation coding, test writing, debugging (contextvar vs runtime user_id mismatch root cause analysis), migration script, PR description
- [x] I've read and understand every line of code in this PR


willem-bd (commented):
Hi @ech0hol — thanks for the thorough write-up. I focused on the per-user isolation boundary and found one issue worth addressing before merge.

[P2] Custom-skill enabled/disabled state is global-by-name, breaking per-user isolation
- Location: `backend/packages/harness/deerflow/config/extensions_config.py` (`is_skill_enabled`) + `backend/app/gateway/routers/skills.py` (`update_skill`, which now accepts CUSTOM toggles after the 400 guard was removed).
- Problem: This PR makes custom-skill *content* per-user, but the enabled/disabled *state* is still stored in the single global `extensions_config.json` keyed only by `skill_name` (`self.skills.get(skill_name)`). With per-user storage, two users can each own a distinct custom skill with the same name. When user A toggles their skill, the global entry is rewritten by name and user B's same-named skill inherits the state — so user A can disable (or re-enable) a skill that affects user B. This is a cross-user integrity breach of the very isolation boundary the PR establishes, and it is net-new: previously custom toggles were rejected with 400 and custom skills were a single global object, so name collisions across users could not exist.
- Evidence: `is_skill_enabled` resolves purely by name with no user dimension; `load_skills()` applies that state to every category by `(skill.name, skill.category)`; `update_skill` writes the toggle into `ExtensionsConfig.resolve_config_path()` (process-global file). The new `_is_disabled_skill_path()` in `sandbox/tools.py` reads the same global state, so the bleed also reaches the sandbox ls/read blocking.
- Suggested fix: Scope custom/legacy skill enabled-state per user — e.g. key the state by `(user_id, skill_name)` for non-public categories, or persist custom-skill enabled flags inside the per-user storage (`users/{user_id}/skills/...`) rather than the shared `extensions_config.json`. Public skills can remain global-by-name.
- Test: Two users each create a custom skill named `report-gen`; user A disables theirs; assert user B's `report-gen` stays enabled in `load_skills()`, is still present in the prompt section, and is not blocked by `_is_disabled_skill_path()`.

Maintainer note (not blocking): `UserScopedSkillStorage.__init__` calls `_validate_user_id()` (which *rejects* non-`[A-Za-z0-9_-]` ids), whereas thread/memory/agent dirs go through `make_safe_user_id()`/`prepare_user_dir_for_raw_id()` (which *sanitize*). If any effective user_id reaching skill storage (e.g. the `X-DeerFlow-Owner-User-Id` value from IM channels) is a raw, unsanitized identity, `get_or_new_user_skill_storage()` will raise and break skill listing/agent creation for that user. Worth confirming IM owner ids are normalized before they reach the contextvar.

ech0hol:
@willem-bd Thanks for the review. The force push (`03f41a92`) addresses both points.

### [P2] Custom-skill enabled/disabled state is global-by-name

Fixed. Custom and LEGACY skill toggle state is now stored per-user in `users/{user_id}/skills/_skill_states.json`, keyed by skill name. `extensions_config.json` is no longer used for non-public categories.

- `UserScopedSkillStorage` owns the per-user state file; `is_skill_enabled()` for custom/legacy reads from it, not the global config.
- `load_skills()` applies the per-user state only to that user's custom/legacy skills. Public skill state remains global.
- `_is_disabled_skill_path()` in `sandbox/tools.py` reads the same per-user state, so sandbox ls/read blocking is also isolated.
- Added `test_user_scoped_skill_storage.py` with a cross-user isolation test: two users each create a custom skill named `report-gen`; user A disables theirs; assert user B's stays enabled in `load_skills()`, prompt section, and `_is_disabled_skill_path()`.

### Maintainer note — `_validate_user_id()` vs `make_safe_user_id()`

Fixed. The per-user storage cache key is now `make_safe_user_id(user_id)` in both `get_or_new_user_skill_storage()` and `reset_user_skill_storage()`, so reset always matches the cached key even for raw IM channel IDs with characters that get sanitized. `_validate_user_id()` still rejects post-sanitize IDs that don't match `[A-Za-z0-9_-]` (which shouldn't happen after `make_safe_user_id()`), as a defense-in-depth guard rather than a user-facing failure.

---

The force push (`9cbf2ce`) addresses Copilot's three comments.

### 1. `reset_user_skill_storage()` cache key mismatch (`storage/__init__.py:174`)

Fixed. Both `get_or_new_user_skill_storage()` and `reset_user_skill_storage()` now use `make_safe_user_id(user_id)` as the cache key. A reset for any user_id will match the cached entry regardless of whether the id contains characters that get normalized.

### 2. `str(matching.category)` returns `"SkillCategory.PUBLIC"` (`sandbox/tools.py:214`)

Fixed. `_is_disabled_skill_path()` now uses `matching.category.value` instead of `str(matching.category)` to extract the category string. This is more explicit and defensive — while `StrEnum.__str__()` already returns the value, `.value` makes the intent unambiguous and avoids relying on `StrEnum`'s implicit behaviour.

### 3. Error message references `skills/custom/` (`user_scoped_skill_storage.py:222`)

Fixed. The error message no longer directs users to `skills/custom/`. It now points to `skill_manage` as the supported way to create/edit custom skills, consistent with the per-user storage layout and the updated `skill-creator/SKILL.md` guidance.

ech0hol:
@WillemJiang Fixed all 9 backend-unit-tests failures. Here's a summary of the root causes and fixes:

1. **`test_existing_skill_rejected`** — After switching `client.install_skill()` to use `UserScopedSkillStorage`, the security scan fires before the "already exists" check, producing an unexpected error message. Fix: added `allow_skill_security_scan` fixture and patched `deerflow.client.get_or_new_user_skill_storage` to return a `LocalSkillStorage` so the duplicate check is reached first.

2. **`test_lead_agent_skills.py` × 5** — `get_skills_prompt_section()` calls `get_or_new_user_skill_storage(app_config=None)`, which falls through to `get_app_config()` and raises `FileNotFoundError` for the missing `config.yaml`. Fix: added a `_mock_skill_storages()` helper that monkeypatches `get_app_config` and the storage factory functions with lightweight mocks.

3. **`test_user_scoped_skill_storage.py::TestFactory` × 3** — Same missing `app_config` issue. Fix: added the `config` fixture as a parameter and passed `app_config=config` to all `get_or_new_user_skill_storage()` calls.

4. **`test_mcp_file_migration.py` race condition** — Two `write_bytes` calls occurring in the same nanosecond with the same file length produce identical `(mtime_ns, size)` signatures, so the change detector misses the modification. Fix: added `time.sleep(0.05)` to ensure mtime advances, and changed the content from `b"new"` to `b"new_content"` (different length guarantees a size change even if timing coincides).

Verified locally in the Docker (Linux) environment: **102 passed** ✅

WillemJiang:
@ech0hol Could you resolve the conflicts with main branch?

ech0hol:
@WillemJiang I've resolved the conflicts with main. Could you please take a look when you get a chance? Thanks!

WillemJiang:
@ech0hol  please take a look at the failed unit tests.

ech0hol:
@WillemJiang It's been resolved. Could you please review it once more?

backend/packages/harness/deerflow/sandbox/local/local_sandbox_provider.py
```diff
@@ -83,38 +83,64 @@ def _setup_path_mappings(self) -> list[PathMapping]:
...
+                        read_only=True,
                     )
                 )
 
+            # Legacy skills: global custom (pre-migration), read-only —
+            # static, shared by all threads. These are visible as
+            # SkillCategory.LEGACY to users who have no per-user custom
+            # skills yet, and mounted at /mnt/skills/legacy/<name>/ so
+            # their supporting files are accessible in the sandbox.
+            legacy_skills_path = skills_path / "custom"
```
willem-bd: [P1] Static `/mnt/skills/legacy` mount serves hidden skills to every user

- Location: `local_sandbox_provider.py:122-130` (`_setup_path_mappings`)
- Problem: The legacy `PathMapping` (`skills/custom/` -> `/mnt/skills/legacy`, read-only) is added to **every** thread's sandbox unconditionally whenever the global `skills/custom/` holds any `SKILL.md` — it has no `user_id` dimension. But `UserScopedSkillStorage._iter_skill_files` only surfaces `SkillCategory.LEGACY` to users who have **no** per-user custom skills. So a user who *does* have their own custom skills (and therefore cannot see LEGACY skills in their listing/prompt) can still `read_file("/mnt/skills/legacy/<skill>/SKILL.md")` or `cat` it in the shell and read content the listing layer told them doesn't exist — including another user's pre-migration custom skills left in the global `skills/custom/`.
- Evidence: The mount condition is purely `legacy_skills_path.exists() and any(...SKILL.md...)`; it is prepended to every sandbox in `acquire()` regardless of `user_id`. The only path-access gate, `_is_disabled_skill_path()` in `sandbox/tools.py`, returns `False` for these — a hidden LEGACY skill has no per-user `_skill_states.json` entry, so `get_skill_enabled_state` defaults `True` ("not disabled") and nothing blocks the read. The listing layer and the sandbox layer disagree about visibility, and the sandbox layer is the more permissive one.
- Suggested fix: Gate the legacy `PathMapping` on the acquiring user — only add `/mnt/skills/legacy` for users who have no per-user custom skills (mirror the `user_custom_exists` check from `_iter_skill_files`), or block reads of non-enumerated LEGACY paths in `_is_disabled_skill_path`/`validate_local_tool_path`.
- Test: User A with per-user custom skills + a global `skills/custom/shared/SKILL.md`; assert `read_file("/mnt/skills/legacy/shared/SKILL.md")` and `cat` via shell are denied/empty for A, while a user with no per-user skills can still read it.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -150,9 +150,95 @@ def _is_skills_path(path: str) -> bool:
...
+
+            ext_config = ExtensionsConfig.from_file()
+            return not ext_config.is_skill_enabled(skill_name, category)
+        else:
+            # CUSTOM / LEGACY: use per-user state
+            effective_uid = user_id or get_effective_user_id()
+            storage = get_or_new_user_skill_storage(effective_uid)
+            return not storage.get_skill_enabled_state(skill_name)
+    except Exception:
+        # If we can't determine the enabled state, allow access (fail-open)
```
willem-bd: [P1] `_is_disabled_skill_path` fails open — disabled skills become readable on any config/state read error

- Location: `sandbox/tools.py:226-228`
- Problem: The disabled-skill access gate wraps every `ExtensionsConfig.from_file()` / `get_or_new_user_skill_storage()` / `get_skill_enabled_state()` call in `try/except Exception: return False`. `ls_tool` and `read_file_tool` only short-circuit on a `True` return, so a `False` lets the read proceed. A transient `JSONDecodeError`/`OSError`/`KeyError` while reading `_skill_states.json` or `extensions_config.json` (mid-write race, permissions, or a malformed entry like `{"enabled": "yes"}`) makes the gate return "not disabled" and the skill's files are served. Disabling a skill only flips a flag — the files stay on disk and readable — so there is no secondary guard.
- Evidence: `except Exception: # ... allow access (fail-open) return False`; the callers fall through to `sandbox.list_dir/read_file` on `False`.
- Suggested fix: For an access-control check, fail closed — on read error, log loudly and deny (return `True`), or at minimum refuse to serve when state cannot be determined. At least narrow the caught exception.
- Test: Corrupt/truncate `_skill_states.json` after disabling a skill; assert `read_file`/`ls` on that skill's path still returns the disabled error rather than contents.

backend/packages/harness/deerflow/skills/storage/user_scoped_skill_storage.py
```diff
@@ -0,0 +1,351 @@
...
+        # Let the parent do full discovery + global enabled-state merge.
+        # The overridden _iter_skill_files() routes custom reads to
+        # _user_custom_root and legacy reads to _global_custom_root.
+        skills = super().load_skills(enabled_only=False)
+
+        # Override enabled state for CUSTOM / LEGACY with per-user state.
+        # PUBLIC skill state remains as-is (from global extensions_config).
+        for skill in skills:
+            if skill.category != SkillCategory.PUBLIC:
+                skill.enabled = self.get_skill_enabled_state(skill.name)
```
willem-bd: [P2] `load_skills` discards global extensions_config disables for CUSTOM/LEGACY skills

- Location: `user_scoped_skill_storage.py:175-177`
- Problem: `load_skills` calls `super().load_skills()` (which applies `extensions_config.is_skill_enabled` to every category, including CUSTOM/LEGACY), then unconditionally overwrites `skill.enabled = self.get_skill_enabled_state(name)` for every `category != PUBLIC`. `get_skill_enabled_state` returns the hardcoded default `True` whenever the user's `_skill_states.json` has no entry for the skill. Net effect: any global disable of a custom/legacy skill (e.g. a pre-upgrade `extensions_config.json` entry, or an operator hard-disabling a shared legacy skill) is silently turned back into `enabled=True` for every user who never individually toggled it. This also contradicts the `is_skill_enabled` docstring ("All skill categories respect extensions_config").
- Evidence: `super().load_skills(enabled_only=False)` applies global state; then `if skill.category != SkillCategory.PUBLIC: skill.enabled = self.get_skill_enabled_state(skill.name)`; `get_skill_enabled_state` returns `True` on a missing entry.
- Suggested fix: Make the per-user state *refine* the global default rather than erase it — fall back to `extensions_config.is_skill_enabled(name, category)` when no per-user entry exists, instead of a hard `True`. If global config is intentionally no longer authoritative for non-public, skip the `super()` extensions_config pass for those categories and fix the docstring.
- Test: Seed `extensions_config.json` with a custom/legacy skill `enabled=false`; for a user with no per-user entry assert `load_skills()` reports it disabled; then have the user explicitly enable it and assert the per-user override wins.

backend/packages/harness/deerflow/skills/storage/user_scoped_skill_storage.py
```diff
@@ -0,0 +1,351 @@
...
+            if isinstance(data, dict):
+                return data
+        except (json.JSONDecodeError, OSError):
+            logger.warning("Failed to read skill states file %s", self._skill_states_file)
+        return {}
+
+    def _write_skill_states(self, states: dict[str, dict[str, bool]]) -> None:
+        """Persist per-user skill enabled states to ``_skill_states.json``."""
+        self._user_skills_root.mkdir(parents=True, exist_ok=True)
+        with open(self._skill_states_file, "w", encoding="utf-8") as f:
```
willem-bd: [P2] `_write_skill_states` is non-atomic — a mid-write crash re-enables every disabled skill

- Location: `user_scoped_skill_storage.py:109-113`
- Problem: `_write_skill_states` opens `_skill_states.json` with `"w"` (truncating) then `json.dump`s, with no temp-file + `os.replace`. On crash/disk-full/SIGTERM mid-write the file is truncated/corrupt; `_read_skill_states` catches `JSONDecodeError` and returns `{}`, and `get_skill_enabled_state` then returns `True` for every skill — silently and permanently re-enabling everything the user disabled (the disabled-set is lost on the next successful overwrite).
- Evidence: `with open(self._skill_states_file, "w", ...) as f: json.dump(states, f, ...)`; `_read_skill_states` returns `{}` on `JSONDecodeError`; default `True`. Notably `write_custom_skill` in the same file already uses `NamedTemporaryFile` + `tmp_path.replace(target)`, so the atomic pattern is a local precedent.
- Suggested fix: Write to `NamedTemporaryFile(dir=self._user_skills_root, delete=False)` then `os.replace`/`Path.replace` onto the final path (mirror `write_custom_skill`).
- Test: Inject a failure during `set_skill_enabled_state` (or truncate the file mid-write) and assert the prior disabled-state survives, or that reads fail closed rather than defaulting all-enabled.

backend/app/gateway/auth_middleware.py
```diff
@@ -88,7 +88,17 @@ async def dispatch(self, request: Request, call_next: Callable) -> Response:
...
         if is_valid_internal_auth_token(request.headers.get(INTERNAL_AUTH_HEADER_NAME)):
-            internal_user = get_internal_user()
+            # Extract the channel owner user ID from the trusted header.
+            # When present, the synthetic internal user carries the actual
+            # owner identity so that get_effective_user_id() and per-user
+            # filesystem paths (custom skills, memory, thread data) resolve
+            # to the IM channel user instead of falling back to "default".
+            from app.gateway.internal_auth import INTERNAL_OWNER_USER_ID_HEADER_NAME
+
+            owner_user_id = request.headers.get(INTERNAL_OWNER_USER_ID_HEADER_NAME)
```
willem-bd, line 98: [P2] `X-DeerFlow-Owner-User-Id` is trusted raw for identity at the trust boundary

- Location: `auth_middleware.py:90-101` (+ `internal_auth.get_internal_user`, `services.py:471`)
- Problem: When a valid internal token is present, the `X-DeerFlow-Owner-User-Id` header is `.strip()`-ed and stamped verbatim into `user.id` (no `_validate_user_id`/`make_safe_user_id`). That raw value reaches `get_effective_user_id()`, the contextvar, thread-ownership checks, `RunManager.create_or_reject(user_id=...)`, and `thread_store.update_owner`. Filesystem path traversal is contained (`make_safe_user_id`/`_validate_user_id` sanitize at every path-building site), so `../../etc` cannot escape — but **identity/thread-ownership impersonation is real** if the per-worker internal token leaks, and `docker/nginx` does **not** strip `X-DeerFlow-Internal-Token`/`X-DeerFlow-Owner-User-Id` from external clients, so the only barrier is token secrecy.
- Evidence: `get_internal_user(owner_user_id=...)` does `SimpleNamespace(id=owner_user_id or DEFAULT_USER_ID)` with no validation; `services.py:471` re-stamps `set_current_user(SimpleNamespace(id=owner_user_id))`; nginx `proxy_set_header` directives set only Host/IP/X-Forwarded-*.
- Suggested fix: Normalize/validate the header at the boundary (`make_safe_user_id` before stamping, or reject values failing `_validate_user_id`), and/or have nginx clear these headers for external-facing locations. This complements the cache-key normalization already fixed in `get_or_new_user_skill_storage`.
- Test: Send a forged `X-DeerFlow-Owner-User-Id` with control chars/path segments and assert it is rejected or normalized before reaching `user.id`/contextvar; assert thread-ownership checks see a sanitized identity.

backend/packages/harness/deerflow/agents/lead_agent/prompt.py
```diff
@@ -128,33 +128,44 @@ def get_cached_enabled_skills() -> list[Skill]:
...
             if cached_config is app_config:
                 return list(cached_skills)
 
-    skills = list(get_or_new_skill_storage(app_config=app_config).load_skills(enabled_only=True))
+    if user_id:
+        skills = list(get_or_new_user_skill_storage(user_id, app_config=app_config).load_skills(enabled_only=True))
+    else:
+        skills = list(get_or_new_skill_storage(app_config=app_config).load_skills(enabled_only=True))
     with _enabled_skills_lock:
         _enabled_skills_by_config_cache[cache_key] = (app_config, skills)
```
willem-bd, line 174: [P2] `_enabled_skills_by_config_cache` is now unbounded (one entry per user)

- Location: `lead_agent/prompt.py:151-160` (key `cache_key = (id(app_config), user_id or "default")`)
- Problem: This PR adds the `user_id` dimension to the cache key, so the dict now grows one entry per distinct `(app_config, user)` with no maxsize/TTL/eviction. Entries are removed only on explicit skill mutation by that same user; users who never mutate skills accumulate forever. Each entry holds a strong reference to `app_config` plus a `list[Skill]`, pinning them against GC. In a long-lived multi-tenant gateway (many distinct IM users) this is a slow memory leak. Pre-PR the key was `id(app_config)` only (bounded by config count).
- Evidence: `_enabled_skills_by_config_cache` is a plain `dict`; insertion `_enabled_skills_by_config_cache[cache_key] = (app_config, skills)`; the only removal paths are the global `.clear()` in `_invalidate_enabled_skills_cache` and the per-user `pop` in `invalidate_user_skill_cache` — no size/time bound.
- Suggested fix: Bound it the same way the sibling per-user storage cache is bounded (`OrderedDict` LRU with `move_to_end`/`popitem(last=False)`, max ~64-256), or key off `make_safe_user_id(user_id)` and cap.
- Test: Simulate N distinct users building the agent once each (no mutations) in a long-running process; assert the cache size stays bounded.

willem-bd (commented):
Thanks @ech0hol. I found a few issues that should be addressed before this is ready. Building on @willem-bd's earlier review (the global-by-name enabled-state bleed and the cache-key normalization, both fixed), these are net-new findings focused on the per-user isolation boundary, the sandbox access gate, and cache invalidation.

Most severe: the static `/mnt/skills/legacy` mount and the `_is_disabled_skill_path` fail-open both allow reads the listing layer hides — worth treating as security/access-control gaps. The rest are correctness/data-integrity (non-atomic state write, global-disable silently lost) and cache/memory hygiene (unbounded cache, partial PUBLIC-toggle invalidation). Details inline.

backend/app/gateway/routers/skills.py
```diff
@@ -328,33 +348,59 @@ async def update_skill(skill_name: str, body: SkillUpdateRequest, request: Reque
...
+                extensions_config.skills[skill_name] = SkillStateConfig(enabled=body.enabled)
+                config_data = {
+                    "mcpServers": {name: server.model_dump() for name, server in extensions_config.mcp_servers.items()},
+                    "skills": {name: {"enabled": skill_config.enabled} for name, skill_config in extensions_config.skills.items()},
+                }
+                with open(config_path, "w", encoding="utf-8") as f:
+                    json.dump(config_data, f, indent=2)
+                reload_extensions_config()
+
+        await refresh_user_skills_system_prompt_cache_async(get_effective_user_id())
```
willem-bd: [P2] Toggling a PUBLIC skill only invalidates the caller's prompt cache

- Location: `routers/skills.py:401` (`update_skill`)
- Problem: PUBLIC enabled-state is written to the global `extensions_config.json` (shared across all users), but `update_skill` calls `refresh_user_skills_system_prompt_cache_async(get_effective_user_id())`, which removes only the caller's `(id(app_config), user_id)` entries from `_enabled_skills_by_config_cache`. Every other user's cached enabled-state stays stale until process restart (the cache has no TTL). Pre-PR the invalidation cleared all users (`_enabled_skills_by_config_cache.clear()`).
- Evidence: The PUBLIC branch writes `extensions_config.json` + `reload_extensions_config()`; then `await refresh_user_skills_system_prompt_cache_async(get_effective_user_id())`; `invalidate_user_skill_cache` filters `key[1] == user_id`.
- Suggested fix: For a PUBLIC (global) toggle, clear all users' entries (or the whole cache) since the underlying state is shared — keep per-user invalidation only for custom/legacy.
- Test: Admin A disables a PUBLIC skill; user B (cached before) next builds the prompt; assert B no longer sees the disabled PUBLIC skill.

backend/packages/harness/deerflow/client.py
```diff
@@ -1024,32 +1027,55 @@ def update_skill(self, name: str, *, enabled: bool) -> dict:
...
+            extensions_config.skills[name] = SkillStateConfig(enabled=enabled)
 
-        self._atomic_write_json(config_path, config_data)
+            config_data = {
+                "mcpServers": {n: s.model_dump() for n, s in extensions_config.mcp_servers.items()},
+                "skills": {n: {"enabled": sc.enabled} for n, sc in extensions_config.skills.items()},
+            }
+
+            self._atomic_write_json(config_path, config_data)
+            reload_extensions_config()
```
willem-bd, line 1167: [P2] `DeerFlowClient.update_skill` does not invalidate the skills prompt cache at all

- Location: `client.py:1027-1075` (`update_skill`)
- Problem: Unlike the router's `update_skill`, this path writes the PUBLIC toggle to global `extensions_config.json` (line 1054 `reload_extensions_config()`) but never calls `refresh_user_skills_system_prompt_cache_async(...)` / `invalidate_user_skill_cache(...)` for any user — not even the caller. So even the requesting caller's `_enabled_skills_by_config_cache` stays stale until process restart.
- Evidence: The method body runs `reload_extensions_config()` for the PUBLIC branch and `storage.set_skill_enabled_state(...)` for custom/legacy, but contains no `refresh_*skills*` / `invalidate*skill*cache` call.
- Suggested fix: Mirror the router — after the write, call the appropriate invalidation (global for PUBLIC, per-user `refresh_user_skills_system_prompt_cache_async(user_id)` for custom/legacy).
- Test: Toggle a skill via `DeerFlowClient.update_skill`; assert the next `get_enabled_skills_for_config` reflects the new state without a process restart.

WillemJiang:
@ech0hol Please resolve the conflicts from the main branch.

ech0hol:
Hi @WillemJiang, thanks again for the thorough second-round review — all 8 points were spot-on. Here's what I changed for each one, grouped by priority.

## P1 — Security: Cross-User Data Leakage

### P1-1: Static `/mnt/skills/legacy` mount exposed hidden skills to all users

**Problem**: `LocalSandboxProvider.build()` registered the legacy mount as a static `PathMapping`, so every thread/sandbox saw `/mnt/skills/legacy` regardless of user identity. This bypassed `UserScopedSkillStorage`'s visibility rule — users who can't see legacy in the listing could still read files via the sandbox layer.

**Fix** (`local_sandbox_provider.py` L301–323):

Moved the legacy mount from `build()`'s static mappings into `_build_thread_path_mappings()`, making it per-thread-conditional: we now only add the legacy mount when `user_has_no_custom_skills`, matching `UserScopedSkillStorage._iter_skill_files`'s visibility rule exactly.

```python
# Per-thread: only mount legacy for users without their own custom skills
user_has_no_custom_skills = (
    not any(p.is_dir() and not p.name.startswith(".")
            for p in user_custom_path.iterdir())
    if user_custom_path.exists() else True
)
if user_has_no_custom_skills and legacy_skills_path.exists() and …:
    mappings.append(PathMapping(…))
```

The listing layer and the sandbox layer now agree on who can see legacy.

Tests: `test_local_sandbox_provider_mounts.py` — conditional legacy mount scenarios.

---

### P1-2: `_is_disabled_skill_path` failed open on error

**Problem**: The `except` block silently swallowed all exceptions, returning `False` (not disabled → accessible). A corrupted `_skill_states.json` or a read-race would silently serve files from a disabled skill.

**Fix** (`sandbox/tools.py` L231–237):

Flipped to **fail-closed**. The handler now catches four explicit exception types (`OSError`, `ValueError`, `KeyError`, `json.JSONDecodeError`), logs a warning, and returns `True` (treat as disabled, deny access).

```python
except (OSError, ValueError, KeyError, json.JSONDecodeError) as exc:
    # Access-control check must fail closed
    logger.warning("Failed to determine enabled state, denying access: %s", exc)
    return True
```

Tests: `test_sandbox_tools_security.py` — corrupt file and missing file scenarios.

---

## P2 — Correctness: Edge Cases + Cache Consistency

### P2-1: `load_skills` discarded global `extensions_config` disabled state

**Problem**: `UserScopedSkillStorage.load_skills()` used only `get_skill_enabled_state()` (per-user state) for CUSTOM/LEGACY skills, ignoring what `extensions_config.json` had set. If someone disabled a shared custom skill in `extensions_config.json` before upgrading to per-user storage, the upgrade would **silently re-enable** it — the per-user state file simply had no entry for that skill, defaulting to `True`.

**Fix** (`user_scoped_skill_storage.py` L204–212):

AND-merge the two states:

```python
per_user_state = self.get_skill_enabled_state(skill.name)
global_state = extensions_config.is_skill_enabled(skill.name, category)
skill.enabled = per_user_state and global_state
```

A skill is now enabled only when **both** the per-user toggle and the global switch say so. PUBLIC skills continue to be governed solely by `extensions_config` (handled by `super().load_skills()`).

Tests: `test_user_scoped_skill_storage.py` — AND-merge scenarios.

---

### P2-2: `_write_skill_states` was non-atomic

**Problem**: Direct `open()` + `json.dump()` on `_skill_states.json`. A crash/SIGTERM/disk-full mid-write would leave a truncated or empty file. `_read_skill_states()` then catches `json.JSONDecodeError` and returns `{}`, which would **silently re-enable every disabled skill**.

**Fix** (`user_scoped_skill_storage.py` L109–137):

Atomic write via `tempfile.mkstemp()` + `Path.replace()`, matching the pattern already used by `LocalSkillStorage.write_custom_skill` in the same module:

```python
fd, tmp_path_str = tempfile.mkstemp(dir=str(self._user_skills_root), …)
try:
    with os.fdopen(fd, "w", encoding="utf-8") as f:
        json.dump(states, f, indent=2)
    tmp_path.replace(self._skill_states_file)
except Exception:
    tmp_path.unlink(missing_ok=True)  # best-effort cleanup
    raise
```

`Path.replace` is POSIX-atomic on the same filesystem, so the file is never observed in a half-written state.

Tests: `test_user_scoped_skill_storage.py` — atomic write failure-cleanup scenarios.

---

### P2-3: `X-DeerFlow-Owner-User-Id` not normalized inside the trusted boundary

**Problem**: `get_internal_user()` used the caller-supplied `owner_user_id` raw, without running it through `make_safe_user_id()`. Internally-trusted callers are still expected to produce plausible values, but defense-in-depth at the boundary is cheap and eliminates an entire class of risks (path-traversal fragments, case-variants landing in different storage buckets, etc.).

**Fix** (`internal_auth.py` L63–67):

```python
if owner_user_id:
    effective_id = make_safe_user_id(owner_user_id)
else:
    effective_id = DEFAULT_USER_ID
```

Tests: `test_internal_auth.py` — path-traversal, special characters, and fallback-to-default scenarios.

---

### P2-4: `_enabled_skills_by_config_cache` had unbounded growth

**Problem**: Keyed by `(app_config_hash, user_id)`, this `dict` grew by one entry per unique user, never evicting anything. In a long-running multi-tenant process it leaks memory proportional to the number of distinct identities the process has ever seen.

**Fix** (`prompt.py` L21–34):

- Switched from `dict` to `OrderedDict`
- Capped at 256 entries (matches `_user_scoped_storages` in `deerflow.skills.storage`)
- `move_to_end()` on each access for LRU ordering
- `popitem(last=False)` evicts the least-recently-used entry on overflow

```python
_ENABLED_SKILLS_BY_CONFIG_CACHE_MAXSIZE = 256
_enabled_skills_by_config_cache: OrderedDict = OrderedDict()

with _enabled_skills_lock:
    _enabled_skills_by_config_cache.move_to_end(key)
    # …
    if len(_enabled_skills_by_config_cache) > maxsize:
        _enabled_skills_by_config_cache.popitem(last=False)
```

Tests: `test_lead_agent_prompt.py` — LRU eviction boundary tests.

---

### P2-5: PUBLIC skill toggle only cleared the caller's cache

**Problem**: When a PUBLIC skill's enabled state changed, `update_skill` only called `refresh_user_skills_system_prompt_cache_async()` — which only invalidates the calling user's cache. But PUBLIC skills are shared across all users, so every user's prompt cache was left stale.

**Fix** (`routers/skills.py` L406–413):

Branched by category:

```python
if skill.category == SkillCategory.PUBLIC:
    await asyncio.to_thread(clear_skills_system_prompt_cache)
else:
    await refresh_user_skills_system_prompt_cache_async(get_effective_user_id())
```

PUBLIC toggle → flush the global cache (all users). CUSTOM/LEGACY toggle → flush only that user's cache. `clear_skills_system_prompt_cache` is sync, so we dispatch it via `asyncio.to_thread` to keep the event loop free.

Tests: `test_skills_custom_router.py` — both branches verified.

---

### P2-6: `DeerFlowClient.update_skill` didn't invalidate skill caches

**Problem**: `client.py`'s `update_skill` wrote to `extensions_config.json` but never invalidated the enabled-skills prompt cache. After a programmatic skill update, the agent would serve stale enabled/disabled states until the next process restart.

**Fix** (`client.py` L1193–1207):

Added cache invalidation after the config write, mirroring `routers/skills.py`:

```python
try:
    from deerflow.agents.lead_agent.prompt import (
        clear_skills_system_prompt_cache,
        invalidate_user_skill_cache,
    )
    if skill_category_value == SkillCategory.PUBLIC.value:
        clear_skills_system_prompt_cache()
    else:
        invalidate_user_skill_cache(get_effective_user_id())
except Exception as exc:
    logging.getLogger(__name__).warning(
        "Failed to invalidate skills prompt cache after update_skill: %s", exc
    )
```

The try/except prevents a cache-invalidation failure from masking the actual write success (stale cache is better than a false error).

Tests: `test_client.py` — `test_update_skill_resets_config_key` verifies invalidation.

---

## Status

- **Conflicts with main**: resolved (merged upstream/main as of commit `0b2bf563`).
- **Tests**: all related test suites pass — `test_user_scoped_skill_storage`, `test_internal_auth`, `test_skills_custom_router`, `test_client`, `test_lead_agent_prompt`, `test_sandbox_tools_security`, `test_local_sandbox_provider_mounts`, `test_sandbox_windows_path_normalization`. Total: 228 tests, 0 failures.
- **Lint / format**: `ruff check` and `ruff format --check` both clean.

All changes are in commit `b26a19ae` on top of the merge. Happy to address any follow-up questions!

WillemJiang:
@ech0hol  please solve the conflicts with the main branch.

ech0hol:
@WillemJiang It's been resolved. Could you please review it once more?

ech0hol:
@WillemJiang Hi, I've fixed the failing `test_skill_container_path_defaults` unit test , could you please take a look when you have a moment? Thanks!

WillemJiang (approved):

PR #3898, opened 2026-07-01
feat: add scheduled tasks MVP

fancyboi999:
## Summary

Adds a first-class scheduled-tasks MVP to DeerFlow.

Scope in this PR:
- durable scheduled task persistence + task-run history
- scheduler service that reuses the existing DeerFlow run lifecycle
- workspace management page at `/workspace/scheduled-tasks`
- manual actions: create, pause, resume, trigger, delete
- schedule types: `once` and `cron`
- execution context modes:
  - `fresh_thread_per_run` (default)
  - `reuse_thread`

### Backend
- added scheduled task/task-run persistence models, repositories, and migrations
- added scheduler config section
- added `/api/scheduled-tasks` routes and thread-scoped listing route
- added scheduler service in `backend/app/scheduler/`
- reused the existing Gateway run lifecycle for scheduled executions
- fixed task-run status semantics so history is not marked success before the real DeerFlow run reaches terminal state
- fixed manual trigger semantics so paused cron tasks stay paused after a manual trigger
- added `context_mode` support so scheduled runs can either reuse a thread or create a fresh thread per run

### Frontend
- added `/workspace/scheduled-tasks` workspace page
- added sidebar entry and thread-level entry link
- added create form, filters, detail pane, edit flow, run history rendering, and task actions
- added support for `fresh_thread_per_run` vs `reuse_thread` in the UI

### Docs
- updated `README.md`, `AGENTS.md`, `backend/AGENTS.md`, and `backend/docs/CONFIGURATION.md`
- added spec + implementation plan under `docs/superpowers/`

## Evidence policy

- No screenshots are committed into the repo or included in the code diff.
- PR evidence is kept as text: command outputs, real HTTP responses, and real-browser verification notes.
- If maintainers want visual proof later, screenshots should be attached in a separate PR comment, not committed into git.

### Backend automated
```bash
cd backend && PYTHONPATH=. uv run pytest   tests/test_scheduled_task_models.py   tests/test_scheduled_task_repository.py   tests/test_scheduled_task_schedules.py   tests/test_scheduled_task_claims.py   tests/test_scheduled_task_service.py   tests/test_scheduled_task_router.py   tests/test_scheduled_task_lifecycle.py   tests/test_scheduled_task_router_behavior.py   tests/test_persistence_bootstrap.py::test_empty_branch_creates_all_and_stamps_head -q
```
Result: `28 passed`

### Frontend static + unit
```bash
cd frontend && ./node_modules/.bin/eslint   src/app/workspace/scheduled-tasks/page.tsx   src/core/scheduled-tasks/api.ts   src/core/scheduled-tasks/hooks.ts   src/core/scheduled-tasks/types.ts   src/components/workspace/thread-scheduled-tasks-link.tsx   tests/e2e/scheduled-tasks.spec.ts   tests/e2e/utils/mock-api.ts   tests/unit/core/scheduled-tasks/hooks.test.ts   playwright.config.ts && ./node_modules/.bin/tsc --noEmit

cd frontend && ./node_modules/.bin/rstest tests/unit/core/scheduled-tasks/hooks.test.ts
```
Result: passed

### Real HTTP verification
Verified against a live local Gateway using the configured real models/API setup:
- login
- create thread
- create scheduled task (`cron`)
- create scheduled task (`once`)
- list tasks
- pause / resume
- manual trigger
- task run history
- delete
- automatic scheduler trigger for a real `once` task
- `fresh_thread_per_run` verified to create a new execution thread and persist `last_thread_id`

### Real UI verification
Verified in the in-app browser against the real local app:
- open `/workspace/scheduled-tasks`
- create task
- pause task
- trigger task
- delete task
- edit task
- see `once`/`cron` controls
- see `fresh_thread_per_run` / `reuse_thread` controls

## Residual gap

- The Playwright suite is now much closer to green, but still has selector-stability debt in a couple of assertions. This is test-harness polish, not a known functional blocker in the verified real HTTP / real UI flows.


fancyboi999:
## Scheduled-task form: preset-driven UX

Pushed in `1ca27a73`. Replaces the raw cron-expression input with a preset-driven form (addresses the "raw cron is too abstract" feedback).

### After — preset Select + structured inputs + live preview

**Daily** (default) — browser timezone auto-detected, no more hard-coded UTC:
daily

**Weekly** — weekday toggles, multi-select + ordered, at least one kept:
weekly

**Custom cron** — raw input + crontab.guru link:
custom

**One-time** — `datetime-local` (replaces the raw ISO string):
once

### What changed
- New `cron.ts` — `serializeCron` / `parseCron` / `describeSchedule` plus DST-safe zoned↔UTC helpers. Zero new deps (pure `Intl`).
- New `ScheduledTaskScheduleInput` component — reused by both the create and edit forms.
- `once` → `<input type="datetime-local">`; timezone selector defaults to the browser timezone.
- Live preview row (`每天 09:00 (Asia/Shanghai)` / `Every Mon, Wed at 09:00`).
- Backend contract **unchanged** — still `{cron}` / `{run_at}`.

### Verification
- **Unit**: 27 cron tests — 5-preset serialize, parse round-trip, describe en/zh, DST season flip (EST/EDT), half-hour offset (Asia/Kolkata), weekly-empty fallback. All green.
- **E2E**: 67 Playwright tests passed (2.2m), including the scheduled-tasks suite with the new `One-time` / `getByLabel("Run at")` selectors.
- **Real browser** (chrome-devtools, in-app): all 5 presets + once + timezone + preview exercised; i18n `zh-CN` confirmed.
- `pnpm check`: 0 errors.

### Code-review fixes applied
An independent verifier + `/code-review` (3 finder angles) ran before push. Fixed:
- edit-form refetch wiping in-progress edits (effect deps now keyed on task id, not object reference)
- weekly-with-no-weekday producing a preview that disagreed with the serialized cron
- deduped `pad2`, memoized the timezone list at module level, lazy `parseCron` initializer

### Known gap (out of scope for this commit)
The rest of the page (filters, detail-pane labels, action buttons) still carries pre-existing hard-coded English; only the schedule section was i18n'd here to keep the diff reviewable. Full-page i18n + executor/recipe work for the trending / news / GitHub-issue use cases is tracked as follow-up.

fancyboi999:
## Real end-to-end verification (local backend + gpt-5.4)

Ran the scheduler against a live local Gateway with `model: gpt-5.4` via a
ccg gateway (key in local `config.yaml`, gitignored — **never** enters any
commit / PR / screenshot). All through the public `/api/scheduled-tasks` REST
surface, no test doubles.

### Scheduled run actually fires the agent

Created a task, triggered it manually, watched the run go `running → success`
in ~4.5s. Backend log shows the real model call:

```
lead_agent.agent: Create Agent(default) -> model_name: gpt-5.4-ccg
langchain_openai: POST <gateway>/chat/completions  "HTTP/1.1 200 OK"
worker: Run 96e08a7e-... streaming with modes ['values']
```

### Agent output is persisted to the run's thread

**Trending task** (prompt asked for `web_search`): the agent honestly reported
the tool gap instead of hallucinating — the scheduled-run path surfaces real
tool availability:

> I can do that, but I don't have a `web_search` tool available in this environment.

**Weekly-report task** (no external tool needed): produced a clean markdown
report, stored against the run's thread:

```
# Weekly Status Report

## Accomplished this week
- [Placeholder for accomplishment 1]
- [Placeholder for accomplishment 2]
- [Placeholder for accomplishment 3]

## Currently blocked
- [Placeholder for blocker 1]
- [Placeholder for blocker 2]

## Top 3 priorities for next week
- [Placeholder for priority 1]
- [Placeholder for priority 2]
- [Placeholder for priority 3]
```

### What this proves / doesn't prove

- ✅ create → trigger → dispatch → real model call → terminal status → message
  persistence: the full scheduled-run lifecycle works against a live backend,
  not just mocked tests.
- ⚠️ `web_search` isn't configured in this local `extensions_config` (MCP tooling
  is environment-supplied, not part of this PR), so the trending recipe can't
  fetch the live page here. With `web_search` wired in, the same recipe would
  produce a real digest.
- ❌ Not run in CI (CI can't carry the api key). This is local evidence only.

frontend/src/core/scheduled-tasks/api.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export async function fetchThreadScheduledTasks(
+  threadId: string,
+): Promise<ScheduledTask[]> {
+  const response = await fetch(
+    `${getBackendBaseURL()}/api/threads/${encodeURIComponent(threadId)}/scheduled-tasks`,
+  );
+  return response.json();
+}
```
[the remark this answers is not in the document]
fancyboi999: Agreed — fetchThreadScheduledTasks now checks response.ok and surfaces the error like the other fetchers. Fixed in a017afb1.

backend/app/gateway/routers/scheduled_tasks.py
```diff
@@ -0,0 +1,300 @@
...
+        raise HTTPException(status_code=401, detail="Authentication required")
+    task = await repo.get(task_id, user_id=str(user.id))
+    if task is None:
+        raise HTTPException(status_code=404, detail="Scheduled task not found")
+    result = await service.dispatch_task(task, now=datetime.now(UTC), trigger="manual")
+    if result["outcome"] == "conflict":
+        raise HTTPException(status_code=409, detail=result["error"] or "Scheduled task trigger conflicted with an active run")
+    if result["outcome"] == "failed":
+        raise HTTPException(status_code=409, detail=result["error"] or "Scheduled task trigger failed")
+    return {"id": task_id, "triggered": True}
```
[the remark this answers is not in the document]
fancyboi999: Good catch — outcome="failed" now returns 502; 409 stays for real conflicts. Fixed in a017afb1.

frontend/tests/e2e/utils/mock-api.ts
```diff
@@ -188,6 +229,190 @@ export function mockLangGraphAPI(page: Page, options?: MockAPIOptions) {
...
+  });
+
+  void page.route("**/api/scheduled-tasks", (route) => {
+    if (route.request().method() === "GET") {
+      return route.fulfill({
+        status: 200,
+        contentType: "application/json",
+        body: JSON.stringify(mutableScheduledTasks),
+      });
+    }
```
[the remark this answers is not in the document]
fancyboi999: Agreed — the GET/POST/thread-scoped mocks now normalize objects with context_mode, last_thread_id, and nullable thread_id. Fixed in a017afb1.

frontend/tests/e2e/utils/mock-api.ts
```diff
@@ -188,6 +229,190 @@ export function mockLangGraphAPI(page: Page, options?: MockAPIOptions) {
...
+        parts[parts.indexOf("threads") + 1] ?? "",
+      );
+      return route.fulfill({
+        status: 200,
+        contentType: "application/json",
+        body: JSON.stringify(
+          mutableScheduledTasks.filter((task) => task.thread_id === threadId),
+        ),
+      });
+    }
```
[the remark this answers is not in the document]
fancyboi999: Same normalization applied to the thread-scoped endpoint (shared handler logic). Fixed in a017afb1.

backend/packages/harness/deerflow/persistence/migrations/versions/0003_scheduled_tasks.py
```diff
@@ -0,0 +1,97 @@
+"""scheduled tasks.
+
+Revision ID: 0003_scheduled_tasks
+Revises: 0002_runs_token_usage
+Create Date: 2026-07-01
+"""
```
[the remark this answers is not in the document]
fancyboi999: Renamed 0002_scheduled_tasks.py -> 0003_scheduled_tasks.py to match the revision id. Fixed in a017afb1.

backend/docs/CONFIGURATION.md
```diff
@@ -218,6 +218,27 @@ tool_groups:
   - name: file:read    # Read-only file operations
   - name: file:write   # Write file operations
   - name: bash         # Shell command execution
+
+### Scheduler
```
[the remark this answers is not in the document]
fancyboi999: Fixed — the Tool Groups yaml fence is now closed before the Scheduler heading. Fixed in a017afb1.

backend/docs/CONFIGURATION.md
```diff
@@ -218,6 +218,27 @@ tool_groups:
...
+```
+
+Notes:
+
+- `enabled: false` keeps background polling off by default.
+- The MVP supports thread reuse and fresh-thread-per-run execution modes.
+- The MVP supports only `once` and `cron`.
+- Manual trigger uses the same scheduled-task resource and run lifecycle.
+- Scheduled task definitions and task-run history are persisted in the application database.
 ```
```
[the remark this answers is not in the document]
fancyboi999: Fixed — dropped the stray fence after the Scheduler notes. Fixed in a017afb1.

fancyboi999:
## Architecture review pass — fixes pushed in `f84284e5`

Did a full review pass (backend + frontend + peer-project comparison against QwenPaw / hermes-agent / claude-code / codex automations). The overall architecture is sound — clean harness/app split, correct reuse of the Gateway run lifecycle, no IDOR, migration-compliant. The issues below are edge-case hardening and config-convention alignment, not design rework.

### P0 — real bugs, fixed

1. **Lease dead-end** — `claim_due_tasks` only selected `status == 'enabled'`, so a crash between claim and dispatch left the task permanently stuck in `'running'`. Pause/resume/update all 409'd; only delete worked. Fixed: the claim query now also reclaims `'running'` rows whose lease has expired. Added `SKIP LOCKED` for Postgres multi-worker safety.

2. **Poller crash kills scheduler silently** — `_run_loop` had no try/except around `run_once`, and `task_run_repo.create` ran outside the try block. A transient SQLite "database is locked" would terminate the asyncio task for the rest of the process life. Fixed: wrapped the poll body in try/except with a logged warning.

3. **Manual trigger eats a future `once` task** — a failed manual trigger (409 overlap or 502) flipped a `once` task to `'failed'`, consuming its scheduled run before it ever fired. Fixed: failed manual triggers now preserve the task's prior status.

### P1 — hardening, fixed

4. **Stale-run sweeper** — on scheduler start, orphaned `queued`/`running` task-run rows (left by a process crash) are marked `'failed'` so the run history does not show ghost executions.

5. **Config conventions** — registered `scheduler` in `reload_boundary.STARTUP_ONLY_FIELDS`, bumped `config_version` 17→18, added the `scheduler:` section to `config.example.yaml`, and declared `croniter` as a direct harness dependency (was transitive via `langgraph-runtime-inmem`).

6. **Migration cleanup** — collapsed 0004/0005 into 0003 (this PR is not merged yet, so no need to keep the incremental evolution history). Removed dead schema columns (`misfire_policy`, `max_runs`), the unused `release_lease` method, and the unused `clock.py` module.

7. **`once` timezone** — a naive `run_at` is now interpreted in the task's declared timezone instead of hardcoded UTC, matching how cron schedules work.

8. **Skip semantics** — a skipped `once` task (overlap) is now marked `'failed'` instead of `'completed'`, since the single occurrence never executed.

### Frontend fixes

9. **Server errors were silently swallowed** — all mutations now surface errors via sonner toast (`error.message` carries the backend `detail` extracted by `throwScheduledTaskApiError`). The list query shows an inline error row.

10. Added 15s polling, delete confirmation dialog, Textarea for prompt, locale-aware timestamp formatting, `cn()` for conditional classes, completed/failed filter buttons, plural fix (`1 run` not `1 runs`), form reset after creation.

11. Fixed a render-loop footgun in `ScheduledTaskScheduleInput` (onChange stored in a ref instead of useEffect deps). Replaced the placeholder hooks unit test with real fetch-mock tests. Added PATCH to the e2e mock.

### Deliberately deferred (RFC phase 2+, documented in the spec)

- **Missed-run / misfire policy**: peer projects (QwenPaw, hermes, claude-code) all define this explicitly — recurring skips catch-up, one-shot gets a short grace window. The MVP's implicit behavior (restart runs one missed occurrence, cron collapses to next tick) is deterministic but should be made explicit + tested in a follow-up.
- **Recursive self-scheduling guard**: hermes hard-disables the cron toolset for cron-spawned agents. The MVP only excludes `ask_clarification`; a per-user task-count cap is the cheap first defense.
- **Run-level timeout / same-job overlap**: the existing `RunStatus.timeout` covers wall-clock kill, but there is no scheduler-level timeout or "previous run still active" guard for the case where a 1-min cron meets a 5-min agent run.
- **Result delivery**: fresh-thread-per-run results land in an unobserved thread. IM/GitHub dispatch targets are scoped to RFC phase 2.

### Verification

```
backend: 160 passed (incl. 3 new claim tests + 1 stale-sweeper test)
backend: ruff check + format clean
frontend: tsc --noEmit clean, rstest 36 passed, playwright 6 passed
frontend: eslint + prettier clean
```

willem-bd (commented):
Reviewed the scheduled-tasks MVP. Findings below are focused on the task-run lifecycle, the `non_interactive` exposure, overlap semantics, and a DST bug in the frontend timezone conversion. Most-severe first. Happy to dig into any of these further.

backend/app/scheduler/service.py
```diff
@@ -0,0 +1,270 @@
...
+                },
+            )
+            next_at = next_run_at(
+                task["schedule_type"],
+                task["schedule_spec"],
+                task["timezone"],
+                now=now,
+            )
+            if task["schedule_type"] == "once":
+                task_status = "completed"
```
willem-bd: **A `once` task is marked `completed` at launch and stays that way if the run fails off the happy path.**

`dispatch_task` sets `status="completed"` for `once` tasks immediately after `_launch_run` returns (before the run finishes). `handle_run_completion` is the only thing that corrects it back to `"failed"`, but two gaps leave it stuck at `"completed"`:

- `mark_stale_active_runs` (called at startup, ~line 241) only flips the `scheduled_task_runs` row to `failed` — it never touches the parent `scheduled_tasks` row. So a gateway crash before the completion hook fires leaves the `once` task `completed` while its run row shows `failed`.
- The `on_run_completed` hook swallows exceptions as non-fatal (`worker.py`), so a DB error during correction leaves the task `completed` with `last_error=None`.

Net effect: a user sees a "completed" one-time task whose only run actually failed. Consider deferring the `once` → `completed` transition to `handle_run_completion`, or having `mark_stale_active_runs` also reconcile the parent task's status.

backend/app/gateway/services.py
```diff
@@ -137,6 +142,7 @@ def normalize_input(raw_input: dict[str, Any] | None) -> dict[str, Any]:
         "max_concurrent_subagents",
         "agent_name",
         "is_bootstrap",
+        "non_interactive",
```
willem-bd: **`non_interactive` is now client-controllable via `body.context`.**

Before this PR the key was dropped by `merge_run_context_overrides`; whitelisting it here forwards `body.context={"non_interactive": true}` (a client-controlled field on `RunCreateRequest`) into the lead-agent config, which then strips `ask_clarification` from the toolset (`lead_agent/agent.py`). The scheduler sets it internally via `launch_scheduled_thread_run`, but nothing restricts it to that path — an HTTP/IM client can force autonomous, no-clarification execution on an otherwise-interactive run.

Consider gating `non_interactive` so it can only be set by the internal scheduler path (e.g. only honor it when `request.state.auth_source == "internal"`), rather than accepting it from arbitrary client `body.context`.

backend/app/scheduler/service.py
```diff
@@ -0,0 +1,270 @@
...
+        return "enabled"
+
+    async def dispatch_task(
+        self,
+        task: dict[str, Any],
+        *,
+        now: datetime,
+        trigger: str,
+    ) -> dict[str, Any]:
+        execution_thread_id = task.get("thread_id")
```
willem-bd, line 88: **`overlap_policy="skip"` is silently ignored for `fresh_thread_per_run`.**

Skip relies on `multitask_strategy="reject"` raising `ConflictError` on a same-thread inflight run, but the line below mints a new `thread_id` per run, so it never conflicts and the skip branch (in the `dispatch_task` except) is never taken. A fast cron (e.g. every minute) with long runs (e.g. 5 min) piles up concurrent executions — bounded only by `poll_interval × max_concurrent_runs` per cycle, accumulating across cycles.

Worth either documenting that `skip` is a no-op for fresh-thread mode, or enforcing a per-task concurrency cap for fresh-thread runs.

frontend/src/core/scheduled-tasks/cron.ts
```diff
@@ -0,0 +1,327 @@
...
+ */
+export function zonedLocalToUtcIso(
+  localValue: string,
+  timezone: string,
+): string {
+  const [date, time] = localValue.split("T");
+  const [y, mo, d] = (date ?? "").split("-").map(Number);
+  const [h, mi] = (time ?? "00:00").split(":").map(Number);
+  const refMs = Date.UTC(y ?? 1970, (mo ?? 1) - 1, d ?? 1, h ?? 0, mi ?? 0);
+  const offsetMs = tzOffsetMs(timezone, new Date(refMs));
```
willem-bd: **DST offset bug: `once` tasks fire an hour late around spring-forward, and the create→edit round-trip breaks.**

The offset is computed at the wall-time-as-UTC instant (`tzOffsetMs(timezone, new Date(refMs))`), which is stale for wall times after a spring-forward transition. Example: `2026-03-08 03:30 America/New_York` → `08:30Z` instead of the correct `07:30Z` (fires 1h late), because at `03:30Z` NY is still EST (-5) but the resolved instant is EDT (-4).

`utcToZonedLocalInput` computes the offset at the correct instant, so the round-trip diverges: `03:30 → 08:30Z → 04:30`. This is the actual `once` submission/edit path. Fix: iterate (recompute the offset at `utcMs` and resolve again) or use a proper tz library.

backend/app/scheduler/service.py
```diff
@@ -0,0 +1,270 @@
...
+                task["timezone"],
+                now=now,
+            )
+            if task["schedule_type"] == "once":
+                task_status = "completed"
+            elif trigger == "manual" and task.get("status") == "paused":
+                task_status = "paused"
+            else:
+                task_status = "enabled"
+            await self._task_run_repo.update_status(
```
willem-bd, line 147: **Race: this `update_status("running")` can overwrite a terminal status already written by `handle_run_completion`.**

`_launch_run` returns after scheduling the run task; this `await update_status(... "running")` yields the loop. Under DB contention (SQLite "database is locked" — explicitly anticipated per the `_run_loop` comment), a fast-failing run can reach `handle_run_completion` and commit `"failed"` before this `update_status` resumes, which then overwrites it with `"running"` (and clears `error`), sticking the row until the next restart's stale sweep.

Consider setting the run row to `"running"` *before* launching (or making this write a no-op once the row is already terminal) so launch-path writes can't clobber completion-hook writes.

backend/app/scheduler/service.py
```diff
@@ -0,0 +1,270 @@
...
+        task_run_id = metadata.get("scheduled_task_run_id")
+        user_id = record.user_id
+        if not isinstance(task_id, str) or not isinstance(task_run_id, str) or not user_id:
+            return
+
+        terminal_status: Literal["success", "failed"] | None
+        if record.status.value == "success":
+            terminal_status = "success"
+            error = None
+        elif record.status.value in {"error", "timeout", "interrupted"}:
```
willem-bd: **Interrupted scheduled runs are recorded as `failed` with `error=None`, and a `once` task's single occurrence is consumed by an interrupt.**

`interrupted` is mapped to `failed` here, but `worker.py` sets the interrupted status without an error message, so the run history shows a failed run indistinguishable from a real failure. For a `once` task, `handle_run_completion` also flips the task itself to `failed` (line ~234) and it can't fire again — even if the interrupt was just a user cancel or a concurrent interrupt-strategy run on the same reused thread.

Consider treating `interrupted` as a distinct, non-terminal outcome (or at least not consuming the `once` occurrence on interrupt).

backend/packages/harness/deerflow/persistence/scheduled_task_runs/sql.py
```diff
@@ -0,0 +1,123 @@
...
+            result = await session.execute(stmt)
+            rows = list(result.scalars())
+            for row in rows:
+                row.status = "failed"
+                row.error = error
+                row.finished_at = now
+            await session.commit()
+            return len(rows)
+
+    async def update_by_run_id(
```
willem-bd: **`update_by_run_id` appears to be dead code.**

`handle_run_completion` updates run rows by `run_record_id` via `update_status`; no caller uses a `run_id`-based lookup (grep finds only this definition). If a `run_id` lookup isn't needed, remove it; if it is, wire the completion hook to use it.

frontend/src/core/scheduled-tasks/api.ts
```diff
@@ -0,0 +1,173 @@
+import { fetch } from "@/core/api/fetcher";
+import { getBackendBaseURL } from "@/core/config";
+
+import type { ScheduledTask, ScheduledTaskRun } from "./types";
+
+function scheduledTasksUrl(path: string): string {
+  return `${getBackendBaseURL()}/api/scheduled-tasks${path}`;
+}
+
+async function throwScheduledTaskApiError(
```
willem-bd: **`throwScheduledTaskApiError` duplicates `throwChannelApiError`.**

This is a verbatim copy of the error helper in `frontend/src/core/channels/api.ts` (parse `response.json().detail`, throw with fallback). If the Gateway error envelope ever changes, both must be updated in lockstep. A shared helper in a `core/api` errors module would keep them aligned.

backend/app/scheduler/service.py
fancyboi999: Fixed in 2c4a8f1a. The `once` → `completed` transition now lives only in `handle_run_completion`; at launch the task stays `running` (its lease is cleared and `next_run_at` is consumed, so it cannot be re-claimed). For the crash window you pointed out, startup now runs `cancel_stuck_once_tasks` alongside the run-row sweep: any `once` task orphaned in `running` with no lease is reconciled to `cancelled` with the restart message. Verified against a live gateway: launched a real once run, `kill -9` the gateway mid-run, restarted — task ends `cancelled` / run row `interrupted`, both carrying "interrupted: gateway restarted before the run reached a terminal state".

backend/app/gateway/services.py
fancyboi999: Fixed in 2c4a8f1a. `non_interactive` moved out of the public whitelist into an internal-only set that `merge_run_context_overrides` honors only when `request.state.auth_source == AUTH_SOURCE_INTERNAL` (the scheduler's `launch_scheduled_thread_run` path stamps exactly that). Live check on a running gateway: a client run with `body.context={"non_interactive": true}` still lists `ask_clarification` among its tools, while a scheduler-launched run does not.

backend/app/scheduler/service.py
fancyboi999: Fixed in 2c4a8f1a. `dispatch_task` now pre-checks `has_active_runs(task_id)` (queued/running rows) before launching when `overlap_policy=skip` and the trigger is scheduled, so skip holds for `fresh_thread_per_run` where the same-thread ConflictError can never fire; the ConflictError branch stays for user-owned runs on a reused thread. Live check: an every-minute cron with a real 70s bash run recorded the next cycle as `skipped` ("a previous run of this task is still active") instead of piling up a concurrent execution.

frontend/src/core/scheduled-tasks/cron.ts
fancyboi999: Fixed in 2c4a8f1a. `zonedLocalToUtcIso` re-derives the offset at the resolved instant and corrects once — your exact case now converges: `2026-03-08 03:30 America/New_York` → offset at `03:30Z` is EST(-5) → candidate `08:30Z` → offset there is EDT(-4) → `07:30Z`, and the round-trip with `utcToZonedLocalInput` is stable. Unit tests added for spring-forward, the ambiguous fall-back hour (resolves to the first occurrence), the create→edit round-trip, and a no-DST zone.

backend/app/scheduler/service.py
fancyboi999: Fixed in 2c4a8f1a. The launch-path write now passes `protect_terminal=True`: if the row is already in a terminal status (`success`/`failed`/`skipped`/`interrupted`) it keeps the status and error and only backfills `run_id`/`started_at` when missing, so a fast-failing run's completion write can no longer be clobbered back to `running`. Pinned by a repository test that commits the terminal state first and asserts the late `running` write is a no-op on status/error.

backend/app/scheduler/service.py
fancyboi999: Fixed in 2c4a8f1a, with one deliberate deviation from the suggestion. `interrupted` is now a distinct run status (defaulting to "run was interrupted before completion" when the record carries no error), rendered separately in the run history, and an interrupted `once` task ends `cancelled` rather than `failed`. I kept the occurrence consumed rather than making interrupts non-terminal: the run did launch, so re-arming a past-due `once` task would immediately re-fire after a user cancel (and risks duplicate side effects after a partial execution). `cancelled` + the run-history entry keeps it distinguishable and the user can re-trigger manually. Live check: cancelling an in-flight scheduled run yields run `interrupted` / task `cancelled`.

backend/packages/harness/deerflow/persistence/scheduled_task_runs/sql.py
fancyboi999: Removed in 2c4a8f1a — the completion hook keys off `run_record_id` carried in run metadata, so a `run_id` lookup has no caller.

frontend/src/core/scheduled-tasks/api.ts
fancyboi999: Fixed in 2c4a8f1a. Extracted `throwGatewayApiError` into `core/api/errors.ts`; both the channels and scheduled-tasks API modules now share it, so the Gateway error envelope is parsed in one place.

WillemJiang (approved):

willem-bd (commented):
Medium-effort review of the scheduled-tasks MVP. The design is generally careful (lease/claim mechanism, `protect_terminal` on the run-row write, internal-only `non_interactive` gating), but several of those invariants are incompletely enforced -- most notably the `non_interactive` guard is bypassable via `body.config` (#1), and a fast-failing `once` run can be clobbered back to "running" because the parent-task write lacks the `protect_terminal` guard the sibling run-row write has (#2). Inline comments below are ordered most-severe first.

backend/app/gateway/services.py
```diff
@@ -561,7 +578,8 @@ async def start_run(
         # The ``context`` field is a custom extension for the langgraph-compat layer
         # that carries agent configuration (model_name, thinking_enabled, etc.).
         # Only agent-relevant keys are forwarded; unknown keys (e.g. thread_id) are ignored.
-        merge_run_context_overrides(config, getattr(body, "context", None))
+        is_internal_caller = getattr(getattr(request, "state", None), "auth_source", None) == AUTH_SOURCE_INTERNAL
+        merge_run_context_overrides(config, getattr(body, "context", None), internal=is_internal_caller)
```
willem-bd, line 597: **Security: the `non_interactive` guard is bypassable via `body.config`.** This `merge_run_context_overrides(config, body.context, internal=is_internal_caller)` call -- and the `_INTERNAL_ONLY_CONTEXT_KEYS` guard it enables -- only filters the top-level `body.context`. But `build_run_config` (services.py:~339) copies `body.config["context"]` (and `configurable`) **verbatim** into `config["context"]` with no allowlist, and `RunCreateRequest.config` is a free-form `dict[str, Any]`.

So a session-authenticated (non-internal) client can `POST /api/threads/{id}/runs` with `{"config": {"context": {"non_interactive": true}}}` (and no top-level `context`). `_make_lead_agent` then reads `cfg["non_interactive"]=True` and strips `ask_clarification`, forcing autonomous execution from an untrusted caller -- exactly what the comment on `_INTERNAL_ONLY_CONTEXT_KEYS` (line 183) says must be prevented. Note `merge_run_context_overrides` uses `setdefault`, so even via `body.context` it only *adds* keys; it never strips keys already planted by `build_run_config`.

Fix: apply the `_INTERNAL_ONLY_CONTEXT_KEYS` filter to `body.config.context` / `body.config.configurable` inside `build_run_config` (keyed on `is_internal_caller`), or run the same filtering over the union of `body.context` and `body.config.context`/`configurable`.

backend/app/scheduler/service.py
```diff
@@ -0,0 +1,326 @@
...
+            await self._task_run_repo.update_status(
+                task_run_id,
+                status="running",
+                run_id=result["run_id"],
+                started_at=now,
+                # A fast-failing run can reach handle_run_completion before this
+                # write resumes; never clobber its terminal status.
+                protect_terminal=True,
+            )
+            await self._task_repo.update_after_launch(
```
willem-bd, line 156: **Race: `update_after_launch` can clobber a terminal task status written by `handle_run_completion`.** The sibling run-row write two lines up (137) uses `protect_terminal=True` precisely because "a fast-failing run can reach handle_run_completion before this write resumes." But this parent-task `update_after_launch` has no equivalent guard.

If the background `run_agent` fails fast and its `finally` calls `handle_run_completion` (setting a `once` task to `failed`/`completed`/`cancelled`) before this commits, `update_after_launch` overwrites it back to `running` with `next_run_at=None` and `last_error=None`. The task is then stuck -- `claim_due_tasks` requires `next_run_at IS NOT NULL` -- and only `cancel_stuck_once_tasks` at the next gateway restart recovers it; the parent and child rows also disagree. The same gap exists in the `except`-block `update_after_launch` (line 174) and `update_status` (line 167, no `protect_terminal`).

Fix: mirror `protect_terminal` onto `update_after_launch` (skip the `status`/`last_error` write when the row is already terminal), or re-read the task status immediately before writing.

backend/app/gateway/routers/scheduled_tasks.py
```diff
@@ -0,0 +1,300 @@
...
+    _ensure_task_mutable(existing)
+    updated = await repo.update(task_id, user_id=str(user.id), updates={"status": "enabled"})
+    if updated is None:
+        raise HTTPException(status_code=404, detail="Scheduled task not found")
+    return updated
+
+
+@router.post("/scheduled-tasks/{task_id}/trigger")
+@require_permission("threads", "write")
+async def trigger_scheduled_task(task_id: str, request: Request):
```
willem-bd, line 254: **Manual trigger skips the mutability/overlap check, allowing a duplicate concurrent run on `fresh_thread_per_run` tasks.** Unlike pause/resume/update, `trigger_scheduled_task` doesn't call `_ensure_task_mutable`, and the `has_active_runs` skip in `dispatch_task` is gated to `trigger == "scheduled"`. For `fresh_thread_per_run`, each dispatch mints a new `thread_id`, so `multitask_strategy="reject"` (which is per-thread) never fires.

Triggering a task that's already mid-run therefore launches a second concurrent run -- duplicate side effects, and both runs' `handle_run_completion` race on the parent task row. This contradicts the spec's required "trigger rejected due to active in-flight run" case; the existing `test_manual_overlap_conflict_returns_conflict` only covers `reuse_thread`, where the per-thread `ConflictError` accidentally hides the gap.

Fix: check for an active run (or call `_ensure_task_mutable`) before dispatching a manual trigger, regardless of `context_mode`.

backend/app/gateway/routers/scheduled_tasks.py
```diff
@@ -0,0 +1,300 @@
...
+        except ValueError as exc:
+            raise HTTPException(status_code=422, detail=str(exc)) from exc
+        if existing["schedule_type"] == "once" and next_run_at is None:
+            raise HTTPException(status_code=422, detail="once schedule must be in the future")
+        if existing["schedule_type"] == "once" and next_run_at is not None and (next_run_at - datetime.now(UTC)).total_seconds() < config.scheduler.min_once_delay_seconds:
+            raise HTTPException(
+                status_code=422,
+                detail=(f"once schedule must be at least {config.scheduler.min_once_delay_seconds} seconds in the future"),
+            )
+        updates["schedule_spec"] = schedule_spec
```
willem-bd, line 201: **PATCH on a terminal `once` task is a silent no-op: it recomputes `next_run_at` but never resets `status` to `enabled`.** `_ensure_task_mutable` only blocks `status == "running"`, so `completed`/`failed`/`cancelled` tasks are mutable. This branch sets `updates["next_run_at"]` (and even validates the new `run_at` is in the future, raising 422 otherwise) but never sets `updates["status"]`.

`claim_due_tasks` admits only `enabled` (with lease checks) or `running`-with-expired-lease, so a terminal task with a fresh future `next_run_at` matches neither branch and silently never fires. The endpoint returns 200 with a phantom `next_run_at`; only a separate `/resume` call recovers it, and the caller gets no signal that's needed.

Fix: reset `status` to `"enabled"` when `next_run_at` is advanced on a terminal task, or reject the PATCH on terminal tasks (mirroring the future-time validation).

backend/app/scheduler/service.py
```diff
@@ -0,0 +1,326 @@
...
+        self._lease_owner = f"{socket.gethostname()}:{uuid.uuid4().hex}"
+        self._task: asyncio.Task | None = None
+        self._stop = asyncio.Event()
+
+    async def run_once(self, *, now: datetime) -> None:
+        claimed = await self._task_repo.claim_due_tasks(
+            now=now,
+            lease_owner=self._lease_owner,
+            lease_seconds=self._lease_seconds,
+            limit=self._max_concurrent_runs,
```
willem-bd: **`max_concurrent_runs` is only the per-poll claim `LIMIT`, not a global concurrency cap.** Each claimed task spawns a fire-and-forget `asyncio.create_task(run_agent(...))` (via `_launch_run` -> `start_run`), and the poller runs every `poll_interval_seconds`. With N due tasks and runs longer than the poll interval, active runs accumulate across polls far beyond `max_concurrent_runs` -- there's no semaphore and no active-run-count check anywhere before claiming or launching. The per-task `has_active_runs` overlap check only prevents the *same* task double-firing; it doesn't bound cross-task concurrency.

Example: `max_concurrent_runs=3`, `poll_interval=5s`, 60s runs, 10+ due tasks -> ~12 concurrent runs. Risk: LLM rate-limit / sandbox / DB-connection exhaustion. The field name implies a cap the code doesn't enforce (the docs do accurately call it a per-poll batch limit, so a rename would help too).

Fix: gate claiming (or launching) on a global active-run count / semaphore bounded by `max_concurrent_runs`.

backend/packages/harness/deerflow/persistence/scheduled_task_runs/sql.py
```diff
@@ -0,0 +1,130 @@
...
+            status=status,
+            created_at=datetime.now(UTC),
+        )
+        async with self._sf() as session:
+            session.add(row)
+            await session.commit()
+            await session.refresh(row)
+            return self._row_to_dict(row)
+
+    async def list_by_task(self, task_id: str) -> list[dict[str, Any]]:
```
willem-bd: **`list_by_task` returns the entire run history with no LIMIT/pagination.** There's no `.limit()`, no retention cap, and no cron-frequency floor (`* * * * *` is accepted); the UI (`page.tsx`) renders every returned row. A per-minute cron yields ~525,600 rows/year, all ORM-instantiated and JSON-serialized into one response per `GET /scheduled-tasks/{id}/runs` request, growing forever with no trim.

This deviates from the codebase's normal pagination -- `thread_runs`/`threads` use `limit`/`offset`/cursors (e.g. `limit: int = Query(default=50, le=200)`).

Fix: add a `limit` query param (default ~50-100, capped) and an offset/cursor; consider a retention trim for old run rows.

backend/app/gateway/routers/scheduled_tasks.py
```diff
@@ -0,0 +1,300 @@
...
+
+@router.get("/threads/{thread_id}/scheduled-tasks")
+@require_permission("threads", "read", owner_check=True)
+async def list_thread_scheduled_tasks(thread_id: str, request: Request):
+    repo = get_scheduled_task_repo(request)
+    user = await get_optional_user_from_request(request)
+    if user is None:
+        raise HTTPException(status_code=401, detail="Authentication required")
+    tasks = await repo.list_by_user(str(user.id))
+    return [task for task in tasks if task["thread_id"] == thread_id]
```
willem-bd: **`list_thread_scheduled_tasks` loads every task the user owns, then filters by `thread_id` in Python.** `repo.list_by_user` returns all the user's tasks; this list comprehension discards all but one thread's worth. Both `user_id` and `thread_id` are indexed (migration 0003), so a SQL `WHERE user_id = ? AND thread_id = ?` would be a cheap indexed lookup instead of an O(total tasks per user) scan + materialize.

Fix: add a `repo.list_by_user_and_thread(user_id, thread_id)` method and push the `thread_id` filter into the query.

backend/app/gateway/services.py
```diff
@@ -593,6 +611,57 @@ async def start_run(
...
+    # SimpleNamespace stands in for the Pydantic run-request body that the
+    # HTTP path parses. If start_run gains a new body.* attribute that it reads
+    # directly, add the matching field here so the scheduler path stays in sync.
+    body = SimpleNamespace(
+        assistant_id=assistant_id,
+        input={"messages": [{"role": "user", "content": prompt}]},
+        command=None,
+        metadata=metadata or {},
+        config=None,
+        context={"non_interactive": True},
```
willem-bd: **Scheduled runs never stamp `config['context']['user_id']`, unlike IM-channel runs.** Here `body.context = {"non_interactive": True}` has no `user_id`; `merge_run_context_overrides` only propagates `user_id` when it's already in `body.context`; and `inject_authenticated_user_context` returns early for the internal user (`system_role == INTERNAL_SYSTEM_ROLE`). IM channels set `user_id` in `body.context`; the scheduler doesn't.

`GuardrailMiddleware._build_request` reads `context.get("user_id")` with **no** ContextVar fallback (the one consumer that doesn't use `resolve_runtime_user_id`), so a custom user-scoped `GuardrailProvider` silently fails to match owner-scoped rules for scheduled runs while working for IM and interactive runs. The owner *is* available via the `X-DeerFlow-Owner-User-Id` header -> ContextVar -> `record.user_id`, so every other consumer is fine -- only the guardrail path is exposed. No built-in provider breaks, so this is latent.

Fix: stamp `user_id` into `body.context` here (or have `inject_authenticated_user_context` stamp it for internal callers from the owner header).

backend/app/gateway/services.py
fancyboi999: Fixed in 2859564b. Gating `body.context` alone was indeed incomplete — added `strip_internal_context_keys`, called in `start_run` for every non-internal caller after config assembly, so `non_interactive` is scrubbed from both `config['context']` and `config['configurable']` regardless of whether it arrived via `body.context` or the verbatim `body.config` copy. Live check: a session-authenticated run with `{"config": {"context": {"non_interactive": true}}}` still lists `ask_clarification` in the agent toolset; the scheduler path is unaffected.

backend/app/scheduler/service.py
fancyboi999: Fixed in 2859564b. `update_after_launch` now takes the same `protect_terminal` guard as the run-row write: when the parent task is already `completed`/`failed`/`cancelled` it keeps the hook's status and `last_error`, while still recording launch bookkeeping (`last_run_id`/`last_run_at`/`run_count`) and clearing the lease. Pinned by a repository test that finalizes the task first and asserts the late launch write is a no-op on status/error.

backend/app/gateway/routers/scheduled_tasks.py
fancyboi999: Fixed in 2859564b. The `has_active_runs` pre-check now applies to manual triggers too (still honoring `overlap_policy` — `allow` tasks may overlap deliberately): a manual trigger against an active run returns outcome `conflict` without creating a run row, which the router maps to 409, matching the spec's rejected-trigger case. Live check: second trigger during a real 70s run → `409 {"detail": "task already has an active run"}`. Added a fresh-thread regression test alongside the existing reuse-thread one.

backend/app/gateway/routers/scheduled_tasks.py
fancyboi999: Fixed in 2859564b. When a PATCH recomputes a non-null `next_run_at` and the task is in a terminal status, the update now resets `status` to `enabled` so `claim_due_tasks` can admit it. Live check: completed once task → PATCH with a future `run_at` → status `enabled`, and it actually fired again (`run_count` 1→2). Prompt/title-only PATCHes on terminal tasks intentionally do not re-arm.

backend/app/scheduler/service.py
fancyboi999: Fixed in 2859564b. `run_once` now counts global active (queued/running) run rows via `count_active_runs()` and claims only into the remaining budget (`max_concurrent_runs - active`; zero budget skips the claim entirely), so long runs accumulating across poll cycles are bounded by the cap rather than by `poll_interval × limit`. Updated `config.example.yaml` and `CONFIGURATION.md` to document the field as a global cap.

backend/packages/harness/deerflow/persistence/scheduled_task_runs/sql.py
fancyboi999: Fixed in 2859564b. `GET /scheduled-tasks/{id}/runs` now takes `limit` (default 50, capped at 200) and `offset`, matching the thread-runs convention, and `list_by_task` applies them in SQL. The UI reads the default page (most recent 50). A retention trim for old rows and a cron-frequency floor are real follow-ups I'd rather not squeeze into this PR — happy to file an issue for retention if you want it tracked.

backend/app/gateway/routers/scheduled_tasks.py
fancyboi999: Fixed in 2859564b — added `list_by_user_and_thread(user_id, thread_id)` doing the filter in SQL on the indexed columns; the route no longer materializes the user's full task list.

backend/app/gateway/services.py
fancyboi999: Fixed in 2859564b. `launch_scheduled_thread_run` now stamps `context.user_id = owner_user_id` in `body.context` (same shape IM channels use), so `merge_run_context_overrides` propagates it to `runtime.context` and user-scoped guardrail providers resolve the owning user for scheduled runs. `inject_authenticated_user_context` still skips the internal user, so the owner id is not overwritten.

PR #3902, opened 2026-07-01
feat(observability): add trace-id correlation and enhanced logging

### LINKED ISSUE #3901: [feat]  Enhance production observability with trace-id based log and LLM tracing correlation
### Before you start

## 背景

这是一个生产环境可观测性增强需求。

在生产环境中，可观测性不仅包含 LLM tracing，还应覆盖系统 metrics、traces、logs 等多类信号。当前 DeerFlow 已经支持 Langfuse/LangSmith 等 LLM 侧观测能力，但从一次真实用户请求出发，仍然缺少一个稳定的 request-level trace id，用于串联 Gateway 请求、系统日志、运行时上下文以及 LLM metadata。

本次 issue 主要聚焦两个方向：

1. 增强日志模块的扩展能力，为后续更结构化的日志、更多上下文字段预留空间。
2. 补齐现有 LLM 可观测链路中的缺口，使一次请求可以通过同一个 trace id 关联 HTTP response、日志和 Langfuse metadata。

## 与之前 PR 的关系

之前提交过一个相关 PR：

- https://github.com/bytedance/deer-flow/pull/3827

该 PR 主要将 suggestion、memory 等 LLM 调用关联到 `session_id` 和 `user_id`，用于提升 Langfuse 侧的会话/用户维度可观测性。

本 issue 是它的扩展：不仅关注 LLM 调用本身，也关注生产系统中的 request-level correlation。对应 PR 会是 #3827 的超集，目标是让一次请求可以从 Gateway 到日志，再到 Langfuse metadata 使用统一 trace id 串起来。

## 需求说明

希望引入可配置的 request trace correlation 能力：

- 在 Gateway 请求入口生成或继承 `X-Trace-Id`
- 将 trace id 写回 HTTP response header
- 将 trace id 注入日志上下文，例如 `trace_id`
- 将 trace id 注入 Langfuse metadata，例如 `deerflow_trace_id`
- 在运行时、memory、suggestion、subagent/task 等链路中尽量延续该 trace id
- 保持默认行为不变，只有显式开启增强配置时才改变日志/响应输出

## 日志扩展性

本次先加入 `trace_id` 作为最小可用字段，用于打通跨模块关联。

后续可以继续扩展：

- 关联更多 request/user/session/thread/run 相关字段
- 支持更结构化的日志格式，例如 JSON logs
- 支持更美观、更易读的 text log format
- 为生产环境接入 OpenTelemetry、ELK 等系统预留更清晰的结构

## 行为边界

该功能应保持向后兼容：

- 当 `logging.enhance.enabled=false` 时，行为应尽量与当前现状对齐
- 默认不改变已有 HTTP response header
- 默认不改变已有日志格式
- 默认不影响现有 Langfuse/LangSmith tracing 行为
- 新增 trace id 只作为增强 metadata，不应替代现有 `run_id`、`thread_id`、Langfuse native trace id 等语义

## 预期收益

引入 request-level trace id 后，在生产排障时可以更方便地回答：

- 某个前端/API 请求对应哪些 Gateway 日志？
- 某次请求触发了哪些 LLM 调用？
- suggestion、memory、subagent/task 等二级调用是否来自同一个用户请求？
- 用户反馈的 response header 中的 `X-Trace-Id` 能否快速定位到后端日志和 Langfuse metadata？

这会显著降低生产环境中排查 LLM 调用、异步任务、memory/suggestion 调用和系统日志之间关系的成本。

### Affected area(s)

### Alternatives considered

_No response_

### Additional context

_No response_

AnoobFeng:
- add opt-in gateway request trace correlation via X-Trace-Id
- enhance logging with configurable trace_id-aware formatting
- propagate deerflow_trace_id into runtime context and Langfuse metadata
- keep enhanced logging disabled by default to preserve existing behavior

Closes #3901 

## Why

Production observability should not stop at LLM tracing. In real deployments, troubleshooting usually requires correlating HTTP requests, gateway logs, runtime execution, background LLM calls, and Langfuse metadata.

A previous PR, https://github.com/bytedance/deer-flow/pull/3827, improved Langfuse metadata for suggestion and memory calls by attaching session/user information. This PR extends that direction by adding request-level trace correlation so logs and LLM metadata can be connected through a stable DeerFlow trace id.

## Root Cause

Before this change, DeerFlow had useful LLM observability metadata such as session/user/run/thread information, but it did not have a consistent request-level correlation id that flowed across:

- Gateway HTTP requests
- Gateway/system logs
- runtime context
- Langfuse metadata
- suggestion, memory, task/subagent-related LLM calls

## What changed

- Added configurable enhanced logging support under `logging.enhance`.
- Added request trace context utilities for creating, normalizing, and propagating DeerFlow trace ids.
- Added Gateway trace middleware:
  - inherits inbound `X-Trace-Id` when present
  - generates one when missing
  - writes `X-Trace-Id` to HTTP responses
  - supports streaming/SSE responses by writing the header at response start
- Injected `trace_id` into enhanced log records.
- Injected `deerflow_trace_id` into Langfuse metadata.
- Propagated trace metadata through runtime context and relevant LLM call paths.
- Updated config example and documentation.
- Kept the feature disabled by default to preserve current behavior.

## Compatibility

When `logging.enhance.enabled=false`, the existing behavior remains unchanged:

- no additional response header is emitted by default
- existing log format remains unchanged
- existing tracing behavior remains unchanged

The new `deerflow_trace_id` is a DeerFlow correlation id. It does not replace `run_id`, `thread_id`, subagent trace ids, or Langfuse's native trace id.

## Validation

- Added/updated unit tests for logging configuration, trace middleware, Langfuse metadata propagation, memory/suggestion/task/subagent trace propagation, and worker runtime context.
- Ran targeted tests:

```bash
uv run pytest tests/test_trace_middleware.py tests/test_logging_config.py tests/test_worker_langfuse_metadata.py -q
```

## AI assistance

**Tool(s) used:** Codex

### DIFF
--- a/backend/app/gateway/trace_middleware.py
+++ b/backend/app/gateway/trace_middleware.py
@@ -1,0 +1,61 @@
+"""Gateway request trace middleware."""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Callable
+from typing import Any
+
+from starlette.datastructures import Headers, MutableHeaders
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+from deerflow.trace_context import TRACE_ID_HEADER, request_trace_context
+
+logger = logging.getLogger(__name__)
+
+TraceEnabledGetter = Callable[[], bool]
+
+
+class TraceMiddleware:
+    """Bind a request-level trace id and write it to HTTP response headers."""
+
+    def __init__(self, app: ASGIApp, *, enabled_getter: TraceEnabledGetter):
+        self.app = app
+        self.enabled_getter = enabled_getter
+
+    def _enabled(self) -> bool:
+        try:
+            return bool(self.enabled_getter())
+        except Exception:
+            logger.debug("Trace middleware disabled because config lookup failed", exc_info=True)
+            return False
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http" or not self._enabled():
+            await self.app(scope, receive, send)
+            return
+
+        headers = Headers(scope=scope)
+        incoming_trace_id = headers.get(TRACE_ID_HEADER)
+
+        with request_trace_context(incoming_trace_id) as trace_id:
+
+            async def send_with_trace(message: Message) -> None:
+                if message["type"] == "http.response.start":
+                    response_headers = MutableHeaders(scope=message)
+                    response_headers[TRACE_ID_HEADER] = trace_id
+                await send(message)
+
+            await self.app(scope, receive, send_with_trace)
+
+
+def make_trace_enabled_getter(config_getter: Callable[[], Any]) -> TraceEnabledGetter:
+    """Return a lazy getter for ``logging.enhance.enabled``."""
+
+    def _enabled() -> bool:
+        config = config_getter()
+        logging_config = getattr(config, "logging", None)
+        enhance = getattr(logging_config, "enhance", None)
+        return bool(getattr(enhance, "enabled", False))
+
+    return _enabled
--- a/backend/app/gateway/trace_middleware.py
+++ b/backend/app/gateway/trace_middleware.py
@@ -3,35 +3,37 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import Callable
 from typing import Any
 
 from starlette.datastructures import Headers, MutableHeaders
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
 
+from deerflow.config.app_config import is_trace_correlation_enabled
 from deerflow.trace_context import TRACE_ID_HEADER, request_trace_context
 
 logger = logging.getLogger(__name__)
 
-TraceEnabledGetter = Callable[[], bool]
-
 
 class TraceMiddleware:
-    """Bind a request-level trace id and write it to HTTP response headers."""
+    """Bind a request-level trace id and write it to HTTP response headers.
 
-    def __init__(self, app: ASGIApp, *, enabled_getter: TraceEnabledGetter):
+    The ``enabled`` flag is a **startup snapshot** rather than a per-request
+    live read: ``logging`` is registered as restart-required in
+    ``deerflow.config.reload_boundary.STARTUP_ONLY_FIELDS`` because
+    ``configure_logging()`` only installs the trace-context filter and
+    formatter during app.py lifespan startup. Reading ``logging.enhance.enabled``
+    live here would let a runtime config edit surface the response
+    ``X-Trace-Id`` header and Langfuse ``deerflow_trace_id`` immediately while
+    the log formatter stays on its startup value, contradicting the
+    restart-required contract IDE hover surfaces on ``AppConfig.logging``.
+    """
+
+    def __init__(self, app: ASGIApp, *, enabled: bool):
         self.app = app
-        self.enabled_getter = enabled_getter
-
-    def _enabled(self) -> bool:
-        try:
-            return bool(self.enabled_getter())
-        except Exception:
-            logger.debug("Trace middleware disabled because config lookup failed", exc_info=True)
-            return False
+        self.enabled = bool(enabled)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        if scope["type"] != "http" or not self._enabled():
+        if scope["type"] != "http" or not self.enabled:
             await self.app(scope, receive, send)
             return
 
@@ -49,13 +51,13 @@ class TraceMiddleware:
             await self.app(scope, receive, send_with_trace)
 
 
-def make_trace_enabled_getter(config_getter: Callable[[], Any]) -> TraceEnabledGetter:
-    """Return a lazy getter for ``logging.enhance.enabled``."""
+def resolve_trace_enabled(config: Any) -> bool:
+    """Read ``logging.enhance.enabled`` from an ``AppConfig``-like object.
 
-    def _enabled() -> bool:
-        config = config_getter()
-        logging_config = getattr(config, "logging", None)
-        enhance = getattr(logging_config, "enhance", None)
-        return bool(getattr(enhance, "enabled", False))
-
-    return _enabled
+    Thin backwards-compatible alias around
+    :func:`deerflow.config.app_config.is_trace_correlation_enabled`, kept so
+    existing gateway callers and tests do not have to switch imports. Both
+    the Gateway middleware and the embedded ``DeerFlowClient`` resolve the
+    gate through the same harness helper so their behaviour cannot drift.
+    """
+    return is_trace_correlation_enabled(config)
--- a/backend/packages/harness/deerflow/client.py
+++ b/backend/packages/harness/deerflow/client.py
@@ -44,6 +44,7 @@ from deerflow.models import create_chat_model
 from deerflow.runtime.user_context import get_effective_user_id
 from deerflow.skills.storage import get_or_new_skill_storage
 from deerflow.tools.builtins.tool_search import assemble_deferred_tools
+from deerflow.trace_context import DEERFLOW_TRACE_METADATA_KEY, ensure_trace_context, get_current_trace_id
 from deerflow.tracing import build_tracing_callbacks, inject_langfuse_metadata
 from deerflow.uploads.manager import (
     claim_unique_filename,
@@ -519,6 +520,17 @@ class DeerFlowClient:
         thread_id: str | None = None,
         **kwargs,
     ) -> Generator[StreamEvent, None, None]:
+        """Stream a conversation turn with a DeerFlow request trace context."""
+        with ensure_trace_context():
+            yield from self._stream_without_trace_context(message, thread_id=thread_id, **kwargs)
+
+    def _stream_without_trace_context(
+        self,
+        message: str,
+        *,
+        thread_id: str | None = None,
+        **kwargs,
+    ) -> Generator[StreamEvent, None, None]:
         """Stream a conversation turn, yielding events incrementally.
 
         Each call sends one user message and yields events until the agent
@@ -610,6 +622,7 @@ class DeerFlowClient:
             config["callbacks"] = [*existing_callbacks, *tracing_callbacks]
 
         configurable = config.get("configurable") or {}
+        deerflow_trace_id = get_current_trace_id()
         inject_langfuse_metadata(
             config,
             thread_id=thread_id,
@@ -617,12 +630,15 @@ class DeerFlowClient:
             assistant_id=self._agent_name or "lead-agent",
             model_name=configurable.get("model_name") or self._model_name,
             environment=self._environment or os.environ.get("DEER_FLOW_ENV") or os.environ.get("ENVIRONMENT"),
+            deerflow_trace_id=deerflow_trace_id,
         )
 
         self._ensure_agent(config)
 
         state: dict[str, Any] = {"messages": [HumanMessage(content=message)]}
         context = {"thread_id": thread_id}
+        if deerflow_trace_id:
+            context[DEERFLOW_TRACE_METADATA_KEY] = deerflow_trace_id
         if self._agent_name:
             context["agent_name"] = self._agent_name
 
--- a/backend/packages/harness/deerflow/client.py
+++ b/backend/packages/harness/deerflow/client.py
@@ -37,14 +37,14 @@ from deerflow.agents.lead_agent.agent import build_middlewares
 from deerflow.agents.lead_agent.prompt import apply_prompt_template
 from deerflow.agents.thread_state import ThreadState
 from deerflow.config.agents_config import AGENT_NAME_PATTERN
-from deerflow.config.app_config import get_app_config, reload_app_config
+from deerflow.config.app_config import get_app_config, is_trace_correlation_enabled, reload_app_config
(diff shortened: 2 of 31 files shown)


WillemJiang:
@AnoobFeng, thanks for your contribution. Please fix the lint error.

backend/packages/harness/deerflow/config/app_config.py
```diff
@@ -98,6 +111,7 @@ class AppConfig(BaseModel):
             field_doc="Logging level for deerflow and app modules (debug/info/warning/error); third-party libraries are not affected.",
         ),
     )
+    logging: LoggingConfig = Field(default_factory=LoggingConfig, description="Structured logging and request trace correlation settings.")
```
willem-bd: **`logging.enhance` is restart-required in practice but isn't registered as startup-only, so a runtime toggle yields inconsistent behavior.**

`configure_logging()` runs only in `lifespan` (startup), but `TraceMiddleware` reads `logging.enhance.enabled` live on every request via `make_trace_enabled_getter(get_app_config)`. If an operator flips `enabled: true` in `config.yaml` at runtime: the `X-Trace-Id` response header and Langfuse `deerflow_trace_id` light up immediately (live reads), but the root-handler formatter/filter is never re-installed, so **logs do not gain `trace_id`** until a restart — contradicting the README promise ("logs include `trace_id`"). The reverse toggle is symmetric: after disabling at runtime, logs keep `[trace_id=-]` because the trace formatter is never reverted.

This also trips the repo's own convention. `backend/AGENTS.md` / `reload_boundary.py` state: *"Adding a new restart-required field requires updating the registry; drift is pinned by `tests/test_reload_boundary.py`."* `logging` qualifies (same shape as the already-registered `log_level` — both are consumed once during startup), yet it's neither in `STARTUP_ONLY_FIELDS` nor marked `"startup-only:"` in this `Field(description=...)`.

Suggested fix (pick one): (1) register `logging` in `STARTUP_ONLY_FIELDS` with a reason pointing at `configure_logging()` and add the `startup-only:` marker here via `format_field_description("logging", ...)` so the drift test pins it and operators get a restart hint; or (2) re-run `configure_logging(...)` on config hot-reload so the formatter/filter tracks the live `enhance.enabled` value the middleware already honors.

willem-bd (commented):
Reviewed the trace-correlation propagation end to end. The core wiring is sound: `run_agent` is spawned via `asyncio.create_task` inside the HTTP handler (`services.py`), so the task inherits the middleware's `ContextVar`, and background paths (memory `Timer`, subagent thread pools) carry the id explicitly instead of relying on the contextvar. I also empirically confirmed the inbound `X-Trace-Id` can't crash the response (Starlette decodes header bytes latin-1, so the re-encode round-trips) and that `normalize_trace_id`'s C0/DEL rejection prevents CRLF injection. One consistency/conventions issue below.

backend/app/gateway/trace_middleware.py
```diff
@@ -0,0 +1,61 @@
...
+        self.enabled_getter = enabled_getter
+
+    def _enabled(self) -> bool:
+        try:
+            return bool(self.enabled_getter())
+        except Exception:
+            logger.debug("Trace middleware disabled because config lookup failed", exc_info=True)
+            return False
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
```
willem-bd, line 35: **Blocking file IO on every HTTP request, even when `logging.enhance.enabled=false`.**

`TraceMiddleware.__call__` is `async` and runs inline on the event loop. On every request it calls `self._enabled()` → `enabled_getter()` → `get_app_config()`. `get_app_config()` in `config/app_config.py:511` unconditionally runs `Path.stat()` twice **and** streams the full `config.yaml` through a SHA-256 in 1 MiB chunks for cache-invalidation (`_get_config_mtime` at L470, `_get_config_signature` at L479–495, invoked at L528–530 on every call).

Existing FastAPI routes reach `get_app_config()` through a sync dependency, so anyio parks it on a threadpool. This ASGI middleware bypasses that and stalls the loop for every request — and users who never opt in still pay the cost because `_enabled()` must run to decide whether to bypass. The repo's own `blocking-io-guard` skill and `tests/blocking_io/` exist to prevent exactly this class of regression.

Suggested fix: cache the "enabled" snapshot at startup (and refresh on explicit config reload), or offload the config lookup via `starlette.concurrency.run_in_threadpool`. Either way, guard this with a blocking-IO anchor test.

backend/packages/harness/deerflow/trace_context.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def normalize_trace_id(value: object) -> str | None:
+    """Return a safe trace id string, or ``None`` when *value* is unusable."""
+    if not isinstance(value, str):
+        return None
+    trace_id = value.strip()
+    if not trace_id or len(trace_id) > _MAX_TRACE_ID_LENGTH:
+        return None
+    if any(ord(ch) < 32 or ord(ch) == 127 for ch in trace_id):
```
willem-bd: **`normalize_trace_id` accepts codepoints above 0x7E — a caller-crafted `X-Trace-Id` can force a 500 on any endpoint.**

The check here only rejects `ord(ch) < 32` and `== 127`. Any other codepoint (Chinese, Japanese, emoji, or bytes 0x80–0xFF C1 controls) passes and is bound to the request context. `trace_middleware.py:47` then sets it on the response via `MutableHeaders`, which encodes header values as latin-1: codepoints > 255 raise `UnicodeEncodeError` before `send(message)` is dispatched, and C1 controls (0x80–0x9F) technically pass through but are rejected by hardened intermediaries (nginx/envoy/cloudfront), silently breaking the response.

Repro: `curl -H 'X-Trace-Id: 请求-1' https://.../any/endpoint` → server-side `UnicodeEncodeError` on `_get_str` in `MutableHeaders.__setitem__` → forced 500 (or truncated stream on SSE endpoints).

Suggested fix: reject any codepoint > 0x7E in `normalize_trace_id`, or ASCII-encode with `errors='ignore'` at set time. Add a test covering the crafted-header case.

backend/packages/harness/deerflow/client.py
```diff
@@ -518,6 +519,17 @@ def stream(
         *,
         thread_id: str | None = None,
         **kwargs,
+    ) -> Generator[StreamEvent, None, None]:
+        """Stream a conversation turn with a DeerFlow request trace context."""
```
willem-bd: **`DeerFlowClient.stream()` unconditionally binds a trace context, so embedded/CLI callers get `metadata.deerflow_trace_id` on every Langfuse trace even with `logging.enhance.enabled=false`.**

The PR body says: "When `logging.enhance.enabled=false`, the existing behavior remains unchanged — no additional response header is emitted by default, existing tracing behavior remains unchanged." The gateway path correctly gates this behind the middleware. This entry point does not — `ensure_trace_context()` runs regardless, so a fresh id is generated and `inject_langfuse_metadata(..., deerflow_trace_id=deerflow_trace_id)` at L631 emits the new key on every LLM call.

Result: any embedded/TUI user on the default config sees a schema change on their Langfuse traces (new indexed key, new filter values) that they opted out of.

Suggested fix: read `logging.enhance.enabled` (or a dedicated toggle) inside `stream()` and skip the binding + metadata injection when disabled.

willem-bd: **Sync generator + ContextVar-modifying `with` is a footgun: leaks the id into the caller's context between yields, and risks `Token was created in a different Context` on GC-driven close.**

Python sync generators share the caller's context — they do not create a per-generator context. So `_current_trace_id.set(...)` inside `ensure_trace_context()` mutates the **caller's** context, and the corresponding `reset(token)` only runs when the generator is exhausted or `close()`d.

Two concrete failure modes:
1. **Cross-context reset (unraisable exception).** A caller does `gen = client.stream(...)`, pulls a few events, abandons it, and later `gen` is closed by the GC in a different Context than where `.set(...)` returned the token. `ContextVar.reset(token)` raises `ValueError: <Token> was created in a different Context`, surfaced via CPython's unraisable-exception handler on every GC pass.
2. **Cross-request bleed inside a single Context.** While the generator is paused between yields, any caller code that runs and reads `get_current_trace_id()` (e.g. a follow-up `inject_langfuse_metadata` for unrelated work) picks up this stream's id.

Suggested fix: bind/reset the ContextVar inside the internal `_stream_without_trace_context` startup/teardown (which is entered and exited within a single method call), not around a `yield from`.

backend/packages/harness/deerflow/logging_config.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class TraceContextFilter(logging.Filter):
+    """Inject the current request trace id into every log record."""
+
+    name = _TRACE_FILTER_NAME
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        record.trace_id = get_current_trace_id() or "-"
+        return True
```
willem-bd, line 26: **`TraceContextFilter` reads only the ContextVar, so enhanced-log records from the memory-update `threading.Timer` thread show `trace_id=-` despite the value being carefully plumbed through the queue.**

The memory pipeline threads `deerflow_trace_id` across the Timer boundary via `ConversationContext` (`agents/memory/queue.py`, `updater.py`) precisely because the ContextVar doesn't propagate to a new thread. But here `record.trace_id = get_current_trace_id() or "-"` — inside the Timer thread and inside the `_run_sync_in_thread` executor thread, `get_current_trace_id()` returns None, so every enhanced log line from `MemoryUpdater._do_update_memory_sync` ("Memory updated successfully…", model-error logs, tracing-callback logs) is emitted as `[trace_id=-]`. The Langfuse metadata is correct; only the logs are broken — on precisely the async path the extra plumbing exists to serve.

Suggested fix: at the entry of `_do_update_memory_sync`, wrap the body in `request_trace_context(conversation_context.deerflow_trace_id)` so the ContextVar is set for the duration of the update; or add a per-thread fallback (`threading.local` set by the Timer entry point) that the filter also consults.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -91,8 +92,12 @@ def _install_runtime_context(config: dict, runtime_context: dict[str, Any]) -> N
     if isinstance(existing_context, dict):
         existing_context.setdefault("thread_id", runtime_context["thread_id"])
         existing_context.setdefault("run_id", runtime_context["run_id"])
+        if DEERFLOW_TRACE_METADATA_KEY in runtime_context:
+            existing_context.setdefault(DEERFLOW_TRACE_METADATA_KEY, runtime_context[DEERFLOW_TRACE_METADATA_KEY])
         if "app_config" in runtime_context:
             existing_context["app_config"] = runtime_context["app_config"]
+        if "__run_journal" in runtime_context:
+            existing_context["__run_journal"] = runtime_context["__run_journal"]
```
willem-bd: **Unrelated behavior change bundled into a trace-correlation PR: `__run_journal` is now overwritten on the caller's `config['context']`, breaking harnesses that pre-populated it.**

The adjacent trace/thread/run keys use `setdefault` (preserve caller value); this line uses direct assignment. A test harness or higher-level runner that constructs `config={'context': {'thread_id': 't', '__run_journal': FakeJournal()}}` to intercept audit writes — legitimate today because `_install_runtime_context` did not touch `__run_journal` before this PR — silently loses that override after the merge. `test_worker_langfuse_metadata.py` exercises trace metadata but adds no assertion for this branch, so the regression rides quietly.

This change is orthogonal to X-Trace-Id correlation. Suggested fix: either (a) split it into a separate commit/PR with a dedicated test asserting the intended `__run_journal` merge semantics, or (b) match the adjacent `setdefault` pattern so caller-supplied journals win.

willem-bd (commented):
Review findings from an automated pass. Focus: blocking IO on the request path, response-header safety, and trace propagation for the memory background thread. See inline comments.

AnoobFeng:
Thanks for the wonderful review.

1. Based on issues 1 and 2, I made `logging.enhance.enabled` a restart-required startup snapshot and removed the per-request `get_app_config()` lookup from `TraceMiddleware`, keeping `X-Trace-Id`, enhanced log `trace_id`, and Langfuse `deerflow_trace_id` behavior coherent while eliminating blocking file IO from the HTTP request path.

2. Based on issue 3, I tightened `normalize_trace_id()` to accept only printable ASCII characters, so crafted non-ASCII / C1-control `X-Trace-Id` values are rejected and replaced with a fresh safe id before they can reach response headers, logs, or Langfuse metadata.

3. Based on issue 4, I gated implicit trace-id creation in `DeerFlowClient.stream()` behind `logging.enhance.enabled`, while preserving explicit caller opt-in through `request_trace_context(...)`.

4. Based on issue 5, I changed the sync-generator trace binding in `DeerFlowClient.stream()` to per-step `set/reset`, so the trace id is active during LangGraph execution but does not leak into the caller context between yields or fail on cross-context generator close.

5. Based on issue 6, I rebound the captured `deerflow_trace_id` inside the memory queue/updater worker paths, so enhanced log records emitted from `threading.Timer` and executor threads carry the correct trace id instead of `-`.

6. Based on issue 7, I removed the unrelated `__run_journal` overwrite from `_install_runtime_context()`, preserving caller-supplied `config["context"]` values and keeping the trace-correlation PR scoped to observability behavior.

WillemJiang:
@AnoobFeng  please take a look at the failed tests.

WillemJiang (approved):

WillemJiang (approved):

PR #3903, opened 2026-07-01
fix(gateway): clamp client-supplied recursion_limit to prevent runaway runs

DengY11:
## Summary

The unified Gateway `build_run_config()` copied every top-level key from the
client request body (except `configurable`/`context`) verbatim into the
LangGraph `RunnableConfig`, including `recursion_limit`. The server default of
`100` was therefore fully overridable by the caller with no upper bound.

A caller could send:

```json
{ "config": { "recursion_limit": 100000000 } }
```

and cause a single run to execute an effectively unbounded number of LangGraph
super-steps — each of which triggers at least one LLM call — leading to runaway
API cost and a denial-of-service / cost-exhaustion vector. This is reachable
from all HTTP run-creation paths (`/api/threads/{id}/runs*`, `/api/runs/*`), and
the frontend already sends `recursion_limit: 1000` by default, confirming the
value is taken from the (untrusted) client.

## Fix

`build_run_config()` now validates and clamps a client-supplied
`recursion_limit` server-side instead of trusting it verbatim:

- Valid positive integers are capped at a **configurable** ceiling
  (`AppConfig.max_recursion_limit`, default `1000`).
- Invalid values (non-int, `bool`, `None`, non-positive) fall back to the
  server default (`100`).
- Applied on both the `configurable` and the LangGraph >= 0.6.0 `context`
  passthrough paths, after the generic key passthrough so it always wins.
- A `WARNING` is logged when a value is clamped, for observability.

The default ceiling of `1000` matches the existing frontend / public-skill /
`chat.sh` default, so legitimate deeply-nested subagent runs keep working
unchanged while extreme values are rejected. Operators can tune the ceiling via
`max_recursion_limit` in `config.yaml`.

The IM channel path (`app/channels/manager.py`) is unaffected: its run config
comes from operator-controlled session/channel/user layers (not untrusted
end-user input) and already starts from a fixed `recursion_limit: 100`.

## Config / Docs

- Exposes `max_recursion_limit` in `config.example.yaml` (bumped
  `config_version` to 16).
- Documents the ceiling/fallback in `backend/docs/API.md`.

## Security impact

Fixes a cost-amplification / DoS vector for any caller able to create runs. No
API contract change for legitimate clients (values <= ceiling are unchanged).

### DIFF
--- a/backend/app/gateway/services.py
+++ b/backend/app/gateway/services.py
@@ -204,6 +204,41 @@ def resolve_agent_factory(assistant_id: str | None):
     return make_lead_agent
 
 
+# Lead-agent recursion budget bounds. The Gateway must NOT trust a
+# client-supplied ``recursion_limit`` verbatim: an arbitrarily large value lets
+# a single run execute unbounded LangGraph super-steps (each at least one LLM
+# call), enabling runaway API cost / DoS. ``_DEFAULT_RECURSION_LIMIT`` is the
+# server default when the client sends nothing; the hard ceiling any client
+# value is clamped to is configurable via ``AppConfig.max_recursion_limit``.
+_DEFAULT_RECURSION_LIMIT = 100
+_DEFAULT_MAX_RECURSION_LIMIT = 1000
+
+
+def _resolve_max_recursion_limit() -> int:
+    """Resolve the clamp ceiling from ``AppConfig.max_recursion_limit``.
+
+    Falls back to ``_DEFAULT_MAX_RECURSION_LIMIT`` when the app config cannot be
+    loaded (e.g. no ``config.yaml`` in a bare unit-test environment) so that the
+    clamp still applies rather than crashing the run-config assembly.
+    """
+    try:
+        return get_app_config().max_recursion_limit
+    except Exception:
+        return _DEFAULT_MAX_RECURSION_LIMIT
+
+
+def _clamp_recursion_limit(value: Any, max_limit: int) -> int:
+    """Clamp a client-supplied ``recursion_limit`` into a safe server range.
+
+    Non-integer values (including ``bool``, an ``int`` subclass) and non-positive
+    values fall back to ``_DEFAULT_RECURSION_LIMIT``; valid positive integers are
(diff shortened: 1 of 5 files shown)


WillemJiang:
@DengY11, thanks for your contribution. Please fix the unit test error.

backend/app/gateway/services.py
```diff
@@ -233,7 +254,7 @@ def build_run_config(
     # subagent inside ONE lead tools-node step, and subagents enforce their own
     # limit via `subagents.max_turns`. Do not conflate this 100 with the
     # general-purpose subagent's max_turns.
-    config: dict[str, Any] = {"recursion_limit": 100}
+    config: dict[str, Any] = {"recursion_limit": _DEFAULT_RECURSION_LIMIT}
```
willem-bd, line 271: Minor observation: the server default (`100`) isn't itself subject to the configured ceiling.

`config` is initialized here with `recursion_limit: _DEFAULT_RECURSION_LIMIT` (100), and the clamp below only runs when the client explicitly sends `recursion_limit` (`if "recursion_limit" in request_config`). So if an operator tightens `max_recursion_limit` below 100 (e.g. `50`), a run whose client omits the field still runs at `100`, silently exceeding the operator's ceiling — and it inverts the clamp (sending nothing → `100`, sending `200` → `50`).

The field description says "ceiling for a client-supplied run recursion_limit", so this may be intentional (100 is a fixed safe default). If so, consider either:
- adding `ge=100` (or `ge=_DEFAULT_RECURSION_LIMIT`) to `AppConfig.max_recursion_limit` so it can't be configured below the default, or
- clarifying the description to state the default is not subject to the ceiling.

If the intent is a true hard ceiling, the default should be clamped as well.

willem-bd (commented):
Reviewing the recursion_limit clamp. One minor design observation below; the core clamp logic, type handling (bool excluded before int), passthrough ordering, single-entry-point coverage, and test matrix all look correct.

DengY11:
Hi @WillemJiang , tests fixed

WillemJiang (approved):

WillemJiang (approved):

Issue #3146, opened 2026-05-22
Follow-up: use structured subtask status instead of parsing backend text


ShenAC-SAC:
Split from #3138.

Related:

- #3107
- #3131
- #3138

#3131 fixed the observed stale subtask-running UI by recognizing more terminal result strings.

The remaining maintainability concern is that the frontend still derives subtask status from display text, for example:

- `Task Succeeded. Result:`
- `Task failed.`
- `Task timed out`
- `Task cancelled by user.`
- `Task polling timed out`
- leading `Error:`

This works for the current backend messages, but it is fragile. A future backend wording change can break frontend reconstruction again.

Suggested follow-up:

Add a structured status field for subtask tool messages, for example in `ToolMessage.additional_kwargs`, and let the frontend use that field instead of parsing text.

The display text can remain useful for users or debugging, but it should not be the main state contract between backend and frontend.

This is a maintainability follow-up, not an immediate release blocker.

PR #3906, opened 2026-07-02
feat: emit structured runtime metadata (follow-up#3887)

ShenAC-SAC:
## Why

This is a follow-up to #3887.

#3887 moved durable runtime facts out of checkpointed `messages` and into explicit thread-state channels such as `summary_text`, `delegations`, and `skill_context`. That fixed the persistence/projection side of durable context, but the PR body intentionally called out one remaining capture-side coupling:

- delegation results were still derived from `task` result text such as `Task Succeeded. Result:`;
- frontend subtask cards still inferred terminal state by matching those same text prefixes;
- skill references were still recoverable by re-parsing the paired `ToolMessage.content` for `SKILL.md` reads;
- middleware could convert ordinary task result text back into structured fields, which kept the text protocol alive behind a structured-looking wrapper.

That meant the durable state channels existed, but some of their facts still came from mutable transcript text. A wording change in task output, wrapper output, or skill-read content could still make backend durable context or frontend task state drift.

This PR finishes that decoupling for new runtime output: producer boundaries now emit the structured facts, and consumers read those facts directly.

## Design Summary

The contract after this change is:

1. `ToolMessage.content` remains model-visible/display text.
2. Runtime facts are emitted in `ToolMessage.additional_kwargs` at the authoritative producer boundary.
3. Durable-context capture and frontend task rendering consume those structured fields.
4. Consumers no longer parse task result text or re-parse skill-read content to reconstruct state.

This keeps the transcript useful for the model and user, but removes it as the backend/frontend wire protocol for task state and durable runtime capture.

### Structured task-result metadata at the task boundary

`task_tool` now returns a `Command(update={"messages": [...]})` with an explicit `ToolMessage` for each terminal path. The visible content string is preserved, but the durable facts are attached separately:

- `subagent_status`
- `subagent_error`
- `subagent_result_brief`
- `subagent_result_sha256`

The metadata is bounded before it is written to the message. Completed task results carry a bounded brief and a SHA-256 digest of the full result. Failed/cancelled/timed-out paths carry bounded error metadata.

Pre-execution task failures, normal completion, subagent failure, cancellation, execution timeout, and polling timeout all use the same helper so the terminal branches do not drift independently.

### Task exceptions remain producer-owned

`ToolErrorHandlingMiddleware` still handles exceptions raised before a tool can return normally. For the `task` tool, that middleware is the producer of the error wrapper, so it stamps `subagent_status=failed` and `subagent_error` on that wrapper.

It no longer parses ordinary successful task result text and backfills metadata. Normal task output is owned by `task_tool`; exception wrappers are owned by the middleware.

### Delegation ledger consumes only structured task metadata

`delegation_ledger.extract_delegations()` still captures task dispatches from `AIMessage.tool_calls`, but terminal status/result/error facts now come only from structured task metadata.

The old prefix parsing path was removed:

- no `_parse_task_result()`;
- no task-result prefix constants;
- no backend status inference from `ToolMessage.content`.

If a paired task `ToolMessage` lacks structured metadata, the dispatch remains `in_progress` rather than being promoted to a terminal durable fact from display text. Status-only metadata can update the status, but result/error briefs are not back-filled from content.

### Skill-read metadata at the read-tool boundary

Successful configured read tools now stamp compact `skill_context_entry` metadata when the read targets a `SKILL.md` under the configured skills root.

The metadata contains only:

- `name`
- `path`
- frontmatter `description`

`DurableContextMiddleware` / `skill_context.extract_skills()` validates the metadata against the original read tool call path, rebuilds the skill name from the validated path, and stores only the compact reference.

The extractor no longer reparses `ToolMessage.content` when metadata is missing or mismatched. That makes the read-tool boundary the only place where `SKILL.md` content is interpreted for durable skill references.

### Frontend subtask parser reads structured fields only

`parseSubtaskResult()` now treats task result text as display content, not a protocol. It reads:

- `subagent_status` for the card status;
- `subagent_error` for failure details;
- `subagent_result_brief` for completed result previews.

The old prefix parser and cross-language prefix fixture cases were removed. The shared contract file now pins only the structured status enum values.

### Docs and tests

Backend and frontend docs were updated to state that task result text is display/model content, while durable capture and subtask rendering use structured metadata.

Tests were updated around the new boundary:

- task tool tests assert terminal branches return `Command`/`ToolMessage` metadata;
- delegation tests assert text-only terminal-looking content does not create terminal durable state;
- middleware tests assert only task exception wrappers are stamped by middleware;
- skill-context tests assert missing or mismatched metadata is not recovered by parsing content;
- frontend tests assert terminal-looking text without structured metadata stays `in_progress`.

## Problems Solved

This removes the main follow-up weakness left by #3887:

- Durable context no longer depends on task-result wording.
- Frontend task state no longer depends on backend text prefixes.
- Skill reference capture no longer depends on durable extraction re-reading raw tool content.
- Middleware no longer acts as a glue layer that converts ordinary task display text back into state.
- The structured contract is smaller and easier to review: producer writes state, consumers read state.

This also reduces future maintenance risk. Changes to task result copy or wrapper wording should no longer require coordinated backend ledger and frontend parser updates.

## Scope Notes

This PR intentionally does not redesign the durable state channels introduced by #3887. It keeps the same `delegations` and `skill_context` surfaces, and changes where their facts come from.

This PR also intentionally does not preserve the old task-result text parser for new capture/rendering behavior. The tradeoff is deliberate: old text compatibility would keep the duplicate protocol alive and make future changes drift again. New runtime output is expected to carry structured metadata.

The remaining transcript-derived part is task dispatch discovery from `AIMessage.tool_calls`, which is the authoritative record that the model requested the `task` tool. The removed coupling is terminal status/result/error inference from text.

## Testing

Validated on a clean worktree branched from latest `upstream/main` (`1f740829`):

- `cd backend && uv run pytest tests/test_subagent_status_contract.py tests/test_tool_error_handling_subagent_stamp.py tests/test_tool_error_handling_middleware.py tests/test_task_tool_core_logic.py tests/test_delegation_ledger.py tests/test_skill_context.py tests/test_durable_context_middleware.py tests/test_delegation_ledger_live.py tests/test_thread_state_reducers.py tests/test_create_deerflow_agent.py -q`
  - `226 passed, 1 skipped, 1 warning`
- `cd backend && uv run ruff check packages/harness/deerflow/agents/middlewares/delegation_ledger.py packages/harness/deerflow/agents/middlewares/skill_context.py packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py packages/harness/deerflow/subagents/status_contract.py packages/harness/deerflow/tools/builtins/task_tool.py tests/test_delegation_ledger.py tests/test_skill_context.py tests/test_subagent_status_contract.py tests/test_tool_error_handling_subagent_stamp.py tests/test_durable_context_middleware.py tests/test_task_tool_core_logic.py tests/test_tool_error_handling_middleware.py tests/test_create_deerflow_agent.py`
  - `All checks passed!`
- `cd backend && uv run ruff format --check packages/harness/deerflow/agents/middlewares/delegation_ledger.py packages/harness/deerflow/agents/middlewares/skill_context.py packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py packages/harness/deerflow/subagents/status_contract.py packages/harness/deerflow/tools/builtins/task_tool.py tests/test_delegation_ledger.py tests/test_skill_context.py tests/test_subagent_status_contract.py tests/test_tool_error_handling_subagent_stamp.py tests/test_durable_context_middleware.py tests/test_task_tool_core_logic.py tests/test_tool_error_handling_middleware.py tests/test_create_deerflow_agent.py`
  - `13 files already formatted`
- `cd frontend && pnpm test tests/unit/core/tasks/subtask-result.test.ts`
  - `22 passed`
- `cd frontend && pnpm check`
  - `0 errors, 3 existing warnings` in `message-list-item.tsx` and `message-list.tsx`
- `git diff --cached --check && git diff --check`
  - passed


frontend/src/core/tasks/subtask-result.ts
```diff
@@ -39,89 +45,37 @@ const STRUCTURED_STATUS_TO_SUBTASK: Record<string, SubtaskStatus> = {
...
 export function parseSubtaskResult(
-  text: string,
+  _text: string,
   additionalKwargs?: Record<string, unknown> | null,
 ): SubtaskResultUpdate {
-  const fromText = parseFromText(text.trim());
   const structured = readStructuredStatus(additionalKwargs);
   if (!structured) {
-    return fromText;
+    return { status: "in_progress" };
```
willem-bd, line 85: **[P1] Removing the task-result text fallback degrades subtask cards for every previously-checkpointed thread**

- **Problem:** `subagent_result_brief` is introduced by *this* PR, so any thread persisted before it ships has no `subagent_result_brief`/`subagent_result_sha256`. For an existing completed-task thread, `readStructuredResultBrief()` returns `undefined`, so `update.result` is never set and the card's result-detail step renders `null` — `subtask-card.tsx` renders `task.result`, not the raw tool content. Threads created before the #3146 stamping commit carry no `subagent_status` either, so `parseSubtaskResult` now returns `{status:"in_progress"}` and those cards stay stuck on "executing" forever.
- **Evidence:** `subagent_result_brief`/`subagent_result_sha256` are added in this PR (`status_contract.py`); the old `parseFromText` back-filled `result`/`error` from the `Task Succeeded. Result:` / `Task failed. Error:` content and this PR deletes it. The card body is sourced from `task.result`/`task.error`.
- **Suggested fix:** The PR body calls dropping the text parser "deliberate" for *new* output, but this also regresses *existing* threads on deploy. Either (a) ship a one-time backfill that derives `subagent_result_brief`/`subagent_status` from historical content, or (b) keep a degraded text fallback *only* when structured metadata is entirely absent (pre-stamping threads) so cards don't stick on `in_progress`.
- **Test:** render a fixture thread whose `task` ToolMessages carry only `content` (no `additional_kwargs`) and assert the card reaches a terminal state with a non-empty body.

willem-bd (commented):
Thanks @ShenAC-SAC. I found a few issues that should be addressed before this is ready — details inline, ranked by severity. The main one (#1) is a compatibility regression for previously-checkpointed threads; the producer/consumer wiring itself looks solid.

backend/packages/harness/deerflow/agents/middlewares/skill_context.py
```diff
@@ -114,17 +156,17 @@ def extract_skills(
...
         tool_call_id = str(message.tool_call_id) if message.tool_call_id else ""
-        path = skill_paths_by_id.get(tool_call_id)
-        if path is None:
+        expected_path = skill_paths_by_id.get(tool_call_id)
+        if expected_path is None:
             continue
-        content = message.content if isinstance(message.content, str) else str(message.content)
-        if _is_tool_error_text(content):
+        metadata = read_skill_entry_metadata(message.additional_kwargs)
+        if metadata is None or metadata["path"] != expected_path:
```
willem-bd: **[P2] `extract_skills` silently drops all skill references when the stamper is absent or misconfigured**

- **Problem:** `extract_skills` now hard-requires `skill_context_entry` metadata (stamped only by `ToolErrorHandlingMiddleware`) with no fallback, warning, or assertion. If the stamper isn't in the chain, or its config doesn't match the read tool/root, `extract_skills()` returns `[]` and the model silently forgets which skills are active mid-thread.
- **Evidence:** `create_deerflow_agent` (`factory.py:216`) constructs `ToolErrorHandlingMiddleware()` with `app_config=None`, so stamping falls back to hardcoded `/mnt/skills` + `{read_file,read,view,cat}` — an embedder with a different `skills.container_path` or a read tool outside that set gets nothing. The PR's own tests had to insert `ToolErrorHandlingMiddleware()` (`test_durable_context_middleware.py:442,470`) to keep skill capture working, confirming the new hard dependency.
- **Suggested fix:** at minimum emit a warning when a skill-read candidate (by read-tool name + skills-root path) produces no metadata, so the regression is observable; ideally assert the stamper is registered when `DurableContextMiddleware` is.
- **Test:** assemble `middleware=[DurableContextMiddleware()]` alone and read a `SKILL.md`; assert capture is empty **and** a warning is emitted (not silent).

backend/packages/harness/deerflow/subagents/status_contract.py
```diff
@@ -46,41 +45,30 @@
...
+def _bound_metadata_text(text: str, cap: int = SUBAGENT_METADATA_TEXT_MAX_CHARS) -> str:
+    cleaned = text.strip()
+    if len(cleaned) <= cap:
+        return cleaned
+    marker = "\n...\n"
+    if cap <= len(marker):
+        return cleaned[:cap]
+    head = cap * 2 // 3
+    tail = cap - head - len(marker)
+    return f"{cleaned[:head]}{marker}{cleaned[-tail:]}"
```
willem-bd, line 73: **[P2] `_bound_metadata_text` is missing the `tail <= 0` guard the sibling bounders have**

- **Problem:** when a small `cap` makes `tail = cap - head - len(marker)` ≤ 0, `cleaned[-tail:]` becomes a forward slice (`cleaned[-0:] == cleaned[0:]`, the whole string), so the returned text can exceed `cap`.
- **Evidence:** e.g. `_bound_metadata_text('A'*100, cap=15)` → head=10, tail=0 → returns ~115 chars. `delegation_ledger._bound_text` and `durable_context_middleware._bound_text` both guard with `if tail <= 0: return text[:cap]`; this new helper dropped it. Latent today (all callers use the default `cap=2000` → head=1333, tail=662, safe), but the next caller passing a small cap gets corrupt/oversized metadata.
- **Suggested fix:** add the same guard. (This is also the 4th copy of the head/tail bounding idiom — consider sharing one helper.)
- **Test:** unit-test `_bound_metadata_text` with `cap` values 0, 1, and just above `len(marker)`.

backend/packages/harness/deerflow/agents/middlewares/delegation_ledger.py
```diff
@@ -128,21 +113,20 @@ def extract_delegations(messages: list[AnyMessage]) -> list[DelegationEntry]:
...
-        entry.update(
-            {
-                "status": status,
-                "result_brief": _bound_text(result_text),
-                "result_sha256": hashlib.sha256(result_text.encode("utf-8")).hexdigest(),
-                "result_ref": result_ref,
-            }
-        )
+        entry["status"] = structured["status"]
+        result_text = structured.get("result_brief") or structured.get("error")
```
willem-bd: **[P2] Terminal delegations with status but no result_brief/error lose their ledger result detail**

- **Problem:** a terminal delegation carrying only `subagent_status` (e.g. cancelled with blank `result.error`) gets `status` set but `result_brief`/`result_sha256`/`result_ref` left at defaults, weakening the model-facing "prior attempt" guidance.
- **Evidence:** the cancelled branch passes `error=result.error`; `make_subagent_additional_kwargs` omits `subagent_error` when blank, so `additional_kwargs` is just `{subagent_status:"cancelled"}`. Then `result_text = result_brief or error` is falsy and the `if result_text:` block is skipped. The old `_parse_task_result` fallthrough populated `result_brief` from the `Task cancelled by user.` content.
- **Suggested fix:** low-impact (rendering uses `.get`), but consider deriving a brief from `content` for status-only terminal delegations, or document that status-only metadata yields a status-only ledger entry.

frontend/src/core/tasks/subtask-result.ts
```diff
@@ -39,89 +45,37 @@ const STRUCTURED_STATUS_TO_SUBTASK: Record<string, SubtaskStatus> = {
...
+ * carries no structured stamp.
  * LangChain only ever emits a `ToolMessage` once the tool itself has
  * returned (success or wrapped exception), so an unknown shape means
  * "the contract changed underneath us" — surfacing it as still-running
  * prompts the operator to investigate, where eagerly marking it
  * terminal-failed would mask the drift.
  */
 export function parseSubtaskResult(
-  text: string,
+  _text: string,
```
willem-bd: **[P2] `parseSubtaskResult` keeps an unused `_text` param; its caller still pays `extractTextFromMessage` per render**

- **Problem:** after deleting `parseFromText`, `_text` is unused, but `message-list.tsx:505` still calls `parseSubtaskResult(extractTextFromMessage(message), ...)` for every tool message on every render — the text extraction is computed and discarded.
- **Suggested fix:** drop the parameter from `parseSubtaskResult` and remove the `extractTextFromMessage(message)` argument at the call site.
- **Test:** existing `parseSubtaskResult` tests still pass; just update the signature.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -73,25 +60,56 @@ def _build_error_message(self, request: ToolCallRequest, exc: Exception) -> Tool
...
-        # frontend would otherwise have to fall back to prefix-matching
-        # ``Error: Tool 'task' failed ...`` on the wire. The ``subagent_error``
-        # carries the same ``ExcClass: detail`` shape the wrapper string
-        # uses so debugging artifacts stay aligned.
+        # This middleware is the producer for exception wrappers, so task
+        # failures raised before task_tool can build its own Command still
+        # carry the same structured metadata.
         structured_error = f"{exc.__class__.__name__}: {detail}"
-        return _stamp_task_subagent_status(message, tool_name=tool_name, error=structured_error)
+        return _stamp_task_exception_status(message, tool_name=tool_name, error=structured_error)
```
willem-bd, line 83: **[P2] `_skill_read_tool_names()` allocates a new frozenset on every tool-call result**

- **Problem:** `_maybe_stamp` runs on the exit path of every `wrap_tool_call`/`awrap_tool_call`, and its first check rebuilds the frozenset each time (the None-config branch also re-allocates `frozenset({…})` per call). `app_config` is fixed for the middleware's lifetime.
- **Suggested fix:** cache `self._skill_read_tool_names` and `self._skills_root` in `__init__`, as `DurableContextMiddleware` already does (`durable_context_middleware.py:127`).

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -73,25 +60,56 @@ def _build_error_message(self, request: ToolCallRequest, exc: Exception) -> Tool
...
 
-    @staticmethod
-    def _maybe_stamp(result: ToolMessage | Command, request: ToolCallRequest) -> ToolMessage | Command:
-        """Apply the subagent stamp to successful task tool returns.
+    def _skill_read_tool_names(self) -> frozenset[str]:
+        if self._app_config is None:
+            return frozenset({"read_file", "read", "view", "cat"})
```
willem-bd: **[P2] Default skill-read tool-name set is duplicated in three places**

- **Problem:** `frozenset({"read_file","read","view","cat"})` is hardcoded here as the None-config fallback, while the same set lives in `durable_context_middleware.py:28` (`_DEFAULT_SKILL_READ_TOOL_NAMES`) and is the schema default in `summarization_config.py:55`.
- **Risk:** if the default ever changes, the producer (this middleware) and the consumer (`DurableContextMiddleware`) can drift — one stamps `skill_context_entry` on a read the other doesn't treat as a skill read, and loaded skills silently vanish from durable context.
- **Suggested fix:** import `_DEFAULT_SKILL_READ_TOOL_NAMES` (or expose one canonical default from `summarization_config`) in both middlewares.

backend/packages/harness/deerflow/tools/builtins/task_tool.py
```diff
@@ -184,14 +190,36 @@ def _merge_skill_allowlists(parent: list[str] | None, child: list[str] | None) -
     return [skill for skill in child if skill in parent_set]
 
 
+def _task_result_command(
```
willem-bd, line 195: **[P2] `_task_result_command` takes both `content` and `result`/`error`, but `content` is derivable — a new text/structured drift surface**

- **Problem:** at all 7 call sites `content` is a fixed function of `status`+payload (failed→`f"Error: {error}"`, completed→`f"Task Succeeded. Result: {result}"`, etc.; the polling_timed_out branch even passes the same string as both). This re-creates the exact display-text/structured-metadata drift the PR removed from `delegation_ledger`: a future branch can hand-write `content` that disagrees with the structured fields.
- **Suggested fix:** build `content` inside `_task_result_command` from a status→format table, so callers pass only `result`/`error`.
- **Test:** existing per-branch assertions on `message.content` still pin the rendered text.

backend/packages/harness/deerflow/agents/middlewares/skill_context.py
```diff
@@ -114,17 +156,17 @@ def extract_skills(
...
             continue
         entries.append(
             {
-                "name": _skill_name_from_path(path),
-                "path": path,
-                "description": _parse_description(content),
+                "name": _skill_name_from_path(expected_path),
```
willem-bd, line 176: **[P2] The producer's `name` field is dead on the wire**

- **Problem:** `build_skill_entry_metadata_from_read` stamps `name` into `skill_context_entry`, but `extract_skills` rebuilds the name from the validated path (`_skill_name_from_path(expected_path)`) and discards `metadata["name"]`. The stamped `name` (typed required on `SkillEntryMetadata`) is never read.
- **Suggested fix:** either drop `name` from the producer payload, or — since the path is validated against the read call — trust `metadata["name"]` after the path check and drop the rebuild. If the rebuild is kept (anti-tamper, per `test_extract_skills_rebuilds_name_from_validated_read_path`), a one-line comment "name rebuilt, description trusted" would prevent future churn.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -73,25 +60,56 @@ def _build_error_message(self, request: ToolCallRequest, exc: Exception) -> Tool
...
+            return message
+        if getattr(message, "status", "success") == "error":
+            return message
+        content = message.content if isinstance(message.content, str) else None
+        if content is None:
+            return message
+        args = request.tool_call.get("args")
+        if not isinstance(args, dict):
+            return message
+        path = next((args[key] for key in ("path", "file_path", "filepath") if isinstance(args.get(key), str) and args.get(key)), None)
```
willem-bd: **[P2] Inline path-arg lookup duplicates `_tool_call_path` from the same module**

- **Problem:** the `next((args[key] for key in ("path","file_path","filepath") …))` lookup reimplements `_tool_call_path(tool_call)` defined in `skill_context.py:42` — the very module this file already imports `SKILL_CONTEXT_ENTRY_KEY`/`build_skill_entry_metadata_from_read` from.
- **Risk:** the accepted-key set and guards must stay in sync by hand; if a read tool adopts a new path-arg convention, one site is patched and the other silently stops stamping skill metadata.
- **Suggested fix:** call `_tool_call_path(request.tool_call)`.

ShenAC-SAC:
@willem-bd @WillemJiang Thanks for the reviews and the reminders, especially around old-session compatibility, the structured metadata boundary, and the P2 cleanup items where coupling could drift over time.

This round has been addressed and rebased onto the latest `upstream/main`:

- Kept the legacy fallback only as read compatibility for old checkpointer / old-session data. The new path still prefers structured metadata, so existing sessions do not lose subtask status after upgrade.
- `skill_context` no longer falls back to parsing `ToolMessage.content` when the stamper metadata is missing or the metadata path does not match the tool-call path. It now logs a warning and skips capture, so transcript text is not treated as the durable-state protocol.
- Task result display text is centralized in `status_contract.format_subagent_result_message()`. Both normal `task_tool` results and the `ToolErrorHandlingMiddleware` task exception wrapper use the same status/result/error formatter, avoiding two drifting string protocols.
- Terminal delegations that have only a status and no `result_brief/error` now get canonical details for `failed`, `cancelled`, `timed_out`, and `polling_timed_out`, so the ledger keeps a readable terminal result.
- The default `/mnt/skills` container path now lives in the lightweight `deerflow.constants.DEFAULT_SKILLS_CONTAINER_PATH`. Runtime middleware imports it directly instead of going through `skills_config`, and a source-level regression test prevents the exact literal from being reintroduced across production code.
- The default skill-read tool names are unified through the config constant. The middleware caches read-tool names and the skills root at initialization, and path argument extraction now reuses the common `skill_context` helper.
- The `skill_context_entry` producer no longer emits the dead `name` field; consumers derive the name from the validated skill path.
- `_bound_metadata_text()` now handles small caps where the omission marker leaves no room for a tail segment.

Rebase conflict note: `sandbox/tools.py` keeps the upstream secret-context behavior and also uses the shared skills-container constant from this PR.

Verification:

- Local backend full: `5706 passed, 21 skipped`
- Local frontend unit: `440 passed`
- Local `pnpm check`: `0 errors` (the existing 3 warnings remain)
- GitHub CI: backend unit / blocking IO / frontend unit / lint / e2e / replay E2E / CLA all passed

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -26,39 +33,31 @@
...
-    content = message.content if isinstance(message.content, str) else ""
-    status = extract_subagent_status(content)
-    if status is None:
-        # Non-terminal streaming chunks or unrecognised shapes leave the
-        # field unset so the frontend can keep the card on its in-progress
-        # placeholder until a real terminal frame arrives.
-        return message
-    stamp = make_subagent_additional_kwargs(status, error=error)
+    content, metadata_error = format_subagent_result_message("failed", error=error)
+    message.content = content
```
willem-bd: Task-tool exceptions lose the model-facing recovery guidance. `_build_error_message` builds `"Error: Tool 'task' failed with RuntimeError: foo. Continue with available context, or choose an alternative tool."`, but `_stamp_task_exception_status` then overwrites `message.content` with `format_subagent_result_message("failed", ...)` → `"Task failed. Error: RuntimeError: foo"`. Every non-task tool exception still surfaces the "Continue with available context, or choose an alternative tool." recovery hint, but task-tool exceptions — exactly the ones where the model most needs to be told to retry with a different subagent or approach — silently lose it. `test_task_exception_wrapper_uses_subagent_result_formatter` pins the shorter string, so this is by construction. If the intent was only to align the content with the structured `failed` status, the `Continue with available context, or choose an alternative tool.` suffix could be preserved (e.g. append it to the formatted message for the exception-wrapper path).

backend/packages/harness/deerflow/subagents/status_contract.py
```diff
@@ -91,12 +81,72 @@ def make_subagent_additional_kwargs(
...
+    if raw_status not in SUBAGENT_STATUS_VALUES:
+        return None
+    status = raw_status
+    payload: StructuredSubagentResult = {"status": status}
+    raw_result = additional_kwargs.get(SUBAGENT_RESULT_BRIEF_KEY)
+    raw_hash = additional_kwargs.get(SUBAGENT_RESULT_SHA256_KEY)
+    raw_error = additional_kwargs.get(SUBAGENT_ERROR_KEY)
+    if status == "completed" and isinstance(raw_result, str) and raw_result.strip():
+        payload["result_brief"] = _bound_metadata_text(raw_result)
+        if isinstance(raw_hash, str) and len(raw_hash) == 64:
```
willem-bd: `result_sha256` is validated by length only (`len(raw_hash) == 64`), so any 64-character string passes. The producer always emits a real hex digest, but a faulty checkpoint serializer or a third-party tool wrapper that relays the message could surface a non-hex value, and this reader would store it as `result_sha256` in the delegation ledger. One-line tighten, e.g. `bool(re.fullmatch(r"[0-9a-f]{64}", raw_hash))`, given the PR's own round-trip test implies the field should be strict.

willem-bd (commented):
Reviewing the structured-metadata follow-up. The decoupling is clean and the contract slimming looks right. Two findings worth a look below; neither is a blocker.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
ShenAC-SAC: Fixed in ff39ad97. The recovery suffix is now a shared `_RECOVERY_HINT` constant, and `_stamp_task_exception_status` appends it after the formatted task-failure message — so the exception-wrapper path keeps the same model-facing guidance as every other tool, while the structured `subagent_error` stays clean (existing assertion pins that). `test_task_exception_wrapper_uses_subagent_result_formatter` now pins the full content.

backend/packages/harness/deerflow/subagents/status_contract.py
ShenAC-SAC: Fixed in e219d852. `read_subagent_result_metadata` now requires a full lowercase-hex match (`re.fullmatch` on `[0-9a-f]{64}`) instead of length-only, matching the producer's `hexdigest()` shape. Non-hex, uppercase, and 63/65-char values are dropped — covered by `test_read_subagent_result_metadata_rejects_non_hex_sha256`.

WillemJiang (approved):

PR #1143, opened 2026-03-15
feat: add citation/reference support to deep research reports

lailoo:
## Summary

Closes #1141

Adds citation/reference information to deep research reports, making sources traceable and links visually distinguishable.

### Backend
- **Lead agent prompt** (`prompt.py`): Enhanced `<citations>` section with mandatory citation requirements after `web_search`/`web_fetch` tool usage. Added format examples for both inline citations and Sources sections.
- **TitleMiddleware fix** (`agent.py`): Only register TitleMiddleware when `title.enabled=true` in config, preventing async errors in sync execution mode.

### Frontend
- **Markdown link styling** (`markdown-content.tsx`): Added visual styling (color, underline, hover effect) to regular markdown links so they are clearly distinguishable from plain text.

### Skills
- **GitHub Deep Research skill** (`SKILL.md`): Added citation best practices and good/bad examples.
- **Report template** (`report_template.md`): Added citation hints in Executive Summary and Key Analysis sections.

## How to Test

1. Start the application with a web search tool configured (e.g. Tavily)
2. Trigger a deep research report (e.g. "深度调研 DeerFlow 项目")
3. Verify:
   - Inline citations appear as blue badges in the report body
   - Sources section at the end contains clickable styled links
   - Regular links are visually distinct from plain text (colored + underlined)

## Screenshots

image
image

PR #3907, opened 2026-07-02
feat(frontend): add citation sources evidence panel

18062706139fcz:
## Why

**The trigger.** DeerFlow already renders inline `[citation:Title](URL)` links as
small badges inside AI messages and markdown artifacts (added in #1143 / the
`[citation:Title](URL)` work). That's great for spot-checking a single claim, but
it breaks down for the exact case DeerFlow is built for — deep research reports and
long agent answers that cite **many** sources.

**The pain.** When a message references dozens of citations:

- The badges are scattered *inline* throughout the prose. There is no single place
  that answers "what did this answer actually draw on?" — the reader has to scroll
  the whole message and eyeball every badge.
- The **same source cited five times shows up as five separate badges**, with no
  signal that they're the same URL or how heavily a source was leaned on.
- There's **no way to grab a source** for reuse — to cite it elsewhere you have to
  hand-copy the title and URL out of the rendered text.
- Long reports turn into a wall of inline chips, which is visually noisy and makes
  the citations feel like decoration rather than verifiable evidence.

This PR adds a consolidated, **deduplicated evidence panel** so a reader can see the
full source list, how many times each was cited, and copy any of them — without
changing how inline citations already look.

## What changed

From a user's perspective:

- **New "sources" panel** appears below AI messages and below markdown artifact
  previews **whenever the content contains `[citation:…](url)` links**. If there are
  no citations, nothing renders (zero visual cost for non-research answers).
- The panel is a **collapsible `<details>`** summarizing `Used N sources`
  (`使用了 N 个来源`). Expanding it shows a numbered, deduplicated list:
  - Each row shows the **source title**, its **domain**, and a **cite count**
    (`3 cites` / `3 次引用`) so heavily-referenced sources are obvious.
  - Sources are **deduplicated by normalized URL** — the same link cited many times
    becomes one row with `count = N`, not N rows.
  - Each row is a link that opens the source in a new tab, plus a **copy button**
    that copies a reusable `[Title](URL)` markdown reference to the clipboard.
- **Behavior with many citations** (the case this targets): the list is capped to an
  internal scroll area (`max-h-80` + `overflow-y-auto` + `overscroll-contain`), so a
  report citing 50 sources shows a tidy, self-contained, scrollable panel instead of
  an unbounded list that pushes the rest of the conversation off-screen. Dedup + cite
  counts mean the panel length tracks *distinct* sources, not raw citation volume.

Inline citation badges are unchanged — this is purely additive.

## How it works (implementation notes)

- `frontend/src/core/citations/sources.ts` — pure extraction logic:
  - Regex-matches `[citation:Title](http/https URL)` links; **skips image links**
    (`![citation:…]`) and **masks fenced code blocks** so example snippets aren't
    scraped as real sources.
  - Normalizes URLs (only `http`/`https`), dedupes by `url`, records every
    occurrence (index + per-occurrence title) and a total `count`.
  - Falls back to the **domain** when the citation label is generic
    (`Source` / `来源`), so rows never render a meaningless "Source" title.
  - `formatCitationMarkdownReference()` produces the `[Title](URL)` copy payload.
- `citation-sources-panel.tsx` — the collapsible UI + per-source copy button with
  transient "copied" feedback and localized aria/title labels.
- Wired in two places:
  - AI message content (`message-list-item.tsx`) — extracts from the displayed
    (non-human) message content.
  - Markdown artifact preview (`artifact-file-detail.tsx`) — extracts only for
    `markdown` artifacts; the container also gains `overflow-auto` so the panel and
    long reports scroll properly.
- i18n: added a `citations` block (`sourcesSummary`, `citeCount`, `copyReference`,
  `copiedReference`) to `types.ts`, `en-US.ts`, and `zh-CN.ts`.

## Surface area

- [x] **Frontend UI** — new `CitationSourcesPanel`, wired into message + artifact views, plus i18n strings
- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [ ] **Default behavior change** — additive only; panel is hidden when a message has no citations, inline badges are unchanged

## Validation

- [x] `cd frontend && pnpm test` — 43 files, 423 tests passed (incl. 5 new: `core/citations/sources.test.ts` + `components/workspace/citations/citation-sources-panel.test.ts`)
- [x] `cd frontend && pnpm typecheck` — clean
- [x] `cd frontend && pnpm lint` — no new errors/warnings (the 3 warnings shown are pre-existing on `main` in `message-list-item.tsx`/`message-list.tsx`)
- [x] `pnpm prettier --check` on all changed files — clean
- [ ] Manual: open a citation-heavy report and confirm the panel dedupes repeated URLs, shows correct cite counts, scrolls internally, and copy yields a valid `[Title](URL)` reference

image

image


WillemJiang:
@18062706139fcz  Please resolve the conflict with the main branch.

frontend/src/components/workspace/messages/message-list-item.tsx
```diff
@@ -275,6 +278,17 @@ function MessageContent_({
         <MessageImage {...props} threadId={threadId} maxWidth="90%" />
       ),
       a: ({ href, ...props }: AnchorHTMLAttributes<HTMLAnchorElement>) => {
+        if (typeof props.children === "string") {
+          const match = /^citation:(.+)$/.exec(props.children);
+          if (match) {
+            const [, text] = match;
```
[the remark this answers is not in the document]
18062706139fcz: Good catch — fixed in 8cb13156. The fallback branch now mirrors `MarkdownContent`'s default `a` renderer: primary-color underline styling plus `target="_blank"`/`rel="noopener noreferrer"` for external (`http(s)://`) links, while preserving the passed `className`/`target`/`rel` when present. `citation:` and `/mnt/` handling are unchanged.

frontend/src/core/citations/sources.ts
```diff
@@ -0,0 +1,100 @@
...
+  id: string;
+  title: string;
+  url: string;
+  domain: string;
+  count: number;
+  occurrences: CitationOccurrence[];
+};
+
+const CITATION_LINK_RE =
+  /(^|[^!])\[citation:\s*([^\]]+?)\]\((https?:\/\/[^\s)]+(?:\([^\s)]*\)[^\s)]*)?)\)/gi;
```
willem-bd: Two correctness issues with this regex:

**1. Adjacent citations drop every other source.** The `(^|[^!])` prefix *consumes* the char before `[citation:`, so when two citation links are directly adjacent the separating `)` is eaten by the first match and the second can't match. Verified:
```
[citation:A](a)[citation:B](b)[citation:C](c)  →  extracts only a, c   (B vanishes)
```
Dense reports that emit citations with no interleaving prose will undercount sources and show wrong cite counts. A non-consuming lookbehind `(?<!!)` (or `(?<![!\]])`) avoids eating the boundary char.

**2. URLs with two parenthetical groups are truncated.** The URL sub-pattern only tolerates a single balanced `(...)` group. Verified:
```
[citation:W](https://en.wikipedia.org/wiki/Foo_(a)_(b))  →  https://en.wikipedia.org/wiki/Foo_(a
```
The panel row then links to and copies a broken reference that disagrees with the correctly-rendered inline badge. Wikipedia-style disambiguation URLs are exactly what deep-research reports cite.

willem-bd (commented):
Reviewed the citation sources panel. Solid feature with good test coverage for the common cases. My main concern is that `extractCitationSources` re-parses markdown with a hand-rolled regex that disagrees with the actual streamdown parser on a few inputs, so the panel can misrepresent the sources actually cited. Inline comments below.

**Findings (most severe first):**
1. Back-to-back citations with no separator drop every other source (`sources.ts` regex).
2. URLs with two parenthetical groups are truncated to a broken URL (`sources.ts` regex).
3. The link-renderer `a` component is copy-pasted between `message-list-item.tsx` and `markdown-content.tsx` (duplication).
4. Only fenced code is masked — inline-code and unclosed fences leak example citations into the panel.

Findings 1 and 2 are the ones I'd fix before merge — both produce a panel that misrepresents the cited sources. 3 is cleanup; 4 is an edge/streaming concern.

frontend/src/core/citations/sources.ts
```diff
@@ -0,0 +1,100 @@
...
+
+function extractDomain(url: string): string {
+  try {
+    return new URL(url).hostname.replace(/^www\./i, "");
+  } catch {
+    return url;
+  }
+}
+
+function maskFencedCodeBlocks(markdown: string): string {
```
willem-bd, line 106: Only *fenced* code is masked, so inline-code and unclosed fences leak example citations into the panel. Verified: `` `[citation:X](https://x.com)` `` (inline code) and an unclosed ``` fence during streaming both get scraped as real sources. The panel then lists sources that render as literal code (or aren't yet real links), a mismatch between what the reader sees inline and what the panel claims was "used." Lower severity than the regex issues, but worth masking inline spans too.

frontend/src/components/workspace/messages/message-list-item.tsx
```diff
@@ -275,6 +278,17 @@ function MessageContent_({
         <MessageImage {...props} threadId={threadId} maxWidth="90%" />
       ),
       a: ({ href, ...props }: AnchorHTMLAttributes<HTMLAnchorElement>) => {
```
willem-bd: This `a` renderer block (citation-detection regex + external-link styling with `target`/`rel` defaults) is a near-verbatim duplicate of the one already in `markdown-content.tsx`. Two copies of the same link logic will drift — the styling classes, the `https?` test, and the `rel` defaults now have to be kept in sync by hand. A shared `createMarkdownLinkComponent(threadId)` factory returning the `a` renderer would remove the duplication and keep the two render paths consistent.

WillemJiang (approved):

PR #3908, opened 2026-07-02
Fix(frontend): stale run reconnect and cancel handling

### LINKED ISSUE #3905: [bug] 前端残留 lg:stream run_id 导致 409 "is not  cancellable(status: success)"，第一条消息被路由到 stop
### Before you start

### Problem summary

在后端刚跑出结果、前端尚未正常收尾的那一刻刷新页面（这正是前端通常会卡死的时机），sessionStorage 里残留的 lg:stream:<thread> key 会让刷新后的第一次"发送"被路由到stop（取消一个已 success 的 run）而非 submit，抛出 409 Run X is not cancellable  (status: success)，直到该残留 key 被清除才恢复。

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

 后端 run 到达 success 与前端 SSE 流正常收尾（onSuccess → 清除 lg:stream key、
  isLoading 翻回 false）之间存在一个时间窗口。在这个窗口内刷新页面：

  1. SDK 的 onSuccess 没有机会执行，lg:stream:<thread> 把那个 run_id 残留在
     sessionStorage 里。
  2. 后端 run 已经是 success（因 on_disconnect="continue" 继续跑完）。
  3. 刷新后，useStream 在 reconnectOnMount: true 下读到残留 key，调用
     client.runs.joinStream(threadId, runId)。此时 thread.isLoading 仍为 true，
     第一次点"发送"会被 input-box.tsx 路由到 onStop() → thread.stop() →
     client.runs.cancel(threadId, runId)（SDK stream.lgp.js 的 stop()），从而 409：

     HTTP 409: Run <run_id> is not cancellable (status: success)

     随后 SDK 的 stop() 会 removeItem 掉这个 key，于是下一次发送才能正常发出
     ——即刷新后第一条消息被吞掉，只有第二条才真正发送；刷新后短时间内页面表现为卡死。

  后端从 _cancel_conflict_detail（thread_runs.py）返回 409 是有意的：
  RunManager.cancel() 对终态 run 返回 False。缺陷在前端：client.runs.cancel 没有
  对终态 409 做静默 no-op 处理，joinStream 也没有预检跳过已结束的 run。

Image

Image

### Expected behavior

在结果刚出、刷新后，第一次"发送"应当提交一条新消息，而不是被路由到 stop。对一个已经                                   
 结束的 run 发起（或尝试发起）取消应当是静默 no-op，绝不应向用户暴露 409，也不应让页面                                 
 在刷新后短时间内卡死；重连到一个已结束的 run 不应把 isLoading 卡在 true。

### Steps to reproduce

  1. make docker-start（或 make dev），打开 http://localhost:2026                                                       
  2. 打开一个会话，发一条会产生多秒 run 的提问（例如会触发工具调用 / 长回答的问题）。                                   
  3. 开启langfuse辅助判断时机，在后端刚跑出结果、前端尚未正常收尾的那一刻按 F5 刷新页面                                                 
     （这正是前端通常会卡死的时机——后端已有结果，但前端 SSE 还没 drain 完）。                                           
  4. 刷新完成后，立刻点击发送一条新消息。                                                                               
  5. 观察：第一次点击被当作 Stop，浏览器控制台 / 网络面板显示：                                                         
                                                                                                                        
     POST /api/langgraph/threads/<thread_id>/runs/<run_id>/cancel → 409                                                 
     {"detail":"Run <run_id> is not cancellable (status: success)"}                                                     
                                                                                                                        
     且刷新后短时间内页面表现为卡死、无法正常发送。                                                                     
  6. 再发送同一条消息——这次能正常提交（残留 key 已被 SDK stop() 在失败的                                                
     cancel 之后清除）。

### Relevant logs

```shell
无
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Support bundle summary

_No response_

### Additional context

_No response_

lllyfff:
Fixes #3905

  ## Why

  After a page reload, a stuck + 409 bug appears: when a reload lands inside the
  window where the previous run has already reached `success` on the backend but
  the frontend SSE stream has not yet drained normally (`onSuccess` has not run,
  so the `lg:stream:<thread>` key is still set), the stale run_id left in
  sessionStorage routes the first post-reload "send" to `stop` (cancelling an
  already-succeeded run) instead of `submit`, surfacing:

      HTTP 409: Run <run_id> is not cancellable (status: success)

  For a few seconds after reload the page appears stuck and cannot send, until
  the SDK's `stop()` clears the stale key after the failed cancel, after which
  the second message finally goes through. This is the reload timing users hit
  most often — the backend already has a result, the UI looks frozen, and the
  user instinctively reloads, which is exactly what triggers the bug.

  See the linked issue for the repro. Root cause is on the frontend:
  `client.runs.cancel` does not treat the terminal-state 409 as a no-op, and
  `joinStream` has no preflight to skip a run that has already finished
  (reconnecting to a finished run whose in-memory stream bridge has been reaped
  blocks forever, pinning `isLoading` true). The backend 409 is intentional
  (`_cancel_conflict_detail`, pinned by `test_cancel_successful_run_returns_409`)
  and is not changed here.

  ## What changed

  - When reconnecting (`joinStream`) to a run that has already reached a terminal
    state (`success` / `error` / `timeout`), the frontend now preflights with a
    `GET /runs/{run_id}` and, on a terminal status, skips the join and clears the
    stale reconnect key so the SDK runs its `onSuccess` path and flips `isLoading`
    back to false. Net effect: after reload the send button is no longer stuck in
    the stop state and the first message sends normally — no more freeze.
  - The `is not cancellable (status: ...)` 409 raised when stop is clicked during
    the finish window (backend already `success`, SSE not yet drained) is now
    silently treated as a no-op and the stale key is cleared, instead of
    surfacing as an unhandled rejection.
  - Both paths match only the backend's terminal-state branch phrasing. The
    multi-instance 409 for a run still running on another worker
    (`not active on this worker and cannot be cancelled`) is still rethrown, so a
    real cancel failure on a live run is never masked.
  - Normal flows (stopping a still-running run, reconnecting to an active stream)
    are unchanged.

  ## Surface area

  - [x] **Frontend UI** — LangGraph client wrapper layer in
        `frontend/src/core/api/api-client.ts` (`runs.cancel` / `runs.joinStream`)
        plus unit tests.
  - [ ] **Backend API**
  - [ ] **Agents / LangGraph**
  - [ ] **Sandbox**
  - [ ] **Skills**
  - [ ] **Dependencies**

  ## Screenshots / Recording

  No visual UI change. The fix changes post-reload send-button interaction
  behavior (no longer stuck in stop state, no more 409), which a static
  screenshot cannot convey; see the linked issue for repro steps and the 409
  network response.

  ## Bug fix verification

  - Test path reproducing the bug: `frontend/tests/unit/core/api/api-client.test.ts`
    - `swallows terminal-state cancel 409 and clears stale key`
    - `rethrows not-active-on-worker cancel 409` (guards against masking the
      multi-instance real error)
    - `short-circuits reconnect to a terminal run`
    - `falls back to join when preflight cannot resolve the run`
  - Red on `main`, green on this branch: yes. The `isRunNotCancellableError` /
    terminal-preflight assertions fail on unfixed `main` (cancel 409 surfaces as
    a rejection; a terminal run would proceed to join) and pass on this branch.
    Full `pnpm test` suite: 347 passing, 0 failing.
  - What a red test could not cheaply cover: the reload-timing race itself
    depends on real SSE timing. Unit tests mock the terminal fetch response to
    lock down the wrapper behavior; end-to-end reload timing is left to manual
    verification.

  ## Validation

  What I actually ran:
  - `cd frontend && pnpm test -- tests/unit/core/api/api-client.test.ts` → 347 passed, 0 failed
  - `cd frontend && pnpm exec eslint src/core/api/api-client.ts tests/unit/core/api/api-client.test.ts` → passed
  - `cd frontend && pnpm typecheck` → `src/` clean (only pre-existing
    `@rstest/core` / `@rsbuild/plugin-react` module-resolution warnings under
    `tests/` and `rstest.config.ts`, unrelated to this change)

  Please run before merge (I did not run these locally; listing for review):
  - `cd frontend && BETTER_AUTH_SECRET=local-dev-secret pnpm build`

  ## AI assistance

  **Tool(s) used:** Claude Code

  **How you used it:** Used Claude Code to locate the root cause (reading SDK
  `stream.lgp.js` / `manager.js` and backend `thread_runs.py` / `runs/manager.py`),
  draft the `api-client.ts` cancel / joinStream wrappers and unit tests. I
  reviewed every line and tightened the match conditions (narrowed to only the
  terminal-state branch phrasing), added the `clearReconnectRun` calls, and
  extracted the shared helper.

  - [x] I've read and understand every line of this change and take responsibility for it — it's not unreviewed AI
  output.


frontend/src/core/api/api-client.ts
```diff
@@ -99,8 +180,35 @@ function createCompatibleClient(isMock?: boolean): LangGraphClient {
...
+    }
+  }) as typeof client.runs.cancel;
+
   const originalJoinStream = client.runs.joinStream.bind(client.runs);
   client.runs.joinStream = async function* (threadId, runId, options) {
+    // Short-circuit reconnects to runs that have already finished: otherwise a
+    // reload after the backend's stream bridge is reaped blocks forever on a
+    // drained condition variable, pinning ``isLoading`` true so the first
+    // post-reload message is routed to ``stop`` instead of ``submit``.
+    if (threadId && (await shouldSkipReconnect(client, threadId, runId))) {
```
willem-bd, line 215: **Behavior nuance (low severity): terminal short-circuit for `error`/`timeout` skips the live-bridge error replay.**

`TERMINAL_RUN_STATUSES` includes `error`/`timeout`, so `shouldSkipReconnect` short-circuits those reconnects too. In the narrow window where a run reaches `error`/`timeout` and the user reloads *before the in-memory bridge is reaped* (~60s), the old path rejoined the live bridge and replayed the buffered error event → `onError` (toast via `getStreamErrorMessage`, `hooks.ts:1003`). The new preflight sees the terminal status and short-circuits to `onSuccess`, so that error toast is no longer shown on reload.

The persisted error state still loads from the checkpoint via `useThreadHistory`, so only the transient toast is dropped in this window — arguably desirable on reload — but it is a behavior change beyond the `success` case the PR description focuses on. Worth a conscious sign-off (and maybe a one-line comment noting the intent for `error`/`timeout`).

WillemJiang:
@lllyfff thanks for the contribution, please fix the lint error and take a look at the review comments.

lllyfff:
Thanks @willem-bd — good catch, and conscious sign-off on this. Including error/timeout is intentional: on reload the 
persisted error state still loads from the checkpoint via useThreadHistory, so only the transient onError toast is    
dropped, which I think is preferable to a stale error toast firing on every reload. Added a comment block above       
TERMINAL_RUN_STATUSES noting the intent. The latest commit also fixes a prettier --check failure in the test file.

willem-bd (commented):
Reviewed the core mechanism against both the SDK internals and the backend (`_cancel_conflict_detail`, `worker.py`): the cancel wrapper is on the SDK `stop()` path, an empty `joinStream` generator correctly falls through to `onSuccess` (clearing the key + flipping `isLoading`), the matcher narrowing correctly excludes the multi-instance "not active on this worker" branch, and `interrupted` is genuinely terminal in DeerFlow (only set on abort/cancel paths, not HITL). Nothing blocking — the inline comments below are non-blocking observations.

frontend/src/core/api/api-client.ts
```diff
@@ -38,7 +38,50 @@ function injectCsrfHeader(_url: URL, init: RequestInit): RequestInit {
...
+ *
+ * Every passed ``needles`` substring must be present; this AND semantics is what
+ * lets a caller distinguish sibling conflict branches by phrase (e.g. the
+ * terminal-state cancel branch from the still-active-on-another-worker branch).
+ *
+ * Match strings until the API exposes a structured error code; the source of
+ * truth is ``_cancel_conflict_detail`` / the store-only response in
+ * ``backend/app/gateway/routers/thread_runs.py``.
+ */
+function isRunConflictError(error: unknown, ...needles: string[]): boolean {
```
willem-bd, line 84: Non-blocking, but worth a follow-up: this matcher keys off the human-readable `detail` strings from `_cancel_conflict_detail` (`"is not cancellable"`, `"not active on this worker"`, `"cannot be streamed"`). Any rewording or future i18n of those strings in `backend/app/gateway/routers/thread_runs.py` silently re-breaks both the cancel swallow and the inactive-stream clear, and no test pins the backend-phrasing contract (the tests construct their own strings). The docstring already flags this ("Match strings until the API exposes a structured error code") — the durable fix is a structured error code from the gateway so this coupling isn't load-bearing on prose. Filed as an observation, not a merge blocker.

frontend/src/core/api/api-client.ts
```diff
@@ -38,7 +38,50 @@ function injectCsrfHeader(_url: URL, init: RequestInit): RequestInit {
...
+// nothing left to stream — its state lives in the checkpoint, fetched
+// independently by ``useThreadHistory``, and resuming means a fresh ``submit``.
+//
+// ``error``/``timeout`` are terminal too, so a reload within the ~60s
+// bridge-reap window no longer replays the buffered error event through
+// ``onError`` — the transient error toast (``getStreamErrorMessage``) is
+// dropped. The persisted error state still loads from the checkpoint via
+// ``useThreadHistory``, so only the toast is lost; that is intentional, since
+// surfacing a stale error toast on every reload is noise rather than signal.
+const TERMINAL_RUN_STATUSES = new Set([
```
willem-bd, line 63: Heads-up on a user-visible behavior change for explicit sign-off: because `error`/`timeout` are terminal, a reconnect now short-circuits and the buffered `error` SSE event is never replayed through `onError`. Within the ~60s bridge-reap window, reloading a run that *just* errored will no longer surface the error toast via the stream path — only the persisted checkpoint state loads via `useThreadHistory`. The block comment declares this intentional ("surfacing a stale error toast on every reload is noise"), which is reasonable, but flagging it because it's the kind of change a user could notice ("I reloaded and my error message disappeared"). If the transient-on-error reload case matters, an `error`/`timeout`-only replay (or only short-circuiting `success`/`interrupted`) would preserve it.

frontend/src/core/api/api-client.ts
```diff
@@ -52,16 +95,61 @@ export function isInactiveRunStreamError(error: unknown): boolean {
...
+ * nothing to rejoin. Returns ``true`` when the caller should skip the
+ * underlying ``joinStream`` so the SDK's ``onSuccess`` path runs and
+ * ``isLoading`` flips back to false — instead of blocking forever on a drained
+ * stream bridge.
+ *
+ * Any error (404 for an evicted record, network blip, auth hiccup, …) falls
+ * back to the original join so a legitimately active reconnect is never
+ * silently suppressed.
+ */
+async function shouldSkipReconnect(
```
willem-bd, line 140: Two minor notes on the preflight (non-blocking):

1. `client.runs.get` is called without the SDK's abort `signal`, so navigating away mid-reconnect lets the GET run to completion — negligible, but inconsistent with the underlying join, which is signal-bound.
2. Every reload with a stored run now pays one extra synchronous GET before join can start; this sits on the reload hot path.

The catch-all fallback to join on any preflight error is the right call (a 404/eviction/auth blip never suppresses a legitimate reconnect). Just calling the two items out for awareness.

WillemJiang (approved):

PR #3927, opened 2026-07-04
fix(channels): offload blocking filesystem IO in Discord channel

hata33:
## Why

The backend runs a strict "no blocking IO on the event loop" gate (`make detect-blocking-io` + the Blockbuster suite in `tests/blocking_io/`). The scanner flagged `backend/app/channels/discord.py` with 7 blocking-IO findings — synchronous filesystem calls sitting on async entry points:

- `start()` → `_load_active_threads` (`exists` + `read_text`) restoring thread mappings on startup
- `_on_message` → `_save_thread` (`exists` + `read_text` + `mkdir` + `write_text`) when a new Discord thread is created
- `send_file` → `open()` reading outbound attachments directly on the loop

Each trips `BlockingError` under the gate and stalls the Gateway's asyncio loop while the Discord channel does its file IO. Tool-found via `make detect-blocking-io` (no associated issue; wechat/skills/agents blocking-IO already have PRs, discord did not).

## What changed

Offloaded every synchronous filesystem call to a worker thread via `await asyncio.to_thread(...)`, matching the pattern in `channels/wechat.py` and `channels/manager.py`. The sync helpers (`_load_active_threads`, `_save_thread`) keep their signatures; only the async call sites wrap them:

- `start()`: `await asyncio.to_thread(self._load_active_threads)`
- `_on_message` (3 sites): `await asyncio.to_thread(self._save_thread, channel_id, thread_id)`
- `send_file`: reads the attachment via a new threaded helper `_read_attachment_bytes`, then hands discord.py an in-memory `BytesIO` buffer instead of a live file handle. The bytes are consumed during `target.send` on `_discord_loop`, after which the buffer is reclaimable — this also avoids leaking a file handle on both success and failure paths.

`__init__` was already IO-free (only path computation: `Path.home()` / `store._path.parent`), so `ChannelService._start_channel()` constructing the channel inline on the async path does not block. No behavior change: thread mappings are persisted/restored identically, attachments uploaded identically — just off the loop.

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API** — change is under `backend/app/channels`, but it's an internal IO offload; no endpoint / SSE event / request-response shape change.
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [x] **Docs / tests / CI only**

## Bug fix verification

- Test path: `backend/tests/blocking_io/test_discord_channel_state.py` — three async tests under the strict Blockbuster gate covering the IO-free constructor (the production `_start_channel` path), the `_save_thread` write path, and the `_load_active_threads` read path.

## Validation

- `cd backend && make detect-blocking-io` → `discord.py` drops out of the top offending files (was 7 findings).
- `cd backend && PYTHONPATH=. uv run pytest tests/blocking_io/test_discord_channel_state.py -v` → 3/3 pass.
- `cd backend && PYTHONPATH=. uv run pytest tests/ -k discord` → 13 passed, 2 skipped (no existing discord tests broken).
- `cd backend && uv run ruff check` and `uv run ruff format --check` → clean.
- Branch rebased onto the latest `upstream/main` (`94c852ac`); three-dot diff is exactly 2 files.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** I drove the work end-to-end. The AI used `make detect-blocking-io` to locate the blocking-IO sites, applied the established `asyncio.to_thread` offload pattern matching `channels/wechat.py` / `manager.py`, and wrote the regression tests. I reviewed every line of the diff.

### DIFF
--- a/backend/app/channels/discord.py
+++ b/backend/app/channels/discord.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import io
 import json
 import logging
 import threading
@@ -108,7 +109,7 @@ class DiscordChannel(Channel):
 
         self._thread = threading.Thread(target=self._run_client, daemon=True)
         self._thread.start()
-        self._load_active_threads()
+        await asyncio.to_thread(self._load_active_threads)
         logger.info("Discord channel started")
 
     def _load_active_threads(self) -> None:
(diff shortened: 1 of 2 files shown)


hata33:
Thanks @copilot-pull-request-reviewer — valid catch, addressed in the force-push (`60f81067`).

The race was real: `_save_thread` updated `_active_thread_ids` *inside* the worker thread, so during the `await` an inbound message in the newly created thread could miss `_on_message`'s membership check (`thread_id in self._active_thread_ids`) and be misclassified as orphaned — then a duplicate thread would be created.

Fix: split `_save_thread`'s two responsibilities apart —
- **`_record_thread_mapping(channel_id, thread_id)`** — synchronous, runs on the event loop (no IO, no await). Updates `_active_threads` + `_active_thread_ids` (discards the replaced id, adds the new one), so the new thread is recognizable immediately.
- **`_persist_thread_mappings()`** — pure filesystem IO, offloaded via `asyncio.to_thread`.

The 3 call sites in `_on_message` now do `_record_thread_mapping(...)` then `await asyncio.to_thread(self._persist_thread_mappings)`. Memory visibility no longer depends on persistence latency.

Added a regression anchor — `test_discord_record_thread_mapping_visible_before_persist` — that records a mapping *without* awaiting persistence and asserts `_active_thread_ids` already contains it (and the file is not written yet), pinning the race closed.

WillemJiang (approved):

Issue #3875, opened 2026-06-29
[bug] Subagents can loop unbounded on trivial tasks (no loop-detection / summarization / budget)


ggnnggez:
### What happened

A smoke test (Ultra mode): ask the agent to spawn 3 subagents, each writing one sentence to a local file, then have the lead combine them into a new file. This trivial task burned **~4.4M tokens over ~16 min** and had to be **manually interrupted**.

From the run record:
- glm-5.2: **input 4,439,523 / output 4,543** (input:output ≈ 1000:1).
- The lead made only **1** real dispatch call (clean — 3 `task` calls, ~12.5K input); `subagent_tokens` = **4.43M (99.7%)**.
- Estimated **~110 turns per subagent** (context re-sent every turn, ~12 output tokens/turn) — i.e. each "write one sentence" subagent fell into a degenerate loop.

### Root cause

Subagents inherit **none** of the lead's runaway guards. `build_subagent_runtime_middlewares` attaches no `LoopDetectionMiddleware`, `SummarizationMiddleware`, or `TokenBudgetMiddleware` — all three are lead-only (added in `build_middlewares`). So:
- no loop detection → a degenerate loop runs unchecked;
- no summarization → each turn re-sends a growing context (input explodes);
- `general-purpose` subagent has `max_turns=150`.

The only thing that stopped it was a manual interrupt (the 4 `CancelledError`s in the run).

### Expected

A subagent given a trivial task should finish in a few turns. A degenerate loop should be caught early (loop detection) and capped (turn/token budget) instead of burning millions of tokens.

### Notes

- Structural gap in the subagent middleware chain, not specific to the task content.
- Files: `packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py::build_subagent_runtime_middlewares`; `packages/harness/deerflow/subagents/builtins/general_purpose.py` (`max_turns=150`).
- Tracked as the prerequisite item in #3857.

fancyboi999:
The diagnosis matches the current middleware split.

`build_subagent_runtime_middlewares()` currently builds the shared runtime chain and then adds only subagent-specific pieces such as vision/deferred-tool filtering/safety-finish handling. The lead-only chain is where `DeerFlowSummarizationMiddleware`, `LoopDetectionMiddleware`, and `TokenBudgetMiddleware` are appended. `general-purpose` also still has `max_turns=150`, so a degenerate subagent can burn a lot of input before the existing timeout is the only hard backstop.

[...]

PR #3931, opened 2026-07-04
fix(subagents): inherit LoopDetectionMiddleware to break tool loops (#3875)

hata33:
## Summary

Phase 1 of #3875: subagents now inherit `LoopDetectionMiddleware`, so a degenerate subagent tool loop is detected and broken instead of running unchecked until `max_turns` (the reported ~4.4M-token burn over ~110 turns/subagent).

## Root cause

`build_subagent_runtime_middlewares()` attached **none** of the lead's runaway guards — no `LoopDetectionMiddleware`, `SummarizationMiddleware`, or `TokenBudgetMiddleware` (all three are lead-only). So a subagent stuck in a tool loop re-sent a growing context every turn until `max_turns=150` (`general-purpose`).

## Fix (Phase 1 — smallest blast radius)

Append `LoopDetectionMiddleware.from_config(app_config.loop_detection)` to the subagent chain, gated by the existing `loop_detection.enabled` config (default on) — **no new config field**.

- Registered **before** `SafetyFinishReasonMiddleware`, matching the lead's after-model ordering (the safety middleware's docstring requires placement after `LoopDetection`).
- Subagents disallow `task`, so only the tool-loop heuristic can fire — no recursive-delegation false positives (per @fancyboi999's triage note).

## Why Phase 1 only

Per the phased plan I floated on #3875, Phase 1 is the structural gap with zero new config decisions. Phases 2–3 need sign-off:

- **Phase 2 — deterministic turn/token budget with a lead-visible stop reason.** Open questions:
  1. Budget config: a new `subagents.*` field, or reuse `loop_detection`?
  2. Should "capped" map to a new `subagent_status` enum value so the lead can distinguish "finished" from "capped"?
- **Phase 3 — summarization** (deferred; "summarization alone can hide the loop").

## Test plan

- `test_subagent_runtime_middlewares_attach_loop_detection_when_enabled` — default chain carries the guard.
- `test_subagent_runtime_middlewares_omit_loop_detection_when_disabled` — `enabled=False` drops it (mirrors the lead gate).
- `test_subagent_runtime_middlewares_place_loop_detection_before_safety_finish` — ordering matches the lead chain.
- Updated the existing `len(middlewares)` baseline (`+1` for the new default-on middleware).

`LoopDetectionMiddleware` behavior itself is covered by `test_loop_detection_middleware.py`; this PR is the wiring, not new detection logic.

`ruff format` + `ruff check` clean; `detect-blocking-io` unaffected (no IO in the change).

cc @fancyboi999 — Phase 1 ready for review; the Phase 2 config/status questions above are what I'd like your call on before starting Phase 2.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
@@ -252,6 +252,26 @@ def build_subagent_runtime_middlewares(
 
         middlewares.append(DeferredToolFilterMiddleware(deferred_setup.deferred_names, deferred_setup.catalog_hash))
 
+    # LoopDetectionMiddleware — subagents inherit none of the lead's runaway
+    # guards today (see #3875): with no loop detection a degenerate subagent tool
+    # loop runs unchecked until ``max_turns``, re-sending a growing context each
+    # turn (the reported 4.4M-token burn). Mirror the lead chain so the loop is
+    # detected and broken. Subagents disallow ``task``, so only the tool-loop
+    # heuristic can fire here — no recursive-delegation path to false-positive on.
+    # Registered before SafetyFinishReasonMiddleware (earlier in the list).
+    # LangChain dispatches after_model hooks in REVERSE registration order, so
+    # SafetyFinishReasonMiddleware (appended below) executes first and strips
+    # safety-terminated tool_calls; LoopDetectionMiddleware then accounts on the
+    # cleaned message. This is the placement SafetyFinishReasonMiddleware's
+    # docstring requires ("register after LoopDetection") and mirrors the lead
+    # chain (``lead_agent/agent.py``). Phase 1 of #3875; a deterministic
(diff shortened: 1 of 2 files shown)


hata33:
Thanks @copilot-pull-request-reviewer — fair point, addressed in the force-push (`958fab78`).

The code ordering is intentional and correct: `LoopDetectionMiddleware` is registered before `SafetyFinishReasonMiddleware`, so reverse `after_model` dispatch runs `SafetyFinishReasonMiddleware` first (stripping safety-terminated `tool_calls`) and `LoopDetectionMiddleware` second (accounting on the cleaned message) — exactly the placement `SafetyFinishReasonMiddleware`'s docstring requires and what the lead chain (`lead_agent/agent.py`) uses.

Only the wording was wrong: the comment and the test docstring described registration order as if it were execution order. Reworded both to state registration order explicitly and call out the reverse dispatch, so a future reader doesn't try to "fix" the list order and break it. The test assertion (`loop_idx < safety_idx`, pinning registration order) is unchanged.

WillemJiang (approved):

PR #3933, opened 2026-07-04
fix(redis): stream retention recovery

### LINKED ISSUE #3932: Redis stream bridge: mid-run worker/pod death leaves SSE client hung and stream key orphaned
Follow-up to #3191 (Redis stream bridge), which resolved cross-worker SSE delivery and `Last-Event-ID` replay. This issue tracks a gap in the **producer-death** path that the bridge doesn't close end-to-end. cc @Janlaywss (author of #3191); parts of this were flagged during #3191 review by @fancyboi999 and @willem-bd.

## Summary

`RedisStreamBridge` lets a client reattach to buffered events from any worker, but if the producing worker dies **mid-run**, the SSE stream never terminates cleanly: no `__end__` marker is published, the run stays `running` in the store, and a reconnecting client hangs on heartbeats indefinitely. Separately, the per-run stream key can leak in Redis because cleanup is a fire-and-forget `DELETE` with no `EXPIRE` fallback.

## What's in place today

- `publish_end` is called from exactly one site — the normal completion path at `worker.py:594`. If the worker exits before reaching it, the stream gets no `__end__` marker.
- `cleanup` is `asyncio.create_task(bridge.cleanup(run_id, delay=60))` at `worker.py:595`, and the Redis impl is `DELETE` (`redis.py:206`). No `EXPIRE` is set anywhere on the stream key, so if the process dies before the 60s task fires, the key persists.
- There is an orphan-recovery path — `reconcile_orphaned_inflight_runs` (`manager.py:696`, invoked at gateway startup from `deps.py:219`) — but:
  - it is **startup-only**, not continuous, so a mid-run pod death in a live multi-pod cluster isn't reaped until the next gateway restart;
  - it marks the run `error` in the RunStore but does **not** call `bridge.publish_end()`, so the Redis stream still has no end marker.
- `_terminal_record_stream_missing` (`app/gateway/services.py`) only short-circuits the consumer when the run is terminal **and** the stream is gone. For a mid-run death the run is non-terminal; even after the startup reaper marks it `error`, `stream_exists` is still true (buffered events, no end marker), so the consumer keeps looping on `HEARTBEAT_SENTINEL` (`redis.py:189`).

Net effect: the store eventually says `error` (on next restart), but the SSE client says "still waiting" — forever.

This matters most for multi-pod / K8s deployments with real pod churn (rolling updates, autoscaling, node drains) — i.e. the scenario #3191's Docker / multi-worker scope doesn't fully cover.

## Proposed

Two independent pieces:

1. **TTL safety net on the stream key.** Set `EXPIRE` on `deerflow:stream_bridge:{run_id}` — e.g. at `publish_end` with a multi-hour TTL (still allows late replay, bounds memory), or a rolling TTL refreshed on each `XADD`. This bounds Redis independently of the Python cleanup path. (Raised in #3191 review by @fancyboi999.)

2. **Reaper publishes `__end__` on behalf of dead workers.** The cheapest improvement is to extend `reconcile_orphaned_inflight_runs` to also `bridge.publish_end(run_id)` (and `cleanup`) for each reaped run, so subscribers receive `END_SENTINEL` and terminate. That only fires at restart. The full fix is a continuous liveness scan (with some worker-ownership / heartbeat notion in `RunRecord`) that does the same. A cheaper consumer-side fallback is to give up after N consecutive heartbeats with no stream growth on a non-terminal run.

## Tests that would lock this in

- Mid-run producer death (kill the worker after `publish`, before `publish_end`) → a reconnecting client terminates (receives `end`) within a bounded window, and the run ends `error` in the store.
- Stream key is reclaimed by Redis after the TTL even when `cleanup` never runs.

Janlaywss:
## Summary

Fixes #3932 by adding a Redis Stream retention safety net and making startup recovery terminate retained streams for orphaned runs.

## Changes

- Add a rolling Redis Stream TTL that refreshes on every publish and END marker, with `stream_ttl_seconds` config and a zero-value opt-out.
- Publish END markers during gateway startup recovery for orphaned in-flight runs, and schedule delayed stream cleanup with configurable `recovered_stream_cleanup_delay_seconds`.
- Avoid recreating already-expired streams during recovery by checking stream existence before publishing END.
- Document the TTL as a memory safety net, not a subscriber termination mechanism, and update config examples.
- Add unit and Redis integration coverage for TTL refresh, TTL disablement, key reclamation, and recovery END behavior.

## Root Cause

Redis StreamBridge retained stream keys only relied on explicit cleanup. If cleanup did not run, abandoned stream keys could remain indefinitely, while recovered orphaned runs did not proactively unblock retained stream subscribers with an END marker.

## Validation

- `cd backend && make test` -> `5799 passed, 23 skipped, 11 warnings`
- `cd backend && uv run pytest tests/test_gateway_run_recovery.py tests/test_stream_bridge.py -q` -> `36 passed, 5 skipped, 1 warning`
- `cd backend && uv run ruff format --check app/gateway/deps.py tests/test_gateway_run_recovery.py packages/harness/deerflow/runtime/stream_bridge/async_provider.py packages/harness/deerflow/runtime/stream_bridge/redis.py tests/test_stream_bridge.py`
- `cd backend && uv run ruff check app/gateway/deps.py tests/test_gateway_run_recovery.py packages/harness/deerflow/runtime/stream_bridge/async_provider.py packages/harness/deerflow/runtime/stream_bridge/redis.py tests/test_stream_bridge.py`
- Redis integration tests were run against a local Docker Redis earlier in the branch.


Janlaywss:
cc @WillemJiang @zhfeng

WillemJiang (approved):

PR #3934, opened 2026-07-04
feat(frontend): add side conversations for quoted follow-ups

18062706139fcz:
## Why

DeerFlow works well for linear conversations, but users often need to drill into a specific part of an assistant response without interrupting the main thread. This PR adds Codex-like side conversations: users can select text from an assistant message, open a side chat, and ask follow-up questions grounded in that quoted material while preserving the main conversation.

## What's New

- Adds two actions for selected assistant text:
  - `Add to conversation`: attach the quote to the main composer.
  - `Ask in side chat`: open a right-side follow-up conversation.
- Supports multiple quoted text fragments.
- Stores side conversations as independent threads, keeping the main thread clean.
- Keeps a persistent toolbar entry after a side conversation exists, so users can reopen or close it.
- Aligns the side composer with the main composer: attachments, mode selection, model selection, upload limit checks, and send state.
- Allows selecting text inside side conversation messages as well.
- Shows reference summaries on visible user messages so users can see when a message used quoted material.
- Shows token usage in side conversations.

## How It Works

- Adds a new `sidecar` core module for quote context, sidecar thread metadata, reference metadata, and sidecar thread lookup.
- Creates sidecar threads with metadata such as `deerflow_sidecar`, `parent_thread_id`, and `referenced_message_ids`, making them easy to inspect and trace.
- Sends quoted context through hidden context messages:
  - Main conversation quotes use hidden `conversation_quote_context`.
  - Side conversations use hidden `sidecar_context`.
  - Visible user messages keep `referenced_message_contexts` for UI rendering and debugging.
- Includes recent parent conversation context in side chat prompts as read-only background, while instructing the model to answer only the latest side question.
- Filters sidecar threads out of primary thread lists by default.
- Deletes attached sidecar threads when deleting their parent thread.
- Keeps the side panel mutually exclusive with the artifacts panel.

## UX Details

- Quote chips and tooltips are compact, with long references clamped instead of locally scrollable.
- The side panel opens and closes with width/opacity transitions, avoiding unwanted scroll animation when restored history is shown.
- Long model names are hidden on narrow widths to avoid composer overflow.
- Main and side composers are aligned in height, bottom spacing, and model/send-button spacing.

## Verification

- `pnpm check`
- `git diff --check`
- `pnpm test tests/unit/core/sidecar tests/unit/core/threads/delete-thread.test.ts tests/unit/core/threads/send-message.test.ts tests/unit/core/messages/usage.test.ts`
- `PLAYWRIGHT_SKIP_WEB_SERVER=1 PLAYWRIGHT_BASE_URL=http://127.0.0.1:3002 pnpm exec playwright test tests/e2e/sidecar-chat.spec.ts`

image

image

image

image

image


frontend/src/components/workspace/input-box.tsx
```diff
@@ -559,6 +633,30 @@ export function InputBox({
...
+              }),
+            ],
+          }
+        : undefined;
+      const submit = () => {
+        const result = onSubmit?.(message, submitOptions);
+        if (quoteIds.length > 0) {
+          void Promise.resolve(result)
+            .then(() => {
+              sidecar?.clearConversationQuotes(quoteIds);
```
willem-bd: **Quotes are cleared even when the send is dropped by the in-flight guard.**

`submit()` captures `const result = onSubmit?.(message, submitOptions)` and then unconditionally runs `Promise.resolve(result).then(() => clearConversationQuotes(quoteIds))`. The bound `onSubmit` ultimately calls `sendMessage`, which early-returns `undefined` when `sendInFlightRef.current` is true (`core/threads/hooks.ts` ~line 1208). `Promise.resolve(undefined).then(clear)` then fires on the next microtask regardless of whether anything was actually sent.

**Failure scenario:** user attaches a quote and hits Enter during the window between `sendInFlightRef` being set and `status` flipping to `"streaming"` (the two guards are separate). `sendMessage` no-ops, but the `.then(clear)` wipes the quote from the composer — the quote is gone and was never delivered, with no UI signal that the send was dropped. Better to clear only once the send has genuinely proceeded (e.g. have `sendMessage` return a sentinel, or clear inside its success path).

frontend/src/components/workspace/messages/message-list.tsx
```diff
@@ -370,237 +502,307 @@ export function MessageList({
...
-    </Conversation>
+          )}
+          <div style={{ height: `${paddingBottom}px` }} />
+        </ConversationContent>
+      </Conversation>
+      {selectionToolbar && sidecar && (
+        <div
+          className="bg-popover text-popover-foreground border-border fixed z-50 flex -translate-x-1/2 -translate-y-full items-center gap-1 rounded-full border p-1 shadow-lg"
+          data-sidecar-selection-toolbar
+          style={{ left: selectionToolbar.x, top: selectionToolbar.y }}
```
willem-bd, line 814: **The selection toolbar is clipped above the viewport when the selected text sits near the top.**

The toolbar is `position: fixed` with `-translate-x-1/2 -translate-y-full` and `style={{ top: selectionToolbar.y }}`, where `y = Math.max(12, rect.top - 8)` (around line 362). The `Math.max(12, …)` clamp doesn't account for the `-translate-y-full` shift, which moves the element up by its own height.

**Failure scenario:** scroll an assistant message so the selected phrase sits within ~50px of the viewport top, so `rect.top - 8` clamps to 12. After the full-height upward translate, the toolbar's top edge lands above `y=0` and the whole pill is off-screen — both "Add to conversation" and "Ask in side chat" are unreachable (Escape/scroll are the only ways out). Either flip the toolbar *below* the selection when `rect.top` is small, or subtract the toolbar height from the clamp.

frontend/src/core/sidecar/reference-metadata.ts
```diff
@@ -0,0 +1,81 @@
...
+  return value === "user" || value === "assistant";
+}
+
+export function buildReferenceMessageMetadata(
+  contexts: SidecarContext[],
+): ReferenceMessageMetadata {
+  return {
+    referenced_message_count: contexts.length,
+    referenced_message_ids: Array.from(
+      new Set(
```
willem-bd: **`referenced_message_ids` is deduped, but `referenced_message_roles` and `referenced_message_count` are not — the arrays can diverge.**

`referenced_message_ids` runs through `Array.from(new Set(...))`, while `referenced_message_roles` is `contexts.map(c => c.role)` and `referenced_message_count` is `contexts.length`.

**Failure scenario:** the user quotes two non-overlapping excerpts of the *same* assistant message. `appendSidecarReference` only dedups on identical content, so both fragments survive with the same `messageId`. The persisted metadata then has `ids.length === 1` but `roles.length === 2` and `count === 2` — non-parallel arrays and `ids.length !== count`. Any consumer (now or later) that zips `ids` with `roles`, or asserts `ids.length === count`, mis-attributes roles or reads past the end of `ids`. Either drop the `Set` so all three stay parallel, or derive `count` from the deduped set and document `ids` as "unique source message ids".

frontend/src/components/workspace/messages/message-list.tsx
```diff
@@ -250,6 +285,104 @@ export function MessageList({
...
+        setSelectionToolbar(null);
+        return;
+      }
+
+      const container = event.currentTarget;
+      if (
+        !selection.anchorNode ||
+        !selection.focusNode ||
+        !container.contains(selection.anchorNode) ||
+        !container.contains(selection.focusNode)
```
willem-bd: **A selection that spans two assistant messages silently yields no toolbar.**

`onMouseUp` is bound per-message, and the handler bails when `!container.contains(selection.focusNode)`, where `container` is the current message wrapper.

**Failure scenario:** in a multi-message assistant turn, the user drags a selection that starts in the first AI message and ends in the second. `focusNode` lives in the second message's wrapper, so the check fails; the native highlight stays on screen, no toolbar appears, and there's no feedback that the quote is unusable. Worth either widening the containment check to the shared turn container, or showing a brief "selection crosses messages" hint.

frontend/src/core/sidecar/api.ts
```diff
@@ -0,0 +1,70 @@
...
+  isSidecarThread,
+} from "./thread";
+
+type SidecarThreadSearchClient = {
+  threads: {
+    search: (query: Record<string, unknown>) => Promise<AgentThread[]>;
+  };
+};
+
+export async function createSidecarThread({
```
willem-bd, line 26: **The find-then-create pattern isn't atomic, so two callers can create duplicate sidecar threads for one parent.**

The lookup-or-create flow is two independent round-trips — `findLatestSidecarThread({ parentThreadId })` → null → `createSidecarThread(...)` — with no upsert and no client-side lock.

**Failure scenario:** a double-click on "Ask in side chat", or two tabs open on the same parent thread: both callers find `null` and both create a sidecar. The two resulting threads share `parent_thread_id`; later `findLatestSidecarThread` calls return whichever the backend sorts first, orphaning the other (it still satisfies `isSidecarThread`, so it's filtered out of primary lists and becomes unreachable). An upsert/dedup on the backend, or a client-side in-flight create promise keyed by `parentThreadId`, would close the gap.

frontend/src/components/workspace/sidecar/sidecar-trigger.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import { Tooltip } from "../tooltip";
+
+import { useMaybeSidecar } from "./context";
+
+export function SidecarTrigger() {
+  const { t } = useI18n();
+  const sidecar = useMaybeSidecar();
+
+  if (!sidecar?.sidecarThreadId) {
```
willem-bd, line 18: **The trigger stays visible after the sidecar is deleted elsewhere, and reopening points at a dead thread.**

Visibility is keyed off the cached in-memory `sidecar?.sidecarThreadId` (`if (!sidecar?.sidecarThreadId) return null`), and that state has no listener for external deletion (another client, a thread-management view, or server-side cleanup).

**Failure scenario:** the side thread is deleted through any other surface. `sidecarThreadId` stays set, so the toolbar button remains visible indefinitely; clicking it reopens the panel pointed at a dead thread, where `useThreadStream` errors or shows an empty history with no recovery path. Re-running `restoreSidecarThread` on trigger click and hiding the trigger if it returns null would make this self-healing.

WillemJiang:
@18062706139fcz  please take a look at the review comments and resolve the conflicts with the main branch.

WillemJiang (approved):

PR #3938, opened 2026-07-04
feat(skills): bind request-scoped secrets for autonomously-invoked skills (A+)

fancyboi999:
## Problem

`required-secrets` (#3861) only injects when a skill is activated with `/slash`. In real use — natural-language chat, IM channels, or a single turn that needs several skills — the user does not type `/skill-name`; the model autonomously reads a SKILL.md when the task matches. In that path the skill never received its declared secrets. Gap 1 of #3914.

The concern that blocked this originally was: does letting autonomously-invoked skills receive secrets weaken the security boundary? We researched it before designing (see the [design comment on #3914](https://github.com/bytedance/deer-flow/issues/3914#issuecomment-4880680000)): read the implementations of 7 peer systems (Claude Code, OpenAI Codex CLI, opencode, pi, LangChain deepagents, hermes-agent, QwenPaw) and 4 spec families (agentskills.io, MCP 2025-06-18/2025-11-25, Claude API Skills, Copilot/OpenAI/Dify/n8n credential models). Findings:

- The industry trust boundary is **enable-time consent + caller-scoped credentials, resolved silently at execution outside the model context** — not per-invocation ceremony (MCP 2025-11-25 makes "secrets never transit the LLM context" a MUST; Copilot/OpenAI inject a per-request token bound by a one-time OAuth).
- **No surveyed system scopes secrets to an "activation turn."** #3861's binding point A was already stricter than the whole field; keeping "autonomous = never" was our conservative default, not an industry norm.
- The real cautionary tale for declaration-based schemes is hermes-agent's GHSA-rhgp-j443-p4rf: a malicious skill declared a platform key to exfiltrate it. **#3861 is immune by construction** — injected values come only from the caller's `context.secrets`; the host env is scrubbed first, so a declared name never harvests a host credential.

## Solution — binding point A+ (in-context binding)

`SkillActivationMiddleware._resolve_secret_bindings` recomputes the injection set (`runtime.context["__active_skill_secrets"]`) on **every** model call and replaces it, unioning two sources:

1. **Slash** — the run's most recent `/skill` activation, persisted on the run context so the whole tool loop after the activation call keeps the binding (preserves #3861 semantics; validated once at activation).
2. **In-context** — skills the model actually loaded in this thread (`ThreadState.skill_context`, the existing durable record of read SKILL.md references), re-validated against the live registry every call.

Authorization stays **three-gated regardless of activation style**:

| Gate | Who | When |
|---|---|---|
| enabled | operator installs & enables the skill | enable-time (the industry consent point) |
| supplied | caller sends values in `context.secrets` on **this** request (never persisted server-side, never from host env) | per-request (Copilot/OpenAI-style scoping) |
| declared | only names in `required-secrets` can bind (∩ semantics) | declared |

A+ only widens the *activation* dimension (slash → slash ∪ in-context); all three gates hold. This directly answers the "must stay usable long-term" requirement: as long as the skill is in `skill_context` and the caller keeps supplying values, the binding persists across turns; eviction (capacity) or the caller ceasing supply revokes it on the next model call.

**Prompt-injection mitigation, stated plainly:** a hostile tool output can induce the model to *read* a SKILL.md, putting it in context. Binding ≠ leak — exfiltration still requires the operator to have enabled that skill, the caller to be supplying its secret this request, and a further injection to emit an exfil command (which already exists inside a slash turn). To bound the added surface: a new frontmatter field **`secrets-autonomous: false`** (default `true`; malformed → fail closed to `false`) keeps ceremony-only binding for high-sensitivity skills; binding changes emit a `middleware:skill_secrets` audit event (names only, never values).

**Non-goals** (unchanged from #3861): no server-side vault; subagents don't inherit the set; MCP per-user credentials (#3322) is a sibling.

## Validation

**Unit** — new `TestInContextBindsSecrets` (autonomous bind / eviction revokes / disabled / allowlist / opt-out vs slash / source merge / no-supply / audit names-only / unchanged-not-re-recorded / slash-persists-across-run) + `TestSecretsAutonomousParsing` (default/explicit/fail-closed). All failed on `main` before the change.

```
$ uv run pytest tests/test_skill_request_scoped_secrets.py tests/test_slash_skills.py \
    tests/test_skills_parser.py tests/test_lead_agent_skills.py -q
149 passed
$ make lint  → All checks passed
```
Full backend suite: 5842 passed; the only 7 failures are pre-existing on clean `main` (verified by reverting the changed files in-place — identical set).

**Real-model end-to-end** (local Gateway + `gpt-5.4` via the ccg gateway, `LocalSandboxProvider`, a test skill declaring `required-secrets: [TEST_ERP_TOKEN]`, **no `/slash` used**):

- Turn 1 "帮我导出我的 ERP 季度报表" → model autonomously loads the skill → script runs → real token `sk-e2e-user-alpha-9527` lands in the output file: `erp-report-for-token=[sk-e2e-user-alpha-9527]`.
- Turns 2–3 same thread, no skill mention, caller re-supplies the secret → still injected (long-lived binding across turns).
- Turns 4–5 caller stops supplying `context.secrets` → next injection resolves empty: `erp-report-for-token=[]` (supply-gate revocation, live).
- Leak surfaces re-checked on the real run: token absent from every API-returned message, from the run records / `kwargs` echo, and from the gateway log (grep count 0 on all three).

**Independent adversarial verification** (fresh reviewer agent, pre-PR) — all invariants passed, including cross-run isolation (`runtime.context` rebuilt per run), confused-deputy via forged `skill_context` (a name-squat entry can only bind the named skill's own declared secret, never a union or another skill's), async thread-safety, and parser fail-closed.

**Not verified** (fail-loud): a real Feishu/Slack webhook-to-sandbox run (verified each link in code + the bash path live, not a live IM message); AIO sandbox path (unchanged `bash_tool`, its own regression tests untouched); Windows.

Part of #3914 (gap 1). Gap 3 shipped in #3926; gap 2 was withdrawn by the issue author.


backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
```diff
@@ -262,59 +276,173 @@ def _prepare_model_request(self, request: ModelRequest, *, hook: str) -> ModelRe
...
+        for skill in skills:
+            by_path[skill.get_container_file_path(container_root)] = skill
+            by_name.setdefault(skill.name, skill)
+
+        sources: list[tuple[str, tuple[SecretRequirement, ...]]] = []
+        seen: set[str] = set()
+        for entry in entries:
+            if not isinstance(entry, dict):
+                continue
+            skill = by_path.get(entry.get("path")) or by_name.get(entry.get("name"))
```
fancyboi999: P2: this `or by_name.get(...)` fallback can bind secrets for a different skill than the one the model actually loaded. `skill_context` is intentionally path-validated and drops mismatched metadata (`backend/tests/test_skill_context.py` covers that), but here a missing/stale path silently degrades to name matching. That becomes observable because DeerFlow already allows same-name shadowing across skill roots (`load_skills()` de-dupes by name, and `UserScopedSkillStorage` explicitly lets custom skills shadow built-ins). In the shadowing case, a thread that previously read `/mnt/skills/public/foo/SKILL.md` can start binding the later custom `foo` skill's declared secrets even though that custom skill was never read in this thread. I think in-context binding needs to stay path-based only, or at least reject name fallback when multiple roots can legitimately share a name.

fancyboi999 (commented):
One behavior-level concern in the in-context secret binding path; see inline comment.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
fancyboi999: Good catch — fixed in 052060ef by dropping the by-name fallback entirely. In-context binding now matches strictly by the exact container file path the model read (which equals the registry's own `get_container_file_path`); a path that no longer resolves simply does not bind. This closes the shadowing confused-deputy (thread read public/foo → cannot bind custom/foo's secrets) and is robuster against the recent user-isolation path changes. Added two regression tests: `test_shadowing_name_does_not_bind_unread_skill` and `test_stale_path_does_not_fall_back_to_name`.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
```diff
@@ -262,59 +276,180 @@ def _prepare_model_request(self, request: ModelRequest, *, hook: str) -> ModelRe
...
+
+        if activation is not None:
+            context[_SLASH_SECRET_SOURCE_KEY] = {
+                "skill_name": activation.skill_name,
+                "requirements": [(req.name, req.optional) for req in activation.required_secrets],
+            }
 
         request_secrets = extract_request_secrets(context)
+        sources: list[tuple[str, tuple[SecretRequirement, ...]]] = []
+        slash_source = context.get(_SLASH_SECRET_SOURCE_KEY)
```
willem-bd: **Security: a forged `_SLASH_SECRET_SOURCE_KEY` bypasses the enabled / allowlist / `secrets-autonomous` gates.**

`build_run_config` (`app/gateway/services.py:318`) merges the caller's `config.context` into `runtime.context` via `dict(context_value)` with no `__`-key filtering, and `RunCreateRequest.config` is an unfiltered `dict[str, Any]` (`app/gateway/routers/thread_runs.py:63`). So a caller can POST `config.context.__slash_skill_secret_source = {"skill_name": "x", "requirements": [["ADMIN_TOKEN", false]]}` with **no** `/slash` in the messages and no skill loaded.

This slash-source path reads the key here (line 331) and appends its requirements to the injection set with **no** `skill.enabled` check, **no** `self._available_skills` allowlist check, and **no** `secrets_autonomous` check — all three are applied only on the in-context path (lines 431, 433). As a result `ADMIN_TOKEN` is written into `ACTIVE_SECRETS_CONTEXT_KEY` (the bash subprocess env) even when the skill is disabled, not allowlisted, or opted out via `secrets-autonomous: false` — the latter being this PR's primary prompt-injection mitigation.

The docstring claims the source is "validated once, at activation" and that authorization is "three-gated regardless of activation style"; both are false for a source that was never activated. The old (#3861) code only acted on the message-validated `_Activation` object, so this is a new forgeability surface.

Suggested fix: don't trust a caller-mergeable context key as authoritative — e.g. strip `__`-prefixed keys from caller `config.context` in `build_run_config`, or re-validate the slash source against the live registry (enabled + allowlist) on read.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
```diff
@@ -262,59 +276,180 @@ def _prepare_model_request(self, request: ModelRequest, *, hook: str) -> ModelRe
...
+        if isinstance(slash_source, dict) and slash_source.get("requirements"):
+            sources.append(
+                (
+                    str(slash_source.get("skill_name") or ""),
+                    tuple(SecretRequirement(name=name, optional=optional) for name, optional in slash_source["requirements"]),
```
willem-bd: **Robustness: unguarded unpack — a malformed `requirements` crashes the model call (DoS).**

`slash_source["requirements"]` is caller-controlled (same merge path as the gate-bypass comment at line 331). `for name, optional in slash_source["requirements"]` here raises `ValueError: not enough values to unpack` for any entry that isn't a 2-element iterable (e.g. `[["X"]]`, or a string `"abc"` that iterates into single chars).

There's no `try/except` in `_resolve_secret_bindings`, `_handle_model_request` (line 284), or `wrap_model_call`, so a single crafted `config.context.__slash_skill_secret_source = {"requirements": [["X"]]}` propagates as an unhandled exception and 500s the run. The middleware itself always writes well-formed 2-tuples (`[(req.name, req.optional) for req in activation.required_secrets]`), so normal operation is safe — the crash is purely caller-reachable.

Suggested fix: validate the entry shape before unpacking, and/or wrap the slash-source read in a `try/except` that fails closed (drops the source) rather than crashing.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
```diff
@@ -262,59 +276,180 @@ def _prepare_model_request(self, request: ModelRequest, *, hook: str) -> ModelRe
...
+        # and it equals the registry's own get_container_file_path. A by-name
+        # fallback would be a confused deputy: DeerFlow lets a custom skill shadow
+        # a same-named public/legacy one (load_skills de-dupes by name, custom
+        # wins), so a thread that read public/foo could bind the *custom* foo's
+        # declared secrets even though the custom skill was never loaded. A path
+        # that no longer resolves (e.g. the skill was moved/removed) simply does
+        # not bind — the safe direction.
+        by_path: dict[str, Skill] = {}
+        for skill in skills:
+            by_path[skill.get_container_file_path(container_root)] = skill
```
willem-bd: **Latent bug: path-normalization asymmetry silently kills in-context binding on non-canonical config.**

- Capture side stores `posixpath.normpath(path)` (`agents/middlewares/skill_context.py:47,116`), so `entry["path"]` is always canonical, e.g. `/mnt/skills/public/foo/SKILL.md`.
- Match side keys `by_path` here (line 420) by `skill.get_container_file_path(container_root)`, which builds raw f-strings (`skills/types.py::get_container_file_path`) from `container_root = storage.get_container_root()` — the config's `container_path` verbatim, no `normpath` (`skills/storage/skill_storage.py:182`).

So an operator setting `skills.container_path: /mnt/skills/` (trailing slash — a natural misconfiguration) gets a `by_path` key of `/mnt/skills//public/foo/SKILL.md` vs `entry["path"]` of `/mnt/skills/public/foo/SKILL.md` → `by_path.get(...)` returns `None` → the skill is skipped → in-context secret binding never fires for the whole install. No error, no log; the feature is silently dead. The default `/mnt/skills` happens to match, so the e2e test passed.

Suggested fix: `posixpath.normpath` the `container_root` (and/or the `entry["path"]`) before matching.

willem-bd (commented):
Review of PR #3938 — bind request-scoped secrets for autonomously-invoked skills (A+).

Three findings, all in `SkillActivationMiddleware`. The two most severe share one root cause: `_SLASH_SECRET_SOURCE_KEY` is read from `runtime.context`, which `build_run_config` (`app/gateway/services.py:318`) merges from the caller's `config.context` via `dict(context_value)` with no `__`-key filtering (`RunCreateRequest.config` is an unfiltered `dict[str, Any]`, `app/gateway/routers/thread_runs.py:63`). A caller can forge that key to (1) bypass the enabled/allowlist/`secrets-autonomous` gates and inject arbitrary caller secrets into the bash subprocess env, and (2) crash the run with a single malformed request. The third finding is a path-normalization asymmetry that silently disables the headline feature on a trailing-slash config.

Details inline. Also examined and rejected (not flagged): the AIMessage failure path leaving a stale `ACTIVE_SECRETS_CONTEXT_KEY` — the run ends before any bash tool runs, so the stale key is unread; the audit-state `previous == audit_state` dedup silencing re-activations — `runtime.context` is request-scoped (rebuilt per run), so `previous` resets and the warning re-fires each run; the default `secrets_autonomous=True` — intended PR behavior; the per-call `load_skills` cost — documented intentional revocation-safety tradeoff, gated behind `request_secrets` + `skill_context`.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
fancyboi999: Confirmed and fixed in cee1757b — thanks, this was a real hole. I reproduced it (forged `__slash_skill_secret_source` + no `/slash` → `ADMIN_TOKEN` injected into `__active_skill_secrets` for a skill that doesn't exist / isn't allowlisted).

**Root cause you nailed:** the slash source was trusted as authoritative — its stored `requirements` were injected directly, with no registry re-validation, on a caller-mergeable context key.

**Fix (defense in depth, both layers):**
1. The slash source now records **only the activated skill's canonical container path** — never its declared secrets. Both the slash and in-context sources resolve the live registry skill by `posixpath.normpath`'d container path (`_resolve_registry_skill`) and bind **only that real, enabled, allowlisted skill's own declared secrets**. A forged path resolves to nothing; a forged path that happens to hit a real skill still only binds that skill's own declarations (and the caller already holds the value it supplied). Slash is exempt from `secrets-autonomous` (the ceremony), but enabled + allowlist are enforced for both paths.
2. `build_run_config` strips caller-supplied `__`-prefixed context keys at the gateway boundary, so the internal channels (`__slash_skill_secret_source`, `__active_skill_secrets`, `__run_journal`) can't be seeded at all. Verified no legitimate caller `__`-context key exists.

I verified each layer **independently** (middleware rejects the forged source even with the gateway strip hypothetically absent) and traced every caller path into `runtime.context` (thread/stateless runs → single `build_run_config` chokepoint; `merge_run_context_overrides` is whitelist-only; IM channels build context server-side; subagents never copy secrets and don't run this middleware). Regression: `test_forged_slash_source_cannot_bypass_gates`, `test_forged_slash_source_ignores_caller_requirements_and_allowlist`, `test_build_run_config_strips_caller_dunder_context_keys`.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
fancyboi999: Fixed in cee1757b. The middleware no longer unpacks caller-provided requirement data at all — `slash_source["requirements"]` is gone; the code reads only `.get("path")` and pulls declarations from the live registry. A malformed source fails closed (binds nothing) instead of raising. Regression `test_malformed_slash_source_does_not_crash` throws list/str/int/None sources, non-dict `path`, and malformed `skill_context` entries at it — no crashes.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
fancyboi999: Fixed in cee1757b. Both the registry keys and the lookup path now go through `posixpath.normpath`, so a `container_path` with a trailing slash (or other non-canonical form) matches the canonical path captured in `skill_context`. Regression `test_trailing_slash_container_root_still_binds` sets `container_root=/mnt/skills/` and confirms binding still fires. (Good catch that the default `/mnt/skills` masked this in the e2e.)

WillemJiang (approved):

Issue #3936, opened 2026-07-04
[feat] BoxLite as a lightweight micro-VM SandboxProvider — open to an integration?


DorianZheng:
### The ask

Would the team be open to a lightweight **micro-VM** `SandboxProvider` for DeerFlow, backed by [BoxLite](https://github.com/boxlite-ai/boxlite)? Happy to prototype and send a PR — wanted to check interest and constraints first.

### Why

A couple of open issues point at the same pain with the default AIO docker sandbox:

- #3439 — cold start is slow and the all-in-one container is resource-heavy; the ask is for a lighter sandbox that DeerFlow can switch to.
- #3213 — each sandbox sits around 500–830 MiB idle; a contributor is already profiling alternative runtimes (OpenSandbox, agent-sandbox, CubeSandbox, gVisor / Kata / Firecracker).

BoxLite is in that micro-VM family, so it's a natural candidate to benchmark against that list.

### What BoxLite is

A **Box** is a hardware-isolated micro-VM (libkrun/KVM on Linux, Hypervisor.framework on macOS) that runs any OCI image unchanged and persists across turns.

- **Daemonless** — embeds as a library, no root, no background service (optional server mode).
- **OCI-native** — you'd keep your own sandbox image; BoxLite just runs it.
- **Own kernel per box** — stronger isolation than a shared-kernel container, lighter than a full VM.
- **Controlled egress** (`allow_net` allow-list) and secret injection (real values never enter the VM).
- Python/Node/Go/Rust/C SDKs, Apache-2.0.

### How it would plug in

DeerFlow already resolves providers dynamically (`sandbox.use: deerflow.community.aio_sandbox:AioSandboxProvider`), so this stays additive — a new `deerflow.community.boxlite:BoxliteProvider` behind the same interface, no change to the default AIO path.

### Open questions / caveats (want your read before building)

- **Host requirements**: micro-VMs need KVM on Linux hosts (nested virt when DeerFlow itself runs inside a cloud VM). Is that acceptable for the deployments you care about, or is a container-only host a hard requirement?
- **Tool-surface parity**: the AIO image bundles browser / Jupyter / code-server / MCP and specific `/mnt/user-data/*` mount semantics. To be a drop-in, a BoxLite provider needs to reproduce those — matching the compat checklist raised in #3213 (bash, file IO, browser, MCP, uploads/artifacts, idle-timeout/orphan cleanup). Which of these are must-haves vs. optional for a first provider?
[...]

Issue #3439, opened 2026-06-08
[feat] aio docker sandbox性能很慢有没考虑其它轻量级sandbox实现方案


cw1427:
### Before you start

- [x] I searched [existing issues](https://github.com/bytedance/deer-flow/issues?q=is%3Aissue) and this is not a duplicate.

### Problem / motivation

在公司内部署了deerflow最新代码使用 aio docker sandbox给试用，都反馈agent启动过程要等好久，特别是冷启动第一次问问题的时候。看了下agent 启动过程主要是要准备sandbox环境执行代码。还有就是aiodocker sandbox吃资源特别大，单台多用户压力很大。又不想上k8s(本身运维就费力)。

建议有没轻量级的sandbox方案， deerflow可以支持方便切换

### Proposed solution

建议有没轻量级的sandbox方案， deerflow可以支持方便切换

### Affected area(s)

Not sure

zwj110610:
我这边实测deepseekv4pro大概在13s左右

zwj110610:
# Why The Baseline Is Slow

The core baseline problem is that the AIO sandbox is started lazily. The backend only starts the container after the model has already decided to call sandbox tools such as `bash`, `read_file`, or `ls`.

The first-question flow is roughly:

```text
User sends message
-> Model does its first reasoning step
-> Model decides to call bash
-> Backend starts the AIO sandbox
-> Wait for the container to become ready
-> Execute pwd
-> Model generates the final answer
```

The main latency sources are:

- AIO container cold start averages about `3.7s`.
- In the baseline, this `3.7s` happens after the model decides to call a tool, so it cannot overlap with the model's first reasoning step.
- The first tool execution adds about `0.8s`.
- Real requests also include model API latency, final-answer generation, and container daemon scheduling variance.

# Proposed Change

Keep the correctness of lazy sandbox acquisition, but start the container earlier for requests that are very likely to use the sandbox. This lets container cold start overlap with the model's first reasoning step.

The new flow is:

```text
User sends message
-> Backend quickly checks whether this looks like a sandbox scenario
-> If matched, backend starts the AIO sandbox in the background
-> Model does its first reasoning step at the same time
-> When the model actually calls bash, the sandbox is likely already ready
-> Reuse the prewarmed sandbox
```

[...]

PR #3940, opened 2026-07-04
feat(sandbox): add BoxLite micro-VM sandbox provider (scaffold)

DorianZheng:
Refs #3936 · relates to #3439, #3213

## Why

**Trigger:** #3439 (the AIO docker sandbox is slow to cold-start and resource-heavy) and #3213 (~500–830 MiB per sandbox; a maintainer invited PRs and a contributor is already profiling alternative runtimes including Firecracker/Kata/gVisor). I opened #3936 to align on approach — this draft makes it concrete.

**Pain:** a lighter, faster-booting sandbox that DeerFlow can switch to via `sandbox.use`, without standing up k8s.

## What changed

A new **optional** community sandbox provider, `deerflow.community.boxlite_sandbox:BoxliteSandboxProvider`, backed by [BoxLite](https://github.com/boxlite-ai/boxlite) — a daemonless, OCI-native micro-VM runtime (libkrun/KVM on Linux, Hypervisor.framework on macOS). Each sandbox is a hardware-isolated VM with its own kernel that runs any OCI image unchanged.

- **Opt-in only** via `sandbox.use`; existing providers untouched, no default-behavior change.
- `execute_command` (bash) is wired end to end; the remaining `Sandbox` methods are defined stubs that raise `NotImplementedError`.
- Async→sync bridge: a private asyncio loop on a daemon thread marshals BoxLite's async SDK via `run_coroutine_threadsafe` — safe under DeerFlow's `asyncio.to_thread` worker pool, and avoids BoxLite's greenlet sync facade (which refuses to run inside an async context).

**Scaffold status — draft for approach review, not production-ready.** Deliberately deferred until the approach is agreed: the file `Sandbox` methods (`read_file`/`write_file`/`download_file`/`update_file`/`list_dir`/`glob`/`grep`, planned via exec + `deerflow.sandbox.search`, mirroring `e2b_sandbox`), warm pooling / idle reaping / mount syncing, the optional `[boxlite]` extra + `uv.lock` update (kept off the shared lockfile for now — install is `pip install boxlite`), and benchmarks vs. the AIO sandbox.

Open questions for maintainers are in #3936: host/KVM acceptability (nested virt when DeerFlow itself runs inside a cloud VM), tool-surface parity with the AIO image, and in-tree vs. external packaging.

## Surface area

- [x] **Sandbox** — new optional execution provider under `community/`
- [ ] **Frontend UI**
- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Skills**
- [ ] **Dependencies** — none added to `pyproject.toml` (deferred, see above)
- [ ] **Default behavior change** — no; opt-in only

## Screenshots / Recording

N/A — backend provider, no UI.

## Validation

- `ruff check .` and `ruff format --check .` — pass against deer-flow's `ruff.toml`.
- `pyproject.toml` and `uv.lock` are untouched, so `uv-lock-check` is not triggered.
- Package is import-safe without `boxlite` installed (lazy import), so unit-test collection is unaffected.
- **Not run:** `make test` end-to-end and a live `execute_command` against a booted box — that needs a KVM/HVF host running the harness, which I couldn't stand up while writing this. Treat the runtime paths as **unverified**; end-to-end validation and benchmarks are exactly what I'd finish once you confirm interest.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Scaffold generated with Claude Code by mirroring the existing `community/e2b_sandbox` provider against the BoxLite SDK; the async-bridge design and the deliberately-narrow scope were directed, and the diff is under review.


backend/packages/harness/deerflow/community/boxlite_sandbox/boxlite_sandbox.py
```diff
@@ -0,0 +1,152 @@
...
+        stderr = result.stderr or ""
+        if stdout and stderr:
+            output = f"{stdout}\n{stderr}"
+        else:
+            output = stdout or stderr
+        if result.exit_code not in (0, None) and not output:
+            output = f"Command exited with code {result.exit_code}"
+        return output if output else "(no output)"
+
+    # ── Not implemented yet (scaffold) ──────────────────────────────────
```
fancyboi999: P2: this provider is exposed as a normal `SandboxProvider`, but every non-bash file API still raises `NotImplementedError`, and those APIs are on DeerFlow's default path rather than being optional extras. Once someone opts into `sandbox.use: deerflow.community.boxlite_sandbox:BoxliteSandboxProvider`, ordinary flows like `read_file` / `write_file` / `ls` / `glob` / `grep` and upload sync can hit these methods immediately (`backend/packages/harness/deerflow/sandbox/tools.py`, `backend/app/gateway/routers/uploads.py`). That turns the provider into a deterministic regression for common agent/tool workflows, even though the PR description frames it as an additive scaffold. I think this needs either a stronger gate that keeps the provider out of normal selection until the file/tool surface is implemented, or a first pass that actually covers the contract DeerFlow already assumes.

fancyboi999 (commented):
One deterministic P2 below. The BoxLite direction itself may be fine, but this scaffold is still exposed as a normal `SandboxProvider`, and the current missing file/tool surface looks reachable from ordinary DeerFlow flows once someone opts into it.

backend/packages/harness/deerflow/community/boxlite_sandbox/boxlite_sandbox.py
DorianZheng: Good catch — fixed by implementing the contract rather than gating.

The 7 file methods no longer raise: `read_file` / `write_file` / `update_file` / `download_file` / `list_dir` / `glob` / `grep` now run as shell commands inside the box (`cat` / `find` / `grep` / chunked `base64`), reusing `deerflow.sandbox.search` and busybox-portable flags so any OCI image works. `download_file` keeps the `/mnt/user-data` prefix + traversal guards, and the provider `mkdir`s the virtual dirs on box start so those paths resolve natively. So once selected, `bash` / `read_file` / `ls` / `glob` / `grep` and upload sync all work instead of regressing.

Verified end-to-end through deer-flow's real `get_sandbox_provider()` against a live box (boxlite 0.9.7): text + append + binary round-trips (all 256 byte values), a 100 KB chunked write, list/glob/grep, the download prefix-guard, missing-file handling, and `execute_command` — all pass.

Also renamed to `BoxliteProvider` / `BoxliteBox` (module `deerflow.community.boxlite`) so it reads as a compute backend rather than "a sandbox".

DorianZheng:
Hi @zhfeng @fancyboi999 all comments fixed, PTAL

WillemJiang:
@DorianZheng Thanks for your contribution. Please take a look at the unit test errors and review comments from Copliot.

DorianZheng:
Thanks for the thorough pass — addressed in the latest commit:

- **grep `literal` bug** — real bug, fixed. grep now gets the raw pattern (`-F` literal / `-E` regex); the `re.escape` was only meant for the Python-side `re.compile` validation, which now runs for the non-literal case only. Added a regression check: `literal=True` matches `a.b.c` literally, `literal=False` treats `.` as a wildcard.
- **env re-resolution** — fixed. Dropped the local `$VAR` resolution; `AppConfig.resolve_env_variables` already resolves recursively and raises on a missing var, so `environment` is used as-is.
- **unit tests** — added `tests/test_boxlite_provider.py` (runs in CI without BoxLite): asserts the actionable `pip install boxlite` ImportError on the lazy import, a clean acquire-failure + idempotent `shutdown()`, and the traversal / `download_file` prefix guards.
- **env-key validation** — added. After updating this branch with `main`, `_validate_extra_env` now exists on the `Sandbox` contract and `local`/`e2b`/`aio` all call it; `execute_command` now calls `_validate_extra_env(env)` too, so a non-POSIX key fails fast with `ValueError` before any exec (covered by a unit test).
- **README vs PR description** — the README is the accurate one (contract fully implemented; verified end-to-end on macOS/HVF, Linux/KVM pending in #3936). Updating the stale PR description to match.

Also fixed the red CI: `provider.py` hardcoded the `/mnt/skills` literal, which a guard test bans outside `deerflow/constants.py` — now imports `DEFAULT_SKILLS_CONTAINER_PATH`.

DorianZheng:
Hi @WillemJiang. Thanks for the review! I have fixed the CI issues. Could you kick off the CI against the newest commit again

WillemJiang (approved):

PR #3942, opened 2026-07-04
fix(web_fetch): add SSRF guard for self-hosted providers

DengY11:
## Summary

This PR hardens self-hosted `web_fetch` providers against SSRF.

DeerFlow already had URL safety validation for the Browserless web tool path: it rejects loopback, private, link-local, cloud metadata, reserved, multicast, and unresolved targets by default, with an explicit `allow_private_addresses: true` opt-in for intentional internal access.

However, later-added self-hosted `web_fetch` providers, especially Crawl4AI, did not reuse that protection. An attacker could potentially guide the agent into calling `web_fetch` on internal or metadata endpoints such as:

```text
http://169.254.169.254/latest/meta-data/
http://localhost:...
http://10.x.x.x/...
http://*.svc.cluster.local/...
```

When Browserless/Crawl4AI runs inside the deployment network, this can turn `web_fetch` into an SSRF primitive and leak sensitive internal information.

## What Changed

- Extracted the URL safety check into `deerflow.community.url_safety`.
- Applied the shared guard to Browserless `web_fetch`.
- Applied the shared guard to Crawl4AI `web_fetch`.
- Kept the existing explicit `allow_private_addresses: true` opt-in for trusted internal targets.
- Updated `config.example.yaml` to document the SSRF guard for Browserless and Crawl4AI `web_fetch`.

## Validation

I started the Gateway with `make gateway` and ran runtime smoke tests against a local fake Browserless/Crawl4AI service.

The smoke test verified that normal public `web_fetch` still reaches the fetch service, metadata URLs are blocked before reaching the service, and `allow_private_addresses: true` still permits intentional private fetches. This confirms the feature remains functional and the new validation logic is effective.

Related Browserless/Crawl4AI tests and formatting/lint checks were also run successfully.


backend/packages/harness/deerflow/community/crawl4ai/tools.py
fancyboi999, line 1: This still leaves one self-hosted `web_fetch` path unprotected: `fastcrw.web_fetch_tool()` is documented as supporting `base_url: http://localhost:3000` for self-hosting, but it still forwards arbitrary URLs straight into `client.scrape(...)` without the new SSRF gate. That means the attack this PR is fixing remains reachable whenever users switch to the fastCRW provider, so the hardening is incomplete for the repo's current self-hosted fetch surface. Please either extend the shared `validate_public_http_url(...)` guard to `backend/packages/harness/deerflow/community/fastcrw/tools.py` as well (and add matching tests), or narrow the PR/docs claim so it does not imply self-hosted providers are now covered broadly.

fancyboi999 (dismissed):
Left one concrete follow-up: the new shared SSRF guard only covers Browserless and Crawl4AI, but `fastcrw.web_fetch_tool()` is still documented/configured for self-hosted use and still forwards arbitrary URLs without equivalent screening. That leaves the same internal-target fetch primitive available through another current `web_fetch` provider, so I don't think the hardening is complete yet.

DengY11:
@fancyboi999 Thanks for the catch. I extended the shared SSRF guard to fastCRW web_fetch as well, including metadata/private/DNS-to-private rejection tests and the explicit allow_private_addresses opt-in path. I also updated `config.example.yaml` to document the fastCRW `allow_private_addresses` option.

WillemJiang (approved):

PR #3945, opened 2026-07-04
feat: add workspace change review for agent runs

18062706139fcz:
## Why

Agent runs can create or modify files in the workspace, but users currently need to inspect artifacts or filesystem state manually to understand what changed. This makes file-producing runs harder to review, especially when an agent edits multiple files or when a maintainer/debugger needs to understand a run after the fact.

This PR adds a run-scoped workspace change summary so users can immediately see which files were created, modified, or deleted, with additions/deletions and a detailed diff panel when needed.

## What Changed

- Record workspace file changes for each agent run.
- Show an inline "Edited N files" card under the assistant message.
- Display changed file paths with per-file `+/-` counts.
- Add a "View changes" action that opens a detailed diff panel.
- Support created, modified, deleted, binary, large, sensitive, and truncated file states.
- Hide sensitive file contents from diffs.
- Keep the summary card lightweight by fetching file metadata without full diffs first.
- Fetch full diffs only when the user opens the review panel.
- Preserve modified-file diffs without retaining full baseline text in memory.
- Add tests for backend diff recording, sensitive-path handling, summary-only API responses, frontend API query behavior, and E2E UI flow.
- Refine the summary card visual scale so it fits better inside the chat timeline.

## Implementation Notes

- Backend captures a pre-run workspace snapshot, compares it with a post-run snapshot, and stores the result as a `workspace_changes` run event.
- The API exposes run-scoped workspace changes through the existing thread/run route.
- The frontend badge requests `include_diff=false` for the lightweight timeline card.
- The side panel requests full diff data only when opened.
- Sensitive paths such as `.env`, keys, tokens, passwords, credentials, and API key files are never exposed as raw diffs.
- Baseline file text is cached temporarily for accurate modified-file diffs, then cleaned up after recording.

## Validation

- `backend`: `pytest tests/test_workspace_changes.py -q`
- `backend`: `ruff format --check .`
- `backend`: `ruff check .`
- `frontend`: `pnpm check`
- `frontend`: workspace changes unit tests
- `frontend`: `tests/e2e/workspace-changes.spec.ts`
- Manual demo with real API: created two files from an agent run and verified the inline change card plus review panel.

image

image


frontend/src/components/workspace/messages/message-list-item.tsx
```diff
@@ -413,6 +417,11 @@ function MessageContent_({
         components={components}
       />
       <CitationSourcesPanel sources={citationSources} />
+      <WorkspaceChangeBadge
+        threadId={threadId}
```
fancyboi999: `buildVisibleHistoryMessages(...)` intentionally carries the same `run_id` onto historical human messages, and `message-list.tsx` passes that `runId` into every `MessageListItem`. Rendering the badge unconditionally here means the same workspace-change summary will show up under the user's message as soon as a run touches files, which contradicts the feature contract in this PR body. Please gate this to assistant messages (or only pass `runId` through on assistant items) before shipping.

fancyboi999 (changes requested):
One deterministic UI regression remains: the workspace-change badge is currently wired into every message item that carries a run id, so it will also render under the user's prompt instead of only under the assistant turn described in the PR.

WillemJiang:
@18062706139fcz  Please fix the lint error and address the review comments.

WillemJiang (approved):

PR #3949, opened 2026-07-05
fix(subagents): surface turn-budget cap as MAX_TURNS_REACHED with partial result (#3875 Phase 2)

hata33:
Phase 2 of #3875. Builds on the Phase 1 subagent `LoopDetectionMiddleware` (#3931).

**What this PR does not do:** it adds **no new limit and no new stop**. The turn budget (`max_turns`, default 150 for `general-purpose`) already existed, and LangGraph already stopped the run on `recursion_limit` by raising `GraphRecursionError`. This PR only changes **how that pre-existing stop is reported**, and stops it from throwing away the work the subagent had already produced.

> Not to be confused with `TIMED_OUT`: that is the **wall-clock** limit (`subagents.timeout_seconds`, 1800s) and is a separate path that already worked. This PR is about the **turn-count** limit (`max_turns`), a different axis.

## Why
When `max_turns` is exhausted, `GraphRecursionError` propagated from `agent.astream` into the generic `except Exception as e:` in `_aexecute`, so the run was reported as `SubagentStatus.FAILED` with `error=str(exception)` and the partial work already streamed into `final_state` was **discarded**. Two consequences:

- The lead agent cannot tell **"subagent broken"** from **"subagent out of budget"**, so it tends to retry blindly and burn the same tokens again.
- The partial work the subagent produced before the cap is lost — the lead gets an empty failure instead of the partial result.

The stop itself was correct; the **misclassification + lost partial result** was the bug. Reported in #3875.

## Root cause
`backend/packages/harness/deerflow/subagents/executor.py::_aexecute` had a single `except Exception as e:` around the `agent.astream` loop that mapped every exception — including `GraphRecursionError` — to `SubagentStatus.FAILED` with no result. The `final_state` captured from the last streamed chunk was never consulted on this path.

## What changed
- **executor.py**: catch `GraphRecursionError` specifically (before the generic handler) and set a **distinct** `SubagentStatus.MAX_TURNS_REACHED` terminal status, recovering the partial result from `final_state` via a new shared `_extract_final_result` / `_stringify_message_content` helper (refactored out of the normal-completion path so both paths render content identically). `MAX_TURNS_REACHED` added to the `SubagentStatus` enum and its `is_terminal` set.
- **status_contract.py**: added `"max_turns_reached"` to `SubagentStatusValue` / `SUBAGENT_STATUS_VALUES`, a `format_subagent_result_message` branch (`Task reached max turns. <cap> Partial result: <partial>`), and extended `make_subagent_additional_kwargs` / `read_subagent_result_metadata` via a `_RESULT_BEARING_STATUSES` set so a capped run carries `subagent_result_brief` + `subagent_result_sha256` (the recovered work, like `completed`) **and** the cap notice on `subagent_error` — the one status that carries both.
- **task_tool.py**: a `MAX_TURNS_REACHED` polling branch that returns the result via the shared `_task_result_command(status="max_turns_reached", result=partial, error=cap)`; `_is_subagent_terminal` includes the new status so cleanup proceeds.
- **delegation_ledger.py**: `_status_guidance` gives the lead explicit options (reuse the partial result / retry tighter / raise per-agent `max_turns`). `extract_delegations` already prefers `result_brief`, so the partial work is what the durable ledger shows.
- **contracts/subagent_status_contract.json**: added `max_turns_reached` to `valid_status_values`.
- **frontend**: `STRUCTURED_STATUS_TO_SUBTASK` maps `max_turns_reached` → `failed` (card collapses, cap notice on `error`).

The distinct status is the load-bearing part: it is what lets the lead distinguish budget-exhaustion from a crash and decide (reuse / retry tighter / raise budget) instead of blindly retrying. The partial-result recovery is the other half.

Externally observable behavior for a normal run is unchanged — only the previously-bare `FAILED`-on-cap path now yields a distinct, partial-result-carrying status.

## Side effect analysis
- **Default values**: unchanged. `general-purpose` `max_turns` stays 150; `timeout_seconds` stays 1800. This only changes how hitting the turn limit is reported, not the limit itself.
- **Backward compatibility**: existing 5 status values behave identically. `max_turns_reached` is purely additive. Old clients that don't know it fall through to the frontend's documented "unknown future status" path.
- **Downstream consumers**: the structured-status contract is the protocol; `read_subagent_result_metadata` and the frontend reader handle the new value. The delegation ledger treats it as terminal (via the derived `TERMINAL_STATUSES = frozenset(SUBAGENT_STATUS_VALUES)`), so a later in-progress write cannot downgrade it.
- **Security boundary**: no new parsing, network, or execution surface. `GraphRecursionError` was already raised and caught; this just routes it to a more specific branch.

## Validation
- `PYTHONPATH=. uv run pytest tests/test_subagent_executor.py tests/test_task_tool_core_logic.py tests/test_delegation_ledger.py tests/test_subagent_status_contract.py -q` → **145 passed**
- broad sweep `pytest tests/ -q -k "subagent or task_tool or delegation or status_contract or thread_state or subtask"` → **409 passed, 3 skipped**
- `uv run ruff check` + `ruff format --check` on changed files → clean
- `cd frontend && pnpm typecheck` → clean; `pnpm test tests/unit/core/tasks/subtask-result.test.ts` → **25 passed**; `eslint` on changed files → clean

## AI assistance
**Tool(s) used:** Claude Code
**How:** AI analyzed #3875, located the root cause (GraphRecursionError falling through to the generic FAILED handler), implemented the distinct status + partial-result recovery across the contract/executor/task_tool/ledger/frontend, and wrote the regression tests. I reviewed every line and am responsible for the change.
- [x] I have read and understand every line of this change — it is not unreviewed AI output.

Note: the first implementation attempt was written against the pre-#3906 status contract (`extract_subagent_status` / prefix-table based); I caught this before pushing, discarded it, and re-implemented against the current `format_subagent_result_message` / structured-metadata contract on a clean `upstream/main` branch.


backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -137,6 +140,75 @@ def try_set_terminal(
             return True
 
 
+def _stringify_message_content(content: Any) -> str:
```
willem-bd: Cleanup — this helper duplicates `deerflow/utils/messages.py::message_content_to_text`, which already normalizes the same three LangChain `content` shapes (bare `str` → list of `str`/`{"text": …}` blocks → `str(content)` fallback). There's also a third near-identical sibling in `utils/llm_text.py::extract_response_text`. This PR rewrote the whole inline block, so it was the natural moment to consolidate onto the existing helper instead of crystallizing a third copy — future content-shape fixes now have to land in three places.

Caveat on a straight swap: the new helper concatenates consecutive raw-string chunks without a separator (then joins across text-block boundaries with `\n`) and returns a `"No text content in response"` sentinel when no text is found, whereas `message_content_to_text` joins every part with `\n` and returns `""`. The new behavior matches the old inline code being replaced, so importing the existing helper would change output for edge inputs (empty/no-text content → sentinel vs `""`; consecutive bare strings → concatenated vs `\n`-joined — atypical for a final AIMessage). Suggested fix: lift this helper into `utils/messages.py` as the canonical version (reconciling the sentinel/join behavior across call sites) and import it here, rather than keeping a private third copy.

hata33: Good catch — consolidated in ac706fc5. Dropped `_stringify_message_content`; `_extract_final_result` now delegates to `utils/messages.py::message_content_to_text`, with the `"No response generated"` sentinel pushed down to the consumer (the canonical helper returns `""` for no-text, matching its other three call sites). Net effect on real `AIMessage` content is unchanged; only the atypical edge inputs you flagged — consecutive bare-string list items (now `\n`-joined, not concatenated) and empty content (now `""` → sentinel at the call site) — move to the canonical behavior every other caller already relies on.

`extract_response_text` is deliberately left as-is: it filters by OpenAI content-block `type` (`{"text","output_text"}`), a different shape with many callers (`channels/manager`, `gateway/routers/suggestions`, `runtime/goal`) and its own dedicated test suite. Unifying it is a separate, riskier refactor outside this PR's scope.

backend/packages/harness/deerflow/tools/builtins/task_tool.py
```diff
@@ -483,6 +483,27 @@ async def task_tool(
...
+                # final answer. ``_task_result_command`` formats a distinct
+                # ``Task reached max turns`` message that carries the partial
+                # result the executor recovered, and stamps ``result_brief`` +
+                # the cap notice on ``subagent_error`` so the delegation ledger
+                # and frontend card keep both. The polling loop emits
+                # ``task_failed`` so any live listener transitions the card
+                # out of running; the structured status is the precise reason.
+                _cache_subagent_usage(tool_call_id, usage, enabled=cache_token_usage)
+                _report_subagent_usage(runtime, result)
+                writer({"type": "task_failed", "task_id": task_id, "error": f"Reached max turns ({config.max_turns})", "usage": usage})
```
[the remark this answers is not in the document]
hata33: Fixed in ac706fc5 — the live `task_failed` event now emits `Reached max_turns={n}`, matching the logger, the structured `error=`, and the executor, so the cap notice is identical across event/log/metadata.

WillemJiang (approved):

PR #3950, opened 2026-07-05
feat: add branching support for assistant turns

18062706139fcz:
## Why

Users often need to explore an alternative follow-up path from a previous assistant response without losing the current conversation. Before this change, they had to manually copy context into a new thread, which is error-prone and loses the relationship to the original conversation.

This PR adds a branch-from-turn workflow for main chat conversations. Users can create a new thread from a completed assistant turn and continue independently from that point.

## What

- Adds backend support for creating a branch from a thread turn.
- Adds a main-chat-only branch action on completed assistant turns.
- Creates the new thread from the selected assistant turn checkpoint.
- Preserves branch provenance metadata for maintainability and debugging.
- Navigates users directly to the newly created branch.
- Keeps the branch title aligned with the source thread title, without adding a `Branch:` prefix.
- Adds i18n strings, tests, and documentation updates.

### Backend

- Adds `POST /api/threads/{thread_id}/branches`.
- Validates that the target message is a visible assistant message.
- Rejects unsupported sidecar threads and stale branch targets.
- Resolves the historical checkpoint for the selected assistant turn.
- Creates a new thread checkpoint from that state.
- Stores branch metadata, including the parent thread id, parent checkpoint id, parent message id, and branch creation time.
- Best-effort copies user data so the branched thread keeps the same workspace context.

### Frontend

- Adds `branchThreadFromTurn` API support and a `useBranchThread` mutation hook.
- Adds a branch action to completed assistant turns in the main message list.
- Uses a branch-specific icon and tooltip for the action.
- Shows a loading/disabled state while the branch is being created.
- Invalidates relevant thread caches and navigates to the new branch after creation.
- Keeps sidecar conversations excluded from the branch action.

## UX

The branch action appears alongside existing assistant-turn actions such as copy and regenerate. It is only available for completed assistant turns in the main conversation, so it does not affect streaming turns or sidecar workflows.

## Tests

- `PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run pytest tests/test_threads_router.py -k branch -q`
- `cd backend && make lint`
- `cd frontend && pnpm check`
- `cd frontend && pnpm test tests/unit/core/threads/api.test.ts`
- `cd frontend && PLAYWRIGHT_BASE_URL=http://127.0.0.1:3130 PORT=3130 pnpm test:e2e tests/e2e/branch-thread.spec.ts --project=chromium`
- `git diff --check`

image


backend/app/gateway/routers/threads.py
```diff
@@ -367,6 +514,88 @@ async def create_thread(body: ThreadCreateRequest, request: Request) -> ThreadRe
...
+            assistant_id=source_record.get("assistant_id"),
+            display_name=display_name,
+            metadata=branch_metadata,
+            **thread_owner_kwargs,
+        )
+    except Exception:
+        logger.exception("Failed to write branch thread_meta for %s", sanitize_log_param(new_thread_id))
+        raise HTTPException(status_code=500, detail="Failed to create branch") from None
+
+    workspace_clone_mode = await _copy_branch_user_data(thread_id, new_thread_id)
```
fancyboi999: P2: this breaks the PR's own "branch from a historical assistant turn" contract for any thread that has mutated workspace after that turn. `checkpoint_tuple` is correctly resolved to the selected assistant turn, but `_copy_branch_user_data(thread_id, new_thread_id)` always clones the *current* thread user-data tree, not the state that existed at `parent_checkpoint_id`. I verified this in an isolated review worktree by seeding checkpoints `[human-1, ai-1]` and `[human-1, ai-1, human-2, ai-2]`, creating `workspace/future.txt` after the later checkpoint existed, and then branching from `ai-1`: the new branch gets `parent_checkpoint_id='0001'` but still contains `future.txt`. That means a branch can show the old message history while exposing files that only appeared in a later timeline, which is a real behavior mismatch for coding / artifact workflows. At minimum this needs a regression test for branching from a non-latest turn plus either checkpoint-scoped workspace restore or a narrower product contract that stops claiming historical-turn state restoration.

fancyboi999 (commented):
One deterministic P2 remains: historical-turn branching currently restores the selected checkpoint's messages but still clones the latest thread workspace, so a branch from an older assistant turn can inherit files that only exist in a later timeline.

WillemJiang:
@18062706139fcz  please fix the lint error.

WillemJiang (approved):

Issue #123, opened 2025-05-13
Feature Request: OpenMemory


zackarychapple:
I discovered [OpenMemory](https://github.com/mem0ai/mem0) and its quite interesting for sharing MCP context and updating a persistent memory that can be used across MCPs. Is there any possibility to integrate it with Deer Flow or have something similar?

fancyboi999:
Closing as completed by #4528, which added an opt-in mem0 HTTP memory backend for hosted and self-hosted mem0 deployments.

PR #3951, opened 2026-07-05
feat(sandbox): warm pool for boxlite

Vanzeren:
<!-- Reference a related issue with #123. Use Fixes / Closes / Resolves to
     auto-close it on merge. Delete this line if the PR doesn't reference an issue. -->

## Why

<!-- Why are you opening this PR? Cover two things:
       - The trigger — what made you write this? A bug you hit, a feature you need,
         tech debt, or a prod issue?
       - The pain being addressed — user-facing problem, or what it unblocks.
     For non-trivial features, please open an issue/discussion first to align on
     scope before writing code. -->
BoxLite sandboxes were destroyed after every agent turn because `SandboxMiddleware.after_agent()`    
 calls `release()` after each run. That made BoxLite pay a cold micro-VM startup cost again on the next 
 turn, unlike AIO sandbox which already has warm-pool behavior.                                         
                                                                                                        
   This PR gives BoxLite the same production lifecycle shape: keep released VMs warm for the same       
 user/thread, reclaim them on the next acquire, and bound resource use with replica and idle-timeout    
 controls.

## What changed

- BoxLite sandboxes now use deterministic per-`user_id:thread_id` sandbox IDs.                       
- Released BoxLite VMs are parked in an in-process warm pool instead of being destroyed immediately. 
- The same user/thread can reclaim its warm VM on the next acquire.                                  
- Different users/threads cannot reclaim each other’s VMs.                                           
- Warm VMs are health-checked before reuse; dead warm VMs are evicted and recreated.                 
- `sandbox.replicas` now caps active + warm BoxLite VMs per gateway process.                         
- `sandbox.idle_timeout` controls when idle warm BoxLite VMs are stopped; `0` disables idle reaping. 
- Shutdown/reset now closes active and warm BoxLite VMs safely.                                      
- Added BoxLite SDK dependency and documentation/config examples for warm-pool settings. 

## Surface area

- [x] **Dependencies** — new/upgraded entry in `backend/pyproject.toml` or `frontend/package.json` (say what it buys us)


## Screenshots / Recording

## Validation

Ran:                                                                                                 
                                                                                                        
   ```bash                                                                                              
   cd backend                                                                                           
   PYTHONPATH=. uv run pytest tests/test_boxlite_provider.py -v                                         
   make lint
``` 
Results:                                                                                               
                                                                                                        
 - tests/test_boxlite_provider.py -v → 21 passed                                                        
 - make lint → ruff check passed; ruff format --check passed                                            
 - uv run --project backend python scripts/scan_changed_blocking_io.py --base main → no blocking-IO     
 candidates introduced                                                                                  
                                                                                                        
 Full backend test suite was not run in the final verification pass. 
## AI assistance

**Tool(s) used:** Codex / Oh My Pi coding harness, subagents

**How you used it:** AI helped draft the implementation plan, implement BoxLite warm-pool lifecycle        
 changes, add tests, respond to review feedback, and update docs/config examples. The final branch was  
 verified with focused tests, lint, and blocking-IO scan.

### CODE REVIEW

File: backend/packages/harness/deerflow/community/boxlite/provider.py [hunk 1]
```diff
@@ -159,28 +275,101 @@ def get(self, sandbox_id: str) -> Sandbox | None:
...
+        return sandbox_id
 
     def reset(self) -> None:
         with self._lock:
+            active = list(self._boxes.values())
+            warm = [box for box, _ in self._warm_pool.values()]
             self._boxes.clear()
+            self._warm_pool.clear()
             self._thread_boxes.clear()
+            self._acquire_locks.clear()
```
fancyboi999: P2: `reset()` is supposed to drop provider-owned state so `reset_sandbox_provider()` can swap in a fresh provider cleanly, but this implementation only clears the box maps. The idle reaper thread and private event-loop thread started in `__init__` keep running after `reset()` until a later `shutdown()`. I verified that directly on this branch: after `BoxliteProvider.reset()`, both `_idle_checker_thread.is_alive()` and `_loop._thread.is_alive()` stay `True`, unlike after `shutdown()`. That means repeated reset/recreate cycles leak background threads and leave the old provider's housekeeping running against a supposedly discarded instance. I think `reset()` needs to stop the reaper and close the loop too, or delegate to the same cleanup path as `shutdown()`.

fancyboi999 (commented):
One deterministic P2 below. The warm-pool lifecycle itself makes sense, but `reset()` currently leaves provider-owned background resources alive after the singleton is discarded, which breaks DeerFlow's sandbox-provider reset contract.

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -91,14 +94,38 @@ class BoxliteProvider(SandboxProvider):
...
+    DEFAULT_REPLICAS = 3
+
+    @staticmethod
+    def _sandbox_id(thread_id: str, user_id: str) -> str:
+        """Deterministic sandbox ID from user/thread scope.
+
+        Includes user_id so a box created for one user's bucket cannot be
+        reclaimed by another user's thread with the same thread_id.
+        """
+        return hashlib.sha256(f"{user_id}:{thread_id}".encode()).hexdigest()[:8]
```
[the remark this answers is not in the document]
Vanzeren: It is the same way to define deterministic sandbox id as aio_sandbox.

backend/packages/harness/pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
     "tiktoken>=0.8.0",
     "ddgs>=9.10.0",
     "duckdb>=1.4.4",
+    "boxlite>=0.9.7",
     "langchain-google-genai>=4.2.1",
```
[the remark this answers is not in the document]
Vanzeren: Fixed by making `boxlite` an optional `deerflow-harness[boxlite]` extra instead of a core dependency.
  The lazy-import fallback remains in place, and the README now documents the optional install plus
  supported-host limitation.

backend/packages/harness/deerflow/community/boxlite/README.md
```diff
@@ -16,13 +16,14 @@ sandbox:
...
+  idle_timeout: 600        # warm VM idle seconds before stop; 0 disables reaping
   environment:              # injected into every command
     PYTHONUNBUFFERED: "1"
 ```
 
-```bash
-pip install boxlite   # an optional `[boxlite]` extra + uv.lock update will follow once the approach lands
-```
+`boxlite` is bundled as a `deerflow-harness` dependency; no separate install
+step is needed once the BoxLite provider is selected.
```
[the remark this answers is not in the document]
Vanzeren: Updated the README to clarify that BoxLite is an optional `deerflow-harness[boxlite]` extra, not part
  of the default install, and that it only works on BoxLite-supported host platforms/architectures.
  Unsupported hosts such as Windows should use another sandbox provider or run DeerFlow from a supported Linux/macOS environment.

backend/tests/test_boxlite_provider.py
```diff
@@ -70,3 +116,468 @@ def _fail_run(_coro: object) -> None:
...
+    # Use a very short check interval so the reaper runs quickly
+    monkeypatch.setattr(BoxliteProvider, "IDLE_CHECK_INTERVAL", 0.1)
+    provider = BoxliteProvider()
+    provider._loop.run = _fake_run
+
+    # Acquire and release a box into the warm pool
+    sid = provider.acquire("thread-1", user_id="u1")
+    box, _ = provider._warm_pool.get(sid, (None, None)) if sid in provider._warm_pool else (None, None)
+    provider.release(sid)
+
```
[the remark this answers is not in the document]
Vanzeren: Fixed. Removed the unused `box` assignment before `release()`. The test still covers the intended
  warm-pool behavior, and `tests/test_boxlite_provider.py` passes.

backend/packages/harness/deerflow/community/boxlite/provider.py
Vanzeren: Fixed by delegating `reset()` to `shutdown()`, so it now stops the idle reaper, closes the private
  event loop, and clears provider-owned state through the same cleanup path.

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -159,28 +276,89 @@ def get(self, sandbox_id: str) -> Sandbox | None:
...
+        Returns sandbox_id on success, None if not found or dead.
+        """
+        with self._lock:
+            if sandbox_id not in self._warm_pool:
+                return None
+            box, _ = self._warm_pool[sandbox_id]
+
+        # Health check: run a simple command to verify the VM is alive
+        try:
+            result = box.execute_command("echo ok")
```
willem-bd: The warm-VM health check here blocks indefinitely with no timeout, and it runs while `acquire()` holds the per-thread `acquire_lock` — so a hung warm VM pins every subsequent turn for that `(user_id, thread_id)`.

Tracing the timeout layers:
- This call passes no `timeout`.
- `BoxliteBox.execute_command` (box.py:135) forwards `timeout` to `box.exec(...)` but calls `self._run(...)` **without** forwarding it.
- `_EventLoopThread.run` defaults to `timeout=None` → `run_coroutine_threadsafe(...).result(None)` blocks forever.

Failure path: a warm micro-VM whose runtime is hung (OOM-killed process with a live IPC socket, unresponsive OCI runtime, host pause) → `execute_command("echo ok")` never returns → the calling `acquire()` holds `acquire_lock` for that user/thread → every later turn queues on the same lock → agent turn hangs. The idle reaper won't touch the box for ~`idle_timeout` (600s default) and, even then, closing the box on the private loop doesn't cancel the in-flight exec future.

Fix: pass an explicit `timeout=` here (e.g. `timeout=5`) **and** forward `timeout` through `_EventLoopThread.run` / `execute_command` so `.result()` is bounded at both layers, not just at the SDK exec layer.

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -159,28 +276,89 @@ def get(self, sandbox_id: str) -> Sandbox | None:
...
+            warm_entry = self._warm_pool.pop(sandbox_id, None)
+            if warm_entry is None:
+                return None  # Raced with another thread
+            box, _ = warm_entry
+            self._boxes[sandbox_id] = box
+
+        logger.info("Reclaimed warm-pool box %s", sandbox_id)
+        return sandbox_id
+
+    def reset(self) -> None:
```
willem-bd, line 442: Routing `reset()` to `shutdown()` breaks the `reset_sandbox_provider()` contract. `sandbox_provider.py:115-129` documents `reset` as clearing the cached instance *"without calling shutdown"* — distinct from `shutdown_sandbox_provider()`, which is the teardown path — and says providers override `reset()` to *"clear any module-level state they keep alive across instances ... Without it, config/mount changes would not take effect on the next acquire()"*. `AioSandboxProvider` doesn't override `reset()`; `LocalSandboxProvider` keeps `reset()` (lightweight clear) and `shutdown()` separate.

Here `reset()` closes all VMs, stops the idle reaper, and closes the private event loop, leaving the instance permanently unusable (`_shutdown_called=True`, loop closed). The documented "switch config and re-acquire" use of `reset_sandbox_provider()` would then crash on the next acquire (`self._loop.run` on a closed loop → `RuntimeError`). It's safe today only because every caller discards the instance after `reset()`.

If teardown is genuinely desired here, the cleaner shape is the `LocalSandboxProvider` inversion — `shutdown()` calls a shared `_teardown()`, `reset()` only clears in-memory state (`_boxes`/`_warm_pool`/`_thread_boxes`/`_acquire_locks`) without `box.close()` and without stopping the loop — so reset stays the lightweight "clear cached state" operation the contract describes.

backend/tests/test_harness_packaging.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import annotations
+
+import tomllib
+from pathlib import Path
+
+
+def test_boxlite_is_optional_harness_dependency() -> None:
+    """BoxLite should not make core harness installs platform-dependent."""
+    pyproject = tomllib.loads(Path("packages/harness/pyproject.toml").read_text(encoding="utf-8"))
```
willem-bd: This test resolves `pyproject.toml` via a cwd-relative path, so it only passes when pytest is invoked from `backend/`. Run from an IDE, from repo root, or `pytest backend/tests/test_harness_packaging.py` directly → `FileNotFoundError`.

The existing sibling test `tests/test_checkpointer.py` solves this with an `__file__`-relative path:

```python
pyproject_path = Path(__file__).resolve().parents[1] / "packages" / "harness" / "pyproject.toml"
```

Worth matching here so the test is cwd-independent like the rest of the suite.

backend/uv.lock
```diff
@@ -4264,8 +4279,8 @@ name = "standard-aifc"
 version = "3.13.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "audioop-lts", marker = "python_full_version >= '3.13'" },
-    { name = "standard-chunk", marker = "python_full_version >= '3.13'" },
+    { name = "audioop-lts" },
```
willem-bd: This `standard-aifc` marker change looks unrelated to the `boxlite` extra and is probably lockfile drift from regenerating against a different interpreter. The `python_full_version >= '3.13'` marker on `audioop-lts`/`standard-chunk` was dropped even though nothing in this PR's `pyproject.toml` changes touch `standard-aifc`.

Cost: the lockfile diff misrepresents the scope of the PR, and on Python 3.12 installs it now pulls `audioop-lts`/`standard-chunk` that were previously gated out. If unintentional, revert this hunk and regenerate with a minimal `uv lock` scoped to the boxlite extra (e.g. `uv lock --upgrade-package boxlite`).

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -91,14 +94,38 @@ class BoxliteProvider(SandboxProvider):
     uses_thread_data_mounts = False
     needs_upload_permission_adjustment = True
 
+    # ── Warm pool constants (mirrors AioSandboxProvider) ─────────────────
+
+    DEFAULT_IDLE_TIMEOUT = 600
```
willem-bd: The warm-pool / idle-reaper / replica-enforcement pattern in this provider is a near-verbatim copy of `AioSandboxProvider` (community/aio_sandbox/aio_sandbox_provider.py): the `DEFAULT_IDLE_TIMEOUT=600` / `IDLE_CHECK_INTERVAL=60` / `DEFAULT_REPLICAS=3` constants, `_replica_count()`, `_evict_oldest_warm()` (same `min(..., key=lambda sid: ...[1])` shape), the idle-reaper thread management, and the `_create_box` replica-cap warning log are all line-for-line duplicates — and the boxlite copy has already drifted (it omits AIO's `_log_replicas_soft_cap` helper, inlining the warning instead).

The storage types genuinely differ (AIO carries `SandboxInfo` for cross-process backend discovery; boxlite carries a loop-affine `BoxliteBox`), so a full merge isn't free. But the replica-count + idle-reaper skeleton is extractable into a shared base/mixin parameterised by the warm-entry type and a `_destroy_warm_entry(entry)` hook, leaving each provider to specialise only entry construction, health-check, and destroy. Worth doing now while both copies are in sync, before AIO's next eviction/reaping edge-case hardening has to be ported in two places.

backend/packages/harness/deerflow/community/boxlite/provider.py
Vanzeren: Old ```reset()``` → ```shutdown()``` already breaks the lightweight-clear contract by closing VMs and stopping   
the loop. But a pure memory-state clear (```_boxes.clear()``` etc.) would be worse: it orphans the running 
BoxLite VMs. ```_reconcile_orphans()``` on the next provider construction would re-adopt them into the     
warm pool — but that's a cleanup path, not a correctness guarantee. The current ```reset()``` parks active 
 boxes into ```_warm_pool``` without closing them, keeping them visible to this instance's idle reaper and  
atexit shutdown. No orphans, no leaked resources, and the next ```acquire()``` can still reclaim from the  
warm pool. That's the motivation behind the new design.

WillemJiang (approved):

PR #3964, opened 2026-07-06
feat(agent): support config-declared lead middlewares

### LINKED ISSUE #3923: [feat]  Extension points for downstream integrators: config-declared middlewares, SSE stream wrapper, and run-level model override hooks
### Before you start

### Problem / motivation

 We run a production-grade, correctness-critical enterprise assistant on top of a deer-flow fork. To integrate our domain layer we currently have to patch deer-flow source files directly (anchor-based text insertion via a replay script) in three places:

    1. lead_agent/agent.py — register custom AgentMiddleware instances (domain guards: deterministic output validation/repair, tool-call gating, latency stamping, an FSM dispatch middleware) into the middleware chain built by build_middlewares().
    2. gateway/services.py — apply a deterministic, intent-based model override before a run is created (e.g. route greetings/read-only turns to a fast model when the user did not explicitly pick one).
    3. gateway/routers/thread_runs.py — wrap the SSE generator returned by sse_consumer(...) to capture request-level observability (TTFT / total latency / frame count) without altering the stream.

    This works, but source patching is fragile in practice:

    - Anchors drift whenever upstream refactors these files; every upgrade requires manual re-verification.
    - We hit a real incident where a hand-edited block plus a re-applied patch produced double middleware registration (guards executed twice; risk of graph node-name collisions).
    - It makes the integration impossible to share as a clean, copyable pattern for other teams.

我们在企业场景中基于 deer-flow 集成了一组领域守卫中间件,目前只能用锚点方式直接改源码(agent.py 注册中间件 / services.py 模型覆写 / thread_runs.py 包装 SSE 流),升级易碎且踩过双注册事故。建议增加 config.yaml 驱动的扩展点(中间件列表、SSE 包装钩子、run 级模型覆写钩子),importlib 动态加载,缺包 INFO 跳过、真错 ERROR 带栈,完全向后兼容。我们愿意提交实现 PR。

### Proposed solution

 A small, backward-compatible extension API, driven by config.yaml, so integrations become declarative instead of patched:

    extensions:
      # Loaded and appended by build_middlewares(), in order, after built-in middlewares
      middlewares:
        - my_pkg.middlewares:MyDomainGuardMiddleware
        - my_pkg.middlewares:MyLatencyMiddleware
      # Optional: wraps the SSE generator at the stream endpoints (pass-through signature)
      sse_wrapper: my_pkg.observability:traced_stream     # (gen, record, thread_id) -> async generator
      # Optional: called at run creation when the client did not specify a model
      run_model_override: my_pkg.routing:model_override    # (last_user_text) -> {"model_name":..., "thinking_enabled":...} | {}

    Implementation sketch (happy to contribute this as a PR):

    - Resolve entries via importlib ("module.path:attr" form).
    - Error semantics: ModuleNotFoundError → log at INFO and skip (extension package absent is a legal deployment); any other exception → log at ERROR with full traceback (a broken extension should be loud, never a silent downgrade to "running without guards").
    - No behavior change when the extensions section is absent.

### Affected area(s)

Config / setup

### Alternatives considered

- Anchor-based source patching (current approach): fragile across upgrades; caused a double-registration incident; not shareable as a pattern.
- Subclassing / re-exporting the gateway app: heavier to maintain and still requires forking internal assembly functions.
- Python entry-points based plugin discovery: also viable, but config-declared loading is simpler, explicit, and easier to audit in enterprise settings.

### Additional context

 - The middleware chain in create_agent (LangChain AgentMiddleware with before_model / after_model / wrap_model_call / jump_to) is already an excellent extension surface — the only missing piece is a supported way to register into it without editing source.
    - Our use case: deterministic guard middlewares for a weak-model, numbers-critical domain (tool-call parameter backfill/merging/circuit-breaking, output validate-and-repair, per-request TTFT capture written into persisted state). All of it lives cleanly in an external package today; only the wiring requires patches.
    - We're glad to submit a PR implementing the three hooks above if maintainers agree on the shape.

ZenCGL:
Fixes #3923

## Why

The lead-agent middleware chain is currently only extensible by changing DeerFlow source code or passing programmatic `custom_middlewares` from embedded callers. That makes downstream integrations harder to maintain when they need domain-specific guardrails, tool-call governance, request stamping, or observability hooks.

This PR adds a config-declared extension point so deployments can add lead-agent `AgentMiddleware` classes without patching `lead_agent/agent.py` or carrying a forked runtime chain.

## What changed

- `config.yaml` now supports `extensions.middlewares`, a list of zero-argument `AgentMiddleware` class paths in `module.path:ClassName` format.
- Configured middlewares are loaded into the lead-agent runtime after built-in and programmatic custom middlewares, and before the safety/clarification tail.
- Missing optional extension packages are skipped, so shared configs can work across deployments with different optional integrations.
- Invalid middleware paths, invalid classes, and broken extension modules fail loudly during agent creation.
- The existing tail invariant is preserved: `ClarificationMiddleware` remains last.
- `config.example.yaml`, README, and backend agent docs now describe the extension point.

## Surface area

## Screenshots / Recording

N/A. This PR does not change frontend UI.

## Bug fix verification

N/A. This PR adds an extension point rather than fixing a user-facing bug.

## Validation

```bash
cd backend && PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 .venv/bin/python -m pytest tests/test_lead_agent_model_resolution.py tests/test_config_version.py -q
```
Result:
27 passed
```bash
cd backend && PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 .venv/bin/python -m pytest tests/test_lead_agent_model_resolution.py tests/test_tool_error_handling_middleware.py -k 
'configured_extension_middlewares or invalid_configured_extension_middleware or runtime_middlewares or subagent_runtime or lead_runtime' -q
```
Result:
15 passed
```bash
cd backend && .venv/bin/python -m ruff check packages/harness/deerflow/agents/lead_agent/agent.py packages/harness/deerflow/config/extensions_config.py tests/test_lead_agent_model_resolution.py
```
Result:
All checks passed!

## AI assistance

Tool(s) used: Codex
How you used it: Used for implementation assistance, test drafting, and validation command iteration. I reviewed the diff and adjusted the final behavior, docs, and tests before opening the PR.

I've read and understand every line of this change and take responsibility for it — it's not unreviewed AI output.


config.example.yaml
```diff
@@ -30,6 +30,20 @@ logging:
...
 
+# ============================================================================
+# Agent Extensions
+# ============================================================================
+# Optional lead-agent AgentMiddleware classes loaded into the runtime middleware
+# chain after built-in lead middlewares and any programmatic custom middlewares,
+# but before the safety/clarification tail. Missing optional extension packages
+# are skipped; invalid classes or broken modules fail loudly at agent creation.
+extensions:
+  middlewares: []
```
willem-bd: **Finding 1 — this config.yaml block is silently dropped on load (the feature is inert as documented).**

`AppConfig.from_file()` runs `config_data["extensions"] = ExtensionsConfig.from_file().model_dump()` (app_config.py:307-308), and `ExtensionsConfig.from_file()` reads **only** `extensions_config.json`, unconditionally overwriting whatever the user puts under `extensions:` in config.yaml. The README and backend/AGENTS.md item 27 both point operators to this block, but an operator who follows them gets zero middlewares loaded — no error, no warning.

The added tests pass only because they construct `AppConfig` directly (`app_config.extensions = ExtensionsConfig(middlewares=[...])`) and bypass `from_file()`, so CI never catches this.

Fix: either move `middlewares` off `ExtensionsConfig` onto a config.yaml-backed model, merge instead of overwrite at app_config.py:308, or move the example/docs into `extensions_config.json` (and add a `middlewares` entry to `extensions_config.example.json`).

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -79,6 +79,10 @@ class SkillStateConfig(BaseModel):
 class ExtensionsConfig(BaseModel):
     """Unified configuration for MCP servers and skills."""
 
+    middlewares: list[str] = Field(
```
willem-bd, line 135: **Finding 2 — toggling any public skill silently deletes `middlewares` from extensions_config.json.**

Once this field is live, `PUT /api/skills/{name}` wipes it: the PUBLIC-skill write path (`app/gateway/routers/skills.py` ~line 375) and the legacy fallback rebuild `config_data = {"mcpServers": ..., "skills": ...}` and `json.dump` it, dropping every other top-level key — including `middlewares`. So enabling/disabling any public skill via the UI unloads all configured extension middlewares until the next manual edit.

The MCP router (`mcp.py`) preserves unknown top-level keys via `raw_other_keys`, so these two skills write paths are inconsistent with it, and `DeerFlowClient.update_skill_status` (`client.py` ~line 1177) has the same lossy rebuild.

Fix: include `middlewares` in the rebuilt dict (e.g. round-trip through `ExtensionsConfig.model_dump()`), or reuse the MCP router's key-preservation pattern.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -391,6 +450,10 @@ def build_middlewares(
     if custom_middlewares:
         middlewares.extend(custom_middlewares)
 
+    configured_middlewares = _load_configured_extension_middlewares(resolved_app_config)
```
willem-bd: **Finding 3 — configured middlewares apply to the lead agent only; subagents silently bypass them.**

Subagents are built via `build_subagent_runtime_middlewares` (`subagents/executor.py`), which never calls `_load_configured_extension_middlewares`. A `general-purpose` subagent (all tools except `task`, up to 150 turns) forwards the same conversation history to the model with **none** of the operator-declared guardrails. The PR pitches the feature as "domain guardrails / tool-call governance," so a guardrail that silently doesn't apply to delegated work is a false sense of security.

If the feature is intentionally lead-only, the README/AGENTS.md "guardrail" framing should say so explicitly; otherwise wire the loader into the subagent builder too.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -256,6 +258,63 @@ def _create_todo_list_middleware(is_plan_mode: bool) -> TodoMiddleware | None:
...
 
 
+def _is_missing_declared_extension_module(exc: ModuleNotFoundError, module_path: str) -> bool:
+    missing_name = getattr(exc, "name", None)
+    if not missing_name:
+        return False
+    return module_path == missing_name or module_path.startswith(f"{missing_name}.")
+
+
+def _load_configured_extension_middlewares(app_config: AppConfig) -> list[AgentMiddleware]:
```
willem-bd: **Finding 4 — this reimplements the project's canonical plugin loader.**

`deerflow.reflection.resolvers.resolve_class(class_path, base_class)` (`reflection/resolvers.py:73`) does exactly the parse-`module:attr` → import → validate-subclass contract, and is already used by `models/factory.py`, `sandbox/sandbox_provider.py`, `skills/storage/__init__.py`, `tools/tools.py`, `mcp/tools.py`, and the guardrails middleware.

Composing `resolve_class(middleware_path, AgentMiddleware)` + a small instantiation wrapper would replace ~50 lines here and would **inherit `resolve_class`'s actionable install hint** (`uv add <pkg>`) on `ModuleNotFoundError` — which this hand-rolled version replaces with a bare "module is not installed" skip.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -256,6 +258,63 @@ def _create_todo_list_middleware(is_plan_mode: bool) -> TodoMiddleware | None:
     return TodoMiddleware(system_prompt=system_prompt, tool_description=tool_description)
 
 
+def _is_missing_declared_extension_module(exc: ModuleNotFoundError, module_path: str) -> bool:
```
willem-bd: **Finding 5 — this predicate misclassifies config typos as "optional package not installed" and skips them silently.**

For a submodule-name typo like `my_company.middlewares_typo:DomainGuard`, `import_module` raises `ModuleNotFoundError(name="my_company.middlewares_typo")`, and `module_path == missing_name` returns `True` → silent info-level `continue`. That contradicts the docstring's "invalid classes fail loudly" promise, and an operator cannot tell a typo from a deliberately-absent optional package.

The same swallow happens when an installed module transitively imports a missing dep that shares the module's own prefix (e.g. `foo.bar` importing `foo.bar.qux`).

Worth tightening the skip condition to only the declared top-level package, and/or surfacing these at warning level rather than info.

willem-bd (commented):
Review of the config-declared lead-middlewares feature. The middleware-chain ordering invariants are correctly preserved and test-pinned — these comments focus on the config plumbing and reach of the feature.

1. `extensions.middlewares` in config.yaml is silently dropped on load (the feature is inert as documented).
2. Toggling any public skill wipes `middlewares` from extensions_config.json.
3. Configured middlewares apply to the lead agent only; subagents bypass them.
4. The loader reimplements `deerflow.reflection.resolve_class`.
5. `_is_missing_declared_extension_module` misclassifies config typos as 'optional package not installed'.

config.example.yaml
ZenCGL: 1. extensions.middlewares in config.yaml was overwritten
Fixed. AppConfig.from_file() now loads extensions_config.json first, then merges any explicitly declared extensions fields from config.yaml, with config.yaml taking precedence. This makes config.yaml -> extensions.middlewares active instead of inert.
Added tests:
tests/test_app_config_reload.py
test_app_config_preserves_config_yaml_extension_middlewares
test_app_config_loads_extension_middlewares_from_extensions_config

backend/packages/harness/deerflow/config/extensions_config.py
ZenCGL: Fixed. ExtensionsConfig now has to_file_dict(), and both the skill toggle path and client write-back path serialize from the full extension config instead of manually rebuilding only mcpServers / skills. This preserves middlewares and other top-level extension keys.
Covered paths:
backend/app/gateway/routers/skills.py
backend/packages/harness/deerflow/client.py

backend/packages/harness/deerflow/agents/lead_agent/agent.py
ZenCGL: Fixed. Configured middleware loading was extracted into a shared helper:
deerflow.agents.middlewares.configured_extensions.load_configured_extension_middlewares
Both the lead agent and subagent runtime middleware builders now load the same extensions.middlewares.
Added test:
tests/test_tool_error_handling_middleware.py::test_subagent_runtime_middlewares_inject_configured_extension_middlewares

backend/packages/harness/deerflow/agents/lead_agent/agent.py
ZenCGL: Fixed. The custom import / getattr / subclass validation logic was removed. The configured middleware loader now reuses deerflow.reflection.resolve_class(middleware_path, AgentMiddleware), matching the project’s other extension loading paths.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
ZenCGL: Fixed. Missing packages, misspelled modules, invalid classes, and broken modules now fail loudly during agent creation instead of being silently skipped. The error handling now comes from the shared resolve_class path.
Added test:
tests/test_lead_agent_model_resolution.py::test_build_middlewares_rejects_missing_configured_extension_module

WillemJiang:
@ZenCGL  please resolve the conflict with the main branch.

ZenCGL:
> @ZenCGL please resolve the conflict with the main branch.
I have resolved the conflict with the main branch

backend/packages/harness/deerflow/config/app_config.py
```diff
@@ -319,9 +319,16 @@ def from_file(cls, config_path: str | None = None) -> Self:
...
+        # Load extensions config separately (it's in a different file), while
+        # preserving any config.yaml-backed extension fields. config.yaml wins
+        # when it explicitly declares a field because those values are part of
+        # the main AppConfig hot-reload contract.
+        yaml_extensions = config_data.get("extensions")
         extensions_config = ExtensionsConfig.from_file()
-        config_data["extensions"] = extensions_config.model_dump()
+        extensions_data = extensions_config.model_dump(by_alias=True)
+        if isinstance(yaml_extensions, Mapping):
+            extensions_data.update(yaml_extensions)
```
willem-bd: **`config.yaml` `extensions.middlewares: []` default silently clobbers `extensions_config.json` middlewares**

`extensions_data.update(yaml_extensions)` makes config.yaml win unconditionally - including the `extensions.middlewares: []` that `config.example.yaml` now ships with. So a deployment that keeps that shipped default and instead configures middlewares in `extensions_config.json` (the file this PR also adds the field to, and which the API round-trips via `to_file_dict()`) will silently have those JSON middlewares ignored. The JSON path only takes effect if the whole `extensions:` block is removed from `config.yaml`.

The two new tests cover the sources in isolation - yaml non-empty wins (`test_app_config_preserves_config_yaml_extension_middlewares`), and json is loaded when `config.yaml` has *no* `extensions` section (`test_app_config_loads_extension_middlewares_from_extensions_config`) - but neither covers the realistic shipped-default case: yaml `[]` clobbering a non-empty json list.

Fix options (one of):
- ship the `extensions.middlewares` example **commented out** in `config.example.yaml` so `yaml_extensions` is absent unless opted in;
- skip empty/default values in the merge (only override when the yaml list is non-empty);
- document `config.yaml` as the sole effective source and drop `middlewares` from `extensions_config.example.json`.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -321,6 +321,10 @@ def build_subagent_runtime_middlewares(
 
         assert_mcp_routing_before_deferred_filter(middlewares)
 
+    from deerflow.agents.middlewares.configured_extensions import load_configured_extension_middlewares
+
+    middlewares.extend(load_configured_extension_middlewares(app_config))
```
willem-bd: **Ordering inconsistency vs the lead chain**

In `build_middlewares` (lead), configured extensions are appended *after* `LoopDetectionMiddleware` / `TokenBudgetMiddleware`. Here in the subagent they're appended *before* them - the loop/token guards are added further down in this function. The `backend/AGENTS.md` entry says configured extensions run "after built-in runtime middlewares", which holds for the lead but not for the subagent, where `LoopDetection`/`TokenBudget` are built-in runtime middlewares that fire *after* the configured ones.

Harmless for an observability/stamping middleware, but a governance middleware that expects to see post-cap state would be surprised. Either move this `extend()` to after `TokenBudgetMiddleware` to mirror the lead, or tighten the AGENTS.md wording to "before the safety tail" (the README already phrases it that way).

backend/packages/harness/deerflow/agents/middlewares/configured_extensions.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    middlewares: list[AgentMiddleware] = []
+    for middleware_path in list(app_config.extensions.middlewares or []):
+        middleware_cls = resolve_class(middleware_path, AgentMiddleware)
+        try:
+            middleware = middleware_cls()
+        except Exception:
+            logger.exception("Failed to instantiate configured extension middleware %s", middleware_path)
+            raise
+        if not isinstance(middleware, AgentMiddleware):
```
willem-bd: **Effectively-dead guard + redundant cast**

`resolve_class(middleware_path, AgentMiddleware)` already raises `ValueError` if the resolved object isn't an `AgentMiddleware` subclass, so `middleware_cls()` is guaranteed to produce an `AgentMiddleware` instance (barring a pathological `__new__`). This `isinstance` check / `TypeError` branch therefore can't fire in practice and has no test coverage, and the `cast(AgentMiddleware, middleware)` on the next line is redundant once the type is narrowed.

The `try/except` that logs the failing path before re-raising is the valuable part - keep that; this block can go. (`test_build_middlewares_rejects_invalid_configured_extension_middleware` exercises `resolve_class`'s `ValueError`, not this `TypeError`.)

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -143,6 +147,10 @@ class ExtensionsConfig(BaseModel):
     )
     model_config = ConfigDict(extra="allow", populate_by_name=True)
 
+    def to_file_dict(self) -> dict[str, Any]:
```
willem-bd, line 150: **Nice incidental fix - worth a test**

Switching the skill/MCP write paths to `to_file_dict()` (= `model_dump(by_alias=True)`) now preserves `extra="allow"` fields like `mcpInterceptors` when writing `extensions_config.json`. The previous hand-built `{"mcpServers": …, "skills": …}` dict in `skills.py` / `client.py` dropped `mcpInterceptors` on every skill/MCP API write.

Worth a line in the PR description, and a regression test asserting `mcpInterceptors` survives a skill toggle would lock this in - the new e2e assertions only check that `middlewares` is preserved.

willem-bd (commented):
Reviewed the config-declared lead/subagent middleware extension point. Overall well-built - `resolve_class` reuse gives consistent actionable failure hints, and the `ClarificationMiddleware`-last tail invariant is preserved (verified by the `-5:` ordering assertion).

Findings (inline):
1. **Medium** - shipped `extensions.middlewares: []` in `config.example.yaml` silently clobbers any middlewares set in `extensions_config.json`, because the yaml merge wins unconditionally. The realistic default case isn't covered by tests.
2. **Low/Medium** - configured extensions sit *before* `LoopDetection`/`TokenBudget` in the subagent chain but *after* them in the lead chain; the AGENTS.md wording only matches the lead.
3. **Low** - redundant `isinstance`/`TypeError`/`cast` in `configured_extensions.py` (already guaranteed by `resolve_class`).
4. **Positive** - `to_file_dict()` incidentally preserves `mcpInterceptors` on skill/MCP writes (previously dropped); worth a regression test.

The one I'd want resolved before merge is #1.

backend/packages/harness/deerflow/config/extensions_config.py
ZenCGL: I have addressed middleware extension review

willem-bd (commented):
Code review of the config-declared lead/subagent middlewares extension, verified against head `0a069ec3`.

Net: clean implementation that reuses `resolve_class` for consistent failure paths, preserves the `ClarificationMiddleware`-last invariant, and incidentally fixes a latent bug where `update_skill`/`update_mcp_config` dropped `mcpInterceptors` (now round-tripped via `to_file_dict()`). Approve with the four inline notes below - none blocking.

backend/packages/harness/deerflow/agents/middlewares/configured_extensions.py
```diff
@@ -0,0 +1,34 @@
...
+    ``module.path:ClassName`` format. Import, attribute, and subclass validation
+    intentionally go through the shared reflection resolver so failures carry
+    the same actionable dependency hints as models, tools, sandbox providers,
+    and guardrail providers.
+    """
+    middlewares: list[AgentMiddleware] = []
+    for middleware_path in list(app_config.extensions.middlewares or []):
+        middleware_cls = resolve_class(middleware_path, AgentMiddleware)
+        try:
+            middleware = middleware_cls()
```
willem-bd, line 29: **Trust-boundary note (not a blocker, consistent with existing patterns).** `extensions.middlewares` loads and instantiates arbitrary classes from config - i.e. config-declared code execution. This matches DeerFlow's existing model (models/tools/sandbox/guardrail providers, `mcpInterceptors`, and MCP servers already run arbitrary commands), so it's not a new vulnerability class, but two things are worth making explicit:

- The Gateway skill/MCP toggle endpoints only mutate `skills`/`mcpServers` and preserve the rest via `to_file_dict()`; they don't let an API caller set `middlewares`. Good - no new API injection vector. Worth a sentence in the docs so a future contributor doesn't add a `middlewares` write path without considering it.
- Operators must keep treating `config.yaml`/`extensions_config.json` as trusted (already required by MCP). The README/AGENTS additions mention "fail loudly" but don't state the trust assumption; a one-liner would help.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -369,6 +369,10 @@ def build_subagent_runtime_middlewares(
 
         middlewares.append(TokenBudgetMiddleware.from_config(token_budget_config))
 
+    from deerflow.agents.middlewares.configured_extensions import load_configured_extension_middlewares
+
+    middlewares.extend(load_configured_extension_middlewares(app_config))
```
willem-bd, line 374: **Same middlewares run in both lead and subagent chains, zero-arg.** `load_configured_extension_middlewares(app_config)` is called here with the same list used by the lead agent, and classes are instantiated with no arguments. Two consequences worth documenting:

- A middleware written for the lead agent (assuming lead-only state/tools) also runs in every subagent. AGENTS.md notes subagents "receive the same configured extension middlewares," but the `config.example.yaml` comment doesn't - a one-line "applies to both lead and subagent runtimes" there would set expectations.
- Zero-arg construction means no per-context parameterization; a deployment that needs lead-only vs subagent-only guards can't express it. Fine for an MVP, but worth flagging as a known limitation.

backend/packages/harness/deerflow/config/app_config.py
willem-bd: **This merge operates on raw yaml dict keys (alias-form pitfall).** `extensions_data` is `model_dump(by_alias=True)` (keys: `mcpServers`, `middlewares`, `skills`, `mcpInterceptors`), but `yaml_extensions` is the unvalidated yaml dict, so a user who writes `extensions.mcp_servers:` (snake_case) in yaml adds a `mcp_servers` key *alongside* the file's `mcpServers` instead of overriding it. The example only uses `middlewares` (no alias), so the common path is fine, but consider validating `yaml_extensions` through `ExtensionsConfig` first, or documenting that yaml must use the public aliased keys.

Also worth stating explicitly: this is **replace-per-field, not list-concat** - if both file and yaml list `middlewares`, yaml's list wins entirely (verified by `test_app_config_preserves_config_yaml_extension_middlewares`). Reasonable, but users may expect merging.

backend/tests/test_lead_agent_model_resolution.py
```diff
@@ -723,6 +769,46 @@ def test_build_middlewares_allows_runtime_subagent_total_limit_override(monkeypa
     assert limit.max_total == 5
 
 
+def test_build_middlewares_rejects_invalid_configured_extension_middleware(monkeypatch):
```
willem-bd, line 781: **Coverage gaps.** This test uses `_make_model` (a function), which hits `resolve_variable`'s `expected_type=type` check (`"not an instance of type"`). It doesn't cover `resolve_class`'s other branch - a real class that isn't an `AgentMiddleware` subclass (`"is not a subclass of AgentMiddleware"`). Consider adding a case pointing at a non-middleware class to lock that branch.

Also no test for the instantiation-failure path in `load_configured_extension_middlewares` (`middleware_cls()` raising -> `except Exception: logger.exception(...); raise`). A middleware whose `__init__` raises is an expected failure mode worth a regression test.

ZenCGL: I have  tighten middleware extension docs and tests

WillemJiang (approved):

PR #3966, opened 2026-07-06
fix(backend): stop classifying UTF-16 markdown files as binary

### LINKED ISSUE #3963: [bug] 工作区变更区域中，部分md文件被识别为二进制文件
### Before you start

### Problem summary

工作区变更区域中，部分md文件被识别为二进制文件

### Affected area(s)

### What happened?

在对话页面中，执行一个任务，生成了7个md文件，页面上”工作区变更“区域中，部分md文件被识别为二进制文件，显示无法diff。

Image

Image

[code_review_guide.md](https://github.com/user-attachments/files/29690931/code_review_guide.md)
上传了一个被识别为二进制的md文件，供参考定位。

### Expected behavior

md文件可以被正常识别。

### Steps to reproduce

1、对话页面上执行生成式任务，生成几个md文件。
2、查看工作区变更。

### Relevant logs

```shell
无
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Support bundle summary

_No response_

### Additional context

_No response_

Vanzeren:
<!-- Reference a related issue with #123. Use Fixes / Closes / Resolves to
     auto-close it on merge. Delete this line if the PR doesn't reference an issue. -->
Fixes #3963

## Why

<!-- Why are you opening this PR? Cover two things:
       - The trigger — what made you write this? A bug you hit, a feature you need,
         tech debt, or a prod issue?
       - The pain being addressed — user-facing problem, or what it unblocks.
     For non-trivial features, please open an issue/discussion first to align on
     scope before writing code. -->
Some generated .md files were showing up as “binary file, diff unavailable” in Workspace Changes,      
 which made the change review UI much less useful for document-heavy runs.                              
                                                                                                        
 The immediate pain is user-facing: even though the outputs are Markdown, reviewers cannot inspect what 
 changed from the workspace diff panel. This PR makes those text files render as diffs instead of being 
 misclassified as binary.

## What changed

Workspace Changes now treats BOM-encoded UTF-16 Markdown files as text and shows normal diffs for them 
 instead of marking them as binary.                                                                     
                                                                                                        
 The change also normalizes cached text snapshots to UTF-8 so modified-file diffs keep working when the 
 baseline file was stored in UTF-16. True binary files are still reported as binary and remain          
 undisplayable in the diff panel. 

## Surface area

## Screenshots / Recording

**Before** :
<img width="1815" height="675" alt="屏幕截图_20260706_163329" src="https://github.com/user-attachments/assets/b8202ca1-57c4-4d4b-bf0f-b86497fcdb7c" />
**After** :
<img width="1755" height="1615" alt="屏幕截图_20260706_163232" src="https://github.com/user-attachments/assets/25452e3c-523d-4155-a1ce-125a0bdcd828" />
## Bug fix verification

<!-- Skip (delete) this section if this PR is not a bug fix.

     Bugs should be encoded as a failing test that goes red before the fix.
     Confirm:
       - Did it go red on `main` and green on this branch? (yes / no)
       - If a red test wasn't cheap to write, explain why and what you did instead. -->
   - backend/tests/test_workspace_changes.py                                                            
   - Added:                                                                                             
     - test_compare_snapshots_treats_utf16_markdown_as_text                                             
     - test_compare_snapshots_reads_cached_utf16_markdown_baseline                                      
 - Did it go red on main and green on this branch?                                                      
   - yes                                                                                                
 - Notes:                                                                                               
   - Before the fix, the new tests failed with binary=True / diff_unavailable_reason="binary".          
   - After the fix, the same tests pass and existing binary-file coverage still passes.


## Validation

Actually ran:                                                                                          
                                                                                                        
 ```bash                                                                                                
   cd backend && uv run pytest tests/test_workspace_changes.py -q                                       
   uv run --project backend python scripts/scan_changed_blocking_io.py --base origin/main               
 ```                                                                                                    
                                                                                                        
 Observed:                                                                                              
                                                                                                        
 - backend/tests/test_workspace_changes.py → 15 passed                                                  
 - backend/tests/blocking_io via make test-blocking-io → 36 passed                                      
 - blocking-IO scan → No blocking-IO candidates introduced by this change 

## AI assistance

**Tool(s) used:** <!-- e.g. Claude Code, Cursor, GitHub Copilot, Codex, Windsurf, or "none" --> Codex

**How you used it:** <!-- e.g. "generated the module from a spec", "autocomplete only",
     "AI wrote tests, I wrote the impl". A prompt or conversation link is great too. --> Used Codex to investigate the root cause, write regression tests, implement the       
 backend fix, and run targeted verification. I reviewed the resulting changes and verification output   
 manually.   


Huixin615:
@Vanzeren Thanks for the fix and regression tests. I found two blocking issues:
1. This does not fix the actual reproduction from #3963. The attached Markdown file is valid UTF-8, but byte 4096 cuts through a multi-byte Chinese character. _decode_text_bytes(sample) therefore raises unexpected end of data, and both the base commit and this PR still classify the file as binary.
2. Replacing the explicit NUL-byte check introduces a regression: content such as b"\x00\x01\x02" is valid UTF-8 and is now classified as text.

Please preserve the NUL-byte check and make sample decoding tolerate an incomplete multi-byte character at the end of the sample, for example by using an incremental decoder with final=False.
Could you also add regression tests for:
- a valid UTF-8 Markdown file whose multi-byte character crosses the 4096-byte sample boundary;
- binary content containing NUL bytes but no invalid UTF-8 sequence?

The existing UTF-16 tests pass, but they cover a different case from the issue’s attached reproduction.

Vanzeren:
Thanks. I will tighten _looks_binary() to keep the NUL-byte binary check and make UTF-8   
 sample detection tolerate an incomplete trailing multibyte character, and add regressions for both   
 cases.

Huixin615 (approved):
Thanks for the follow-up. I re-reviewed the latest commit d6ce3419 and verified it against the original file attached to #3963.
Both previous blockers are resolved:
- valid UTF-8 text crossing the 4096-byte sample boundary is now correctly treated as text;
- NUL-containing binary content remains classified as binary.

I also verified the UTF-16 BOM and invalid UTF-8 paths. The workspace-change tests and blocking-I/O tests pass, and Ruff is clean.
No further blocking findings from me. LGTM!

backend/packages/harness/deerflow/workspace_changes/scanner.py
```diff
@@ -238,11 +242,38 @@ def _sha256_file(path: Path) -> str:
     return digest.hexdigest()
 
 
-def _looks_binary(sample: bytes) -> bool:
-    if b"\x00" in sample:
-        return True
+def _decode_text_bytes(data: bytes) -> str | None:
+    for encoding in ("utf-8", "utf-8-sig"):
```
willem-bd: The `("utf-8", "utf-8-sig")` order makes the `utf-8-sig` branch dead code and leaves a stray BOM in the decoded text.

Python's plain `utf-8` codec **accepts** the BOM bytes `ef bb bf` and decodes them as the `U+FEFF` character — it does not raise `UnicodeDecodeError`. So the loop returns on the first iteration and the `utf-8-sig` arm (which strips the BOM) is never reached:

```
b"\xef\xbb\xbf# hello".decode("utf-8")     # -> "﻿# hello"   (no error)
b"\xef\xbb\xbf# hello".decode("utf-8-sig") # -> "# hello"
```

Consequence: a UTF-8-with-BOM Markdown file is now correctly classified as **text** (good), but its decoded content — and therefore `change.diff` and the cached baseline written by `_cache_text_file` — starts with an invisible zero-width `﻿`. The first added diff line renders as `+﻿# Title`, and the BOM persists into the UTF-8 cache across snapshots.

The same dead ordering exists in `_looks_binary` (line 275–277): `_sample_decodes_as_text(sample, "utf-8")` already returns `True` for a BOM-prefixed sample, so the trailing `sample.startswith(b"\xef\xbb\xbf") and _sample_decodes_as_text(sample, "utf-8-sig")` branch is unreachable too (harmless there since the file is still classified as text, but misleading).

**Failure scenario:** user saves `guide.md` as UTF-8 with BOM (common on Windows editors); the Workspace Changes diff begins with a stray `﻿` on the first line.

**Suggested fix:** reverse the order to `("utf-8-sig", "utf-8")` in `_decode_text_bytes` (sig-aware first, plain utf-8 as fallback) and drop the now-redundant utf-8-sig special-case in `_looks_binary`. A small regression test (`test_compare_snapshots_strips_utf8_bom`) would lock it in.

(Posted from a `/review` pass — this is the only finding that survived verification; the `OSError`-only catch in `diff.py` and the `binary = True` reassignment after a full-file decode failure both check out.)

WillemJiang (approved):

PR #3969, opened 2026-07-06
feat: add manual context compaction

18062706139fcz:
## Summary
- add `POST /api/threads/{thread_id}/compact` to manually compact older active thread context into `summary_text`
- reuse the existing summarization middleware for forced compaction and write a new checkpoint with updated `messages` / `summary_text` channel versions
- add `/compact` and `/context compact` composer commands with localized UI feedback
- document the user-facing command and backend/frontend development behavior

## Behavior
- full visible chat history remains available in the UI
- future model calls use the compacted summary plus the retained recent message window
- compaction is skipped for empty/new threads and blocked while a thread has a run in flight

## Tests
- `PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run pytest tests/test_context_compaction.py tests/test_summarization_summary_text.py -q`
- `uvx ruff check app/gateway/routers/threads.py packages/harness/deerflow/agents/middlewares/summarization_middleware.py packages/harness/deerflow/runtime/context_compaction.py tests/test_context_compaction.py tests/test_summarization_summary_text.py`
- `uvx ruff format --check app/gateway/routers/threads.py packages/harness/deerflow/agents/middlewares/summarization_middleware.py packages/harness/deerflow/runtime/context_compaction.py tests/test_context_compaction.py tests/test_summarization_summary_text.py`
- `pnpm check`
- `pnpm test tests/unit/components/workspace/input-box-helpers.test.ts`
- `pnpm format`


backend/app/gateway/routers/threads.py
```diff
@@ -807,6 +841,49 @@ async def clear_thread_goal(thread_id: str, request: Request) -> ThreadGoalRespo
...
+        preserved_message_count=result.preserved_message_count,
+        summary_updated=result.summary_updated,
+        checkpoint_id=result.checkpoint_id,
+        total_tokens=result.total_tokens,
+    )
+
+
+@router.post("/{thread_id}/compact", response_model=ThreadCompactResponse)
+@require_permission("threads", "write", owner_check=True, require_existing=True)
+async def compact_thread(thread_id: str, body: ThreadCompactRequest, request: Request) -> ThreadCompactResponse:
```
willem-bd, line 860: **`/compact` doesn't acquire `goal_thread_lock(thread_id)`, racing with `/goal` writes.**

`set_thread_goal` / `clear_thread_goal` (lines 818, 834) both wrap their checkpoint read-modify-write in `async with goal_thread_lock(thread_id)`; this handler doesn't. Both paths read checkpoint V, rebuild the full `channel_values` from V, and `aput` — last writer wins. If `/goal` writes V+1 (new goal) while `/compact` is still computing from V, compaction's V+2 carries the *old* `channel_values` and the goal silently vanishes (or, in the reverse order, the summary/compaction is lost). Consider holding `goal_thread_lock(thread_id)` around the whole compaction.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -601,6 +610,46 @@ export function InputBox({
     ],
   );
 
+  const handleCompactCommand = useCallback(async (): Promise<void> => {
```
willem-bd, line 618: **No AbortController / staleness guard, unlike `handleGoalCommand`.**

`handleGoalCommand` uses `beginGoalRequest(...)` (creates an `AbortController`, tracks `threadId`) and guards each state mutation with `isCurrentGoalRequest(...)`. This handler has neither. The summarization model call takes several seconds; if the user switches threads mid-flight, the resolved callback still fires `setInput("")`, `setFollowups([])`, `setFollowupsHidden(false)`, `setFollowupsLoading(false)`, and resets the prompt-history refs — against the *new* thread's composer, wiping an in-progress draft. `frontend/AGENTS.md` documents the `/goal` AbortController pattern for exactly this case. Wire `/compact` through the same request-lifecycle tracking.

backend/app/gateway/routers/threads.py
```diff
@@ -807,6 +841,49 @@ async def clear_thread_goal(thread_id: str, request: Request) -> ThreadGoalRespo
...
+    """Manually summarize old thread context while preserving the visible history."""
+    run_manager = get_run_manager(request)
+    if await run_manager.has_inflight(thread_id):
```
willem-bd: **`has_inflight` is a check-then-act (TOCTOU) with no follow-through lock.**

Between this returning `False` and `compact_thread_context`'s `aput` committing, the user can submit a normal message that admits a run reading the pre-compaction checkpoint V. Compaction then writes V+1 (compacted); the run later writes V+2 from V's `channel_values`, clobbering the summary and the restored preserved messages — so compaction silently has no effect. This is the same class of hazard as the `/goal` race above and can be fixed by the same per-thread serialization.

backend/packages/harness/deerflow/runtime/context_compaction.py
```diff
@@ -0,0 +1,191 @@
...
+        return current_version + 1.0
+    if isinstance(current_version, str):
+        try:
+            return str(int(current_version) + 1)
+        except ValueError:
+            return f"{current_version}.1"
+    return 1
+
+
+def _create_compaction_middleware(
```
willem-bd, line 40: **Near-duplicate of `_create_summarization_middleware` (`agents/lead_agent/agent.py:88`), and drops the `if config.model_name:` fallback.**

The live factory branches: when `config.model_name` is empty it calls `create_chat_model(thinking_enabled=False, app_config=...)` with no `name`, to resolve the default model. This copy always passes `name=config.model_name` (i.e. `None`/`""` when unset), so when `summarization.model_name` is unconfigured manual compaction either raises (opaque 500) or summarises on a different model than the one that produced prior `summary_text`. Worth extracting a shared factory (parameterised by an optional `keep` override) so the two can't drift.

backend/packages/harness/deerflow/runtime/context_compaction.py
```diff
@@ -0,0 +1,191 @@
...
+    """Summarize old messages in a thread and write a compacted checkpoint."""
+    resolved_app_config = app_config or get_app_config()
+    middleware = _create_compaction_middleware(app_config=resolved_app_config, keep=keep)
+
+    read_config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
+    checkpoint_tuple = await checkpointer.aget_tuple(read_config)
+    if checkpoint_tuple is None:
+        raise LookupError(f"Thread {thread_id} checkpoint not found")
+
+    checkpoint: dict[str, Any] = copy.deepcopy(getattr(checkpoint_tuple, "checkpoint", {}) or {})
```
willem-bd: **`copy.deepcopy` of the whole checkpoint is wasteful and fragile.**

`write_thread_goal` (`runtime/goal.py:456`) uses shallow `dict(...)` copies and deep-copies only the single value it mutates. Here the entire checkpoint — including the full `messages` history and every channel value — is deep-copied, then `channel_values["messages"]` is replaced wholesale below, so the deep-copied history is allocated only to be discarded (O(history) per `/compact`). Any channel value that doesn't survive `deepcopy` (a lock/handle/lazy proxy living in `sandbox`/`artifacts`) will raise, get caught by the route's generic handler, and surface as a 500. Mirror `write_thread_goal`'s shallow-copy approach.

backend/packages/harness/deerflow/runtime/context_compaction.py
```diff
@@ -0,0 +1,191 @@
...
+        return ThreadCompactionResult(thread_id=thread_id, compacted=False, reason="not_enough_messages")
+
+    state = {
+        "messages": list(messages),
+        "summary_text": channel_values.get("summary_text"),
+    }
+    if middleware._prepare_compaction(state, force=force) is None:
+        return ThreadCompactionResult(thread_id=thread_id, compacted=False, reason="not_enough_messages")
+
+    runtime = SimpleNamespace(context={"thread_id": thread_id, "user_id": user_id})
```
willem-bd: **`SimpleNamespace` runtime leaves `memory_flush_hook`'s `agent_name=None`.**

`_resolve_agent_name(runtime)` reads `runtime.context["agent_name"]` (absent here), then falls back to `get_config()`, which raises `RuntimeError` outside a graph context (caught) → `agent_name=None`. The automatic summarization path carries the real agent name, so when `memory.enabled` manual compaction enqueues summarized messages into the memory queue with `agent_name=None`, the attribution diverges from auto-summarization. Set `agent_name` on the runtime context if memory attribution matters.

backend/packages/harness/deerflow/runtime/context_compaction.py
```diff
@@ -0,0 +1,191 @@
...
+    channel_values: dict[str, Any] = dict(checkpoint.get("channel_values", {}) or {})
+    messages = channel_values.get("messages")
+    if not isinstance(messages, list) or not messages:
+        return ThreadCompactionResult(thread_id=thread_id, compacted=False, reason="not_enough_messages")
+
+    state = {
+        "messages": list(messages),
+        "summary_text": channel_values.get("summary_text"),
+    }
+    if middleware._prepare_compaction(state, force=force) is None:
```
willem-bd: **`_prepare_compaction` runs twice on the hot path.**

This pre-flight call runs the full token-count + cutoff + partition + reminder-preservation pipeline; `acompact_state` (line 142) then calls it again internally, and only that second result is used. For a long thread this doubles compaction latency/CPU (and double-counts anything `token_counter` meters). The pre-flight exists only to avoid building the `SimpleNamespace`; let `acompact_state` return `None` and map that to the no-op result instead.

backend/packages/harness/deerflow/runtime/context_compaction.py
```diff
@@ -0,0 +1,191 @@
...
+    force: bool = True,
+    user_id: str | None = None,
+    app_config: AppConfig | None = None,
+) -> ThreadCompactionResult:
+    """Summarize old messages in a thread and write a compacted checkpoint."""
+    resolved_app_config = app_config or get_app_config()
+    middleware = _create_compaction_middleware(app_config=resolved_app_config, keep=keep)
+
+    read_config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
+    checkpoint_tuple = await checkpointer.aget_tuple(read_config)
```
willem-bd: **Checkpoint I/O bypasses `goal.py`'s `_call_checkpointer_method`, and version bumping is re-implemented.**

`runtime/goal.py:374` routes `aget_tuple` / `aput` through `_call_checkpointer_method`, which offloads sync checkpointers via `asyncio.to_thread` (the blocking-IO gate documented in `backend/AGENTS.md`). This calls `checkpointer.aget_tuple` / `aput` directly, and `_bump_channel_version` re-implements channel-version bumping with extra bool/float/str branches the existing code (`goal.py:467`) never needed. Reuse the helper so a sync / partially-sync checkpointer doesn't block the ASGI event loop during compaction, and to avoid two sources of truth for version bumping.

willem-bd (commented):
Reviewed the manual-context-compaction diff. The `summarization_middleware.py` refactor (extracting `_prepare_compaction` / `compact_state` / `acompact_state`) is faithful — no invariants were dropped from the automatic path. The findings below are in `compact_thread_context` / the `/compact` route (how it diverges from the established `/goal` checkpoint-write pattern in the same `runtime/` package) and in the frontend `handleCompactCommand` (how it diverges from `handleGoalCommand`).

Most severe: the missing `goal_thread_lock` (#1), the missing frontend staleness guard (#2), and the inflight TOCTOU (#3) — these look like blockers. The rest are reuse / efficiency cleanups.

xg-gh-25:
**Why manual compaction matters:** when context grows beyond the summarization window, the *user knows* they're at a natural breakpoint (e.g., pivot to new topic) — but the system waits for the next model call. This PR gives the user control: compact NOW, before the next expensive round-trip.

**Production pattern we've validated:** *user-triggered context operations* beat auto-threshold every time when:
1. The user can see a semantic boundary the model can't (e.g., "done with research, now drafting").
2. Early compaction reduces next-call latency (summary computed upfront, not inline during generation).

**Why this implementation is sound:**
```python
# Reuses existing summarization middleware
POST /api/threads/{thread_id}/compact
  → force-compact older active messages into summary_text
  → write new checkpoint with updated channel versions
  → skip if thread is empty OR run in flight (safe guard)
```

The key design choice: **full visible history stays in UI**, only *future model calls* use the compacted summary + recent window. Users still browse the uncompacted transcript — no data loss, purely a token-budget optimization.

**Our similar approach:** we expose `/context compact` as a composer command (not just API) because *agents working with humans benefit from explicit checkpoints*. The user saying "compact now" is a *semantic signal* (topic boundary) that pure token-count thresholds miss. This PR nails that.

**Test coverage confirms no regressions:**
```bash
pytest tests/test_context_compaction.py tests/test_summarization_summary_text.py
ruff check + format (all pass)
pnpm check + pnpm test (frontend coverage)
```

Risk label `risk:high` + `needs-validation` is appropriate — compaction is user-visible state mutation. Recommend: merge after manual QA on a long-running thread (50+ messages) to verify UI stays intact post-compact.

**Edge case to watch:** if a user spams `/compact` repeatedly on the same thread, does each compact re-summarize the already-summarized summary (lossy cascade)? The PR description doesn't clarify if `summary_text` is preserved or regenerated. Ideal: compaction is idempotent (second compact on unchanged messages is a no-op).

---
*Implemented similar user-triggered compaction in session checkpoints. Discussion: [T-MEM: Memory is the Moat](https://github.com/xg-gh-25/SwarmAI/discussions/4)*

WillemJiang (approved):

PR #3971, opened 2026-07-06
feat: add Lark CLI integration

### LINKED ISSUE #3957: [feat] 建议新增 official Lark/Feishu CLI  内置在沙箱中, 启动时候方便用户快速接入 feishu 生态，（界面也可以优化下, 一键集成飞书生态）
### Before you start

### Problem / motivation

目前非技术用户无法自己安装  lark-cli， 但是这个功能对办公场景非常重要, 其他大厂产品已集成：如 workbuddy：

Image

参考官方 lark cli: https://github.com/larksuite/cli

### Proposed solution

目前问题：
现在不能纯界面操作，需要三步，其中第 1 步涉及环境安装：

  第 1 步：在环境里安装 lark-cli 二进制

  你用的是 Docker Dev 模式（deer-flow-dev-gateway），需要在 Gateway 容器或沙箱中安装：

  # 进入 Gateway 容器
  docker exec -it deer-flow-gateway sh

  # 安装 lark-cli（macOS/Linux 二进制）
  # 参考 https://github.com/larksuite/cli 的安装说明

  或者更简单：在 DeerFlow 的沙箱镜像中预装 lark-cli，这样 Agent 的 bash 工具可以直接调用。

  第 2 步：把 Skills 复制进 DeerFlow 的 skills/custom/

  Lark CLI 仓库的 skills/ 目录下有 20+ 个 skill（lark-doc、lark-im、lark-base、lark-calendar、lark-sheets、lark-task、lark-mail、lark-approval、lark-wiki 等），结构是：

  lark-cli/skills/
  ├── lark-doc/
  │   ├── SKILL.md
  │   └── references/
  ├── lark-im/
  │   ├── SKILL.md
  │   └── references/
  ├── lark-base/
  │   ├── SKILL.md
  │   └── references/
  ...

  DeerFlow 的 custom skills 目录在 deer-flow/skills/custom/。需要把每个 lark-* 目录复制进去，同时把 SKILL.md 中的 ../lark-shared/ 交叉引用修正为绝对路径（因为 DeerFlow 的 skill 路径是
  /mnt/skills/custom/lark-shared/...）。

  第 3 步：在前端界面启用

  复制完成后，打开 DeerFlow 前端 → Settings → Skills → Custom 标签，每个 Lark skill 都会出现在列表中，一键开关即可启用。

  然后 Agent 就能通过 /lark-doc、/lark-base、/lark-im 等斜杠命令激活对应能力。

  总结

  用户现在能做的（纯 UI）:
    ✅ 开关 Skill（Settings → Skills → 切换）

  用户现在不能做的（需要手动）:
    ❌ 安装 lark-cli 二进制到容器/沙箱
    ❌ 把 20+ skills 目录复制到 skills/custom/
    ❌ 批量上传 SKILL.md（只支持 .skill 压缩包安装单个 skill）

  最现实的方案是写一个安装脚本，一键完成：拉取 Lark CLI skills → 复制到 skills/custom/ → 重启 Gateway。

### Affected area(s)

Sandbox / Docker, Frontend (UI / Next.js)

### Alternatives considered

_No response_

### Additional context

_No response_

18062706139fcz:
## Why

This closes the usability gap described in #3957. Feishu/Lark is a core office workflow surface, but the previous path was not realistic for non-technical users:

- `lark-cli` had to be installed manually inside the Gateway/container or sandbox runtime.
- The 20+ official `lark-*` skills had to be copied by hand into `skills/custom/`.
- Skill path references and enablement had to be handled manually.
- User authorization required terminal commands such as `lark-cli config init` and `lark-cli auth login`.

That made the integration environment-dependent, easy to misconfigure, and effectively limited to developers/admins. The goal of this PR is to turn Lark into a managed first-party integration: install the official skill pack once from the UI, then let each user connect their own Lark/Feishu account through the browser.

### Managed Lark CLI Integration

- Adds a managed Lark/Feishu CLI integration package backed by the official `lark-*` skills from `@larksuite/cli`.
- Installs the skills into a user-scoped integration directory instead of mixing them into custom user-authored skills:
  - host: `.deer-flow/users/{user_id}/skills/integrations/lark-cli/...`
  - sandbox: `/mnt/skills/integrations/lark-cli/...`
- Marks these skills as integration skills so they are treated as read-only, managed content rather than mutable custom skills.
- Writes an install manifest with the pinned source/version and installed skill names.

### Gateway API

Adds `/api/integrations/lark/*` endpoints for the complete connection lifecycle:

- `GET /status` checks skill-pack install state, enabled skills, Gateway `lark-cli` availability, app config, and user auth status.
- `POST /install` installs or reinstalls the official managed Lark skill pack and refreshes the user skill prompt cache.
- `POST /config/start` starts first-time Lark/Feishu app setup through a browser device-flow link.
- `POST /config/complete` completes that app setup after the user approves it in the browser and persists the app credentials through `lark-cli`.
- `POST /auth/start` starts browser-based user OAuth authorization with `lark-cli auth login --no-wait --json`.
- `POST /auth/complete` completes the user authorization after the browser approval.

The first-time app setup is intentionally internalized behind the "Connect Lark" action, so users do not need to understand or run `lark-cli config init`.

### UI / User Experience

- Adds `Settings -> Integrations -> Lark / Feishu CLI`.
- Shows the three relevant readiness states in one card:
  - managed skill pack installed
  - Gateway `lark-cli` available
  - user Lark authorization connected
- Adds an admin-only install/reinstall action.
- Adds a single "Connect Lark" flow that:
  - opens the browser setup/auth link
  - shows a copyable fallback URL
  - asks the user to click "continue" only after finishing approval in the browser
  - avoids a hidden long-polling state that looks like page refresh or UI hanging
- Adds `?settings=integrations` deep-link support so agent guidance can take users directly to the integration settings page.
- Adds mock API routes and tests for the integration UI.

### Sandbox / Runtime

- Updates local and AIO sandbox providers to mount managed integration skills under `/mnt/skills/integrations`.
- Extends sandbox path resolution so skill references under `/mnt/skills/integrations/...` resolve safely to the current user's managed integration skill directory.
- Keeps Lark skill support files available to tools without exposing them as ordinary writable custom skills.
- Adds DeerFlow-specific guidance to `lark-shared` so agents direct users to the UI integration flow when auth is missing, instead of asking them to run terminal commands.

### Docker / Deployment

- Installs the pinned official Lark CLI package in the Gateway image so the integration can work out of the box in Docker/dev deployments.
- Adds the integration skills mount to docker-compose volumes.

### Docs

- Documents managed integration skill storage and the Lark/Feishu CLI connection flow in README and agent guidance docs.
- Updates backend/frontend orientation docs with the new integration router, skill category, and frontend module location.

## Security and Isolation

- Integration install is admin-gated.
- Lark CLI config and data directories are scoped per DeerFlow user to prevent local trusted-mode credentials from bleeding across accounts.
- Lark app secret material stays on the Gateway side and is not sent to the browser.
- Managed integration skills are read-only and separated from custom user skills.
- Archive installation validates the official skill pack before replacing the existing install:
  - rejects unsafe zip paths and symlinks
  - rejects executable binary members
  - enforces archive and extracted-size limits
  - requires every expected `lark-*` skill to be present
  - validates each `SKILL.md` before publishing the staging directory
- Sandbox path mapping is restricted to the current user's integration skill root.

## Review Notes

The most important behavior to review is the end-to-end connection flow:

1. Admin installs the managed Lark skill pack from Settings.
2. User clicks "Connect Lark".
3. If app setup is missing, DeerFlow opens the first-time setup browser link and waits for the user to confirm completion.
4. DeerFlow then starts the user OAuth browser authorization.
5. After the user confirms authorization, the status card reports the authenticated Lark account.

This PR does not add a new DB migration. State is stored in the existing per-user DeerFlow home layout and the Lark CLI config/data directories.

## Tests

- `cd backend && uv run ruff check .`
- `cd backend && uv run pytest tests/test_lark_cli_integration.py tests/test_sandbox_tools_security.py -q`
- `cd frontend && PATH=/Users/bytedance/.nvm/versions/node/v22.21.1/bin:$PATH pnpm exec rstest run tests/unit/core/integrations/lark/api.test.ts`
- `cd frontend && PATH=/Users/bytedance/.nvm/versions/node/v22.21.1/bin:$PATH pnpm check`
- `cd frontend && PATH=/Users/bytedance/.nvm/versions/node/v22.21.1/bin:$PATH PORT=3001 PLAYWRIGHT_BASE_URL=http://localhost:3001 pnpm test:e2e tests/e2e/integrations.spec.ts --project=chromium`
- `git diff --check`

Closes #3957

image


backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,857 @@
...
+
+def _resolve_or_download_archive() -> Path:
+    if env_archive := os.getenv(LARK_CLI_SOURCE_ARCHIVE_ENV):
+        return Path(env_archive)
+
+    fd, archive_name = tempfile.mkstemp(prefix="lark-cli-skills-", suffix=".zip")
+    os.close(fd)
+    archive_path = Path(archive_name)
+    try:
+        with urllib.request.urlopen(DEFAULT_LARK_CLI_ARCHIVE_URL, timeout=LARK_CLI_DOWNLOAD_TIMEOUT_SECONDS) as response:
```
willem-bd: **[P1] Downloaded skill archive has no integrity verification**

- **Problem:** The install path fetches `DEFAULT_LARK_CLI_ARCHIVE_URL` over HTTPS and extracts it as trusted first-party content, but nothing verifies authenticity. The structural checks (zip-slip, symlinks, executable-binary prefix, size, skill completeness) only validate *shape*, not *origin*. These skills become agent-readable instructions/reference that the model follows, so a supply-chain compromise of `larksuite/cli`'s release — or a cached/intermediated download — injects malicious skill content into every user's integration dir.
- **Evidence:** `_resolve_or_download_archive()` streams the zip straight to disk and hands it to the extractor with no checksum/signature step.
- **Suggested fix:** Pin an expected SHA-256 (`EXPECTED_LARK_CLI_ARCHIVE_SHA256`) and verify the downloaded bytes before extraction; fail the install on mismatch. Continue to honor `DEER_FLOW_LARK_CLI_SKILLS_ARCHIVE`, but still require the hash to match.
- **Test:** tamper the archive bytes → install fails with a clear error and nothing is written under the user's integration dir.

willem-bd (commented):
Thanks @18062706139fcz. I found a few issues that should be addressed before this is ready.

The overall design is solid — per-user credential isolation via `LARKSUITE_CLI_CONFIG_DIR`/`DATA_DIR`, the app secret kept server-side (`--app-secret-stdin`), and the staging-dir + atomic-swap install are well done. The items below are mostly hardening around the trust boundary plus a couple of fragile couplings. The most important one is the archive integrity check (P1).

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,857 @@
...
+        return tuple(sorted(extracted))
+    except Exception:
+        if backup is not None and backup.exists() and not target.exists():
+            backup.rename(target)
+        raise
+    finally:
+        shutil.rmtree(staging_parent, ignore_errors=True)
+
+
+def _extract_lark_skills(zf: zipfile.ZipFile, destination: Path) -> set[str]:
```
willem-bd, line 1526: **[P2] Security-critical extractor has only happy-path tests**

- **Problem:** `_extract_lark_skills` / `_resolve_lark_skill_member` / `_validate_extracted_lark_skills` carry the entire security claim of the PR, but `test_lark_cli_integration.py` only exercises the valid archive. The rejection paths (zip-slip, symlink member, executable-binary prefix, oversize archive, missing/renamed skill) are unverified.
- **Suggested fix:** Add focused negative tests — one malformed archive per rejection branch — asserting each raises `ValueError` and leaves the target dir untouched.
- **Test:** these *are* the tests; cover every `raise ValueError(...)` in the extractor and validator.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,857 @@
...
+    if not device_code:
+        raise ValueError("device_code is required.")
+    parsed_brand = _normalize_lark_brand(brand)
+    result = _poll_lark_app_registration(
+        device_code=device_code,
+        brand=parsed_brand,
+        interval=interval or 5,
+        expires_in=expires_in or 300,
+    )
+    if not result.get("client_secret") and _tenant_brand(result) == "lark":
```
willem-bd, line 926: **[P2] Lark-vs-Feishu re-poll branch is untested and its contract undocumented**

- **Problem:** The first `_poll_lark_app_registration` already returns on success (`client_id` present), yet `client_secret`-absent + `tenant_brand == "lark"` triggers a full second poll against the Lark endpoint. The Lark response shape that motivates this (does the feishu poll omit `client_secret` for lark tenants? is `user_info.tenant_brand` always populated on a successful first poll?) isn't documented, and no test exercises the re-poll — only the fully-mocked happy path is covered.
- **Suggested fix:** Add a comment capturing the exact Lark response contract, and a test where the first poll returns `client_id` without `client_secret` under `tenant_brand="lark"` asserting the second poll fires and the final credentials are saved.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,857 @@
...
+
+from deerflow.config.app_config import AppConfig
+from deerflow.config.paths import get_paths
+from deerflow.skills.installer import is_executable_binary_prefix, is_symlink_member, is_unsafe_zip_member
+from deerflow.skills.parser import parse_skill_file
+from deerflow.skills.permissions import make_skill_tree_sandbox_readable
+from deerflow.skills.types import SKILL_MD_FILE, SkillCategory
+
+INTEGRATION_ID = "lark-cli"
+DEFAULT_LARK_CLI_VERSION = "v1.0.65"
```
willem-bd: **[P2] Two independent sources of truth for the Lark CLI version**

- **Problem:** `DEFAULT_LARK_CLI_VERSION = "v1.0.65"` (used for the GitHub archive URL, the manifest, and the `lpv`/`ocv` verification-URL params) is independent from `LARK_CLI_NPM_VERSION=1.0.65` in `backend/Dockerfile` (the npm-installed runtime binary). Bumping one without the other silently drifts the installed skill pack from the runtime `lark-cli`, and the `lpv`/`ocv` params won't match the actual CLI version.
- **Suggested fix:** Drive both from a single value, or add a startup compatibility check comparing the manifest version to `lark-cli --version`.
- **Test:** assertion that the Python constant and the Dockerfile ARG resolve to the same version.

backend/app/gateway/routers/integrations.py
```diff
@@ -0,0 +1,315 @@
...
+        logger.error("Failed to get Lark integration status: %s", e, exc_info=True)
+        raise HTTPException(status_code=500, detail=f"Failed to get Lark integration status: {str(e)}")
+
+
+@router.post("/lark/install", response_model=LarkInstallResponse, summary="Install Lark/Feishu Skill Pack")
+async def install_lark(request: Request, config: AppConfig = Depends(get_config)) -> LarkInstallResponse:
+    await require_admin_user(request, detail=_ADMIN_REQUIRED_DETAIL)
+    try:
+        user_id = get_effective_user_id()
+        result = await asyncio.to_thread(install_lark_integration, user_id, config)
```
willem-bd: **[P2] No `tests/blocking_io/` anchor for the new router**

- **Problem:** All six handlers correctly offload `subprocess`/`urllib`/FS work via `asyncio.to_thread`, but this repo gates blocking IO behind a teeth-verified anchor (e.g. `tests/blocking_io/test_skills_install.py` for skill install). The new `integrations` router has no such anchor, so a future refactor that drops `to_thread` here would block the event loop with no regression guard.
- **Suggested fix:** Add a `tests/blocking_io/test_integrations_router.py` anchor that drives `install` / `auth/complete` against a stubbed `lark-cli` and asserts the blocking phases run off the loop.
- **Validation:** `cd backend && make test-blocking-io`.

backend/Dockerfile
```diff
@@ -95,7 +102,8 @@ ENV PYTHONIOENCODING=utf-8
 COPY --from=builder /usr/bin/node /usr/bin/node
 COPY --from=builder /usr/lib/node_modules /usr/lib/node_modules
 RUN ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/bin/npm \
-    && ln -s ../lib/node_modules/npm/bin/npx-cli.js /usr/bin/npx
+    && ln -s ../lib/node_modules/npm/bin/npx-cli.js /usr/bin/npx \
+    && ln -s ../lib/node_modules/@larksuite/cli/scripts/run.js /usr/bin/lark-cli
```
willem-bd: **[P2] Final-image `lark-cli` symlink hardcodes internal package layout**

- **Problem:** `ln -s ../lib/node_modules/@larksuite/cli/scripts/run.js /usr/bin/lark-cli` assumes the package's internal `scripts/run.js` entry path. The builder stage verifies `lark-cli --version` via npm's own bin, but this hand-rolled final-stage symlink is a separate assumption with no runtime check in the final image — an upstream layout change yields a dead symlink discovered only at probe time (`probe_lark_cli` → "not on PATH").
- **Suggested fix:** Derive the bin target from the installed `package.json` `bin` field, or copy the symlink npm already generated, and add a `lark-cli --version` check in the final stage.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -291,6 +301,18 @@ def _resolve_skills_path(path: str) -> str:
             return str(user_custom_dir / custom_relative)
         return str(user_custom_dir)
 
+    if relative == "integrations" or relative.startswith("integrations/"):
```
willem-bd, line 339: This new `integrations/` branch joins `integrations_relative` (everything after `integrations/`) to the user's integration dir without an `is_relative_to(...)` check — a container path like `/mnt/skills/integrations/../../<x>` would lexically resolve outside the user's integration dir on the host.

It mirrors the existing `custom/` branch, so it's not a *new* hole (presumably `_reject_path_traversal` runs upstream and the read-only bind mount contains it), but since this extends a `risk:high` sandbox surface it's worth either confirming that guard runs before this for every caller, or adding the same `resolved.is_relative_to(user_integrations_dir)` defense the extractor already uses. A negative test (`/mnt/skills/integrations/../../etc/passwd` → rejected) would lock it in.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,887 @@
...
+def probe_lark_auth(user_id: str) -> LarkAuthProbe:
+    path = _resolve_lark_cli_path()
+    if path is None:
+        return LarkAuthProbe(status="unavailable", message="lark-cli is not installed on the Gateway")
+    app_config = read_lark_app_config(user_id)
+    if not app_config["configured"]:
+        return LarkAuthProbe(status="not_configured", message="Lark app is not configured")
+    try:
+        result = subprocess.run(
+            [path, "auth", "status", "--json", "--verify"],
```
willem-bd: `--verify` forces a live token check against Lark on **every** `GET /lark/status`. Combined with the `lark-cli --version` probe (5s timeout) this makes each status fetch cost up to ~13s and a network round-trip per poll/refresh. Consider using cached token presence for status and reserving `--verify` for the explicit "complete authorization" step, or briefly caching the probe result.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,887 @@
...
+    if not device_code:
+        raise ValueError("device_code is required.")
+    parsed_brand = _normalize_lark_brand(brand)
+    result = _poll_lark_app_registration(
+        device_code=device_code,
+        brand=parsed_brand,
+        interval=interval or 5,
+        expires_in=expires_in or 300,
+    )
+    if not result.get("client_secret") and _tenant_brand(result) == "lark":
```
willem-bd, line 926: This re-poll against the Lark accounts host when `client_secret` is missing is coupled to undocumented Lark account-host behavior (Feishu host can return `tenant_brand: lark` with `client_id` but no `client_secret`). It's tested and the comment helps, but it's fragile if Lark changes the response shape — worth a monitoring note, or a pointer to the upstream `lark-cli` behavior it mirrors.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,887 @@
...
+
+def complete_lark_auth(user_id: str, config: AppConfig, *, device_code: str) -> LarkAuthCompleteResult:
+    """Complete a Lark device authorization flow after the user approves it."""
+    device_code = device_code.strip()
+    if not device_code:
+        raise ValueError("device_code is required.")
+
+    path = _require_lark_cli_path()
+    _run_lark_cli_json(
+        [path, "auth", "login", "--device-code", device_code, "--json"],
```
willem-bd, line 1007: `complete_lark_auth` is fired automatically by the frontend right after `auth/start` (`completeAuthorization(..., { automatic: true })`), and this call blocks a worker thread for up to the 45s `lark-cli auth login --device-code` timeout. The 504 → "still pending" toast handling is correct and the blocking-IO anchor covers the event loop, but many concurrent connects could pressure the `to_thread` pool. Worth a note in the design docs; not a blocker.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,887 @@
...
+
+def _resolve_or_download_archive() -> Path:
+    if env_archive := os.getenv(LARK_CLI_SOURCE_ARCHIVE_ENV):
+        return Path(env_archive)
+
+    fd, archive_name = tempfile.mkstemp(prefix="lark-cli-skills-", suffix=".zip")
+    os.close(fd)
+    archive_path = Path(archive_name)
+    try:
+        with urllib.request.urlopen(DEFAULT_LARK_CLI_ARCHIVE_URL, timeout=LARK_CLI_DOWNLOAD_TIMEOUT_SECONDS) as response:
```
willem-bd: Install downloads the skill archive from GitHub at install time, so air-gapped deployments must pre-stage `DEER_FLOW_LARK_CLI_SKILLS_ARCHIVE`. That escape hatch is hash-checked (safe — not a bypass), but the internet dependency isn't surfaced in the README or the settings UI. A one-line operator note plus a clear "download failed" message in the UI would help.

backend/app/gateway/routers/integrations.py
```diff
@@ -0,0 +1,315 @@
...
+        return _install_to_response(result)
+    except FileNotFoundError as e:
+        raise HTTPException(status_code=404, detail=str(e))
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e))
+    except HTTPException:
+        raise
+    except Exception as e:
+        logger.error("Failed to install Lark integration: %s", e, exc_info=True)
+        raise HTTPException(status_code=500, detail=f"Failed to install Lark integration: {str(e)}")
```
willem-bd: The catch-all 500 echoes internal exception text (`str(e)` — host paths, `lark-cli` stderr, etc.) to the client. The repo convention for *unexpected* failures leans toward generic 500 details (e.g. the threads-goal endpoint returns a generic 500). The `ValueError`/`FileNotFoundError` → 400/404 mappings are fine (expected, user-facing); consider a generic message for these catch-all 500s and logging the detail server-side only (it's already logged on the line above).

willem-bd (commented):
Review of `feat: add Lark CLI integration` — a large, high-risk PR executed to a high standard. The security-critical archive-install path is properly hardened (SHA-256 pin, zip-slip/symlink/executable/size guards, atomic stage→swap install, per-user isolation, secret-via-stdin, admin-gated install) and the test suite — including blocking-IO anchors and e2e — is thorough. None of the inline notes below are blockers.

Main things worth addressing before/after merge:
- Confirm the upstream traversal guard covers the new `integrations/` resolution branch in `_resolve_skills_path` (and add a negative test) — `sandbox/tools.py:304`.
- Reduce the `--verify` network cost on the status probe — `lark_cli.py:222`.
- Genericize the catch-all 500 detail messages — `integrations.py:228`.

Coverage gaps to consider: a negative traversal test for `_resolve_skills_path`, and a test asserting install is idempotent across reinstalls (exercises the backup/swap-with-existing-target path).

Two design notes: the version-pin coupling (archive URL, SHA-256, 27-entry skill list, Dockerfile/compose version) must move together on upgrade — the `test_python_and_docker_lark_cli_versions_match` check covers the Python↔Dockerfile slice, but the SHA/skill-list can't be auto-checked, so a short upgrade checklist in the module docstring would help. And given the `needs-validation` label, the end-to-end browser device-flow should be exercised against real Lark/Feishu accounts before release — the unit/e2e tests stub the CLI and HTTP layer.

WillemJiang:
@18062706139fcz, it's better to check the Latest release of the Lark CLI from GitHub releases instead of hard-coding it in the Python file backend/packages/harness/deerflow/integrations/lark_cli.py.

18062706139fcz:
## Summary

### Track 1: Lark integration authorization (install + UI + least-privilege)

**Install & versioning** (`lark_cli.py`, `routers/integrations.py`, `README.md`, `AGENTS.md`)
- Install now **resolves the newest `larksuite/cli` release from GitHub `releases/latest`** at install time and downloads that version's skills, instead of a hardcoded version; falls back to a bottom-line pinned version if the lookup fails.
- Dropped the pinned archive-byte SHA (GitHub source archives aren't byte-stable); integrity now relies on **official-host restriction + per-member structural guards + a recorded content SHA** of the extracted skills, so reinstalls are auditable.
- Status API adds `latest_available_version` and `runtime_version_mismatch`; the UI shows installed vs. newest version and skill-pack/runtime drift.
- Auth probe split: `/status` does a cheap offline check by default (`auth status --json`); live `--verify` runs only on the explicit "complete authorization" step.

**Auth UI & least-privilege** (`integrations-settings-page.tsx`, i18n trio, e2e/unit, mocks)
- No longer requests recommended scopes by default: `recommend` defaults to `false` on both frontend and backend (base sign-in only).
- Scope selector expanded from 4 to the **full 21 lark-cli `--domain` business domains + All**, using native `--domain` (read+write); removed the read-only hardcoded mapping for calendar/docs/drive.
- Removed the common-scope dropdown, kept the exact-scope input; i18n refactored from flat keys to an `authDomains` record.

### Track 2: Sandbox lark-cli reuses host authorization (config-isolation fix)

> Fixes: Settings shows authorized, but lark-cli inside the chat sandbox reports unauthorized — the two sides read different config/data dirs.

- **Path helpers** (`config/paths.py`): added `host_user_integration_config_dir/data_dir` + `integration_id` validation (path-traversal guard).
- **AIO sandbox** (`aio_sandbox_provider.py`): mounts `users/{user}/integrations/lark-cli/{config,data}` into fixed container paths so sandbox and Gateway share one auth profile.
- **bash tool** (`sandbox/tools.py`): injects the matching config/data env vars when a command invokes `lark-cli` (host paths locally, container paths in sandbox), plus an `is_relative_to` traversal guard on the `integrations` skill path.
- **Provisioner/K8s** (`remote_backend.py`, `docker/provisioner/app.py`, compose): `create_sandbox` accepts an allowlisted `extra_mounts` (count cap, container-path allowlist, host path must live under the DeerFlow state root, PVC subPath mapping); `thread_id` relaxed to optional; new `DEER_FLOW_HOST_BASE_DIR` env var.
- Also tidied provisioner error strings to single-line form.

### Tests
- Backend: `test_lark_cli_integration.py`, `test_integrations_router.py`, `test_aio_sandbox_provider.py`, `test_remote_sandbox_backend.py`, `test_sandbox_tools_security.py`, `test_provisioner_pvc_volumes.py` (extra_mounts / version resolution / env isolation / traversal coverage).
- Frontend: `api.test.ts`, `integrations.spec.ts` updated contracts and assertions.


** @willem-bd  — requesting a re-review.** Please focus on: (1) the least-privilege trade-off of defaulting `recommend=false`; (2) whether the provisioner `extra_mounts` allowlist and host-path boundary checks are strict enough; and (3) consistency of the sandbox lark-cli env injection and host/container path switching across both AIO and K8s modes. Thanks!

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,1081 @@
...
+        make_skill_tree_sandbox_readable(staging_target)
+
+        if target.exists():
+            backup = parent / f".replacing-{INTEGRATION_ID}-{os.getpid()}"
+            if backup.exists():
+                shutil.rmtree(backup)
+            target.rename(backup)
+        staging_target.rename(target)
+        if backup is not None:
+            shutil.rmtree(backup)
```
willem-bd: **[P2] Post-install backup cleanup can report a successful install as failed**

- **Problem:** After `staging_target.rename(target)` succeeds, the new skills are already live, but the next line `shutil.rmtree(backup)` (no `ignore_errors=True`) can raise on a transient FS error (e.g. EACCES on one file inside the old tree). Control jumps to `except Exception`, whose restore guard `backup.exists() and not target.exists()` is False (target *does* exist with the new content), so the original exception is re-raised and the admin sees an error even though the install fully succeeded. The next install's `if backup.exists(): shutil.rmtree(backup)` then operates on a half-deleted backup.
- **Evidence:** `lark_cli.py:960`. The sibling cleanup at line 967 already uses `shutil.rmtree(staging_parent, ignore_errors=True)`, so this is an inconsistency — the post-install backup deletion is the only rmtree without it.
- **Suggested fix:** `shutil.rmtree(backup, ignore_errors=True)` so post-install cleanup cannot flip a success into a failure.
- **Test:** force `shutil.rmtree` to raise after a successful rename and assert `install_lark_integration` returns success with the new skills on disk and the backup gone.

frontend/src/app/workspace/workspace-content.tsx
```diff
@@ -37,6 +38,7 @@ export async function WorkspaceContent({
         </SidebarInset>
       </SidebarProvider>
       <CommandPalette />
+      <WorkspaceSettingsDeepLink />
```
willem-bd, line 43: **[P2] Two `<SettingsDialog>` instances are mounted — they can open at the same time with racing auth flows**

- **Problem:** `<WorkspaceSettingsDeepLink/>` (always rendered here) mounts its own `<SettingsDialog>`, and `workspace-nav-menu.tsx` mounts another. They share no open-state, so when `?settings=integrations` is present the deep-link dialog is open while the user can also open the nav-menu dialog. If both land on Integrations, two `LarkIntegrationCard` instances run their own automatic-retry loops and blank-browser-window flows against the same `device_code` — duplicate `completeLarkAuthorization` calls and duplicate toasts (per-instance `authToastIdRef` defeats dedup).
- **Evidence:** `workspace-content.tsx:41` -> `<WorkspaceSettingsDeepLink/>` -> `workspace-settings-deep-link.tsx` renders `<SettingsDialog open={open}>`; separately `workspace-nav-menu.tsx:67` renders `<SettingsDialog open={settingsOpen}>`.
- **Suggested fix:** share one open/section store (have the deep-link drive the existing nav-menu dialog instead of mounting its own), or only mount the deep-link dialog when a `?settings=` param is actually present.
- **Test:** Playwright — open `?settings=integrations` then open the nav-menu settings; assert only one dialog is visible and only one `auth/start` request fires.

frontend/src/components/workspace/settings/integrations-settings-page.tsx
```diff
@@ -0,0 +1,822 @@
...
+  const scheduleAuthorizationRetry = (
+    deviceCode: string,
+    attemptId: number,
+  ) => {
+    clearAuthRetryTimer();
+    if (Date.now() >= authDeadlineRef.current) {
+      toast.info(t.settings.integrations.lark.authorizationStillPending);
+      return;
+    }
+    authRetryTimeoutRef.current = setTimeout(() => {
```
willem-bd, line 427: **[P2] Automatic-retry timer can be re-installed after unmount**

- **Problem:** The unmount cleanup (effect clearing `authRetryTimeoutRef`) runs once, but a `completeAuth.mutate` that is in-flight at unmount still resolves its `onSuccess`/`onError`, which call `scheduleAuthorizationRetry` -> `setTimeout(...,1500)` written into the ref the cleanup already cleared. That timer then fires `completeAuthorization` on an unmounted component, which reschedules until `authDeadlineRef` (up to ~`expires_in`, e.g. 600s): toasts / `setPendingFlow` / `refetch` fire after unmount.
- **Evidence:** `integrations-settings-page.tsx:408` (`authRetryTimeoutRef.current = setTimeout(...)`) with the single unmount effect at lines 176-183; react-query fires mutation callbacks after the component unmounts.
- **Suggested fix:** guard the callbacks with an `isMounted`/abort flag (return early when unmounted), or clear `authRetryTimeoutRef` inside every callback path before scheduling.
- **Test:** start the auth flow, unmount the dialog mid-flight, advance timers; assert no `completeLarkAuthorization` call or toast after unmount.

frontend/src/components/workspace/settings/integrations-settings-page.tsx
```diff
@@ -0,0 +1,822 @@
...
+        return withoutAll.filter((item) => item !== domain);
+      }
+      return [...withoutAll, domain];
+    });
+  };
+
+  const handleConnect = async () => {
+    if (!data) return;
+    authRequestRef.current = buildAuthRequest();
+    const browserWindow =
```
willem-bd: **[P2] `handleConnect` decides the popup from stale data -> the browser blocks the auth window**

- **Problem:** `browserWindow` is chosen from cached `data`: when the cache shows `authenticated` with no extra permission request it is set to `null` and no tab is pre-opened. The subsequent `await refetch()` may reveal the session actually expired, so execution falls through to `startConnectionFlow` -> `openAuthorizationUrl(url, null)` -> `window.open(...)` *after* the await — outside the user-gesture window, so Safari/Chrome block it; the verification URL only shows in-dialog and never auto-opens.
- **Evidence:** `integrations-settings-page.tsx:313` (`const browserWindow = data.auth.status === "authenticated" && !hasAdditionalPermissionRequest ? null : openPendingBrowserWindow();`), followed by `await refetch()` and a later `openAuthorizationUrl(..., null)`.
- **Suggested fix:** decide the popup from `latestStatus` (post-refetch), or always pre-open the blank tab whenever the user is not provably connected with no extra permission request.
- **Test:** Playwright — cache shows authenticated, expire server-side, click Connect; assert the auth URL opens (or is surfaced cleanly) rather than being silently blocked.

backend/packages/harness/deerflow/config/paths.py
```diff
@@ -37,6 +38,13 @@ def _validate_user_id(user_id: str) -> str:
     return user_id
 
 
+def _validate_integration_id(integration_id: str) -> str:
```
willem-bd, line 41: **[P2] `_validate_integration_id` accepts `.` and `..`**

- **Problem:** The regex `^[A-Za-z0-9_.\-]+$` admits `.` and `..` as valid integration IDs. Not exploitable today (`integration_id` is the constant `"lark-cli"`), but `host_user_integration_config_dir` / `host_user_integration_data_dir` build the path with `_join_host_path(..., "integrations", integration_id, ...)`, so a future caller passing `..` would resolve one level above the per-integration namespace. `_validate_thread_id` / `_validate_user_id` already forbid `.` / `..`, so this validator is inconsistent.
- **Evidence:** `paths.py:15` (`_SAFE_INTEGRATION_ID_RE`) and `:41` (`_validate_integration_id`).
- **Suggested fix:** reject `.` / `..` explicitly (match the other two validators).
- **Test:** `_validate_integration_id("..")` and `_validate_integration_id(".")` raise `ValueError`.

willem-bd (commented):
Thanks @18062706139fcz. A few more issues worth addressing before this lands — install cleanup, the settings dialog mounting, the auth-retry lifecycle, the connect popup, and an integration-id validator gap.

18062706139fcz:
@willem-bd 麻烦再帮忙 review 一轮。这个 PR 的目标是让非技术用户不需要在终端手动安装/配置 Lark CLI，所以我这轮把“技能包安装、Gateway CLI、授权 UI、对话执行环境”的边界重新收口了一下。

本次新增提交：

`bfb07f80 fix(lark): address managed integration review feedback`

## 主要改动

### 1. Managed Gateway lark-cli

- `POST /api/integrations/lark/install` 现在会先尝试安装/更新 DeerFlow-managed Gateway `lark-cli`：
  `.deer-flow/integrations/lark-cli/gateway-cli`
- `_resolve_lark_cli_path()` 优先使用 DeerFlow-managed CLI，其次才 fallback 到系统 PATH。
- 这样用户不需要在 Gateway/container 里手动执行 `npm install -g @larksuite/cli`。
- local sandbox 的 Lark env overlay 会把 managed CLI bin 注入 PATH，让本地对话执行也能复用托管 CLI。

### 2. 版本对齐

- skill pack 下载版本改为对齐当前 managed/runtime `lark-cli --version`。
- 避免出现 UI 重装后 skill pack 追到 latest，但实际执行 CLI 仍是旧版本的错位。
- UI 文案也改成：admin reinstall 会更新 managed Gateway CLI 和 skill pack。

### 3. Review hardening

- `_validate_integration_id` 显式拒绝 `.` / `..`。
- Lark skill pack post-install backup cleanup 改为 best-effort，避免清理旧 backup 失败导致已成功安装被报失败。
- SettingsDialog 改成单例 store/host，deep link、nav menu、command palette 不再分别 mount dialog，避免两个 Integrations card 同时跑授权轮询。
- Lark auth 自动重试加 unmount guard，避免组件卸载后继续 toast/refetch/reschedule。
- Connect 流程点击时同步预开窗口，避免 stale cache 经过 `await refetch()` 后再 `window.open` 被浏览器拦截。

## 需要重点确认的 blocker

对话里的 Lark 业务调用链路是：

Agent 读取 `lark-*` skill → 通过 sandbox `bash` 执行 `lark-cli ...`

所以 Gateway managed CLI 只能覆盖：

- Settings 授权
- Gateway 状态探测
- local sandbox PATH 注入

但 **remote/AIO sandbox 里实际执行业务命令时，`lark-cli` 必须存在于 sandbox 容器内**。

因此我认为这应该作为 release blocker：

- 官方/默认 AIO sandbox image 必须内置 `@larksuite/cli`
- 或者默认/推荐 `SANDBOX_IMAGE` 必须切到一个已发布的 Lark-enabled AIO sandbox image

否则用户可以在 UI 完成安装和授权，但 remote/AIO 对话执行仍可能失败：

`lark-cli: command not found`

当前 repo 只发布 backend/frontend 镜像，不发布 AIO sandbox 镜像，所以我没有直接改成一个未确认存在的 sandbox tag，避免部署拉不到镜像。这里需要维护者确认默认 `all-in-one-sandbox` 是否已经包含 `lark-cli`，或者提供/发布对应 Lark-enabled sandbox image。

## 测试

已验证：

- `cd backend && uv run ruff check packages/harness/deerflow/integrations/lark_cli.py app/gateway/routers/integrations.py tests/test_lark_cli_integration.py`
- `cd backend && uv run pytest tests/test_lark_cli_integration.py tests/test_paths_user_isolation.py -q` (`82 passed`)
- `cd frontend && PATH=/Users/bytedance/.nvm/versions/node/v22.21.1/bin:$PATH pnpm check`
- `cd frontend && PATH=/Users/bytedance/.nvm/versions/node/v22.21.1/bin:$PATH pnpm exec rstest run tests/unit/components/workspace/settings/settings-dialog-store.test.ts`
- `cd frontend && PATH=/Users/bytedance/.nvm/versions/node/v22.21.1/bin:$PATH pnpm exec rstest run tests/unit/core/integrations/lark/api.test.ts`
- `git diff --check`

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,1201 @@
...
+        data_dir = lark_cli_data_dir(user_id)
+        config_dir.mkdir(parents=True, exist_ok=True)
+        data_dir.mkdir(parents=True, exist_ok=True)
+    overlay = {
+        "LARKSUITE_CLI_CONFIG_DIR": str(config_dir),
+        "LARKSUITE_CLI_DATA_DIR": str(data_dir),
+        "LARKSUITE_CLI_NO_UPDATE_NOTIFIER": "1",
+        "LARKSUITE_CLI_NO_SKILLS_NOTIFIER": "1",
+    }
+    if not sandbox_paths and _lark_cli_managed_path() is not None:
```
willem-bd, line 545: **[Needs validation] `lark-cli` may not be on PATH inside the AIO sandbox**

For `sandbox_paths=True` (AIO sandbox), this overlay sets `LARKSUITE_CLI_CONFIG_DIR/DATA_DIR` to the mounted `/mnt/integrations/lark-cli/{config,data}` dirs (correctly mounted by `_get_lark_cli_runtime_mounts`), but intentionally does not touch `PATH`. The **local** sandbox is fine — commands run on-host via `subprocess` and the Gateway image now ships `lark-cli` (plus the managed `gateway-cli` dir, preferred by `_resolve_lark_cli_path`).

But the AIO sandbox runs inside the prebuilt `all-in-one-sandbox:latest` image, which this PR does not modify and which is not shown to contain `@larksuite/cli`. If that image lacks `lark-cli`, every `lark-*` skill fails at runtime in AIO mode (the skills assume `lark-cli` exists; the injected `lark-shared` guidance only covers auth, not install). Can you confirm the AIO image ships `lark-cli`, or else mount the managed binary and prepend it to `PATH` here for the `sandbox_paths=True` branch?

backend/app/gateway/routers/integrations.py
```diff
@@ -0,0 +1,319 @@
...
+    except Exception as e:
+        logger.error("Failed to get Lark integration status: %s", e, exc_info=True)
+        raise HTTPException(status_code=500, detail="Failed to get Lark integration status.")
+
+
+@router.post("/lark/install", response_model=LarkInstallResponse, summary="Install Lark/Feishu Skill Pack")
+async def install_lark(request: Request, config: AppConfig = Depends(get_config)) -> LarkInstallResponse:
+    await require_admin_user(request, detail=_ADMIN_REQUIRED_DETAIL)
+    try:
+        user_id = get_effective_user_id()
```
willem-bd: **[Design] Per-user install + admin gate blocks non-admin users**

`install_lark_integration` writes the skill pack into `get_effective_user_id()`'s namespace (`.deer-flow/users/{this_admin}/skills/integrations/...`), and install is admin-gated. A non-admin user's `GET /status` reports `installed=False`, and the frontend's `authDisabled = !data?.installed` blocks their Connect button — but they can't install (403). So in multi-user deployments only admin users can use Lark, which contradicts the PR/README framing of "admin installs once, each user connects their own account."

This is fine for the common single-user / single-admin deployment. If multi-user is intended, the pack needs to land in a shared location, or install needs to be a global admin action. Worth reconciling the docs with the per-user implementation either way.

backend/packages/harness/deerflow/community/aio_sandbox/remote_backend.py
```diff
@@ -29,6 +29,33 @@
 
 logger = logging.getLogger(__name__)
 
+_PROVISIONER_EXTRA_MOUNT_PATHS = {
```
willem-bd, line 32: **[Maintainability] Two allowlists must stay in sync**

This `_PROVISIONER_EXTRA_MOUNT_PATHS` set is a duplicate of `ALLOWED_EXTRA_MOUNT_PATHS` in `docker/provisioner/app.py`. If one is updated without the other, a mount is silently dropped (filtered out here, or rejected by the provisioner) with no test catching the drift. Consider sharing one constant, or add a test asserting the two sets are equal.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,1201 @@
...
+
+def _tenant_brand(result: dict[str, Any]) -> str | None:
+    user_info = result.get("user_info")
+    if not isinstance(user_info, dict):
+        return None
+    brand = str(user_info.get("tenant_brand") or "").strip().lower()
+    return brand if brand in {"feishu", "lark"} else None
+
+
+def _save_lark_app_config_with_cli(user_id: str, *, app_id: str, app_secret: str, brand: str) -> None:
```
willem-bd, line 1208: **[Security] App secret persisted to disk — verify per-user dir perms**

Good: the secret is passed via stdin (`--app-secret-stdin`) and never returned to the browser. But `lark-cli config init` then writes `appSecret` in plaintext to `…/users/{user}/integrations/lark-cli/config/config.json`. On a shared Gateway host, a world-readable file there would leak the secret across local users. This PR doesn't chmod the integration config dir. Can you confirm the `…/users/{user}/` tree is created `0700` (the existing user-isolation layout likely does, but a secret now lives there, so worth an explicit check/guard)?

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,1201 @@
...
+    staging_target.mkdir(parents=True, exist_ok=True)
+
+    backup: Path | None = None
+    try:
+        with zipfile.ZipFile(archive_path, "r") as zf:
+            archive_version = version or _infer_lark_archive_version(zf)
+            extracted = _extract_lark_skills(zf, staging_target)
+        _validate_extracted_lark_skills(staging_target, extracted)
+        _append_deerflow_lark_shared_guidance(staging_target)
+        content_sha = _content_sha256(staging_target, extracted)
```
willem-bd, line 1501: **[Accuracy] `content_sha256` is computed after injecting DeerFlow guidance**

`_append_deerflow_lark_shared_guidance(staging_target)` runs on the line above, so the manifest `content_sha256` hashes DeerFlow-modified content, not the upstream archive. It still achieves the stated goal (detecting content change across reinstalls), but the module docstring frames the hash as integrity over "the extracted skill tree," which is slightly overstated. Either hash before the append, or tweak the docstring to note the hash includes the injected guidance.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,1201 @@
...
+    """Complete a Lark device authorization flow after the user approves it."""
+    device_code = device_code.strip()
+    if not device_code:
+        raise ValueError("device_code is required.")
+
+    path = _require_lark_cli_path()
+    _run_lark_cli_json(
+        [path, "auth", "login", "--device-code", device_code, "--json"],
+        user_id=user_id,
+        timeout=45,
```
willem-bd: **[UX] Auto-detection latency can be ~45s per cycle**

`complete_lark_auth` blocks on `lark-cli auth login --device-code …` for up to 45s, then runs a `--verify` status probe (8s). The frontend only reschedules its 1.5s retry from `onSuccess`/`onError`, so each pending poll cycle can take the full ~45s before retrying. The UI copy ("DeerFlow will detect completion automatically") implies near-real-time. If lark-cli doesn't return quickly on pending, perceived latency will be tens of seconds; worth confirming against real lark-cli behavior (the manual "I completed authorization" button is a good fallback).

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,1201 @@
...
+        _append_deerflow_lark_shared_guidance(staging_target)
+        content_sha = _content_sha256(staging_target, extracted)
+        _write_manifest(staging_target, extracted, version=archive_version, content_sha256=content_sha)
+        make_skill_tree_sandbox_readable(staging_target)
+
+        if target.exists():
+            backup = parent / f".replacing-{INTEGRATION_ID}-{os.getpid()}"
+            if backup.exists():
+                shutil.rmtree(backup)
+            target.rename(backup)
```
willem-bd, line 1509: **[Robustness] Concurrent reinstalls can race**

Two simultaneous admin reinstalls use pid-based backup names (good), but the `target.rename(backup)` / `staging_target.rename(target)` sequence can interleave between processes: one process can see `target` missing (skip backup) while another has already renamed it, yielding last-writer-wins and a lingering `.replacing-…` backup beside the target. Rare (admin-only, manual), but a per-install lockfile would make it deterministic.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,1201 @@
...
+        app_configured=bool(app_config["configured"]),
+        app_id=app_config["app_id"],
+        app_brand=app_config["brand"],
+        skills_expected=len(LARK_SKILL_NAMES),
+        skills_installed=len(installed_skills),
+        installed_skills=installed_skills,
+        enabled_skills=enabled_skills,
+        install_path=str(root),
+        cli=cli,
+        auth=probe_lark_auth(user_id, verify=verify_auth),
```
willem-bd, line 721: **[Nit] Status "authenticated" reflects token presence, not validity**

The frequently-polled `/status` endpoint calls `probe_lark_auth(verify=False)`, so `auth.status == "authenticated"` means a token exists locally, not that it's live. An expired token still shows "Connected" in the card until a `--verify` runs (only on `auth/complete`). Reasonable for a cheap status probe — just flagging that the card can be stale; consider a periodic verify or wording that hedges.

willem-bd (commented):
Review of the Lark CLI integration. Strong, security-conscious implementation with good test coverage and careful hardening on the archive-install and provisioner paths. Leaving inline notes below.

Top items to address before merge:
- **[Needs validation]** Confirm `lark-cli` is present in the AIO sandbox image (otherwise lark skills are non-functional in AIO mode).
- **[Design]** Reconcile the "admin installs once, each user connects" framing with the per-user, admin-gated install.

The rest are low-severity polish. No DB migration and state lives in the existing per-user layout, as noted.

fancyboi999 (changes requested):
Thanks @18062706139fcz. I found one issue that should be addressed before this is ready.

[P1] Avoid adding the per-user custom skills mount twice

- Location: `backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py:321-324` (`_get_extra_mounts` appends `_get_user_skill_mounts` after `_get_skills_mounts`)
- Problem: provisioner-backed AIO sandbox creation fails before the sandbox starts because the combined mount list contains `/mnt/skills/custom` twice.
- Evidence: `_get_skills_mounts()` already unconditionally returns the current user's custom directory at `/mnt/skills/custom` (lines 388-407), while the new `_get_user_skill_mounts()` returns the same target again (lines 441-443). `RemoteSandboxBackend` forwards both because `/mnt/skills/custom` is allowlisted, and `docker/provisioner/app.py::_validated_extra_mounts()` rejects duplicate container paths with HTTP 400. The helper tests exercise the mount producers separately, so the duplicate in the composed list is currently missed.
- Suggested fix: have `_get_user_skill_mounts()` add only the global integration-skills mount (or fold that mount into `_get_skills_mounts()`), and defensively ensure the composed list has unique container paths before it reaches either backend.
- Test: build the full `_get_extra_mounts()` list with Lark runtime present, pass its provisioner-filtered payload through `_validated_extra_mounts()`, and assert every container path is unique while the integration/config/data/runtime mounts remain present.


Reviewed head SHA: `ccfdf8fa95231994cffc83add380654398f79492`

zhfeng:
Following the `lark-cli` availability discussion - the Gateway side is cleanly out of the box (image-baked `@larksuite/cli` + managed npm install with offline fallback), but the AIO/K8s sandbox execution path still depends on either:

- (a) downloading Linux binaries from GitHub at install time into `.deer-flow/integrations/lark-cli/sandbox-cli` and mounting them via hostPath/PVC (`_ensure_managed_sandbox_lark_cli` + `_get_lark_cli_runtime_mounts`), or
- (b) baking `lark-cli` into a custom `all-in-one-sandbox` derivative.

Both have friction: (a) couples to `DEER_FLOW_HOST_BASE_DIR` + a shared PVC and has no offline fallback - it hard-fails the whole install if GitHub is unreachable; (b) forks the upstream sandbox image and re-couples lark-cli versioning to it.

A third option that fits the K8s provisioner cleanly: **an init container that provisions `lark-cli` into a shared `emptyDir`.**

## Pattern A - init container as binary provider

- The provisioner adds an init container whose image has `@larksuite/cli` baked in, plus an `emptyDir` volume mounted at `/mnt/integrations/lark-cli/runtime` in both the init container and the main sandbox container.
- The init container copies the `bin/lark-cli` launcher + `linux-amd64/` + `linux-arm64/` layout (the one `_validate_lark_cli_sandbox_runtime` already enforces) into the shared volume, then exits.
- The main sandbox container then sees `lark-cli` at `/mnt/integrations/lark-cli/runtime/bin/lark-cli` - exactly where `lark_cli_env_overlay(sandbox_paths=True)` already points `PATH`.

Why this is attractive:

- **Zero gateway-side and zero skill changes** - the PATH-injection / mount contract is unchanged; only the *source* of the mounted binary moves from hostPath to an init image.
- Removes the `DEER_FLOW_HOST_BASE_DIR` + shared-PVC coupling for the runtime binary (the `extra_mounts`/provisioner surface around @fancyboi999's P1).
- Removes the install-time GitHub download for the sandbox binary, so install no longer hard-fails offline (and `_ensure_managed_sandbox_lark_cli` can be skipped for K8s).
- Decouples `lark-cli` versioning from the sandbox image - bump the init image to upgrade, keep `all-in-one-sandbox` upstream-tracked.

Cost: a maintained "lark-cli init" image + a provisioner change to optionally add the init container + `emptyDir` (signaled by a flag on `CreateSandboxRequest`) instead of the runtime hostPath mount.

## Pattern B - long-running sidecar as credential broker (longer term)

This lines up with the `lark_cli_env_overlay` docstring: *"Auth Proxy support can later replace these directories for sandbox execution without changing the status API contract."* A sidecar in the same Pod holds `lark-cli` + the per-user OAuth credentials and serves it over `localhost`; the sandbox container has a small `lark-cli` shim on PATH that forwards to it.

Beyond Pattern A, this stops mounting plaintext `LARKSUITE_CLI_CONFIG_DIR` / `DATA_DIR` (app secret + OAuth tokens) into the sandbox container - addressing the credential-bleeding concern @willem-bd raised - and enables short-lived/refreshed tokens. Cost is larger: the sidecar HTTP contract, the shim, and broker semantics. Closer to a v2 of the integration.

## Trade-off summary

|  | Bake into sandbox image | Init container (A) | Sidecar broker (B) |
|---|---|---|---|
| Provisioner changes | none | small | larger |
| lark-cli version decoupled from sandbox image | no | yes | yes |
| Keeps sandbox image upstream-tracked | no | yes | yes |
| Avoids install-time GitHub download | yes | yes | yes |
| Avoids mounting plaintext creds in sandbox | no | no | yes |
| Needs a shim in the sandbox | no | no | yes |

## Suggestion

For the K8s default, Pattern A looks like a near-pure win: it directly resolves the "[Needs validation] is `lark-cli` present in the AIO image" blocker without requiring a custom sandbox image, and it removes the host-path/PVC fragility - while leaving the gateway/skill contract untouched. It might be worth considering before merge, or as a fast-follow. Happy to help prototype the provisioner change if there's interest.

(Also still think the highest-leverage near-term fix is surfacing a sandbox-runtime readiness signal in `/status` so a green UI can't hide a `command not found` at chat time - independent of which provisioning pattern wins.)

WillemJiang:
@18062706139fcz  Could you take a look at the review comments from @zhfeng? I think we can create a follow-up issue to keep track of the installation issue.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
```diff
@@ -839,6 +874,59 @@ def _get_skills_mounts(*, user_id: str | None = None) -> list[tuple[str, str, bo
...
+        ``lark-cli`` inside the sandbox, so those same directories must be
+        mounted into the container or the CLI sees a separate unauthenticated
+        profile.
+        """
+        try:
+            paths = get_paths()
+            effective_user_id = AioSandboxProvider._effective_acquire_user_id(user_id)
+            ensure_lark_cli_credential_tree(effective_user_id, paths=paths)
+            mounts = [
+                (paths.host_user_integration_config_dir(effective_user_id, LARK_CLI_INTEGRATION_ID), LARK_CLI_SANDBOX_CONFIG_DIR, False),
```
willem-bd: **[Security] Lark app_secret + OAuth tokens are readable by arbitrary sandbox processes**

These two mounts bind the per-user `config` dir (which holds `appSecret` in `config.json` - written by `lark-cli config init`) and the `data` dir (OAuth tokens) into the AIO sandbox **read-write** (`read_only=False`). Host-side `0700/0600` hardening from `ensure_lark_cli_credential_tree` does not help inside the container: `bash_tool` gives the agent arbitrary command execution there, so any code the agent runs - including code reached via prompt-injection in a tool result - can `cat /mnt/integrations/lark-cli/config/config.json` and exfiltrate the app secret, or read/refresh tokens from `/mnt/integrations/lark-cli/data`. The trusted `lark-cli` binary itself isn't the threat; the co-located process surface is. A compromised tool output is sufficient.

The `lark_cli_env_overlay` docstring already points at a future "Auth Proxy" to replace these dirs for sandbox execution. Until that lands, could you:
 (a) document this as an explicit sandbox trust boundary in `backend/AGENTS.md` / README - the README currently only says "secret stays on the Gateway side", which is true for the browser but not for sandbox processes;
 (b) track the auth-proxy follow-up so this exposure isn't the steady state?

If `config` (app_secret) can be mounted read-only and only `data` (refreshable tokens) stays read-write, that shrinks the exposed long-lived secret surface. Medium-high.

backend/packages/harness/deerflow/integrations/lark_cli.py
```diff
@@ -0,0 +1,1532 @@
...
+    install_root = lark_cli_managed_gateway_dir()
+    install_root.mkdir(parents=True, exist_ok=True)
+    result = subprocess.run(
+        [
+            npm,
+            "install",
+            "--prefix",
+            str(install_root),
+            "--no-audit",
+            "--no-fund",
```
willem-bd, line 1073: **[Security] `npm install` runs the package's install scripts on the Gateway**

The managed-gateway CLI install runs `npm install --prefix <root> --no-audit --no-fund @larksuite/cli@<ver>` with no `--ignore-scripts`, so the package's `preinstall`/`postinstall` scripts execute with Gateway privileges during an admin-triggered install. `@larksuite/cli` is the official/trusted package (and the Dockerfile install is consistent), but this is the supply-chain blast radius: a compromise of the package or a transitive dep runs arbitrary code on the Gateway.

If `--ignore-scripts` is feasible for this package, prefer it; if the CLI needs postinstall to fetch platform binaries (likely), leave the behavior but add a one-line comment noting install scripts execute with Gateway privileges so the tradeoff is explicit. Low severity given the official package + admin gating.

backend/app/gateway/routers/integrations.py
```diff
@@ -0,0 +1,330 @@
...
+        latest_available_version=status.latest_available_version,
+        runtime_version_mismatch=status.runtime_version_mismatch,
+        app_configured=status.app_configured,
+        app_id=status.app_id,
+        app_brand=status.app_brand,
+        skills_expected=status.skills_expected,
+        skills_installed=status.skills_installed,
+        installed_skills=list(status.installed_skills),
+        enabled_skills=list(status.enabled_skills),
+        install_path=status.install_path,
```
willem-bd: **[Security] `install_path` and `cli.path` (host filesystem paths) returned to any authenticated user by `GET /lark/status`**

`get_lark_status` is not admin-gated (only `/install` is), and `_status_to_response` echoes `install_path` (host absolute path, e.g. `/home/deer-flow/.deer-flow/integrations/skills/lark-cli`) and `cli.path` in the 200 body. Any authenticated user can map the Gateway's host directory layout this way. The UI doesn't render these fields (verified), but the API still discloses them.

Consider omitting `install_path`/`cli.path` for non-admin callers, or mapping `install_path` to a virtual/relative path. Low severity - info disclosure to authenticated users only.

willem-bd (commented):
Security findings from review (3 inline comments below). All are non-blocking for the core integration but worth a decision before merge.

- **S1 (Medium-High)** `aio_sandbox_provider.py` - the per-user Lark `config` (app_secret) and `data` (OAuth tokens) dirs are mounted **read-write** into the AIO sandbox, where `bash_tool` gives the agent arbitrary command execution. Prompt-injection-driven code can exfiltrate the app secret / tokens. Host-side `0700/0600` doesn't help inside the container. Recommend documenting this as a sandbox trust boundary and tracking the documented "Auth Proxy" follow-up.
- **S2 (Low)** `lark_cli.py` - `npm install` of `@larksuite/cli` runs without `--ignore-scripts`, so package install scripts execute with Gateway privileges. Official package + admin-gated, but the supply-chain tradeoff should be explicit.
- **S3 (Low)** `integrations.py` - `GET /lark/status` (not admin-gated) returns `install_path` and `cli.path` (host filesystem paths) to any authenticated user. UI doesn't render them; consider omitting for non-admins.

Diffed against the 25 existing comments on this PR to avoid duplicate threads; these three surfaces are not already raised.

18062706139fcz:
Thanks @zhfeng, this is a great breakdown — agreed on both counts.

**Pattern A (init container + emptyDir) for now.** It's the cleanest fit for the K8s default: it removes the install-time GitHub download and the `DEER_FLOW_HOST_BASE_DIR`/PVC coupling, keeps the gateway/skill PATH contract untouched, and decouples `lark-cli` versioning from the upstream `all-in-one-sandbox` image. I'll take this one and switch the current implementation over to Pattern A (init container + `emptyDir` + a flag on `CreateSandboxRequest`) — please hold on the review until I've pushed that change.

**Pattern B (sidecar credential broker) as a follow-up.** It's the right long-term direction since it stops mounting plaintext config/data into the sandbox and addresses the credential-bleeding concern @willem-bd raised — but the shim + broker contract is really a v2 of the integration. Tracked in #4338.

**Agreed on the `/status` sandbox-runtime readiness signal** regardless of which pattern wins — surfacing it so a green UI can't hide a `command not found` at chat time is the highest-leverage near-term fix. I'll fold that into this line of work.

zhfeng (approved):
Please fix the lint check failures and other things LGTM

WillemJiang (approved):

PR #3976, opened 2026-07-07
Fix/card tool message bug

### LINKED ISSUE #3975: [bug] Human Input Card 在上下文压缩后渲染出现问题
### Before you start

### Problem summary

之前的PR #3956 引入了Human Input Card（目前仅针对ask_clarification），但当上下文压缩的时候，前端UI上渲染不出来卡片

### Affected area(s)

## 问题总结

Human Input Card 在上下文压缩后可能出现历史恢复不完整：

1. `ask_clarification` request card 可能丢失。
2. request card 恢复后，卡片内的 `answeredResponse` 可能丢失，导致已回答卡片重新变成未回答状态。

## 根因

`ask_clarification` 的 request payload 由 `ClarificationMiddleware` 生成 `ToolMessage.artifact.human_input`。middleware 通过 `Command(goto=END)` 短路工具执行时，LangChain 不会触发正常的 `on_tool_end`，因此 `RunJournal` 原本可能没有把该 `ToolMessage` 写入 `run_events`。

`answeredResponse` 丢失是另一条链路：前端提交回答时会发送一个 `hide_from_ui: true` 的 `HumanMessage`，其 `additional_kwargs.human_input_response` 承载结构化回答。压缩前 checkpoint 里还能看到这条 hidden reply；压缩后 checkpoint 只保留 tail。如果这条 hidden reply 没有写入 `run_events`，前端就只能恢复 request card，无法恢复 answered 状态。

### Expected behavior

预期是所见即所得，压缩前后不应该变化。

### Steps to reproduce

1. 触发ask_clarification
2. 长任务触发上下文压缩

### Relevant logs

```shell
无
```

Other

### Operating system

Other

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Support bundle summary

_No response_

## 修复方向

- 对 `ask_clarification` request `ToolMessage`：通过 `RunJournal` root `on_chain_end` 做窄白名单最终 reconciliation，补写缺失的 `llm.tool.result`。
- 对 Human Input Card hidden reply：允许 `RunJournal` 持久化白名单 source 的 hidden `human_input_response`，目前只开放 `ask_clarification`，仍过滤普通内部 hidden context。

## 其他问题

目前发现上下文压缩后，UI显示上还是出现乱套的问题，不过非本次、上次Card PR引入，这个issue不讨论，后续可以单独开PR处理。

AnoobFeng:
Fixes #3975 Closes #3975 

## Why

Human Input Cards losing their answered state after context compaction. The request card can be recovered from `run_events`, but the user’s hidden answer was not persisted, so `answeredResponse` became `null` after compaction.

## Root Cause

`RunJournal.on_chat_model_start()` previously persisted only UI-visible `HumanMessage`s and filtered out every message with `hide_from_ui: true`. Human Input Card answers are genuine user input, but they are intentionally submitted as hidden messages to avoid rendering bridge text as a normal user bubble. As a result, they were missing from `run_events`.

Before compaction, the checkpoint could still contain the hidden reply. After compaction, checkpoint messages are only a retained tail, so the frontend could no longer derive the answered card state.

## What changed

- Added narrow allowlist-based persistence in `RunJournal`:
  - Visible `HumanMessage`s are still persisted as before.
  - Hidden `HumanMessage`s are persisted only when they contain a valid `human_input_response` and the response `source` is allowlisted.
  - The initial allowlist contains only `ask_clarification`.
- Kept generic internal hidden messages, such as summaries, reminders, and dynamic context, filtered out.
- Added regression tests for:
  - hidden `ask_clarification` responses being persisted as `llm.human.input`;
  - the latest hidden response winning over an older visible prompt in the prompt history;
  - non-allowlisted hidden responses staying filtered.
- Updated `backend/AGENTS.md` with the Human Input Card request/response persistence semantics.

## Validation

```shell
uv run --project backend pytest backend/tests/test_run_journal.py backend/tests/test_human_input.py -q
uv run --project backend ruff check backend/packages/harness/deerflow/runtime/journal.py backend/tests/test_run_journal.py
```

## AI assistance

**Tool(s) used:** Codex


backend/packages/harness/deerflow/runtime/journal.py
```diff
@@ -208,7 +221,7 @@ def on_chat_model_start(
         if caller == "lead_agent" and not self._first_human_msg and messages:
             for batch in reversed(messages):
                 for m in reversed(batch):
-                    if _is_user_visible_human_message(m):
+                    if _should_persist_human_input_message(m):
```
willem-bd, line 224: **Observation (not a blocker):** `_should_persist_human_input_message` returns `True` for hidden `ask_clarification` replies, and the reversed scan here treats them as persistable — so `set_first_human_message(m.text)` (line 225) can now record the hidden reply string (e.g. `'For your clarification "...", my answer is: staging'`) as the run's `first_human_msg` for a clarification-continuation run.

`_first_human_msg` flows into `get_completion_data()["first_human_message"]` (line 718), which is persisted into the `runs` table (`persistence/run/sql.py`) and surfaced in the console runs dashboard / progress snapshots. So post-clarification runs will display the hidden reply there instead of the original visible user prompt.

The new test `test_hidden_human_input_response_wins_over_older_visible_prompt` asserts this, so it looks intentional — but in case the intent was only to persist the answered-card state to `run_events` (which the `llm.human.input` `_put` already does on the next lines) without also overwriting run-level `first_human_message`, splitting the two (emit the event for hidden replies, but keep `_first_human_msg` visible-only) would preserve the dashboard semantics. If overwriting is the desired behavior, feel free to dismiss.

willem-bd (commented):
Reviewed via automated pass (line-by-line, removed-behavior, cross-file, reuse/simplification/efficiency, altitude, conventions). No correctness, cleanup, or conventions finding survived verification — reconciliation ordering, identity-based dedup, allowlist gating, and `_record_message_summary` caller semantics are all preserved. One behavioral observation below.

backend/packages/harness/deerflow/runtime/journal.py
AnoobFeng: Thanks for the observation. I think keeping this behavior is meaningful. The key question is how we define the card UX after submission:

1. Keep the card after submission.

   In this model, the original card remains the final UI representation of the user’s answer. The hidden clarification reply mainly exists as metadata to mark that card as answered. If we choose this direction, then it may be reasonable to keep `first_human_message` visible-only, because the card itself is the user-facing answer.

2. Do not keep the card after submission.

   In this model, the card is only an input surface. Once the user submits it, the action becomes a normal human input continuation and starts a real follow-up run. Under this semantics, the hidden `human_input_response` is the actual user-authored input for that run, so persisting it as `first_human_message` is meaningful and accurate. Otherwise the run would either have no human input summary or could be incorrectly attributed to an older visible prompt from checkpoint history.

WillemJiang (approved):

PR #3978, opened 2026-07-07
feat(memory): conversational memory tools + vector retrieval

NextE-Moffatt:
## Summary

Conversational memory control tools + vector-based semantic retrieval for fact ranking.

### New Tools
- **`remember`** — lets users explicitly persist facts during conversation (immediate write, bypasses debounce queue)
- **`forget`** — delete facts by ID or semantic search with user confirmation on multiple matches
- **`search_memory`** — agent proactively searches memory using combined keyword + vector scoring

### Vector Retrieval
- **`embeddings.py`** — embedding computation supporting API models (OpenAI-compatible `/v1/embeddings`) with zero-dependency char n-gram fallback
- **`search.py`** — shared ranking utilities: keyword scoring, vector similarity, combined ranking, formatted output
- **Save-time embedding** — `updater.py` computes embeddings when facts are created (both LLM-extracted and manual)
- **Injection-time vector ranking** — `format_memory_for_injection` accepts optional `query` param; ranks facts by cosine similarity to current conversation instead of static confidence
- **Config** — new `memory.embedding_model` field in `MemoryConfig`

### Refactoring
- `forget_tool` refactored to use shared `search.py` utilities (removed duplicate `_find_facts_by_content`)
- Call chain wired through: `DynamicContextMiddleware` → `_get_memory_context()` → `format_memory_for_injection(query=...)`

### Files Changed
```
12 files changed, 694 insertions(+), 21 deletions(-)

A  embeddings.py          ← embedding computation
A  search.py               ← shared search/ranking
A  remember_tool.py        ← remember tool
A  forget_tool.py          ← forget tool (refactored)
A  search_memory_tool.py   ← search tool
M  updater.py              ← save-time embedding
M  prompt.py               ← vector ranking in injection
M  dynamic_context_middleware.py  ← query extraction
M  memory_config.py        ← embedding_model field
M  __init__.py / tools.py  ← registration
```

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/backend/packages/harness/deerflow/agents/lead_agent/agent.py
+++ b/backend/packages/harness/deerflow/agents/lead_agent/agent.py
@@ -29,7 +29,7 @@ from langchain_core.runnables import RunnableConfig
 from deerflow.agents.lead_agent.prompt import apply_prompt_template
 from deerflow.agents.memory.summarization_hook import memory_flush_hook
 from deerflow.agents.middlewares.clarification_middleware import ClarificationMiddleware
-from deerflow.agents.middlewares.loop_detection_middleware import LoopDetectionMiddleware
+from deerflow.agents.middlewares.resource_guard_middleware import ResourceGuardMiddleware
 from deerflow.agents.middlewares.memory_middleware import MemoryMiddleware
 from deerflow.agents.middlewares.safety_finish_reason_middleware import SafetyFinishReasonMiddleware
 from deerflow.agents.middlewares.subagent_limit_middleware import SubagentLimitMiddleware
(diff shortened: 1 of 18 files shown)


WillemJiang:
@NextE-Moffatt, thanks for your contribution. Please click the CLA assistant button to sign the CLA first.  
Please address the review feedback and solve the conflict with the main branch.

xg-gh-25:
This is a really clean architecture for conversational memory control: giving users (`remember`/`forget`) and agents (`search_memory`) explicit handles on what gets persisted, while the system still does autonomic extraction in the background. The split between immediate writes (user-driven `remember`) and debounced extraction (LLM-driven) is smart — it means users can force-persist facts without waiting for the next extraction cycle.

The vector-ranking-at-injection-time part is especially interesting: instead of injecting facts in static confidence order, you're ranking them by cosine similarity to the *current conversation turn*. That's a lightweight way to get query-aware retrieval without needing a full semantic search index at read time — you just need embeddings computed at save time, which `updater.py` already does.

**One edge case to watch:** what happens when `format_memory_for_injection(query=...)` is called with a query that doesn't semantically match *any* facts? Do you fall back to confidence-only ranking, or does the injection end up empty? (The PR doesn't show the fallback logic, so it might already be there.) We've seen cases where vector-only ranking can miss high-confidence structural facts (e.g. "user's timezone is UTC-8") because the query isn't semantically close to them, even though they're always-relevant.

**The `embeddings.py` zero-dependency fallback** (char n-grams when no API is available) is a nice touch — it means the system degrades gracefully instead of failing silently if the embedding model is misconfigured or unreachable.

**Refactoring `forget_tool` to use shared `search.py`** is good hygiene — removing the duplicate `_find_facts_by_content` means there's one canonical "find facts by semantic search" path, which makes future changes easier (e.g. adding fuzzy matching or threshold tuning).

We've been collecting similar memory-control patterns (explicit user commands vs. autonomic extraction, vector retrieval for fact ranking, dedup/eviction strategies) in **[T-MEM: Durable Memory Systems](https://github.com/xg-gh-25/SwarmAI/discussions/238)**. This PR is a great example of how to give both users and agents control over the same memory surface without them stepping on each other.

---
*Patterns for conversational memory with user and agent control. [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-MEM: Durable Memory Systems](https://github.com/xg-gh-25/SwarmAI/discussions/238)*

xg-gh-25:
The dual-mode memory design here (conversational + semantic vector retrieval) maps well to long-running agentic workflows where context grows beyond the prompt window.

Two observations:
1. **Compaction trigger**: Line 84 checks `len(messages) > threshold`, but conversational bloat isn't always linear — have you considered tracking *token density* instead? A 50-message thread with short ACKs differs from 50 deep exchanges.
2. **Vector stale-read risk**: If `retrieve_relevant_context()` runs on an outdated index while new messages stream in, you might inject out-of-order summaries. Does the retrieval snapshot the index at call-time, or does it see live writes?

The `MemoryMode` enum is a nice escape hatch for debugging. In production, surfacing *why* a fallback happened (quota vs. latency vs. relevance score) helps operators tune thresholds without blind iteration.

---
*Part of the [Agent Memory Systems](https://github.com/xg-gh-25/SwarmAI/discussions/12) discussion on [SwarmAI](https://github.com/xg-gh-25/SwarmAI).*

PR #3980, opened 2026-07-07
fix(subagents): unify guardrail caps on additive stop_reason + token_budget (#3875 Phase 2)

hata33:
## What & why

Phase 2 of #3875 — stops subagents looping/burning tokens on trivial tasks by surfacing *why* a run ended early, and adds the missing **token-budget** backstop.

Two guardrail axes can end a subagent run early:

| Axis | Mechanism | Raises? |
|------|-----------|---------|
| **Turn** | `recursion_limit == max_turns` → `GraphRecursionError` | yes |
| **Token** | `TokenBudgetMiddleware` hard-stop | **no** — strips tool_calls, forces a final answer |

Both now report the cap on **one additive `subagent_stop_reason` field** (`token_capped` / `turn_capped` / `loop_capped`) instead of a status enum.

## Course-corrects Phase 1 (#3949)

#3949 shipped the turn-budget cap as a `max_turns_reached` **status enum**. Per @ggnnggez's Phase 2 design, this replaces that enum with an **optional `stop_reason` field**:

> a new enum value breaks v1 consumers; an additive field is ignored by older frontends and ledger readers.

So this PR **removes** `max_turns_reached` and `SubagentStatus.MAX_TURNS_REACHED`, and the cap reason now travels on `subagent_stop_reason`. A capped run that produced a final answer stays `status=completed`; a capped run with no usable output is `status=failed`. Either way the lead (and UI) can tell "out of budget" from "broken subagent" without parsing result text.

## What changed

- **`subagents.token_budget` config** (default: enabled, **2,000,000 tokens**, warn 0.7) with per-agent override. `TokenBudgetMiddleware` is now attached in `build_subagent_runtime_middlewares`, so the cost-ceiling backstop engages for **every** subagent (built-in + custom) unless explicitly disabled.
- **`executor.py`**: reads `consume_stop_reason()` on normal completion → stamps `completed + token_capped` when the budget fired; on `GraphRecursionError` recovers the last `AIMessage` partial (`completed + turn_capped`) or, if nothing usable survived, `failed + turn_capped`. `SubagentResult` gains `stop_reason`.
- **Contract** (`status_contract.py` / `contracts/subagent_status_contract.json` v2 / `frontend/.../subtask-result.ts`): additive `subagent_stop_reason`, pinned by `test_status_values_match_contract` + `test_stop_reason_values_match_contract`.
- **`task_tool.py` + `delegation_ledger.py`**: dropped the `max_turns_reached` paths; the ledger captures `stop_reason` and renders model-facing "capped" guidance so the lead reuses a capped completion knowingly instead of mistaking it for clean.

## ⚠️ Tunable default — flagging for review

The **2,000,000-token** default is deliberately loose. It would have roughly halved the reported 4.4M pathological burn while leaving legitimate deep-research runs (`max_turns=150`, no summarization yet) room to breathe. Happy to tighten it (e.g. 1M) or gate differently per maintainer preference. Subagent summarization is the proper long-term fix and is out of scope here.

## Backward compatibility

Older frontends/ledger readers that only read `subagent_status` are unaffected — they never see the additive `subagent_stop_reason` field. The contract fixture version bumped to v2; `config_version` bumped 19 → 20.

## Tests

- Backend: `test_subagent_status_contract`, `test_delegation_ledger`, `test_subagent_executor`, `test_task_tool_core_logic`, `test_token_budget_middleware`, `test_tool_error_handling_middleware` — all green (439 in the contract-related sweep).
- Frontend: `subtask-result.test.ts` (incl. new stop_reason + contract-fixture cases) green; `pnpm check` (lint + typecheck) clean.
- `ruff format` / `ruff check` clean.

---

Part of #3875 (Phase 2). Does **not** close it — Phase 3 (subagent summarization) is still pending; this mirrors how Phase 1 (#3949) referenced the same umbrella.


frontend/src/core/tasks/subtask-result.ts
```diff
@@ -56,7 +77,6 @@ const STRUCTURED_STATUS_TO_SUBTASK: Record<string, SubtaskStatus> = {
   cancelled: "failed",
   timed_out: "failed",
   polling_timed_out: "failed",
```
willem-bd, line 89: Removing `max_turns_reached: "failed"` from this map strands historical thread checkpoints as perpetually `in_progress`.

Phase 1 (#3949, already shipped) wrote `subagent_status: "max_turns_reached"` into the `task` tool result's `additional_kwargs`, which is checkpointed in thread message history. On this build, `readStructuredStatus("max_turns_reached")` → `null` (value no longer in the map) → `parseSubtaskResult` falls into the `if (!structured)` branch, but `hasStructuredSubagentMetadata` still returns `true` (the old payload carries the other structured keys) → the card resolves to `{ status: "in_progress" }`.

Result: opening a thread that had a turn-capped subagent under the #3949 build shows a spinning "running" card forever, instead of the terminal failed pill Phase 1 produced.

Suggested fix: keep `max_turns_reached: "failed"` here as a deprecated alias so existing data still resolves to a terminal pill (and/or normalize the old value on the backend read path).

willem-bd (commented):
Automated review pass. One backward-compatibility finding below; everything else verified clean.

frontend/src/core/tasks/subtask-result.ts
hata33: Good catch — addressing both read paths since `max_turns_reached` is checkpointed in thread history from Phase 1 (#3949), so removing it stranded historical data in two places, not just the card.

Pushed in 51cbbc7e:

1. **Frontend (this line)** — kept `max_turns_reached: "failed"` as a deprecated read-only alias in `STRUCTURED_STATUS_TO_SUBTASK`. Old turns now resolve to the terminal failed pill exactly as Phase 1 rendered them, instead of spinning `in_progress`. No code path produces the value anymore — pure read-side tolerance.

2. **Backend delegation ledger** — the same root cause affected `read_subagent_result_metadata` (returns `None` for the legacy value → the durable ledger stranded the delegation as `in_progress` → the lead saw "already delegated; wait for result" for an already-terminal task). Added `_LEGACY_STATUS_NORMALIZATION` so the reader maps the legacy value to its Phase 2 cap equivalent: `completed + turn_capped` when a `result_brief` survived (preserves the partial the lead could reuse, matching Phase 1's result-bearing intent) or `failed + turn_capped` otherwise.

The producer side (`SUBAGENT_STATUS_VALUES`, `make_subagent_additional_kwargs`, the contract fixture) stays clean — `max_turns_reached` is gone from everything the backend emits; only the two readers tolerate it for historical data.

Regression tests on both sides: `test_read_subagent_result_metadata_normalizes_legacy_max_turns_reached` (backend) and "renders legacy max_turns_reached … as a terminal failed pill, not spinning in_progress" (frontend).

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -419,8 +431,19 @@ def _create_agent(self, tools: list[BaseTool] | None = None, *, deferred_setup:
...
+            app_config=app_config,
+            model_name=self.model_name,
+            lazy_init=True,
+            deferred_setup=deferred_setup,
+            agent_name=self.config.name,
+        )
+        # Capture the TokenBudgetMiddleware (if attached) so _aexecute can read
+        # consume_stop_reason after the run and surface token_capped to the
+        # lead. Duck-typed so this file needs no import of the middleware class.
+        self._token_budget_middleware = next((m for m in middlewares if hasattr(m, "consume_stop_reason")), None)
```
ggnnggez: This only captures the token-budget guard, but the v2 contract now advertises `loop_capped` as well. Per #3875 Phase 2, a loop hard-stop should also surface through `subagent_stop_reason`; today `LoopDetectionMiddleware` strips tool calls and lets the run complete, but it has no `consume_stop_reason()` producer, so the lead/ledger will see a clean `completed` result instead of `completed + loop_capped`. Could we make this collect all guard middlewares with `consume_stop_reason()` and add the symmetric producer on `LoopDetectionMiddleware`, with a real-path test asserting the final task ToolMessage carries `subagent_stop_reason=loop_capped`?

ggnnggez (commented):
Left one inline comment on the missing `loop_capped` stop-reason path from the #3875 acceptance criteria.

backend/packages/harness/deerflow/subagents/executor.py
hata33: Done in fd7f8155 — wired exactly as you described:

1. **Symmetric producer on `LoopDetectionMiddleware`** — added `_stop_reason` (BoundedDict, keyed by `run_id`) + `consume_stop_reason()`, stamped `loop_capped` in the `_apply` hard-stop path (covers both the hash-based and per-tool-frequency detectors, since they share that path). Mirrors `TokenBudgetMiddleware` 1:1; `reset()` clears it, `after_agent` does not (so the executor can read it post-run).

2. **Collect all guard middlewares** — `_token_budget_middleware = next(...)` became `_stop_reason_middlewares = [m for m in middlewares if hasattr(m, "consume_stop_reason")]`, and `_consume_token_stop_reason()` → `_consume_guard_stop_reason()` iterates all of them and returns the first non-`None`. So a future guard plugs in with no executor change.

3. **Real-path test** — `test_task_result_command_carries_loop_capped_from_real_loop_detection` drives the actual `LoopDetectionMiddleware` to a hard stop (4 identical tool calls), feeds the produced `loop_capped` through `_task_result_command`, and asserts the final `ToolMessage` carries `subagent_stop_reason=loop_capped` (+ the `(capped: repeated tool-call loop)` note in content). Plus middleware-level coverage that the hard stop really stamps it (and a warn-only run does not), and an executor test that two guards are both checked (token-budget returns `None`, loop-detection returns `loop_capped` → surfaced).

Minor refactor: extracted the `BoundedDict` both guards now share into `agents/middlewares/_bounded_dict.py` rather than duplicating or cross-importing between middleware modules. Backend test suite green (256 in the affected sweep), ruff clean.

hata33: Follow-up: amended to 95c4c66e (fd7f8155 was force-pushed over). One hardening on top of the above — the `loop_capped` stamp in `LoopDetectionMiddleware._apply` is now written under `self._lock`, matching `TokenBudgetMiddleware`'s locking. The lead agent's middleware instance is shared across concurrent Gateway threads, so without the lock the bounded-dict write had a theoretical concurrent-write race (subagent runs are unaffected — fresh instance per run, single-threaded). Tests re-run green.

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -708,32 +756,54 @@ async def _aexecute(self, task: str, result_holder: SubagentResult | None = None
...
+                    text = message_content_to_text(m.content).strip()
+                    if text:
+                        usable_partial = text
+                    break
+            records = collector.snapshot_records() if collector is not None else None
+            if usable_partial is not None:
+                result.try_set_terminal(
+                    SubagentStatus.COMPLETED,
+                    result=usable_partial,
+                    stop_reason="turn_capped",
```
willem-bd: The `GraphRecursionError` handler hardcodes `stop_reason="turn_capped"` on both arms (here and line 804), bypassing the guard-reason lookup the normal-completion path uses at line 764 (`stop_reason = self._consume_guard_stop_reason()`).

If the token-budget (or loop-detection) hard-stop already fired earlier in this run — setting `_stop_reason[run_id] = "token_capped"` and stripping `tool_calls` to force a final answer — and `recursion_limit` (= `max_turns`) is then tripped on the following super-step before the model emits that final answer, this handler mislabels the run as `turn_capped` even though the token budget was the binding constraint. The real reason is also left orphaned in the (bounded) dict because it's never popped.

The whole point of this PR is accurate cap-reason surfacing, so a one-liner keeps the two paths consistent:

```python
stop_reason = self._consume_guard_stop_reason() or "turn_capped"
```

Trigger is narrow (both budgets must bind on consecutive turns) and both reasons still classify as "capped", so severity is low — but the fix is trivial and the divergence from the completion path is clear.

willem-bd (commented):
Reviewed the diff — the design is clean: additive `subagent_stop_reason` preserves v1 consumers, the fresh-middleware-per-run lifecycle avoids cross-run contamination, `_stop_reason` writes are lock-guarded and excluded from `after_agent` cleanup, and the legacy `max_turns_reached` read-side normalization is thoughtful. Contract fixture v2 + config_version 19→20 + full tests on both sides all check out. One low-severity correctness-adjacent edge below.

backend/packages/harness/deerflow/subagents/executor.py
hata33: Good catch — applied exactly that in 5a3d7035. The exception handler now computes `stop_reason = self._consume_guard_stop_reason() or "turn_capped"` once and uses it on both arms, so:

- a guard that already fired this run (token/loop hard-stop stripped tool_calls to force a final answer) wins when `GraphRecursionError` then trips on the next super-step — its reason was the binding constraint, and `consume_stop_reason` pops it so it is no longer orphaned;
- a pure turn-budget exhaustion (no guard fired) still falls back to `turn_capped`.

Both `stop_reason` paths in the executor (normal completion + this exception handler) now go through the same `_consume_guard_stop_reason()` lookup. Added `test_aexecute_recursion_error_prefers_guard_stop_reason_over_turn_capped` covering the guard-fired-then-recursion-tripped case. Tests green.

(And thanks for the broader sign-off on the design — appreciate it.)

WillemJiang (approved):

PR #3981, opened 2026-07-07
feat(frontend): render slash-skill activations as inline chips

18062706139fcz:
## Summary

- Show an explicit `/skill` activation as a compact **inline chip** in both the composer and the chat transcript, instead of raw slash text. The chip sits inline with the input text so it reads naturally and stays aligned.
- In the composer, selecting a skill suggestion stores it as a removable chip; the `/skill ` prefix is only re-attached at submit time, so the **backend activation protocol is unchanged**. Backspace on an empty input (or the chip close button) clears it, and prompt-history navigation is suppressed while a chip is active.
- Human messages in the transcript that begin with `/skill` now render the skill as a read-only chip followed by the task text.
- Add a shared `core/skills/slash.ts` (`parseSlashSkillReference` + `resolveSlashSkillDisplay`) that mirrors the backend `slash.py` gate, so the transcript only renders a chip when the skill **actually exists and is enabled** — plain `/word` text and disabled/removed skills fall back to plain text. This also removes a duplicated regex/reserved-name list, keeping display semantics consistent with backend activation.

## Notes

- Multi-skill activation is intentionally **not** introduced here: the backend only resolves a single leading `/skill`, so the UI keeps single-skill semantics rather than faking multi-select.

## Test plan

- [x] `pnpm check` (eslint + tsc) clean
- [x] `pnpm test` — 570 unit tests pass, incl. new `tests/unit/core/skills/slash.test.ts` covering plain `/word`, unknown skill, and disabled skill all falling back to plain text
- [x] `chat.spec.ts` e2e extended to assert the composer still submits `/skill <task>` after showing a chip

image


willem-bd (commented):
Automated review pass. The core round-trip (chip select → `/skill ` prefix at submit → re-parse for display) and the parser parity with the backend are verified clean. A few second-order findings below.

frontend/src/core/skills/slash.ts
```diff
@@ -0,0 +1,65 @@
...
+ */
+export function resolveSlashSkillDisplay(
+  text: string,
+  skills: Skill[],
+): SlashSkillReference | null {
+  const reference = parseSlashSkillReference(text);
+  if (!reference) {
+    return null;
+  }
+  const enabled = skills.some(
```
willem-bd, line 65: `resolveSlashSkillDisplay` only checks `skill.enabled && name === reference.name`, but the backend resolver also enforces a per-agent allowlist (`available_skills`), which is a strict subset for any custom agent that declares a `skills: [...]` list (and for bootstrap agents). The frontend reads only the global skills catalog and has no way to learn that allowlist.

Effect: on a custom-agent thread whose allowlist omits a globally-enabled skill, the composer still suggests it as a chip and the transcript renders it as an activation, but the backend rejects it and replies "Skill /X is not available for this agent." — the chip misrepresents what actually ran. It's not silent (the user sees the error reply), but worth either filtering suggestions by the active agent's allowlist or exposing the allowlist via the skills endpoint.

frontend/src/core/skills/slash.ts
```diff
@@ -0,0 +1,65 @@
+import type { Skill } from "./type";
+
+/**
+ * Composer control commands that own the leading slash. They must never be
+ * shown as skill activations. Keep in sync with the backend
+ * `RESERVED_SLASH_SKILL_NAMES` in `deerflow/skills/slash.py`.
+ */
+export const RESERVED_SLASH_SKILL_NAMES = new Set([
```
willem-bd, line 12: `RESERVED_SLASH_SKILL_NAMES` and `SLASH_SKILL_RE` are duplicated from the backend parser with only a "Keep in sync" comment guarding the link. If a reserved command is added on the backend side and this file isn't updated, the frontend will start rendering `/newcmd …` submissions as activation chips in the transcript while the backend treats them as control commands. The repo already has a `contracts/` directory for cross-language JSON contracts — worth driving these from a shared contract rather than a manual copy.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -1200,20 +1257,35 @@ export function InputBox({
...
-            onFocus={() => setTextareaFocused(true)}
-            onKeyDown={handlePromptTextareaKeyDown}
-            ref={textareaRef}
-          />
-        </PromptInputBody>
+        <div className="min-h-16 w-full min-w-0 px-3 py-3">
+          <div className="flex w-full min-w-0 items-center gap-2">
+            {selectedSlashSkill && (
+              <button
+                aria-label={`Remove /${selectedSlashSkill.name}`}
```
willem-bd: This chip's Tailwind classes are near-identical to the transcript chip in `message-list-item.tsx` (same border/bg/font/shadow, with only a hover state added here). Worth extracting a shared `SlashSkillChip` so the two stay in lockstep. Minor: the `/8` and `/12` opacity steps here fall outside the standard token scale used by the rest of the palette.

frontend/src/components/workspace/messages/message-list-item.tsx
```diff
@@ -212,6 +214,28 @@ function MessageImage({
 
 const clientTurnDurations = new Map<string, number>();
 
+function HumanMessageText({ content }: { content: string }) {
```
willem-bd, line 221: `HumanMessageText` subscribes to the skills query per human message. Two observable costs: on initial thread load `skills` is empty until the query resolves, so a stored `/skill …` message paints as plain text and then re-renders as a chip (flicker); and toggling any skill's enabled state re-renders every human turn at once. Resolving the lookup once in the parent and passing the result down (or memoizing per content) avoids both.

WillemJiang (approved):

frontend/src/components/workspace/input-box.tsx
```diff
@@ -1200,20 +1258,29 @@ export function InputBox({
...
-            autoFocus={autoFocus}
-            defaultValue={initialValue}
-            onBlur={() => setTextareaFocused(false)}
-            onChange={handlePromptTextareaChange}
-            onFocus={() => setTextareaFocused(true)}
-            onKeyDown={handlePromptTextareaKeyDown}
-            ref={textareaRef}
-          />
-        </PromptInputBody>
+        <div className="min-h-16 w-full min-w-0 px-3 py-3">
```
willem-bd, line 1657: **Composer body layout change — worth a visual check across states.**

`PromptInputBody` is just `<div className={cn("contents", className)}>`, so the `absolute top-0 right-0 left-0 z-3` classes it had were effectively no-ops on a `display:contents` box — the textarea was a direct flex child of `PromptInput` with `size-full`. This change replaces it with a real `<div className="min-h-16 … px-3 py-3">` box and a `min-h-6! p-0! leading-6!` textarea, which shifts the box model (a fixed `min-h-16` where the textarea previously filled the parent).

The included screenshot covers one state. Could you confirm the composer still renders correctly with: attachments present, follow-ups visible, the welcome suggestion list, IME composition, long multi-line input, and on narrow/mobile widths? `pnpm check` and the e2e test won't surface visual regressions here.

PR #3982, opened 2026-07-07
fix(artifacts): honor trusted owner-user-id header

d33kayyy:
## Why

The artifact endpoint (`GET /api/threads/{id}/artifacts/{path}`) resolved filesystem paths only via the *effective* user. A trusted internal caller acting on behalf of an owner (e.g. an IM channel worker carrying `X-DeerFlow-Owner-User-Id`) therefore read the synthetic **internal** user's storage and 404'd on files the owner's run had actually written. This left artifacts generated during owner-scoped runs unreachable through internal callers, even though the memory and threads routers already honor the same header.

## What changed

- The artifact endpoint now resolves the owner via
`get_trusted_internal_owner_user_id(request)` and serves files from that
owner's user-scoped storage — matching the existing behavior of the
memory and threads routers.
- The header is honored **only after the internal token validates**; browser/API
callers send no such header, resolve to `None`, and fall back to the effective
user. **No change to browser/API behavior.**
- `resolve_thread_virtual_path()` gains an optional `user_id` param (defaults to
the effective user when omitted), so callers can resolve on behalf of a
specific owner explicitly.

## Surface area

## Bug fix verification

- Test path that reproduces the bug: `backend/tests/test_artifacts_router.py`
- `test_get_artifact_scopes_to_trusted_owner_header` — internal caller with a
    valid token + owner header reads the owner's storage
- `test_get_artifact_without_owner_header_falls_back_to_effective_user` — no
    header → effective-user behavior unchanged
- `test_get_artifact_ignores_owner_header_for_non_internal_caller` — header is
    ignored unless the internal token validated
- Did it go red on `main` and green on this branch? **yes** — the scoping test
fails against the pre-fix path resolution and passes with the owner threaded through.

## Validation

```
```

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Assisted with implementation and test authoring

### DIFF
--- a/backend/app/gateway/routers/artifacts.py
+++ b/backend/app/gateway/routers/artifacts.py
@@ -11,6 +11,7 @@ from fastapi.responses import FileResponse, PlainTextResponse, Response
 from app.gateway.authz import require_permission
 from app.gateway.internal_auth import get_trusted_internal_owner_user_id
 from app.gateway.path_utils import resolve_thread_virtual_path
+from deerflow.config.paths import make_safe_user_id
 
 logger = logging.getLogger(__name__)
 
@@ -184,10 +185,13 @@ async def get_artifact(thread_id: str, path: str, request: Request, download: bo
     """
     # Trusted internal callers may act on behalf of a thread's owner via the
     # owner-user-id header (honored only after the internal token validates).
-    # Resolve artifacts under that owner so internal callers read the same
-    # user-scoped storage the run wrote to, rather than the synthetic internal
-    # user. Browser/API callers get None here and fall back to the effective user.
-    owner_user_id = get_trusted_internal_owner_user_id(request)
+    # The header carries the raw platform owner id, while runs store files
+    # under the make_safe_user_id bucket (the same normalization the channel
+    # file pipeline and the memory router apply), so resolution uses the
+    # normalized id. Browser/API callers get None here and fall back to the
+    # effective user.
+    raw_owner_user_id = get_trusted_internal_owner_user_id(request)
+    owner_user_id = make_safe_user_id(raw_owner_user_id) if raw_owner_user_id else None
 
     # Check if this is a request for a file inside a .skill archive (e.g., xxx.skill/SKILL.md)
     if ".skill/" in path:
--- a/backend/app/gateway/routers/artifacts.py
+++ b/backend/app/gateway/routers/artifacts.py
@@ -9,6 +9,7 @@ from fastapi import APIRouter, HTTPException, Request
 from fastapi.responses import FileResponse, PlainTextResponse, Response
 
 from app.gateway.authz import require_permission
+from app.gateway.internal_auth import get_trusted_internal_owner_user_id
 from app.gateway.path_utils import resolve_thread_virtual_path
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


backend/app/gateway/routers/artifacts.py
```diff
@@ -181,6 +182,13 @@ async def get_artifact(thread_id: str, path: str, request: Request, download: bo
         - Download file: `/api/threads/abc123/artifacts/mnt/user-data/outputs/data.csv?download=true`
         - Active web content such as `.html`, `.xhtml`, and `.svg` artifacts is always downloaded
     """
+    # Trusted internal callers may act on behalf of a thread's owner via the
+    # owner-user-id header (honored only after the internal token validates).
+    # Resolve artifacts under that owner so internal callers read the same
+    # user-scoped storage the run wrote to, rather than the synthetic internal
+    # user. Browser/API callers get None here and fall back to the effective user.
+    owner_user_id = get_trusted_internal_owner_user_id(request)
```
willem-bd: The owner id from the trusted header isn't normalized through `make_safe_user_id` before being used as a filesystem user bucket — diverging from the memory router this PR says it mirrors.

Memory does: `raw_owner = get_trusted_internal_owner_user_id(request)` → `return make_safe_user_id(raw_owner)` (`app/gateway/routers/memory.py:39-41`). Here, `owner_user_id` is passed raw into `resolve_thread_virtual_path(..., user_id=owner_user_id)`.

The channel worker deliberately sends the **raw** owner id in this header — `app/channels/manager.py:549-551` documents that `_owner_headers` "deliberately sends the *raw* owner id (no sanitize, no platform fallback) over HTTP for gateway to re-resolve." The gateway is expected to re-resolve it; memory does, artifacts doesn't.

Concrete failure for any owner whose raw platform id has chars outside `[A-Za-z0-9_-]` (Feishu `open_id`s with dots, emails, anything with `@`/`:`/`/`):
- `_validate_user_id` raises `ValueError` (`packages/harness/deerflow/config/paths.py:33-37`) → `resolve_thread_virtual_path` turns it into **HTTP 400**.
- Even when validation passes, the lookup targets `users/{raw}/...` while the run wrote files under `users/{make_safe_user_id(raw)}/...` (`_channel_storage_user_id` → `prepare_user_dir_for_raw_id`) → **404**.

The added tests only use the already-safe id `"owner-123"`, so they never exercise this. Suggested fix: `owner_user_id = make_safe_user_id(owner_user_id)` when present, matching memory.

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ with **180 merged pull requests** since the first 2.0 milestone tag.
 - **memory:** Add `memory.token_counting` config to opt out of tiktoken for
   network-restricted deployments. ([#3465])
 - **suggest:** Make AI follow-up question suggestions optional. ([#3591])
+- **persistence:** wire alembic migrations, bootstrap schema on startup ([#3706])
```
willem-bd: This `persistence` entry (plus the `2.0.0` date bump `2026-06-15`→`2026-06-25` at line 8 and the PR-count bump `180`→`182` at line 16) doesn't belong to this PR — `#3706` is a separate persistence/migrations PR, unrelated to the artifacts owner-header fix in #3982. Looks like a rebase/merge artifact; dropping it keeps the changelog tied to this PR's actual change.

willem-bd (commented):
Reviewed the diff — the fix's shape and the internal-token gating are correct, but one correctness issue and two bundled-but-unrelated changes stand out.

**Most important:** the trusted owner id isn't normalized through `make_safe_user_id`, unlike the memory router this PR says it mirrors. The channel worker sends the *raw* owner id in the header and expects the gateway to re-resolve it, so for any IM-channel owner whose platform id falls outside `[A-Za-z0-9_-]` (dots, `@`, `:`, `/`, …) the artifact request 400s or reads the wrong bucket — i.e. the fix fails for exactly the callers it targets. Details inline.

backend/uv.lock
```diff
@@ -2162,7 +2162,7 @@ wheels = [
 
 [[package]]
 name = "langsmith"
-version = "0.8.0"
+version = "0.8.18"
```
willem-bd, line 2165: `langsmith` `0.8.0`→`0.8.18` and `pydantic-settings` `2.14.0`→`2.14.2` are bumped here with no matching change in `pyproject.toml` and no mention in the PR body. Unrelated to the artifacts fix — please ship as a separate dependency-bump PR if intended.

WillemJiang:
@d33kayyy  thanks for your contribution. Please take a look at the review comments and resolve the conflicts with the main branch.

d33kayyy:
sorry I was a little bit busy, will look into this today

d33kayyy:
updated per review @willem-bd

WillemJiang (approved):

PR #3985, opened 2026-07-07
fix(provisioner): gate legacy skills mount by user visibility

### LINKED ISSUE #3977: [Bug] AioSandboxProvider 未跟随 #3889 三路挂载（public/custom/legacy），per-user 与 legacy custom skill 在 Docker 沙箱下 read_file 404
## 背景

在 #3914 缺口 1（自然语言自主调用注入 request-scoped secrets）的真实环境验证中，@yong326 发现自然语言触发 skill 时 `read_file(SKILL.md)` 返回 404（复现日志见 [该评论](https://github.com/bytedance/deer-flow/issues/3914#issuecomment-4889661744)）。深入代码后确认这是一个独立于自主注入功能本身的沙箱挂载缺口。

## 根因

PR #3889（per-user custom skill isolation）为 skill 引入了 `public`/`custom`/`legacy` 三种容器路径（`Skill.get_container_path()` 据此拼出 `/mnt/skills/{public,custom,legacy}/<name>/`），并同步更新了 `LocalSandboxProvider._build_thread_path_mappings`（`backend/packages/harness/deerflow/sandbox/local/local_sandbox_provider.py`）来做三路挂载：

- `public`：全局静态挂载
- `custom`：per-user 目录 `{base_dir}/users/{user_id}/skills/custom/` 挂到 `{container}/custom`（per-thread，因为不同用户对应不同宿主目录）
- `legacy`：仅当用户没有 per-user custom skill 时，把全局 `skills/custom/` 挂到 `{container}/legacy`（兼容旧数据）

但 **`AioSandboxProvider._get_skills_mount()`**（`backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py:339-356`）**没有跟着更新**——它仍是 #3889 之前的老逻辑：一个无参数的 static method，把 `config.skills.get_skills_path()`（全局 `skills/` 目录）整体挂到 `container_path`（如 `/mnt/skills`），不区分三路分类，也不感知 `user_id`。

## 影响

在 AIO/Docker 沙箱（生产环境的推荐部署方式）下：

1. **Legacy 全局 custom skill**：harness 按分类算出容器路径 `/mnt/skills/legacy/<name>/SKILL.md`，但 AIO 实际把宿主机 `skills/custom/<name>/` 原样镜像挂在 `/mnt/skills/custom/<name>/` → 路径不匹配，`read_file` 404。
2. **per-user custom skill**（#3889 引入的新隔离目录）：`_get_skills_mount()` 只挂全局 `skills_path`，从未挂载任何 per-user 目录 → 这类 skill 在 AIO 沙箱里**完全不可读**，不是路径错，是内容压根没挂进去。

`LocalSandboxProvider` 不受影响（已随 #3889 一起修复）。

## 建议修复方向

给 `AioSandboxProvider` 补上等价的三路挂载逻辑，参考 `LocalSandboxProvider._build_thread_path_mappings` 的实现：`_get_skills_mount()` 需要感知 `user_id`（当前是 static method，需要改造），并在 `_get_extra_mounts` / `_get_thread_mounts` 里追加 per-user custom + legacy 兼容两个 volume mount（Docker `-v` 参数），而不是维持现在的单一全局挂载。

## 验证计划

- Local 与 AIO 两种 sandbox provider 下，分别用（a）public skill、（b）per-user custom skill、（c）legacy 全局 custom skill 三种 skill，验证模型自主 `read_file` 均能正确解析。
- 回归 #3914 缺口 1 的自然语言自主调用 secrets 注入链路。

Vanzeren:
Fixes #3977

## Why

Provisioner hostPath mode was always mounting the legacy global `skills/custom`
  directory into every sandbox. That created two regressions after the three-way
  skills layout change:

  1. Fresh K8s installs with no legacy `skills/custom` directory could fail before
     the sandbox Pod started, because the unconditional `skills-legacy` hostPath
     required an existing directory.
  2. Users who already had per-user custom skills could still access hidden legacy
     skills through sandbox filesystem paths, even though `UserScopedSkillStorage`
     intentionally stops surfacing legacy skills once a user has their own custom
     skills.

  This PR aligns provisioner behavior with the existing local/AIO/user-scoped
  storage visibility rules and prevents the sandbox filesystem from being less
  restrictive than skill discovery.

## What changed

Provisioner sandbox creation now treats legacy skills as conditional instead of
  always mounting them.

  - The backend now decides whether a sandbox should see legacy skills and passes
    that decision to the provisioner via `include_legacy_skills`.
  - K8s hostPath sandboxes now always mount public skills and the current user's
    custom skill root, but only mount `/mnt/skills/legacy` when that user should
    actually see legacy fallback skills.
  - Fresh installs without a legacy `skills/custom` directory no longer require
    that directory just to start sandbox Pods.
  - Added regression coverage for provisioner helper behavior, route-level Pod
    spec creation, remote backend request payloads, and documented the verified
    manual K8s end-to-end flow.

## Surface area

## Bug fix verification

    - `backend/tests/test_provisioner_pvc_volumes.py`
    - `backend/tests/test_provisioner_request_threading.py`
    - `backend/tests/test_remote_sandbox_backend.py`
  - Did it go red on `main` and green on this branch? (yes / no)
    - Unit/regression coverage was written against the buggy behavior and passes on this branch.
    - Real verification was also done with manual end-to-end runs in both Kubernetes provisioner mode
    and local Docker sandbox mode.
    - Full K8s provisioner failure depends on a live cluster plus hostPath state, so I also ran a real
    `kind`-backed manual .

## Validation

 - `cd backend && backend/.venv/bin/python -m ruff check backend`
  - `cd backend && backend/.venv/bin/python -m ruff format --check backend`
  - `cd backend && UV_CACHE_DIR=/tmp/uv-cache PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run
  pytest tests/test_provisioner_pvc_volumes.py tests/test_remote_sandbox_backend.py -q`
  - `cd backend && UV_CACHE_DIR=/tmp/uv-cache PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run
  pytest tests/
  test_provisioner_request_threading.py::test_create_sandbox_route_builds_expected_skills_mount_layout[w
  ithout-legacy] tests/
  test_provisioner_request_threading.py::test_create_sandbox_route_builds_expected_skills_mount_layout[w
  ith-legacy] -q`
  - Manual K8s E2E against local `kind` cluster:
    - verified `include_legacy_skills=false` creates a running Pod without `skills-legacy`
    - verified `include_legacy_skills=true` creates a running Pod with `skills-legacy`
  - Manual local Docker E2E:
    - verified user without custom skills gets `public + custom + legacy`
    - verified user with custom skills gets `public + custom` and cannot read `legacy`

## AI assistance

**Tool(s) used:**  Codex

**How you used it:**  Used Codex to inspect the existing sandbox/storage logic, implement the
  provisioner/backend changes, add regression tests, run targeted validation, and document the manual
  K8s verification flow. I reviewed the resulting changes and used additional real K8s and local Docker
  end-to-end checks to confirm behavior.


docker/provisioner/app.py
```diff
@@ -250,26 +252,79 @@ def _sandbox_url(node_port: int) -> str:
...
-        skills_vol = k8s_client.V1Volume(
-            name="skills",
-            persistent_volume_claim=k8s_client.V1PersistentVolumeClaimVolumeSource(
-                claim_name=SKILLS_PVC_NAME,
-                read_only=True,
-            ),
+        # PVC mode: three-way subPath not yet supported; fall back to
+        # single-volume mount for backward compatibility.
+        logger.warning(
+            "SKILLS_PVC_NAME is set — three-way skills layout is not "
```
willem-bd, line 276: ⚠️ **Scope heads-up for reviewers:** this security/isolation fix only fully applies to **hostPath mode**. In PVC mode the whole skills PVC is still mounted at `/mnt/skills`, so:

- Legacy `skills/custom` stays filesystem-readable by users who already have their own custom skills — the exact leak this PR closes in hostPath mode.
- Per-user custom skills still won't resolve (`Skill.get_container_path()` produces `/mnt/skills/custom/<name>/`, but PVC mode has no per-user subPath).

This is a reasonable scoping decision (and the warning helps), but it isn't called out in the PR body and isn't flagged under Surface area. Could you add a note under Surface area and/or open a follow-up issue so PVC-mode operators know they don't pick up the fix?

backend/packages/harness/deerflow/community/aio_sandbox/remote_backend.py
```diff
@@ -22,13 +22,21 @@
...
 from .backend import SandboxBackend
 from .sandbox_info import SandboxInfo
 
 logger = logging.getLogger(__name__)
 
 
+def _should_mount_legacy_skills(user_id: str) -> bool:
+    """Return whether this user should see legacy global-custom skills."""
+    storage = get_or_new_user_skill_storage(user_id)
+    return any((skill.category.value if hasattr(skill.category, "value") else skill.category) == SkillCategory.LEGACY.value for skill in storage.load_skills(enabled_only=False))
```
willem-bd: The legacy-visibility decision is computed **two different ways** across providers:

- **Here (remote):** `_should_mount_legacy_skills` → `load_skills()` → checks for a `LEGACY`-category skill.
- **AIO / Local** (`aio_sandbox_provider.py`, `local_sandbox_provider.py`): filesystem heuristic — *any* non-hidden subdir in the user's custom dir suppresses legacy.

These can diverge: a user custom subdir **without** `SKILL.md` makes the filesystem heuristic suppress legacy (dir exists) while storage still yields `LEGACY` (no valid custom skill file) → remote mounts legacy, AIO/Local don't.

The remote approach is actually the more correct one (it uses the same source of truth as discovery, so the mount can't be less restrictive than what the model is told). Not a regression — AIO just mirrors existing Local behavior — but the divergence is a maintenance trap. Worth a shared helper (or having AIO/Local delegate to storage) in a follow-up.

backend/tests/test_remote_sandbox_backend.py
```diff
@@ -148,13 +149,15 @@ def mock_create(thread_id: str, sandbox_id: str, extra_mounts=None, *, user_id=N
 
 def test_provisioner_create_returns_sandbox_info(monkeypatch):
     backend = RemoteSandboxBackend("http://provisioner:8002")
+    monkeypatch.setattr(remote_backend_mod, "_should_mount_legacy_skills", lambda user_id: True)
```
willem-bd: Coverage gap: every test here monkeypatches `_should_mount_legacy_skills`, so the actual decision logic (the `hasattr(category, "value")` dance + reliance on `load_skills`'s visibility rule) is only exercised by manual E2E.

A focused unit test would lock in the contract cheaply:
- user **with** a custom skill → `False`
- user **without** custom skills + a legacy dir containing a skill → `True`
- user **without** custom skills + empty legacy dir → `False`

willem-bd (commented):
Three review notes from a `/review` pass — posting inline. Overall this is a solid fix with strong test coverage; these are scope/consistency observations, not blockers.

docker/provisioner/app.py
Vanzeren: I have already discussed about this matter in [3977](https://github.com/bytedance/deer-flow/issues/3977)

WillemJiang (approved):

PR #3986, opened 2026-07-07
feat: add composer input polishing

18062706139fcz:
## Why

Users often draft short or rough prompts in the Web UI composer and then send them directly to the agent. This PR adds an explicit pre-send polish action so users can turn a rough draft into a clearer instruction before starting a run, without creating a thread message or mutating thread state.

The intended UX is lightweight: the user stays in control, the rewrite only replaces the local draft after clicking the polish button, and the rewritten text can be undone before sending.

## What changed

- Added an authenticated Gateway endpoint, `POST /api/input-polish`, guarded by `runs:create` permission.
- The endpoint performs a short LLM rewrite request and returns `{ rewritten_text, changed }`; it does not create a LangGraph run, persist a message, or modify thread state.
- Added `input_polish` app config with:
  - `enabled` to hide/disable the endpoint when needed.
  - `max_chars` to cap draft size.
  - `model_name` so deployments can choose a fast/cheap model for this UX path.
- Added a composer polish button in the Web UI.
  - Empty input, reserved builtin commands like `/goal` and `/help`, streaming state, mock mode, and active slash-suggestion state do not trigger polishing.
  - Slash skill prompts such as `/web-dev ...` remain eligible, and the backend prompt explicitly preserves leading slash prefixes.
  - While a polish request is in flight, the composer shows a compact status pill, disables editing/actions, and aborts stale requests on text/thread changes.
  - When a rewrite is applied, the same button becomes an undo action.
- Added i18n strings for English and Chinese. The transient status is now concise: `Polishing input...` / `正在优化输入...`.
- Updated README and AGENTS docs to describe the new composer helper and Gateway route.
- Added backend unit tests, frontend helper unit coverage, and Playwright coverage for polish + undo flows.

## Surface area

## Screenshots / Recording

Not attached. The UI change is a small composer action button plus a transient in-composer status pill; the interaction is covered by the Playwright polish and undo tests listed below.

## Validation

- `cd backend && PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run pytest tests/test_input_polish_router.py -v`
- `cd backend && uvx ruff check app/gateway/routers/input_polish.py packages/harness/deerflow/config/input_polish_config.py tests/test_input_polish_router.py && uvx ruff format --check app/gateway/routers/input_polish.py packages/harness/deerflow/config/input_polish_config.py tests/test_input_polish_router.py`
- `cd frontend && nvm use 22.21.1 && pnpm test tests/unit/components/workspace/input-box-helpers.test.ts`
- `cd frontend && nvm use 22.21.1 && pnpm check`
- `cd frontend && nvm use 22.21.1 && PORT=3100 PLAYWRIGHT_BASE_URL=http://localhost:3100 pnpm test:e2e tests/e2e/chat.spec.ts -g "polishes draft input before sending|undoes polished draft from the polish button"`

Note: the targeted Playwright run was executed on port `3100` because an existing local process was already listening on `localhost:3000`.

image

image


frontend/src/components/workspace/input-box.tsx
```diff
@@ -1247,14 +1383,22 @@ export function InputBox({
...
           className,
         )}
-        disabled={disabled}
+        disabled={composerLocked}
         globalDrop
         multiple
         onSubmit={handleSubmit}
         {...props}
       >
+        {polishingInput && (
```
willem-bd, line 1413: While a polish request is in flight the entire composer is hard-locked with no cancel path. `composerLocked` disables the textarea, submit, attachments, mode/model menus, and the polish button itself (`inputPolishDisabled` is true while `polishingInput`); the only abort trigger is `handlePromptTextareaChange`, which can't fire because the textarea is `disabled`. With a slow/hung provider, `model.ainvoke` blocks up to `stream_chunk_timeout` (~240s default in `factory.py`), leaving the user stuck for minutes with no escape but a page reload — which loses the draft. Consider adding a cancel affordance on this polishing overlay that calls `abortInputPolishRequest()`.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -967,6 +1020,87 @@ export function InputBox({
...
+        return;
+      }
+
+      const rewrittenText = result.rewritten_text.trim();
+      if (!rewrittenText || !result.changed) {
+        toast.info(t.inputBox.inputPolishNoChanges);
+        return;
+      }
+
+      setPromptHistoryValue(rewrittenText);
```
willem-bd, line 1077: `handlePolishInput` replaces the textarea value here without resetting `promptHistoryIndexRef`, leaving stale history-browse state. If the user pressed ArrowUp to browse history (index = N-2) and then clicks polish, the value updates but the index stays at N-2 (only `handlePromptTextareaChange`, which polish bypasses, clears it). The next ArrowDown reuses the stale index and overwrites the polished text with `promptHistory[N-1]`, silently losing the rewrite. Worth resetting `promptHistoryIndexRef.current = null` (and `promptHistoryDraftRef`) alongside applying the rewrite.

backend/app/gateway/routers/input_polish.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+_extract_response_text = llm_text.extract_response_text
+_strip_markdown_code_fence = llm_text.strip_markdown_code_fence
+_strip_think_blocks = llm_text.strip_think_blocks
+
+
+def _clean_rewritten_text(text: str) -> str:
+    candidate = _strip_think_blocks(text)
+    candidate = _strip_markdown_code_fence(candidate)
```
willem-bd: `strip_think_blocks` will truncate any legitimate rewrite that contains a literal `<think>` substring. After removing complete blocks, `_OPEN_THINK_RE.search` truncates from any dangling `<think>` open tag (see `llm_text.strip_think_blocks`). A draft like "rewrite: explain what the `<think>` tag does" produces model output containing `<think>`, which gets cut at the tag → a truncated rewrite, or an empty result → a spurious `503 Failed to polish input` despite a successful LLM call.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -881,6 +919,21 @@ export function InputBox({
...
     skillSuggestions.length > 0 &&
     dismissedSkillSuggestionValue !== textInput.value;
+  const isComposerDisabled = disabled === true;
+  const isMockThread = isMock === true;
+  const composerLocked = isComposerDisabled || polishingInput;
+  const inputPolishUndoAvailable =
+    !polishingInput &&
+    inputPolishUndo !== null &&
+    (textInput.value ?? "") === inputPolishUndo.rewrittenText;
+  const inputPolishDisabled =
```
willem-bd, line 938: Polish is not disabled while a human-input card is open. `frontend/AGENTS.md` states: "Composer entry points should disable normal bottom input while `hasOpenHumanInputRequest(...)` is true so users answer through the card and preserve response metadata." `inputPolishDisabled` checks streaming/slash/mock/disabled but not `hasOpenHumanInputRequest`, so a user can overwrite their in-progress card-reply draft mid-request.

frontend/src/components/workspace/input-box-helpers.ts
```diff
@@ -185,6 +185,16 @@ export function parseGoalCommand(value: string): GoalCommand | null {
   return { kind: "set", objective: args };
 }
 
+export function canPolishInput(value: string): boolean {
+  const trimmed = value.trim();
+  if (!trimmed) {
+    return false;
+  }
+  return (
+    !/^\/(?:goal|help)(?:\s+|$)/i.test(trimmed) && !parseCompactCommand(trimmed)
```
willem-bd: Reserved builtin commands are hardcoded here rather than read from the existing registry. The component already keeps `builtinSlashCommands` (`goal`, `compact`), and `/help` is a phantom (no `/help` parser exists in the frontend) — yet a third parallel list is introduced. When the next builtin lands, a dev must remember to patch this regex too, or the polish button will silently rewrite a reserved command line into prose and route it to the LLM as ordinary input.

backend/app/gateway/routers/input_polish.py
```diff
@@ -0,0 +1,117 @@
...
+
+    max_chars = config.input_polish.max_chars
+    if len(body.text) > max_chars:
+        raise HTTPException(status_code=400, detail=f"Input text exceeds {max_chars} characters")
+
+    model_name = config.input_polish.model_name
+    try:
+        model = create_chat_model(name=model_name, thinking_enabled=False, app_config=config)
+        invoke_config: dict = {"run_name": "input_polish"}
+        inject_langfuse_metadata(
```
willem-bd: The `create_chat_model` → `inject_langfuse_metadata` → `ainvoke([SystemMessage, HumanMessage])` sequence plus the `_strip_think_blocks`/`_strip_markdown_code_fence` cleaning is a near-verbatim clone of `suggestions.py` (and the title-generation path). There's no shared "single non-graph LLM rewrite" helper, so the next content-wrapper stripping fix or tracing-metadata field must be applied in both routers and will silently regress in whichever is forgotten.

backend/app/gateway/routers/input_polish.py
```diff
@@ -0,0 +1,117 @@
...
+    if not text:
+        raise HTTPException(status_code=400, detail="Input text is required")
+
+    max_chars = config.input_polish.max_chars
+    if len(body.text) > max_chars:
+        raise HTTPException(status_code=400, detail=f"Input text exceeds {max_chars} characters")
+
+    model_name = config.input_polish.model_name
+    try:
+        model = create_chat_model(name=model_name, thinking_enabled=False, app_config=config)
```
willem-bd: `create_chat_model` is rebuilt on every polish click. This is an interactive, latency-sensitive affordance explicitly meant for a fast/cheap model, yet it re-runs provider reflection/config resolution per request. A small LRU keyed by `(model_name, thinking_enabled)` would remove repeated instantiation without affecting hot-reload.

backend/app/gateway/routers/input_polish.py
```diff
@@ -0,0 +1,117 @@
...
+
+    if not config.input_polish.enabled:
+        raise HTTPException(status_code=404, detail="Input polishing is disabled")
+
+    text = body.text.strip()
+    if not text:
+        raise HTTPException(status_code=400, detail="Input text is required")
+
+    max_chars = config.input_polish.max_chars
+    if len(body.text) > max_chars:
```
willem-bd: `max_chars` is checked against raw `body.text`, while the empty-text check uses `text = body.text.strip()` (line ~81) and `_build_user_content` sends the unstripped text. A draft over the limit only due to padding is rejected, and a padded draft just under the limit is sent to the model with whitespace intact — the user-facing boundary and the model input disagree. Consider validating against one normalized view of the input.

willem-bd (commented):
Reviewed the input-polish PR. Core mechanics (abort/sequence race protection, auth decorator, `name=None` fallback, `inject_langfuse_metadata` with `environment=None`) check out, and docs/tests/config bump are all in order. Findings below are mostly UX edge cases plus one altitude concern — see inline comments, most-severe first.

WillemJiang (approved):

Issue #3929, opened 2026-07-04
[security][sandbox] K8s provisioner exposes code-execution sandboxes via NodePort — propose ClusterIP


zhfeng:
## Summary

The K8s sandbox provisioner exposes each sandbox Pod via a **NodePort** Service and hands the gateway `http://{NODE_HOST}:{NodePort}`. On a real multi-node cluster this binds the sandbox — a service that **executes arbitrary agent-generated code** — onto every node's interfaces, including externally-reachable ones. I'd like to discuss switching the per-sandbox Service to **ClusterIP** and returning a cluster-DNS URL. This removes a meaningful external-exposure surface and the per-deploy `nodeHost` knob, and works on any cluster shape.

Filing this as a discussion / RFC rather than a bug — the current design is correct and stable for the local-Docker-Desktop / OrbStack / minikube single-node case it was built for; the concern is how it behaves on a real cluster.

## Current behavior

`docker/provisioner/app.py`:

- `_build_service()` (`app.py:373`) creates `type="NodePort"`; K8s auto-allocates a port from 30000–32767.
- `_get_node_port()` (`app.py:404`) reads the allocated port back.
- `_sandbox_url()` (`app.py:239`) returns `f"http://{NODE_HOST}:{node_port}"`.
- `NODE_HOST` (`app.py:77`) defaults to `host.docker.internal`.

The provisioner README documents this as the intended flow: *"A NodePort Service is created to expose the Pod … The provisioner returns `http://host.docker.internal:{NodePort}` to the backend."*

NodePort was chosen originally (commit `300e5a51`, *Adds Kubernetes sandbox provisioner support*) because the provisioner targeted **local single-node K8s** (Docker Desktop / OrbStack / minikube) where the gateway container and sandbox Pods share the VM network and `host.docker.internal` resolves to the VM host. NodePort is the K8s analogue of a Docker *published port* — a faithful translation of the compose DooD pattern. On single-node local clusters it is correct.

## The security concern (the main reason I'm raising this)

A NodePort is bound on **every node's interfaces**, including external ones. A code-execution sandbox — which by design runs arbitrary agent-generated shell/code, so prompt-injection or tool misuse can land there — becomes reachable from outside the cluster. On managed cloud K8s (GKE / EKS / AKS) NodePorts are opened on the node's **external** IP by default unless network policy / firewall rules specifically block it. ClusterIP, by contrast, is cluster-internal only: the gateway→sandbox hop never leaves the cluster network.

[...]

PR #3987, opened 2026-07-08
feat(deploy): first-class Helm chart for Kubernetes deployment

### LINKED ISSUE #3930: [RFC][deploy] First-class Helm chart for Kubernetes deployment
## Summary

DeerFlow's only production deployment story today is `docker/docker-compose.yaml`. This RFC proposes adding a **first-class Helm chart** (`deploy/helm/deer-flow/`) that translates the compose stack to native Kubernetes resources, plus CI to publish the chart and images as OCI artifacts. I have a working chart prepared on a local branch; I'm opening this RFC to get maintainer buy-in and align on the open decisions *before* opening the PR, since this is a sizeable addition.

Filing as an RFC per the pattern of #3739 / #3804 / #2634.

## Motivation

- **No native K8s path.** The repo ships a compose stack and a K8s sandbox provisioner, but no manifests/chart for running the *gateway/frontend/nginx* plane on Kubernetes. K8s users currently have to hand-roll it.
- **An emerging K8s user base with no first-class path.** A string of recent issues touch K8s deployment — #2480 (extra mounts in k8s provisioner), #2600 (sandbox host-gateway reachability), #3929 (NodePort exposure of sandboxes). There's clear demand, and no canonical answer to point people at.
- **compose is the wrong ceiling for prod.** Rolling updates, autoscaling, RBAC, multi-tenancy, and secret management all want K8s primitives. A chart lets prod deployments stop reinventing this.

## Chart design (summary)

The chart deploys five workloads, mirroring the compose services:

| Workload    | Role                                                                 |
| ----------- | ------------------------------------------------------------------- |
| `gateway`   | Backend + embedded LangGraph runtime                                |
| `frontend`  | Next.js                                                              |
| `nginx`     | Internal reverse proxy preserving the compose routing (`/api/*` etc.) |
| `provisioner` | K8s-native sandbox — spawns code-execution Pods on demand         |
| `postgresql`| Bundled StatefulSet (disable to use an external DB)                 |

Highlights, all of which are motivated by production hardening rather than novelty:

- **Config-checksum annotations** on gateway/nginx pod templates so config-only upgrades trigger a rolling restart (subPath mounts don't get live updates).
- **Pod security contexts**: all workloads non-root (gateway/frontend/provisioner uid 1000, nginx uid 101, postgres uid 999), drop ALL capabilities, `seccompProfile: RuntimeDefault`, no privilege escalation.
- **Resource requests/limits** defaults for all five workloads.
- **OCI publishing**: chart published as an OCI artifact to GHCR on `v*` tags, images published alongside. CI uses SHA-pinned actions + build provenance.
- **No existing repo files modified** — the chart is purely additive.

## Open decisions I'd like maintainer input on

These are the choices I'd rather debate now than bake into the PR and re-litigate in review.

### 1. Chart location

Currently `deploy/helm/deer-flow/`. The Helm community convention is `charts/<name>/`. `deploy/helm/` groups it with the existing `docker/` deploy artifacts and reads well next to `docker-compose`; `charts/` is more conventional for repos whose chart is a primary artifact. Mild preference for `deploy/helm/` (additive, discoverable next to compose). Maintainers' call.

### 2. OCI publishing target & versioning

Plan is to publish chart + images to `ghcr.io/<owner>/deer-flow*` on `v*` tags. Two sub-questions:

- **Owner / package visibility.** For an official chart this should land under `ghcr.io/bytedance/deer-flow`, which needs the bytedance side to own the GHCR packages and flip them public (or document a pull secret). If that's not palatable, the chart can stay consumer-build-only (the `scripts/build-and-push.sh` path) and skip OCI publishing. Which does the team prefer?
- **Version policy.** Chart `version` vs `appVersion` vs `image.tag`. Today the chart enforces `chart-version == git tag` for releases and treats `image.tag` as independently pinnable. Is that the policy the maintainers want, or should `image.tag` default to the chart version?

### 3. Bundled vs external Postgres

The chart bundles a postgres StatefulSet (`postgresql.enabled: true` by default) for ease of first install, with `enabled: false` to bring your own managed DB. Two reasonable defaults:

- **Default-on** (current) — one `helm install` gives a working stack. Better first-run experience; worse for prod where most will use a managed DB.
- **Default-off** — chart assumes an external DB; bundled postgres is opt-in for dev/testing.

Mild preference for default-on (matches compose's "it just works" ethos; prod users will override anyway). Maintainers' call.

### 4. Multi-replica streaming gap — the big architectural caveat

This is the one I most want eyes on. The gateway's **stream bridge is in-process** (`type: memory`, the default; the `redis` backend is scaffolded but raises `NotImplementedError`). Normal chatting works fine on >1 gateway replica because `POST /{thread_id}/runs/stream` starts the run *and* streams SSE on the same request — pinned to one pod. But:

- **Cross-pod routing** — `GET /runs/{run_id}/join` (reconnect) and `POST /runs/{run_id}/cancel` can land on a pod that doesn't own the run. Cheap fix: nginx consistent hashing on `thread_id` (~5 lines in the chart's nginx config, no new infra). Makes 2 replicas fully usable for testing.
- **Pod death** — the pod owning a stream dies → its in-process queue is lost → the in-flight stream cannot resume on any pod. Only a shared stream store (Redis) fixes this; nginx stickiness cannot.

So the chart can ship with `gateway.replicas: 1` safely today, and with nginx stickiness it's usable at 2 replicas for testing, but **prod multi-replica with rolling updates / autoscaling needs the Redis stream bridge**, which is separate backend work. Questions for maintainers:

- Should the chart **ship nginx stickiness on `thread_id` by default** (it's cheap and only helps)?
- Should **Redis be a chart dependency** (subchart or values pointing at an external Redis), gated on the backend landing the `redis` backend — or stay out of scope until that lands?
- Is "replicas: 1 by default; document the streaming caveat" an acceptable posture for a first chart release?

### 5. Provisioner scope & the NodePort question

The chart includes the K8s sandbox provisioner (`provisioner.enabled: true`), which today exposes each sandbox via a **NodePort** Service + a per-deploy `provisioner.nodeHost` knob. As raised in **#3929**, NodePort binds the sandbox (a code-execution surface) onto every node's external interfaces — the wrong default for a real cluster. Proposal: have the chart adopt **ClusterIP** from the start (per #3929) so we don't ship a known-exposure default. Tied to this:

- **RBAC** — the chart's provisioner Role grants `services`/`pods` create/get/list/watch/delete; that covers ClusterIP Services, no change needed.
- **RWO `persistence.home`** — the sandbox PVC is `ReadWriteOnce`, pinning each sandbox Pod to one node (limits multi-node scheduling). Orthogonal to the Service type, but the chart should document it as a known limitation. True multi-node sandbox scheduling needs RWX (NFS) or a per-node local-volume strategy — out of scope here.

### 6. Ingress default

Ship with `ingress.enabled` + a sane default (className, host, TLS secret), or leave ingress fully to user values? Mild preference for shipping a working default with `ingress.className: nginx` and TLS opt-in.

## What's ready

A working chart exists (local branch `helm-prod-hardening`, not yet pushed): full templates, `values.yaml` with the knobs above, `README` + `NOTES`, and drafted CI for image + chart publishing. Will open the PR — with tests (`helm lint`, `helm template`, kubeval/kyverno) — once the direction on the open decisions is agreed. No existing repo files are modified.

## Non-goals

- Not replacing docker-compose for dev — `make dev` / `make up` stay as-is.
- Not implementing the Redis stream bridge — that's separate backend work; the chart only needs to *accommodate* it (open question 4).
- Not changing the provisioner's core behavior — only the chart's defaults around it (open question 5, tied to #3929).

## Related

- #3929 — K8s provisioner NodePort → ClusterIP (the security concern baked into open question 5).
- #2480 — extra mounts not supported in k8s provisioner.
- #2600 — sandbox host-gateway reachability (Docker mode, same "don't route sandbox via host IP" theme).

Would love the maintainers' read on questions 1–6 — especially 4 (streaming posture) and 5 (ClusterIP default). Happy to open the PR once there's rough agreement.

zhfeng:
Closes #3930.

Implements the Helm chart proposed in the RFC: a native-Kubernetes translation of the production `docker-compose` stack, plus CI to publish the chart and container images as OCI artifacts to GHCR. The chart itself is purely additive — the only pre-existing file touched is a one-line pointer in `AGENTS.md`.

## What's added

`deploy/helm/deer-flow/` — a chart deploying five workloads mirroring the compose services:

| Workload      | Role                                                          |
| ------------- | ------------------------------------------------------------- |
| `gateway`     | Backend + embedded LangGraph runtime                          |
| `frontend`    | Next.js                                                       |
| `nginx`       | Internal reverse proxy preserving compose routing (`/api/*`)  |
| `provisioner` | K8s-native sandbox — spawns code-execution Pods on demand     |
| `postgresql`  | Bundled StatefulSet (disable to bring your own DB)            |

Production-hardening features:

- **Config-checksum annotations** on gateway/nginx pod templates so config-only upgrades trigger a rolling restart (subPath mounts don't get live updates).
- **Pod security contexts**: all workloads non-root (gateway/frontend/provisioner uid 1000, nginx uid 101, postgres uid 999), drop ALL capabilities, `seccompProfile: RuntimeDefault`, no privilege escalation.
- **Resource requests/limits** defaults for all five workloads.
- **OCI publishing** on `v*` tags: chart + images to GHCR, SHA-pinned actions + build provenance.

## RFC open decisions — resolved

1. **Chart location** → `deploy/helm/deer-flow/` (next to `docker/`).
2. **OCI publishing & versioning** → chart + images to `ghcr.io/<owner>/deer-flow*` on `v*` tags; `image.tag` defaults to `latest` and is independently pinnable; chart `version` and `appVersion` must equal the git tag.
3. **Postgres** → bundled StatefulSet, `postgresql.enabled: true` by default; set `enabled: false` for an external DB.
4. **Multi-replica streaming** → ships `gateway.replicas: 1` as the safe default; multi-replica is gated on #3948's run-control work (cancel/dedup/reconcile). Documented in the chart README.
5. **Provisioner sandbox exposure** → **deferred, not resolved here.** The chart ships the provisioner unchanged: each sandbox Pod is still exposed via a **NodePort** Service and the gateway reaches it via `provisioner.nodeHost` (a single node IP). Moving per-sandbox Services to **ClusterIP** (#3929) is a provisioner-code change (`docker/provisioner/app.py`), tracked as follow-up after this hardening batch. (The chart's own Service in front of the provisioner *deployment*, port 8002, is ClusterIP — but that's unrelated to #3929.)
6. **Ingress** → enabled by default with `className: nginx` and a placeholder host; TLS is opt-in.

## Release version gating (second commit)

Adds a reusable `verify-versions` workflow that both `container.yaml` and `chart.yaml` call on `v*` tags. It fails the release — skipping all image and chart publishing — unless `Chart.yaml` (`version` + `appVersion`), `backend/pyproject.toml`, and `frontend/package.json` all match the tag. Ships with `scripts/bump_version.sh <ver>` (bumps all four sources in lockstep) and `scripts/verify_versions.sh` (runnable locally); flow documented in `RELEASING.md`. This is what keeps the chart from shipping at a stale version — it originally sat at `0.1.0` while the project was at `2.1.0`.

## Verification

- `helm lint deploy/helm/deer-flow` — passes.
- Version scripts tested: `verify_versions.sh` pass/fail cases, `bump_version.sh` round-trip with a clean diff.
- Workflow YAML parses; gate wiring confirmed in both `chart.yaml` and `container.yaml`.

## Related

- #3929 — K8s provisioner NodePort → ClusterIP. **Not** resolved here; deferred as follow-up (provisioner-code change).
- #3948 — run-control work that unblocks safe multi-replica gateway.

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -3,6 +3,11 @@ name: Publish Helm Chart
 # Publishes the DeerFlow Helm chart as an OCI artifact to GHCR alongside the
 # container images (see container.yaml). Triggers on the same `v*` tags.
 #
+# On pull requests touching the chart or config.example.yaml, `validate-chart`
+# runs lint + template render + a config_version drift check (the chart's
+# embedded config_version must not lag config.example.yaml) so a broken or
+# stale chart fails the PR, not the release.
+#
 # Users then install with:
 #   helm install deer-flow oci://ghcr.io/${{ owner }}/deer-flow --version <ver>
 
@@ -10,16 +15,65 @@ on:
   push:
     tags:
       - "v*"
+  pull_request:
+    paths:
+      - "deploy/helm/deer-flow/**"
+      - "config.example.yaml"
+      - ".github/workflows/chart.yaml"
 
 jobs:
+  validate-chart:
+    # Runs on PRs and release tags: catch a broken render or a stale
+    # config_version before merge / publish. A broken chart published under a
+    # vX.Y.Z tag is an immutable OCI artifact (GHCR won't let you overwrite
+    # --version), so a regression must fail here, not on install.
+    if: github.event_name == 'pull_request' || startsWith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
+
+      # ubuntu-latest ships with helm 3 preinstalled - no setup-helm action needed.
+      - name: Lint chart
+        run: helm lint deploy/helm/deer-flow
+      - name: Validate templates render
+        run: helm template deer-flow deploy/helm/deer-flow --include-crds >/dev/null
+
+      # The chart's `config:` block embeds a config_version that must not fall
+      # behind config.example.yaml. A stale version is silent in-cluster (the
+      # image ships no example to compare against, so _check_config_version
+      # never warns) but means the chart's config is authored against an older
+      # schema. Bump it in values.yaml and the README example. config_version
+      # gates no runtime behavior - it only drives the outdated-warning - so a
+      # bare version bump needs no field changes.
+      - name: config_version drift check
+        run: |
+          set -eu
+          example=$(grep -E '^config_version:[[:space:]]+[0-9]+' config.example.yaml | head -1 | awk '{print $2}')
+          chart=$(awk '/^config:[[:space:]]*\|/{f=1; next} f && /^[[:space:]]+config_version:[[:space:]]+[0-9]+/ {print $2; exit}' deploy/helm/deer-flow/values.yaml)
+          echo "config.example.yaml config_version=$example"
+          echo "chart values.yaml     config_version=$chart"
+          if [ -z "$example" ] || [ -z "$chart" ]; then
+            echo "::error::could not parse config_version from one of the files"
+            exit 1
+          fi
+          if [ "$chart" -lt "$example" ]; then
+            echo "::error::chart config_version ($chart) is behind config.example.yaml ($example). Bump 'config_version' in deploy/helm/deer-flow/values.yaml (and the README example) to $example."
+            exit 1
+          fi
+
   verify-versions:
     # Gate the release: every version source must match the v* tag. A forgotten
     # bump in Chart.yaml, pyproject.toml, or package.json fails here and skips
     # the publish. See scripts/verify_versions.sh.
+    if: startsWith(github.ref, 'refs/tags/v')
     uses: ./.github/workflows/verify-versions.yml
 
   publish-chart:
-    needs: verify-versions
+    if: startsWith(github.ref, 'refs/tags/v')
+    needs: [verify-versions, validate-chart]
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -27,16 +81,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
-
-      # ubuntu-latest ships with helm 3 preinstalled — no setup-helm action needed.
-
-      # Gate packaging on template validity: a broken chart published under a
-      # vX.Y.Z tag is an immutable OCI artifact (GHCR won't let you overwrite
-      # --version), so a render regression must fail here, not on install.
-      - name: Lint chart
-        run: helm lint deploy/helm/deer-flow
-      - name: Validate templates render
-        run: helm template deer-flow deploy/helm/deer-flow --include-crds >/dev/null
 
       - name: Log in to GHCR
         run: |
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -12,7 +12,14 @@ on:
       - "v*"
 
 jobs:
+  verify-versions:
+    # Gate the release: every version source must match the v* tag. A forgotten
+    # bump in Chart.yaml, pyproject.toml, or package.json fails here and skips
+    # the publish. See scripts/verify_versions.sh.
+    uses: ./.github/workflows/verify-versions.yml
+
   publish-chart:
+    needs: verify-versions
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -27,17 +34,6 @@ jobs:
         run: |
           echo "${{ secrets.GITHUB_TOKEN }}" | \
             helm registry login ghcr.io -u ${{ github.actor }} --password-stdin
-
-      - name: Verify chart version matches tag
-        run: |
-          CHART_VERSION=$(helm show chart deploy/helm/deer-flow | awk '/^version:/ {print $2}')
-          TAG_VERSION="${GITHUB_REF_NAME#v}"
-          echo "Chart version: $CHART_VERSION"
-          echo "Tag version:    $TAG_VERSION"
-          if [ "$CHART_VERSION" != "$TAG_VERSION" ]; then
-            echo "::error::Chart version ($CHART_VERSION) does not match git tag ($GITHUB_REF_NAME). Bump Chart.yaml's version to $TAG_VERSION before tagging."
-            exit 1
-          fi
 
       - name: Package chart
         run: helm package deploy/helm/deer-flow --destination ./packages
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -1,0 +1,50 @@
+name: Publish Helm Chart
+
+# Publishes the DeerFlow Helm chart as an OCI artifact to GHCR alongside the
+# container images (see container.yaml). Triggers on the same `v*` tags.
+#
+# Users then install with:
+#   helm install deer-flow oci://ghcr.io/${{ owner }}/deer-flow --version <ver>
+
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  publish-chart:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # ubuntu-latest ships with helm 3 preinstalled — no setup-helm action needed.
+
+      - name: Log in to GHCR
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | \
+            helm registry login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Verify chart version matches tag
+        run: |
+          CHART_VERSION=$(helm show chart deploy/helm/deer-flow | awk '/^version:/ {print $2}')
+          TAG_VERSION="${GITHUB_REF_NAME#v}"
+          echo "Chart version: $CHART_VERSION"
+          echo "Tag version:    $TAG_VERSION"
+          if [ "$CHART_VERSION" != "$TAG_VERSION" ]; then
+            echo "::error::Chart version ($CHART_VERSION) does not match git tag ($GITHUB_REF_NAME). Bump Chart.yaml's version to $TAG_VERSION before tagging."
+            exit 1
+          fi
+
+      - name: Package chart
+        run: helm package deploy/helm/deer-flow --destination ./packages
+
+      - name: Push chart to GHCR
+        run: |
+          for pkg in ./packages/*.tgz; do
+            echo "--- pushing $pkg"
+            helm push "$pkg" oci://ghcr.io/${{ github.repository_owner }}
+          done
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -26,9 +26,17 @@ jobs:
       packages: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
 
       # ubuntu-latest ships with helm 3 preinstalled — no setup-helm action needed.
+
+      # Gate packaging on template validity: a broken chart published under a
+      # vX.Y.Z tag is an immutable OCI artifact (GHCR won't let you overwrite
+      # --version), so a render regression must fail here, not on install.
+      - name: Lint chart
+        run: helm lint deploy/helm/deer-flow
+      - name: Validate templates render
+        run: helm template deer-flow deploy/helm/deer-flow --include-crds >/dev/null
 
       - name: Log in to GHCR
         run: |
--- a/deploy/helm/deer-flow/templates/nginx-deployment.yaml
+++ b/deploy/helm/deer-flow/templates/nginx-deployment.yaml
@@ -1,0 +1,85 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "deer-flow.fullname" . }}-nginx
+  namespace: {{ include "deer-flow.namespace" . }}
+  labels:
+    {{- include "deer-flow.labels" . | nindent 4 }}
+    app.kubernetes.io/component: nginx
+spec:
+  replicas: {{ .Values.nginx.replicas }}
+  selector:
+    matchLabels:
+      {{- include "deer-flow.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: nginx
+  template:
+    metadata:
+      labels:
+        {{- include "deer-flow.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: nginx
+      annotations:
+        # Roll pods when the nginx ConfigMap changes (subPath mount).
+        checksum/nginx: {{ include "deer-flow.nginxChecksum" . }}
+    spec:
+      securityContext:
+        # Non-root: nginx:alpine ships a `nginx` user (uid 101). /etc/nginx and
+        # /var/cache/nginx are root-owned, so the command writes the rendered
+        # config to /tmp and we mount an emptyDir on the cache dir (below).
+        runAsNonRoot: true
+        runAsUser: 101
+        runAsGroup: 101
+        fsGroup: 101
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: nginx
+          image: {{ include "deer-flow.nginxImage" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: ["ALL"]
+          ports:
+            - name: http
+              containerPort: 2026
+          # Strip the IPv6 listen line when IPv6 is unavailable (mirrors compose).
+          # Writes the config to /tmp (1777, writable by uid 101) instead of
+          # /etc/nginx (root-owned), then loads it with `nginx -c`.
+          command: ["sh", "-c"]
+          args:
+            - >-
+              cp /etc/nginx/nginx.conf.template /tmp/nginx.conf &&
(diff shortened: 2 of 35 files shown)


deploy/helm/deer-flow/templates/nginx-deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+            - name: http
+              containerPort: 2026
+          # Strip the IPv6 listen line when IPv6 is unavailable (mirrors compose).
+          # Writes the config to /tmp (1777, writable by uid 101) instead of
+          # /etc/nginx (root-owned), then loads it with `nginx -c`.
+          command: ["sh", "-c"]
+          args:
+            - >-
+              cp /etc/nginx/nginx.conf.template /tmp/nginx.conf &&
+              test -e /proc/net/if_inet6 || sed -i '/^[[:space:]]*listen[[:space:]]\+\[::\]:2026;/d' /tmp/nginx.conf &&
```
willem-bd: The IPv6-listen strip pattern can never match, so this container crash-loops on pods without IPv6.

The sed requires a `;` immediately after `2026`: `\[::\]:2026;`. But the rendered config (`configmap-nginx.yaml:49`) emits `listen [::]:2026 default_server;` — there's a space + `default_server` between `2026` and `;`. So the line is never deleted. When `/proc/net/if_inet6` is absent the `[::]:2026` listen survives into `nginx -c /tmp/nginx.conf`, and nginx fails with `socket() :::2026 failed (97: Address family not supported)` → crash-loop.

This regex is copied verbatim from `docker/docker-compose-dev.yaml:94`, so the latent bug exists in compose too — but it's far more likely to trigger in K8s, where loopback IPv6 is commonly absent, than in Docker Desktop.

Suggested fix: drop the trailing `;` from the pattern, e.g. `/^[[:space:]]*listen[[:space:]]\+\[::\]:2026[[:space:]]/d`.

deploy/helm/deer-flow/templates/postgres-secret.yaml
```diff
@@ -0,0 +1,43 @@
...
+{{- $password := "" -}}
+{{- $dsn := "" -}}
+{{- if .Values.postgresql.enabled -}}
+{{- /* bundled: password persists across upgrades via lookup */ -}}
+{{- $prev := lookup "v1" "Secret" (include "deer-flow.namespace" .) (include "deer-flow.databaseUrlSecret" .) -}}
+{{- if $prev -}}
+{{- $password = index $prev.data "postgres-password" | default "" | b64dec -}}
+{{- end -}}
+{{- if not $password -}}{{- $password = randAlphaNum 32 -}}{{- end -}}
+{{- $dsn = printf "postgresql://%s:%s@%s:5432/%s" .Values.postgresql.auth.username $password (include "deer-flow.postgresFullname" .) .Values.postgresql.auth.database -}}
```
willem-bd: The password is spliced into the DSN verbatim, so a password containing URL-special characters breaks the connection string.

`postgresql://%s:%s@...` inserts `postgresql.auth.password` (or the persisted random value) with no URL-encoding. The default `randAlphaNum` is URL-safe, but an operator who sets `postgresql.auth.password` to something containing `@`, `:`, `/`, `#`, or `%` (reasonable for a managed-DB-style secret) produces a malformed DSN → the gateway fails to connect with a confusing parse error rather than a clear auth error.

The same applies to `redis-secret.yaml:23` (`redis://:%s@...` with `redis.auth.password`).

Suggested fix: percent-encode the password before interpolation (e.g. a `urlqueryescape`/`quote`-based helper), or document that `auth.password` must be URL-safe.

deploy/helm/deer-flow/templates/provisioner-deployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+              value: {{ include "deer-flow.namespace" . | quote }}
+            - name: SANDBOX_IMAGE
+              value: {{ .Values.provisioner.sandboxImage | quote }}
+            # In-cluster API access via the mounted ServiceAccount token.
+            # K8S_API_SERVER is intentionally unset (do NOT use host.docker.internal).
+            - name: NODE_HOST
+              {{- if .Values.provisioner.nodeHost }}
+              value: {{ .Values.provisioner.nodeHost | quote }}
+              {{- else }}
+              value: "gateway"
```
willem-bd: Defaulting `NODE_HOST` to `"gateway"` ships a silently-broken sandbox path: the gateway cannot reach sandbox NodePorts at that address.

The provisioner returns `http://${NODE_HOST}:{NodePort}` so the agent can reach its sandbox. But the `gateway` Service is ClusterIP on port 8001 only — it knows nothing of a sandbox NodePort, so `gateway:<nodeport>` won't route. A user who skips the NOTES/README caveat gets unreachable sandboxes with no error at install time.

It's documented as a caveat (there's no node IP known at render time, so a working default is hard), but the silent fallback to a value that can't work is the rough edge. Consider leaving `NODE_HOST` unset and having the provisioner fail loudly when no host is configured, rather than defaulting to `"gateway"`.

deploy/helm/deer-flow/templates/nginx-deployment.yaml
zhfeng: Good catch - fixed in eec66bc9. Dropped the trailing `;` from the pattern so it matches `listen [::]:2026 default_server;`:

```
sed -i '/^[[:space:]]*listen[[:space:]]\+\[::\]:2026[[:space:]]/d'
```

Verified the pattern now deletes the IPv6 line while leaving `listen 2026 default_server;` intact (on GNU sed, the `nginx:alpine` runtime - `\+` is GNU, equivalent to `\{1,\}`). Same fix applied to `docker-compose-dev.yaml:94`, where the bug was copied from.

deploy/helm/deer-flow/templates/postgres-secret.yaml
zhfeng: Fixed in eec66bc9. Added a `deer-flow.urlEscape` helper (chained `replace` calls - Sprig lacks `urlqueryescape`, and `regexReplaceAllLiteral` treats the replacement as a regex template so `[`/`]`/`?` break it) and applied it to the password in both DSNs:

- `postgres-secret.yaml`: `postgresql://user:(urlEscape $password)@host:5432/db`
- `redis-secret.yaml`: `redis://:(urlEscape $password)@host:6379/0`

`%` is encoded first to avoid double-encoding the percent signs emitted for the other chars. The raw `postgres-password` / `redis-password` keys (POSTGRES_PASSWORD / REDIS_PASSWORD) are left unencoded - they're not URL segments.

One nuance: in bundled mode the postgres password is always `randAlphaNum 32` (URL-safe) and `.Values.postgresql.auth.password` isn't currently wired into the DSN, so the encoding there is defensive. The redis path reads the user-settable password directly - that's the case that actually bites - and render-checks correctly: `r@d:pass/#?[] %x` -> `r%40d%3Apass%2F%23%3F%5B%5D%20%25x`.

deploy/helm/deer-flow/templates/provisioner-deployment.yaml
zhfeng: Fixed in eec66bc9 - went a step beyond "fail loudly": gave it a default that actually routes. When `provisioner.nodeHost` is empty, `NODE_HOST` now comes from the downward API (`status.hostIP`) - the provisioner pod's node IP. A NodePort is exposed on *every* node, so `<node-IP>:<NodePort>` routes from the gateway on most clusters, no operator action needed:

```yaml
- name: NODE_HOST
  valueFrom:
    fieldRef:
      fieldPath: status.hostIP
```

`provisioner.nodeHost` stays as an override for CNIs/network policies that block pod->node-IP traffic (the one real gap, which #3929 calls out). NOTES.txt, `values.yaml`, and the chart README are updated.

This is a stopgap for the NodePort path. #3929 (ClusterIP + cluster-DNS URL) remains the real fix - it removes `NODE_HOST` and the external NodePort exposure entirely. Left it open as a follow-up rather than folding it into this chart PR.

.github/workflows/chart.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+      # ubuntu-latest ships with helm 3 preinstalled — no setup-helm action needed.
+
+      - name: Log in to GHCR
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | \
+            helm registry login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Package chart
+        run: helm package deploy/helm/deer-flow --destination ./packages
```
willem-bd, line 91: **Medium: no `helm lint` / `helm template` gate before publish.**

A template regression ships as an immutable OCI artifact — GHCR won't let you overwrite `--version X.Y.Z`, so a broken chart published under a tag is stuck there. `helm lint` passes today (per the PR body), but CI doesn't re-enforce it. Suggest gating before packaging:

```yaml
- run: helm lint deploy/helm/deer-flow
- run: helm template deer-flow deploy/helm/deer-flow --include-crds >/dev/null
```

Given the repo's TDD convention, a `ct lint` job or a small `helm-unittest` suite under `deploy/helm/deer-flow/tests/` would strengthen this further.

.github/workflows/chart.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  publish-chart:
+    needs: verify-versions
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
willem-bd: **Medium: action pinning is inconsistent and the PR body overstates it.**

- **Pin mismatch across workflows:** this is `actions/checkout@v4` while `container.yaml` uses `@v6`. Pick one major version across both.
- **"SHA-pinned actions" is half-true:** the `docker/*` actions are SHA-pinned, but `actions/checkout@v4`/`@v6` and `actions/attest-build-provenance@v2` are tag-pinned. Either SHA-pin all of them or soften the claim in the PR body.

Tag-pinned actions are mutable, which is exactly what pinning is meant to defend against — worth getting right on a `risk:high` supply-chain change.

deploy/helm/deer-flow/templates/provisioner-rbac.yaml
```diff
@@ -0,0 +1,77 @@
...
+metadata:
+  name: {{ include "deer-flow.fullname" . }}-provisioner
+  namespace: {{ include "deer-flow.namespace" . }}
+  labels:
+    {{- include "deer-flow.labels" . | nindent 4 }}
+    app.kubernetes.io/component: provisioner
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "pods/log", "services"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
willem-bd: **Medium: provisioner RBAC is broader than the sandbox lifecycle needs.**

With `update`/`patch`/`delete` on all pods here plus `pods/exec` (line 25) and cluster-wide namespace `create` (line 60), the provisioner SA can mutate or exec into the gateway/frontend/postgres/redis pods in this namespace — not just its sandbox pods — and create arbitrary namespaces cluster-wide.

I see narrowing is listed as deferred in the README, so this is a heads-up rather than a blocker. For production I'd at least (a) scope these verbs to sandbox pods via a label selector or run the provisioner in a dedicated namespace, and (b) soften `NOTES.txt`, which currently understates the grant as just "create Pods and Services in this namespace."

.github/workflows/chart.yaml
zhfeng: Fixed in 57697589. Added a lint + render gate before `helm package`:

```yaml
- run: helm lint deploy/helm/deer-flow
- run: helm template deer-flow deploy/helm/deer-flow --include-crds >/dev/null
```

A broken template now fails CI instead of shipping as an immutable OCI artifact. (`ct lint` / `helm-unittest` deferred - happy to add a small `helm-unittest` suite under `deploy/helm/deer-flow/tests/` in a follow-up if you'd like.)

.github/workflows/chart.yaml
zhfeng: Fixed in 57697589 - both points addressed:

- **Pin mismatch:** chart.yaml was `@v4`, container.yaml `@v6` - now both SHA-pinned to the same commit (`actions/checkout` v6.0.3, `df4cb1c0`), so the mismatch is gone.
- **"SHA-pinned actions" made true:** SHA-pinned `actions/checkout` (v6.0.3) and `actions/attest-build-provenance` (v2.4.0, `e8998f94`) across the publishing workflows (`chart.yaml`, `container.yaml`, `verify-versions.yml`), matching the existing `docker/*` pattern. The PR-body claim is now accurate.

Scoped to the three workflows this PR adds/modifies (`chart.yaml`, `container.yaml`, `verify-versions.yml`); the pre-existing workflows (`backend-unit-tests`, `e2e-tests`, `lint-check`, etc.) are out of scope and left tag-pinned.

deploy/helm/deer-flow/templates/provisioner-rbac.yaml
zhfeng: Fixed in 57697589 - did both:

(a) **Narrowed the Role:** dropped `update`/`patch` from the pods/services verbs and removed the `pods/exec` and `events` rules entirely. Audited against `docker/provisioner/app.py` - it only calls `get`/`create`/`delete` on pods and `get`/`list`/`create`/`delete` on services; `update`/`patch`/`exec`/`events` were all unused. The Role now grants exactly `get`/`list`/`watch`/`create`/`delete` on pods + pods/log + services.

(b) **Softened NOTES.txt** - it now lists the actual verbs and calls out that they apply to *all* Pods in the namespace, not just sandbox Pods (pointing to the README deferred section). The README's RBAC bullet and deferred-narrowing section are updated to match.

The remaining scope concern (verbs apply to all pods, not just sandbox pods) is still deferred - RBAC can't scope by label, so the realistic options are a dedicated sandbox namespace or admission control (OPA/Kyverno), now noted in the README.

WillemJiang (approved):

PR #3988, opened 2026-07-08
feat(backend): queue rapid same-thread messages and preserve topic card previews

### LINKED ISSUE #3806: [feat] Improve Feishu rapid-message handling with per-message response cards and per-thread queuing
### Before you start

### Problem / motivation

This proposal focuses on Feishu IM interaction UX when users send multiple messages quickly in the same topic/thread.

Image

### Proposed solution

Improve the Feishu channel UX for rapid consecutive messages:

1. Serialize or queue channel-triggered runs per DeerFlow `thread_id`.
   - If one run is active for a Feishu topic, later messages in the same mapped DeerFlow thread should wait instead of immediately triggering the generic busy reply.
   - Different DeerFlow threads should still be able to run concurrently.

2. Preserve Feishu's topic/thread reply workflow for bot responses.
   - The running/queued card should be sent as an in-thread reply to the specific Feishu user message that triggered the run.
   - A new direct message to the bot should continue to create/open a Feishu topic where the user can continue the conversation.
   - For messages already sent inside a Feishu topic, the bot card should include a compact source-message preview rendered as a markdown blockquote, so users can distinguish rapid consecutive questions in the same topic.
   - Stream and final updates should patch that same in-topic card, preserving the source-message preview without switching to ordinary non-topic replies.
 Example card content inside an existing Feishu topic:
Image

### Affected area(s)

### Alternatives considered

_No response_

### Additional context

_No response_

Zhengcy05:
Fixes #3806

## Why

Feishu topic/thread conversations get confusing when users send multiple messages quickly in the same mapped DeerFlow thread.

Before this change, a later message in the same Feishu topic could hit the generic busy reply instead of waiting behind the active run. On top of that, topic replies did not consistently preserve a per-message visual anchor across queued, running, and final card updates, which made rapid consecutive questions hard to distinguish.

## What changed

Feishu-triggered runs are now serialized per mapped DeerFlow `thread_id` inside the channel manager, so rapid follow-up messages in the same topic wait their turn instead of surfacing the generic busy reply.

For Feishu topic/thread replies, DeerFlow now keeps one response card per source message and preserves a compact source-message preview as a markdown blockquote across queued, running, and final patches. Different DeerFlow threads still run concurrently.

## Surface area

## Validation

- `cd backend && ../backend/.venv/bin/python -m ruff check .`
- `PYTHONPATH=backend backend/.venv/bin/python -m pytest backend/tests/test_channels.py -q`

## AI assistance

**Tool(s) used:** Codex

**How you used it:** I used Codex to Locate the files that need modification and identify any missing parts., add regression tests, and update the related docs. I reviewed and refined the resulting changes before preparing this PR text.


willem-bd (commented):
Reviewed the per-thread serialization and the Feishu source-preview card plumbing. The refactor cleanly extracts `_handle_chat_on_thread`, the preview key survives `_slim_metadata` into queued/running/final patches, and the eager GitHub run-policy import is circular-import-safe. Two findings below.

backend/app/channels/manager.py
```diff
@@ -1439,6 +1508,50 @@ async def _handle_chat(
...
+        )
+        serial_lock_acquired = False
+        if queued:
+            await self._publish_progress_update(
+                msg,
+                thread_id,
+                "Queued behind another request in this conversation. I’ll start working on this as soon as it finishes.",
+            )
+        if serial_state is not None:
+            await serial_state.lock.acquire()
```
willem-bd: This `await serial_state.lock.acquire()` (and the "Queued..." `_publish_progress_update` on L1517) sits **outside** the `try/finally` that calls `_finish_serialized_thread_run`. But `_begin_serialized_thread_run` at L1511 already mutated shared state: it created the `_serialized_thread_runs[(channel, thread_id)]` entry and incremented `state.waiters`.

If either await raises — the realistic case is a queued Feishu turn whose task is cancelled while waiting on `lock.acquire()` (e.g. manager shutdown), or the queued-progress publish raising — the `finally` never runs, so:
- `state.waiters` is never decremented (permanently +1 for that thread), and
- the `_serialized_thread_runs` entry is never popped (the pop in `_finish_serialized_thread_run` is gated on `waiters == 0`).

That thread's lock state then leaks for the lifetime of the manager — one entry per thread that ever has a cancelled waiter.

Fix: move both awaits inside the `try` (start the `try` right after `_begin`), keeping `serial_lock_acquired` so the `finally` still releases only when actually held:

```python
serial_state, queued = self._begin_serialized_thread_run(...)
serial_lock_acquired = False
try:
    if queued:
        await self._publish_progress_update(msg, thread_id, "Queued…")
    if serial_state is not None:
        await serial_state.lock.acquire()
        serial_lock_acquired = True
    if queued:
        await self._publish_progress_update(msg, thread_id, "thinking...")
    await self._handle_chat_on_thread(...)
finally:
    self._finish_serialized_thread_run(..., lock_acquired=serial_lock_acquired)
```

backend/app/channels/manager.py
```diff
@@ -841,6 +856,60 @@ def _resolve_session_layer(self, msg: InboundMessage) -> tuple[dict[str, Any], d
         user_layer = _as_dict(users_layer.get(msg.user_id))
         return channel_layer, user_layer
 
+    @staticmethod
+    def _channel_requires_serial_thread_runs(channel_name: str) -> bool:
+        return channel_name == "feishu"
```
willem-bd: This hardcodes the per-channel serialization knob as an if-branch on `channel_name`. The same module documents the established pattern a few inches below (the `_apply_channel_policy` block at ~L1450): _"Driven by CHANNEL_RUN_POLICY so each new webhook channel is a one-row registration, not a fresh if-branch here."_

Serialization is exactly that kind of per-channel behavior, and it belongs as a field on `ChannelRunPolicy` (e.g. `serialize_thread_runs=True`) registered once per channel — otherwise every channel that later wants queue-instead-of-busy will grow another branch here, against the repo's own convention.

backend/app/channels/manager.py
Zhengcy05: I have made corrections, and once you confirm this part is correct, I will close this comment

backend/app/channels/manager.py
Zhengcy05: I've moved the queue-same-thread behavior into `ChannelRunPolicy` via a new `serialize_thread_runs` flag, added a Feishu policy registration entry, and updated the manager to read that policy instead of special-casing Feishu by name. That keeps this aligned with the existing `fire_and_forget` / `requires_bound_identity` pattern and makes it easier to add the same behavior for other channels later without growing more manager branches.

Please let me know how you feel after you review it.

WillemJiang (approved):

PR #3990, opened 2026-07-08
Return the full-resolution image URL from image_search

DaoyuanLi2816:
## What does this PR do?

The `image_search` tool set **both** `image_url` and `thumbnail_url` to the same field:

```python
"image_url": r.get("thumbnail", ""),
"thumbnail_url": r.get("thumbnail", ""),
```

So it never returned the full-resolution image, even though the tool's docstring and `usage_hint` promise reference-quality images ("Use the 'image_url' values as reference images in image generation"). A result that has an `image` but no `thumbnail` also comes back with an empty `image_url`.

DDGS `.images()` results expose the full-resolution source under the separate `image` key (vs `thumbnail` for the preview). This reads `image` for `image_url`, matching the sibling providers `serper` (`imageUrl` vs `thumbnailUrl`) and `brave` (`properties.url` vs `thumbnail.src`) which already keep the two distinct. `image_search` is the default image-search provider in `config.example.yaml`.

Adds a regression test.

### DIFF
--- a/backend/packages/harness/deerflow/community/image_search/tools.py
+++ b/backend/packages/harness/deerflow/community/image_search/tools.py
@@ -119,7 +119,7 @@ def image_search_tool(
     normalized_results = [
         {
             "title": r.get("title", ""),
-            "image_url": r.get("thumbnail", ""),
+            "image_url": r.get("image", ""),
             "thumbnail_url": r.get("thumbnail", ""),
         }
         for r in results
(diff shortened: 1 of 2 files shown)


WillemJiang:
@DaoyuanLi2816 Thanks for your contribution. Please click the CLA button to sign the CLA before we start the review process.

WillemJiang (approved):

PR #3991, opened 2026-07-08
Fix circuit breaker wedging after a non-retriable half-open probe

DaoyuanLi2816:
## Summary

`LLMErrorHandlingMiddleware`'s circuit breaker can wedge permanently in the `half_open` state after a single **non-retriable** probe failure, causing every subsequent LLM call to fast-fail forever.

## The bug

When the circuit is `half_open`, `_check_circuit()` admits exactly one probe call by setting `_circuit_probe_in_flight = True`. The probe call then runs through `wrap_model_call` / `awrap_model_call`. The failure path only ever resets that flag on two branches:

- `_record_success()` — on a successful probe (→ `closed`)
- `_record_failure()` — on a **retriable** probe failure (→ back to `open`)

```python
if retriable:
    self._record_failure()
return self._build_user_fallback_message(exc, reason)
```

If the probe raises a **non-retriable** error (quota / auth / generic), `retriable` is `False`, so:

- `_record_failure()` is correctly **not** called — business errors must not trip the breaker (this invariant is already covered by `test_circuit_breaker_does_not_trip_on_non_retriable_errors`), **but**
- nothing resets `_circuit_probe_in_flight`.

The circuit is now stuck at `half_open` with `_circuit_probe_in_flight = True`. From then on `_check_circuit()` returns `True` (fast fail) for **every** call, and because the handler never runs again, no call can ever reach `_record_success` / `_record_failure` to clear the flag. The breaker is wedged permanently — a transient quota blip during a recovery probe takes the model offline until the process restarts.

The existing `GraphBubbleUp` handler already recognises this exact hazard and releases the probe:

```python
except GraphBubbleUp:
    with self._circuit_lock:
        if self._circuit_state == "half_open":
            self._circuit_probe_in_flight = False
    raise
```

The non-retriable exception path was simply missing the equivalent reset.

## The fix

Release the half-open probe on the non-retriable path too, mirroring the `GraphBubbleUp` handler, on both the sync and async paths:

```python
if retriable:
    self._record_failure()
else:
    # A non-retriable error consumed the half-open probe without
    # recording a failure; release the probe so the circuit can
    # admit the next probe instead of fast-failing forever.
    with self._circuit_lock:
        if self._circuit_state == "half_open":
            self._circuit_probe_in_flight = False
return self._build_user_fallback_message(exc, reason)
```

The breaker still never trips on non-retriable errors (semantics unchanged); it just no longer leaks the probe slot, so the next call admits a fresh probe.

## Tests

Adds `test_circuit_half_open_non_retriable_error_resets_probe` and its async mirror `test_async_circuit_half_open_non_retriable_error_resets_probe`. Each drives the circuit to `half_open`, admits a probe, raises a non-retriable error during it, and asserts:

- the probe is released (`_circuit_probe_in_flight is False`),
- the circuit stays `half_open` (non-retriable errors don't trip it),
- a graceful fallback `AIMessage` is still returned, and
- the real `_check_circuit()` re-admits the next probe.

Verified fail-before / pass-after: both tests fail (`assert True is False` on the wedged flag) without the fix and pass with it. Full file: `31 passed`.

```
cd backend && PYTHONPATH=. pytest tests/test_llm_error_handling_middleware.py
```

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/llm_error_handling_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/llm_error_handling_middleware.py
@@ -354,6 +354,13 @@ class LLMErrorHandlingMiddleware(AgentMiddleware[AgentState]):
                 )
                 if retriable:
                     self._record_failure()
+                else:
+                    # A non-retriable error consumed the half-open probe without
+                    # recording a failure; release the probe so the circuit can
+                    # admit the next probe instead of fast-failing forever.
+                    with self._circuit_lock:
+                        if self._circuit_state == "half_open":
+                            self._circuit_probe_in_flight = False
                 return self._build_user_fallback_message(exc, reason)
 
     @override
@@ -406,6 +413,13 @@ class LLMErrorHandlingMiddleware(AgentMiddleware[AgentState]):
                 )
                 if retriable:
                     self._record_failure()
+                else:
+                    # A non-retriable error consumed the half-open probe without
+                    # recording a failure; release the probe so the circuit can
+                    # admit the next probe instead of fast-failing forever.
+                    with self._circuit_lock:
+                        if self._circuit_state == "half_open":
+                            self._circuit_probe_in_flight = False
                 return self._build_user_fallback_message(exc, reason)
 
 
--- a/backend/packages/harness/deerflow/agents/middlewares/llm_error_handling_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/llm_error_handling_middleware.py
@@ -182,6 +182,17 @@ class LLMErrorHandlingMiddleware(AgentMiddleware[AgentState]):
                         self.circuit_recovery_timeout_sec,
                     )
 
+    def _release_half_open_probe(self) -> None:
+        """Release the in-flight half-open probe without recording a failure.
+
+        Used when something other than a classified success/failure consumes the probe (a
+        GraphBubbleUp control-flow signal, or a non-retriable error), so the circuit can admit
+        the next probe instead of fast-failing forever.
+        """
+        with self._circuit_lock:
+            if self._circuit_state == "half_open":
+                self._circuit_probe_in_flight = False
+
     def _classify_error(self, exc: BaseException) -> tuple[bool, str]:
         detail = _extract_error_detail(exc)
         lowered = detail.lower()
@@ -326,9 +337,7 @@ class LLMErrorHandlingMiddleware(AgentMiddleware[AgentState]):
                 return response
             except GraphBubbleUp:
                 # Preserve LangGraph control-flow signals (interrupt/pause/resume).
-                with self._circuit_lock:
-                    if self._circuit_state == "half_open":
-                        self._circuit_probe_in_flight = False
+                self._release_half_open_probe()
                 raise
             except Exception as exc:
                 retriable, reason = self._classify_error(exc)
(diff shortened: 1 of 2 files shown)


backend/packages/harness/deerflow/agents/middlewares/llm_error_handling_middleware.py
```diff
@@ -354,6 +354,13 @@ def wrap_model_call(
                 )
                 if retriable:
                     self._record_failure()
+                else:
+                    # A non-retriable error consumed the half-open probe without
+                    # recording a failure; release the probe so the circuit can
+                    # admit the next probe instead of fast-failing forever.
+                    with self._circuit_lock:
```
willem-bd: Optional polish (not blocking): this 3-line probe-release snippet now appears 4× in this file — the two pre-existing `GraphBubbleUp` handlers plus the two `else` branches this PR adds. A tiny helper would centralize the invariant and make future probe-state tweaks a one-liner:

```python
def _release_half_open_probe(self) -> None:
    with self._circuit_lock:
        if self._circuit_state == "half_open":
            self._circuit_probe_in_flight = False
```

Then all four sites become `self._release_half_open_probe()`. Fine to keep inline too — that just matches the existing inlined `GraphBubbleUp` style, so it's a judgment call. (Same block is mirrored in the async path at line 416.)

backend/tests/test_llm_error_handling_middleware.py
```diff
@@ -224,6 +224,73 @@ async def handler(_request) -> AIMessage:
...
+    """A non-retriable error during a half-open probe must release the probe.
+
+    Regression: the non-retriable branch neither recorded a failure (correct —
+    business errors like quota/auth must not trip the breaker) nor reset
+    ``_circuit_probe_in_flight``. So one non-retriable probe left the circuit
+    stuck at half_open with probe_in_flight=True, and every subsequent call
+    fast-failed forever because no later call could ever run the handler to
+    reach ``_record_success`` / ``_record_failure``.
+    """
+    import unittest.mock
```
willem-bd, line 237: Minor nit (optional): `import unittest.mock` inside the function body is a local pattern already used by the `GraphBubbleUp` tests above, so this is consistent. If the file is ever tidied, a top-level `import unittest.mock` (or the `monkeypatch` fixture used elsewhere in this file) would read slightly cleaner. No action needed for this PR.

willem-bd (commented):
**Verdict: looks good — real, high-severity fix. Approving.**

The non-retriable half-open path was the only exit from `wrap_model_call`/`awrap_model_call` that didn't release the probe slot, so a single transient quota/auth blip during a recovery probe wedged the circuit at `half_open` with `_circuit_probe_in_flight=True` forever — every later call fast-fails in `_check_circuit()`, so nothing ever reaches `_record_success`/`_record_failure` to recover. The fix mirrors the existing `GraphBubbleUp` handler byte-for-byte, preserves the "non-retriable errors don't trip the breaker" invariant (still no `_record_failure()` call), and the `if self._circuit_state == "half_open"` guard correctly no-ops in `closed`/`open` states. Sync + async both covered.

Tests are solid: both new tests assert the flag is released **and** that the real `_check_circuit()` re-admits the next probe — both assertions fail without the fix and pass with it. Good redundancy.

The two inline comments are optional polish only.

DaoyuanLi2816:
Extracted `_release_half_open_probe()` and switched all four call sites (both `GraphBubbleUp` handlers, both non-retriable `else` branches) to use it, per your suggestion. Left the `import unittest.mock` placement as-is per your "no action needed" note.

WillemJiang (approved):

PR #3994, opened 2026-07-08
fix(config): Make the sync checkpointer honor the unified database config

DaoyuanLi2816:
## Summary

The **sync** checkpointer factory ignores the unified `database:` config section, so embedded (`DeerFlowClient`) and TUI runs silently drop their checkpoints to an in-memory saver even when `database:` selects sqlite/postgres.

## The bug

`runtime/checkpointer/provider.py` — both `get_checkpointer()` (singleton) and `checkpointer_context()` (one-shot) read **only** the legacy `checkpointer:` section and fall back to `InMemorySaver` when it is absent:

```python
# get_checkpointer()
config = get_checkpointer_config()      # legacy checkpointer: section only
if config is None:
    ... return InMemorySaver()

# checkpointer_context()
config = get_app_config()
if config.checkpointer is None:
    yield InMemorySaver(); return
```

Every sibling resolver already honors `database:`:

- async checkpointer `async_provider.py::make_checkpointer` — legacy `checkpointer:` first, else `database:`, else memory
- sync Store `store/provider.py::_resolve_store_config` (+ `store_context` / `get_store`)
- async Store `store/async_provider.py::make_store`

So the sync checkpointer is the lone outlier. With `database: {backend: sqlite|postgres}` and **no** legacy `checkpointer:` section:

- the Store (same process, same config) persists to sqlite/postgres, but
- the sync checkpointer silently returns `InMemorySaver`.

This contradicts `backend/AGENTS.md`:

> the unified `database` section selects the Gateway's LangGraph checkpointer, LangGraph Store, and DeerFlow SQL repositories

The Gateway is unaffected (it uses the async `make_checkpointer` path). The reachable failures are the embedded/sync callers:

- `DeerFlowClient` (`client.py`) when no explicit checkpointer is passed
- the TUI (`tui/session.py`): `persistence.py` writes a `threads_meta` row to sqlite so the thread shows up in the Web UI sidebar, but the conversation checkpoints go to `InMemorySaver` and are lost when the process exits.

## The fix

Mirror the sync Store provider. Add `_resolve_checkpointer_config` (legacy `checkpointer:` precedence, else translate `database:` into a `CheckpointerConfig`) and `_get_checkpointer_config` (the lock-free loader with the `FileNotFoundError` → memory fallback that embedded callers rely on), then route both the singleton and the context manager through them:

```python
def _resolve_checkpointer_config(app_config: AppConfig) -> CheckpointerConfig:
    if app_config.checkpointer is not None:
        return app_config.checkpointer
    database = app_config.database
    if database.backend == "memory":
        return CheckpointerConfig(type="memory")
    if database.backend == "sqlite":
        return CheckpointerConfig(type="sqlite", connection_string=database.checkpointer_sqlite_path)
    if database.backend == "postgres":
        if not database.postgres_url:
            raise ValueError("database.postgres_url is required for the postgres backend")
        return CheckpointerConfig(type="postgres", connection_string=database.postgres_url)
    raise ValueError(f"Unknown database backend: {database.backend!r}")
```

This makes the sync checkpointer resolution identical to the sync Store's, and consistent with the async checkpointer's documented priority. `memory` backend still yields `InMemorySaver`; the legacy `checkpointer:` section still takes precedence.

## Tests

Adds `TestCheckpointerDatabaseConfig`, mirroring the existing `TestStoreDatabaseConfig` suite:

- `test_sync_checkpointer_context_uses_database_config` — one-shot honors `database: postgres`
- `test_sync_checkpointer_singleton_uses_database_config` — singleton honors `database: postgres`
- `test_sync_checkpointer_falls_back_to_memory_when_config_file_is_missing` — embedded no-config fallback preserved
- `test_legacy_checkpointer_config_takes_precedence` — legacy `checkpointer:` overrides `database:`
- `test_explicit_memory_database_uses_in_memory_saver` — explicit memory mode stays non-persistent

Verified fail-before / pass-after: the two `uses_database_config` tests return `InMemorySaver` (assertion fails) before the fix and pass after. Full file: `49 passed`; TUI-persistence + persistence-bootstrap suites also green.

```
cd backend && PYTHONPATH=. pytest tests/test_checkpointer.py
```

### DIFF
--- a/backend/packages/harness/deerflow/runtime/checkpointer/provider.py
+++ b/backend/packages/harness/deerflow/runtime/checkpointer/provider.py
@@ -26,8 +26,8 @@ from collections.abc import Iterator
 
 from langgraph.types import Checkpointer
 
-from deerflow.config.app_config import get_app_config
-from deerflow.config.checkpointer_config import CheckpointerConfig, ensure_config_loaded
+from deerflow.config.app_config import AppConfig, get_app_config
+from deerflow.config.checkpointer_config import CheckpointerConfig, ensure_config_loaded, get_checkpointer_config
 from deerflow.runtime.store._sqlite_utils import ensure_sqlite_parent_dir, resolve_sqlite_conn_str
 
 logger = logging.getLogger(__name__)
@@ -41,6 +41,51 @@ POSTGRES_INSTALL = (
     "langgraph-checkpoint-postgres is required for the PostgreSQL checkpointer. Install the package extra with: pip install 'deerflow-harness[postgres]' (or use: uv sync --all-packages --extra postgres when developing locally)"
 )
 POSTGRES_CONN_REQUIRED = "checkpointer.connection_string is required for the postgres backend"
+
+
+# ---------------------------------------------------------------------------
+# Config resolution
+# ---------------------------------------------------------------------------
+
+
+def _resolve_checkpointer_config(app_config: AppConfig) -> CheckpointerConfig:
+    """Resolve the checkpointer backend from legacy or unified application config.
+
+    The legacy ``checkpointer`` section remains authoritative when present so
+    Checkpointer and Store keep using the same backend. Otherwise the unified
+    ``database`` section drives the checkpointer, matching the async
+    :func:`~deerflow.runtime.checkpointer.async_provider.make_checkpointer`
+    factory and the sync Store provider's ``_resolve_store_config``.
+    """
+    if app_config.checkpointer is not None:
+        return app_config.checkpointer
+
+    database = app_config.database
+    if database.backend == "memory":
+        return CheckpointerConfig(type="memory")
+    if database.backend == "sqlite":
+        return CheckpointerConfig(type="sqlite", connection_string=database.checkpointer_sqlite_path)
+    if database.backend == "postgres":
+        if not database.postgres_url:
+            raise ValueError("database.postgres_url is required for the postgres backend")
+        return CheckpointerConfig(type="postgres", connection_string=database.postgres_url)
+    raise ValueError(f"Unknown database backend: {database.backend!r}")
+
+
+def _get_checkpointer_config() -> CheckpointerConfig:
+    """Load checkpointer config without holding the provider singleton lock."""
+    ensure_config_loaded()
+
+    # Preserve callers that initialise the legacy config singleton directly.
+    legacy_config = get_checkpointer_config()
+    if legacy_config is not None:
+        return legacy_config
+    try:
+        app_config = get_app_config()
+    except FileNotFoundError:
+        return CheckpointerConfig(type="memory")
+    return _resolve_checkpointer_config(app_config)
+
 
 # ---------------------------------------------------------------------------
 # Sync factory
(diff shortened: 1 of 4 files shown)


backend/packages/harness/deerflow/runtime/checkpointer/provider.py
```diff
@@ -42,6 +42,51 @@
...
+    Checkpointer and Store keep using the same backend. Otherwise the unified
+    ``database`` section drives the checkpointer, matching the async
+    :func:`~deerflow.runtime.checkpointer.async_provider.make_checkpointer`
+    factory and the sync Store provider's ``_resolve_store_config``.
+    """
+    if app_config.checkpointer is not None:
+        return app_config.checkpointer
+
+    database = app_config.database
+    if database is None or database.backend == "memory":
```
willem-bd, line 64: Minor divergence from the mirrored store resolver: this adds a `database is None` guard that `_resolve_store_config` in `store/provider.py` omits (there it's just `if database.backend == "memory":`, which would raise `AttributeError` on `None`). On a real `AppConfig`, `database` is never `None` (`default_factory=DatabaseConfig` plus `_apply_database_defaults`), so this branch only fires for tests using `SimpleNamespace`/`MagicMock`. The extra defensiveness is fine; if the intent is to keep the two resolvers identical, consider backfilling the same guard into the store resolver, or dropping it here and having the tests pass a real `DatabaseConfig(backend="memory")`. Non-blocking.

backend/tests/test_checkpointer.py
```diff
@@ -705,6 +705,108 @@ async def test_database_sqlite_creates_parent_dir_via_to_thread(self):
...
+    ``make_checkpointer`` factory and the sync Store provider.
+
+    Regression: ``get_checkpointer`` / ``checkpointer_context`` previously read
+    only the legacy ``checkpointer`` section and fell back to ``InMemorySaver``,
+    silently ignoring ``database``. Embedded callers (``DeerFlowClient``) and the
+    TUI then persisted Store rows to sqlite/postgres while checkpoints went to an
+    in-memory saver and were lost on exit.
+    """
+
+    def test_sync_checkpointer_context_uses_database_config(self):
```
willem-bd, line 720: Coverage gap: this new class exercises the `postgres` and `memory` branches of the database-to-checkpointer mapping, but not `sqlite` (`connection_string=database.checkpointer_sqlite_path`). A parallel `database: sqlite` case (mirroring the store suite) would cover that one-liner. Minor, not blocking.

DaoyuanLi2816:
Addressed both (non-blocking but worth fixing):
- `store/provider.py`: backfilled the `database is None` guard into `_resolve_store_config`, matching `_resolve_checkpointer_config`. Both docstrings already state the intent to keep the two resolvers identical, so this closes that gap rather than dropping the guard on the checkpointer side.
- `test_checkpointer.py`: added `test_sync_checkpointer_context_uses_sqlite_database_config`, mirroring the existing postgres test in `TestCheckpointerDatabaseConfig` but for the sqlite branch.

WillemJiang (approved):

PR #3996, opened 2026-07-08
feat(memory): add memory consolidation to synthesize fragmented facts

sontianye:
## Why

The memory system extracts facts atomically — each conversation adds individual entries like "uses React", "uses Python", "uses PostgreSQL". Over many sessions a single category can accumulate dozens of fragments that describe facets of the **same** picture. With `max_facts=100`, fragmented low-value entries crowd out higher-value new facts and waste injection budget.

This is the last piece of the memory lifecycle: **extraction → guaranteed injection → staleness review → consolidation**. The first three exist; consolidation lets the agent actively improve its own knowledge representation quality rather than only adding or removing facts.

## What changed

- Memory updater now detects **fragmented categories** (≥ `consolidation_min_facts` entries) during the normal memory-update LLM call — same invocation, zero extra API cost
- The LLM reviews each group and decides whether related facts can be **synthesized into a single richer fact** (e.g., 5 tech-stack facts → 1 "full-stack engineer: React + Python + PostgreSQL")
- Prompt surfaces the most fragmented categories first, capped at `consolidation_max_groups_per_cycle` groups with at most `consolidation_max_sources` facts each — aligned with apply-time guardrails so the LLM never sees groups it cannot act on
- Apply-time guardrails: source IDs must exist and not overlap across groups, consolidated confidence capped at source maximum (prevents inflation), results below `fact_confidence_threshold` are rejected, double-consume protection
- Feature-gated at both prompt and apply time via `consolidation_enabled` (default: `true`)
- New config fields: `consolidation_enabled`, `consolidation_min_facts`, `consolidation_max_groups_per_cycle`, `consolidation_max_sources`
- 26 tests covering candidate selection, normalization, apply guardrails, and prompt integration

## Surface area

- [x] **Agents / LangGraph** — memory updater prompt + apply logic

## Validation

```
cd backend && make lint && ruff format --check
cd backend && pytest tests/test_memory_consolidation.py tests/test_memory_staleness_review.py
# 61 passed, 0 failed
```


backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -321,12 +340,52 @@ def _normalize_memory_update_data(update_data: dict[str, Any]) -> dict[str, Any]
...
+    consolidation_raw = update_data.get("factsToConsolidate")
+    normalized_consolidation: list[dict[str, Any]] = []
+    if isinstance(consolidation_raw, list):
+        for entry in consolidation_raw:
+            if not isinstance(entry, dict):
+                continue
+            source_ids = entry.get("sourceIds")
+            if not isinstance(source_ids, list) or not source_ids:
+                continue
+            clean_ids = [sid for sid in source_ids if isinstance(sid, str) and sid]
```
willem-bd: **Duplicate `sourceIds` bypass the "≥2 distinct sources" guardrail.**

`clean_ids` filters out non-string/empty IDs but doesn't deduplicate. An LLM response like `{"sourceIds": ["f1", "f1"], ...}` passes the `len(clean_ids) < 2` check here and the `2 <= len(source_ids) <= max_sources` check at apply time. At apply, `ids_consumed.update(["f1","f1"])` collapses to `{"f1"}`, so exactly one fact is removed and one consolidated fact is written — effectively rewriting a single fact's content/category through the consolidation path, which the normal ingestion path never permits (it dedups by content and never mutates existing facts).

Suggested fix — dedupe and reject when fewer than 2 unique IDs remain:
```python
clean_ids = list(dict.fromkeys(sid for sid in source_ids if isinstance(sid, str) and sid))
```

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -807,6 +932,81 @@ def _apply_updates(
...
+                    merge_count += 1
+                    logger.info(
+                        "Consolidation merged %d facts into: %s",
+                        len(source_ids),
+                        content.strip()[:80],
+                    )
+
+                if ids_consumed:
+                    current_memory["facts"] = [f for f in current_memory.get("facts", []) if f.get("id") not in ids_consumed]
+                    current_memory["facts"].extend(new_consolidated)
```
willem-bd, line 1100: **The `max_facts` trim at the end of `_apply_updates` can drop a freshly-consolidated fact *after* its sources were deleted here — unrecoverable loss.**

Consolidation removes the N source facts (line 1007) and appends the consolidated fact (line 1008) before new facts are added and before the final `max_facts` trim. The consolidated fact's confidence is capped at `max_source_conf` (the max of its sources); if new facts arrive at higher confidence and push the count over `max_facts`, the stable confidence-descending sort in the trim can drop the consolidated fact — but its N sources are already gone. Net: all N originals and their synthesized replacement vanish, with no log line.

Consider running consolidation after the trim, or excluding `source == "consolidation"` facts from the trim for the cycle they were created.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -464,6 +523,60 @@ def _build_staleness_section(
...
+    parts: list[str] = []
+    for cat, group in sorted_candidates[:max_groups]:
+        lines: list[str] = []
+        for fact in group[:max_sources]:
+            fid = fact.get("id", "?")
+            raw_conf = fact.get("confidence", 0.0)
+            conf = float(raw_conf) if isinstance(raw_conf, (int, float)) and not isinstance(raw_conf, bool) else 0.0
+            content = str(fact.get("content", ""))
+            lines.append(f'- [{fid} | {conf:.2f}] "{content}"')
+        parts.append(f'<consolidation_candidates category="{cat}" count="{len(group)}">\n' + "\n".join(lines) + "\n</consolidation_candidates>")
```
willem-bd: **`count="{len(group)}"` advertises the full category size, but only `group[:max_sources]` facts are listed above.**

When a category holds more than `consolidation_max_sources` facts (e.g. 20 facts, max_sources=8), the LLM is told `count=20` while seeing only 8 IDs — contradicting this function's own docstring ("the LLM is never shown groups it cannot act on"). No data corruption (apply guardrails reject hallucinated IDs), but the LLM may waste output on IDs it can't see or attempt consolidations that get silently dropped.

Fix: use `min(len(group), max_sources)` for `count`, or add a note like `shown 8 of 20`.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -464,6 +523,60 @@ def _build_staleness_section(
...
+        return {}
+    by_category: dict[str, list[dict[str, Any]]] = {}
+    for fact in facts:
+        if not isinstance(fact, dict):
+            continue
+        cat = fact.get("category", "context")
+        if isinstance(cat, str) and cat.strip():
+            by_category.setdefault(cat.strip(), []).append(fact)
+    threshold = config.consolidation_min_facts
+    return {cat: group for cat, group in by_category.items() if len(group) >= threshold}
```
willem-bd: **No protected-category exemption, unlike the sibling staleness feature.**

`_select_stale_candidates` skips categories in `staleness_protected_categories` (default `["correction"]` — "explicit user feedback [that] should not be auto-pruned"), but this selector groups every category unconditionally. If a protected category ever reaches `consolidation_min_facts`, consolidation actively surfaces it for merging — the exact category the system is designed to preserve.

Low likelihood at default `min_facts=8` (corrections are rare), but the contract should be honored in both features. A one-line fix: skip categories in `config.staleness_protected_categories` (or add a parallel `consolidation_protected_categories`).

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -321,12 +340,52 @@ def _normalize_memory_update_data(update_data: dict[str, Any]) -> dict[str, Any]
...
+                _norm_conf = 0.9
+            else:
+                _f = float(_raw_conf)
+                _norm_conf = _f if math.isfinite(_f) else 0.9
+            normalized_consolidation.append(
+                {
+                    "sourceIds": clean_ids,
+                    "consolidated": {
+                        "content": content.strip(),
+                        "category": consolidated.get("category", "context") if isinstance(consolidated.get("category"), str) else "context",
```
willem-bd: **Consolidated `category` isn't stripped/emptiness-checked, unlike the `newFacts` path (`_normalize_memory_update_fact`).**

A value like `"  "` or `" preference "` is persisted verbatim, while the rest of the system strips and defaults empty categories. A padded stored category can later miss category-based matching (e.g. guaranteed-injection slots), diverging from how every other fact is normalized.

Strip + default empty to match `_normalize_memory_update_fact`.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -807,6 +932,81 @@ def _apply_updates(
...
+                        continue
+
+                    new_fact = {
+                        "id": f"fact_{uuid.uuid4().hex[:8]}",
+                        "content": content.strip(),
+                        "category": consolidated.get("category", "context"),
+                        "confidence": fact_confidence,
+                        "createdAt": now,
+                        "source": "consolidation",
+                        "consolidatedFrom": list(source_ids),
```
willem-bd, line 1081: **`sourceError` metadata is not propagated into the consolidated fact.**

This `new_fact` dict copies content/category/confidence/createdAt/source/consolidatedFrom, but facts may carry `sourceError` (the "what went wrong" guidance attached to corrections). Consolidating such facts loses that metadata silently — against the consolidation prompt's own "preserve ALL key details" rule.

Propagate `sourceError` when any source fact carries it.

willem-bd (commented):
Reviewed the memory-consolidation changes. Apply-time guardrails (source-ID existence, no-overlap, size cap, confidence cap, threshold floor) look solid and the brace-escaping in both prompts is correct. Six findings below, most-severe first. The two actionable items are #1 (dedupe source IDs) and #2 (consolidation-vs-trim ordering) — both real data-integrity gaps.

sontianye:
Thank you for the thorough and detailed review — the findings are all spot-on!

All six issues have been addressed in the latest commit:

1. **Duplicate `sourceIds`** — added `dict.fromkeys` dedup so `["f1","f1"]` collapses to `["f1"]` and is correctly rejected.
2. **Consolidation-vs-trim ordering** — moved the consolidation block to run *after* the `max_facts` trim, so evicted source facts are absent from `fact_index` and rejected by the existence guardrail before any deletion happens.
3. **`count=` mismatch** — changed to `min(len(group), max_sources)` so the LLM is only told about the IDs it can actually reference.
4. **Protected-category exemption** — `_select_consolidation_candidates` now skips categories in `staleness_protected_categories`, honouring the same contract as the staleness feature.
5. **Category not stripped** — added `strip()` + empty→`"context"` fallback in the normalization block, matching the pattern in `_normalize_memory_update_fact`.
6. **`sourceError` not propagated** — collected `sourceError` from all source facts (deduped), and carried it into the consolidated fact.

Regression tests for all six cases have been added to `test_memory_consolidation.py`.

Thanks again for the careful read!

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -852,6 +983,95 @@ def _apply_updates(
...
+                for decision in consolidation_decisions:
+                    if merge_count >= max_groups:
+                        break
+
+                    source_ids = decision.get("sourceIds", [])
+                    consolidated = decision.get("consolidated", {})
+
+                    # Guardrail: all source IDs must exist in the post-trim index
+                    # and not already consumed by an earlier merge this cycle.
+                    if any(sid in ids_consumed or sid not in fact_index for sid in source_ids):
```
willem-bd: [P1] Protected categories can still be consolidated at apply time (defense-in-depth gap vs staleness)

- Location: `updater.py:1019` (consolidation apply guardrail loop); contrast staleness apply at `updater.py:909-917`.
- Problem: Following up on the earlier protected-category point — `_select_consolidation_candidates` now excludes `staleness_protected_categories` at prompt time, but the apply path does **not** re-check. The guardrails here only verify source existence, non-consumption, size, content, confidence, and threshold. The prompt still embeds the full `current_memory` dump (`{current_memory}`), so the LLM sees correction facts even though they aren't surfaced as candidates. An LLM slip that proposes a correction fact as a consolidation source is accepted here — silently deleting/rewriting an explicit user correction.
- Evidence: The staleness pass closes exactly this hole by intersecting removals with the candidate set at apply time — `candidate_ids = {f["id"] for f in _select_stale_candidates(current_memory, config)}` (`updater.py:909-917`) — so protected/non-aged facts can never be deleted "regardless of model behavior." The consolidation loop has no equivalent. `test_protected_category_not_selected` only asserts candidate *selection*, not apply, so the gap is untested.
- Suggested fix: mirror staleness — build `allowed_ids` from `_select_consolidation_candidates(current_memory, config)` at apply time and reject any decision whose `sourceIds` aren't all in it; or at minimum reject sources whose `fact_index[sid]["category"]` is in `staleness_protected_categories`.
- Test: feed `_apply_updates` a `factsToConsolidate` entry whose `sourceIds` include a correction fact and assert it is not merged (and the correction survives).

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -852,6 +983,95 @@ def _apply_updates(
...
+
+                    source_confidences = [_coerce_source_confidence(fact_index[sid]) for sid in source_ids]
+                    max_source_conf = min(max(source_confidences), 1.0)
+
+                    # Use the LLM's returned confidence (already validated by normalization),
+                    # capped at the source maximum so consolidation cannot inflate confidence.
+                    # Falls back to max_source_conf when the field is absent or malformed.
+                    raw_llm_conf = consolidated.get("confidence")
+                    if isinstance(raw_llm_conf, (int, float)) and not isinstance(raw_llm_conf, bool) and math.isfinite(float(raw_llm_conf)):
+                        fact_confidence = min(float(raw_llm_conf), max_source_conf)
```
willem-bd: [P2] Two apply-time guardrails lack direct tests

- Location: `updater.py:1037` (confidence cap) and the `if config.consolidation_enabled:` apply gate (~line 996).
- Problem: Two guardrails introduced by this PR have no test coverage:
  1. **Confidence cap** — `fact_confidence = min(float(raw_llm_conf), max_source_conf)` prevents consolidation from inflating confidence. No test asserts that an LLM returning confidence `1.0` over sources maxing at `0.8` is capped to `0.8` (nor that a sub-threshold capped result is rejected).
  2. **Apply-time `consolidation_enabled=False`** — the real safety net when the flag is toggled mid-flight. There's a prompt-time "omitted when disabled" test, but no test that feeds `factsToConsolidate` into `_apply_updates` with the flag off and asserts nothing is merged.
- Evidence: scanned `test_memory_consolidation.py` — apply tests cover max_groups, nonexistent IDs, over-max-sources, double-consume, post-trim rejection, and sourceError; none exercise the confidence-cap value or the apply-time disabled gate.
- Suggested fix: add a test for the confidence cap (LLM conf > source max -> capped; capped value < threshold -> rejected) and one for apply-time disabled (`factsToConsolidate` present, `consolidation_enabled=False` -> no merge).

willem-bd (commented):
Thanks @sontianye. Following up on the earlier review — the duplicate-IDs, post-trim ordering, `count` attribute, category-strip, and `sourceError` points are all addressed in this head (with regression tests). Two net-new items remain: the protected-category guardrail is enforced at selection time but not at apply time (defense-in-depth gap vs the staleness pass), and two apply-time guardrails lack direct tests. Details inline.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Thanks again for the precise analysis!

**P1 — apply-time protected-category guardrail:**
Mirrored the staleness-pass pattern exactly.  is now built from `_select_consolidation_candidates(current_memory, config)` at the top of the consolidation apply block, and the inner guard checks `sid not in allowed_source_ids` alongside the existing checks. A protected or below-threshold fact proposed by the LLM is rejected regardless of model behaviour — same defense-in-depth guarantee the staleness pass provides.

Test added: `test_protected_category_rejected_at_apply_time` — feeds `factsToConsolidate` targeting correction facts into `_apply_updates` with the category below the selection threshold and asserts all correction facts survive untouched.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: **P2 — tests for confidence cap and apply-time disabled gate:**

Both tests are now in `TestReviewerFindings`:

1. **`test_confidence_cap_and_threshold_gate`** — asserts that an LLM returning confidence `1.0` over sources maxing at `0.75` is capped to `0.75`; also asserts that when the capped value (`0.65`) falls below `fact_confidence_threshold` the merge is rejected and both source facts survive.

2. **`test_apply_gate_consolidation_disabled`** — passes `factsToConsolidate` into `_apply_updates` with `consolidation_enabled=False` and asserts nothing is merged (both source facts survive untouched).

Total tests: 35, all passing.

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -133,6 +133,35 @@ class MemoryConfig(BaseModel):
         description=("Fact categories exempt from staleness review. Correction facts represent explicit user feedback and should not be auto-pruned based on age alone."),
     )
 
+    # ── Memory consolidation ────────────────────────────────────────────
+    consolidation_enabled: bool = Field(
+        default=True,
```
willem-bd: [P2] `consolidation_enabled` defaults to `true` for a lossy, irreversible mutation

- Location: `memory_config.py:137-145`
- Problem: Consolidation deletes source facts and replaces them with an LLM-synthesized fact. `consolidatedFrom` preserves the source IDs but not their original content, so a bad merge is real data loss with no in-memory rollback (the atomic temp-file+rename gives a consistent single write, not history). Shipping this ON by default for every deployment — on a `risk:high` PR — means the first bad consolidation a user hits is unrecoverable from memory alone.
- Suggested fix: consider defaulting `false` (opt-in) for one release, or document the audit/rollback story (e.g. memory-file backups) before flipping the default. The flag itself is fine; this is only about the default.
- Test: n/a (config default).

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -464,6 +527,62 @@ def _build_staleness_section(
...
+        return ""
+    # Prioritise the most fragmented categories; alphabetical tiebreak for stability.
+    sorted_candidates = sorted(candidates.items(), key=lambda kv: (-len(kv[1]), kv[0]))
+    parts: list[str] = []
+    for cat, group in sorted_candidates[:max_groups]:
+        lines: list[str] = []
+        for fact in group[:max_sources]:
+            fid = fact.get("id", "?")
+            raw_conf = fact.get("confidence", 0.0)
+            conf = float(raw_conf) if isinstance(raw_conf, (int, float)) and not isinstance(raw_conf, bool) else 0.0
```
willem-bd: [P2] Confidence coercion defaults diverge: `0.5` at apply vs `0.0` in the prompt formatter

- Location: `updater.py:577-578` (`_build_consolidation_section`) vs `_coerce_source_confidence` (~`updater.py:102`)
- Problem: `_coerce_source_confidence` (used at apply time for the confidence cap) defaults a missing/invalid source confidence to `0.5`, but this prompt formatter defaults it to `0.0`. A fact stored with `confidence: null` is therefore shown to the LLM as `0.00` in the consolidation section yet treated as `0.5` when capping the merged fact's confidence — same concept, two defaults, so a fact can read low-value in the prompt while still lifting the capped result.
- Suggested fix: reuse `_coerce_source_confidence(fact)` here instead of the inline coercion, so display and apply agree.
- Test: assert a null-confidence source renders with the same value used in the cap.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -852,6 +983,104 @@ def _apply_updates(
...
+                    # facts that the normal ingestion path would reject.
+                    if fact_confidence < config.fact_confidence_threshold:
+                        continue
+
+                    new_fact: dict[str, Any] = {
+                        "id": f"fact_{uuid.uuid4().hex[:8]}",
+                        "content": content.strip(),
+                        "category": consolidated.get("category", "context"),
+                        "confidence": fact_confidence,
+                        "createdAt": now,
```
willem-bd: [P2] Consolidated fact uses `now`, resetting the staleness clock

- Location: `updater.py:1061`
- Problem: The consolidated fact is stamped `createdAt: now`, so even when its sources were aged (stale candidates), the synthesized fact won't be eligible for staleness review for another `staleness_age_days`. Consolidation can effectively "launder" aged info into a fresh fact that evades the staleness pass.
- Suggested fix: likely acceptable since the LLM just reviewed the sources during consolidation — but make it a conscious choice. If aged info should stay reviewable, carry forward the newest source's `createdAt` (or add a `consolidatedAt` alongside `createdAt`).
- Test: assert the consolidated `createdAt` policy is intentional (either `now` or newest-source) under a stale-source scenario.

willem-bd (commented):
Thanks @sontianye. The apply-time protected-category guardrail and the two missing tests from the prior review are addressed in af7d44c0 — the `allowed_source_ids` intersection mirrors staleness, and the confidence-cap / disabled-gate tests cover both guardrails. Looks good. A few minor items remain; posting them as follow-up nits.

backend/packages/harness/deerflow/config/memory_config.py
sontianye: Good call — agreed. Consolidation is a lossy, irreversible mutation: source content is permanently replaced and only the source IDs are preserved in `consolidatedFrom`. Shipping that **on** by default is the wrong risk posture for a new feature.

Changed `consolidation_enabled` default to `false` in `memory_config.py`, and updated `config.example.yaml` with a note explaining the trade-off so operators know exactly what they are opting into when they flip it.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Correct — the divergence was real. `_build_consolidation_section` used an inline `float(raw_conf) if ... else 0.0` that defaulted missing/invalid confidence to `0.0`, while `_coerce_source_confidence` (used at apply time) defaults to `0.5`. A null-confidence fact would show `0.00` to the LLM while the cap treated it as `0.5`.

Fixed by replacing the inline coercion with a direct call to `_coerce_source_confidence(fact)` — same function, same default, prompt and apply now agree. A regression test (`test_null_confidence_renders_consistently_with_cap`) verifies both sides use `0.50`.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Made it a conscious choice: the consolidated fact now carries the **newest source's `createdAt`** rather than `now`.

The semantic: the fact's age should reflect when the underlying information was last recorded, not when synthesis happened. This means aged information that happens to get consolidated stays eligible for the next staleness pass — the clock is never reset. `consolidatedAt: now` is added alongside as an explicit audit field, so the merge timestamp is preserved without affecting staleness arithmetic.

A regression test (`test_consolidated_created_at_tracks_newest_source`) asserts `createdAt == max(source.createdAt)` and that `consolidatedAt > max(source.createdAt)`.

backend/AGENTS.md
```diff
@@ -572,6 +572,7 @@ The cached value is reused for both the blocking (`runs.wait`) and streaming (`_
...
+5b. **Consolidation pass** (same LLM invocation as step 3, no extra API call): when `consolidation_enabled` is `true` and at least one category holds `consolidation_min_facts` or more facts, `_select_consolidation_candidates` identifies fragmented categories and surfaces at most `consolidation_max_groups_per_cycle` of them (largest first) in the prompt. The LLM decides which groups to merge and proposes a synthesised fact per group. `_apply_updates` enforces guardrails: source IDs must exist and must not overlap across groups, group size is capped at `consolidation_max_sources`, the merged fact's confidence cannot exceed the source maximum, and facts below `fact_confidence_threshold` are not written.
```
willem-bd, line 575: **Doc gap (safety property).** This step enumerates the apply-time guardrails but omits a key one: protected categories (`staleness_protected_categories`, default `correction`) are excluded from consolidation candidates — enforced both in `_select_consolidation_candidates` and the apply-time allowlist (`allowed_source_ids`). "Explicit user feedback is never silently merged away" is the central safety property here, so worth calling it out, mirroring how the staleness step documents its own protected-category exclusion.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -852,6 +982,110 @@ def _apply_updates(
...
+                    # same gate applied to newFacts, so consolidation never admits
+                    # facts that the normal ingestion path would reject.
+                    if fact_confidence < config.fact_confidence_threshold:
+                        continue
+
+                    # Carry the newest source's createdAt so the staleness clock
+                    # reflects the age of the underlying information, not when
+                    # synthesis happened.  consolidatedAt records the merge time
+                    # for audit without resetting staleness eligibility.
+                    source_created_at = max(fact_index[sid].get("createdAt") or now for sid in source_ids)
```
willem-bd: This string `max()` over `createdAt` is the one spot that doesn't get the "corrupted / manually-edited memory file" hardening that `_coerce_source_confidence` was added for. Two concerns:
1. If a file stores a numeric `createdAt` (e.g. a unix timestamp), `max([str, int])` raises `TypeError` and aborts the whole memory update.
2. String comparison assumes a uniform `Z` suffix; a file mixing `Z` and `+00:00` orders wrong lexicographically (same instant → the `Z` one is picked as "newer"), skewing the merged fact's staleness clock.

The system already has `_parse_fact_datetime` for correct chronological comparison — routing through it (or a `_coerce`-style guard) would make this crash-proof and consistent with how staleness reasons about age.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -852,6 +982,110 @@ def _apply_updates(
...
+                    # Guardrail: 2..max_sources per group
+                    if not (2 <= len(source_ids) <= max_sources):
+                        continue
+
+                    content = consolidated.get("content", "")
+                    if not isinstance(content, str) or not content.strip():
+                        continue
+
+                    source_confidences = [_coerce_source_confidence(fact_index[sid]) for sid in source_ids]
+                    max_source_conf = min(max(source_confidences), 1.0)
```
willem-bd: `_coerce_source_confidence` already clamps to `[0, 1]`, so the outer `min(..., 1.0)` can never bind — dead code that slightly mis-states intent (it reads as if the clamp were needed). Nit: either drop it or leave a note that it's purely defensive.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -321,12 +340,56 @@ def _normalize_memory_update_data(update_data: dict[str, Any]) -> dict[str, Any]
...
+                continue
+            # Normalize confidence: reject booleans (bool subclasses int, so the
+            # isinstance check alone would silently accept True/False), coerce to float,
+            # and reject non-finite values — matching _normalize_memory_update_fact.
+            _raw_conf = consolidated.get("confidence", 0.9)
+            if isinstance(_raw_conf, bool) or not isinstance(_raw_conf, (int, float)):
+                _norm_conf = 0.9
+            else:
+                _f = float(_raw_conf)
+                _norm_conf = _f if math.isfinite(_f) else 0.9
```
willem-bd, line 372: Two small nits on the consolidated-confidence path:
1. Unlike `_normalize_memory_update_fact`, this keeps out-of-range values (e.g. `1.5`) — only finiteness / non-bool is checked. Not exploitable today because the apply-time cap pins it to `max_source_conf <= 1.0`, but if that cap is ever loosened the divergence becomes a foot-gun.
2. The null/bool/non-finite coercion now lives in four places (`_validate_confidence`, `_coerce_source_confidence`, here, and the apply-time check at ~L1046). `_coerce_source_confidence` could subsume the latter two.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -852,6 +982,110 @@ def _apply_updates(
...
+                    max_source_conf = min(max(source_confidences), 1.0)
+
+                    # Use the LLM's returned confidence (already validated by normalization),
+                    # capped at the source maximum so consolidation cannot inflate confidence.
+                    # Falls back to max_source_conf when the field is absent or malformed.
+                    raw_llm_conf = consolidated.get("confidence")
+                    if isinstance(raw_llm_conf, (int, float)) and not isinstance(raw_llm_conf, bool) and math.isfinite(float(raw_llm_conf)):
+                        fact_confidence = min(float(raw_llm_conf), max_source_conf)
+                    else:
+                        fact_confidence = max_source_conf
```
willem-bd, line 1054: **Test gap.** This fallback branch (LLM omits / malforms `confidence`) isn't directly exercised — `test_confidence_cap_and_threshold_gate` only covers the present-confidence path. Since this sits on the loss-protection side, an explicit test (e.g. LLM returns no `confidence` field → result uses `max_source_conf`) would be worthwhile.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -464,6 +527,61 @@ def _build_staleness_section(
...
+    # Prioritise the most fragmented categories; alphabetical tiebreak for stability.
+    sorted_candidates = sorted(candidates.items(), key=lambda kv: (-len(kv[1]), kv[0]))
+    parts: list[str] = []
+    for cat, group in sorted_candidates[:max_groups]:
+        lines: list[str] = []
+        for fact in group[:max_sources]:
+            fid = fact.get("id", "?")
+            conf = _coerce_source_confidence(fact)
+            content = str(fact.get("content", ""))
+            lines.append(f'- [{fid} | {conf:.2f}] "{content}"')
```
willem-bd, line 579: Raw fact content is injected inside `"..."`. Content containing a literal `"` or a closing `</consolidation_candidates>` tag could confuse the model. This matches the existing staleness section (same pattern), so it's not a regression — but given the recent prompt-injection hardening focus (b36d7194), worth a deliberate decision on whether memory-fact content warrants the same neutralization.

willem-bd (commented):
Thanks for this — really nicely structured. The consolidation pass faithfully mirrors the staleness-review pattern: candidates are recomputed at apply time and used as an allowlist, so hallucinated IDs / protected categories / over-merging / confidence inflation are all rejected regardless of model behavior. Default-off is the right call for a lossy mutation, the post-trim ordering correctly prevents the only real data-loss vector, and the test suite (38 tests) is thorough.

Posting a few inline comments below — mostly docs and minor hardening. The one non-doc item I'd address before merge is the `createdAt` handling.

Two notes that don't attach to a single line:
- **PR description says `consolidation_enabled` defaults to `true`, but the code / `config.example.yaml` / `test_consolidation_enabled_defaults_to_false` all default to `false`.** The code is the safer choice; the description is stale and should be corrected so reviewers don't conclude this mutates memory on first deploy.
- **The core risk to sign off on** is the acknowledged irreversibility: `consolidatedFrom` keeps source IDs but not their text, so a bad synthesis permanently destroys detail. The default-off / opt-in framing handles this correctly — just flagging it as the gate for ever flipping the default.

backend/AGENTS.md
sontianye: Good catch — the protected-category exclusion is the central safety property of this feature and deserved explicit documentation. Expanded the guardrails comment to say:

> must be in `allowed_source_ids` — the set built from `_select_consolidation_candidates`, which excludes categories in `staleness_protected_categories` (default: `correction`). This mirrors the staleness apply-time check and ensures explicit user feedback is never silently merged away regardless of model behaviour.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Both concerns are real and fixed. Replaced the string `max()` with a datetime-aware comparison via `_parse_fact_datetime`, mirroring exactly how staleness computes age:

```python
_fallback_dt = _parse_fact_datetime(now) or datetime.now(UTC)
_source_dts = [_parse_fact_datetime(fact_index[sid].get("createdAt") or "") or _fallback_dt for sid in source_ids]
source_created_at = max(_source_dts).isoformat().removesuffix("+00:00") + "Z"
```

A numeric `createdAt` is caught by `_parse_fact_datetime`'s `TypeError` guard and falls back to `_fallback_dt`; mixed `Z`/`+00:00` formats are now compared as aware `datetime` objects rather than as strings. The existing `test_consolidated_created_at_tracks_newest_source` still exercises this path.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Correct — dead code removed. `_coerce_source_confidence` already guarantees its return is in `[0.0, 1.0]` via the final `max(0.0, min(val, 1.0))`, so `max(source_confidences) ≤ 1.0` by contract and the outer `min(..., 1.0)` could never bind. Changed to plain `max(source_confidences)` with a comment explaining the invariant.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Both points addressed:

1. **Out-of-range values** — added `max(0.0, min(float(raw_llm_conf), 1.0))` clamping before applying the source cap, so values like `1.5` are clipped to `1.0` first. Safe today via the cap, but now defensively correct if the cap is ever relaxed.

2. **Refactoring** — kept the coercion inline rather than folding it into `_coerce_source_confidence`. The two contexts have genuinely different fallback semantics: source facts default to `0.5` (uncertain but plausible, and the system can't improve on that); the LLM's returned confidence defaults to `max_source_conf` (the natural conservative ceiling). A combined helper would need a caller-supplied default parameter, making it less readable than the two explicit call sites. The duplication is minimal (one conditional expression each).

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Added `test_confidence_fallback_to_max_source_when_llm_omits_field`: sources at 0.85/0.75, LLM returns a `consolidated` dict with no `confidence` key, asserts `merged[0]["confidence"] == pytest.approx(0.85)` (the `max_source_conf` fallback).

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Noted — and deliberately left as-is for this PR. The pattern (raw fact content inside `"..."` in prompt XML) is identical to the existing staleness section, so this is not a regression. A proper fix would require escaping all fact content used across both sections, which is a meaningful separate change. Happy to follow up in a dedicated PR once this one lands.

WillemJiang (approved):

PR #3997, opened 2026-07-08
fix(runtime): Atomic create_or_reject across workers, fixing race conditions for multiple active Runs.

leoredfish:
# 问题

在多 Worker 部署环境下（`GATEWAY_WORKERS > 1`），`create_or_reject` 使用 `asyncio.Lock` 仅在单进程内生效。当两个请求被分发到不同 Worker 时，各自进程内的内存锁无法互斥，导致**同一 Thread 上可以同时存在多个活跃 Run**，违反了 `multitask_strategy="reject"` 的预期行为。

# 方案

在 `RunModel` 表上添加**部分唯一索引**（partial unique index），在数据库层面强制约束：每个 Thread 最多存在一个状态为 `pending` 或 `running` 的 Run。`asyncio.Lock` 保留为进程内快速路径，DB 唯一索引作为跨 Worker 的最终防线。

## 改动点

1. **`model.py`** — 添加 `ix_one_active_run_per_thread` 部分唯一索引
   - 约束条件：`thread_id` unique，`WHERE status IN ('pending', 'running')`
   - PostgreSQL（`postgresql_where`）与 SQLite（`sqlite_where`）双方言支持
2. **`0004_one_active_run_per_thread` 迁移（新增）** — 为存量数据库创建索引
   - 建索引前先执行 `_cleanup_duplicate_active_runs()`：用 `ROW_NUMBER()` 窗口函数对每个 thread 保留最新的 active run，其余标记为 `interrupted` 并写入清理说明，避免脏数据（正是本竞态产生的）阻塞 `alembic upgrade head` / Gateway 启动
   - 幂等：索引已存在则跳过
3. **`manager.py`**
   - `_persist_new_run_to_store` 捕获 `IntegrityError` → 转为 `ConflictError`（HTTP 409）
   - **interrupt/rollback 调序**：先将所有 inflight run（含 `finalizing` 的活跃 task run，写 `interrupted` 是幂等的）持久化为 `interrupted`，再插入新 run，避免撞索引产生误导性 409
   - `_persist_status` 失败不再静默：收集失败的 run_id 并抛出 `RuntimeError`，显式暴露内存/DB 不一致，而非滑向误导性 409
   - 保留 `persisted` 标志 + `finally` 回滚：任何持久化失败（含 cancellation）都不会在内存中留下 phantom run

## Review 意见处理（@willem-bd）

| Review 意见 | 处理 |
| --- | --- |
| [P1] interrupt/rollback 回归 409 | 调序：先持久化旧 run 中断状态再插入新 run；含活跃 task（`finalizing`）路径 |
| [P1] 缺少 Alembic 迁移 | 新增 `0004` 迁移 |
| [P1] 迁移在存量脏数据下失败 | `_cleanup_duplicate_active_runs()` 先 reconcile 再建索引 |
| [P1] `except ConflictError` 收窄导致 phantom run 泄漏 | 恢复 `finally` 兜底回滚（覆盖 cancellation），两个既有测试恢复通过 |
| [P2] `_persist_status` 返回值被静默忽略 | 检查返回值，失败抛 `RuntimeError` |

## 测试

- **`test_cross_worker_race.py`（新增 8 个）**：reject 拒绝第二个活跃 run、完成后放行、跨 thread 独立、IntegrityError→ConflictError 转换与回滚、interrupt/rollback 策略取消旧 run 并创建新 run、**活跃 task 场景下 interrupt 先持久化再插入**（回归测试）
- **`test_persistence_bootstrap.py`（新增 1 个）**：端到端迁移测试——升级到 0003 → seed 同 thread 两条 pending 行 → 升级到 0004 → 断言旧行被 reconcile、索引建成且生效、跨 thread 不受影响
- **`test_run_manager.py`（新增 1 个，参数化 2 case）**：`_persist_status` 两种失败模式（`update_status` 抛异常 / 行丢失且 snapshot 恢复失败）均抛 `RuntimeError` 而非 409，无 phantom run，内存/DB 分歧显式暴露
- 适配存量用例：`test_owner_isolation.py`、`test_run_repository.py` 等约 40 个用例适配新约束
- 全量：**6636 passed, 0 failures**，lint（ruff check + format）全绿

## 真实环境验证

| 方式 | 结果 |
| --- | --- |
| 4 Worker + SQLite，12 并发请求 | 修复前：4 × 200 → 修复后：1 × 200 + 11 × 409 |

## 滚动部署说明

迁移（建索引 + 数据清理）由第一个新版本 Worker 启动时执行。在滚动发布窗口期内，仍在运行旧代码的 Worker 不会将 `IntegrityError` 转换为 409：

- 旧 Worker 上的并发 create 撞索引会返回 500（而非 409）
- 旧 Worker 的 interrupt 路径（旧顺序：先插新 run）撞索引也会返回 500
- 数据清理可能将旧 Worker 内存中仍在执行的 run 在 DB 中标记为 `interrupted`，run 结束时会被覆写为终态（无害，日志可见状态变化）

属一次性升级窗口风险。对可用性敏感的部署，建议在低峰期发布，或采用先停流量再滚动的方式。

Part of #3948 (Work Item 2).

### DIFF
--- a/backend/packages/harness/deerflow/persistence/migrations/versions/0004_one_active_run_per_thread.py
+++ b/backend/packages/harness/deerflow/persistence/migrations/versions/0004_one_active_run_per_thread.py
@@ -1,0 +1,64 @@
+"""Add partial unique index to enforce at most one active run per thread.
+
+Revision ID: 0004_one_active_run_per_thread
+Revises: 0003_scheduled_tasks
+Create Date: 2026-07-09
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+from deerflow.persistence.migrations._helpers import _inspector
+
+revision: str = "0004_one_active_run_per_thread"
+down_revision: str | Sequence[str] | None = "0003_scheduled_tasks"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+INDEX_NAME = "ix_one_active_run_per_thread"
+
+
+def _index_exists() -> bool:
+    insp = _inspector()
+    if "runs" not in insp.get_table_names():
+        return True  # nothing to add/drop
+    for idx in insp.get_indexes("runs"):
+        if idx.get("name") == INDEX_NAME:
+            return True
+    return False
+
+
+def upgrade() -> None:
+    if _index_exists():
+        return
+
+    bind = op.get_bind()
+    dialect_name = bind.dialect.name
+
+    if dialect_name == "postgresql":
+        op.create_index(
+            INDEX_NAME,
+            "runs",
+            ["thread_id"],
+            unique=True,
+            postgresql_where=sa.text("status IN ('pending', 'running')"),
+        )
+    else:
+        # SQLite and other dialects
+        op.create_index(
+            INDEX_NAME,
+            "runs",
+            ["thread_id"],
+            unique=True,
+            sqlite_where=sa.text("status IN ('pending', 'running')"),
+        )
+
+
+def downgrade() -> None:
+    if not _index_exists():
+        return
+    op.drop_index(INDEX_NAME, table_name="runs")
--- a/backend/packages/harness/deerflow/persistence/migrations/versions/0004_one_active_run_per_thread.py
+++ b/backend/packages/harness/deerflow/persistence/migrations/versions/0004_one_active_run_per_thread.py
@@ -7,6 +7,7 @@ Create Date: 2026-07-09
 
 from __future__ import annotations
 
+import logging
 from collections.abc import Sequence
 
 import sqlalchemy as sa
@@ -14,12 +15,16 @@ from alembic import op
 
 from deerflow.persistence.migrations._helpers import _inspector
 
+logger = logging.getLogger(__name__)
+
 revision: str = "0004_one_active_run_per_thread"
 down_revision: str | Sequence[str] | None = "0003_scheduled_tasks"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
 INDEX_NAME = "ix_one_active_run_per_thread"
+
+_CLEANUP_ERROR = "Cleaned up during migration 0004: duplicate active runs per thread"
 
 
 def _index_exists() -> bool:
@@ -32,12 +37,58 @@ def _index_exists() -> bool:
     return False
 
 
+def _cleanup_duplicate_active_runs(connection, dialect_name: str) -> int:
+    """Mark duplicate active runs as interrupted, keeping only the newest per thread.
+
+    If a thread has multiple pending/running rows (e.g. from a pre-fix
+    multi-worker race), the partial unique index cannot be created.  This
+    helper keeps the most-recently-created active run per thread and marks
+    the rest as ``interrupted``.
+    """
+    if dialect_name == "postgresql":
+        time_expr = "NOW()"
+    else:
+        time_expr = "datetime('now')"
+
+    result = connection.execute(
+        sa.text(f"""
+            UPDATE runs SET
+                status = 'interrupted',
+                error = :error,
+                updated_at = {time_expr}
+            WHERE run_id IN (
+                SELECT run_id FROM (
+                    SELECT run_id,
+                        ROW_NUMBER() OVER (
+                            PARTITION BY thread_id ORDER BY created_at DESC
+                        ) AS rn
+                    FROM runs
+                    WHERE status IN ('pending', 'running')
+                ) ranked
+                WHERE rn > 1
+            )
+        """),
+        {"error": _CLEANUP_ERROR},
+    )
+    cleaned = result.rowcount
+    if cleaned:
+        logger.warning(
+            "Migration 0004: cleaned up %d duplicate active run(s) before creating unique index",
+            cleaned,
+        )
+    return cleaned
+
+
 def upgrade() -> None:
     if _index_exists():
         return
 
     bind = op.get_bind()
     dialect_name = bind.dialect.name
+
+    # Clean up any pre-existing duplicate active runs that would prevent
+    # the unique index from being created.
+    _cleanup_duplicate_active_runs(bind, dialect_name)
 
     if dialect_name == "postgresql":
         op.create_index(
--- a/backend/packages/harness/deerflow/persistence/run/model.py
+++ b/backend/packages/harness/deerflow/persistence/run/model.py
@@ -47,4 +47,13 @@ class RunRow(Base):
     created_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), default=lambda: datetime.now(UTC))
     updated_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), default=lambda: datetime.now(UTC), onupdate=lambda: datetime.now(UTC))
 
-    __table_args__ = (Index("ix_runs_thread_status", "thread_id", "status"),)
+    __table_args__ = (
+        Index("ix_runs_thread_status", "thread_id", "status"),
+        Index(
+            "ix_one_active_run_per_thread",
+            "thread_id",
+            unique=True,
+            postgresql_where=(status.in_(["pending", "running"])),
+            sqlite_where=(status.in_(["pending", "running"])),
+        ),
+    )
--- a/backend/packages/harness/deerflow/runtime/runs/manager.py
+++ b/backend/packages/harness/deerflow/runtime/runs/manager.py
@@ -231,14 +231,26 @@ class RunManager:
         Unlike follow-up status/model updates, failures are propagated so the
         caller can treat creation as failed. Rollback is the caller's
         responsibility after inserting the record into ``_runs``.
+
+        When a partial unique index on ``(thread_id, status)`` rejects a
+        duplicate active run (cross-worker race), translate the database
+        constraint violation into a ``ConflictError`` so the caller can
+        respond with an appropriate HTTP 409.
         """
+        from sqlalchemy.exc import IntegrityError
+
         if self._store is None:
             return
-        await self._call_store_with_retry(
-            "put",
-            record.run_id,
-            lambda: self._store.put(record.run_id, **self._store_put_payload(record)),
-        )
+        try:
+            await self._call_store_with_retry(
+                "put",
+                record.run_id,
+                lambda: self._store.put(record.run_id, **self._store_put_payload(record)),
+            )
+        except IntegrityError:
+            raise ConflictError(
+                f"Thread {record.thread_id} already has an active run"
+            ) from None
 
     async def _persist_to_store(self, record: RunRecord, *, error: str | None = None) -> bool:
         """Best-effort persist run record to backing store."""
@@ -619,14 +631,16 @@ class RunManager:
         already has a pending/running run.  For ``interrupt``/``rollback``,
         cancels inflight runs before creating.
 
-        This method holds the lock across both the check and the insert,
-        eliminating the TOCTOU race in separate ``has_inflight`` + ``create``.
+        For interrupt/rollback the old-run status is persisted to the store
+        *before* inserting the new run so the partial unique index does not
+        fire.  This means old runs are committed as interrupted even when the
+        subsequent new-run insert fails — a trade-off that is correct because
+        the caller explicitly requested cancellation of inflight work.
         """
         run_id = str(uuid.uuid4())
         now = _now_iso()
 
         _supported_strategies = ("reject", "interrupt", "rollback")
-        interrupted_records: list[RunRecord] = []
 
         async with self._lock:
             if multitask_strategy not in _supported_strategies:
@@ -644,6 +658,22 @@ class RunManager:
                     thread_id,
                     multitask_strategy,
                 )
+                for r in inflight:
+                    if r.finalizing:
+                        continue
+                    r.abort_action = multitask_strategy
+                    r.abort_event.set()
+                    task_active = r.task is not None and not r.task.done()
+                    r.finalizing = task_active
+                    if task_active:
+                        r.task.cancel()
+                    r.status = RunStatus.interrupted
+                    r.updated_at = now
+                # Persist interruptions before inserting the new run so
+                # the DB partial unique index won't see two active rows.
+                for r in inflight:
+                    if not r.finalizing:
+                        await self._persist_status(r, RunStatus.interrupted)
 
             record = RunRecord(
                 run_id=run_id,
@@ -665,8 +695,10 @@ class RunManager:
             try:
                 await self._persist_new_run_to_store(record)
                 persisted = True
+            except ConflictError:
+                raise
             except Exception:
-                logger.warning("Failed to persist run %s; rolled back in-memory record", run_id, exc_info=True)
+                logger.warning("Failed to persist run %s; rolling back in-memory record", run_id, exc_info=True)
                 raise
             finally:
                 # Also covers cancellation, which bypasses ``except Exception``.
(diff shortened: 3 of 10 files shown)


backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -231,14 +231,26 @@ async def _persist_new_run_to_store(self, record: RunRecord) -> None:
...
-            lambda: self._store.put(record.run_id, **self._store_put_payload(record)),
-        )
+        try:
+            await self._call_store_with_retry(
+                "put",
+                record.run_id,
+                lambda: self._store.put(record.run_id, **self._store_put_payload(record)),
+            )
+        except IntegrityError:
+            raise ConflictError(
```
willem-bd: ⚠️ This catch (together with the partial unique index) breaks the `interrupt` and `rollback` multitask strategies.

In `create_or_reject`, the **new** run is persisted to the DB *before* the existing inflight runs are de-activated:

1. new `RunRecord(status=pending)` is built and indexed in `_runs`;
2. `await self._persist_new_run_to_store(record)` — INSERTs a second `pending` row on the same `thread_id`;
3. only *after* this succeeds are the old inflight runs flipped to `interrupted` in memory and persisted via `_persist_status`.

At step 2 the old run is still `pending`/`running` in the DB, so `ix_one_active_run_per_thread` fires `IntegrityError` → `ConflictError` right here. The cancellation block (after the persist) is never reached, and the caller gets a 409 instead of an interrupt.

These strategies are part of the public API (`thread_runs.py:76`: `Literal["reject","rollback","interrupt","enqueue"]`) and are only used *when an inflight run exists* — i.e. exactly the case they're meant to handle.

Fix: persist the old runs as `interrupted` (`update_status`) **before** inserting the new run in `create_or_reject`. The new tests only exercise `reject`, which is why this slips through.

backend/packages/harness/deerflow/persistence/run/model.py
```diff
@@ -47,4 +47,13 @@ class RunRow(Base):
...
     updated_at: Mapped[datetime] = mapped_column(DateTime(timezone=True), default=lambda: datetime.now(UTC), onupdate=lambda: datetime.now(UTC))
 
-    __table_args__ = (Index("ix_runs_thread_status", "thread_id", "status"),)
+    __table_args__ = (
+        Index("ix_runs_thread_status", "thread_id", "status"),
+        Index(
+            "ix_one_active_run_per_thread",
+            "thread_id",
+            unique=True,
+            postgresql_where=(status.in_(["pending", "running"])),
```
willem-bd, line 56: ⚠️ No Alembic revision ships this index — existing deployments get no protection, which is exactly the population this PR targets.

The index is added to the ORM model only; no new file exists under `migrations/versions/` (still `0001`/`0002`/`0003`). The bootstrap layer (`bootstrap.py:10-11`) explicitly states: *"Any new ORM column / table / index must ship as a revision under `migrations/versions/`."* Tracing the three bootstrap branches:

- **Existing / versioned prod DB** (the target — `GATEWAY_WORKERS > 1`): `alembic upgrade head` is a no-op at `0003`; the index is never created.
- **Legacy DB**: `create_all(checkfirst=True)` skips the already-existing `runs` table wholesale; SQLAlchemy emits `CREATE INDEX` only alongside `CREATE TABLE`, never as a standalone add to an existing table.
- **Fresh / empty DB**: the index is created — but a brand-new deployment isn't the one with the race.

So the deployments this PR claims to fix upgrade and ship **without** the constraint, and `RunManager.create_or_reject` silently relies on a guarantee that doesn't exist. `test_cross_worker_race.py` passes because each test spins up a fresh tmp SQLite DB (the empty branch), and no drift test enforces index parity (`grep compare_metadata` → 0 hits).

Fix: add a `0004_*` revision using the existing dialect-aware helpers in `migrations/_helpers.py` to `op.create_index("ix_one_active_run_per_thread", "runs", ["thread_id"], unique=True, postgresql_where=..., sqlite_where=...)` (and drop it on downgrade).

willem-bd (commented):
Reviewed at medium effort. The approach — a DB-level partial unique index as the cross-worker backstop that an in-process `asyncio.Lock` can't provide — is the right idea. Two issues survive verification, and each independently blocks this PR (details inline):

1. **`interrupt`/`rollback` regress to HTTP 409** whenever an inflight run exists, because the new run is persisted *before* the old runs are de-activated in the DB.
2. **No Alembic migration ships the index**, so existing multi-worker deployments (the ones that actually have the race) never get the constraint — only fresh DBs do.

Neither is caught by the new test suite: it only covers `reject`, and every test spins up a fresh temp SQLite DB (the empty-bootstrap branch).

backend/packages/harness/deerflow/persistence/run/model.py
leoredfish: The user wants a concise English response to this specific CR comment about the missing Alembic migration. I need to reply in a GitHub PR comment style.
Good catch, completely agree — a migration is necessary for this to have the intended effect in production. Added `0004_one_active_run_per_thread.py`:

- `op.create_index` on the existing `runs` table with dialect-aware partial unique index (`postgresql_where` / `sqlite_where`)
- `_index_exists()` guard for idempotency — safe against re-run and fresh `create_all` deployments that already have it
- `downgrade()` drops the index for clean rollback

Tested: existing SQLite DB at revision 0003 → `alembic upgrade head` → index created; `downgrade -1` → index dropped; re-upgrade idempotent.

backend/packages/harness/deerflow/runtime/runs/manager.py
leoredfish: Good catch, thanks! You're spot on — the ordering meant the partial unique index rejected all strategies, not just reject.

Fix: Reordered create_or_reject so _persist_status(r, interrupted) for inflight runs happens before _persist_new_run_to_store for the new run. The DB sees zero active rows by the time the INSERT runs.

Tests added:

test_interrupt_strategy_cancels_old_and_creates_new — pending predecessor, interrupt succeeds
test_rollback_strategy_cancels_old_and_creates_new — same for rollback
test_interrupt_after_completed_is_safe — already-finished predecessor, no-op interrupt → normal create
Local 4-worker end-to-end: 15 rounds × 6 concurrent, every round 1 × 200 + 5 × 409 ✅

backend/packages/harness/deerflow/persistence/migrations/versions/0004_one_active_run_per_thread.py
```diff
@@ -0,0 +1,64 @@
...
+
+    bind = op.get_bind()
+    dialect_name = bind.dialect.name
+
+    if dialect_name == "postgresql":
+        op.create_index(
+            INDEX_NAME,
+            "runs",
+            ["thread_id"],
+            unique=True,
```
willem-bd, line 98: [P1] Migration can fail (and block upgrade/startup) on DBs with pre-existing duplicate active runs

- Location: `0004_one_active_run_per_thread.py:43-58` (`upgrade()` -> `op.create_index(..., unique=True, ...)`)
- Problem: `CREATE UNIQUE INDEX ... WHERE status IN ('pending','running')` raises if the `runs` table already holds two+ rows with the same `thread_id` in pending/running. That state is exactly what this PR fixes (the cross-worker race) plus crashed-worker "stuck" rows — both plausible on a long-running prod DB. The failure aborts `alembic upgrade head`, and since `bootstrap_schema` runs `upgrade head` at Gateway startup (versioned branch), it can block deployment or startup for affected tenants.
- Evidence: `upgrade()` calls `op.create_index(..., unique=True, postgresql_where=...)` with no prior reconciliation. `_index_exists()` only guards the index *name*, not data validity, so it doesn't help. `RunManager.reconcile_orphaned_inflight_runs` exists for precisely the stuck-run case but isn't invoked here.
- Suggested fix: before `op.create_index`, reconcile violating rows — for each `thread_id` with >1 active run, mark all but the newest as `error`/`interrupted` (reuse the `reconcile_orphaned_inflight_runs` logic), then create the index.
- Test: seed a DB with two pending rows on the same thread, run `alembic upgrade head`, assert the index is created and the duplicate active rows are reconciled (not left pending).

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -637,13 +653,32 @@ async def create_or_reject(
...
+                    r.finalizing = task_active
+                    if task_active:
+                        r.task.cancel()
+                    r.status = RunStatus.interrupted
+                    r.updated_at = now
+                # Persist interruptions before inserting the new run so the
+                # DB partial unique index won't see two active rows.
+                for r in inflight:
+                    if not r.finalizing:
+                        await self._persist_status(r, RunStatus.interrupted)
```
willem-bd: [P2] Best-effort interrupted-status persist is now a silent gate for the new insert

- Location: `manager.py:679-681` (the persist-interrupted loop inside `create_or_reject`)
- Problem: `_persist_status` is best-effort — it catches all exceptions, logs a warning, and returns `False` on failure (it never raises). Its return value is ignored here. If the `update_status` write fails (transient SQLite lock, missing-row recovery failure), the old run is already `interrupted` in memory but still `pending`/`running` in the DB. The new-run INSERT that immediately follows then trips `ix_one_active_run_per_thread` -> `ConflictError` (409), so the interrupt/rollback fails for the caller, and memory/DB disagree on the old run's status.
- Evidence: `_persist_status` returns `bool` and swallows exceptions (`logger.warning(...); return False`); this call site discards the result. The reorder made this write a hard precondition for the new insert but kept it best-effort.
- Suggested fix: check the return value; on `False`, either retry/raise so the caller sees a real error instead of a misleading 409, or make the policy explicit — but don't silently leave memory and DB inconsistent.
- Test: mock `store.update_status` to fail during an interrupt `create_or_reject` and assert the outcome is intentional (not an incidental 409 over a half-cancelled old run).

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -661,35 +696,13 @@ async def create_or_reject(
...
-            persisted = False
             try:
                 await self._persist_new_run_to_store(record)
-                persisted = True
-            except Exception:
-                logger.warning("Failed to persist run %s; rolled back in-memory record", run_id, exc_info=True)
+            except ConflictError:
+                self._runs.pop(run_id, None)
+                self._unindex_run_locked(run_id, record.thread_id)
                 raise
```
willem-bd, line 713: [P1] Narrowed `except ConflictError:` leaks the in-memory run on non-conflict failures — two existing tests now fail

- Location: `manager.py:699-704` (`create_or_reject` new-run persist try/except)
- Problem: The old code used `except Exception` + a `finally` that always popped the new record from `_runs` / the thread index — explicitly covering cancellation ("Also covers cancellation, which bypasses `except Exception`"). The reorder replaced that with `except ConflictError:` only. Now a non-ConflictError failure during `_persist_new_run_to_store` (DB down, non-retryable OperationalError) or a cancellation propagates WITHOUT rolling back, leaving a phantom `pending` run in `_runs` with no DB row. That phantom is returned by `get`/`list_by_thread` and, worse, makes the in-memory `inflight` check reject future `reject` runs on the thread (409) until the process restarts — a transient blip wedges the thread.
- Evidence: reproduced at this head. Both pre-existing tests fail:
  - `test_create_or_reject_does_not_interrupt_old_run_when_new_run_store_write_fails` — `store.put` raises `RuntimeError`; new run not rolled back -> `assert list(manager._runs) == [old.run_id]` fails ("Left contains one more item").
  - `test_create_or_reject_does_not_interrupt_old_run_when_new_run_store_write_is_cancelled` — `store.put` raises `CancelledError`; same phantom-leak failure. This is exactly the cancellation case the deleted `finally` handled.
- Suggested fix: restore a `finally` (or broaden to `except Exception`) that pops/unindexes the new record on any non-success exit, and keep the `ConflictError` branch for the 409 path. Preserve the cancellation guarantee the old comment called out.
- Test: the two tests above already encode the invariant — they should pass again once rollback is restored.

willem-bd (commented):
Thanks @leoredfish. The `0004` migration and the reorder in `create_or_reject` address the earlier comments. A few new issues remain on this head — the most concrete: the narrowed `except ConflictError:` drops the rollback-on-failure / cancellation that the old `finally` provided, and two pre-existing tests now fail at this head (reproduced locally). Details inline.

WillemJiang:
@leoredfish  Please fix the lint errors, unit test errors and address the review comments when you have time.

WillemJiang:
@leoredfish  please resolve the conflicts with the main branch.

backend/packages/harness/deerflow/persistence/migrations/versions/0004_one_active_run_per_thread.py
leoredfish: Good catch — this is a real risk. Addressed in 8eeb855c, following your suggested approach (reconcile before index creation).

**Fix**: `upgrade()` now reconciles before creating the index:

```python
# Clean up any pre-existing duplicate active runs that would prevent
# the unique index from being created.
_cleanup_duplicate_active_runs(bind, dialect_name)
```

`_cleanup_duplicate_active_runs()`:
- Uses `ROW_NUMBER() OVER (PARTITION BY thread_id ORDER BY created_at DESC)` to keep the newest active run per thread
- Marks all others (`rn > 1`) as `interrupted` with `error: "Cleaned up during migration 0004: duplicate active runs per thread"`
- Handles both PostgreSQL (`NOW()`) and SQLite (`datetime('now')`) timestamp expressions
- Logs a warning with the count of cleaned-up runs
- Runs before `op.create_index()`, so the partial unique index creation is guaranteed to succeed

This covers both duplicate sources you mentioned: pre-fix cross-worker races and crashed-worker stuck rows.

**Test (as requested)**: `test_migration_0004_cleans_up_duplicate_active_runs` in `test_persistence_bootstrap.py`:
- Seeds a DB at rev `0003` with two `pending` rows on the same `thread_id` (asserting the index doesn't exist yet as a precondition)
- Runs `alembic upgrade head` through `0004`
- Asserts the older run is reconciled to `interrupted` with the expected error message, and the newer stays `pending`
- Asserts the partial unique index is functional: a third active insert on the same thread raises `IntegrityError`, while a different thread is unaffected

**On reusing `reconcile_orphaned_inflight_runs`**: we opted for a dedicated window-function approach instead — that function's scope is different (crashed-worker stuck rows with stale timestamps vs. deterministic duplicate resolution), and keeping the migration self-contained avoids coupling it to runtime code that may evolve independently.

The new test passes along with the full suite (6636 passed, 0 failures).

backend/packages/harness/deerflow/runtime/runs/manager.py
leoredfish: Fixed in 8eeb855c — following your suggestion: the persist result is now checked, and failure raises an explicit error instead of falling through to a misleading 409.

**Fix** (persist loop in `create_or_reject`):

```python
failed_interruptions: list[str] = []
for r in inflight:
    ok = await self._persist_status(r, RunStatus.interrupted)
    if not ok:
        failed_interruptions.append(r.run_id)

if failed_interruptions:
    logger.error(
        "Failed to persist interrupted status for runs %s on thread %s; "
        "cannot safely create new run because old active runs may still "
        "be visible in the database",
        failed_interruptions,
        thread_id,
    )
    raise RuntimeError(
        f"Failed to interrupt active runs {failed_interruptions} for thread {thread_id}"
    )
```

- `_persist_status`'s return value is no longer discarded; failures are collected and surfaced as a `RuntimeError` before the new-run insert is attempted — the caller sees a real error, never an incidental `ConflictError` from the unique index.
- The policy is now explicit: at the moment of failure the old run is already marked `interrupted` in memory (task cancellation requested, record updated — the caller explicitly requested interruption), while the DB row is untouched. This divergence is logged at error level with the affected run IDs rather than silently swallowed; the worker's cancellation cleanup persists the terminal status once the store recovers.

**Test (as requested)**: `test_create_or_reject_interrupt_raises_when_interrupted_persist_fails` in `backend/tests/test_run_manager.py`, parametrized over both `_persist_status` failure modes:

- `update_raises` — `store.update_status` raises (e.g. transient SQLite lock)
- `row_missing_and_recovery_fails` — `update_status` reports a missing row and the snapshot-recovery `put` also fails

Both cases assert: `RuntimeError` matching `"Failed to interrupt active runs"` is raised (not a 409); no phantom new run is left in `_runs`; and the divergence is explicit — memory shows `interrupted` while the store row remains `running`.

leoredfish:
fix at #4003

PR #4001, opened 2026-07-08
perf(boxlite): add benchmark-driven warm-pool reclaim tuning

Vanzeren:
## Why                                                                                               
                                                                                                       
  #3951  merged the first BoxLite warm-pool lifecycle, which removed the worst-case “destroy and      
 cold-boot every turn” behavior. This follow-up started from measurement rather than intuition: I added 
 a provider-agnostic sandbox benchmark harness and used it to profile the BoxLite warm path directly.   
                                                                                                        
   From the committed benchmark runs on **real BoxLite 0.9.7** (`python:3.12-slim`, `replicas=3`,       
 `idle_timeout=600`), two things stood out:                                                             
                                                                                                        
   - `cold_unique_thread / noop`: `acquire p50 = 855.8ms`, `p95 = 902.5ms`                              
   - `warm_same_thread / noop`: `acquire p50 = 14.6ms`, `total p50 = 29.9ms`                            
   - `warm_same_thread / python_small`: `acquire p50 = 14.2ms`, `total p50 = 38.1ms`                    
                                                                                                        
   So once BoxLite can reclaim a warm VM, the hot path is already about **60x faster** than a cold      
 acquire on the noop workload.                                                                          
                                                                                                        
   Those same measurements also showed the next bottleneck clearly: after cold boot cost is removed,    
 the reclaim-time health check becomes a visible part of the warm-path latency budget. In practice, a   
 healthy warm reclaim was still paying about **~14ms** on the noop path, mostly from the `echo ok`      
 validation round-trip rather than VM startup itself.                                                   
                                                                                                        
   That benchmark result is what motivated the tuning in this PR. The idea here is not “blindly skip    
 safety checks”; it is:                                                                                 
                                                                                                        
   - keep reclaim validation **on by default** for correctness;                                         
   - make dead/stale BoxLite VMs get invalidated instead of being reused; and                           
   - expose an **opt-in** low-latency path for boxes that were just released by this provider instance, 
 where the extra validation round-trip is often pure overhead.                                          
                                                                                                        
   The local tuning run captured in this branch showed the effect of that opt-in path on                
 `warm_same_thread / noop / 20 iterations`:                                                             
                                                                                                        
   - `acquire p50: 14.9ms -> 0.0ms`                                                                     
   - `total p50:   29.9ms -> 14.0ms`                                                                    
                                                                                                        
   So this PR is benchmark-driven in two directions:                                                    
   1. **correctness** — invalidate dead warm boxes and validate reclaim by default;                     
   2. **performance** — add a measured, opt-in skip window for very recent releases.                    
                                                                                                        
   ### Risk                                                                                             
                                                                                                        
   This PR touches sandbox lifecycle and warm-pool state transitions, so it is not risk-free.           
                                                                                                        
   Main risks:                                                                                          
   - **false-positive invalidation**: if a recoverable execution error is misclassified as terminal, we 
 may throw away a healthy warm VM and pay an unnecessary cold boot later;                               
   - **over-aggressive skip behavior**: if reclaim validation is skipped too broadly, a dead VM could   
 be handed back as if it were healthy;                                                                  
   - **extra maintenance surface**: the benchmark harness and accounting add code, tests, and reviewer  
 surface area.                                                                                          
                                                                                                        
   To bound that risk, the runtime stays conservative by default:                                       
   - reclaim validation remains **enabled by default**;                                                 
   - the skip window is **opt-in** via `sandbox.health_check_skip_seconds`;                             
   - the skip path only applies to boxes that were **just released by this provider instance**, not     
 startup-adopted/orphaned boxes.                                                                             
                                                                                                       
  ## What changed                                                                                      
                                                                                                       
  - BoxLite now invalidates cached sandboxes when command execution fails with terminal                
transport/runtime errors, instead of leaving dead VMs in active or warm-pool state.                    
  - Warm-pool reclaims are validated before reuse by default, so stale or already-closed BoxLite VMs   
are evicted and recreated instead of being handed back to the caller.                                  
  - Added an optional `sandbox.health_check_skip_seconds` setting for a low-latency reclaim path. When 
enabled, BoxLite may skip the reclaim health check for VMs that were just released by this provider    
instance; the default is `0.0`, which keeps reliability-first validation.                              
  - Startup-adopted/orphaned BoxLite VMs still require validation before reuse; the skip window only   
applies to boxes this provider actually released into its own warm pool.                               
  - Added a provider-agnostic sandbox benchmark CLI and summary tooling under                          
`backend/scripts/benchmark/` so BoxLite and other providers can be compared with the same              
acquire/run/release measurements and JSONL output.                                                     
  - Added focused regression coverage for warm-pool invalidation, reclaim validation, stale adapter    
handling, and benchmark accounting behavior.                                                           
  - Updated the BoxLite README/config example to document the new reclaim-tuning option and its        
reliability/performance tradeoff.                                                                      
                                                                                                       
  ## Surface area                                                                                      
                                                                                                       
(say what it buys us)                                                                                  
                                                                                                       
  ## Screenshots / Recording                                                                           
                                                                                                       
  N/A — backend sandbox/runtime change only.                                                           
                                                                                                       
  ## Bug fix verification                                                                             
                                                                                                       
  ## Validation                                                                                        
 **Test environment**                                                                                 
                                                                                          
   - Focused regression/lint verification for this PR was run on:                                       
     - OS: Linux x64                                                                                    
     - CPU: Intel Core Ultra 5 225H                                                                     
   - Benchmark environment captured in this branch:                                                     
     - runtime: BoxLite `0.9.7`                                                                         
     - sandbox image: `python:3.12-slim`                                                                
     - benchmark config: `replicas=3`, `idle_timeout=600`
                                                                               
  Ran:                                                                                                 
                                                                                                       
  ```bash                                                                                              
  cd backend                                                                                           
  uv run ruff check \                                                                                  
    packages/harness/deerflow/community/boxlite/box.py \                                               
    packages/harness/deerflow/community/boxlite/provider.py \                                          
    scripts/benchmark/bench_sandbox_provider.py \                                                      
    tests/test_boxlite_provider.py \                                                                   
    tests/test_bench_sandbox_provider.py                                                               
                                                                                                       
  uv run pytest \                                                                                      
    tests/test_boxlite_provider.py \                                                                   
    tests/test_bench_sandbox_provider.py    
```                                                                                                    
                                                                                                        
 Results:                                                                                               
                                                                                                        
 - ruff check on touched BoxLite / benchmark files → passed                                             
 - pytest tests/test_boxlite_provider.py tests/test_bench_sandbox_provider.py → 39 passed               
                                                                                                        
 ## AI assistance                                                                                          
                                                                                                        
 Tool(s) used: Codex / Oh My Pi coding harness                                                          
                                                                                                        
 How you used it: AI helped inspect the BoxLite warm-pool follow-up branch, review the touched runtime  
 and benchmark changes, extract the committed benchmark data, summarize reviewer-facing                 
 risks/tradeoffs, and draft this PR description from the verified branch state and test results.        
                                                                                                        
 unreviewed AI output.

### CODE REVIEW

**Review by 18062706139fcz — COMMENTED** [review]
Thanks for the benchmark-driven BoxLite follow-up. The runtime direction looks good overall: default reclaim validation stays conservative, terminal transport failures are invalidated, and the stale-adapter tests cover the important lifecycle edges. I found a few follow-up items worth addressing before merge:

File: backend/scripts/benchmark/bench_sandbox_provider.py [hunk 1]
```diff
@@ -0,0 +1,888 @@
...
+
+
+def main(argv: list[str] | None = None) -> int:
+    args = _parse_args(argv)
+    if args.workload == "state_reuse" and (args.scenario != "warm_same_thread" or args.concurrency != 1):
+        raise SystemExit("state_reuse requires --scenario warm_same_thread --concurrency 1")
+
+    output_path = Path(args.output)
+
+    config: dict[str, Any] = {
```
18062706139fcz, line 659: This benchmark config does not expose or forward `health_check_skip_seconds`, so the CLI cannot reproduce the opt-in low-latency reclaim path introduced by this PR. With the provider default now `0.0`, runs through this script only cover validated reclaim by default, and the JSONL output cannot distinguish validated vs skipped reclaim runs.

Can we add a `--health-check-skip-seconds` flag, pass it through to `sandbox_attrs`, and include the value in `config_used` / output metadata?

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -196,13 +197,15 @@ def _opt(name: str, default: Any = None) -> Any:
...
         idle_timeout = _opt("idle_timeout")
+        health_check_skip_seconds = _opt("health_check_skip_seconds")
         return {
             "image": _opt("image") or DEFAULT_IMAGE,
             "memory_mib": _opt("memory_mib"),
             "cpus": _opt("cpus"),
             "environment": dict(_opt("environment") or {}),
             "replicas": replicas if replicas is not None else self.DEFAULT_REPLICAS,
             "idle_timeout": idle_timeout if idle_timeout is not None else self.DEFAULT_IDLE_TIMEOUT,
+            "health_check_skip_seconds": float(health_check_skip_seconds if health_check_skip_seconds is not None else 0.0),
```
18062706139fcz, line 208: Since `health_check_skip_seconds` is now a user-facing sandbox option, it would be good to expose it through the normal config surface as well. Right now it is accepted via `extra="allow"`, but `SandboxConfig` and the BoxLite section in `config.example.yaml` do not document or type it, so `make config-upgrade` will not surface the option and invalid values are only caught here by `float(...)`.

Can we add a typed `health_check_skip_seconds: float | None` field to `SandboxConfig`, document it in `config.example.yaml`, and bump `config_version` if needed?

backend/scripts/benchmark/bench_sandbox_provider.py
```diff
@@ -0,0 +1,888 @@
...
+        box = self._loop.run(_make())
+        return BoxliteBox(
+            sandbox_id,
+            box,
+            self._loop.run,
+            default_env=self._config["environment"],
+            on_terminal_failure=self._invalidate_box,
+        )
+
+    BoxliteProvider._create_box = _patched_create_box
```
18062706139fcz: This class-level patch is never restored, and the benchmark CLI also imports `pytest.MonkeyPatch` outside the test suite. It works for a one-shot script, but it makes the committed benchmark tool depend on dev-only pytest and can leak patched `BoxliteProvider` behavior if `main()` is called from tests or another long-lived Python process.

Could we replace this with a small local context manager / explicit try-finally restore, or avoid class-level patching by moving the BoxLite 0.9.7 shim workaround behind an injectable factory?

Vanzeren:
@18062706139fcz thanks for reviews. I've added ```--health-check-skip-seconds``` to ```bench_sandbox_provider.py``` exposed ```health_check_skip_seconds``` as a typed ```SandboxConfig``` field and also removed the benchmark script’s ```pytest.MonkeyPatch``` dependency and class-scope _create_box patching.

Vanzeren:
@18062706139fcz thanks for reviews. I've added ```--health-check-skip-seconds``` to ```bench_sandbox_provider.py``` exposed ```health_check_skip_seconds``` as a typed ```SandboxConfig``` field and also removed the benchmark script’s ```pytest.MonkeyPatch``` dependency and class-scope _create_box patching.

backend/packages/harness/deerflow/community/boxlite/box.py
```diff
@@ -64,25 +64,65 @@ def __init__(
...
 
-    # ── bridge helpers ──────────────────────────────────────────────────
+    @staticmethod
+    def _is_terminal_box_failure(error: Exception) -> bool:
+        if isinstance(error, (BrokenPipeError, ConnectionError, EOFError)):
+            return True
+        if not isinstance(error, RuntimeError | OSError):
+            return False
+        msg = str(error).lower()
+        markers = (
```
willem-bd: **Medium: terminal-failure classification relies on substring matching against hardcoded error messages.**

The markers are broad (`"transport"`, `"disconnected"`, …) and the consequence — invalidate → cold boot — is paid by the warm path this PR is optimizing. If BoxLite (or an intermediate lib) ever raises a *recoverable* `RuntimeError`/`OSError` whose message contains one of these (e.g. a transient "transport not ready" that would succeed on retry), a healthy VM gets thrown away — eroding the 60× warm-path win.

Suggestions:
- Prefer matching on exception **type/cause** where BoxLite exposes a typed error, and keep the marker list as a fallback.
- Add a regression test asserting a *non-terminal* transport-flavored message (e.g. a busy/retryable error) is **not** invalidated — the current tests only cover the clearly-terminal and clearly-benign ends.
- Make the marker list overridable (config or class attr) so a future BoxLite release's wording can't silently break classification.

backend/scripts/benchmark/bench_sandbox_provider.py
```diff
@@ -0,0 +1,908 @@
...
+    if "memory_mib" in config:
+        sandbox_attrs["memory_mib"] = config["memory_mib"]
+    if "cpus" in config:
+        sandbox_attrs["cpus"] = config["cpus"]
+    if "environment" in config:
+        sandbox_attrs["environment"] = config["environment"]
+
+    # -- BoxLite 0.9.7 shim permission workaround --
+
+    def _patched_create_box(self: Any, sandbox_id: str) -> Any:
```
willem-bd: **Medium: `_patched_create_box` reaches into BoxLite 0.9.7 private internals.**

`box._box`, `box._box.id`, `box._box.__aenter__()`, and `box._started` (lines ~196-208) are implementation details. Two concerns:
- It's pinned to a specific internal layout and will break silently on the next BoxLite release.
- It means the benchmark exercises a *different* box-create path than the production `_create_box`, so create-path numbers aren't fully representative.

The shim-+x workaround for "0.9.7 extracts the shim without the execute bit" reads like a real BoxLite bug — if so, it belongs upstream (or in the provider, gated by a version check) rather than forked into the benchmark. At minimum, add a version guard that loud-fails on non-0.9.7 runtimes instead of mis-chmodding.

backend/packages/harness/deerflow/community/boxlite/box.py
```diff
@@ -132,12 +177,8 @@ def execute_command(
         """
         _validate_extra_env(env)  # POSIX env-var key rule; raises ValueError on a bad key
         merged_env = {**self._default_env, **(env or {})} or None
-        with self._lock:
-            if self._closed:
-                return "Error: sandbox has been closed"
-            box = self._box
         try:
-            result = self._run(box.exec("sh", "-lc", command, env=merged_env, timeout=timeout), timeout=timeout)
+            result = self._exec("sh", "-lc", command, env=merged_env, timeout=timeout)
```
willem-bd, line 191: **Low: delegating to `_exec` makes closed-box access hit `logger.error`.**

Before this change the closed-box check returned `"Error: sandbox has been closed"` *before* the try block, so it never logged. Now a closed box flows through `_exec` (which raises) → this `except` → `logger.error("Failed to execute command…")`. A caller holding a stale reference (or retrying) will spam error logs for a benign "already closed" condition. Consider an `is_closed` short-circuit before delegating, or downgrade this case to `debug`.

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -419,12 +477,14 @@ def _reclaim_warm_pool(self, sandbox_id: str) -> str | None:
                 logger.warning("Warm pool box %s health check failed: %s", sandbox_id, result)
                 with self._lock:
                     self._warm_pool.pop(sandbox_id, None)
+                    self._skip_health_check_warm_ids.discard(sandbox_id)
```
willem-bd: **Low: reclaim health-check-failure path duplicates `_destroy_warm_entry` logic inline.**

This `pop` + `discard` + `close` sequence (and the matching one in the `except` branch below) is exactly what `_destroy_warm_entry` already encapsulates — and it's now a second place that must remember to `discard` from `_skip_health_check_warm_ids` on close. That kind of duplication is what caused the lifecycle bugs this PR is fixing. Recommend `self._destroy_warm_entry(sandbox_id, box, reason="health_check_failed")`.

WillemJiang:
@Vanzeren  please update the deploy/helm/deer-flow/values.yaml version with the latest config version to fix the Helm charter error.

Vanzeren:
cc @WillemJiang

WillemJiang (approved):

WillemJiang (approved):

WillemJiang:
@Vanzeren  could you drop me an email (willem.jiang AT Gmail), so we can have more discussion about how to enhance the DeerFlow test harness.

Vanzeren:
@WillemJiang Sure, I am glad to discuss about it. But I may send you an email on weekends.

PR #3960, opened 2026-07-06
feat(gateway): refuse multi-worker startup on non-Postgres backends

heart-scalpel:
Fixes #3948

## Why

Issue #3948 lays out four correctness breakers that ship today when an
operator bumps `GATEWAY_WORKERS` past 1. This PR is work item 1 of that
plan — the foundation gate that unblocks the rest.

The acute pain: with `database.backend=sqlite` and
`GATEWAY_WORKERS > 1`, multiple Uvicorn workers share a single SQLite
file with no startup guard. Production hits intermittent
`database is locked`, the runs table drifts between workers, and
there is **no signal at boot** that the configuration is unsupported.
Operators discover it from random failures in the middle of a deploy.

The fix is intentionally narrow: refuse to start. Operators get a
clear, actionable error before any traffic flows; everyone else gets
to ship the deeper fixes (cross-process run ownership, lease
heartbeats, cancel routing) on top of a stable foundation.

## What changed

- **No behavior change for `GATEWAY_WORKERS=1`** (the documented default
  in `docker/docker-compose.yaml` and `tests/test_compose_default_workers.py`).
  Single-worker Gateway runs are byte-identical to before.
- **`GATEWAY_WORKERS > 1` + non-Postgres backend** now fails fast at
  lifespan startup. The process exits with a message naming both
  remediation paths:
  > `GATEWAY_WORKERS=2 requires database.backend='postgres', but database.backend is 'sqlite'. SQLite cannot support concurrent multi-process access. Set GATEWAY_WORKERS=1 or switch to Postgres.`
- The gate runs inside `langgraph_runtime()` **before**
  `init_engine_from_config()`, `make_stream_bridge()`, and every other
  expensive startup step, so a misconfigured deploy never opens a
  listener or writes to disk.
- Malicious / typo'd `GATEWAY_WORKERS` values (empty string, `"auto"`,
  floats) fall back to `1` instead of crashing startup — uvicorn still
  gets to do its own validation, the gate just stays inert.

Redis is **not** enforced. `stream_bridge.type=redis` remains the
operator's call: it affects SSE delivery across workers (already
shipped in #3191), not run-data correctness.

## Surface area

- [x] **Backend API** — startup-time gate inside
  `app/gateway/deps.py::langgraph_runtime`; no endpoint / SSE / request
  shape changes.

## Validation

```
cd backend && make lint
cd backend && PYTHONPATH=. uv run pytest tests/test_multi_worker_postgres_gate.py -v
cd backend && make test
```

[...]

PR #4003, opened 2026-07-08
fix(runs): close multi-worker ownership gaps in run atomicity (#3948)

### LINKED ISSUE #3948: [feat] 多 Worker 数据正确性基线:Postgres 强制 + Run 跨进程原子化
### Before you start

### Problem / motivation

PR #3191 落地了 Redis StreamBridge,SSE 跨进程投递和重放已解决。但 `docker/docker-compose.yaml:92-97` 仓库自身的边界声明仍是:

> The Redis stream bridge solves cross-worker SSE delivery and reconnect, but **run cancel, request dedup, and per-worker IM channel services remain worker-local.**

`tests/test_compose_default_workers.py` 也仍钉死默认 `GATEWAY_WORKERS=1`。本 feat 只处理"开多 worker 会破坏 run 数据正确性"的 P0 问题——sticky routing、IM leader election、subagent 全局信号量、memory pub/sub 等平行工作留给后续 feat,以便本 feat 能独立 review、独立 merge、独立灰度。

四个 P0 破坏面,按依赖顺序:

### 1. SQLite + 多 worker 静默挂掉

`database.backend=sqlite` + `GATEWAY_WORKERS>1` 时多进程写同一文件偶发 `database is locked`,无启动拒绝。全仓 grep 无启动校验。生产环境随机挂,无任何提示。**阻塞所有后续多 worker 工作。**

### 2. `create_or_reject` 跨进程竞态 → 同 thread 两个 run

`runtime/runs/manager.py:631` 的 `asyncio.Lock` 只在单进程有效。两个 worker 同时收到 thread T 的 run 请求,各自查 `_runs` dict 都没 inflight,各自创建 → checkpoint 互相覆盖。

触发场景:用户连发两条消息、多端登录、IM 渠道连续推送、前端网络重试、scheduled task 和用户操作重叠、subagent task() 期间用户再发、rolling update 期间 sticky miss。

`multitask_strategy` 参数(reject / interrupt / rollback)的存在就是仓库已经承认并发是真实场景——单 worker 下这个参数正常工作,多 worker 下完全失效。**checkpoint 数据破坏不可恢复。**

### 3. Postgres 路径无 reconciliation,Pod 崩溃后 run 永远 running

`deps.py:252` 把 `reconcile_orphaned_inflight_runs` 门控在 `backend == "sqlite"`,Postgres 完全不跑。Pod 崩溃 / OOM / rolling update / 节点驱逐后,该 worker 持有的 run 行永远停在 `pending` / `running`,UI 永远转圈,只能 DBA 手工 SQL 清。K8s 部署下 pod 死亡是常态。

**不能直接去掉 sqlite 门控**:当前逻辑是"本进程无 task 即孤儿",搬到 Postgres 会让 Worker B 启动时误杀 Worker A 的活跃 run。必须先把 lease 列加上,改造判定逻辑,再开启 Postgres 路径。**K8s 部署下高频。**

### 4. Cancel 落到非 owner pod → 409

`cancel()` 依赖 `record.task`(asyncio.Task)在内存。多 worker 下 POST `/runs/{run_id}/cancel` 落在非 owner pod,该 pod 的 RunRecord 是从 store `store_only=True` hydrate 的,无 task/abort_event → 返回 409。AGENTS.md 已声明此限制。无 sticky routing 时任意 pod 都可能收到 cancel,大概率失败,**取消按钮失效**。

### 依赖关系

四个破坏面互相耦合:破坏面 1 是基础设施,破坏面 2 是数据正确性根因,破坏面 3+4 是破坏面 2 落地后的延伸修复——lease 列加完不发心跳就检测不到崩溃,加了心跳不修 reconciliation 还会卡死。

### 工作项 1 — Postgres 强制校验

启动期校验:`GATEWAY_WORKERS > 1` 且 `database.backend != "postgres"` → 启动失败 exit code 1。位置在 `langgraph_runtime()` 调 `init_engine_from_config()` 之前。`GATEWAY_WORKERS=1` 行为不变。

不强制 Redis——`stream_bridge.type=redis` 仍由用户自己配,Redis 缺失只影响 SSE 多 worker,不影响 run 数据正确性。

### 工作项 2 — Run 跨进程原子化(核心)

新增 alembic revision `0004_run_ownership`:

- `runs.owner_worker_id VARCHAR(24)` NULL
- `runs.lease_expires_at TIMESTAMPTZ` NULL
- `idx_runs_lease (lease_expires_at)`
- 部分唯一索引 `uq_runs_thread_active ON runs(thread_id) WHERE status IN ('pending', 'running')`

改造 `create_or_reject`:

- `reject` 策略:直接 INSERT,捕获 `UniqueViolation` → `ConflictError`
- `interrupt` / `rollback`:`SELECT FOR UPDATE` 锁冲突行 → 检查 lease 存活性 → cancel 或标 error → retry INSERT

实现可参照 `persistence/scheduled_tasks/sql.py:111-156` `claim_due_tasks` 的 `FOR UPDATE SKIP LOCKED` + 双条件模式(已在仓库内被 PR #3898 验证过)。

### 工作项 3 — Lease 心跳 + Postgres reconciliation

心跳任务:`heartbeat_enabled=true` 时,每个 worker 启动后台 task,周期 = `lease_seconds / 3`,对本 worker 持有的 active run 续约 `lease_expires_at`。

改造 `reconcile_orphaned_inflight_runs`:只把 `lease_expires_at IS NULL OR lease_expires_at < NOW() - INTERVAL 'grace_seconds'` 的 run 标 error(`IS NULL` 兼容旧数据)。

去掉 `deps.py:252` sqlite-only 门控让 Postgres 也跑——**和上面的逻辑改造必须同 commit**,否则会临时引入误杀。

启动清理 stale run 参照 `app/scheduler/service.py:308` `mark_stale_active_runs`。

```yaml
run_ownership:
  lease_seconds: 30
  grace_seconds: 10
  heartbeat_enabled: false  # GATEWAY_WORKERS=1 时不开
```

### 工作项 4 — Cancel 降级到 lease 接管

`cancel()` 收到请求:

- 查 store 获取 `owner_worker_id` 和 `lease_expires_at`
- `owner == self` → 本地 cancel(现有路径)
- `owner != self` + lease 过期 → 本 worker 标 error(接管)
- `owner != self` + lease 未过期 → 409 + `Retry-After` 头

不实现 HTTP 内部 endpoint 跨进程转发(引入 RPC 故障模式,后续 sticky routing 命中率 99%+ 后极端 miss 走 lease 接管即可)。

### 功能开关

`GATEWAY_WORKERS=1` 时所有功能开关默认行为 = 当前代码,**零回归**。

## 范围外

| 项 | 推迟原因 |
|---|---|
| nginx sticky + 前端 `X-Thread-ID` | UX 优化,不影响数据正确性 |
| `worker_id` 注入日志 + `X-Worker-ID` 响应头 | 可观测性,留下个 feat |
| IM Channel leader election | IM 重复回复是运营问题,不是 run 数据正确性 |
| Subagent 全局信号量 | 资源 / rate limit 问题 |
| Memory cache pub/sub | UX 一致性问题 |

拆分原则:本 feat 落地后,运维可以**安全开启 `GATEWAY_WORKERS=2` 跑 run 路径**,IM 渠道 / 全局并发等仍待后续 feat。用户在多 worker 初期可临时 `channels.enabled: false` 规避 IM 重复消费。

### Affected area(s)

### Alternatives considered

**Redis 分布式锁替代 DB 唯一索引**:放弃。锁超时 / 网络分区 / GC 停顿都可能失效,Redis 故障时无法回退。DB 唯一索引是 ground truth,代码 bug 也不会破坏一致性。Redis 在本 feat 中不参与数据正确性,只参与 SSE 投递(PR #3191 已落地)。

**把 sticky routing 一起做**:放弃。sticky routing 是 UX 优化,不是正确性。sticky miss 在 K8s rolling update 时是必然事件,破坏面 2 必须用 DB 约束解决,不能依赖 sticky。把 sticky 拆出本 feat 也可以让前端 `X-Thread-ID` 改造有独立的 review 周期。

## 测试要求

- 并发 100 个 `create_or_reject`,reject 策略仅 1 个成功
- interrupt 策略下旧 run 被 cancel 后 retry 成功
- Worker B 的 reconciliation 跳过 Worker A 持有的未过期 lease run
- Worker A 崩溃后,B 在 `lease_seconds + grace_seconds` 内清理孤儿
- Cancel 非 owner pod:lease 未过期 → 409 + `Retry-After`;lease 过期 → 接管成功
- `GATEWAY_WORKERS=1` + `heartbeat_enabled=false` 行为不变
- `GATEWAY_WORKERS=2` + SQLite → 启动失败 exit code 1

## 灰度

| Stage | 内容 | Workers |
|---|---|---|
| Stage 0 | 工作项 1 落地 | 1 |
| Stage 1 | 工作项 2/3/4 落地,`heartbeat_enabled=true` | 2 |
| Stage 2 | 7 天观察 + 告警校准 | 2 |
| Stage 3 | 开放生产 `GATEWAY_WORKERS=4` | 4 |

回滚:`GATEWAY_WORKERS=1` 立即生效,无需代码回滚。

## 现状代码引用

**已有内部范例(可参考)**:

- `persistence/migrations/versions/0003_scheduled_tasks.py:48-49` — `lease_owner` / `lease_expires_at` 同名列(在 scheduled_tasks 表)
- `persistence/scheduled_tasks/sql.py:111-156` — `claim_due_tasks`,`FOR UPDATE SKIP LOCKED` + 双条件
- `app/scheduler/service.py:35` — per-process `lease_owner = f"{hostname}:{uuid}"` 标识
- `app/scheduler/service.py:308` — `mark_stale_active_runs` 启动清理

**待改造**:

- `runtime/runs/manager.py:604-694` — `create_or_reject`
- `runtime/runs/manager.py:696-` — `reconcile_orphaned_inflight_runs`
- `runtime/runs/manager.py` `cancel()` — 加 owner 判定 + lease 接管分支
- `app/gateway/deps.py:252` — 去掉 sqlite-only 门控
- `persistence/run/sql.py` — 新增 `update_lease` / 改造 `list_inflight`
- `config/app_config.py` — 新增 `RunOwnershipConfig`
- `config/reload_boundary.py` — 注册 `run_ownership.heartbeat_enabled` 为 startup-only

## 相关

- PR #3191 — Redis StreamBridge(已合并,本 feat 不重复其工作)
- PR #3898 — Scheduled tasks MVP(已合并,本 feat 复用其 lease 模式)
- `docker/docker-compose.yaml:92-97` — 当前仓库的多 worker 边界声明,本 feat 落地后需要更新

heart-scalpel:
Fixes #3948

## Why

PR #3191 landed Redis StreamBridge, which solved cross-worker SSE delivery. But the repository's own boundary declaration (`docker/docker-compose.yaml`) still says:

> The Redis stream bridge solves cross-worker SSE delivery and reconnect, but **run cancel, request dedup, and per-worker IM channel services remain worker-local.**

This PR closes the run-data-correctness slice of that gap: with these changes, operators can safely enable `GATEWAY_WORKERS=2` for the run path. Work item 1 (Postgres startup gate, #3960) already landed; this PR implements **work items 2 and 3** from `docs/multi_worker.md`. Work item 4 (cancel lease takeover) and the parallel surfaces (IM leader election, subagent global semaphore, memory pub/sub) stay out of scope so this can be reviewed and rolled out independently.

The four P0 breakers the plan addresses (in dependency order):

1. SQLite + multi-worker silent lock failure — already blocked by #3960.
2. `create_or_reject` cross-process race → two runs for the same thread, checkpoint corruption. **Solved by this PR (work item 2).**
3. Postgres deployments with no reconciliation → pod crash leaves runs `pending`/`running` forever, UI spins, DBA-only SQL cleanup. **Solved by this PR (work item 3).**
4. Cancel landing on a non-owner pod → 409, cancel button broken. Out of scope (work item 4). 

### Work item 2 — cross-process atomic `create_or_reject`

- `RunRow` gains `owner_worker_id` and `lease_expires_at` columns plus two indexes: `ix_runs_lease` for reconciliation queries, and `uq_runs_thread_active` — a partial unique index on `(thread_id) WHERE status IN ('pending', 'running')`. The partial unique index **is** the cross-process atomicity guarantee: at most one active run per thread, enforced by the database, not by an in-process lock.
- The index lives in `RunRow.__table_args__` with both `sqlite_where` and `postgresql_where` (mirroring the existing `uq_channel_connection_active_identity` pattern). The empty-DB bootstrap path runs `Base.metadata.create_all` + `alembic stamp head` **without executing any revision's `upgrade()`**, so the index must come from ORM metadata to land on fresh deployments. Migration `0004_run_ownership` also creates it idempotently for legacy / versioned upgrades using the same dialect-aware `create_index` call.
- `RunRepository.create_run_atomic` is the new atomic primitive:
  - **reject**: INSERT directly. The partial unique index catches duplicate active runs; the manager surfaces the result as `ConflictError`.
  - **interrupt/rollback**: `SELECT FOR UPDATE` the conflicting rows, skip rows whose lease is still valid **and** owned by another live worker (raise `ConflictError` — the INSERT would have failed on the index anyway, and a retry loop cannot make progress), cancel the rest in the same transaction, then INSERT the new row. Rows owned by *this* worker are interruptible regardless of lease state, so a worker can replace its own active run.
- `RunManager.create_or_reject` dispatches to the store under the existing local lock; same-worker in-memory cancellation runs after the store commit succeeds. `MemoryRunStore` mirrors the same semantics for tests and `database.backend=memory`.

### Work item 3 — lease heartbeat + Postgres reconciliation

- New `RunOwnershipConfig` (`lease_seconds=30`, `grace_seconds=10`, `heartbeat_enabled=false` by default). Registered as a startup-only field in `reload_boundary.STARTUP_ONLY_FIELDS` because the heartbeat background task is created once in `langgraph_runtime()` and is not rebuilt on `config.yaml` edits; `AppConfig.run_ownership` Field description carries the `startup-only:` marker so the restart requirement surfaces in IDE hover.
- When `heartbeat_enabled`, each worker starts a background task that renews leases on its own active runs with `interval = lease_seconds / 3`. The loop is bounded and stop-event-cancellable so shutdown is prompt.
- `reconcile_orphaned_inflight_runs` now runs on **every** backend — the sqlite-only gate in `app/gateway/deps.py` is dropped in the same commit so there is no window where Postgres would mis-mark live Worker A runs as orphans. Reconciliation errors only runs whose lease is `NULL` (legacy pre-ownership rows) or older than `grace_seconds`. In single-worker mode (heartbeat off, NULL leases) all inflight rows reclaim immediately, preserving the pre-ownership recovery latency.
- Heartbeat starts **after** startup reconciliation and stops **before** the in-flight run drain on shutdown so the two cannot race (a heartbeat tick mid-drain would otherwise renew a lease for a run whose task is about to be cancelled).

### About work item 4
This closes items 2–3, but cancel (item 4) is still open — cancel landing on a non-owner pod still 409s. Worth stating so "close ownership gaps" in the title isn't read as all-closed.

### Backward compatibility

`GATEWAY_WORKERS=1` with `heartbeat_enabled=false` (the default) keeps current behavior — runs still get `owner_worker_id` populated, but `lease_expires_at` stays `NULL`, heartbeat does not start, and reconciliation reclaims every inflight row at startup exactly as before.

## Surface area

The runtime behavior changes only fire under `GATEWAY_WORKERS > 1` + `heartbeat_enabled=true`, which is not the default. No user-facing API changes.

## Validation

Backend (from `backend/`):

```
make lint           # ruff check — clean
make format         # ruff format — clean

PYTHONPATH=. uv run pytest \
  tests/test_multi_worker_run_ownership.py \
  tests/test_persistence_bootstrap.py \
  tests/test_persistence_bootstrap_concurrency.py \
  tests/test_run_manager.py \
  tests/test_run_repository.py \
  tests/test_gateway_run_recovery.py \
  tests/test_reload_boundary.py \
  tests/test_owner_isolation.py
# 170 passed

PYTHONPATH=. uv run pytest tests/ --ignore=tests/test_client_live.py
# 6700 passed, 25 skipped
# (the 5 test_client_live cases need Docker sandbox + Langfuse)
```

Test coverage in `tests/test_multi_worker_run_ownership.py`:

- `reject` raises `ConflictError` when thread has an active run (local and store-enforced).
- `interrupt` cancels old runs and creates a new one.
- `create_run_atomic` raises `ConflictError` when interrupt encounters a valid lease owned by another worker; allows interrupt when the existing run is owned by the same worker.
- Reconciliation claims expired-lease and NULL-lease runs; skips valid-lease runs; skips locally-active runs.
- Heartbeat renews leases on this worker's active runs; skips runs owned by other workers; is a no-op when disabled.
- `GATEWAY_WORKERS=1` + `heartbeat_enabled=false` behavior unchanged.
- `test_empty_branch_creates_all_and_stamps_head` now also asserts `uq_runs_thread_active` and `ix_runs_lease` exist on a fresh DB.

Operator validation still pending (cannot run from this environment):

- Real Postgres, `GATEWAY_WORKERS=2`, `heartbeat_enabled=true`:
  - `create_or_reject("reject")` racing on the same thread across two workers yields exactly one success and one `ConflictError`.
  - `create_or_reject("interrupt")` against a thread whose active run is owned by the same worker succeeds and cancels the old run.
  - `create_or_reject("interrupt")` against a thread whose active run is owned by another *live* worker returns `ConflictError`.
  - After Worker A's lease expires, Worker B's reconciliation reclaims the run as `error` within `lease_seconds + grace_seconds`.
  - `\di runs` in psql shows `uq_runs_thread_active` on a freshly-bootstrapped DB.

## E2E Validation

E2E tested against real Postgres with `GATEWAY_WORKERS=2`, `heartbeat_enabled=true`, `lease_seconds=30`, `grace_seconds=10`.

  ### 1. Concurrent reject — cross-process atomicity

  Two requests hitting the same thread simultaneously. Exactly one succeeds, the other gets 409.

  Request 1 → 200 (stream started, run created)
  Request 2 → 409 "Thread xxx already has an active run"

  The partial unique index `uq_runs_thread_active` enforces the invariant at the DB level.

  ### 2. Lease heartbeat

  Two distinct worker IDs started heartbeat tasks, each independently renewing leases every `lease_seconds/3`.

  heartbeat started for worker b420f5a37bc4:b0b9830a...
  heartbeat started for worker b420f5a37bc4:b99e2ad9...

  ### 3. Reconciliation — orphan recovery after crash

  Inserted three rows with different lease states directly into Postgres, then restarted the gateway to trigger `reconcile_orphaned_inflight_runs`.

  | Run | Lease | Expected | Actual |
  |-----|-------|----------|--------|
  | `orphan-*` | Expired 60s ago | `error` | `error` |
  | `valid-*` | Valid (5min remaining) | `running` | `running` |
  | `null-*` | NULL (legacy) | `error` | `error` |

  Log: `Recovered 2 orphaned inflight run(s) as error` — expired-lease and NULL-lease runs reclaimed; valid-lease run correctly skipped.
[multi_worker_test_results.md](https://github.com/user-attachments/files/29892708/multi_worker_test_results.md)

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Implementation + test authoring. Claude read `docs/multi_worker.md`, implemented work items 2 and 3 (migration, store changes, manager changes, config, lifespan wiring), wrote the test suite in `tests/test_multi_worker_run_ownership.py`, and updated existing repository/bootstrap tests for the new partial unique index invariant. A self-review pass before opening the PR caught four issues (partial unique index missing from ORM metadata, interrupt vs. valid lease edge case, missing `reload_boundary` registration, misleading test docstring); those were folded back into the same implementation rather than landing as follow-up patches. Commit message and PR description also drafted by Claude.


backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -31,6 +34,37 @@
...
+    """
+    pending: list[BaseException] = [exc]
+    seen: set[int] = set()
+    while pending:
+        current = pending.pop()
+        if id(current) in seen:
+            continue
+        seen.add(id(current))
+        message = str(current).lower()
+        if "unique" in message and "violat" in message:
```
willem-bd: **Substring heuristic doesn't match SQLite's `IntegrityError` format.**

`sqlite3` raises `UNIQUE constraint failed: runs.uq_runs_thread_active` — lowercased it contains `"unique"` but **not** `"violat"` (SQLite says "failed", not "violates") and not `"duplicate"`. So on SQLite this returns `False`, and the `reject` path re-raises a raw `IntegrityError` (HTTP 500) instead of `ConflictError` (409). Postgres's `duplicate key value violates unique constraint` matches fine.

Low real-world impact (SQLite isn't multi-worker, and the local-lock check catches same-worker dups before the store is hit), but the index *is* created on SQLite by `create_all`, and this conversion is the load-bearing piece of the "store is source of truth" design — yet it's untested (every atomic test uses `MemoryRunStore`, which raises `ConflictError` directly and never reaches this branch).

Suggestion: detect by dialect/exception type rather than message — match `orig` against `sqlite3.IntegrityError` / `psycopg.errors.UniqueViolation` (pgcode 23505). At minimum add `"constraint failed"` to the SQLite branch, and add a SQLite-backed `RunRepository.create_run_atomic` reject-conflict test.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -619,63 +683,129 @@ async def create_or_reject(
...
+                                    kwargs=kwargs,
+                                    created_at=now,
+                                    grace_seconds=grace_seconds,
+                                ),
+                            )
+                            break
+                        except Exception as exc:
+                            if _is_unique_violation(exc) and attempt + 1 < max_retries:
+                                continue
+                            raise
```
willem-bd, line 848: **Exhausted retries re-raise raw `IntegrityError` (→ 500), inconsistent with the `reject` path.**

The `reject` branch above converts `_is_unique_violation → ConflictError` (→ 409). Here, on the 3rd unique-violation (`attempt + 1 >= max_retries`), control falls through to `raise` and surfaces the raw `IntegrityError`.

If this race is reachable enough to warrant a 3× retry loop, the exhausted-retry outcome should also surface as `ConflictError` for consistent caller handling — e.g. convert one more time after the loop, mirroring the `reject` handling.

backend/packages/harness/deerflow/config/run_ownership_config.py
```diff
@@ -0,0 +1,29 @@
...
+
+    lease_seconds: int = Field(
+        default=30,
+        ge=5,
+        description="Seconds before a run lease expires if not renewed.",
+    )
+    grace_seconds: int = Field(
+        default=10,
+        ge=0,
+        description="Extra seconds past lease expiry before an orphaned run is reclaimed.",
```
willem-bd: **Clock-skew tolerance is thin for a mechanism that reclaims a peer worker's in-flight run as `error` (discarding its work).**

`reconcile_orphaned_inflight_runs` compares another worker's UTC lease timestamp against this worker's `datetime.now(UTC)`. The only skew budget is `grace_seconds` (default 10s). Two workers >10s apart under load can cause Worker B to kill Worker A's healthy run. There's no documented NTP assumption or clock-discipline guard.

Suggestion: note the clock-sync requirement in this description (or in `config.example.yaml`), and consider a larger default `grace_seconds` for multi-worker mode.

backend/packages/harness/deerflow/runtime/runs/store/base.py
```diff
@@ -142,3 +144,66 @@ async def aggregate_tokens_by_thread(self, thread_id: str, *, include_active: bo
...
+        self,
+        *,
+        before: str | None = None,
+        grace_seconds: int = 10,
+    ) -> list[dict[str, Any]]:
+        """Return active runs whose lease has expired (or is NULL for pre-ownership rows)."""
+        pass
+
+    @abc.abstractmethod
+    async def claim_inflight_runs(
```
willem-bd: **`claim_inflight_runs` has no caller anywhere in the repo** — not in `RunManager`, not in tests. `create_run_atomic` does its own inline claim (`SELECT FOR UPDATE` + cancel) and the manager only ever calls that.

This adds ~40 lines across `base.py`, `memory.py`, and `sql.py` (plus an unused `now_iso` parameter in both impls), and forces every future `RunStore` implementation to provide it. Either wire it in or remove it (and the unused `now_iso` param).

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -762,9 +899,100 @@ async def cleanup(self, run_id: str, *, delay: float = 300) -> None:
...
+
+        for run_id, record in active_runs:
+            try:
+                updated = await self._store.update_lease(
+                    run_id,
+                    owner_worker_id=self._worker_id,
+                    lease_expires_at=new_expiry,
+                )
+                if updated:
+                    record.lease_expires_at = new_expiry
```
willem-bd, line 1086: **`record.lease_expires_at = new_expiry` is written without `self._lock`**, while `set_status` / `_persist_status` can touch the same record concurrently.

It's a single string field so not corruption-level, but for consistency with the locked snapshot taken above, the write should happen under the lock (or carry a short comment explaining why the unsynchronized write is benign).

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -762,9 +899,100 @@ async def cleanup(self, run_id: str, *, delay: float = 300) -> None:
...
+
+    async def _renew_leases(self) -> None:
+        """Renew the lease on every locally-owned active run."""
+        if self._store is None or self._run_ownership_config is None:
+            return
+        lease_seconds = self._run_ownership_config.lease_seconds
+        new_expiry = (datetime.now(UTC) + timedelta(seconds=lease_seconds)).isoformat()
+
+        async with self._lock:
+            active_runs = [(rid, record) for rid, record in self._runs.items() if record.status in (RunStatus.pending, RunStatus.running) and record.owner_worker_id == self._worker_id and record.task is not None and not record.task.done()]
```
willem-bd: **The renewal filter requires `record.task is not None and not record.task.done()`.** A run sitting in `pending` before its task is spawned won't be renewed; if it stays pending longer than `lease_seconds`, another worker's reconciliation can reclaim it as `error`.

Probably fine given fast pending→running transitions, but worth either a comment acknowledging the window or broadening the filter to pending-state records owned by this worker (i.e. drop the `task is not None` requirement for `pending` runs).

willem-bd (commented):
Reviewing the multi-worker ownership changes — a few findings (severity-ranked). The design is sound; these are mostly around the `IntegrityError` handling and some dead code.

heart-scalpel:
I still have a few issues to resolve and will submit the revised content in batches.

backend/packages/harness/deerflow/persistence/migrations/versions/0004_run_ownership.py
```diff
@@ -0,0 +1,60 @@
...
+            batch_op.create_index("ix_runs_lease", ["lease_expires_at"], unique=False)
+    if "uq_runs_thread_active" not in existing:
+        with op.batch_alter_table("runs", schema=None) as batch_op:
+            batch_op.create_index(
+                "uq_runs_thread_active",
+                ["thread_id"],
+                unique=True,
+                sqlite_where=sa.text("status IN ('pending', 'running')"),
+                postgresql_where=sa.text("status IN ('pending', 'running')"),
+            )
```
willem-bd, line 117: **Migration creates the partial UNIQUE index without deduping pre-existing duplicate active rows (upgrade blocker).**

`upgrade()` issues `create_index("uq_runs_thread_active", unique=True, ...)` directly with no prior cleanup. Any DB that already has two+ `pending`/`running` rows for the same `thread_id` fails: `could not create unique index` (Postgres) / `UNIQUE constraint failed` (SQLite), aborting `alembic upgrade head` so the gateway can't start. That dirty state is exactly what this race fix is about, and it's reachable in the field — Postgres deployments had reconciliation skipped entirely by the old sqlite-only gate, and anyone who tried `GATEWAY_WORKERS>1` before this PR can have duplicates.

Fix: add a dedupe/cancel pass before `create_index`, e.g. keep the newest active row per thread and set the rest to `error` (`... GROUP BY thread_id HAVING COUNT(*) > 1`).

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -762,9 +931,100 @@ async def cleanup(self, run_id: str, *, delay: float = 300) -> None:
...
```
willem-bd: **Heartbeat skips pending runs that haven't been assigned a task yet, so their lease can expire while legitimately queued.**

`_renew_leases` filters on `record.task is not None and not record.task.done()`. A run just created by `create_or_reject` is registered with `task=None` until the launch path assigns it; under event-loop load that window can approach `lease_seconds` (30s). With no renewal, a peer worker's `create_or_reject("interrupt")` then sees the run's lease as expired, interrupts it, and starts a new one — silently killing a legitimately-queued run.

Consider renewing any run owned by this worker that is `pending`/`running`, regardless of `task`.

backend/packages/harness/deerflow/runtime/runs/store/memory.py
```diff
@@ -166,3 +167,156 @@ async def aggregate_tokens_by_thread(self, thread_id: str, *, include_active: bo
...
+                        lease_dt = datetime.fromisoformat(existing_lease)
+                        if lease_dt >= cutoff and r.get("owner_worker_id") != owner_worker_id:
+                            # Live run owned by another worker — cannot
+                            # interrupt, and the partial unique index would
+                            # reject the INSERT anyway. Surface as ConflictError
+                            # so the caller gets a clean signal.
+                            raise ConflictError(f"Thread {thread_id} already has an active run owned by another worker")
+                    except (ValueError, TypeError):
+                        pass
+                r["status"] = "interrupted"
```
willem-bd, line 302: **`create_run_atomic` interrupt loop mutates earlier rows before raising `ConflictError` on a later one — no rollback, diverging from the transactional SQL store.**

The loop sets `r["status"] = "interrupted"` (line 298) as it iterates, then raises `ConflictError` on a subsequent valid-lease-other-worker row (line 295). The already-interrupted rows stay interrupted. `sql.py` is correct because the whole thing is one transaction (a raise rolls everything back). Production (SQL) is fine, but the entire `test_multi_worker_run_ownership.py` suite runs against `MemoryRunStore`, so this divergence gives false confidence — a multi-row interrupt scenario passes in memory and behaves differently in SQL. Make the memory path atomic, or decide before mutating.

backend/packages/harness/deerflow/persistence/run/sql.py
```diff
@@ -376,3 +382,177 @@ async def aggregate_tokens_by_thread(self, thread_id: str, *, include_active: bo
...
+                    select(RunRow)
+                    .where(
+                        RunRow.thread_id == thread_id,
+                        RunRow.status.in_(("pending", "running")),
+                    )
+                    .with_for_update()
+                )
+                result = await session.execute(stmt)
+                for row in result.scalars():
+                    if row.lease_expires_at is not None and row.lease_expires_at >= cutoff and row.owner_worker_id != owner_worker_id:
```
willem-bd: **tz-naive `lease_expires_at` comparison raises `TypeError` if heartbeat is ever enabled on SQLite.**

This line compares the ORM attribute `row.lease_expires_at` (naive on SQLite — this file's own comment at line 76 says SQLite drops tzinfo on read) against `cutoff = datetime.now(UTC) - ...` (aware) **in Python** -> `TypeError: can't compare offset-naive and offset-aware datetimes`, surfacing as an opaque 500.

Harmless under defaults (heartbeat off -> leases always NULL -> `is not None` short-circuits), but there's no guard against `heartbeat_enabled=true` on SQLite. Either reject that combination or coerce tzinfo before comparing. (`list_inflight_with_expired_lease` is safe — its comparison is SQL-side.)

backend/packages/harness/deerflow/runtime/runs/store/base.py
willem-bd: **`claim_inflight_runs` is dead code mirrored across all three store implementations.**

Nothing calls it — `create_or_reject` uses `create_run_atomic`, which inlined the claim (SELECT FOR UPDATE + cancel + INSERT in one transaction). This abstract def plus the matching impls in `sql.py` and `memory.py` (~60 lines total) must be kept in sync with `create_run_atomic`'s lease logic forever and are exercised by nothing (only a no-op fake-store stub in `test_run_repository.py`). Recommend removing all three definitions.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -30,6 +33,62 @@
...
+            return True
+        if getattr(current, "sqlite_errorcode", None) == _SQLITE_UNIQUE_ERRORCODE:
+            return True
+
+        message = str(current).lower()
+        if "unique constraint failed" in message:
+            return True
+        if "unique" in message and "violat" in message:
+            return True
+        if "duplicate key" in message:
```
willem-bd: **`_is_unique_violation` message fallbacks can misclassify a non-IntegrityError as a unique violation.**

The driver-code path (`pgcode`/`sqlite_errorcode`) is reliable and load-bearing. But these fallbacks fire whenever the driver exception isn't reachable through the cause chain, so any application exception whose `str()` contains `"duplicate key"` (or `"unique"` + `"violat"`) — e.g. a CHECK constraint or a validation error — would be converted to `ConflictError` (HTTP 409) instead of surfacing as a 500.

Gate the message fallbacks on `isinstance(current, IntegrityError)` before trusting the substrings.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -762,9 +931,100 @@ async def cleanup(self, run_id: str, *, delay: float = 300) -> None:
...
+        stop = self._heartbeat_stop
+
+        while not stop.is_set():
+            try:
+                await asyncio.wait_for(stop.wait(), timeout=interval)
+                break  # stop event was set
+            except TimeoutError:
+                pass  # interval elapsed, renew leases
+
+            await self._renew_leases()
```
willem-bd: **`_renew_leases()` is the only store call that bypasses `_call_store_with_retry`, and the call is unguarded in the loop.**

Every other store call in the manager goes through the retry wrapper; `update_lease` is called directly inside `_renew_leases`, and `_renew_leases` is invoked here with no `try/except` (only the per-run updates are guarded). A transient error from the lock/snapshot path (or a `CancelledError` racing shutdown) would kill the heartbeat task silently — after which no lease is ever renewed again and every active run eventually looks orphaned.

Wrap the call (retry + `except Exception: logger.warning(...)`).

willem-bd (commented):
Reviewed for correctness/concurrency on the atomic-create, lease heartbeat, and migration paths. The core transactional design (`create_run_atomic` SELECT-FOR-UPDATE + cancel + INSERT), the driver-code `_is_unique_violation` detection, ConflictError/retry semantics, and the SQLite single-worker backward-compat path all check out. Findings below, most severe first.

Cleanup-only (not posted inline):
- `list_inflight` is now dead in production (only `test_run_repository.py` references it) — consider removing.
- The migration idempotency comment (lines 33-35) misstates the reason: the guard is for crash-recovery between `create_all` and `stamp head`, not "the legacy bootstrap path runs create_all" (legacy uses a restricted baseline `create_all` that skips the runs table).
- The four new `@abc.abstractmethod`s on `RunStore` are a breaking interface change for any third-party `RunStore` implementation — worth a changelog note.

willem-bd (commented):
Multi-worker run ownership review — posting findings inline. Overall the design is sound: the partial unique index as the cross-process atomicity boundary, lease-based reconciliation, and lifecycle ordering (reconcile before heartbeat start; heartbeat stop before in-flight drain) are all correct, and the test suite is thorough. One medium operational note (NULL-lease reclaim during rolling upgrades) and the rest are low/nit. Detail below.

backend/app/gateway/deps.py
```diff
@@ -287,20 +287,29 @@ async def langgraph_runtime(app: FastAPI, startup_config: AppConfig) -> AsyncGen
...
-            await _publish_recovered_run_stream_end(app.state.stream_bridge, recovered_runs, cleanup_delay=cleanup_delay)
-            await _mark_latest_recovered_threads_error(app.state.run_manager, app.state.thread_store, recovered_runs)
+        run_ownership_config = getattr(config, "run_ownership", None)
+        app.state.run_manager = RunManager(
+            store=app.state.run_store,
+            run_ownership_config=run_ownership_config,
+        )
+        # Startup recovery: mark inflight runs whose lease has expired as error.
+        # In single-worker mode (SQLite / backend=memory), no run has a lease, so
+        # all inflight rows are reclaimed (unchanged behaviour). In multi-worker
```
willem-bd, line 315: **⚠️ Operational note (medium) — NULL-lease reclaim is unconditional.**

Reconciliation now runs on every backend and reclaims *all* inflight rows whose lease is NULL **or** expired. That's correct for crash recovery, but during a **rolling upgrade** an old-code worker (which never sets leases) still owns live runs with NULL leases — a newly-upgraded worker starting up will see those NULL leases, find no local task for them, and mark them `error`, silently killing live runs on the other worker.

The PR body's "no window where Postgres would mis-mark live Worker A runs as orphans" holds for *lease-backed* runs but not for NULL-lease ones.

Suggestion: note in `docs/multi_worker.md` / the operator checklist that all workers must be upgraded together before enabling `heartbeat_enabled`, since first-boot reconciliation will reclaim pre-existing (old-code) NULL-lease inflight rows.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -381,6 +463,8 @@ async def create(
             user_id=user_id,
             created_at=now,
             updated_at=now,
+            owner_worker_id=self._worker_id,
```
willem-bd, line 476: **(low) `create()` bypasses the atomic primitive.**

`create()` still persists via `_persist_new_run_to_store` → `store.put` (full-row upsert), not `create_run_atomic`. The production run path uses `create_or_reject` (services.py), so this isn't a production regression — but `create` is public, and if it's ever called for a thread that already has an active row, the partial unique index now rejects the INSERT with a raw `IntegrityError` (HTTP 500) instead of the `ConflictError` (409) that `create_or_reject` yields.

Consider routing `create` through `create_run_atomic`, or noting the "assumes no active run exists" assumption in its docstring.

backend/packages/harness/deerflow/persistence/run/sql.py
```diff
@@ -376,3 +382,149 @@ async def aggregate_tokens_by_thread(self, thread_id: str, *, include_active: bo
...
+        lease_expires_at: str,
+    ) -> bool:
+        lease_dt = datetime.fromisoformat(lease_expires_at)
+        values: dict[str, Any] = {
+            "owner_worker_id": owner_worker_id,
+            "lease_expires_at": lease_dt,
+            "updated_at": datetime.now(UTC),
+        }
+        async with self._sf() as session:
+            result = await session.execute(update(RunRow).where(RunRow.run_id == run_id, RunRow.status.in_(("pending", "running"))).values(**values))
```
willem-bd: **(low) `update_lease` doesn't verify ownership.**

This filters by `run_id` + `status` but **sets** `owner_worker_id` to whatever the caller passes without checking the existing owner. The manager always filters `record.owner_worker_id == self._worker_id` before calling it, so it's safe today — but a misuse would silently *transfer* ownership of a run. Consider scoping the `WHERE` to also match `owner_worker_id` so the primitive is safe by construction.

backend/packages/harness/deerflow/persistence/run/model.py
```diff
@@ -44,7 +44,25 @@ class RunRow(Base):
     # Follow-up association
     follow_up_to_run_id: Mapped[str | None] = mapped_column(String(64))
 
+    # Multi-worker run ownership
+    owner_worker_id: Mapped[str | None] = mapped_column(String(128), nullable=True)
```
willem-bd, line 48: **(low) Column width.**

`String(128)` holds `hostname + ":" + uuid4.hex`. K8s pod names are short so this is fine in practice, but Postgres `VARCHAR(128)` *errors* on overflow (it does not truncate) — a pathological hostname (>95 chars) would crash run creation. Consider `String(255)` if hostnames aren't guaranteed bounded.

backend/packages/harness/deerflow/runtime/runs/store/memory.py
```diff
@@ -146,10 +151,6 @@ async def aggregate_tokens_by_thread(self, thread_id: str, *, include_active: bo
                     entry["tokens"] += usage.get("total_tokens", 0)
                     entry["runs"] += 1
             else:
-                # Fallback for rows written before per-model accounting landed:
-                # attribute the whole run to its single ``model_name``. Keeps
-                # the legacy lead-only behavior for old data instead of
-                # silently dropping it.
                 model = r.get("model_name") or "unknown"
```
willem-bd, line 157: **(nit) Unrelated change.**

The explanatory comment block that lived here ("Fallback for rows written before per-model accounting landed…") was deleted with no code change, and it's outside this PR's scope (multi-worker ownership). The context was useful — consider restoring it, or moving the deletion to a separate cleanup PR.

backend/packages/harness/deerflow/persistence/run/sql.py
```diff
@@ -376,3 +382,149 @@ async def aggregate_tokens_by_thread(self, thread_id: str, *, include_active: bo
...
+            claimed: list[dict[str, Any]] = []
+
+            if multitask_strategy in ("interrupt", "rollback"):
+                stmt = (
+                    select(RunRow)
+                    .where(
+                        RunRow.thread_id == thread_id,
+                        RunRow.status.in_(("pending", "running")),
+                    )
+                    .with_for_update()
```
willem-bd, line 498: **(note) Real-Postgres verification gap.**

The interrupt path's "raise `ConflictError` mid-transaction → roll back earlier row cancellations" semantics are only tested on the memory store; for SQL they rely on transaction semantics by construction. Given `risk:high`, the cross-worker races against real Postgres `SELECT FOR UPDATE` + partial unique index behavior are exactly what's most likely to diverge from the SQLite fixtures — worth landing the operator validation (concurrent reject/interrupt, lease-expiry reclaim on real Postgres) before rollout.

heart-scalpel:
This PR cannot be marked ready for merge until real multi-worker validation is completed. I will finish implementing multi-worker correctness test cases, generate the test report, and address all feedback from the code review by July 12.

rayhpeng:
**Verdict: Request changes** — the implementation is solid and the design is the right one; there's one production-safety gap I'd want closed before merge, plus two nits.

### Overall

Strong PR. It correctly closes the prerequisite that #3191 explicitly called out ("*Do not broaden this into a shared-database multi-pod reaper without adding worker ownership/liveness first*"). The architecture is sound: cross-process atomicity comes from a **DB-level partial unique index** (`uq_runs_thread_active`) rather than an in-process lock, and orphan recovery is handled by **lease + heartbeat + reconciliation**. Backward compatibility for single-worker (`NULL` lease → immediate reclaim) is clean, and test coverage is broad.

### What's solid

- **Partial unique index for atomicity** — reuses the proven `uq_channel_connection_active_identity` pattern; the invariant is enforced by the database, not a process lock. Correct call.
- **Migration self-heals dirty data** — `_dedupe_active_runs_per_thread()` cancels superseded active rows (keeping newest per thread, deterministic tiebreak) *before* building the unique index, with an e2e regression test seeding a realistic pre-0004 multi-worker DB. This defuses the most concrete upgrade risk: `CREATE UNIQUE INDEX` aborting the alembic upgrade on existing duplicates.
- **`_is_unique_violation` is driver/locale-robust** — prefers `pgcode`/`sqlite_errorcode`, and gates the message-substring fallback on an `IntegrityError`-typed node so unrelated exceptions can't be misclassified as 409.
- **Heartbeat is defensive** — can't die silently, handles the "pending run whose task isn't spawned yet" window, and orders start/stop (reconcile → start; stop → drain) to avoid racing shutdown.

### Findings

#### 🟡 Blocker (medium) — nothing enforces `GATEWAY_WORKERS>1 ⟹ heartbeat_enabled`

The #3960 startup gate (`app/gateway/deps.py`) only checks `backend == postgres`. There is **no** check tying multi-worker to the heartbeat. So this config passes the gate and starts normally:

```
GATEWAY_WORKERS>1  +  Postgres  +  heartbeat_enabled=false
```

But with heartbeat off, `_compute_lease_expires_at()` always returns `None`, so every run has a `NULL` lease → `list_inflight_with_expired_lease` treats **every NULL-lease row as an orphan** → on startup, Worker B's reconciliation marks **Worker A's live runs as `error`**. The local-task skip in `reconcile_orphaned_inflight_runs` doesn't help here — Worker B has no in-memory record for Worker A's runs.

This fires on **every rolling update / HPA scale-up** (each new pod reconciles and kills runs in flight on older pods). Right now it's guarded only by a one-line comment in `config.example.yaml`, and the existing Postgres gate creates a false sense that multi-worker safety is already covered.

**Fix:** extend that same gate to fail-fast (or hard-warn) when `GATEWAY_WORKERS>1` and `heartbeat_enabled=false`, or default heartbeat on under multi-worker. Cheap to add, and it's a data-correctness issue with a common trigger.

#### 🟢 Nit — `ConflictError` layering

`store/sql.py` and `store/memory.py` import `ConflictError` from `runtime.runs.manager` at module top level, but `RunManager` owns the store (store should be the lower layer). No runtime cycle today, but the direction is backwards — consider moving `ConflictError` down into `store/base.py` or a dedicated exceptions module.

#### 🟢 Nit — `MemoryRunStore` compares `created_at` as strings

`list_inflight_with_expired_lease` uses `r.get("created_at","") > now` (ISO string lexical compare) while the SQL store compares `datetime`. Fine when formats match, fragile if a naive / `Z` / offset value slips in. Memory backend only, so low risk — but worth aligning on `datetime.fromisoformat`.

### Accepted trade-offs (not blocking)

Clock skew (documented in config + docstring, NTP advised), cross-worker task cancellation (work item 4, explicitly out of scope), SQLite `SELECT FOR UPDATE` being ignored (SQLite is single-worker by the gate).

### Before merge

1. Close the heartbeat-enforcement gap above.
2. Note for rollout: this closes items 2/3 but **cancel (item 4) is still open** — under multi-worker, "cancel landing on a non-owner pod → 409" persists, so `GATEWAY_WORKERS=2` is safe for **data correctness** but cancel UX isn't complete yet.
3. The author's "real Postgres, two workers" race checks are still pending — worth running once before merge (esp. concurrent `reject` yielding exactly one success, and B reclaiming after A's lease expires).

backend/packages/harness/deerflow/runtime/runs/manager.py
heart-scalpel: create() docstring notes it bypasses atomic create_run_atomic and
  assumes no active run exists for the thread

backend/packages/harness/deerflow/persistence/run/model.py
heart-scalpel: Approved: String(128) is sufficient for hostname:uuid4.hex. It reserves buffer for edge-case long hostnames with negligible storage cost.

backend/packages/harness/deerflow/persistence/run/sql.py
heart-scalpel: Agreed, this gap is acknowledged. Operator validations covering cross-worker concurrent reject/interrupt races and lease expiry reclaim against real Postgres are in progress and will be completed before rollout.

heart-scalpel:
@rayhpeng  CI is green now. Could you please help review this again when you have a chance? Thanks!

heart-scalpel:
@rayhpeng 
On item 4 (cancel):
Agreed, this is deliberately deferred. This PR already covers DB migration, create_or_reject atomicity, lease heartbeat and cross-process reconciliation. Adding cross-worker cancel takeover would bloat the diff and complicate reviews. Item 4 is self-contained and can ship separately. Currently cancel routed to a non-owner pod returns 409 — suboptimal UX but safe for data consistency, matching the design scope boundary. It will be addressed in a follow-up PR.
On real Postgres two-worker race validation:
Race conditions are covered in test_multi_worker_run_ownership.py using dual RunManager + MemoryRunStore to simulate cross-process behaviour. Before merge I will complete manual validation against real Postgres with two uvicorn workers and share results here.
Thanks for the careful scrutiny. Minor review feedback (create() docstring, update_lease ownership guard, missing comments) has been resolved in the latest commit.

willem-bd (commented):
Left a few inline comments from review. Substantive findings:

1. **Reconciliation is startup-only** — a crashed worker replaced before its lease expires leaves orphaned runs until the next pod restart; the heartbeat loop is the natural place for a periodic sweep.
2. **`_is_unique_violation` misses psycopg3's `sqlstate`** — on the only supported multi-worker backend, detection falls through to the message-substring fallback. One-line hardening.
3. **Postgres single-worker behavior change** — inflight runs are now reclaimed at startup where they previously weren't; worth a release note + checking the "Default behavior change" box.
4. **Local lock held across store calls with retries** — serializes run creation; non-blocking, flagging for throughput awareness.

Overall the design (partial unique index as the cross-process atomicity guarantee, lease-based reconciliation, default-off gating) is solid, and test coverage is strong on SQLite/MemoryRunStore. The highest-value pre-merge item is #2 (Postgres native-code detection), since it's load-bearing and untested in CI.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -762,9 +949,132 @@ async def cleanup(self, run_id: str, *, delay: float = 300) -> None:
...
+            # silently. The per-run try/except inside ``_renew_leases``
+            # already handles store failures, but a transient error from
+            # the snapshot path or an unexpected exception must not take
+            # the heartbeat down — a dead heartbeat means no lease is
+            # renewed again, and every active run eventually looks
+            # orphaned to peers. ``except Exception`` lets
+            # ``CancelledError`` (a ``BaseException`` since 3.8) propagate
+            # so shutdown cancellation still works.
+            try:
+                await self._renew_leases()
```
willem-bd, line 1032: **Reconciliation is startup-only — a crashed worker replaced before its lease expires leaves runs orphaned.**

`reconcile_orphaned_inflight_runs` has a single production caller (the lifespan startup in `deps.py`); this heartbeat loop only renews *this* worker's leases and never reclaims expired leases from dead peers.

Failure path in a stable cluster:
1. Worker A holds run R, lease renewed to T+30s.
2. Worker A hard-crashes at T+1s.
3. Replacement A′ starts at T+5s and runs startup reconciliation — R's lease (T+30s) is still valid, so it's **skipped**.
4. A′ never reconciles again. R's lease expires at T+30s but nothing reclaims it; R stays `pending`/`running` until the *next* pod restart of any worker.

So breaker #3 is reduced to "recovered on next restart," not fully closed. The heartbeat loop (already firing every `lease_seconds / 3`) is the natural place to also sweep for expired leases. If periodic reconciliation is intentionally out of scope here, worth stating that explicitly in `docs/multi_worker.md`.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -30,6 +35,70 @@
...
+    seen: set[int] = set()
+    while pending:
+        current = pending.pop()
+        if id(current) in seen:
+            continue
+        seen.add(id(current))
+
+        if getattr(current, "pgcode", None) == _UNIQUE_PGCODE:
+            return True
+        if getattr(current, "sqlcode", None) == _UNIQUE_PGCODE:
```
willem-bd, line 75: **`_is_unique_violation` doesn't match psycopg3's native code attribute.**

This checks `pgcode` (psycopg2) and `sqlcode` (Oracle/DB2) but not `sqlstate` (psycopg3 — the modern default for SQLAlchemy 2.0 + Postgres). On a psycopg3-backed deployment the native-code path misses and detection falls through to the message-substring fallback, which the docstring itself calls "belt-and-suspenders" and fragile across locales.

The message path happens to work today ("duplicate key value violates unique constraint" contains "duplicate key"), but Postgres is the *only* supported multi-worker backend and the test suite can't exercise this — every DB-backed test uses SQLite. One-line hardening closes the gap:

```python
if getattr(current, "sqlstate", None) == _UNIQUE_PGCODE:
    return True
```

backend/app/gateway/deps.py
```diff
@@ -287,20 +305,29 @@ async def langgraph_runtime(app: FastAPI, startup_config: AppConfig) -> AsyncGen
...
+        # mode (Postgres), only runs with an expired lease are reclaimed; runs
+        # owned by another live worker are skipped.
+        from deerflow.utils.time import now_iso
+
+        recovered_runs = await app.state.run_manager.reconcile_orphaned_inflight_runs(
```
willem-bd, line 320: **Behavior change for Postgres *single-worker* deployments — worth a release note.**

Dropping the sqlite-only gate means Postgres now reclaims inflight runs at startup where it previously did not. This is the intended fix for breaker #3, but it's a default behavior change for existing single-worker Postgres users: on the next restart, every pre-existing `pending`/`running` row is marked `error`. Desirable, but surprising if anyone relied on those rows sitting there.

The PR body's "Default behavior change" checkbox is currently unchecked — I'd check it and add a CHANGELOG line so operators aren't caught off guard.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -619,63 +726,136 @@ async def create_or_reject(
...
+                        raise
+                    except Exception as exc:
+                        if _is_unique_violation(exc):
+                            raise ConflictError(f"Thread {thread_id} already has an active run") from exc
+                        raise
+                else:
+                    # Interrupt / rollback: store-side claim + insert in one
+                    # transaction. Retry on IntegrityError in case another
+                    # worker races us between our SELECT FOR UPDATE and INSERT.
+                    max_retries = 3
```
willem-bd, line 816: **`self._lock` is held across the store call with up to 3 retry attempts.**

For interrupt/rollback, the local lock is held across `create_run_atomic`, which can do up to 3 attempts of `SELECT FOR UPDATE` + INSERT against Postgres. This serializes all run creation/cancellation on the worker.

Not a regression (the old code also held the lock across the store write) and it's correctness-motivated by the lock-ordering invariant in the docstring, but under concurrency with DB round-trips it adds head-of-line latency. Flagging in case throughput becomes a concern — not blocking.

heart-scalpel: Acknowledged. The lock-ordering invariant in the docstring requires the local lock to span check → store insert → register, otherwise a same-worker ConflictError could leave a leaked pending row. The retry loop only fires on cross-worker IntegrityError (rare in practice); the first attempt almost always succeeds. If throughput becomes a concern later, we could narrow the held section, but correctness first — not blocking.

backend/app/gateway/deps.py
heart-scalpel: Good catch — checked the "Default behavior change" box and added a note to the rollout section. The previous backend == "sqlite" gate was a bug, not a feature: single-worker Postgres deployments had no recovery path for inflight runs after a crash. Aligning Postgres with the existing SQLite behavior is the intended fix for breaker #3. For operators upgrading, the surface is narrow — runs that were pending/running at restart are marked error with a clear message, which is strictly better than an indefinite UI spinner that only a DBA can clear.

heart-scalpel:
@WillemJiang hi CI is green. All findings from the last round are addressed.Ready for another pass — thanks for the thorough review.

rayhpeng:
<p>Read the whole change plus the full test suite, and <strong>checked out the PR branch and reproduced the one blocking issue end-to-end</strong> (real SQLite, real <code>uq_runs_thread_active</code> index, real <code>reconcile</code>, two <code>RunManager</code>s = two pods sharing one DB). This is a well-scoped, high-quality PR that does exactly what #3191 said had to come first — adds run <strong>ownership + liveness</strong> — and correctly limits itself to work items 2–3. My feedback is one blocker (with a repro), plus a few non-blocking items.</p>
<h3>🟢 What's solid</h3>
<ul>
<li><strong>Cross-process atomicity via a partial unique index (<code>uq_runs_thread_active</code>), not a process lock</strong> — the right call, reusing the proven <code>uq_channel_connection_active_identity</code> pattern; the in-process lock stays as a same-worker fast path.</li>
<li><strong>Lease + heartbeat for liveness</strong> cleanly lets <code>reconcile</code> run on every backend and drops the sqlite-only gate.</li>
<li><strong>Migration self-heals</strong> dirty data (<code>_dedupe_active_runs_per_thread</code> + e2e test) so unique-index creation can't abort the upgrade.</li>
<li><strong><code>_is_unique_violation</code> is robust and well-tested</strong> — driver-native codes first, message fallback gated on an <code>IntegrityError</code> node, with direct tests for both the real-SQLite path and the "don't misclassify an app exception as 409" guard.</li>
<li>Bonus: you found and regression-tested a real SQLite tz-naive-vs-aware <code>TypeError</code> in the interrupt comparison.</li>
</ul>
<h3>🔴 Blocking — nothing enforces <code>GATEWAY_WORKERS&gt;1 ⇒ heartbeat_enabled</code> (reproduced)</h3>
<p>The #3960 gate only checks <code>backend == postgres</code>. There's no check tying multi-worker to the heartbeat, so <code>GATEWAY_WORKERS&gt;1 + Postgres + heartbeat_enabled=false</code> starts cleanly and then corrupts run state: with heartbeat off, <code>_compute_lease_expires_at()</code> returns <code>None</code> → every run has a <code>NULL</code> lease → <code>list_inflight_with_expired_lease</code> treats <strong>every NULL-lease row as an orphan</strong> → on boot, <strong>Worker B marks Worker A's live runs as <code>error</code></strong>. Fires on every rolling update / scale-up.</p>
<p><strong>Reproduced on this branch</strong> — two <code>RunManager</code>s (different <code>worker_id</code>s) sharing one real SQLite DB; the only variable is <code>heartbeat_enabled</code>:</p>

reading | ❌ heartbeat=false | ✅ heartbeat=true
-- | -- | --
lease_expires_at in DB | None | 2026-…T16:46:49+00:00
Worker A's local view | running | running
DB status after B's reconcile | error | running
B reclaimed | ['051ee0ba…'] | []


<p>The failure mode is the exact tension this PR exists to fix, turned against itself: <strong>the same run is <code>running</code> in Worker A's memory and <code>error</code> in the shared DB.</strong> Worker A's asyncio task is still executing — it has no idea its run was declared dead — so the user sees a phantom failure while A keeps burning compute on a run the DB already killed. The two rows differ by one boolean, which pins the causality: same two workers, same startup reconcile, <code>heartbeat</code> off kills, on is safe.</p>
&lt;details&gt;&lt;summary&gt;Repro script (runs from &lt;code&gt;backend/&lt;/code&gt;: &lt;code&gt;PYTHONPATH=. uv run python repro.py&lt;/code&gt;)&lt;/summary&gt;
<pre><code class="language-python">import asyncio, shutil, tempfile
from pathlib import Path
from deerflow.config.run_ownership_config import RunOwnershipConfig
from deerflow.persistence.engine import close_engine, get_session_factory, init_engine
from deerflow.persistence.run import RunRepository
from deerflow.runtime import RunManager, RunStatus

async def run_scenario(heartbeat_enabled: bool, tmp: Path) -&gt; dict:
    await init_engine("sqlite", url=f"sqlite+aiosqlite:///{tmp/'runs.db'}", sqlite_dir=str(tmp))
    shared_db = RunRepository(get_session_factory())   # one DB, shared by both "pods"
    def cfg(): return RunOwnershipConfig(lease_seconds=30, grace_seconds=10, heartbeat_enabled=heartbeat_enabled)

    # Worker A: a live pod executing a run on thread-1
    worker_a = RunManager(store=shared_db, run_ownership_config=cfg(), worker_id="worker-A")
    run = await worker_a.create("thread-1", user_id="repro-user")
    await worker_a.set_status(run.run_id, RunStatus.running)
    if heartbeat_enabled:
        await worker_a._renew_leases()                 # A's heartbeat renewing its lease

    # Worker B: a second pod boots (scale-up / rolling deploy) and runs startup reconcile
    worker_b = RunManager(store=shared_db, run_ownership_config=cfg(), worker_id="worker-B")
    recovered = await worker_b.reconcile_orphaned_inflight_runs(error="Gateway restarted…")

    row = await shared_db.get(run.run_id, user_id=None)
    await close_engine()
    return {"lease": row["lease_expires_at"], "a_view": str(worker_a._runs[run.run_id].status),
            "db_status": row["status"], "reclaimed": [r.run_id for r in recovered]}

async def main():
    for label, hb in [("heartbeat=FALSE", False), ("heartbeat=TRUE", True)]:
        tmp = Path(tempfile.mkdtemp())
        try: res = await run_scenario(hb, tmp)
        finally: shutil.rmtree(tmp, ignore_errors=True)
        print(label, res, "  &lt;-- BUG" if res["db_status"] == "error" else "  &lt;-- safe")

asyncio.run(main())
</code></pre>
&lt;/details&gt;
<p><strong>Why the green tests miss it:</strong> <code>test_heartbeat_disabled_crashed_run_reclaimed_immediately</code> asserts this exact reclaim as <em>correct</em> — but from a single-worker viewpoint (<code>manager_a</code> crashed, <code>manager_b</code> restarted). Neither the test nor the code can tell "B is A's restart" from "B is a concurrent peer," so the suite actively vouches for the path that corrupts data under multi-worker. The ambiguity has to be removed at the config layer.</p>
<p><strong>Ask:</strong> extend the existing <code>GATEWAY_WORKERS</code> gate to fail-fast (or hard-warn) when <code>GATEWAY_WORKERS&gt;1 &amp;&amp; heartbeat_enabled=false</code>, or default heartbeat on under multi-worker. Small change; closes a data-correctness footgun with a very common trigger. Happy to send a patch — the repro flips from <code>error</code> to a refused startup once the gate lands.</p>
<h3>🟡 Non-blocking</h3>
<ul>
<li><strong>No automated cross-process race test.</strong> Atomicity tests are serial (real SQLite, sequential inserts) — the "two workers INSERT the same thread, one wins" guarantee rides on the pending operator validation, so the real-Postgres/two-worker checks in the description are a hard prerequisite, not optional.</li>
<li><strong>Heartbeat loop isn't exercised end-to-end</strong> — <code>test_heartbeat_renews_active_run_leases</code> bypasses the loop and calls <code>_renew_leases()</code> directly, so the interval timing and the die-proof <code>except Exception</code> never run. A <code>lease_seconds=5</code> test that lets the loop tick once would lock both.</li>
</ul>
<h3>🟢 Nits</h3>
<ul>
<li><strong><code>ConflictError</code> layering</strong> — <code>store/sql.py</code> and <code>store/memory.py</code> import it from <code>runtime.runs.manager</code>, but the manager owns the store; consider moving it down to <code>store/base.py</code> or an exceptions module.</li>
<li><strong><code>MemoryRunStore</code> compares <code>created_at</code> as strings</strong> while the SQL store uses <code>datetime</code> — memory backend only, low risk, worth aligning.</li>
</ul>
<h3>Accepted trade-offs (not blocking)</h3>
<p>Clock skew (documented + NTP advised), SQLite <code>SELECT FOR UPDATE</code> ignored (single-worker by the gate), cross-worker cancel (item 4, out of scope).</p>
<h3>Before merge</h3>
<ol>
<li><strong>Add the <code>GATEWAY_WORKERS&gt;1 ⇒ heartbeat_enabled</code> enforcement</strong> (the blocker above).</li>
<li>Run the real-Postgres / two-worker race checks from the description.</li>
<li>Rollout note: this closes items 2–3, but <strong>cancel (item 4) is still open</strong> — cancel landing on a non-owner pod still 409s. Worth stating so "close ownership gaps" in the title isn't read as all-closed.</li>
</ol>
<p><strong>Verdict: Request changes</strong> — solid, merge-quality implementation on top of the Redis StreamBridge foundation. The single hold is the heartbeat-enforcement gate: without it, a routine misconfig silently marks live runs <code>error</code> across workers on every deploy — reproduced above — and the current tests vouch for that path. Everything else is non-blocking.</p>

heart-scalpel:
> Read the whole change plus the full test suite, and **checked out the PR branch and reproduced the one blocking issue end-to-end** (real SQLite, real `uq_runs_thread_active` index, real `reconcile`, two `RunManager`s = two pods sharing one DB). This is a well-scoped, high-quality PR that does exactly what #3191 said had to come first — adds run **ownership + liveness** — and correctly limits itself to work items 2–3. My feedback is one blocker (with a repro), plus a few non-blocking items.
> 
> ### 🟢 What's solid
> * **Cross-process atomicity via a partial unique index (`uq_runs_thread_active`), not a process lock** — the right call, reusing the proven `uq_channel_connection_active_identity` pattern; the in-process lock stays as a same-worker fast path.
> * **Lease + heartbeat for liveness** cleanly lets `reconcile` run on every backend and drops the sqlite-only gate.
> * **Migration self-heals** dirty data (`_dedupe_active_runs_per_thread` + e2e test) so unique-index creation can't abort the upgrade.
> * **`_is_unique_violation` is robust and well-tested** — driver-native codes first, message fallback gated on an `IntegrityError` node, with direct tests for both the real-SQLite path and the "don't misclassify an app exception as 409" guard.
> * Bonus: you found and regression-tested a real SQLite tz-naive-vs-aware `TypeError` in the interrupt comparison.
> 
> ### 🔴 Blocking — nothing enforces `GATEWAY_WORKERS>1 ⇒ heartbeat_enabled` (reproduced)
> The #3960 gate only checks `backend == postgres`. There's no check tying multi-worker to the heartbeat, so `GATEWAY_WORKERS>1 + Postgres + heartbeat_enabled=false` starts cleanly and then corrupts run state: with heartbeat off, `_compute_lease_expires_at()` returns `None` → every run has a `NULL` lease → `list_inflight_with_expired_lease` treats **every NULL-lease row as an orphan** → on boot, **Worker B marks Worker A's live runs as `error`**. Fires on every rolling update / scale-up.
> 
> **Reproduced on this branch** — two `RunManager`s (different `worker_id`s) sharing one real SQLite DB; the only variable is `heartbeat_enabled`:
> 
> reading	❌ heartbeat=false	✅ heartbeat=true
> lease_expires_at in DB	None	2026-…T16:46:49+00:00
> Worker A's local view	running	running
> DB status after B's reconcile	error	running
> B reclaimed	['051ee0ba…']	[]
> The failure mode is the exact tension this PR exists to fix, turned against itself: **the same run is `running` in Worker A's memory and `error` in the shared DB.** Worker A's asyncio task is still executing — it has no idea its run was declared dead — so the user sees a phantom failure while A keeps burning compute on a run the DB already killed. The two rows differ by one boolean, which pins the causality: same two workers, same startup reconcile, `heartbeat` off kills, on is safe.
> 
> <details><summary>Repro script (runs from <code>backend/</code>: <code>PYTHONPATH=. uv run python repro.py</code>)</summary>
> ```
> import asyncio, shutil, tempfile
> from pathlib import Path
> from deerflow.config.run_ownership_config import RunOwnershipConfig
> from deerflow.persistence.engine import close_engine, get_session_factory, init_engine
> from deerflow.persistence.run import RunRepository
> from deerflow.runtime import RunManager, RunStatus
> async def run_scenario(heartbeat_enabled: bool, tmp: Path) -> dict:
> await init_engine("sqlite", url=f"sqlite+aiosqlite:///{tmp/'runs.db'}", sqlite_dir=str(tmp))
> shared_db = RunRepository(get_session_factory())   # one DB, shared by both "pods"
> def cfg(): return RunOwnershipConfig(lease_seconds=30, grace_seconds=10, heartbeat_enabled=heartbeat_enabled)
> ```
> # Worker A: a live pod executing a run on thread-1
> worker_a = RunManager(store=shared_db, run_ownership_config=cfg(), worker_id="worker-A")
> run = await worker_a.create("thread-1", user_id="repro-user")
> await worker_a.set_status(run.run_id, RunStatus.running)
> if heartbeat_enabled:
>     await worker_a._renew_leases()                 # A's heartbeat renewing its lease
> 
> # Worker B: a second pod boots (scale-up / rolling deploy) and runs startup reconcile
> worker_b = RunManager(store=shared_db, run_ownership_config=cfg(), worker_id="worker-B")
> recovered = await worker_b.reconcile_orphaned_inflight_runs(error="Gateway restarted…")
> 
> row = await shared_db.get(run.run_id, user_id=None)
> await close_engine()
> return {"lease": row["lease_expires_at"], "a_view": str(worker_a._runs[run.run_id].status),
>         "db_status": row["status"], "reclaimed": [r.run_id for r in recovered]}
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> async def main():
> for label, hb in [("heartbeat=FALSE", False), ("heartbeat=TRUE", True)]:
> tmp = Path(tempfile.mkdtemp())
> try: res = await run_scenario(hb, tmp)
> finally: shutil.rmtree(tmp, ignore_errors=True)
> print(label, res, "  <-- BUG" if res["db_status"] == "error" else "  <-- safe")
> ```
> 
> `asyncio.run(main())
> `
> 
> 
> </details>
> **Why the green tests miss it:** `test_heartbeat_disabled_crashed_run_reclaimed_immediately` asserts this exact reclaim as _correct_ — but from a single-worker viewpoint (`manager_a` crashed, `manager_b` restarted). Neither the test nor the code can tell "B is A's restart" from "B is a concurrent peer," so the suite actively vouches for the path that corrupts data under multi-worker. The ambiguity has to be removed at the config layer.
> 
> **Ask:** extend the existing `GATEWAY_WORKERS` gate to fail-fast (or hard-warn) when `GATEWAY_WORKERS>1 && heartbeat_enabled=false`, or default heartbeat on under multi-worker. Small change; closes a data-correctness footgun with a very common trigger. Happy to send a patch — the repro flips from `error` to a refused startup once the gate lands.
> 
> ### 🟡 Non-blocking
> * **No automated cross-process race test.** Atomicity tests are serial (real SQLite, sequential inserts) — the "two workers INSERT the same thread, one wins" guarantee rides on the pending operator validation, so the real-Postgres/two-worker checks in the description are a hard prerequisite, not optional.
> * **Heartbeat loop isn't exercised end-to-end** — `test_heartbeat_renews_active_run_leases` bypasses the loop and calls `_renew_leases()` directly, so the interval timing and the die-proof `except Exception` never run. A `lease_seconds=5` test that lets the loop tick once would lock both.
> 
> ### 🟢 Nits
> * **`ConflictError` layering** — `store/sql.py` and `store/memory.py` import it from `runtime.runs.manager`, but the manager owns the store; consider moving it down to `store/base.py` or an exceptions module.
> * **`MemoryRunStore` compares `created_at` as strings** while the SQL store uses `datetime` — memory backend only, low risk, worth aligning.
> 
> ### Accepted trade-offs (not blocking)
> Clock skew (documented + NTP advised), SQLite `SELECT FOR UPDATE` ignored (single-worker by the gate), cross-worker cancel (item 4, out of scope).
> 
> ### Before merge
> 1. **Add the `GATEWAY_WORKERS>1 ⇒ heartbeat_enabled` enforcement** (the blocker above).
> 2. Run the real-Postgres / two-worker race checks from the description.
> 3. Rollout note: this closes items 2–3, but **cancel (item 4) is still open** — cancel landing on a non-owner pod still 409s. Worth stating so "close ownership gaps" in the title isn't read as all-closed.
> 
> **Verdict: Request changes** — solid, merge-quality implementation on top of the Redis StreamBridge foundation. The single hold is the heartbeat-enforcement gate: without it, a routine misconfig silently marks live runs `error` across workers on every deploy — reproduced above — and the current tests vouch for that path. Everything else is non-blocking.

Thanks for the repro — it cleanly pins the failure mode. Two notes on the
  current state and a question on direction.

  Already landed: the startup gate in _enforce_postgres_for_multi_worker
  (deps.py:49, commit 91271c1a) now rejects GATEWAY_WORKERS>1 unless both
  database.backend=='postgres' AND run_ownership.heartbeat_enabled==true.
  The repro still fires because it constructs RunManager directly, bypassing
  the deployment boundary — but the production path (Gateway lifespan) exits
  before any persistence engine initialises.

  Your "default heartbeat on under multi-worker" suggestion got me thinking
  about going further: what if heartbeat_enabled defaults to true for
  *all* deployments, not just multi-worker? That would eliminate the NULL
  lease ambiguity entirely — every run gets a lease, reconciliation only
  reclaims genuinely expired ones, single-worker and multi-worker follow
  the same path.

  Trade-off I want to sanity-check with you: single-worker crash recovery
  latency shifts from "immediate on next restart" (NULL lease → reclaim) to
  "at most lease_seconds + grace_seconds" (~40s default). That delay is the
  intended lease safety budget (GC pause, event-loop saturation), but it is
  a behaviour change for existing single-worker Postgres users.

  DB cost on single-worker is negligible — one UPDATE per active run every
  lease_seconds/3 (~10s), zero when idle.

  Would you prefer (A) keep default=false + keep the gate as the sole guard,
  or (B) flip default=true and let the gate become defense-in-depth against
  explicit heartbeat_enabled=false overrides? Lean toward B myself — fewer
  config knobs to get wrong — but the single-worker recovery latency shift
  is real and worth your call.

heart-scalpel:
@rayhpeng hi
  Blocking — heartbeat enforcement gate
  Landed in 91271c1a. _enforce_postgres_for_multi_worker (deps.py:49) now
  rejects GATEWAY_WORKERS>1 unless both database.backend=='postgres' AND
  run_ownership.heartbeat_enabled==true. Tests:
  test_gate_rejects_multi_worker_without_heartbeat,
  test_gate_rejects_multi_worker_without_run_ownership_config,
  test_gate_heartbeat_check_not_triggered_for_single_worker.

  Nits
  - ConflictError: store layer now lazy-imports it from manager inside
    create_run_atomic (no module-level upward dependency).
  - MemoryRunStore created_at: switched to datetime.fromisoformat with
    tz-naive handling aligned with the SQL store. Test added.

  Real Postgres / two-worker race checks
  Ran end-to-end against GATEWAY_WORKERS=2 + Postgres. Results posted in
  the PR description: concurrent reject yields exactly one 200 + one 409;
  Worker B reclaims after Worker A's lease expires; valid-lease runs
  skipped.

  Rollout note
  PR description now states item 4 (cancel landing on non-owner pod → 409)
  is intentionally out of scope for this PR and will follow up separately.

  Separate question for you: would you prefer heartbeat_enabled default to
  true for all deployments (not just multi-worker)? It eliminates the NULL
  lease ambiguity entirely and turns the gate into defense-in-depth against
  explicit overrides. Trade-off is single-worker crash recovery latency
  shifting from immediate to lease_seconds + grace_seconds (~40s). Lean
  toward default=true myself, but wanted your call before changing user-
  visible defaults.

WillemJiang (approved):

PR #4004, opened 2026-07-08
feat: add MCP routing hints

### LINKED ISSUE #3640: [feat] 工具搜索
### Before you start

### Problem / motivation

我在使用的过程中发现一个问题，就是，我添加了一个mcp,里面的工具是查询数据库的，但是每次当我跟智能体对话的时候，智能体默认联网搜索，或者自己回答，怎么样才能让她默认取mcp里面找工具呢，mcp返回的工具列表里面，描述也写的很清楚了，就是不走mcp

### Proposed solution

能不能有一个配置，来排序他们优先级

### Affected area(s)

MCP

### Alternatives considered

_No response_

### Additional context

_No response_

18062706139fcz:
## Summary

Implements PR1 for MCP routing hints, addressing #3640 by adding a soft preference layer for MCP tool selection without changing tool binding.

- add `routing` and per-tool `tools.<name>.routing` schema to `extensions_config.json`
- attach effective MCP routing metadata to MCP tools during discovery
- render `<mcp_routing_hints>` in lead, embedded client, and subagent prompts
- preserve `routing`, `tools`, `tool_call_timeout`, and advanced server extra fields across Gateway MCP config round-trips
- document routing hint behavior and update example config

## Scope

This is intentionally a soft guidance layer. It does not hard-disable `web_search`/`web_fetch`, does not change bound tool schemas, and does not implement auto-promotion for `tool_search.enabled=true`.

## Tests

- `uvx ruff format . && uvx ruff check .`
- `PYTHONPATH=. PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run pytest tests/test_mcp_config_secrets.py tests/test_mcp_routing_config.py tests/test_mcp_routing_metadata.py tests/test_mcp_routing_prompt.py -q`
- earlier related checks: MCP session/sync wrapper tests and client path tests passed during development

Fixes #3640

### DIFF
--- a/backend/app/gateway/routers/mcp.py
+++ b/backend/app/gateway/routers/mcp.py
@@ -1,8 +1,9 @@
 import json
 import logging
 import os
+import re
 from pathlib import Path
-from typing import Literal
+from typing import Any, Literal
 
 from fastapi import APIRouter, HTTPException, Request, status
 from pydantic import BaseModel, ConfigDict, Field
@@ -85,6 +86,55 @@ class McpCacheResetResponse(BaseModel):
 
 
 _MASKED_VALUE = "***"
+_SENSITIVE_EXTRA_KEY_RE = re.compile(
+    r"(^|_)(api_key|apikey|access_key|private_key|client_secret|secret|token|password|passwd|credential|credentials|authorization|bearer)(_|$)",
+    re.IGNORECASE,
+)
+
+
+def _normalize_config_key(key: str) -> str:
+    with_boundaries = re.sub(r"(.)([A-Z][a-z]+)", r"\1_\2", key)
+    with_boundaries = re.sub(r"([a-z0-9])([A-Z])", r"\1_\2", with_boundaries)
+    return re.sub(r"[^a-z0-9]+", "_", with_boundaries.lower()).strip("_")
+
+
+def _is_sensitive_extra_key(key: str) -> bool:
+    return bool(_SENSITIVE_EXTRA_KEY_RE.search(_normalize_config_key(key)))
+
+
+def _mask_sensitive_extra_value(value: Any) -> Any:
+    if isinstance(value, dict):
+        return {key: _MASKED_VALUE if _is_sensitive_extra_key(str(key)) else _mask_sensitive_extra_value(nested) for key, nested in value.items()}
+    if isinstance(value, list):
+        return [_mask_sensitive_extra_value(item) for item in value]
+    return value
+
+
+def _merge_extra_value_preserving_masked(key: str, incoming_value: Any, existing_value: Any, *, existing_present: bool) -> Any:
+    if incoming_value == _MASKED_VALUE and _is_sensitive_extra_key(key):
+        if existing_present:
+            return existing_value
+        raise HTTPException(
+            status_code=400,
+            detail=f"Cannot set extra config key '{key}' to masked value '***'; provide a real value.",
+        )
+
+    if isinstance(incoming_value, dict) and isinstance(existing_value, dict):
+        merged: dict[str, Any] = {}
+        for nested_key, nested_value in incoming_value.items():
+            nested_present = nested_key in existing_value
+            merged[nested_key] = _merge_extra_value_preserving_masked(
+                str(nested_key),
+                nested_value,
+                existing_value.get(nested_key),
+                existing_present=nested_present,
+            )
+        return merged
+
+    if isinstance(incoming_value, list) and isinstance(existing_value, list) and len(incoming_value) == len(existing_value):
+        return [_merge_extra_value_preserving_masked(key, nested_value, existing_value[index], existing_present=True) for index, nested_value in enumerate(incoming_value)]
+
+    return incoming_value
 
 
 def _allowed_stdio_commands() -> set[str]:
@@ -154,11 +204,13 @@ def _mask_server_config(server: McpServerConfigResponse) -> McpServerConfigRespo
                 "refresh_token": None,
             }
         )
+    masked_extra = {key: _MASKED_VALUE if _is_sensitive_extra_key(key) else _mask_sensitive_extra_value(value) for key, value in (server.model_extra or {}).items()}
     return server.model_copy(
         update={
             "env": masked_env,
             "headers": masked_headers,
             "oauth": masked_oauth,
+            **masked_extra,
         }
     )
 
@@ -228,6 +280,15 @@ def _merge_preserving_secrets(
         update["routing"] = existing.routing
     if "tools" not in incoming.model_fields_set:
         update["tools"] = existing.tools
+    incoming_extra = incoming.model_extra or {}
+    existing_extra = existing.model_extra or {}
+    for key, value in incoming_extra.items():
+        update[key] = _merge_extra_value_preserving_masked(
+            key,
+            value,
+            existing_extra.get(key),
+            existing_present=key in existing_extra,
+        )
     for key, value in (existing.model_extra or {}).items():
         if key not in (incoming.model_extra or {}):
             update[key] = value
--- a/backend/app/gateway/routers/mcp.py
+++ b/backend/app/gateway/routers/mcp.py
@@ -5,10 +5,10 @@ from pathlib import Path
 from typing import Literal
 
 from fastapi import APIRouter, HTTPException, Request, status
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field
 
 from app.gateway.deps import require_admin_user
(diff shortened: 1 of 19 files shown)


backend/app/gateway/routers/mcp.py
```diff
@@ -215,13 +219,19 @@ def _merge_preserving_secrets(
...
-    )
+    update = {
+        "env": merged_env,
+        "headers": merged_headers,
+        "oauth": merged_oauth,
+    }
+    if "routing" not in incoming.model_fields_set:
+        update["routing"] = existing.routing
+    if "tools" not in incoming.model_fields_set:
+        update["tools"] = existing.tools
```
willem-bd, line 282: [P2] `tool_call_timeout` is not preserved when a PUT omits it

- **Problem:** `routing` and `tools` get defensive `model_fields_set` preservation here, but `tool_call_timeout` — now a declared field rather than an `extra` entry — does not. If a client PUTs a server config without `tool_call_timeout` (e.g. toggling `enabled`), `incoming.tool_call_timeout` defaults to `None` and `model_copy` writes `null` to disk, silently dropping a configured stdio per-call timeout.
- **Evidence:** `load_server_tools` reads `server_cfg.tool_call_timeout` for stdio wrapping, so the wipe changes real runtime behavior. It only survives today because GET now echoes it back and the documented flow is a full round-trip.
- **Suggested fix:** mirror the sibling handling:
  ```python
  if "tool_call_timeout" not in incoming.model_fields_set:
      update["tool_call_timeout"] = existing.tool_call_timeout
  ```
- **Test:** add a round-trip test asserting a pre-existing `tool_call_timeout` survives a partial PUT (next to the new `_preserves_*` tests).

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -546,6 +546,7 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
     if non_interactive:
         filtered = [tool for tool in filtered if tool.name not in _NON_INTERACTIVE_DISABLED_TOOL_NAMES]
     final_tools, setup = assemble_deferred_tools(filtered, enabled=resolved_app_config.tool_search.enabled)
+    mcp_routing_hints_section = get_mcp_routing_hints_prompt_section(filtered)
```
willem-bd: [P2] Routing hints can point at tools the model can't call when `tool_search` is enabled

- **Problem:** the hint section is built from `filtered` (pre-deferral), but `build_deferred_tool_setup` defers *every* MCP tool (`deferred = [t for t in filtered_tools if is_mcp_tool(t)]`). With `tool_search.enabled=true`, the rendered hint tells the model to "prefer the `postgres_query` tool" while that tool is not bound — it's only reachable via the search/promotion tool.
- **Evidence:** PR1 explicitly defers auto-promotion, so this is a known limitation, but the hint text doesn't coordinate with the deferral mechanism, so the model may attempt a direct call and fail.
- **Suggested fix:** either build hints from the bound `final_tools` (hints go quiet when tools are deferred), or have the hint cross-reference promotion; at minimum note the interaction in the "Routing Hints" docs. The same applies to `client.py` and `subagents/executor.py`.
- **Test:** cover the `tool_search.enabled=true` + routed MCP tool combination.

backend/app/gateway/routers/mcp.py
```diff
@@ -53,6 +53,10 @@ class McpServerConfigResponse(BaseModel):
     headers: dict[str, str] = Field(default_factory=dict, description="HTTP headers to send (for sse or http type)")
     oauth: McpOAuthConfigResponse | None = Field(default=None, description="OAuth configuration for MCP HTTP/SSE servers")
     description: str = Field(default="", description="Human-readable description of what this MCP server provides")
+    routing: McpRoutingConfig = Field(default_factory=McpRoutingConfig, description="Soft routing hints for tools from this MCP server")
+    tools: dict[str, McpToolOverride] = Field(default_factory=dict, description="Per-original-tool MCP configuration overrides")
+    tool_call_timeout: float | None = Field(default=None, description="Timeout in seconds for individual stdio MCP tool calls")
+    model_config = ConfigDict(extra="allow")
```
willem-bd, line 60: [P2] `extra="allow"` surfaces operator-authored extra fields unmasked in GET

- **Problem:** with `extra="allow"`, GET `/mcp/config` now serializes unknown server fields (e.g. a custom `api_key`, `cwd`) via `McpServerConfigResponse(**server.model_dump())`, but `_mask_server_config` only masks `env`/`headers`/`oauth`. Before this change those fields were dropped by the response model; now they're returned in plaintext.
- **Evidence:** the new `test_update_mcp_configuration_preserves_server_extra_fields` confirms `cwd`/`customFlag` round-trip — which also confirms they appear on read. GET is admin-gated, so impact is bounded, but the masking layer exists for defense-in-depth.
- **Suggested fix:** either extend `_mask_server_config` to scrub sensitive-looking `model_extra` keys, or document that extra fields are returned verbatim so operators know not to store secrets there.

backend/tests/test_mcp_routing_metadata.py
```diff
@@ -0,0 +1,123 @@
...
+            "keywords": ["订单"],
+            "auto_promote_top_k": None,
+        },
+    )
+
+    assert get_mcp_routing(tool) is None
+
+
+@pytest.mark.asyncio
+async def test_get_mcp_tools_tags_effective_routing_metadata():
```
willem-bd: [P2] Missing coverage for the stdio wrapper's metadata preservation

- **Problem:** this test only exercises the HTTP transport branch (unwrapped tool). The stdio branch — which the docs' postgres example uses — routes tools through `_make_session_pool_tool`, which reconstructs the `StructuredTool`. Nothing guards against a future refactor of that wrapper dropping `metadata` and silently disabling hints for every stdio server.
- **Evidence:** `_make_session_pool_tool` currently copies `metadata=tool.metadata`, but that's an implementation detail with no regression test.
- **Suggested fix:** add a `transport: stdio` variant asserting `get_mcp_routing(wrapped_tool)` is non-`None` after `get_mcp_tools()`.

backend/packages/harness/deerflow/tools/builtins/tool_search.py
```diff
@@ -219,3 +220,32 @@ def get_deferred_tools_prompt_section(*, deferred_names: frozenset[str] = frozen
...
+            continue
+        hints.append((int(routing.get("priority", 0)), candidate.name, [str(keyword) for keyword in keywords]))
+
+    if not hints:
+        return ""
+
+    lines = ["<mcp_routing_hints>"]
+    for priority, tool_name, keywords in sorted(hints, key=lambda item: (-item[0], item[1])):
+        lines.append(f"When the user's request involves {_format_keyword_list(keywords)}:")
+        lines.append(f"  prefer the `{tool_name}` tool (priority {priority}).")
```
willem-bd: [P2 · nit, non-blocking] Priority number is rendered into the model-facing prompt

- **Problem:** `prefer the `{tool_name}` tool (priority {priority})` exposes an internal ordering key to the model. `priority` is only meant to order hints (enforced by the sort above); showing it adds prompt noise and could lead the model to interpret it literally.
- **Suggested fix:** drop `(priority {priority})` from the rendered text and keep `priority` as a sort key only.

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -1,14 +1,58 @@
...
+    )
+    priority: int = Field(
+        default=0,
+        description="Ordering key for routing hints. Higher values are rendered first.",
+    )
+    keywords: list[str] = Field(
+        default_factory=list,
+        description="Operator-authored keywords that describe when this MCP tool should be preferred.",
+    )
+    auto_promote_top_k: int | None = Field(
```
willem-bd: [P2 · nit, non-blocking] `auto_promote_top_k` is committed schema with no behavior in PR1

- **Problem:** this field is accepted/validated/serialized (and asserted in tests) but nothing reads it — auto-promotion is explicitly out of scope for PR1. Because `McpRoutingConfig` is `extra="forbid"`, this is now a public, hard-to-remove schema contract.
- **Suggested fix:** drop the field until auto-promotion actually lands so the config surface ships with real behavior; or, if kept as a forward-compat placeholder, add a `# no-op until PR2` comment so reviewers don't hunt for the implementation.

willem-bd (commented):
Thanks @18062706139fcz. I found a few issues that should be addressed before this is ready — most relate to config round-trip preservation and the `tool_search` interaction. The soft-hint design (single render path, no schema mutation) is a good base; these are fixes on top of it.

WillemJiang (approved):

PR #4009, opened 2026-07-09
fix(subagents): inherit summarization middleware and harden step capture (#3875 Phase 3)

hata33:
Part of #3875 (Phase 3 — the "proper long-term fix" Phase 2 explicitly deferred to).

## Why

Phases 1 + 2 added loop detection, a token budget, and an additive `stop_reason` channel. But the underlying cause of the 4.4M-token burn remained: **subagents had no context compaction**. A deep-research subagent (`max_turns` up to 150) re-sends a growing context every turn, so even with Phase 2's budget capping the *tail*, the run can legitimately accumulate >1M cumulative input before `max_turns`/`timeout`/`token_budget` engage. Phase 2 flagged this: *"Subagent summarization is the proper long-term fix and is out of scope here."* This is that fix.

## What changed

- **Subagent chain inherits `DeerFlowSummarizationMiddleware`** — gated on the **same `app_config.summarization.enabled` switch the lead reads** (per [@ggnnggez's Phase 3 scoping](https://github.com/bytedance/deer-flow/issues/3875) — *"gate the subagent chain on the same `summarization.enabled`"*). One config covers both chains; no separate `subagents.summarization` field. Trigger/keep/model/prompt all come from the shared `summarization` config, so the two chains cannot drift. The shared `create_summarization_middleware` factory returns `None` when disabled, so this is a pure no-op when summarization is off.
- **`capture_new_step_messages` hardened against history contraction** (validation point (a)) — summarization rewrites the messages channel via `RemoveMessage(id=REMOVE_ALL_MESSAGES)`, which shrinks `len(messages)` below the step-capture cursor mid-run. Without a reset, **every step appended after the compaction point was silently dropped** until the list length overtook the stale cursor. The cursor now resets to the new tail; `capture_step_message`'s id/content dedup prevents re-emitting pre-compaction steps. This is the #3845 interaction the maintainer called out.
- **Tightened `subagents.token_budget` default `max_tokens` 2M → 1M** (validation point (b)) — now that summarization compacts the running context before it reaches pathological size, a tighter ceiling still covers legitimate deep research while catching degenerate runs earlier.

The summarization middleware does **not** implement `consume_stop_reason`, so it does not interfere with the Phase 2 guard-cap stop-reason channel — no contract change, no new enum value.

## What I did *not* do

- **Did not delete `_create_summarization_middleware` in `lead_agent/agent.py`.** The maintainer noted it "currently lives in `lead_agent/agent.py`" and should be "moved somewhere shared." The underlying factory `create_summarization_middleware` **already lives in the shared `summarization_middleware.py`** and the lead wrapper is a one-line passthrough to it — so the sharing goal is already met. The wrapper remains a stable `monkeypatch` target for 6 existing tests in `test_lead_agent_model_resolution.py` / `test_durable_context_middleware.py`; deleting it would break those for no behavioral gain. Happy to remove it if a reviewer prefers, with the test patches migrated.

## ⚠️ Tunable default — flagging for review

The **1,000,000-token** ceiling (tightened from Phase 2's 2M) assumes summarization keeps a deep-research run's running context well under 1M cumulative input. If real-world deep-research runs hit the new ceiling prematurely, this should be raised again. Flagged tunable.

## Tests

- `test_subagent_step_events.py` — new `test_capture_new_step_messages_handles_history_contraction` pins the contraction-reset (pre-compaction steps survive, summary captured, post-compaction steps **not** dropped).
- `test_subagent_executor.py` — new `test_aexecute_step_capture_survives_history_contraction` exercises the same path through the executor's real `astream` loop with a scripted `values`-mode sequence that snaps the channel down mid-run.
- `test_tool_error_handling_middleware.py` — new attach/omit pair mirrors the Phase 1 loop-detection wiring tests; the shared factory is patched at its source module so no real chat model is needed.
- `test_subagent_timeout_config.py` — pins the new 1M default.
- Regression: `test_subagent_status_contract`, `test_token_budget_middleware`, `test_loop_detection_middleware`, `test_durable_context_middleware`, `test_lead_agent_model_resolution`, `test_summarization_middleware` — **all green (327 across the sweep)**.
- `ruff format` / `ruff check` clean.

## AI assistance

Authored with ZCode. I read every line of the diff and wrote all tests by hand against the existing fake-model patterns from Phase 1/2.


willem-bd (commented):
Reviewed Phase 3 (#3875) against the head commit. The core mechanism is sound: the shared factory genuinely can't drift from the lead, and the `capture_new_step_messages` contraction-reset fixes a real bug (post-compaction super-steps appending >1 message dropped all but the last). Verified per-run middleware isolation in `executor._aexecute`.

5 inline comments below. The two I'd want resolved before merge:
1. The 1M ceiling default isn't coupled to `summarization.enabled` (which defaults to `False`) — premature capping risk for summarization-off deployments.
2. Inheriting the shared factory silently enables subagent→parent-thread memory flushes via `memory_flush_hook` — unstated side effect needing sign-off.

The other three are minor (comment accuracy, test fidelity, a latent invariant worth documenting). Happy to dig into any of them.

backend/packages/harness/deerflow/config/subagents_config.py
```diff
@@ -10,17 +10,19 @@
...
+    exist). ``max_tokens`` was 2M in Phase 2 — a deliberately loose ceiling
+    chosen because, without summarization, legitimate deep-research runs
+    (``max_turns=150``) could genuinely accumulate >1M cumulative input. Now
+    that Phase 3 lands ``DeerFlowSummarizationMiddleware`` on the subagent
+    chain, the running context is compacted before it reaches pathological
+    size, so a tighter 1M ceiling still covers legitimate deep research while
+    catching degenerate runs earlier. Flagged tunable in the PR description.
     """
-    return TokenBudgetConfig(enabled=True, max_tokens=2_000_000, warn_threshold=0.7)
+    return TokenBudgetConfig(enabled=True, max_tokens=1_000_000, warn_threshold=0.7)
```
willem-bd: ⚠️ **Substantive** — the 1M ceiling isn't coupled to `summarization.enabled`, but its justification ("compaction keeps the running context below pathological size") only holds when summarization is ON.

- `config.example.yaml` ships `summarization.enabled: true`, so standard deployments are fine.
- But `SummarizationConfig.enabled` defaults to `False`. A deployment with no/missing `summarization` block — or one that disables it — gets the tighter 1M ceiling with NO compaction.
- Phase 2's own docstring noted legitimate deep-research runs (`max_turns=150`, no summarization) "can genuinely accumulate >1M cumulative input." `max_tokens` bounds cumulative **total** tokens (`token_budget_middleware.py:228`), so ~1M total ≈ 1M input — i.e. summarization-off deep-research runs that Phase 2 acknowledged exceed 1M now get prematurely capped.

The PR body flags 1M as tunable, but frames that as "if real-world runs hit the ceiling prematurely" — which is exactly the summarization-off case, and it's not called out.

**Suggestion:** couple the default to the switch (1M when `summarization.enabled`, else 2M), or state in the docstring that 1M assumes summarization is enabled. As written, turning summarization off silently loses the headroom Phase 2 deliberately gave.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -354,4 +354,27 @@ def build_subagent_runtime_middlewares(
...
+    # ``summarization`` config the lead reads, so the two chains cannot drift.
+    # The middleware rewrites history via ``RemoveMessage(id=REMOVE_ALL_MESSAGES)``,
+    # which shrinks the messages channel mid-run;
+    # ``capture_new_step_messages`` must tolerate that contraction (see
+    # ``step_events.py``) or it drops steps captured after the compaction
+    # point. It does not implement ``consume_stop_reason``, so it does not
+    # interfere with the Phase 2 guard-cap stop-reason channel.
+    from deerflow.agents.middlewares.summarization_middleware import create_summarization_middleware
+
+    summarization_middleware = create_summarization_middleware(app_config=app_config)
```
willem-bd: **Behavioral side effect to confirm:** the shared `create_summarization_middleware` factory attaches `memory_flush_hook` whenever `app_config.memory.enabled`. That hook (`agents/memory/summarization_hook.py`) flushes pre-compaction messages into the durable memory queue, keyed by `thread_id` + `agent_name`.

Subagents share the **parent** `thread_id` in context (executor sets `context["thread_id"] = self.thread_id`), so a subagent compaction now writes its internal turns (the "Task" human message + intermediate AI/tool turns) into the **parent thread's** durable memory — including the correction/reinforcement detection path.

This isn't mentioned in the PR body, whose stated goal is purely context compaction. It may be desirable (research persists) or undesirable (subagent-internal exchanges pollute the user's memory). Needs a maintainer sign-off. If undesired, the factory needs a subagent-aware flag to skip the hook.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -354,4 +354,27 @@ def build_subagent_runtime_middlewares(
 
         middlewares.append(SafetyFinishReasonMiddleware.from_config(safety_config))
 
+    # DeerFlowSummarizationMiddleware — subagents inherit none of the lead's
```
willem-bd, line 382: **Minor / comment accuracy:** the block says it "mirror[s] the lead chain," but the placement differs. In the lead chain (`lead_agent/agent.py:280`) summarization is appended **before** LoopDetection (329) / TokenBudget (336) / SafetyFinishReason (349); here it's appended **last**, after the guard trio. (The relative order of the guard trio itself is preserved.)

I think this is benign — compaction runs in `before_model` regardless of relative position, and the guard middlewares account in `after_model` — but the comment overstates fidelity. Either align the placement (append before the guards) or soften the comment to "mirrors the lead's gating, placed after the guards."

backend/tests/test_subagent_step_events.py
```diff
@@ -243,6 +243,49 @@ def test_capture_new_step_messages_is_noop_on_values_reyield():
...
+        AIMessage(content="done turn", id="ai-2"),
+    ]
+    processed = capture_new_step_messages(before, captured, seen, 0)
+    assert processed == 4
+    assert [c["id"] for c in captured] == ["ai-1", "tool-1", "ai-2"]
+
+    # Compaction rewrites the channel to a single summary AIMessage — length
+    # drops from 4 to 1, below the cursor. The summary is a NEW step (new id),
+    # so it must be captured, and the cursor must reset to the new tail.
+    compacted = [AIMessage(content="summary of prior turns", id="ai-summary")]
```
willem-bd: **Test fidelity (minor):** both new tests model the compaction summary as an `AIMessage` inserted into the messages channel and assert it's captured. But the real `_maybe_summarize` (`summarization_middleware.py:333-343`) puts the summary into a **separate `summary_text` state key** — the messages channel after compaction holds only `[*preserved_messages]` (the retained recent tail), with no summary message. And `capture_step_message` only captures `AIMessage`/`ToolMessage`, so a real summary (`HumanMessage(name="summary")`) would be skipped anyway.

The fix itself is correctly validated — the **post-compaction growth** assertions (`tool-2`, `ai-3` not dropped) are the real regression coverage. But the "ai-summary is captured" assertion tests a scenario that doesn't occur in production and could mislead a future reader. More faithful: model the compaction chunk as the already-seen preserved tail rather than a synthetic summary AIMessage.

backend/packages/harness/deerflow/subagents/step_events.py
```diff
@@ -95,8 +95,18 @@ def capture_new_step_messages(
...
+    When the history *contracted* (``total < processed_count``) — which happens
+    when ``DeerFlowSummarizationMiddleware`` rewrites the channel via
+    ``RemoveMessage(id=REMOVE_ALL_MESSAGES)`` (#3875 Phase 3) — reset the cursor
+    to the new tail and let ``capture_step_message``'s id/content dedup prevent
+    re-emitting steps captured before the compaction. Without this reset, every
+    step appended after the compaction point is dropped until ``total`` overtakes
+    the stale cursor.
     """
     total = len(messages)
+    if total < processed_count:
```
willem-bd, line 117: **Minor / latent:** after `processed_count = total`, the no-growth branch only re-examines `messages[-1]`. A genuinely new `AIMessage`/`ToolMessage` inserted at an index *below* the reset cursor in a compacted list would be missed.

Not reachable with the real middleware today (compaction retains only already-seen tail messages; new content goes to the out-of-channel `summary_text`), so it's latent — but the correctness here depends on that assumption. A one-line comment noting "compaction never inserts a new capturable message below the cursor" would make the invariant explicit.

hata33:
@willem-bd Thanks for the thorough review — all 5 points are addressed in the force-push (`61520a03`). Core mechanism unchanged; the two substantive points are fixed.

## 1. 1M ceiling not coupled to `summarization.enabled` — **fixed**
You're right, this was a real regression risk. The default is now **coupled to the switch**: **1M when `summarization.enabled=True`, 2M when False** (preserves Phase 2's deliberate headroom for summarization-off deep-research runs that can exceed 1M).

Implementation detail worth flagging: the model-level `default_factory` can't read the sibling `summarization.enabled` field, so `default_subagent_token_budget(summarization_enabled=False)` falls back to 2M, and `build_subagent_runtime_middlewares` recomputes via `get_token_budget_for(agent_name, summarization_enabled=app_config.summarization.enabled)`. A **user-set** budget (global or per-agent) always wins regardless of the switch — a deployment that pinned a value is never silently changed by flipping summarization on/off. The `_token_budget_is_default` flag (set from `model_fields_set` in `__init__`, preserved across the `load_subagents_config_from_dict` reload path) gates the recompute.

## 2. `memory_flush_hook` polluting parent thread memory — **fixed**
Added `skip_memory_flush: bool = False` to `create_summarization_middleware`; the subagent path passes `True`. The lead path keeps the hook (research should persist). This stops a subagent's internal turns (the "Task" human message + intermediate AI/tool turns) from being flushed into the parent thread's durable memory queue. Two new factory tests pin both paths (`test_factory_attaches_memory_flush_hook_by_default`, `test_factory_skip_memory_flush_omits_hook`).

## 3. "Mirror the lead chain" comment overstates placement — **fixed**
Softened to "Gated on the SAME switch the lead reads" and added an explicit note that placement differs (lead appends summarization before the guard trio; subagent appends it after), with the rationale that it's benign (compaction in `before_model`, guards in `after_model`).

## 4. Test modeled summary as AIMessage — **fixed**
Rewrote both tests (`test_subagent_step_events.py` + the executor integration test) to model the compaction chunk as the **preserved recent tail** (already-seen messages, deduped), not a synthetic summary AIMessage. This matches the real `_maybe_summarize` — summary goes into the separate `summary_text` state key; the messages channel post-compaction holds only retained tail messages. The real regression coverage (post-compaction `tool-2`/`ai-3` not dropped) is unchanged and now asserts against faithful data.

## 5. Latent invariant below the reset cursor — **fixed**
Added an explicit `INVARIANT` block to `capture_new_step_messages`'s docstring: after the reset, the no-growth branch only re-examines `messages[-1]`, so a new capturable message inserted below the reset cursor would be missed — but this is unreachable today (compaction never inserts a new capturable message below the cursor; summary lives out-of-channel). Flagged that a future middleware violating this would need a full re-scan.

## Validation
- Targeted: `test_subagent_step_events`, `test_summarization_middleware`, `test_subagent_timeout_config`, `test_subagent_executor`, `test_tool_error_handling_middleware` — all green.
- Regression sweep incl. `test_app_config_reload` (the reload round-trip for `_token_budget_is_default`), `test_lead_agent_model_resolution`, `test_durable_context_middleware`, `test_token_budget_middleware`, `test_loop_detection_middleware`, `test_subagent_status_contract` — **346 total, all green**.
- `ruff format` / `ruff check` clean.

backend/packages/harness/deerflow/config/subagents_config.py
```diff
@@ -165,18 +195,34 @@ def get_skills_for(self, agent_name: str) -> list[str] | None:
...
+        any explicitly configured ``token_budget`` (global or per-agent)
+        wins regardless, so a deployment that pinned a value is never
+        silently changed by flipping the summarization switch.
         """
         override = self.agents.get(agent_name)
         if override is not None and override.token_budget is not None:
             return override.token_budget
+        # Only recompute when the caller is using the default (no explicit
+        # global token_budget was set). A user-set global is respected as-is.
+        if self._token_budget_is_default:
```
willem-bd, line 224: This cached `_token_budget_is_default` flag is redundant with `model_fields_set` and introduces a staleness hazard: `model_copy(update={"token_budget": ...})` copies `__pydantic_private__` verbatim without re-running `__init__`, so the flag could silently disagree with the new value.

You can read it live here and drop both the private field and the `__init__` override:

```python
if "token_budget" not in self.model_fields_set:
    return default_subagent_token_budget(summarization_enabled=summarization_enabled)
return self.token_budget
```

`model_fields_set` is always accurate to the last construction/validation, so this also stays correct after the `load_subagents_config_from_dict` drop-key path. The existing coupling tests assert behavior, not mechanism, so they pass unchanged.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -346,7 +346,17 @@ def build_subagent_runtime_middlewares(
...
+    # Default-ceiling coupling (#3875 Phase 3 review): the default ``max_tokens``
+    # is re-coupled to ``summarization.enabled`` — 1M when compaction is on, 2M
+    # when off. This ONLY applies to the default; a user-set budget (global or
+    # per-agent) always wins, so a deployment that pinned a value is never
+    # silently changed by flipping the summarization switch.
+    summarization_enabled = app_config.summarization.enabled
+    if agent_name is not None:
+        token_budget_config = app_config.subagents.get_token_budget_for(agent_name, summarization_enabled=summarization_enabled)
+    else:
+        token_budget_config = app_config.subagents.token_budget
```
willem-bd, line 366: The `else` (`agent_name is None`) branch bypasses the summarization coupling — it returns `app_config.subagents.token_budget`, which is the model-level default (always 2M, since `default_factory` can't read the summarization switch). The only caller passes `agent_name=self.config.name` (executor.py), so this is likely unreachable in practice — but if it ever is, a summarization-on deployment silently keeps the loose 2M ceiling, partially defeating the tightening.

Consider routing this through `default_subagent_token_budget(summarization_enabled=...)` for consistency, or adding a short note that `agent_name` is always provided at the single call site.

backend/packages/harness/deerflow/config/subagents_config.py
```diff
@@ -191,6 +237,18 @@ def get_subagents_app_config() -> SubagentsAppConfig:
...
+    # ``config.subagents.model_dump()``, which serializes a default
+    # ``token_budget`` into the dict. Re-constructing from that dict would make
+    # ``model_fields_set`` contain ``token_budget`` and flip
+    # ``_token_budget_is_default`` to False — breaking the
+    # summarization-coupled recompute in ``get_token_budget_for`` (#3875 Phase
+    # 3). Drop the key when its value still equals the no-compaction default so
+    # the default_factory fires on reconstruction and the "user did not set
+    # this" signal is preserved.
+    tb = config_dict.get("token_budget")
+    if tb is not None and tb == default_subagent_token_budget(summarization_enabled=False).model_dump():
```
willem-bd, line 250: Minor edge case worth a one-line comment: if a deployment explicitly pins `token_budget` to a value that *happens to equal* the 2M no-compaction default, this equality check treats it as "unset" after a reload and re-couples it to the summarization switch (so it would flip to 1M when summarization turns on). The user is effectively indifferent, but a comment would preempt a future "why did my pinned budget change?" question.

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -362,4 +372,45 @@ def build_subagent_runtime_middlewares(
...
+    #
+    # The middleware rewrites history via ``RemoveMessage(id=REMOVE_ALL_MESSAGES)``,
+    # which shrinks the messages channel mid-run;
+    # ``capture_new_step_messages`` must tolerate that contraction (see
+    # ``step_events.py``) or it drops steps captured after the compaction
+    # point. It does not implement ``consume_stop_reason``, so it does not
+    # interfere with the Phase 2 guard-cap stop-reason channel.
+    from deerflow.agents.middlewares.summarization_middleware import create_summarization_middleware
+
+    summarization_middleware = create_summarization_middleware(
```
willem-bd, line 416: The relative-order asymmetry vs. the lead chain is documented in the comment block above and is benign today (summarization has only `before_model`; the guards account in `after_model`). But there's no test pinning this placement, so a future refactor that mirrors the lead's "summarization-before-guards" order would change compaction timing relative to the guards with nothing catching it. A small assertion in `test_tool_error_handling_middleware.py` (e.g. `SummarizationMiddleware` index > `TokenBudgetMiddleware`/`LoopDetectionMiddleware` index, or explicitly equal to the lead's relative order if you'd rather mirror it) would lock the intended contract.

willem-bd (commented):
## Overview

Phase 3 of #3875 — gives subagents the same context-compaction the lead agent already has, so a deep-research subagent (`max_turns` up to 150) stops re-sending a monotonically growing context every turn. Four pieces:

1. Wires `DeerFlowSummarizationMiddleware` into `build_subagent_runtime_middlewares`, gated on the **same** `summarization.enabled` switch the lead reads (one config, no drift), with a new `skip_memory_flush` flag so a subagent's internal turns aren't flushed into the **parent** thread's durable memory.
2. Hardens `capture_new_step_messages` against history contraction (compaction shrinks `len(messages)` below the cursor → post-compaction steps were silently dropped).
3. Couples the default `token_budget.max_tokens` to the summarization switch (1M on / 2M off); user-set budgets always win.
4. Focused unit tests for each.

## Correctness — verified against the surrounding code

- `DeerFlowSummarizationMiddleware` implements **only** `before_model` — no `after_model`, no `consume_stop_reason`, no `awrap_model_call`. So appending it after the guard trio (vs. before in the lead chain) cannot disturb the Phase 2 stop-reason channel or the guards' `after_model` accounting. The ordering asymmetry flagged in the comment is real but harmless.
- The contraction fix is correct: after reset the cursor snaps to the new tail, `messages[-1]` is re-examined (deduped), and `max(processed_count, total)` returns the tail so subsequent growth is captured. The `INVARIANT` note is the right call.
- The reload round-trip is real — `app_config.py` literally calls `load_subagents_config_from_dict(config.subagents.model_dump())` — so the drop-key logic is necessary, not speculative.
- `skip_memory_flush` reasoning is sound: a subagent's result reaches the parent as a delegation ToolMessage the lead's memory path already captures, so flushing internal turns would be pollution.

## Gaps not tied to a specific line

- **No integration test drives the real compaction during a subagent run.** The unit tests verify *wiring* (factory gating, builder attach/omit, config coupling) and the contraction *logic* via scripted `values` chunks — but nothing exercises the real `DeerFlowSummarizationMiddleware` through the executor to confirm compaction fires mid-run and the run still produces a usable final answer afterward. Given `risk:high` + `needs-validation`, that's the gap I'd most want filled.
- **`AGENTS.md` is now stale.** The Subagent System → "Guardrail caps" paragraph states the token-budget default is "default 2,000,000 tokens" (now 1M-on / 2M-off coupled), and the Middleware Chain description of the subagent chain should mention `DeerFlowSummarizationMiddleware` + the new contraction behavior in Step capture. Per the repo's docs-sync convention this should ship in the same change set.
- **The 1M ceiling is the highest-risk behavioral change** (and the one you flag). Summarization-off deployments keep 2M (no regression), but summarization-on deployments drop to 1M — if compaction doesn't keep running context comfortably under 1M cumulative input, legitimate deep-research runs will hit `token_capped`. Worth validating against a real deep-research run before merge.

Overall: strong engineering quality and the core fix is correct. Findings below are a simplification plus a few consistency/doc gaps — nothing blocking.

hata33:
@willem-bd Thanks for the second pass and the "nothing blocking" call — the three gaps are addressed in `5135fd0c`.

## Gap 1 — real integration test driving live compaction
Added `test_subagent_summarization_fires_mid_run_and_produces_usable_result`: it runs the REAL `DeerFlowSummarizationMiddleware` (the exact instance the subagent chain gets via `create_summarization_middleware(skip_memory_flush=True)`) through a `create_agent` stream with 6 messages past a `trigger=messages:4` + `keep=messages:2` threshold, then asserts both:
- **(a) compaction fired**: the `before_model` hook emitted a `summary_text` + led with a `RemoveMessage`;
- **(b) the run survived**: a usable final `AIMessage` ("final answer after compaction") was produced after compaction.

I drive the factory directly rather than the full `build_subagent_runtime_middlewares` chain because the chain's sandbox/thread-data stubs aren't `AgentMiddleware`-compatible for a live run — but it's the same middleware instance the builder appends, and the wiring test already proves it lands on the chain. One subtlety worth recording: the trigger also needs `keep` small enough that there's something to compact (a 6-message run with default `keep=20` has nothing to trim), so the test sets `keep=messages:2`.

## Gap 2 — AGENTS.md now in sync
- **Subagent System → Token axis**: updated the default from "2,000,000 tokens" to the coupled "**1,000,000 when `summarization.enabled`, 2,000,000 when off; a user-set budget always wins** (#3875 Phase 3)".
- Added a new **"Context compaction (#3875 Phase 3)"** paragraph covering: same-switch gating, `skip_memory_flush=True` rationale (parent thread_id pollution), placement asymmetry vs the lead chain and why it's harmless (`before_model`-only, no `consume_stop_reason`), and the `RemoveMessage`→contraction interaction.
- Appended the **history-contraction** behavior to the Step capture paragraph so `capture_new_step_messages`'s contraction-reset is documented alongside the append-only baseline.

## Gap 3 — 1M ceiling real-world validation
Being upfront: I can't run a real deep-research subagent against a live LLM from here, so I have **not** empirically confirmed that a summarization-on deep-research run keeps cumulative input comfortably under 1M. What I have:
- The unit tests prove the coupling logic (1M-on / 2M-off / user-set-wins) and the guard fires correctly.
- The integration test above proves compaction actually fires and the run survives.

But the open question you flagged stands: **if real-world summarization-on deep research genuinely needs >1M cumulative input even with compaction, it will hit `token_capped` and the user will need to raise `subagents.token_budget.max_tokens`.** I'd rather we validate that against a real run before merge — if you (or the CI) have a deep-research scenario handy, that's the most valuable check. If it turns out 1M is too tight, raising the on-ceiling (or reverting to a flat 2M with just a doc note) is a one-line follow-up.

## Validation
- New integration test + all prior tests green: **349 total** across the Phase-3 sweep (`test_tool_error_handling_middleware`, `test_subagent_step_events`, `test_subagent_executor`, `test_subagent_timeout_config`, `test_summarization_middleware`, + the regression batch incl. `test_app_config_reload`).
- `ruff format` / `ruff check` clean.

WillemJiang (approved):

PR #4012, opened 2026-07-09
fix(gateway): live-tail malformed redis reconnect ids

### LINKED ISSUE #4011: Redis stream bridge: `Last-Event-ID: -1` on reconnect triggers full-buffer replay, hanging the UI
Follow-up to #3191 (Redis stream bridge) and #3932 (producer-death hang). A third reconnect-path gap: when the reconnecting client sends an invalid/missing `Last-Event-ID`, the bridge replays the **entire** retained buffer, hanging the UI.

## Summary

When `useStream` (`@langchain/langgraph-sdk`) reconnects after a page refresh, it calls `joinStream(runId)` **without** a tracked `lastEventId`; the SDK defaults it to `"-1"` (`react/stream.lgp.js:326`), so the reconnect sends `Last-Event-ID: -1`. The Redis bridge passes `-1` straight to `XREAD`, Redis rejects it (`Invalid stream ID specified as stream command argument`), and the `except ResponseError` fallback resets to `0-0` and replays **every retained event from the start**. For a long run (~256 retained events at `MAXLEN`), this re-streams the whole response on every reconnect, hanging the chat UI until the run completes.

## Reproduction

1. Deploy with `stream_bridge.type: redis`, `gateway.replicas: 2` (Helm chart), no nginx stickiness.
2. Send a prompt that streams ~30-60s (e.g. "write a detailed 2000-word essay about the history of AI").
3. Refresh the browser mid-stream.
4. UI hangs; gateway logs show the warning + `ResponseError` below.

## Expected / Actual

**Expected:** reconnect resumes from the live tail (new events only) — the frontend already loads thread state from the checkpoint via `fetchStateHistory`, so replaying the whole stream is redundant.

**Actual:** full-buffer replay from `0-0` on every reconnect.

## Evidence

```
07:12:53 - deerflow.runtime.stream_bridge.redis - WARNING - Redis rejected Last-Event-ID '-1' for stream bridge; replaying from earliest retained event
Traceback (most recent call last):
  File ".../stream_bridge/redis.py", line 179, in subscribe
    response = await self._redis.xread({key: stream_id}, count=_XREAD_COUNT, block=block_ms)
  ...
redis.exceptions.ResponseError: Invalid stream ID specified as stream command argument
```

Redis during the run: `deerflow:stream_bridge:<run_id>` type=stream XLEN=256 (MAXLEN-capped rolling buffer). The run itself survives the refresh (`on_disconnect: continue`) and completes normally — only the reconnect path is broken.

## Root cause (two sides)

1. **SDK** (`langchain-ai/langgraphjs`, `@langchain/langgraph-sdk` 1.6.0): `useStream` reconnect calls `joinStream(reconnectKey.runId)` with no `lastEventId` (`stream.lgp.js:388`), defaulting it to `"-1"` (`:326`). The SDK doesn't track the last received SSE event id, so it can't send a real resumption offset.
2. **deer-flow bridge** (`backend/packages/harness/deerflow/runtime/stream_bridge/redis.py:165-196`): `subscribe` does `stream_id = last_event_id or "0-0"` (`:173`) — an invalid id like `-1` is truthy, round-trips through a `ResponseError`, then falls back to `0-0` (full replay) instead of the live tail.

## Proposed fix

In `redis.py:subscribe`, validate `last_event_id` *before* `XREAD`. If it isn't a valid Redis stream ID (`^\d+-\d+$`), start from the live tail (`$`), not `0-0`:

```python
stream_id = last_event_id if _is_valid_stream_id(last_event_id) else "$"
```

The frontend short-circuits terminal runs (`shouldSkipReconnect` in `api-client.ts`), so live-tail only applies to *active* runs where it's correct — new events arrive as the run continues, and `publish_end` still delivers the `__end__` marker. This eliminates both the `ResponseError` round-trip and the full-buffer replay, and makes reconnect robust regardless of what the SDK sends.

## Environment

- deer-flow with Redis stream bridge (`stream_bridge.type: redis`), Helm chart, `gateway.replicas: 2`
- `@langchain/langgraph-sdk` 1.6.0
- redis-py 7.4.0

## Related

- #3191 — Redis stream bridge (cross-worker SSE delivery + Last-Event-ID replay)
- #3932 — Redis stream bridge: producer-death hang (closed, milestone 2.1.0)
- #3260 — Gateway mode streaming reconnect not multi-worker safe (original reconnect issue)
- #3948 — Multi-worker run-control baseline (closed)
- SDK root cause: `langchain-ai/langgraphjs` `useStream` reconnect doesn't track `lastEventId`

Huixin615:
Fixes #4011

## Why

When the Redis stream bridge is enabled, a reconnecting LangGraph SDK client can send `Last-Event-ID: -1` after a page refresh. The bridge previously passed that value directly to Redis `XREAD`; Redis rejected it, and the error fallback reset the stream id to `0-0`.

That fallback replayed the retained stream buffer from the beginning, so long active runs could re-stream the whole retained response on every reconnect and hang the chat UI.

## What changed

Malformed Redis stream reconnect ids now live-tail new events instead of replaying retained history.

- Valid Redis stream ids (`<ms>-<seq>`) still resume after the provided `Last-Event-ID`.
- Missing `Last-Event-ID` still starts from `0-0`, preserving initial subscribe behavior.
- Malformed values such as `-1` now start from Redis `$`, so active reconnects wait for new events only.
- Redis `ResponseError` no longer falls back to `0-0`, avoiding accidental full-buffer replay.
- Redis stream bridge docs now mention the malformed reconnect-id behavior.

## Surface area

- [x] **Backend API** — SSE reconnect behavior through the Redis stream bridge
- [x] **Default behavior change** — malformed Redis reconnect ids now live-tail instead of replaying retained history

## Screenshots / Recording

N/A. Backend Redis stream reconnect behavior only.

## Bug fix verification

- `backend/tests/test_stream_bridge.py::test_redis_invalid_last_event_id_tails_live_events`
- `backend/tests/test_stream_bridge.py::test_redis_integration_invalid_last_event_id_tails_live_events`

- Not rerun on `main` in this checkout. The previous implementation passed malformed ids to Redis, handled the resulting `ResponseError`, and reset to `0-0`; the updated tests assert that malformed ids skip retained `metadata` and receive only newly published events.

If a red test wasn't cheap to write, explain why and what you did instead.

- A focused fake Redis regression covers the SDK-style `-1` reconnect path without requiring Redis. The real Redis integration test is also updated for environments with `DEER_FLOW_TEST_REDIS_URL`.

## Validation

```bash
cd backend && UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_stream_bridge.py -k redis
# 14 passed, 5 skipped, 20 deselected

cd backend && UV_CACHE_DIR=/tmp/uv-cache uv run ruff check packages/harness/deerflow/runtime/stream_bridge tests/test_stream_bridge.py
# All checks passed

cd backend && UV_CACHE_DIR=/tmp/uv-cache uv run ruff format --check packages/harness/deerflow/runtime/stream_bridge tests/test_stream_bridge.py
# 6 files already formatted

git diff --check
# passed

### CODE REVIEW

**Review by zhfeng — APPROVED** [review]
Thanks for this - clean fix, and the `xrevrange`-to-concrete-id approach is nicer than a naive `$` (uniform `XREAD` loop, clean empty-stream -> `0-0` fallback, smaller miss window). CI is green and the regression tests (fake + real-redis) look valid. Approving with two notes below.

**Non-blocking: malformed `Last-Event-ID` on an already-terminal run now hangs until TTL.**

The replaced test (`test_redis_integration_invalid_last_event_id_falls_back`) subscribed to a *completed* run with a malformed id and relied on the `0-0` fallback to replay the retained events + END. With the new resolver, a malformed id on a terminal run whose stream is still buffered (within `stream_ttl_seconds`, default 86400s) resolves to the **END marker's id** -> `XREAD` after it returns nothing -> the subscriber heartbeats until the TTL expires, because `_terminal_record_stream_missing` stays `False` while the stream still exists.

This isn't reachable from the frontend (`shouldSkipReconnect` skips terminal runs before `joinStream`, and the SDK only sends `-1` on reconnect), so it only affects a direct API client sending a bogus `Last-Event-ID` to a finished run. If that's acceptable, a one-line note in the resolver docstring would help; if not, a small guard works: in `_resolve_start_stream_id`, if the `xrevrange` last entry's `kind == _KIND_END`, return `"0-0"` (bounded replay delivers END and returns - no hang, since no new events pile on a terminal run; the hang only bites *active* runs).

**Trivial: dead `$` branch in `_FakeRedis.xread`.**

The `if last_id == "$": …` branch isn't reachable, since the resolver produces a concrete id via `xrevrange` and never `"$"`. Harmless, but can be dropped for clarity.

File: backend/packages/harness/deerflow/runtime/stream_bridge/redis.py [hunk 1]
```diff
@@ -162,6 +164,17 @@ async def stream_exists(self, run_id: str) -> bool:
         """Return whether Redis still has retained stream data for *run_id*."""
         return bool(await self._redis.exists(self._stream_key(run_id)))
 
+    async def _resolve_start_stream_id(self, key: str, last_event_id: str | None) -> str:
+        if last_event_id is None:
+            return "0-0"
+        if _REDIS_STREAM_ID_RE.fullmatch(last_event_id):
+            return last_event_id
+        entries = await self._redis.xrevrange(key, count=1)
```
willem-bd, line 172: The PR body says malformed ids "start from **Redis `$`**", but this resolves to the **last retained entry's id** via `XREVRANGE` instead. They're equivalent for live-tailing a non-empty stream, and the `XREVRANGE` approach is arguably the better choice here — it also covers the empty-stream case by returning `0-0`, whereas `$` on a non-existent stream has murkier `BLOCK` semantics.

One loose end: `_FakeRedis.xread` gained a `last_id == "$"` branch (test_stream_bridge.py:49) that is now dead — nothing in production or tests ever passes `$` (confirmed by grep). It reads like an incomplete refactor. Suggest aligning the PR-body wording to "live-tail from the last retained id" and either dropping the fake's `$` branch or leaving a comment that it's intentional future-proofing.

backend/packages/harness/deerflow/runtime/stream_bridge/redis.py
```diff
@@ -32,6 +33,7 @@
 
 _KIND_EVENT = "event"
 _KIND_END = "end"
+_REDIS_STREAM_ID_RE = re.compile(r"^\d+-\d+$")
```
willem-bd: Minor: this is stricter than Redis's own stream-id grammar. Redis accepts bare-ms partial ids like `1234` (== `1234-0`), but this regex rejects them, so such an id now falls into the malformed path and live-tails instead of resuming — silently dropping any retained events with id >= that value. No known client emits partial ids (the SDK echoes the server's full `<ms>-<seq>`), so the risk is theoretical, but `^\d+(-\d+)?$` would preserve resume semantics if you want belt-and-suspenders.

Nit: with `fullmatch`, the `^…$` anchors are redundant.

backend/packages/harness/deerflow/runtime/stream_bridge/redis.py
```diff
@@ -162,6 +164,17 @@ async def stream_exists(self, run_id: str) -> bool:
         """Return whether Redis still has retained stream data for *run_id*."""
         return bool(await self._redis.exists(self._stream_key(run_id)))
 
+    async def _resolve_start_stream_id(self, key: str, last_event_id: str | None) -> str:
+        if last_event_id is None:
+            return "0-0"
+        if _REDIS_STREAM_ID_RE.fullmatch(last_event_id):
+            return last_event_id
+        entries = await self._redis.xrevrange(key, count=1)
+        if not entries:
```
willem-bd, line 173: Coverage gap: this empty-stream fallback (`XREVRANGE -> [] -> "0-0"`) isn't exercised — both new tests pre-populate the stream, so the malformed path always takes the `XREVRANGE` branch. A small test subscribing with a malformed id to an empty/expired stream would cover this branch and confirm it live-tails future events.

willem-bd (commented):
Reviewed the malformed reconnect-id fix. The core change is sound: pre-validating the id before `XREAD`, live-tailing on a malformed token, and propagating `ResponseError` (now safe because validation happens up front) all check out, and the two regression tests are meaningful. None of the items below are blocking — they're accuracy / coverage / cleanup nits. Details inline.

WillemJiang (approved):

PR #4016, opened 2026-07-09
feat(provisioner): support ClusterIP services and scoped skills PVC mounts

kongxiangxin:
## Summary
- Add ClusterIP service support so in-cluster provisioner deployments can return stable DNS sandbox URLs.
- Add opt-in `SKILLS_PVC_SUBPATH_TEMPLATE` for user/thread-scoped skills mounts while preserving the default skills PVC root mount.
- Document the new provisioner deployment options and cover them with regression tests.

## Test Plan
- `uv run pytest tests/test_provisioner_pvc_volumes.py tests/test_provisioner_request_threading.py -q`

### DIFF
--- a/docker/provisioner/app.py
+++ b/docker/provisioner/app.py
@@ -63,6 +63,7 @@ THREADS_HOST_PATH = os.environ.get("THREADS_HOST_PATH", "/.deer-flow/threads")
 DEER_FLOW_HOST_BASE_DIR = os.environ.get("DEER_FLOW_HOST_BASE_DIR", "/.deer-flow")
 SKILLS_PVC_NAME = os.environ.get("SKILLS_PVC_NAME", "")
 USERDATA_PVC_NAME = os.environ.get("USERDATA_PVC_NAME", "")
+SKILLS_PVC_SUBPATH_TEMPLATE = os.environ.get("SKILLS_PVC_SUBPATH_TEMPLATE", "")
 SANDBOX_CONTAINER_PORT_RAW = os.environ.get("SANDBOX_CONTAINER_PORT", "8080")
 SANDBOX_SERVICE_TYPE = os.environ.get("SANDBOX_SERVICE_TYPE", "NodePort")
 try:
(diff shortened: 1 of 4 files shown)


docker/provisioner/app.py
```diff
@@ -521,15 +512,21 @@ def _build_service(sandbox_id: str) -> k8s_client.V1Service:
...
         svc = core_v1.read_namespaced_service(_svc_name(sandbox_id), K8S_NAMESPACE)
-        for port in svc.spec.ports or []:
-            if port.name == "http":
-                return port.node_port
-    except ApiException:
-        pass
+    except ApiException as exc:
+        if exc.status == 404:
+            return None
+        raise
```
willem-bd, line 541: **[P2] `_sandbox_access_url` now raises on transient non-404 K8s errors, but the create-sandbox callers relied on the old swallow-and-retry**

- **Problem:** the old `_get_node_port` caught *any* `ApiException` and returned `None`; the readiness polling loop (`for _ in range(20): ... time.sleep(0.5)`) and the idempotent fast path both leaned on that to tolerate transient failures. The new `_sandbox_access_url` only returns `None` on 404 — every other status (`429`, `503`, `500`, expired-token `401`) now `raise`s, and that raise is uncaught in both the fast path and the polling loop.
- **Evidence:** during the up-to-10s polling window after Pod+Service creation, a transient `429`/`503` on `read_namespaced_service` previously made the loop sleep and retry until it succeeded; now `create_sandbox` raises → unhandled → FastAPI 500, and the already-created Pod + Service are orphaned (the cleanup-on-failure block only wraps the Service *create* call, not the polling loop). The fast path has the same regression: a transient read error on an *existing* sandbox turns an idempotent `POST /api/sandboxes` into a 500, where the old code fell through to a 409-swallowed create and succeeded.
- **Suggested fix:** the 404→`None` narrowing is the right call for the existence check, but the wait path should keep tolerating transient errors — e.g. catch the raise inside the polling loop and `time.sleep(0.5)`/retry unless `exc.status` is in a small terminal set (404/403/401), or add a flag so `_sandbox_access_url` returns `None` for retryable statuses when called from the loop.
- **Test:** extend the `test_provisioner_request_threading.py` `_RecordingCoreV1.read_namespaced_service` to raise `ApiException(status=503)` for the first couple of calls then return the service, and assert `create_sandbox` still returns 200 with a URL (it currently raises).

docker/provisioner/app.py
```diff
@@ -685,27 +674,29 @@ def list_sandboxes():
...
-                SandboxResponse(
-                    sandbox_id=sid,
-                    sandbox_url=_sandbox_url(node_port),
-                    status=_get_pod_phase(sid),
-                )
+        sandbox_url = None
+        if SANDBOX_SERVICE_TYPE == "ClusterIP":
+            sandbox_url = _sandbox_url(sid)
+        else:
+            for port in svc.spec.ports or []:
```
willem-bd: **[P2] NodePort port-scan is now duplicated between `list_sandboxes` and `_sandbox_access_url`**

- **Problem:** the `for port in svc.spec.ports: if port.name == "http" and port.node_port` scan (plus the NodePort-vs-ClusterIP branching) now lives in two places. The duplication is understandable — `list_sandboxes` already holds the `svc` objects from the list call and wants to avoid N extra `read_namespaced_service` reads — but it means a future change to port name or the ClusterIP URL shape can drift between the two paths.
- **Suggested fix:** factor a small `_url_from_service(svc, sandbox_id)` helper that takes the already-fetched `svc` object and have both `_sandbox_access_url` (after its read) and `list_sandboxes` call it, so the access-mode branching stays in one spot.

willem-bd (commented):
Thanks @kongxiangxin. The ClusterIP + scoped skills subPath additions look reasonable, and the startup validation plus the sync-handler contract are handled well. I found a couple of issues worth addressing before this is ready — details inline.

docker/provisioner/app.py
kongxiangxin: Fixed in f7436e62. Factored `_url_from_service(svc, sandbox_id)` and now both `_sandbox_access_url` and `list_sandboxes` use it, so NodePort/ClusterIP URL construction stays in one place.

docker/provisioner/app.py
kongxiangxin: Fixed in f7436e62. `_sandbox_access_url` now has a retry-tolerant mode used by the create-sandbox fast path and readiness loop, while direct status reads still surface non-404 errors. Added a regression test that raises 503 and 429 before the service becomes readable.

WillemJiang (approved):

PR #3377, opened 2026-06-04
feat(tool):Add structured synopsis for oversized tool output previews

HaotianChen616:
## Why

Oversized tool results are currently externalized to thread storage, but the model-facing preview is still a raw head/tail slice. That makes common structured outputs, especially JSON, hard to inspect: the preview often shows partial syntax instead of useful shape, keys, counts, or representative fields.

This PR is inspired by OpenViking's tool result stub capability, but implements the behavior independently for DeerFlow's existing `ToolOutputBudgetMiddleware` path. The goal is to keep large raw tool outputs recoverable while giving the agent a compact, typed preview it can reason about.

## What changed

- Oversized externalized tool outputs now get a deterministic structured synopsis instead of only raw head/tail truncation.
- The synopsis detects common output kinds: JSON, CSV/TSV, YAML, XML, code, text, and unknown.
- JSON previews include top-level keys, container sizes, structural paths, line/byte anchors, and representative scalar values.
- CSV/TSV previews include row/column counts and headers.
- XML/YAML/code/text previews include lightweight shape and sample summaries.
- The existing raw output recovery path is preserved: the full content is saved under `.tool-results`, and the preview tells the agent to use `read_file` with line ranges.
- When durable tool-result storage is unavailable, the middleware still falls back to the conservative bounded sample behavior instead of pretending the omitted raw content is recoverable.
- Config comments and focused backend tests were updated for the new preview behavior.

## Surface area

- [ ] **Frontend UI** — page / component / setting / interaction under `frontend/`
- [ ] **Backend API** — endpoint / SSE event / request-response shape under `backend/app`
- [x] **Agents / LangGraph** — agent node, graph wiring, `langgraph.json`, or prompt change
- [ ] **Sandbox** — `docker/` or sandboxed execution
- [ ] **Skills** — change under `skills/`
- [ ] **Dependencies** — new/upgraded entry in `backend/pyproject.toml` or `frontend/package.json` (say what it buys us)
- [x] **Default behavior change** — changes existing behavior without the user opting in (default model, default setting, data shape)
- [ ] **Docs / tests / CI only** — no runtime behavior change

## Screenshots / Recording

Manual end-to-end terminal / embedded-client validation screenshots are included below. There are no frontend UI changes.

User prompt:
image

Raw tool-result:
08088b94b2a27a8b4ae626f9c17923c1

Tool-result preview:
38bd2b6ac4d1b9e45cc6812dc1348986

Read back prompt:
c508b2170b381ca030f1b7f3388d105b

[...]

Issue #3289, opened 2026-05-28
[Bug] DeerFlow 客户端未防护外接工具超大输出导致上下文溢出且线程无法恢复


heart-scalpel:
## 问题描述
当外接工具返回超大结果时，DeerFlow客户端没有任何前置防护，直接将完整结果透传给LLM，导致上下文溢出，整个线程执行完全中断且无法恢复。

### 复现步骤
1. Agent因语法错误，未按Skill规范生成带过滤条件的精确命令
2. 执行了无过滤的全局查询命令
3. 外接工具 `fdiTools_basic_remote_executor` 返回 **1.2MB** 超大结果
4. DeerFlow客户端直接将完整结果传入LLM上下文
5. 触发GPT-4o 262k token限制，返回400 BadRequestError
6. 整个线程崩溃，无法从已有checkpoint继续执行

### 关键日志
```
[DeerFlowClient] Tool fdiTools_basic_remote_executor completed, result length: 1213479

openai.BadRequestError: Error code: 400 - {'error': {'message': "This model's maximum context length is 262144 tokens..."}}
```

<img width="1442" height="846" alt="Image" src="https://github.com/user-attachments/assets/5dc7b472-9c14-4282-8e93-4606e08d3132" />

## 核心问题
1. **fdiTools是外接工具**，我们无法控制其输出大小和行为
2. **AI生成不符合Skill规范的命令是不可避免的**
3. 一旦发生，DeerFlow没有任何兜底防护，直接导致任务完全失败

## 建议修复
DeerFlow客户端应该提供最基础的兜底防护：
1. 对所有工具返回结果添加**默认50KB大小限制**，超过自动截断
2. 截断后返回标准化提示：`[输出已截断，共X字节] 请使用过滤条件缩小查询范围`
3. 检测到LLM上下文溢出错误时，自动回滚到上一个checkpoint重试

heart-scalpel:
---

## Checkpoint 86

- **ID**: `1f15a736-855a-66b4-8054-580985e6ef9c`
- **Parent ID**: `1f15a736-68bd-6bc1-8053-32e91c36d6d0`
- **Time**: 2026-05-28T08:58:34.337815+00:00

### [AI]

[...]

PR #4017, opened 2026-07-09
fix(agent): snap tool-output tail forward so fallback truncation respects max_chars

AnnaSuSu:
## Why

`ToolOutputBudgetMiddleware` exists so that a single oversized tool result cannot blow the model's context window (issue #3289; guard added in #3303). Its fallback path — taken when disk persistence is unavailable — documents a hard guarantee:

> `_build_fallback`: "The returned string is guaranteed to be no longer than *max_chars*."

It can return many times that.

`_snap_to_line_boundary()` moves an offset **backwards** to the preceding newline. That is correct for the head's *end* offset: moving back shortens the head. `_build_fallback()` also applies it to the tail's *start* offset, where moving backwards **lengthens** the tail.

The overshoot depends on where the last newline falls, and it grows with the size of the output — precisely when the guard matters most. With the shipped defaults (`fallback_max_chars=30000`, `fallback_head_chars=8000`, `fallback_tail_chars=3000`), on output that is line-oriented for its first ~60% and then emits one long unbroken line — a minified JSON body, a base64 artifact, a single-line stack dump; all everyday `bash` and `web_fetch` output:

| input size | returned | `max_chars` |
|---|---|---|
| 50,000 | 28,127 | 30,000 |
| 200,000 | **88,128** | 30,000 |
| 1,000,000 | **408,128** | 30,000 |

So the guard that exists to protect the context window can itself paste ~40% of the original payload back into it.

`TestBuildFallback::test_result_never_exceeds_max_chars` already asserts this invariant, but it passes `"x" * 50000` — newline-free content, so the snapping branch is never exercised.

## What changed

From a caller's perspective: an oversized tool result that falls back to inline truncation now honours `tool_output.fallback_max_chars` regardless of where newlines fall in the output. Previously a large, mostly-single-line result could still push hundreds of KB into the model context.

Internally, this adds `_snap_start_to_line_boundary()` — the forward-snapping mirror of `_snap_to_line_boundary()` — and uses it for the tail offset. Head snapping is unchanged. The search window mirrors the existing helper (the first half of `text[pos:]`, versus the second half of `text[:pos]`), so the tail is never shortened below half its requested length. The old `tail_start_snapped > head_end` guard is dropped: it existed to stop the backward snap from eating into the head, and a forward snap can only move the tail start further away from it.

`_build_preview()` currently contains the same three lines. It is deliberately **not** touched here: open PR #3377 replaces that function body wholesale with `render_tool_output_preview()`, which does not line-snap at all. Fixing it here would conflict for no net gain.

## Surface area

- [x] **Agents / LangGraph** — `ToolOutputBudgetMiddleware` (`backend/packages/harness/deerflow/agents/middlewares/tool_output_budget_middleware.py`)

## Bug fix verification

- Test path: `backend/tests/test_tool_output_budget_middleware.py::TestBuildFallback::test_result_never_exceeds_max_chars_with_newlines`
- Did it go red on `main` and green on this branch? **yes** — red on `main` with `AssertionError: total=200000: got 88128` / `assert 88128 <= 30000`; green on this branch.
- Also added `TestSnapStartToLineBoundary` (5 cases), including `test_never_moves_backwards`, which pins the property that made the original code wrong.

## Validation

- `cd backend && make lint` — `All checks passed!`, 797 files already formatted.
- `cd backend && make test` — **6750 passed, 26 skipped, 8 failed**. The 8 failures are pre-existing and unrelated: `test_browserless_client.py`, `test_crawl4ai_tools.py` and `test_fastcrw_tools.py` `web_fetch` cases fail identically on a clean `main` checkout in my environment, because it has no outbound DNS and the SSRF guard therefore rejects `https://example.com` with `Error: Refusing to fetch a private, loopback, or metadata address`. Baseline on `main`: 6744 passed / same 8 failed. The delta of +6 passing is exactly the 6 tests added here.
- `uv run pytest tests/test_tool_output_budget_middleware.py tests/test_tool_output_truncation.py` — 137 passed.
- Beyond the added regression test, I checked the guarantee as a property: 3,000 randomized inputs (newline densities 0 → 0.3, including "newlines only in the first 60%" shapes, CJK/emoji content, randomized `max_chars`/`head`/`tail`) produced **0 violations** of `len(result) <= max_chars` after the fix.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Claude Code assisted with locating the root cause and writing the fix and the regression tests. The contributor reviewed every line and takes responsibility for it.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/tool_output_budget_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/tool_output_budget_middleware.py
@@ -77,11 +77,31 @@ def _snap_to_line_boundary(text: str, pos: int) -> int:
     Used so that previews and truncations end on a complete line when
     possible.  If no newline exists in the second half of ``text[:pos]``
     the original *pos* is returned unchanged.
+
+    Only valid for an *end* offset: moving backwards shortens the slice that
+    ends here.  Use :func:`_snap_start_to_line_boundary` for a start offset.
     """
     if pos <= 0 or pos >= len(text):
         return pos
     half = pos // 2
     nl = text.rfind("\n", half, pos)
+    if nl >= 0:
+        return nl + 1
+    return pos
+
+
+def _snap_start_to_line_boundary(text: str, pos: int) -> int:
+    """Return *pos* or the nearest following newline+1, whichever is closer.
+
+    The start-offset mirror of :func:`_snap_to_line_boundary`. Snapping a start
+    backwards would *lengthen* the slice beginning there, so the tail of a
+    budgeted preview must snap forward instead. If no newline exists in the
+    first half of ``text[pos:]`` the original *pos* is returned unchanged.
+    """
+    if pos <= 0 or pos >= len(text):
+        return pos
+    half = pos + (len(text) - pos) // 2
+    nl = text.find("\n", pos, half)
     if nl >= 0:
         return nl + 1
     return pos
@@ -258,10 +278,7 @@ def _build_fallback(
     effective_tail = min(tail_chars, max(0, budget - effective_head))
 
     head_end = _snap_to_line_boundary(content, min(effective_head, total))
-    tail_start = max(head_end, total - effective_tail)
-    tail_start_snapped = _snap_to_line_boundary(content, tail_start)
-    if tail_start_snapped > head_end:
-        tail_start = tail_start_snapped
+    tail_start = _snap_start_to_line_boundary(content, max(head_end, total - effective_tail))
 
     head = content[:head_end]
     tail = content[tail_start:] if tail_start < total else ""
--- a/backend/tests/test_tool_output_budget_middleware.py
+++ b/backend/tests/test_tool_output_budget_middleware.py
@@ -27,6 +27,7 @@ from deerflow.agents.middlewares.tool_output_budget_middleware import (
     _needs_budget,
     _patch_model_messages,
     _sanitize_tool_name,
+    _snap_start_to_line_boundary,
     _snap_to_line_boundary,
     _tool_message_over_budget,
 )
@@ -37,6 +38,20 @@ from deerflow.config.tool_output_config import ToolOutputConfig
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
+
+
+def _lines_then_long_line(total: int, newline_ratio: float = 0.6) -> str:
+    """Content that is line-oriented for the first *newline_ratio*, then one unbroken line.
+
+    Mirrors real bash/web_fetch output that logs progress lines and then dumps a
+    single-line artifact (minified JSON, base64 blob). The last newline lands in
+    the second half of the content, which is what exercises line snapping around
+    the tail offset.
+    """
+    head_len = int(total * newline_ratio)
+    lines = "".join(f"[info] step {i} ok\n" for i in range(head_len // 18 + 1))[:head_len]
+    lines = lines[:-1] + "\n" if not lines.endswith("\n") else lines
+    return lines + "A" * (total - len(lines))
 
 
 def _make_request(tool_name: str = "remote_executor", tool_call_id: str = "tc-1", outputs_path: str | None = None) -> SimpleNamespace:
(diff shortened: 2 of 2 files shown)


willem-bd:
### Follow-up suggestion (non-blocking): a `_build_fallback` test that exercises the forward-snap path

The committed regression test (`test_result_never_exceeds_max_chars_with_newlines`) pins the `max_chars` bound nicely, but its content shape — line-oriented for the first ~60%, then one unbroken line — means the tail's forward-snap window `[tail_start, tail_start + (total − tail_start)//2)` contains **no newline**. So in that test `_snap_start_to_line_boundary` returns `pos` unchanged: it passes because the harmful *backward* snap is gone, not because the *forward* snap actually moved the offset.

The forward-snap-move path is covered by the `TestSnapStartToLineBoundary` unit tests, but nothing exercises it end-to-end through `_build_fallback`. A small test placing a newline inside the tail's snap window would close that gap and also pin the snap *direction*:

```python
def test_fallback_forward_snaps_tail_onto_line_boundary(self):
    max_chars, head_chars, tail_chars = 30_000, 8_000, 3_000
    total = 100_000
    nl_pos = total - 2_000  # lands inside the tail's forward-snap window
    content = "A" * nl_pos + "\n" + "B" * (total - nl_pos - 1)
    result = _build_fallback(
        content, tool_name="bash",
        max_chars=max_chars, head_chars=head_chars, tail_chars=tail_chars,
    )
    assert len(result) <= max_chars
    # tail begins right after the newline (forward snap fired);
    # on main the backward snap leaves the tail starting with "A"
    tail = result.split("[...", 1)[1].split("]\n\n", 1)[1]
    assert tail.startswith("B")
```

With `nl_pos = total − 2000`, the raw tail offset is `total − 3000 = 97000` and the snap window is `[97000, 98500)`, so the newline at `98000` is found and the tail start moves forward to `98001`. The `tail.startswith("B")` assertion fails on `main` (old code leaves `tail_start = 97000`, tail begins with `"A"`), so it doubles as a direction pin.

Totally optional given the primitive coverage — just flagging the end-to-end gap. Nice fix overall.

AnnaSuSu:
Verified and applied — thanks, this is a real gap.

You're right that the committed bound test never moves the tail offset. I checked all three sizes: with `effective_tail = 3_000` the snap window is `[total − 3_000, total − 1_500)`, which for `_lines_then_long_line(total, 0.6)` sits entirely inside the trailing unbroken line, so `_snap_start_to_line_boundary` returns `pos` unchanged every time. Deleting the snap call outright leaves that test green — it pins the `max_chars` bound, not the direction.

Added `test_fallback_forward_snaps_tail_onto_line_boundary` in 8fae91f, close to your sketch. I take the tail with `result.rsplit("]\n\n", 1)[1]` since head and tail are newline-free by construction. It is a genuine direction pin: reverting just the `tail_start` line to the old backward snap — leaving the helper defined, so the failure is an assertion rather than an `ImportError` — gives

```
E   AssertionError: tail begins mid-line: 'AAAAAAAAAAAAAAAAAAAA'
```

### On the `needs-validation` label

Rather than only exercising the primitive, I drove the real entry point: production `ToolOutputConfig()` defaults (30_000 / 8_000 / 3_000), the real `ToolOutputBudgetMiddleware.awrap_tool_call`, and the stdout of an actual `bash -c` — a few passes of `find /usr/share`, then a minified-JSON dump. That yields 3,966,630 chars whose last newline is 92% through, so the trailing artifact is one 310,890-char line. I reached "persistent storage unavailable" two ways: no `outputs_path` in `thread_data`, and an `outputs_path` whose directory is unwritable so `_externalize` swallows the `OSError`.

| | model receives | within the 30k budget |
|---|---|---|
| `main` | 318,998 chars | ✗ 10.6× over |
| this PR | 11,108 chars | ✓ |

Identical in both scenarios, and the tail stays a true suffix of the real output.

One caveat worth recording: the shape matters. My first attempt used a 1.57M-char output whose last newline was only 33% through — that one is **not** affected on `main` either, because the backward snap finds no newline in its search window. The bug needs log-heavy output followed by a moderate single-line artifact, which is exactly what `bash` and `web_fetch` produce. I did not commit that as a test: it depends on the contents of `/usr/share` and builds a 4 MB string, so the committed coverage stays deterministic.

Full backend suite: 6751 passed. The 8 failures on my machine are pre-existing and unrelated — `web_fetch` cases that need outbound DNS, so the SSRF guard rejects `example.com`. Ruff check and format are clean.

WillemJiang (approved):

fancyboi999 (commented):
Post-merge review (review was requested from me, but the merge landed first — closing the loop). Read the full diff.

The fix is sound: `_snap_start_to_line_boundary` can only move the tail start *forward*, so the tail can only shrink and the `max_chars` bound holds by construction. Dropping the old `tail_start_snapped > head_end` guard is also correct — a forward snap starting from `max(head_end, total - effective_tail)` can never land before `head_end`. `test_never_moves_backwards` pins exactly the property the original code violated, and `test_fallback_forward_snaps_tail_onto_line_boundary` pins the direction so a regression back to backward-snapping can't pass. Nice, surgical fix. LGTM.

PR #4019, opened 2026-07-09
feat(mcp): auto-promote deferred MCP tools from routing hints

18062706139fcz:
## Summary

Second layer of the MCP routing work (follows #4004). When
`tool_search.enabled=true` defers MCP tool schemas, PR1 routing hints still
force the model to spend a `tool_search` discovery round trip before it can
call the tool the routing metadata already points at. This adds a
`McpRoutingMiddleware` that matches the latest user message against PR1
routing keywords and promotes the matching deferred schemas **before** the
model call, removing that round trip.

Soft routing, opt-in, purely additive. **No behavior change when
`tool_search.enabled=false`.**

## Design

- **Latest user message only** — matches the latest real `HumanMessage` via a
  shared `is_real_user_message` helper (also reused by
  `SkillActivationMiddleware` so the two cannot drift). Case-insensitive
  substring match; no tokenizer dependency; Chinese and English use the same
  rule.
- **Ordering + breadth** — priority desc, then tool name asc; capped by the new
  global `tool_search.auto_promote_top_k` (default 3, clamped 1..5). PR1 schema
  is unchanged: no per-tool `auto_promote_top_k` is added or consumed (a
  per-tool value is ignored with a DEBUG note).
- **State** — returns a plain `{"promoted": ...}` state update (not a
  `Command`) and relies on `ThreadState.merge_promoted` for union/dedupe, so
  auto-promote and a model-triggered `tool_search` converge on the same catalog
  hash.
- **Placement** — installed before `DeferredToolFilterMiddleware` on every
  deferred-tool path (lead agent, subagent, embedded client, webhook via shared
  builders). A construction-time assert rejects the reversed order.
  `catalog_hash is None` / no routing index is a complete no-op, so bootstrap
  and ACP skip it.
- **Privacy** — never executes tools, never promotes policy-filtered tools,
  adds no routing keywords or matched tool names to trace metadata or
  INFO/WARN logs.

## Real-LLM validation

End-to-end against a real DeepSeek model (`deepseek-chat-fast`), same user
message hitting a routing keyword, deferred loading on:

- User message: `Query the orders table for the 5 most recent orders.`
- Routed deferred MCP-like tool: `postgres_query`
- Routing keywords: `orders`, `order`, `SQL`, `database`

| Metric | Baseline (auto-promote off) | Auto-promote on |
|---|---:|---:|
| First tool schemas bound to model | `web_search`, `tool_search` | `postgres_query`, `web_search`, `tool_search` |
| Model's first action | `tool_search` | `postgres_query` |
| `tool_search` discovery rounds | 4 | **0** |
| Tool invocation trace | `tool_search×4 + web_search + postgres_query×3` | `postgres_query×1` |
| Total model calls | 9 | **2** |
| Wall time | 14.41s | **5.81s** |

This validates the intended positive optimization: the routed schema is visible
on the first model call, the model can call it directly, and the discovery round
trip is removed.

## Test plan

- [x] Unit: routing index construction (deferred + `prefer` + non-empty
      keywords only; ignores per-tool `auto_promote_top_k`; clamps global top-k)
- [x] Unit: matching semantics (latest message only, casefold, Chinese,
      priority tie-break, top-k, structured content extraction, skips
      `summary` / `hide_from_ui`)
- [x] Unit: middleware state updates (`before_model` / `abefore_model`,
      no-match / no-hash no-op, minimal delta, no ToolMessage)
- [x] Integration: same-cycle deferred-filter interaction, direct call without
      `tool_search`, explicit `tool_search` merges with auto-promoted names
- [x] Runtime paths: lead agent + subagent order invariant, embedded-client
      builder wiring, bootstrap/ACP skip, reversed order rejected
- [x] Config: `auto_promote_top_k` clamping + `config.example.yaml`
      parseability regression
- [x] Privacy: no keywords / tool names in trace metadata or INFO/WARN logs
- [x] `223 passed` across all affected suites; `ruff format --check` clean


18062706139fcz:
## Real-LLM validation details

Adding the full experiment notes here so reviewers can see why this PR is a positive optimization, not just a mechanical middleware change.

### What the experiment validates

The PR is intended to solve this exact behavior under `tool_search.enabled=true`:

1. Deferred loading hides MCP tool schemas from the first model call.
2. PR1 routing metadata already tells us which deferred MCP tool is relevant.
3. Without auto-promote, the model must spend at least one `tool_search` discovery round before it can call the routed tool.
4. With auto-promote, the routed tool schema is visible on the first model call, so the model can call it directly.

The experiment compares those two paths with the same real LLM, same user message, same routed MCP-like tool, and same deferred-tool setup.

### Experiment setup

- Real model: `deepseek-chat-fast` via the local `config.yaml` model factory.
- User message: `Query the orders table for the 5 most recent orders.`
- Deferred MCP-like tool: `postgres_query`
- Routing metadata: `mode=prefer`, `priority=100`, `keywords=["orders", "order", "SQL", "database"]`
- Control tool: `web_search`
- Deferred loading: forced on via `assemble_deferred_tools(..., enabled=True)`

Two scenarios were run:

| Scenario | Middleware chain |
|---|---|
| Baseline | `[DeferredToolFilterMiddleware]` |
| Auto-promote | `[McpRoutingMiddleware, DeferredToolFilterMiddleware]` |

The script also wraps the real model with a small proxy that records every `bind_tools(...)` call. This is important because the first bound tool list proves whether `postgres_query` was visible to the model before its first response.

### Result

| Metric | Baseline (auto-promote off) | Auto-promote on |
|---|---:|---:|
| First tool schemas bound to model | `web_search`, `tool_search` | `postgres_query`, `web_search`, `tool_search` |
| Model's first action | `tool_search` | `postgres_query` |
| `tool_search` discovery rounds | 4 | **0** |
| Tool invocation trace | `tool_search×4 + web_search + postgres_query×3` | `postgres_query×1` |
| Total model calls | 9 | **2** |
| Wall time | 14.41s | **5.81s** |

This shows the intended optimization clearly:

- Baseline hides the routed tool on the first model call, so the model starts with `tool_search` discovery and even detours through `web_search`.
- Auto-promote exposes the routed schema on the first model call, so the model directly calls `postgres_query` with zero `tool_search` discovery rounds.

### Reproduction script used locally

This was a temporary local scratch script and was not committed. It is included here for reviewability. It expects the repo's model config to resolve `DEEPSEEK_API_KEY` from the environment; no key is printed or embedded.

```python
"""Real-LLM end-to-end check for PR2 MCP routing auto-promote.

Compares two middleware configs against a REAL DeepSeek model, using the same
user message that hits a routing keyword:

  A) auto-promote ON : [McpRoutingMiddleware, DeferredToolFilterMiddleware]
  B) baseline OFF    : [DeferredToolFilterMiddleware] only

We record, for each first model call, which tool schemas were bound and whether
the model's first action was tool_search (a discovery round trip) or the routed
tool directly. This is a manual scratch script, not a pytest test.
"""

import asyncio
import os
import time

from langchain.agents import create_agent
from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
from langchain_core.tools import tool as as_tool

from deerflow.agents.middlewares.deferred_tool_filter_middleware import DeferredToolFilterMiddleware
from deerflow.agents.thread_state import ThreadState
from deerflow.config.app_config import get_app_config
from deerflow.models import create_chat_model
from deerflow.tools.builtins.tool_search import assemble_deferred_tools, build_mcp_routing_middleware
from deerflow.tools.mcp_metadata import tag_mcp_routing, tag_mcp_tool

MODEL_NAME = "deepseek-chat-fast"
USER_MESSAGE = "Query the orders table for the 5 most recent orders."
SYSTEM_PROMPT = (
    "You are a data assistant. Once a tool returns the data you need, reply to "
    "the user with the answer in text. Do not call the same tool more than once "
    "for the same request."
)


@as_tool
def postgres_query(sql: str) -> str:
    "Run a read-only SQL query against the production Postgres database."
    return (
        "5 rows returned:\n"
        "order_id=1005 total=42.00 created=2026-07-09\n"
        "order_id=1004 total=17.50 created=2026-07-09\n"
        "order_id=1003 total=88.10 created=2026-07-08\n"
        "order_id=1002 total=5.00 created=2026-07-08\n"
        "order_id=1001 total=63.20 created=2026-07-07"
    )


@as_tool
def web_search(query: str) -> str:
    "Search the public web."
    return "no results"


def routed_postgres():
    tag_mcp_tool(postgres_query)
    tag_mcp_routing(
        postgres_query,
        {
            "mode": "prefer",
            "priority": 100,
            "keywords": ["orders", "order", "SQL", "database"],
        },
    )
    return postgres_query


def build_recording_model(bound_calls: list[list[str]]):
    app_config = get_app_config()
    model = create_chat_model(
        name=MODEL_NAME,
        thinking_enabled=False,
        app_config=app_config,
        attach_tracing=False,
    )

    class RecordingProxy:
        def __init__(self, inner):
            object.__setattr__(self, "_inner", inner)

        def bind_tools(self, tools, **kwargs):
            bound_calls.append([getattr(tool, "name", None) for tool in tools])
            return self._inner.bind_tools(tools, **kwargs)

        def __getattr__(self, name):
            return getattr(self._inner, name)

    return RecordingProxy(model)


def summarize(label: str, result: dict, bound_calls: list[list[str]], elapsed: float):
    messages = result["messages"]
    tool_messages = [message for message in messages if isinstance(message, ToolMessage)]
    ai_messages = [message for message in messages if isinstance(message, AIMessage)]

    first_tool_call = None
    for message in ai_messages:
        if getattr(message, "tool_calls", None):
            first_tool_call = message.tool_calls[0]["name"]
            break

    tool_search_calls = sum(1 for message in tool_messages if message.name == "tool_search")

    print(f"\n===== {label} =====")
    print(f"elapsed: {elapsed:.2f}s")
    print(f"first bound tools ({len(bound_calls[0]) if bound_calls else 0}): {bound_calls[0] if bound_calls else None}")
    print(f"model's FIRST tool call: {first_tool_call}")
    print(f"# tool_search invocations: {tool_search_calls}")
    print(f"tools invoked (in order): {[message.name for message in tool_messages]}")
    print(f"promoted state: {result.get('promoted')}")
    print(f"total model calls (bind_tools count): {len(bound_calls)}")
    return {"first_tool_call": first_tool_call, "tool_search_calls": tool_search_calls}


async def run_scenario(label: str, *, auto_promote: bool):
    final_tools, setup = assemble_deferred_tools([routed_postgres(), web_search], enabled=True)
    bound_calls: list[list[str]] = []
    model = build_recording_model(bound_calls)

    middleware = []
    if auto_promote:
        routing_middleware = build_mcp_routing_middleware(final_tools, setup, top_k=3)
        assert routing_middleware is not None
        middleware.append(routing_middleware)
    middleware.append(DeferredToolFilterMiddleware(setup.deferred_names, setup.catalog_hash))

    graph = create_agent(
        model=model,
        tools=final_tools,
        system_prompt=SYSTEM_PROMPT,
        middleware=middleware,
        state_schema=ThreadState,
    )

    started = time.time()
    result = await graph.ainvoke(
        {"messages": [HumanMessage(content=USER_MESSAGE)]},
        {"recursion_limit": 25},
    )
    return summarize(label, result, bound_calls, time.time() - started)


async def main():
    print(f"model={MODEL_NAME!r} user={USER_MESSAGE!r}")
    baseline = await run_scenario("B) BASELINE (auto-promote OFF)", auto_promote=False)
    promoted = await run_scenario("A) AUTO-PROMOTE ON", auto_promote=True)

    print("\n########## VERDICT ##########")
    print(f"Baseline    : first_call={baseline['first_tool_call']} tool_search_rounds={baseline['tool_search_calls']}")
    print(f"Auto-promote: first_call={promoted['first_tool_call']} tool_search_rounds={promoted['tool_search_calls']}")


if __name__ == "__main__":
    # Required: DEEPSEEK_API_KEY (or equivalent model config env var) in the environment.
    assert os.environ.get("DEEPSEEK_API_KEY"), "Set DEEPSEEK_API_KEY before running this scratch validation."
    asyncio.run(main())
```

backend/packages/harness/deerflow/agents/middlewares/mcp_routing_middleware.py
```diff
@@ -0,0 +1,132 @@
...
+            return []
+
+        text = get_original_user_content_text(target.content, target.additional_kwargs)
+        if not text:
+            return []
+
+        haystack = text.casefold()
+        matched: list[tuple[int, str]] = []
+        for name, (priority, keywords) in self._routing_index.items():
+            if any(keyword.casefold() in haystack for keyword in keywords):
```
willem-bd, line 95: **Substring matching can over-match short keywords (minor).** `keyword.casefold() in haystack` is a raw substring test, so e.g. `"api"` matches `rapid`/`capital`, `"go"` matches `going`/`ago`. Since promotion is soft/additive (it only *exposes* a schema and never disables other tools), the blast radius is one extra bound schema — not a routing failure — so this is fine as-is.

Worth a one-line note in `MCP_SERVER.md` (the only place `keywords` semantics are documented) advising distinctive keywords, so operators authoring them know it's a substring match rather than token/word-boundary matching.

backend/packages/harness/deerflow/tools/builtins/tool_search.py
```diff
@@ -202,6 +207,59 @@ def assemble_deferred_tools(filtered_tools: list[BaseTool], *, enabled: bool) ->
...
 
 
+def _routing_priority(value: Any) -> int:
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return 0
+
+
+def _routing_keywords(value: Any) -> list[str]:
```
willem-bd, line 220: **Duplicated normalization logic (minor, defense-in-depth).** `_routing_priority`/`_routing_keywords` here produce typed data, then `McpRoutingMiddleware._normalize_index` re-parses the same `priority`/`keywords` fields. The middleware's docstring justifies the defensive re-parse (built to accept serialized data), but it's two copies of the normalization logic that can drift; in practice `_normalize_index` is a no-op over the builder's output. Not blocking — if both stay, a cross-reference comment between them would help.

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -437,13 +437,27 @@ def _create_agent(self, tools: list[BaseTool] | None = None, *, deferred_setup:
...
-            lazy_init=True,
-            deferred_setup=deferred_setup,
-            agent_name=self.config.name,
-        )
+        mcp_routing_middleware = None
+        if deferred_setup is not None and deferred_setup.deferred_names:
+            from deerflow.tools.builtins.tool_search import build_mcp_routing_middleware
+
+            tool_search_config = getattr(app_config, "tool_search", None)
+            auto_promote_top_k = getattr(tool_search_config, "auto_promote_top_k", 3)
```
willem-bd: **Defensive `getattr` is inconsistent with the other two paths (nit).** The lead-agent and embedded-client paths access `resolved_app_config.tool_search.auto_promote_top_k` directly. `AppConfig.tool_search` is always present (pydantic with defaults), so the `getattr(..., None)` / `getattr(..., 3)` here reads as if the field is optional and is more defensive than the type warrants. Direct access would match the other paths and surface config typos at the right layer.

backend/packages/harness/deerflow/agents/middlewares/mcp_routing_middleware.py
```diff
@@ -0,0 +1,132 @@
...
+            return []
+
+        matched.sort(key=lambda item: (-item[0], item[1]))
+        return [name for _, name in matched[: self._top_k]]
+
+    def _state_update(self, state: Mapping[str, Any] | None) -> dict[str, Any] | None:
+        names = self._matched_names(state)
+        if not names:
+            return None
+        logger.debug(
```
willem-bd, line 108: **Minor note on the privacy claim.** The PR description scopes 'no matched tool names in logs' to INFO/WARN, which is accurate — but at DEBUG this line logs `names=...` (the matched tool names). Tool names like `postgres_query` aren't generally sensitive and routing keywords are never logged, so this is reasonable; just flagging that 'no matched tool names in logs' isn't true at DEBUG, in case DEBUG is expected on some deployment/debug path.

willem-bd (commented):
## Review: auto-promote deferred MCP tools from routing hints

**Verdict: Looks good — no correctness bugs.** I traced the full `before_model` state-update → `merge_promoted` reducer → `DeferredToolFilterMiddleware._filter_tools` path and confirmed the promotion is visible in the same model cycle (your `test_auto_promote_makes_schema_visible_in_same_model_cycle` proves this end-to-end). Verified against PR1 (#4004, merged) and the established `before_model` pattern used by 17+ existing middlewares.

**Strengths**
- Clean use of the existing `merge_promoted` reducer so auto-promote and a model-triggered `tool_search` converge on one unioned/deduped set.
- Good refactor: extracting `is_real_user_message` so skill-activation and MCP routing can't drift (it's byte-identical to the old `_is_user_activation_target`).
- Construction-time `assert_mcp_routing_before_deferred_filter` across all three builders is a solid guardrail.
- Fail-closed posture holds: only post-policy-filter deferred tools are indexed, tools are never executed, keywords are never logged.
- Docs + `config_version` bump + example-parseability test all kept in sync.

Four minor/nit findings posted inline below — none blocking.

backend/packages/harness/deerflow/agents/middlewares/mcp_routing_middleware.py
```diff
@@ -0,0 +1,137 @@
...
+    @override
+    def before_model(self, state: AgentState, runtime: Runtime) -> dict[str, Any] | None:
+        return self._state_update(state)
+
+    @override
+    async def abefore_model(self, state: AgentState, runtime: Runtime) -> dict[str, Any] | None:
+        return self._state_update(state)
+
+
+def assert_mcp_routing_before_deferred_filter(middlewares: Sequence[AgentMiddleware]) -> None:
```
willem-bd, line 130: This assert is enforced, and `test_order_invariant_rejects_reversed_middlewares` verifies it fires — but nothing demonstrates that a *reversed* order actually changes behavior. `McpRoutingMiddleware` only implements `before_model`, and `DeferredToolFilterMiddleware` only implements `wrap_model_call`, so a `before_model` state update is normally applied before the `wrap_model_call` chain reads `request.state` regardless of registration order.

If the order isn't functionally load-bearing, the assert is over-strict (rejects a valid configuration); if it is, a test showing reversed order fails to promote — or a one-line comment naming the LangGraph hook-timing assumption that makes the order matter — would justify keeping it. Worth resolving so a future reader doesn't misunderstand or silently weaken the invariant.

backend/packages/harness/deerflow/agents/middlewares/mcp_routing_middleware.py
```diff
@@ -0,0 +1,137 @@
...
+            if any(keyword.casefold() in haystack for keyword in keywords):
+                matched.append((priority, name))
+
+        if not matched:
+            return []
+
+        matched.sort(key=lambda item: (-item[0], item[1]))
+        return [name for _, name in matched[: self._top_k]]
+
+    def _state_update(self, state: Mapping[str, Any] | None) -> dict[str, Any] | None:
```
willem-bd, line 104: Question: on a multi-turn tool loop the latest real `HumanMessage` is unchanged, so `_state_update` re-emits the same `promoted` update every `before_model`. It's correct (monotonic union via `merge_promoted`), but the reducer returns a fresh dict each time even when the set is unchanged, which may register as a channel update each turn (possible per-turn checkpoint churn on a long subagent run).

If that matters, an optional early-out — read `state["promoted"]` and return `None` when the matched names are already a subset under the same `catalog_hash` — would make steady-state turns a true no-op. Low priority given how small the value is.

backend/packages/harness/deerflow/agents/middlewares/mcp_routing_middleware.py
```diff
@@ -0,0 +1,137 @@
...
+            return []
+
+        text = get_original_user_content_text(target.content, target.additional_kwargs)
+        if not text:
+            return []
+
+        haystack = text.casefold()
+        matched: list[tuple[int, str]] = []
+        for name, (priority, keywords) in self._routing_index.items():
+            if any(keyword.casefold() in haystack for keyword in keywords):
```
willem-bd, line 95: Minor: keywords are re-`casefold()`-ed on every `before_model` call here, because `_normalize_index` stores them in their original case. Pre-casefolding in `_normalize_index` (store `tuple(k.casefold() for k in keywords)`) would avoid the repeat. Micro-optimization — keyword sets are tiny — just noting.

backend/packages/harness/deerflow/tools/builtins/tool_search.py
```diff
@@ -202,6 +207,64 @@ def assemble_deferred_tools(filtered_tools: list[BaseTool], *, enabled: bool) ->
     return final_tools, deferred_setup
 
 
+def _routing_priority(value: Any) -> int:
```
willem-bd, line 210: The priority/keyword coercion lives in two places — here (`_routing_priority`/`_routing_keywords`) and `McpRoutingMiddleware._normalize_index` — each carrying a "keep in sync" comment. The defensive re-normalization in the middleware is reasonable (it accepts serialized data), but it's a two-place maintenance hazard. Centralizing the coercion in one helper imported by both would remove the sync burden.

willem-bd (commented):
## Overview

Second layer of the MCP routing work (follows #4004). Adds `McpRoutingMiddleware` that matches the latest user message against PR1 routing keywords and promotes matching deferred schemas **before** the model call, removing the `tool_search` discovery round-trip. Soft, opt-in, purely additive — no behavior change when `tool_search.enabled=false`. Promotes via a plain `{"promoted": ...}` update through the existing `merge_promoted` reducer, so auto-promote and explicit `tool_search` converge on one catalog hash.

## Correctness — verified against the code

- `final_tools` does contain the deferred MCP tools (`assemble_deferred_tools` keeps them for `ToolNode`; the filter only hides their schemas from `bind_tools`), so `build_mcp_routing_middleware` correctly indexes them.
- The reducer path is sound: `merge_promoted` unions on matching hash and `DeferredToolFilterMiddleware._promoted` reads `state["promoted"]` keyed on that hash — auto-promote and explicit promotion behave identically.
- `is_real_user_message` correctly excludes the summarization trigger (`_SUMMARY_TRIGGER_MESSAGE_NAME = "summary"` == `SUMMARY_MESSAGE_NAME`), so the refactor preserves skill-activation behavior and now covers routing too. Nice consolidation that removes drift risk between the two.
- Pure no-op when disabled: empty `deferred_names` -> builder returns `None` -> neither routing nor deferred-filter middleware is appended; bootstrap/ACP likewise.

## Strengths

- Strong security posture: never executes tools, only indexes tools that survived policy filtering, holds no `BaseTool` objects (serialized routing index only), and leaks no keywords/tool names into trace metadata or INFO/WARN logs. Matching on operator-authored keywords can't escalate — worst case is one extra bound schema the model may ignore.
- Docs fully synced (README, backend/README, MCP_SERVER.md, AGENTS.md renumbered 23->31, config.example.yaml, `config_version` 20->21).
- Layered tests including a real-graph integration test, plus the real-LLM validation table in the description (9->2 model calls, 14.4s->5.8s).

## One body-level note

The headline latency/model-call wins are reported only in the PR description, not as a committed CI test. Using `GenericFakeChatModel` in the integration tests is the right call (no external flakiness), but it means the optimization itself isn't regression-protected — acceptable given `needs-validation`, just worth being aware of.

Overall: high quality, correct core mechanism, docs in sync. The one item I'd want resolved before merge is the ordering-assert question (first inline) — either a behavioral test proving reversed order breaks promotion, or a comment documenting the LangGraph hook-timing assumption. The rest is optional polish.

WillemJiang (approved):

18062706139fcz:
## Follow-up live bench evidence using the Harbor validation pattern

Adding one more post-merge validation note for the auto-promote optimization in this PR. I used the evidence style from the DeerFlow Harbor adapter live smoke comment here as the reference pattern:

- https://github.com/harbor-framework/harbor/pull/2106#issuecomment-4932477096

That Harbor comment is useful because it does not just say "tests pass". It records a real model-backed run, the task/runtime context, the adapter path being exercised, the observed output, the runner summary, token usage, and caveats about what the smoke does and does not prove. I mirrored that style for this PR's optimization bench: one real model, one controlled task, two middleware configurations, and explicit evidence for whether the routed tool schema is visible on the first model call.

### What I ran

This was a real-LLM local bench against PR #4019's head commit, not a pytest-only check and not a full Harbor CLI batch run. The goal was to reproduce the important Harbor-style evidence boundary for this PR: prove that the agent can skip the deferred-tool discovery round and call the routed MCP tool directly.

Setup steps:

```bash
git fetch origin pull/4019/head:refs/remotes/origin/pr/4019
git worktree add --detach /private/tmp/deer-flow-pr4019-auto-bench-... origin/pr/4019
cd /private/tmp/deer-flow-pr4019-auto-bench-.../backend
```

I then ran a scratch script in that isolated worktree with the repo's real local model config:

```bash
set -a
source /Users/bytedance/deer-flow/.env
set +a

PYTHONUNBUFFERED=1 \
DEER_FLOW_PROJECT_ROOT=/Users/bytedance/deer-flow \
DEER_FLOW_CONFIG_PATH=/Users/bytedance/deer-flow/config.yaml \
BENCH_MODEL_NAME=deepseek-chat-fast \
uv run python tmp_mcp_auto_promote_bench.py
```

The local config resolved `deepseek-chat-fast` to the configured DeepSeek chat model. API keys were only loaded through environment variables; the script printed no key values.

### Harbor-style task shape

The bench uses the same kind of controlled task shape as the Harbor smoke evidence: a real model must complete a concrete tool-backed task, and we inspect the runtime trace rather than only asserting internal state.

User message:

```text
Query the orders table for the 5 most recent orders.
```

Tools supplied to the agent:

- `postgres_query`: MCP-tagged deferred database tool. It returns five deterministic order rows.
- `web_search`: normal control tool returning `no results`.

Routing metadata on `postgres_query`:

```python
mode = "prefer"
priority = 100
keywords = ["orders", "order", "SQL", "database"]
```

The script wraps the real chat model with a small recording proxy around `bind_tools(...)`. That is the key Harbor-style evidence point here: the first bound tool list proves whether `postgres_query` was actually visible to the model before the first response.

### Compared scenarios

| Scenario | Middleware chain |
|---|---|
| Baseline | `[DeferredToolFilterMiddleware]` |
| Auto-promote | `[McpRoutingMiddleware, DeferredToolFilterMiddleware]` |

Both scenarios use the same model, same user message, same tools, same routing metadata, and same deferred-tool setup. The only intended difference is whether PR #4019's routing middleware writes the `promoted` state before deferred schema filtering.

### Result

| Metric | Baseline | Auto-promote |
|---|---:|---:|
| First tool schemas bound to model | `web_search`, `tool_search` | `postgres_query`, `web_search`, `tool_search` |
| Model's first tool call | `tool_search` | `postgres_query` |
| `tool_search` discovery rounds | 4 | **0** |
| Tool invocation trace | `tool_search x4` | `postgres_query x1` |
| Total model calls | 5 | **2** |
| Wall time | 8.079s | **3.480s** |
| Total tokens | 3502 | **1551** |
| Final outcome | Did not reach the DB tool | Successfully returned the five orders |

Observed optimized answer included the expected five deterministic rows:

```text
1005  $42.00  2026-07-09
1004  $17.50  2026-07-09
1003  $88.10  2026-07-08
1002  $5.00   2026-07-08
1001  $63.20  2026-07-07
```

### Takeaway

This validates the intended optimization in PR #4019:

- Without auto-promote, the first model call cannot see `postgres_query`, so it starts with `tool_search` and spends discovery rounds trying to find the missing schema.
- With auto-promote, the routing keyword match promotes `postgres_query` before model binding, so the model calls the target MCP tool directly on the first action.
- In this run, that changed the path from an unsuccessful discovery loop into a successful one-tool execution.

Quantitatively, this run reduced model calls by 60%, wall time by about 57%, and total tokens by about 56%.

Caveat: this was a local Harbor-pattern live bench rather than a full `harbor run` invocation. The part intentionally borrowed from Harbor is the evidence discipline: real model, isolated task, explicit runtime context, trace-derived metrics, and clear caveats. The isolated worktree and scratch script were removed after the run.

Issue #2450, opened 2026-04-22
Memory Module Roadmap


ShenAC-SAC:
# DeerFlow Memory Module — Roadmap

## 1. Purpose

This document lays out the main problems in the memory module as of `main` today, maps them to existing RFCs / PRs, and proposes the order we should tackle them.

It is not an exhaustive checklist. It covers the most important architectural issues, their dependencies, and the sequencing.

Three states matter:

- merged into upstream `main`
- open RFC
- open PR

If it has not landed on `main`, it is not shipped.

---

## 2. Current state and problems

There are seven things that need fixing. The first five are structural; the last two are about quality and visibility of what we already have.

### P1 — Write path is not multi-worker safe

The write path today is an in-process `list` queue with `threading.Timer` debounce (`queue.py`), writing to a single JSON file via `FileMemoryStorage`. There is no durable queue, no writer lease, and no stale-task recovery.

On top of that, `updater.py` runs a `ThreadPoolExecutor(max_workers=4)` to bridge async-to-sync memory updates. This means up to four concurrent writers can hit `FileMemoryStorage.save()` at the same time. The mtime-based cache in `storage.py` has a TOCTOU window under concurrent writes — a stale read can silently overwrite a newer write.

There are also two separate entry points into the queue that both need to migrate together:

- `MemoryMiddleware.after_agent()` → `queue.add()` (debounced)
- `summarization_hook.memory_flush_hook()` → `queue.add_nowait()` (immediate)

[RFC #2283](https://github.com/bytedance/deer-flow/issues/2283) / [PR #2403](https://github.com/bytedance/deer-flow/pull/2403) are meant to fix this. Neither has landed on `main`.

### P2 — Read path is confidence-first, not query-aware

`format_memory_for_injection()` in `prompt.py` sorts facts by confidence descending and packs them into a token budget. The query itself is not considered — every prompt gets the same ranking regardless of what the user is asking about.

### P3 — Retrieval is not measurable

There is no per-injection trace of which facts were selected, no replay mechanism, and no baseline eval. Without numbers, any change to retrieval ranking is guesswork.

### P4 — Schema is scattered

The memory data shape is defined independently in at least seven places:

| Layer | Location |
|---|---|
| Storage default | `storage.py` → `create_empty_memory()` |
| LLM prompt contract | `prompt.py` → `MEMORY_UPDATE_PROMPT` JSON template |
| Update application | `updater.py` → `_apply_updates()` |
[...]

PR #4023, opened 2026-07-09
feat(memory): add memory tool sets

Vanzeren:
## Why                                                                                               
                                                                                                        
   The current native memory path is middleware-driven, and its main pain points are:                   
                                                                                                        
   1. **The write path is too heuristic and too hardcoded.**                                            
      The current middleware flow relies on hardcoded substring / regex-style signal detection to       
 decide whether a turn contains corrections, reinforcement, or memory-worthy content. This makes the    
 behavior brittle, language-dependent, and hard to evolve safely. This broader limitation is already    
 called out in the memory roadmap issue: [#2450](https://github.com/bytedance/deer-flow/issues/2450).   
                                                                                                        
   2. **The passive middleware path adds non-trivial overhead on every turn.**                          
      In middleware mode, DeerFlow always pays the memory write-path cost after the agent turn: message 
 filtering, queue/debounce, and background summarization/update work. This happens even when the task   
 may not benefit from memory writes at all.                                                             
                                                                                                        
   This PR introduces an alternative **tool-driven memory mode**:                                       
                                                                                                        
   - `memory.mode: middleware` keeps the existing passive flow                                          
   - `memory.mode: tool` skips `MemoryMiddleware` and exposes:                                          
     - `memory_search`                                                                                  
     - `memory_add`                                                                                     
     - `memory_update`                                                                                  
     - `memory_delete`                                                                                  
                                                                                                        
   The architectural idea is intentionally narrow:                                                      
   - keep the same **storage backend**                                                                  
   - keep the same **memory injection path**                                                            
   - keep the same **staleness review / updater backend**                                               
   - only swap the **control / write path**                                                             
                                                                                                        
   Instead of forcing memory decisions through middleware heuristics, tool mode lets the model          
 explicitly decide when to search, add, update, or delete memory.                                       
                                                                                                        
   This design is inspired by                                                                           
 [NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent), which treats memory as an   
 explicit agent capability instead of only a passive background pipeline.                               
                                                                                                        
 **Architecture**                                                                                
                                                                                                        
   ```text                                                                                              
                       ┌─── memory.mode ───┐                                                            
                       │                    │                                                           
                 "middleware"            "tool"                                                         
                       │                    │                                                           
            ┌─────────▼──────────┐   ┌─────▼──────────────┐                                             
            │ MemoryMiddleware   │   │ MemoryToolSet      │                                             
            │ (after_agent hook) │   │ (tools registered  │                                             
            │ queue → debounce   │   │  to model)         │                                             
            │ → LLM summarize    │   │                    │                                             
            └────────┬───────────┘   └────────┬───────────┘                                             
                     │                        │                                                         
            ┌────────▼────────────────────────▼───────────┐                                             
            │                Shared Backend               │                                             
            │  MemoryStorage        |   MemoryUpdater     │                                             
            │  (memory.json)        |   (CRUD/staleness)  │                                             
            └───────────────────────┬─────────────────────┘                                             
                                    │                                                                   
                          ┌─────────▼──────────┐                                                        
                          │ format_memory_for  │                                                        
                          │ _injection()       │                                                        
                          │ (system prompt)    │                                                        
                          └────────────────────┘                                                        
 ```                                                                                                    
                                                                                                        
 Current risks / limitations                                                                            
                                                                                                        
 This PR adds a new control path, but it does not fully solve retrieval quality yet.                    
                                                                                                        
 1. Tool mode depends heavily on model capability.                                                      
 If the model does not proactively call memory tools at the right time, memory quality may be worse     
 than the middleware path. Tool mode gives the model more agency, but also shifts more responsibility   
 onto model behavior.                                                                                   
 2. memory_search is still weak retrieval.                                                              
 The current implementation is only a simple substring match over stored fact content. That is enough   
 for a first tool surface, but it is not a strong retrieval layer and should not be treated as a final  
 memory-search design.                                                                                  
 3. Prompt injection is still preserved, so search is not yet a strong standalone primitive.            
 This PR intentionally keeps memory injection in both modes. That means the model may already see       
 injected memory in the prompt and therefore may not get much practical benefit from memory_search yet. 
 In other words, tool-mode search exists, but the system is still partially relying on the old          
 injection-based recall path.                                                                           
 4. A more specialized memory framework is still future work.                                           
 If DeerFlow wants high-quality query-aware retrieval, ranking, and long-term memory optimization, this 
 will likely require a more professional memory architecture than the current substring-based fact      
 search.   


## What changed

Memory now supports an opt-in **tool-driven mode** in addition to the existing middleware mode.      
                                                                                                        
   - A new config field, `memory.mode`, controls how memory works:                                      
     - `middleware` — existing default behavior                                                         
     - `tool` — experimental mode                                                                       
   - In `memory.mode: tool`, the agent gets explicit memory tools:                                      
     - `memory_search`                                                                                  
     - `memory_add`                                                                                     
     - `memory_update`                                                                                  
     - `memory_delete`                                                                                  
   - In tool mode, DeerFlow skips the passive `MemoryMiddleware` write path and lets the model decide   
 when to search or modify memory.                                                                       
   - Memory storage, prompt injection, and staleness handling remain shared between both modes.         
   - Existing users see **no behavior change by default** because `memory.mode` defaults to             
 `middleware`.                                                                                          
                                                                                                        
   Also included:                                                                                       
   - tighter tool-mode scoping so memory operations stay bound to the active user / agent context       
   - duplicate fact rejection for `memory_add`                                                          
   - safer limit handling for `memory_search`                                                           
   - docs updated to explain both modes and position tool mode as experimental 

## Surface area

## Screenshots / Recording

## Validation

### Backend                                                                                          
                                                                                                        
   ```bash                                                                                              
   cd backend && make lint                                                                              
   cd backend && uv run pytest tests/test_memory_tools.py tests/test_memory_search.py                   
   cd backend && uv run pytest tests/test_lead_agent_skills.py                                          
 tests/test_tool_args_schema_no_pydantic_warning.py                                                     
 ```                                                                                                    
                                                                                                        
 ### Config example                                                                                     
                                                                                                        
 ```bash                                                                                                
   uv run --project backend python - <<'PY'                                                             
   from pathlib import Path                                                                             
   import yaml                                                                                          
                                                                                                        
   with Path('config.example.yaml').open() as f:                                                        
       data = yaml.safe_load(f)                                                                         
                                                                                                        
   assert data['memory']['mode'] == 'middleware'                                                        
   print('memory.mode', data['memory']['mode'])                                                         
   PY                                                                                                   
 ```                                                                                                    
                                                                                                        
 ### Frontend docs checks                                                                               
                                                                                                        
 ```bash                                                                                                
   cd frontend && pnpm install --frozen-lockfile                                                        
   cd frontend && pnpm lint                                                                             
   cd frontend && pnpm typecheck                                                                        
 ```                                                                                                    
                                                                                                        
 ### Results                                                                                            
                                                                                                        
 - backend lint passed                                                                                  
 - targeted backend regression tests passed                                                             
 - frontend lint passed                                                                                 
 - frontend typecheck passed                                                                            
 - config.example.yaml parses and memory.mode defaults to middleware                                    
                                                                                                
                                                                                                       
## AI assistance

**Tool(s) used:** Codex

**How you used it:** <!-- e.g. "generated the module from a spec", "autocomplete only",
     "AI wrote tests, I wrote the impl". A prompt or conversation link is great too. --> Used Codex to implement the tool-mode memory fixes, add regression tests,       
 update the related docs, and help draft the PR description. I then reviewed the changes, refined the   
 final wording/behavior, and ran the verification commands myself.


backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -0,0 +1,223 @@
...
+
+    Returns:
+        JSON string with "fact_id" and "status": "added".
+        On duplicate content, returns "error" with explanation.
+    """
+    agent_name, user_id = _resolve_scope(runtime)
+    try:
+        normalized_content = content.strip()
+        existing_key = _memory_content_key(normalized_content)
+        existing_facts = search_memory_facts(
```
willem-bd: Duplicate-detection can miss exact duplicates. The pre-check uses `search_memory_facts(..., limit=10)` sorted by confidence desc — if more than 10 facts contain the new content as a substring and the exact-match duplicate has lower confidence (ranking past position 10), it won't be found and the duplicate gets persisted. The middleware path (`_apply_updates`) does a full-scan dedup; a direct full-scan equality check here would close the gap.

backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -0,0 +1,223 @@
...
+            normalized_content,
+            limit=10,
+            agent_name=agent_name,
+            user_id=user_id,
+        )
+        if any(_memory_content_key(str(fact.get("content", ""))) == existing_key for fact in existing_facts):
+            return json.dumps({"error": "Duplicate fact"})
+
+        updated_memory = create_memory_fact(
```
willem-bd: `max_facts` cap is unenforced in tool mode. `create_memory_fact` just appends — the `max_facts` trim (default 100) only runs in `_apply_updates` (middleware path, `updater.py:847`). So `memory_add` can grow `memory.json` past the configured cap without limit. Consider enforcing the cap in `create_memory_fact` (or in this wrapper), trimming lowest-confidence entries like the middleware does.

backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -0,0 +1,223 @@
...
+            normalized_content,
+            category=category,
+            confidence=confidence,
+            agent_name=agent_name,
+            user_id=user_id,
+        )
+        # Extract the newly created fact's id (last one in the list)
+        facts = updated_memory.get("facts", [])
+        fact_id = facts[-1]["id"] if facts else "unknown"
```
willem-bd: `fact_id` extraction is order-coupled: `facts[-1]` assumes the newly created fact is last — true today since `create_memory_fact` appends, but a future sort or insert-at-index would silently return the wrong id. Cleaner to have `create_memory_fact` return the new fact (or its id) directly, so the caller doesn't depend on list ordering.

backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -0,0 +1,223 @@
...
+        return json.dumps({"error": f"Fact not found: {fact_id}"})
+    except ValueError as exc:
+        return json.dumps({"error": str(exc)})
+    except Exception as exc:
+        logger.exception("memory_update_tool failed")
+        return json.dumps({"error": str(exc)})
+
+
+@tool("memory_delete", parse_docstring=True)
+def memory_delete_tool(runtime: Runtime, fact_id: str) -> str:
```
willem-bd, line 192: Tool mode bypasses the staleness guardrails. In middleware mode, deletions are age-gated, category-protected (`staleness_protected_categories`, default `correction`), and per-cycle capped (`staleness_max_removals_per_cycle`). `memory_delete` and `memory_update` call the raw CRUD primitive with none of these — the model can delete protected `correction` facts and unlimited facts per turn. Worth either applying the same protections here, or adding an explicit code comment + docs note so operators understand what they're opting out of when switching to tool mode.

backend/packages/harness/deerflow/agents/factory.py
```diff
@@ -242,9 +242,19 @@ def _assemble_from_features(
         if isinstance(feat.memory, AgentMiddleware):
             chain.append(feat.memory)
         else:
-            from deerflow.agents.middlewares.memory_middleware import MemoryMiddleware
+            from deerflow.config.memory_config import get_memory_config
 
-            chain.append(MemoryMiddleware(agent_name=name))
+            memory_cfg = get_memory_config()
```
willem-bd: Two config sources for the same `mode` decision. Here the factory reads the global `get_memory_config()` singleton, but `_make_lead_agent` reads `resolved_app_config.memory`. After a hot-reload of `memory.mode` these can diverge unless `set_memory_config()` is re-invoked on every reload. Both decision points should read the same source — prefer `resolved_app_config.memory`, which is the per-request hot-reload-aware view.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -492,6 +508,11 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
         final_tools, setup = assemble_deferred_tools(filtered, enabled=resolved_app_config.tool_search.enabled)
         if skill_setup.describe_skill_tool:
             final_tools.append(skill_setup.describe_skill_tool)
+        if resolved_app_config.memory.mode == "tool" and resolved_app_config.memory.enabled:
+            from deerflow.agents.memory.tools import get_memory_tools
+
+            final_tools.extend(get_memory_tools())
+        final_tools = _dedupe_tools_by_name(final_tools)
```
willem-bd: `_dedupe_tools_by_name` runs over the entire toolset, not just memory tools (same at the second call site at line 577). It silently drops any duplicate-named tool (first wins, rest logged as a warning) — a behavior change with scope beyond memory that also applies in `middleware` mode. If two legitimately distinct tools share a name (two MCP servers, or a custom `tool_groups` entry colliding with a builtin), the wrong one may be kept. Consider scoping the dedup to the memory tools, or documenting the first-wins policy.

willem-bd (commented):
Reviewing the tool-driven memory mode. Overall a clean, well-tested change — good architectural split (shared storage / injection / staleness backend), follows the existing `@tool` + `Runtime` convention, and correct user/agent scoping via `resolve_runtime_user_id`. Inline notes below on a few correctness/consistency points; the highest-impact are the `max_facts` cap gap and the staleness-guardrail asymmetry in tool mode.

willem-bd (commented):
Reviewed the tool-driven memory mode. Nice work — the architecture is conservative (only the write/control path is swapped; storage, isolation, injection, and staleness backend are shared), user-scoping via `resolve_runtime_user_id` is correct, and the `max_facts`-eviction + duplicate-rejection handling in `memory_add` is thoughtful.

Findings below are mostly nits and one effectiveness suggestion (prompt guidance). Nothing blocking from my side.

backend/packages/harness/deerflow/agents/features.py
```diff
@@ -26,6 +29,7 @@ class RuntimeFeatures:
 
     sandbox: bool | AgentMiddleware = True
     memory: bool | AgentMiddleware = False
+    memory_config: MemoryConfig | None = None
```
willem-bd, line 34: **`memory_config` appears to be dead in production.** This field is read in `factory.py` (`feat.memory_config or get_memory_config()`), but nothing in production code ever populates it — every `RuntimeFeatures(...)` construction omits it (only `test_explicit_memory_config_drives_factory_mode` sets it). So the embedded/client path always falls back to the global `get_memory_config()`, and the "explicit config wins" branch is effectively dead outside tests.

Either wire it up at the `RuntimeFeatures(...)` construction sites (`factory.py:125`, `client.py`), or drop the field and read `get_memory_config()` directly so the dataclass doesn't advertise a capability that isn't there.

config.example.yaml
```diff
@@ -1387,6 +1387,13 @@ summarization:
 # Stores user context and conversation history for personalized responses
 memory:
   enabled: true
+  # Memory operation mode:
+  #   middleware (default) - passive background extraction after each turn.
+  #   tool - experimental opt-in; the model calls memory_search/memory_add/
+  #     memory_update/memory_delete directly. This gives the model agency over
+  #     memory writes, but effectiveness depends on model tool-use behavior.
+  # Only one mode runs at a time.
+  mode: middleware
```
willem-bd, line 1396: **Schema change — bump `config_version`.** Adding `memory.mode` is a config schema change, and the repo convention is to bump `config_version` (currently 21 → 22) so the startup staleness warning stays accurate. `make config-upgrade` auto-merges the field for existing users either way, so impact is low, but the bump keeps the outdated-config signal meaningful.

backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -0,0 +1,225 @@
...
+    Args:
+        query: Natural language query to match against fact content.
+            Case-insensitive substring matching.
+        category: Optional category filter (e.g. "preference", "correction",
+            "context"). Only facts with this exact category are returned.
+        limit: Maximum results to return (default 10).
+
+    Returns:
+        JSON string with "results" (list of fact objects) and "count".
+        Each fact has id, content, category, confidence, createdAt.
```
willem-bd: **Minor: docstring vs. returned payload.** The docstring says each fact has `id, content, category, confidence, createdAt`, but `search_memory_facts` returns the full fact dict, which also includes `source`. Either add `source` to the docstring or project the returned facts to the documented fields so the model's expectations match what it actually receives.

backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -0,0 +1,225 @@
...
+
+    Returns:
+        JSON string with "fact_id" and "status": "added".
+        On duplicate content, returns "error" with explanation.
+    """
+    agent_name, user_id = _resolve_scope(runtime)
+    try:
+        normalized_content = content.strip()
+        existing_key = _memory_content_key(normalized_content)
+        existing_facts = get_memory_data(agent_name, user_id=user_id).get("facts", [])
```
willem-bd, line 116: **Minor: double read + small TOCTOU window.** The duplicate check reads storage here, then `create_memory_fact_with_created_fact` reads it again internally. Storage is cached so it's cheap, but it's redundant, and there's a theoretical window between this scan and the write where a concurrent call could insert the same fact. Fine for single-user chat; worth a comment if the writer ever broadens beyond one in-flight call per user.

backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -0,0 +1,225 @@
...
+    except KeyError:
+        return json.dumps({"error": f"Fact not found: {fact_id}"})
+    except ValueError as exc:
+        return json.dumps({"error": str(exc)})
+    except Exception as exc:
+        logger.exception("memory_delete_tool failed")
+        return json.dumps({"error": str(exc)})
+
+
+def get_memory_tools() -> list:
```
willem-bd, line 218: **Effectiveness: the model gets no prompt guidance for tool mode.** In tool mode the model's only signal that memory tools exist is these tool docstrings — the lead-agent system prompt still only describes the injected `<memory>` block, and tool mode intentionally keeps injection. So the model sees memory in its prompt *plus* four memory tools, with no instruction on when/why to call them, which directly undercuts the PR's "effectiveness depends on model behavior" caveat.

A short tool-mode-only prompt section guiding when to prefer `memory_search`/`memory_add` over relying on injected memory would materially improve quality and is low-risk. This is the single highest-leverage change for making tool mode actually useful.

backend/packages/harness/deerflow/agents/factory.py
```diff
@@ -242,9 +244,25 @@ def _assemble_from_features(
         if isinstance(feat.memory, AgentMiddleware):
             chain.append(feat.memory)
         else:
-            from deerflow.agents.middlewares.memory_middleware import MemoryMiddleware
+            from deerflow.config.memory_config import get_memory_config
+
+            memory_cfg: MemoryConfig = feat.memory_config or get_memory_config()
+            if memory_cfg.mode == "tool" and memory_cfg.enabled:
```
willem-bd: **Behavior nit: `mode: tool` + `enabled: false` silently falls back to middleware.** When `mode == "tool"` but `enabled` is false, this condition is False → `MemoryMiddleware` is appended (no-ops at runtime since disabled) and no memory tools register. Memory ends up fully off, which is defensible, but the silent switch *off* of the tool path is surprising for an operator who explicitly chose `mode: tool`. Consider logging a warning when `mode == "tool"` and `enabled` is false so the misconfiguration is visible.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -296,8 +309,9 @@ def build_middlewares(
     # Add TitleMiddleware
     middlewares.append(TitleMiddleware(app_config=resolved_app_config))
 
-    # Add MemoryMiddleware (after TitleMiddleware)
-    middlewares.append(MemoryMiddleware(agent_name=agent_name, memory_config=resolved_app_config.memory))
+    # Add MemoryMiddleware (after TitleMiddleware) — skipped in tool mode
+    if not (resolved_app_config.memory.mode == "tool" and resolved_app_config.memory.enabled):
```
willem-bd: **Duplication: the `mode == "tool" and enabled` check now lives in 4 places.** `factory.py:250`, here in `build_middlewares`, and the two `_make_lead_agent` branches at `agent.py:525` and `agent.py:590`. A small `should_use_memory_tools(memory_config)` helper would keep them from drifting and make a future fifth site hard to get wrong. (Non-blocking — the logic is correct in all four today.)

backend/packages/harness/deerflow/agents/features.py
Vanzeren: I kept `RuntimeFeatures.memory_config` but clarified its scope instead of wiring it into the         
 lead-agent production path.                                                                            
                                                                                                        
   The production lead-agent path does not use `RuntimeFeatures` as its config transport; it already    
 reads the hot-reload-aware `resolved_app_config.memory` and passes that directly into                  
 `build_middlewares()` / `MemoryMiddleware`. Routing that path back through                             
 `RuntimeFeatures(memory_config=...)` would add an extra layer without changing behavior.               
                                                                                                        
   `memory_config` is intended for the direct factory API:                                              
   `create_deerflow_agent(features=RuntimeFeatures(memory=True, memory_config=...))`.                   
                                                                                                        
   That path does not have an `AppConfig` argument, so removing the field would force direct factory    
 callers back onto the global `get_memory_config()` singleton and make isolated tests/embedded callers  
 harder to control.                                                                                     
                                                                                                        
   I added a comment on the field to make this boundary explicit:                                       
   - lead-agent AppConfig path: uses `resolved_app_config.memory`                                       
   - direct factory path: may use `RuntimeFeatures.memory_config`                                       
   - otherwise falls back to `get_memory_config()`

WillemJiang (approved):

PR #4026, opened 2026-07-10
fix(sandbox): scrub abbreviated *_PASS vars and Postgres PGPASSFILE from skill env

AnnaSuSu:
## Why

`sandbox/env_policy.py` (added in #3871 for #3861) scrubs secret-looking variables out of the environment a skill subprocess inherits, so platform credentials in the Gateway's `os.environ` cannot simply be read by skill scripts. #4018 extended it to the password variables MySQL and Redis clients read directly with no flag — `MYSQL_PWD`, `REDISCLI_AUTH`.

The same threat model leaves two gaps.

**1. The abbreviated `_PASS` form.** `_SECRET_NAME_PATTERNS` blocks `*PASSWORD*` and `*PASSWD*`, but both require the full spelling. The ubiquitous abbreviated form is missed:

```python
>>> os.environ["DB_PASS"] = "prod-db-password"
>>> "DB_PASS" in build_sandbox_env()
True
```

Verified leaking today: `DB_PASS`, `SMTP_PASS`, `MYSQL_PASS`, `REDIS_PASS`, `FTP_PASS`, `MAIL_PASS`. Unlike a connection string, the value here **is** the plaintext password, so nothing needs parsing. Secrets operators (Vault Agent, External Secrets) and `.env` conventions routinely materialize passwords under exactly these names.

**2. Postgres's file-based credential sources.** `PGPASSFILE` names a `.pgpass` file (`host:port:db:user:password`) and `PGSERVICEFILE` names a `pg_service.conf` that may carry a `password` field. libpq reads both with no flag — the psql analog of the `MYSQL_PWD` / `REDISCLI_AUTH` sources #4018 added. Neither contains `KEY` / `SECRET` / `TOKEN` / `PASSWORD` / `PASSWD` / `CREDENTIAL` / `DSN`, so no wildcard catches them and neither is in `_BLOCKED_EXACT_NAMES`.

A skill script — or the agent's own `bash` tool, whose environment comes from the same `build_sandbox_env()` — can therefore read the operator's password with `echo $DB_PASS`, or authenticate straight to the platform database with `psql -h prod-db -U app -c '...'`.

Like #4018, this is a gap in a defence-in-depth layer rather than a default-deployment vulnerability: it only bites when the Gateway process environment actually carries one of these variables.

## What changed

`*PASSWORD*` and `*PASSWD*` collapse into a single `*PASS*` pattern. It subsumes both full spellings, covers the whole abbreviated family, and catches `PGPASSFILE` — one token instead of an ever-growing exact-name list. It deliberately does **not** touch `PWD` / `OLDPWD`, which carry no `PASS` substring (this is why the original design reached for exact names: a `*PWD*` wildcard would have stripped them).

`PGSERVICEFILE` is added to `_BLOCKED_EXACT_NAMES`, since it shares no token with anything blockable by wildcard.

`*PASS*` also catches the `*_ASKPASS` credential helpers — `GIT_ASKPASS`, `SSH_ASKPASS`, `SUDO_ASKPASS`. Each names a *program whose purpose is to hand the caller a credential*, so inheriting the pointer is the same leak class as inheriting the value; a skill can simply invoke it. Scrubbing them is therefore intended rather than incidental, and all three are pinned in the blocked-names test. Incidental `PASS`-bearing names (`COMPASS_*`, `BYPASS_*`) are scrubbed too — over-scrubbing is this module's fail-safe direction, with `required-secrets` as the escape hatch.

Capability handles that carry no `PASS` substring — `KUBECONFIG`, `DOCKER_CONFIG`, `SSH_AUTH_SOCK` — remain untouched. They are a broader, non-value-bearing class and are out of scope for this fix.

From a caller's perspective, nothing about the escape hatch changes: a skill that legitimately needs one of these values declares it via `required-secrets`, the caller supplies it through `context.secrets`, and the injected value still wins over the host environment.

## Surface area

- [x] **Sandbox** — sandboxed execution (`backend/packages/harness/deerflow/sandbox/env_policy.py`)
- [x] **Default behavior change** — a skill subprocess that previously inherited `$DB_PASS` / `$PGPASSFILE` from the host no longer receives it. That inheritance is the leak this PR closes, but it is a behavior change without an opt-in, so it is flagged here: such a skill must now declare the value via `required-secrets`, exactly as it already must for `DATABASE_URL` and `MYSQL_PWD`.

`backend/AGENTS.md` is updated in the same change set to keep the documented pattern set in sync.

## Bug fix verification

- Test path: `backend/tests/test_skill_request_scoped_secrets.py::TestEnvPolicy::test_secret_like_names_are_blocked`
- Did it go red on `main` and green on this branch? **yes** — the eleven added parametrizations (`DB_PASS`, `SMTP_PASS`, `MYSQL_PASS`, `REDIS_PASS`, `FTP_PASS`, `MAIL_PASS`, `PGPASSFILE`, `PGSERVICEFILE`, `GIT_ASKPASS`, `SSH_ASKPASS`, `SUDO_ASKPASS`) each fail on `main` with `AssertionError: assert False is True`, and all pass on this branch.
- The over-scrubbing direction is pinned explicitly. `test_benign_names_are_allowed` contains no `PASS`-bearing name, so on its own it asserts nothing about this broadening; its docstring now records that as a deliberate decision, and `PWD`/`OLDPWD` are the boundary it does pin. What `*PASS*` newly catches is locked by the three `*_ASKPASS` names added to the blocked-names test — each red on `main`, green here.

## Validation

- `cd backend && make lint` — `All checks passed!`, 799 files already formatted.
- `cd backend && make test` — **6799 passed, 26 skipped, 8 failed**. The 8 failures are pre-existing and unrelated: the `web_fetch` cases in `test_browserless_client.py`, `test_crawl4ai_tools.py` and `test_fastcrw_tools.py`. I ran all eight on a clean `origin/main` worktree and they fail identically there — my environment has no outbound DNS, so the SSRF guard rejects `https://example.com`. Baseline on `main`: **6788 passed / the same 8 failed**. The delta of +11 passing is exactly the 11 parametrizations added here.
- Adjacent suites (`test_sandbox_tools_security.py`, `test_channel_user_id_env.py`, `test_aio_sandbox.py`): 162 passed.
- End-to-end through a real `bash` subprocess (`LocalSandbox.execute_command` → `build_sandbox_env` → `subprocess.Popen(env=...)`), with the credentials set in the host environment:
  - on `main`, the subprocess sees `DB_PASS=[prod-db-plaintext-pw]`, `SMTP_PASS=[smtp-plaintext-pw]`, `PGPASSFILE=[/etc/deerflow/pg/.pgpass]`, `PGSERVICEFILE=[/etc/deerflow/pg/service.conf]`;
  - on this branch all four are empty, benign `PGHOST` / `PGUSER` are still inherited, and an injected `env={"DB_PASS": ...}` still overrides the host value.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Claude Code assisted with locating the gap, writing the fix, and writing the regression tests. The contributor reviewed every line and takes responsibility for it.


backend/packages/harness/deerflow/sandbox/env_policy.py
```diff
@@ -25,8 +25,12 @@
...
     "*SECRET*",
     "*TOKEN*",
-    "*PASSWORD*",
-    "*PASSWD*",
+    # ``*PASS*`` subsumes the full ``PASSWORD``/``PASSWD`` spellings *and* the
+    # ubiquitous abbreviated form (``DB_PASS``, ``SMTP_PASS``, ``MYSQL_PASS``, ...),
+    # whose plaintext value is the password itself. It also covers ``PGPASSFILE``
+    # (libpq's ``.pgpass`` locator). It does not touch benign ``PWD``/``OLDPWD``,
+    # which carry no ``PASS`` substring.
+    "*PASS*",
```
willem-bd, line 43: probe

willem-bd, line 43: Heads-up: `*PASS*` now also scrubs `GIT_ASKPASS` and `SSH_ASKPASS`, which the PR body (and the module comment's "non-goals" framing) explicitly list as capability-handle vars **"intentionally not touched here."**

Verified:
- `GIT_ASKPASS` — was kept, now dropped
- `SSH_ASKPASS` — was kept, now dropped

Not a runtime bug — `*_ASKPASS` points at a credential-prompt *program*, not a secret, so dropping it is harmless (arguably good). But the stated intent is now factually wrong, and a future reader could be misled. Either:
- accept the broader scrub and correct the "not touched" claim in the PR body (+ pin `GIT_ASKPASS` in `test_secret_like_names_are_blocked` so the behavior is locked), or
- if `*_ASKPASS` must pass through, it can't under `*PASS*` — needs a deliberate decision, not silence.

Same incidental over-scrub hits niche benign names (`COMPASS_PATH`, `BYPASS_PROXY`) — fine, worth a one-line acknowledgement.

backend/tests/test_skill_request_scoped_secrets.py
```diff
@@ -167,6 +167,20 @@ class TestEnvPolicy:
             "MYSQL_PWD",  # read directly by mysql / libmysqlclient
             "REDISCLI_AUTH",  # read directly by redis-cli
             "REDIS_AUTH",
+            # Abbreviated ``_PASS`` password vars: value-bearing plaintext passwords
```
willem-bd, line 170: The positive direction here is well covered (8 new names, red→green ✓). The open gap is the over-scrubbing direction for `*PASS*`.

`test_benign_names_are_allowed` (line ~212) pins `PATH, HOME, SHELL, USER, LANG, LC_ALL, PWD, OLDPWD, TMPDIR, VIRTUAL_ENV, PYTHONPATH, DEERFLOW_PLAIN_VAR, NEXT_PUBLIC_BASE_URL, SERVICE_ENDPOINT` — **none contains a `PASS` substring**, so it asserts nothing about the broadening this PR makes. The "no new test needed for over-scrubbing" argument is circular: it assumes no benign var carries `PASS`, which is exactly the assumption being changed.

Low practical risk (no known benign `PASS` var DeerFlow cares about), but worth either pinning a representative benign `PASS`-bearing name that must still pass through, or adding a comment to the benign list documenting the accepted over-scrub so it's a conscious decision rather than an unnoticed one.

backend/AGENTS.md
```diff
@@ -457,7 +457,7 @@ Lets a caller pass per-request, short-lived end-user credentials (e.g. an ERP to
...
-- **Inherited-env scrub**: `execute_command` no longer leaks the Gateway's `os.environ` to skill subprocesses — `env_policy.build_sandbox_env` drops secret-looking names (`*KEY*`/`*SECRET*`/`*TOKEN*`/`*PASSWORD*`/`*CREDENTIAL*`/`*DSN*` + a connection-string denylist like `DATABASE_URL`/`REDIS_URL`/`GH_PAT`) so platform credentials never reach a skill; a skill that needs one must declare it.
+- **Inherited-env scrub**: `execute_command` no longer leaks the Gateway's `os.environ` to skill subprocesses — `env_policy.build_sandbox_env` drops secret-looking names (`*KEY*`/`*SECRET*`/`*TOKEN*`/`*PASS*`/`*CREDENTIAL*`/`*DSN*` + a connection-string denylist like `DATABASE_URL`/`REDIS_URL`/`GH_PAT`, plus no-flag credential sources like `MYSQL_PWD`/`REDISCLI_AUTH`/`PGSERVICEFILE`) so platform credentials never reach a skill; a skill that needs one must declare it.
```
willem-bd: Nit: this lists `PGSERVICEFILE` among the no-flag sources but omits its sibling `PGPASSFILE` — the other half of what this PR adds (now caught by `*PASS*`). Mentioning both keeps the pair together for readers, e.g. "…like `MYSQL_PWD`/`REDISCLI_AUTH`/`PGPASSFILE`/`PGSERVICEFILE`".

AnnaSuSu:
Thanks — all three verified, all three were right. Pushed a follow-up commit.

**1. `*_ASKPASS` (env_policy.py)**

Confirmed, and the PR body was factually wrong: I listed `GIT_ASKPASS` under "intentionally not touched" while my own pattern scrubs it. I also found a third: `SUDO_ASKPASS`. `KUBECONFIG`, `DOCKER_CONFIG` and `SSH_AUTH_SOCK` genuinely are untouched (no `PASS` substring), so only that one name in the list was wrong — but wrong is wrong.

I've taken your first option, and I'd argue slightly more strongly than "harmless": `GIT_ASKPASS` / `SSH_ASKPASS` / `SUDO_ASKPASS` each name a *program whose purpose is to hand the caller a credential*. Inheriting the pointer is the same leak class as inheriting the value — a skill can just invoke it. So scrubbing them is **intended**, not incidental collateral. The module comment now says that explicitly, and all three are pinned in `test_secret_like_names_are_blocked` (red on `main`, green here). The incidental `COMPASS_*` / `BYPASS_*` over-scrub is acknowledged in the same comment, with the rationale that over-scrubbing is this module's fail-safe direction and `required-secrets` is the escape hatch.

**2. The over-scrubbing direction was untested — and my argument for skipping it was circular**

You're right, and empirically so: none of the 14 names in `test_benign_names_are_allowed` contains a `PASS` substring, so that test passes identically before and after the broadening. It asserted nothing about the change. My "no new test needed" line assumed exactly the premise under revision.

This is the same defect you flagged on #4017 — a regression test that anchors the bound but not the thing that actually changed. I clearly didn't generalise the lesson; noted.

The fix pins the direction from both sides. `test_secret_like_names_are_blocked` now carries the three `*_ASKPASS` names, so what `*PASS*` newly catches is locked. `test_benign_names_are_allowed` gains a docstring recording that its lack of a `PASS`-bearing name is a decision, not an oversight: no benign `PASS` name is expected to pass through, and `PWD` / `OLDPWD` are the boundary it does pin.

I considered narrowing to a `*_PASS` suffix pattern instead (it would spare `GIT_ASKPASS`, `COMPASS_PATH`, `BYPASS_PROXY`), but it misses `DBPASS`-style names, needs `PGPASSFILE` back as an exact entry, and — given the reasoning in (1) — sparing the askpass helpers is the wrong outcome anyway. Happy to switch if you disagree.

**3. `AGENTS.md`**

Fixed — the line now reads `MYSQL_PWD`/`REDISCLI_AUTH`/`PGPASSFILE`/`PGSERVICEFILE`.

**Verification after the change**

- `make lint` — `All checks passed!`, 799 files already formatted.
- `make test` — 6799 passed, 26 skipped, 8 failed. Same pre-existing `web_fetch` failures (no outbound DNS here; identical on a clean `main` worktree, baseline 6788 passed). The +11 delta is exactly the 11 parametrizations this PR adds.
- The three `*_ASKPASS` names each fail on `main` (`is_blocked_env_name` returns `False`) and pass here, so they are real behaviour pins rather than no-op assertions.

The PR body is updated to drop the incorrect "not touched" claim.

WillemJiang (approved):

PR #4028, opened 2026-07-10
fix(security): html-escape fact content in memory prompt sections

sontianye:
## Summary

Raw memory fact content was injected verbatim into the staleness-review and consolidation prompt XML sections. A fact containing a literal `"` could break the `"..."` delimiter, and a closing tag like `</consolidation_candidates>` could prematurely end the XML block, both potentially confusing the model.

- Apply `html.escape()` to `content` in `_build_staleness_section` and `_build_consolidation_section`
- Also escape `cat` in the consolidation section's XML `category="..."` attribute
- 5 new tests covering: special characters in content, XML tag injection in content, and XML attribute injection via category name

## Background

Flagged by @willem-bd as a prompt-injection nit on #3996 (comment [3555513053](https://github.com/bytedance/deer-flow/pull/3996#discussion_r3555513053)), referencing the recent `neutralize_untrusted_tags` work in b36d7194. As noted there, `neutralize_untrusted_tags` only covers a predefined blocklist of framework tags and does not handle arbitrary XML like `</consolidation_candidates>` — `html.escape()` is the right primitive here.

Deliberately deferred from #3996 to keep both sections fixed together in a single clean change.

## Test plan

- [x] `uv run pytest tests/test_memory_staleness_review.py::TestBuildStalenessSection -v` — 5 tests, all pass
- [x] `uv run pytest tests/test_memory_consolidation.py::TestBuildConsolidationSection -v` — 6 tests, all pass
- [x] `uvx ruff check` — no lint errors

### DIFF
--- a/backend/packages/harness/deerflow/agents/memory/updater.py
+++ b/backend/packages/harness/deerflow/agents/memory/updater.py
@@ -675,6 +675,7 @@ class MemoryUpdater:
         # conscious accept: json.dumps escapes " and \ but not < > & — lower-risk than
         # staleness/consolidation (read-only context, not delete/merge instructions).
         # If revisited, escape at fact-content insert time rather than blanket-escaping the JSON blob.
+        # Tracked in #4044.
         prompt = MEMORY_UPDATE_PROMPT.format(
             current_memory=json.dumps(current_memory, indent=2, ensure_ascii=False),
             conversation=conversation_text,
--- a/backend/packages/harness/deerflow/agents/memory/updater.py
+++ b/backend/packages/harness/deerflow/agents/memory/updater.py
@@ -516,7 +516,7 @@ def _build_staleness_section(
     lines: list[str] = []
     for fact in stale_candidates:
         fid = fact.get("id", "?")
-        cat = str(fact.get("category", "context")).strip() or "context"
+        cat = html.escape(str(fact.get("category", "context")).strip() or "context")
         conf = fact.get("confidence", 0.0)
         created_raw = fact.get("createdAt", "")
         created_short = created_raw[:10] if isinstance(created_raw, str) and len(created_raw) >= 10 else created_raw
@@ -672,6 +672,9 @@ class MemoryUpdater:
                     max_sources=config.consolidation_max_sources,
                 )
 
+        # conscious accept: json.dumps escapes " and \ but not < > & — lower-risk than
+        # staleness/consolidation (read-only context, not delete/merge instructions).
+        # If revisited, escape at fact-content insert time rather than blanket-escaping the JSON blob.
         prompt = MEMORY_UPDATE_PROMPT.format(
             current_memory=json.dumps(current_memory, indent=2, ensure_ascii=False),
             conversation=conversation_text,
--- a/backend/packages/harness/deerflow/agents/memory/updater.py
+++ b/backend/packages/harness/deerflow/agents/memory/updater.py
@@ -4,6 +4,7 @@ import asyncio
 import atexit
 import concurrent.futures
 import copy
+import html
 import json
 import logging
 import math
@@ -519,7 +520,7 @@ def _build_staleness_section(
         conf = fact.get("confidence", 0.0)
         created_raw = fact.get("createdAt", "")
         created_short = created_raw[:10] if isinstance(created_raw, str) and len(created_raw) >= 10 else created_raw
-        content = str(fact.get("content", ""))
+        content = html.escape(str(fact.get("content", "")))
         lines.append(f'- [{fid} | {cat} | {conf:.2f} | {created_short}] "{content}"')
     return STALENESS_REVIEW_PROMPT.format(
         stale_facts="\n".join(lines),
@@ -575,10 +576,10 @@ def _build_consolidation_section(
         for fact in group[:max_sources]:
             fid = fact.get("id", "?")
             conf = _coerce_source_confidence(fact)
-            content = str(fact.get("content", ""))
+            content = html.escape(str(fact.get("content", "")))
             lines.append(f'- [{fid} | {conf:.2f}] "{content}"')
         shown = min(len(group), max_sources)
-        parts.append(f'<consolidation_candidates category="{cat}" count="{shown}">\n' + "\n".join(lines) + "\n</consolidation_candidates>")
+        parts.append(f'<consolidation_candidates category="{html.escape(cat)}" count="{shown}">\n' + "\n".join(lines) + "\n</consolidation_candidates>")
     return CONSOLIDATION_PROMPT.format(consolidation_groups="\n\n".join(parts), max_groups=max_groups)
 
 
--- a/backend/packages/harness/deerflow/agents/memory/updater.py
+++ b/backend/packages/harness/deerflow/agents/memory/updater.py
@@ -672,10 +672,8 @@ class MemoryUpdater:
                     max_sources=config.consolidation_max_sources,
                 )
 
-        # conscious accept: json.dumps escapes " and \ but not < > & — lower-risk than
-        # staleness/consolidation (read-only context, not delete/merge instructions).
-        # If revisited, escape at fact-content insert time rather than blanket-escaping the JSON blob.
-        # Tracked in #4044.
+        # conscious accept: json.dumps leaves < > & unescaped in fact content (tracked in #4044);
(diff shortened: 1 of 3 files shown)


backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -519,7 +520,7 @@ def _build_staleness_section(
         conf = fact.get("confidence", 0.0)
         created_raw = fact.get("createdAt", "")
         created_short = created_raw[:10] if isinstance(created_raw, str) and len(created_raw) >= 10 else created_raw
-        content = str(fact.get("content", ""))
+        content = html.escape(str(fact.get("content", "")))
```
willem-bd, line 523: Good — this closes the content-quote / closing-tag injection in this section.

One related surface worth a conscious decision: the always-on `MEMORY_UPDATE_PROMPT` is formatted with `current_memory=json.dumps(...)` (updater.py ~line 676), and the template wraps it in `<current_memory>{current_memory}</current_memory>`. `json.dumps` escapes `"` and `\` (so quote-based structural breaks are handled) but **not** `<`, `>`, `&` — so a fact `content` of `</current_memory>` or `<injected>…</injected>` lands verbatim inside that block on *every* update, the same class being fixed here. Note also this `current_memory` path runs unconditionally, whereas these two sections are conditionally rendered (only when staleness/consolidation are enabled and thresholds met).

It's genuinely lower severity: the `current_memory` block is for the model to *read*, whereas staleness/consolidation instruct it to *delete/merge* based on embedded content (higher stakes for injection). So leaving it is defensible risk-prioritization. Two small asks:
1. Confirm this is a conscious accept, ideally with a comment at the `current_memory=json.dumps(...)` call site so a future reader doesn't miss it or "fix" it inconsistently.
2. If it's closed later, avoid blanket-escaping the entire JSON blob (it would distort every fact on the hot path) in favor of a targeted escape at fact-content insert time.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -519,7 +520,7 @@ def _build_staleness_section(
...
         lines.append(f'- [{fid} | {cat} | {conf:.2f} | {created_short}] "{content}"')
```
willem-bd, line 524: Minor consistency point: `cat` is escaped in the consolidation section (it sits in an XML attribute) but not here. Here it's inside a bracketed list item `[id | cat | conf | date]`, so a `cat` containing `]` could close the bracket early, or `|` could shift fields. Categories are LLM-derived (semi-untrusted), so for defense-in-depth / symmetry it'd be reasonable to escape (or sanitize) `cat` here too — or at least leave a one-line note that the bracket context is intentionally treated as lower-risk than the attribute case.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -575,10 +576,10 @@ def _build_consolidation_section(
         for fact in group[:max_sources]:
             fid = fact.get("id", "?")
             conf = _coerce_source_confidence(fact)
-            content = str(fact.get("content", ""))
+            content = html.escape(str(fact.get("content", "")))
```
willem-bd, line 579: Heads-up on the readability tradeoff (not a blocker): `html.escape` also mangles *legitimate* `<`, `>`, `&`, `"`, `'` in facts — e.g. `"prefers x < y"` reaches the model as `x &lt; y`, and apostrophes/quotes become `&#x27;`/`&quot;`. Modern models usually decode these, and for these rarely-fired sections it's an acceptable security tradeoff; just worth knowing that facts about code/HTML/math get visually distorted in the prompt.

willem-bd (commented):
Reviewing the html-escaping fix for `_build_staleness_section` / `_build_consolidation_section`. The change is correct, minimal, and well-tested for what it touches — `html.escape` (default `quote=True`) is the right primitive and covers the content-delimiter, closing-tag, and attribute vectors. The test cases assert both the negative (raw break sequence absent) and positive (escaped entity present), which is nice.

Three inline notes below; none are blockers. The most substantive is the first — the same injection class still exists on the always-on `current_memory` path and is worth a conscious decision + follow-up.

One additional (non-inline) note on test scope: the plan runs only the two touched test classes + `ruff`. Since `updater.py` is shared across the whole memory pipeline, running the full `tests/test_memory*.py` set would be more reassuring for a change to a shared module.

sontianye:
Addressed in ed7c6d3:

1. **`cat` in staleness section** — applied `html.escape` to `cat` in `_build_staleness_section` for symmetry (both sections now consistently escape LLM-derived category values).

2. **`current_memory` conscious accept** — added a comment at the `json.dumps` call site documenting the deliberate decision: `json.dumps` leaves `<`/`>`/`&` unescaped, accepted because this block is read-only context for the model (not delete/merge instructions like staleness/consolidation). The comment also notes to fix at fact-content insert time rather than blanket-escaping the JSON blob, if revisited.

3. **Readability tradeoff** — understood. For these conditionally-rendered, rarely-fired sections the security benefit outweighs the visual distortion. Modern models handle HTML entities well in practice.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -671,6 +672,9 @@ def _prepare_update_prompt(
                     max_sources=config.consolidation_max_sources,
                 )
 
+        # conscious accept: json.dumps escapes " and \ but not < > & — lower-risk than
```
willem-bd: **Worth a tracking issue — this is the largest unescaped surface, not just a lower-risk one.** The deferral is sound (you can't blanket-escape the JSON blob without corrupting its structure, and the per-field escape this comment already describes is the right fix). But `content` — the field most influenced by user/conversation input — still reaches the model verbatim here, so a fact like `</current_memory><evil>...</evil>` lands in the read-only context block unescaped, even though it is now neutralized in the staleness/consolidation action sections above.

Your risk-tiering (action sections are directly destructive, so protect those first) is defensible. Just suggesting an explicit tracking issue so the per-field `html.escape` on each fact `content` before `json.dumps` eventually lands, rather than living only as a comment.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -515,11 +516,11 @@ def _build_staleness_section(
     lines: list[str] = []
     for fact in stale_candidates:
         fid = fact.get("id", "?")
-        cat = str(fact.get("category", "context")).strip() or "context"
+        cat = html.escape(str(fact.get("category", "context")).strip() or "context")
```
willem-bd, line 519: **Minor / optional: `fid` (line 518) and `created_short` (line 522) in this loop are the two interpolated fact fields still unescaped.** Both are safe by construction — id is `fact_<uuid-hex>` and createdAt is a server-set ISO timestamp — so this is genuinely low-risk and probably not worth changing.

Flagging only if you want the `every interpolated fact field is escaped` invariant to hold; the consolidation builder (line 577) has the same `fid` gap. No action expected unless that invariant is desired.

willem-bd (commented):
Reviewed the escaping fix. Correct primitive (`html.escape`) applied at the right layer (insertion time, not on stored data), soundly scoped to the attacker-derived `content` and LLM-derived `category` in both action sections, with good tests covering quotes / tag-injection-in-content / attribute-injection-in-category. Approve.

Two inline notes:
- The `current_memory` JSON blob is the largest unescaped surface (all fact `content`s) — deferral is reasonable, suggest a tracking issue.
- `fid` / `created_short` remain unescaped; safe by construction, optional defense-in-depth only.

The comment-placement nit (3-line `conscious accept` comment sits tight against the `.format()` call) is readability-only, not worth a change.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Good suggestion — opened #4044 to track the per-field escape work. Also updated the call-site comment to reference it (2fb6ddc) so a future reader lands on the issue directly.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Confirmed — `fid` is `fact_<uuid-hex>` (our own generated value) and `created_short` is a server-set ISO timestamp sliced to 10 chars, so neither is user-influenced. No intention to hold the 'every interpolated field is escaped' invariant for values that are safe by construction.

WillemJiang (approved):

PR #4033, opened 2026-07-10
fix: ensure visible response after tool runs

### LINKED ISSUE #4027: [bug] Tool runs can finish successfully without a visible final assistant response
### Before you start

### Problem summary

After tool execution, a run can end with an empty AIMessage, be persisted as successful, and leave the user with only tool steps and no visible final response.

### Affected area(s)

Frontend (UI / Next.js), Backend API (gateway / endpoints / SSE), Agents / LangGraph (graph, prompts, langgraph.json)

### What happened?

In tool-heavy conversations, especially after Playwright or shell tools return an
error, DeerFlow can finish the run without displaying a final assistant answer.

I reproduced two related cases in the same thread:

1. The model returned a terminal `AIMessage` with:

   - `content=""`
   - `tool_calls=[]`
   - `finish_reason="stop"`
   - `output_tokens=2`

   LangChain treated this as the normal agent-loop exit condition because there
   were no tool calls. DeerFlow then persisted the run with `status="success"`,
   even though `last_ai_message` was `NULL`.

2. Earlier AI messages contained both user-visible text and tool calls, for
   example a failure explanation followed by "let me try another approach".
   These messages were classified as `assistant:processing`, but their text was
   not rendered as an assistant response. The UI therefore showed only the tool
   execution card.

The final empty message was present in the LangGraph checkpoint, so this does
not appear to be checkpoint persistence loss. However, the JSON export drops
messages with empty content, which makes the exported conversation appear as
if no final assistant message existed.

Relevant code paths:

- `backend/packages/harness/deerflow/agents/lead_agent/agent.py`: DeerFlow uses
  `langchain.agents.create_agent`.
- LangChain's installed `agents/factory.py` exits the loop when
  `len(last_ai_message.tool_calls) == 0`, without checking for visible content.
- `backend/packages/harness/deerflow/runtime/runs/worker.py`: a normally
  exhausted stream is marked `success` unless it was aborted or contains a
  recognized LLM error fallback.
- `backend/packages/harness/deerflow/runtime/journal.py`: empty AI content does
  not populate `last_ai_message`.
- `frontend/src/core/messages/utils.ts`: an AI message becomes a normal
  assistant bubble only when `hasContent(message) && !hasToolCalls(message)`.
- `frontend/src/core/threads/export.ts`: empty messages are omitted from JSON
  exports.

The system prompt already states that the agent must always provide a visible
response, so this is not caused by a missing prompt instruction.

The immediate empty response may originate from the model provider, the
OpenAI-compatible gateway, or the LangChain adapter. The persisted checkpoint
only proves that DeerFlow received an empty terminal `AIMessage`. Raw provider
SSE data was unavailable because run events were configured with the in-memory
backend.

### Expected behavior

A tool run should never complete successfully without a visible terminal
assistant response.

If the model returns an empty message with no tool calls, DeerFlow should do
one of the following:

- retry once with an explicit request to summarize the tool result;
- produce a user-visible fallback explaining that the model returned no answer;
- mark the run as an error instead of success.

Text included in an AI message that also contains tool calls should either be
displayed in the processing group or otherwise remain visible to the user.

### Steps to reproduce

1. Check out `main` at commit
   `11415875c49bd5830dc8b5da4a4138ba9938c3fb`.
2. Start DeerFlow in Docker development mode with `make docker-start`.
3. Configure an OpenAI-compatible `ChatOpenAI` model with thinking enabled.
   The reproduced case used `qwen3.5-plus`.
4. Enable a tool that can return an error, such as a Playwright MCP browser tool.
5. Ask the agent to perform a browser action.
6. Let the tool return an error, then allow the agent to process the tool result.
7. Repeat the tool turn if necessary. This occurred twice in one conversation.
8. Observe that the run stops and the UI shows only tool-processing steps.
9. Inspect the latest checkpoint and run row. The final message can have empty
   content and no tool calls while the run status is `success`.
10. Export the thread as JSON. The empty final AI message is absent from the
    export.

### Relevant logs

```shell
Run A:
status=success
model_name=qwen35-plus
llm_call_count=2
last_ai_message=NULL
total_input_tokens=40151
total_output_tokens=79

Final checkpoint message:
type=AIMessage
content=""
tool_calls=[]
finish_reason=stop
input_tokens=20132
output_tokens=2
reasoning_tokens=0

Run B:
status=success
model_name=qwen35-plus
llm_call_count=8

Final checkpoint message:
type=AIMessage
content=""
tool_calls=[]
finish_reason=stop
input_tokens=21204
output_tokens=2
reasoning_tokens=0
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

Windows x86_64, PowerShell, Docker Desktop with Linux containers

### Python version

Python 3.12 container image

### Node.js version

Node.js 22 container image

### pnpm version

_No response_

### uv version

_No response_

### Git state

branch: main
commit: 11415875c49bd5830dc8b5da4a4138ba9938c3fb

### Support bundle summary

_No response_

### Additional context

The issue is not specific to Playwright. Playwright errors made it easier to
reproduce because they create multiple model/tool cycles, but the missing
completion invariant applies to any tool.

A screenshot and a redacted JSON conversation export are available. The JSON
export preserves intermediate AI text but omits the terminal empty AIMessage,
while the UI hides intermediate text when the same message also contains tool
calls.

[使用 Playwright 访问 Docker 内部主机.json](https://github.com/user-attachments/files/29874151/Playwright.Docker.json)

Image

Huixin615:
Fixes #4027

## Why

A tool-heavy agent run can finish with an empty terminal `AIMessage` containing
neither visible content nor tool calls. LangChain treats the absence of tool
calls as the normal agent-loop exit condition, so DeerFlow previously persisted
the run as successful even though the user received no visible final assistant
response.

This is especially visible after tool failures, where the UI can show only the
tool-processing steps while the final empty assistant message remains hidden.

## What changed

- Added a lead-agent terminal-response middleware that detects empty terminal
  responses after tool execution.
- Retry the model once with a hidden recovery prompt asking it to summarize the
  existing tool results into a visible final response.
- If the retry is also empty, replace the empty message with a visible fallback
  and mark it with DeerFlow's existing LLM error-fallback metadata.
- Reuse the run worker's existing fallback detection so the degraded run is
  persisted as an error instead of a silent success.
- Keep retry state isolated by thread and run, and clean it up when the agent
  finishes.
- Preserve normal text responses, tool routing, invalid tool-call handling,
  clarification flows, and safety-termination behavior.
- Added graph-level sync and async regression coverage using a real
  `create_agent` loop.
- Updated the runtime middleware documentation and user-facing behavior
  description.

## Surface area

## Screenshots / Recording

Not applicable. This change affects the lead-agent runtime and does not modify
the frontend UI.

## Bug fix verification

  `backend/tests/test_terminal_response_middleware.py`
- Did it go red on `main` and green on this branch? **Yes.** Without the new
  terminal-response guard, LangChain exits after the first empty no-tool-call
  `AIMessage`; the graph-level regression tests require one retry followed by
  either a visible model response or a visible error fallback.
- The tests exercise the real `langchain.agents.create_agent` routing path,
  including synchronous and asynchronous execution, instead of testing only a
  helper function.

## Validation

- `cd backend && UV_CACHE_DIR=/tmp/uv-cache make lint`
- Targeted middleware/runtime suite:
  `146 passed`
- Full backend suite:
  `6369 passed, 26 skipped`; 12 external-network tests initially failed under
  restricted DNS/LLM access.
- Re-ran the affected Browserless, live-client, Crawl4AI, and FastCRW files with
  network access:
  `78 passed, 1 skipped`
- `ruff format --check`: passed
- `git diff --check`: passed


WillemJiang:
@Huixin615  thanks for the contribution, please resolve the conflict error with main branch.

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,194 @@
...
+
+        key = self._key(runtime)
+        with self._lock:
+            retry_count = self._retry_counts.get(key, 0)
+            if retry_count == 0:
+                self._retry_counts[key] = 1
+                self._pending_prompts.add(key)
+
+        if retry_count == 0:
+            return {"jump_to": "model"}
```
willem-bd: This is the success-retry branch, and it has an asymmetry with the fallback branch below: returning `{"jump_to": "model"}` re-runs the model, which appends a *new* AIMessage (new `id`), while the empty message from this turn stays in `messages`. The fallback branch (~line 128) avoids this by `model_copy`-ing the empty message (same `id`) so the reducer replaces it in place — but nothing removes the empty message on the success path.

So a recovered run persists `[..., ToolMessage, AIMessage(empty), AIMessage(content)]`. The test only asserts `result["messages"][-1]`, so this isn't caught. Two effects: (1) if the frontend doesn't filter empty content, the user sees a blank assistant bubble between the tool step and the answer — a variant of the bug this fixes; (2) the empty message pollutes context on later turns.

Could strip `messages[-2]` when the post-retry `after_model` sees visible content in `messages[-1]` (and the prior message is the empty one this middleware created), or remember the empty message's `id` in run state and remove it on success. Worth confirming the frontend's empty-message rendering either way.

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def _apply(self, state: AgentState, runtime: Runtime) -> dict[str, Any] | None:
+        messages = list(state.get("messages") or [])
+        if not messages or not isinstance(messages[-1], AIMessage):
+            return None
+
+        last = messages[-1]
+        if _has_visible_content(last) or _has_tool_call_intent_or_error(last):
+            return None
+        if not _tool_result_in_current_turn(messages):
```
willem-bd, line 116: This gate scopes the guard to post-tool responses, so an empty *first* model response (no tools run) still exits via the no-tool-call router as a silent success — the same class as #4027, just not the tool case. If the intent is "no empty terminal response ever persists as success," this leaves a gap; if it's specifically the post-tool regression, a one-line comment here stating the scope would prevent future confusion.

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,194 @@
...
+            return None
+
+        key = self._key(runtime)
+        with self._lock:
+            retry_count = self._retry_counts.get(key, 0)
+            if retry_count == 0:
+                self._retry_counts[key] = 1
```
willem-bd, line 126: Because the counter is keyed by run and only reset in `after_agent`, the budget is one recovery per run, not per empty occurrence: a run that goes empty → retry → tool call → empty again hits the fallback on the second empty without a fresh recovery attempt. That's a reasonable cost bound, but it's slightly surprising given the recovery prompt says "do not call another tool unless strictly necessary." Worth a comment documenting that the budget is per-run.

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,194 @@
...
+    @override
+    async def awrap_model_call(
+        self,
+        request: ModelRequest,
+        handler: Callable[[ModelRequest], Awaitable[ModelResponse]],
+    ) -> ModelCallResult:
+        return await handler(self._augment_request(request))
+
+    @override
+    def after_agent(self, state: AgentState, runtime: Runtime) -> dict | None:
```
willem-bd, line 207: `after_agent` cleanup only runs when the agent finishes. If a run sets retry state and then `ClarificationMiddleware` interrupts (`Command(goto=END)`), `after_agent` may not fire, leaving `retry_count=1` for that `(thread, run)`. If the same `run_id` is later resumed (rather than spawned as a new run), `before_agent`'s `_clear_other_runs` won't clear it (it targets other `run_id`s), so a later empty could fall straight to fallback. Low probability (clarification usually fires before tool execution), but consider clearing current-run state in `before_agent` too, or resetting the counter when a non-empty/tool-call message is observed.

willem-bd (commented):
Reviewing `TerminalResponseMiddleware` (fixes #4027). The fix is correctly layered: it detects empty post-tool terminal AIMessages in `after_model`, retries the model once with a hidden recovery prompt, and on a second empty replaces the message in-place with a visible fallback marked `deerflow_error_fallback` — which the existing run-worker detection (`_extract_llm_error_fallback_message` → `error_reason`) picks up to persist the run as an error. I verified the persistence contract (the in-place replace keeps the same message `id`, so it isn't masked by the worker's `pre_existing_ids` filter), the `after_model` ordering (dispatch is reverse-registration, so Safety's visible explanation is added before this middleware sees the message — no misfire), the bounded retry, and the langchain/langgraph version floor (1.2.15 / 1.1.9 support `hook_config`/`jump_to`/`Runtime.context` — the codebase's first use of `jump_to`).

Four inline notes below, none blocking. The main one is #1 (the success-retry path leaves an empty intermediate message in history).

A few minor/style notes (not inline):
- `_FALLBACK_CONTENT` is hardcoded English. This follows the existing `SafetyFinishReasonMiddleware` convention, so it's consistent — flagging only because DeerFlow ships localized READMEs and this is user-facing.
- `_key`'s fallback collapses all context-less runs to one bucket (`str(id(runtime))`); fine as a defensive fallback since production always provides context, but a one-line comment would help.
- `_tool_result_in_current_turn`: when there's no real (non-hidden) user message, `latest_user_index == -1` makes the slice `messages[0:]`, so *any* ToolMessage satisfies it — an unlikely but loose edge; returning `False` when `latest_user_index == -1` would be stricter.
- `threading.Lock` around instant dict ops in the async path is fine (no I/O); no change needed.

Verdict: approve with minor follow-ups. Recommend addressing #1 (decide whether to strip the empty intermediate message, or confirm the frontend hides it) before or shortly after merge.

Huixin615:
@willem-bd Thanks for your review! I have updated the branch, could you please take another look when you have time? Thanks!

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,212 @@
...
+    return any(isinstance(message, ToolMessage) for message in messages[latest_user_index + 1 :])
+
+
+class TerminalResponseMiddleware(AgentMiddleware[AgentState]):
+    """Retry one empty post-tool response, then persist a visible error fallback."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._lock = threading.Lock()
+        self._retry_counts: dict[tuple[str, str], int] = {}
```
willem-bd: Consider BoundedDict (from deerflow.agents.middlewares._bounded_dict) for _retry_counts and _pending_prompts, matching LoopDetectionMiddleware / TokenBudgetMiddleware.

In the Gateway path the middleware is instantiated per-run so there's no leak, but in the embedded-client / TUI path (client.py) the middleware instance is a singleton reused across runs. _clear_other_runs only sweeps entries for the current thread_id, so entries left by runs on abandoned threads (crash/cancel skipping after_agent) accumulate without bound. BoundedDict(maxsize=1000) is the established safety net.

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,212 @@
...
+        self._pending_prompts: set[tuple[str, str]] = set()
+
+    @staticmethod
+    def _key(runtime: Runtime) -> tuple[str, str]:
+        context = getattr(runtime, "context", None)
+        if isinstance(context, dict):
+            thread_id = str(context.get("thread_id") or "unknown-thread")
+            run_id = str(context.get("run_id") or context.get("run_attempt_id") or id(runtime))
```
willem-bd, line 88: Heads up that run_id isn't always present in Runtime.context: the embedded client (client.py) and the compaction path (context_compaction.py) build context with only thread_id. The id(runtime) fallback handles it (same trick TokenBudgetMiddleware uses), but id() can be recycled after GC — mitigated by _clear always running in before_agent. An end-to-end test invoking with only thread_id would lock this in.

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,212 @@
...
+def _tool_result_in_current_turn(messages: list[Any]) -> bool:
+    """Return whether a tool result follows the latest real user message."""
+    latest_user_index = -1
+    for index, message in enumerate(messages):
+        if not isinstance(message, HumanMessage):
+            continue
+        if (message.additional_kwargs or {}).get("hide_from_ui"):
+            continue
+        latest_user_index = index
+    # This guard intentionally covers the post-tool regression from #4027,
```
willem-bd: Known gap: scheduled/non-interactive runs (launched internally, no real user message) and compaction runs can still end in a silent-success empty terminal after tools, since latest_user_index == -1 short-circuits here. The scoping is reasonable for #4027, but a one-line note in the PR description (or a TODO) would keep it from being rediscovered later.

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,212 @@
...
+
+        last = messages[-1]
+        if _has_visible_content(last) or _has_tool_call_intent_or_error(last):
+            return None
+        if not _tool_result_in_current_turn(messages):
+            return None
+
+        key = self._key(runtime)
+        with self._lock:
+            # The recovery budget is once per run, not once per empty message.
```
willem-bd, line 121: Worth confirming/documenting: _retry_counts is never reset after a successful recovery, so a run like tool -> empty (retry) -> visible + tool_call -> tool -> empty yields a fallback with no second retry. The loop-prevention rationale is sound, but this is the most surprising behavior and isn't directly tested. A test for that boundary (or resetting the counter on a non-empty response) would make the intent explicit.

willem-bd (commented):
Reviewed against the run worker's fallback contract and the langchain 1.2.15 middleware lifecycle. The design is sound: it correctly reuses the deerflow_error_fallback gate convention, mirrors TodoListMiddleware for the jump_to=model + wrap_model_call recovery injection, keeps the recovery prompt transient (verified absent from final state), and bounds the retry so it can't loop. Graph-level sync+async tests are great.

Inline findings are all minor — the one I'd most like addressed before merge is switching the state dicts to BoundedDict for consistency with the other guard middlewares and singleton-path leak safety. The rest are nits / doc / optional test additions.

backend/packages/harness/deerflow/agents/middlewares/terminal_response_middleware.py
```diff
@@ -0,0 +1,212 @@
...
+            # in checkpoint history or future model context.
+            message_updates = [RemoveMessage(id=last.id)] if last.id else []
+            return {"messages": message_updates, "jump_to": "model"}
+
+        additional_kwargs = dict(last.additional_kwargs or {})
+        additional_kwargs.update(
+            {
+                "deerflow_error_fallback": True,
+                "error_reason": "Model returned an empty terminal response after one retry",
+                "empty_terminal_response": True,
```
willem-bd: Minor: empty_terminal_response isn't read by any consumer today — the run worker gates purely on deerflow_error_fallback (worker.py:_try_extract_from_message), with error_reason supplying the message text. Fine as diagnostic metadata, but if nothing consumes it, consider dropping it or wiring it into a reader.

README.md
```diff
@@ -615,6 +615,10 @@ Use it as-is. Or tear it apart and make it yours.
 
 ## Core Features
 
+Tool-using runs are guarded against empty provider completions: DeerFlow retries
```
willem-bd: Nit: this sits under '## Core Features', which reads as a capabilities list (Skills, Tools, Memory...). 'Retries an empty final response once' is a reliability detail that feels slightly out of place at this altitude — might fit better lower down or just in the backend guide (which was already updated).

backend/tests/test_lead_agent_model_resolution.py
```diff
@@ -373,9 +373,17 @@ def test_build_middlewares_uses_resolved_model_name_for_vision(monkeypatch):
...
-    #   ..., custom, SafetyFinishReasonMiddleware, ClarificationMiddleware
-    # so the custom mock sits at index [-3].
-    assert len(middlewares) > 0 and isinstance(middlewares[-3], MagicMock)
+    #   ..., custom, TerminalResponseMiddleware, SafetyFinishReasonMiddleware,
+    #   ClarificationMiddleware, so the custom mock sits at index [-4].
+    assert len(middlewares) > 0 and isinstance(middlewares[-4], MagicMock)
+
+    from deerflow.agents.middlewares.clarification_middleware import ClarificationMiddleware
+    from deerflow.agents.middlewares.safety_finish_reason_middleware import SafetyFinishReasonMiddleware
+    from deerflow.agents.middlewares.terminal_response_middleware import TerminalResponseMiddleware
```
willem-bd, line 382: Nit: the comment 'custom mock sits at index [-4]' assumes SafetyFinishReasonMiddleware is enabled (it's optional, gated on safety_config.enabled). It holds for this test's config, but reads as a general invariant.

WillemJiang (approved):

PR #4034, opened 2026-07-10
fix(memory): coerce stored confidence in the three remaining raw reads

AnnaSuSu:
> **Rebased onto `938391c1`.** #4023 (`feat(memory): add memory tool sets`) landed while this was open and reshaped one of the three sites — see [Rebase](#rebase-onto-4023) below. The site *count* is still three, but the membership changed. Everything below is re-measured against the new base, not carried over.

## Why

#3996 added `_coerce_source_confidence` to `agents/memory/updater.py`, documenting exactly why a stored `confidence` cannot be trusted:

> *"This helper guards against null, bool, non-numeric, and non-finite values from corrupted or manually edited memory files."*

`memory.json` is user-editable (served and reloaded through `/api/memory`) and has been written by more than one version of the extractor, so a non-float value is reachable — that is the premise the helper was written on. Consolidation routes every stored read through it. **Three reads of a stored fact's confidence in the same file are still raw:**

| Site | Code | Raises |
|---|---|---|
| `_build_staleness_section` | `f"{conf:.2f}"` on `fact.get("confidence", 0.0)` | `ValueError` (str) / `TypeError` (None) |
| `_apply_updates`, staleness cap | `sort(key=lambda f: f.get("confidence", 0))` | `TypeError: '<' not supported between 'str' and 'float'` |
| `search_memory_facts` (`memory_search` tool) | `sort(key=lambda f: f.get("confidence", 0), reverse=True)` | same |

`_do_update_memory_sync` wraps the whole cycle in `except Exception: return False`. So a raise at either of the first two:

1. aborts the **entire** memory-update cycle — no facts extracted, no context updated, no staleness pruning, no consolidation;
2. surfaces only as a generic `Memory update failed` log line;
3. in the `_build_staleness_section` case, happens **before** `_get_model()`, so the run never even reaches the LLM;
4. leaves the offending fact on disk, because the update that would have rewritten `memory.json` is the one that just failed.

Every subsequent cycle re-reads the same fact and aborts again. Memory updates stop permanently for that user, with `staleness_review_enabled` defaulting to `true`.

The third site is not on that path — `memory_search` is a model-facing tool, so a str confidence fails the tool call instead.

### The two sorts also mis-rank *without* raising

`str` and `None` raise. `true` / `false` / `inf` / `nan` / an absent key do **not** — `bool` subclasses `int`, and `inf`/`nan` compare against floats without error. Under the raw key each lands at an *end* of the order, so it decides which fact the cap deletes and which result `memory_search` returns first:

| stored `confidence` | old rank | new rank | effect on a neighbouring genuine fact |
|---|---|---|---|
| key absent, `false` | `0` — worst | `0.5` | deleted itself ahead of a genuine `0.1` |
| `true` | `1.0` — best | `0.5` | **deleted a genuine `0.9`** |
| `inf` | `+inf` — best | `0.5` | **deleted a genuine `1.0`** |
| `nan` | undefined (compares false against everything) | `0.5` | outcome depended on the fact's position in the file |

So the two sorts carry a second, quieter defect than the crash: **silent loss of a genuine high-confidence fact**, with no exception and no log line — and, on `memory_search`, the wrong facts handed to the model within `limit`. Both reproduced end-to-end below. This is what makes `0.5` load-bearing rather than cosmetic — it is the only rank that is neither best nor worst. The two sorts run in **opposite directions** (the cap deletes the lowest, the search returns the highest), so the default has to hold from both ends.

## What changed

All three reads now go through `_coerce_source_confidence`, which is defined in this file — no new import, three lines.

`_coerce_source_confidence` rather than `prompt.py`'s `_coerce_confidence`, because **0.5 is already this module's default for an unknown confidence** (`create_memory_fact(confidence=0.5)`, `_normalize_memory_update_fact`'s `fact.get("confidence", 0.5)`, and the helper's own docstring). The `0.0` / `0` sentinels appear only in the buggy reads. That default is not cosmetic: the staleness cap **removes the lowest-confidence facts first**, so ranking an unreadable confidence at `0.0` makes it the first fact deleted. `0.5` — unknown — is the honest rank. It also treats a JSON `true` as corrupt data (0.5) rather than as full confidence (`_coerce_confidence` would yield `1.00`, since `bool` subclasses `int`).

Reads of **LLM-produced** data are deliberately left alone (`_normalize_memory_update_fact`, `_normalize_memory_update_data`, the `newFacts` threshold check, the consolidated-confidence cap). Different trust boundary, and those values already pass through `_normalize_memory_update_fact`, which drops unparseable entries — verified, not assumed. Every read of the `confidence` field under `agents/memory/` was enumerated to confirm nothing else is left raw; `prompt.py` already routes all of its reads through `_coerce_confidence`.

Because this changes a mapping and not just a crash, the full behaviour delta on the staleness prompt is:

| stored `confidence` | before | after |
|---|---|---|
| `"0.9"` / `"high"` / `""` / `None` | raises | `0.90` / `0.50` / `0.50` / `0.50` |
| key absent | `0.00` | `0.50` |
| `True` / `False` | `1.00` / `0.00` | `0.50` / `0.50` |
| `1.5` / `-0.3` | `1.50` / `-0.30` | `1.00` / `0.00` |
| `inf` / `nan` | `inf` / `nan` | `0.50` / `0.50` |
| valid float, `int` | unchanged | unchanged |

## Rebase onto #4023

#4023 refactored `_apply_updates`, extracting the `max_facts` trim into `_trim_facts_to_max()` — and wrote that helper with `key=_coerce_source_confidence`. That was one of this PR's original three sites, so **that hunk is dropped: upstream already fixed it.** The same PR introduced `search_memory_facts` with a fresh raw read of the same field, which is **added here** in its place. Net: still three sites, different membership. The two staleness sites are untouched by #4023 and unchanged from the original submission.

The three tests that anchored the `max_facts` trim now pass on `main` without this patch — they guard nothing — so they are not kept as-is. The same delta matrix is repointed at `search_memory_facts`, where it goes red (verified per-test below).

The earlier #4028 / #3993 rebases mentioned in this thread are both merged and no longer relevant.

## Surface area

- [x] **Agents / LangGraph** — memory updater prompt section, apply-time ranking, and the `memory_search` tool (`backend/packages/harness/deerflow/agents/memory/updater.py`)
- [x] **Default behavior change** — a stored confidence outside `[0, 1]`, non-finite, boolean, or absent now normalises to the module's `0.5` / clamped value instead of being printed and ranked raw. This changes which facts the staleness cap keeps and which order `memory_search` returns. It is the intended effect (see above) and is pinned by tests, but it is a change without an opt-in, so it is flagged here.

No documentation change is needed: `backend/AGENTS.md` makes no claim about the trustworthiness or domain of the stored `confidence` field.

## Bug fix verification

Both directions are anchored, at all three sites.

- **Crash direction**
  - `tests/test_memory_staleness_review.py::TestBuildStalenessSection::test_non_float_confidence_does_not_raise` — `"0.9"`, `None`, `"high"`, `""`.
  - `tests/test_memory_staleness_review.py::TestApplyUpdatesStaleness::test_safety_cap_sort_survives_non_float_stored_confidence`.
  - `tests/test_memory_updater.py::test_search_memory_facts_sort_survives_non_float_stored_confidence`.
- **Mapping direction (prompt rendering)** — `...::TestBuildStalenessSection::test_confidence_is_normalised_like_every_other_stored_read` and `::test_missing_confidence_key_renders_unknown_not_zero`. These values format fine on `main` (they print raw), so without them the behaviour change would ship unnoticed.
- **Ranking direction (both sorts)** — the crash tests above use a `str`, which *raises* on `main`; they therefore cannot go red for any value that mis-ranks silently. Added, one per sort site:
  - `...::TestApplyUpdatesStaleness::test_safety_cap_ranks_unusable_confidence_as_unknown` and `...::test_search_memory_facts_ranks_unusable_confidence_as_unknown` — parametrized over the four inputs that fall to the `0.5` default without raising (absent, `false`, `true`, `inf`), each pitted against a genuine neighbour chosen so the survivor *flips*. The two sorts run in opposite directions, so the one matrix covers eviction from both ends.
  - `...::TestApplyUpdatesStaleness::test_safety_cap_with_nan_confidence_is_order_independent` and `...::test_search_memory_facts_with_nan_confidence_is_order_independent` — `nan` is excluded from the parametrize above on purpose: under the raw key its rank is not an end of the order but *undefined*, so a single fixed input order happens to produce the correct survivor. Pinning `nan` by outcome would have been a test that cannot go red. The property it actually has is order-independence, so that is what is asserted, across both fact orders.
- **Does every new test actually go red?** Measured per-test, not assumed. Reverting the three fixed lines back to the raw key on this branch and re-running the new tests:

  ```
  $ pytest tests/test_memory_updater.py tests/test_memory_staleness_review.py -k "confidence or normalised or unknown"
  25 failed, 5 passed, 95 deselected
  ```

  The 25 failures are **exactly** the 25 test ids this PR adds (6 on `search_memory_facts`, 19 on the staleness sites); the 5 that pass are pre-existing tests caught by the keyword filter. Restoring the fix: `149 passed` across `test_memory_updater.py`, `test_memory_staleness_review.py`, and `test_memory_tools.py`.

## Validation

- `ruff format --check` — 3 files already formatted. `ruff check` — All checks passed!
- `pytest tests/ --ignore=tests/blocking_io` — **6940 passed, 26 skipped, 8 failed**. The 8 are pre-existing `web_fetch` cases (`test_browserless_client.py`, `test_crawl4ai_tools.py`, `test_fastcrw_tools.py`) — this machine has no outbound DNS, so the SSRF guard rejects `https://example.com`. Running those three files on a pristine `origin/main` worktree gives **the same 8 failed, 52 passed**, so the failure set is unchanged in both directions.
- `git merge-tree origin/main HEAD` — no conflict.

### End-to-end, against a real on-disk `memory.json`

Real cycle, only the LLM stubbed: `_do_update_memory_sync` → `_prepare_update_prompt` → real `FileMemoryStorage` read → `_build_staleness_section` → `create_chat_model` → `_apply_updates` → real `FileMemoryStorage` write. A probe records whether the model was ever constructed.

**A. Crash path.** Three aged facts, one with `"confidence": "0.9"`:

| | `main` | this branch |
|---|---|---|
| cycle #1 / #2 `_do_update_memory_sync()` | `False` / `False` | `True` / `True` |
| LLM model constructed | **`False`** | `True` |
| `workContext` learned | `''` | `'backend eng'` |
| facts on disk | 3 | 4 |

`LLM model constructed: False` on `main` is the point: the crash precedes `_get_model()`. Cycle #2 reproduces cycle #1 exactly — the offending fact is still on disk, so the failure is self-perpetuating.

**B. Silent-loss path (staleness cap).** Two aged facts — `fact_corrupt` with a hand-edited `"confidence": true`, `fact_good` with `0.9` — the LLM proposes removing both, `staleness_max_removals_per_cycle=1` forces the cap to rank them. Nothing raises on either side; the cycle returns `True` and rewrites the file both times:

| | `main` | this branch |
|---|---|---|
| cycle returned | `True` | `True` |
| fact left on disk | **`fact_corrupt`** | `fact_good` |

On `main` the genuine `0.9` fact is the one deleted, silently. That is the defect the `str`-based tests could never have caught.

**C. `memory_search`.** Same two facts, real storage, `limit=1`:

| | `main` | this branch |
|---|---|---|
| top result | **`fact_corrupt`** | `fact_good` |

The model asks for the best-ranked fact and gets the corrupt one.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Claude Code assisted with enumerating every read of the stored `confidence` field, choosing the helper, writing the fix, and writing the regression tests. The contributor reviewed every line and takes responsibility for it.

### DIFF
--- a/backend/packages/harness/deerflow/agents/memory/updater.py
+++ b/backend/packages/harness/deerflow/agents/memory/updater.py
@@ -239,7 +239,7 @@ def search_memory_facts(
             continue
         matched.append(fact)
 
-    matched.sort(key=lambda f: f.get("confidence", 0), reverse=True)
+    matched.sort(key=_coerce_source_confidence, reverse=True)
     return matched[:limit]
 
 
@@ -592,7 +592,7 @@ def _build_staleness_section(
     for fact in stale_candidates:
         fid = fact.get("id", "?")
         cat = html.escape(str(fact.get("category", "context")).strip() or "context")
-        conf = fact.get("confidence", 0.0)
+        conf = _coerce_source_confidence(fact)
         created_raw = fact.get("createdAt", "")
         created_short = created_raw[:10] if isinstance(created_raw, str) and len(created_raw) >= 10 else created_raw
         content = html.escape(str(fact.get("content", "")))
@@ -1038,7 +1038,7 @@ class MemoryUpdater:
                 max_stale = config.staleness_max_removals_per_cycle
                 if len(stale_ids_to_remove) > max_stale:
                     stale_facts = [f for f in current_memory.get("facts", []) if f.get("id") in stale_ids_to_remove]
-                    stale_facts.sort(key=lambda f: f.get("confidence", 0))
+                    stale_facts.sort(key=_coerce_source_confidence)
                     stale_ids_to_remove = {f["id"] for f in stale_facts[:max_stale]}
 
                 current_memory["facts"] = [f for f in current_memory.get("facts", []) if f.get("id") not in stale_ids_to_remove]
--- a/backend/tests/test_memory_staleness_review.py
+++ b/backend/tests/test_memory_staleness_review.py
@@ -12,6 +12,8 @@ Covers:
 from datetime import UTC, datetime, timedelta
 from unittest.mock import MagicMock, patch
 
+import pytest
+
 from deerflow.agents.memory.updater import (
     MemoryUpdater,
     _build_staleness_section,
@@ -22,6 +24,10 @@ from deerflow.agents.memory.updater import (
 from deerflow.config.memory_config import MemoryConfig
 
 # ── Helpers ────────────────────────────────────────────────────────────────
+
+
+_ABSENT = object()
+"""Sentinel: the fact carries no ``confidence`` key at all."""
 
 
 def _memory_config(**overrides: object) -> MemoryConfig:
@@ -244,6 +250,65 @@ class TestBuildStalenessSection:
         assert 'pref<"erences>' not in section
         assert "pref&lt;&quot;erences&gt;" in section
 
+    @pytest.mark.parametrize("stored_confidence", ["0.9", None, "high", ""])
+    def test_non_float_confidence_does_not_raise(self, stored_confidence):
+        """A stored ``confidence`` that is not a float must not abort the update.
+
+        ``memory.json`` is user-editable and written across versions, which is why
+        ``_coerce_source_confidence`` exists. Formatting it raw raises ValueError on
+        a str and TypeError on None; ``_do_update_memory_sync``'s ``except Exception``
+        turns that into a silent ``return False`` that aborts the whole memory-update
+        cycle -- permanently, since the offending fact is then never rewritten.
+        """
+        fact = _make_fact("fact_x", "Some fact", "knowledge", 0.9, days_ago=120)
+        fact["confidence"] = stored_confidence
+
+        section = _build_staleness_section([fact], 90)
+
+        assert "fact_x" in section
+        assert "Some fact" in section
+
+    @pytest.mark.parametrize(
+        ("stored_confidence", "rendered"),
+        [
+            ("0.9", "0.90"),
+            ("high", "0.50"),
+            (None, "0.50"),
+            (True, "0.50"),
+            (1.5, "1.00"),
+            (-0.3, "0.00"),
+            (float("inf"), "0.50"),
+            (float("nan"), "0.50"),
+        ],
+    )
+    def test_confidence_is_normalised_like_every_other_stored_read(self, stored_confidence, rendered):
+        """Pins the mapping, not just the absence of a crash.
+
+        ``0.5`` is this module's default for an unknown confidence
+        (``create_memory_fact``, ``_normalize_memory_update_fact``,
+        ``_coerce_source_confidence``). Before this change the staleness prompt
+        rendered ``1.5`` as ``1.50``, ``inf`` as ``inf``, and a ``True`` as ``1.00``
+        -- disagreeing with the consolidation prompt, which reads the same field
+        through the same helper.
+        """
+        fact = _make_fact("fact_x", "Some fact", "knowledge", 0.9, days_ago=120)
+        fact["confidence"] = stored_confidence
+
+        section = _build_staleness_section([fact], 90)
+
+        assert f"| {rendered} |" in section
+
+    def test_missing_confidence_key_renders_unknown_not_zero(self):
+        """An absent key is *unknown* (0.50), not *worthless* (0.00).
+
+        The staleness cap removes the lowest-confidence facts first, so ranking an
+        unreadable confidence at 0.00 would make that fact the first one deleted.
+        """
+        fact = _make_fact("fact_x", "Some fact", "knowledge", 0.9, days_ago=120)
+        del fact["confidence"]
+
+        assert "| 0.50 |" in _build_staleness_section([fact], 90)
+
 
 # ── _apply_updates with staleness removals ─────────────────────────────────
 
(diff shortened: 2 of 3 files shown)


backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -516,7 +516,7 @@ def _build_staleness_section(
     for fact in stale_candidates:
         fid = fact.get("id", "?")
         cat = str(fact.get("category", "context")).strip() or "context"
-        conf = fact.get("confidence", 0.0)
+        conf = _coerce_source_confidence(fact)
```
willem-bd, line 595: Merge note: #4028's `_build_staleness_section` hunk html-escapes `content` and carries this line as an unchanged context neighbor, so expect a textual conflict here. (Already flagged in the PR body — recording it at the conflict point so it isn't lost at merge time.) Happy either way on the rebase order.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -978,7 +978,7 @@ def _apply_updates(
             # Sort by confidence and keep top ones
             current_memory["facts"] = sorted(
                 current_memory["facts"],
-                key=lambda f: f.get("confidence", 0),
+                key=_coerce_source_confidence,
```
willem-bd: Worth being explicit about the mirror of the `0.0 -> 0.5` change on this side. The PR body spells out the staleness cap (removes-lowest-first) direction in detail; this `max_facts` trim is the less-visible half: with `reverse=True`, a missing-key fact (now `0.5`) survives the trim over a *genuinely* low-confidence fact (e.g. `0.1`), whereas before both ranked ~`0` and ordering fell to sort stability. It's correct and intended — `test_max_facts_trim_sort_survives_non_float_stored_confidence` confirms the direction — but a sentence in the body (or a word in the `# Sort by confidence and keep top ones` comment above) would make the full behavior delta equally visible on both sides. Non-blocking.

willem-bd (commented):
Verified the diagnosis and fix independently against the PR head — clean and complete.

- **Crash premise holds.** `_build_staleness_section` (line 658) runs *before* `_get_model()` (line 768), inside the `except Exception -> return False` handler — so a bad `confidence` aborts the whole cycle before the LLM is even constructed, and leaves the offending fact on disk (self-perpetuating). ✓
- **These are the only three stored-`confidence` raw reads.** The other four reads (lines 268 / 367 / 944 / 1050) are all LLM-produced data already normalized by `_normalize_memory_update_fact` or the consolidation-output path. Leaving them alone is correct. ✓
- **Helper mapping matches the documented table**, and both sort directions are preserved (staleness cap: ascending → removes lowest; `max_facts` trim: `reverse=True` → keeps highest). ✓
- **Tests pin both directions** — crash-prevention *and* the exact mapping (incl. the `bool`-subclasses-`int` edge via `True -> "0.50"`, and `inf`/`nan`), and the sort tests assert ranking *outcome* rather than just absence of exception.

The `0.0 -> 0.5` default change is the right call (unknown ≠ worthless) and is coherent in both trim directions, not just the cap. Two minor notes inline; neither blocks merge. Recommend approve.

AnnaSuSu:
Thanks — and the second note is sharper than it reads. Chasing it turned up something the PR was wrong about.

**The `max_facts` trim delta is real, and my tests did not cover it.** You wrote that `test_max_facts_trim_sort_survives_non_float_stored_confidence` "confirms the direction". It does not. All three of its facts carry a `confidence` key, so the `0.5` default is never exercised — the test pins the `str` → float coercion and nothing else. It passes unchanged whichever default the helper uses. That is the same circular-anchor mistake you flagged on #4017, one layer down: the test I cited as proof of the mapping could not go red for the mapping.

**Enumerating the actual delta set made the bug bigger, not smaller.** `str` and `None` raise. `true` / `false` / `inf` / `nan` / an absent key do not — so under the raw key they never reached the `except Exception` handler at all. They silently ranked at an *end* of the order:

| stored | old rank | new rank |
|---|---|---|
| absent, `false` | `0` — worst | `0.5` |
| `true` | `1.0` — best (`bool` subclasses `int`) | `0.5` |
| `inf` | `+inf` — best | `0.5` |
| `nan` | undefined | `0.5` |

So the two sort sites carry a second defect the PR never claimed: not an aborted cycle, but **silent deletion of a genuine fact**. End-to-end on a real `memory.json`, `max_facts=1`, one fact hand-edited to `"confidence": true` and one genuine at `0.9` — on `main` the cycle returns `True`, rewrites the file, and the surviving fact is the corrupted one. No exception, no log line. Your framing of `0.5` as "unknown ≠ worthless" is what makes this come out right in both directions; I had only argued it for the cap.

Pushed:

- **Ranking tests at both sort sites**, parametrized over the four non-raising inputs, each pitted against a genuine neighbour chosen so the survivor flips.
- `nan` is deliberately **not** in that parametrize. Its old rank is not an end of the order but undefined — it compares false against everything, so for the ascending cap sort a single fixed input order happens to yield the correct survivor on `main`. Pinning it by outcome would have been another test that cannot go red. Its real property is order-independence, so that is what the two dedicated tests assert, across both fact orders. (On `main`, storing the corrupted fact *second* makes the cap delete the genuine one instead.)
- The comment you asked for at the trim site, plus the delta table and the second e2e in the body.

Measured: the two test files give **25 failed / 93 passed** on a clean `origin/main` (`c0b917cc`) and **118 passed** here. Reverting only the two sort keys on this branch turns exactly the 10 new tests red, isolating the sorts from the formatter. Full suite **6881 passed, 8 pre-existing `web_fetch` failures** — unchanged from the parent's `6871 / 8`.

On the first note: #3993 has since merged; this branch rebases onto it cleanly. #4028 is still open and I'll take the conflict at that line whenever it lands — happy to go second.

AnnaSuSu:
Rebased onto `main` now that #4028 has landed, as offered in the description.

The conflict was purely textual: #4028 touched `cat` and `content` in `_build_staleness_section`, which sit immediately above and below the `conf` line this PR changes. Resolution keeps both — `cat = html.escape(...)` from #4028, `conf = _coerce_source_confidence(fact)` from here. The test-file conflict was two additions at the same point in the same class; both sets are kept.

Worth noting #4028 did not incidentally fix this: `conf` is still read raw on `main`, so `f"{conf:.2f}"` still raises `ValueError` on a string-valued stored confidence.

Re-verified after the rebase: reverting the three source hunks turns all 25 of this PR's new test cases red, while #4028's three escaping tests stay green — the two changes are orthogonal and each is still anchored. Full backend suite shows no new failures.

AnnaSuSu:
Rebased onto `938391c1` — no conflicts, ready for review.

@WillemJiang one thing worth flagging, since it changed what this PR touches: #4023 landed while this was open and extracted the `max_facts` trim into `_trim_facts_to_max()`, writing it with `key=_coerce_source_confidence`. That was one of my three sites, so **upstream already fixed it and I dropped that hunk.** The same PR introduced `search_memory_facts()` with a fresh raw read of the same field, which I've added in its place. Still three sites, different membership — the two staleness ones are unchanged from the original submission.

The three tests that anchored the trim now pass on `main` without the patch, so they'd guard nothing; I repointed that delta matrix at `search_memory_facts`, where it does go red. Re-verified per-test: reverting the three fixed lines turns exactly the 25 new test ids red, no more and no less.

Body is fully re-measured against the new base, including the end-to-end runs.

WillemJiang (approved):

AnnaSuSu:
@WillemJiang gentle ping — this was approved on Jul 11 and is still open. CI is green and it's mergeable on the current base; is there anything left on my side, or is it just waiting on a merge?

The two open review threads are both stale, in case they're what's holding it: the `updater.py:595` merge note anticipated the conflict with #4028, which has since landed and I rebased past it (and past #4023's `_trim_facts_to_max()` extraction after that). The other thread is already marked outdated. Happy to resolve both if you'd rather the PR show clean.

PR #4036, opened 2026-07-10
feat(frontend): add voice dictation input

18062706139fcz:
## Summary

- Add browser-based voice dictation to the Web UI composer using the Web Speech API.
- Keep voice input as a local draft flow: transcribed text is inserted into the existing prompt box and users can review/edit before sending.
- Add privacy-safe copy: DeerFlow receives only transcribed text, while audio handling is delegated to the browser/system speech-recognition service.
- Add focused unit coverage for speech-recognition helper behavior.

## Behavior

- Shows a microphone button when browser speech recognition is available.
- Toggles to a stop control while listening.
- Aborts voice recognition before submit/manual edits/thread cleanup so late speech results cannot repopulate the next draft.
- Keeps existing `/goal`, `/compact`, slash skill, attachment, input polish, and submit flows on the same text-input path.

## Tests

Ran with Node v22.21.1 via PATH:

- `pnpm check`
- `pnpm format`
- `pnpm test tests/unit/core/voice-input/speech-recognition.test.ts`


frontend/src/components/workspace/input-box.tsx
```diff
@@ -2153,6 +2344,50 @@ export function InputBox({
...
+  supported: boolean;
+  onToggle: () => void;
+}) {
+  const { t } = useI18n();
+  const tooltipContent = !supported
+    ? t.inputBox.voiceInputUnsupported
+    : listening
+      ? t.inputBox.voiceInputListening
+      : t.inputBox.voiceInputStart;
+  const label = listening
```
willem-bd, line 2454: **a11y: the accessible name is the full privacy blurb.** `label` resolves to `voiceInputStart` — _"Dictate with voice. DeerFlow receives only transcribed text; audio is handled by your browser or system speech service."_ — so a screen reader announces that entire sentence every time the button is focused.

The tooltip already carries the explanation for sighted users; the `aria-label` should be short. Suggest adding a concise label (e.g. `voiceInputStartLabel: "Dictate with voice"`) and using the long string only for the tooltip. Worth doing before merge since the PR is clearly aiming for good a11y.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -1003,6 +1048,140 @@ export function InputBox({
...
+    };
+    recognition.onerror = (event) => {
+      const message = getVoiceInputErrorMessage(
+        mapSpeechRecognitionError(event.error),
+      );
+      if (message) {
+        toast.error(message);
+      }
+    };
+    recognition.onend = () => {
```
willem-bd: **`continuous` won't survive Chrome's auto-stop.** Chrome stops a `continuous` session after a silence timeout (~`no-speech`) or ~60s and then fires `onend`. This handler only cleans up and flips the button back to idle, so a user pausing to think — or dictating anything longer than a sentence — silently ends the session and has to re-click.

For a dictation feature that's the main use case, so worth a follow-up: consider auto-restarting here when the session was user-started and not aborted/stopped. Not a blocker.

frontend/src/core/voice-input/speech-recognition.ts
```diff
@@ -0,0 +1,144 @@
...
+  return (
+    maybeWindow.SpeechRecognition ?? maybeWindow.webkitSpeechRecognition ?? null
+  );
+}
+
+export function getSpeechRecognitionLanguage(locale: string): string {
+  if (locale.toLowerCase().startsWith("zh")) {
+    return "zh-CN";
+  }
+  return "en-US";
```
willem-bd: **Silent English fallback for any non-`zh` locale.** Only `en-US`/`zh-CN` ship today, so this is correct now — but adding `ja`/`fr` later would silently recognize English.

Since the helper already takes the locale, consider deriving the BCP-47 tag directly (with an allow-list) rather than hard-coding two outcomes. Note the test currently asserts `fr-FR → en-US`, which codifies the brittle behavior.

frontend/src/components/workspace/input-box.tsx
```diff
@@ -1003,6 +1048,140 @@ export function InputBox({
...
+      return;
+    }
+    try {
+      recognition.stop();
+    } catch {
+      cleanupVoiceRecognition(recognition);
+    }
+  }, [cleanupVoiceRecognition]);
+
+  const toggleVoiceInput = useCallback(() => {
```
willem-bd, line 1234: **Coverage gap: the lifecycle isn't tested.** The pure helpers are well covered, but `toggleVoiceInput` / abort-on-submit / abort-on-manual-edit / abort-on-thread-switch / the `composerLocked` stop effect — i.e. the parts where regressions will actually appear — have no test.

`data-testid="voice-input-button"` is already in place, and `getSpeechRecognitionConstructor` accepts `unknown`, so injecting a fake constructor in a component test is straightforward. A follow-up here would de-risk this significantly.

willem-bd (commented):
Reviewed the voice dictation change. Overall it's solid — the Web Speech lifecycle (stale handlers, late results repopulating state, abort on submit/edit/thread-switch) is handled carefully, the cumulative-transcript recompute avoids duplication, the constructor lookup is SSR-safe, and the pure helpers are well tested. No blocking correctness bugs.

Inline notes below; the short version:
- **a11y:** `aria-label` is the full privacy sentence — should be concise (suggest fixing before merge)
- `continuous` sessions die on Chrome's silence/timeout auto-stop — worth a follow-up auto-restart
- language fallback forces `en-US` for any non-`zh` locale
- no component-level tests for the toggle/abort lifecycle

Minor nits (not blocking): `getVoiceInputErrorMessage` deps enumerate six leaf strings (`[t]` is cleaner if `t` is referentially stable per locale — worth checking the i18n hook); the `requestAnimationFrame` focus callback isn't cancelled on unmount/thread-switch (guarded by `?.focus()`, so harmless).

WillemJiang (approved):

PR #4037, opened 2026-07-10
feat(skills): add skill review quality gate

18062706139fcz:
## Summary

- Add a read-only skill review core with package snapshot readers, deterministic facts, resource/eval analysis, report rendering, and versioned contracts under `contracts/skill_review/`.
- Add the built-in `review_skill_package` tool plus the public `skills/public/skill-reviewer` skill, keeping review isolated from skill activation, required secrets, target scripts, and target dependency installation.
- Add a diff-scoped GitHub Actions workflow and reusable runner that review only changed `skills/public/**/SKILL.md` owning packages and hard-fail on deterministic `error` or `blocker` findings via `--fail-on error`.
- Update README and AGENTS docs for the new governance path.

## Actual Effect

- PRs that touch `skills/public/<skill>/SKILL.md` now run deterministic skill review only for that skill package.
- PRs with no public `SKILL.md` changes skip review work cleanly.
- The review path is static and read-only: it does not activate the reviewed skill, bind target secrets, execute target scripts, install target dependencies, or use LLM semantic review as a CI hard gate.

## Verification

- `PYTHONPATH=.:../scripts PYTHONIOENCODING=utf-8 PYTHONUTF8=1 uv run pytest tests/test_review_changed_public_skills.py tests/test_skill_review_core.py tests/test_review_skill_package_tool.py tests/test_skill_reviewer_public_skill.py -q` -> 19 passed in 8.90s
- `uv run ruff check ../scripts/review_changed_public_skills.py tests/test_review_changed_public_skills.py packages/harness/deerflow/skills/review packages/harness/deerflow/tools/builtins/review_skill_package_tool.py tests/test_skill_review_core.py tests/test_review_skill_package_tool.py tests/test_skill_reviewer_public_skill.py` -> passed
- `uv run ruff format --check ../scripts/review_changed_public_skills.py packages/harness/deerflow/skills/review/resource_graph.py` -> passed
- `uv run python ../scripts/review_changed_public_skills.py --before HEAD --after HEAD` -> skipped no-op diff with exit 0
- `uv run python -m deerflow.skills.review.cli ../skills/public/skill-reviewer --format text --fail-on error` -> 0 blocker(s), 0 error(s), 4 warning-level fixture findings


18062706139fcz:
## RFC: Skill Review quality gate

This PR adds a production path for Skill quality governance: deterministic review facts, a read-only review tool, a public reviewer skill, and a diff-scoped CI gate for changed public skills.

### Why this matters

Before this change, Skill quality checks were mostly split between manual review and safety scanning. That left several gaps:

- A Skill could be structurally valid enough to load, but still have unclear triggers, broken resources, weak evidence, or unsafe review boundaries.
- Reading a target Skill for review could be confused with loading that Skill for execution unless the two paths were explicitly separated.
- CI had no reusable, machine-readable facts layer for public Skill quality gates.
- Reviewer feedback was hard to standardize across authors, maintainers, and future automation.

### What this PR changes

- Adds a read-only review core under `backend/packages/harness/deerflow/skills/review/`.
- Adds versioned JSON contracts under `contracts/skill_review/` for review facts and reports.
- Adds the built-in `review_skill_package` tool, which returns `review_subject_entry` metadata instead of `skill_context_entry`, so reviewing a Skill does not activate it.
- Adds `skills/public/skill-reviewer/` as the human-friendly semantic review entry point.
- Adds `scripts/review_changed_public_skills.py` plus `.github/workflows/skill-review-ci.yml` so PRs only review changed `skills/public/**/SKILL.md` owning packages.
- Uses `--fail-on error`, so deterministic `error` and `blocker` findings fail CI while warning-level calibration findings remain advisory.

### Key boundaries

- Review is read-only.
- Review does not install the target Skill.
- Review does not execute target scripts.
- Review does not bind target `required-secrets`.
- Review does not apply the target Skill's `allowed-tools` to the reviewer.
- Model-based semantic review remains advisory; deterministic facts are the CI gate.

### Benefits

- **PR-level quality gate**: public Skill changes now get an automated deterministic check before merge.
- **Smaller CI blast radius**: the workflow is diff-scoped, so it reviews only changed public Skill packages instead of scanning every Skill on every PR.
- **Safer architecture**: review and execution are separated, which avoids context, secret, and tool-policy contamination.
- **Reusable facts layer**: `review-facts.v1` can be consumed by CI, future UI reports, installation preflight, and `skill-creator` workflows.
- **Lower reviewer burden**: maintainers get consistent blocker/error/warning output instead of re-deriving basic package quality issues manually.
- **Clear upgrade path**: deterministic gates can stay strict while semantic readiness scoring evolves independently.

### Verification

Local verification covered:

- focused Skill Review tests: `19 passed`;
- bundled Skill and SkillScan fixture-boundary tests: `44 passed`;
- `make lint` for backend formatting and ruff checks;
- no-op diff runner smoke;
- real `skill-reviewer` package review with `--fail-on error`.

The CI failure found after opening this PR was useful: it proved the new gate was active. The follow-up fixes now exclude reviewer calibration fixtures from being treated as real public Skill packages while keeping the actual root Skill review enforced.

backend/packages/harness/deerflow/skills/review/readers.py
```diff
@@ -0,0 +1,365 @@
...
+                        continue
+                    rel_path = self._normalize_archive_name(info.filename, snapshot)
+                    if rel_path is None:
+                        continue
+                    total_bytes += max(info.file_size, 0)
+                    if total_bytes > self.limits.max_total_bytes:
+                        snapshot["truncated"] = True
+                        snapshot["reader_errors"].append({"code": "total_size_exceeded", "path": rel_path, "message": "Archive total size exceeds the review limit"})
+                        break
+                    if info.file_size > self.limits.max_file_bytes:
```
willem-bd: **Size limits trust the declared `info.file_size`, not actual decompressed bytes.**

`max_file_bytes`/`max_total_bytes` gate on `info.file_size` from the zip's central directory (lines 279, 284), but `zf.read(info)` (line 293) decompresses the real stream. A crafted archive can declare a small `file_size` while its deflate stream decompresses far more, bypassing both limits and exhausting gateway memory. Since this tool is a builtin reachable on any allowlisted `.skill` under `/tmp`/cwd/skills-root, a user-supplied malicious archive reviewed via the tool is a plausible DoS path.

Classic high-ratio bombs (honest huge `file_size`) are already caught; this is the spoofed-declared-size gap. Consider streaming `zf.open(info)` in chunks and aborting + marking `truncated` once actual decompressed bytes exceed the limit, rather than trusting `info.file_size`.

backend/packages/harness/deerflow/skills/review/readers.py
```diff
@@ -0,0 +1,365 @@
...
+            return None
+
+    @staticmethod
+    def _sort_snapshot(snapshot: dict[str, Any]) -> dict[str, Any]:
+        snapshot["files"] = sorted(snapshot["files"], key=lambda item: item["path"])
+        snapshot["reader_errors"] = sorted(snapshot["reader_errors"], key=lambda item: (str(item.get("path") or ""), str(item.get("code") or "")))
+        return snapshot
+
+
+class ArchivePackageReader:
```
willem-bd, line 255: **The security-critical readers are the least tested.**

`ArchivePackageReader` (zip path normalization, symlink members via `_zip_member_is_symlink`), `parse_skill_uri` (`skill://public/../../etc` rejection), and `normalize_relative_path` (`..`/absolute rejection) have no direct tests. The traversal defenses look correct, but they're exactly the code that most warrants direct adversarial tests — and the `.skill` archive path currently has zero coverage.

Suggested cases: archive member with `../` escape, absolute zip member name, symlink member pointing outside the package, and a `skill://` target with traversal segments.

backend/tests/test_skill_review_core.py
```diff
@@ -0,0 +1,103 @@
...
+    return f"---\nname: {name}\ndescription: {description}\nallowed-tools: []\n---\n\n# Demo\n\nFollow the steps and stop.\n"
+
+
+def test_review_core_accepts_minimal_valid_skill(tmp_path):
+    _write(tmp_path / "SKILL.md", _valid_skill())
+
+    snapshot = LocalDirectoryReader(tmp_path).read()
+    facts = analyze_skill_package(snapshot)
+
+    assert facts["schema_version"] == "deerflow.skill-review.facts.v1"
```
willem-bd, line 44: This asserts the `schema_version` string but doesn't validate the emitted facts against `contracts/skill_review/review_facts.v1.schema.json`. The PR ships three versioned schemas as contracts, yet no test confirms snapshots/facts/reports actually conform — a silent contract drift (renamed required field, type change) wouldn't be caught.

Consider adding a `jsonschema`-validated golden test per schema (snapshot, facts, report) so the contracts are enforced, not just documented.

willem-bd (commented):
Reviewing the skill review quality gate (#4037).

**Overall: sound design and strong security posture.** Read-only, no script execution/network/dep-install, SkillScan runs on an ephemeral temp copy, the non-activation boundary is enforced by metadata key naming (`review_subject_entry` vs `skill_context_entry`) and tested, path-traversal defenses are present across all three readers, and CI is least-privilege with SHA refs (no injection) and `skills/public/**/SKILL.md` path filtering.

**Inline findings** (ranked): the archive reader trusts declared `info.file_size` for size limits (zip-bomb gap) and is the least-tested path; plus contract-conformance tests, the `SKILL_LOADING_TOOL_NAMES` naming, unsanitized review output (by design), the SkillScan `HIGH`→warning CI bar, and a few cleanups.

**Minor notes not posted inline:**
- `eval_schema.py`: `path in _EVAL_MANIFEST_NAMES or (path.startswith("evals/") and path.endswith(".json"))` — the first clause is a subset of the second; the `or` is redundant.
- `cli.py`: `--fail-on <level>` means "this severity **or worse**" (so `--fail-on error` also fails on `blocker`). Correct, but the help text doesn't state it.
- `analyzer.py::_valid_skill_name` does a function-local `import re`; hoist it.
- `_ensure_local_target_allowed` allows all of `Path.cwd()`; `target="."` walks the whole repo (bounded by limits, read-only, and `read_file` already grants equivalent access) — consider requiring a `SKILL.md` in the target tree to avoid accidental whole-repo walks.
- `_scan_with_skillscan` writes every text file to a temp dir then scans; bounded by `PackageLimits` so fine for CI/typical use, just noting the cost scales with package size.

Recommendation: approve with follow-ups; prioritize (1) archive-reader traversal tests, (2) cap actual decompressed bytes, (3) schema-conformance tests before merge.

backend/packages/harness/deerflow/skills/tool_policy.py
```diff
@@ -10,7 +10,7 @@ class NamedTool(Protocol):
     name: str
 
 
-SKILL_LOADING_TOOL_NAMES = frozenset({"read_file"})
+SKILL_LOADING_TOOL_NAMES = frozenset({"read_file", "review_skill_package"})
```
willem-bd: Adding `review_skill_package` to `SKILL_LOADING_TOOL_NAMES` (the `always_allowed_tool_names` set passed to `filter_tools_by_skill_allowed_tools`) means no skill's `allowed-tools` whitelist can exclude the review tool — it's unconditionally available to the lead whenever any skill restricts tools.

Functionally fine (it's a read-only builtin, consistent with `BUILTIN_TOOLS`), but the name `SKILL_LOADING_TOOL_NAMES` implies "infrastructure needed to load skills" — `read_file` qualifies, a review tool doesn't. Consider a separate `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES` or a comment stating the override intent, so the policy is explicit rather than implicit.

backend/packages/harness/deerflow/tools/builtins/review_skill_package_tool.py
```diff
@@ -0,0 +1,161 @@
...
+            "display_ref": facts["subject"]["display_ref"],
+            "package_digest": facts["subject"]["package_digest"],
+            "profile": profile,
+            "scope": scope,
+        }
+        return Command(
+            update={
+                "messages": [
+                    ToolMessage(
+                        content=stable_json_dumps(payload),
```
willem-bd: The target `SKILL.md` content (including adversarial prompts, per the `prompt-injection` fixture) is returned in the `ToolMessage` content/artifacts and is **not** run through `ToolResultSanitizationMiddleware` (which targets `web_fetch`/`web_search`). Mitigation is prompt-level: the skill's "treat as untrusted" instructions + the `untrusted_review_data: True` marker.

This is by design — sanitizing would defeat the review — and the prompt-injection fixture tests the boundary. Flagging only so the deliberate untrusted-content entry point is explicit for a `risk:high` PR; no change required unless you want framework-level treatment for review data.

backend/packages/harness/deerflow/skills/review/models.py
```diff
@@ -0,0 +1,126 @@
...
+ProfileName = Literal["deerflow", "agentskills"]
+
+SEVERITY_RANK: dict[str, int] = {
+    "blocker": 0,
+    "error": 1,
+    "warning": 2,
+    "info": 3,
+}
+
+SKILLSCAN_SEVERITY_MAP: dict[str, Severity] = {
```
willem-bd, line 25: `HIGH` maps to `warning`, and the CI gate uses `--fail-on error` (fails on `blocker`/`error` only). So a `HIGH`-severity SkillScan finding does **not** fail CI. This is consistent with existing SkillScan Phase 1 ("blocks CRITICAL"), but for a "quality gate" it's worth confirming the bar is intended — e.g., a suspicious-pattern `HIGH` finding would pass the gate as a warning.

backend/packages/harness/deerflow/skills/review/readers.py
```diff
@@ -0,0 +1,365 @@
...
+    if len(data) > limits.max_file_bytes:
+        snapshot["truncated"] = True
+        snapshot["reader_errors"].append(
+            {
+                "code": "file_too_large",
+                "path": "SKILL.md",
+                "message": "Inline SKILL.md exceeds the per-file review limit",
+            }
+        )
+        content = content[: limits.max_file_bytes]
```
willem-bd: `content[: limits.max_file_bytes]` slices by code point, but `max_file_bytes` is a byte budget — multibyte UTF-8 content can exceed the byte limit after truncation. For consistency with the byte-based guard, slice the encoded bytes and re-decode (`errors="ignore"`). Low impact at 64 MiB, but the char/byte mismatch is a latent foot-gun.

backend/packages/harness/deerflow/skills/parser.py
```diff
@@ -137,22 +133,23 @@ def parse_skill_file(skill_file: Path, category: SkillCategory, relative_path: P
...
         try:
-            metadata = yaml.safe_load(front_matter_text)
+            yaml.safe_load(front_matter_text)
         except yaml.YAMLError as exc:
             logger.error("%s", _format_yaml_error(skill_file, exc, front_matter_text))
             return None
 
-        if not isinstance(metadata, dict):
-            logger.error("Front-matter in %s is not a YAML mapping", skill_file)
+        parts, error = split_skill_markdown(content)
```
willem-bd: The frontmatter is parsed twice: the parser's own regex + `yaml.safe_load` (line 143, for rich line-number errors), then `split_skill_markdown(content)` (line 148) re-matches with an identical regex and re-parses. The success-path re-parse is effectively dead — same regex, so if the first succeeded the second will too.

Either drop the redundant `split_skill_markdown` call and reuse the already-parsed metadata, or drive everything through the shared helper and move the rich formatting into it.

backend/packages/harness/deerflow/skills/skillscan/orchestrator.py
```diff
@@ -271,6 +271,14 @@ def _scan_file_package_properties(rel_path: str, file_bytes: bytes, file_size: i
     return findings
 
 
+def _is_eval_fixture_skill_md(path: PurePosixPath) -> bool:
```
willem-bd: This is the third `_is_eval_fixture_skill_md` implementation in the repo (also in `scripts/review_changed_public_skills.py` and `tests/test_skills_bundled.py`), each with a different part-count threshold and base-path assumption. They agree on the standard `skills/public/<pkg>/evals/fixtures/<name>/SKILL.md` layout but will drift silently. Consider extracting one shared helper.

willem-bd (commented):
Reviewed the head commit. Well-architected, security-conscious foundation — clean layering (review core / built-in tool / public skill / CI gate), solid path-traversal and symlink defenses, honest completeness signaling, versioned JSON contracts, and good test coverage (19 new tests). No critical correctness bug; the inline comments below are policy/calibration and robustness items.

The two I'd want resolved before merge: **#1** (HIGH-severity SkillScan findings map to `warning` and pass the gate — including `declaration-prompt-override` and sensitive-path reads) and **#2** (non-`SKILL.md` package edits bypass review entirely). **#3** (cwd allow-root + the tool being globally available to every skill) is worth an explicit decision.

Smaller notes (not anchored):
- `is_zero_sha` (`set(value) == {"0"}`) accepts any-length all-zero string; a length check would be stricter.
- `build_inline_snapshot` truncates by *character* count then re-encodes with `errors="ignore"` — multi-byte UTF-8 split bytes are silently dropped. Cosmetic at the 64 MB default.
- `_semantic_artifacts` marks the *payload* `untrusted_review_data: true`, but individual `artifacts[*]` aren't tagged — if a consumer splits the payload, the marker is lost.
- `scope` is a free-form `list[str]`; `profile` is correctly `Literal`-constrained.
- The `legacy` category resolves to `get_skills_root_path()/"custom"/rel_path` — an implicit layout assumption worth a comment.
- Push path: `github.event.before` can orphan after a force-push → `git diff` fails non-zero. Edge case.

backend/packages/harness/deerflow/skills/review/models.py
```diff
@@ -0,0 +1,126 @@
...
+    "CRITICAL": "blocker",
+    "HIGH": "warning",
```
willem-bd: The CI gate runs `--fail-on error`, which fails only on severity rank ≤ `error` (`cli.py::_exit_code`). Because `HIGH` maps to `warning` (rank 2), HIGH-severity SkillScan findings do **not** fail the gate. That currently includes `declaration-prompt-override`, `declaration-sensitive-capability`, `declaration-sensitive-path`, `python-dynamic-import`, `python-subprocess`, and `secret-env-assignment` — all surfaced as passing warnings.

The skillscan core itself only blocks `CRITICAL`, but the review layer is free to be stricter: mapping `HIGH → error` would let the quality gate catch prompt-override and sensitive-path findings. If leaving `HIGH` as a passing warning is deliberate, a one-line note here would make the calibration explicit.

scripts/review_changed_public_skills.py
```diff
@@ -0,0 +1,227 @@
...
+
+import argparse
+import os
+import subprocess
+import sys
+from collections.abc import Sequence
+from dataclasses import dataclass
+from pathlib import Path, PurePosixPath
+
+PUBLIC_SKILL_MD_PATHSPEC = ":(glob)skills/public/**/SKILL.md"
```
willem-bd: The diff pathspec is `skills/public/**/SKILL.md`, and `select_skill_packages` only queues a package when its `SKILL.md` appears in the diff. A PR that edits `skills/public/<skill>/references/foo.md` or adds a `scripts/` / `templates/` file — without touching `SKILL.md` — is skipped entirely, so a secret or missing-resource introduced via a non-`SKILL.md` file bypasses review.

Consider widening the pathspec to cover package content (at least `references/`, `scripts/`, `templates/`, `evals/`) while still keying package selection off any changed file under the package.

backend/packages/harness/deerflow/tools/builtins/review_skill_package_tool.py
```diff
@@ -0,0 +1,161 @@
...
+    path = Path(target).expanduser()
+    _ensure_local_target_allowed(path)
+    if path.suffix == ".skill":
+        return ArchivePackageReader(path).read()
+    return LocalDirectoryReader(path).read()
+
+
+def _ensure_local_target_allowed(path: Path) -> None:
+    resolved = path.resolve()
+    allowed_roots: list[Path] = [Path.cwd().resolve(), Path("/tmp").resolve()]
```
willem-bd, line 120: `Path.cwd()` is an allowed root, alongside `/tmp` and the skills root. Combined with `review_skill_package` being added to `SKILL_LOADING_TOOL_NAMES` (`tool_policy.py:13`) — i.e. available to **every** skill, not just `skill-reviewer` — any skill can read and surface up to 80 KB (`_MAX_SEMANTIC_ARTIFACT_CHARS`) of text from any file under cwd into its context.

If the gateway cwd is broad (repo root / workspace), this is a read primitive for skills that otherwise declare a restricted `allowed-tools`. Confirm cwd is the intended trust boundary, and add a test asserting the allow-scope — the only current test covers `/etc` rejection. Note `expanduser()` runs before the check, so `~/…` targets are reachable only if home sits under an allowed root; worth a test too.

backend/packages/harness/deerflow/skills/skillscan/orchestrator.py
willem-bd: This is the third copy of "is eval-fixture SKILL.md": `scripts/review_changed_public_skills.py` uses `parts[3]/parts[4]`, `tests/test_skills_bundled.py` uses `parts[1]/parts[2]`, and the analyzer uses a package-relative `startswith("evals/fixtures/")`. Each hard-codes indices against a different root, so they'll drift as layouts change.

This positional form (any-position `evals/fixtures` + `parts[-1]=="SKILL.md"`) is the most robust — extract it as one shared helper and reuse everywhere.

backend/packages/harness/deerflow/skills/parser.py
```diff
@@ -137,22 +133,23 @@ def parse_skill_file(skill_file: Path, category: SkillCategory, relative_path: P
...
+        # error string; tests and authoring UX depend on the line-specific hint.
+        front_matter_match = re.match(r"^---\s*\n(.*?)\n---\s*\n?", content, re.DOTALL)
         if not front_matter_match:
             return None
-
         front_matter_text = front_matter_match.group(1)
-
         try:
-            metadata = yaml.safe_load(front_matter_text)
+            yaml.safe_load(front_matter_text)
```
willem-bd: This bare `yaml.safe_load(front_matter_text)` (result discarded) exists only to produce the richer `_format_yaml_error` diagnostic. Then `split_skill_markdown(content)` below re-runs the **same** regex and `yaml.safe_load`. Every skill load now parses YAML twice and matches the frontmatter regex twice.

Restructure to parse once — e.g. have `split_skill_markdown` surface the `YAMLError`, or call it first and only fall back to the detailed formatter on failure.

.github/workflows/skill-review-ci.yml
```diff
@@ -0,0 +1,70 @@
...
+
+jobs:
+  skill-review:
+    if: github.event_name != 'pull_request' || github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
willem-bd, line 40: Worth confirming `actions/checkout@v6`, `actions/setup-python@v6` (line below), and `astral-sh/setup-uv@v7` are published majors — they're ahead of the commonly pinned versions, and if a major doesn't exist the workflow fails on first run. Low confidence this is wrong given the 2026 date, but a quick check avoids a red CI on merge.

backend/packages/harness/deerflow/tools/builtins/review_skill_package_tool.py
```diff
@@ -0,0 +1,170 @@
...
+            continue
+        path = str(entry.get("path"))
+        if not _is_semantic_artifact(path):
+            continue
+        content = str(entry.get("content") or "")
+        truncated = False
+        if len(content) > remaining:
+            content = content[:remaining]
+            truncated = True
+        artifacts.append({"path": path, "content": content, "truncated": truncated, "untrusted_review_data": True})
```
willem-bd, line 177: **[Medium] Reviewed content reaches the model context without the sanitization remote-content tools get.**

The default `include_content="semantic-review"` bundles up to 80 KB of reviewed `SKILL.md` / `references` / `evals` text into `ToolMessage.content`, flagged `untrusted_review_data: true`. `ToolResultSanitizationMiddleware` only neutralizes framework/injection tags (`<system-reminder>`, boundary markers, …) for **remote-content** tools (`web_fetch` / `web_search` / `image_search`) — local tool output (bash / `read_file`, and this tool) is left untouched. But this tool's whole purpose is to ingest packages that are likely adversarial (the `prompt-injection` fixture literally embeds "Ignore the reviewer instructions…"), so a reviewed `SKILL.md` can forge framework tags into the lead-agent context with no defense beyond the model honoring the `untrusted_review_data` marker.

This is the same trust class as `read_file` / `bash`, but the reviewed content is far more likely to be adversarial. Consider routing review artifacts through the same neutralization path remote-content tools use, or document explicitly why this boundary is acceptable. Worth a maintainer decision before merge.

backend/packages/harness/deerflow/tools/builtins/review_skill_package_tool.py
```diff
@@ -0,0 +1,170 @@
...
+        snapshot = _snapshot_for_target(target, runtime=runtime, inline_content=inline_content)
+        facts = analyze_skill_package(snapshot, profile=profile)
+        artifacts = _semantic_artifacts(snapshot, include_content=include_content)
+        static_report = build_static_report(facts, scope=scope)
+        payload = {
+            "untrusted_review_data": True,
+            "facts": facts,
+            "artifacts": artifacts,
+            "static_report": static_report,
+            "markdown": {
```
willem-bd, line 60: **[Low–Med] Large `content` payload; consider trimming.**

`payload` serializes `facts` + ≤80 KB of `artifacts` + `static_report` + **both** `en` and `zh` markdown renders into `ToolMessage.content` (and again into `artifact`). If any tool-output budget / truncation middleware acts on `content`, the JSON is cut mid-string and the skill-reviewer cannot parse the facts it's told to read.

Suggestions:
- Don't embed both locales in `content` — the skill can render Markdown from `facts`, or the tool picks one locale from the run.
- Keep `content` to a compact summary and put the full payload only in `artifact`.

backend/packages/harness/deerflow/skills/tool_policy.py
```diff
@@ -10,7 +10,10 @@ class NamedTool(Protocol):
     name: str
 
 
-SKILL_LOADING_TOOL_NAMES = frozenset({"read_file"})
+# Framework built-ins that remain available even when an active skill declares
+# allowed-tools. They support controlled framework workflows rather than
+# extending the reviewed/activated skill's own tool authority.
+ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES = frozenset({"read_file", "review_skill_package"})
```
willem-bd, line 16: **[Low] `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES` now includes `review_skill_package`.**

The rename from `SKILL_LOADING_TOOL_NAMES` + addition is justified: a non-reviewer skill (e.g. `skill-creator`) needs to invoke review mid-turn even when it declares its own `allowed-tools`. Flagging because it widens the always-available surface for **every** active skill that declares `allowed-tools` — even a narrowly-scoped skill now exposes `review_skill_package` to the model. Blast radius is limited (read-only / static), but combined with the untrusted-content note above it's worth an explicit callout in the PR description rather than being implied by a rename.

scripts/review_changed_public_skills.py
```diff
@@ -0,0 +1,274 @@
...
+def run_review(package: Path, repo_root: Path, python_executable: str) -> int:
+    package_rel = package.relative_to(repo_root).as_posix()
+    command = [
+        python_executable,
+        "-m",
+        "deerflow.skills.review.cli",
+        package_rel,
+        "--format",
+        "text",
+        "--fail-on",
```
willem-bd, line 232: **[Low] Truncated packages pass the CI gate.**

`--fail-on error` only fails on `error` / `blocker`. Truncation produces `warning`s plus a `truncated` / `not_assessed` completeness flag, so a public skill that grows past `max_file_bytes` / `max_total_bytes` passes this gate silently. The `partial-package` fixture expects `readiness: revise`, but that softer signal never reaches the runner's exit code. Intentional? If not, consider escalating `--fail-on` (e.g. to `warning`) or at least surfacing truncation in the runner's success summary.

backend/packages/harness/deerflow/skills/review/readers.py
```diff
@@ -0,0 +1,410 @@
...
+                            snapshot["files"].append({"path": rel_path, "kind": "binary", "size": actual_size, "sha256": "", "content": None})
+                            snapshot["reader_errors"].append({"code": "file_too_large", "path": rel_path, "message": "Archive member exceeds the per-file review limit"})
+                            continue
+                        snapshot["reader_errors"].append({"code": "total_size_exceeded", "path": rel_path, "message": "Archive total size exceeds the review limit"})
+                        break
+
+                    total_bytes += actual_size
+                    if _zip_member_is_symlink(info):
+                        target = data.decode("utf-8", errors="replace")
+                        snapshot["files"].append({"path": rel_path, "kind": "symlink", "size": actual_size, "sha256": _sha256(data), "target": target})
```
willem-bd: **[Nit] Symlink `size` diverges between the two readers.**

`ArchivePackageReader` records `actual_size` here, while `LocalDirectoryReader` (around line 231) records `size: 0` for symlinks. Both are schema-valid, but inconsistent for any consumer keying on `size`. `0` is arguably more correct since a symlink carries no payload — worth aligning.

scripts/review_changed_public_skills.py
```diff
@@ -0,0 +1,274 @@
...
+
+def review_env(repo_root: Path) -> dict[str, str]:
+    env = os.environ.copy()
+    harness_path = repo_root / "backend" / "packages" / "harness"
+    existing_pythonpath = env.get("PYTHONPATH")
+    env["PYTHONPATH"] = str(harness_path) if not existing_pythonpath else f"{harness_path}{os.pathsep}{existing_pythonpath}"
+    return env
+
+
+def is_zero_sha(value: str) -> bool:
```
willem-bd, line 271: **[Nit] `is_zero_sha` only matches 40-char (SHA-1) SHAs.**

A SHA-256 repo's zero SHA is 64 chars, so `build_diff_args` won't take the empty-tree fast path directly for those repos. The force-push fallback (`build_force_push_fallback_diff_args`) still recovers because the primary `git diff` then fails on the bad object — so it's functionally fine today. Just noting the implicit SHA-1 assumption in case it matters later.

backend/packages/harness/deerflow/skills/review/eval_schema.py
```diff
@@ -0,0 +1,105 @@
...
+        manifest = _classify_manifest(payload)
+        manifest["path"] = path
+        aggregate["manifests"].append(manifest)
+        schemas.add(manifest["schema"])
+        aggregate["case_count"] += manifest["case_count"]
+        aggregate["positive_trigger_cases"] += manifest["positive_trigger_cases"]
+        aggregate["negative_trigger_cases"] += manifest["negative_trigger_cases"]
+
+    if schemas:
+        aggregate["schema"] = sorted(schemas)[0] if len(schemas) == 1 else "mixed"
```
willem-bd: **[Nit] Redundant `sorted()` when there is exactly one schema.**

`sorted(schemas)[0] if len(schemas) == 1 else "mixed"` — the sort only matters in the multi-element branch, which returns the literal `"mixed"` anyway. `next(iter(schemas)) if len(schemas) == 1 else "mixed"` is clearer (and avoids imposing an ordering on a single element).

willem-bd (commented):
Review of the skill-review quality gate.

Overall this is well-architected: clean harness/app split (review core imports only `deerflow.*`), versioned JSON contracts validated against their schemas in tests, strong path-traversal / zip-bomb defenses, consistent eval-fixture exclusion across the analyzer / SkillScan / resource graph / CI runner, and the non-activation + non-execution claims check out (results carry `review_subject_entry`, not `skill_context_entry`; no secret binding, no `allowed-tools` application, no script execution).

Verified the new sync `review_skill_package` tool does **not** block the event loop: `langchain_core` 1.3.3 `StructuredTool._arun` offloads sync-only tools via `run_in_executor`, so no `tests/blocking_io/` anchor is strictly required.

Inline notes below, in priority order. The main item I'd want resolved before merge is the untrusted-content sanitization gap on `_semantic_artifacts`.

WillemJiang (approved):

PR #4038, opened 2026-07-10
fix(front): resolve relative artifact image paths

chaoxi007:
## Why

Assistant responses may reference generated images using bare relative paths, for example:

    Overview

Although the referenced file exists in the thread's output artifacts, the frontend previously passed the relative path directly to the browser. The browser then resolved it relative to the chat page instead of the thread artifact API, resulting in a broken image and a 404 response.

This affects both newly generated responses and existing conversation history that contains relative artifact image paths.

## What changed

- Resolve relative Markdown image paths against the current thread's registered artifacts.
- Rewrite a relative path only when it uniquely matches an artifact, preventing incorrect resolution when an artifact is missing or the filename is ambiguous.
- Preserve the existing behavior for absolute `/mnt/...` artifact paths.
- Leave external, unregistered, ambiguous, and path-traversal image references unchanged.
- Update the agent prompt to require:
  - Complete `/mnt/user-data/outputs/...` paths for output images.
  - Calling `present_files` before referencing an output image.
  - Avoiding bare filenames and workspace-relative paths.
- Add regression tests for successful relative-path resolution and guarded fallback behavior.

## Surface area

- [x] **Frontend UI** - page / component / setting / interaction under `frontend/`
- [ ] **Backend API** - endpoint / SSE event / request-response shape under `backend/app`
- [x] **Agents / LangGraph** - agent node, graph wiring, `langgraph.json`, or prompt change
- [x] **Default behavior change** - changes existing behavior without the user opting in

### Before

Relative artifact image failing to render before the fix

### After

Relative artifact image rendering correctly after the fix

## Bug fix verification

**Test path:** `frontend/tests/unit/core/artifacts/utils.test.ts`

**Did the test go red on `main` and green on this branch?** No.

The test targets the artifact path resolution utility introduced by this change, so it could not be applied directly to `main` without part of the implementation. Instead, the issue was reproduced manually on the latest `main` using an exported conversation whose registered artifacts were referenced by bare filenames.

The branch was then built and validated with Docker. The same image references rendered successfully through the thread artifact API.

Unit tests cover unique relative-path resolution and verify that missing, ambiguous, external, and unsafe paths are not rewritten.

## Validation

Frontend:

- `frontend/tests/unit/core/artifacts/utils.test.ts`: 3 tests passed
- TypeScript `tsc --noEmit`: passed
- ESLint on modified frontend files: passed
- Prettier check on modified frontend files: passed

Backend:

- `backend/tests/test_lead_agent_prompt.py`: 20 tests passed
- Ruff check on modified backend files: passed

Additional validation:

- `git diff --check`: passed
- Docker development stack started successfully from the latest `main`-based branch
- Manual browser verification confirmed that relative artifact images render correctly
- Artifact requests succeeded through `/api/threads/{thread_id}/artifacts/...`

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Codex was used to trace the failure from the exported conversation through the thread artifact state, frontend Markdown rendering, and artifact API. It assisted with implementing the fix, adding regression tests, migrating the change onto the latest `main`, and running validation. I reviewed the resulting diff and manually verified the behavior in Docker.


willem-bd (commented):
Thanks @chaoxi007. I found a couple of issues that should be addressed before this is ready.

frontend/tests/unit/core/artifacts/utils.test.ts
```diff
@@ -73,4 +73,46 @@ describe("artifact URL helpers", () => {
...
+      resolveMessageImageURL("shared.png", "thread-1", [
+        "/mnt/user-data/outputs/first/shared.png",
+        "/mnt/user-data/outputs/second/shared.png",
+      ]),
+    ).toBe("shared.png");
+    expect(
+      resolveMessageImageURL("https://example.com/image.png", "thread-1", [
+        "/mnt/user-data/outputs/image.png",
+      ]),
+    ).toBe("https://example.com/image.png");
```
willem-bd, line 121: [P2] Add a regression test for the path-traversal guard

- **Problem:** The PR description states path-traversal image references are left unchanged, and the guard exists (`normalizedPath.split("/").includes("..")`), but no test exercises it.
- **Evidence:** The new test at `utils.test.ts:100` covers missing, ambiguous, and external-URL cases, but not `..` (nor absolute `/` or `//`). The guard could be removed and the suite would still pass.
- **Suggested fix:** add a case such as:
```ts
expect(
  resolveMessageImageURL("../etc/secret.png", "thread-1", [
    "/mnt/user-data/outputs/secret.png",
  ]),
).toBe("../etc/secret.png");
```
- **Test:** this is the test — it asserts the guard returns the input unchanged.

frontend/src/components/workspace/messages/message-list.tsx
```diff
@@ -730,6 +730,7 @@ export function MessageList({
                           groupIndex === groupedMessages.length - 1
                         }
                         threadId={threadId}
+                        artifactPaths={thread.values.artifacts ?? []}
```
willem-bd: [P2] Reuse the existing artifact helper

- **Problem:** `thread.values.artifacts ?? []` is inlined here, but `extractArtifactsFromThread(thread)` in the same module (`core/artifacts/utils.ts`) already returns exactly that.
- **Evidence:** The component already imports from `@/core/artifacts/utils`; inlining duplicates the `?? []` fallback so the two expressions can drift.
- **Suggested fix:** import `extractArtifactsFromThread` and pass `artifactPaths={extractArtifactsFromThread(thread)}`.

frontend/tests/unit/core/artifacts/utils.test.ts
chaoxi007: Addressed in the latest commit. Added regression coverage for path-traversal image references.

frontend/src/components/workspace/messages/message-list.tsx
chaoxi007: Addressed in the latest commit. Reused extractArtifactsFromThread and narrowed its parameter type so it also accepts the streamed thread shape.

WillemJiang (approved):

willem-bd (commented):
Thanks @chaoxi007. Two post-merge follow-ups worth addressing now that this is on `main` — details inline.

frontend/src/core/artifacts/utils.ts
```diff
@@ -22,7 +22,9 @@ export function urlOfArtifact({
   return `${getBackendBaseURL()}/api/threads/${threadId}/artifacts${filepath}${download ? "?download=true" : ""}`;
 }
 
-export function extractArtifactsFromThread(thread: AgentThread) {
+export function extractArtifactsFromThread(thread: {
+  values: Pick<AgentThreadState, "artifacts">;
+}) {
   return thread.values.artifacts ?? [];
```
willem-bd, line 28: [P2] `artifactPaths` reference instability can re-render historical messages during streaming

- Location: `frontend/src/core/artifacts/utils.ts:28` (root cause); also the per-message call in `frontend/src/components/workspace/messages/message-list.tsx` and the `useMemo([artifactPaths, threadId])` dep in `message-list-item.tsx`.
- Problem: `extractArtifactsFromThread` returns `thread.values.artifacts ?? []`. When `artifacts` is `undefined`, `?? []` mints a fresh array on every render. That value is now passed as the `artifactPaths` prop to the memoized `MessageContent` and is a `useMemo` dep for the markdown `components` map, so a fresh reference each render defeats the `memo` and re-renders historical messages' markdown on every streaming delta. Before this PR `MessageContent` had no `artifactPaths` prop, so this re-render path did not exist.
- Evidence: `MessageList` re-renders on each streaming `messages` delta and calls `extractArtifactsFromThread(thread)` once per message inside the `.map`; when `artifacts` is nullish the returned `[]` is a new reference each time. `EMPTY_THREAD_VALUES.artifacts` is `[]`, but threads whose checkpoint state omits `artifacts` hit the `?? []` fallback.
- Suggested fix: return a module-level stable empty array for the nullish case, and hoist the extraction out of the `.map`:
  ```ts
  const EMPTY_ARTIFACT_PATHS: readonly string[] = [];
  export function extractArtifactsFromThread(thread: {
    values: Pick<AgentThreadState, "artifacts">;
  }) {
    return thread.values.artifacts ?? EMPTY_ARTIFACT_PATHS;
  }
  ```
  and in `message-list.tsx`, compute `const artifactPaths = extractArtifactsFromThread(thread);` once, then pass it to each `MessageListItem`.
- Test: a render test that forces two `MessageList` renders with `artifacts` undefined and asserts an unchanged historical `MessageContent` does not re-render (e.g. via a render-count spy).

frontend/tests/unit/core/artifacts/utils.test.ts
```diff
@@ -73,4 +73,51 @@ describe("artifact URL helpers", () => {
       resolveArtifactURL("/mnt/user-data/outputs/style.css", "thread-1"),
     ).toBe("/demo/threads/thread-1/user-data/outputs/style.css");
   });
+
+  test("resolves a relative message image when it uniquely matches an artifact", async () => {
```
willem-bd, line 77: [P2] Test coverage gaps for `resolveMessageImageURL` branches

- Location: `frontend/tests/unit/core/artifacts/utils.test.ts`
- Problem: The new tests cover unique-match, `./` + hash, missing, ambiguous, traversal, and external URL, but two likely real-world shapes are not asserted.
- Evidence: No test exercises the `/mnt/` absolute-path branch (the `if (src.startsWith("/mnt/"))` early return delegating to `resolveArtifactURL`), and none exercises a multi-segment relative path such as `outputs/chart.png` matching `/mnt/user-data/outputs/chart.png` (the same `endsWith("/" + normalizedPath)` logic, but unverified at this entry point).
- Suggested fix: add two assertions:
  - `resolveMessageImageURL("/mnt/user-data/outputs/chart.png", "thread-1", artifacts)` → `/api/threads/thread-1/artifacts/mnt/user-data/outputs/chart.png`
  - `resolveMessageImageURL("outputs/chart.png", "thread-1", ["/mnt/user-data/outputs/chart.png"])` → resolves to the same artifact URL.
- Test: the two assertions above.

chaoxi007: Covered by #4134 . The follow-up adds regression coverage for both absolute `/mnt/...` image paths and multi-segment relative paths such as `outputs/chart.png`.

frontend/src/core/artifacts/utils.ts
chaoxi007: Thanks for the post-merge follow-up. The render-stability issue was addressed in #4094, and the remaining image-path coverage gap is addressed in #4134 .

PR #4040, opened 2026-07-10
fix(subagents): inject durable context before compaction

### LINKED ISSUE #4039: [bug] durable context should be injected after summarization compaction during the runtime of subagents
## Summary

Subagents inherit `DeerFlowSummarizationMiddleware`, but they do not inherit
`DurableContextMiddleware`. When summarization compacts the message channel,
the generated summary is stored in `ThreadState.summary_text` and the removed
messages are replaced only by the preserved tail. Unlike the lead-agent chain,
the subagent chain never projects `summary_text` back into subsequent model
requests.

With a message-count keep policy, the preserved tail can therefore begin with
an assistant tool-call followed by tool results. Strict OpenAI-compatible
providers reject that request shape with HTTP 400 `InvalidParameter`.

## Reproduction

1. Enable summarization with a token trigger and
   `keep: {type: messages, value: 4}`.
2. Run a subagent whose first model turn emits three parallel tool calls.
3. Let all three tool results return and trigger compaction before the next
   model call.
4. The four preserved messages are the assistant tool-call followed by the
   three tool results.
5. The original system/user messages are summarized into `summary_text`, but
   no subagent middleware injects that summary.
6. The next provider request starts with `assistant -> tool -> tool -> tool`
   and may be rejected as an invalid message sequence.

A two-tool-call control case succeeds because the last four messages still
include the human task: `human -> assistant -> tool -> tool`.

## Root cause

The lead chain registers `DurableContextMiddleware` immediately before
summarization. Its model wrapper injects a bounded summary as hidden
durable-context data. `build_subagent_runtime_middlewares` registers
summarization without that durable-context projection layer.

There is a related status issue: `LLMErrorHandlingMiddleware` converts a
non-retriable provider 400 into an AI fallback message, and the subagent
executor then treats that text as a successful final result. That should be
handled separately unless it can be safely classified without broadening this
fix.

## Proposed fix

Register `DurableContextMiddleware` in the subagent middleware chain before
`DeerFlowSummarizationMiddleware`, using the same app-config skill path and
skill-read tool settings as the lead chain.

## Acceptance criteria

- After subagent compaction, a non-empty `summary_text` is projected into the
  next model request.
- The durable summary is injected before a preserved assistant/tool tail, so
  the provider request does not begin with assistant/tool history.
- The injected durable context remains ephemeral and is not appended to
  checkpointed `messages`.
- Subagents preserve the same prompt-injection authority boundary used by the
  lead agent.
- Regression coverage reproduces the three-parallel-tool-call + `keep=4`
  case through the public middleware/agent path.
- Existing subagent middleware tests and formatting pass.

## Context

Regression introduced/exposed by #4009, which made subagents inherit
summarization.

ggnnggez:
## Summary

- add `DurableContextMiddleware` to the subagent runtime chain immediately
  before summarization
- reuse the lead chain's configured skills root and skill-read tool names
- add a graph-level regression for compaction after three parallel tool calls
- document durable summary projection for long-running subagents

## Root cause

Subagent summarization stores compacted history in `ThreadState.summary_text`,
but the subagent middleware chain did not project that state back into later
model requests. With `keep: {type: messages, value: 4}`, a turn containing
three parallel tool calls could compact into an assistant tool-call followed by
three tool results. Strict providers then received an assistant-first history
and rejected it with HTTP 400 `InvalidParameter`.

The lead chain already prevents this by registering
`DurableContextMiddleware` before summarization. This change mirrors that
composition for subagents while keeping the summary ephemeral and preserving
the existing durable-context prompt-injection boundary.

## User impact

Long-running subagents can continue after automatic context compaction even
when the preserved recent window contains only assistant/tool activity. Their
generated summary is available to the next model call instead of remaining in
state only.

## Validation

- `uv run pytest tests/test_tool_error_handling_middleware.py tests/test_durable_context_middleware.py tests/test_subagent_executor.py -q`
  - 117 passed
- `make lint`
  - Ruff lint passed
  - 802 files already formatted

Fixes #4039


backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
```diff
@@ -379,6 +379,22 @@ def build_subagent_runtime_middlewares(
...
+    # ``summary_text`` state channel instead of writing a summary message back
+    # into ``messages``. Mirror the lead chain so subagents project that summary
+    # into subsequent model requests; otherwise a message-count keep policy can
+    # leave an assistant tool-call + tool-result tail with no leading user
+    # context, which strict providers reject. The same middleware also keeps
+    # skill references durable when their original read results are compacted.
+    from deerflow.agents.middlewares.durable_context_middleware import DurableContextMiddleware
+
+    middlewares.append(
+        DurableContextMiddleware(
```
willem-bd, line 392: **Subagent chain is missing `SystemMessageCoalescingMiddleware`, which the lead chain pairs with `DurableContextMiddleware`.**

`_inject` always inserts a `SystemMessage(_AUTHORITY_CONTRACT)` alongside the durable `HumanMessage` (see `_insert_after_leading_system_messages` in `durable_context_middleware.py`). In the **lead** chain, `SystemMessageCoalescingMiddleware` is appended later (innermost) and merges `request.system_message` + every `SystemMessage` in `request.messages` into a single leading system message — its docstring notes vLLM/SGLang/Qwen/Anthropic reject non-leading or duplicate system messages. The subagent chain (used directly at `executor.py:458`, nothing appended) has no coalescer.

After this change, a post-compaction subagent request looks like `[System(system_prompt), System(authority_contract), Human(durable_data), Assistant(tool_calls), Tool, Tool, Tool]`. The `Human` before `Assistant` fixes the assistant-first rejection ✅, but the request now carries **two** system messages where it previously carried one. On a strict backend that requires a single leading system message, this can trade the #4039 HTTP 400 for an equivalent one.

**Suggestion:** append `SystemMessageCoalescingMiddleware` to the subagent chain (innermost, mirroring the lead chain). It's provider-agnostic, only touches the request payload, and would coalesce `system_prompt + authority_contract` into one leading system message while preserving the injected `Human` before the assistant tail — making the fix robust on the strict backends it targets. At minimum, confirm the #4039 backend tolerates a second leading system message before merging.

backend/tests/test_tool_error_handling_middleware.py
```diff
@@ -662,7 +670,92 @@ def fake_create_summarization_middleware(*, app_config=None, keep=None, skip_mem
...
+
+        def bind_tools(self, tools, **kwargs):
+            return self
+
+        def _generate(self, messages, stop=None, run_manager=None, **kwargs):
+            if self.require_durable_summary:
+                first_ai = next(i for i, message in enumerate(messages) if isinstance(message, AIMessage))
+                durable = [(i, message) for i, message in enumerate(messages) if isinstance(message, HumanMessage) and message.additional_kwargs.get("durable_context_data")]
+                assert durable, "compacted summary must be injected into the subagent request"
+                assert durable[0][0] < first_ai, "durable summary must precede the assistant/tool tail"
```
willem-bd, line 727: The `_StaticModel` validates that the durable block precedes the first `AIMessage`, but it doesn't validate the request stays well-formed for a strict backend after injection (e.g., a single leading `SystemMessage`). If the coalescer concern on the append site is real, this test would still pass while a strict provider rejects the two-system-message request.

Consider also asserting the final request has exactly one leading `SystemMessage` (or that system messages are coalesced), so the regression guard covers the strict-backend shape that motivated #4039 — not just the human-before-assistant ordering.

backend/tests/test_tool_error_handling_middleware.py
```diff
@@ -148,18 +148,21 @@ def __init__(self, *, app_config):
...
-    assert len(middlewares) == 13
+    assert len(middlewares) == 14
     assert isinstance(middlewares[0], FakeMiddleware)  # InputSanitizationMiddleware stub
     assert isinstance(middlewares[1], ToolOutputBudgetMiddleware)
     assert any(isinstance(m, ToolErrorHandlingMiddleware) for m in middlewares)
     # The token-budget backstop is attached by default so the cap engages (#3875).
     assert any(isinstance(m, TokenBudgetMiddleware) for m in middlewares)
-    assert isinstance(middlewares[-1], SafetyFinishReasonMiddleware)
+    assert any(isinstance(m, SafetyFinishReasonMiddleware) for m in middlewares)
+    assert isinstance(middlewares[-1], DurableContextMiddleware)
```
willem-bd: `middlewares[-1] is DurableContextMiddleware` only holds because the default `_make_app_config()` has `summarization.enabled=False` (so `create_summarization_middleware` returns `None`). When summarization is enabled — the actual fix scenario — the last middleware is `DeerFlowSummarizationMiddleware`, not `DurableContextMiddleware`.

The placement is covered separately by the `index(durable) < index(sentinel)` assertion, so this is fine functionally, but a one-line comment noting "last only when summarization is disabled" would prevent future confusion.

backend/tests/test_tool_error_handling_middleware.py
```diff
@@ -662,7 +670,92 @@ def fake_create_summarization_middleware(*, app_config=None, keep=None, skip_mem
     # skip_memory_flush=True so subagent-internal turns are not flushed into the
     # PARENT thread's durable memory (#3875 Phase 3 review).
     assert captured["skip_memory_flush"] is True
-    assert sentinel in middlewares
+    durable = [middleware for middleware in middlewares if isinstance(middleware, DurableContextMiddleware)]
+    assert len(durable) == 1
+    assert durable[0]._skills_root == app_config.skills.container_path
```
willem-bd: Minor: `_skills_root` is `posixpath.normpath(container_path or DEFAULT)`, so this equality only holds when `container_path` is already normalized. A trailing slash or `.`/`..` in the config would fail it.

Test-only and low impact, but could compare against `posixpath.normpath(app_config.skills.container_path)` to be robust.

willem-bd (commented):
Reviewing the subagent durable-context fix for #4039.

**The core fix is correct and well-targeted.** `DurableContextMiddleware` is placed before `DeerFlowSummarizationMiddleware` (mirroring the lead chain), so its `before_model` capture runs pre-compaction and its `wrap_model_call` injection reads the freshly-populated `summary_text` after compaction — projecting the summary ahead of the preserved assistant/tool tail. Constructor args match the lead chain exactly, the middleware has no `consume_stop_reason` (so the Phase 2 guard-cap stop-reason channel is undisturbed), and the new graph-level regression reproduces the reported failure shape well.

**One substantive concern** (inline on the append site): the subagent chain adds `DurableContextMiddleware` without the `SystemMessageCoalescingMiddleware` the lead chain pairs with it. The injected authority `SystemMessage` can leave a two-system-message request that strict backends may reject — potentially on the same backends #4039 targets. Recommend adding the coalescer for parity, or confirming the target backend tolerates it.

**Minor notes** (inline): a test gap on strict-backend request shape, the `middlewares[-1]` assertion only holding when summarization is off, and a fragile `_skills_root` equality.

Also checked and sound: `skill_context` retention across compaction (only overwritten when non-empty), prompt-injection boundary preserved by reuse (authority contract + bounded/escaped hidden `HumanMessage`), and no new per-turn cost beyond what the lead chain already incurs (`extract_skills` walks messages each `before_model` — same as lead; worth noting for `max_turns=150` subagents but accepted).

backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py
ggnnggez: Fixed in 2d54ab2a. Confirmed the hazard is real and reachable: the subagent passes `create_agent(system_prompt=None)` and carries its prompt as a leading `SystemMessage` in `messages`, so `DurableContext._inject` inserts `SystemMessage(authority)` right after it → two adjacent system messages whenever `data_block` is non-empty (summary/skill_context present). Appended `SystemMessageCoalescingMiddleware` innermost on the subagent chain (after the optional summarization middleware, so it is unconditionally last), mirroring the lead chain. It merges every `SystemMessage` into one leading `system_message` on the outgoing request, and sits inner of `DurableContextMiddleware` so it observes the injected message.

backend/tests/test_tool_error_handling_middleware.py
ggnnggez: Fixed in 2d54ab2a. Added `test_subagent_chain_coalesces_durable_authority_system_message`, which drives the real `build_subagent_runtime_middlewares` output through a strict `_generate` and asserts `system_indices == [0]` (exactly one leading `SystemMessage`). I verified it bites: temporarily removing the coalescer append makes the strict model see `[System(base), System(authority), ...]` and the test fails. Also added the single-leading-system assertion to the existing compaction integration test.

backend/tests/test_tool_error_handling_middleware.py
ggnnggez: Resolved in 2d54ab2a. Rather than only commenting, the coalescer is now appended unconditionally after the optional summarization middleware, so `middlewares[-1]` is always `SystemMessageCoalescingMiddleware` regardless of `summarization.enabled` — the summarization-dependence ambiguity is gone. Updated the assertion to check the coalescer is last and that `DurableContextMiddleware` precedes it, with a comment noting durable was only last when summarization was off.

backend/tests/test_tool_error_handling_middleware.py
ggnnggez: Fixed in 2d54ab2a. Changed the assertion to compare against `posixpath.normpath(app_config.skills.container_path)`, matching `_normalize_skills_root`, so a trailing slash / `.` / `..` in config no longer breaks it.

WillemJiang (approved):

PR #4042, opened 2026-07-10
fix(subagents): classify LLM error fallbacks as failed

### LINKED ISSUE #4041: [bug] subagent executor records the run which return Error 400 as COMPLETED
## Problem

When a subagent model request fails, `LLMErrorHandlingMiddleware` converts the provider exception into an `AIMessage` with `additional_kwargs.deerflow_error_fallback=true`. The graph then ends normally, and `SubagentExecutor` unconditionally records the run as `COMPLETED`.

That produces `task_completed` / `subagent_status=completed`, so the UI shows a completed subagent even though its displayed result is an error such as:

```
LLM request failed: Error code: 400 - ...
```

This was observed on thread `4f433b2e-7379-497d-91e9-fcf672d59345`.

## Root cause

The executor distinguishes raised exceptions from successful graph completion, but it does not inspect the structured fallback marker already attached by `LLMErrorHandlingMiddleware`. A handled provider failure therefore looks like an ordinary final assistant response.

The frontend is following the backend structured status correctly and should not need to parse error text.

## Expected behavior

- If the final subagent AI message has `deerflow_error_fallback=true`, terminate the subagent as `failed`.
- Surface a useful error through the existing `subagent_error` contract.
- Emit `task_failed` and persist `subagent.end(status=failed)`.
- Do not classify ordinary assistant text by matching the phrase `LLM request failed`.
- Preserve existing behavior for clean completions and additive guardrail `stop_reason` values.

## Proposed approach

Detect the structured fallback marker at the subagent executor terminalization boundary and map it to the existing `SubagentStatus.FAILED` path. No frontend status enum or configuration change should be necessary.

## Validation

Add regression coverage proving:

1. A structured LLM error fallback becomes `failed`.
2. Similar plain text without the marker remains `completed`.
3. Existing normal completion and guardrail-cap behavior remains unchanged.

ggnnggez:
## Summary

- classify terminal subagent `AIMessage` fallbacks marked with `deerflow_error_fallback=true` as `SubagentStatus.FAILED`
- surface the middleware-generated user-facing error through the existing `subagent_error` path
- keep ordinary error-looking assistant prose as a completed result when the structured marker is absent
- document the terminal-status contract for users and maintainers

## Root cause

`LLMErrorHandlingMiddleware` intentionally converts provider/model exceptions into marked `AIMessage` responses so an agent graph can end cleanly. `SubagentExecutor` previously treated every clean graph termination as `COMPLETED`, even when the final message was that structured error fallback. The task tool then emitted `task_completed`, and the Web UI correctly rendered the incorrect backend status.

The executor now translates the existing marker at its terminalization boundary. That reuses the normal `FAILED` → `task_failed` → `subagent_status=failed` flow, so no frontend status enum or text parsing is needed.

Closes #4041.

## Validation

- `uv run pytest tests/test_subagent_executor.py tests/test_task_tool_core_logic.py tests/test_subagent_status_contract.py -q` — 126 passed
- `uv run ruff check .` — passed
- `uv run ruff format --check .` — 802 files already formatted
- `make test` — 6881 passed, 24 skipped, 6 unrelated local-environment failures: five auth-secret persistence/config isolation assertions and one repository-root detector assertion; all subagent-related tests passed


willem-bd (commented):
Reviewed the head commit. The fix is correct and well-scoped — I verified the load-bearing assumptions:

- `LLMErrorHandlingMiddleware` **is** in the subagent middleware chain (`build_subagent_runtime_middlewares` → `_build_runtime_middlewares` appends it at `tool_error_handling_middleware.py:200`), so the fallback path is actually reachable in production.
- Marker keys match exactly (`deerflow_error_fallback` bool + `error_detail` str) against `_build_error_fallback_message`.
- `try_set_terminal` is first-terminal-wins, so it's race-safe against the concurrent timeout/cancellation paths.
- No status enum or `contracts/subagent_status_contract.json` change, so v1 consumers are unaffected.

The inline comments below are all nits — the substantive one is pinning down *why* last-message-only is correct given the shared-`thread_id` replay.

One design note (not a blocker): the LLM-fallback path is unconditionally `FAILED` with no partial-result recovery, unlike the `GraphRecursionError` path which surfaces surviving work as `COMPLETED + stop_reason=turn_capped`. Defensible — a dead provider is a different failure class than an exhausted turn budget — but a subagent that did N useful turns then hit a transient provider death on turn N+1 discards all of it. If preserving partial work on provider death is ever wanted, this is the spot.

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -192,6 +192,41 @@ def _extract_final_result(final_state: Any, *, trace_id: str, name: str) -> str:
...
 
+def _extract_llm_error_fallback(final_state: Any) -> str | None:
+    """Return the user-facing error for a terminal LLM fallback message.
+
+    ``LLMErrorHandlingMiddleware`` converts provider exceptions into marked
+    ``AIMessage`` objects so the graph can terminate cleanly. Clean graph
+    termination is not task success, however: subagent callers need the
+    structured marker translated into the existing failed terminal state.
+
+    Only the last assistant message is authoritative. Earlier handled errors
```
willem-bd: The "last message only" check is what makes this safe, but the stated rationale doesn't match the actual guarantee. A fallback `AIMessage` carries no `tool_calls`, so it always **terminates** the run — there is no "later assistant turn" to recover within a single subagent invocation.

The real reason last-only is correct: subagents **share the parent's `thread_id`** (`run_config["configurable"] = {"thread_id": self.thread_id}` in `_aexecute`), and LangGraph replays the full message history through `stream_mode="values"`. The lead-agent run path handles this exact hazard by masking stale markers via `pre_existing_message_ids` (`runtime/runs/worker.py::_extract_llm_error_fallback_message`) — because *that* path scans **all** messages. Here, since a subagent always appends ≥1 AIMessage of its own, the terminal assistant message is never a stale parent-history marker, so no masking is needed.

Worth stating explicitly (and cross-referencing worker.py) so a future maintainer doesn't "fix" this by scanning all messages and reintroducing the stale-marker false positive.

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -777,12 +812,20 @@ async def _aexecute(self, task: str, result_holder: SubagentResult | None = None
             # happened so we can mark the completed result with the cap reason
             # (token_capped / loop_capped) for the lead (#3875 Phase 2).
             stop_reason = self._consume_guard_stop_reason()
-            result.try_set_terminal(
-                SubagentStatus.COMPLETED,
-                result=final_result,
-                stop_reason=stop_reason,
-                token_usage_records=token_usage_records,
-            )
+            llm_error = _extract_llm_error_fallback(final_state)
```
willem-bd, line 824: Two values are computed/popped and then discarded on the `FAILED` branch:

- `final_result` (line above) is extracted but unused when `llm_error` is set.
- `stop_reason` (line 814) is **popped** from the guard middlewares via `consume_stop_reason`, but is never passed to `try_set_terminal` on this branch.

The `stop_reason` pop is the surprising one — it has a side effect on a path that doesn't use the value. Consider moving both `final_result` and `stop_reason` into the `else` branch so only the values actually used get computed.

(Harmless today: a fallback returns no `tool_calls`, so no guard hard-stop can have co-occurred — the popped reason would be `None` anyway.)

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -192,6 +192,41 @@ def _extract_final_result(final_state: Any, *, trace_id: str, name: str) -> str:
...
+            return None
+
+        content = message_content_to_text(message.content).strip()
+        if content:
+            return content
+
+        detail = metadata.get("error_detail")
+        if isinstance(detail, str) and detail.strip():
+            return detail.strip()
+        return "LLM request failed"
```
willem-bd, line 241: These two fallbacks (`error_detail`, then `"LLM request failed"`) are unreachable in practice: `LLMErrorHandlingMiddleware._build_error_fallback_message` always sets a non-empty `content` (the user-facing message) **and** a non-empty `error_detail` (`_extract_error_detail` falls back to the class name).

Fine as defensive guards — a one-line comment noting "defensive; middleware always populates content" would signal intent and stop a reader from assuming there's a real code path producing an empty fallback.

backend/tests/test_subagent_executor.py
```diff
@@ -714,6 +714,58 @@ async def test_aexecute_success(self, classes, base_config, mock_agent, msg):
...
+        with patch.object(executor, "_create_agent", return_value=mock_agent):
+            result = await executor._aexecute("Do something")
+
+        assert result.status == SubagentStatus.FAILED
+        assert result.error == fallback_text
+        assert result.result is None
+        assert result.stop_reason is None
+
+    @pytest.mark.anyio
+    async def test_aexecute_does_not_infer_llm_failure_from_message_text(self, classes, base_config, mock_agent, msg):
```
willem-bd, line 752: Both new tests use a fresh `final_state` with no replayed history. Given subagents share the parent's `thread_id` and LangGraph replays history, the correctness of "last AIMessage only" rests on a stale parent-history marker **not** being terminal — and that invariant isn't locked in here.

Suggest adding a case like:
```python
final_state = {"messages": [stale_fallback_aimessage, msg.human("..."), msg.ai("real result")]}
```
→ asserts `COMPLETED`, `result == "real result". It's the regression guard for the whole "no masking needed" argument in `_extract_llm_error_fallback`.

backend/packages/harness/deerflow/subagents/executor.py
ggnnggez: Fixed in 407f33f2. Rewrote the docstring to state the real guarantee: a fallback `AIMessage` carries no `tool_calls` so it always terminates the run, and a subagent always appends its own terminal assistant message — so the last `AIMessage` is never a stale parent-history marker replayed via the shared `thread_id`. Cross-referenced `worker.py::_extract_llm_error_fallback_message` and added an explicit warning against "fixing" this by scanning all messages.

backend/packages/harness/deerflow/subagents/executor.py
ggnnggez: Fixed in 407f33f2. Moved both `final_result` (`_extract_final_result`) and the `stop_reason` pop (`_consume_guard_stop_reason`) into the `else` (COMPLETED) branch, so the side-effecting pop no longer fires on the FAILED path where the value was discarded. Added a comment noting a fallback carries no `tool_calls`, so no guard hard-stop can co-occur on that branch.

backend/packages/harness/deerflow/subagents/executor.py
ggnnggez: Fixed in 407f33f2. Added a `# Defensive:` comment above those branches noting `_build_error_fallback_message` always sets a non-empty `content` and `error_detail` (`_extract_error_detail` falls back to the exception class name), so the paths only guard against a future middleware emitting an empty fallback.

backend/tests/test_subagent_executor.py
ggnnggez: Added in 407f33f2 as `test_aexecute_ignores_stale_parent_history_fallback_marker`: `final_state = {"messages": [stale_fallback, msg.human("Do something"), msg.ai("real result")]}` → asserts `COMPLETED` and `result == "real result"`. This locks the "no masking needed" invariant behind the tail-only scan.

WillemJiang (approved):

PR #4049, opened 2026-07-10
feat(subagents): show effective model and token usage on task cards

### LINKED ISSUE #4048: [feat] Show effective model and token usage on sub-agent task cards
## Problem

When the lead agent delegates to sub-agents, the collapsed sub-agent task cards in the Web UI show the task description and status, but not **which model** the sub-agent ran on or **how many tokens** it consumed. For deep-research runs that fan out into many sub-agents, there's no per-sub-agent visibility into model selection or token spend without digging into traces — and after a page reload that context is gone entirely.

## Proposal

Surface runtime metadata on each sub-agent task card:

- **Effective model** — the DeerFlow model actually selected for that sub-agent run (which can differ from the lead agent's model).
- **Cumulative token usage** — a running token total that updates after each completed sub-agent LLM call while the sub-agent is in flight, when the provider returns usage metadata.

Requirements:

- **Live updates during the run** — the collapsed card reflects the model as soon as the sub-agent starts and the token total as it climbs, without re-accounting the parent run's totals.
- **Survives reload** — the model and final token total are restored from persisted terminal metadata after a page refresh, so historical cards stay informative.
- **Additive / backward-compatible** — absent provider usage stays absent (never estimated as zero); older frontends that don't read the new fields are unaffected.
- **Token display gated** — the token total respects the existing `token_usage` config switch.

## Notes

- The effective model can ride on the existing `task_started` event; cumulative usage on `task_running`; and both on the terminal `ToolMessage` metadata (`subagent_model_name` / `subagent_token_usage`) plus the persisted `subagent.end` run event for reload.
- Known limitation to track separately: on a **user-stop + reload**, the cancel path produces no terminal `ToolMessage` and the reload backfill reads only step events, so metadata for a stopped sub-agent is not restored after refresh. This needs coordinated backend + frontend work and is out of scope for the initial implementation.

A PR implementing this follows.

ggnnggez:
Closes #4048.

## Summary

Adds runtime metadata (effective model + cumulative token usage) to sub-agent task cards.

Two commits:

1. **`feat(subagents): show runtime metadata on task cards`** — collapsed sub-agent cards show the effective model and, when the provider returns usage metadata, a cumulative token total that updates after each completed sub-agent LLM call and persists after reload. `task_started` carries the effective `model_name`; `task_running` carries a cumulative usage snapshot; terminal `ToolMessage` metadata (`subagent_model_name` / `subagent_token_usage`) restores the values after reload. Absent provider usage stays absent (never estimated as zero); the token total respects the `token_usage` config switch. The cross-language status contract is extended additively so older consumers are unaffected.

2. **`fix(subagents): stop task-card render loop and dedupe model fetches`** — code-review fixes (below).

## Review fixes in commit 2

**P1 — infinite render loop.** The terminal `ToolMessage` is re-parsed on every `MessageList` render and always carries `modelName`/`usage`, so the presence-based `setTasks` condition fired a fresh state object each render → `Maximum update depth exceeded`. Fix:
- `computeNextSubtask` returns a `changed` flag that **value-compares** `usage` (a re-parse rebuilds it into a new object each render, so reference comparison would always look changed).
- A new pure `subtaskNotification(task, {becameTerminal, changed})` → `"eager" | "deferred" | "none"` routes terminal transitions through the deferred after-render path and drops no-op re-parses, replacing the loop-prone inline condition.

**P2 — per-card `useModels` refetch.** Each subtask card mounted its own `useModels()` observer with default `staleTime: 0`, so newly-mounted cards refetched `/api/models`. Added `staleTime: Infinity` (model config is session-stable) so all cards share one fetch.

## Known limitation (follow-up)

Runtime metadata is **not** restored after a **user-stop + reload**: the `except asyncio.CancelledError` path produces no terminal `ToolMessage`, and the reload backfill reads only `subagent.step` events, so a backend-only change would write data nothing reads on reload. A complete fix needs coordinated backend + frontend work and is deferred (noted in #4048).

## Testing

- New unit tests in `subtask-update.test.ts` pin the render-loop regression (`changed` value-comparison + all three `subtaskNotification` routing cases, including the terminal re-parse → `"none"` case).
- `pnpm test tests/unit/core/tasks`, `pnpm typecheck`, `pnpm lint`, and the backend subagent suites pass.

> The render-loop fix is anchored at the state-management layer (a pure, testable function) since a full React-render regression test would require adding `@testing-library/react`, which is not currently a dependency.

---

Note: the branch is based on `main` a few commits back; the i18n locale files and README/AGENTS may need a rebase against current `main` (voice-dictation PR #4036 touched overlapping files). Happy to rebase.


backend/packages/harness/deerflow/subagents/step_events.py
```diff
@@ -244,3 +251,15 @@ def subagent_run_event(chunk: Any) -> dict[str, Any] | None:
         }
 
     return None
+
+
+def _normalized_usage(value: Any) -> dict[str, int] | None:
```
willem-bd: **Duplicated token-usage normalization.**

This `_normalized_usage` is identical to `status_contract.py::_normalize_token_usage` — same three keys, same bool/negative/missing-key rejection. Worth extracting to one shared helper so the two can't drift; e.g. one side later accepting `cache_tokens` or a `prompt_tokens` alias while the other rejects it would silently drop usage on one path.

(The same shape is mirrored again on the frontend — `lifecycle.ts::normalizeUsage` and `subtask-result.ts::readStructuredTokenUsage` — so one shared validator per side covers all four.)

frontend/src/core/tasks/lifecycle.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  return null;
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null;
+}
+
+function normalizeUsage(value: unknown) {
```
willem-bd: **Mirror of `subtask-result.ts::readStructuredTokenUsage`.**

The same `{input,output,total}_tokens` validation (plus `nonNegativeNumber`) lives in both modules. If these drift — e.g. one accepts an extra token field — a running-event usage and a terminal-metadata usage would normalize differently. Consider a single shared `normalizeUsage`.

backend/packages/harness/deerflow/tools/builtins/task_tool.py
```diff
@@ -434,17 +455,26 @@ async def task_tool(
                             "message": message,
                             "message_index": i + 1,  # 1-based index for display
                             "total_messages": current_message_count,
+                            "usage": usage,
```
willem-bd, line 458: **Live usage only refreshes when a new AI message arrives.**

`task_running` is emitted inside the `for i in range(last_message_count, current_message_count)` loop, so the card's token total updates per new assistant message. Meanwhile `update_token_usage_records` fires on *every* super-step — so during a tool-execution-only iteration the freshest snapshot isn't pushed to the card until the next AI turn. The terminal snapshot is authoritative, so this is a live-UX cadence gap, not a correctness bug.

Worth confirming it matches "updates after each completed sub-agent LLM call": it holds for calls that yield an AI message (should be all of them), but a tool-call-only turn wouldn't refresh the live counter.

frontend/src/core/threads/hooks.ts
```diff
@@ -1016,6 +1017,19 @@ export function useThreadStream({
...
     },
     onCustomEvent(event: unknown) {
+      const taskUpdate = taskEventToSubtaskUpdate(event);
+      if (taskUpdate) {
+        updateSubtask(taskUpdate);
+        if (
+          typeof event === "object" &&
+          event !== null &&
+          "type" in event &&
+          event.type === "task_started"
```
willem-bd: **Minor: re-narrows the event type after `taskEventToSubtaskUpdate` already validated it.**

`taskEventToSubtaskUpdate` only returns non-null for `task_started` / `task_running` and has already checked `event.type`, so this `typeof event === "object" && ... && event.type === "task_started"` block is a second source of truth for "is this a started event." It could be derived from the returned `taskUpdate` (e.g. `modelName` present without `usage`) or a small helper, so the two checks can't drift.

willem-bd (commented):
Posted a few inline comments. Summary:

- **Rebase required (merge blocker):** the branch is based on `main` a few commits back and overlaps current `main` — `task_tool.py` already references `effective_model` in a `writer(...)` around line 357 on `main`, and the i18n / README / AGENTS files overlap with the voice-dictation PR (#4036). Worth regenerating the diff against current `main` before merge so the `task_started` / `effective_model` region doesn't double-add the field on rebase.
- **DRY:** token-usage normalization is duplicated 2× backend (`step_events.py` / `status_contract.py`) and 2× frontend (`lifecycle.ts` / `subtask-result.ts`).
- **Live-usage cadence:** `task_running` (and thus the card's live total) only fires on new AI messages — fine in practice, flagged to confirm intent.
- **Nit:** `onCustomEvent` re-narrows the event type that `taskEventToSubtaskUpdate` already validated.

Overall a well-executed, additive change. The render-loop fix is correctly reasoned and pinned by tests, cumulative-snapshot semantics defend against double-counting/reorder, and `staleTime: Infinity` is safe (no `useModels` consumer invalidates the query). Approve pending the rebase.

WillemJiang:
@ggnnggez  please reslove the conficts with the main branch.

backend/packages/harness/deerflow/subagents/step_events.py
ggnnggez: Done in 53a61152 — extracted a single `normalize_token_usage()` into `status_contract.py`. Both `step_events.py` and the terminal-metadata builder call it now, so the two can no longer drift.

frontend/src/core/tasks/lifecycle.ts
ggnnggez: Done in 53a61152 — added `messages/usage.ts::normalizeTokenUsage`. Both `lifecycle.ts` (live `task_running`) and `subtask-result.ts::readStructuredTokenUsage` (terminal metadata) delegate to it now.

frontend/src/core/threads/hooks.ts
ggnnggez: Done in 53a61152 — `onCustomEvent` narrows `event.type` once into `eventType` and the redundant `task_started` re-check (already validated by `taskEventToSubtaskUpdate`) is dropped.

backend/packages/harness/deerflow/tools/builtins/task_tool.py
ggnnggez: Intended, and you read it right. `task_running` is the per-AI-message step channel, so the live counter refreshes on each completed sub-agent LLM call (every LLM call yields an AI message). A tool-execution-only super-step emits no new AI message, so it does not push a fresh live snapshot — but the terminal snapshot (and the persisted `subagent.end` metadata) is authoritative, so the final total is always correct, and gating the live push on new AI messages avoids re-emitting an unchanged cumulative snapshot on every super-step. Happy to add a dedicated usage-only refresh if you want tighter live cadence during long tool-only stretches.

ggnnggez:
@WillemJiang @willem-bd rebased onto current `main` and pushed. GitHub now reports the branch mergeable (the LLM-error-fallback fix #4042 had landed on `main`, which was the source of the `README.md` / `test_subagent_executor.py` overlap — those now carry both `main`'s content and this PR's additions).

Changes since the review (commit `53a61152`):

- **DRY:** one shared token-usage validator per side — backend `status_contract.normalize_token_usage()` (used by both the terminal metadata and `step_events.py`) and frontend `messages/usage.normalizeTokenUsage()` (used by both `lifecycle.ts` and `subtask-result.ts`).
- **Nit:** `onCustomEvent` narrows `event.type` once; the redundant `task_started` re-check is gone.
- **Live-usage cadence:** replied inline — it's intentional; the terminal snapshot is authoritative.

Verification: backend subagent suites (151 passed), frontend `tests/unit/core/tasks` (66 passed), `pnpm typecheck` / `pnpm lint` / `ruff` all clean.

WillemJiang (approved):

PR #4050, opened 2026-07-11
ci: add nightly build for images + helm chart

zhfeng:
## Summary

Adds a scheduled + manually-dispatchable workflow (`.github/workflows/nightly.yaml`) that builds and pushes the three DeerFlow container images (backend, frontend, provisioner) and the Helm chart to GHCR from `main`, so consumers can track unreleased changes between `v*` releases.

## Behavior

Mirrors the existing tag-driven release workflows (`container.yaml` + `chart.yaml`) but for unreleased `main`:

- **Triggers:** `schedule: cron "0 16 * * *"` (UTC 16:00, adjustable) + `workflow_dispatch`. `concurrency: nightly, cancel-in-progress: false` (never cancel a running nightly mid-push).
- **Images (amd64):** `ghcr.io/<owner>/deer-flow-{backend,frontend,provisioner}` tagged `:nightly` (rolling), `:nightly-YYYYMMDD` (dated), `:sha-<short>`. `latest` is **not** touched - it stays pinned to the last `v*` release. Backend still bakes `UV_EXTRAS=postgres`. Build provenance attested.
- **Chart:** version patched **in-workflow only** (never committed) to `<base>-nightly.YYYYMMDD` (valid semver prerelease; unique per day for OCI immutability), with `image.registry`/`image.tag` defaults repointed to the nightly build so the nightly chart self-installs against the matching nightly images.
- **Job graph:** `prepare` (date) -> `build-images` (matrix) || `validate-chart` -> `publish-chart` (gated on images + validation). No `verify-versions` gate (no `v*` tag to match).
- **Restricted to upstream:** every job is gated on `github.repository == 'bytedance/deer-flow'`, so scheduled runs / manual dispatches on forks skip all jobs.

Install the nightly chart:
```bash
helm install deer-flow oci://ghcr.io/bytedance/deer-flow --version 2.1.0-nightly.20260710
```

## Docs

`RELEASING.md` gains a "Nightly builds" section documenting the artifacts, install command, and the same-day re-run immutability caveat.

## Verification

- **Local:** dry-ran the chart `sed` patching (valid YAML), `helm lint` + `helm template` on the patched chart, workflow YAML parse.
- **Fork (`zhfeng/deer-flow`):** dispatched via `workflow_dispatch` (guard temporarily relaxed on a throwaway branch); all jobs green - images + chart pushed to `ghcr.io/zhfeng`, confirmed by anonymous `helm show` / `docker manifest inspect`.

## Tradeoffs / notes

- New self-contained file; the `v*` release workflows are untouched. Build steps intentionally duplicate `container.yaml` (isolation over a reusable-workflow refactor).
- **Same-day re-run:** `:nightly` image tags refresh (mutable); the chart version is OCI-immutable, so `helm push` fails on a same-day re-dispatch. Append a short SHA to the chart version if same-day re-publish is needed.
- amd64-only, matching the existing release builds. Multi-arch can be added later if there's demand.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,0 +1,202 @@
+name: Nightly Build
+
+# Nightly build of the three DeerFlow container images (backend, frontend,
+# provisioner) and the Helm chart, published to GHCR from the default branch.
+#
+# This mirrors the tag-driven release workflows (container.yaml + chart.yaml)
+# but trades the `v*` tag for date-based tags, since nightly ships unreleased
+# `main`. The `verify-versions` gate is intentionally skipped - there is no tag
+# to match - and `latest` is left untouched so it keeps tracking the last `v*`
+# release. Images are amd64-only to match the release builds.
+#
+# Artifacts (under the running repo's owner):
+#   ghcr.io/<owner>/deer-flow-{backend,frontend,provisioner}:nightly
+#   ghcr.io/<owner>/deer-flow-{backend,frontend,provisioner}:nightly-YYYYMMDD
+#   oci://ghcr.io/<owner>/deer-flow   chart version <base>-nightly.YYYYMMDD
+#
+# The nightly chart defaults image.tag=nightly and image.registry=ghcr.io/<owner>
+# (patched in-workflow, never committed), so installing it pulls the matching
+# nightly images with no values overrides.
+#
+# Restricted to the upstream repo: every job is gated on
+# `github.repository == 'bytedance/deer-flow'`, so a scheduled run or manual
+# dispatch on a fork skips all jobs rather than pushing to the fork's own GHCR
+# namespace. Scheduled workflows also only fire on the default branch.
+
+on:
+  schedule:
+    - cron: "0 16 * * *" # 16:00 UTC daily (adjust as needed)
+  workflow_dispatch:
+
+concurrency:
+  group: nightly
+  # Don't cancel a running nightly - a mid-cancel leaves a half-pushed set.
+  cancel-in-progress: false
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  prepare:
+    if: github.repository == 'bytedance/deer-flow'
+    runs-on: ubuntu-latest
+    outputs:
+      date: ${{ steps.date.outputs.date }}
+    steps:
+      - name: Set nightly date (UTC)
+        id: date
+        # Shared by build-images (image tag) and publish-chart (chart version)
+        # so the two stay in sync for a given run.
+        run: echo "date=$(date -u +%Y%m%d)" >> "$GITHUB_OUTPUT"
+
+  validate-chart:
+    if: github.repository == 'bytedance/deer-flow'
+    # Catch a broken render or a stale config_version before publish. A broken
+    # chart published under an OCI version is immutable (GHCR won't let you
+    # overwrite --version), so a regression must fail here, not on install.
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
+
+      - name: Lint chart
+        run: helm lint deploy/helm/deer-flow
+      - name: Validate templates render
+        run: helm template deer-flow deploy/helm/deer-flow --include-crds >/dev/null
+
+      # The chart's `config:` block embeds a config_version that must not fall
+      # behind config.example.yaml. Mirrors chart.yaml's drift check.
+      - name: config_version drift check
+        run: |
+          set -eu
+          example=$(grep -E '^config_version:[[:space:]]+[0-9]+' config.example.yaml | head -1 | awk '{print $2}')
+          chart=$(awk '/^config:[[:space:]]*\|/{f=1; next} f && /^[[:space:]]+config_version:[[:space:]]+[0-9]+/ {print $2; exit}' deploy/helm/deer-flow/values.yaml)
+          echo "config.example.yaml config_version=$example"
+          echo "chart values.yaml     config_version=$chart"
+          if [ -z "$example" ] || [ -z "$chart" ]; then
+            echo "::error::could not parse config_version from one of the files"
+            exit 1
+          fi
+          if [ "$chart" -lt "$example" ]; then
+            echo "::error::chart config_version ($chart) is behind config.example.yaml ($example). Bump 'config_version' in deploy/helm/deer-flow/values.yaml (and the README example) to $example."
+            exit 1
+          fi
+
+  build-images:
+    needs: prepare
+    if: github.repository == 'bytedance/deer-flow'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - component: backend
+            context: .
+            file: backend/Dockerfile
+            # Bake the `postgres` extra so multi-replica (K8s/Helm) deployments
+            # can use shared Postgres. Matches container.yaml's release image.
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,202 @@
...
+        run: |
+          set -eu
+          CHART=deploy/helm/deer-flow
+          BASE=$(grep -m1 '^version:' "$CHART/Chart.yaml" | awk '{print $2}')
+          NIGHTLY="${BASE}-nightly.${DATE}"
+          echo "Nightly chart version: ${NIGHTLY}"
+          sed -i "s|^version:.*|version: ${NIGHTLY}|" "$CHART/Chart.yaml"
+          sed -i "s|^appVersion:.*|appVersion: \"${NIGHTLY}\"|" "$CHART/Chart.yaml"
+          sed -i "s|^  registry: \"\".*|  registry: \"ghcr.io/${OWNER}\"|" "$CHART/values.yaml"
+          sed -i 's|^  tag: "latest"|  tag: "nightly"|' "$CHART/values.yaml"
```
willem-bd, line 177: These two `sed` patches (lines 183–184) are exact-string matches against the current `values.yaml`. `sed` exits 0 even when it matches zero lines, and `set -eu` won't catch that — so if `values.yaml` drifts (indentation, `'latest'` vs `"latest"`, or a non-empty default `registry`), the packaged chart ships with `registry: ""` / `tag: "latest"` and **silently does not pull the nightly images**, with no failure signal. The debug `echo` on line 186 aids eyeballing but isn't a gate.

Suggestion — assert the patch actually landed before packaging:
```bash
grep -q '^  registry: "ghcr.io/' "$CHART/values.yaml" || { echo "::error::registry sed did not apply"; exit 1; }
grep -q '^  tag: "nightly"' "$CHART/values.yaml"      || { echo "::error::tag sed did not apply"; exit 1; }
```
(or drive the defaults with `yq` instead of `sed`).

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,202 @@
...
+      - name: Patch chart to a nightly version + nightly image defaults
+        # Bumps Chart.yaml version/appVersion to <base>-nightly.YYYYMMDD (valid
+        # semver prerelease; unique per day for OCI immutability) and repoints
+        # the chart's default image registry/tag at the nightly build. These
+        # patches are in-workflow only - nothing is committed back.
+        run: |
+          set -eu
+          CHART=deploy/helm/deer-flow
+          BASE=$(grep -m1 '^version:' "$CHART/Chart.yaml" | awk '{print $2}')
+          NIGHTLY="${BASE}-nightly.${DATE}"
```
willem-bd: Since OCI chart versions are immutable, a same-day manual re-dispatch refreshes the mutable `:nightly` images but then **fails** at `helm push` (line 201), turning the run red. The PR body acknowledges this and suggests appending a short SHA. Worth resolving now rather than reacting to the first red nightly — e.g. `NIGHTLY="${BASE}-nightly.${DATE}-${GITHUB_SHA::7}"` keeps re-dispatch resilient at the cost of a slightly noisier version string.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,202 @@
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@902fa8ec7d6ecbf8d84d538b9b233a880e428804 #v5.7.0
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          # `nightly` rolls forward each run; `nightly-YYYYMMDD` pins a day.
```
willem-bd: Minor: GHCR image tags are mutable, so `:nightly-<date>` is only pinned *until* a same-day re-dispatch overwrites it — `:sha-<short>` is the only truly immutable tag. The comment "pins a day" slightly overstates the guarantee; a one-line caveat pointing reproducibility-minded users at `:sha-*` would set the right expectation.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,202 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
+
+      - name: Lint chart
+        run: helm lint deploy/helm/deer-flow
+      - name: Validate templates render
+        run: helm template deer-flow deploy/helm/deer-flow --include-crds >/dev/null
+
+      # The chart's `config:` block embeds a config_version that must not fall
+      # behind config.example.yaml. Mirrors chart.yaml's drift check.
```
willem-bd: This `config_version` drift-check block is now byte-for-byte duplicated from `chart.yaml` (which already runs it on every PR touching the chart/config, so `main` is already validated before nightly runs). It's cheap belt-and-suspenders and the isolation tradeoff is documented — just flagging that the parsing logic now lives in two workflows, so a future fix has to touch both.

willem-bd (commented):
## Review summary

Solid, well-documented workflow. The safety properties are all correct: `latest` is preserved (no `type=raw,value=latest`), `verify-versions` is correctly skipped (no tag to match), every job is fork-gated on `github.repository == 'bytedance/deer-flow'`, and `publish-chart` is gated on all images + validation. Pinned action SHAs and per-job least-privilege permissions match the existing release workflows, and the image/chart namespaces are self-consistent end-to-end.

The inline comments below are robustness/operability nits — the main one worth acting on before merge is the silent-no-op risk in the image-default `sed` patches (#1). #2 is a maintainer decision; the rest are minor.

.github/workflows/nightly.yaml
zhfeng: Addressed in bd7dcdd8. Added four `grep -q` assertions right after the seds (version, appVersion, registry, tag) that emit `::error::` and `exit 1` if any patch missed, so a drifted Chart.yaml/values.yaml can no longer silently ship a chart that pulls the release `latest` images.

Verified both directions locally: the gate passes on the current values.yaml, and fails loudly when I drift the registry line to a non-empty value the sed won't match. Kept sed rather than pulling in yq - the assertions make a silent no-op impossible.

.github/workflows/nightly.yaml
zhfeng: Addressed in bd7dcdd8. The nightly chart version is now `<base>-nightly.<date>-<short-sha>` (`SHORT_SHA="${GITHUB_SHA::7}"`), so each dispatch produces a unique OCI version and a same-day re-dispatch re-publishes cleanly instead of failing at `helm push`.

Confirmed `2.1.0-nightly.20260710-77a3652` parses as a valid semver prerelease and the patched chart lints + renders. Updated the RELEASING.md caveat (the "fails at helm push" note is gone, since this resolves it).

.github/workflows/nightly.yaml
zhfeng: Addressed in bd7dcdd8. Rewrote the image-tags comment to state that `:nightly-<date>` is pinned to a day but mutable within it (a same-day re-dispatch overwrites it), and `:sha-<short>` is the only truly immutable pin. Mirrored the same note in RELEASING.md's artifacts list.

.github/workflows/nightly.yaml
zhfeng: Addressed in bd7dcdd8. Extracted the drift check into `scripts/check_config_version.sh` (mirroring the existing `scripts/verify_versions.sh` convention - ROOT-resolved paths, `::error::` annotations) and both `chart.yaml` and `nightly.yaml` now `run: bash scripts/check_config_version.sh`, so the parsing logic lives in one place.

Also added the script to `chart.yaml`'s `pull_request.paths` so a script-only edit is still exercised on PRs. `validate-chart` passed on this PR, confirming the shared script runs correctly in both workflows.

WillemJiang (approved):

PR #4051, opened 2026-07-11
fix(sandbox): align AIO bash working directory

kongxiangxin:
## Problem

`AioSandboxProvider` supports local Docker containers and remote provisioner-managed Pods. Their default working directories depend on different runtime settings (`WORKDIR` versus Pod/container configuration), so the same relative bash command can run from different directories. This is especially visible for request-scoped environment commands, which use a fresh `bash.exec` session and cannot rely on prior shell cwd state.

## Solution

- Prefix non-local sandbox bash commands with `cd /mnt/user-data/workspace &&`.
- Apply the prefix before the channel identity prefix, so `export`/`unset DEERFLOW_CHANNEL_USER_ID` always executes even if the workspace mount is unavailable.
- Keep LocalSandbox behavior unchanged; it still uses the host thread workspace through `_apply_cwd_prefix`.
- Confirm the provisioner contract: thread `user-data` is mounted at `/mnt/user-data`, which provides `/mnt/user-data/workspace`.

## Review Follow-up

The initial CI failure was caused by existing channel identity tests asserting the pre-cwd command format. Those tests now cover the new ordering, dropped-user `unset` behavior, and the LocalSandbox host workspace path.

## Test Plan

- `pytest tests/test_channel_user_id_env.py tests/test_skill_request_scoped_secrets.py -q` (122 passed)
- `ruff check packages/harness/deerflow/sandbox/tools.py tests/test_channel_user_id_env.py tests/test_skill_request_scoped_secrets.py`
- Existing provisioner mount code verified in `docker/provisioner/app.py` (`/mnt/user-data`).

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/tools.py
+++ b/backend/packages/harness/deerflow/sandbox/tools.py
@@ -1741,6 +1741,7 @@ def bash_tool(runtime: Runtime, description: str, command: str) -> str:
                 max_chars,
             )
         ensure_thread_directories_exist(runtime)
+        command = f"cd {VIRTUAL_PATH_PREFIX}/workspace; {command}"
         if identity_prefix:
             command = identity_prefix + command
         try:
--- a/backend/tests/test_channel_user_id_env.py
+++ b/backend/tests/test_channel_user_id_env.py
@@ -91,13 +91,13 @@ class TestBashToolChannelIdentityPrefix:
         sandbox = _run_bash(monkeypatch, _aio_runtime({"channel_user_id": "ou_feishu_123"}))
 
         assert len(sandbox.calls) == 1
-        assert sandbox.calls[0]["command"] == f"export {CHANNEL_USER_ID_ENV}=ou_feishu_123; echo hi"
+        assert sandbox.calls[0]["command"] == f"export {CHANNEL_USER_ID_ENV}=ou_feishu_123; cd /mnt/user-data/workspace; echo hi"
         assert sandbox.calls[0]["env"] is None
 
-    def test_no_channel_user_id_leaves_command_unchanged(self, monkeypatch):
+    def test_no_channel_user_id_omits_identity_prefix(self, monkeypatch):
         sandbox = _run_bash(monkeypatch, _aio_runtime({"thread_id": "t1"}))
 
-        assert sandbox.calls[0]["command"] == "echo hi"
+        assert sandbox.calls[0]["command"] == "cd /mnt/user-data/workspace; echo hi"
         assert sandbox.calls[0]["env"] is None
 
     def test_per_call_identity_follows_current_context(self, monkeypatch):
@@ -114,10 +114,10 @@ class TestBashToolChannelIdentityPrefix:
         sandbox = _run_bash(monkeypatch, _aio_runtime({"channel_user_id": "x'; rm -rf /tmp/y; '"}))
 
         command = sandbox.calls[0]["command"]
-        assert command.endswith("; echo hi")
+        assert command.endswith("; cd /mnt/user-data/workspace; echo hi")
         # shlex.quote wraps the value; the raw injection payload must not appear
         # as executable syntax outside the quoted region.
-        assert "export " + CHANNEL_USER_ID_ENV + "='x'\"'\"'; rm -rf /tmp/y; '\"'\"''; echo hi" == command
+        assert "export " + CHANNEL_USER_ID_ENV + "='x'\"'\"'; rm -rf /tmp/y; '\"'\"''; cd /mnt/user-data/workspace; echo hi" == command
 
     def test_secrets_and_identity_compose(self, monkeypatch):
         """Active skill secrets keep the env= channel; the identity keeps the
@@ -132,7 +132,7 @@ class TestBashToolChannelIdentityPrefix:
 
         call = sandbox.calls[0]
         assert call["env"] == {"ERP_TOKEN": "secret-value"}
-        assert call["command"].startswith(f"export {CHANNEL_USER_ID_ENV}=ou_1; ")
+        assert call["command"] == f"export {CHANNEL_USER_ID_ENV}=ou_1; cd /mnt/user-data/workspace; echo hi"
         assert "secret-value" not in call["command"]
 
     def test_non_im_run_leaves_command_untouched(self):
@@ -159,8 +159,8 @@ class TestBashToolChannelIdentityPrefix:
         a = _run_bash(monkeypatch, _aio_runtime({"channel_user_id": "sender-a"}))
         b = _run_bash(monkeypatch, _aio_runtime({"channel_user_id": "b" * 5000}))
 
-        assert a.calls[0]["command"] == f"export {CHANNEL_USER_ID_ENV}=sender-a; echo hi"
-        assert b.calls[0]["command"] == f"unset {CHANNEL_USER_ID_ENV}; echo hi"
+        assert a.calls[0]["command"] == f"export {CHANNEL_USER_ID_ENV}=sender-a; cd /mnt/user-data/workspace; echo hi"
+        assert b.calls[0]["command"] == f"unset {CHANNEL_USER_ID_ENV}; cd /mnt/user-data/workspace; echo hi"
         assert b.calls[0]["env"] is None
 
     def test_windows_local_sandbox_skips_prefix(self, monkeypatch):


WillemJiang:
@kongxiangxin Thanks for your contribution. Please take a look at the Unit test errors.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -1694,6 +1694,7 @@ def bash_tool(runtime: Runtime, description: str, command: str) -> str:
         ensure_thread_directories_exist(runtime)
         if identity_prefix:
             command = identity_prefix + command
+        command = f"cd {VIRTUAL_PATH_PREFIX}/workspace && {command}"
```
willem-bd: **Ordering: this `cd ... && ` wrapper breaks the `identity_prefix` chain.** As written, an IM-channel run on a non-local sandbox produces `cd /mnt/user-data/workspace && export DEERFLOW_CHANNEL_USER_ID=<id>; <cmd>`, which the shell parses as `(cd ... && export ...) ; <cmd>`. If the `cd` fails, the `&&` short-circuits the `export`/`unset` — reopening the stale-group-chat-identity window `_channel_identity_prefix` exists to close ("a sender whose id is dropped inherits the previous sender's value") — while `<cmd>` still runs (via `;`) in the wrong directory.

Fix: apply the `cd` prefix **before** the identity prefix so identity is never gated on `cd` success:
```python
command = f"cd {VIRTUAL_PATH_PREFIX}/workspace && {command}"
if identity_prefix:
    command = identity_prefix + command
# -> "export ...; cd /mnt/user-data/workspace && <cmd>"
```
This keeps `export`/`unset` unconditional and leaves the user command fail-closed on a missing workspace.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -1694,6 +1694,7 @@ def bash_tool(runtime: Runtime, description: str, command: str) -> str:
         ensure_thread_directories_exist(runtime)
```
willem-bd, line 1743: **A missing `/mnt/user-data/workspace` silently fails the whole command.** Because the prefix is `cd X && <cmd>`, if the dir isn't present in the sandbox the command doesn't run at all (just a shell `cd` error). `ensure_thread_directories_exist` should guarantee it for AIO (host-created + volume-mounted), but the PR's stated motivation is the **provisioner/K8s path**, where the mount is "controlled by different runtime configuration." If the Pod doesn't mount at exactly `/mnt/user-data/workspace`, every bash command silently no-ops. Worth confirming the provisioner mounts at this exact virtual path, and/or surfacing a clearer error than a bare `cd` failure.

backend/tests/test_skill_request_scoped_secrets.py
```diff
@@ -943,11 +944,13 @@ def execute_command(self, command, env=None, timeout=None):
 
     def test_active_secret_forwarded_as_env(self):
         out, captured = self._run_bash({"__active_skill_secrets": {"ERP_TOKEN": "tok-123"}})
+        assert captured["command"] == "cd /mnt/user-data/workspace && echo hi"
```
willem-bd: **Coverage gap for a `risk:high` change.** These two assertions only exercise the **no-identity** path (context `{}` -> `_channel_identity_prefix` returns `None`) through a `FakeSandbox`. Not covered:
1. The headline claim that **local** bash does *not* receive `cd /mnt/user-data/workspace` (it should still use the host workspace path via `_apply_cwd_prefix`) — nothing pins this regression.
2. The **identity_prefix + `cd` composition** (the case flagged on `tools.py:1697`), which is the one real branching the new line introduces.

A local-path regression test + a `channel_user_id`-in-context test asserting both the `export`/`unset` and the `cd` are present and correctly ordered would close both gaps.

willem-bd (commented):
Reviewed the AIO bash cwd alignment. The scoping is correct — the new prefix is in the non-local branch only, so `LocalSandboxProvider` is genuinely unchanged (still uses `_apply_cwd_prefix` with the host workspace path), and only `bash_tool` is touched (`ls`/`read_file`/`write_file` resolve via path mapping, unaffected).

The main ask is the **ordering issue on `tools.py:1697`**: as written, the `cd ... && ` wrapper composes badly with `identity_prefix` and can silently drop the channel-user-id `export`/`unset` when the `cd` fails — a one-line reorder fixes it. Plus two smaller notes on workspace-existence and test coverage.

kongxiangxin:
Review follow-up applied in commit `3bc74826`.

- Moved the AIO cwd prefix before `identity_prefix`, so channel `export`/`unset` is unconditional and the user command remains gated by `cd ... &&`.
- Updated channel identity expectations and added a LocalSandbox regression test confirming the host workspace cwd remains in use.
- Confirmed provisioner mounts thread `user-data` at `/mnt/user-data`, so `/mnt/user-data/workspace` is the intended virtual workspace.
- Local targeted verification: 122 passed; Ruff passed.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -1692,6 +1692,7 @@ def bash_tool(runtime: Runtime, description: str, command: str) -> str:
                 max_chars,
             )
         ensure_thread_directories_exist(runtime)
+        command = f"cd {VIRTUAL_PATH_PREFIX}/workspace && {command}"
```
willem-bd: **Consider: failure mode when the workspace mount is absent.** With `&&`, if `/mnt/user-data/workspace` doesn't exist in a container, `cd` fails and the user command never runs — the agent only sees `cd: ...: No such file or directory`. The codebase already treats this path as canonical (the docstring points agents at `/mnt/user-data/workspace/.venv`, and the provisioner mounts `user-data` at `/mnt/user-data` per `docker/provisioner/app.py:388`), so the assumption is shared rather than new — but the blast radius is now *every* non-local bash command. No change required if the contract is trusted; a diagnostic message or guard would make a missing mount less opaque.

backend/tests/test_channel_user_id_env.py
```diff
@@ -91,13 +91,13 @@ def test_identity_exported_and_env_stays_none(self, monkeypatch):
         sandbox = _run_bash(monkeypatch, _aio_runtime({"channel_user_id": "ou_feishu_123"}))
 
         assert len(sandbox.calls) == 1
-        assert sandbox.calls[0]["command"] == f"export {CHANNEL_USER_ID_ENV}=ou_feishu_123; echo hi"
+        assert sandbox.calls[0]["command"] == f"export {CHANNEL_USER_ID_ENV}=ou_feishu_123; cd /mnt/user-data/workspace && echo hi"
         assert sandbox.calls[0]["env"] is None
 
     def test_no_channel_user_id_leaves_command_unchanged(self, monkeypatch):
```
willem-bd: **Nit: test name is now stale.** The command is no longer "unchanged" — it now carries the `cd /mnt/user-data/workspace &&` prefix. The underlying intent ("no `channel_user_id` → no identity `export` prefix") still holds, but the name misleads. Consider renaming to `test_no_channel_user_id_omits_identity_prefix`.

willem-bd (commented):
Reviewed the AIO bash cwd alignment. Core change is correct and well-scoped — identity-before-`cd` ordering is the right call (the `export`/`unset` runs unconditionally via `;`, while `&&` only short-circuits the user command if the mount is absent). Two inline notes below: a stale test name (nit) and a consideration on the silent-failure surface. Overall solid.

kongxiangxin:
Follow-up applied in commit `49c884b5`:

- Changed the non-local cwd wrapper from `cd /mnt/user-data/workspace && command` to `cd /mnt/user-data/workspace; command`. This preserves backward compatibility: if the workspace mount is absent, the shell still reports the native `cd` diagnostic, but an existing bash command that does not depend on the workspace can continue to run instead of being blocked by the newly added cwd setup.
- When the workspace exists, commands still start from `/mnt/user-data/workspace`.
- Renamed `test_no_channel_user_id_leaves_command_unchanged` to `test_no_channel_user_id_omits_identity_prefix`.
- The PR branch is based on the latest `main` at the time of the update.
- Targeted tests: 121 passed; Ruff passed.

backend/packages/harness/deerflow/sandbox/tools.py
kongxiangxin: We changed the wrapper to use `;` instead of `&&` in commit `49c884b5`.

The main reason is backward compatibility: adding the cwd setup must not make an existing bash command that does not depend on the workspace fail just because `/mnt/user-data/workspace` is unavailable in a particular container. With `;`, the shell still reports the native `cd` error, but the original command continues to run, preserving the previous execution behavior. When the workspace mount exists, commands still start from `/mnt/user-data/workspace` as intended.

We also renamed the stale test to `test_no_channel_user_id_omits_identity_prefix`. The targeted tests pass (121 passed), and the branch is based on the latest `main`.

WillemJiang (approved):

PR #4055, opened 2026-07-11
Fix require_mention gating on whitespace-only bot_login/mention_login

DaoyuanLi2816:
## Summary

A whitespace-only `github.bot_login` or per-trigger `mention_login` (e.g. `"   "`) silently defeats the `require_mention` fallback chain: the trigger gates on the literal whitespace string forever and can never fire, even when the real, working fallback handle is correctly `@mentioned`.

## The bug

AGENTS.md documents the full `require_mention` precedence chain as treating whitespace-only defaults as unset:

> Mention-handle precedence for `require_mention` triggers is `trigger.mention_login` → `github.bot_login` → `channels.github.default_mention_login` → `agent.name`. Whitespace-only defaults are treated as unset.

That contract was only half-implemented. `channels.github.default_mention_login` (the third link) is correctly normalized in `dispatcher.py`:

```python
operator_default = (operator_default_mention_login or "").strip() or None
default_mention_login = github.bot_login or operator_default or agent.name
```

— and is pinned by `test_operator_default_blank_string_treated_as_none`. But the two *higher-precedence* links were used raw:

- `backend/app/gateway/github/dispatcher.py`: `default_mention_login = github.bot_login or operator_default or agent.name` read `github.bot_login` unstripped.
- `backend/app/gateway/github/triggers.py`: `login = trigger.mention_login or default_mention_login` read `trigger.mention_login` unstripped.

A whitespace-only string is truthy in Python, so `"   " or anything` evaluates to `"   "` — the `or` chain never falls through. Concretely, an agent with `github.bot_login: "   "` and `require_mention: true` rejected every mention with `mention required for @   ` (a literal 3-space string no comment can ever contain), permanently and silently — even though its own self-event gate (`_is_self_event`, a few lines above in the same file) already defensively stripped this same field.

## The fix

Normalize both fields once, at the config layer, instead of patching each call site: `GitHubTriggerConfig.mention_login` and `GitHubAgentConfig.bot_login` (`packages/harness/deerflow/config/agents_config.py`) each get a `field_validator` that blank-to-nones a whitespace-only value, mirroring the `@field_validator` pattern already used elsewhere in the config package (`extensions_config.py`, `tool_search_config.py`):

```python
def _blank_to_none(value: str | None) -> str | None:
    if value is None:
        return None
    stripped = value.strip()
    return stripped or None
```

This fixes both `dispatcher.py` and `triggers.py` without touching their `or`-chains at all — every reader of `github.bot_login` / `trigger.mention_login` (today's and any future one) now sees an honest `None` instead of a literal whitespace string. Added a short comment at each existing `or`-chain noting the normalization now happens upstream, for anyone reading those call sites without `agents_config.py` open.

## Tests

Adds `test_bot_login_whitespace_only_treated_as_none` and `test_trigger_mention_login_whitespace_only_treated_as_none` to `test_github_dispatcher.py`, mirroring the existing `test_operator_default_blank_string_treated_as_none` shape (same fixtures, same `fanout_event` harness, no FastAPI dependency).

Verified fail-before / pass-after: reverting only the `agents_config.py` change reproduces the bug exactly — both new tests fail with `'reason': 'mention required for @   '` — and both pass again with the fix restored. Full suite: `test_github_dispatcher.py` 22 passed, `test_github_agents_config.py` 12 passed (no regressions from the added validators).

```
cd backend && PYTHONPATH=. pytest tests/test_github_dispatcher.py tests/test_github_agents_config.py
```

### DIFF
--- a/backend/app/gateway/github/dispatcher.py
+++ b/backend/app/gateway/github/dispatcher.py
@@ -201,6 +201,14 @@ async def fanout_event(
         # ``coder`` with ``require_mention: true`` and no per-trigger or
         # per-agent override silently required ``@coder`` mentions instead
         # of ``@deerflow-bot``.
+        #
+        # ``github.bot_login`` is normalized (whitespace-only -> None) by
+        # ``GitHubAgentConfig``'s field validator, so this ``or`` chain
+        # correctly falls through a misconfigured ``bot_login: "   "``
+        # instead of comparing mentions against a literal whitespace
+        # string. ``operator_default_mention_login`` is a plain function
+        # argument (not a validated model field), so it is normalized here
+        # explicitly.
         operator_default = (operator_default_mention_login or "").strip() or None
         default_mention_login = github.bot_login or operator_default or agent.name
         fire, reason = event_should_fire(event, payload, trigger, default_mention_login)
--- a/backend/app/gateway/github/triggers.py
+++ b/backend/app/gateway/github/triggers.py
@@ -186,6 +186,11 @@ def event_should_fire(
             return True, f"allow_authors:{author}"
 
     if trigger.require_mention:
+        # ``trigger.mention_login`` is normalized (whitespace-only -> None)
+        # by ``GitHubTriggerConfig``'s field validator, so this ``or`` falls
+        # through a misconfigured ``mention_login: "   "`` to
+        # ``default_mention_login`` instead of gating on a literal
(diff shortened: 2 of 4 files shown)


zhfeng (approved):
LGTM

WillemJiang (approved):

PR #4056, opened 2026-07-11
fix(subagents): classify recursion-capped LLM error fallbacks as failed

DaoyuanLi2816:
## Summary

#4042 taught `SubagentExecutor._aexecute`'s normal-completion path to recognize a terminal `AIMessage` marked by `LLMErrorHandlingMiddleware` (`additional_kwargs.deerflow_error_fallback=True`) via `_extract_llm_error_fallback`, and classify the run as `SubagentStatus.FAILED` instead of `COMPLETED`. The sibling `except GraphRecursionError:` block — hit when a subagent exhausts its `max_turns` budget — has its own, older logic that never calls `_extract_llm_error_fallback`, so the same fallback message is still misclassified as a completed task when it coincides with hitting the turn cap.

## The bug

`_aexecute`'s `except GraphRecursionError:` block decides `usable_partial` purely from whether the trailing `AIMessage` has non-empty text:

```python
messages = (final_state or {}).get("messages", [])
usable_partial: str | None = None
for m in reversed(messages):
    if isinstance(m, AIMessage):
        text = message_content_to_text(m.content).strip()
        if text:
            usable_partial = text
        break
records = collector.snapshot_records() if collector is not None else None
stop_reason = self._consume_guard_stop_reason() or "turn_capped"
if usable_partial is not None:
    result.try_set_terminal(
        SubagentStatus.COMPLETED,
        result=usable_partial,
        stop_reason=stop_reason,
        token_usage_records=records,
    )
else:
    ...
```

`_extract_llm_error_fallback` is never consulted here, unlike the normal-completion branch a few lines above it. A fallback `AIMessage` stamped by `LLMErrorHandlingMiddleware` always carries non-empty user-facing text (it's a rendered error explanation, e.g. `"LLM request failed: ..."`), so when the subagent's last turn before hitting `GraphRecursionError` happens to be that fallback, `usable_partial` is set to the fallback text and the run is classified `COMPLETED` — the exact bug #4042 fixed for the normal-completion path, still present on this one.

## The fix

Consult `_extract_llm_error_fallback` in this except-block too, before falling back to the raw-text scan, and classify `FAILED` with the same `stop_reason` (`turn_capped`, or a guard's reason if one already fired) the block already computes for its other branches:

```python
llm_error = _extract_llm_error_fallback(final_state)
if llm_error is not None:
    result.try_set_terminal(
        SubagentStatus.FAILED,
        error=llm_error,
        stop_reason=stop_reason,
        token_usage_records=records,
    )
else:
    messages = (final_state or {}).get("messages", [])
    usable_partial: str | None = None
    ...
```

Same contract #4042 established for the normal-completion path — only the structured marker is authoritative, ordinary error-looking prose without it still counts as a completed result — just applied to the second terminalization path that missed it.

## Tests

Adds `test_aexecute_recursion_error_with_llm_error_fallback_surfaces_failed`: streams a fallback `AIMessage` (`additional_kwargs={"deerflow_error_fallback": True, ...}`) and then raises `GraphRecursionError` on the next step, mirroring the existing `test_aexecute_recursion_error_with_partial_surfaces_completed_turn_capped` fixture style.

Verified fail-before / pass-after: the test asserts `SubagentStatus.FAILED`; without the fix it gets `COMPLETED` with the fallback text as `result.result` and `result.error is None`, and passes with the fix.

```
cd backend && PYTHONPATH=. pytest tests/test_subagent_executor.py tests/test_task_tool_core_logic.py tests/test_subagent_status_contract.py -q
```

128 passed (the same three suites #4042 validated with, plus this PR's new test).

`ruff format --check --line-length 240` and `ruff check --line-length 240` are clean on the changed files.

### DIFF
--- a/backend/packages/harness/deerflow/subagents/executor.py
+++ b/backend/packages/harness/deerflow/subagents/executor.py
@@ -865,30 +865,46 @@ class SubagentExecutor:
             # consistent and pops the reason so it is not orphaned in the dict.
             max_turns = self.config.max_turns
             logger.warning(f"[trace={self.trace_id}] Subagent {self.config.name} reached max_turns={max_turns} (GraphRecursionError); recovering partial result")
-            messages = (final_state or {}).get("messages", [])
-            usable_partial: str | None = None
-            for m in reversed(messages):
-                if isinstance(m, AIMessage):
-                    text = message_content_to_text(m.content).strip()
-                    if text:
-                        usable_partial = text
-                    break
             records = collector.snapshot_records() if collector is not None else None
             stop_reason = self._consume_guard_stop_reason() or "turn_capped"
-            if usable_partial is not None:
+
+            # A handled LLM provider failure (#4042) carries non-empty
+            # user-facing text on its terminal ``AIMessage`` just like genuine
+            # partial output, so it must be checked here too or it is
+            # indistinguishable from the raw-text scan below and gets
+            # misclassified as a completed task. Consult the same marker the
+            # normal-completion path above uses, before falling back to that scan.
+            llm_error = _extract_llm_error_fallback(final_state)
+            if llm_error is not None:
                 result.try_set_terminal(
-                    SubagentStatus.COMPLETED,
-                    result=usable_partial,
+                    SubagentStatus.FAILED,
+                    error=llm_error,
                     stop_reason=stop_reason,
                     token_usage_records=records,
                 )
             else:
-                result.try_set_terminal(
-                    SubagentStatus.FAILED,
-                    error=f"Reached max_turns={max_turns}",
-                    stop_reason=stop_reason,
-                    token_usage_records=records,
-                )
+                messages = (final_state or {}).get("messages", [])
+                usable_partial: str | None = None
+                for m in reversed(messages):
+                    if isinstance(m, AIMessage):
+                        text = message_content_to_text(m.content).strip()
+                        if text:
+                            usable_partial = text
+                        break
+                if usable_partial is not None:
+                    result.try_set_terminal(
+                        SubagentStatus.COMPLETED,
+                        result=usable_partial,
+                        stop_reason=stop_reason,
+                        token_usage_records=records,
+                    )
+                else:
+                    result.try_set_terminal(
+                        SubagentStatus.FAILED,
+                        error=f"Reached max_turns={max_turns}",
+                        stop_reason=stop_reason,
+                        token_usage_records=records,
+                    )
 
         except Exception as e:
             logger.exception(f"[trace={self.trace_id}] Subagent {self.config.name} async execution failed")
--- a/backend/tests/test_subagent_executor.py
+++ b/backend/tests/test_subagent_executor.py
@@ -1143,6 +1143,56 @@ class TestAsyncExecutionPath:
         assert result.completed_at is not None
 
     @pytest.mark.anyio
+    async def test_aexecute_recursion_error_with_llm_error_fallback_surfaces_failed(self, classes, base_config, mock_agent, msg):
+        """A structured LLM error fallback that coincides with hitting
+        ``max_turns`` must still classify as ``failed``, not ``completed``.
+
+        ``_extract_llm_error_fallback`` (#4042) marks a terminal ``AIMessage``
+        as a handled provider failure via
+        ``additional_kwargs.deerflow_error_fallback``, and the
+        normal-completion branch above already consults it before falling
+        back to ``_extract_final_result``. This except-block must apply the
+        same check before recovering ``usable_partial`` from raw non-empty
+        ``AIMessage`` text: a fallback message always carries non-empty
+        user-facing text, so without checking the marker first it is
+        indistinguishable from genuine partial output and gets misclassified
+        as a completed task rather than the failed provider error it is.
+        """
+        from langgraph.errors import GraphRecursionError
+
+        AIMessage = classes["AIMessage"]
+        SubagentExecutor = classes["SubagentExecutor"]
+        SubagentStatus = classes["SubagentStatus"]
+
+        fallback_text = "LLM request failed: provider rejected the request"
+        fallback_message = AIMessage(
+            content=fallback_text,
+            additional_kwargs={
+                "deerflow_error_fallback": True,
+                "error_type": "BadRequestError",
+                "error_reason": "generic",
+                "error_detail": "Error code: 400 - InvalidParameter",
+            },
+        )
+        fallback_state = {"messages": [msg.human("Task"), fallback_message]}
+
+        async def mock_astream(*args, **kwargs):
+            yield fallback_state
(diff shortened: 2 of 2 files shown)


ggnnggez (approved):
### Code review — Approve

Correct, well-scoped follow-up to #4042. It closes the second terminalization path that `_extract_llm_error_fallback` never covered: the `except GraphRecursionError` block now consults the same structured marker before the raw-text scan, mirroring the normal-completion path.

https://github.com/bytedance/deer-flow/blob/63a1d591c66e7d68af9c5f21f8b5ad3fdf170fa7/backend/packages/harness/deerflow/subagents/executor.py#L873-L886

**Reachability empirically verified.** I built a real `langchain.create_agent` graph with the actual `LLMErrorHandlingMiddleware` (`wrap_model_call`, produces the fallback inside the model node) plus a real `after_model` middleware, and drove a scripted model that fails on the turn where `recursion_limit` is hit. `GraphRecursionError` is raised while `final_state`'s tail is the fresh fallback marker (`deerflow_error_fallback=True`, no `tool_calls`, `content="LLM request failed: ..."`). Pre-fix logic scans that tail and classifies `COMPLETED` with the error text as `result`; post-fix classifies `FAILED`. Notably the collision reproduces even without the extra `after_model` node — it is intrinsic to `create_agent`'s multi-super-step-per-turn structure vs. `recursion_limit` counting super-steps, so this is not a narrow edge case.

**Not a partial-work regression.** This extends #4042's `fallback → FAILED, no partial recovery` choice to the recursion path. The old block's `for m in reversed(messages): ... break` only ever inspected the last `AIMessage`, which in the collision case is the error text — so no genuinely useful earlier turn was ever surfaced as `result`. The change strictly improves it (error-as-`COMPLETED-result` → error-as-`FAILED`). Re: willem-bd's #4042 note about preserving partial work on provider death — if that is ever wanted, both terminal paths would need it together.

**Non-blocking:** the fallback-`FAILED` branch stamps `stop_reason=turn_capped`. The primary failure cause is the LLM error rather than the turn cap, so `turn_capped` is slightly imprecise, but `stop_reason` is additive/informational, the run genuinely did hit the recursion limit, and `FAILED + turn_capped` is already an existing combination — acceptable as-is.

Contract-safe (no status enum / `subagent_status_contract.json` change), `try_set_terminal` first-terminal-wins preserved, and TDD test covers fail-before/pass-after.

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -865,30 +865,46 @@ async def _aexecute(self, task: str, result_holder: SubagentResult | None = None
...
-                )
-            else:
+
+            # A handled LLM provider failure (#4042) carries non-empty
+            # user-facing text on its terminal ``AIMessage`` just like genuine
+            # partial output, so it must be checked here too or it is
+            # indistinguishable from the raw-text scan below and gets
+            # misclassified as a completed task. Consult the same marker the
+            # normal-completion path above uses, before falling back to that scan.
+            llm_error = _extract_llm_error_fallback(final_state)
```
willem-bd, line 877: **Worth validating in a real run that this coincidence is reachable.** The test forces the sequence (yield fallback state, then raise `GraphRecursionError`), so it proves the classification logic but not that the live graph produces this ordering. `_extract_llm_error_fallback`'s own docstring says a fallback "always terminates the run," which at face value makes this path look unreachable. It *should* be reachable — LangGraph checks the recursion limit at the super-step boundary, so a fallback produced on the turn that lands on `max_turns` trips the limit on the next transition before clean termination finalizes — but it'd be good to confirm with a real subagent run where the provider error lands on the final allowed turn. (This is the real-path check the `needs-validation` label is asking for.)

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -865,30 +865,46 @@ async def _aexecute(self, task: str, result_holder: SubagentResult | None = None
...
+            if llm_error is not None:
                 result.try_set_terminal(
                     SubagentStatus.FAILED,
-                    error=f"Reached max_turns={max_turns}",
+                    error=llm_error,
                     stop_reason=stop_reason,
```
willem-bd, line 882: **Intentional asymmetry to confirm.** This FAILED branch passes `stop_reason` (`turn_capped` or a guard reason), whereas the normal-completion FAILED branch above omits it — its comment notes a fallback + guard hard-stop can't co-occur. Here it's deliberate and more informative, since the turn cap is the binding context that put us in this except block, and the test pins `stop_reason == "turn_capped"`. Flagging so reviewers can confirm the lead agent treats `status=FAILED` as taking precedence over `stop_reason=turn_capped` (additive color, not "out of budget").

willem-bd (commented):
Reviewed the recursion-capped fallback classification fix. Correct and well-scoped — applies the same marker-first contract #4042 established for the normal-completion path to the `GraphRecursionError` branch; the non-fallback path is behaviorally unchanged (`final_state` is None-safe via the line-783 init; `_consume_guard_stop_reason` still consumed once). Two inline notes: a request for the real-path validation the `needs-validation` label already asks for, and a confirmation point on the `stop_reason` asymmetry between the two FAILED branches.

DaoyuanLi2816:
@willem-bd Confirmed the coincidence is reachable in a real run, from the executor side (corroborating @ggnnggez's `create_agent` repro).

Drove the real `SubagentExecutor._aexecute` end-to-end through the real `create_agent` graph + real subagent middleware chain (real `LLMErrorHandlingMiddleware`), with a scripted model that emits a `bash` tool call for the first two turns, then raises a non-retriable provider error on the third. The middleware catches it in `awrap_model_call` and returns the marked fallback (`deerflow_error_fallback=True`, no `tool_calls`, content `"LLM request failed: ..."`). Because `create_agent` spends several super-steps per turn, sweeping `max_turns` puts the fallback on the final allowed turn at `max_turns` 17–21 (error on model-call 3): `GraphRecursionError` is raised from `astream` with that fresh fallback as the tail of the last streamed chunk — i.e. the `except GraphRecursionError` block, with `final_state`'s tail confirmed as the marker (`AIMessage`, `tool_calls=[]`, `deerflow_error_fallback=True`).

At `max_turns=18`:
- origin/main: `status=COMPLETED`, `result="LLM request failed: provider rejected the request"`, `error=None` — the provider-error text is presented as the subagent's successful output.
- this branch: `status=FAILED`, `error="LLM request failed: provider rejected the request"`, `result=None`, `stop_reason=turn_capped`.

Below `max_turns=17` the erroring call never fits before the limit, so the tail is an ordinary tool-call message and both revisions report the plain `Reached max_turns` failure — the divergence is specifically the fallback-on-final-turn case. On the `stop_reason` point from your inline note: I observe `FAILED + turn_capped` — the run did hit the recursion limit, so the turn cap is the additive context while the status carries the failure.

WillemJiang (approved):

PR #4057, opened 2026-07-11
fix(skills): close AST literal-only shell=True bypass in SkillScan

DaoyuanLi2816:
## Summary

SkillScan's Python analyzer only classifies a `subprocess.*(...)` call as the CRITICAL, hard-blocked `python-shell-exec` rule when its `shell=` keyword is the literal AST constant `True`. Any other value that behaves identically at runtime - a variable or a call expression - falls through to the HIGH, non-blocking `python-subprocess` classification instead, which is reviewed only by the probabilistic LLM scanner rather than being hard-blocked by `enforce_static_scan`.

## The bug

`_scan_python()` decides the rule for a `subprocess.*` call like this:

```python
elif call_name in {"os.system", "os.popen"} or (call_name.startswith("subprocess.") and _call_has_shell_true(node)):
    findings.append(_finding_for_node("python-shell-exec", rel_path, node, call_name))
elif call_name.startswith("subprocess."):
    findings.append(_finding_for_node("python-subprocess", rel_path, node, call_name))
```

and the old `_call_has_shell_true()` only matched a literal:

```python
def _call_has_shell_true(node: ast.Call) -> bool:
    return any(keyword.arg == "shell" and isinstance(keyword.value, ast.Constant) and keyword.value.value is True for keyword in node.keywords)
```

`isinstance(keyword.value, ast.Constant)` is `False` for a `Name` node or a `Call` node, so:

```python
shell_flag = True
subprocess.run(cmd, shell=shell_flag)   # -> python-subprocess (HIGH, not blocked)

subprocess.run(cmd, shell=bool(1))      # -> python-subprocess (HIGH, not blocked)
```

both classify as the non-blocking `python-subprocess` HIGH finding instead of `python-shell-exec` CRITICAL, even though both execute through the shell exactly like `shell=True` at runtime. `enforce_static_scan()` only raises `StaticScanBlockedError` on `CRITICAL` findings, so a skill using either form installs and runs without ever hitting the deterministic gate - the whole point of a static, hard-blocking check is defeated by a one-line indirection.

## The fix

`_call_has_shell_true` is renamed to `_call_shell_may_be_true` and inverts the check to fail closed on ambiguity: a `shell=` keyword is only safe when it is the literal, statically-provable constant `False`; any other value under that keyword (`True`, any other literal, or a non-literal expression) is treated the same as `shell=True`. A call with no `shell=` keyword at all is unaffected, since `subprocess.*` already defaults to `shell=False` in that form.

```python
def _call_shell_may_be_true(node: ast.Call) -> bool:
    for keyword in node.keywords:
        if keyword.arg == "shell":
            return not (isinstance(keyword.value, ast.Constant) and keyword.value.value is False)
    return False
```

The lookup stays scoped to the `shell=` keyword of the specific `Call` node already gated by `call_name.startswith("subprocess.")`, so it cannot pick up unrelated `shell=` usage elsewhere (dict literals, kwargs on non-subprocess calls, etc.).

## Tests

Adds to `backend/tests/test_skillscan_native.py`:
- `test_python_subprocess_shell_via_variable_blocks` - `shell=shell_flag` (`shell_flag = True` assigned nearby) now classifies as `python-shell-exec` / CRITICAL and `enforce_static_scan` raises.
- `test_python_subprocess_shell_via_expression_blocks` - `shell=bool(1))` likewise.
- `test_python_subprocess_shell_false_literal_warns_not_block` - locks in that literal `shell=False` still classifies as the non-blocking `python-subprocess` HIGH finding, so the fix doesn't over-block legitimate `shell=False` usage.

Verified fail-before / pass-after: with the fix reverted, the two bypass tests fail (`python-shell-exec` missing from findings, `enforce_static_scan` does not raise); with the fix applied, both pass. The existing literal `shell=True` coverage in `test_enforce_static_scan_blocks_only_critical_findings` continues to pass unchanged, as does `test_python_subprocess_without_shell_warns` (no `shell=` keyword at all still warns, not blocks) and `test_bundled_public_skills_have_no_critical_findings` (no false positives on the real bundled skills). Full file: `24 passed`.

```
cd backend && PYTHONPATH=. pytest tests/test_skillscan_native.py
```

`ruff check --line-length 240` and `ruff format --check --line-length 240` are clean on both changed files.

### DIFF
--- a/backend/packages/harness/deerflow/skills/skillscan/orchestrator.py
+++ b/backend/packages/harness/deerflow/skills/skillscan/orchestrator.py
@@ -369,7 +369,7 @@ def _scan_python(rel_path: str, text: str) -> list[SecurityFinding]:
         call_name = _python_call_name(node, aliases)
         if call_name in {"eval", "exec"} or (call_name == "compile" and _compile_mode_is_exec(node)):
             findings.append(_finding_for_node("python-dynamic-exec", rel_path, node, call_name))
-        elif call_name in {"os.system", "os.popen"} or (call_name.startswith("subprocess.") and _call_has_shell_true(node)):
+        elif call_name in {"os.system", "os.popen"} or (call_name.startswith("subprocess.") and _call_shell_may_be_true(node)):
             findings.append(_finding_for_node("python-shell-exec", rel_path, node, call_name))
         elif call_name.startswith("subprocess."):
             findings.append(_finding_for_node("python-subprocess", rel_path, node, call_name))
@@ -649,8 +649,18 @@ def _compile_mode_is_exec(node: ast.Call) -> bool:
     return any(keyword.arg == "mode" and isinstance(keyword.value, ast.Constant) and keyword.value.value == "exec" for keyword in node.keywords)
 
 
-def _call_has_shell_true(node: ast.Call) -> bool:
-    return any(keyword.arg == "shell" and isinstance(keyword.value, ast.Constant) and keyword.value.value is True for keyword in node.keywords)
+def _call_shell_may_be_true(node: ast.Call) -> bool:
+    # Fail closed on ambiguity: a ``shell=`` keyword is only safe when it is a
+    # literal, statically-provable ``False``. Any other value under that keyword,
+    # the literal ``True``, any other literal, or a non-literal expression such as
+    # a variable or a function call (``shell=shell_flag``, ``shell=bool(1)``),
+    # cannot be proven safe by static analysis, so it is treated the same as an
+    # explicit ``shell=True`` rather than silently falling through to the
+    # non-blocking ``python-subprocess`` classification.
+    for keyword in node.keywords:
+        if keyword.arg == "shell":
+            return not (isinstance(keyword.value, ast.Constant) and keyword.value.value is False)
+    return False
 
 
 def _call_is_network_sink(call_name: str) -> bool:
--- a/backend/packages/harness/deerflow/skills/skillscan/orchestrator.py
+++ b/backend/packages/harness/deerflow/skills/skillscan/orchestrator.py
@@ -658,6 +658,15 @@ def _call_shell_may_be_true(node: ast.Call) -> bool:
     # explicit ``shell=True`` rather than silently falling through to the
     # non-blocking ``python-subprocess`` classification.
     for keyword in node.keywords:
+        if keyword.arg is None:
+            # ``**mapping`` / ``**kwargs`` unpacking: represented in the AST as a
+            # keyword with ``arg is None``. The mapping's contents (and whether it
+            # even carries a ``shell`` key) are not knowable by static analysis, so
+            # this fails closed the same as a variable/expression shell= value.
+            # Deliberate, documented over-block: a harmless kwargs-unpack with no
+            # ``shell`` key also blocks, since the alternative (inspecting the
+            # unpacked mapping's contents) is not generally possible statically.
+            return True
         if keyword.arg == "shell":
             return not (isinstance(keyword.value, ast.Constant) and keyword.value.value is False)
     return False
--- a/backend/tests/test_skillscan_native.py
+++ b/backend/tests/test_skillscan_native.py
@@ -189,6 +189,55 @@ def test_python_subprocess_shell_via_expression_blocks(tmp_path: Path) -> None:
     assert _finding_by_rule(excinfo.value.findings, "python-shell-exec")["severity"] == "CRITICAL"
 
 
+def test_python_subprocess_shell_via_kwargs_unpacking_blocks(tmp_path: Path) -> None:
+    # A ``**``-unpacked mapping can carry a ``shell=True`` key that is invisible
+    # to a plain ``keyword.arg == "shell"`` scan: in the AST, a ``**mapping``
+    # argument is represented as a keyword with ``arg is None``. Its effect is
+    # statically indistinguishable from a literal ``shell=True``, so it must be
+    # classified and blocked the same way, not silently downgraded to the
+    # non-blocking python-subprocess warning.
+    skill_dir = tmp_path / "demo-skill"
+    _write_skill(skill_dir)
+    scripts_dir = skill_dir / "scripts"
+    scripts_dir.mkdir()
+    (scripts_dir / "run.py").write_text(
+        "import subprocess\nopts = {'shell': True}\nsubprocess.run(['whoami'], **opts)\n",
+        encoding="utf-8",
+    )
+
+    findings = scan_skill_dir(skill_dir)["findings"]
+
+    assert _finding_by_rule(findings, "python-shell-exec")["severity"] == "CRITICAL"
+    assert not [item for item in findings if item["rule_id"] == "python-subprocess"]
+
+    with pytest.raises(StaticScanBlockedError) as excinfo:
+        enforce_static_scan(skill_dir, skill_name="demo-skill")
+    assert _finding_by_rule(excinfo.value.findings, "python-shell-exec")["severity"] == "CRITICAL"
+
+
+def test_python_subprocess_kwargs_unpacking_without_shell_key_still_blocks(tmp_path: Path) -> None:
+    # Known, deliberate over-block, documented here rather than in a code
+    # comment alone: a ``**``-unpacked mapping that provably carries no
+    # ``shell`` key (only ``check`` below) is still treated as shell-ambiguous
+    # and blocked, because what a ``**``-unpacked mapping contains is not
+    # knowable by static analysis in general. Failing closed on every
+    # ``**``-unpack is the conservative, defensible choice over trying to
+    # inspect the unpacked mapping's contents.
+    skill_dir = tmp_path / "demo-skill"
+    _write_skill(skill_dir)
+    scripts_dir = skill_dir / "scripts"
+    scripts_dir.mkdir()
+    (scripts_dir / "run.py").write_text(
+        "import subprocess\nopts = {'check': True}\nsubprocess.run(['whoami'], **opts)\n",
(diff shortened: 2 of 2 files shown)


willem-bd (commented):

The core fix is sound — renaming to `_call_shell_may_be_true` and failing closed on any non-literal-`False` `shell=` value correctly closes the `shell=variable` / `shell=bool(1)` downgrade, and the new tests lock that in well. But the renamed helper's "fail closed on ambiguity" posture is applied inconsistently: a sibling form of the same ambiguity still passes silently, which leaves a same-class gap in a PR whose explicit purpose is closing shell-exec evasion in the deterministic blocker.

[P1] `**`-keyword-unpacking still downgrades shell execution to non-blocking HIGH
- Location: `backend/packages/harness/deerflow/skills/skillscan/orchestrator.py:652` (`_call_shell_may_be_true`) and the classification at `orchestrator.py:372-375`.
- Problem: A `subprocess.*` call that supplies `shell` via `**`-unpacking (any `**mapping` / `**kwargs` argument) is invisible to the new check. In the AST, a `**`-unpacked argument is represented as a `keyword` with `arg is None`, so the helper's `if keyword.arg == "shell"` branch never matches it, the loop completes, and the function returns `False`. The call then falls through to the `elif call_name.startswith("subprocess.")` branch and is classified as `python-subprocess` (HIGH) — exactly the silent, non-blocking downgrade this PR is trying to eliminate. Since `enforce_static_scan` only hard-blocks on `CRITICAL`, such a skill installs and runs and only hits the probabilistic LLM scan, defeating the deterministic pre-gate that gates `.skill` archive install and agent-managed skill writes. This is statically indistinguishable-in-effect from the `shell=variable` case the diff correctly closes — the shell flag is supplied indirectly and its value is unknowable to static analysis — so it should fail closed for the same reason.
- Evidence: Verified against the head AST: for any subprocess call carrying a `**`-unpacked argument, `keyword.arg` is `None` for the unpack entry, `_call_shell_may_be_true` returns `False`, and the call lands in the `python-subprocess` HIGH branch rather than `python-shell-exec` CRITICAL. This is the same detection surface the diff touches; the diff establishes a fail-closed posture for the keyword *value* but not for keyword *delivery via unpacking*.
- Suggested fix: Extend the fail-closed rule to a `**`-unpacked argument as well — if the call has any `keyword.arg is None` entry, the shell flag may be arriving through it and cannot be proven safe, so treat it the same as a non-literal `shell=` value:
  ```python
  def _call_shell_may_be_true(node: ast.Call) -> bool:
      for keyword in node.keywords:
          if keyword.arg is None:  # **mapping / **kwargs unpacking
              return True
          if keyword.arg == "shell":
              return not (isinstance(keyword.value, ast.Constant) and keyword.value.value is False)
      return False
  ```
- Test: Add a case to `tests/test_skillscan_native.py` mirroring `test_python_subprocess_shell_via_variable_blocks`, where the shell flag is delivered via a `**`-unpacked mapping on a `subprocess.*` call; assert the finding is `python-shell-exec` / CRITICAL, that no `python-subprocess` finding is emitted, and that `enforce_static_scan` raises `StaticScanBlockedError`. A companion negative case (a `**`-unpack that provably carries no `shell`, if you want to keep the over-block intentional) should be documented as a known false positive, or the unpack rule scoped accordingly.

DaoyuanLi2816:
@willem-bd Good catch, and agreed on the reasoning — a shell flag delivered through `**`-unpacking is statically just as unprovable as `shell=variable`, so it belongs on the same fail-closed path.

Pushed 07640f8c:

- Extended `_call_shell_may_be_true` to treat any `**`-unpacked keyword (`keyword.arg is None`) as shell-ambiguous and return `True`, matching your suggested fix exactly.
- Added `test_python_subprocess_shell_via_kwargs_unpacking_blocks`: `subprocess.run(['whoami'], **opts)` with `opts = {'shell': True}` now classifies as `python-shell-exec` / CRITICAL (no `python-subprocess` finding emitted), and `enforce_static_scan` raises `StaticScanBlockedError`.
- Added `test_python_subprocess_kwargs_unpacking_without_shell_key_still_blocks` as the companion negative case you asked for: a `**opts` unpack that provably carries no `shell` key (only `check`) still blocks as CRITICAL. Documented inline as a deliberate over-block — the unpacked mapping's contents aren't knowable statically, so scoping the rule to "no shell key present" isn't something we can prove rather than guess, and the unconditional `arg is None -> True` is the more defensible choice.

Full `tests/test_skillscan_native.py` passes, 26/26, including the existing `shell=True` / `shell=shell_flag` / `shell=bool(1)` / literal `shell=False` cases from the original diff — no regressions.

backend/packages/harness/deerflow/skills/skillscan/orchestrator.py
```diff
@@ -649,8 +649,27 @@ def _compile_mode_is_exec(node: ast.Call) -> bool:
...
+def _call_shell_may_be_true(node: ast.Call) -> bool:
+    # Fail closed on ambiguity: a ``shell=`` keyword is only safe when it is a
+    # literal, statically-provable ``False``. Any other value under that keyword,
+    # the literal ``True``, any other literal, or a non-literal expression such as
+    # a variable or a function call (``shell=shell_flag``, ``shell=bool(1)``),
+    # cannot be proven safe by static analysis, so it is treated the same as an
+    # explicit ``shell=True`` rather than silently falling through to the
+    # non-blocking ``python-subprocess`` classification.
+    for keyword in node.keywords:
+        if keyword.arg is None:
```
willem-bd, line 661: **Minor: first-match short-circuit makes `shell=` vs `**kwargs` order-dependent.**

The loop returns on the *first* `shell=` keyword *or* the first `**`-unpack, whichever comes first. So `f(shell=False, **opts)` returns `False` (never inspects `**opts`), while `f(**opts, shell=False)` returns `True` (blocks). The asymmetry only matters when `opts` actually carries a `shell` key - and that's a runtime `TypeError` (duplicate keyword) - so it's practically harmless. Flagging only because the order-sensitivity could surprise a future reader; a one-line comment on the first-match semantics (or scanning all keywords) would make the intent explicit. Not a defect.

backend/packages/harness/deerflow/skills/skillscan/orchestrator.py
```diff
@@ -649,8 +649,27 @@ def _compile_mode_is_exec(node: ast.Call) -> bool:
...
+            # ``**mapping`` / ``**kwargs`` unpacking: represented in the AST as a
+            # keyword with ``arg is None``. The mapping's contents (and whether it
+            # even carries a ``shell`` key) are not knowable by static analysis, so
+            # this fails closed the same as a variable/expression shell= value.
+            # Deliberate, documented over-block: a harmless kwargs-unpack with no
+            # ``shell`` key also blocks, since the alternative (inspecting the
+            # unpacked mapping's contents) is not generally possible statically.
+            return True
+        if keyword.arg == "shell":
+            return not (isinstance(keyword.value, ast.Constant) and keyword.value.value is False)
```
willem-bd, line 671: **Minor: falsy-but-not-`False` literals are over-blocked.**

`value is False` is an identity check, not a truthiness check, so `shell=0`, `shell=None`, `shell=""`, `shell=0.0` - all falsy at runtime, i.e. no shell execution - are classified CRITICAL here. This is consistent with the PR's stated "any other literal blocks" fail-closed philosophy and is the conservative choice for a security scanner; just flagging it as a false-positive surface for third-party skills that might write `shell=0`. `test_bundled_public_skills_have_no_critical_findings` confirms no bundled skill trips it. Defensible as-is.

willem-bd (commented):
Code review - **approve**. Solid, well-scoped security fix: closes a one-line-indirection bypass in the deterministic `shell=True` gate and extends fail-closed coverage to `**kwargs` unpacking. Verified statically against the head commit - the rename is complete (one caller, no stale refs), rule severities match (`python-shell-exec`=CRITICAL / `python-subprocess`=HIGH), and alias resolution (`import ... as` + `from ... import`) means the fix applies through import aliases.

Couldn't re-run the test suite locally (transient sandbox-classifier outage), so the "24 passed" claim is taken as author-verified, not independently confirmed.

Two minor inline notes below (both defensible as-is; one is the fail-closed over-block, the other an order-sensitivity wart). Two further observations belong here in the summary rather than inline:

- **Test gap (optional):** no test pins down the `shell=False, **opts` mixed-ordering behavior or the falsy-literal (`shell=0` / `None`) cases. Given the deliberate over-block philosophy these aren't required, but a test locking in the ordering behavior would prevent future confusion.
- **Pre-existing, out of scope:** dynamically-dispatched calls like `getattr(subprocess, "run")(cmd, shell=True)` still bypass entirely (`call_name` resolves to `""`). Not introduced here; flagging only because the PR is about closing bypasses. Appropriately left for a separate effort.

WillemJiang:
@DaoyuanLi2816 Please check out the latest review comments.

DaoyuanLi2816:
@WillemJiang thanks for the nudge -- following up on willem-bd's two inline notes:

- **First-match short-circuit (`shell=` vs `**kwargs` order):** agreed the asymmetry could surprise a future reader, but as you noted a `**opts` carrying a `shell` key is already a runtime `TypeError` (duplicate keyword) before this code path matters, so there's no real-world input that hits it. Leaving as-is per your "not a defect" read.
- **`value is False` over-blocking falsy-but-not-`False` literals:** agreed this is the conservative, fail-closed choice for a security scanner, and `test_bundled_public_skills_have_no_critical_findings` confirms no bundled skill trips it today. Leaving as-is per your "defensible as-is" read.

No code changes on either -- both were already flagged as non-defects in the original review, just making sure they're acknowledged rather than silently sitting.

WillemJiang (approved):

PR #3869, opened 2026-06-29
fix(sandbox): normalize Windows backslash paths to forward slashes in bash commands

bjtolo:
## Summary

- Normalize all virtual path resolutions in bash commands to forward slashes on Windows
- Fixes `replace_virtual_paths_in_command()` in `tools.py` (3 callbacks: skills, ACP, user-data)
- Fixes `LocalSandbox._resolve_paths_in_command()` in `local_sandbox.py` (custom mount paths)

## Problem

On Windows, virtual paths like `/mnt/skills/custom/...` are resolved to host paths with backslashes: `C:\Users\admin\deer-flow\skills\custom\...`. Bash interprets `\U`, `\a`, `\d`, `\s`, `\n`, `\t` as escape sequences, mangling the path to `C:Usersadmindeer-flowskillscustom...`. The agent wastes 13+ commands trying different paths before giving up.

## Fix

Add `.replace("\\", "/")` to all path resolution callbacks. Forward slashes work universally in bash on all platforms, and Windows APIs / Python's `open()` accept them natively. The sibling method `_resolve_paths_in_content()` already does this — `_resolve_paths_in_command()` was just inconsistent.

## Test plan

- [x] All 118 existing tests in `test_sandbox_tools_security.py` pass
- [ ] Manual test on Windows: agent runs `python /mnt/skills/custom/.../script.py` successfully

Closes #3865

PR #4058, opened 2026-07-11
fix(sandbox): use os.sep in reverse-resolve containment check on Windows

DaoyuanLi2816:
## Summary

`LocalSandbox._reverse_resolve_path` leaks the raw host path (real username, full directory tree) instead of the virtual `/mnt/user-data/...` path on Windows, for any nested path. This function backs `list_dir`, `glob`, `grep`, and bash-output masking.

## The bug

```python
normalized_path = path.replace("\\", "/")
path_str = str(Path(normalized_path).resolve())

for mapping in self._mappings_by_local_specificity:
    local_path_resolved = self._resolved_local_paths[mapping]
    if path_str == local_path_resolved or path_str.startswith(local_path_resolved + "/"):
        ...
```

`Path.resolve()` always renders with the native separator regardless of the forward-slash normalization a few lines above -- on Windows that's a backslash. `local_path_resolved` (`_resolved_local_paths`, also built via `Path(...).resolve()`) is backslash-separated too. So the containment check's hardcoded `+ "/"` can never match: `path_str.startswith(local_path_resolved + "/")` compares a backslash-joined string against a forward-slash-suffixed prefix. Only the exact-root case (`path_str == local_path_resolved`, no separator involved) ever matches. Every nested path falls through to `return path_str` at the bottom, returning the raw host path.

`_is_read_only_path` does the equivalent containment check correctly, via `os.sep`:

```python
if resolved == local_resolved or resolved.startswith(local_resolved + os.sep):
```

## The fix

Align `_reverse_resolve_path` with `_is_read_only_path`'s pattern: compare with `os.sep` instead of a hardcoded `"/"`. Since the extracted relative portion is native-separated on Windows but container paths are always POSIX-style, it's additionally normalized to forward slashes before being spliced into the result:

```python
if path_str == local_path_resolved or path_str.startswith(local_path_resolved + os.sep):
    relative = path_str[len(local_path_resolved) :].lstrip(os.sep).replace(os.sep, "/")
    resolved = f"{mapping.container_path}/{relative}" if relative else mapping.container_path
    return resolved
```

On POSIX `os.sep == "/"`, so this is behavior-identical there.

## Same-file cosmetic bug (`list_dir` overlay)

While auditing this function I found `list_dir`'s "virtual sub-directory overlay" (added to backfill mount children that the underlying scan skips) has a related but separate bug:

```python
existing_dirs = {e.rstrip("/") for e in result if e.endswith("/")}
for mapping in self.path_mappings:
    if mapping.container_path.startswith(container_path + "/"):
        child_rel = mapping.container_path[len(container_path) + 1 :]
        if "/" not in child_rel and child_rel not in existing_dirs:
            ...
            result.append(f"{mapping.container_path}/")
```

`existing_dirs` holds full container paths (e.g. `/mnt/user-data/workspace`), but `child_rel` is the bare child name (e.g. `workspace`). The comparison never matches, so the "already listed" guard is a no-op. For the standard `/mnt/user-data` mount (aggregate parent + workspace/uploads/outputs as real nested subdirectories, all found by the plain scan), each of the three gets appended a second time by the overlay. Fixed by comparing the mapping's full container path instead of the bare name.

## Context

Same separator-bug class already fixed in this file by #3869 (forward-direction command-string resolution) and #4035 (reverse-direction regex segment boundary). Neither touches this containment check -- #3869 fixes `_resolve_paths_in_command`, and #4035 fixes which substring `_reverse_output_patterns` extracts from output, not what `_reverse_resolve_path` does with it once extracted.

## Tests

Fail-before / pass-after, on the pre-existing suite (no mocking -- real per-thread `LocalSandboxProvider` mappings and real filesystem paths):

- `tests/test_local_sandbox_virtual_path_contract.py::test_list_dir_via_public_api_with_virtual_path`, `test_glob_with_virtual_path`, `test_grep_with_virtual_path` -- all failed before (host path leaked, e.g. `matches[0].path` was `C:\Users\...\workspace\findme.txt` instead of ending in `/mnt/user-data/workspace/findme.txt`), pass after.
- `tests/test_local_sandbox_path_regex_cache.py::test_reverse_resolve_output_maps_local_back_to_container` and the `\win\p` case of `test_reverse_resolve_still_matches_root_before_non_slash_boundaries` -- same failure shape, pass after.

Added `test_list_dir_on_user_data_root_does_not_duplicate_subdir_mounts` for the overlay bug (no existing test called `list_dir` on the aggregate `/mnt/user-data` parent directly); confirmed it fails with `workspace listed 2 time(s)` etc. against the old comparison and passes with the fix.

Full `test_local_sandbox_virtual_path_contract.py` + `test_local_sandbox_path_regex_cache.py`: 45 tests, 43 passed. The remaining 2 (`test_command_paths_resolved_to_local`, `test_forward_resolution_behavior_unchanged`) fail identically with or without this change -- they assert forward-direction (`_resolve_path`/`_resolve_paths_in_command`) output against a hardcoded expected string built from a native-separator `tmp_path`, unrelated to the reverse-direction code this PR touches.

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
+++ b/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
@@ -344,9 +344,20 @@ class LocalSandbox(Sandbox):
         # Try each mapping (longest local path first for more specific matches)
         for mapping in self._mappings_by_local_specificity:
             local_path_resolved = self._resolved_local_paths[mapping]
-            if path_str == local_path_resolved or path_str.startswith(local_path_resolved + "/"):
-                # Replace the local path prefix with container path
-                relative = path_str[len(local_path_resolved) :].lstrip("/")
+            # ``Path.resolve()`` always renders with the native separator
+            # (backslash on Windows), regardless of the forward-slash
+            # normalization above, so the containment check must compare with
+            # ``os.sep`` here too -- mirroring ``_is_read_only_path`` -- instead
+            # of a hardcoded "/". A hardcoded "/" can never match a
+            # backslash-joined nested path on Windows, so every nested path
+            # silently fell through to the "no mapping found" branch below and
+            # leaked the raw host path (real username, full directory tree).
+            if path_str == local_path_resolved or path_str.startswith(local_path_resolved + os.sep):
+                # Replace the local path prefix with container path. Container
+                # paths are always POSIX-style, so the extracted relative
+                # portion (native-separated on Windows) is normalized to
+                # forward slashes before being spliced in.
+                relative = path_str[len(local_path_resolved) :].lstrip(os.sep).replace(os.sep, "/")
                 resolved = f"{mapping.container_path}/{relative}" if relative else mapping.container_path
                 return resolved
 
@@ -650,8 +661,14 @@ class LocalSandbox(Sandbox):
             # 2. It is NOT already present in the result (was skipped by list_dir)
             if mapping.container_path.startswith(container_path + "/"):
                 child_rel = mapping.container_path[len(container_path) + 1 :]
-                # Only direct children (no further slashes), e.g. "public", "custom"
-                if "/" not in child_rel and child_rel not in existing_dirs:
+                # Only direct children (no further slashes), e.g. "public", "custom".
+                # Compare the mapping's full container path -- not the bare child
+                # name -- against existing_dirs, which holds full paths (e.g.
+                # "/mnt/user-data/workspace"). Comparing the bare name here would
+                # never match, so an already-listed mount (the common case: real
+                # nested workspace/uploads/outputs subdirectories under
+                # /mnt/user-data) would be appended a second time.
+                if "/" not in child_rel and mapping.container_path.rstrip("/") not in existing_dirs:
                     # Verify the host path exists so we don't add phantom entries
                     try:
                         if Path(mapping.local_path).resolve().is_dir():
--- a/backend/tests/test_local_sandbox_path_regex_cache.py
+++ b/backend/tests/test_local_sandbox_path_regex_cache.py
@@ -9,6 +9,7 @@ from pathlib import Path
 
 import pytest
 
+from deerflow.sandbox.local import local_sandbox as local_sandbox_module
 from deerflow.sandbox.local.local_sandbox import LocalSandbox, PathMapping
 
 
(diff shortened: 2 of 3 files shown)


willem-bd (commented):

The core `_reverse_resolve_path` fix (os.sep in the containment check + `lstrip(os.sep).replace(os.sep, "/")`) is correct: on POSIX `os.sep == "/"` so it is behavior-identical to the old code, and on Windows it now correctly matches a backslash-joined `path_str` against a backslash-joined `local_path_resolved`. I also confirmed it does not overlap with the already-merged #4035 — that PR fixed `_reverse_output_patterns` (regex extraction), while this PR fixes what `_reverse_resolve_path` does with the extracted substring, so they are complementary. The `list_dir` overlay fix and its new test are also correct.

One test gap below.

[P2] No CI-verifiable regression test for the os.sep containment fix
- Location: `backend/tests/test_local_sandbox_virtual_path_contract.py` (new test) and `backend/tests/test_local_sandbox_path_regex_cache.py`
- Problem: The primary Windows fix has no test that would fail if it were reverted. CI runs only on `ubuntu-latest` (`.github/workflows/backend-unit-tests.yml`), where `os.sep == "/"`. On POSIX, `local_path_resolved + os.sep` is identical to the old `local_path_resolved + "/"`, `lstrip(os.sep)` is identical to `lstrip("/")`, and `replace(os.sep, "/")` is a no-op — so the pre-fix and post-fix code are observationally identical on the CI platform. The "fail-before / pass-after" cases the PR body cites (`test_list_dir_via_public_api_with_virtual_path`, `test_glob_with_virtual_path`, `test_grep_with_virtual_path`, the regex-cache cases) only fail on Windows, which CI never runs; they pass on Linux both before and after this change. The one test this PR actually adds — `test_list_dir_on_user_data_root_does_not_duplicate_subdir_mounts` — covers the second (list_dir overlay) bug, not the os.sep change. Neither changed test file references `os.sep`, `sys.platform`, `monkeypatch`, or `mock`.
- Evidence: `grep -rn 'os.sep|monkeypatch|sys.platform|mock'` over both changed test files returns nothing; backend CI is `runs-on: ubuntu-latest` only. So a future refactor that reverts the `os.sep` back to a hardcoded `"/"` (the exact regression this PR exists to fix) would keep CI fully green.
- Suggested fix: Add a portable regression test that forces the Windows separator path independent of host OS. Options: (a) monkeypatch `deerflow.sandbox.local.local_sandbox.os.sep` to `"\\"` and stub `Path.resolve`/`_resolved_local_paths` to return backslash-joined strings, then assert `_reverse_resolve_path("C:\\host\\workspace\\sub\\f.txt")` returns `/mnt/user-data/workspace/sub/f.txt` rather than the raw host path; or (b) extract the containment + relative-extraction into a pure helper taking an explicit `sep` argument and parametrize the test over `("/", "\\")`. Either makes the Linux run genuinely prove the fix.
- Test: the new test must fail when `os.sep` in the containment check (line 355) and the `lstrip(os.sep).replace(os.sep, "/")` (line 360) are reverted to the hardcoded `"/"` form, when run on Linux.

DaoyuanLi2816:
@willem-bd Thanks for the detailed review — you're right that neither changed test file discriminated the `os.sep` fix on Linux CI. Added `test_reverse_resolve_path_matches_windows_backslash_containment` to `test_local_sandbox_path_regex_cache.py` (be3b2a47), using your option (a): monkeypatch `os.sep` to `"\\"` plus stub both the module's `Path` name and the sandbox's cached `_resolved_local_paths` to return backslash-joined strings, mirroring what real `WindowsPath.resolve()` produces. No filesystem or actual Windows host involved, so it runs identically on `ubuntu-latest`.

Verified the discrimination directly: with the `os.sep` fix reverted to the hardcoded `"/"` form (both the containment check and the `lstrip/replace` line), the new test fails with the raw host path leaking through:

```
AssertionError: assert 'C:\Users\test\workspace\sub\f.txt' == '/mnt/user-data/workspace/sub/f.txt'
```

With the fix restored, it passes.

Full `test_local_sandbox_virtual_path_contract.py` + `test_local_sandbox_path_regex_cache.py` also pass with zero regressions (46 tests total, including the new one).

Also ran `ruff check --line-length 240` and `ruff format --check --line-length 240` on the changed file — clean.

WillemJiang (approved):

WillemJiang (approved):

PR #4059, opened 2026-07-11
fix(agent): fix lost loop_capped stop reason when a subagent's run_id is None

DaoyuanLi2816:
## Summary

A subagent's `loop_capped` stop reason can be silently lost when the subagent's `run_id` is `None` (e.g. an embedded/TUI-client-dispatched subagent, where `run_id` is never assigned). The lead just sees a plain `completed` status with no indication the subagent was actually hard-stopped for looping.

## The bug

`LoopDetectionMiddleware._get_run_id` used a truthiness check that treats a present-but-`None` `run_id` the same as a totally absent one:

```python
def _get_run_id(self, runtime: Runtime) -> str:
    """Extract run_id from runtime context for per-run warning scoping."""
    run_id = runtime.context.get("run_id") if runtime.context else None
    if run_id:
        return str(run_id)
    return "default"
```

Meanwhile `SubagentExecutor` sets `context["run_id"] = self.run_id` unconditionally (no truthiness guard), so `run_id` can genuinely be `None` in the subagent's own runtime context:

```python
context["run_id"] = self.run_id
```

and later reads the stop reason back with that same raw, un-normalized attribute:

```python
reason = mw.consume_stop_reason(self.run_id)
```

So when a loop-detection hard-stop fires for a `run_id=None` subagent:
- the middleware's write side collapses `run_id=None` to the literal string `"default"` and records `self._stop_reason["default"] = "loop_capped"`
- the executor's read side calls `consume_stop_reason(None)`, which looks up `self._stop_reason[None]` — a different key

The `loop_capped` signal is silently dropped. `TokenBudgetMiddleware._get_run_id` (the sibling guard added for the same #3875 Phase 2 stop-reason mechanism) already avoids this class of bug by keying on presence instead of truthiness:

```python
@staticmethod
def _get_run_id(runtime: Runtime) -> str:
    ctx = getattr(runtime, "context", None)
    if isinstance(ctx, dict) and "run_id" in ctx:
        return ctx["run_id"]
    # Fallback to runtime object ID to prevent collisions across embedded client runs
    return str(id(runtime))
```

## The fix

Align `LoopDetectionMiddleware._get_run_id` with the same key-presence-based pattern, so a present `run_id` (including `None`) round-trips exactly, and an absent `run_id` falls back to a per-runtime-unique key (instead of a literal shared with the present-but-`None` case, which was the source of the collision):

```python
def _get_run_id(self, runtime: Runtime) -> str:
    ctx = getattr(runtime, "context", None)
    if isinstance(ctx, dict) and "run_id" in ctx:
        return ctx["run_id"]
    # Fallback to runtime object ID to prevent collisions across embedded client runs
    return str(id(runtime))
```

## Tests

Adds `test_hard_stop_stamps_loop_capped_with_explicit_none_run_id`: drives a real `LoopDetectionMiddleware` with `runtime.context={"run_id": None, ...}` through a real hard-stop, then calls `consume_stop_reason(None)` exactly as the executor does, and asserts it returns `"loop_capped"`.

Verified fail-before / pass-after: before the fix this test fails with `assert None == 'loop_capped'`; the fix makes it pass. `test_missing_run_id_uses_default_pending_scope` (covering the separate, already-correct "no `run_id` key at all" fallback) is updated to `test_missing_run_id_uses_per_runtime_pending_scope` since that fallback now also matches `TokenBudgetMiddleware`'s per-runtime key instead of a shared `"default"` literal. Full file: `69 passed`.

```
cd backend && PYTHONPATH=. pytest tests/test_loop_detection_middleware.py
```

Also re-ran `tests/test_token_budget_middleware.py`, `tests/test_loop_detection_config.py`, `tests/test_subagent_executor.py`, and `tests/test_task_tool_core_logic.py` for regressions: `194 passed` combined.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
@@ -274,11 +274,30 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
         return "default"
 
     def _get_run_id(self, runtime: Runtime) -> str:
-        """Extract run_id from runtime context for per-run warning scoping."""
-        run_id = runtime.context.get("run_id") if runtime.context else None
-        if run_id:
-            return str(run_id)
-        return "default"
+        """Extract run_id from runtime context for per-run warning scoping.
+
+        Keyed by presence, not truthiness: ``SubagentExecutor`` sets
+        ``context["run_id"] = self.run_id`` unconditionally (no truthiness
+        guard), so an embedded/TUI-dispatched subagent — whose ``run_id`` is
+        never assigned per ``AGENTS.md``'s description of the embedded
+        ``DeerFlowClient`` — runs with a context that legitimately carries
+        ``run_id=None`` (the key is *present*, not absent). The executor
+        later reads the stop reason back with the raw attribute,
+        ``consume_stop_reason(self.run_id)``, so this must return exactly
+        that value (``None`` included) when the key is present, rather than
+        collapsing it to a shared fallback indistinguishable from an absent
+        key. A truthiness check (``if run_id:``) previously conflated
+        "present but None/falsy" with "absent", both mapping to the same
+        literal ``"default"`` — so a genuine ``run_id=None`` hard-stop was
+        recorded under ``"default"`` here but looked up under ``None`` by
+        the executor, silently losing the ``loop_capped`` stop reason.
+        Mirrors ``TokenBudgetMiddleware._get_run_id``.
+        """
+        ctx = getattr(runtime, "context", None)
+        if isinstance(ctx, dict) and "run_id" in ctx:
+            return ctx["run_id"]
+        # Fallback to runtime object ID to prevent collisions across embedded client runs
+        return str(id(runtime))
 
     def consume_stop_reason(self, run_id: str | None) -> str | None:
         """Pop and return the stop reason the hard-stop set for this run.
--- a/backend/tests/test_loop_detection_middleware.py
+++ b/backend/tests/test_loop_detection_middleware.py
@@ -2,6 +2,7 @@
 
 import copy
 from collections import OrderedDict
+from types import SimpleNamespace
 from typing import Any
 from unittest.mock import MagicMock
 
@@ -304,8 +305,14 @@ class TestLoopDetection:
         mw.wrap_model_call(request_a, handler)
         assert any(isinstance(message, HumanMessage) and message.name == "loop_warning" for message in captured[1].messages)
 
-    def test_missing_run_id_uses_default_pending_scope(self):
-        """When runtime has no run_id, warning handling falls back to the default run scope."""
+    def test_missing_run_id_uses_per_runtime_pending_scope(self):
+        """When runtime.context has no ``run_id`` key at all, warning handling
+        falls back to a key scoped to the runtime object's identity —
+        mirroring ``TokenBudgetMiddleware._get_run_id``'s fallback — instead
+        of a shared literal like the old ``"default"``, which would collide
+        across concurrent runs that both lack a run_id (the ``_stop_reason``
+        dict this same key derivation feeds is keyed by run_id alone, with
+        no thread scoping)."""
         mw = LoopDetectionMiddleware(warn_threshold=3, hard_limit=10)
         runtime = MagicMock()
         runtime.context = {"thread_id": "test-thread"}
@@ -314,7 +321,8 @@ class TestLoopDetection:
         for _ in range(3):
             mw._apply(_make_state(tool_calls=call), runtime)
 
-        assert mw._pending_warnings.get(_pending_key(run_id="default"))
+        fallback_run_id = str(id(runtime))
+        assert mw._pending_warnings.get(_pending_key(run_id=fallback_run_id))
 
         request = _make_request([AIMessage(content="hi")], runtime)
         captured, handler = _capture_handler()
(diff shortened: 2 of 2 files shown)


backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
```diff
@@ -274,11 +274,30 @@ def _get_thread_id(self, runtime: Runtime) -> str:
...
+        literal ``"default"`` — so a genuine ``run_id=None`` hard-stop was
+        recorded under ``"default"`` here but looked up under ``None`` by
+        the executor, silently losing the ``loop_capped`` stop reason.
+        Mirrors ``TokenBudgetMiddleware._get_run_id``.
+        """
+        ctx = getattr(runtime, "context", None)
+        if isinstance(ctx, dict) and "run_id" in ctx:
+            return ctx["run_id"]
+        # Fallback to runtime object ID to prevent collisions across embedded client runs
+        return str(id(runtime))
```
willem-bd, line 300: **Heads-up: this fallback change is somewhat orthogonal to the stated bug.** The actual fix only needed the presence-based check above — for subagents the `run_id` key is always present, so this `str(id(runtime))` branch isn't even reached on the bug path. The change from `"default"` to `str(id(runtime))` affects a *different* path (absent `run_id`, non-executor usage), bundled in as sibling alignment. It's a reasonable improvement (avoids cross-run collisions when `run_id` is absent), and I confirmed nothing reads the literal `"default"` directly — all access flows through `_get_run_id` / `consume_stop_reason`. One theoretical caveat: `id()` can be reused after GC, so a fresh runtime at a recycled address could pick up a stale key — bounded in practice by pop-on-read, `before_agent` clearing `_pending_warnings`, and the `BoundedDict(1000)` cap. Matches `TokenBudgetMiddleware`, so consistent — just flagging the broader-than-the-fix blast radius.

willem-bd (commented):
Reviewed the `loop_capped`-when-`run_id`-is-`None` fix. Correct diagnosis and sound fix — verified the write side (`_get_run_id(runtime)` at line 543) and read side (`consume_stop_reason(self.run_id)` at executor.py:542) now agree on the `None` key, and that `consume_stop_reason` already accepted `str | None` so there's no contract change. Mirrors the already-correct `TokenBudgetMiddleware._get_run_id`, and the regression test reproduces the exact write/read mismatch. Two inline notes: a scope/awareness flag on the `id()`-based fallback (orthogonal to the actual bug), and a request for the real-path validation the `needs-validation` label asks for.

backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
```diff
@@ -274,11 +274,30 @@ def _get_thread_id(self, runtime: Runtime) -> str:
         return "default"
 
     def _get_run_id(self, runtime: Runtime) -> str:
```
willem-bd, line 276: **Worth a real-path validation pass (the `needs-validation` label).** The unit test proves the write/read keys now match for `run_id=None`, but it'd be good to confirm end-to-end in a live embedded/TUI subagent run that (a) `run_id` is genuinely `None` through the whole path, and (b) a loop-detection hard-stop there now surfaces `loop_capped` to the lead instead of a clean `completed`. The test exercises the middleware in isolation; this confirms the executor actually instantiates it and reads the reason in the `None`-run_id embedded path.

DaoyuanLi2816:
@willem-bd Ran the real-path validation.

**(a) `run_id` is genuinely `None` on the embedded/TUI path.** The gateway path always injects a real `run_id` — `runtime/runs/worker.py::_build_runtime_context` builds `{"thread_id", "run_id"}` with `run_id` a required `str`. The embedded `DeerFlowClient.stream` does not: it builds `context = {"thread_id": thread_id}` with no `run_id` (`client.py:762`; the TUI is a shell over this client). A subagent dispatched from there reads `run_id = parent_context.get("run_id")` → `None` (`task_tool.py:340`), constructs `SubagentExecutor(run_id=None)`, and sets `context["run_id"] = self.run_id` → `None` (`executor.py:778`) — key present, value `None`, the exact case the old truthiness check collapsed.

**(b) A loop hard-stop now surfaces `loop_capped` to the lead.** Drove the real `SubagentExecutor._aexecute` end-to-end through the real `create_agent` graph + real subagent middleware chain (real `LoopDetectionMiddleware`), with `run_id=None` and a scripted model that emits an identical `bash` tool call every turn (`hard_limit` lowered to 3). The hard-stop fires, strips `tool_calls`, and the run completes. Watching the executor's own read — `consume_stop_reason(self.run_id)` i.e. `consume_stop_reason(None)`:

- origin/main: the hard-stop is recorded under the literal `"default"` (`_stop_reason == {"default": "loop_capped"}`), the executor reads with key `None` and gets `None` → the lead sees `status=COMPLETED, stop_reason=None` — a clean completed, `loop_capped` lost.
- this branch: recorded under `None` (`_stop_reason == {None: "loop_capped"}`), read with `None` returns `loop_capped` → the lead sees `status=COMPLETED, stop_reason=loop_capped`.

Same result on an isolated `create_agent(model, tools, middleware=[LoopDetectionMiddleware])` graph where the hard-stop routes cleanly to `END` (3 model calls, no recursion): before → `consume_stop_reason(None)` is `None`; after → `loop_capped`.

The lead→`task_tool` dispatch hop rests on the code path cited in (a); everything from the subagent's `run_id=None` context through the middleware write-key and the executor's read was exercised live on a real graph.

WillemJiang (approved):

PR #1403, opened 2026-03-26
feat(gateway): implement LangGraph Platform API in Gateway, replace langgraph-cli

rayhpeng:
## Summary

Implements a Gateway-native LangGraph Platform API Runtime, allowing the
development stack to run without `langgraph-cli`. The Gateway (FastAPI)
now directly serves all thread / runs / assistants endpoints, eliminating
a heavyweight closed-source dependency.

**Process reduction**: 4 services → 3 (no more `langgraph dev` process).  
**Concurrency**: each run executes as an independent `asyncio.Task` instead
of a serial worker queue.

---

### New harness packages (`packages/harness/deerflow/agents/`)

| Package | Role |
|---|---|
| `stream_bridge` | In-memory pub/sub: decouples agent execution (producer) from SSE endpoints (consumer). Each run gets its own `asyncio.Queue`. |
| `runs/manager.py` | In-memory run registry with atomic `create_or_reject` (no TOCTOU race). |
| `runs/worker.py` | Spawns a background `asyncio.Task`, publishes Platform-style SSE events, injects `Runtime(context={"thread_id": ...})` so existing middlewares work correctly. |
| `runs/schemas.py` | `RunStatus`, `DisconnectMode`, `CancelAction` enum definitions. |

### New Gateway endpoints

| Group | Endpoints |
|---|---|
| Threads | `POST /api/threads`, `GET /api/threads/{tid}`, `POST /api/threads/search`, `DELETE /api/threads/{tid}`, `GET /api/threads/{tid}/state`, `POST /api/threads/{tid}/state`, `POST /api/threads/{tid}/history` |
| Runs | `POST /api/threads/{tid}/runs`, `POST /api/threads/{tid}/runs/stream` (SSE), `POST /api/threads/{tid}/runs/wait`, `GET /api/threads/{tid}/runs/{rid}/join`, `POST /api/threads/{tid}/runs/{rid}/cancel` |
| Assistants compat | `POST /api/assistants/search`, `GET /api/assistants/{aid}`, `GET /api/assistants/{aid}/graph`, `GET /api/assistants/{aid}/schemas` |
| Stateless Runs | `POST /api/runs/stream`, `POST /api/runs/wait` (via `routers/runs.py`) |

### Infrastructure

Two fully isolated routing modes are provided based on `SKIP_LANGGRAPH_SERVER`:

**`SKIP_LANGGRAPH_SERVER=1` (default — compact routing)**
- nginx routes `/api/langgraph-compat/*` → Gateway
[...]

PR #4064, opened 2026-07-11
fix(runs): cancel degrades to lease takeover for multi-worker

### LINKED ISSUE #3948: [feat] 多 Worker 数据正确性基线:Postgres 强制 + Run 跨进程原子化

heart-scalpel:
Fixes #3948

## Why

Under `GATEWAY_WORKERS > 1`, `POST /runs/{run_id}/cancel` landing on a non-owning worker returns 409 — the stop button silently fails. `cancel()` requires the current worker to hold the in-memory `task` / `abort_event`; without sticky routing any pod can receive the request and most will miss. This lands work item 4 from the multi-worker epic: cancel degrades to lease takeover when the owner is unreachable, instead of refusing.

## What changed

**Behaviour**

- `RunManager.cancel()` returns a `CancelOutcome` enum instead of `bool`, so the router can map each path to the right HTTP response:
  - `cancelled` — local cancel (existing path).
  - `taken_over` — owner's lease expired past grace; this worker atomically marks the run `error` and returns 202.
  - `lease_valid_elsewhere` — owner is still alive; router returns 409 + `Retry-After` (= `lease_expires_at + grace_seconds` from now).
  - `not_active_locally` — heartbeat disabled; preserves the original 409. `not_cancellable` — terminal state. `unknown` — not found.
- New store primitive `claim_for_takeover(run_id, grace_seconds, error)`: one atomic conditional `UPDATE … WHERE status IN ('pending','running') AND (lease_expires_at IS NULL OR lease_expires_at < NOW() - grace)`. The conditional WHERE closes the stale-read / concurrent-heartbeat race — if the owner renews between our read and write, the UPDATE matches 0 rows and we surface `lease_valid_elsewhere`.
- HTTP cancel + stream-join endpoints route on outcome. `taken_over` returns 202 immediately without entering `StreamingResponse` — the run is terminal on another worker, so SSE would hang waiting for an END_SENTINEL that will never arrive.

**Refactors**

- `RunManager.grace_seconds` exposed as a public property; the router no longer reaches into `_run_ownership_config`.
- `_is_lease_expired` extracted to a module-level function, shared by `RunManager.cancel()` and `MemoryRunStore.claim_for_takeover()` (previously duplicated).

**Zero-regression**

`heartbeat_enabled=false` (the default, single-worker mode) short-circuits to `not_active_locally` before touching the store. The existing `test_cancel_store_only_run_returns_409` passes unchanged.

## Surface area

- [x] **Backend API** — `POST /api/threads/{tid}/runs/{rid}/cancel` and `POST /api/threads/{tid}/runs/{rid}/stream?action=…` gain a `Retry-After` header on 409 when the run is active on another worker; `taken_over` returns 202. Success status codes unchanged.
- [ ] **Frontend UI** — no change here; honoring `Retry-After` on the client is left for a follow-up.
- [ ] **Agents / LangGraph** — no change.
- [ ] **Sandbox** — no change.
- [ ] **Skills** — no change.
- [ ] **Dependencies** — no change.
- [ ] **Default behavior change** — no change when `heartbeat_enabled=false` (the default). New behaviour only activates for operators who opted into multi-worker mode.
- [x] **Docs / tests / CI only** — `backend/AGENTS.md` RunManager contract updated; 15 new tests.

## Bug fix verification

- Test path that reproduces the bug: `tests/test_multi_worker_run_ownership.py::test_cancel_refuses_active_lease_from_other_worker` (non-owning worker, valid lease → `lease_valid_elsewhere`, store row untouched) and `::test_cancel_takeover_from_crashed_worker` (expired lease → `taken_over`, store row marked `error`).
- Did it go red on `main` and green on this branch? **Yes** — `cancel()` on `main` returns `False` for any non-local run, so both tests would assert against a `bool` instead of `CancelOutcome` and fail. On this branch they pass.
- Race condition pinned by `::test_cancel_takeover_race_owner_renewed_lease` (monkey-patches `claim_for_takeover` to simulate a heartbeat renewal between read and write → `lease_valid_elsewhere`).
- SSE-hang regression guarded by `::test_http_stream_action_interrupt_takeover_returns_202_not_hang` — caught during self-review before pushing.

## Validation

```bash
cd backend && make format && make lint && make test
```

- `make format` / `make lint`: clean.
- 244 directly-related tests pass (7 test files + boundary tests).
- Blocking-IO gate: 36/36 pass.
- Full suite: pre-existing failures unrelated (CSRF/auth, live-HTTP e2e, `GATEWAY_WORKERS=2` env in test runner). Zero failures in files touched by this PR.

## E2E Validation

  Setup: Docker, 2 gateway workers, Postgres, Redis, heartbeat_enabled=true, lease_seconds=30, grace_seconds=10.

  Local cancel (owner worker): POST /cancel → 202. Run transitions to interrupted. Idempotent — second cancel also 202.

  Cross-worker cancel, valid lease: POST /cancel lands on non-owning worker → 409 with Retry-After: 39 (computed from fresh lease_expires_at +
  grace_seconds). Store row untouched. Before this branch the same request returned plain 409 with no retry hint.

  Takeover (dead owner): covered by 6 unit tests — claim_for_takeover atomic path, stale-read race defense, update_status conditional guard, and
  heartbeat self-termination.

  Regression: local cancel, already-interrupted, completed-run, unknown-run, and store-only-with-heartbeat-off all return the same status codes as
  before.
[multi_worker_e2e_validation.md](https://github.com/user-attachments/files/29935655/multi_worker_e2e_validation.md)

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Generated the design from the spec in `docs/multi_worker.md`, implemented the manager / store / router changes and the test matrix. I reviewed every diff, caught the SSE-hang bug during self-review, and take responsibility for the final code.


backend/packages/harness/deerflow/persistence/run/sql.py
```diff
@@ -404,6 +404,30 @@ async def update_lease(
...
+                update(RunRow)
+                .where(
+                    RunRow.run_id == run_id,
+                    RunRow.status.in_(("pending", "running")),
+                    or_(
+                        RunRow.lease_expires_at.is_(None),
+                        RunRow.lease_expires_at < cutoff,
+                    ),
+                )
+                .values(status="error", error=error, updated_at=datetime.now(UTC))
```
willem-bd, line 433: ⚠️ **Split-brain risk: this row flip to `error` can be silently overwritten by a late owner.**

`claim_for_takeover` marks the row `error`, but `RunStore.update_status` is unconditional (`WHERE run_id == X`, no status guard — sql.py:164). If the owning worker is actually still alive (GC pause / event-loop saturation past `lease_seconds + grace_seconds`, but the run task still progressing):

1. The owner's heartbeat calls `update_lease`, which **is** guarded by `status IN (pending,running)` → returns `False`. But `_renew_leases` only treats that as "skip updating in-memory `lease_expires_at`" and keeps the task running.
2. When the owner's task finishes, `_persist_status(record, success)` → `update_status` overwrites `error` → `success`.

Net: the client got 202 (run presumed dead) but the row later flips to `success`, and the owner kept streaming to any client attached to it.

I realize this is the same class of issue as `reconcile_orphaned_inflight_runs` (so not a novel regression), and the grace window makes false positives rare — but cancel makes it directly triggerable by a user action with immediate 202 feedback.

**Suggested fix (closes the hole for both takeover AND reconciliation):** in `_renew_leases`, treat `update_lease(...) is False` on a run this worker still believes it owns as "lease lost/stolen" → `record.task.cancel()` + mark `interrupted`. `updated=False` is a strong signal there (transient DB errors are already retried/raised by `_call_store_with_retry`).

Either way, a one-line comment here (or near `_renew_leases`) documenting the accepted tradeoff would help future readers.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -512,24 +531,44 @@ async def cancel_run(
...
+            try:
+                await record.task
+            except asyncio.CancelledError:
+                pass
+            return Response(status_code=204)
+        return Response(status_code=202)
+
+    # The run is still alive on another worker — tell the client when
+    # to retry (the lease will expire by then).
+    if outcome == CancelOutcome.lease_valid_elsewhere:
```
willem-bd, line 587: **Minor: in the race-lost path, `Retry-After` is computed from a stale value.**

When `cancel()` returns `lease_valid_elsewhere` because `claim_for_takeover` matched 0 rows (owner renewed between our read and write), the store now holds a freshly-renewed lease — but `record.lease_expires_at` was fetched at request start and still holds the *old, expired* value. `_compute_retry_after` then yields `max(1, negative)` ≈ `1`.

Effect: client gets `Retry-After: 1`, retries immediately, and on the second call sees the valid lease → correct value. Self-corrects in one round trip, so non-blocking — just a slightly misleading hint. Could re-fetch the row in the race-lost branch, or have `cancel()` surface the fresh `lease_expires_at`.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -662,41 +682,117 @@ async def update_model_name(self, run_id: str, model_name: str | None) -> None:
...
+        if store_status not in ("pending", "running"):
+            return CancelOutcome.not_cancellable
+
+        grace_seconds = self.grace_seconds
+        lease_expires_at: str | None = row.get("lease_expires_at")
+
+        if not _is_lease_expired(lease_expires_at, grace_seconds=grace_seconds):
+            return CancelOutcome.lease_valid_elsewhere
+
+        take_over_msg = f"Run reclaimed by worker {self._worker_id}: the owning worker ({row.get('owner_worker_id')}) stopped renewing its lease and is presumed dead."
```
willem-bd: Nit: for NULL-lease rows (pre-ownership data), `row.get('owner_worker_id')` is `None`, so the message renders "the owning worker (None) stopped renewing its lease". Log/error-field only, but an `(unknown)` fallback would read cleaner.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -967,6 +1063,11 @@ def heartbeat_enabled(self) -> bool:
             return False
         return self._run_ownership_config.heartbeat_enabled
 
+    @property
+    def grace_seconds(self) -> int:
+        """Return the configured grace seconds (or ``10`` when ownership is unconfigured)."""
+        return self._run_ownership_config.grace_seconds if self._run_ownership_config else 10
```
willem-bd, line 1114: Nit: the `else 10` branch is unreachable on this path — `grace_seconds` is only read after the `if not self.heartbeat_enabled: return not_active_locally` guard, and `heartbeat_enabled` is `False` whenever `_run_ownership_config is None`. Harmless defensive code, just flagging.

willem-bd (commented):
Reviewed against the PR head (`1589979`) — the local checkout is behind the base, so line numbers are verified against head, not local.

Overall this is a clean, well-tested change. The conditional `WHERE` in `claim_for_takeover` is the right place to close the stale-read/heartbeat race, the SSE-hang guard (`taken_over` → 202 before `StreamingResponse`) is a good self-catch, and the `bool`→`CancelOutcome` migration is safe: I audited every `run_mgr.cancel()` caller — only the two router sites here consume the return value; the two calls in `services.py:806,860` discard it (fire-and-forget disconnect cleanup).

Inline comments below for one substantive concern (split-brain on late owner completion) and a few minor nits.

**Test-coverage gaps worth considering** (no single line to anchor these to):
- No test pins the split-brain behavior either way — an explicit test asserting what happens on late owner completion would lock in the intended semantics.
- `claim_for_takeover` has no SQL-store test (only `MemoryRunStore` is exercised); the `or_(lease IS NULL, lease < cutoff)` SQL is the production path and deserves one integration test.
- `_compute_retry_after` (None / unparseable / already-expired) is a trivially-testable pure function with no unit tests.

heart-scalpel:
@WillemJiang hi
 Thanks for the thorough review. All feedback addressed:

  - Split-brain defence: update_status now guards on status IN ('pending','running') so a late owner cannot overwrite a peer's takeover.
  _persist_status recovery path checks row existence before attempting put(). Heartbeat cancels the local task when update_lease returns False.
  - Stale Retry-After: re-fetched the store row in the 409 path so the header is computed from the owner's fresh lease.
  - Nits: (None) → (unknown) in the takeover message; removed unreachable else 10 from grace_seconds.
  - Tests: added 3 unit tests pinning each split-brain guard, plus HTTP-level tests for Retry-After and the SSE-hang regression caught during
  self-review.

  Also ran E2E validation against a real 2-worker Docker stack with Postgres + Redis + heartbeat enabled. Cross-worker cancel returned 409 +
  Retry-After: 39 as expected; local cancel and all regression paths unchanged.

  CI is green.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -340,6 +360,19 @@ async def _persist_status(self, record: RunRecord, status: RunStatus, *, error:
...
             )
             if updated is False:
+                # ``update_status`` is now guarded by ``status IN ('pending','running')``.
+                # False can mean either:
+                #   (a) the row was never persisted (initial ``put()`` failed) → recreate.
+                #   (b) the row is terminal because another worker took it over → skip.
+                existing = await self._store.get(record.run_id)
+                if existing is not None:
+                    logger.warning(
+                        "Run %s status update to %s skipped: store row already at %s (likely takeover)",
```
willem-bd: This "likely takeover" warning fires for *any* `update_status` False-return where the row exists - but False can also mean a normal **cancel-vs-completion race**, not a peer takeover.

Sequence:

1. `cancel()` writes `interrupted` to the store (guard: running→interrupted, OK).
2. The run worker, about to finish, calls `set_status(success)` → `update_status(interrupted→success)` returns False (interrupted ∉ pending/running) → recovery logs "likely takeover".

That's a user cancel winning, not a peer takeover. Since users routinely hit stop as a run finishes, this can fire with non-trivial frequency and noise up operator logs with false "takeover" signals.

`set_status` also sets the in-memory status unconditionally and ignores `_persist_status`'s return, so in-memory (`success`) and store (`interrupted`) diverge in this window - pre-existing in shape (in-memory wins for `get()`), but the new guard flips the store winner from "last write" to "first terminal write".

**Suggestion:** branch the message on `existing["status"]` - `error` from another worker is a real takeover; `interrupted`/`success` is a local cancel/completion race.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -662,41 +695,117 @@ async def update_model_name(self, run_id: str, model_name: str | None) -> None:
...
+                if task_active:
+                    record.task.cancel()
+                record.status = RunStatus.interrupted
+                record.updated_at = _now_iso()
+
+        # Persist outside the lock so store calls don't block other mutations.
+        if record is not None:
+            await self._persist_status(record, RunStatus.interrupted)
+            logger.info("Run %s cancelled (action=%s)", run_id, action)
+            return CancelOutcome.cancelled
```
willem-bd, line 762: This returns `cancelled` (→ HTTP 202) even when `_persist_status` was skipped because a peer had already taken the run over to `error`.

If another worker's `claim_for_takeover` flipped the row to `error` between this worker's in-memory cancel and the `update_status` call, the guard rejects the `interrupted` write (row is terminal), the recovery path returns `False`, and `cancel()` still returns `cancelled` → 202. The store row says `error`, but the client sees "cancelled".

Same shape as the original (which returned `True` regardless of persist outcome), so not a regression - but the outcome could be more precise here. The recovery path already fetches `existing`; when `existing["status"] == "error"`, surfacing `CancelOutcome.taken_over` instead of `cancelled` would let the router return a status consistent with the store.

willem-bd (commented):
Reviewed against the PR head (3fc74bf) - 156 directly-related tests pass, ruff clean. The core mechanism (conditional claim_for_takeover, CancelOutcome mapping, split-brain guards) is sound and the test matrix is thorough. Two low-severity precision/observability nits inline; neither blocks merge.

heart-scalpel:
@WillemJiang  hi 
Thanks for the review catches! Both issues are fixed in commit 24c9f41b.

Reduced log noise: _persist_status checks existing["status"]. Real peer takeover logs as WARNING, expected interrupt/success races log as INFO, eliminating false takeover alerts when stopping completed runs.

Fixed inaccurate cancel outcome: cancel() reloads store state if _persist_status fails. Returns taken_over if the record turned to error mid-operation to align client view with storage. Added race test coverage.

CI is green.

rayhpeng:
Took a close pass over this on top of the merged #4003 infra — the review turnaround here is solid: the split-brain guard, the stale `Retry-After` re-fetch, the log-severity branching, and the precise `taken_over` outcome all landed cleanly, and CI is green. One regression worth fixing before merge, plus a couple of non-blocking notes.

## 🔴 `action=rollback` now persists as `interrupted`, losing the `error` state (regression)

The new `update_status` guard (`WHERE status IN ('pending','running')`) is correct for blocking a late owner from overwriting a takeover's `error`, but it also silently drops a **legitimate** transition on the rollback path.

`worker.py` is untouched by this PR, and its rollback finalization (`set_status(RunStatus.error, "Rolled back by user")`, from #1403) has been the established behavior for months. Sequence:

1. `cancel(action="rollback")` → store `running → interrupted` (guard passes), in-memory `interrupted`.
2. The task raises `CancelledError` → `worker.py:532` calls `set_status(error, "Rolled back by user")`.
3. `update_status(interrupted → error)` → guard is `status IN (pending,running)`, row is `interrupted` → **0 rows → False**.
4. `_persist_status` recovery sees `existing == "interrupted"` (not `error`) → returns False without recreating.

Net: the store stays `interrupted` and the `"Rolled back by user"` message is lost, while the in-memory record is `error` — the two diverge (and `runs` vs `threads_meta` diverge too). Deterministic on `MemoryRunStore`, and the likely outcome on SQL since the `interrupted` write lands before the task's `error` write.

Root cause is altitude: the guard encodes "don't overwrite a takeover's `error`" but is implemented as "don't write any non-`pending`/`running` row," on the shared `update_status` primitive — so it also blocks the same-worker `interrupted → error` finalize. Underneath that is a state-machine ambiguity: `interrupted` is terminal for `interrupt` but an *intermediate* state for `rollback`.

**Suggested fix:** widen the whitelist to `status IN ('pending','running','interrupted')` (SQL + memory). A takeover's `error` still stays locked (it's not in the set, so L1 protection is unchanged), while `interrupted → error` is allowed. Worth a dedicated `action=rollback` test asserting the store's terminal status, and adjusting `test_update_status_rejects_terminal_row` accordingly.

## 🟡 Non-blocking

- **`claim_for_takeover` 0-row result conflates two causes** (`manager.py:828`): the owner renewed the lease (→ genuinely `lease_valid_elsewhere`) vs. the row already went terminal between the read and the claim (run has ended). Both return `lease_valid_elsewhere`, so a finished run gets a `409 + Retry-After: 1` telling the client to retry. It self-corrects in one round trip (the next cancel returns `not_cancellable`), same class as the stale-`Retry-After` already accepted here. A re-read after a lost claim (`error → taken_over`, other terminal → `not_cancellable`) would make it precise.
- **Test coverage:** the SQL `RunRepository.claim_for_takeover` and the new `update_status` guard are only exercised against `MemoryRunStore`; the production SQL path — including the `NULL`-lease / tz-naive comparison in the conditional `WHERE` — has no direct test. The stream endpoint's `lease_valid_elsewhere` 409 branch and `_compute_retry_after`'s NULL/unparseable inputs are also untested.
- There are a few duplication/cold-path cleanups (the `lease_valid_elsewhere` 409 block is copy-pasted across both router endpoints; a taken-over cancel does 2–3 `store.get`s of the same row; the lease-expiry predicate lives in ~4 places). Happy to open a separate follow-up if useful — none block this PR.

heart-scalpel:
@rayhpeng hi 
Thanks for the review.
Rollback regression, deduplication & test coverage
Expanded update_status filter to include interrupted to preserve rollback interrupted → error transition. Protection against overwriting takeover error stays effective.
claim_for_takeover reloads state on failure to distinguish lease_valid_elsewhere / not_cancellable / taken_over.
Extracted shared helpers to remove duplicated 409 blocks and lease expiry logic.
Added 11 tests for SQL claim_for_takeover, retry calculation, recheck logic and stream 409 branch.
Non-blocking items kept as-is per review:
Cold-path repeated store.gets — refactoring would introduce extra coupling with limited gains.
Inline lease cutoff calculation — switching to shared helper would recalculate datetime.now() per row and hurt performance. Both items deferred to follow-up.
CI is green.

rayhpeng:
Verified the latest commit locally — all three fixes land correctly. The widened guard keeps error/success locked (split-brain protection intact, pinned by test_update_status_rejects_terminal_row) while allowing the rollback interrupted → error finalize; the claim_for_takeover re-read cleanly separates the three outcomes; and the SQL claim_for_takeover now has direct coverage. 169 tests green. The two deferred items are reasonable. LGTM — thanks for the thorough turnaround. 🚀

rayhpeng (approved):

backend/packages/harness/deerflow/runtime/runs/store/memory.py
```diff
@@ -194,6 +199,28 @@ async def update_lease(
...
         return True
 
+    async def claim_for_takeover(
+        self,
+        run_id: str,
+        *,
+        grace_seconds: int,
+        error: str,
+    ) -> bool:
+        from deerflow.runtime.runs.manager import _is_lease_expired
```
willem-bd: This backward import (store layer reaching into the manager for a private, `_`-prefixed function) is a layering smell, and it duplicates the lease-expiry parsing that `_compute_retry_after` in the router re-implements a third time. The SQL store sidesteps both by using its own local `_lease_expired_or_null`.

Consider extracting one shared helper (e.g. `parse_lease_expiry(s) -> datetime | None` plus `is_lease_expired(s, grace_seconds) -> bool`) into a util module (`runtime/runs/lease.py` or `utils/time.py`). That removes this backward import, the `_`-private reach-across, and the router/manager duplication in one move.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -967,6 +1114,16 @@ def heartbeat_enabled(self) -> bool:
...
 
+    @property
+    def grace_seconds(self) -> int:
+        """Return the configured grace seconds.
+
+        Only called downstream of ``heartbeat_enabled``, which is False
+        whenever ``_run_ownership_config`` is None, so the attribute is
+        always present here.
+        """
+        return self._run_ownership_config.grace_seconds  # type: ignore[union-attr]
```
willem-bd: The `# type: ignore[union-attr]` relies on the documented invariant that this is only called downstream of `heartbeat_enabled` (which is `False` whenever `_run_ownership_config` is `None`). Safe today -- the router only reaches `grace_seconds` via `lease_valid_elsewhere`, which requires heartbeat on -- but a future caller that invokes it when heartbeat is disabled gets an `AttributeError` at runtime instead of a clear signal.

Consider `return 0 if self._run_ownership_config is None else self._run_ownership_config.grace_seconds` (or raise a typed error) -- costs nothing and drops the type-ignore.

backend/packages/harness/deerflow/runtime/runs/manager.py
```diff
@@ -662,41 +705,145 @@ async def update_model_name(self, run_id: str, model_name: str | None) -> None:
...
+                # re-check the store to see if a peer takeover flipped the
+                # row to ``error`` between our in-memory cancel and the
+                # guarded ``update_status``. If so, surface ``taken_over``
+                # so the client sees a status consistent with the store.
+                try:
+                    existing = await self._store.get(run_id)
+                except Exception:
+                    existing = None
+                if existing is not None and existing.get("status") == "error":
+                    logger.info("Run %s local cancel superseded by peer takeover", run_id)
```
willem-bd, line 759: When a local cancel's persist fails because the store row flipped to `error` (peer takeover), this returns `taken_over` -- but `record.status` was already set to `interrupted` under the lock and the local task was already cancelled. The in-memory record now says `interrupted` while the store says `error`.

Not harmful: the `_persist_status` guard prevents the late finalization from overwriting `error`, and it's pinned by `test_cancel_returns_taken_over_when_peer_claims_during_local_cancel`. But for consistency you could flip `record.status` here to reflect the takeover, or leave a one-line comment noting the transient stale in-memory state is intentional.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -145,6 +146,49 @@ def _cancel_conflict_detail(run_id: str, record: RunRecord) -> str:
...
+    if lease_expires_at is None:
+        return None
+    try:
+        dt = datetime.fromisoformat(lease_expires_at)
+        if dt.tzinfo is None:
+            dt = dt.replace(tzinfo=UTC)
+    except (ValueError, TypeError):
+        return None
+    remaining = (dt - datetime.now(UTC)).total_seconds() + grace_seconds
+    return max(1, int(remaining))
```
willem-bd, line 169: The `max(1, ...)` floor means a lease about to expire yields `Retry-After: 1`. No client consumes this header yet (deferred per the PR body), but when one does, a valid-but-slow owner could drive a tight 409 -> retry -> 409 loop.

Worth a one-line comment noting the floor is a lower bound, not a suggested poll interval -- and when client-side retry lands, apply a minimum backoff / jitter rather than retrying every 1s.

willem-bd (commented):
Reviewed PR head `cd8536ee`.

Well-executed `risk:high` fix. The atomic `claim_for_takeover` (conditional `UPDATE ... WHERE status IN ('pending','running') AND lease expired`) correctly closes the stale-read / concurrent-heartbeat race, and the 0-row-claim re-read distinguishes "owner renewed" from "row went terminal." The SSE-hang on `taken_over` was caught in self-review and guarded. Single-worker mode (`heartbeat_enabled=False`, the default) short-circuits to `not_active_locally`, so the existing 409 path is preserved.

Verified the risky spots:
- `update_status` guard (`status IN ('pending','running','interrupted')`) is safe -- audited all `.update_status()` callers; the only RunStore caller is `_persist_status` (already updated). The others are `ThreadStore` / `TaskRunRepository`, untouched.
- `cancel()` callers in `services.py:806/860` are fire-and-forget (no return-value check), so the `bool` -> `CancelOutcome` change is safe. Since `StrEnum` members are truthy, a missed `if not cancelled:` would have silently broken -- but there are none.
- Takeover is lease-gated: can't be forced on a healthy run.

Test matrix is excellent (28 new tests across SQL + memory stores, HTTP, and manager -- including the stale-read race, SSE-hang regression, the `update_status` guard, heartbeat lease-loss, and the local-cancel-superseded path).

Inline findings below are all low-severity cleanups -- none are merge blockers. Approve.

backend/packages/harness/deerflow/runtime/runs/manager.py
heart-scalpel: Comment the transient in-memory interrupted vs store error state when
  a local cancel is superseded by a peer takeover.

backend/app/gateway/routers/thread_runs.py
heart-scalpel: Comment the max(1, ...) floor in _compute_retry_after — the floor is
  a lower bound, not a poll interval; clients should apply jitter.

PR #4066, opened 2026-07-11
feat(memory): bulk-load sample for registered users

zhangbububu:
## Why

The sample memory loader wrote to the legacy shared memory path, while authenticated users now read from per-user memory storage. As a result, users could run the loader successfully but still see no memory in Settings.

The CLI cannot safely infer which browser user is signed in, so bulk loading must use the registered user database explicitly.

## What changed

- require exactly one of `--target PATH` or `--all-users`
- add `--all-users` to enumerate registered users from the configured persistent database
- back up each user's existing memory before replacing it
- resolve the configured memory manager through the public `get_memory_manager()` API
- honor normal config resolution, including `DEER_FLOW_CONFIG_PATH`
- reject non-persistent database mode and imports that do not persist the sample
- directly test the `load_sample_for_all_users()` runtime wiring
- document UI import for one user and CLI bulk import for disposable review environments

## Surface area

- [x] **Default behavior change** — the loader no longer uses an ambiguous implicit target; callers select `--target` or `--all-users`

## Screenshots / Recording

Not applicable; there is no frontend UI change.

## Bug fix verification

- Test paths: `backend/tests/test_load_memory_sample.py`, `backend/tests/test_auth.py`
- Regression coverage directly calls `load_sample_for_all_users()` with the manager, repository, and persistence boundaries patched
- The documented `--all-users` command was run against the local persistent review database and loaded two users with backups

## Validation

- `PYTHONPATH=. .venv/bin/pytest tests/test_load_memory_sample.py tests/test_auth.py -q` — 78 passed, 1 warning
- `.venv/bin/ruff check ../scripts/load_memory_sample.py tests/test_load_memory_sample.py` — passed
- `.venv/bin/ruff format --check ../scripts/load_memory_sample.py tests/test_load_memory_sample.py` — passed
- full backend suite — 9,612 passed, 45 skipped, 10 failed in 18m 19s; nine failures were live LLM tests with OpenAI connection/SSL errors, and one was `test_create_agent_with_model_and_tool_groups` returning 422 instead of 201

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Investigated the per-user memory path mismatch, implemented the CLI and repository changes, addressed review feedback, wrote tests, updated documentation, and ran verification.


fancyboi999 (changes requested):
Thanks @zhangbububu. I found a few issues that should be addressed before this is ready.

[P1] Run the bulk path through the current memory-manager API

- Location: `scripts/load_memory_sample.py:107` and `scripts/load_memory_sample.py:130-138`
- Problem: the documented `--all-users` command cannot start on the current PR head. It imports `deerflow.agents.memory.updater`, but that module no longer exists after the memory-manager abstraction landed.
- Evidence: running `python scripts/load_memory_sample.py --all-users` at `b2592bfa` raises `ModuleNotFoundError: No module named 'deerflow.agents.memory.updater'`. The current public contract is `get_memory_manager().get_memory(...)` / `.import_memory(...)` in `deerflow.agents.memory.manager`. The focused 67-test selection stays green because it exercises only helper callables and never invokes `load_sample_for_all_users()` or the documented CLI entry point.
- Suggested fix: resolve the configured `MemoryManager` once and pass its `get_memory` / `import_memory` methods into the bulk helper. Avoid importing DeerMem-private updater functions so alternate managers remain supported.
- Test: execute `main(["--all-users"])` against an isolated persistent user database and isolated memory root, then assert every registered user's document was replaced and backed up.

[P2] Honor DeerFlow's configured runtime path before overwriting every user

- Location: `scripts/load_memory_sample.py:116`
- Problem: the loader hard-codes `<repo>/config.yaml`, bypassing the documented `DEER_FLOW_CONFIG_PATH` and project-root resolution. An operator running the command for an alternate review environment can therefore enumerate and overwrite users from the repository-root database instead of the Gateway configuration they selected.
- Evidence: `AppConfig.resolve_config_path()` gives an explicit argument higher priority than `DEER_FLOW_CONFIG_PATH`; this call always supplies the repository-root path. The manager singleton is then initialized from that forced config, so the whole bulk operation consistently targets the wrong environment rather than failing visibly.
- Suggested fix: use the normal `AppConfig.from_file()` resolution, or add an explicit `--config` option that is the sole override and print the resolved config/state paths before starting the bulk write.
- Test: provide two isolated configs with different databases and memory roots, set `DEER_FLOW_CONFIG_PATH` to the second, and assert only the second environment is enumerated and modified.

[P2] Do not report a no-op import as a successful user load

- Location: `scripts/load_memory_sample.py:87-93`
- Problem: `load_sample_for_users()` ignores the document returned by `import_memory` and increments the count unconditionally. A supported `noop` manager, or another manager that declines/normalizes the import, therefore produces `Loaded sample memory for N registered user(s)` even when the sample was not persisted.
- Evidence: `NoopMemoryManager.import_memory()` returns an empty document. Passing an equivalent no-op callback to the current helper returns `1` with zero persisted facts.
- Suggested fix: define and verify the post-import success condition by checking the returned/reloaded document, or explicitly reject managers that cannot persist this replacement operation. Keep the existing backup-before-write ordering.
- Test: cover the `noop` manager and a backend whose import raises or returns a document without the sample, and require a nonzero exit without a success count.

scripts/load_memory_sample.py
```diff
@@ -4,60 +4,166 @@
...
+    sample: dict[str, Any],
+    *,
+    no_backup: bool,
+) -> tuple[int, Path | None]:
+    backend_dir = repo_root / "backend"
+    sys.path.insert(0, str(backend_dir))
+    sys.path.insert(0, str(backend_dir / "packages" / "harness"))
+
+    from app.gateway.auth.repositories.sqlite import SQLiteUserRepository
+    from deerflow.agents.memory.updater import get_memory_data, import_memory_data
```
willem-bd: P1 (blocking): this import path is broken, so `--all-users` crashes at runtime.

`deerflow.agents.memory.updater` is not a real module - there's no `updater.py` under `deerflow/agents/memory/`, and the package `__init__.py` explicitly states that the private DeerMem symbols (`get_memory_data`, `MemoryUpdater`, ...) are "NOT re-exported here." So the first `--all-users` invocation raises `ModuleNotFoundError: No module named 'deerflow.agents.memory.updater'`.

Even with the path corrected, `get_memory_data` / `import_memory_data` are **instance methods** on `MemoryUpdater` (`core/updater.py:559`), reached through a `DeerMem` instance via `self._updater.get_memory_data(...)` / `self._updater.import_memory_data(...)`. They cannot be called as bare functions: `load_memory(user_id=...)` (line ~136) would be missing `self`, and `import_memory(sample, user_id=...)` (line ~137) would bind `sample` to `self`. So the call sites in `load_sample_for_users` are wrong regardless of the import path.

This is the headline feature of the PR and it's uncaught by validation because: (a) `test_load_memory_sample.py` loads the script via `importlib`, which runs only top-level code - the body of `load_sample_for_all_users` (where this import lives) is never executed; (b) the tests mock `load_memory`/`import_memory` and never call `load_sample_for_all_users`; (c) the PR's validation lists pytest suites but not an actual `python scripts/load_memory_sample.py --all-users` run against a DB.

Fix direction: obtain the memory manager through its public factory and call its instance methods, e.g. `from deerflow.agents.memory.manager import get_memory_manager` then `manager = get_memory_manager()`, and call `manager.get_memory(user_id=uid)` / `manager.import_memory(sample, user_id=uid)` (the `MemoryManager` ABC defines `get_memory`/`import_memory`). That also routes through the configured backend so per-user isolation holds. Then add a test that actually exercises `load_sample_for_all_users` with `get_memory_manager` patched so the import wiring is covered.

backend/tests/test_load_memory_sample.py
```diff
@@ -0,0 +1,108 @@
...
+import importlib.util
+from pathlib import Path
+
+import pytest
+
+SCRIPT_PATH = Path(__file__).resolve().parents[2] / "scripts" / "load_memory_sample.py"
+SPEC = importlib.util.spec_from_file_location("load_memory_sample", SCRIPT_PATH)
+assert SPEC is not None and SPEC.loader is not None
+loader = importlib.util.module_from_spec(SPEC)
+SPEC.loader.exec_module(loader)
```
willem-bd, line 14: Test-coverage gap that let the `--all-users` import bug slip through.

`exec_module` runs the script's top-level code only - function bodies are not executed on import. Combined with the mocked `load_memory`/`import_memory` and the fact that no test calls `loader.load_sample_for_all_users`, the entire `--all-users` path (including its `from deerflow.agents.memory.updater import ...` inside the function body) is never exercised. That's why a `ModuleNotFoundError` on a non-existent module survived green tests.

Consider adding a test that drives `load_sample_for_all_users` with `get_memory_manager` (and the user repo / engine init) patched - even a smoke test that just asserts the function's imports resolve and it calls the manager's `get_memory`/`import_memory` per user would catch this class of bug. The existing `load_sample_for_users` tests are good unit coverage for the backup/import ordering, but they can't validate the wiring inside `load_sample_for_all_users`.

willem-bd (commented):
Fresh review. One P1 blocking issue: the `--all-users` path imports `get_memory_data`/`import_memory_data` from `deerflow.agents.memory.updater`, which is not a real module (no `updater.py` under `deerflow/agents/memory/`; the package `__init__.py` explicitly states these private DeerMem symbols are NOT re-exported), and those names are instance methods on `MemoryUpdater`, not module-level functions - so `--all-users` raises `ModuleNotFoundError` at runtime. The unit tests don't catch it because they mock the memory functions and never execute `load_sample_for_all_users`. Suggested fix: use `get_memory_manager()` and call its instance methods. Details inline.

WillemJiang:
@zhangbububu  please take a look at the latest code review comment.

zhangbububu:
@WillemJiang @fancyboi999 
Addressed the remaining review feedback in `50a6944c`:

- use the public configured `MemoryManager` for `--all-users`
- directly test `load_sample_for_all_users()`
- honor `DEER_FLOW_CONFIG_PATH` through normal config resolution
- reject imports whose returned document does not contain the sample

Focused validation: 78 passed; Ruff check and format check passed. I also updated the PR description with the full backend-suite result and its remaining environment/test failures.

scripts/load_memory_sample.py
zhangbububu: Fixed in `50a6944c`. The bulk path now resolves the configured manager once through `get_memory_manager()` and passes its `get_memory` / `import_memory` methods into the helper.

backend/tests/test_load_memory_sample.py
zhangbububu: Fixed in `50a6944c`. Added a direct `load_sample_for_all_users()` test that patches the manager, user repository, and persistence lifecycle, then verifies both users are backed up and imported.

WillemJiang (approved):

PR #4068, opened 2026-07-11
fix(frontend): avoid highlighting code while streaming

### LINKED ISSUE #4061: [bug] 任务执行中的时候，该网页标签切换为其他标签页，再返回后，原始整个页面是空白的，需要重复刷新，体验不佳
### Before you start

### Problem summary

用户点击其他标签页后，再返回原始任务页面后，出现空白，需要手动刷新，才能显示。

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

返回原始页面后，是空白的
Image

Image

### Expected behavior

我希望，切换页面回来后，能够直接显示，而不是需要手动刷新，或者等待很久，谢谢！

### Steps to reproduce

1、执行长一点的任务，分析，多智能体任务
2、生成 html 复杂报告
3、生成的过程中，我点开查看其他网页上方其他标签页
4、返回原始任务页面，后，是空白的，任务还在执行，但是需要我重复刷新呢

### Relevant logs

```shell
同上
```

Docker (make docker-start)

### Operating system

macOS

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Support bundle summary

_No response_

### Additional context

_No response_

Huixin615:
Fixes #4061

## Why

During long-running tasks that generate complex HTML reports, the assistant may stream a large fenced code block over many updates.

Previously, every streaming update caused Streamdown to re-render the growing code block and Shiki to regenerate both light and dark syntax-highlighted HTML. When the page returned from a background browser tab, the accumulated rendering work could make the entire DeerFlow workspace unresponsive or appear blank until refreshed.

## What changed

- Render fenced code blocks as lightweight plain `<pre><code>` content while the assistant response is still streaming.
- Preserve inline-code rendering and custom Markdown link/image components during streaming.
- Restore the existing Streamdown and Shiki syntax highlighting after streaming finishes.
- Add regression coverage for streaming fenced code, inline code, completed-response highlighting, and custom renderers.

This does not change response content, backend streaming, task execution, or artifact rendering.

## Surface area

## Screenshots / Recording

The original behavior and blank workspace screenshot are available in #4061.

The issue occurs while a long HTML response is actively streaming, so a short before/after recording is more representative than a static screenshot. Manual verification used the original reproduction flow: generate a complex HTML report, switch browser tabs during execution, and return while the response is still streaming.

## Bug fix verification

  `frontend/tests/unit/components/workspace/messages/markdown-content.test.ts`
- Did it go red on `main` and green on this branch? **Yes**
- Before the fix, streaming fenced HTML produced Streamdown/Shiki code-block containers. After the fix, streaming uses the lightweight renderer and restores the default highlighted renderer once the response finishes.
- A Docker-based manual reproduction also confirmed that large HTML streaming could make Chrome report the page as unresponsive.

## Validation

Executed:

```bash
cd frontend
pnpm exec rstest tests/unit/components/workspace/messages/markdown-content.test.ts
pnpm test
pnpm format
pnpm check
cd ..
git diff --check

Results:
Targeted regression tests: 4 passed
Full frontend unit suite: 608 passed
Prettier: passed
ESLint: passed
TypeScript: passed
git diff --check: passed

### DIFF
--- a/frontend/src/components/workspace/messages/markdown-content.tsx
+++ b/frontend/src/components/workspace/messages/markdown-content.tsx
@@ -1,6 +1,13 @@
 "use client";
 
-import { type ComponentProps, type ReactNode, useMemo } from "react";
+import {
+  createContext,
+  type ComponentProps,
+  isValidElement,
+  type ReactNode,
+  useContext,
+  useMemo,
+} from "react";
 
 import { type ClipboardSafeStreamdownProps } from "@/components/ai-elements/streamdown";
 import {
@@ -22,26 +29,48 @@ export type MarkdownContentProps = {
 };
 
 type StreamingCodeProps = ComponentProps<"code"> & {
-  node?: {
-    position?: {
-      start: { line: number };
-      end: { line: number };
-    };
-  };
+  node?: unknown;
   children?: ReactNode;
 };
+
+const StreamingCodeBlockContext = createContext(false);
+
+function StreamingPre({ children }: ComponentProps<"pre">) {
+  const childClassName = isValidElement<{ className?: string }>(children)
+    ? children.props.className
+    : undefined;
+  const language =
+    /(?:^|\s)language-([^\s]+)/.exec(childClassName ?? "")?.[1] ?? "";
+
+  return (
+    <div
+      className="my-4 w-full overflow-hidden rounded-xl border"
+      data-language={language}
+      data-streaming-code-block="true"
+    >
+      {language && (
+        <div className="bg-muted/80 text-muted-foreground p-3 text-xs">
+          <span className="ml-1 font-mono lowercase">{language}</span>
+        </div>
+      )}
+      <pre className="bg-muted/40 overflow-x-auto border-t p-4 font-mono text-xs">
+        <StreamingCodeBlockContext.Provider value={true}>
+          {children}
+        </StreamingCodeBlockContext.Provider>
+      </pre>
+    </div>
+  );
+}
 
 function StreamingCode({
   children,
   className,
-  node,
+  node: _node,
   ...props
 }: StreamingCodeProps) {
-  const isInline =
-    node?.position?.start.line === node?.position?.end.line &&
-    !className?.includes("language-");
+  const isBlock = useContext(StreamingCodeBlockContext);
 
-  if (isInline) {
+  if (!isBlock) {
     return (
       <code
         {...props}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: frontend/src/components/workspace/messages/markdown-content.tsx [hunk 1]
```diff
@@ -20,6 +21,62 @@ export type MarkdownContentProps = {
...
+};
+
+function StreamingCode({
+  children,
+  className,
+  node,
+  ...props
+}: StreamingCodeProps) {
+  const isInline =
+    node?.position?.start.line === node?.position?.end.line &&
```
willem-bd: **Nit (non-blocking):** The inline/block heuristic misclassifies one narrow case — a fenced block with a single line of content and no language (e.g. ` ``` \nx\n ``` `) would be treated as inline during streaming, since `start.line === end.line` and there's no `language-` class. It self-heals the moment streaming finishes (so cosmetic only), but if you want to be robust you could key off the parent node (fenced `<code>` lives under a `<pre>`) rather than position + class alone.

frontend/src/components/workspace/messages/markdown-content.tsx
```diff
@@ -39,11 +96,18 @@ export function MarkdownContent({
...
       a: createMarkdownLinkComponent(),
       ...componentsFromProps,
     };
-  }, [componentsFromProps]);
+    if (!isLoading) {
+      return baseComponents;
+    }
+    return {
+      ...baseComponents,
+      code: StreamingCode,
```
willem-bd: **Nit (non-blocking):** During streaming this unconditionally overwrites a `code` component a caller might pass via `components`. No current caller does (message-group/subtask-card only pass `a`/`img`-style renderers), so it's not a live bug — but if you want it defensive: `code: componentsFromProps?.code ?? StreamingCode`.

willem-bd (commented):
Reviewed the streaming code-highlight bypass. Overall this is a solid, well-scoped fix for #4061 — overriding the `code` component correctly bypasses Streamdown/Shiki on the hot path, the `isLoading` memo dep is correctly added, and the regression coverage (marker swap + escaped content + custom-renderer preservation) is good. Two non-blocking nits inline.

One broader note: the streaming block uses different chrome than the final Shiki block, so long code "snaps" on stream completion. Acceptable tradeoff for responsiveness, just a visible default-behavior change (correctly noted under Surface area).

WillemJiang (approved):

PR #4072, opened 2026-07-11
fix(middleware): window LoopDetection tool-frequency counter so long runs don't false-trip

nankingjing:
## Summary
`LoopDetectionMiddleware`'s per-tool frequency guard counted tool-name usage with a monotonic counter that never decayed. A long, legitimate run that calls a tool many times spread across the whole conversation would eventually trip the loop hard-stop even though the calls were not a tight loop.

## Problem
`freq[name] += 1` accumulated for the life of the thread. There was no sliding window, so the "same tool used too often" heuristic conflated "used 20 times over 200 turns" (normal) with "used 20 times in 20 turns" (an actual loop), producing false hard-stops on long research runs.

## Fix
Replace the monotonic counter with a windowed `deque` (`_tool_name_history`) so only recent calls count toward the frequency threshold. The existing "warn once" de-dup guarantee is preserved with a decay-aware `_tool_freq_warned` guard: a tool is warned once while over threshold and re-armed only after it decays back below, with eviction/reset cleanup so per-thread state cannot leak.

## Test
`backend/tests/test_loop_detection_middleware.py` updated to cover windowed decay and the warn-once-per-decay semantics. Full file passes (70 passed).

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
@@ -55,7 +55,7 @@ import hashlib
 import json
 import logging
 import threading
-from collections import OrderedDict, defaultdict, deque
+from collections import Counter, OrderedDict, defaultdict, deque
 from collections.abc import Awaitable, Callable
 from copy import deepcopy
 from typing import TYPE_CHECKING, override
@@ -199,12 +199,14 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
             Default: 20.
         max_tracked_threads: Maximum number of threads to track before
             evicting the least recently used. Default: 100.
-        tool_freq_warn: Number of calls to the same tool *type* (regardless
-            of arguments) before injecting a frequency warning. Catches
-            cross-file read loops that hash-based detection misses.
-            Default: 30.
-        tool_freq_hard_limit: Number of calls to the same tool type before
-            forcing a stop. Default: 50.
+        tool_freq_warn: Maximum number of same-tool-type calls within a
+            sliding window of ``_tool_freq_window`` before injecting a
+            frequency warning. Catches cross-file read loops that
+            hash-based detection misses. Default: 30 (within a window
+            of 50).
+        tool_freq_hard_limit: Maximum number of same-tool-type calls within
+            a sliding window of ``_tool_freq_window`` before forcing a
+            stop. Default: 50 (within a window of 50).
         tool_freq_overrides: Per-tool overrides for frequency thresholds,
             keyed by tool name. Each value is a ``(warn, hard_limit)`` tuple
             that replaces ``tool_freq_warn`` / ``tool_freq_hard_limit`` for
@@ -241,7 +243,7 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
         # the largest hard limit in play (global + every per-tool override) so a
         # tight burst can actually reach it while spread-out calls still decay
         # out of the window. Warn thresholds are intentionally excluded: a sane
-        # config has warn <= hard (covered by sizing to hard), and a misconfig
+        # config enforces warn <= hard (covered by sizing to hard), and a misconfig
         # with warn > hard would hard-stop first anyway, so an unreachable warn
         # is harmless and must not inflate the window.
         self._tool_freq_window = max(
@@ -255,7 +257,13 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
         # Windowed per-tool-type frequency: recent tool names per thread,
         # trimmed to ``window_size`` so the count decays instead of growing
         # monotonically (replaces the old monotonic ``_tool_freq`` integer).
-        self._tool_name_history: dict[str, deque[str]] = {}
+        self._tool_name_history: defaultdict[str, deque[str]] = defaultdict(deque)
+        # Per-thread Counter mirroring the deque so freq_count is O(1) instead
+        # of scanning the whole window on every tool call. A single high
+        # per-tool override (e.g. bash: {hard_limit: 1000}) inflates the window
+        # globally, so the scan would cost 1000 per call for every tool; Counter
+        # increments on append and decrements on popleft.
+        self._tool_name_counter: defaultdict[str, Counter[str]] = defaultdict(Counter)
         # Per-thread set of tool names already warned about in Layer 2, so a
         # frequency warning is enqueued once rather than on every subsequent
         # call. Cleared per name when the windowed count decays back below the
@@ -474,7 +482,8 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
                     return _WARNING_MSG, False
 
             # --- Layer 2: per-tool-type frequency (windowed) ---
-            tool_name_history = self._tool_name_history.setdefault(thread_id, deque())
+            tool_name_history = self._tool_name_history[thread_id]
+            name_counter = self._tool_name_counter[thread_id]
             for tc in tool_calls:
                 name = tc.get("name", "")
                 if not name:
@@ -482,11 +491,18 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
                 # Windowed counting: append the name and trim to the frequency
                 # window (>= the largest threshold) so the count can reach the
                 # warn/hard limits on a tight burst yet still decay for
-                # spread-out calls.
+                # spread-out calls. A mirrored Counter gives O(1) freq_count
+                # even when a per-tool override inflates the window globally.
                 tool_name_history.append(name)
+                name_counter[name] += 1
                 while len(tool_name_history) > self._tool_freq_window:
-                    tool_name_history.popleft()
-                freq_count = sum(1 for n in tool_name_history if n == name)
+                    old = tool_name_history.popleft()
+                    c = name_counter[old] - 1
+                    if c <= 0:
+                        del name_counter[old]
+                    else:
+                        name_counter[old] = c
+                freq_count = name_counter.get(name, 0)
 
                 if name in self._tool_freq_overrides:
                     eff_warn, eff_hard = self._tool_freq_overrides[name]
--- a/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
@@ -55,7 +55,7 @@ import hashlib
 import json
 import logging
 import threading
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict, defaultdict, deque
 from collections.abc import Awaitable, Callable
 from copy import deepcopy
 from typing import TYPE_CHECKING, override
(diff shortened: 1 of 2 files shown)


willem-bd (commented):

[P1] Windowing silently disables the Layer-2 frequency guard under the shipped default config
- Location: `backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py:449-470` (windowed `freq_count`), with defaults from `loop_detection_config.py:45-57` and `config.example.yaml:939-945`.
- Problem: The windowed `freq_count` is bounded by `window_size` (default **20**), but both Layer-2 thresholds are strictly larger: `tool_freq_warn` (default **30**) and `tool_freq_hard_limit` (default **50**). Since `freq_count = sum(1 for n in tool_name_history if n == name)` can never exceed the deque length, and the deque is trimmed to `window_size` (20), it can never reach 30 or 50. With the shipped defaults and no per-tool overrides (the `tool_freq_overrides` block is commented out in `config.example.yaml`), Layer 2 never warns and never hard-stops — for any tool.
- Evidence:
  - Defaults: `_DEFAULT_WINDOW_SIZE = 20`, `_DEFAULT_TOOL_FREQ_WARN = 30`, `_DEFAULT_TOOL_FREQ_HARD_LIMIT = 50` (lines 79-82); identical in `LoopDetectionConfig` and `config.example.yaml`.
  - The deque is capped at `window_size`: `while len(tool_name_history) > self.window_size: tool_name_history.popleft()` (lines 450-451) ⟹ `freq_count ≤ 20`.
  - So `freq_count` (≤20) `< eff_warn` (30) and `< eff_hard` (50) for every tool without an override ⟹ the `if freq_count >= eff_hard` (line 459) and `if freq_count >= eff_warn` (line 470) branches are dead code under default config.
  - The pre-PR counter was unbounded (`freq[name] += 1`, monotonic for the life of the thread) and would reach 30 then 50; the windowed counter cannot. This regression is introduced purely by this diff.
  - Impact: a genuine tight loop that calls one tool type with *varying arguments* — exactly the case Layer 2 was added to catch, since Layer 1 hashes name+args and misses varying-args loops — is no longer broken by Layer 2 and runs until the recursion/token backstop. The PR's stated goal is to keep the guard firing for tight bursts while relaxing for spread-out calls; under the default config it does neither.
  - The two new tests pass only because they manually set `window_size=5, tool_freq_hard_limit=4` (window > hard). No test exercises the shipped default ratio (`window=20 < warn=30 < hard=50`), which is exactly where the bug lives.
- Suggested fix: Don't reuse the Layer-1 `window_size` for the Layer-2 frequency window — size the frequency window to the hard limit instead. Minimal option: bound the frequency deque by `max(self.window_size, self.tool_freq_hard_limit)` (and the max of every per-tool override's `hard_limit`), or add a dedicated `tool_freq_window_size` config field defaulting to `>= tool_freq_hard_limit`. If you intend to keep sharing `window_size`, add a validator (e.g. in `LoopDetectionConfig.validate_thresholds`) requiring `window_size >= tool_freq_warn` / `>= tool_freq_hard_limit`, so a configuration that silently disables the guard is rejected rather than shipped.
- Test: Add a test using the real defaults (or `LoopDetectionMiddleware.from_config(LoopDetectionConfig())`) that calls one tool type with distinct args in a tight burst with no interleaving, and asserts the hard stop fires. This fails today under default config and would guard the regression.

nankingjing:
Great catch @willem-bd — you're right, reusing `window_size` (20) capped `freq_count` below `tool_freq_warn` (30) and `tool_freq_hard_limit` (50), so Layer 2 was dead code under the shipped defaults, and my two tests only passed because they set `window > hard`.

Fixed by sizing a dedicated frequency window independent of Layer 1:
```python
self._tool_freq_window = max(
    self.window_size,
    self.tool_freq_hard_limit,
    *(hard for _, hard in self._tool_freq_overrides.values()),
)
```
so the deque can actually reach the hard limit on a tight burst while spread-out calls still decay out of the window. I excluded the *warn* thresholds from the max on purpose: a sane config has `warn <= hard` (covered by sizing to hard), and a `warn > hard` misconfig hard-stops first anyway, so an unreachable warn is harmless and shouldn't inflate the window.

Added the regression tests you asked for, all against the real default config (`from_config(LoopDetectionConfig())`):
- `test_freq_window_sized_to_hard_limit_under_defaults` — window >= hard/warn
- `test_freq_window_covers_largest_override_hard_limit`
- `test_tight_burst_hard_stops_under_default_config` — one tool type, distinct args (so Layer 1's hash never fires), tight burst, asserts the hard stop fires at the limit and not before

Full file: 73 passed.

nankingjing:
Thanks @willem-bd — you're right, the Layer-2 frequency window was sized to `window_size` (20), which can never reach the default `tool_freq_hard_limit` (50), so the guard was dead under shipped defaults.

Fixed in 571c94f: the per-tool-type frequency deque now uses a dedicated `_tool_freq_window = max(window_size, tool_freq_hard_limit, *per-tool-override hard limits)`, so `freq_count` can reach the warn/hard thresholds on a tight burst while Layer-1's own `window_size` behavior is untouched.

Added `test_tight_burst_hard_stops_under_default_config` (one tool called with distinct args under the real default config → hard stop fires — the exact varying-args loop Layer 1's name+args hash misses), plus `test_freq_window_sized_to_hard_limit_under_defaults` and `test_freq_window_covers_largest_override_hard_limit`.

WillemJiang:
@nankingjing, please fix the lint error. BTW, don't flood the PR with AI.

nankingjing:
Lint fixed in 0544ddf — the test file `test_loop_detection_middleware.py` had a few extra blank lines that `ruff format --check` flagged. Verified locally with `ruff format --check . && ruff check .` both pass. Sorry for the noise on the PR thread.

nankingjing:
Lint fix from your earlier note is in `0544ddf`; `flake8` + `mypy` clean in the local container, regression test still green. Could you re-run the gating CI when convenient? Branch `fix/loop-detection-freq-decay`, no further changes since that commit.

willem-bd (commented):
Overall: the core fix is correct and well-tested - approve with minor changes. The windowed decay, warn-once-per-decay re-arming (lazy `discard` on the next sub-threshold call), and thread-safety all check out, and I verified the full `test_loop_detection_middleware.py` passes (73 tests) against this branch. `_tool_freq` is fully replaced (no stale refs, no external readers), and `LoopDetectionConfig` enforces `warn <= hard` globally and per-override, so the window-sizing assumption holds.

Inline: docs (the one I'd most want updated before merge), perf, alternating-loop tradeoff, and a `setdefault` nit.

One more non-blocking note not tied to a specific line: small test gap - there's no end-to-end test of the warn-once-then-rearm-after-decay path (warn at threshold, count decays below, burst again -> warns again). The logic is correct by trace, but a test would lock it in alongside the existing warn-once test.

backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
```diff
@@ -233,10 +233,33 @@ def __init__(
...
+        # compared against — otherwise the hard-stop branch is dead code. Do NOT
+        # reuse Layer 1's ``window_size`` (which is unrelated and defaults below
+        # the freq thresholds, e.g. 20 < hard 50); size the frequency window to
+        # the largest hard limit in play (global + every per-tool override) so a
+        # tight burst can actually reach it while spread-out calls still decay
+        # out of the window. Warn thresholds are intentionally excluded: a sane
+        # config has warn <= hard (covered by sizing to hard), and a misconfig
+        # with warn > hard would hard-stop first anyway, so an unreachable warn
+        # is harmless and must not inflate the window.
+        self._tool_freq_window = max(
```
willem-bd, line 249: The one thing I'd ask to fix before merge: the user-facing docs still describe the *old* monotonic semantics. The `__init__` docstring (`tool_freq_hard_limit`: "Number of calls to the same tool type before forcing a stop. Default: 50.") and the config field description ("Number of calls to the same tool type before forcing a stop") both read as a plain lifetime counter - a user tuning this (or reading telemetry showing `count=50`) will expect "50 total, ever," not "50 within a sliding window of 50." For a `risk:high` behavioral change on a safety guard, both should state the windowed semantics and point at this sizing rule.

(Side note on this block's comment: "a sane config has warn <= hard" is actually *enforced* by `LoopDetectionConfig.validate_thresholds` and `ToolFreqOverride`, so it could say "enforced" rather than "sane" - minor.)

backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
```diff
@@ -431,44 +454,54 @@ def _track_and_check(self, state: AgentState, runtime: Runtime) -> tuple[str | N
                     )
                     return _WARNING_MSG, False
 
-            # --- Layer 2: per-tool-type frequency ---
-            freq = self._tool_freq[thread_id]
+            # --- Layer 2: per-tool-type frequency (windowed) ---
+            tool_name_history = self._tool_name_history.setdefault(thread_id, deque())
```
willem-bd: Minor: `deque()` is constructed as the `default` arg on every call and thrown away when the key already exists (the common case after a thread's first call). This is a hot path (every tool call). `self._tool_name_history: dict[str, deque[str]] = defaultdict(deque)` avoids the per-call allocation - and the file already uses that idiom for `_tool_freq_warned` (`defaultdict(set)`) and the old `_tool_freq` (`defaultdict(lambda: defaultdict(int))`).

backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
```diff
@@ -431,44 +454,54 @@ def _track_and_check(self, state: AgentState, runtime: Runtime) -> tuple[str | N
...
                 if not name:
                     continue
-                freq[name] += 1
-                tc_count = freq[name]
+                # Windowed counting: append the name and trim to the frequency
+                # window (>= the largest threshold) so the count can reach the
+                # warn/hard limits on a tight burst yet still decay for
+                # spread-out calls.
+                tool_name_history.append(name)
+                while len(tool_name_history) > self._tool_freq_window:
```
willem-bd, line 498: Design tradeoff to confirm (the `risk:high` label makes me want explicit sign-off): the windowed decay means a loop that *alternates* tools with **distinct args** - e.g. `read_file(x) -> bash(y) -> read_file(z) -> ...` - never reaches `tool_freq_hard_limit` same-tool calls in any single window (each tool sits around `window/2`), so Layer 2 no longer hard-stops it. The old monotonic counter would have stopped at the 50th `read_file`. Layer 1's hash detector still catches this if args repeat; only varying-arg alternation slips through. `test_windowed_frequency_decay_avoids_hard_stop_when_interleaved` asserts no hard-stop on exactly this pattern, so the intent is clear - just flagging that varying-arg alternating loops are now a class of loop Layer 2 won't catch. Acceptable given they're genuinely ambiguous (could be legit research)?

backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
```diff
@@ -431,44 +454,54 @@ def _track_and_check(self, state: AgentState, runtime: Runtime) -> tuple[str | N
...
+                    tool_name_history.popleft()
+                freq_count = sum(1 for n in tool_name_history if n == name)
```
willem-bd: This `sum(...)` scans the whole deque every call - O(`_tool_freq_window`) per tool call. Fine at the default window of 50, but `_tool_freq_window = max(window_size, tool_freq_hard_limit, *override_hard_limits)`, so a single high per-tool override (e.g. `bash: {hard_limit: 1000}`) inflates the window *globally*: every tool call on every thread then scans 1000 entries (the inflation isn't per-tool). Not a correctness issue - the count threshold is unchanged, so detection isn't delayed - just per-call cost. A `collections.Counter` (increment on `append`, decrement on `popleft`) would make this O(1) amortized and sidestep the global-inflation cost entirely. Worth it if high overrides are expected to be common.

nankingjing:
Fixed the docstrings to describe the actual windowed semantics (not just "history"). Switched from O(n) list scans to defaultdict+Counter for the per-name frequency count within the window. The warn <= hard invariant is already enforced by the config layer — just changed the comment to reflect that.

nankingjing:
All review comments addressed — lint fixed, docstrings updated for windowed semantics, defaultdict+Counter optimization applied. CI green. Ready for re-review when convenient.

nankingjing:
All review comments addressed: lint fixed, docstrings updated for windowed semantics, defaultdict+Counter optimization applied. CI green. Ready for re-review when convenient.

nankingjing:
Rebased onto latest main (81b3ed01) to resolve merge-base drift. No code changes, CI was already green.

WillemJiang (approved):

PR #4074, opened 2026-07-11
fix(memory): coerce null source.confidence so it no longer blocks memory updates

nankingjing:
## Summary
A memory fact whose `source.confidence` is `null` (absent) caused numeric comparisons in the memory updater to fail, blocking otherwise-valid memory updates.

## Problem
`updater.py` compared raw `confidence` values that could be `None`, so a single unscored fact from the model aborted the update cycle (mixed `None`/`float` comparison).

## Fix
Route confidence reads through `_coerce_source_confidence(...)` at the two comparison sites (~L595 and ~L1041), coercing missing/invalid confidence to a safe default so updates proceed deterministically.

## Test
`backend/tests/test_memory_updater.py` adds coverage for null-confidence facts (3 passed).


WillemJiang (approved):

WillemJiang:
@nankingjing, please fix the lint error.

nankingjing:
Lint fixed in bcdfa02 — `tests/test_memory_updater.py` was missing a `ruff format` pass. Verified locally: `ruff format --check . && ruff check .` both pass.

nankingjing:
Thanks @WillemJiang for the approval! Lint fix landed in bcdfa02 (and `tests/test_memory_updater.py` is now ruff-clean). Anything I can do to help push this through to merge — rebase, squash, or anything else?

PR #4075, opened 2026-07-11
fix(memory): make factsToRemove optional so add-only memory updates aren't rejected

nankingjing:
## Summary
The memory-update response parser required `factsToRemove` as a mandatory top-level key. Valid model responses that only added facts (no removals) were rejected, silently dropping legitimate memory updates.

## Problem
`_REQUIRED_MEMORY_UPDATE_TOP_LEVEL_KEYS` included `factsToRemove`, which represents an optional operation, so the strict gate over-rejected add-only payloads.

## Fix
Narrow the required set to `frozenset({"user", "history", "newFacts"})` and treat `factsToRemove` as optional (default empty). Add-only updates now parse and apply.

## Test
`backend/tests/test_memory_updater.py` covers an add-only payload without `factsToRemove` (2 passed).

### DIFF
--- a/backend/packages/harness/deerflow/agents/memory/updater.py
+++ b/backend/packages/harness/deerflow/agents/memory/updater.py
@@ -323,7 +323,7 @@ def _extract_text(content: Any) -> str:
     return str(content)
 
 
-_REQUIRED_MEMORY_UPDATE_TOP_LEVEL_KEYS = frozenset({"user", "history", "newFacts", "factsToRemove"})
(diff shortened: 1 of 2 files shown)


nankingjing:
The "Approve and run" button on the bytedance-cla fork-PR gate is what's blocking CI here — the workflow fires only on a maintainer click. Could @WillemJiang or any other maintainer with write access hit it when convenient? Branch `fix/factsToRemove-parse-gate`, no review comments yet.

WillemJiang (approved):

PR #4077, opened 2026-07-11
fix(runtime): serialize SQLite event-store writes to prevent per-thread seq collisions

nankingjing:
## Summary
The SQLite run-event store could assign colliding sequence numbers when two coroutines wrote events for the same thread concurrently, corrupting event ordering and violating the `UNIQUE(thread_id, seq)` invariant.

## Problem
`put()` / `put_batch()` computed the next seq and inserted without a per-thread write lock, so interleaved writers on the same thread could read the same max seq and both write it. Two different runs within the same thread collide too, which is why the lock must be thread-scoped rather than run-scoped.

## Fix
Add a per-`thread_id` `asyncio.Lock` registry (`_write_locks` + `_get_write_lock()`) and wrap `put`/`put_batch` so seq assignment + insert is atomic per thread. Locks are keyed by `thread_id`, matching the `UNIQUE(thread_id, seq)` constraint, so different threads still write in parallel. `delete_by_thread` evicts the lock entry afterward so `_write_locks` doesn't grow unbounded over the store's lifetime.

## Test
`backend/tests/test_run_event_store.py` adds a concurrent-write regression plus a lock-eviction test.

### DIFF
--- a/backend/packages/harness/deerflow/runtime/events/store/db.py
+++ b/backend/packages/harness/deerflow/runtime/events/store/db.py
@@ -321,6 +321,14 @@ class DbRunEventStore(RunEventStore):
             if count > 0:
                 await session.execute(delete(RunEventRow).where(*count_conditions))
                 await session.commit()
+            # Evict the per-thread seq-assignment lock so ``_write_locks`` does
+            # not grow unbounded over the (long-lived, singleton) store's
+            # lifetime. Only pop when no writer is mid-flight; a later write
+            # recreates the lock lazily and seq restarts correctly from the
+            # now-deleted thread.
+            lock = self._write_locks.get(thread_id)
+            if lock is not None and not lock.locked():
+                self._write_locks.pop(thread_id, None)
             return count
 
     async def delete_by_run(
--- a/backend/packages/harness/deerflow/runtime/events/store/db.py
+++ b/backend/packages/harness/deerflow/runtime/events/store/db.py
@@ -6,6 +6,7 @@ at ``max_trace_content`` bytes to avoid bloating the database.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 from datetime import UTC, datetime
@@ -26,6 +27,20 @@ class DbRunEventStore(RunEventStore):
     def __init__(self, session_factory: async_sessionmaker[AsyncSession], *, max_trace_content: int = 10240):
         self._sf = session_factory
         self._max_trace_content = max_trace_content
+        # Per-thread asyncio locks serialize seq assignment for concurrent
+        # in-process writers on the same thread. The DB-level FOR UPDATE /
+        # advisory lock guards cross-process races; this guards the common
+        # single-process case where two coroutines interleave between the
+        # max(seq) read and the INSERT and would otherwise collide on seq.
+        self._write_locks: dict[str, asyncio.Lock] = {}
+
+    def _get_write_lock(self, thread_id: str) -> asyncio.Lock:
+        """Return (creating if needed) the per-thread seq-assignment lock."""
+        lock = self._write_locks.get(thread_id)
+        if lock is None:
+            lock = asyncio.Lock()
+            self._write_locks[thread_id] = lock
+        return lock
 
     @staticmethod
     def _row_to_dict(row: RunEventRow) -> dict:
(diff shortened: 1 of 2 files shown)


willem-bd (commented):

The core fix looks correct: the `max(seq)` read and the INSERT now both run inside the per-`thread_id` `asyncio.Lock` critical section, so two in-process coroutines writing the same thread can no longer both compute the same next seq. Locking per `thread_id` (rather than per `run_id`) is the right granularity, since the uniqueness invariant is `UNIQUE(thread_id, seq)` and two different runs in the same thread would otherwise still collide — the in-code comment is accurate on this. The `async with` releases the lock on exception (transaction rolls back, so no seq gap), and the in-process lock is a strict subset of the Postgres `pg_advisory_xact_lock` protection, so the Postgres path is unaffected. `_get_write_lock` is a plain sync method with no `await`, so the check-then-create is atomic under the single-threaded event loop.

[P2] `_write_locks` registry grows unbounded — orphaned locks after `delete_by_thread`
- Location: `backend/packages/harness/deerflow/runtime/events/store/db.py:38` (`_get_write_lock`) and `delete_by_thread` (line ~308)
- Problem: `_write_locks` is a per-instance `dict[str, asyncio.Lock]` that is only ever inserted into (`self._write_locks[thread_id] = lock`); there is no eviction anywhere. `DbRunEventStore` is a long-lived singleton, and the keys are `thread_id`s, so over the process lifetime the registry accumulates one Lock per thread ever seen and never releases it. `delete_by_thread` removes the events for a thread but leaves the orphaned lock entry behind, so a server that creates then deletes many threads still leaks proportional to lifetime-unique threads.
- Evidence: Introduced by this diff — `grep` of the new file shows `_write_locks` written only at `db.py:42` and never read-removed; `delete_by_thread`/`delete_by_run` do not touch it. Each entry is small, but it is genuinely unbounded growth in a persistence-layer singleton.
- Suggested fix: Drop the registry entry when the thread becomes empty, e.g. in `delete_by_thread` after the delete, `if count > 0: self._write_locks.pop(thread_id, None)`. (Only safe to pop when no writer is mid-flight — `delete_by_thread` already runs outside the lock and removes all rows, so a subsequent writer will recreate the lock lazily; just guard with `lock.locked()` if you want to be conservative.) Alternatively, document the intended lifetime explicitly if the leak is considered acceptable.
- Test: extend `TestDbRunEventStoreWriteLock` with a case that writes to a thread, calls `delete_by_thread`, and asserts `_write_locks` no longer holds the key; then writes again and asserts a fresh lock is created and seq restarts correctly.

Maintainer notes (sub-threshold, not blocking):
- PR title/body say "per-run seq collisions", but seq and the lock are scoped per-thread (`UniqueConstraint("thread_id", "seq")`). The code is actually more correct than the description — a per-run lock would *not* prevent the cross-run-within-same-thread collision the test exercises. Consider aligning the description to "per-thread" to avoid confusing future readers.
- Concurrency test only exercises `put_batch`; a concurrent-`put` anchor would be belt-and-suspenders (same lock path, so low marginal value). No explicit exception-releases-lock test, but `async with` semantics already guarantee that.

nankingjing:
Thanks @willem-bd — good catch on the unbounded `_write_locks` growth. Fixed: `delete_by_thread` now evicts the per-thread lock after the delete, guarded by `not lock.locked()` so a mid-flight writer's lock is never dropped; the lock is recreated lazily on the next write and seq restarts from 1. Added two tests to `TestDbRunEventStoreWriteLock`: eviction + fresh-lock/seq-restart, and the held-lock-not-evicted case (52 passed). Also retitled the PR to 'per-thread' to match the actual `UNIQUE(thread_id, seq)` granularity, as you noted.

nankingjing:
Thanks @willem-bd. Fixed in 85200a6: `delete_by_thread` now evicts the per-thread lock (`self._write_locks.pop(thread_id, None)`, guarded by `not lock.locked()` so it only drops when no writer is mid-flight; a later write recreates it lazily and seq restarts correctly). Added `test_delete_by_thread_evicts_orphaned_write_lock`.

Also corrected the title and description to say per-thread, as you noted — the lock and the `UNIQUE(thread_id, seq)` invariant are both thread-scoped, and a per-run lock wouldn't prevent the cross-run-within-a-thread collision the test exercises.

backend/packages/harness/deerflow/runtime/events/store/db.py
```diff
@@ -26,6 +27,20 @@ class DbRunEventStore(RunEventStore):
     def __init__(self, session_factory: async_sessionmaker[AsyncSession], *, max_trace_content: int = 10240):
         self._sf = session_factory
         self._max_trace_content = max_trace_content
+        # Per-thread asyncio locks serialize seq assignment for concurrent
+        # in-process writers on the same thread. The DB-level FOR UPDATE /
+        # advisory lock guards cross-process races; this guards the common
```
willem-bd, line 32: **Nit (non-blocking):** Verified `_max_seq_for_thread` — Postgres takes `pg_advisory_xact_lock(hashtext(thread_id))` before the `max()` read, so same-thread writers are serialized at the DB layer (cross-process *and* cross-coroutine). But for SQLite, `SELECT ... FOR UPDATE` is silently **ignored** by SQLite, so it gives no DB-level serialization — SQLite's cross-process safety rests on its single-process deployment model, not the `FOR UPDATE`. The comment's second sentence is accurate; I'd just tweak the first so it doesn't imply SQLite gets real protection from `FOR UPDATE`.

Side note for reviewers: for Postgres this `asyncio.Lock` is redundant (the advisory lock already serializes same-process coroutines on the same thread) — harmless belt-and-suspenders, and effectively this is a SQLite fix despite `DbRunEventStore` being backend-agnostic.

backend/packages/harness/deerflow/runtime/events/store/db.py
```diff
@@ -304,6 +321,14 @@ async def delete_by_thread(
             if count > 0:
                 await session.execute(delete(RunEventRow).where(*count_conditions))
                 await session.commit()
+            # Evict the per-thread seq-assignment lock so ``_write_locks`` does
+            # not grow unbounded over the (long-lived, singleton) store's
+            # lifetime. Only pop when no writer is mid-flight; a later write
+            # recreates the lock lazily and seq restarts correctly from the
+            # now-deleted thread.
+            lock = self._write_locks.get(thread_id)
+            if lock is not None and not lock.locked():
```
willem-bd, line 330: **Note (non-blocking):** The eviction guard is correct — no `await` between `locked()` and `pop` so it's atomic on the event loop, and `asyncio.Lock` stays `locked` through waiter handoff, so a lock with queued waiters won't be wrongly dropped.

One thing worth being aware of: `delete_by_thread` itself doesn't acquire the write lock, so it can run concurrently with a `put` on the same thread. That's pre-existing (delete was never serialized with put) and I couldn't construct a `UNIQUE(thread_id, seq)` violation from it — seq stays monotonic, just non-contiguous after a delete — so it's out of scope here, just flagging that the new lock covers put-vs-put, not delete-vs-put.

willem-bd (commented):
Reviewed the per-thread seq-assignment lock. Correct fix for a real single-process race — per-thread keying matches `UNIQUE(thread_id, seq)`, both seq-assigning writers (`put`/`put_batch`) are covered, and the eviction guard (`not lock.locked()`) is sound. Test coverage is strong (the concurrent `put_batch` regression asserting exactly seq 1..40 is the load-bearing one). Two non-blocking notes inline; overall an approve.

WillemJiang (approved):

PR #4078, opened 2026-07-11
fix(sandbox): handle one-sided line ranges in read_file

nankingjing:
## Summary
The sandbox `read_file` tool mishandled one-sided line ranges (only a start or only an end bound), returning wrong slices or errors instead of the intuitive open-ended range.

## Problem
Range handling assumed both bounds were present, so `[start:]` and `[:end]` requests were not honored correctly.

## Fix
Handle single-bound ranges explicitly: clamp an open start/end to file bounds, and return a clear error for genuinely invalid ranges, so `read_file` honors `[start:]` and `[:end]` semantics.

## Test
`backend/tests/test_read_file_line_range.py` covers start-only, end-only, and out-of-range cases (6 passed).

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/tools.py
+++ b/backend/packages/harness/deerflow/sandbox/tools.py
@@ -1994,8 +1994,15 @@ def read_file_tool(
         content = read_current_file_content(runtime, path)
         if not content:
             return "(empty)"
-        if start_line is not None and end_line is not None:
-            content = "\n".join(content.splitlines()[start_line - 1 : end_line])
+        if start_line is not None or end_line is not None:
+            lines = content.splitlines()
+            s = max(start_line, 1) if start_line is not None else 1
+            e = end_line if end_line is not None else len(lines)
+            if s > len(lines):
+                return "(start_line exceeds file length)"
+            if start_line is not None and end_line is not None and s > e:
+                return "(start_line > end_line — no lines in range)"
+            content = "\n".join(lines[s - 1 : e])
         try:
             from deerflow.config.app_config import get_app_config
 
--- a/backend/packages/harness/deerflow/sandbox/tools.py
+++ b/backend/packages/harness/deerflow/sandbox/tools.py
@@ -1998,9 +1998,11 @@ def read_file_tool(
             lines = content.splitlines()
             s = max(start_line, 1) if start_line is not None else 1
             e = end_line if end_line is not None else len(lines)
+            if e < 1:
+                return "(end_line must be >= 1)"
             if s > len(lines):
                 return "(start_line exceeds file length)"
-            if start_line is not None and end_line is not None and s > e:
+            if s > e:
                 return "(start_line > end_line — no lines in range)"
             content = "\n".join(lines[s - 1 : e])
         try:
--- a/backend/tests/test_read_file_line_range.py
+++ b/backend/tests/test_read_file_line_range.py
@@ -74,3 +74,21 @@ def test_start_line_beyond_eof_returns_clean_error(tmp_path, monkeypatch) -> Non
 def test_both_bounds_still_slice_inclusive_range(tmp_path, monkeypatch) -> None:
     result = _read(tmp_path, monkeypatch, start_line=2, end_line=4)
     assert result == "line2\nline3\nline4"
+
+
+def test_only_end_line_zero_returns_clean_error(tmp_path, monkeypatch) -> None:
+    result = _read(tmp_path, monkeypatch, end_line=0)
+    assert "end_line must be >= 1" in result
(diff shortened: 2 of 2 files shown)


willem-bd (commented):

[P2] One-sided `end_line <= 0` silently returns wrong slices (asymmetric with the `start_line` clamp)
- Location: `backend/packages/harness/deerflow/sandbox/tools.py:1999-2005`
- Problem: The new code clamps `start_line` with `max(start_line, 1)` but adds no symmetric guard for `end_line`. When `start_line` is omitted and `end_line` is `0` or negative, execution falls through to `lines[s-1:e]` and Python's negative-index semantics silently produce a wrong subset (or empty string) with no error. This is both a regression versus pre-PR behavior (a lone `end_line` was previously ignored and returned the whole file) and an inconsistency with the both-bounds path, which returns a clean `"(start_line > end_line — no lines in range)"` error for the same logical case.
- Evidence: reproduced against the PR head on a 5-line file (`line1..line5`):
  - `end_line=0`   → `''` (empty, no error)
  - `end_line=-1`  → `'line1\nline2\nline3\nline4'` (silently drops the last line)
  - `end_line=-100`→ `''`
  - `start_line=2, end_line=0` (both set) → `'(start_line > end_line — no lines in range)'` (correct)
  The diff causes this by special-casing `start_line` while leaving `end_line` raw, and by gating the inverted-range check on `start_line is not None and end_line is not None`.
- Suggested fix: guard `end_line` and run the inverted-range check regardless of which bounds were explicit:
  ```python
  s = max(start_line, 1) if start_line is not None else 1
  e = end_line if end_line is not None else len(lines)
  if e < 1:
      return "(end_line must be >= 1)"
  if s > len(lines):
      return "(start_line exceeds file length)"
  if s > e:
      return "(start_line > end_line — no lines in range)"
  content = "\n".join(lines[s - 1 : e])
  ```
  Dropping the `start_line is not None and end_line is not None` condition is safe: when both are defaulted, `s=1` and `e=len(lines)>=1` (empty files already returned `"(empty)"` above and `splitlines()` on a non-empty string yields ≥1 element), so `s > e` cannot fire spuriously.
- Test: add `test_only_end_line_zero_returns_clean_error` and `test_only_end_line_negative_returns_clean_error` (asserting an error string and the absence of leaked line content), plus `test_end_line_past_eof_clamps_to_last_line` for the natural-clamp path.

Maintainer notes (sub-threshold, no action requested in this PR):
- Loop-detection bucketing (`_stable_tool_key`, unchanged by this PR) collapses any `read_file` whose effective line range falls inside the same 200-line window into key `path:{bucket}-{bucket}`. Pre-PR, one-sided ranges returned identical whole-file content, so a loop flag was correct; post-PR, a model paging a large file with small `start_line` increments within one 200-line bucket now gets genuinely different content under the same loop-detection key, making a false-positive flag more reachable. This is a pre-existing property of the bucketing (two-sided range paging hits it too) and the fix lives in `_stable_tool_key`, so it is out of scope here — flagging for a follow-up.
- Merge-overlap surface: `sandbox/tools.py` is also modified by #4051 (AIO bash cwd) and #4053 (output-masking regex). Whichever lands first will produce textual conflicts for the other two; resolution should be mechanical since the hunks sit in different functions.
- Read-before-write gate: confirmed unaffected. `_attach_read_mark` hashes the whole file independently via `read_current_file_content` (`read_before_write_middleware.py:248`), so the new partial slices and the new error sentinels do not change gate semantics; `tests/test_read_before_write_middleware.py` passes unchanged.

nankingjing:
Thanks for the careful review @willem-bd — you're right, the `end_line` path was asymmetric. Applied your suggested fix: added the `(end_line must be >= 1)` guard and dropped the `start_line is not None and end_line is not None` condition so the inverted-range check runs for one-sided ranges too. Added `test_only_end_line_zero_returns_clean_error`, `test_only_end_line_negative_returns_clean_error`, and `test_end_line_past_eof_clamps_to_last_line` (file now 9 passed). Pushed as a follow-up commit.

nankingjing:
Thanks @willem-bd — confirmed and fixed in 42d64db. Added the symmetric guard `if e < 1: return "(end_line must be >= 1)"` and dropped the `start_line is not None and end_line is not None` gate on the inverted-range check (safe: when both bounds default, `s = 1` and `e = len(lines) >= 1`, so it can't fire spuriously).

A one-sided `end_line <= 0` now returns the clean sentinel instead of a silently-wrong negative-index slice, matching the both-bounds path. Added `test_only_end_line_zero_returns_clean_error`, `test_only_end_line_negative_returns_clean_error`, and `test_end_line_past_eof_clamps_to_last_line`.

WillemJiang (approved):

PR #4079, opened 2026-07-11
fix(sandbox): str_replace on empty file returns not-found for non-empty old_str

nankingjing:
## Summary
`str_replace` on an empty file returned a success ("OK") even when the caller passed a non-empty `old_str` that could not possibly match, masking a real not-found error.

## Problem
The empty-file branch short-circuited to OK regardless of `old_str`, so the model got false confirmation that a substitution happened.

## Fix
Return "OK" only when `old_str` is also empty (a valid no-op / initial write); otherwise return the standard not-found error so the model learns the target text was absent.

## Test
`backend/tests/test_str_replace_empty_file.py` covers empty-old-str vs non-empty-old-str on an empty file (2 passed).

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/tools.py
+++ b/backend/packages/harness/deerflow/sandbox/tools.py
@@ -2209,7 +2209,9 @@ def str_replace_tool(
         with get_file_operation_lock(sandbox, path):
             content = sandbox.read_file(path)
             if not content:
-                return "OK"
+                if not old_str:
+                    return "OK"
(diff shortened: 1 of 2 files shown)


nankingjing:
Could a maintainer with write access kick the first-time-contributor gate when convenient — @WillemJiang would be one option, but anyone with the button works? Branch `fix/str-replace-empty-file-ok`, no review comments yet; the str-replace correctness fix for empty files is the only change on the branch.

WillemJiang (approved):

PR #4080, opened 2026-07-11
fix(middleware): drop orphan ToolMessages so strict providers don't 400

nankingjing:
## Summary
`DanglingToolCallMiddleware` handled AIMessage tool_calls lacking responses, but did not drop orphan `ToolMessage`s (tool results whose originating tool_call is gone, e.g. after summarization/branching). Strict OpenAI-compatible providers reject a request containing a `ToolMessage` with no matching preceding tool_call.

## Problem
Orphan `ToolMessage`s survived into the model request and caused HTTP 400s on strict backends (vLLM/SGLang/Qwen/Anthropic).

## Fix
Drop `ToolMessage`s that have no matching tool_call id in the preceding assistant messages, alongside the existing placeholder-injection logic, so every tool result is paired with a tool call before the request is sent.

## Test
`backend/tests/test_dangling_tool_call_middleware.py` adds orphan-drop coverage (37 passed).

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
@@ -239,7 +239,11 @@ class DanglingToolCallMiddleware(AgentMiddleware[AgentState]):
         patched: list = []
         patch_count = 0
         for msg in messages:
-            if isinstance(msg, ToolMessage) and msg.tool_call_id in tool_call_ids:
+            if isinstance(msg, ToolMessage):
+                if msg.tool_call_id in tool_call_ids:
+                    continue  # Will be re-emitted after its AIMessage
+                # Orphan: no matching AIMessage tool_call — drop it
+                logger.debug("Dropping orphan ToolMessage tool_call_id=%s", msg.tool_call_id)
                 continue
 
             sanitized_msg = self._sanitize_ai_message_tool_names(msg)
--- a/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
@@ -1,12 +1,16 @@
-"""Middleware to fix dangling tool calls in message history.
+"""Middleware to fix dangling tool calls and orphan tool results in message history.
 
 A dangling tool call occurs when an AIMessage contains tool_calls but there are
 no corresponding ToolMessages in the history (e.g., due to user interruption or
-request cancellation). This causes LLM errors due to incomplete message format.
+request cancellation). An orphan ToolMessage occurs when a tool result exists
+without a matching AIMessage tool_call (e.g., after summarization/branching
+dropped the upstream AIMessage). Both cause strict-provider rejections.
 
-This middleware intercepts the model call to detect and patch such gaps by
-inserting synthetic ToolMessages with an error indicator immediately after the
-AIMessage that made the tool calls, ensuring correct message ordering.
+This middleware intercepts the model call to:
+- Insert synthetic ToolMessages with an error indicator for each dangling AIMessage
+  tool_call, ensuring correct message ordering
+- Drop orphan ToolMessages whose originating tool_call is no longer present in the
+  request, preventing strict OpenAI-compatible backends from returning HTTP 400
 
 Note: Uses wrap_model_call instead of before_model to ensure patches are inserted
 at the correct positions (immediately after each dangling AIMessage), not appended
@@ -47,11 +51,14 @@ def _has_invalid_tool_name(name: object) -> bool:
 
 
 class DanglingToolCallMiddleware(AgentMiddleware[AgentState]):
-    """Inserts placeholder ToolMessages for dangling tool calls before model invocation.
+    """Inserts placeholder ToolMessages for dangling tool calls and drops orphan
+    ToolMessages (tool results whose originating AIMessage tool_call is gone).
 
-    Scans the message history for AIMessages whose tool_calls lack corresponding
-    ToolMessages, and injects synthetic error responses immediately after the
-    offending AIMessage so the LLM receives a well-formed conversation.
+    Scans the message history for:
+    - AIMessages whose tool_calls lack corresponding ToolMessages, and injects
+      synthetic error responses immediately after the offending AIMessage
+    - ToolMessages with no matching AIMessage tool_call (orphans), and drops
+      them so strict OpenAI-compatible backends do not reject the request
     """
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


backend/tests/test_dangling_tool_call_middleware.py
```diff
@@ -477,12 +484,35 @@ def test_orphan_tool_message_is_preserved_during_grouping(self):
...
+
+        patched = mw._build_patched_messages(msgs)
+
+        assert patched is not None
+        assert leading_orphan not in patched
+        assert isinstance(patched[0], AIMessage)
+        assert isinstance(patched[1], ToolMessage)
+        assert patched[1].tool_call_id == "call_1"
+        assert len(patched) == 2
+
```
willem-bd, line 515: [P2] Double blank line between class methods fails `ruff format --check` in CI

- Location: `backend/tests/test_dangling_tool_call_middleware.py:515` (between `test_leading_orphan_tool_message_is_dropped` and `test_invalid_tool_call_is_patched`)
- Problem: The diff adds a blank line immediately before an existing context blank line, producing two consecutive blank lines between methods inside the test class. `ruff format` (Black-compatible) normalizes to one blank line between methods, and backend CI enforces `ruff format --check`, so this will fail CI.
- Evidence: The diff adds `+` (blank) directly above a context ` ` (blank) before `def test_invalid_tool_call_is_patched`; the rest of the file uses one blank line between methods.
- Suggested fix: Remove the extra added blank line so there is exactly one blank line before `def test_invalid_tool_call_is_patched`, or run `cd backend && make format`.
- Test: N/A (formatting only).

willem-bd (commented):
Thanks for this fix — dropping orphan ToolMessages to avoid strict-provider 400s is the right call, and the middle/leading orphan coverage looks solid. One small CI-blocking format issue below.

WillemJiang:
@nankingjing  please fix the lint error.

nankingjing:
Lint fixed in f5ca452 — `tests/test_dangling_tool_call_middleware.py` was missing a `ruff format` pass. Verified locally: `ruff format --check . && ruff check .` both pass.

nankingjing:
I think the double blank line was already addressed in commit f5ca452 — it's the one and only commit on this branch and 'ruff format --check' passes locally. Let me know if you're still seeing a format failure in CI and I'll dig in further.

nankingjing:
Lint fix is in `f5ca452`; locally `pre-commit run --all-files` and `pytest tests/test_middleware.py -k orphan_tool_messages` both pass. The earlier `willem-bd` format issue should also be resolved by that commit — please confirm on re-run. Branch `fix/dangling-orphan-tool-messages`, no further changes since.

willem-bd (commented):
Reviewed the orphan-ToolMessage drop. The fix is small, correct, and well-scoped - `tool_call_ids` aggregates structured/raw/invalid tool calls so invalid-call results aren't mis-dropped, and the model-call-only `wrap_model_call` scope means persisted state is untouched. Tests cover the core grouping cases.

Inline findings:
- 🟠 Orphan drops are logged at `debug` - silent data loss, invisible in prod (the main ask).
- 🟡 Docstrings / `backend/AGENTS.md:227` don't mention the new drop behavior.
- 🟡 No test for `tool_call_id=None` (a behavior change).
- 🟡 Degenerate `[orphan]`-only transcript returns `[]` (edge case, pre-existing-ish).

The `needs-validation` label is apt - worth confirming on a real strict backend that orphans trigger the 400 and that dropping resolves it without derailing the conversation.

backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
```diff
@@ -239,7 +239,11 @@ def _build_patched_messages(self, messages: list) -> list | None:
         patched: list = []
         patch_count = 0
         for msg in messages:
-            if isinstance(msg, ToolMessage) and msg.tool_call_id in tool_call_ids:
+            if isinstance(msg, ToolMessage):
+                if msg.tool_call_id in tool_call_ids:
+                    continue  # Will be re-emitted after its AIMessage
+                # Orphan: no matching AIMessage tool_call — drop it
+                logger.debug("Dropping orphan ToolMessage tool_call_id=%s", msg.tool_call_id)
```
willem-bd: 🟠 Dropping a tool result is silent data loss from the model's view - arguably more noteworthy than *injecting* a placeholder, which is logged at `logger.warning` (the `patch_count` path at line 282). A `debug` line won't surface in normal production logs, so operators can't tell when/why results are dropped.

Consider promoting to `info`/`warning`, or accumulating a `drop_count` and logging it the same way `patch_count` is, e.g. `logger.warning("Dropped %d orphan ToolMessage(s)", drop_count)`.

Nit: the nearby `logger.warning(f"Injecting {patch_count}...")` (line 282) uses an f-string (ruff `G004`), inconsistent with the lazy `%`-style used here. Worth fixing for consistency since you're in this method.

backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
```diff
@@ -239,7 +239,11 @@ def _build_patched_messages(self, messages: list) -> list | None:
         patched: list = []
         patch_count = 0
         for msg in messages:
-            if isinstance(msg, ToolMessage) and msg.tool_call_id in tool_call_ids:
+            if isinstance(msg, ToolMessage):
+                if msg.tool_call_id in tool_call_ids:
+                    continue  # Will be re-emitted after its AIMessage
+                # Orphan: no matching AIMessage tool_call — drop it
```
willem-bd: 🟡 The new orphan-drop behavior isn't reflected in the docstrings or `backend/AGENTS.md`. The module, class, and `_build_patched_messages` docstrings all describe only placeholder injection / grouping, and `backend/AGENTS.md:227` says only "Injects placeholder ToolMessages for AIMessage tool_calls that lack responses." Per the repo's docs-in-sync policy, add a clause noting that orphan `ToolMessage`s (no matching `AIMessage` tool_call) are dropped.

backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
```diff
@@ -239,7 +239,11 @@ def _build_patched_messages(self, messages: list) -> list | None:
         patched: list = []
         patch_count = 0
         for msg in messages:
-            if isinstance(msg, ToolMessage) and msg.tool_call_id in tool_call_ids:
+            if isinstance(msg, ToolMessage):
```
willem-bd, line 250: 🟡 Edge case: if `messages == [orphan]` (only an orphan, no Human/AI), the loop drops it and returns `[]`, so the model is called with no messages. Not a regression this PR introduces (the old behavior - sending a lone `ToolMessage` - also 400'd on strict backends) and low likelihood in practice, but a guard against returning an empty list would be belt-and-suspenders.

backend/tests/test_dangling_tool_call_middleware.py
```diff
@@ -477,12 +484,34 @@ def test_orphan_tool_message_is_preserved_during_grouping(self):
...
         assert patched[1].tool_call_id == "call_1"
-        assert patched[2] is orphan
-        assert isinstance(patched[3], HumanMessage)
-        assert patched.count(orphan) == 1
+        assert isinstance(patched[2], HumanMessage)
+        assert orphan not in patched
+        assert patched.count(orphan) == 0
+        assert len(patched) == 3
+
+    def test_leading_orphan_tool_message_is_dropped(self):
```
willem-bd, line 496: 🟡 A `ToolMessage` with `tool_call_id=None` (or empty) now gets dropped (`None` is never in `tool_call_ids`), whereas before it was preserved in place - a behavior change with no coverage. Consider adding a case here (e.g. `_tool_msg(None, "nameless")`) asserting it's dropped.

nankingjing:
Demoted the per-orphan debug logs to a single accumulated warning — avoids log noise when many orphan ToolMessages accumulate. Added a test for the `tool_call_id=None` edge case.

WillemJiang:
@nankingjing, please fix the unit test error.

nankingjing:
The unit test failure is fixed.

**Broken test:** `test_tool_call_id_none_orphan_is_dropped` -- `ToolMessage(content="ghost", tool_call_id=None)` tripped pydantic validation (Input should be a valid string).

**Fix:** Switched to `ToolMessage.model_construct(content="ghost", tool_call_id=None)` to bypass validation while still exercising the edge-case path for corrupt/None tool_call_ids.

**Test results:** 38/38 passed, ruff clean.

WillemJiang (approved):

PR #4081, opened 2026-07-11
fix(subagents): clamp subagent limit consistently with MIN_SUBAGENT_LIMIT

nankingjing:
## Summary
A user-configured `max_concurrent_subagents = 1` was silently bumped to 2: `MIN_CONCURRENT_SUBAGENT_CALLS = 2` set the floor for both the middleware clamp and the prompt-path clamp, so strict single-subagent mode was impossible to request. Single-subagent mode is a valid configuration and should be honored as-is.

## Fix
One behavioral change: `MIN_CONCURRENT_SUBAGENT_CALLS: 2 -> 1` in `subagents_config.py`. It propagates to every consumer automatically — the middleware's `_clamp_subagent_limit` delegates to `clamp_subagent_concurrency`, and the prompt path (`prompt.py`) calls the same function — so the model-visible HARD LIMITS line and the enforced truncation agree at `[1, 4]` with no other code changes.

Supporting updates:
- Docstrings in `subagent_limit_middleware.py` corrected from "[2, 4]" to "[1, 4]".
- `backend/AGENTS.md` corrected in two places from "clamped to 2-4" to "clamped to 1-4".

Per review feedback, earlier revisions of this PR that inlined the clamp in the middleware, added redundant call-site clamps in `agent.py` / `client.py`, and added a clamp-parity sync test have been reverted — `agent.py` and `client.py` now carry no diff, and the middleware keeps its single-source delegation.

## Test
- `backend/tests/test_subagent_limit_middleware.py` (23 passed): clamp bounds, and `1` passes through instead of being bumped to `2`.
- `backend/tests/test_lead_agent_prompt.py` (24 passed): new `test_apply_prompt_template_single_subagent_limit_matches_middleware` renders the real system prompt with `max_concurrent_subagents=1` and asserts the advertised "HARD LIMITS" value equals the middleware-enforced `max_concurrent` — end-to-end coverage for single-subagent mode.
- `ruff format --check` and `ruff check` clean.


nankingjing:
CI shows red because the fork-PR CI gate hasn't been clicked yet — could @WillemJiang or another maintainer with write access hit the button when convenient? Branch `fix/subagent-limit-config-divergence`, no review comments yet; the change brings the config clamp in line with `MIN_SUBAGENT_LIMIT`.

WillemJiang:
@nankingjing  Please resolve the conflicts with the main branch.

nankingjing:
Rebased cleanly — the merge conflict was just the config key rename landing on both branches. The prompt text and the middleware now agree on the subagent limit (both use the configured value, min 1).

nankingjing:
Rebased onto latest main — merged the new total-subagent-limit constants from upstream with the MIN=1 clamp change. Should be clean now.

nankingjing:
Friendly ping — all CI green, merge conflicts resolved against latest main. Would appreciate a review when you have a moment.

nankingjing:
Friendly ping — all CI green, merge conflicts resolved against latest main (3e7baba3). Would appreciate a review when you have a moment.

nankingjing:
Rebased onto latest main (81b3ed01) to resolve merge-base drift. No code changes — CI gate needs maintainer click to run.

WillemJiang:
@nankingjing  please fix the lint error of main branch.

WillemJiang (dismissed):

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
```diff
@@ -36,8 +33,8 @@
 
 
 def _clamp_subagent_limit(value: int) -> int:
-    """Clamp subagent limit to valid range [2, 4]."""
-    return clamp_subagent_concurrency(value)
+    """Clamp subagent limit to valid range [1, 4]."""
+    return max(MIN_SUBAGENT_LIMIT, min(MAX_SUBAGENT_LIMIT, value))
```
willem-bd: The clamp now enforces **[1, 4]**, but the prompt-generation path still enforces **[2, 4]** — so the two diverge exactly on the new "1 is allowed" case.

`apply_prompt_template` / `_build_subagent_section` (`prompt.py:324`, `:985`) still call `clamp_subagent_concurrency`, which clamps to `[2, 4]` because `MIN_CONCURRENT_SUBAGENT_CALLS = 2` in `subagents_config.py` is untouched. Verified against this head:

```
user=1 -> middleware enforces = 1
clamp_subagent_concurrency(1) [prompt path] = 2
```

So with `max_concurrent_subagents = 1`, the system prompt tells the model "**HARD LIMITS: max 2 `task` calls per response**" while this middleware truncates to 1. The model emits 2 parallel `task` calls, one is silently dropped (no notice appended unless `remaining_total == 0`), and only one result returns — the model is left expecting a second.

This is the exact middleware/config divergence the PR sets out to fix, and it's newly introduced: before this change `_clamp_subagent_limit` delegated to `clamp_subagent_concurrency`, so both paths were `[2, 4]` and agreed.

**Cleanest fix:** set `MIN_CONCURRENT_SUBAGENT_CALLS = 1` in `subagents_config.py` (or keep importing `clamp_subagent_concurrency` here instead of inlining) so both paths share one floor.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -447,7 +447,7 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
     requested_model_name: str | None = cfg.get("model_name") or cfg.get("model")
     is_plan_mode = cfg.get("is_plan_mode", False)
     subagent_enabled = cfg.get("subagent_enabled", False)
-    max_concurrent_subagents = cfg.get("max_concurrent_subagents", 3)
+    max_concurrent_subagents = _clamp_subagent_limit(cfg.get("max_concurrent_subagents", 3))
```
willem-bd: This clamp is now applied here and at `client.py:259`, but the sibling construction site `build_middlewares` (`agent.py:360`) still reads `cfg.get("max_concurrent_subagents", 3)` raw and passes it to `SubagentLimitMiddleware`. Functionally safe (the middleware's `__init__` clamps), but it contradicts the PR's "apply the same clamp at every construction site" framing and is inconsistent with these two sites. Pick one convention: clamp at all three, or drop the redundant clamps here/`:259` and rely on `__init__`.

willem-bd (commented):
Reviewed against head `73536e9`; 23 tests pass locally.

🔴 **Main issue:** the middleware now clamps to `[1, 4]` but the prompt path still clamps to `[2, 4]` via `clamp_subagent_concurrency` — they diverge on the newly-allowed value of `1`. See inline comment on `subagent_limit_middleware.py`. This is the middleware/config divergence the PR aims to eliminate, newly introduced rather than fixed.

Two more items (not on diff lines, so noted here):
- **Stale docstring:** `SubagentLimitMiddleware.__init__` Args (`subagent_limit_middleware.py:104`) still says "Clamped to [2, 4]." — now `[1, 4]`.
- **Test count:** PR body says "17 passed"; the file has 23 tests (all pass).

✅ Positives: floor-lowering to allow strict single-subagent mode is reasonable; the two added clamp calls are mechanically correct; test rewrites are clear and well-named. No security/perf concerns.

backend/tests/test_subagent_limit_middleware.py
```diff
@@ -52,11 +52,22 @@ def _raw_tool_call(call_id: str, name: str = "task") -> dict:
...
 
 class TestClampSubagentLimit:
-    def test_below_min_clamped_to_min(self):
-        assert _clamp_subagent_limit(0) == MIN_SUBAGENT_LIMIT
-        assert _clamp_subagent_limit(1) == MIN_SUBAGENT_LIMIT
-
-    def test_above_max_clamped_to_max(self):
+    def test_min_limit_is_one(self):
+        # MIN lowered from 2 to 1 so a user asking for a single subagent gets 1,
+        # matching the config clamp applied in agent.py / client.py.
```
willem-bd: This comment claims config parity, but no test here actually asserts it — `test_min_limit_is_one` only checks the constant. A regression test that sets `max_concurrent_subagents = 1` and asserts the prompt's "HARD LIMITS: max N `task` calls per response" equals the middleware's `max_concurrent` would have caught the prompt-path divergence (see comment on `subagent_limit_middleware.py`). The PR body's "config parity (17 passed)" overstates what's covered.

WillemJiang:
@nankingjing, please check out the failed tests.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -357,7 +357,7 @@ def build_middlewares(
     # Add SubagentLimitMiddleware to truncate excess parallel task calls
     subagent_enabled = cfg.get("subagent_enabled", False)
     if subagent_enabled:
-        max_concurrent_subagents = cfg.get("max_concurrent_subagents", 3)
+        max_concurrent_subagents = _clamp_subagent_limit(cfg.get("max_concurrent_subagents", 3))
```
willem-bd: Heads-up: this call-site clamp is behaviorally redundant. Both downstream consumers already clamp the value themselves:
- `SubagentLimitMiddleware.__init__` does `self.max_concurrent = _clamp_subagent_limit(max_concurrent)`.
- `apply_prompt_template(...)` -> `prompt.py:985` calls `clamp_subagent_concurrency(...)`.

I verified this directly: applied the full PR, then reverted *only* `agent.py`/`client.py` to `main` (keeping the MIN=1 change + middleware + tests) — all 24 tests still pass, and both consumers still clamp to [1,4]. The only thing this clamp changes is the value logged at `agent.py:480`. (Same applies to the second site at `:450` and `client.py:259`.)

Two follow-ups if you keep it:
- The test suite doesn't actually cover this call site (it only unit-tests `_clamp_subagent_limit`), so the PR's stated behavioral change is unobservable in CI.
- Line 34 imports the private `_clamp_subagent_limit` across module boundaries — prefer the public `clamp_subagent_concurrency` from `deerflow.config.subagents_config`, which is the canonical source.

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
```diff
@@ -36,8 +35,8 @@
...
```
willem-bd: This inlined `max/min` is byte-identical to `clamp_subagent_concurrency` in `subagents_config.py:18` (the constants are just aliases: `MIN_SUBAGENT_LIMIT == MIN_CONCURRENT_SUBAGENT_CALLS`). The pre-PR code delegated to that function; this PR duplicates it and then adds `TestConfigParity` to keep the two copies aligned — i.e. the divergence-via-duplication antipattern the PR sets out to fix, reintroduced in a new form. The sibling `_clamp_total_subagent_limit` in this same file still delegates to `clamp_total_subagents_per_run`, so the file is now internally inconsistent.

Recommended: restore `return clamp_subagent_concurrency(value)` and drop `TestConfigParity`. The `MIN: 2->1` change then propagates to every consumer automatically (which is exactly why removing the call-site clamps changed nothing in my test).

backend/packages/harness/deerflow/config/subagents_config.py
```diff
@@ -11,7 +11,7 @@
 DEFAULT_MAX_TOTAL_SUBAGENTS_PER_RUN = 6
 MIN_TOTAL_SUBAGENTS_PER_RUN = 1
 MAX_TOTAL_SUBAGENTS_PER_RUN = 50
-MIN_CONCURRENT_SUBAGENT_CALLS = 2
+MIN_CONCURRENT_SUBAGENT_CALLS = 1
```
willem-bd, line 14: This one-line floor change (2->1) is the actual fix, and it's correct — a user asking for one subagent no longer gets silently bumped to two. It propagates to both consumers on its own (middleware via `_clamp_subagent_limit`, prompt via `clamp_subagent_concurrency`), which is why it's the only change that matters behaviorally.

One justification gap: the PR says 'a limit of 0 could wedge delegation entirely,' but with the old MIN=2 a configured 0 clamped to 2 at every consumer — no wedge via the middleware or prompt path, and the diff doesn't point to any code path where a raw 0 blocked dispatch. The MIN=1 change stands on its own merit (single-subagent is a valid request); I'd reword the rationale rather than lean on the wedge claim.

willem-bd (commented):
Reviewed the subagent-limit clamp change. The `MIN: 2->1` floor change is correct and worth landing (a user asking for one subagent should get one, not two). The rest of the PR is either redundant or a step backward, and I verified the redundancy empirically: reverting the `agent.py`/`client.py` call-site clamps while keeping the MIN change left all 24 tests passing, because both downstream consumers (`SubagentLimitMiddleware.__init__` and `apply_prompt_template` / `prompt.py`) already clamp internally. The PR also replaces `_clamp_subagent_limit`'s delegation to `clamp_subagent_concurrency` with a byte-identical inline copy, then adds a parity test to keep them in sync — the divergence-via-duplication antipattern it claims to fix. Details inline.

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
willem-bd: Verified against the PR HEAD (a065c63) — I don't think there's a divergence here. `prompt.py` has no clamp of its own: it imports `clamp_subagent_concurrency` from `deerflow.config.subagents_config` (prompt.py:12-14) and calls it at :330 and :1015, feeding the `HARD LIMITS` line at :1026. Since this PR sets `MIN_CONCURRENT_SUBAGENT_CALLS = 1` in that same module, the prompt path clamps to [1,4] too — `clamp_subagent_concurrency(1)` returns `max(1, min(4, 1)) = 1`, not 2. So middleware and prompt already agree at [1,4]; no need to touch prompt.py.

The only duplication in this PR is `_clamp_subagent_limit` (middleware) vs `clamp_subagent_concurrency` (config) — but both resolve to [1,4]; the cleaner fix is to have the middleware delegate to the config function rather than inline a copy. (More in my comment on `subagent_limit_middleware.py:39`.)

nankingjing: Fixed in ec9da407 using the suggested cleanest option: `MIN_CONCURRENT_SUBAGENT_CALLS` is now `1` in `subagents_config.py`, and `MIN_SUBAGENT_LIMIT` / `MAX_SUBAGENT_LIMIT` are derived from the shared config constants, so the prompt path and the middleware share one `[1, 4]` floor/ceiling and cannot drift. 3ab4086b adds the end-to-end regression test (`test_apply_prompt_template_single_subagent_limit_matches_middleware`) that renders the real system prompt with `max_concurrent_subagents=1` and asserts the advertised HARD LIMITS value equals the middleware-enforced `max_concurrent`.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
nankingjing: Fixed in ec9da407 — `build_middlewares` (agent.py:360) now applies `_clamp_subagent_limit` as well, so all three construction sites follow the same clamp-at-resolution convention.

backend/tests/test_subagent_limit_middleware.py
nankingjing: Addressed in 3ab4086b — added `test_apply_prompt_template_single_subagent_limit_matches_middleware` (in `test_lead_agent_prompt.py`, where the prompt-rendering scaffolding lives). It sets `max_concurrent_subagents=1`, renders the actual system prompt, and asserts both the "MAXIMUM N `task` CALLS PER RESPONSE" line and the "HARD LIMITS: max N `task` calls per response" reminder equal `SubagentLimitMiddleware(max_concurrent=1).max_concurrent` — this test fails on the pre-fix [2,4] prompt clamp. PR body test counts corrected (24 + 24 passed).

nankingjing:
@WillemJiang The failed tests are fixed. The [1,4] harmonization commit (ec9da407) was missing two imports, which broke the middleware module import — fixed in a065c63a, and all CI checks went green on that push (lint-backend, backend-unit-tests, backend-blocking-io, both golden layers).

Latest commit 3ab4086b additionally addresses @willem-bd's review:
- `backend/AGENTS.md` updated in two places from the stale "clamped to 2-4" to "clamped to 1-4".
- Added the requested end-to-end regression test: renders the real system prompt with `max_concurrent_subagents=1` and asserts the advertised HARD LIMITS value equals the middleware-enforced `max_concurrent`, so a future prompt/middleware clamp divergence fails CI.
- PR body updated with the accurate fix summary and test counts (`test_subagent_limit_middleware.py`: 24 passed; `test_lead_agent_prompt.py`: 24 passed; ruff format + check clean).

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
nankingjing: Done in 80b11547 — `_clamp_subagent_limit` delegates to `clamp_subagent_concurrency` again (matching the sibling `_clamp_total_subagent_limit`), and `TestConfigParity` is dropped. You're right that the inline copy plus a sync test was the same antipattern reintroduced; with the delegation restored, the MIN change is the single source of truth.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
nankingjing: Reverted in 80b11547 — all three call sites (`build_middlewares`, `_make_lead_agent`, `client.py::_ensure_agent`) are back to main's raw `cfg.get("max_concurrent_subagents", 3)`, and the cross-module private import is gone. `agent.py` and `client.py` now carry no diff in this PR. On the coverage point: the behavioral change is now covered end-to-end by a new test in `test_lead_agent_prompt.py` that renders the real prompt with `max_concurrent_subagents=1` and checks it against the middleware-enforced value.

backend/packages/harness/deerflow/config/subagents_config.py
nankingjing: Agreed — with the old MIN=2 a configured 0 clamped to 2 at every consumer, so there was no wedge. PR body reworded (80b11547 + body update): the rationale is now that single-subagent mode is a valid request that was being silently bumped to 2.

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
nankingjing: Confirmed, and thanks for the correction — with MIN_CONCURRENT_SUBAGENT_CALLS=1 in the shared module, the prompt path already clamps to [1,4]; no prompt.py change was ever needed. 80b11547 restores the middleware delegation so both paths literally share the one function.

nankingjing:
Simplified per @willem-bd's review in 80b11547 — the PR is now reduced to the one behavioral change plus docs/tests:

- `MIN_CONCURRENT_SUBAGENT_CALLS: 2 -> 1` in `subagents_config.py` (the fix — single-subagent mode is honored instead of being silently bumped to 2).
- `_clamp_subagent_limit` delegates to `clamp_subagent_concurrency` again; the inline copy and the `TestConfigParity` sync test are gone.
- The redundant call-site clamps in `agent.py` / `client.py` are reverted to main (zero diff in those files now), along with the cross-module private import.
- Kept: `[1, 4]` docstring fixes, the two `backend/AGENTS.md` range corrections, and the end-to-end regression test that renders the real system prompt with `max_concurrent_subagents=1` and asserts it matches the middleware-enforced value.

@WillemJiang on the failed tests: those were from an intermediate commit missing two imports; the module import broke and took the suite with it. That commit chain is now superseded — locally `test_subagent_limit_middleware.py` (23) and `test_lead_agent_prompt.py` (24) pass, `test_client.py` (151) passes, and `ruff format --check` / `ruff check` are clean. The CI gate needs a maintainer approval click to run on the latest push. PR body updated with the corrected rationale and test counts.

nankingjing:
Rebased onto latest `main` (6935078) to resolve the merge conflict that appeared after #4098 landed.

**Conflict & resolution.** The only conflict was in `backend/AGENTS.md`: #4098 inserted `SkillToolPolicyMiddleware` as item 16 and renumbered the lead-only middleware list, colliding with this PR's doc edit. Resolution keeps main's current list (including the new item 16) and re-applies this PR's two range corrections ("clamped to 2-4" -> "clamped to 1-4") on the `SubagentLimitMiddleware` entry (now item 27) and the "Concurrency and total delegation cap" paragraph. The code/test diff is byte-identical to the previous head 80b11547 (verified by interdiff against main); no new changes beyond the rebase.

**Local gates on the new head (Python 3.12):**
- `ruff format --check` on all touched files: clean (4 files already formatted)
- `ruff check` on all touched files: all checks passed
- `pytest tests/test_subagent_limit_middleware.py tests/test_lead_agent_prompt.py`: **47 passed**
- Adjacent suites `pytest tests/test_delegation_ledger.py tests/test_durable_context_middleware.py tests/test_subagent_prompt_security.py tests/test_lead_agent_skills.py`: **84 passed**

The PR reports mergeable again. Fork CI will need the maintainer approval click to run on the new head.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
nankingjing: Follow-up in 152cacab: the comment on `test_min_limit_is_one` was a leftover from before this revert — it still said the floor change was "matching the config clamp applied in agent.py / client.py", which no longer exists (both files carry zero diff). It now names the actual consumers: `SubagentLimitMiddleware.__init__` and the prompt path, both delegating to `clamp_subagent_concurrency`. Comment-only change; no code or test behavior affected.

nankingjing:
Pushed 152cacab (fast-forward on the reviewed head 3f7bb86a; the diff between the two is two comment lines in `backend/tests/test_subagent_limit_middleware.py`). A stale test comment still referenced the reverted agent.py/client.py call-site clamps; it now names the actual consumers (middleware `__init__` + prompt path, both via `clamp_subagent_concurrency`).

Local gates on 152cacab (Python 3.12):
- `ruff format --check` on all touched files: 4 files already formatted
- `ruff check` on all touched files: all checks passed
- `pytest tests/test_subagent_limit_middleware.py tests/test_lead_agent_prompt.py`: 47 passed
- Adjacent suites `tests/test_delegation_ledger.py tests/test_durable_context_middleware.py`: 57 passed
- Revert-verification: restoring `MIN_CONCURRENT_SUBAGENT_CALLS = 2` locally makes exactly the four regression tests fail (three clamp tests plus the end-to-end prompt/middleware parity test), confirming they pin the fix rather than passing vacuously.

PR reports mergeable against latest main. Fork CI still needs the maintainer approval click to run on the new head.

WillemJiang (approved):

PR #4082, opened 2026-07-11
fix(runtime): re-buffer subagent events on flush failure to avoid losing steps

nankingjing:
## Summary
`_SubagentEventBuffer` in the run worker dropped buffered subagent events if a batch flush to the `RunEventStore` failed, permanently losing subagent step/end events for that run.

## Problem
On flush failure the pending batch was discarded rather than retained for the next attempt, so a transient store error erased subagent history that the frontend's fetch-on-expand relies on.

## Fix
Re-buffer on failure (`self._pending = batch + self._pending`) so a transient store error is retried on the next flush / in the worker `finally`, instead of silently losing subagent history.

## Test
`backend/tests/test_worker_subagent_persistence.py` covers flush-failure re-buffering (11 passed).

### DIFF
--- a/backend/packages/harness/deerflow/runtime/runs/worker.py
+++ b/backend/packages/harness/deerflow/runtime/runs/worker.py
@@ -201,6 +201,9 @@ class _SubagentEventBuffer:
         try:
             await self._event_store.put_batch(batch)
         except Exception:
+            # Re-buffer the failed batch (ahead of any events queued since) so a
+            # transient store error does not silently drop subagent step events.
+            self._pending = batch + self._pending
             logger.warning("Run %s: failed to persist %d subagent step event(s)", self._run_id, len(batch), exc_info=True)
 
 
--- a/backend/packages/harness/deerflow/runtime/events/store/jsonl.py
+++ b/backend/packages/harness/deerflow/runtime/events/store/jsonl.py
@@ -28,6 +28,7 @@ import logging
 import re
 from datetime import UTC, datetime
 from pathlib import Path
+from typing import Any
 
 from deerflow.runtime.events.store.base import RunEventStore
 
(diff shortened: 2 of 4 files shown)


willem-bd (commented):
Thanks for this fix. The re-buffering approach is sound for atomic stores, but I found one issue for non-atomic `RunEventStore` implementations.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -201,6 +201,9 @@ async def flush(self) -> None:
         try:
             await self._event_store.put_batch(batch)
         except Exception:
+            # Re-buffer the failed batch (ahead of any events queued since) so a
+            # transient store error does not silently drop subagent step events.
+            self._pending = batch + self._pending
```
willem-bd, line 206: [P2] Re-buffering the whole batch can duplicate events when `put_batch` is not atomic

- Location: `backend/packages/harness/deerflow/runtime/runs/worker.py:206`
- Problem: On flush failure the entire `batch` is re-buffered (`self._pending = batch + self._pending`). This assumes `put_batch` is atomic (all-or-nothing). `JsonlRunEventStore.put_batch` is not atomic — it writes events one at a time via `self.put(**ev)` in a loop (`runtime/events/store/jsonl.py:156-163`), each appending a record to disk. A mid-batch failure (disk full / I/O error after writing some events) leaves those events persisted, but the re-buffer re-queues the whole batch, so the next successful flush re-writes the already-persisted events as duplicates (with new `seq` values), producing duplicate step entries in the subagent card's reload history.
- Evidence: `jsonl.py:160-162` — `for ev in events: record = await self.put(**ev)` with no transaction wrapping; `worker.py:206` — `self._pending = batch + self._pending` re-queues unconditionally; `runtime/events/store/base.py:43` — `RunEventStore.put_batch` documents no atomicity guarantee. Before this diff a failed flush silently dropped the batch (no duplicates); the diff introduces the duplicate path. The DB store (`db.py:144`) is atomic (single transaction), so this affects the JSONL store.
- Suggested fix: Either (a) document on `RunEventStore.put_batch` that implementations must be atomic and make `JsonlRunEventStore.put_batch` atomic (buffer all records and write them in one `_write_record` call under the existing write lock), or (b) have `put_batch` return/raise the count of successfully written events so `flush` re-buffers only the unwritten tail.
- Test: Add a test with a store whose `put_batch` succeeds for the first N events then raises, and assert the next flush does not re-write the first N (no duplicates in the persisted records).

nankingjing:
Great catch — fixed in commit 4b1b869. JsonlRunEventStore.put_batch now reserves all seq numbers under the per-thread write lock and appends every record in a single open().write() call, so a mid-batch failure cannot persist half the batch. Added a regression test (test_put_batch_failure_rolls_back_no_partial_records) that monkey-patches _append_records to fail half-way and asserts a subsequent retry produces 4 unique monotonic seqs with no duplicates.

WillemJiang (approved):

PR #4085, opened 2026-07-11
fix(streaming): drop silent delta-discard in _merge_stream_text

nankingjing:
## Problem

The dual-mode `_merge_stream_text` used short-circuits `chunk == existing` and `existing.endswith(chunk)` that silently dropped legitimate delta chunks in messages-tuple mode:

- **CJK reduplication**: '谢谢' tokenized as ['谢','谢'] → buffer stays '谢'  
- **Repeated tokens**: 'gogo' as ['go','go'] → buffer stays 'go'  
- **Suffix-matching tails**: 'hello' as ['hel','l','o'] → drops 'l'

Delta payloads must always append; only strictly-longer cumulative snapshots should replace.

## Fix

Both copies of `_merge_stream_text` are fixed:

- **`app/channels/manager.py`** (IM streaming to Feishu/Telegram/etc.): Gates cumulative-replace with `len(chunk) > len(existing)`, keeps exact-resend guard for values-snapshot no-ops.
- **`deerflow/tui/view_state.py`** (TUI client rendering): Simplified to always-append delta logic; exact-re-send detection moved to caller `reduce()` with `len(action.text) > 1` guard so single-char CJK deltas aren't mistaken for re-sends.

## Tests

- `test_tui_view_state.py`: +8 regression tests (CJK reduplication, repeated tokens, suffix-matching, cumulative, empty, newline split, genuine append)
- `test_channels.py`: +8 equivalent regression tests

All existing tests continue to pass (27/27 TUI, 258/258 channels).

### DIFF
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -370,12 +370,17 @@ def _merge_stream_text(existing: str, chunk: str) -> str:
     """Merge either delta text or cumulative text into a single snapshot."""
     if not chunk:
         return existing
-    if not existing or chunk == existing:
-        return chunk or existing
-    if chunk.startswith(existing):
+    if not existing:
         return chunk
-    if existing.endswith(chunk):
-        return existing
+    # Cumulative re-delivery: strictly longer and starts with existing.
+    if len(chunk) > len(existing) and chunk.startswith(existing):
+        return chunk
+    # Everything else is a delta — always append, even when the delta
+    # happens to match the buffer suffix (e.g. 'hel' + 'l') or equals
+    # the buffer (CJK reduplication: '谢' + '谢' = '谢谢'). Channels feed
+    # only delta ('messages-tuple') events to this function; 'values'
+    # snapshots are consumed via a separate branch, so a same-content
+    # delta (chunk == existing) still represents a fresh token to keep.
     return existing + chunk
 
 
--- a/backend/packages/harness/deerflow/tui/view_state.py
+++ b/backend/packages/harness/deerflow/tui/view_state.py
@@ -222,11 +222,13 @@ def _apply_assistant_delta(state: ViewState, action: AssistantDelta) -> ViewStat
         # match here anyway — the guard is belt-and-suspenders to keep an error
         # row from being merged into if a future change ever gives it an id.
         if isinstance(row, AssistantRow) and row.id == action.id and not row.error:
-            merged = _merge_stream_text(row.text, action.text)
-            if merged == row.text:
-                # No-op re-send (e.g. a values snapshot re-emitting history) —
-                # don't mark this as the actively-streaming message.
+            # Exact re-send of the same full text (e.g. a values snapshot
+            # re-emitting history after reconnection): no-op.  Only multi-char
+            # matches are treated as re-sends so single-char deltas that happen
+            # to equal the buffer (CJK reduplication) are NOT mistaken for no-ops.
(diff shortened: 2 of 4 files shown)


backend/app/channels/manager.py
```diff
@@ -370,12 +370,17 @@ def _merge_stream_text(existing: str, chunk: str) -> str:
...
-        return chunk or existing
-    if chunk.startswith(existing):
+    if not existing:
         return chunk
-    if existing.endswith(chunk):
+    # Cumulative re-delivery: strictly longer and starts with existing.
+    if len(chunk) > len(existing) and chunk.startswith(existing):
+        return chunk
+    # Exact re-send (values snapshot re-emitting history): no-op.
+    if len(chunk) == len(existing) and chunk == existing:
```
willem-bd: [P1] Exact-match no-op check drops legitimate equal-chunk deltas — the PR's own tests fail

- Location: `backend/app/channels/manager.py:379` (`_merge_stream_text`)
- Problem: The added exact-match check `if len(chunk) == len(existing) and chunk == existing: return existing` fires before the delta-append at line 384. For an incoming delta that happens to equal the accumulated buffer (CJK reduplication `'谢' + '谢'`, repeated tokens `'go' + 'go'`, split newlines `'\n' + '\n'`), it returns `existing` unchanged, discarding the delta. The PR's own added comment at lines 381-383 states this case should append (`'谢' + '谢' = '谢谢'`), and the new tests in `test_channels.py` assert it — but they fail against this implementation.
- Evidence: `_get_merge_stream_text()` imports `_merge_stream_text` from `app.channels.manager` (`test_channels.py:7320`), and `test_merge_stream_text_cjk_reduplication` asserts `_merge("谢", "谢") == "谢谢"` (`test_channels.py:7328`). Tracing the head code: for `("谢","谢")`, `len(chunk) == len(existing)` is `1 == 1` → True and `chunk == existing` → True, so line 380 returns `"谢"` instead of `"谢谢"`. The same breaks `test_merge_stream_text_repeated_token_append` (`"go"+"go"` → returns `"go"`) and `test_merge_stream_text_newline_split` (`"\n"+"\n"` → returns `"\n"`). The `view_state.py` version was fixed correctly (no exact-match check); `manager.py` was not.
- Suggested fix: Remove the exact-match check (lines 378-380) from `manager.py`'s `_merge_stream_text` to match `view_state.py`. The channels path feeds only delta (`messages-tuple`) events to this function — `values` snapshots are handled in the separate `elif event == "values"` branch at line 1732 — so the no-op guard serves no purpose here and only drops legitimate equal-chunk deltas. The strictly-longer cumulative check at line 376 already handles true cumulative re-delivery.
- Test: The three tests above (`test_merge_stream_text_cjk_reduplication`, `test_merge_stream_text_repeated_token_append`, `test_merge_stream_text_newline_split`) already cover this — they need to pass. `cd backend && make test`.

willem-bd (commented):
Thanks for this fix. The `view_state.py` change is correct, but I found one issue in the `manager.py` path that breaks the PR's own tests.

nankingjing:
Thanks for the careful review! Fixed in commit 8d007e4 — removed the exact-match no-op at manager.py:378-380 so channels/manager.py now matches view_state.py. All 8 tests in test_channels.py (test_merge_stream_text_*) pass locally with this change.

WillemJiang (approved):

PR #4088, opened 2026-07-11
fix(goal): prevent continuation_count regression from racing continuations

nankingjing:
## Problem

Goal `continuation_count` was computed from a stale goal snapshot outside the per-thread lock (`_prepare_goal_continuation_input` line 919):

```python
next_count = int(goal.get(continuation_count, 0)) + 1
```

Two concurrent continuations reading the same count would each compute `next_count = N + 1` and both write the same value inside `_persist_goal_evaluation` — effectively losing one continuation event against the cap.

## Fix

Inside `_persist_goal_evaluation`, after re-reading `current_goal` under the lock, the `continuation_count` is now computed defensively:

```python
if continuation_count is not None:
    current_count = int(current_goal.get(continuation_count, 0))
    continuation_count = max(continuation_count, current_count + 1)
```

- **No race**: caller's count = current_count + 1, max is a no-op
- **Race**: current_count already bumped by a racing continuation, current_count + 1 > caller's count, the higher (correct) value wins

## Tests

- `test_persist_goal_evaluation_does_not_regress_continuation_count_on_race` — simulates a racing continuation bumping the count before the late writer acquires the lock; asserts count=3 not 2
- `test_persist_goal_evaluation_no_race_uses_caller_count` — confirms normal (no-race) case is unaffected

### DIFF
--- a/backend/packages/harness/deerflow/runtime/runs/worker.py
+++ b/backend/packages/harness/deerflow/runtime/runs/worker.py
@@ -732,6 +732,12 @@ async def _persist_goal_evaluation(
             current_goal = _read_checkpoint_goal(checkpoint_tuple)
             if current_goal is None or not _goal_instance_matches(goal, current_goal):
                 return None
+            # Defensive: compute continuation_count from the fresh current_goal
+            # inside the lock.  The caller computed it from a possibly-stale goal
+            # snapshot; a racing continuation may have already bumped the count.
+            if continuation_count is not None:
+                current_count = int(current_goal.get("continuation_count", 0))
+                continuation_count = max(continuation_count, current_count + 1)
             expected_checkpoint_id = _checkpoint_id(checkpoint_tuple)
             updated_goal = attach_goal_evaluation(
                 current_goal,
--- a/backend/tests/test_goal_worker.py
+++ b/backend/tests/test_goal_worker.py
@@ -597,6 +597,90 @@ async def test_run_agent_reuses_goal_evaluator_model_for_goal_loop(monkeypatch):
 
 
 @pytest.mark.asyncio
+async def test_persist_goal_evaluation_does_not_regress_continuation_count_on_race():
+    """A racing continuation must not overwrite a higher count with a lower one.
+
+    Scenario: two goal continuations run concurrently.  Continuation A reads
+    continuation_count=1, computes next=2.  Continuation B reads the same
+    count=1, computes next=2, but acquires the lock first and writes count=2.
+    When A acquires the lock, the current_goal already has count=2.  Without
+    the defensive guard, A would write count=2 again (stale computation),
+    effectively losing one continuation event.  The guard must compute
+    ``max(stale_next, current_count + 1)`` so A writes count=3.
+    """
+    checkpointer = InMemorySaver()
+    thread_id = "race-count-thread"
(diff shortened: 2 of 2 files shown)


WillemJiang (approved):

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -732,6 +732,12 @@ async def _persist_goal_evaluation(
             current_goal = _read_checkpoint_goal(checkpoint_tuple)
             if current_goal is None or not _goal_instance_matches(goal, current_goal):
                 return None
+            # Defensive: compute continuation_count from the fresh current_goal
+            # inside the lock.  The caller computed it from a possibly-stale goal
+            # snapshot; a racing continuation may have already bumped the count.
+            if continuation_count is not None:
+                current_count = int(current_goal.get("continuation_count", 0))
+                continuation_count = max(continuation_count, current_count + 1)
```
willem-bd, line 740: Heads-up: this `max` over-counts on the line-934 stand-down re-persist path.

The "no-race → no-op" invariant assumes the caller's `continuation_count` is ahead of `current_count` (i.e. `stale_count + 1`). That holds for the primary caller (line 920), but **not** for the "thread changed before continuation" stand-down re-persist at line 934:

```python
# line 920: persist succeeds, writes count = next_count
updated_goal = await _persist(goal, evaluation, no_progress_count, continuation_count=next_count)
...
# line 934: re-persist the SAME next_count just to annotate stand_down_reason
await _persist(latest_goal, evaluation, no_progress_count,
               continuation_count=next_count,
               stand_down_reason="thread_changed_before_continuation")
```

Line 920's persist (in the *same* call) already bumped `current_goal.continuation_count` to `next_count`. So at line 934, `current_count == next_count`, and `max(next_count, next_count + 1) = next_count + 1` — the continuation is counted **twice** against `max_continuations`. I confirmed empirically: seed count=1 → line-920 persist → count=2 → line-934 re-persist (same `continuation_count=2` + `stand_down_reason`) → count=3.

**Severity: low.** The path is rare (needs a user message to land between line 920's commit and line 927's re-read), the over-count is by 1, it's in the conservative direction (cap reached one step *earlier*), and `stand_down_reason` is still recorded correctly. It's also an improvement over the old behavior, which in a racing case would *overwrite* a higher count with `next_count`.

**Suggested fix (one line):** pass `continuation_count=None` at line 934 so `attach_goal_evaluation` preserves the fresh under-lock `current_count` (correct in both race and no-race cases) — line 920 already recorded the increment and line 934 is only annotating with `stand_down_reason`. Alternatively, if counting "wasted" continuations against the cap is intentional, a test documenting that would help.

(The core fix itself is correct: `_goal_instance_matches` doesn't compare `continuation_count`, so the late writer reaches this code; `attach_goal_evaluation` sets the count absolutely; `goal_thread_lock` serializes in-loop writes while `GoalWriteConflict` OCC covers cross-loop races. Both new tests pass.)

willem-bd (commented):
Reviewed the concurrency fix. The core race is correctly handled (verified independently: `_goal_instance_matches` ignores `continuation_count` so the late writer reaches the fix; `attach_goal_evaluation` sets the count absolutely; `goal_thread_lock` serializes in-loop writes while `GoalWriteConflict` OCC covers cross-loop races; both new tests pass, 15/15 in the file, ruff clean). One follow-up worth considering on a related caller — see inline.

PR #4090, opened 2026-07-11
fix(DB): legacy backfill creates missing Index objects on existing tables

nankingjing:
## Problem

`_run_baseline_create_all_sync` (the legacy-branch backfill) calls `Base.metadata.create_all(sync_conn, tables=baseline_tables, checkfirst=True)`. SQLAlchemy's `Table.create(checkfirst=True)` skips a table and **all its `Index` objects** when the table already exists. An index added to the ORM model after the table was first provisioned would therefore never be created.

Specifically, the partial unique index `uq_channel_connection_active_identity` (`WHERE status != 'revoked'`) on `channel_connections` was added to the model but a legacy DB that already had the `channel_connections` table (from an earlier `create_all`) would never get this index. Stamping `0001_baseline` skips alembic's own `create_index` call, so the index is permanently absent. Missing this index silently breaks the single-active-owner invariant enforced at the DB layer by `ChannelConnectionRepository.upsert_connection`.

The same class of bug affects every `Index` (not `UniqueConstraint`) on every baseline table — any index added to a model after the table was first provisioned is silently lost on legacy-backfilled DBs.

## Fix

After `create_all`, explicitly create every `Index` on every baseline table using `Index.create(sync_conn, checkfirst=True)`. This checks for each index independently (not gated on the parent table's existence), so indexes present on the table get a no-op and missing indexes are created.

## Tests

Two new regression tests in `test_persistence_bootstrap.py`:

1. **`test_legacy_backfill_creates_missing_index_on_existing_table`** — Creates the full schema, drops only the `uq_channel_connection_active_identity` index, runs the backfill helper, and asserts the index is recreated.
2. **`test_legacy_backfill_idempotent_when_index_already_exists`** — The explicit index-creation pass must not raise when every index is already present (idempotent).

### DIFF
--- a/backend/packages/harness/deerflow/persistence/bootstrap.py
+++ b/backend/packages/harness/deerflow/persistence/bootstrap.py
@@ -368,9 +368,7 @@ def _run_baseline_create_all_sync(sync_conn: Any) -> None:
                 idx.create(sync_conn, checkfirst=True)
             except Exception:
                 logger.warning(
-                    "bootstrap: failed to create baseline index %r on %r -- "
-                    "the DB may contain rows that violate the index constraint. "
-                    "Address the duplicate data, then re-run bootstrap.",
+                    "bootstrap: failed to create baseline index %r on %r -- the DB may contain rows that violate the index constraint. Address the duplicate data, then re-run bootstrap.",
                     idx.name,
                     table.name,
                 )
--- a/backend/packages/harness/deerflow/persistence/bootstrap.py
+++ b/backend/packages/harness/deerflow/persistence/bootstrap.py
@@ -299,6 +299,19 @@ def _run_baseline_create_all_sync(sync_conn: Any) -> None:
     baseline_tables = [Base.metadata.tables[name] for name in _BASELINE_TABLE_NAMES if name in Base.metadata.tables]
     Base.metadata.create_all(sync_conn, tables=baseline_tables, checkfirst=True)
 
+    # ``create_all`` with ``checkfirst=True`` skips a table and all its
+    # subordinate ``Index`` objects when the table already exists.  An index
+    # that was added to the ORM model after the table was first provisioned
+    # would therefore never be created, and because the legacy branch stamps
+    # ``0001_baseline`` before running upgrade, alembic's own
+    # ``batch_op.create_index`` for baseline-era indexes is skipped too.
+    # Explicitly creating every ``Index`` on every baseline table (each with
+    # its own ``checkfirst=True``) guarantees each index exists regardless of
+    # whether its parent table was just created or already present.
+    for table in baseline_tables:
+        for idx in table.indexes:
+            idx.create(sync_conn, checkfirst=True)
+
 
 def _stamp(cfg: AlembicConfig, revision: str) -> None:
     """Synchronous alembic stamp; callers must wrap in ``asyncio.to_thread``."""
--- a/backend/packages/harness/deerflow/persistence/bootstrap.py
+++ b/backend/packages/harness/deerflow/persistence/bootstrap.py
@@ -139,6 +139,45 @@ _BASELINE_TABLE_NAMES: frozenset[str] = frozenset(
     }
 )
 
+# ``test_baseline_index_names_constant_matches_0001`` pins this set against
+# what 0001 actually creates -- editing 0001 without updating this constant
+# (or vice versa) fires that test.
+_BASELINE_INDEX_NAMES: frozenset[str] = frozenset(
+    {
+        # channel_connections
+        "idx_channel_connections_event_lookup",
+        "ix_channel_connections_owner_user_id",
+        "ix_channel_connections_provider",
+        "uq_channel_connection_active_identity",
+        # channel_conversations
+        "ix_channel_conversations_connection_id",
+        "ix_channel_conversations_owner_user_id",
+        "ix_channel_conversations_provider",
+        "ix_channel_conversations_thread_id",
+        # channel_oauth_states
+        "ix_channel_oauth_states_owner_user_id",
+        "ix_channel_oauth_states_provider",
+        # feedback
+        "ix_feedback_run_id",
+        "ix_feedback_thread_id",
+        "ix_feedback_user_id",
+        # run_events
+        "ix_events_run",
+        "ix_events_thread_cat_seq",
+        "ix_run_events_user_id",
+        # runs
+        "ix_runs_thread_id",
+        "ix_runs_thread_status",
+        "ix_runs_user_id",
+        # threads_meta
+        "ix_threads_meta_assistant_id",
+        "ix_threads_meta_user_id",
+        # users
+        "idx_users_oauth_identity",
+        "ix_users_email",
+    }
+)
+
 
 # Per-engine SQLite bootstrap locks. Per-engine (not module-global) so each
 # engine instance pairs with a lock bound to the event loop that uses that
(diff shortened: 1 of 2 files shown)


backend/packages/harness/deerflow/persistence/bootstrap.py
```diff
@@ -299,6 +299,19 @@ def _run_baseline_create_all_sync(sync_conn: Any) -> None:
...
+    # that was added to the ORM model after the table was first provisioned
+    # would therefore never be created, and because the legacy branch stamps
+    # ``0001_baseline`` before running upgrade, alembic's own
+    # ``batch_op.create_index`` for baseline-era indexes is skipped too.
+    # Explicitly creating every ``Index`` on every baseline table (each with
+    # its own ``checkfirst=True``) guarantees each index exists regardless of
+    # whether its parent table was just created or already present.
+    for table in baseline_tables:
+        for idx in table.indexes:
+            idx.create(sync_conn, checkfirst=True)
```
willem-bd: **Nit (non-blocking) — worth a one-line addition to the comment above.**

This index-level backfill has the same forward-looking risk the module already documents for tables (the `_BASELINE_TABLE_NAMES` restriction): if a future **post-baseline** revision adds an index to a **baseline** table via `op.create_index(...)` *without* `checkfirst=True`, and that index is also on the ORM model, the legacy backfill will now pre-create it here and that revision will collide with `"index already exists"` on the subsequent `upgrade head`.

Today there's no such revision — 0002 is column-only, and 0003's indexes are all on tables that 0003 itself creates (non-baseline) — so this is purely a note for the next contributor. The conventional guard, mirroring `safe_add_column`, is for such a revision to use `checkfirst=True` (or a future `safe_create_index` helper). A sentence in the comment block pointing this out would prevent the surprise.

backend/tests/test_persistence_bootstrap.py
```diff
@@ -591,6 +591,71 @@ async def test_legacy_backfill_skips_non_baseline_tables(tmp_path: Path) -> None
...
+# ---------------------------------------------------------------------------
+
+
+async def _channel_connections_index_names(engine) -> set[str]:
+    async with engine.connect() as conn:
+        return await conn.run_sync(lambda c: {ix["name"] for ix in sa.inspect(c).get_indexes("channel_connections")})
+
+
+@asyncio_test
+async def test_legacy_backfill_creates_missing_index_on_existing_table(tmp_path: Path) -> None:
```
willem-bd, line 641: **Nit (non-blocking).** This regression test (and the rest of the file) exercises the SQLite path only. The `has_index` / `Index.create(checkfirst=True)` mechanism is equally supported on Postgres 2.x, so the same fix applies there — but the partial-unique-index enforcement story (`uq_channel_connection_active_identity`, the headline motivation for this PR) is arguably most interesting on PG deployments.

Not a blocker since it's consistent with the SQLite-only convention used everywhere else in this file; just flagging in case a PG-backed variant is worth adding in a follow-up.

nankingjing:
Thanks @willem-bd — both nits acknowledged:

1. **bootstrap.py:313 comment future-proofing (`checkfirst=True` in revisions)** — happy to add a one-line forward-looking note pointing future contributors at `safe_add_column` as the convention. Will push shortly.

2. **Postgres-backed test variant** — agreed the partial-unique `uq_channel_connection_active_identity` story is most interesting on PG, but this test file follows the SQLite-only convention used elsewhere. A PG test for `TestLegacyBackfillCreatesMissingIndex*` is best as a follow-up PR (membership in pg-backed `tests/test_persistence_bootstrap_pg.py` — I do not see that file yet, so I would open it as a sibling). Deferring unless you want it bundled here.

nankingjing:
Nit #1 addressed in `7ffb53b` — added a short forward-looking block under the legacy index-level backfill comment (`backend/packages/harness/deerflow/persistence/bootstrap.py:312-320`) pointing at the same shape risk `_BASELINE_TABLE_NAMES` already carries, and asking future contributors to mirror `safe_add_column` with `checkfirst=True` (or a future `safe_create_index`). Doc-only, no behavior change.

Nit #2 (PG-backed test variant): leaving as a separate follow-up per the reply above. Thanks for the heads-up on the existing SQLite-only convention.

willem-bd (commented):
Code review of the legacy-backfill index fix. The bug is real and the fix is correct in principle (new tests pass), but the loop's scope is too broad and introduces a verified regression - details inline.

**Headline:** the loop iterates the *current model's* `table.indexes` (a superset of baseline-era indexes), so it pre-creates post-baseline indexes like `uq_runs_thread_active` before revision `0004` can run its `_dedupe_active_runs_per_thread()`. On legacy DBs with duplicate active rows (the case `0004`'s dedupe was written for), `idx.create` raises `IntegrityError` and bricks bootstrap - confirmed against the PR code.

Suggested fix: scope the loop to baseline-era indexes (those `0001` creates) + add a duplicate-data regression test. Everything else is solid.

backend/packages/harness/deerflow/persistence/bootstrap.py
```diff
@@ -299,6 +299,28 @@ def _run_baseline_create_all_sync(sync_conn: Any) -> None:
...
+    #
+    # Forward-looking note: this index-level backfill has the same shape risk
+    # the module already documents for tables (``_BASELINE_TABLE_NAMES``
+    # restriction). If a future post-baseline revision adds an index to a
+    # baseline table via ``op.create_index(...)`` without ``checkfirst=True``,
+    # that revision will collide with ``"index already exists"`` on ``upgrade
+    # head`` because the backfill above already pre-created it. Use
+    # ``checkfirst=True`` (or a future ``safe_create_index`` helper) in such a
+    # revision, mirroring ``safe_add_column``.
+    for table in baseline_tables:
```
willem-bd, line 363: ⚠️ **Critical - this loop over-reaches into post-baseline indexes and crashes bootstrap on legacy DBs with duplicate active runs (verified).**

`table.indexes` is the *current model's* index set - a superset of baseline-era indexes. It includes post-baseline indexes that their owning revision creates with data prerequisites. Specifically, `uq_runs_thread_active` (partial `UNIQUE` on `runs(thread_id) WHERE status IN ('pending','running')`) lives in `run/model.py` `__table_args__` but is created by revision **`0004`**, which runs `_dedupe_active_runs_per_thread()` *first* to cancel duplicate active rows, then creates the index (`0004_run_ownership.py:99-117`).

This backfill now creates `uq_runs_thread_active` **before** `0004` runs - i.e. before the dedupe. On a legacy DB with two+ active rows for the same thread (the exact case `0004`'s dedupe was written for), `idx.create(checkfirst=True)` raises:
```
IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: runs.thread_id
[SQL: CREATE UNIQUE INDEX uq_runs_thread_active ON runs (thread_id) WHERE status IN ('pending', 'running')]
```

I confirmed this against the PR code: seed two pending runs on one thread, then run `_run_baseline_create_all_sync` -> it crashes before `0004` ever runs, bricking Gateway startup on upgrade.

**Before this PR:** `create_all` skipped the index -> `0004` deduped -> created it. Worked in both clean and duplicate cases.

**Suggestion:** scope the loop to baseline-era indexes only (the set `0001_baseline.upgrade()` actually creates), leaving post-baseline indexes to their owning revisions. A `_BASELINE_INDEX_NAMES` constant + guard test (mirroring `_BASELINE_TABLE_NAMES`) would prevent drift.

backend/packages/harness/deerflow/persistence/bootstrap.py
```diff
@@ -299,6 +299,28 @@ def _run_baseline_create_all_sync(sync_conn: Any) -> None:
...
+        for idx in table.indexes:
+            idx.create(sync_conn, checkfirst=True)
```
willem-bd: Same class of issue affects this PR's target index. If a legacy DB already has duplicate non-revoked rows for the same `(provider, external_account_id, workspace_id)`, `idx.create(checkfirst=True)` for `uq_channel_connection_active_identity` raises the same `IntegrityError: UNIQUE constraint failed`.

Unlike `uq_runs_thread_active`, **no revision dedupes channel connections** - so an upgrade over violating data bricks startup with a raw `IntegrityError` and no self-heal path. Before this PR the index was silently missing (the bug being fixed); after, it's a hard crash for affected DBs.

Failing loudly is arguably better than silent absence, but bricking bootstrap with no remediation is rough. Consider a dedupe-then-create step here (mirroring `0004`'s `_dedupe_active_runs_per_thread`) or at least catching the error and surfacing an actionable message.

backend/packages/harness/deerflow/persistence/bootstrap.py
```diff
@@ -299,6 +299,28 @@ def _run_baseline_create_all_sync(sync_conn: Any) -> None:
...
+    # its own ``checkfirst=True``) guarantees each index exists regardless of
+    # whether its parent table was just created or already present.
+    #
+    # Forward-looking note: this index-level backfill has the same shape risk
+    # the module already documents for tables (``_BASELINE_TABLE_NAMES``
+    # restriction). If a future post-baseline revision adds an index to a
+    # baseline table via ``op.create_index(...)`` without ``checkfirst=True``,
+    # that revision will collide with ``"index already exists"`` on ``upgrade
+    # head`` because the backfill above already pre-created it. Use
+    # ``checkfirst=True`` (or a future ``safe_create_index`` helper) in such a
```
willem-bd: Minor: alembic's `op.create_index` / `batch_op.create_index` doesn't accept a `checkfirst` parameter, so the suggestion to use `checkfirst=True` here points at a non-existent API. The established pattern in this repo is `0004`'s manual `sa.inspect(op.get_bind()).get_indexes(...)` + `if name not in existing` guard (see `0004_run_ownership.py:99-103`) - or a future `safe_create_index` helper mirroring `safe_add_column`.

Also: this forward-looking contract is documentation-only. There's no guard test pinning it (unlike `test_baseline_table_names_constant_matches_0001` for the table set), so a future contributor adding `op.create_index(...)` to a baseline table will only discover the collision on a legacy `upgrade head` - which CI on fresh/empty DBs won't catch. Worth either a `safe_create_index` helper or a guard test.

backend/tests/test_persistence_bootstrap.py
willem-bd, line 641: Coverage gaps that let the regression above slip through:

- **Empty table.** Both new tests create the schema and drop the index, but insert **no rows** - so the duplicate-data failure mode (`uq_runs_thread_active` / `uq_channel_connection_active_identity` on violating data) is never exercised. Consider seeding violating rows (e.g. two pending runs on one thread, or two non-revoked channel connections with the same identity) and asserting the backfill doesn't crash.
- **SQLite-only.** The index loop now runs on Postgres too, but the `postgresql_where` predicate rendering is untested. (Consistent with the rest of the file - just noting the gap.)
- **Name-only assertion.** The test checks the index name exists, not that it's `UNIQUE`/partial with the correct `WHERE` predicate. `idx.create` does render from metadata so this is a test-precision nit, not a correctness concern.

WillemJiang:
@nankingjing  please take a look at the recent review comments.

nankingjing:
Pushed a fix for the bootstrap crash — scoped the index loop to `_BASELINE_INDEX_NAMES` so post-baseline unique constraints are not re-created during legacy backfill. Also wrapped `idx.create()` in try/except for safety against duplicate data. Added two regression tests covering the exact crash case.

nankingjing:
All of willem-bd's inline comments have been addressed in the latest push — the loop is scoped to _BASELINE_INDEX_NAMES (same pattern as _BASELINE_TABLE_NAMES), post-baseline indexes like uq_runs_thread_active and uq_channel_connection_active_identity are excluded, the checkfirst=True comment is corrected, and two regression tests cover the duplicate-data crash scenario.

WillemJiang:
@nankingjing  Please fix the lint and unit test errors.

nankingjing:
The lint and unit test failures are fixed.

**Broken tests:**
- `test_legacy_backfill_rejects_post_baseline_indexes` -- used invalid columns `users.name` and `runs.task` (not in baseline schema)
- `test_legacy_backfill_duplicate_channel_connections_does_not_crash` -- used invalid columns `users.name` and `channel_connections.name` (not in baseline schema)

**Fix:**
- Removed nonexistent `name` column from all INSERT fixtures
- Replaced `runs.task` with `runs.multitask_strategy`
- Added all NOT NULL columns without SQL server defaults to raw INSERTs (message_count, total_*_tokens, etc.)
- Used distinct `owner_user_id` values in channel_connections INSERT to avoid `uq_channel_connection_owner_provider_identity` violation
- Collapsed multi-line warning string in bootstrap.py to pass ruff format

**Test results:** 27/27 passed, ruff clean.

WillemJiang (approved):

PR #4092, opened 2026-07-11
fix landing hero background positioning

zhangbububu:
## What changed

- establish the Hero root as the positioning context for its absolute background layers
- remove the vertical translation from the deer-shaped flickering grid

## Why

The absolute Galaxy and FlickeringGrid layers had no positioned Hero ancestor, so they were positioned against a higher containing block. The grid was also shifted downward by `translate-y-8`.

This keeps both background layers aligned to the Hero bounds and centers the deer mask without an extra offset.

## Validation

- `pnpm --dir frontend test` — 604 tests passed
- ESLint on `src/components/landing/hero.tsx`
- Prettier check on `src/components/landing/hero.tsx`
- `pnpm typecheck`
- `git diff --check`

### DIFF
--- a/frontend/src/components/landing/hero.tsx
+++ b/frontend/src/components/landing/hero.tsx
@@ -32,7 +32,7 @@ export function Hero({ className }: { className?: string }) {
   return (
     <div
       className={cn(
-        "flex size-full flex-col items-center justify-center",
+        "relative flex size-full flex-col items-center justify-center",
         className,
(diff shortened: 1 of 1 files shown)


zhangbububu:
### Before

The deer background extends into the next section:

CleanShot 2026-07-11 at 22 10 32@2x

WillemJiang (approved):

PR #4096, opened 2026-07-11
fix(sandbox): stop glob/grep/ls from surfacing disabled skills' files

AnnaSuSu:
## Why

`_is_disabled_skill_path` gates the path a tool is **given**. `ls`, `glob` and `grep` all **descend** from that path and return paths other than the one they were given, so a root anywhere above a disabled skill still serves its files.

`glob` and `grep` never call the gate at all, and `ls` calls it only on its own argument. Driving all four read tools against a real disabled skill (real `extensions_config.json`, real `LocalSandbox`, nothing stubbed):

| root | `ls` | `glob` | `grep` | `read_file` |
|---|---|---|---|---|
| `/mnt/skills` | dir name only | **leaks file paths** | **leaks paths + contents** | n/a |
| `/mnt/skills/public` | **leaks file paths** | **leaks file paths** | **leaks paths + contents** | n/a |
| `/mnt/skills/public/secret-skill` | blocked | **leaks file paths** | **leaks paths + contents** | blocked |

```
$ grep SECRET_PROCEDURE /mnt/skills          # secret-skill is disabled
/mnt/skills/public/secret-skill/SKILL.md:5: SECRET_PROCEDURE = step-1-step-2
```

The error message the gate hands back is the repo's own words — *"Access to its **files** is blocked"* — and `grep` prints those files. The fail-closed comment added in #3889's review is more explicit still:

> Access-control check must fail closed … refuse access rather than **silently serving a disabled skill's files**.

`read_file` is the only one of the four whose gate is complete, because it does not descend: the path it is given *is* the path it reads.

## What changed

**Two clauses, because one root gate cannot cover a recursive tool.**

1. `glob` and `grep` gain the entry gate `ls`/`read_file` already carry, so pointing straight at a disabled skill returns the same actionable error instead of results.
2. `ls`, `glob` and `grep` filter their **results** through `_drop_disabled_skill_paths`.

The filter adds no new authorization logic — it reuses `_is_disabled_skill_path` verbatim, including its fail-closed behaviour, and just applies that existing decision to the paths being returned rather than only to the path requested.

Clause 2 is what actually closes the leak; clause 1 exists so the direct case still explains *why* it was refused, instead of degrading to "no matches found".

**Memoization is required, not an optimization.** `ExtensionsConfig.from_file()` re-reads and re-parses the file on every call — it is not cached — so a per-match gate call would turn a 100-match grep into 100 config reads. Keying the verdict on the skill makes the cost track distinct skills instead of results:

| call | config reads | results |
|---|---|---|
| `glob **/*.py` under `/mnt/user-data/workspace` | **0** | 96 |
| `glob **/*.md` under `/mnt/skills` (12 skills) | **12** | 88 |

Non-skills paths never reach the gate, so the ordinary workspace path pays nothing. The tool bodies run under `await asyncio.to_thread(...)`, so these reads are off the event loop; `make test-blocking-io` stays green.

## Surface area

- [x] **Sandbox** — `backend/packages/harness/deerflow/sandbox/tools.py`

No documentation change needed: `backend/AGENTS.md` does not describe this gate (its Sandbox Tools list does not cover `glob`/`grep` at all), so there is no claim to bring back in line.

## Bug fix verification

Both clauses are anchored separately, and I wrote down what each anchor should do before running it.

**Fix direction** — reverting only `tools.py` turns exactly these five red, each checked individually rather than in aggregate:

```
5 failed, 23 deselected

FAILED ::test_glob_tool_blocks_disabled_skill_root
FAILED ::test_grep_tool_blocks_disabled_skill_root
FAILED ::test_glob_tool_does_not_surface_disabled_skill_from_ancestor_root
FAILED ::test_grep_tool_does_not_surface_disabled_skill_content_from_ancestor_root
FAILED ::test_ls_tool_does_not_surface_disabled_skill_from_category_root
```

The first two guard the entry gate; the last three guard the result filter — reverting only the filter leaves the first two green, which is why both clauses needed their own tests.

**Over-blocking direction** — this narrows what the tools return, so the other side has to be pinned too. Mutating the filter to deny every skill turns six red, and **three of them are tests that already existed**:

```
FAILED ::test_glob_tool_supports_skills_virtual_paths                          ← pre-existing
FAILED ::test_ls_tool_masks_skills_host_paths                                  ← pre-existing
FAILED ::test_ls_tool_skills_path_uses_sandbox_mapping_user_id_not_contextvar  ← pre-existing
FAILED ::test_glob_tool_does_not_surface_disabled_skill_from_ancestor_root
FAILED ::test_grep_tool_does_not_surface_disabled_skill_content_from_ancestor_root
FAILED ::test_ls_tool_does_not_surface_disabled_skill_from_category_root
```

Each of my three new tests asserts the *enabled* sibling skill is still returned from the same root that no longer surfaces the disabled one — those assertions are green on `main` too. They do not guard the bug; they guard the fix from over-reaching.

**Delta set** — running the old and new predicates over a candidate list that deliberately includes names I expected *not* to be dropped:

```
DROPPED  /mnt/skills/public/secret-skill/SKILL.md
DROPPED  /mnt/skills/public/secret-skill/
DROPPED  /mnt/skills/public/secret-skill/references/deep/notes.txt
DROPPED  /mnt/skills/custom/priv-skill/SKILL.md          (per-user disabled custom skill)

kept     /mnt/skills/public/secret-skill-2/SKILL.md      ← name-prefix of the disabled skill
kept     /mnt/skills-extra/data.txt                      ← sibling dir sharing the /mnt/skills prefix
kept     /mnt/skills/public/open-skill/SKILL.md          ← enabled sibling
kept     /mnt/skills/custom/open-custom/SKILL.md
kept     /mnt/skills/public/   /mnt/skills/custom/   /mnt/skills/     (category roots, not skills)
kept     /mnt/user-data/workspace/app.py   /mnt/acp-workspace/notes.md
kept     /mnt/skills/public/nonexistent/SKILL.md
```

Nothing is withheld except a disabled skill's own paths. The two boundary cases (`secret-skill-2`, `skills-extra`) survive because the check keys on path segments, not on a string prefix.

**One pre-existing test needed a fixture, and it is worth saying why rather than burying it.** `test_ls_tool_skills_path_uses_sandbox_mapping_user_id_not_contextvar` lists a *category root*, which now descends into the skills beneath it and resolves each one's enabled state. That lookup needs app config, and the test provided none — so the gate hit its existing fail-closed branch and the listing came back empty, for a reason unrelated to what the test asserts. It now patches `get_paths` / `get_app_config` the way `TestSkillToggleIsolation` already does; **its assertions are unchanged**. It still passes on a clean `main` with the fixture added (so it is not being bent to accommodate this change), and it is one of the three pre-existing tests in the over-blocking red set above (so it is still a live guard).

## Validation

- `cd backend && make format` — all checks passed, 829 files left unchanged.
- `cd backend && make lint` — `ruff check`: all checks passed; `ruff format --check`: 829 files already formatted.
- `cd backend && make test` — **7109 passed, 26 skipped, 8 failed**. The same 8 fail on a clean `origin/main` checkout (pre-existing `web_fetch` cases — `test_browserless_client`, `test_crawl4ai_tools`, `test_fastcrw_tools`; no outbound DNS on this machine). The failure sets are byte-identical; this branch adds none. Passing count is `main`'s 7104 plus the five new tests.
- `cd backend && make test-blocking-io` — 36 passed. The added config reads run inside the sync tool body, which the async wrappers dispatch via `asyncio.to_thread`, so nothing new touches the event loop.
- End-to-end, nothing stubbed — a real `extensions_config.json` marking `secret-skill` disabled, a real `LocalSandbox` with `/mnt/skills` mapped, driving the real `ls` / `glob` / `grep` / `read_file`. On `main` the matrix at the top of this PR reproduces; on this branch every cell is clean, while the enabled sibling `open-skill/SKILL.md` is still returned from every root.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Claude Code assisted with driving the four read tools against a real disabled skill to map which access shapes actually leak (this is what showed the bug was not simply "two missing call sites" — `ls` carries the gate and still leaked from a category root), writing the fix, measuring the config-read cost, and writing the regression tests. The contributor reviewed every line and takes responsibility for it.

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/tools.py
+++ b/backend/packages/harness/deerflow/sandbox/tools.py
@@ -173,8 +173,10 @@ def _extract_skill_name_from_skills_path(path: str) -> str | None:
     if not relative:
         return None
     # Expected patterns: "public/<name>/...", "custom/<name>/...", "legacy/<name>/..."
-    # or "<name>/..." (direct skill access)
-    parts = relative.split("/")
+    # or "<name>/..." (direct skill access). Empty segments are dropped so a
+    # directory entry ("public/", as `ls` emits for dirs) is still recognized as
+    # a category root rather than yielding an empty skill name.
+    parts = [part for part in relative.split("/") if part]
     if len(parts) >= 2 and parts[0] in ("public", "custom", "legacy"):
         return parts[1]
     if len(parts) == 1 and parts[0] in ("public", "custom", "legacy"):
@@ -1773,8 +1775,9 @@ def ls_tool(runtime: Runtime, description: str, path: str) -> str:
         path: The **absolute** path to the directory to list.
     """
     try:
+        user_id = resolve_runtime_user_id(runtime)
         # Block access to disabled skill directories
-        if _is_disabled_skill_path(path, user_id=resolve_runtime_user_id(runtime)):
+        if _is_disabled_skill_path(path, user_id=user_id):
             skill_name = _extract_skill_name_from_skills_path(path) or "unknown"
             return f"Error: Skill '{skill_name}' is disabled. Access to its files is blocked. Enable the skill in settings before using it."
         sandbox = ensure_sandbox_initialized(runtime)
@@ -1802,7 +1805,7 @@ def ls_tool(runtime: Runtime, description: str, path: str) -> str:
             output = mask_local_paths_in_output(output, thread_data)
         # The gate above only covers `path` itself; the listing descends into
         # children, so a root above a disabled skill still exposes its files.
-        entries = _drop_disabled_skill_paths(output.splitlines(), user_id=resolve_runtime_user_id(runtime))
+        entries = _drop_disabled_skill_paths(output.splitlines(), user_id=user_id)
         if not entries:
             return "(empty)"
         output = "\n".join(entries)
--- a/backend/packages/harness/deerflow/sandbox/tools.py
+++ b/backend/packages/harness/deerflow/sandbox/tools.py
@@ -241,6 +241,39 @@ def _is_disabled_skill_path(path: str, *, user_id: str | None = None) -> bool:
         return True
 
 
+def _drop_disabled_skill_paths(paths: list[str], *, user_id: str | None = None) -> list[str]:
+    """Filter out paths that belong to a disabled skill.
+
+    ``_is_disabled_skill_path`` gates the *requested* path, which is enough for
+    ``read_file`` but not for the tools that descend: ``ls``, ``glob`` and
+    ``grep`` return paths other than the one they were given, so a root anywhere
+    above a disabled skill still surfaces its files.  This applies the same check
+    to the results.
+
+    The enabled-state lookup re-reads ``extensions_config.json`` (or the per-user
+    skill state) on every call, so the verdict is memoized per skill — a 100-match
+    grep must not become 100 config reads.
+    """
+    skills_prefix = _get_skills_container_path()
+    verdicts: dict[tuple[str, str], bool] = {}
+    kept: list[str] = []
+    for path in paths:
+        skill_name = _extract_skill_name_from_skills_path(path)
+        if skill_name is None:
+            kept.append(path)
+            continue
+        # Leading segment (the category, or the skill itself in the direct
+        # layout) plus the name identifies the skill the same way
+        # _is_disabled_skill_path does, so paths sharing a key share a verdict.
+        category = path[len(skills_prefix) :].lstrip("/").split("/")[0]
+        key = (category, skill_name)
+        if key not in verdicts:
+            verdicts[key] = _is_disabled_skill_path(path, user_id=user_id)
+        if not verdicts[key]:
+            kept.append(path)
+    return kept
+
+
 def _resolve_skills_path(path: str) -> str:
     """Resolve a virtual skills path to a host filesystem path.
 
@@ -1767,6 +1800,12 @@ def ls_tool(runtime: Runtime, description: str, path: str) -> str:
         output = "\n".join(children)
         if thread_data is not None:
             output = mask_local_paths_in_output(output, thread_data)
+        # The gate above only covers `path` itself; the listing descends into
+        # children, so a root above a disabled skill still exposes its files.
+        entries = _drop_disabled_skill_paths(output.splitlines(), user_id=resolve_runtime_user_id(runtime))
+        if not entries:
+            return "(empty)"
+        output = "\n".join(entries)
         try:
             from deerflow.config.app_config import get_app_config
 
@@ -1811,6 +1850,11 @@ def glob_tool(
         max_results: Maximum number of paths to return. Default is 200.
     """
     try:
+        user_id = resolve_runtime_user_id(runtime)
+        # Block access to disabled skill directories
+        if _is_disabled_skill_path(path, user_id=user_id):
+            skill_name = _extract_skill_name_from_skills_path(path) or "unknown"
+            return f"Error: Skill '{skill_name}' is disabled. Access to its files is blocked. Enable the skill in settings before using it."
         sandbox = ensure_sandbox_initialized(runtime)
         ensure_thread_directories_exist(runtime)
         requested_path = path
@@ -1829,6 +1873,9 @@ def glob_tool(
         matches, truncated = sandbox.glob(path, pattern, include_dirs=include_dirs, max_results=effective_max_results)
         if thread_data is not None:
             matches = [mask_local_paths_in_output(match, thread_data) for match in matches]
+        # The gate above only covers `path` itself; the search descends into it,
+        # so a root above a disabled skill still surfaces its files.
+        matches = _drop_disabled_skill_paths(matches, user_id=user_id)
         return _format_glob_results(requested_path, matches, truncated)
     except SandboxError as e:
         return f"Error: {e}"
@@ -1887,6 +1934,11 @@ def grep_tool(
         max_results: Maximum number of matching lines to return. Default is 100.
     """
     try:
+        user_id = resolve_runtime_user_id(runtime)
+        # Block access to disabled skill directories
+        if _is_disabled_skill_path(path, user_id=user_id):
+            skill_name = _extract_skill_name_from_skills_path(path) or "unknown"
+            return f"Error: Skill '{skill_name}' is disabled. Access to its files is blocked. Enable the skill in settings before using it."
         sandbox = ensure_sandbox_initialized(runtime)
         ensure_thread_directories_exist(runtime)
         requested_path = path
@@ -1919,6 +1971,10 @@ def grep_tool(
                 )
                 for match in matches
             ]
+        # The gate above only covers `path` itself; the search descends into it,
+        # so a root above a disabled skill still surfaces its file contents.
+        allowed = set(_drop_disabled_skill_paths([match.path for match in matches], user_id=user_id))
+        matches = [match for match in matches if match.path in allowed]
         return _format_grep_results(requested_path, matches, truncated)
     except SandboxError as e:
         return f"Error: {e}"
--- a/backend/tests/test_sandbox_search_tools.py
+++ b/backend/tests/test_sandbox_search_tools.py
@@ -1,7 +1,9 @@
+import json
 from types import SimpleNamespace
 from unittest.mock import patch
 
 from deerflow.community.aio_sandbox.aio_sandbox import AioSandbox
+from deerflow.config.paths import Paths
 from deerflow.sandbox.local.local_sandbox import LocalSandbox, PathMapping
 from deerflow.sandbox.search import GrepMatch, find_glob_matches, find_grep_matches
 from deerflow.sandbox.tools import glob_tool, grep_tool, ls_tool
@@ -500,16 +502,33 @@ def test_ls_tool_skills_path_uses_sandbox_mapping_user_id_not_contextvar(tmp_pat
     )
     monkeypatch.setattr("deerflow.sandbox.tools.ensure_sandbox_initialized", lambda runtime: sandbox)
 
+    # Listing a category root descends into the skills below it, so the
+    # disabled-skill gate now resolves each one's enabled state. That lookup
+    # needs app config; without it the gate fails closed (see PR #3889) and the
+    # listing would be empty for a reason unrelated to what this test asserts.
+    skills_root = tmp_path / "skills"
+    (skills_root / "custom").mkdir(parents=True)
+    app_config = SimpleNamespace(
+        skills=SimpleNamespace(
+            get_skills_path=lambda: skills_root,
+            container_path="/mnt/skills",
+            use="deerflow.skills.storage.local_skill_storage:LocalSkillStorage",
+        ),
+        skill_evolution=SimpleNamespace(enabled=False),
+    )
+
     # Leave contextvar unset → get_effective_user_id() returns "default"
     # Before the fix, _resolve_skills_path would resolve to default_custom (empty)
     # After the fix, the sandbox PathMapping resolves to user-abc_custom (has my-skill)
     token = set_current_user(SimpleNamespace(id="default"))  # contextvar says "default"
     try:
-        result = ls_tool.func(
-            runtime=_make_runtime(tmp_path),
-            description="list custom skills",
-            path="/mnt/skills/custom",
-        )
+        with patch("deerflow.config.paths.get_paths", return_value=Paths(base_dir=base_dir)):
+            with patch("deerflow.config.get_app_config", return_value=app_config):
+                result = ls_tool.func(
+                    runtime=_make_runtime(tmp_path),
+                    description="list custom skills",
+                    path="/mnt/skills/custom",
+                )
 
         # Must show user-abc's skill (sandbox mapping), NOT default's empty dir (contextvar)
         assert "my-skill" in result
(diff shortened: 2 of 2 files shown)


backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -241,6 +241,39 @@ def _is_disabled_skill_path(path: str, *, user_id: str | None = None) -> bool:
         return True
 
 
+def _drop_disabled_skill_paths(paths: list[str], *, user_id: str | None = None) -> list[str]:
+    """Filter out paths that belong to a disabled skill.
+
+    ``_is_disabled_skill_path`` gates the *requested* path, which is enough for
+    ``read_file`` but not for the tools that descend: ``ls``, ``glob`` and
```
willem-bd, line 250: **Scope / residual-risk note (not a regression):** this filter closes the leak for `ls`/`glob`/`grep`, but `_is_disabled_skill_path` is only referenced from these four read tools. The `bash` tool has no enabled-state gate, so `cat /mnt/skills/public/secret-skill/SKILL.md`, `grep -r ... /mnt/skills`, or `find /mnt/skills -name SKILL.md -exec cat {} \;` still reads a disabled skill's files.

That gap predates this PR (#3889 gated `read_file` but never `bash`), and this PR is explicitly scoped to the structured read tools — but given the `risk:high` label and a title promising to "stop surfacing disabled skills' files," it's worth explicitly acknowledging the residual exposure so the gate isn't mistaken for airtight. The durable fix is at the mount/path-mapping layer (don't mount disabled skills into the sandbox at all), which is a larger change — a follow-up issue would be enough here.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -241,6 +241,39 @@ def _is_disabled_skill_path(path: str, *, user_id: str | None = None) -> bool:
...
+
+    The enabled-state lookup re-reads ``extensions_config.json`` (or the per-user
+    skill state) on every call, so the verdict is memoized per skill — a 100-match
+    grep must not become 100 config reads.
+    """
+    skills_prefix = _get_skills_container_path()
+    verdicts: dict[tuple[str, str], bool] = {}
+    kept: list[str] = []
+    for path in paths:
+        skill_name = _extract_skill_name_from_skills_path(path)
```
willem-bd, line 263: **Latent edge case:** `_extract_skill_name_from_skills_path` returns `""` (not `None`) for a category directory that carries a trailing slash. `LocalSandbox.list_dir` appends `/` to dirs, so `ls /mnt/skills` yields entries like `/mnt/skills/public/`, for which `relative = "public/"` → `parts = ["public", ""]` → returns `parts[1] == ""`. That empty string then skips the `if skill_name is None` short-circuit here and falls through to a config read.

It currently lands on the right outcome only because `is_skill_enabled("", "public")` defaults unknown skills to **enabled**, so the category dir is kept — but it's fragile (a future default change, or a literal `""` key, would silently drop `public`/`custom`/`legacy` from `ls /mnt/skills`) and wastes one config read per category dir on the security path. The `len(parts) == 1` category-root guard a few lines up in `_extract_skill_name_from_skills_path` already encodes the intent "category root → None"; it just misses the trailing-slash form. Normalizing (e.g. strip the trailing `/` before splitting, or filter empty parts) would close it.

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -1767,6 +1800,12 @@ def ls_tool(runtime: Runtime, description: str, path: str) -> str:
         output = "\n".join(children)
         if thread_data is not None:
             output = mask_local_paths_in_output(output, thread_data)
+        # The gate above only covers `path` itself; the listing descends into
+        # children, so a root above a disabled skill still exposes its files.
+        entries = _drop_disabled_skill_paths(output.splitlines(), user_id=resolve_runtime_user_id(runtime))
```
willem-bd: **Minor:** `resolve_runtime_user_id(runtime)` is evaluated a second time here — the entry gate at ~1777 already called it. `glob_tool`/`grep_tool` hoist it into a local `user_id`; worth doing the same in `ls_tool` for consistency (one resolution per call).

backend/packages/harness/deerflow/sandbox/tools.py
```diff
@@ -1919,6 +1971,10 @@ def grep_tool(
                 )
                 for match in matches
             ]
+        # The gate above only covers `path` itself; the search descends into it,
+        # so a root above a disabled skill still surfaces its file contents.
+        allowed = set(_drop_disabled_skill_paths([match.path for match in matches], user_id=user_id))
```
willem-bd, line 1979: **Minor / cosmetic:** `truncated` was computed before this filter runs. If the matches that got truncated belonged to a disabled skill, the formatted result reports "truncated" while showing fewer lines than the model might expect. Not a correctness issue, but recomputing (or annotating) after the filter would make the flag match the visible output. Same applies to `glob_tool`.

backend/tests/test_sandbox_search_tools.py
```diff
@@ -536,3 +555,112 @@ def test_ls_tool_filters_upload_staging_files(tmp_path, monkeypatch) -> None:
     assert "/mnt/user-data/uploads/report.txt" in result
     assert "/mnt/user-data/uploads/.upload-note.txt" in result
     assert ".upload-active.part" not in result
+
+
+def _make_skills_sandbox(tmp_path, monkeypatch, *, disabled: str):
```
willem-bd, line 560: **Coverage gaps worth considering:**
- This fixture only sets up **public** skills, so all five new tests exercise the `ExtensionsConfig` path. The PR body's manual "delta set" verifies a disabled **custom** (per-user `_skill_states.json`) skill surfaced from `/mnt/skills/custom`, but that case isn't committed as an automated test — and custom/legacy resolve enabled state through a different store (`get_or_new_user_skill_storage`), so it's the path most likely to regress unnoticed.
- All new tests use `LocalSandbox`. The filter is sandbox-agnostic (operates on virtual `/mnt/skills/...` paths), so AIO/remote should behave the same, but nothing pins it.

willem-bd (commented):
Reviewed the diff. The core fix is correct: the result filter (reusing `_is_disabled_skill_path` with per-skill memoization) closes the real leak for the descending tools, and the test suite pins both the fix direction and the over-blocking direction — including asserting the enabled sibling is still returned.

Inline notes below: one residual scope item (bash bypass, pre-existing), one latent trailing-slash edge case in `_extract_skill_name_from_skills_path`, and a few minor nits. None block the in-scope fix.

backend/packages/harness/deerflow/sandbox/tools.py
AnnaSuSu: Agreed, and filed as a follow-up: #4107.

I reproduced it before filing so the issue carries a verified repro rather than a claim — with a real `extensions_config.json` (`secret-skill: {enabled: false}`), a real `LocalSandbox` mapping, and the real tools:

```
read_file: Error: Skill 'secret-skill' is disabled. Access to its files is blocked...
bash cat : SECRET_PROCEDURE = step-1-step-2
bash grep: /mnt/skills/public/secret-skill/SKILL.md:SECRET_PROCEDURE = step-1-step-2
```

One nuance worth recording: `is_host_bash_allowed()` disables host bash entirely under the default `LocalSandboxProvider`, so the bypass is **not** reachable out of the box. It is reachable with `sandbox.allow_host_bash: true`, and under `AioSandboxProvider`/Boxlite/E2B, where bash is allowed unconditionally and `_get_skills_mounts()` bind-mounts whole category dirs (`skills/public` → `/mnt/skills/public`) — i.e. exactly the containerized setups where bash is the main execution path.

As you say, the durable fix is at the mount/path-mapping layer (don't mount disabled skills at all), not command-string parsing — `cat`/`grep -r`/`find -exec`/`python -c` are all equivalent and a shell denylist only grows. That is out of scope here; #4107 carries it. Keeping this PR scoped to the structured read tools.

backend/packages/harness/deerflow/sandbox/tools.py
AnnaSuSu: Good catch — fixed in d7e1aca.

Confirmed the mechanism: `relative = "public/"` → `parts = ["public", ""]` → hits `len(parts) >= 2 and parts[0] in (...)` → returns `parts[1]`, i.e. `""` rather than `None`. The empty name skips the `skill_name is None` short-circuit and falls through to a config read, landing on the right outcome only because `is_skill_enabled("", "public")` defaults unknown skills to enabled.

Fix drops empty segments before the category check, so a trailing-slash category root takes the existing `len(parts) == 1 → None` branch. Added `test_extract_skill_name_treats_category_dir_with_trailing_slash_as_root` covering both directions (category roots with/without slash → None; real skills with/without slash → name), plus `test_ls_tool_keeps_category_dirs_when_listing_skills_root`.

Being precise about what that second test does and does not prove: it does **not** go red without the fix — as you noted, the empty name currently lands on "enabled", so category dirs stay visible today either way. It pins the over-blocking direction you flagged as the future risk (if unknown/empty names ever became fail-closed, `public/`, `custom/`, `legacy/` would silently vanish from `ls /mnt/skills`). The helper unit test is the one that actually red-checks this change.

backend/packages/harness/deerflow/sandbox/tools.py
AnnaSuSu: Fixed in d7e1aca — hoisted into a local `user_id`, one resolution per call, matching `glob_tool`/`grep_tool`.

backend/packages/harness/deerflow/sandbox/tools.py
AnnaSuSu: I would rather not change this one, and I think the current behavior is correct.

`truncated` is a property of the **scan**, not of the display: it means `sandbox.glob`/`grep` hit `max_results`, so there may be more matches beyond the cap. Dropping disabled-skill paths from the results does not make the scan un-truncated. Recomputing it to `False` after the filter would tell the model "this is everything" — which is false, since the truncated tail can still hold matches from *enabled* skills. Keeping `True` is the conservative reading, and the "Results truncated. Narrow the path" hint should keep firing.

The rendered output is self-consistent as-is: the N in `Found N paths (showing first N)` is the post-filter count, which is exactly what is printed.

The real limitation is upstream of the flag: the `max_results` budget is spent *before* filtering, so matches inside a disabled skill can crowd out enabled ones (worst case: the cap is filled entirely by a disabled skill and the tool reports "No files matched"). That is not misleading — nothing is claimed to be hidden — and the truncation hint already steers the model to narrow its search. Fixing it properly means filtering inside the scan or over-fetching, which belongs with the mount-layer work in #4107 rather than here.

backend/tests/test_sandbox_search_tools.py
AnnaSuSu: Right, and this is the one I think mattered most — fixed in d7e1aca.

CUSTOM/LEGACY resolve enabled state through `get_or_new_user_skill_storage()` → per-user `_skill_states.json`, a completely different store from the public skills' `extensions_config.json`. The PR body's manual delta covered it; nothing automated did.

Added `_make_custom_skills_sandbox` (real per-user `_skill_states.json`, real per-user `/mnt/skills/custom` PathMapping, no stubbing of the gate) plus `test_grep_tool_does_not_surface_disabled_custom_skill` and `test_ls_tool_does_not_surface_disabled_custom_skill`. Both red-check against the pre-fix `tools.py`:

```
$ PYTHONPATH=. uv run pytest tests/test_sandbox_search_tools.py -q -k custom_skill   # pre-fix tools.py
FAILED test_grep_tool_does_not_surface_disabled_custom_skill
FAILED test_ls_tool_does_not_surface_disabled_custom_skill
2 failed
```

On the AIO/remote half: I left that unpinned deliberately. `_drop_disabled_skill_paths` operates on virtual `/mnt/skills/...` strings and never touches the sandbox implementation, so a fake AIO sandbox would only be re-testing the same pure function through a second test double. Happy to add one if you would rather have it pinned explicitly.

AnnaSuSu:
Thanks for the review — pushed d7e1aca addressing it. Summary of the five inline notes:

| # | Note | Outcome |
|---|---|---|
| 1 | bash bypass (residual risk) | Filed as **#4107** with a verified repro; out of scope here, as you suggested |
| 2 | trailing-slash edge case in `_extract_skill_name_from_skills_path` | Fixed — empty path segments dropped before the category check |
| 3 | `ls_tool` resolves `user_id` twice | Fixed — hoisted into a local, matching `glob`/`grep` |
| 4 | `truncated` computed before the filter | **Not changed** — reasoning inline: `truncated` describes the scan hitting `max_results`, not the display; recomputing it would falsely tell the model it is seeing everything |
| 5 | Coverage gap: CUSTOM / per-user `_skill_states.json` | Fixed — real per-user fixture + two tests, both red-check against the pre-fix code |

On #4107: I reproduced the bash gap before filing rather than restating it. Worth knowing that host bash is disabled entirely under the default `LocalSandboxProvider`, so it is not reachable out of the box — it is reachable with `sandbox.allow_host_bash: true` and under AIO/Boxlite/E2B, where bash is unconditionally allowed and skills are bind-mounted per *category*. That is, the containerized setups where bash is the main execution path.

Verification for this push:

```
$ PYTHONPATH=. uv run pytest tests/test_sandbox_search_tools.py -q
32 passed

$ PYTHONPATH=. uv run pytest tests/ -q
8 failed, 7113 passed, 26 skipped
```

The 8 failures are pre-existing and unrelated (web-fetch tools hitting this machine's network/SSRF guard: `Error: Refused ... metadata address`); they reproduce identically on an unmodified checkout. `ruff format --check` and `ruff check` are clean.

WillemJiang (approved):

PR #4098, opened 2026-07-11
fix(skills): apply allowed-tools only to active skills

### LINKED ISSUE #4095: [bug] Subagent `task` tool stripped from lead agent on every run (skill allowed-tools policy over-applies)
## Summary

The `task` (subagent delegation) tool is **stripped from the lead agent on every run** — even in Ultra mode with `subagent_enabled: true`. Alongside it, `web_search`, `web_fetch`, `write_file`, `str_replace`, `present_files`, etc. are also removed. The lead agent silently degrades to `{bash, read_file, write_todos, review_skill_package}` and can no longer delegate to subagents.

## Root cause

Two defects combine:

**1. Eval fixtures are loaded as real, enabled skills.**
`load_skills()` recursively scans `skills/public/**`, so skill-reviewer's test fixtures become live skills:

| Loaded skill | Source | `allowed-tools` |
|---|---|---|
| `injection-example` | `skills/public/skill-reviewer/evals/fixtures/prompt-injection/SKILL.md` | `[bash]` (+ `required-secrets: [REVIEWER_SECRET]`) |
| `example-safe-skill` | `skills/public/skill-reviewer/evals/fixtures/publish-candidate/SKILL.md` | `[]` |

`injection-example` is explicitly a prompt-injection calibration fixture and must never be a runtime skill.

**2. The tool-policy union clamps the *entire* lead-agent toolset from passive, non-activated skills.**
`_load_enabled_skills_for_tool_policy()` feeds **all enabled skills** into `filter_tools_by_skill_allowed_tools()`. Per `allowed_tool_names_for_skills()`, once *any* enabled skill declares `allowed-tools`, the toolset is restricted to the union of those declarations (+ `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES = {read_file, review_skill_package}`). So a skill that is merely *enabled* — never activated via `/skill` or loaded by the model — collapses the global toolset.

This was dormant before #4037 (no enabled skill declared `allowed-tools`, so `allowed_tool_names_for_skills` returned `None` → allow-all). #4037 shipped `skill-reviewer` (`allowed-tools: [review_skill_package]`) **and** the fixtures above, activating the dormant filter for every run.

## Reproduction / evidence

`get_enabled_skills_for_config()` on a default install yields:

```
union allowed_tool_names = {'bash', 'review_skill_package'}
effective allowed (+framework) = ['bash', 'read_file', 'review_skill_package']
=> task allowed? False
```

Observed in a real Ultra-mode thread: checkpoint metadata confirmed `subagent_enabled: true, mode: "ultra"`, yet 0 `subagent.*` run events and `runs.subagent_tokens = 0`. The agent's toolset was `{write_todos, read_file, bash, review_skill_package}`; it scraped news via `bash`+`curl`/`requests` (no `web_search`), wrote its report with `cat > file` heredocs (no `write_file`), and could not delegate (no `task`).

Note: even removing the fixtures alone is insufficient — the default-enabled `skill-reviewer` (`allowed-tools: [review_skill_package]`) would still collapse the toolset to `{read_file, review_skill_package}`, dropping even `bash`.

## Suggested fixes

- **Stop discovering test fixtures as skills** — exclude `evals/` / `fixtures/` (or don't recurse into a skill's own subdirectories) during `load_skills()`.
- **Narrow the tool-policy scope** — `allowed-tools` should restrict a skill's authority only when that skill is *activated* / loaded by the model, not when it is merely enabled. A passive enabled skill must not clamp the global lead-agent toolset.
- **Never strip framework tools** — consider adding `task` (and other framework builtins) to the always-available set so subagent delegation can't be removed by skill policy.

Add regression tests for both defects.

## Environment

- Regressed by: `41658c5f` (#4037);
- Affects: all runs (Ultra-mode subagent delegation fully broken; web/file tools also stripped)

Huixin615:
Fixes #4095

## Why

Enabled skills were incorrectly treated as active authorization policies when the lead agent graph was built.

Once any enabled skill declared `allowed-tools`, the lead agent's complete toolset was restricted to the union of every enabled skill's declarations. This removed `task`, web tools, and file tools even when the relevant skill had never been activated or loaded. As a result, Ultra mode could have `subagent_enabled: true` while still being unable to delegate.

The issue was made worse by recursive skill discovery treating `skills/public/skill-reviewer/evals/fixtures/**/SKILL.md` files as real enabled runtime skills. This allowed calibration fixtures such as `injection-example` to affect production tool policy.

## What changed

- Treat a directory containing `SKILL.md` as a skill package boundary while preserving nested skill discovery below namespace directories that are not themselves skills.
- Stop applying lead-agent `allowed-tools` declarations from passive enabled skills during graph construction.
- Apply lead-agent tool policy only to slash-activated skills and skills captured in `ThreadState.skill_context`.
  - Filter both model-visible schemas and tool execution.
  - Preserve the existing union, legacy, empty-declaration, framework-tool, and custom-agent allowlist semantics.
  - Keep `task` subject to an active restricted policy instead of permanently exempting it.
- Share slash activation paths through public `runtime.secret_context` helpers instead of a cross-module private constant.
- Store each model step's policy decision as a versioned, JSON-safe dictionary bound to the middleware instance and exact active-path signature.
  - Malformed, old-version, forged-token, or path-mismatched decisions are ignored and recomputed live.
  - The next model call always refreshes the decision.
  - Sync and async tool calls produced by that model step reuse the decision, avoiding repeated registry scans.
- Support mapping and attribute-based/Pydantic-like state objects and warn on malformed state or `skill_context` shapes.
- Preserve fail-closed behavior for a complete registry-load failure, while warning and skipping an individual stale path so other resolvable active skills remain usable.
- Preserve the existing static subagent policy because all configured subagent skills are loaded into context at startup.

Each active model call intentionally reloads the complete live registry. A TTL could temporarily preserve stale enable/disable or frontmatter authorization, while direct path lookup could bypass the existing custom/public name-shadow winner. Tool calls do not repeat this scan because they reuse the model-step decision.

`allowed-tools` remains behavioral scoping rather than a hard security boundary: alternate loads such as `bash cat` are not captured, and bounded `skill_context` can evict old entries.

## Surface area

- [x] **Default behavior change** — passive enabled skills no longer reduce the lead agent's toolset

## Screenshots / Recording

Not applicable. This change affects backend skill discovery and lead-agent tool policy behavior.

## Bug fix verification

- Primary regression tests:
  - `backend/tests/test_skill_tool_policy_middleware.py`
  - `backend/tests/test_skills_loader.py`
  - `backend/tests/test_skills_bundled.py`
  - `backend/tests/test_user_scoped_skill_storage.py`
  - `backend/tests/test_lead_agent_skills.py`
- Did the tests go red on `main` and green on this branch? **Yes**
  - Before the fix, reviewer eval fixtures appeared in the runtime registry.
  - Before the fix, passive enabled skills were applied during lead-agent construction and removed `task`, web, and file tools.
  - After the fix, passive skills preserve the normal toolset while slash-activated and captured skills enforce their declared policy.

Regression coverage includes package boundaries, the bundled fixture tree, real middleware assembly order, public slash-source helpers, JSON round-tripped and forged decisions, mapping and attribute-based state, passive sync/async fast paths, same-call slash activation, durable skill context, custom-agent allowlists, schema filtering, execution blocking, `task` delegation, stale paths, registry failures, model-step decision reuse, and existing subagent behavior.

## Validation

Executed locally:

```bash
cd backend
UV_CACHE_DIR=/tmp/uv-cache uv run ruff check .
UV_CACHE_DIR=/tmp/uv-cache uv run ruff format --check .
UV_CACHE_DIR=/tmp/uv-cache uv run pytest \
  tests/test_lead_agent_model_resolution.py \
  tests/test_lead_agent_skills.py \
  tests/test_memory_tools.py \
  tests/test_skill_request_scoped_secrets.py \
  tests/test_skill_tool_policy_middleware.py \
  tests/test_skills_bundled.py \
  tests/test_skills_loader.py \
  tests/test_user_scoped_skill_storage.py \
  tests/test_slash_skills.py \
  tests/test_durable_context_middleware.py \
  tests/test_subagent_executor.py -q
git diff --check
```

- Ruff check: passed
- Ruff format check: 833 files already formatted
- Complete PR-related and adjacent test set: 385 passed
- `git diff --check`: passed
- Full local backend run: 7,247 passed, 26 skipped, 13 failed. The failures are local-environment/live integration cases in `test_client_live.py`, external Browserless/Crawl4AI/FastCRW fetch configuration, and the existing config-dependent checkpointer case; none exercise skill discovery or policy code. GitHub required checks are the authoritative clean-environment full-suite result.
- GitHub required checks on commit `8ba0ebfa`: all passed, including backend unit tests, backend blocking-I/O, backend/frontend lint, frontend unit tests, both replay E2E layers, and CLA.

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Reproduced the issue, implemented the scoped discovery and active-policy fix, added regression tests, addressed reviewer feedback, updated documentation, and ran validation. I reviewed the resulting behavior and diff.


Huixin615:
cc @18062706139fcz

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,174 @@
...
+            return [], True
+
+        registry = {posixpath.normpath(skill.get_container_file_path(container_root)): skill for skill in skills}
+        active: list[Skill] = []
+        seen: set[str] = set()
+        for path in paths:
+            skill = registry.get(posixpath.normpath(path))
+            if skill is None:
+                logger.warning("Active skill path could not be resolved for allowed-tools policy: %s", path)
+                return [], True
```
willem-bd: **Fail-closed here is harsher than the sibling secrets middleware.**

A single unresolvable active path returns `[], True`, clamping the lead agent to only `{read_file, review_skill_package}` **and abandoning the other resolvable active skills** at the same time. By contrast, `SkillActivationMiddleware._resolve_registry_skill` returns `None` for an unresolvable path (silent skip).

Because `skill_context` is durable and populated by `DurableContextMiddleware` capturing `read_file` calls, a skill physically deleted from disk *after* being loaded would strip the agent's entire toolset on every subsequent turn in that thread — surfacing to the user as "all my tools vanished."

Suggestion: differentiate the two failure modes:
- **Registry *load* failure** (`except Exception` a few lines up): keep fail-closed — genuine authorization uncertainty. ✅ as-is.
- **Individual path miss** (`skill is None` here): `continue` + warn, and resolve the remaining paths — matching the secrets middleware's per-entry tolerance.

This is the only finding I'd want addressed before merge; the rest are perf/style.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,174 @@
...
+        return paths
+
+    def _active_skills(self, request: ModelRequest | ToolCallRequest) -> tuple[list[Skill], bool]:
+        paths = self._active_paths(request)
+        if not paths:
+            return [], False
+
+        try:
+            storage = self._storage()
+            skills = storage.load_skills(enabled_only=False)
```
willem-bd, line 109: **`load_skills` runs on every model call *and* every tool call once a skill is active.**

`_allowed_names → _active_skills` re-scans and re-parses every `SKILL.md` here on each call. The secrets middleware (`_load_skill_registry_by_path`) pays this only per model call and only when request secrets exist; this middleware also pays it per tool call via `_blocked_tool_message`.

For a long run after `/skill` activation with many tool calls, that's many full directory scans + YAML parses on the hot path. Consider caching the resolved allowed-name set keyed on the active-path signature, recomputed only when the active set changes.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,174 @@
...
+            return blocked
+        return handler(request)
+
+    @override
+    async def awrap_tool_call(
+        self,
+        request: ToolCallRequest,
+        handler: Callable[[ToolCallRequest], Awaitable[ToolMessage | Command]],
+    ) -> ToolMessage | Command:
+        blocked = await asyncio.to_thread(self._blocked_tool_message, request)
```
willem-bd: **`awrap_tool_call` always offloads, even with no active skill.**

This spawns a thread via `asyncio.to_thread` on every async tool call regardless of whether any skill is active. A cheap `_active_paths(request)` pre-check before offloading would skip the thread-pool churn on the common (no-active-skill) path.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,174 @@
...
+from langchain.agents import AgentState
+from langchain.agents.middleware import AgentMiddleware
+from langchain.agents.middleware.types import ModelCallResult, ModelRequest, ModelResponse
+from langchain_core.messages import ToolMessage
+from langgraph.prebuilt.tool_node import ToolCallRequest
+from langgraph.types import Command
+
+from deerflow.runtime.secret_context import _SLASH_SECRET_SOURCE_KEY
+from deerflow.skills.storage import get_or_new_skill_storage, get_or_new_user_skill_storage
+from deerflow.skills.storage.skill_storage import SkillStorage
```
willem-bd: **Nit: `SkillStorage` could live under `TYPE_CHECKING`.**

With `from __future__ import annotations`, the `-> SkillStorage` return annotation is a string never evaluated at runtime, yet this is an eager module-level import while `AppConfig` is correctly guarded under `TYPE_CHECKING` just below. Moving it there would be consistent and drop an unnecessary eager import from the middleware import path.

willem-bd (commented):
Reviewed `fix(skills): apply allowed-tools only to active skills` against the PR head.

Overall this is a solid fix with a sound security model (passive = discoverable, active = enforced), correct middleware composition, and thorough tests. Verified against head `fdb7241`: path-resolution mirrors `SkillActivationMiddleware`, `task` is deliberately not in `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`, and the chain ordering (SkillActivation outer → SkillToolPolicy inner) makes slash activation visible to the policy filter on the same model call.

Four inline notes below — the only one I'd want addressed before merge is the fail-closed asymmetry (a per-path miss should arguably skip + warn rather than clamp the whole toolset). The rest are perf/style nits.

ggnnggez:
## Review

Solid fix for a real bug (#4095), in two coherent parts:

1. **Discovery boundary** — `_iter_skill_files` now prunes recursion (`dir_names.clear()`) once a directory contains `SKILL.md`, so nested fixtures like `skill-reviewer/evals/fixtures/**/SKILL.md` stop registering as runtime skills. Applied consistently across all three walkers (local, user-scoped public/custom/legacy).
2. **Policy timing** — the build-time `filter_tools_by_skill_allowed_tools` clamp is removed from `_make_lead_agent`, replaced by a new `SkillToolPolicyMiddleware` that applies `allowed-tools` only for the run's slash-activated skill (via `_SLASH_SECRET_SOURCE_KEY` on the run context) and skills present in `ThreadState.skill_context`, filtering both model-visible schemas and tool execution.

The architecture mirrors the existing secrets-binding design (resolve by canonical container path against the live registry, never trust the stored entry), which is the right pattern here. Docs are updated per project policy, and the middleware chain renumbering in `backend/AGENTS.md` is correct. Test coverage is broad. One description/behavior mismatch is worth resolving before merge, plus a few smaller items.

### Correctness

- **Unknown active paths do *not* fail closed — the PR body, docstrings, and `backend/AGENTS.md` say they do.** In `_active_skills` (`skill_tool_policy_middleware.py`), an unresolvable path logs a warning and is *skipped*; only a storage exception fails closed. If the only active skill's path fails to resolve (skill uninstalled mid-thread, path drift, or disabled), `active` is empty, `allowed_tool_names_for_skills([])` returns `None`, and the agent reverts to the **full unfiltered toolset** — fail-open. `test_unknown_skill_context_path_is_skipped_while_resolvable_skills_apply` pins this skip semantics, so it's intentional code but contradicts the stated contract. Either make an unresolvable path count as a policy failure (fail closed to framework tools, matching the docs), or fix the PR body/AGENTS.md/docstring to say unknown paths are dropped. Given this is an authorization mechanism, I'd lean toward the former, with the deliberate exception of disabled/de-allowlisted skills (dropping those is correct operator-driven loosening).

- **Policy activation is capture-dependent, so it's behavioral scoping, not a hard boundary.** `skill_context` is populated only when the model loads `SKILL.md` via `read_file` (the durable-context capture path). A skill read via `bash cat` never activates its policy, and `merge_skill_context`'s capacity cap means a long run can evict a restricted skill's entry and silently regain the full toolset while the skill's instructions remain in context. Both are inherent to the design and arguably acceptable, but the README/AGENTS.md text presents this as an enforcement policy — a one-line "best-effort, not a security boundary" caveat would prevent operators over-relying on `allowed-tools`.

- The ordering dependency on `SkillActivationMiddleware` (outer) stamping the slash source before this middleware (inner) reads it is real and correctly guaranteed by append order in `build_middlewares`; the composition test (`test_slash_activation_and_policy_compose_on_the_same_model_call`) pins it. Good.

- The bootstrap branch now binds raw tools unfiltered (`filtered = raw_tools`), consistent with the new model since the middleware still applies at runtime. Fine.

### Performance

- **Uncached full registry scan per model call *and* per tool call while any skill is active.** `SkillStorage.load_skills()` walks the skills tree, parses every `SKILL.md` frontmatter, and re-reads `extensions_config.json` on every invocation — and `_allowed_names` runs it in both `wrap_model_call` and `wrap_tool_call`. A turn with several parallel tool calls multiplies this. `SkillActivationMiddleware` already does a similar per-model-call load for secrets, so there's precedent, but this doubles the model-call cost and newly adds the per-tool-call cost. Suggest a short-TTL memo or computing the registry once per model call and reusing it for that turn's tool calls.
- `awrap_model_call` pays `asyncio.to_thread` even when no skill is active, while `awrap_tool_call` has a passive fast-path. Mirror the `if not self._active_paths(request)` guard in `awrap_model_call` — it makes the common no-skill case free and the two async paths symmetric. (The async offloads themselves are correct per the repo's blocking-IO discipline.)

### Conventions & style

- Importing the underscore-private `_SLASH_SECRET_SOURCE_KEY` from `runtime.secret_context` now has two external consumers; consider promoting it to a public name in a follow-up.
- The boundary comment explaining `dir_names.clear()` exists only in `local_skill_storage.py`; the three copies in `user_scoped_skill_storage.py` are bare. A brief comment (or a shared walk helper — the four loops are near-identical) would help, though the refactor is reasonably out of scope.
- The discovery boundary is a **silent behavior change** for anyone who legitimately nested a skill package inside another: the child just disappears from the registry. A one-time `logger.info` when nested `SKILL.md` files are pruned would make that debuggable.

### Tests

- Coverage is genuinely thorough: sync/async, schema filtering + execution blocking, union/legacy/empty-declaration semantics, custom-agent allowlist, fail-closed storage failure, passive fast-path, package-boundary tests for all three storages, and the eval-fixture regression against the real bundled skills tree.
- Weakness: the middleware tests run entirely against hand-rolled `ModelRequestStub`/`ToolRequestStub` and a monkey-patched `_storage`, so nothing exercises a real `ModelRequest`/`ToolCallRequest`/`ToolRuntime`. The real `ToolCallRequest` dataclass does expose `state`/`runtime` as assumed, and the composition test with the real `SkillActivationMiddleware` partially mitigates — but one integration test through the real middleware chain would materially raise confidence for an auth-relevant change. There's also no sync-path counterpart to `test_async_passive_tool_call_skips_storage_and_thread_offload`.

### Security

- Direction of change is a *loosening* (passive skills no longer clamp), which is the intended fix; the compensating runtime enforcement is sound where it activates, and `task` no longer being exempt closes the delegation bypass.
- Eval fixtures (including `injection-example`) leaving the runtime registry is itself a security win.
- The blocked-tool error message and the fail-closed path expose no config internals — consistent with project practice.
- Main residual risks are the two items above: the fail-open unknown-path edge, and the `bash cat` / context-eviction activation gaps.

### Verdict

Approve with changes requested: resolve the fail-closed claim vs. skip behavior for unknown active paths before merge — as written, the documented security posture and the implementation disagree. The per-tool-call registry scan and the async passive fast-path are worth fixing now while the file is fresh; the rest can be follow-ups.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,178 @@
...
+        return paths
+
+    def _active_skills(self, request: ModelRequest | ToolCallRequest) -> tuple[list[Skill], bool]:
+        paths = self._active_paths(request)
+        if not paths:
+            return [], False
+
+        try:
+            storage = self._storage()
+            skills = storage.load_skills(enabled_only=False)
```
willem-bd, line 109: **Performance — registry reloaded on every active call.** When a skill is active (the rest of a slash-activated run, or every call after a `skill_context` capture), `_active_skills()` calls `load_skills(enabled_only=False)` on every model call **and every tool call**. `load_skills` intentionally has no cache ("re-read every call"): it walks the skills tree, parses YAML frontmatter for every `SKILL.md`, and re-reads `extensions_config.json` each time.

For N parallel tool calls in a step, that's N full registry scans — each just to confirm the tool is allowed (the common case, since `wrap_model_call` already filtered the schema). Before this PR, filtering happened once at agent construction, so the active-skill path went from 0 per-call registry loads to `(model_calls + tool_calls)`.

The passive path is unaffected (`_active_paths` early-returns with no IO), so the blast radius is confined to active-skill runs. Still, for a long task under an active restricted skill this is a real per-step tax. Consider memoizing the resolved `(active_skills, policy_failed)` for the request, or caching `load_skills` at the middleware level with a short TTL / mtime guard — at minimum, reuse one allowed-set across the tool calls of a step instead of recomputing per call.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,178 @@
...
+            if skill.name in seen:
+                continue
+            seen.add(skill.name)
+            active.append(skill)
+        return active, False
+
+    def _allowed_names(self, request: ModelRequest | ToolCallRequest) -> set[str] | None:
+        active_skills, policy_failed = self._active_skills(request)
+        if policy_failed:
+            return set(ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES)
```
willem-bd, line 143: **Fail-closed aggressiveness vs. the stated threat model.** On a transient `load_skills` failure while a skill is active, the toolset collapses to just `read_file` + `review_skill_package`, stripping `task`, `bash`, `web_search`, etc. — likely leaving the agent stuck producing tool errors.

Fail-closed is the right default for a hard security boundary, but the PR explicitly frames this as "best-effort behavioral scoping, not a hard security boundary." That framing slightly undercuts the rationale for such an aggressive fail-closed on what may be a transient IO blip. Worth deciding: either (a) fail open (no restriction) on transient errors given the non-security-boundary framing, or (b) keep fail-closed but surface it more loudly than `logger.exception`. Flagging the tension — not a blocker.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,178 @@
...
+    ) -> ModelCallResult:
+        return handler(self._filter_model_request(request))
+
+    @override
+    async def awrap_model_call(
+        self,
+        request: ModelRequest,
+        handler: Callable[[ModelRequest], Awaitable[ModelResponse]],
+    ) -> ModelCallResult:
+        if not self._active_paths(request):
```
willem-bd: Minor: `_active_paths(request)` is computed here for the gate, then again inside `_filter_model_request` → `_allowed_names` → `_active_skills` → `_active_paths`. Cheap (no IO), but you could thread the already-computed paths through to skip the second traversal (same pattern in `awrap_tool_call` at line 173).

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,178 @@
...
+
+        registry = {posixpath.normpath(skill.get_container_file_path(container_root)): skill for skill in skills}
+        active: list[Skill] = []
+        seen: set[str] = set()
+        for path in paths:
+            skill = registry.get(posixpath.normpath(path))
+            if skill is None:
+                logger.warning("Active skill path could not be resolved for allowed-tools policy: %s", path)
+                continue
+            if not skill.enabled or (self._available_skills is not None and skill.name not in self._available_skills):
```
willem-bd: Minor: loading `enabled_only=False` then per-skill `skill.enabled` filtering is equivalent to `enabled_only=True` here — a disabled referenced skill is dropped either way, so loading disabled skills' metadata is wasted work. Harmless as-is; this disappears naturally if `_active_skills` adopts the per-request caching noted on line 83.

willem-bd (commented):
Reviewed the skill-tool-policy changes. Design is sound and well-tested — this correctly fixes #4095. Left a couple of inline notes: one real per-call registry-reload cost in the active-skill path, one fail-closed-vs-threat-model tension, and two minor nits.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,224 @@
...
+                paths.append(entry["path"])
+        return paths
+
+    def _active_skills_for_paths(self, paths: tuple[str, ...]) -> tuple[list[Skill], bool]:
+        if not paths:
+            return [], False
+
+        try:
+            storage = self._storage()
+            skills = storage.load_skills(enabled_only=False)
```
willem-bd, line 109: **Perf: full `load_skills` runs on every model call when a skill is active**

This loads *all* skills (filesystem walk + per-`SKILL.md` YAML parse + `extensions_config.json` re-read — `skill_storage.load_skills` re-reads every call with no in-memory cache) on every model call that has active paths, just to resolve 1–2 active paths into a registry.

Passive calls short-circuit before this and the decision is cached for the step's tool calls, so the overhead only bites when a skill is actually active. But for a long conversation with an active restrictive skill and a non-trivial skill count, this is real per-turn disk+YAML work (the old static filter loaded once at construction).

Could a targeted lookup (e.g. `storage.read_skill_by_container_path(path)`) or a short per-run cache avoid loading the entire registry to resolve a single path?

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -280,6 +279,12 @@ def build_middlewares(
 
     middlewares.append(SkillActivationMiddleware(available_skills=available_skills, app_config=resolved_app_config, user_id=user_id))
 
+    # Enabled skills are only discoverable metadata. Apply allowed-tools at
+    # runtime after explicit slash activation or an actual skill-file load.
+    from deerflow.agents.middlewares.skill_tool_policy_middleware import SkillToolPolicyMiddleware
+
+    middlewares.append(SkillToolPolicyMiddleware(available_skills=available_skills, app_config=resolved_app_config, user_id=user_id))
```
willem-bd: **No test asserts this middleware is wired — or ordered after `SkillActivationMiddleware`**

The new unit tests construct `SkillToolPolicyMiddleware` directly, and the `make_lead_agent` tests monkeypatch `build_middlewares` to `[]`. So if this `append(...)` is removed — or reordered *before* `SkillActivationMiddleware` (which writes `_SLASH_SECRET_SOURCE_KEY` inside `wrap_model_call`) — every test still passes but slash activation silently stops enforcing policy.

The composition test `test_slash_activation_and_policy_compose_on_the_same_model_call` manually chains the two middlewares, so it doesn't guard the real assembly.

Worth a test that calls the real `build_middlewares(...)` and asserts `SkillToolPolicyMiddleware` is present and ordered after `SkillActivationMiddleware`.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,224 @@
...
+        )
+        return await handler(filtered)
+
+    @override
+    def wrap_tool_call(
+        self,
+        request: ToolCallRequest,
+        handler: Callable[[ToolCallRequest], ToolMessage | Command],
+    ) -> ToolMessage | Command:
+        blocked = self._blocked_tool_message(request)
```
willem-bd: **Sync `wrap_tool_call` lacks the empty-paths early return the async variant has**

`awrap_tool_call` short-circuits on `if not self._active_paths(request)` to skip the `to_thread` offload; the sync path always calls `_blocked_tool_message` → `_allowed_names`. It's still correct (empty paths → `allowed=None` → not blocked, and no storage load happens via the early return in `_active_skills_for_paths`), but it does redundant tuple/dict work on every sync tool call.

Mirroring the early return would make the intent obvious and match the async path.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,224 @@
...
+from typing import TYPE_CHECKING, override
+
+from langchain.agents import AgentState
+from langchain.agents.middleware import AgentMiddleware
+from langchain.agents.middleware.types import ModelCallResult, ModelRequest, ModelResponse
+from langchain_core.messages import ToolMessage
+from langgraph.prebuilt.tool_node import ToolCallRequest
+from langgraph.types import Command
+
+from deerflow.runtime.secret_context import _SLASH_SECRET_SOURCE_KEY
```
willem-bd: **Reaches for a private (`_`-prefixed) symbol from another module**

`secret_context` exports public helpers (`read_active_secrets`, `extract_request_secrets`) but no public accessor for the slash-source path, so this imports the underscore-prefixed `_SLASH_SECRET_SOURCE_KEY` and assumes its `{"path": str}` shape — an implicit cross-module contract now shared with `SkillActivationMiddleware` (the writer).

A small public helper in `secret_context` (e.g. `read_slash_skill_source_path(context) -> str | None`) consumed by both sides would centralize the shape.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,224 @@
...
+
+    @staticmethod
+    def _runtime_context(request: ModelRequest | ToolCallRequest) -> dict | None:
+        context = getattr(getattr(request, "runtime", None), "context", None)
+        return context if isinstance(context, dict) else None
+
+    def _store_policy_decision(self, request: ModelRequest, paths: tuple[str, ...], allowed: set[str] | None) -> None:
+        context = self._runtime_context(request)
+        if context is not None:
+            context[_POLICY_DECISION_CONTEXT_KEY] = _PolicyDecision(
```
willem-bd: **Storing a non-serializable dataclass in the shared runtime context**

A `frozen=True, slots=True` dataclass holding a `frozenset` gets written into `context["__skill_tool_policy_decision"]`. Fine today — DeerFlow's own tracer doesn't copy the context (per `secret_context.py`), and this isn't secret so it correctly isn't redacted. But if a future path serializes the context (custom tracers, debug dumps), a `_PolicyDecision`/`frozenset` may not round-trip, and it's visible to any context enumerator.

A plain `tuple`/`dict` (or a non-context side channel) would be more robust. Minor — flagging for awareness.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,224 @@
...
+        paths: list[str] = []
+        context = getattr(getattr(request, "runtime", None), "context", None)
+        if isinstance(context, dict):
+            slash_source = context.get(_SLASH_SECRET_SOURCE_KEY)
+            if isinstance(slash_source, dict) and isinstance(slash_source.get("path"), str):
+                paths.append(slash_source["path"])
+
+        state = getattr(request, "state", None) or {}
+        try:
+            entries = state.get("skill_context") or []
```
willem-bd, line 87: **`state` access assumes a dict-like `.get()`**

If `AgentState` is ever a Pydantic model rather than a dict, `state.get("skill_context")` raises `AttributeError`, gets swallowed by the `try/except`, and `skill_context` is silently ignored — i.e. a captured active skill's policy never applies (fail-open). Works today (state is a dict in practice and in tests), but the silent skip is the weaker direction; a one-line comment noting the dict assumption would help future readers.

airene:
Many people have likely encountered this issue. The workaround prior to the merge is to disable the `skill-reviewer` and `example-safe-skill` skills, or to disable all skills that specify `allowed-tools` (including those with `allowed-tools: []`).

I've found that after the update, the lead_agent has almost no tools left.

Huixin615:
> Many people have likely encountered this issue. The workaround prior to the merge is to disable the `skill-reviewer` and `example-safe-skill` skills, or to disable all skills that specify `allowed-tools` (including those with `allowed-tools: []`).
> 
> I've found that after the update, the lead_agent has almost no tools left.

Thanks for your advice, I will take a look.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,252 @@
...
+                paths.append(entry["path"])
+        return paths
+
+    def _active_skills_for_paths(self, paths: tuple[str, ...]) -> tuple[list[Skill], bool]:
+        if not paths:
+            return [], False
+
+        try:
+            storage = self._storage()
+            skills = storage.load_skills(enabled_only=False)
```
willem-bd, line 109: Performance note (non-blocking): `load_skills(enabled_only=False)` runs on every model call whenever an active skill exists, re-reading and parsing every `SKILL.md` from disk. The trade-off is sound - no TTL means enable/disable, frontmatter edits, and the custom/public name-shadow winner all take effect immediately, and the async path offloads this via `to_thread`. At typical skill counts the cost is negligible vs. LLM latency. If skill counts or network-mounted storage ever grow, an mtime/content-hash-invalidated registry cache would preserve the live-reload correctness without the per-step re-scan.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,252 @@
...
+        )
+        return await handler(filtered)
+
+    @override
+    def wrap_tool_call(
+        self,
+        request: ToolCallRequest,
+        handler: Callable[[ToolCallRequest], ToolMessage | Command],
+    ) -> ToolMessage | Command:
+        if not self._active_paths(request):
```
willem-bd: Minor: `_active_paths(request)` is computed twice on the active path - once here for the guard, then again inside `_blocked_tool_message` -> `_allowed_names`. It's cheap (runtime-context lookup + a `skill_context` iteration), but threading `paths` through into `_blocked_tool_message`/`_allowed_names` would avoid the re-scan. The same double-computation applies to `awrap_tool_call` below.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,252 @@
...
+
+    @override
+    async def awrap_tool_call(
+        self,
+        request: ToolCallRequest,
+        handler: Callable[[ToolCallRequest], Awaitable[ToolMessage | Command]],
+    ) -> ToolMessage | Command:
+        if not self._active_paths(request):
+            return await handler(request)
+        blocked = await asyncio.to_thread(self._blocked_tool_message, request)
```
willem-bd: Minor: when the model-step decision is cached (the common case after the first model call), this still dispatches a thread just to run `_blocked_tool_message` -> `_allowed_names` -> dict lookup + comparison. For steps that fan out many concurrent tool calls that's avoidable thread-pool churn. Fast-pathing the cached-decision check before `to_thread` (or only offloading when a live registry load is actually needed) would skip the dispatch. Negligible at typical counts.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,252 @@
...
+        return active, False
+
+    def _allowed_names_for_paths(self, paths: tuple[str, ...]) -> set[str] | None:
+        active_skills, policy_failed = self._active_skills_for_paths(paths)
+        if policy_failed:
+            return set(ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES)
+        allowed = allowed_tool_names_for_skills(active_skills)
+        if allowed is None:
+            return None
+        return allowed | set(ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES)
```
willem-bd, line 147: Behavior-change note: because `task` is not in `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`, a slash-activated (or `skill_context`-captured) skill that declares a restrictive `allowed-tools` now also blocks `task` delegation for that turn. This is the intended closure of the previous loophole, but it's worth surfacing in release notes so skill authors know: a skill that should still allow subagent delegation under its own policy must now explicitly list `task` in its `allowed-tools`.

backend/packages/harness/deerflow/skills/storage/local_skill_storage.py
```diff
@@ -83,6 +83,12 @@ def _iter_skill_files(self) -> Iterable[tuple[SkillCategory, Path, Path]]:
                 dir_names[:] = sorted(name for name in dir_names if not name.startswith("."))
                 if SKILL_MD_FILE not in file_names:
                     continue
+                # A directory containing SKILL.md is a package boundary. Any
+                # nested SKILL.md files belong to that package's supporting
+                # resources (for example eval fixtures), not to the runtime
+                # skill registry. Namespace directories without SKILL.md still
+                # recurse, preserving layouts such as public/team/helper.
+                dir_names.clear()
```
willem-bd, line 91: Behavior-change note (migration): treating a `SKILL.md`-containing directory as a package boundary is the right fix for the eval-fixture leakage, but it's a mild breaking change for any existing custom skill packages that relied on nested `SKILL.md` discovery - a package can no longer host a legitimately nested runtime skill. Worth a release-note line so authors of unusual nested-skill layouts know to flatten them.

willem-bd (commented):
Reviewed the full diff plus surrounding code (`tool_policy`, storage backends, the durable/skill_context path flow, and middleware assembly). Correctness is solid:
- Path formats are consistent across the registry, the slash source, and `skill_context` (all derive from the canonical container path).
- Fail-closed on a storage load failure vs. fail-open on an individual stale path is the right asymmetry.
- The decision-cache validation (version + `owner_token` + `active_paths` signature + `allowed_names` shape, with `type(version) is not int` guarding against bool) is thorough; forged/stale/path-mismatched decisions fall back to live resolution.
- The slash source is run-scoped, so the restriction correctly does not persist into the next turn.
Test coverage is excellent (passive/active, sync/async, cache reuse, forgery, fail-closed, attribute/mapping state, `task` delegation).

The inline comments below are all non-blocking: one performance note (per-step registry reload), two minor tool-call-path nits, and two behavior-change notes worth a release-note line. No blocking issues - nice work.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,252 @@
...
+from deerflow.skills.tool_policy import ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES, allowed_tool_names_for_skills
+from deerflow.skills.types import Skill
+
+if TYPE_CHECKING:
+    from deerflow.config.app_config import AppConfig
+    from deerflow.skills.storage.skill_storage import SkillStorage
+
+logger = logging.getLogger(__name__)
+
+_POLICY_DECISION_CONTEXT_KEY = "__skill_tool_policy_decision"
```
willem-bd: This key stores the cached policy decision including `owner_token` — the per-middleware-instance secret that authenticates the decision against caller-merged `runtime.context`.

The `owner_token` is load-bearing for anti-forgery: the `active_paths` signature alone doesn't prevent forgery (a caller controls/knows the slash-activation path), so the token is the only value a caller can't synthesize. I confirmed a client *can* inject this key — `build_run_config` copies `body.config['context']` verbatim and `strip_internal_context_keys` only scrubs `non_interactive` — and the token check is what rejects a forged `{"allowed_names": null}` decision.

Unlike the sibling slash-source/binding keys, this key is **not** in `REDACTED_CONTEXT_KEYS` (`secret_context.py`). The `kwargs_json` path is safe — `services.py` stores `redact_config_secrets(body.config)`, capturing pre-run config, not the mid-run mutation written here — but the redaction list exists because runtime-context keys can reach other observable surfaces (traces). If the token leaks there, a caller with the matching token + paths could bypass the active skill's tool policy.

**Suggested fix:** add `_POLICY_DECISION_CONTEXT_KEY` to `REDACTED_CONTEXT_KEYS`. One line, preserves the token's anti-forgery value.

backend/tests/test_skills_bundled.py
```diff
@@ -32,3 +33,11 @@ def test_bundled_skill_frontmatter_is_valid(skill_dir: Path) -> None:
...
 def test_skills_public_dir_has_skills() -> None:
     assert BUNDLED_SKILL_DIRS, f"no SKILL.md found under {SKILLS_PUBLIC_DIR}"
+
+
+def test_runtime_registry_excludes_skill_reviewer_eval_fixtures() -> None:
+    skills = get_or_new_skill_storage(skills_path=SKILLS_PUBLIC_DIR.parent).load_skills(enabled_only=False)
+    names = {skill.name for skill in skills}
+
+    assert "skill-reviewer" in names
+    assert names.isdisjoint({"injection-example", "example-safe-skill", "partial-package-example", "vague-helper", "zh-output-example"})
```
willem-bd: This set is missing one fixture. The actual fixture frontmatter `name`s are:

- `injection-example`, `example-safe-skill`, `partial-package-example`, `vague-helper`, `zh-output-example` (all checked here)
- `blocked-example` (dir `blocked/`) — **not** in the set

If the package boundary regressed and only `blocked-example` leaked into the registry, this test would still pass.

**Suggested fix:** add `"blocked-example"`, or assert no registered skill path contains `/evals/fixtures/` (robust to future fixtures).

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,251 @@
...
+                paths.append(entry["path"])
+        return paths
+
+    def _active_skills_for_paths(self, paths: tuple[str, ...]) -> tuple[list[Skill], bool]:
+        if not paths:
+            return [], False
+
+        try:
+            storage = self._storage()
+            skills = storage.load_skills(enabled_only=False)
```
willem-bd, line 109: Performance: this runs `load_skills(enabled_only=False)` (disk read + frontmatter parse for every skill) on every model step that has active paths. For a long conversation with a restricted skill captured in `skill_context`, that's repeated per-turn I/O. The trade-off is defensible (no stale TTL; preserves the custom/public name-shadow winner), but it's a real per-step cost for skill-heavy deployments. An mtime-based cache (skill dir + frontmatter mtime) would preserve live-edit correctness without re-parsing every `SKILL.md` each step - worth a follow-up if this shows up in profiles.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,251 @@
...
+        paths: tuple[str, ...] | None = None,
+        refresh_decision: bool = False,
+    ) -> ModelRequest:
+        resolved_paths = tuple(self._active_paths(request)) if paths is None else paths
+        allowed = self._allowed_names_for_paths(resolved_paths) if refresh_decision else self._allowed_names(request)
+        if refresh_decision:
+            self._store_policy_decision(request, resolved_paths, allowed)
+        if allowed is None:
+            return request
+        tools = [tool for tool in request.tools if getattr(tool, "name", None) in allowed]
```
willem-bd, line 218: Behavior change: this filter now also drops `describe_skill` when a restricted skill is active. On `main`, `describe_skill` is appended to `final_tools` *after* the construction-time `filter_tools_by_skill_allowed_tools` call in `_make_lead_agent`, so it always survived. After this PR, construction-time filtering is gone and `describe_skill` sits in `request.tools`, so it's subject to this per-call filter - and it isn't in `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES` (`read_file`, `review_skill_package`). Likely fine (you've scoped to one skill), but it's an untested change. If intended, a regression test would lock it in; if not, add `describe_skill` to the exempt set.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -574,7 +579,7 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
     # Build skill search setup from policy-filtered skills (same list used for
     # tool-policy filtering), so describe_skill only exposes allowed skills.
     skill_setup = build_skill_search_setup(
-        skills_for_tool_policy,
+        enabled_skills,
```
willem-bd, line 599: Stale comment: the comment above this (`# Build skill search setup from policy-filtered skills (same list used for tool-policy filtering), so describe_skill only exposes allowed skills.`) is now inaccurate - `enabled_skills` is no longer used for tool-policy filtering (that moved into `SkillToolPolicyMiddleware`), and `describe_skill` exposes the *enabled* catalog set, not tool-policy-"allowed" skills. Rewording to e.g. "Build skill search setup from enabled skills so `describe_skill` only advertises discoverable skills" would avoid conflating the catalog set with the policy set.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,251 @@
...
+
+    @staticmethod
+    def _runtime_context(request: ModelRequest | ToolCallRequest) -> dict | None:
+        context = getattr(getattr(request, "runtime", None), "context", None)
+        return context if isinstance(context, dict) else None
+
+    def _store_policy_decision(self, request: ModelRequest, paths: tuple[str, ...], allowed: set[str] | None) -> None:
+        context = self._runtime_context(request)
+        if context is not None:
+            context[SKILL_TOOL_POLICY_DECISION_CONTEXT_KEY] = {
```
willem-bd, line 158: Assumption worth documenting: this decision-reuse relies on `request.runtime.context` being the same mutable dict from `wrap_model_call` through `wrap_tool_call` (so the stored decision is visible to the tool-call path). `test_real_model_and_tool_requests_share_the_model_step_policy_decision` pins that for the current LangGraph, but it's a contract that could shift on an upgrade. Failure degrades gracefully - each tool call revalidates live (still correct, just slower) - so not a correctness risk; a one-line comment here noting the shared-context assumption would help future readers.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,251 @@
...
+        )
+
+    @override
+    def wrap_model_call(
+        self,
+        request: ModelRequest,
+        handler: Callable[[ModelRequest], ModelResponse],
+    ) -> ModelCallResult:
+        paths = tuple(self._active_paths(request))
+        return handler(self._filter_model_request(request, paths=paths, refresh_decision=True))
```
willem-bd: Sync/async asymmetry: for active paths the sync `wrap_model_call` runs `_filter_model_request` -> `load_skills()` synchronously on the calling thread, while the async path offloads the same work via `asyncio.to_thread`. The lead agent is async, so the async path is the production one, but the asymmetry is a little fragile - if a sync wrapper is ever dispatched from an event loop, this blocks. Either a comment noting sync-only usage or mirroring the `to_thread` offload would make the intent explicit.

willem-bd (commented):
Solid fix for #4095 - passive enabled skills no longer clamp the lead toolset at construction, and the eval-fixture package-boundary change is clean. CI is green and the test coverage (forged/stale/path-mismatched decisions, fail-closed, sync+async decision reuse) is thorough. Inline notes below are mostly low-severity polish; the main one worth tracking is the per-step registry reload.

**Nits** (not anchored inline):
- With `from __future__ import annotations`, `Command`, `ModelCallResult`, `ModelResponse`, `ModelRequest`, `ToolCallRequest`, and `AgentState` are only needed for (string) annotations and could move into the existing `TYPE_CHECKING` block. Cosmetic; ruff doesn't enforce it.
- `_active_paths` is read twice in `awrap_tool_call` (once for the guard, once inside `_blocked_tool_message` -> `_allowed_names`). Both are cheap state/dict reads, so negligible, but caching locally reads cleaner.

ShenAC-SAC:
The items below are new (deduplicated against @willem-bd's rounds — not re-raising the registry-reload cost, thread-hop churn, fail-closed-on-load-failure aggressiveness, `describe_skill` drop, or the stale comment at `agent.py` he already flagged).

**IMPORTANT: the active-set union lets the model widen its own policy via `read_file`, so the #4191 scenario stays one step away.**

`extract_skills` captures any successful `read_file` of a `SKILL.md` under the skills root into `skill_context` — no activation ceremony, no check that the skill relates to the current slash activation. `_active_paths` then unions that skill's `allowed-tools` into the policy. Since `read_file` is in `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`, a restrictive skill can never block the read itself. Concretely, replaying #4191: `/content-research` (no `write_file`) is active; the model reads `content-article-generation/SKILL.md` (a natural move — it's referenced in the same pipeline's HANDOFF); next model call, the union includes `write_file`, and the handwritten-deliverable failure mode is back.

Why this matters: the union semantics were designed for an operator-controlled population (enabled skills). This PR moves the population under model control, which silently changes what the union means as a policy. The README's best-effort caveats cover *uncaptured* loads (bash cat) and eviction, but not this *captured-load widening* direction.

Requested action: decide the intended semantics before merge. Options I see: (a) while a slash activation is live, treat the slash skill's policy as dominant and don't union in passively-read skills; (b) keep the union but document this dilution vector explicitly and add a regression test encoding it as intended. I lean (a) for slash-activated runs — slash is the explicit-ceremony path, same reasoning as the secrets design (#3861/#3938) — but either is workable if it's a recorded decision rather than an accident.

**IMPORTANT: when every active path fails to resolve, the policy silently becomes allow-all.**

In `_active_skills_for_paths`, an unresolvable/disabled/out-of-allowlist path is skipped with a warning; if *all* active paths are skipped, the function returns `([], False)`, `allowed_tool_names_for_skills([])` returns `None`, and `_filter_model_request` applies no filtering at all. So: transient `load_skills()` failure → harshest outcome (framework-only), but a definitively dangling policy reference (skill deleted/renamed/disabled mid-run while still referenced by the live slash activation or `skill_context`) → fully unrestricted toolset, with only a log line. Those two failure modes point in opposite directions.

Requested action: treat the all-paths-unresolved case like a policy failure (framework-safe fallback) or at least make it a deliberate, tested behavior. Current coverage misses it: `test_unknown_skill_context_path_is_skipped_while_resolvable_skills_apply` pairs the stale path with a resolvable one, and `test_active_policy_load_failure_fails_closed_to_framework_tools` covers only the raise branch.

**IMPORTANT: no test exercises enforcement through the assembled chain.**

All new tests construct `SkillToolPolicyMiddleware` directly with hand-built request stubs; the composition test nests two `wrap_model_call` lambdas manually; the assembly test asserts index ordering only. The suite already has the stronger pattern for other middlewares (DurableContext, LoopDetection, TerminalResponse all have tests that build a real `create_agent` graph and invoke it). One end-to-end test — compiled agent, restricted skill active, disallowed tool call comes back with the policy error `ToolMessage` — would catch wiring, ordering, and state-shape regressions that the unit stubs can't, and is the natural home for the all-paths-stale case above.

**QUESTION: is dropping build-time clamping intended for custom agents with an explicit `skills` config?**

The old build-time filter also clamped a custom agent whose `agent_config.skills` pinned a restrictive skill — the operator arguably expressed intent there, distinct from the passive-enabled-skill case #4095 is about. The PR inverts that coverage (`test_make_lead_agent_does_not_apply_passive_skill_policy_when_cache_is_cold` now asserts `bash`/`web_search` present). If intentional, a release-note line for custom-agent authors would help; if not, keeping construction-time filtering only for the explicit `agent_config.skills` case is a contained middle ground.

**NIT: two more stale docs beyond the one already flagged.**

- `assemble_deferred_tools` docstring still says "Call AFTER tool-policy filtering so the deferred catalog never exposes a tool the agent is not allowed to use" — both call sites now pass unfiltered `raw_tools`; the fail-closed guarantee it describes has no input anymore (runtime middleware covers the active case, but the docstring describes a build-time invariant that's gone).
- `backend/AGENTS.md` item 16 says the middleware "must remain immediately after `SkillActivationMiddleware` … a real assembly test pins this order", but the assertion is `activation_idx < policy_idx < durable_idx` — relative order, not adjacency. Either assert adjacency or soften the claim.

**CONSIDER: unify the two fixture-exclusion mechanisms.**

After this PR there are two independent answers to "is this nested SKILL.md a real skill": SkillScan's `is_eval_fixture_skill_md` (path-pattern, `evals/fixtures/` only) and the new structural package-boundary walk (any nested SKILL.md). Scopes differ and nothing ties them together — a nested `examples/SKILL.md` is flagged by SkillScan yet silently dropped from the registry, and `test_skills_bundled.py` hardcodes a third check. A shared predicate (or a consistency test) would prevent drift.

**CONSIDER: extract a shared active-skill resolver.**

`_storage()`, the normpath-keyed registry build, and the enabled/allowlist gates duplicate `SkillActivationMiddleware._load_skill_registry_by_path`/`_resolve_registry_skill` — including the #3938 confused-deputy reasoning, which now lives only in the original's comments. A shared resolver keeps the secrets and tool-policy surfaces from drifting, and as a bonus collapses the two independent full registry reloads on model calls that have both an active skill and caller-supplied secrets (extends the perf thread willem opened).

**NIT:** the three-line package-boundary idiom (`dir_names[:] = …; if SKILL_MD_FILE not in file_names: continue; dir_names.clear()`) now appears four times across the two storage walkers; a shared iterator would keep future boundary-rule changes single-site.

Happy to re-review quickly once the first three are settled — the first one just needs a decision, not necessarily code.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,290 @@
...
+            return _POLICY_SOURCE_SKILL_CONTEXT, tuple(paths)
+        return _POLICY_SOURCE_PASSIVE, ()
+
+    def _active_skills_for_paths(self, paths: tuple[str, ...]) -> tuple[list[Skill], bool]:
+        if not paths:
+            return [], False
+
+        try:
+            storage = self._storage()
+            skills = storage.load_skills(enabled_only=False)
```
willem-bd, line 109: **Perf: full live-registry reload on every model call during an active run.** `_active_skills_for_paths` calls `storage.load_skills(enabled_only=False)` here (parses every `SKILL.md` frontmatter from disk) and rebuilds the path-keyed registry on every model call whenever a skill is slash-active or in `skill_context`. The decision-reuse in `_read_policy_decision` avoids repeating this for tool calls of the same step, but a multi-step slash-activated run still does N full registry scans.

The freshness tradeoff (no stale TTL, no direct-path bypass of the custom/public name-shadow winner) is defensible. For high-step-count runs or large skill trees this is measurable per-step latency -- consider a per-run registry cache invalidated on explicit skill edits rather than per-model-call.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,290 @@
...
+        available_skills: set[str] | None = None,
+        app_config: AppConfig | None = None,
+        user_id: str | None = None,
+        slash_source_owner_token: str | None = None,
+    ) -> None:
+        super().__init__()
+        self._available_skills = set(available_skills) if available_skills is not None else None
+        self._app_config = app_config
+        self._user_id = user_id
+        self._slash_source_owner_token = slash_source_owner_token or secrets.token_urlsafe(24)
```
willem-bd: **Suggestion: make the shared `slash_source_owner_token` required, not optional.** Both `SkillActivationMiddleware` and this middleware accept `slash_source_owner_token: str | None = None` and fall back to `or secrets.token_urlsafe(24)`. `build_middlewares` correctly mints one token and passes it to both (pinned by `test_build_middlewares_orders_skill_activation_before_policy_and_durable_context`), but a future construction path that builds them separately would silently get *different* tokens -> the policy middleware would never see a slash activation -> **fail open to passive** (no enforcement -- not a security hole, but a silent functional bypass).

Making the token required turns mis-pairing into a loud construction error instead of a silent fail-open.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,290 @@
...
+        context = getattr(getattr(request, "runtime", None), "context", None)
+        return context if isinstance(context, dict) else None
+
+    def _store_policy_decision(self, request: ModelRequest, policy: _PolicySignature, allowed: set[str] | None) -> None:
+        context = self._runtime_context(request)
+        if context is not None:
+            source, paths = policy
+            context[SKILL_TOOL_POLICY_DECISION_CONTEXT_KEY] = {
+                "version": _POLICY_DECISION_VERSION,
+                "owner_token": self._decision_owner_token,
```
willem-bd, line 160: **Defense-in-depth note: the `owner_token` is stored in the same runtime context it protects.** Forge-resistance holds because `runtime.context` is server-constructed (a client-injected/forged decision fails the token + source + active_paths checks in `_read_policy_decision`) and the key is in `REDACTED_CONTEXT_KEYS` (stripped by both `redact_secret_context_keys` and `redact_config_secrets`).

But this is "lock next to key": the guarantee depends on **every** serialization/trace/SSE/persistence surface going through `redact_secret_context_keys`/`REDACTED_CONTEXT_KEYS`. The `redact_secret_context_keys` docstring notes the trace builder doesn't copy context today, so redaction is belt-and-suspenders -- worth a grep or seed-test that no path copies `runtime.context` unredacted, since a regression there would leak the token and enable forging a wider `allowed_names` decision.

willem-bd (commented):
Review of the active-skill `allowed-tools` policy fix.

Correctness verified against the PR's commits (not a local checkout): forge-resistance in `_read_policy_decision` (version/owner_token/source/active_paths all checked; `type(...) is not int` rejects bool/float), fail-closed to `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES` on an unresolved active set, passive fast path skips storage, and decision reuse is bound to source+paths so stale/forged/path-or-source-mismatched decisions fall back to live. `task` is correctly not exempt (absent from `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES` at base). Redaction is complete: `redact_config_secrets` -> `redact_secret_context_keys` -> `REDACTED_CONTEXT_KEYS`, which now includes `SKILL_TOOL_POLICY_DECISION_CONTEXT_KEY`. The `package-nested-skill-md` SkillScan rule and eval-fixture exemption pre-exist, so the new scanner test is valid without a scanner change. Test coverage -- including the `_PolicyBypassModel` schema-vs-execution defense-in-depth test -- is excellent.

Three inline notes below: per-step registry-reload cost, the optional shared-token fail-open, and token-in-context redaction completeness.

One behavior change worth flagging in release notes (not a regression -- strictly better than the buggy prior state): slash-activating a skill that declares `allowed-tools` without `task` now blocks `task` delegation for that run with `"Tool 'task' is not allowed by the active skill policy."` Intentional (can't escape policy via subagent), but users who slash-activate a skill and then delegate may be surprised.

ShenAC-SAC (commented):
@Huixin615 Thanks for the thorough fix — the active-vs-passive policy split and the skill package-boundary change both look directionally right. I found one remaining middleware-composition bug that I think should be fixed before merge.

**BLOCKER: an active restrictive skill can lose all access to an explicitly allowed deferred MCP tool.**

`SkillToolPolicyMiddleware._filter_model_request()` keeps only names declared by the active skill plus `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`. However, `tool_search` is appended later by `assemble_deferred_tools()`, is not normally declared by skills, and is not framework-exempt. `DeferredToolFilterMiddleware` then hides the allowed MCP schema until it has been promoted.

I reproduced this on the current PR head with a tagged deferred MCP tool named `calc` and an active skill whose `allowed-tools` is exactly `["calc"]`:

```text
assembled tools:       ["calc", "tool_search"]
after skill policy:    ["calc"]
after deferred filter: []
```

The skill explicitly authorizes `calc`, but on a fresh thread the model receives zero tools unless auto-routing happens to promote it first. This is a loss of discovery infrastructure caused by middleware composition, rather than the intended authorization restriction.

The same mechanism removes `describe_skill` after a restrictive skill becomes active. The existing [inline note](https://github.com/bytedance/deer-flow/pull/4098#discussion_r3576654662) identifies that behavior change; the concrete `tool_search` reproduction shows that this class of issue is functionally reachable, not just untested.

Could we preserve the framework discovery path while continuing to enforce the active policy on what can be promoted and executed? A simple global exemption for `tool_search` should also be checked against denied-tool promotion/schema visibility because the routing middleware runs after the skill policy.

Please add integration coverage for:

- an active skill explicitly allowing a deferred MCP tool: `tool_search` remains usable and the allowed tool can be promoted/called;
- a denied deferred MCP tool cannot bypass schema/execution policy;
- deferred skill discovery remains usable after a restrictive skill becomes active.

I would hold the merge until this path is covered.

Huixin615:
@willem-bd @ShenAC-SAC Thanks for the detailed review. I will think again based on your feedbacks and then make a necessary fix.

backend/packages/harness/deerflow/agents/middlewares/skill_tool_policy_middleware.py
```diff
@@ -0,0 +1,364 @@
...
+            return _POLICY_SOURCE_SKILL_CONTEXT, tuple(paths)
+        return _POLICY_SOURCE_PASSIVE, ()
+
+    def _active_skills_for_paths(self, paths: tuple[str, ...]) -> tuple[list[Skill], bool]:
+        if not paths:
+            return [], False
+
+        try:
+            storage = self._storage()
+            skills = storage.load_skills(enabled_only=False)
```
willem-bd, line 109: Per-model-call cost: this loads the full registry (`enabled_only=False` walks every skill dir and parses each `SKILL.md` frontmatter) on every active model call - the cached decision only covers same-step tool calls, and `wrap_model_call` always refreshes (`refresh_decision=True`). For a long multi-turn run with an active skill, that's repeated O(total skills) I/O per turn.

I see why a TTL/direct-path cache was avoided (enable/disable changes, frontmatter edits, and custom/public name-shadow winners must take effect live), and for typical skill counts this is likely fine - but worth monitoring in production. A content-hash-keyed cache (invalidated when skill files change) or a short TTL would cut the repeated parsing without sacrificing live resolution. Note `test_next_model_call_refreshes_the_policy_decision` pins `storage.load_calls == 2`, so any caching would need that test revised.

backend/packages/harness/deerflow/skills/tool_policy.py
```diff
@@ -11,9 +11,18 @@ class NamedTool(Protocol):
...
 
 # Framework built-ins that remain available even when an active skill declares
-# allowed-tools. They support controlled framework workflows rather than
-# extending the reviewed/activated skill's own tool authority.
-ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES = frozenset({"read_file", "review_skill_package"})
+# allowed-tools. They support controlled file/review/discovery workflows rather
+# than extending the reviewed/activated skill's own business-tool authority.
+# In particular, promotion through tool_search does not restore a tool removed
+# by SkillToolPolicyMiddleware, and describe_skill only returns catalog metadata.
+ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES = frozenset(
```
willem-bd, line 18: Heads-up on a quiet behavior change: `task` is intentionally absent from this set, so a slash-activated (or in-context) restrictive skill that omits `task` from its `allowed-tools` can no longer delegate to subagents - the schema is filtered out by `SkillToolPolicyMiddleware` and the call is blocked. This is correct and documented in CHANGELOG, but any existing skill that declared `allowed-tools` and relied on `task` being implicitly available will silently lose delegation.

Worth a quick sweep of `skills/public/*` (and known custom skills) that declare `allowed-tools` to confirm they list `task` where delegation is expected. The union-semantics test covers the mechanism, but there's no positive test that a skill declaring `task` retains it - might be worth adding.

willem-bd (commented):
Two follow-up notes from review (the rest is positive - solid root-cause fix, thoughtful security design, exceptional tests). Posting inline for tracking.

Huixin615:
@willem-bd Thanks for the follow-up. For the first point, I’m intentionally keeping the live registry reload on each active model call in this PR. Adding a TTL or direct-path cache could delay enable/disable and frontmatter changes or bypass the existing custom/public shadowing semantics; same-step tool calls already reuse the model-step decision. We can revisit caching separately if production profiling shows this is significant.

For the second point, I swept the bundled public skills: skill-reviewer is the only runtime public skill with restrictive allowed-tools, and it does not delegate; skills that require subagents currently use legacy allow-all semantics. I added a parameterized regression test covering both slash activation and skill_context, verifying that explicitly declaring task preserves its schema and allows execution while undeclared tools remain filtered.

WillemJiang (approved):

PR #4101, opened 2026-07-11
fix(auth): resolve email accounts case-insensitively

DaoyuanLi2816:
## Summary

Email lookup was case-sensitive even though local and OIDC registration normalize addresses differently. That could create duplicate accounts that differed only by case and could bypass the existing OIDC collision check.

This change lowercases email writes at the repository boundary and uses a case-insensitive lookup for legacy rows. Registration now checks for mixed-case legacy rows before relying on SQLite's case-sensitive unique constraint. Legacy case-variant duplicates resolve deterministically by `created_at` and `id`; password-only updates do not rewrite an unchanged mixed-case address and therefore do not collide with an existing canonical row.

## Legacy data audit

Deployments upgrading from an earlier version can find case-variant duplicates with:

```sql
SELECT lower(email) AS canonical_email, COUNT(*) AS row_count
FROM users
GROUP BY lower(email)
HAVING COUNT(*) > 1;
```

The read path keeps the oldest row deterministic, but operators should review and consolidate any rows returned by this query before relying on the canonical address for sign-in.

## Tests

Real SQLite repository and OIDC coverage includes case-insensitive lookup, registration against mixed-case legacy rows, duplicate rejection, update normalization, legacy duplicate handling, and cross-case OIDC collision blocking. The complete auth test module passes (65 tests); ruff check and format are clean.


willem-bd (commented):

**[P2] `update_user` can 500 on legacy case-variant duplicate rows**

**Location:** `backend/app/gateway/auth/repositories/sqlite.py` — `update_user`

**Problem:** The new `user.email = _normalize_email(user.email)` in `update_user` rewrites the email to lowercase on *every* write. For a database that already holds two case-variant rows (e.g. `Victim@x.com` + `victim@x.com` — precisely the legacy state the PR's "Migration-safety for existing rows" section prepares for), lowercasing the mixed-case row's email on its next update collides with the already-canonical row and raises `IntegrityError`. `update_user` doesn't catch `IntegrityError`, so this surfaces as a 500 on the change-password / reset-admin paths. Pre-fix, `update_user` stored the email unchanged, so this collision could not occur — i.e. the regression is introduced by this PR.

**Evidence:**
- `update_user` adds `user.email = _normalize_email(user.email)` then `row.email = user.email` with no `IntegrityError` handling (only `UserNotFoundError`).
- The `/change-password` route calls `update_user` on every password change; when `new_email is None` it skips the `get_user_by_email` collision check, so lowercasing in `update_user` is the only thing that can collide.
- The read path was hardened against this exact legacy state (`order_by(created_at).limit(1)` + `.scalars().first()`), but the write path was not, so the migration-safety claim is asymmetric.
- The opportunistic rehash path in `LocalAuthProvider.authenticate` wraps `update_user` in `except Exception`, so login survives; the change-password (`auth.py`) and reset-admin (`reset_admin.py`) call sites do not, so they will 500.

**Suggested fix:** Only rewrite the email when it actually differs from the stored value, so a password-only update on a legacy mixed-case row doesn't touch the email column:
```python
canonical = _normalize_email(user.email)
if canonical != row.email:
    row.email = canonical
```
This still canonicalizes real email changes (covered by `test_update_user_normalizes_email`) but avoids re-writing unchanged legacy emails. Alternatively, catch `IntegrityError` in `update_user` and raise a descriptive error, or document that legacy duplicate pairs require admin dedup before writes will succeed.

**Test:** using the existing legacy-duplicate fixture, call `update_user` on the mixed-case row with a new password (unchanged email) and assert it does not raise.

willem-bd (commented):
Reviewed the email case-insensitivity fix. The approach is correct and well-reasoned: canonicalizing at the repository chokepoint (write) plus a case-insensitive read (`func.lower`) closes the local-vs-OIDC normalization gap, and the non-destructive legacy handling (oldest-wins, no bulk rewrite, guarded `update_user`) degrades gracefully on databases that already tripped the bug. The new tests use the real repo/provider/provisioning stack, which is what exposed the gap the mock-based OIDC tests missed. Approve with minor follow-ups.

Two non-code notes (not inline-anchorable):

- The PR body says "six regression tests" but **seven** are added - `test_update_user_on_legacy_mixed_case_row_does_not_collide` isn't mentioned, and the "Migration-safety" section's claim that a stored email "is only rewritten to canonical form the next time that row is written (e.g. a password/email change)" is imprecise: the code only rewrites on a *genuine* email change, not a password-only change (that's the point of the `canonical_email != _normalize_email(row.email)` guard). The code and the 7th test are correct; the prose should be tightened.

- Normalization now lives in two places: `_normalize_email` here and `email = identity.email.lower()` in `user_provisioning.py:56`. The latter can't be removed (it feeds the domain check before the repo is touched), but if a third normalization rule ever diverges this bug class returns. Centralizing in one utility (or on the `User` model) would be more robust - out of scope for this fix.

Inline comments below cover a tiebreaker + index note on the read query, an `IntegrityError`-handling asymmetry on `update_user`, and the input-mutation side effect.

backend/app/gateway/auth/repositories/sqlite.py
```diff
@@ -82,10 +107,16 @@ async def get_user_by_id(self, user_id: str) -> User | None:
...
 
     async def get_user_by_email(self, email: str) -> User | None:
-        stmt = select(UserRow).where(UserRow.email == email)
+        # Case-insensitive match: an account is keyed by its email regardless of
+        # the case the caller supplies (see ``_normalize_email``). ``.first()``
+        # with a deterministic ``created_at`` ordering resolves to the oldest
+        # account instead of raising if a pre-fix database already holds two
+        # rows differing only in case, so the fix never turns a legacy duplicate
+        # pair into a 500.
+        stmt = select(UserRow).where(func.lower(UserRow.email) == _normalize_email(email)).order_by(UserRow.created_at).limit(1)
```
willem-bd: Two non-blocking follow-ups on this read query (the fix itself is correct):

- **No tiebreaker.** `order_by(UserRow.created_at)` picks "oldest" for legacy duplicates, but if two case-variant rows ever share the same `created_at` the choice is non-deterministic. `created_at` is `NOT NULL` so there's no NULL-ordering issue, and real legacy duplicates are usually far apart in time - but `.order_by(UserRow.created_at, UserRow.id)` makes "oldest wins" fully deterministic.

- **Index bypass.** `func.lower(UserRow.email)` wraps the column, so the existing `email` index can't be used (full scan). Negligible today - `get_current_user_from_request` resolves by `id` (deps.py:465), so this path is only hit on login/SSO/password-change/`reset_admin`, not per request. For a large Postgres deployment a functional index `CREATE INDEX ON users (lower(email))` would help; or a one-time migration to canonicalize legacy rows would let this return to direct equality and reuse the existing index. The non-destructive choice is defensible - just flagging it for scale.

backend/app/gateway/auth/repositories/sqlite.py
```diff
@@ -99,7 +130,24 @@ async def update_user(self, user: User) -> User:
...
+            # change-password / reset-admin paths that do not catch it. Guarding
+            # against the *raw* stored value (``canonical != row.email``) is not
+            # enough: the mixed-case row's canonical form still differs from its
+            # own stored casing, so it would rewrite and collide anyway. A genuine
+            # change still normalises, so the unique constraint keeps enforcing
+            # case-insensitive uniqueness for updated rows; get_user_by_email
+            # already resolves legacy mixed-case rows case-insensitively on read.
+            canonical_email = _normalize_email(user.email)
+            if canonical_email != _normalize_email(row.email):
+                row.email = canonical_email
```
willem-bd, line 155: `update_user` doesn't catch `IntegrityError`, while `create_user` re-raises it as `ValueError`. The change-password endpoint (`routers/auth.py:383-397`) pre-checks uniqueness via `get_user_by_email` - now correctly case-insensitive - but there's a TOCTOU window between that check and the `await session.commit()` below: a concurrent registration creating the colliding account makes the commit raise an uncaught `IntegrityError` -> 500. (The OIDC path handles this same race by catching `ValueError` at `user_provisioning.py:93`.)

Pre-existing, not introduced here, and the canonical write is strictly an improvement (previously this case silently created a duplicate row). But since canonicalization makes the constraint more likely to fire, consider mirroring `create_user` - catch `IntegrityError` -> `ValueError` - so callers can return 409/400 instead of a raw 500.

backend/app/gateway/auth/repositories/sqlite.py
```diff
@@ -65,7 +84,13 @@ def _user_to_row(user: User) -> UserRow:
...
 
     async def create_user(self, user: User) -> User:
-        """Insert a new user. Raises ``ValueError`` on duplicate email."""
+        """Insert a new user. Raises ``ValueError`` on duplicate email.
+
+        The email is canonicalised to lowercase before insert so the existing
+        unique constraint enforces case-insensitive uniqueness for new rows and
+        the returned ``User`` reflects the stored form.
+        """
+        user.email = _normalize_email(user.email)
```
willem-bd, line 93: Minor: `create_user` (`user.email = _normalize_email(...)`) and `update_user` (`user.email = row.email`) both mutate the caller's `User` object. Intentional - the returned object reflects the canonical stored form - and benign for every current caller (verified: `authenticate`, `reset_admin`, change-password, OIDC provisioning). But it's a side effect a caller wouldn't expect from a `create`/`update`. A one-line note on the `UserRepository` interface (`repositories/base.py`) that the returned `User.email` is the canonical stored form would help; alternatively return a fresh object instead of mutating in place.

backend/app/gateway/auth/repositories/sqlite.py
DaoyuanLi2816: Added `id` as the deterministic tiebreaker. A functional lower-email index would require a separate migration strategy for legacy duplicates, so it remains outside this fix.

backend/app/gateway/auth/repositories/sqlite.py
DaoyuanLi2816: Keeping the pre-existing `update_user` error-contract race outside this PR. The case-insensitive pre-check is improved here, but translating repository errors also requires updating the caller contract.

backend/app/gateway/auth/repositories/sqlite.py
DaoyuanLi2816: Documented on the repository interface that create/update return the canonical stored email and may update the supplied model.

willem-bd, line 93: Registration is the one write path without a case-insensitive pre-check. `create_user` dedupes only via the `IntegrityError` from the `email` UNIQUE constraint, which is case-sensitive in SQLite (BINARY collation) and Postgres -- so lowercasing all *new* writes catches new-vs-new collisions, but not a new lowercase write against a pre-existing mixed-case legacy row.

Concretely: legacy `Victim@x.com` + `POST /register` with `victim@x.com` -> `EmailStr` keeps `victim@x.com` -> `create_user` lowercases (no-op) -> inserts `victim@x.com` -> constraint sees `victim@x.com` != `Victim@x.com` -> no conflict, second row created, HTTP 201. The register handler (`routers/auth.py:359`) does no `get_user_by_email` pre-check, unlike change-password (`routers/auth.py:417`) and OIDC (`user_provisioning.py:70`), which both pre-check case-insensitively -- so registration is the odd path out and the stated "one email = one account" invariant doesn't hold here against legacy data.

Suggested fix (mirrors the other two paths), right after this line:

```python
if await self.get_user_by_email(user.email) is not None:
    raise ValueError(f"Email already registered: {user.email}")
```

Distinct from the `update_user` TOCTOU deferred at line 150: that's a concurrent new-vs-new race where the constraint *does* fire; this is legacy mixed-case where it *doesn't*.

backend/app/gateway/auth/repositories/sqlite.py
```diff
@@ -82,10 +107,17 @@ async def get_user_by_id(self, user_id: str) -> User | None:
             return self._row_to_user(row) if row is not None else None
 
     async def get_user_by_email(self, email: str) -> User | None:
-        stmt = select(UserRow).where(UserRow.email == email)
+        # Case-insensitive match: an account is keyed by its email regardless of
+        # the case the caller supplies (see ``_normalize_email``). ``.first()``
+        # with a deterministic ``created_at`` ordering resolves to the oldest
```
willem-bd, line 117: Operational heads-up on the oldest-wins choice: if a legacy case-variant pair exists and a user has actively been logging into the *newer* row (their known password is on the newer account), this fix changes which row `get_user_by_email` resolves to -- so a previously-correct password can start failing after deploy. Worth a release note / AGENTS.md note so operators with legacy duplicate data can audit before users are locked out, e.g.:

```sql
SELECT lower(email) AS e, count(*) FROM user_row GROUP BY e HAVING count(*) > 1;
```

Not a code-change request -- just flagging the user-facing impact of an otherwise-correct deterministic resolution.

backend/app/gateway/auth/repositories/sqlite.py
DaoyuanLi2816: Added a case-insensitive pre-check before insert and a regression test against a mixed-case legacy row. The complete auth test module passes.

backend/app/gateway/auth/repositories/sqlite.py
DaoyuanLi2816: Added the audit query and the oldest-row sign-in caveat to the PR description. The actual table name is `users`.

WillemJiang (approved):

PR #4102, opened 2026-07-11
fix(models): apply stream_chunk_timeout default to all BaseChatOpenAI subclasses

DaoyuanLi2816:
## Summary

The 240s `stream_chunk_timeout` default added for #3189 (in #3195) is scoped to a class-path allowlist of `langchain_openai:ChatOpenAI` and `deerflow.models.patched_openai:PatchedChatOpenAI` only. Every *other* OpenAI-compatible provider in the repo subclasses langchain-openai's `BaseChatOpenAI` and inherits the very same `stream_chunk_timeout` mechanism, but is excluded by that string allowlist:

- `VllmChatModel`
- `MindIEChatModel`
- `PatchedChatDeepSeek`
- `PatchedChatMiMo`
- `PatchedChatStepFun`
- `PatchedChatMiniMax`

For these providers `_apply_stream_chunk_timeout_default` takes the "non-OpenAI" branch, which:

1. **never injects the 240s default** — they silently fall back to langchain-openai's aggressive built-in **120s** chunk-gap timeout (the exact value in the #3189 traceback), and
2. **`pop`s a user's explicit `stream_chunk_timeout` from `config.yaml`** with no warning — so a self-hosted user who sets `stream_chunk_timeout: 300` to survive long reasoning pauses has that override silently discarded.

Notably, #3189 was reported against **`mimo-v2.5`** (`model=mimo-v2.5` in the traceback), which maps to `PatchedChatMiMo` — one of the excluded subclasses. So the original fix did not actually cover the model the issue was filed against.

## What changed

`backend/packages/harness/deerflow/models/factory.py`:

- `_apply_stream_chunk_timeout_default` now gates on `issubclass(model_class, BaseChatOpenAI)` instead of the `_OPENAI_COMPAT_USE_PATHS` string allowlist, and receives the already-resolved `model_class` (rather than the config `use` string).
- Any genuine `BaseChatOpenAI` subclass now gets the 240s default when unset **and** preserves an explicit override.
- Genuinely non-OpenAI clients (e.g. `ChatAnthropic`, which is **not** a `BaseChatOpenAI` subclass) stay on the drop path: they don't declare the field and would divert it into `model_kwargs` and fail at request time, so the kwarg is still popped before construction.

`_normalize_openai_base_url`, `_enable_stream_usage_by_default`, and `_warn_unknown_model_settings` are intentionally left on their existing scoping — this PR only touches the `stream_chunk_timeout` heuristic. (`stream_usage` is already applied to all subclasses via the separate `model_fields`-based block further down in `create_chat_model`.)

`ChatDeepSeek` extends `BaseChatOpenAI` directly (not `ChatOpenAI`), which is why the gate targets `BaseChatOpenAI` rather than `ChatOpenAI`.

## Verification

Effective `stream_chunk_timeout` the constructed model ends up with, from the real `create_chat_model()`:

| Provider | before (unset) | before (config: 300) | after (unset) | after (config: 300) |
|---|---|---|---|---|
| `ChatOpenAI` / `PatchedChatOpenAI` | 240 | 300 | 240 | 300 |
| `VllmChatModel` | 120 (langchain default) | 120 (override dropped) | **240** | **300** |
| `MindIEChatModel` | 120 (langchain default) | 120 (override dropped) | **240** | **300** |
| `PatchedChatDeepSeek` | 120 (langchain default) | 120 (override dropped) | **240** | **300** |
| `PatchedChatMiMo` | 120 (langchain default) | 120 (override dropped) | **240** | **300** |
| `PatchedChatStepFun` | 120 (langchain default) | 120 (override dropped) | **240** | **300** |
| `PatchedChatMiniMax` | 120 (langchain default) | 120 (override dropped) | **240** | **300** |
| `ChatAnthropic` (non-OpenAI) | not injected | dropped (correct) | not injected | dropped (correct) |

Tests (`backend/tests/test_model_factory.py`):

- New parametrized regression over all six subclasses asserting the 240s default when unset **and** that an explicit `stream_chunk_timeout: 300` survives.
- New end-to-end anchor that builds the real `PatchedChatMiMo` (issue #3189's model) and asserts the constructed instance's `stream_chunk_timeout == 240.0` — i.e. accepted as a real field, not diverted into `model_kwargs`.
- Reverting the factory change makes the 13 new subclass assertions fail (`PatchedChatMiMo` resolves to `120.0` instead of `240.0`); re-applying it makes them pass. The existing `ChatOpenAI` / `ChatAnthropic` cases were updated to resolve the real provider classes so the `issubclass` gate is genuinely exercised.
- `test_model_factory.py` + the provider suites (`test_vllm_provider`, `test_mindie_provider`, `test_patched_{deepseek,mimo,minimax,stepfun}`, `test_codex_provider`, `test_claude_provider_*`, `test_model_config`, `test_tracing_factory`, `test_lead_agent_model_resolution`, `test_scheduled_task_models`, `test_token_usage_by_model`): **245 passed**.

### DIFF
--- a/backend/packages/harness/deerflow/models/factory.py
+++ b/backend/packages/harness/deerflow/models/factory.py
@@ -1,6 +1,7 @@
 import logging
 
 from langchain.chat_models import BaseChatModel
+from langchain_openai.chat_models.base import BaseChatOpenAI
 
 from deerflow.config import get_app_config
 from deerflow.config.app_config import AppConfig
@@ -141,20 +142,33 @@ def _warn_unknown_model_settings(model_use_path: str, model_class, model_name: s
 _DEFAULT_STREAM_CHUNK_TIMEOUT_SECONDS: float = 240.0
 
 
-def _apply_stream_chunk_timeout_default(model_use_path: str, model_settings_from_config: dict) -> None:
+def _apply_stream_chunk_timeout_default(model_class: type, model_settings_from_config: dict) -> None:
     """Inject a generous ``stream_chunk_timeout`` for OpenAI-compatible clients.
 
-    The ``stream_chunk_timeout`` kwarg is specific to ``langchain_openai:ChatOpenAI``
-    and is rejected by other providers' constructors as an unexpected keyword
-    argument. Behaviour:
+    ``stream_chunk_timeout`` is a field of langchain-openai's ``BaseChatOpenAI``, so
+    it is accepted by ``ChatOpenAI`` and by every DeerFlow provider that subclasses
+    it: ``PatchedChatOpenAI`` plus the self-hosted / reasoning adapters
+    ``VllmChatModel``, ``MindIEChatModel``, ``PatchedChatDeepSeek``,
+    ``PatchedChatMiMo``, ``PatchedChatStepFun`` and ``PatchedChatMiniMax``. We gate on
+    ``issubclass(model_class, BaseChatOpenAI)`` rather than an explicit class-path
+    allowlist so any OpenAI-compatible subclass inherits the default (and honors an
+    explicit override) automatically. Issue #3189 was reported against ``mimo-v2.5``
+    (``PatchedChatMiMo``); the original fix (#3195) matched only ``ChatOpenAI`` /
+    ``PatchedChatOpenAI``, so those subclasses kept langchain-openai's aggressive
+    built-in chunk-gap timeout and — worse — silently discarded a user's explicit
+    ``stream_chunk_timeout``.
 
-    * OpenAI-compatible path: an explicit value in ``config.yaml`` is preserved.
+    Behaviour:
+
+    * ``BaseChatOpenAI`` subclass: an explicit value in ``config.yaml`` is preserved.
       An explicit ``null`` is dropped upstream by ``model_dump(exclude_none=True)``
       and therefore treated as "unset", so the default is injected.
-    * Non-OpenAI path: drop the key so it is never forwarded to an incompatible
-      constructor (which would raise ``TypeError: unexpected keyword argument``).
+    * Any other client (e.g. ``ChatAnthropic``): drop the key so it is never
+      forwarded to a constructor that does not declare it. Such clients do not reject
+      the unknown kwarg outright — they divert it into ``model_kwargs`` and then fail
+      at request time.
     """
-    if model_use_path not in _OPENAI_COMPAT_USE_PATHS:
+    if not (isinstance(model_class, type) and issubclass(model_class, BaseChatOpenAI)):
         model_settings_from_config.pop("stream_chunk_timeout", None)
         return
     if "stream_chunk_timeout" in model_settings_from_config:
@@ -250,7 +264,7 @@ def create_chat_model(name: str | None = None, thinking_enabled: bool = False, *
     # heuristics (stream_usage / stream_chunk_timeout) see the canonical endpoint key.
     _normalize_openai_base_url(model_config.use, model_settings_from_config)
     _enable_stream_usage_by_default(model_config.use, model_settings_from_config)
-    _apply_stream_chunk_timeout_default(model_config.use, model_settings_from_config)
+    _apply_stream_chunk_timeout_default(model_class, model_settings_from_config)
 
     # For Codex Responses API models: map thinking mode to reasoning_effort
     from deerflow.models.openai_codex_provider import CodexChatModel
--- a/backend/packages/harness/deerflow/models/factory.py
+++ b/backend/packages/harness/deerflow/models/factory.py
@@ -134,7 +134,7 @@ def _warn_unknown_model_settings(model_use_path: str, model_class, model_name: s
 # Default chunk-gap budget for OpenAI-compatible streaming responses.
 #
 # langchain-openai raises ``StreamChunkTimeoutError`` after this many seconds
-# without receiving a chunk. Its own default is 60s, which is too aggressive for
+# without receiving a chunk. Its own default is 120s, which is too aggressive for
 # reasoning models (DeepSeek-R1, Doubao-thinking, GPT-5) whose first chunk can
 # legitimately take 90~150s. We default to 240s so the streaming layer rarely
 # trips on long thinking pauses; the LLMErrorHandlingMiddleware still retries
@@ -164,11 +164,14 @@ def _apply_stream_chunk_timeout_default(model_class: type, model_settings_from_c
       An explicit ``null`` is dropped upstream by ``model_dump(exclude_none=True)``
       and therefore treated as "unset", so the default is injected.
     * Any other client (e.g. ``ChatAnthropic``): drop the key so it is never
-      forwarded to a constructor that does not declare it. Such clients do not reject
-      the unknown kwarg outright — they divert it into ``model_kwargs`` and then fail
-      at request time.
+      forwarded to a constructor that does not declare it. The kwarg is not a
+      declared field of these clients: depending on the client it is either
+      silently dropped (``ChatAnthropic`` declares ``extra="ignore"``) or, for
+      other OpenAI-style clients, diverted into ``model_kwargs`` and rejected
+      at request time. Either way the user's intent is lost, so we drop it
+      proactively instead.
     """
-    if not (isinstance(model_class, type) and issubclass(model_class, BaseChatOpenAI)):
+    if not issubclass(model_class, BaseChatOpenAI):
         model_settings_from_config.pop("stream_chunk_timeout", None)
         return
     if "stream_chunk_timeout" in model_settings_from_config:
--- a/backend/tests/test_model_factory.py
+++ b/backend/tests/test_model_factory.py
@@ -10,6 +10,7 @@ from deerflow.config.model_config import ModelConfig
 from deerflow.config.sandbox_config import SandboxConfig
 from deerflow.models import factory as factory_module
 from deerflow.models import openai_codex_provider as codex_provider_module
+from deerflow.reflection import resolve_class
 
 # ---------------------------------------------------------------------------
 # Helpers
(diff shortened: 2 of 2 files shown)


backend/packages/harness/deerflow/models/factory.py
```diff
@@ -250,7 +264,7 @@ def create_chat_model(name: str | None = None, thinking_enabled: bool = False, *
     # heuristics (stream_usage / stream_chunk_timeout) see the canonical endpoint key.
     _normalize_openai_base_url(model_config.use, model_settings_from_config)
     _enable_stream_usage_by_default(model_config.use, model_settings_from_config)
-    _apply_stream_chunk_timeout_default(model_config.use, model_settings_from_config)
+    _apply_stream_chunk_timeout_default(model_class, model_settings_from_config)
```
willem-bd, line 270: This is the one helper that now gates on `issubclass(model_class, BaseChatOpenAI)`, but the two helpers just above it (`_normalize_openai_base_url`, `_enable_stream_usage_by_default`) still use the `_OPENAI_COMPAT_USE_PATHS` string allowlist.

`stream_usage` is fine - the separate `model_fields`-based block further down in `create_chat_model` already injects it for every `BaseChatOpenAI` subclass. But `_normalize_openai_base_url`'s `api_base` -> `base_url` rename does **not** run for the six subclasses this PR newly covers, so a user who sets `api_base` on e.g. `PatchedChatMiMo` would still have it diverted into `model_kwargs` and fail at request time with `unexpected keyword argument 'api_base'`.

Clearly out of scope for this PR (correctly focused on `stream_chunk_timeout`), but a natural follow-up would be to move those two helpers onto the same `issubclass(model_class, BaseChatOpenAI)` gate for consistency.

backend/packages/harness/deerflow/models/factory.py
```diff
@@ -141,20 +142,33 @@ def _warn_unknown_model_settings(model_use_path: str, model_class, model_name: s
...
+    it is accepted by ``ChatOpenAI`` and by every DeerFlow provider that subclasses
+    it: ``PatchedChatOpenAI`` plus the self-hosted / reasoning adapters
+    ``VllmChatModel``, ``MindIEChatModel``, ``PatchedChatDeepSeek``,
+    ``PatchedChatMiMo``, ``PatchedChatStepFun`` and ``PatchedChatMiniMax``. We gate on
+    ``issubclass(model_class, BaseChatOpenAI)`` rather than an explicit class-path
+    allowlist so any OpenAI-compatible subclass inherits the default (and honors an
+    explicit override) automatically. Issue #3189 was reported against ``mimo-v2.5``
+    (``PatchedChatMiMo``); the original fix (#3195) matched only ``ChatOpenAI`` /
+    ``PatchedChatOpenAI``, so those subclasses kept langchain-openai's aggressive
+    built-in chunk-gap timeout and — worse — silently discarded a user's explicit
```
willem-bd, line 158: Nit (pre-existing, but you're already correcting the same number nearby): the module-level comment above `_DEFAULT_STREAM_CHUNK_TIMEOUT_SECONDS` (line 137) says langchain-openai's "own default is 60s". The actual built-in default is **120s** - the field is declared as `default_factory=lambda: _float_env("LANGCHAIN_OPENAI_STREAM_CHUNK_TIMEOUT_S", 120.0)`.

This PR already drops the stale "60s" claim from the rewritten test docstring; worth fixing the same value in that module-level comment block too.

backend/packages/harness/deerflow/models/factory.py
```diff
@@ -141,20 +142,33 @@ def _warn_unknown_model_settings(model_use_path: str, model_class, model_name: s
 _DEFAULT_STREAM_CHUNK_TIMEOUT_SECONDS: float = 240.0
```
willem-bd, line 142: **Nit (pre-existing, in-scope):** the comment block just above this constant (lines 133-141) says langchain-openai's built-in default is **60s** ("Its own default is 60s"). The actual default is **120s** - `BaseChatOpenAI.stream_chunk_timeout` is a `Field` whose default_factory reads `LANGCHAIN_OPENAI_STREAM_CHUNK_TIMEOUT_S` (default `120.0`), which also matches the value in the #3189 traceback and this PR's description. Since this PR is correcting exactly this area, worth fixing that comment to 120s here so the code stops contradicting the PR body.

backend/packages/harness/deerflow/models/factory.py
```diff
@@ -141,20 +142,33 @@ def _warn_unknown_model_settings(model_use_path: str, model_class, model_name: s
...
+    Behaviour:
+
+    * ``BaseChatOpenAI`` subclass: an explicit value in ``config.yaml`` is preserved.
       An explicit ``null`` is dropped upstream by ``model_dump(exclude_none=True)``
       and therefore treated as "unset", so the default is injected.
-    * Non-OpenAI path: drop the key so it is never forwarded to an incompatible
-      constructor (which would raise ``TypeError: unexpected keyword argument``).
+    * Any other client (e.g. ``ChatAnthropic``): drop the key so it is never
+      forwarded to a constructor that does not declare it. Such clients do not reject
+      the unknown kwarg outright — they divert it into ``model_kwargs`` and then fail
```
willem-bd: **Nit:** the "divert into `model_kwargs` and fail at request time" failure mode is specific to OpenAI clients (it's what `_warn_unknown_model_settings` guards against). `ChatAnthropic` - the example cited here - uses `extra="ignore"`, so it would *silently drop* an unknown kwarg rather than divert-and-crash. The pop is still correct (avoids silently swallowing a user's intended setting), but the cited example doesn't match the described failure mode. Suggest softening to e.g. "the kwarg is not a declared field of these clients (silently dropped, or for OpenAI-style clients diverted into `model_kwargs`); either way the user's intent is lost."

backend/packages/harness/deerflow/models/factory.py
```diff
@@ -141,20 +142,33 @@ def _warn_unknown_model_settings(model_use_path: str, model_class, model_name: s
...
+      at request time.
     """
-    if model_use_path not in _OPENAI_COMPAT_USE_PATHS:
+    if not (isinstance(model_class, type) and issubclass(model_class, BaseChatOpenAI)):
```
willem-bd: **Nit:** `model_class` comes from `resolve_class(model_config.use, BaseChatModel)`, which always returns a type - so `isinstance(model_class, type)` is always `True` and the `and` short-circuits to just the `issubclass` check. Harmless belt-and-suspenders, but it reads as guarding against a state `resolve_class` can't produce. Could simplify to `if not issubclass(model_class, BaseChatOpenAI):`.

backend/packages/harness/deerflow/models/factory.py
willem-bd, line 270: **Observation / follow-up:** now that `stream_chunk_timeout` gates on `issubclass(model_class, BaseChatOpenAI)`, the two sibling helpers invoked just above (`_normalize_openai_base_url` / `_enable_stream_usage_by_default`, lines 265-266) still gate on the 2-class `_OPENAI_COMPAT_USE_PATHS` string list, so the 6 newly-covered providers won't get `api_base`->`base_url` aliasing or unknown-key typo warnings. The narrow scope is defensible (`stream_usage` is already class-based at lines 296-297), but worth a follow-up to reconsider whether the same `issubclass` rationale should extend to those two - `api_base` aliasing in particular deserves a closer look since `ChatDeepSeek` may have its own endpoint-key handling.

willem-bd (commented):
Verified the core thesis against the installed code: all six "excluded" providers (`PatchedChatMiMo`, `PatchedChatStepFun`, `PatchedChatMiniMax`, `VllmChatModel`, `MindIEChatModel` via `ChatOpenAI`; `PatchedChatDeepSeek` via `ChatDeepSeek`) are genuinely `BaseChatOpenAI` subclasses, `stream_chunk_timeout` is a field of `BaseChatOpenAI` itself (so every subclass inherits it), and `ChatAnthropic` / `CodexChatModel` are correctly excluded. Gating on `issubclass(model_class, BaseChatOpenAI)` instead of the string allowlist is the right call - narrower, more maintainable, and covers the model #3189 was actually filed against.

The test refactor from `FakeChatModel`-based stubs to `_capturing_class(<real provider>, ...)` is a real improvement: the old stubs weren't `BaseChatOpenAI` subclasses, so only the string-path branch was ever exercised; the new ones genuinely exercise the `issubclass` gate. The e2e MiMo anchor (real constructor, asserts `instance.stream_chunk_timeout == 240.0`) confirms the value is accepted as a field rather than diverted into `model_kwargs`. CI green.

Recommend **approve with 4 minor nits** below.

DaoyuanLi2816:
Thanks for the detailed review, @willem-bd — pushed 25eea76c addressing the nits.

1. **Fixed** — corrected the module comment above `_DEFAULT_STREAM_CHUNK_TIMEOUT_SECONDS`: langchain-openai's built-in default is 120s (`BaseChatOpenAI.stream_chunk_timeout`'s `default_factory` reads `LANGCHAIN_OPENAI_STREAM_CHUNK_TIMEOUT_S` with a `120.0` fallback), not 60s.
2. **Fixed** — simplified the gate to `if not issubclass(model_class, BaseChatOpenAI):`. Verified the `isinstance(model_class, type)` half is always true in practice: the function has exactly one call site, and the `model_class` passed in there comes from `resolve_class()`, which itself raises `ValueError` before ever returning anything that isn't a `type`. So the extra check was guarding against a state that can't occur.
3. **Fixed** — reworded the "any other client" bullet so it no longer asserts a single failure mode for all non-`BaseChatOpenAI` clients. It now says the kwarg isn't a declared field of these clients, and depending on the client it's either silently dropped (`ChatAnthropic`'s `extra="ignore"`) or, for other OpenAI-style clients, diverted into `model_kwargs` and rejected at request time — either way the user's intent is lost.
4. **Acknowledged, not implemented here** — agreed that bringing `_normalize_openai_base_url` / `_enable_stream_usage_by_default` onto the same `issubclass(model_class, BaseChatOpenAI)` gate is a reasonable next step, particularly the `api_base` aliasing gap you called out. Keeping it out of this PR's scope; can follow up with a separate PR for that if useful.

Reran the full `test_model_factory.py` suite (66 tests) plus `test_lead_agent_model_resolution.py` (21 tests) after both changes — all passing, and `ruff check` / `ruff format --check` are clean on the touched file.

Appreciate you taking the time to dig into this in such detail — genuinely useful review, nits included. Noting for the record that it doesn't formally satisfy the repo's merge/review-required gate on its own, but the technical substance holds regardless.

WillemJiang (approved):

PR #4103, opened 2026-07-11
fix(skills): activate a slash skill once per run, not per model call

DaoyuanLi2816:
## Summary

`SkillActivationMiddleware` is meant to activate a `/skill-name` command **once per run** — its own comment describes "the single activation call" followed by the many model calls a tool loop issues afterward (#3861 semantics). In practice it re-activates on **every** model call of a turn: each call re-parses the slash command, re-reads `SKILL.md` from disk, re-injects the full (multi-KB) skill body, and re-records an `activate` audit event.

## The bug

The activation reminder is added to the model request with `request.override(messages=...)`:

```python
messages = list(request.messages)
messages.insert(target_index, activation_msg)
return request.override(messages=messages), activation
```

LangChain's `create_agent` uses that override for a **single** `model.invoke()` and never writes it back to graph state — `model_node` rebuilds the request from persisted state on every step, and only the model's output is persisted:

```python
# langchain/agents/factory.py
request = ModelRequest(..., messages=state["messages"], state=state, runtime=runtime)
...
result = wrap_model_call_handler(request, _execute_model_sync)
return _build_commands(result.model_response, result.commands)
```

So on the 2nd..Nth model call of a turn the state is `[user, ai(tool_call), tool_result, ...]` — the injected reminder is gone. But the dedup guard only scans `request.messages` for a prior reminder:

```python
def _has_existing_activation_for_target(messages, target_index, target):
    ...
    previous = messages[target_index - 1]
    return is_slash_skill_activation_reminder(previous)
```

Because the reminder is never persisted, that scan can never match on a follow-up call, so `_find_activation_target` re-activates the same slash command every model call.

Reproduced with a real `create_agent` tool loop (scripted model: one tool call ⇒ two model-node invocations in one turn):

```
model-node invocations in ONE turn   : 2
skill disk reads                     : 2
'activate' audit events              : 2
activation-reminder injected per call: [1, 1]
activation reminder in FINAL state?  : False   # confirms it is never persisted
```

## The fix

Key the dedup off `runtime.context`, which LangGraph threads through every model-node call of a run — the same durable signal the request-scoped secret source (`__slash_skill_secret_source`) already uses to keep a slash skill "bound for the rest of the run". The activation call records the slash message's identity in the run context; later calls for the same message skip the redundant re-activation:

```python
# skip if this exact slash message already activated earlier in the run
if isinstance(run_context, dict) and run_context.get(_SLASH_SKILL_ACTIVATION_RUN_KEY) == self._activation_run_key(target):
    return None
...
# on the activation call:
run_context[_SLASH_SKILL_ACTIVATION_RUN_KEY] = self._activation_run_key(target)
```

The key is the user message id (LangGraph assigns a stable id once a message is in state), falling back to a digest of the genuine user text. A **new** user slash command keys differently and still activates. The existing `request.messages` scan is kept for defense in depth. The new context key is registered in `REDACTED_CONTEXT_KEYS` alongside the other middleware run-context keys (it holds a message id / digest, never a secret value).

Secret binding is unchanged: `_resolve_secret_bindings` still runs on every call and re-resolves the injection set from the persisted slash source, so a `required-secrets` skill stays bound across the tool loop even though activation is now skipped after the first call.

After the fix, the same real tool-loop reproduction shows:

```
model-node invocations in ONE turn   : 2
skill disk reads                     : 1
'activate' audit events              : 1
activation-reminder injected per call: [1, 0]
```

## Tests

- `test_skill_activation_middleware_activates_once_across_tool_loop` — rebuilds the **real** 2nd-call state `[user, ai(tool_call), tool_result]` (not a hand-constructed reminder-in-input) with one shared run context, and asserts the reminder is injected on call 1 only, the skill is read from disk once, and exactly one `activate` audit event fires across the turn.
- `test_skill_activation_middleware_reactivates_on_new_user_slash_command` — a second, different slash command in the same run context still activates (two activations recorded), proving the dedup does not over-suppress a genuinely new activation.

Verified fail-before / pass-after by reverting the source fix: `test_..._activates_once_across_tool_loop` re-injects on the 2nd call and fails without the fix, and passes with it. `test_slash_skills.py` and `test_slash_skill_contract.py` pass; `ruff check --line-length 240` and `ruff format --check --line-length 240` are clean.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
@@ -235,7 +235,22 @@ Follow this skill before choosing a general workflow. Load supporting resources 
         context = getattr(runtime, "context", None)
         return context if isinstance(context, dict) else None
 
-    def _find_activation_target(self, messages: list, *, run_context: dict | None = None) -> tuple[int, HumanMessage, _ActivationResolution] | None:
+    @staticmethod
+    def _already_activated(run_context: dict | None, run_key: str) -> bool:
+        """Whether ``run_key`` was already recorded as activated earlier in this run.
+
+        Sibling to ``_has_existing_activation_for_target``: that helper catches an
+        activation reminder still present in the scanned ``messages`` window; this
+        one catches a prior activation recorded on ``run_context`` whose reminder
+        already fell out of that window (the tool-loop case — see
+        ``_SLASH_SKILL_ACTIVATION_RUN_KEY``). ``run_key`` is computed once by the
+        caller (``_find_activation_target``) and reused as-is at the write site in
+        ``_prepare_model_request``, so the same key is always used to check and to
+        record — this helper only ever checks membership, never computes the key.
+        """
+        return isinstance(run_context, dict) and run_context.get(_SLASH_SKILL_ACTIVATION_RUN_KEY) == run_key
+
+    def _find_activation_target(self, messages: list, *, run_context: dict | None = None) -> tuple[int, HumanMessage, _ActivationResolution, str] | None:
         if not messages:
             return None
 
@@ -248,19 +263,22 @@ Follow this skill before choosing a general workflow. Load supporting resources 
             return None
         if self._has_existing_activation_for_target(messages, target_index, target):
             return None
-        # This exact slash message already activated earlier in the run. The message
-        # scan above cannot catch it because the reminder lives only in a per-call
-        # request override, never in state; the run context is the durable signal
-        # (see _SLASH_SKILL_ACTIVATION_RUN_KEY). Skipping here avoids the redundant
-        # skill disk read, reminder re-injection, and duplicate "activate" audit.
-        if isinstance(run_context, dict) and run_context.get(_SLASH_SKILL_ACTIVATION_RUN_KEY) == self._activation_run_key(target):
+        # This exact slash message may have already activated earlier in the run.
+        # The message scan above cannot catch it because the reminder lives only in
+        # a per-call request override, never in state — the run context is the
+        # durable signal (see _already_activated / _SLASH_SKILL_ACTIVATION_RUN_KEY).
+        # Skipping here avoids the redundant skill disk read, reminder re-injection,
+        # and duplicate "activate" audit. run_key is computed once here and threaded
+        # through to the write site in _prepare_model_request.
+        run_key = self._activation_run_key(target)
+        if self._already_activated(run_context, run_key):
             return None
 
         content = get_original_user_content_text(target.content, target.additional_kwargs)
         resolution = self._resolve_activation(content)
         if resolution is None:
             return None
-        return target_index, target, resolution
+        return target_index, target, resolution, run_key
 
     @staticmethod
     def _record_activation(request: ModelRequest, activation: _Activation, *, hook: str) -> None:
@@ -291,7 +309,7 @@ Follow this skill before choosing a general workflow. Load supporting resources 
         if target_and_resolution is None:
             return None, None
 
-        target_index, target, resolution = target_and_resolution
+        target_index, target, resolution, run_key = target_and_resolution
         if resolution.failure_message:
             return AIMessage(content=resolution.failure_message), None
 
@@ -310,9 +328,14 @@ Follow this skill before choosing a general workflow. Load supporting resources 
         # Mark this slash message as activated for the run so the tool loop's later
         # model calls skip the redundant re-activation (#3861: one activation call,
         # many follow-up model calls). A new user slash message keys differently and
-        # still activates.
+        # still activates. Overwrite (`=`), not append/accumulate, is intentional:
+        # _find_activation_target only ever considers the latest real user message as
+        # an activation target, so there is nothing earlier in the run worth
+        # remembering once a new activation replaces it — do not "fix" this into a
+        # set. run_key is the same value already checked in _find_activation_target
+        # (computed once there, threaded through here) rather than recomputed.
         if run_context is not None:
-            run_context[_SLASH_SKILL_ACTIVATION_RUN_KEY] = self._activation_run_key(target)
+            run_context[_SLASH_SKILL_ACTIVATION_RUN_KEY] = run_key
         activation_msg = self._make_activation_message(target, self._build_activation_reminder(activation))
         messages = list(request.messages)
         messages.insert(target_index, activation_msg)
--- a/backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
@@ -20,6 +20,7 @@ from langchain_core.messages import AIMessage, HumanMessage
 from deerflow.runtime.secret_context import (
     _SECRETS_BINDING_AUDIT_KEY,
     _SLASH_SECRET_SOURCE_KEY,
+    _SLASH_SKILL_ACTIVATION_RUN_KEY,
     ACTIVE_SECRETS_CONTEXT_KEY,
     extract_request_secrets,
 )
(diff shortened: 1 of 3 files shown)


backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
```diff
@@ -220,6 +248,13 @@ def _find_activation_target(self, messages: list) -> tuple[int, HumanMessage, _A
             return None
         if self._has_existing_activation_for_target(messages, target_index, target):
             return None
+        # This exact slash message already activated earlier in the run. The message
+        # scan above cannot catch it because the reminder lives only in a per-call
+        # request override, never in state; the run context is the durable signal
+        # (see _SLASH_SKILL_ACTIVATION_RUN_KEY). Skipping here avoids the redundant
+        # skill disk read, reminder re-injection, and duplicate "activate" audit.
+        if isinstance(run_context, dict) and run_context.get(_SLASH_SKILL_ACTIVATION_RUN_KEY) == self._activation_run_key(target):
```
willem-bd: Minor readability nit (non-blocking): this guard reads a bit dense. A tiny `_already_activated(run_context, target) -> bool` helper would mirror the existing `_has_existing_activation_for_target` sibling and make the intent self-documenting. Also note `_activation_run_key(target)` is computed here for the check and again at the write site in `_prepare_model_request`; computing it once and threading it through would make the 'same key for check and write' invariant visually obvious.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
```diff
@@ -271,6 +307,12 @@ def _prepare_model_request(self, request: ModelRequest, *, hook: str) -> tuple[M
             activation.content_hash,
         )
         self._record_activation(request, activation, hook=hook)
+        # Mark this slash message as activated for the run so the tool loop's later
+        # model calls skip the redundant re-activation (#3861: one activation call,
+        # many follow-up model calls). A new user slash message keys differently and
+        # still activates.
+        if run_context is not None:
+            run_context[_SLASH_SKILL_ACTIVATION_RUN_KEY] = self._activation_run_key(target)
```
willem-bd: Non-blocking: worth a half-sentence noting the overwrite (`=`, not an append/set) is intentional — `_find_activation_target` only ever considers the *latest* real user message, so there is no need to remember prior activations. That stops a future reader from 'fixing' this into a set. (Minor: `_activation_run_key(target)` is recomputed here after already being computed in the dedup check above.)

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
```diff
@@ -207,7 +215,27 @@ def _has_existing_activation_for_target(messages: list, target_index: int, targe
...
+        so an id-less message still dedupes within a run. A new user slash message
+        (new id / new text) yields a new key, so it is not suppressed.
+        """
+        if target.id:
+            return target.id
+        content = get_original_user_content_text(target.content, target.additional_kwargs)
+        return "sha256:" + hashlib.sha256(content.encode("utf-8")).hexdigest()
+
+    @staticmethod
+    def _run_context(request: ModelRequest) -> dict | None:
```
willem-bd, line 233: Given the `risk:high` / `needs-validation` labels: when `runtime.context` is non-dict (e.g. the `None` case from #3989), this returns `None` and the middleware gracefully reverts to per-call activation — no worse than before, but there's no test locking that degraded-path contract. A one-liner regression anchor asserting it still activates (rather than silently no-op-ing or crashing) when context is `None` would be cheap insurance.

willem-bd (commented):
Reviewed the diff against head `f48b48e`.

**Verdict: approve / merge.** The diagnosis is correct — the activation reminder lives only in a per-call `request.override(messages=...)` and is never persisted to graph state, so the existing message-scan dedup can't catch the 2nd..Nth model call of a tool loop. Keying the dedup off `runtime.context` is the right fix, and it rides the same durable channel `_SLASH_SECRET_SOURCE_KEY` already relies on (proven by the shipped secret-binding path).

Verified points:
- Secret binding is preserved — `_resolve_secret_bindings` still runs every call and re-resolves from the persisted slash source; skipping re-activation does not drop injection.
- The new context key holds only a message id/content digest (never a secret) and is registered in `REDACTED_CONTEXT_KEYS`.
- The regression test faithfully rebuilds the real 2nd-call state (`[user, ai(tool_call), tool_result]`) with a single shared run context, counts disk reads, and asserts exactly one `activate` event — fails without the fix, passes with it.
- Sync and async paths both route through `_handle_model_request` → `_prepare_model_request`, so both are covered.

Net effect is also a token-cost win (no redundant multi-KB skill re-injections per model call). The three inline comments below are all non-blocking nits/suggestions.

WillemJiang:
@DaoyuanLi2816 Thanks for the contribution. Could you take a look at the review comments?

DaoyuanLi2816:
@WillemJiang thanks for the nudge — @willem-bd's three suggestions are addressed in aea47e55:

1. **Extracted `_already_activated(run_context, run_key) -> bool`** (mirrors the existing `_has_existing_activation_for_target` sibling). `_activation_run_key(target)` is now computed once in `_find_activation_target` and threaded through `_prepare_model_request`, so the check site and the write site provably use the same key instead of two separate calls that happen to agree.
2. **Documented the overwrite** at the `run_context[_SLASH_SKILL_ACTIVATION_RUN_KEY] = run_key` write site: it's `=` not append/accumulate because `_find_activation_target` only ever considers the latest real user message as an activation target, so there's nothing earlier in the run worth remembering once a new activation replaces it.
3. **Added a regression test** for the degraded path: `runtime.context is None` still activates per-call (no crash, no wrong skip). To be transparent — this test passes against the pre-nit code too, since `_run_context`'s `isinstance(context, dict)` normalization already handled it correctly; there was no actual bug, just missing coverage for the contract, which is now locked in.

Full `test_slash_skills.py` (31 tests) plus `test_skill_request_scoped_secrets.py` (the other suite touching this middleware) pass, including both new tests and the existing tool-loop/reactivation regressions from the original PR. No behavior changes — this is a pure readability/test-coverage pass on top of the approved diff.

WillemJiang (approved):

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
DaoyuanLi2816: Implemented in aea47e55 (this PR's own follow-up commit, part of the squash-merge at 2fa05050): `_already_activated(run_context, run_key) -> bool` is now a static helper mirroring `_has_existing_activation_for_target`. `run_key = self._activation_run_key(target)` is computed once in `_find_activation_target`, returned in its result tuple, and reused as-is at the write site in `_prepare_model_request` instead of being recomputed there. Still green on current main (`test_slash_skills.py`, 31/32 passing — the one unrelated failure is a Windows-only `os.symlink` permission limitation in a different test). Resolving — this was addressed within the PR itself, the conversation was just never marked resolved.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
DaoyuanLi2816: Implemented in aea47e55 (this PR's own follow-up commit, part of the squash-merge at 2fa05050): the write site now carries a comment directly above `run_context[_SLASH_SKILL_ACTIVATION_RUN_KEY] = run_key` in `_prepare_model_request` explaining the overwrite (`=`, not append/accumulate) is intentional, since `_find_activation_target` only ever considers the latest real user message as an activation target, so there's nothing earlier in the run worth remembering once a new activation replaces it. Resolving — this was addressed within the PR itself, the conversation was just never marked resolved.

backend/packages/harness/deerflow/agents/middlewares/skill_activation_middleware.py
DaoyuanLi2816: Implemented in aea47e55 (this PR's own follow-up commit, part of the squash-merge at 2fa05050): `test_skill_activation_middleware_activates_per_call_when_run_context_is_none` in `backend/tests/test_slash_skills.py` drives `wrap_model_call` end-to-end with `runtime=SimpleNamespace(context=None)` and asserts the reminder is still injected (no crash, no silent skip) — locking in the degraded-path contract rather than just describing it. Confirmed still green on current main. Resolving — this was addressed within the PR itself, the conversation was just never marked resolved.

PR #4104, opened 2026-07-11
fix(channels): dedupe GitHub webhook redeliveries

DaoyuanLi2816:
## Summary

The inbound dedupe added for the IM channels in #3584 never reached the GitHub channel added later in #3754, so GitHub's native "Redeliver" button and retry-on-timeout re-run the agent — producing real duplicate side effects such as a second PR comment. This routes GitHub deliveries through the *same* `ChannelManager` dedupe the other channels already use, rather than adding a new mechanism.

## The bug

`ChannelManager._is_duplicate_inbound` keys on `_inbound_dedupe_key`, which reads a stable message id from the **top level** of `metadata` (or `metadata["raw_message"]`) plus a workspace id, and fails closed if either is missing:

```python
INBOUND_DEDUPE_METADATA_KEYS = ("event_id", "message_id", "msg_id")
...
for key in INBOUND_DEDUPE_METADATA_KEYS:
    value = metadata.get(key)          # top level only
    if value:
        message_id = str(value)
        break
...
workspace_id = msg.workspace_id or metadata.get("workspace_id") or metadata.get("team_id") or metadata.get("guild_id") or metadata.get("aibotid")
if not workspace_id:
    return None
```

Every IM channel was updated in #3584 to populate both halves (Slack `ts` + `team_id`, Telegram `message_id` + `workspace_id=chat_id`, WeChat/WeCom `message_id` + workspace). The GitHub channel added in #3754 populated neither:

- the `X-GitHub-Delivery` GUID is buried in `metadata["github"]["delivery_id"]`, which `_inbound_dedupe_key` never inspects;
- no `workspace_id` is set.

So `_inbound_dedupe_key` returns `None` for every GitHub delivery and `_is_duplicate_inbound` always returns `False`. GitHub reuses the same `X-GitHub-Delivery` GUID when it retries a timed-out delivery or when a delivery is replayed via the repo/App "Redeliver" button, so those replays fall straight through to a second agent run — and since GitHub agents write back via `gh` from their sandbox, the visible symptom is a duplicate PR comment / duplicated work.

Reproduced against `main` by driving the real `fanout_event` + real `ChannelManager` with two agents bound to one repo/event:

```
=== _is_duplicate_inbound (agent, phase, duplicate?) ===
   ('coder',    'd1',        False)
   ('reviewer', 'd1',        False)
   ('coder',    'redeliver', False)   # <- identical redelivery NOT deduped
   ('reviewer', 'redeliver', False)
GitHub: agent dispatched 4 times across delivery1 + identical redelivery (2 agents).
=== Control: slack redelivery ===
  slack dup?  1st=False  2nd=True     # <- the other channels already dedupe
```

## The fix

`fanout_event` now stamps the identity the manager already consumes:

```python
dedupe_message_id = f"{delivery_id}:{match.user_id}:{agent.name}" if delivery_id else None
msg = InboundMessage(
    ...
    workspace_id=repo,
    metadata={
        "message_id": dedupe_message_id,
        ...
    },
)
```

- `workspace_id = repo` satisfies the manager's fail-closed workspace requirement; the repo is globally unique and always present (mirrors Telegram/WeChat keying the workspace on the chat id).
- `message_id = f"{delivery_id}:{match.user_id}:{agent.name}"`. A single delivery fans out to N agents across potentially N different owning users, so the id is scoped to `(delivery, user, agent)`: an identical redelivery reproduces the same triples and is deduped, while two agents matching the same delivery — including two different users who each bind an agent of the same name to the same repo+event — keep distinct ids and both still fire. `ChannelManager._inbound_dedupe_key` indexes on `(channel, workspace_id, chat_id, message_id)`; workspace_id and chat_id are both the repo here, so `match.user_id` is the only thing that can separate two users in that key. `delivery_id` comes from an optional `Header(default=None)`, so when the header is absent the id is left `None` and the manager fails open exactly as before instead of collapsing distinct deliveries under a constant key.

With the fix the same reproduction dispatches the agent twice (once per agent) and drops the identical redelivery; a genuinely new delivery still fires.

## Tests

`backend/tests/test_github_dispatcher.py` (dispatcher layer):
- `test_delivery_id_populates_inbound_dedupe_identity` — `workspace_id` and top-level `message_id` are stamped where `_inbound_dedupe_key` looks.
- `test_dedupe_identity_stable_across_redelivery_and_distinct_per_agent` — same GUID → identical ids per agent; two agents on one delivery → distinct ids; new GUID → new ids.
- `test_missing_delivery_header_leaves_dedupe_open` — absent header yields no dedupe id (fail-open).

`backend/tests/test_channels.py` (manager layer):
- `test_github_redelivery_is_deduped_like_other_channels` — through the real `ChannelManager`, an identical redelivery dispatches once while a new delivery and a second agent on the same delivery still fire, and the dedupe key matches the other channels' 4-tuple.

Fail-before / pass-after verified by reverting only the dispatcher change: the three new dispatcher tests fail (`KeyError: 'message_id'` / missing `workspace_id`) without it and pass with it. GitHub fan-out suites: `95 passed` (`test_github_dispatcher` / `registry` / `triggers` / `identity` / `prompts`).

```
cd backend && PYTHONPATH=. pytest tests/test_github_dispatcher.py tests/test_channels.py
```

`ruff check` and `ruff format --check` are clean.


zhfeng (approved):
Good catch and it looks good to me. I bypass the redeliver message when debugging the github channel before. so it looks like I missed this case when landing it on the main branch. Thanks for fixing it!

Also it remind me that if the gateway is deployed as replica >= 2 in a k8s cluster for example and want to support HA. I think it could still be an issue if the redelivery messge route to a different pod even after the TTL 10min expired. It needs to investigate this case in the future. 

There is a PR which could resolve it partially. So it maybe similar improvement for the dedupe. And yeah, it denfinely should be with a follow up PR or issue.
- https://github.com/bytedance/deer-flow/pull/4003

zhfeng:
FYI @DaoyuanLi2816 would you like to take the following up issue as well? Thanks in advance!

backend/app/gateway/github/dispatcher.py
```diff
@@ -227,6 +227,21 @@ async def fanout_event(
...
+        # delivery is retried after a timeout or replayed via the repo/App
+        # "Redeliver" button, so keying on it lets the manager absorb those
+        # replays instead of re-running the agent (and its real side effects,
+        # e.g. a duplicate PR comment). One delivery fans out to N agents, so
+        # the per-message id is scoped to (delivery, agent): an identical
+        # redelivery reproduces the same pairs (deduped) while two agents
+        # matching the same delivery keep distinct ids and both still fire.
+        # Left None when the header is absent, so the manager fails open (no
+        # dedupe) exactly as before rather than collapsing distinct deliveries.
+        dedupe_message_id = f"{delivery_id}:{agent.name}" if delivery_id else None
```
willem-bd: The dedupe id isn't scoped by user. `_inbound_dedupe_key` returns `(channel, workspace_id, chat_id, message_id)` — i.e. `("github", repo, repo, f"{delivery_id}:{agent.name}")` — and `owner_user_id` isn't part of it.

The registry indexes by `(repo, event)` across **all** discovered users (`_discover_user_ids` → `lookup_agents`), so if two users each bind an agent named `reviewer` to the same repo+event, both fanned-out messages get an identical key and the manager silently drops the second user's run (`_is_duplicate_inbound` → `True` → `continue`). Before this PR GitHub had no dedupe at all, so both fired — this narrows that to a silent drop of the second user's run.

Low probability (needs multi-user + same agent name on a shared repo), but a real silent-drop regression. Suggested fix: fold `match.user_id` into the id — `f"{delivery_id}:{match.user_id}:{agent.name}"`. It's already in scope (used for `owner_user_id`), stays stable across a redelivery (same delivery + same config), and becomes distinct per `(user, agent)`.

backend/tests/test_github_dispatcher.py
```diff
@@ -945,3 +945,100 @@ async def test_coder_and_reviewer_on_same_pr_get_distinct_threads(base_dir: Path
...
+
+    # Workspace id: the manager fails closed without one; repo is stable + unique.
+    assert msg.workspace_id == "zhfeng/llm-gateway"
+    # Stable per-(delivery, agent) message id the manager keys dedupe on, read
+    # from the top level of metadata (not the nested ``github`` block).
+    assert msg.metadata["message_id"] == "del-abc:reviewer"
+
+
+@pytest.mark.asyncio
+async def test_dedupe_identity_stable_across_redelivery_and_distinct_per_agent(base_dir: Path) -> None:
```
willem-bd, line 1082: This asserts two **different** agent names stay distinct, but doesn't cover the cross-user collision: two users with the **same** agent name on the same repo+event would both produce `f"{delivery}:reviewer"` and the second would be dropped.

Once `user_id` is folded into the dedupe id, a test asserting two same-named agents across two users both fire would lock in the fix.

backend/app/gateway/github/dispatcher.py
zhfeng: Good catch and @DaoyuanLi2816 this must be fixed.

DaoyuanLi2816:
@willem-bd @zhfeng fixed in 304f25d9.

Confirmed the read first: `GitHubAgentMatch` (registry.py) does carry `user_id: str` per match, and `_inbound_dedupe_key` is exactly `(msg.channel_name, str(workspace_id), msg.chat_id, message_id)` with `workspace_id`/`chat_id` both set to the repo for GitHub — so `owner_user_id` was genuinely unrepresented in the key, as described.

Folded `match.user_id` into the id exactly as suggested: `f"{delivery_id}:{match.user_id}:{agent.name}"`.

Added `test_dedupe_identity_distinguishes_same_agent_name_across_users`, reproducing the exact scenario: two users each binding an agent named `reviewer` to the same repo+event. Verified it fails against the pre-fix code (reverted just the source line via a patch file to double check) — both fan-out messages get the identical id `del-cross-user:reviewer`, and `ChannelManager._is_duplicate_inbound` returns `True` on the second, i.e. the silent drop. After the fix the ids differ per-user and neither is flagged as a duplicate in the same dedupe window. A genuine retry (same user, same agent, same delivery) still produces the same id and is still deduped.

`test_dedupe_identity_stable_across_redelivery_and_distinct_per_agent` (two different agent names stay distinct) passes unchanged; `test_delivery_id_populates_inbound_dedupe_identity`'s literal id assertion is updated for the new `delivery:user:agent` shape. Full `test_github_dispatcher.py` (24 tests) plus `test_channels.py` / `test_github_channel.py` / `test_github_registry.py` (336 tests total) all pass.

Thanks for the catch, willem-bd — and thanks zhfeng.

backend/app/gateway/github/dispatcher.py
```diff
@@ -227,6 +227,27 @@ async def fanout_event(
...
+        # reproduces the same triples (deduped) while two agents matching the
+        # same delivery — including two different users who each bind an
+        # agent of the same name to the same repo+event — keep distinct ids
+        # and both still fire. ``ChannelManager._inbound_dedupe_key`` indexes
+        # on (channel, workspace_id, chat_id, message_id); workspace_id and
+        # chat_id are both the repo here, so match.user_id is the only thing
+        # that can separate two users in that key (willem-bd, PR #4104).
+        # Left None when the header is absent, so the manager fails open (no
+        # dedupe) exactly as before rather than collapsing distinct deliveries.
+        dedupe_message_id = f"{delivery_id}:{match.user_id}:{agent.name}" if delivery_id else None
```
willem-bd, line 422: **PR body shows a different key than the code.** The body's "## The fix" block shows `dedupe_message_id = f"{delivery_id}:{agent.name}"` and describes the id as "scoped to (delivery, agent)". The actual code here is `f"{delivery_id}:{match.user_id}:{agent.name}"` - scoped to `(delivery, user, agent)`. The `match.user_id` dimension was added (see `test_dedupe_identity_distinguishes_same_agent_name_across_users`) to fix a real cross-user collision: two users each binding an agent named `reviewer` to the same repo+event would otherwise produce identical ids and the second user's run would be silently dropped as a false-positive duplicate. The body should be updated to the 3-part key so reviewers don't miss this dimension.

backend/app/gateway/github/dispatcher.py
```diff
@@ -237,7 +258,15 @@ async def fanout_event(
             # owner_user_id drives which user bucket the run executes in
             # (custom agent lookup, sandbox, memory).
             owner_user_id=match.user_id,
+            # Tenant key the manager's dedupe requires (it fails closed without
+            # one, to avoid collapsing two workspaces). The repo is globally
+            # unique and always present — mirrors Telegram/WeChat keying the
+            # workspace on the chat id.
+            workspace_id=repo,
```
willem-bd, line 437: **"Redeliver" coverage is bounded by a 10-min in-memory TTL.** `ChannelManager._recent_inbound_events` is an in-memory dict (`INBOUND_DEDUPE_TTL_SECONDS = 600`, 4096-entry cap) and is not persisted to `ChannelStore`. GitHub's immediate timeout-retries (the common duplicate-comment cause) are correctly deduped, but a manual "Redeliver" clicked >10 min after the original - or any redelivery after a Gateway restart - still re-runs the agent and re-posts the duplicate. This is parity with the other IM channels and strictly better than `main`, but the body's implication that "Redeliver" is handled is only true within the 10-min window. Worth documenting the window in the body, or persisting the dedupe key in `ChannelStore` if true idempotency against a late/manual "Redeliver" is a goal.

backend/tests/test_channels.py
```diff
@@ -1039,6 +1039,45 @@ def test_inbound_dedupe_key_fails_closed_without_workspace(self):
...
+        def _gh(delivery: str, agent: str = "reviewer") -> InboundMessage:
+            # Shaped exactly as app.gateway.github.dispatcher.fanout_event emits.
+            return InboundMessage(
+                channel_name="github",
+                chat_id="zhfeng/llm-gateway",
+                user_id="alice",
+                text="@bot please review",
+                topic_id=f"7:{agent}",
+                workspace_id="zhfeng/llm-gateway",
+                metadata={"message_id": f"{delivery}:{agent}", "agent_name": agent},
```
willem-bd, line 1067: **Nit - helper id shape doesn't match production.** This stamps a 2-part id `f"{delivery}:{agent}"`, but the dispatcher emits a 3-part `f"{delivery}:{user}:{agent}"`. The manager only compares for equality so the test still passes, but it wouldn't catch a regression that drops `match.user_id` (covered separately in `test_github_dispatcher.py`). Mirroring the real shape here would make the manager-level test reflect production.

backend/tests/test_github_dispatcher.py
```diff
@@ -945,3 +945,162 @@ async def test_coder_and_reviewer_on_same_pr_get_distinct_threads(base_dir: Path
...
+    _write_agent(base_dir, "default", "reviewer", {"name": "reviewer", "github": {"bindings": [{"repo": "a/b", "triggers": {"pull_request": {"actions": ["opened"]}}}]}})
+    payload = {
+        "action": "opened",
+        "pull_request": {"number": 1, "title": "x", "user": {"login": "u"}, "body": ""},
+        "repository": {"full_name": "a/b"},
+        "sender": {"login": "u"},
+    }
+    await fanout_event(bus, "pull_request", "", payload)
+    (msg,) = await _drain(bus)
+    assert msg.metadata["message_id"] is None
```
willem-bd, line 1195: **Nit - doesn't lock the manager-level fail-open.** This asserts `msg.metadata["message_id"] is None` at the dispatcher layer only. The behavior is correct today (`_inbound_dedupe_key` returns `None` when no id key is truthy -> `_is_duplicate_inbound` returns `False`), but a future change treating a missing id as a real key would slip through. One extra `manager._is_duplicate_inbound(msg) is False` assertion on two header-less deliveries would pin it.

backend/app/gateway/github/dispatcher.py
```diff
@@ -227,6 +227,27 @@ async def fanout_event(
         # (The store accepts a pre-known thread id; manager will fall
         # through to _create_thread() the very first time.)
         topic_id = f"{number}:{agent.name}"
+        # Inbound dedupe identity for ChannelManager._is_duplicate_inbound,
```
willem-bd, line 401: **Nit - comment length / attribution.** The 22-line block is justified by the cross-user subtlety, but could be trimmed by ~1/3. The inline `(willem-bd, PR #4104)` reviewer attribution is unusual in committed code - a pointer like `# cross-user: see test_dedupe_identity_distinguishes_same_agent_name_across_users` would age better than a handle.

willem-bd (commented):
Review of the GitHub webhook dedupe fix. Core logic is sound - verified fail-open on a missing `X-GitHub-Delivery` header, correct 3-part `(delivery, user, agent)` keying incl. the cross-user case, and it rides the existing release-on-failure path so transient errors aren't black-holed. Inline notes: (1) the PR body shows a stale 2-part key - update to match the code; (2) "Redeliver" coverage is bounded by the 10-min in-memory dedupe TTL - worth documenting or persisting; (3)(4)(5) test-shape and comment nits. Overall: approve with the body fix (#1) before merge. Posted as COMMENT (not blocking).

DaoyuanLi2816:
Thanks for the thorough re-review, @willem-bd — both addressed:

1. **Body fix**: updated the "## The fix" section to describe the actual shipped key, `f"{delivery_id}:{match.user_id}:{agent.name}"` scoped to `(delivery, user, agent)`, and explained why `match.user_id` is the only thing that can separate two users in `ChannelManager._inbound_dedupe_key`'s `(channel, workspace_id, chat_id, message_id)` tuple here (workspace_id and chat_id both being the repo). No longer describes the stale 2-part `(delivery, agent)` shape.
2. **Comment attribution**: swapped the inline `(willem-bd, PR #4104)` handle reference for a pointer to the regression test that locks in the cross-user behavior — `test_dedupe_identity_distinguishes_same_agent_name_across_users` — per your suggestion.

Pushed in 81afbcee. Full `test_github_dispatcher.py` (24 tests) and `test_channels.py` (243 tests) still green, `ruff check` / `ruff format --check` clean.

Left the TTL-documentation point and the two test-shape nits as-is for now since they're non-blocking per your review — happy to follow up separately if you'd like those tracked.

WillemJiang:
@DaoyuanLi2816  please resolve the conflict with the main branch.

DaoyuanLi2816:
@WillemJiang Resolved in 6e060549 -- merged main in. The only textual conflict was in `backend/tests/test_github_dispatcher.py`: this branch's 4 new inbound-dedupe-identity tests and main's already-merged additions (#4131's redundant-review-comment-suppression tests, #4055's bot_login tests) had both appended new sections at the same end-of-file anchor point. Kept both in full -- nothing dropped or altered from either side.

Verified: all 35 tests in `test_github_dispatcher.py` pass (24 from this branch + 11 new from main, zero duplicates/losses), the full `test_channels.py` + `test_github_prompts.py` suite (302 tests) is still green, and `ruff check` / `ruff format --check` are clean.

WillemJiang (approved):

PR #4105, opened 2026-07-11
fix(scheduler): close dispatch race that can launch two runs for one task

DaoyuanLi2816:
## Summary

`dispatch_task` used a non-atomic active-run check followed by a separate insert, so concurrent manual or scheduled dispatches could both launch for one task.

This change adds a partial unique index for one queued/running row per `task_id`. Migration `0007_scheduled_run_active_index` resolves legacy duplicates before creating the index. Losing active inserts become the existing manual-conflict or scheduled-skip outcomes, and scheduled skip records are inserted directly as terminal rows.

The index is intentionally status-based while `overlap_policy` only supports `skip`; any future non-skip policy must update the index and conflict handling together.

## Tests

Concurrency, repository, migration, and bootstrap coverage pass (64 tests); ruff check and format are clean.


backend/packages/harness/deerflow/persistence/scheduled_task_runs/sql.py
```diff
@@ -46,7 +67,18 @@ async def create(
...
+            try:
+                await session.commit()
+            except IntegrityError:
+                await session.rollback()
+                # Only active-status inserts can trip the partial unique index
+                # ``uq_scheduled_task_run_active``; a terminal-status row (e.g.
+                # a "skipped" tombstone) is outside its predicate and cannot
+                # conflict, so any IntegrityError there is a genuine fault and
+                # is re-raised untranslated.
+                if status in ACTIVE_RUN_STATUSES:
```
willem-bd, line 79: **Suggestion (non-blocking):** The `except IntegrityError` blanket-translates *any* active-status insert failure into `ActiveScheduledRunConflict`. That's safe today — the only constraint an active insert can violate is `uq_scheduled_task_run_active` (the PK is a fresh UUID per call, and the table has no FKs/CHECKs). The concern is forward-looking: if a later migration adds another constraint to `scheduled_task_runs`, an active-status insert hitting it would be silently misreported as a conflict and collapse to a skip/409 instead of surfacing the real fault.

For comparison, the runs-table repo's `create_run_atomic` lets `IntegrityError` propagate untranslated on the `reject` path (and raises `ConflictError` from an explicit SELECT-FOR-UPDATE check on the `interrupt` path), so it can't mask an unrelated failure.

Two options if you agree it's worth hardening:
- a one-line comment recording the "only the partial unique index can fire here" assumption, or
- a constraint-name check (`exc.orig.diag.constraint_name` on Postgres / the index name in the SQLite message) so only the partial-index violation is translated.

Fine to leave as-is given there are no other constraints today — flagging for the record.

backend/app/scheduler/service.py
```diff
@@ -94,28 +100,37 @@ async def dispatch_task(
...
+        try:
+            await self._task_run_repo.create(
+                run_record_id=task_run_id,
+                task_id=task["id"],
+                thread_id=execution_thread_id,
+                scheduled_for=now,
+                trigger=trigger,
+                status="queued",
+            )
+        except ActiveScheduledRunConflict:
```
willem-bd, line 127: This catch collapses the conflict to skip-policy outcomes (manual -> conflict, scheduled -> skipped) without re-checking `overlap_skip`. Correct for the MVP because `overlap_policy` is fixed to `"skip"` - and the index itself is status-only (not `overlap_policy`-conditional), so it only ever fires under skip today.

Flagging the coupling: when a non-skip policy is added, **both** this catch (needs to handle the non-skip case) **and** the index (must become conditional, e.g. `... AND overlap_policy = 'skip'`, which requires denormalizing `overlap_policy` onto the run row) must be updated together. The PR body and the model's `__table_args__` comment call this out, so it's discoverable - just confirming it's on the radar.

backend/tests/test_scheduled_task_dispatch_race.py
```diff
@@ -0,0 +1,219 @@
...
+        assert len(launched) == 1, launched
+        assert await _active_run_count(run_repo, "task-race-manual") == 1
+        # The manual loser records no run-history row (nothing was scheduled).
+        conflict = next(r for r in results if r["outcome"] == "conflict")
+        assert conflict["task_run_id"] is None
+    finally:
+        await close_engine()
+
+
+async def test_scheduled_and_manual_dispatch_launch_exactly_once(tmp_path):
```
willem-bd, line 128: This barrier test correctly forces both dispatches past `has_active_runs` (so at least one hits the `ActiveScheduledRunConflict` path), but `set(outcomes) <= {"launched", "conflict", "skipped"}` allows either loser - so the scheduled-loser -> `skipped`-via-`ActiveScheduledRunConflict` branch isn't deterministically exercised (it depends on which dispatch wins the race). The manual-loser -> `conflict` branch IS deterministically pinned by `test_two_concurrent_manual_dispatches_launch_exactly_once`, and the index-semantics test covers the `ActiveScheduledRunConflict` raise directly - so coverage is adequate. Minor; a barrier variant that forces the scheduled dispatch to lose would close the gap if deterministic coverage of that branch is wanted.

willem-bd (commented):
## Review summary

Solid fix for a real concurrency bug - the partial unique index `uq_scheduled_task_run_active` makes the DB the atomic arbiter for "at most one active run per task," mirroring the proven #4003 runs-table pattern. I verified the full `dispatch_task` flow at head: the fast-path and race-path collapse to byte-identical outcomes (shared constants), the skip-tombstone is correctly created as terminal `skipped` (outside the index predicate, so it can't self-conflict), the `queued -> running` same-row UPDATE is index-safe, and the migration's dedupe pass + idempotent index creation are correct (standard SQL snapshot semantics, typed `finished_at` bindparam for Py3.12). The `IntegrityError -> ActiveScheduledRunConflict` translation is safe given the schema has no FK / other unique constraint on `scheduled_task_runs` (only the new partial index + a uuid4 PK). Test coverage is strong: barrier-forced race + natural-timing + index-semantics + migration dedupe.

### Finding 1 (residual, pre-existing) - status-update-failure bypass

The index closes the concurrent-`create` TOCTOU race this PR targets, but a separate narrow path can still violate the "one active run per task" invariant: if `_launch_run` succeeds (the run is actually live, line 135) but the subsequent `queued -> running` transition (`update_status(status="running")`, line 162) raises, the `except Exception` handler marks the row terminal `failed` (line 202) - **freeing the slot while the run is still live**. A later dispatch then sees `has_active_runs=False` and launches a *second* run. This requires a transient DB blip in a specific window (`update_status(running)` fails but the failure-path `update_status(failed)` succeeds), the launch-then-update-status pattern predates this PR, and `mark_stale_active_runs` reconciles orphaned active rows on restart (limiting the blast radius to the live process) - so it's out of scope here. Flagging because it's a path that bypasses the invariant this PR establishes; worth a tracking issue if the invariant is meant to be airtight. (Not anchored inline - the failure handler is unchanged code outside the diff hunks.)

### Findings 2 & 3 (inline)

- **`overlap_policy` coupling** - see inline on `service.py:127`.
- **Test determinism** - see inline on `test_scheduled_task_dispatch_race.py:128`.

backend/packages/harness/deerflow/persistence/scheduled_task_runs/sql.py
DaoyuanLi2816: Recorded the current constraint assumption at the catch site: terminal inserts re-raise, while active inserts can only hit the partial active-run index in the present schema.

backend/app/scheduler/service.py
DaoyuanLi2816: Kept the status-only index while `skip` is the only policy and documented that the index and conflict handler must change together when another policy is added.

backend/tests/test_scheduled_task_dispatch_race.py
DaoyuanLi2816: Kept the mixed race assertion broad. The manual-loser test deterministically covers the conflict branch, and the repository test pins the active-insert conflict directly.

WillemJiang (approved):

WillemJiang (approved):

PR #4108, opened 2026-07-12
refactor(sandbox): give the host→virtual output-mask regex a single owner

AnnaSuSu:
## Why

Follow-up to #4053, where I said I would open this separately once that landed: https://github.com/bytedance/deer-flow/pull/4053#discussion_r3564413269

Two call sites rewrite host paths back to their virtual form in text that reaches the model, and each builds the same `escape(base) + boundary + tail` rule from its own copy:

```python
# local_sandbox.py::_reverse_output_patterns — bash output
re.compile(re.escape(resolved_local) + boundary + tail)

# tools.py::_compiled_mask_patterns — glob/grep/ls results
escaped = re.escape(variant).replace(r"\\", r"[/\\]")
re.compile(escaped + boundary + tail)
```

The duplication is not hypothetical — it has already produced two bugs. #4035 added the segment-boundary lookahead to the reverse patterns and missed the masking patterns; #4053 had to add the same boundary to the other copy two days later. Both feed the same downstream contract (a match that stops short of a real segment boundary is rewritten to a container path that forward resolution then refuses to map back), so a third divergence is a bug, not a style problem.

This is @willem-bd's suggestion on #4053 ("Extracting a shared helper would make a future third drift impossible"), implemented.

## What changed

New leaf module `sandbox/path_patterns.py` exporting one function:

```python
def build_output_mask_pattern(base: str, *, separator_agnostic: bool = False) -> re.Pattern[str]
```

Both sites now call it. Neither file imports the other and both already depend on `sandbox/search.py`, so the new module introduces no cycle.

**The extraction is not a pure move**, which is the reason it did not ride along with #4053. The two sites agree on `boundary` and `tail` byte-for-byte, but not on the base:

| site | base source | separators |
|---|---|---|
| `tools.py` | `_path_variants()`, which yields Windows spellings; matched against output whose separators this layer does not control | `re.escape(base).replace(r"\\", r"[/\\]")` — accepts either |
| `local_sandbox.py` | `Path.resolve()`, which already carries the running platform's separator | plain `re.escape` |

A naive `build(base)` would silently widen one side or narrow the other. The difference is now an explicit `separator_agnostic` parameter with the rationale recorded next to it, and pinned by a test. (#4058, in flight, independently confirms the `local_sandbox` premise: `Path.resolve()` renders the native separator on Windows.)

`SEGMENT_BOUNDARY` / `PATH_TAIL` are deliberately **private**: `build_output_mask_pattern` is the only supported spelling of the rule, so a third site cannot import the pieces and hand-roll a variant that drifts — the exact shape this PR exists to remove.

## Scope

I enumerated the family by concept, not by file. Five sites build an escaped path base plus a boundary and tail:

| # | site | direction | this PR |
|---|---|---|---|
| 1 | `tools.py::_compiled_mask_patterns` | host→virtual (glob/grep/ls results) | migrated |
| 2 | `local_sandbox.py::_reverse_output_patterns` | host→virtual (bash output) | migrated |
| 3 | `local_sandbox.py::_command_pattern` | virtual→host (shell-oriented boundary class) | not migrated |
| 4 | `local_sandbox.py::_content_pattern` | virtual→host (text boundary class) | not migrated |
| 5 | `tools.py::replace_virtual_paths_in_command` | virtual→host, **no boundary at all** | separate PR |

**3 and 4** are the opposite direction with a different boundary class and a different tail (`/`-only, no `\`), they live in one file, and they have never drifted. Folding them into the same builder would turn its parameters into a "which rule do you want" switch — that is not removing duplication, it is putting two rules in one function.

**5 is a real bug of the same shape as #4035/#4053, mirrored** (virtual→host instead of host→virtual). Its pattern carries no segment boundary, so a prefix sibling of the virtual root is rewritten into a host path:

```
'cat /mnt/user-data/workspace/a.txt'    -> 'cat /host/t1/user-data/workspace/a.txt'    correct
'cat /mnt/user-data-backup/secret.txt'  -> 'cat /host/t1/user-data-backup/secret.txt'  sibling rewritten
'ls /mnt/user-data2'                    -> 'ls /host/t1/user-data2'                    sibling rewritten
```

That is a behavior change, and mixing it into a behavior-preserving extraction would muddy exactly the property this PR needs reviewed. It follows as its own PR.

## Surface area

- [x] **Sandbox** — `backend/packages/harness/deerflow/sandbox/path_patterns.py` (new), `sandbox/tools.py`, `sandbox/local/local_sandbox.py`

No documentation change needed: `backend/AGENTS.md` describes the virtual-path contract but makes no claim about how these matchers are constructed.

## Verification

This is a refactor, so the property to prove is that the delta set is **empty** — at both sites, for every base shape. I enumerated rather than argued.

**Differential check — a one-time run on the parent commit, not a committed test.** It compared the *real* pre-extraction inline expressions against the new helper: 13 bases (posix, Windows, UNC, spaces, regex metacharacters, non-ASCII, trailing slash, `/`) × 16 text shapes × both separator spellings = 416 probes, checking the compiled pattern source *and* every match span:

```
bases=13  probes=416  differences=0
asymmetry preserved: tools(agnostic) matches /-spelled = True (want True); local(plain) = False (want False)
```

This is what cleared the move, but it is not committable: once this lands there is no old inline expression left to diff against — only the frozen copies in `_legacy_tools_pattern` / `_legacy_local_pattern`. The committed guard is therefore the weaker snapshot in `test_sandbox_path_patterns.py`, and its red-ness rests on those literals, not on the length of `_BASES`. Sizing that list is a separate question, answered below.

**What `_BASES` actually guards** — mutating the helper one clause at a time (12 mutations across the boundary, the tail, and the escape/replace), the 8 committed bases catch all 12. The 7 they started as caught 11: the miss is a helper that *normalizes* its input by rstripping a trailing separator. Only two shapes catch that — a trailing-slash base and the Windows drive root — and `Path.resolve()` / `str(Path(...))` both strip trailing slashes, so neither call site can hand the helper `/host/skills/`. `C:\` survives `resolve()` with its separator intact, so it is the one base worth adding, and it is the only parametrize case that goes red under that mutation. Bases beyond these are shape documentation: both sides of the snapshot compute the same expression, so no base can separate them (5,000 fuzzed bases → zero byte-differences).

**Per-clause red check** — each clause of the new module broken in turn, one at a time:

| mutation | result |
|---|---|
| drop `SEGMENT_BOUNDARY` | 8 red |
| tail loses the `\` separator (posix-only) | 7 red |
| ignore `separator_agnostic` (always plain) | 2 red |
| always separator-agnostic | 2 red |
| re-inline a **byte-identical** copy at a call site | **12 green — not guarded** |
| re-inline a **divergent** copy at a call site (the #4035 shape) | 1 red |

Stating the fifth row plainly: the wiring tests do not catch a byte-identical re-inline, and cannot — an identical copy is not yet a defect. What they catch is the shape of the regression that actually happened: one copy edited, the other left behind. I also wrote and then **deleted** a test asserting the builder composes the exported constants — it stayed green under every mutation, so it guarded nothing; making the constants private addresses that structurally instead.

**End-to-end, nothing stubbed** — real `LocalSandbox`, real bash subprocess, real `mask_local_paths_in_output`:

```
$ echo <host>/skills                 -> /mnt/skills
$ echo <host>/skills/public/a.md     -> /mnt/skills/public/a.md
$ echo <host>/skills-extra/leak.md   -> <host>/skills-extra/leak.md     (sibling not masked — #4035's boundary)
$ echo 'paths: <host>/skills, <host>/skills/public'
                                     -> paths: /mnt/skills, /mnt/skills/public
```

The same script run against a clean `origin/main` worktree produces **byte-identical output** (`diff` reports nothing) — the direct evidence that behavior is unchanged, rather than "the tests are green".

## Validation

- `cd backend && make lint` — `ruff check`: all checks passed; `ruff format --check`: 832 files already formatted.
- `cd backend && make test` — **7130 passed, 26 skipped, 8 failed**. The same 8 fail on a clean `origin/main` checkout (pre-existing `web_fetch` cases — `test_browserless_client`, `test_crawl4ai_tools`, `test_fastcrw_tools`; no outbound DNS on this machine). Failure sets are byte-identical; this branch adds none. The 11 extra passes are the new tests.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Claude Code assisted with enumerating the pattern family across the sandbox layer, building the differential harness that proves the extraction is behavior-preserving, running the per-clause red check, and writing the tests. The contributor reviewed every line and takes responsibility for it.

### DIFF
--- a/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
+++ b/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
@@ -15,6 +15,7 @@ from typing import NamedTuple
 from deerflow.config.paths import VIRTUAL_PATH_PREFIX
 from deerflow.sandbox.env_policy import build_sandbox_env
 from deerflow.sandbox.local.list_dir import list_dir
+from deerflow.sandbox.path_patterns import build_output_mask_pattern
 from deerflow.sandbox.sandbox import Sandbox, _validate_extra_env
 from deerflow.sandbox.search import GrepMatch, find_glob_matches, find_grep_matches
 
@@ -234,9 +235,15 @@ class LocalSandbox(Sandbox):
         #
         # ``$`` is load-bearing: output ending exactly at a mount root would
         # otherwise fail the lookahead and be emitted as the raw host path.
-        boundary = r"(?=/|$|[^\w./-])"
-        tail = r"(?:[/\\][^\s\"';&|<>()]*)?"
-        return [re.compile(re.escape(self._resolved_local_paths[m]) + boundary + tail) for m in self._mappings_by_local_specificity]
+        #
+        # Both are owned by ``deerflow.sandbox.path_patterns`` and shared with
+        # ``sandbox.tools._compiled_mask_patterns``, the other site that rewrites host
+        # paths back to virtual ones — one copy of the rule, so the two cannot drift
+        # again (#4035 added the boundary here and missed that site; #4053 added it
+        # there). Bases stay separator-*sensitive* here: they come from
+        # ``Path.resolve()`` and already carry the platform's separator, so relaxing
+        # them would widen what this masks.
+        return [build_output_mask_pattern(self._resolved_local_paths[m]) for m in self._mappings_by_local_specificity]
 
     @cached_property
     def _resolved_local_paths(self) -> dict[PathMapping, str]:
--- a/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
+++ b/backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
@@ -221,28 +221,21 @@ class LocalSandbox(Sandbox):
     @cached_property
     def _reverse_output_patterns(self) -> list[re.Pattern[str]]:
         """Compiled matchers for local paths in command output (longest local path first)."""
-        # Same segment-boundary lookahead as the forward patterns above, so a mount
-        # root does not match inside a sibling that merely shares its prefix
-        # (``.../skills`` inside ``.../skills-extra``). Without it the regex yields
-        # the bare root, which then *equals* the mount root and so satisfies
-        # ``_reverse_resolve_path``'s own ``+ "/"`` guard — the sibling is rewritten
-        # to a container path that forward resolution refuses to map back.
-        #
-        # The boundary class mirrors ``_content_pattern``'s, not ``_command_pattern``'s:
-        # this runs over arbitrary command output, where a root can legitimately be
-        # followed by ``,`` ``:`` or ``\`` — all of which the shell-oriented class
-        # would reject. The trailing group keeps ``[/\\]`` so Windows paths still match.
-        #
-        # ``$`` is load-bearing: output ending exactly at a mount root would
-        # otherwise fail the lookahead and be emitted as the raw host path.
-        #
-        # Both are owned by ``deerflow.sandbox.path_patterns`` and shared with
+        # The rule — segment boundary plus path tail — is owned by
+        # ``deerflow.sandbox.path_patterns`` and shared with
         # ``sandbox.tools._compiled_mask_patterns``, the other site that rewrites host
-        # paths back to virtual ones — one copy of the rule, so the two cannot drift
-        # again (#4035 added the boundary here and missed that site; #4053 added it
-        # there). Bases stay separator-*sensitive* here: they come from
-        # ``Path.resolve()`` and already carry the platform's separator, so relaxing
-        # them would widen what this masks.
+        # paths back to virtual ones. Its rationale (why the boundary class is
+        # text-oriented rather than shell-oriented like ``_command_pattern``, why ``$``
+        # is load-bearing) lives with the owner rather than in a second copy here, which
+        # is what let the two drift before (#4035 added the boundary here and missed
+        # that site; #4053 added it there).
+        #
+        # What is specific to this site: without the boundary the regex yields the bare
+        # root, which then *equals* the mount root and so satisfies
+        # ``_reverse_resolve_path``'s own ``+ "/"`` guard — the sibling is rewritten to a
+        # container path that forward resolution refuses to map back. And bases stay
+        # separator-*sensitive*: they come from ``Path.resolve()`` and already carry the
+        # platform's separator, so relaxing them would widen what this masks.
         return [build_output_mask_pattern(self._resolved_local_paths[m]) for m in self._mappings_by_local_specificity]
 
     @cached_property
--- a/backend/tests/test_sandbox_path_patterns.py
+++ b/backend/tests/test_sandbox_path_patterns.py
@@ -1,0 +1,119 @@
+"""Tests for the shared host→virtual output-mask pattern (``sandbox/path_patterns.py``).
+
+The rule these pin is not "the regex is correct" — that is #4035/#4053 — but
+"there is exactly one copy of it, and extracting it did not change either call
+site's matching". The two sites differ on one axis only (separator handling),
+and that asymmetry is load-bearing: erasing it would widen ``LocalSandbox``'s
+masking or narrow ``sandbox.tools``'s.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pytest
+
+from deerflow.sandbox.local.local_sandbox import LocalSandbox, PathMapping
+from deerflow.sandbox.path_patterns import build_output_mask_pattern
+from deerflow.sandbox.tools import _compiled_mask_patterns
+
+
+def _legacy_tools_pattern(base: str) -> re.Pattern[str]:
+    """The expression ``_compiled_mask_patterns`` inlined before the extraction."""
+    escaped = re.escape(base).replace(r"\\", r"[/\\]")
+    return re.compile(escaped + r"(?=/|$|[^\w./-])" + r"(?:[/\\][^\s\"';&|<>()]*)?")
+
+
+def _legacy_local_pattern(base: str) -> re.Pattern[str]:
+    """The expression ``_reverse_output_patterns`` inlined before the extraction."""
+    return re.compile(re.escape(base) + r"(?=/|$|[^\w./-])" + r"(?:[/\\][^\s\"';&|<>()]*)?")
+
+
+_BASES = [
+    "/host/skills",
+    "/host/dir with spaces",
+    "/host/re+meta(chars)[x]",
+    "/host/dots.in.name",
+    "/Users/a/.deer-flow/users/u1/threads/t1/user-data",
+    "C:\\host\\skills",
+    "/host/技能",
+]
+
+
+@pytest.mark.parametrize("base", _BASES)
+def test_helper_reproduces_the_pre_extraction_expressions(base: str) -> None:
+    """Byte-identical to what each call site built inline, for both separator modes.
+
+    This is the anchor for the move itself: edit the helper in a way that changes
+    either site's regex and this goes red.
+    """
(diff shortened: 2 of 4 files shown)


AnnaSuSu:
Follow-up filed: the fifth site from the scope table — `replace_virtual_paths_in_command`, the one with no boundary at all — is now #4110, with the host read reproduced through real bash. It is independent of this PR (opposite direction, POSIX-only tail, so it is not a consumer of the helper here); either can land first.

WillemJiang (approved):

backend/tests/test_sandbox_path_patterns.py
```diff
@@ -0,0 +1,119 @@
...
+    escaped = re.escape(base).replace(r"\\", r"[/\\]")
+    return re.compile(escaped + r"(?=/|$|[^\w./-])" + r"(?:[/\\][^\s\"';&|<>()]*)?")
+
+
+def _legacy_local_pattern(base: str) -> re.Pattern[str]:
+    """The expression ``_reverse_output_patterns`` inlined before the extraction."""
+    return re.compile(re.escape(base) + r"(?=/|$|[^\w./-])" + r"(?:[/\\][^\s\"';&|<>()]*)?")
+
+
+_BASES = [
```
willem-bd, line 39: The PR body describes a 416-probe differential harness (13 bases × 16 shapes × 2 spellings), but `_BASES` here has 7 entries — so reading the test alone, it's not clear that the 416-probe matrix was ephemeral rather than committed.

The 7 bases are representative and the byte-identity assertion catches any helper edit that changes output (I re-ran with 12 bases, 0 diffs), so coverage is adequate. Either expand `_BASES` toward the described set, or add a one-line note in the PR body that the differential was a one-time check and this snapshot test is the committed guard.

backend/tests/test_sandbox_path_patterns.py
```diff
@@ -0,0 +1,119 @@
...
+from pathlib import Path
+
+import pytest
+
+from deerflow.sandbox.local.local_sandbox import LocalSandbox, PathMapping
+from deerflow.sandbox.path_patterns import build_output_mask_pattern
+from deerflow.sandbox.tools import _compiled_mask_patterns
+
+
+def _legacy_tools_pattern(base: str) -> re.Pattern[str]:
```
willem-bd, line 28: Maintenance shape worth flagging: `_legacy_tools_pattern` / `_legacy_local_pattern` re-implement the helper's logic, so `test_helper_reproduces_the_pre_extraction_expressions` compares the helper against a copy of itself — a coordinated edit to both stays green.

The two `..._route_through_the_helper` wiring tests mitigate this by anchoring independently to the real call sites, so it's acceptable as-is. Just noting these snapshots are now permanent fixtures to re-read (not just re-run) if the helper ever legitimately changes.

backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
```diff
@@ -234,9 +235,15 @@ def _reverse_output_patterns(self) -> list[re.Pattern[str]]:
...
-        return [re.compile(re.escape(self._resolved_local_paths[m]) + boundary + tail) for m in self._mappings_by_local_specificity]
+        #
+        # Both are owned by ``deerflow.sandbox.path_patterns`` and shared with
+        # ``sandbox.tools._compiled_mask_patterns``, the other site that rewrites host
+        # paths back to virtual ones — one copy of the rule, so the two cannot drift
+        # again (#4035 added the boundary here and missed that site; #4053 added it
+        # there). Bases stay separator-*sensitive* here: they come from
+        # ``Path.resolve()`` and already carry the platform's separator, so relaxing
+        # them would widen what this masks.
+        return [build_output_mask_pattern(self._resolved_local_paths[m]) for m in self._mappings_by_local_specificity]
```
willem-bd, line 239: Now that `path_patterns` owns the boundary rule, the comment above citing `_content_pattern` as the class reference ("The boundary class mirrors `_content_pattern`'s, not `_command_pattern`'s") reads as peer prior-art rather than the canonical source. The rationale is still accurate, so not worth churning — but if touched, a forward-pointer to `path_patterns` (the owner) would age better. Minor.

backend/tests/test_sandbox_path_patterns.py
AnnaSuSu: You are right that the body oversells this, and chasing the fix turned up something sharper than either option.

The differential compares the *real* pre-extraction expressions on the parent commit against the new helper. It is not committable: once this lands there is no old inline expression left to diff against — only the hand-copied literals in the test. So the body was wrong to present it as coverage that ships. Rewritten to say what it was: a one-time check on the parent, with the snapshot as the committed guard.

Expanding `_BASES` toward the described 13, though, runs straight into your other note. I mutated the helper one clause at a time — 12 mutations across the boundary, the tail, and the escape/replace — and asked which bases actually go red:

| candidate addition | ever the sole catcher of a mutation? |
|---|---|
| UNC `\\server\share\skills` | no |
| sibling `/host/skills-extra` | no |
| relative `rel/base` | no |
| posix root `/` | no |
| trailing slash `/host/skills/` | catches the one gap — but unreachable, see below |
| **drive root `C:\`** | **yes** |

The committed 7 catch 11 of the 12. The miss is a helper that *normalizes* its input by rstripping a trailing separator. Only two shapes catch it, and `Path.resolve()` / `str(Path(...))` both strip trailing slashes — so neither call site can ever hand the helper `/host/skills/`, and adding it would guard a case that cannot occur. `C:\` is different: a drive-root mount survives `resolve()` with its separator intact.

So I added exactly that one base. Reverting the helper to normalize its input turns **only** the `C:\` case red; the pre-existing seven stay green:

```
test_helper_reproduces_the_pre_extraction_expressions[C:\host\skills]  PASSED
test_helper_reproduces_the_pre_extraction_expressions[/host/技能]       PASSED
test_helper_reproduces_the_pre_extraction_expressions[C:\]             FAILED   <- the new one, and only it
1 failed, 11 passed
```

backend/tests/test_sandbox_path_patterns.py
AnnaSuSu: Agreed, and it is more fundamental than a caveat: the snapshot *has* to be a copy. "Did the extraction change behavior" can only be anchored against the pre-extraction expression, and after the merge that expression exists nowhere but in these frozen literals.

Worth stating plainly because it also settles the other note: a longer `_BASES` cannot fix the circularity. Both sides compute the same expression, so no base separates them — I fuzzed 5,000 random bases against the named ones and got zero byte-differences, exactly as construction predicts. The red-ness lives entirely in the frozen literals, not in the length of the list. (Sizing it is a different question, and does buy one thing — see the other thread.)

I have put that in the module docstring so the next reader does not have to re-derive it, and left the code alone. The independent anchors stay the ones you identified: the two `..._route_through_the_helper` wiring tests, plus `test_boundary_still_rejects_prefix_siblings_and_accepts_real_segments`, which asserts the rule itself rather than a copy of it.

backend/packages/harness/deerflow/sandbox/local/local_sandbox.py
AnnaSuSu: Taken, and pushed one step further. The comment did not just cite `_content_pattern` as the class reference — it re-stated the whole rationale (why the class is text-oriented rather than shell-oriented, why `$` is load-bearing), all of which now lives in `path_patterns`'s module docstring. That is the same duplication this PR exists to remove, one level up: a second copy of the *explanation* drifts exactly the way the second copy of the regex did.

So I dropped the duplicated half and pointed at the owner, keeping only what is genuinely site-specific — the `_reverse_resolve_path` `+ "/"` interaction, and why bases stay separator-sensitive here.

PR #4109, opened 2026-07-12
Fix: 8 LOW findings from 2026-07 bug bash (fd + rename findings dropped per review)

nankingjing:
## Summary

Bundles LOW-severity findings from the 2026-07-12 bug bash into a single PR.
Each fix is small, targeted, and ships with a unit test where straightforward.
Rebased onto latest `main` (69350787); the memory fixes were ported to their
post-#4122 locations under `agents/memory/backends/deermem/deermem/core/`.

### Defensive validation

- **`backend/packages/harness/deerflow/utils/network.py`** — `PortAllocator.allocate`
  now raises `ValueError` for non-int or out-of-range `start_port` (must be
  in `(0, 65536)`) and non-positive `max_range`.
- **`backend/packages/harness/deerflow/skills/storage/local_skill_storage.py`** —
  `ainstall_skill_from_archive` now fails fast with `FileNotFoundError` when
  the archive path is missing or not a regular file, before scanning
  (`is_file` routed through `asyncio.to_thread`).

### Resource lifecycle

- **`backend/packages/harness/deerflow/skills/installer.py`** —
  `_run_async_install` now reuses a module-level `ThreadPoolExecutor` singleton
  instead of constructing a fresh worker per install. Per review, the comment
  now documents that `max_workers=1` serializes concurrent installs (a real
  behavior change from the per-call executor) and that the pool is
  intentionally never shut down.
- **`backend/packages/harness/deerflow/tools/skill_manage_tool.py`** —
  Removed the redundant async `_to_thread` passthrough wrapper; call sites now
  use `asyncio.to_thread` directly.

### Concurrency

- **`backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py`** —
  `FileMemoryStorage.load` now performs stat + cache lookup + cold-path disk
  read inside a single critical section so concurrent first-time loads cannot
  race past the cache_lock and stomp on each other's cache writes. The comment
  documents the trade-off (cold loads serialize on `_cache_lock`; memory files
  are small) and the invariant that `_load_memory_from_file` must never
  re-acquire the non-reentrant lock.
- **`backend/app/scheduler/service.py`** — `ScheduledTaskService._run_loop`
  now catches known-transient errors (`OSError`, `TimeoutError`,
  `sqlalchemy.exc.OperationalError` — covering SQLite "database is locked")
  narrowly with the concrete exception type in the log line, and keeps a broad
  `except Exception` safety net so an unexpected error (including programming
  errors) is logged loudly as a bug signal but never kills the poller — nothing
  supervises/restarts `_run_loop`, so raising would silently stop all scheduled
  tasks until a process restart. (Per review: the original claim that
  programming errors would "surface" was wrong and has been removed; the
  observable change is better diagnostics, not new crash semantics.)

### Bug fixes

- **`backend/packages/harness/deerflow/community/url_safety.py`** —
  `resolve_host_addresses` now filters out non-INET (`AF_UNIX`, `AF_NETLINK`,
  …) `getaddrinfo` entries up front instead of letting `ip_address` raise and
  the bare `except` silently swallow them.
- **`backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py`** —
  `_normalize_memory_update_data` now `logger.warning`s a redacted summary
  (removal count, `newFacts` type, key count — no fact content, no PII) before
  rejecting a `factsToRemove`-with-malformed-`newFacts` update.

## Tests

- **`backend/tests/test_low_batch_2026_07.py`** — regression coverage
  for `PortAllocator` validation, `resolve_host_addresses` family filter
  (uses a literal non-INET int instead of `AF_APPLETALK`, which is not defined
  on Windows), `LocalSandbox` constructor compatibility (positional and the
  historical `id=` keyword, per review), `FileMemoryStorage.load` concurrency,
  and `ainstall_skill_from_archive` preflight. All 13 tests pass.
- Affected suites green: 324 passed, 2 skipped across the sandbox/memory/
  skill/scheduler test files adjacent to this PR (3 pre-existing Windows-only
  path-separator failures reproduce identically on pristine `main`).
- `ruff format --check` and `ruff check` clean across the diff.

## Dropped findings

- **`uploads/manager.py` fd handling** — dropped per review: the pre-existing
  `fd = -1` after `os.fdopen` + `finally` close already closed the descriptor
  exactly once on every path; the finding was invalid and the earlier rewrite
  introduced (then reverted) a double-close. No change to this file.
- **`local_sandbox.py` `id` → `sandbox_id` rename** — dropped per review (P2):
  renaming the constructor parameter breaks every external harness consumer
  that constructs `LocalSandbox(id="...")` with `TypeError`. The parameter
  keeps its original `id` name (shadowing the built-in inside the tiny
  constructor is harmless), the 27 test call-site updates are reverted (those
  four test files now match `main` byte-for-byte and are out of the diff), and
  `test_low_batch_2026_07.py` pins the public constructor shape (positional +
  `id=` keyword) as the requested regression.

### DIFF
--- a/backend/app/scheduler/service.py
+++ b/backend/app/scheduler/service.py
@@ -8,6 +8,7 @@ from datetime import UTC, datetime
 from typing import Any, Literal
 
 from fastapi import HTTPException
+from sqlalchemy import exc as sa_exc
 
 from deerflow.persistence.scheduled_task_runs import ActiveScheduledRunConflict
 from deerflow.runtime import ConflictError, RunRecord
@@ -389,10 +390,19 @@ class ScheduledTaskService:
         while not self._stop.is_set():
             try:
                 await self.run_once(now=datetime.now(UTC))
+            except (OSError, TimeoutError, sa_exc.OperationalError) as exc:
+                # Known-transient failure (e.g. SQLite "database is locked"
+                # surfaces as sqlalchemy.exc.OperationalError wrapped around
+                # sqlite3.OperationalError): log with the concrete type and
+                # retry on the next poll interval.
+                logger.exception("Scheduled task poll failed (%s); retrying next interval", type(exc).__name__)
             except Exception:
-                # A transient DB error (e.g. SQLite "database is locked") must
-                # not kill the poller task for the rest of the process life.
-                logger.exception("Scheduled task poll failed; retrying next interval")
+                # Anything else (including programming errors such as TypeError
+                # or AttributeError) is unexpected: it is logged loudly as a bug
+                # signal, but the poller deliberately keeps running — nothing
+                # supervises/restarts this task, so raising here would silently
+                # stop all scheduled tasks until a process restart.
+                logger.exception("Scheduled task poll failed with unexpected error; retrying next interval")
             try:
                 await asyncio.wait_for(
                     self._stop.wait(),
--- a/backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
+++ b/backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
@@ -425,7 +425,6 @@ class FileMemoryStorage(MemoryStorage):
         with self._cache_lock:
             return self._scope_locks.setdefault(key, threading.RLock())
 
-<<<<<<< HEAD
     def _get_memory_file_path(self, agent_name: str | None = None, *, user_id: str | None = None) -> Path:
         return memory_file_path(self._config, agent_name, user_id=user_id)
 
--- a/backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
+++ b/backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
@@ -425,6 +425,7 @@ class FileMemoryStorage(MemoryStorage):
         with self._cache_lock:
             return self._scope_locks.setdefault(key, threading.RLock())
 
+<<<<<<< HEAD
     def _get_memory_file_path(self, agent_name: str | None = None, *, user_id: str | None = None) -> Path:
         return memory_file_path(self._config, agent_name, user_id=user_id)
 
@@ -981,15 +982,25 @@ class FileMemoryStorage(MemoryStorage):
                 migration_notifications = self._run_read_migrations_locked(path, agent_name, user_id=user_id)
         for notification_agent, notifications in migration_notifications:
             self._dispatch_retrieval_notifications(notifications, user_id=user_id, agent_name=notification_agent)
-        signature = self._scope_signature(path, agent_name)
+        # Serialise the signature computation + cache lookup + (cold path)
+        # document read under a single critical section. Without this, two
+        # callers racing on a cold cache can both miss, both compute the same
+        # signature, and both schedule a disk read — the second write to
+        # ``_memory_cache`` then clobbers the first read, so the cached value
+        # can disagree with the file's actual contents (the file may have been
+        # touched between the two reads). Trade-off: the disk read + parse now
+        # happen while holding ``_cache_lock``, serialising concurrent cold
+        # loads; memory documents are small so this is acceptable.
+        # ``_scope_signature``/``_read_document`` must never re-acquire
+        # ``_cache_lock`` (it is a non-reentrant ``threading.Lock``).
         with self._cache_lock:
+            signature = self._scope_signature(path, agent_name)
             cached = self._memory_cache.get(key)
             if cached is not None and cached[1] == signature:
                 return copy.deepcopy(cached[0])
-        document = self._read_document(path, agent_name, user_id=user_id)
-        with self._cache_lock:
+            document = self._read_document(path, agent_name, user_id=user_id)
             self._memory_cache[key] = (copy.deepcopy(document), signature)
-        return copy.deepcopy(document)
+            return copy.deepcopy(document)
 
     def reload(self, agent_name: str | None = None, *, user_id: str | None = None) -> dict[str, Any]:
         path = self._get_memory_file_path(agent_name, user_id=user_id)
--- a/backend/packages/harness/deerflow/skills/installer.py
+++ b/backend/packages/harness/deerflow/skills/installer.py
@@ -339,6 +339,20 @@ async def _scan_skill_archive_contents_or_raise(skill_dir: Path, skill_name: str
     return static_findings
 
 
+# Singleton worker pool used to run a fresh asyncio loop inside an existing
+# loop context. Reused across calls so repeated installs do not pay the
+# ThreadPoolExecutor construction/teardown cost (one fresh worker thread per
+# install was the old behaviour).
+# Note: max_workers=1 deliberately serialises concurrent installs on this
+# path (the old per-call executor allowed them to overlap); installs are rare
+# and already mostly serialised by the skill-manage lock. The pool is never
+# shut down — the idle worker exits with the interpreter.
+_ASYNC_INSTALL_EXECUTOR = concurrent.futures.ThreadPoolExecutor(
+    max_workers=1,
+    thread_name_prefix="deerflow-skill-install",
+)
+
+
 def _run_async_install(coro):
     try:
         loop = asyncio.get_running_loop()
@@ -346,6 +360,5 @@ def _run_async_install(coro):
         loop = None
 
     if loop is not None and loop.is_running():
-        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
-            return executor.submit(asyncio.run, coro).result()
+        return _ASYNC_INSTALL_EXECUTOR.submit(asyncio.run, coro).result()
     return asyncio.run(coro)
--- a/backend/tests/test_low_batch_2026_07.py
+++ b/backend/tests/test_low_batch_2026_07.py
@@ -4,8 +4,8 @@ Covers the unit-testable fixes shipped in PR ``fix/low-batch-deerflow``:
 
 - ``PortAllocator.allocate`` argument validation.
 - ``resolve_host_addresses`` ignores non-INET address families.
-- ``LocalSandbox.__init__`` accepts ``sandbox_id`` (and the public
-  ``Sandbox.id`` property still exposes the configured identifier).
+- ``LocalSandbox.__init__`` keeps accepting the historical ``id=`` keyword
+  (the bug-bash rename was dropped in review as an API break).
 - ``FileMemoryStorage.load`` is safe under concurrent first-time loads.
 - ``LocalSkillStorage.ainstall_skill_from_archive`` raises
   ``FileNotFoundError`` for missing/non-file paths before scanning.
@@ -95,27 +95,27 @@ class TestResolveHostAddressesFamilyFilter:
 
 
 # ---------------------------------------------------------------------------
-# LocalSandbox sandbox_id parameter (sandbox/local/local_sandbox.py)
+# LocalSandbox constructor compatibility (sandbox/local/local_sandbox.py)
 # ---------------------------------------------------------------------------
 
 
-class TestLocalSandboxSandboxIdParameter:
-    """The renamed parameter should still surface through the public API."""
+class TestLocalSandboxConstructorCompatibility:
+    """``LocalSandbox`` must keep accepting the historical ``id=`` keyword.
 
-    def test_positional_sandbox_id(self):
+    The 2026-07 bug bash proposed renaming the ``id`` parameter to
+    ``sandbox_id`` to stop shadowing the built-in, but the rename was dropped
+    in review: external consumers of the published harness construct
+    ``LocalSandbox(id="...")``, so removing the keyword is an API break.
+    These tests pin the public constructor shape.
+    """
+
+    def test_positional_id(self):
         sb = LocalSandbox("local:user:thread:42")
         assert sb.id == "local:user:thread:42"
 
-    def test_keyword_sandbox_id(self):
-        sb = LocalSandbox(sandbox_id="local:user:thread:42")
+    def test_keyword_id(self):
+        sb = LocalSandbox(id="local:user:thread:42")
         assert sb.id == "local:user:thread:42"
-
-    def test_id_does_not_shadow_builtin(self):
-        # If the parameter were still named ``id`` the implicit ``id(self)``
-        # call below would resolve to the parameter string and fail with a
-        # TypeError. With the renamed parameter the builtin works again.
-        sb = LocalSandbox(sandbox_id="local:user:thread:42")
-        builtin_id = id(sb)  # noqa: F841 — intentionally exercising the builtin
 
 
 # ---------------------------------------------------------------------------
--- a/backend/tests/test_low_batch_2026_07.py
+++ b/backend/tests/test_low_batch_2026_07.py
@@ -1,0 +1,202 @@
+"""Regression tests for the LOW-severity bug bash batch (2026-07-12).
+
+Covers the unit-testable fixes shipped in PR ``fix/low-batch-deerflow``:
+
+- ``PortAllocator.allocate`` argument validation.
+- ``resolve_host_addresses`` ignores non-INET address families.
+- ``LocalSandbox.__init__`` accepts ``sandbox_id`` (and the public
+  ``Sandbox.id`` property still exposes the configured identifier).
+- ``FileMemoryStorage.load`` is safe under concurrent first-time loads.
+- ``LocalSkillStorage.ainstall_skill_from_archive`` raises
+  ``FileNotFoundError`` for missing/non-file paths before scanning.
+"""
+
+from __future__ import annotations
+
+import socket
+import threading
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+from deerflow.agents.memory.backends.deermem.deermem.config import DeerMemConfig
+from deerflow.agents.memory.backends.deermem.deermem.core.storage import FileMemoryStorage
+from deerflow.community.url_safety import resolve_host_addresses
+from deerflow.sandbox.local.local_sandbox import LocalSandbox
+from deerflow.skills.storage.local_skill_storage import LocalSkillStorage
+from deerflow.utils.network import PortAllocator
+
+# ---------------------------------------------------------------------------
+# PortAllocator validation (utils/network.py)
+# ---------------------------------------------------------------------------
+
+
+class TestPortAllocatorValidation:
+    """``allocate`` must reject obviously invalid arguments up front."""
+
+    def test_rejects_zero_start_port(self):
+        with pytest.raises(ValueError, match="start_port"):
+            PortAllocator().allocate(start_port=0)
+
+    def test_rejects_negative_start_port(self):
+        with pytest.raises(ValueError, match="start_port"):
+            PortAllocator().allocate(start_port=-1)
+
+    def test_rejects_out_of_range_start_port(self):
+        with pytest.raises(ValueError, match="start_port"):
+            PortAllocator().allocate(start_port=65536)
+
+    def test_rejects_zero_max_range(self):
+        with pytest.raises(ValueError, match="max_range"):
+            PortAllocator().allocate(start_port=8080, max_range=0)
+
+    def test_rejects_negative_max_range(self):
+        with pytest.raises(ValueError, match="max_range"):
+            PortAllocator().allocate(start_port=8080, max_range=-5)
+
+    def test_rejects_non_int_start_port(self):
+        with pytest.raises(ValueError, match="start_port"):
+            PortAllocator().allocate(start_port="8080")  # type: ignore[arg-type]
+
+    def test_allocation_still_succeeds_for_valid_args(self):
+        # Use a high start port unlikely to collide with anything in CI.
+        allocator = PortAllocator()
+        port = allocator.allocate(start_port=50000, max_range=10)
+        assert 50000 <= port < 50010
+        allocator.release(port)
+
+
+# ---------------------------------------------------------------------------
+# url_safety.resolve_host_addresses (community/url_safety.py)
+# ---------------------------------------------------------------------------
+
+
+class TestResolveHostAddressesFamilyFilter:
+    """Only AF_INET / AF_INET6 entries should reach ``ip_address``."""
+
+    def test_skips_non_inet_address_families(self):
+        # Inject a fake getaddrinfo result that mixes INET with a non-INET
+        # family. Without the family filter the non-INET entry would raise
+        # ``ValueError`` from ``ip_address`` and the bare ``except`` would
+        # silently drop it; with the filter the entry is rejected upfront.
+        # Use a literal int (999) instead of ``socket.AF_APPLETALK`` which is
+        # not defined on Windows.  Any non-INET/non-INET6 address family value
+        # works — the value only needs to be a valid integer that isn't INET or
+        # INET6 for the family filter to reject it.
+        fake_infos = [
+            (socket.AF_INET, socket.SOCK_STREAM, 0, "", ("10.0.0.1", 0)),
+            (999, socket.SOCK_STREAM, 0, "", (b"\x00\x00\x00\x00",)),
+            (socket.AF_INET6, socket.SOCK_STREAM, 0, "", ("::1", 0, 0, 0)),
+        ]
+        with patch("deerflow.community.url_safety.socket.getaddrinfo", return_value=fake_infos):
+            addrs = resolve_host_addresses("example.com")
+        assert [str(a) for a in addrs] == ["10.0.0.1", "::1"]
+
+
+# ---------------------------------------------------------------------------
+# LocalSandbox sandbox_id parameter (sandbox/local/local_sandbox.py)
+# ---------------------------------------------------------------------------
+
+
+class TestLocalSandboxSandboxIdParameter:
+    """The renamed parameter should still surface through the public API."""
+
+    def test_positional_sandbox_id(self):
+        sb = LocalSandbox("local:user:thread:42")
+        assert sb.id == "local:user:thread:42"
+
+    def test_keyword_sandbox_id(self):
+        sb = LocalSandbox(sandbox_id="local:user:thread:42")
+        assert sb.id == "local:user:thread:42"
+
+    def test_id_does_not_shadow_builtin(self):
+        # If the parameter were still named ``id`` the implicit ``id(self)``
+        # call below would resolve to the parameter string and fail with a
+        # TypeError. With the renamed parameter the builtin works again.
+        sb = LocalSandbox(sandbox_id="local:user:thread:42")
+        builtin_id = id(sb)  # noqa: F841 — intentionally exercising the builtin
+
+
+# ---------------------------------------------------------------------------
+# FileMemoryStorage.load race (agents/memory/backends/deermem/deermem/core/storage.py)
+# ---------------------------------------------------------------------------
+
+
+class TestFileMemoryStorageLoadConcurrency:
+    """Concurrent cold-cache loads must not race past the cache_lock."""
+
+    def test_concurrent_cold_loads_share_one_disk_read(self, tmp_path: Path):
+        # Force a known path that lives under tmp_path so we do not depend on
+        # the global config resolution.
+        target = tmp_path / "mem.json"
+
+        storage = FileMemoryStorage(DeerMemConfig(storage_path=str(tmp_path)))
+
+        read_count = 0
+        read_lock = threading.Lock()
+
+        real_reader = storage._read_document
+
+        def counting_reader(path, agent_name, *, user_id=None):
+            nonlocal read_count
+            with read_lock:
+                read_count += 1
+            return real_reader(path, agent_name, user_id=user_id)
+
+        # Patch the cache key + file path lookup to drive every call at the
+        # same (user_id, agent_name) tuple and the same file, and pin the
+        # scope signature so every thread sees the same cache key state.
+        with (
+            patch.object(storage, "_get_memory_file_path", return_value=target),
+            patch.object(storage, "_cache_key", return_value=("u", "a")),
+            patch.object(storage, "_legacy_agent_memory_path", return_value=None),
+            patch.object(storage, "_global_json_needs_migration", return_value=False),
+            patch.object(storage, "_scope_signature", return_value=("sig",)),
+            patch.object(storage, "_read_document", side_effect=counting_reader),
+        ):
+            barrier = threading.Barrier(8)
+
+            def worker():
+                barrier.wait()
+                data = storage.load("a", user_id="u")
+                assert data["version"] == "1.0"
+
+            threads = [threading.Thread(target=worker) for _ in range(8)]
+            for t in threads:
+                t.start()
+            for t in threads:
+                t.join()
+
+        # All 8 threads raced past the barrier; the single critical section
+        # should have produced exactly one disk read. The original code's
+        # signature-outside-lock + read-outside-lock shape would race and
+        # could hit the disk multiple times.
+        assert read_count == 1, f"expected one disk read, got {read_count}"
+
+
+# ---------------------------------------------------------------------------
+# LocalSkillStorage.ainstall_skill_from_archive (skills/storage/local_skill_storage.py)
+# ---------------------------------------------------------------------------
+
+
+class TestAinstallSkillFromArchivePreflight:
+    """Missing or non-file archive paths must fail fast with ``FileNotFoundError``."""
+
+    def test_missing_path_raises_file_not_found(self, tmp_path: Path):
+        storage = LocalSkillStorage(host_path=str(tmp_path))
+        missing = tmp_path / "does-not-exist.zip"
+        with pytest.raises(FileNotFoundError, match="Skill archive not found"):
+            # ``ainstall_skill_from_archive`` is async; drive it via asyncio.run.
+            import asyncio
+
+            asyncio.run(storage.ainstall_skill_from_archive(missing))
+
+    def test_directory_path_raises_file_not_found(self, tmp_path: Path):
+        storage = LocalSkillStorage(host_path=str(tmp_path))
+        # ``is_file()`` returns False for directories, so this is the other
+        # branch of the preflight check.
+        with pytest.raises(FileNotFoundError, match="Skill archive not found"):
+            import asyncio
+
+            asyncio.run(storage.ainstall_skill_from_archive(tmp_path))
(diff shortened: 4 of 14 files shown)


backend/packages/harness/deerflow/uploads/manager.py
```diff
@@ -214,7 +214,17 @@ def open_upload_file_no_symlink(base_dir: Path, filename: str) -> tuple[Path, ob
...
+            # Transfer ownership of ``fd`` to the file object only once we know
+            # ``fdopen`` succeeded — setting ``fd = -1`` only after ``fdopen``
+            # returns closes the window where ``fdopen`` raised (e.g. a Python
+            # level allocation failure) and the original ``fd`` would have been
+            # closed in the ``finally`` block. The earlier pattern cleared
+            # ``fd`` *before* ``fdopen``, which orphaned the fd on any failure.
+            try:
+                fh = os.fdopen(fd, "wb")
+            except BaseException:
+                os.close(fd)
```
willem-bd: This `except` closes `fd` but never clears it, so the outer `finally` closes the same fd a second time.

```python
try:
    fh = os.fdopen(fd, "wb")
except BaseException:
    os.close(fd)
    raise        # fd still holds the original integer here
fd = -1          # only reached on success
finally:
    if fd >= 0:
        os.close(fd)   # runs again on the failure path -> double close
```

On the fdopen-failure path: the `except` closes `fd`, then `raise` runs the `finally`, which sees `fd >= 0` (still the old value) and closes it **again**. I confirmed empirically - the caller receives `OSError: [Errno 9] Bad file descriptor` with the real failure (e.g. a `RuntimeError` from `fdopen`) lost to `__context__`. Under concurrency the second `os.close()` could close an unrelated fd whose number the OS reused between the two closes - the exact hazard this change was meant to fix.

The comment says "the earlier pattern cleared `fd` *before* `fdopen`, which orphaned the fd on any failure," but the actual original set `fd = -1` *after* `fdopen` and correctly closed once via the `finally`. The original was fine; this change is the regression.

**Fix:** clear `fd` in the except branch before re-raising:

```python
except BaseException:
    os.close(fd)
    fd = -1
    raise
```

backend/app/scheduler/service.py
```diff
@@ -333,10 +333,13 @@ async def _run_loop(self) -> None:
         while not self._stop.is_set():
             try:
                 await self.run_once(now=datetime.now(UTC))
-            except Exception:
+            except (OSError, TimeoutError) as exc:
```
willem-bd: This narrowed catch no longer covers the SQLite "database is locked" case the comment right below it still cites.

The repos use SQLAlchemy `AsyncSession` (`claim_due_tasks` does `session.execute`/`commit`), and "database is locked" surfaces as `sqlalchemy.exc.OperationalError` (wrapping `sqlite3.OperationalError`). Neither is an `OSError` subclass - verified MROs:

- `sqlite3.OperationalError` -> `DatabaseError` -> `Error` -> `Exception`
- `sqlalchemy.exc.OperationalError` -> `…` -> `SQLAlchemyError` -> `Exception`

`count_active_runs()` and `claim_due_tasks()` are called directly in `run_once` with no surrounding try/except, so a >5s lock contention (the `busy_timeout` is 5s per `database_config.py`) now propagates past `except (OSError, TimeoutError)` and **kills the poller task permanently** - exactly what the old `except Exception` was there to survive. The goal of surfacing `TypeError`/`AttributeError` is good, but this drops the documented DB resilience.

**Fix:** include the DB error, e.g. `except (OSError, TimeoutError, sqlalchemy.exc.OperationalError)`, or keep `Exception` and explicitly re-raise programming errors (`TypeError`, `AttributeError`, `KeyError`, …).

backend/tests/test_skill_request_scoped_secrets.py
```diff
@@ -35,37 +35,37 @@ def test_injected_env_visible_to_command(self):
...
+        sandbox = LocalSandbox(sandbox_id="local")
         sandbox.execute_command("true", env={"DEERFLOW_EPHEMERAL": "leaky"})
         out = sandbox.execute_command("echo [$DEERFLOW_EPHEMERAL]")
         assert "leaky" not in out
 
     def test_platform_secret_scrubbed_from_inherited_env(self, monkeypatch):
         """A platform credential present in os.environ must NOT reach the sandbox
         subprocess (the baseline-env leak surface). Without this, scoped injection
-        is security theatre — a skill script could simply read $OPENAI_API_KEY."""
+        is security theatre 鈥?a skill script could simply read $OPENAI_API_KEY."""
```
willem-bd: Heads-up: this `鈥?` is mojibake, not an em-dash - and it appears 27 times across this file (20) and `test_local_sandbox_path_regex_cache.py` (7).

I verified the actual bytes at the PR head: `e9 88 a5 3f` = U+9275 (`鈥`) + `?`, the classic UTF-8-em-dash-decoded-as-GBK corruption. The original lines used clean ASCII hyphens (` - `), so these docstring edits are unrelated to the stated `id`->`sandbox_id` rename (the PR description claims only call-site renames in tests) and look like an editor "smart dashes" + encoding misconfiguration. `test_sandbox_search_tools.py` is clean, so only two of the three files are affected.

It won't break execution (it's in docstrings), but it corrupts existing documentation and adds noise to the diff. Recommend reverting all non-rename docstring edits in these two files.

willem-bd (commented):
Reviewed the bundle against the PR head (2b89e8f). The ~7 defensive fixes are solid and well-tested (175 pass, ruff clean), but two of the changes are themselves regressions and there is widespread mojibake in two test files - details inline. Requesting changes on those three.

nankingjing:
Fixed the fd double-close in uploads/manager.py (set fd=-1 after manual close in except block). Also caught `OperationalError` in the scheduler poller loop — SQLite "database is locked" was killing the poller silently. Cleaned up a mojibake comment while I was there.

nankingjing (commented):
Merge conflicts resolved against latest main (3e7baba3). Additional fixes applied:

1. **Mojibake cleanup**: All remaining 鈫? (arrow mojibake) occurrences in test_skill_request_scoped_secrets.py have been replaced with proper → characters, complementing the 鈥? →  — fixes from the original PR.

2. **Test call site updates**: All LocalSandbox(id=...) calls in test_sandbox_search_tools.py and test_local_sandbox_path_regex_cache.py updated to sandbox_id= to match the renamed parameter. The rebase auto-merge had reverted some of these.

3. **Regression fixes confirmed intact**: The fd double-close fix (manager.py:227 now clears fd after close) and scheduler catch fix (service.py:338 now catches (OSError, TimeoutError, OperationalError) including SQLite locked) are both preserved.

4. **Tests**: 171 pass, 4 fail with pre-existing Windows platform issues (path separators/backslash in glob, file permissions). No new failures introduced.

Ready for re-review.

willem-bd (commented):
Reviewed the 10 LOW-finding batch. Most fixes are clean — the `id`→`sandbox_id` rename has no remaining keyword call sites, and the `resolve_host_addresses` family filter / `PortAllocator` validation / archive preflight / `_to_thread` removal are solid with tests.

Two things I'd ask to address before merge:
- **Test-file corruption**: the three touched test files carry mojibake (`鈥?` / `鈫?`) and spurious whitespace changes unrelated to the rename — please revert non-`sandbox_id` hunks (see inline).
- **`uploads/manager.py` fd rewrite is a no-op**: the pre-existing `finally` already closed the fd on `fdopen` failure; the new `try/except` is equivalent and the stated justification doesn't match the code.

Plus nits/risks inline: scheduler except narrowing, lock-held-across-IO, singleton executor, PII in warning log, `AF_APPLETALK` portability.

backend/tests/test_sandbox_search_tools.py
```diff
@@ -212,7 +212,7 @@ def test_aio_sandbox_glob_include_dirs_filters_nested_ignored(monkeypatch) -> No
                 files=[
                     SimpleNamespace(name="src", path="/mnt/workspace/src"),
                     SimpleNamespace(name="node_modules", path="/mnt/workspace/node_modules"),
-                    # child of node_modules — should be filtered via should_ignore_path
+                    # child of node_modules 鈥?should be filtered via should_ignore_path
```
willem-bd: Mojibake: this line, plus `# ls_tool 鈥?path masking` (L401) and `contextvar unset 鈫?"default"` (L476), show a clean `-` / `->` replaced by `鈥?` / `鈫?` — the classic GBK-locale-editor-misreading-UTF-8 signature. The PR also mangles comment whitespace across `test_local_sandbox_path_regex_cache.py` and `test_skill_request_scoped_secrets.py` (` - ` → `  - `, `-> ` → `->`, plus re-wrapped docstring lines).

None of this relates to the `id`→`sandbox_id` rename and it isn't `ruff format` output. Please revert every non-`sandbox_id` hunk in these three test files so the diff stays focused — it also pollutes `git blame`.

backend/packages/harness/deerflow/uploads/manager.py
```diff
@@ -214,7 +214,18 @@ def open_upload_file_no_symlink(base_dir: Path, filename: str) -> tuple[Path, ob
...
-            fh = os.fdopen(fd, "wb")
+            # Transfer ownership of ``fd`` to the file object only once we know
+            # ``fdopen`` succeeded — setting ``fd = -1`` only after ``fdopen``
+            # returns closes the window where ``fdopen`` raised (e.g. a Python
+            # level allocation failure) and the original ``fd`` would have been
+            # closed in the ``finally`` block. The earlier pattern cleared
+            # ``fd`` *before* ``fdopen``, which orphaned the fd on any failure.
+            try:
+                fh = os.fdopen(fd, "wb")
+            except BaseException:
```
willem-bd: I don't think this fixes an actual leak. The pre-existing code cleared `fd = -1` *after* `os.fdopen` (the unchanged `fd = -1` just below at L229), with `finally: if fd >= 0: os.close(fd)`. On any `fdopen` failure `fd` is still `>= 0`, so the `finally` already closed it — no orphan. The new `except BaseException: os.close(fd); fd = -1; raise` is behaviourally identical (fd closed exactly once).

The PR description's claim that "the earlier pattern cleared fd *before* fdopen" doesn't match the diffed code. Consider dropping this hunk or correcting the comment so we don't record a phantom fix.

backend/app/scheduler/service.py
```diff
@@ -333,10 +334,14 @@ async def _run_loop(self) -> None:
         while not self._stop.is_set():
             try:
                 await self.run_once(now=datetime.now(UTC))
-            except Exception:
-                # A transient DB error (e.g. SQLite "database is locked") must
-                # not kill the poller task for the rest of the process life.
-                logger.exception("Scheduled task poll failed; retrying next interval")
+            except (OSError, TimeoutError, sa_exc.OperationalError) as exc:
```
willem-bd, line 393: Good intent (surface programming errors), but the failure mode is severe: any exception from `run_once` outside `(OSError, TimeoutError, OperationalError)` — e.g. an unexpected `sqlalchemy.exc.DBAPIError` subclass, an `IntegrityError` from schema drift, or a `KeyError`/`TypeError` on a malformed task row — now propagates out of the `while` and kills the poller task, so scheduled tasks silently stop running until process restart. The old broad catch was sloppy but safe.

Would you keep a narrow set plus a logged `except Exception` that continues, or confirm whatever supervises `_run_loop` restarts it on death?

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
```diff
@@ -125,22 +125,25 @@ def load(self, agent_name: str | None = None, *, user_id: str | None = None) ->
...
+                current_mtime = None
+
             cached = self._memory_cache.get(cache_key)
             if cached is not None and cached[1] == current_mtime:
                 return cached[0]
 
-        memory_data = self._load_memory_from_file(agent_name, user_id=user_id)
-
-        with self._cache_lock:
+            memory_data = self._load_memory_from_file(agent_name, user_id=user_id)
```
willem-bd: Race fix looks correct, and `_load_memory_from_file` doesn't re-enter `_cache_lock`, so no deadlock. One tradeoff: the single global `_cache_lock` is now held across `exists()` + `open()` + `json.load()`, so every cold read serialises all `load()` calls across all users/keys.

Fine for small local files; could stall under contention if memory files grow or sit on a network mount. A per-key lock would remove the global stall — acceptable as a LOW fix either way.

backend/packages/harness/deerflow/skills/installer.py
```diff
@@ -309,13 +309,22 @@ async def _scan_skill_archive_contents_or_raise(skill_dir: Path, skill_name: str
     return static_findings
 
 
+# Singleton worker pool used to run a fresh asyncio loop inside an existing
+# loop context. Reused across calls so repeated installs do not pay the
+# ThreadPoolExecutor construction/teardown cost (one fresh worker thread per
+# install was the old behaviour).
+_ASYNC_INSTALL_EXECUTOR = concurrent.futures.ThreadPoolExecutor(
```
willem-bd, line 350: Reasonable reuse. Two caveats:

1. A single `max_workers=1` executor now serialises all `_run_async_install` calls globally — the old per-call executor allowed parallel installs. They're usually already serialised by the skill-manage lock, but `_run_async_install` is also reached via `skill_storage.py` outside that lock.
2. It's never `shutdown()` (harmless — idle worker exits at interpreter shutdown), just noting it's a module-global thread.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -392,6 +392,17 @@ def _normalize_memory_update_data(update_data: dict[str, Any]) -> dict[str, Any]
                 dropped_new_fact = True
 
     if normalized_facts_to_remove and dropped_new_fact:
+        # The LLM asked to remove facts and produced malformed newFacts. Applying
+        # only the removals would let the model quietly drop knowledge it never
+        # intended to replace, so the whole update is rejected. Surface the
+        # rejected payload in the log so operators can correlate the malformed
+        # response with the upstream LLM call without losing the original intent.
+        logger.warning(
+            "Rejecting memory update with %d removal(s) but malformed newFacts (%s); update_data=%s",
```
willem-bd: This logs `json.dumps(update_data)[:512]` at WARNING, which can carry user memory facts (potential PII). The data was already surfaced via the `JSONDecodeError.doc` raised just below, so this additionally writes it to structured logs.

Truncation helps and the path is rare (malformed LLM output), but worth confirming your log sinks are OK with memory content.

backend/tests/test_low_batch_2026_07.py
```diff
@@ -0,0 +1,199 @@
...
+        # Inject a fake getaddrinfo result that mixes INET with a non-INET
+        # family. Without the family filter the non-INET entry would raise
+        # ``ValueError`` from ``ip_address`` and the bare ``except`` would
+        # silently drop it; with the filter the entry is rejected upfront.
+        # Use ``AF_APPLETALK`` since it is platform-portable (Windows does not
+        # expose ``AF_UNIX``); the value only needs to be a real address
+        # family constant that is not INET or INET6.
+        fake_infos = [
+            (socket.AF_INET, socket.SOCK_STREAM, 0, "", ("10.0.0.1", 0)),
+            (socket.AF_APPLETALK, socket.SOCK_STREAM, 0, "", (b"\x00\x00\x00\x00",)),
```
willem-bd: Minor: the comment at L82 claims `AF_APPLETALK` is "platform-portable (Windows does not expose AF_UNIX)", but `AF_APPLETALK` also isn't defined on Windows (Linux/macOS only) — this would `AttributeError` there. The filter only checks `not in (AF_INET, AF_INET6)`, so a literal int like `999` is fully portable and still exercises the branch.

backend/app/scheduler/service.py
```diff
@@ -333,10 +334,18 @@ async def _run_loop(self) -> None:
...
+                # AttributeError, etc.) still surface instead of being silently
+                # looped forever.
+                logger.exception("Scheduled task poll failed (%s); retrying next interval", type(exc).__name__)
             except Exception:
-                # A transient DB error (e.g. SQLite "database is locked") must
-                # not kill the poller task for the rest of the process life.
-                logger.exception("Scheduled task poll failed; retrying next interval")
+                # Safety net: any unexpected exception type that slips through
+                # the narrow catch above must still not kill the poller forever.
+                logger.exception("Scheduled task poll failed with unexpected error; retrying next interval")
```
willem-bd, line 405: **Scheduler: the stated goal isn't achieved.** The PR description says this narrows the broad `except Exception` so genuine programming errors (`TypeError`, `AttributeError`, …) "surface instead of being silently looped forever."

But the broad `except Exception:` (L345) is kept as a safety net *after* the new narrow catch. A `TypeError` isn't in `(OSError, TimeoutError, sa_exc.OperationalError)`, so it falls through to the broad except, gets logged, and the loop continues — exactly the behavior the finding claims to fix. The only observable change is a more specific log line for known-transient errors; genuine bugs still don't surface.

Either drop the broad `except Exception` (if you really want bugs to surface / kill the poller), or keep it and correct the PR description. Side note: `sa_exc.OperationalError` also covers *persistent* conditions (missing table, disk error), not just transient "database is locked" — treating all of them as retryable is pre-existing, but the narrow catch makes that intent look broader than it is.

backend/tests/test_sandbox_search_tools.py
willem-bd: **Encoding corruption introduced here.** `main` has a Unicode em-dash (`# child of node_modules - should be filtered`); this PR replaced it with mojibake `鈥?`. Confirmed at the byte level: `main` is `E2 80 94` (em-dash), the PR head is the GBK-misdecoded sequence. Same pattern at L401 (`# ls_tool 鈥?path masking`) and L798 (`contextvar unset 鈫?"default"`, arrow `->` → `鈫?`).

This is genuine file corruption, and it's unrelated to the `id=`→`sandbox_id=` rename. The same set of files also carries a lot of unrelated whitespace churn (` - ` → `  - `, `-> ` → `->`) and two docstring line-joins in `test_local_sandbox_path_regex_cache.py`.

ruff won't catch any of this (it doesn't reflow comment interiors or detect mojibake). Suggest reverting every non-`sandbox_id` change in these three test files and re-applying only the parameter rename.

backend/packages/harness/deerflow/skills/installer.py
willem-bd, line 350: Two things worth a note on the singleton:

- `max_workers=1` **serializes** concurrent skill installs. Previously each install spun up its own 1-worker executor, so installs could overlap; now they queue on this one thread. Fine if installs are rare, but it's a real behavior change — worth a one-line comment saying so.
- The pool is never shut down (no `atexit`). On a long-lived process that's negligible; just be aware pending `asyncio.run` work can be lost on abrupt exit.

`.result()` still blocks the calling thread exactly as before, so no new event-loop-blocking risk was introduced.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
willem-bd: Locking stat + cache lookup + `_load_memory_from_file` under one critical section is the right fix for the cold-cache race (and the 8-thread test correctly asserts exactly one disk read). One trade-off and one thing to confirm:

- **Trade-off:** the disk read + JSON parse now holds `self._cache_lock`, serializing concurrent cold loads. Acceptable since memory files are small, but worth being deliberate about.
- **Confirm:** `_load_memory_from_file` must not re-acquire `self._cache_lock` — it's a `threading.Lock` (non-reentrant, L71), so a nested acquire would deadlock. The concurrency test patches this method out, so it wouldn't catch a production deadlock here.

willem-bd (commented):
Reviewing the 10 LOW-finding bundle. Most fixes are sound and well-scoped. Two issues to address before merge: (1) the scheduler change doesn't actually achieve its stated goal of letting programming errors surface, and (2) the test files picked up mojibake + unrelated whitespace churn. Details in the inline comments; full write-up below.

**Looks good / no action:** `local_sandbox.py` rename (verified the two production callers in `local_sandbox_provider.py` pass the id positionally, so they're unaffected), `skill_manage_tool.py` `_to_thread` removal (verified it's only referenced in that file), `url_safety.py` family filter, `local_skill_storage.py` archive preflight, `updater.py` warning (verified `logger` is defined at L34), and `network.py` validation (minor: `bool` slips through `isinstance(.., int)` and the end of the range isn't bounded — both cosmetic / out of scope).

**Recommendation:** Request changes — resolve the scheduler description/behavior mismatch and revert the test-file noise; the rest is mergeable.

WillemJiang:
@nankingjing, please resolve the conflict with the main branch.

backend/packages/harness/deerflow/uploads/manager.py
nankingjing: Fixed by dropping the hunk entirely. You're right on both counts: the except-branch rewrite left `fd` holding the closed descriptor so the `finally` double-closed it, and (per your later comment) the original code was already correct — the `fd = -1` after `os.fdopen` plus the `finally` guard closes exactly once on every path. Rather than patching the patch, `uploads/manager.py` is no longer touched by this PR at all, and the description now records the finding as invalid.

backend/app/scheduler/service.py
nankingjing: Fixed — the narrow catch is now `except (OSError, TimeoutError, sa_exc.OperationalError)`, so SQLAlchemy-wrapped SQLite "database is locked" is retried on the next poll interval instead of escaping. A broad `except Exception` safety net is retained after it so nothing can kill the poller (see the discussion on the later comment about the stated goal — the comments/description no longer claim programming errors surface).

backend/tests/test_skill_request_scoped_secrets.py
nankingjing: Fixed. The file was rebuilt from main's pristine copy with only the `LocalSandbox(id=` → `sandbox_id=` rename applied (7 call sites in this file), so every mojibake byte sequence and whitespace/docstring edit is gone — the diff on this file is now exactly the rename lines. Verified at the byte level that no `e9 88 a5`/`e9 88 ab` (GBK-misdecoded em-dash/arrow) sequences remain anywhere in the PR.

backend/tests/test_sandbox_search_tools.py
nankingjing: Fixed the same way as the secrets test file: rebuilt from main's pristine copy and re-applied only the `LocalSandbox(id=` → `sandbox_id=` rename (16 call sites here). Main's em-dash at L215, the `ls_tool — path masking` header, and the `contextvar unset → "default"` arrow are byte-identical to main now; the whitespace churn and docstring line-joins in `test_local_sandbox_path_regex_cache.py` are gone as well.

backend/packages/harness/deerflow/uploads/manager.py
nankingjing: Agreed — it was a no-op (and my first attempt at "fixing" it introduced the double-close you flagged). The hunk is dropped entirely: `uploads/manager.py` is untouched by this PR, and the PR description now lists the finding under "Dropped findings" as invalid instead of recording a phantom fix.

backend/app/scheduler/service.py
nankingjing: Kept the broad `except Exception` as a safety net after the narrow catch — nothing supervises or restarts `_run_loop`, so an escaping `DBAPIError` subclass / `IntegrityError` / `KeyError` would silently stop all scheduled tasks until a process restart, which is worse than looping. The inline comments now say this explicitly: unexpected errors are logged loudly as a bug signal, but the poller deliberately keeps running.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
nankingjing: Kept the single critical section and documented the trade-off in the comment above it: cold loads now serialize on `_cache_lock`, which is acceptable while memory files are small/local; a per-key lock is the natural next step if they grow or move to a network mount. Note the fix now lives at the post-#4122 location (`agents/memory/backends/deermem/deermem/core/storage.py`) after the rebase.

backend/packages/harness/deerflow/skills/installer.py
nankingjing: Both caveats are now documented in the comment on the singleton: (1) `max_workers=1` deliberately serializes concurrent installs on this path — a real behavior change from the per-call executor, acceptable because installs are rare and mostly serialized by the skill-manage lock already (the `skill_storage.py` path outside that lock is noted); (2) the pool is intentionally never shut down — the idle worker exits with the interpreter.

backend/packages/harness/deerflow/agents/memory/updater.py
nankingjing: Fixed — the warning no longer includes any payload content. It now logs only the removal count, the `newFacts` type name, and the number of keys in `update_data`, so no memory facts (potential PII) reach structured logs. The full payload still travels on the `JSONDecodeError.doc` for the immediate caller, same as before.

backend/tests/test_low_batch_2026_07.py
nankingjing: Fixed — the test now uses a literal `999` with a comment explaining that `AF_APPLETALK` is not defined on Windows and that any non-INET/INET6 integer exercises the filter branch. The misleading "platform-portable" claim is gone.

backend/app/scheduler/service.py
nankingjing: You're right — with the safety net in place, programming errors do not surface, and the description claimed otherwise. Taken the "keep it and correct the description" option: the broad `except Exception` stays (nothing supervises/restarts `_run_loop`, so raising would silently stop scheduled tasks until restart), and both the inline comments and the PR description now describe the change honestly as better diagnostics for known-transient errors plus a loudly-logged bug signal for everything else. Also noted your point that `OperationalError` covers persistent conditions (missing table, disk error), not just transient lock contention — retrying those is the pre-existing behavior, now with the concrete type in the log line to make such cases diagnosable.

backend/tests/test_sandbox_search_tools.py
nankingjing: Confirmed genuine corruption and fixed at the root: instead of spot-patching the mojibake, all three test files (plus `test_sandbox_path_patterns.py`) were rebuilt from main's pristine copies with only the `id=` → `sandbox_id=` rename re-applied — 27 keyword call sites total, nothing else. `git diff` on these files now shows exactly the rename lines; main's `E2 80 94` em-dash and `→` arrow bytes are untouched, and the whitespace churn and the two docstring line-joins are gone.

backend/packages/harness/deerflow/skills/installer.py
nankingjing: Added the one-line (well, three-line) comment covering both notes: `max_workers=1` deliberately serializes concurrent installs (behavior change vs. the per-call executor, fine because installs are rare), and the pool is never shut down — the idle worker exits with the interpreter, with the caveat that pending `asyncio.run` work can be lost on abrupt exit being inherent to that. `.result()` blocking semantics are unchanged, as you noted.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
nankingjing: Confirmed: `_load_memory_from_file` only resolves the path, opens, and `json.load`s — it never touches `self._cache_lock`, so no nested acquire is possible. Since the concurrency test patches that method out (as you noted), the invariant is now pinned in the comment directly above the critical section: "``_load_memory_from_file`` must never re-acquire ``_cache_lock`` (it is a non-reentrant ``threading.Lock``)". The serialize-cold-loads trade-off is documented in the same comment. (Post-rebase, this all lives at `agents/memory/backends/deermem/deermem/core/storage.py` — the #4122 refactor moved the file; the logic there was unchanged, so the fix ported cleanly.)

nankingjing:
@WillemJiang Conflicts are resolved — the PR is rebased onto latest `main` (69350787) as a single clean commit (952fc14b) and now reports mergeable. The conflicts came from the #4122 memory refactor, which moved `agents/memory/storage.py` and `updater.py` to `agents/memory/backends/deermem/deermem/core/`; both fixes were ported to the new locations (logic there was unchanged, so they apply 1:1), and the new regression test constructs `FileMemoryStorage(DeerMemConfig(...))` per the new DI signature.

@willem-bd All review feedback is addressed (point-by-point replies inline):

- **Dropped**: the `uploads/manager.py` fd hunk — you were right that the finding was invalid and the rewrite was the actual regression. The file is untouched now; PR retitled to "9 LOW findings".
- **Test-file corruption**: all four touched test files were rebuilt from main's pristine copies with only the `LocalSandbox(id=` → `sandbox_id=` rename re-applied (27 call sites). No mojibake bytes, whitespace churn, or docstring re-wraps remain anywhere in the diff.
- **Scheduler**: narrow catch includes `sqlalchemy.exc.OperationalError` (SQLite "database is locked" survives), broad `except Exception` safety net retained, and the comments/description no longer claim programming errors surface — they are logged loudly but the poller keeps running, since nothing supervises `_run_loop`.
- **PII**: the memory-update rejection warning logs only counts/types, never fact content.
- **Portability**: family-filter test uses a literal `999` instead of `AF_APPLETALK`.
- **Docs-in-code**: installer singleton now documents the serialize-installs trade-off and no-shutdown choice; storage documents the lock-held-across-IO trade-off and the `_load_memory_from_file` non-reentrancy invariant.

**Validation**: the new `test_low_batch_2026_07.py` passes 14/14; the affected suites (sandbox path/search tools, request-scoped secrets, memory storage/updater, skill manage/installer/storage lifecycle, scheduled task service) are green — 300 passed, 2 skipped, with 3 Windows-only path-separator failures that reproduce identically on pristine `main`. `ruff format --check` and `ruff check` are clean on all touched files.

fancyboi999 (changes requested):
Thanks @nankingjing. I found one compatibility regression in the otherwise internal cleanup batch.

[P2] Preserve the existing `id=` constructor keyword

- Location: `backend/packages/harness/deerflow/sandbox/local/local_sandbox.py:181`
- Problem: Renaming the parameter from `id` to `sandbox_id` makes every existing `LocalSandbox(id="...")` caller fail with `TypeError: unexpected keyword argument 'id'`. Updating DeerFlow's own call sites hides the break from CI but does not protect consumers of the published harness module.
- Evidence: The current `origin/main` signature explicitly accepts `id`, and the repository itself uses that keyword throughout its sandbox tests. This diff removes it solely to avoid shadowing the built-in; there is no compatibility alias or deprecation path, and positional construction is the only old form that still works.
- Suggested fix: Keep the `id` parameter name here; shadowing the built-in inside this tiny constructor is harmless and avoids an API break. If a rename is still required, add an explicit compatibility path for `id` with conflict validation and deprecate it separately.
- Test: Retain a regression that constructs `LocalSandbox(id="local")` in addition to the new positional/keyword coverage, so external-style keyword calls remain supported.


Reviewed head SHA: `952fc14b69a8ce270a82296078cd051cbf546dd4`

nankingjing:
@fancyboi999 Addressed in 8916ee6b — the rename is dropped rather than aliased, per your primary suggestion.

- **`local_sandbox.py`**: reverted to `main`'s exact blob (`aafb1c52…`) — the constructor keeps the `id` parameter, so every existing `LocalSandbox(id="...")` caller (internal or external consumers of the published harness) continues to work unchanged. No compatibility shim needed since nothing was renamed.
- **Test call sites**: the 27 `sandbox_id=` updates across `test_sandbox_search_tools.py`, `test_skill_request_scoped_secrets.py`, `test_local_sandbox_path_regex_cache.py`, and `test_sandbox_path_patterns.py` are reverted; those four files now match `main` byte-for-byte and are out of the diff entirely (PR is now 9 files).
- **Requested regression**: `test_low_batch_2026_07.py::TestLocalSandboxConstructorCompatibility` pins the public constructor shape — `test_keyword_id` constructs `LocalSandbox(id="...")` and `test_positional_id` covers positional construction — so an external-style keyword call breaking again would fail CI.
- PR title/description updated: 8 findings, with the rename recorded under "Dropped findings".

**Gates on 8916ee6b** (Windows, py312): `ruff format --check` + `ruff check` clean on all touched files; `test_low_batch_2026_07.py` 13/13; affected adjacent suites (sandbox path/search tools, request-scoped secrets, memory storage/updater + user isolation, scheduled task service, skill manage/installer/storage lifecycle, local skill storage write) — 324 passed, 2 skipped, with the same 3 pre-existing Windows-only path-separator failures that reproduce identically on pristine `main`. Revert-verification: with the storage lock fix, archive preflight, and URL-safety family filter each reverted to the base versions, their regression tests fail (4 failures); restored, all pass — the tests are load-bearing. Byte-scan of every file in the diff: no mojibake sequences, no CR/CRLF, no BOM, valid UTF-8.

@willem-bd FYI: with this commit the three test files you flagged for mojibake/churn are no longer touched at all, which supersedes the earlier "rename re-applied onto pristine copies" state.

nankingjing:

Conflict: `agents/memory/backends/deermem/deermem/core/storage.py` — upstream's `FileMemoryStorage.load` moved to the document/signature model; re-applied the single-critical-section fix against the new shape (signature computation + cache lookup + cold `_read_document` now all under `_cache_lock`) and updated the concurrency regression test to drive `_read_document`.

Checks: `ruff format --check` + `ruff check` clean on touched files; `pytest tests/test_low_batch_2026_07.py` 13/13 passed; `test_skill_request_scoped_secrets.py` + `test_sandbox_path_patterns.py` 122 passed, 1 skipped. New head: 17d3963b.

nankingjing:
All workflows on the current head (`25e832db`) are sitting at `completed / action_required` — the fork workflow gate — so nothing has executed against this head. The last head that did run (`952fc14b`, 2026-07-16) was green on Unit Tests, Frontend Unit Tests and Replay E2E (front-back contract), and this head is still `mergeable: true`.

Could a maintainer click "Approve and run workflows"? That would let CI verify the current head instead of leaving the PR blocked on a run that never starts.

nankingjing:
Status update — nothing has moved on this one since 2026-07-17.

**Every review request is addressed:**

- **@willem-bd** (2026-07-12, and twice on 2026-07-14): the `fd` double-close; the mojibake in the three touched test files, rebuilt from `main`'s pristine copies rather than spot-patched; the scheduler catch that did not actually let programming errors surface, where I took the "keep the safety net and correct the description" option; and the description claim that no longer held.
- **@WillemJiang** (2026-07-15): the conflicts with `main` are resolved — the branch now reports `mergeable: true`.
- **@fancyboi999** (2026-07-16): the `LocalSandbox(id=...)` keyword regression — the rename is dropped rather than aliased, per your primary suggestion, in `8916ee6b`.

**CI has never run on the current head.** Of the six check suites on `25e832db`, one is `completed/success` (`pr-labels`, `reviewing`, `issue-triage`) and **five are `completed/action_required` with zero runs** — queued for maintainer approval, not failed. The green check on the PR page comes from `pr-labels`, which only labels.

If you'd like any of the above re-opened, say which and I'll take it.

Issue #3804, opened 2026-06-26
[RFC] Agent eval harness for replay, trajectory, safety, and outcome regression


hetaoBackend:
> 🇨🇳 **TL;DR:** DeerFlow already has the right raw material for an agent eval harness: `RunJournal`, `RunEventStore`, replay/golden fixtures, Gateway streaming tests, sandbox/guardrail/tool metadata, and token accounting. This RFC proposes turning those pieces into a first-class `deerflow eval` workflow that can catch replay drift, trajectory regressions, safety-policy failures, artifact/output regressions, and cost/latency changes before agent-runtime changes ship.

## Motivation

DeerFlow's harness is now complex enough that ordinary unit tests are necessary but not sufficient:

- Lead-agent behavior is shaped by a long middleware chain: input sanitization, uploads/thread data, sandbox, guardrails, tool-error handling, dynamic context, skill activation, summarization, todos, token usage, title, memory, image handling, deferred MCP tools, loop detection, token budget, safety-finish handling, and clarification.
- Tool behavior spans built-ins, sandbox tools, MCP, community tools, future connectors, and subagents.
- A user-visible success often depends on the **trajectory**, not only the final answer: which tools were chosen, whether deferred tools were promoted safely, whether guardrails denied the right calls, whether artifacts were produced, whether subagent usage was attributed, and whether the stream protocol stayed stable.
- Model/config/prompt changes can silently alter cost, latency, tool ordering, event shapes, or safety posture even when the final text still looks plausible.

There is already a foundation:

- `backend/tests/test_replay_golden.py` drives the real Gateway with a deterministic `ReplayChatModel` and asserts committed SSE event-shape goldens.
- `backend/tests/_replay_fixture.py` already defines hermetic record/replay setup: fixed config, isolated `DEER_FLOW_HOME`, disabled nondeterministic memory/summarization, empty skills/MCP, and stable mode context.
- `RunJournal` captures LLM requests/responses, tool results, token usage, latency, caller buckets, and run lifecycle into `RunEventStore`.
- `RunEventStore` provides a unified message/trace event stream that can be queried after a run.

The missing piece is a cohesive eval harness over these primitives.

## Proposal

Add a first-party **agent eval harness** for DeerFlow that can run deterministic replay suites and optional live-model suites, normalize each run into a trajectory, score it with deterministic assertions first, and emit machine-readable plus human-readable reports.

A future CLI shape could be:

```bash
# Offline, deterministic, CI-friendly
[...]

Issue #3696, opened 2026-06-22
[RFC] Performance benchmark harness for runtime/config hot paths (roadmap: Core Stability → Benchmark)


ly-wang19:
> 🇨🇳 **TL;DR:** Roadmap (#1669) 把 "Benchmark" 列为 Core Stability 🔥🔥🔥🔥🔥,但目前没有任何基准设施。建议加一套 `pytest-benchmark` 的 `benchmarks/` 套件 + `make benchmark`,先覆盖 runtime/config 的每请求热路径,用于量化近期 perf PR 的收益并防止回归。下方有可运行的 PoC 与数据。**先征求 scope / 依赖 / 位置意见,再实现。**

## Motivation

The Q2 roadmap (#1669) lists **Benchmark** under *Core Stability* (🔥🔥🔥🔥🔥), but there's no benchmark infrastructure today.

Several recent fixes removed O(n) / O(n²) work on per-request hot paths — e.g. indexing `MemoryRunStore` / `MemoryRunEventStore` (#3499, #3531, #3562), and a fresh batch (#3686, #3687, #3688). These are real wins, but we currently have no way to (a) **quantify** them or (b) **stop them regressing**. A small, reproducible benchmark harness closes that gap and gives future perf work an objective yardstick.

## Proposal

1. A `backend/benchmarks/` suite using **`pytest-benchmark`** (dev-only dependency), kept **out of `make test`** so the unit suite stays fast.
2. A **`make benchmark`** target — mirroring the established `tests/blocking_io/` + `make test-blocking-io` precedent.
3. Initial coverage of the hottest runtime/config paths: `MemoryRunEventStore` reads, `RunManager.list_by_thread`, `AppConfig.get_*_config`, middleware assembly.
4. **Informational / non-gating to start** (like `detect-blocking-io`), to avoid shared-runner variance; optional `--benchmark-compare` baseline tracking can follow.

## PoC (already working locally)

`make benchmark` on current `main`, worst-case `AppConfig.get_tool_config` lookup at increasing tool counts:

```
test_get_tool_config[10]      0.67 µs    (1.0×)
test_get_tool_config[100]     2.25 µs    (3.4×)
test_get_tool_config[1000]   20.6  µs    (30.9×)   ← linear next() scan; flat once name-indexed
```

[...]

PR #4111, opened 2026-07-12
feat(test): new replay kernel phase1 of #3804

Vanzeren:
## Why

DeerFlow's agent harness is complex enough that unit tests alone can't catch regressions: a middleware 
 chain change can silently alter the SSE protocol shape; a system prompt or graph edit can stale replay 
 fixtures while tests stay green. The existing test_replay_golden.py has replay logic embedded in the   
 test file — not reusable, not importable.                                                              
                                                                                                        
 Issue #3804  proposes a phased eval harness. This PR delivers Phase 1: extract replay into a reusable   
 kernel. Design trade-offs made along the way:                                                          
                                                                                                        
 Location: packages/harness/deerflow/evals/, not tests/. The kernel defines a public API (models,       
 checks, runner) that Phase 2's deerflow eval CLI will import. Placing it in the harness package —      
 alongside deerflow.config, deerflow.sandbox, etc. — makes it a first-class library rather than         
 test-only plumbing. Suite files (eval case definitions) will live under backend/evals/ once Phase 2    
 adds them.                                                                                             
                                                                                                        
 CLI surface: pytest for now, deerflow eval deferred. run_replay_case and run_replay_suite are stable   
 Python APIs that any CLI can wrap. Adding a CLI now would couple the interface to an immature          
 trajectory schema. Deferring keeps the Phase 1 diff small and lets the scorer API stabilize first.     
                                                                                                        
 Only Gateway path in Phase 1. The existing _replay_fixture.py → create_app() → drive_gateway() path is 
 proven and sufficient. Adding DeerFlowClient replay support would duplicate the recording              
 infrastructure before the kernel itself is exercised. The ReplayRuntime callable slots (create_app,    
 drive_gateway) make it swappable later without API changes.                                            
                                                                                                        
 Minimal Trajectory schema. Phase 1 captures only what replay regression needs: events (SSE sequence),  
 replay_misses (hash misses — the primary stale-fixture signal), and metadata. Fields like tool_calls,  
 guardrail_decisions, artifacts, and token_usage are deliberately deferred to Phase 2. A narrow schema  
 now avoids churn when those types are still being designed.                                            
                                                                                                        
 Separate from benchmark RFC (#3696). metadata.generated_at provides a timestamp anchor, but Phase 1    
 does no performance comparison. The two tracks can converge in Phase 2 via shared Trajectory fields. 

## What changed

New replay eval kernel under backend/packages/harness/deerflow/evals/:                                 
                                                                                                        
 - Data models — ReplayEvalCase, Trajectory, CheckResult, ReplayEvalResult, ReplayEvalSuiteResult (all  
 frozen dataclasses)                                                                                    
 - Checks — NoReplayMissesCheck (hash-miss detection), BoundaryEventsCheck (SSE stream starts with      
 metadata, ends with end), SseShapeGoldenCheck (event-by-event golden comparison)                       
 - Runner — ReplayRuntime + run_replay_case + run_replay_suite with per-case tmpdir isolation and       
 replay-miss reset between suite iterations                                                             
 - Report — result_to_json (machine-readable), suite_summary (human-readable)                           
 - Protocols — ReplayMonkeyPatch, ReplayTmpPathFactory re-exported for caller type annotations          
                                                                                                        
 Checks return CheckResult instead of raising on missing metadata, missing event keys, empty events, or 
 corrupted golden files — the runner always gets a structured result it can report.                     
                                                                                                        
 test_replay_golden.py refactored to use the kernel. New test_evaluation_replay.py (16 tests covering   
 all check pass/fail paths). CI workflow runs both.  


## Surface area

## Validation

 cd backend && make lint   # All checks passed                                                        
   cd backend && make test   # 7141 passed, 22 skipped, 0 failed   e2e -->


## AI assistance

**Tool(s) used:** Oh-My-Pi coding harness

**How you used it:** I used omp to scan the code repo and implement this pr.


willem-bd (commented):
Thanks @Vanzeren. One finding on the replay-miss reset in the new kernel.

[P2] run_replay_case doesn't reset replay misses before driving the gateway

- Location: `backend/packages/harness/deerflow/evals/replay_runner.py:83` (`run_replay_case`)
- Problem: `_replay_misses` is a process-wide module-level list (`replay_provider.py:101`). The original `test_replay_golden.py` called `replay_provider.reset_replay_misses()` immediately before `drive_gateway` (original line 76). The extracted `run_replay_case` drops that reset — `runtime.drive_gateway(...)` on line 83 runs with no preceding `runtime.reset_replay_misses()`. Only `run_replay_suite` resets (line 122, per-case). Since the PR description states `run_replay_case` is a "stable Python API that any CLI can wrap," a caller invoking it twice in the same process gets cumulative misses: the second call's `runtime.replay_misses()` (line 91) still contains the first call's misses, so `NoReplayMissesCheck` reports a false failure or an inflated count even when the second case replayed cleanly.
- Evidence: Original test had `replay_provider.reset_replay_misses()` at line 76, before `drive_gateway` at line 78. New `run_replay_case` calls `runtime.drive_gateway` at line 83 with no reset; `runtime.reset_replay_misses()` only appears in `run_replay_suite` at line 122. CI stays green only because the committed fixture is current (zero misses), so `_replay_misses` stays empty — the bug is latent and triggers exactly in the stale-fixture scenario the miss check exists to catch.
- Suggested fix: Add `runtime.reset_replay_misses()` in `run_replay_case` right before the `runtime.drive_gateway(...)` call (before line 83). The reset in `run_replay_suite` becomes redundant but harmless, and `run_replay_case` becomes self-contained.
- Test: Add a test that calls `run_replay_case` twice in the same process with a fixture that produces a miss on the first call (or manually appends to `_replay_misses`), then asserts the second call's `result.trajectory.replay_misses` is empty.

Vanzeren:
Fixed in 4fc7b823.

- `run_replay_case()` now resets replay misses immediately before gateway execution, so every standalone case begins with isolated replay state.
- `run_replay_suite()` relies on the case-level invariant rather than maintaining a duplicate reset.
- Added a regression test that drives two cases in one process: the first records a miss and the second must report none.

Validation: `cd backend && make lint && uv run pytest tests/test_evaluation_replay.py tests/test_replay_golden.py -v` (19 passed).

backend/packages/harness/deerflow/evals/replay_runner.py
```diff
@@ -0,0 +1,133 @@
...
+def _reset_process_singletons(monkeypatch: ReplayMonkeyPatch) -> None:
+    from deerflow.config import app_config as app_config_module
+    from deerflow.config import paths as paths_module
+    from deerflow.persistence import engine as engine_module
+
+    for module, attr in (
+        (app_config_module, "_app_config"),
+        (app_config_module, "_app_config_path"),
+        (app_config_module, "_app_config_mtime"),
+        (paths_module, "_paths_singleton"),
```
willem-bd: This resets the wrong attribute. `paths.py` names the singleton `_paths` (`paths.py:398`; `get_paths()` uses `global _paths`), not `_paths_singleton`. With `raising=False`, `monkeypatch.setattr` silently creates a stray `_paths_singleton` attribute and never invalidates the real `_paths`.

The other resets here (`_app_config`, `_app_config_path`, `_app_config_mtime`, `_engine`, `_session_factory`) are all correctly named - this is an isolated typo, not a pattern.

It doesn't currently break tests because `Paths.base_dir` reads `DEER_FLOW_HOME` lazily when `_base_dir is None` (which is how `get_paths()` constructs it), so the stale cached object still resolves to the new home. But the function's stated purpose is to invalidate the paths cache, and it silently doesn't. Now that it lives in a reusable kernel rather than one test file, worth the one-line fix: `_paths_singleton` -> `_paths`.

backend/packages/harness/deerflow/evals/replay_runner.py
```diff
@@ -0,0 +1,133 @@
...
+    case: ReplayEvalCase,
+    *,
+    tmp_path: Path,
+    monkeypatch: ReplayMonkeyPatch,
+    fixture_dir: Path,
+    runtime: ReplayRuntime,
+) -> ReplayEvalResult:
+    from deerflow.config import app_config as app_config_module
+
+    fixture_path = fixture_dir / f"{case.scenario}.{case.mode}.json"
```
willem-bd, line 66: Two Phase 2 robustness notes on this line and the load below it:

1. `f"{case.scenario}.{case.mode}.json"` interpolates `scenario`/`mode` into a path. In Phase 1 these are author-defined in code, so no traversal risk. But the Phase 2 plan to load suite/case definitions from disk (`backend/evals/`) makes them externally-sourced. At that point validate them against a `[A-Za-z0-9_-]+` pattern (the same pattern `paths.py` uses for user/thread ids) before interpolating.

2. `fixture = json.loads(fixture_path.read_text(...))` (line 67) and the `fixture["prompt"]` / `fixture["context"]` access are unguarded, so a missing fixture file or a fixture missing those keys raises out of `run_replay_case`. Since `run_replay_suite` has no per-case `try/except` (`for case in cases` at line 122), one bad case aborts the whole suite. Fine while pytest isolates each case, but this kernel is the foundation for the `deerflow eval` CLI, where that UX will be poor. Worth a defensive wrap that records a failing `ReplayEvalResult` and continues.

backend/packages/harness/deerflow/evals/replay_checks.py
```diff
@@ -0,0 +1,100 @@
...
+            details={
+                "divergence_index": index,
+                "got_count": got_len,
+                "want_count": want_len,
+            },
+            relevant_slice={"got": got_window, "want": want_window},
+        )
+
+
+def run_checks(trajectory: Trajectory, checks: Iterable[object]) -> list[CheckResult]:
```
willem-bd: `Iterable[object]` (and `ReplayEvalCase.checks: tuple[Any, ...]` in `replay_models.py`) lose the check contract. The kernel defines nice Protocols for the runtime slots (`ReplayMonkeyPatch`, `ReplayTmpPathFactory`) right next to this - the check interface is the most natural place for one and it's missing.

A small `ReplayCheck` Protocol (`run(self, trajectory: Trajectory) -> CheckResult`) would document the contract and let a type-checker catch a malformed check at definition time. `ReplayRuntime`'s seven `Any` callable fields (`replay_runner.py`) would benefit from the same treatment (`Callable[...]` or Protocols) before Phase 2 externalizes the API.

backend/packages/harness/deerflow/evals/replay_checks.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class SseShapeGoldenCheck:
+    name = "sse_shape_golden"
+
+    def run(self, trajectory: Trajectory) -> CheckResult:
+        try:
+            golden_path = Path(trajectory.metadata["golden_path"])
+            golden = json.loads(golden_path.read_text(encoding="utf-8"))["events"]
+        except (KeyError, FileNotFoundError, json.JSONDecodeError, TypeError) as exc:
```
willem-bd, line 69: This `except` branch (missing `golden_path` key, nonexistent file, malformed JSON) is untested. The PR description explicitly calls out "corrupted golden files" as handled, so the path is claimed but not covered. Worth adding cases for: missing `golden_path` in metadata, nonexistent golden file, and malformed JSON - each should return a failing `CheckResult` with the `cannot load golden` message rather than raise.

backend/packages/harness/deerflow/evals/replay_checks.py
```diff
@@ -0,0 +1,100 @@
+from __future__ import annotations
+
+import json
+from collections.abc import Iterable
+from pathlib import Path
+
+from deerflow.evals.replay_models import CheckResult, Trajectory
+
+
+class NoReplayMissesCheck:
```
willem-bd, line 15: The original `test_replay_golden.py` carried a rationale comment that didn't survive the move: *"The gateway swallows a hash-miss into a normal assistant error message, so the SSE shapes below stay green on a stale fixture - the miss list is the only reliable signal at this layer."* That *why* is now only partially recoverable from this check's message string. Worth preserving as a docstring on the class (or the module) so future contributors understand why this check exists separately from `SseShapeGoldenCheck` - it's non-obvious that the golden shape can stay green while the fixture is stale.

backend/tests/test_evaluation_replay.py
```diff
@@ -0,0 +1,304 @@
...
+
+    first_tmp_path = tmp_path / "first"
+    second_tmp_path = tmp_path / "second"
+    first_tmp_path.mkdir()
+    second_tmp_path.mkdir()
+
+    first = run_replay_case(case, tmp_path=first_tmp_path, monkeypatch=monkeypatch, fixture_dir=FIXTURE_DIR, runtime=runtime)
+    second = run_replay_case(case, tmp_path=second_tmp_path, monkeypatch=monkeypatch, fixture_dir=FIXTURE_DIR, runtime=runtime)
+
+    assert first.trajectory.replay_misses == ["first-call-miss"]
```
willem-bd, line 246: This test sets up a first call with a miss and `checks=(NoReplayMissesCheck(),)`, so `first.passed` should be `False` and `first.failure_count == 1` - but only the `replay_misses` list is asserted. If `run_replay_case` had a bug returning `passed=True` always, this test would still pass. Add `assert first.passed is False` / `first.failure_count == 1`.

Separately, `test_run_replay_suite_aggregates_counts` (line 235) only exercises the all-pass suite. No test runs a failing case through `run_replay_suite` end-to-end, so the suite's failure aggregation (`failed_count = len - passed_count`, `overall_passed = failed_count == 0`) is only validated for the all-pass path. A second case with a stubbed miss would cover the failing-suite path.

willem-bd (commented):
Replay-kernel extraction looks sound - clean public API, the refactor preserves (and improves via structured `CheckResult`) the original regression guards, and docs/CI are updated in the same change set. Approve with the following small follow-ups; none are blockers.

Main actionable:
1. `replay_runner.py:48` - `_paths_singleton` is the wrong attribute name (should be `_paths`); the reset is currently a silent no-op.
2. `replay_checks.py:60` - the `SseShapeGoldenCheck` error branch is untested despite being a claimed behavior.
3. `test_evaluation_replay.py:230` - failure-aggregation through the runner/suite isn't asserted; add `passed`/`failure_count` asserts and a failing-case suite run.
4. `replay_checks.py:99` - add a `ReplayCheck` Protocol and tighten `ReplayRuntime`'s `Any` fields before Phase 2 externalizes the API.

Minor / forward-looking (see inline): preserve the lost "why replay misses matter" comment on `NoReplayMissesCheck`; Phase 2 will need `scenario`/`mode` validation and per-case suite resilience.

Minor note (not posted inline): the `@dataclass(frozen=True)` models are only shallowly immutable - the `dict`/`list` fields are still mutable, so `frozen` prevents reassignment but not `events.append(...)`. Fine for data containers; a one-line docstring would prevent future deep-constness assumptions.

Vanzeren:
1. _paths_singleton → _paths — confirmed the actual attribute in paths.py:398.                       
2. SseShapeGoldenCheck error branch — added three tests: missing key, file not found, corrupted      
JSON.                                                                                                
3. Suite failure aggregation — added test_run_replay_suite_reports_failure_counts asserting          
 passed_count=0, failed_count=1, overall_passed=False.                                                
4. ReplayCheck Protocol added in replay_checks.py; ReplayRuntime Any fields tightened to             
Callable[..., T].

backend/packages/harness/deerflow/evals/replay_runner.py
```diff
@@ -0,0 +1,134 @@
...
+def _reset_process_singletons(monkeypatch: ReplayMonkeyPatch) -> None:
+    from deerflow.config import app_config as app_config_module
+    from deerflow.config import paths as paths_module
+    from deerflow.persistence import engine as engine_module
+
+    for module, attr in (
+        (app_config_module, "_app_config"),
+        (app_config_module, "_app_config_path"),
+        (app_config_module, "_app_config_mtime"),
+        (paths_module, "_paths"),
```
willem-bd, line 49: Heads up for reviewers: this isn't an incidental rename. The deleted `test_replay_golden.py` reset `paths_module._paths_singleton`, but `paths.py` has no such attribute - the real singleton is `_paths` (`paths.py:398`, used by `get_paths()`). With `raising=False`, the old `monkeypatch.setattr` was a silent no-op: it created a spurious `_paths_singleton` attr and never touched `_paths`, so the paths cache was never invalidated between runs. That was masked because the old file drove the gateway exactly once per process. This new kernel is now load-bearing for isolation: `test_evaluation_replay.py` drives the gateway ~6x in one process with different `DEER_FLOW_HOME`s, so without this fix the second drive would reuse the first drive's cached `Paths()`. Good catch - just flagging it so it isn't read as cosmetic.

backend/packages/harness/deerflow/evals/replay_checks.py
```diff
@@ -0,0 +1,112 @@
...
+            golden = json.loads(golden_path.read_text(encoding="utf-8"))["events"]
+        except (KeyError, FileNotFoundError, json.JSONDecodeError, TypeError) as exc:
+            return CheckResult(
+                name=self.name,
+                passed=False,
+                message=f"cannot load golden: {exc}",
+                relevant_slice=str(exc),
+            )
+
+        if trajectory.events == golden:
```
willem-bd, line 77: This compares full event-dict equality, not just "shape" (event name + payload keys). Today that's equivalent to a shape check because replay events are shape-only (`{"event": ..., "keys": [...]}`, and `test_run_replay_case_builds_structured_result` asserts `set(events[0]) == {"event","keys"}`). Fine for Phase 1. But the name + message now codify a public API for Phase 2: if Phase 2 events ever carry value-bearing fields, this will fail on value drift that the name implies it should ignore. A one-line docstring clarifying "compares the full committed event sequence (event name + payload keys), not just event names" would pin the semantics before Phase 2 builds on it.

backend/packages/harness/deerflow/evals/replay_models.py
```diff
@@ -0,0 +1,58 @@
...
+# Trajectory.events.append(...) is still possible). This is intentional for
+# data-container usage; do not assume deep constness.
+
+
+@dataclass(frozen=True)
+class ReplayEvalCase:
+    id: str
+    scenario: str
+    mode: str
+    checks: tuple[object, ...]
```
willem-bd: Typed as `object` (to avoid importing `ReplayCheck` from `replay_checks` and creating a cycle, since `replay_checks` imports `replay_models`). Passing a non-`ReplayCheck` fails at runtime (`AttributeError: .run`) instead of at type-check. Consider moving the `ReplayCheck` protocol into `replay_models.py` (or a tiny `replay_protocols.py`) so this can be `tuple[ReplayCheck, ...]` - the cycle is avoidable since the protocol only needs `CheckResult`/`Trajectory`, which live here.

backend/packages/harness/deerflow/evals/replay_runner.py
```diff
@@ -0,0 +1,134 @@
...
+        scenario=case.scenario,
+        mode=case.mode,
+        prompt=fixture["prompt"],
+        context=fixture["context"],
+        events=events,
+        replay_misses=runtime.replay_misses(),
+        metadata={
+            "fixture_path": str(fixture_path),
+            "golden_path": str(events_path),
+            "generated_at": datetime.now(UTC).isoformat(),
```
willem-bd: `datetime.now(UTC).isoformat()` makes `result_to_json()` output non-reproducible. Harmless in Phase 1 (no JSON golden), but if Phase 2 ever golden-compares trajectories or diffs reports, this will flake. Either drop it, isolate it under a `"provenance"` sub-dict, or mark it as intentionally volatile in the docstring.

.github/workflows/replay-e2e.yml
```diff
@@ -62,7 +64,7 @@ jobs:
         run: uv sync --group dev
       - name: Replay golden (backend SSE contract)
         working-directory: backend
-        run: PYTHONPATH=. uv run pytest tests/test_replay_golden.py -v
+        run: PYTHONPATH=. uv run pytest tests/test_replay_golden.py tests/test_evaluation_replay.py -v
```
willem-bd, line 67: Minor CI note: `test_evaluation_replay.py` adds ~6 full-gateway spins (`create_app()` + `drive_gateway`), and `test_run_replay_case_builds_structured_result` + `test_kernel_result_matches_existing_fixture_paths` both drive the same `write_read_file.ultra` fixture with `SseShapeGoldenCheck`, overlapping `test_replay_golden.py`'s golden assertion (~7 spins now vs 1 before). It's path-gated so cost is scoped, and the two files test different surfaces, so acceptable for Phase 1 - just a candidate for dedup once the kernel is trusted.

PR #4113, opened 2026-07-12
fix(agents): surface silent truncation markers and LRU-evict stop-reason state

nankingjing:
## Summary

Five architectural defects from the agent-core hunt land in one batch. All touch the
`ThreadState` reducer / durable-context surface that the lead agent relies on for
no-op-writes invariant, truncation observability, and LRU-style stop-reason eviction.

### D1 — `merge_viewed_images`: empty dict was a state-corrupting reducer

`thread_state.py:59-73` (before) treated `{}` as an explicit "clear all" signal.
A normal `Command(update={"viewed_images": {}})` — the local idiom for a no-op
reset — silently wiped the entire image history with no log line.

The sibling reducers (`merge_todos`, `merge_goal`, `merge_promoted`,
`merge_delegations`, `merge_skill_context`) all share the "no-op writes are
no-ops" invariant. This reducer broke it for one channel only.

**Fix**: empty / `None` writes now preserve existing, matching the sibling
invariant. There is deliberately no clear sentinel: a review pass confirmed no
caller needs explicit clearing today (an earlier `CLEAR_VIEWED_IMAGES` sentinel
was removed as speculative), so clearing requires an explicit opt-in added
alongside the caller that needs it.

### D2 — `merge_delegations`: 50-entry cap was silent

`thread_state.py:165-167` (before) sliced to the most-recent 50 entries with no
marker. Once 51+ delegations happened, the oldest were silently dropped from
durable context the lead re-uses on every continuation, with no signal that
history was truncated. The lead may re-delegate a task whose prior completion is
no longer in the visible ledger.

**Fix**: add a parallel `delegations_truncated_count` channel (max-reducer). The
capture middleware counts evictions **from the merge inputs themselves** — the
net-new delegation ids that overflow the existing capped ledger — and persists
`existing_count + increment`, so the count stays correct even after
summarization compacts old task messages out of `state["messages"]` (it is never
reconstructed by re-scanning the compactable message list). The renderer appends
`... (+N earlier delegations dropped permanently at the 50-entry durable ledger
cap)` so the lead detects the loss; the wording is deliberately distinct from
the render-budget "omitted from this model view" line.

### D3 — Delegation description hard-truncation was invisible

`delegation_ledger.py:17-18, 113` truncated at 200 chars mid-sentence. The model
that reads the durable ledger uses that string to decide whether to re-delegate
or reuse. A `summarize the Q3 payroll report by department` prompt silently
became `summarize the Q3 payroll repo` on the 201st character.

**Fix**: append `... [truncated]` when the source exceeds the cap, and surface
the marker on render.

### D4 — `BoundedDict` was insertion-FIFO, not LRU

`_bounded_dict.py:28-32` only evicted on a NEW key insertion. Updating an existing
key overwrote the value but never moved the key to the tail, so eviction was
insertion-FIFO. Used by `TokenBudgetMiddleware._stop_reason` and
`LoopDetectionMiddleware._stop_reason` (both `maxsize=1000`). Combined with reuse
of `run_id` across replacement runs, a previously-active `run_id` could be evicted
first while a never-touched entry that just arrived was kept.

**Fix**: refresh recency via `move_to_end` on existing-key writes. Mirrors the
explicit pattern already used by `LoopDetectionMiddleware._touch_pending_warning_key_locked`.

### D5 — Skill description truncation was enforced at 4 sites

`thread_state.py:188` (reducer) and three consumers in `skill_context.py`
silently chopped skill descriptions to 500 chars at independent sites. Adding a
new caller risks forgetting the cap (state leak); changing the cap requires
touching four files.

**Fix**: centralize into `description_for_skill_entry(raw) -> (text, truncated)`
so all four readers/writers share the same logic and surface `... [truncated]`
on render.

## Why a single batch

- All five defects touch the same ThreadState surface; reviewers see one diff.
- D1 / D2 / D3 / D5 are variants of the same "silent truncation is the bug"
  pattern; D4 is the LRU bug in the related eviction path. Fixing the pattern
  first makes the next defect obvious.
- Tests are already cross-cutting (`test_thread_state_reducers.py` covers D1 +
  D2; `test_durable_context_middleware.py` covers D2 + D5; the existing
  `test_loop_detection_middleware.py` / `test_token_budget_middleware.py` /
  `test_terminal_response_middleware.py` exercise D4 through the live middlewares,
  plus the dedicated `test_bounded_dict.py`).

## Test plan

- [x] `pytest tests/test_bounded_dict.py tests/test_thread_state_reducers.py tests/test_delegation_ledger.py tests/test_durable_context_middleware.py tests/test_loop_detection_middleware.py tests/test_token_budget_middleware.py tests/test_terminal_response_middleware.py tests/test_skill_context.py tests/test_summarization_middleware.py tests/test_summarization_summary_text.py tests/test_replay_golden.py tests/test_assistant_payload_replay.py` — 285 tests pass.
- [x] New tests:
  - `test_empty_dict_is_noop` — empty-dict / `None` writes preserve existing (D1)
  - `test_bounded_dict.py` — 5 LRU recency-refresh cases incl. updated-oldest-key edge (D4)
  - `test_truncated_count_*` / `TestBoundDescription` — truncation markers asserted in rendered output (D2 / D3)
  - `test_compaction_survivor_eviction_still_increments_truncated_count` — post-compaction eviction still increments the counter and renders the permanent-drop marker (D2)
  - `test_truncated_count_accumulates_across_captures` — `existing_count + increment` write under the max-reducer (D2)
- [x] CI: `ruff format --check` / `ruff check` clean.

## What I did NOT do

- No new dependencies.
- No public API changes.
- No schema migrations (`delegations_truncated_count` rides an existing checkpoint
  schema; the missing field is handled by the `merge_delegations_truncated_count`
  reducer defaulting to `0`).
- No harness / app boundary violations.


willem-bd (commented):
Thanks @nankingjing. Two test-coverage gaps on the new behavior this PR introduces.

[P2] BoundedDict LRU recency-refresh (D4) has no direct test

- Location: `backend/packages/harness/deerflow/agents/middlewares/_bounded_dict.py:33-42`
- Problem: The core behavior change — `move_to_end` on existing-key writes so eviction is LRU not FIFO — is not directly tested. The only eviction test (`test_abandoned_run_state_is_bounded` in `test_terminal_response_middleware.py:264`) inserts 1001 distinct keys and checks the oldest is dropped; it never updates an existing key and then adds new keys to verify the updated key survives. If `move_to_end` is later removed, every existing test still passes.
- Evidence: Searched `backend/tests/` for any test that writes the same BoundedDict key twice then fills to capacity; none exists. The middleware tests exercise `__setitem__` but always with fresh `run_id` keys, so the recency-refresh branch (`if key in self: self.move_to_end(key)`) is never asserted.
- Suggested fix: Add a unit test for `BoundedDict` directly (or extend `test_abandoned_run_state_is_bounded`): insert keys A, B; update A; insert C to trigger eviction; assert B is evicted and A survives. Also test capacity=1 and the boundary where the updated key is the one that would have been evicted.
- Test: `bd = BoundedDict(maxsize=2); bd["a"]=1; bd["b"]=2; bd["a"]=10; bd["c"]=3; assert "a" in bd and "b" not in bd`

[P2] Truncation markers (D2/D3/D5) are not asserted in rendered output

- Location: `backend/packages/harness/deerflow/agents/middlewares/delegation_ledger.py:186` (`render_delegation_ledger` with `truncated_count`), `delegation_ledger.py:54` (`_bound_description`), `thread_state.py:77` (`description_for_skill_entry`)
- Problem: The PR's stated goal is to surface silent truncation, but no test asserts the `... [truncated]` or `... (+N earlier delegations dropped from this ledger)` markers actually appear in rendered output. The D2 test (`test_repeated_capture_after_cap_does_not_reemit_evicted_old_delegation`) only asserts `out.get("delegations_truncated_count") == 1` on the state update — it never calls `_render_durable_context_data` or `render_delegation_ledger` to verify the marker reaches the model-visible string. Likewise no test checks that a delegation description >200 chars or a skill description >500 chars contains the truncation suffix.
- Evidence: Searched `test_delegation_ledger.py`, `test_durable_context_middleware.py`, `test_skill_context.py`, and `test_thread_state_reducers.py` for `[truncated]` or `dropped from this ledger`; zero matches. Existing description-cap tests only assert `len(description) <= 500`, not marker presence.
- Suggested fix: Add tests that call `render_delegation_ledger(entries, truncated_count=2)` and assert the output contains `(+2 earlier delegations dropped`. Add a test that a 201-char delegation description produces `... [truncated]` in the stored entry. Add a test that `_render_durable_context_data(..., truncated_count=1)` includes the marker in the rendered block.
- Test: `assert "(+1 earlier delegations dropped" in render_delegation_ledger(entries, truncated_count=1)`; `assert "... [truncated]" in _bound_description("x" * 201)`

WillemJiang:
@nankingjing  Please take a look at the review comments and fix the unit test errors.

nankingjing:
Added the missing tests: BoundedDict LRU recency-refresh (5 cases) and delegation ledger description truncation + truncated_count rendering.

nankingjing:
Both P2 gaps from willem-bd's review are already addressed in the latest push — test_bounded_dict.py covers LRU recency-refresh (5 cases), and test_delegation_ledger.py now asserts the truncation markers in rendered output. The CI checks need approval to run though (action_required on 5 suites).

WillemJiang:
@nankingjing, please take a look at the failed CI tests.

willem-bd (commented):
Review nits (ruff format + lint and the PR test plan - 213 tests - are already green on this head, so these are suggestions, not blockers). Line-level notes below.

backend/packages/harness/deerflow/agents/middlewares/durable_context_middleware.py
```diff
@@ -232,7 +232,23 @@ def _capture_delegations(self, state: AgentState, runtime: Runtime | None) -> di
...
+    def _compute_truncated_count(self, state: AgentState) -> int | None:
+        """Compute the number of delegation entries silently dropped from the ledger.
+
+        Only emits a non-None count when the channel currently holds exactly the
+        cap of entries AND the in-message observation exceeds the cap. This keeps
+        the parallel ``delegations_truncated_count`` channel stable across
+        checkpoint round-trips: it does not regress when the ledger shrinks back
+        below the cap on a fresh thread. (D2 in the agent-core hunt.)
+        """
+        observed = extract_delegations(state["messages"])
```
willem-bd: Minor perf: `extract_delegations(state["messages"])` walks the full message list on every `before_model` call, even though the vast majority of threads never reach the 50-entry cap. The scan runs *before* the `len(existing) >= _DELEGATION_LEDGER_MAX_ENTRIES` guard, so the common case pays an O(messages) cost each turn on the hot path. Suggested: guard first and return early when the ledger isn't at cap:

```python
existing = state.get("delegations") or []
if len(existing) < _DELEGATION_LEDGER_MAX_ENTRIES:
    return None
observed = extract_delegations(state["messages"])
...
```

backend/packages/harness/deerflow/agents/middlewares/durable_context_middleware.py
```diff
@@ -232,7 +232,23 @@ def _capture_delegations(self, state: AgentState, runtime: Runtime | None) -> di
             existing,
         )
         if delegations:
-            return {"delegations": delegations}
+            updates: dict = {"delegations": delegations}
```
willem-bd: This `updates` local isn't extended anywhere - it's equivalent to the original `return {"delegations": delegations}`. Either revert to the one-liner, or fold the `delegations_truncated_count` write into this dict so `_capture` issues a single state update.

backend/packages/harness/deerflow/agents/middlewares/durable_context_middleware.py
```diff
@@ -232,7 +232,23 @@ def _capture_delegations(self, state: AgentState, runtime: Runtime | None) -> di
...
+
+        Only emits a non-None count when the channel currently holds exactly the
+        cap of entries AND the in-message observation exceeds the cap. This keeps
+        the parallel ``delegations_truncated_count`` channel stable across
+        checkpoint round-trips: it does not regress when the ledger shrinks back
+        below the cap on a fresh thread. (D2 in the agent-core hunt.)
+        """
+        observed = extract_delegations(state["messages"])
+        existing = state.get("delegations") or []
+        if len(observed) > _DELEGATION_LEDGER_MAX_ENTRIES and len(existing) >= _DELEGATION_LEDGER_MAX_ENTRIES:
```
willem-bd: Edge case: the `len(existing) >= cap` guard means a single turn that emits >50 delegations before the ledger reaches cap won't surface a count that turn (and never will if the thread ends right after). The `max`-reducer picks it up on a later capture if more delegations occur, so this is low-risk - just worth a sentence in the docstring noting the gap.

backend/packages/harness/deerflow/agents/middlewares/delegation_ledger.py
```diff
@@ -191,6 +220,16 @@ def render_delegation_ledger(entries: list[DelegationEntry], *, max_chars: int =
         if _fits_budget(lines, omitted_line, max_chars):
             lines.append(omitted_line)
 
+    if truncated_count > 0:
+        marker_line = f"- ... (+{truncated_count} earlier delegations dropped from this ledger because of cap)"
```
willem-bd: Heads-up: this `marker_line` ("… dropped from this ledger because of cap") can render alongside the pre-existing `omitted_line` ("… omitted from this model view because of context budget") when both the durable cap and the render budget are hit. They're semantically distinct but similarly worded - consider tightening the wording so the model doesn't conflate durable-cap truncation with render-budget omission.

backend/packages/harness/deerflow/agents/thread_state.py
```diff
@@ -65,20 +65,35 @@ def merge_artifacts(existing: list[str] | None, new: list[str] | None) -> list[s
...
 
 
+# Sentinel callers must pass to opt into clearing viewed_images. The default
+# "no-op writes are a no-op" invariant (matching merge_todos / merge_goal /
+# merge_promoted / merge_delegations / merge_skill_context) is restored so a
+# reducer-write of ``{}`` does NOT silently wipe the agent's vision context.
+# The legacy "empty dict clears" behavior was a footgun: any node returning
+# ``viewed_images={}`` accidentally wiped the entire image history with no log
+# line and no compensating re-fetch (D1 in the agent-core hunt).
+CLEAR_VIEWED_IMAGES: dict[str, ViewedImageData] = {"__deerflow_clear__": {"base64": "", "mime_type": ""}}
```
willem-bd: Two small notes on this sentinel: (1) it's a public mutable module-level dict, so a caller that mutates it corrupts it globally - low risk by convention, but a comment or copy-on-use would harden it; (2) it currently has zero callers in the tree, so the explicit-clear path is exercised only by its unit test. If it's meant as a stable public API, a line in AGENTS.md would help; otherwise worth confirming the sentinel is pulling its weight.

backend/tests/test_delegation_ledger.py
```diff
@@ -458,3 +458,70 @@ def test_budgeted_render_keeps_newest_delegations(self):
...
+        result = _bound_description("short")
+        assert result == "short"
+        assert _bound_description.__defaults__ is not None  # cap has a default
+
+    def test_long_description_is_truncated_with_suffix(self):
+        from deerflow.agents.middlewares.delegation_ledger import _bound_description
+
+        long = "x" * 201
+        result = _bound_description(long)
+        assert len(result) <= 201
```
willem-bd: Minor: `len(result) <= 201` is looser than the actual contract (`== _DESCRIPTION_CAP`, i.e. 200). Pinning to `== 200` would catch drift in the suffix math.

backend/packages/harness/deerflow/agents/middlewares/durable_context_middleware.py
nankingjing: Good catch - fixed in 86a0dad. The below-cap ledger guard now runs first and returns early, so the common under-cap case skips the extract_delegations scan entirely on the hot before/after-model path.

backend/packages/harness/deerflow/agents/middlewares/durable_context_middleware.py
nankingjing: Went with your second option in 86a0dad: the delegations_truncated_count write is folded into this dict, so the updates local now earns its keep and _capture issues a single merged state update. Side effect: after_model also surfaces the count now, which just means it lands one hook earlier (the channel's max-reducer keeps it idempotent).

backend/packages/harness/deerflow/agents/middlewares/durable_context_middleware.py
nankingjing: Agreed it's low-risk; documented in 86a0dad. The docstring now notes the single-turn over-cap gap and that the max-reducer picks the count up on a later capture once the merged ledger sits at the cap.

backend/packages/harness/deerflow/agents/middlewares/delegation_ledger.py
nankingjing: Reworded in 86a0dad: the durable-cap marker now reads 'dropped permanently at the 50-entry durable ledger cap' (short-budget fallback: 'earlier permanently dropped'), so it can no longer be conflated with the render-budget 'omitted from this model view' line when both render. Docstrings in thread_state.py were synced to the new wording.

backend/packages/harness/deerflow/agents/thread_state.py
nankingjing: You're right that it wasn't pulling its weight - I searched the tree and CLEAR_VIEWED_IMAGES had zero callers outside its own unit test (its payload didn't even match the current ViewedImageData schema, which confirms it was speculative). Removed the sentinel and its test in 86a0dad, keeping only the no-op-write invariant fix; the reducer docstring now says clearing requires an explicit opt-in added alongside a caller that actually needs it. backend/AGENTS.md updated to describe the new merge-only semantics.

backend/tests/test_delegation_ledger.py
nankingjing: Pinned in 86a0dad: the assertion is now len(result) == _DESCRIPTION_CAP (imported from the module), so any drift in the suffix math fails the test.

fancyboi999 (changes requested):


Reviewed head SHA: `8de0e249cb84a64b0e674dd42a58c457f45e01ac`

### [P2] Count ledger evictions after message compaction

- Location: `backend/packages/harness/deerflow/agents/middlewares/durable_context_middleware.py:243-266`
- Problem: `delegations_truncated_count` is inferred from the current raw message list instead of from entries actually evicted by the durable ledger. After summarization compacts old task messages, a new delegation can evict an existing ledger entry while this counter stays at zero or a stale value, so the model-visible marker does not reliably surface the loss this change is meant to expose.
- Evidence: the ledger can retain 50 durable entries while `state["messages"]` contains only the compacted recent tail. On the next new task, `_capture_delegations()` emits a new entry and `merge_delegations()` drops the oldest one, but `_compute_truncated_count()` sees at most the recent-tail task calls; `len(observed) <= 50` returns `None`. The existing regression covers an un-compacted message list with more than 50 visible calls, so it misses this production path.
- Suggested fix: derive the increment from the merge inputs themselves, for example by counting net-new delegation IDs that overflow the existing capped ledger, and add that increment to the persisted counter. Do not reconstruct lifetime evictions from compactable messages.
- Test: start with a 50-entry durable ledger, a compacted message tail containing only one new task call, capture and reduce the update, then assert the oldest entry is evicted and `delegations_truncated_count` increments to 1 and renders the permanent-drop marker.

nankingjing:
@fancyboi999 Thanks — you were right that the counter was reconstructed from a compactable source. Fixed in 7931428.

**What changed** (`durable_context_middleware.py`):
- Removed `_compute_truncated_count` (the `state["messages"]` recount) entirely. Replaced with `_count_ledger_evictions(new_delegations, existing)`, which derives the increment from the merge inputs themselves: the net-new delegation ids that overflow the existing capped ledger, matching `merge_delegations` union-by-id + keep-newest semantics.
- `_capture_delegations` persists `existing_count + increment` in the same single state update as the `delegations` write; the channel's max-reducer keeps replayed writes idempotent.
- Side effects: the single-turn over-cap gap willem-bd asked me to document is now closed (the overflow surfaces on the same capture, not a later one), and the O(messages) at-cap rescan is gone from the hot before/after-model path.

**Tests** (all fail against the previous head, verified by revert):
- `test_compaction_survivor_eviction_still_increments_truncated_count` — your exact scenario: 50-entry durable ledger, compacted message tail containing only one new task call → oldest entry evicted, `delegations_truncated_count` increments to 1, and `render_delegation_ledger` emits the `(+1 earlier delegations dropped permanently ...)` marker.
- `test_truncated_count_accumulates_across_captures` — pins the `existing_count + increment` write (3 → 4).
- `test_repeated_capture_after_cap_does_not_reemit_evicted_old_delegation` — updated: the same-capture overflow now surfaces `delegations_truncated_count == 1` on the first capture, and a capture with no ledger change emits no update (no phantom recount).

`backend/AGENTS.md` ThreadState section now lists `delegations_truncated_count` and documents capture-time eviction counting; the PR description was updated to match.

**Gates on 7931428** (local, `df312` / Python 3.12.13, ruff 0.15.21):
- `ruff format --check` + `ruff check` — clean
- `pytest` on `test_bounded_dict.py`, `test_thread_state_reducers.py`, `test_delegation_ledger.py`, `test_durable_context_middleware.py`, `test_loop_detection_middleware.py`, `test_token_budget_middleware.py`, `test_terminal_response_middleware.py`, `test_skill_context.py`, `test_summarization_middleware.py`, `test_summarization_summary_text.py`, `test_replay_golden.py`, `test_assistant_payload_replay.py` — **285 passed**

CI will need re-approval after this push.

nankingjing:

Conflicts: `backend/AGENTS.md` only (docs) — kept upstream's `merge_message_writes` paragraph and merged the branch's updated `merge_viewed_images`/`delegations_truncated_count` wording. Code files (`thread_state.py`, `test_thread_state_reducers.py`) auto-merged.

Checks: `ruff format --check` + `ruff check` clean on touched files; `pytest tests/test_thread_state_reducers.py tests/test_bounded_dict.py tests/test_delegation_ledger.py tests/test_durable_context_middleware.py` 117/117 passed. New head: 19cb6907.

WillemJiang (dismissed):

WillemJiang:
@nankingjing  please fix the failed CI tests.

nankingjing:
@WillemJiang — you asked three times (2026-07-13 ×2 and 2026-07-29) and never got an answer from me. Sorry about that. Here is the state on the current head `d417ae1e`.

**What the last executed CI run actually reported.** The newest head that ever ran is `f0bf1e48` (2026-08-16); there, Lint Check, Unit Tests, Replay E2E (front-back contract), Frontend Unit Tests and Backend Blocking IO all failed. The head you commented on, `a1be8519` (2026-07-29), had Lint Check green but Unit Tests and Replay E2E red.

**What I verified on the current head `d417ae1e`:**
- `ruff check backend` — clean
- the PR's four touched test files plus the whole replay suite — **146 passed**, including `test_replay_write_read_file_ultra_matches_golden`

I did not re-run the full `Unit Tests` job locally, so I am not claiming that one is green — only the suites this PR touches, plus replay.

**Why it still looks red:** every workflow on `d417ae1e` is sitting at `completed / action_required` — the fork workflow gate. Nothing has executed on this head at all, so the badge is still showing the August result. Could a maintainer click "Approve and run workflows" so the current head actually gets tested?

PR #4114, opened 2026-07-12
fix(front): Show assistant text during tool steps

hetaoBackend:
## Summary

Follow-up to #4027 / #4033 for the frontend rendering path.

Tool-calling AI messages can include user-visible text, for example an error explanation followed by a note that the assistant will try another approach. The message grouping code already keeps those messages in `assistant:processing`, but `MessageGroup` only rendered reasoning and tool-call steps, so the visible text was dropped from the UI.

This PR:

- extracts visible content from AI messages that also contain `tool_calls`
- renders that content as a processing step with the existing markdown renderer
- keeps assistant-text steps visible even when older tool/reasoning steps are collapsed
- adds a component regression test for the hidden-text case
- documents the frontend message-rendering convention in `frontend/AGENTS.md`

## Validation

- `cd frontend && pnpm test` — 605 passed
- `cd frontend && pnpm check`
- `cd frontend && pnpm exec prettier --check src/components/workspace/messages/message-group.tsx tests/unit/components/workspace/messages/message-group.test.ts AGENTS.md`
- `git diff --check`
- `PYTHONPATH=/Users/hetao/workspace/deer-flow/.worktrees/issue-4027-visible-final-response/backend/packages/harness:/Users/hetao/workspace/deer-flow/.worktrees/issue-4027-visible-final-response/backend /Users/hetao/workspace/deer-flow/backend/.venv/bin/python -m pytest backend/tests/test_terminal_response_middleware.py` — 14 passed, 1 existing LangChain deprecation warning

### DIFF
--- a/frontend/src/components/workspace/messages/message-group.tsx
+++ b/frontend/src/components/workspace/messages/message-group.tsx
@@ -8,6 +8,7 @@ import {
   LightbulbIcon,
   ListTodoIcon,
   MessageCircleQuestionMarkIcon,
+  MessageSquareTextIcon,
   NotebookPenIcon,
   SearchIcon,
   SquareTerminalIcon,
@@ -28,6 +29,7 @@ import { useI18n } from "@/core/i18n/hooks";
 import { formatTokenCount } from "@/core/messages/usage";
 import type { TokenDebugStep } from "@/core/messages/usage-model";
 import {
+  extractContentFromMessage,
   extractReasoningContentFromMessage,
   findToolCallResult,
 } from "@/core/messages/utils";
@@ -96,6 +98,11 @@ export function MessageGroup({
     }
     return [];
   }, [lastToolCallStep, steps]);
+  const collapsibleAboveLastToolCallSteps = useMemo(
+    () =>
+      aboveLastToolCallSteps.filter((step) => step.type !== "assistantText"),
+    [aboveLastToolCallSteps],
+  );
   const lastReasoningStep = useMemo(() => {
     if (lastToolCallStep) {
       const index = steps.indexOf(lastToolCallStep);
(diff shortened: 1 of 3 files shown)


hetaoBackend:
Context on overlap with #4033:

#4033 fixed the backend/agent-runtime part of #4027: an empty terminal `AIMessage` after tool execution is retried once and then converted into a visible fallback instead of being persisted as a silent success.

This PR is still a separate frontend follow-up. The original issue also called out intermediate AI messages that contain both user-visible text and `tool_calls`. `core/messages/utils.ts` already groups those messages as `assistant:processing`, but `MessageGroup` only rendered reasoning/tool-call steps, so the text itself stayed hidden in the UI.

I rechecked the latest `upstream/main`: it has #4033, but it does not yet include the `assistantText` rendering path or the new `message-group.test.ts` regression. So #4114 is not duplicating #4033; it covers the remaining frontend visibility gap for text attached to tool-calling AI messages.

frontend/src/components/workspace/messages/message-group.tsx
```diff
@@ -699,12 +734,25 @@ interface CoTToolCallStep extends GenericCoTStep<"toolCall"> {
...
   const steps: CoTStep[] = [];
-  for (const message of messages) {
+  for (const [messageIndex, message] of messages.entries()) {
     if (message.type === "ai") {
+      const content = extractContentFromMessage(message);
+      if (content && message.tool_calls?.length) {
+        steps.push({
+          id: `${message.id ?? `ai-${messageIndex}`}-content`,
+          messageId: message.id,
+          type: "assistantText",
```
willem-bd, line 752: Non-blocking — ordering thought: `assistantText` is pushed *before* `reasoning`, so for a message that carries content + reasoning + tool_calls the visible order becomes text → reasoning → tool call. Chronologically the model usually reasons first, then emits commentary, then acts; you may prefer reasoning → text → tool call. Text-first is defensible (puts user-relevant info up top), just flagging it for a conscious decision. Separately, I didn't see a test covering the content + reasoning + tool_calls combination — worth adding to lock in whichever order you pick.

frontend/src/components/workspace/messages/message-group.tsx
```diff
@@ -258,30 +311,12 @@ export function MessageGroup({
...
-
-              return [
-                renderDebugSummary(step.messageId, stepIndex),
-                renderToolCall(step),
-              ];
-            })}
+          {(showAbove
+            ? aboveLastToolCallSteps
+            : aboveLastToolCallSteps.filter(
+                (step) => step.type === "assistantText",
```
willem-bd, line 317: Non-blocking UX consideration: this filter keeps assistant text always visible, which is exactly the PR's goal. Side effect to be aware of — if a verbose model emits a lot of intermediate prose alongside tool calls, the user has no way to collapse it (only tool/reasoning steps collapse). Fine for now; worth revisiting if these get noisy in practice.

frontend/tests/unit/components/workspace/messages/message-group.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      } as Message,
+    ]);
+
+    expect(html).toContain(
+      "The first tool failed; I will try a narrower search.",
+    );
+    expect(html).toContain(
+      "The second approach should reveal the missing context.",
+    );
+    expect(html).not.toContain("first hidden query");
```
willem-bd, line 90: Minor: this case renders the "more steps" toggle (collapsible count = 1) and exercises `t.toolCalls.moreSteps(...)`, but doesn't assert it. Since the count now intentionally excludes `assistantText` steps, asserting the rendered toggle count is `1` (not `2`) would lock in that the label reflects only collapsible steps and guard against a future regression that re-includes text in the count.

willem-bd (commented):
Reviewed the diff (frontend rendering path follow-up to #4027/#4033). Overall this is a clean, correctly-scoped, well-tested change that fixes a real gap: assistant text on tool-calling messages was being dropped from the UI.

No correctness bugs found — I verified that grouping avoids double-rendering vs. the final assistant bubble (`becomesAssistantBubble = hasContent && !hasToolCalls`), the new step keys are unique (`${message.id}-content` vs. `message.id` / `tool_call.id`), and the token-debug summary dedup (`firstEligibleDebugSummaryStepIndexByMessageId`) still renders once per message.

The three inline notes below are non-blocking polish suggestions. Happy to see this merged as-is.

WillemJiang (approved):

PR #4115, opened 2026-07-12
feat(agent): Add subagent total delegation cap

hetaoBackend:
## Summary
- Add a deterministic per-run subagent delegation cap on top of the existing per-response concurrency cap.
- Wire the cap through config.yaml (`subagents.max_total_per_run`) and runtime context (`max_total_subagents`), and teach the lead prompt about both limits.
- Keep raw provider tool-call metadata in sync when task calls are dropped, and force a visible stop message when the total cap is exhausted.
- Update docs, config_version, and regression coverage for middleware behavior, config, prompt wiring, Gateway context forwarding, and embedded client cache invalidation.

## Root Cause
Issue #3857 identified that the lead agent could obey the per-response concurrency limit while repeatedly launching new legal-sized `task` batches at planning checkpoints. The existing guardrail capped concurrent subagents, but not total delegations across a run.

## Validation
- `python -m ruff check app/gateway/services.py packages/harness/deerflow/agents/lead_agent/agent.py packages/harness/deerflow/agents/lead_agent/prompt.py packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py packages/harness/deerflow/client.py packages/harness/deerflow/config/subagents_config.py tests/test_gateway_services.py tests/test_lead_agent_model_resolution.py tests/test_lead_agent_prompt.py tests/test_subagent_limit_middleware.py tests/test_subagent_timeout_config.py tests/test_client.py tests/test_memory_tools.py tests/test_subagent_prompt_security.py`
- `PYTHONPATH=... python -m pytest tests/test_subagent_limit_middleware.py tests/test_subagent_timeout_config.py tests/test_lead_agent_model_resolution.py tests/test_gateway_services.py tests/test_lead_agent_prompt.py tests/test_client.py tests/test_memory_tools.py tests/test_subagent_prompt_security.py tests/test_config_version.py -q` (372 passed)
- `PYTHONPATH=... python -m pytest tests/ -q --ignore=tests/test_stream_bridge.py --ignore=tests/test_gateway_imports.py` (7030 passed, 21 skipped)
- Manual import checks with the worktree PYTHONPATH for `from app.gateway.app import app` and `from deerflow.subagents import SubagentExecutor, SubagentResult`

## Notes
- Full `tests/` collection in this local reused venv is blocked by missing redis extra in `tests/test_stream_bridge.py`.
- `tests/test_gateway_imports.py` was excluded from the big local run because its subprocess sets only `PYTHONPATH=backend`; when reusing the main checkout venv for this worktree, that subprocess can resolve the installed harness from the main checkout instead of the worktree. The equivalent import commands pass when run with the worktree harness path.


willem-bd (commented):
Thanks @hetaoBackend. Leaving a few minor inline notes from review - none are blocking.

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
```diff
@@ -53,24 +129,37 @@ def _truncate_task_calls(self, state: AgentState) -> dict | None:
...
+            "Truncated %s excess task tool call(s) from model response (concurrent limit: %s; total limit: %s; prior delegations: %s)",
+            dropped_count,
+            self.max_concurrent,
+            self.max_total,
+            prior_delegation_count,
+        )
 
         # Replace the AIMessage with truncated tool_calls (same id triggers replacement)
-        updated_msg = clone_ai_message_with_tool_calls(last_msg, truncated_tool_calls)
+        content = _append_text(last_msg.content, _TOTAL_LIMIT_STOP_MSG) if total_limited else None
```
willem-bd: Minor: `_TOTAL_LIMIT_STOP_MSG` is appended whenever `total_limited` is true, but when `0 < allowed_task_calls < len(task_indices)` some task calls still execute (`tool_calls` is non-empty), so the graph runs them and this content text won't surface as a terminal message. The wording "the subagent delegation limit for this run has been reached" is also inaccurate in that case - the limit is being approached, not exhausted.

Consider gating on `remaining_total == 0` (truly terminal: no task calls run, so the message is the final response), e.g. `content = _append_text(...) if remaining_total == 0 else None`. The hard cap is enforced deterministically either way; this just avoids a misleading / ineffective message.

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
```diff
@@ -53,24 +129,37 @@ def _truncate_task_calls(self, state: AgentState) -> dict | None:
...
+        updated_msg = clone_ai_message_with_tool_calls(last_msg, truncated_tool_calls, content=content)
         return {"messages": [updated_msg]}
 
     @override
     def after_model(self, state: AgentState, runtime: Runtime) -> dict | None:
-        return self._truncate_task_calls(state)
+        return self._truncate_task_calls(state, runtime)
```
willem-bd, line 164: `prior_delegation_count` relies on `SubagentLimitMiddleware.after_model` running *before* `DurableContextMiddleware.after_model` (langchain runs `after_model` in reverse addition order), so the current batch isn't yet in the ledger and isn't double-counted as prior. That ordering is correct and documented in AGENTS.md, but no test asserts it - every limit test pre-populates `state["delegations"]` directly.

If the middleware order ever changes, the count silently breaks (e.g. triggers one batch early). Worth an integration test that runs both middlewares in the real chain and asserts the current response's task calls are not counted as prior delegations.

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
```diff
@@ -1,44 +1,120 @@
...
+    run_id = context.get("run_id")
+    return str(run_id) if run_id else None
+
+
+def _count_prior_delegations(delegations: object, *, run_id: str | None) -> int:
+    if not isinstance(delegations, list):
+        return 0
+    ids = set()
+    for entry in delegations:
+        if run_id is not None and _delegation_run_id(entry) != run_id:
```
willem-bd, line 87: When `run_id` is `None` the `if run_id is not None` guard is skipped and `_count_prior_delegations` counts *all* delegations on the thread, including previous runs. Safe (fail-restrictive), but it could unexpectedly block the first task call on a thread with >=6 historical delegations, and it's hard to diagnose.

Both production paths (Gateway worker and `DeerFlowClient.stream()`) always set `run_id`, so this only affects custom integrations. Consider a debug/warning log in `_runtime_run_id` (or here) when `run_id` is absent so the over-counting is visible.

config.example.yaml
```diff
@@ -15,7 +15,7 @@
 # ============================================================================
 # Bump this number when the config schema changes.
 # Run `make config-upgrade` to merge new fields into your local config.yaml.
-config_version: 22
+config_version: 24
```
willem-bd, line 18: `config_version` jumps 22 -> 24, skipping 23. Assuming 23 is reserved for sibling PR #4113 (same author, overlapping `durable_context_middleware.py` / `thread_state.py`): merge order matters here. If this PR lands first, #4113's version bump will need adjusting to avoid a collision.

Also worth confirming `make config-upgrade` has a migration step for v24 - the new `max_total_per_run` field has a safe default, so existing configs should upgrade cleanly, but the skip is unusual enough to call out.

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
hetaoBackend: Done in `af85b442`. The reached notice is now appended only when `remaining_total == 0`; partial truncation still enforces the cap but leaves the response content unchanged. Added a regression assertion for the partial-cap case.

Verified with the focused middleware tests and the backend suite (`7207 passed, 21 skipped`).

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
hetaoBackend: Done in `af85b442`. Added a real `create_agent` integration test with `DurableContextMiddleware` followed by `SubagentLimitMiddleware`. With two prior delegations and two new task calls under a total cap of three, it asserts that exactly one new task executes and only that task is added to the durable ledger, locking in reverse `after_model` dispatch behavior.

Verified with the focused middleware tests and the backend suite (`7207 passed, 21 skipped`).

config.example.yaml
hetaoBackend: Confirmed. `main` is currently config v22 and open sibling PR #4113 owns the reserved v23 bump, so this PR intentionally uses v24. If merge order changes, the later PR must advance its version to avoid a collision.

I also exercised `scripts/config-upgrade.sh` against a temporary v22 config: it upgraded cleanly to v24. No version-specific migration is required for `max_total_per_run` because the field is optional and `SubagentsAppConfig` supplies the safe default; the generic recursive merge/version bump path is sufficient.

backend/packages/harness/deerflow/agents/middlewares/subagent_limit_middleware.py
hetaoBackend: Done in `af85b442`. Missing `run_id` now emits a warning explaining that all thread delegations are counted and that custom integrations should provide `run_id`. The warning is emitted only when a response actually contains `task` calls, avoiding noise on ordinary tool calls. Added a focused log/behavior regression test and documented the fallback in `backend/AGENTS.md`.

Verified with the focused middleware tests and the backend suite (`7207 passed, 21 skipped`).

WillemJiang (approved):

PR #4117, opened 2026-07-12
fix(frontend): harden artifact and markdown rendering

nankingjing:
## Summary
- sandbox browser-rendered artifact iframes: the HTML preview keeps `sandbox="allow-scripts allow-forms"` (opaque origin; scripts are required by the injected scroll-restoration `postMessage` bridge) and the `urlOfArtifact` iframe (PDF/images/audio/video) gets an empty `sandbox`
- let non-code browser-previewable `write_file` artifacts (e.g. PDF) render in the sandboxed iframe instead of being forced into the code editor
- reject unsafe markdown link schemes (`javascript:`, `data:`, `file:`, …) before rendering anchors — in chat messages (`MarkdownLink`, including the citation branch) and in `.md` artifact previews (`ArtifactLink`); `mailto:`/`tel:` stay allowed
- make local-settings and agent save-hint storage best-effort when browser storage is unavailable (Safari private mode, strict containers)
- add render-level link-component tests, URL/storage regressions, and artifact iframe sandbox e2e assertions

## Scope notes
The thread-history error-surfacing changes were dropped while rebasing: upstream #4065 rewrote `useThreadHistory` as a TanStack `useInfiniteQuery` whose queryFn already throws on `!response.ok` and toasts the failure, so this PR no longer touches `core/threads/hooks.ts`.

## Validation
- `pnpm check` (eslint + tsc) — clean
- `pnpm test` — 653 unit tests pass
- `pnpm format` — clean
- full Playwright e2e suite locally — 95 passed, including the previously failing PDF `urlOfArtifact` sandbox test

### DIFF
--- a/frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
+++ b/frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
@@ -360,6 +360,7 @@ export function ArtifactFileDetail({
         {!isCodeFile && canPreviewInBrowser && (
           <iframe
             className="size-full"
+            sandbox=""
             src={urlOfArtifact({ filepath, threadId, isMock })}
           />
         )}
@@ -528,7 +529,8 @@ export function ArtifactFilePreview({
         ref={iframeRef}
         className="size-full"
         title="Artifact preview"
-        sandbox="allow-scripts allow-forms"
+        // Artifact HTML is untrusted. Keep all iframe capabilities disabled.
+        sandbox=""
         src={htmlPreviewUrl}
       />
     );
--- a/frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
+++ b/frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
@@ -529,8 +529,12 @@ export function ArtifactFilePreview({
         ref={iframeRef}
         className="size-full"
         title="Artifact preview"
-        // Artifact HTML is untrusted. Keep all iframe capabilities disabled.
-        sandbox=""
+        // allow-scripts is needed for the scroll-restoration injected
+        // script (appendHtmlPreviewScrollRestoration) which communicates
+        // via postMessage. allow-same-origin is deliberately omitted: the
+        // opaque origin prevents access to parent.document and cookies,
+        // and postMessage(..., "*") works fine from it.
+        sandbox="allow-scripts allow-forms"
         src={htmlPreviewUrl}
       />
     );
--- a/frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
+++ b/frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
@@ -129,7 +129,13 @@ export function ArtifactFileDetail({
   }, [filepath]);
   const { isCodeFile, language } = useMemo(() => {
     if (isWriteFile) {
-      let language = checkCodeFile(filepath).language;
+      const codeResult = checkCodeFile(filepath);
+      // Non-code browser-previewable files (PDF, images, audio, video)
+      // should render in the sandboxed iframe, not the code editor.
+      if (!codeResult.isCodeFile && canBrowserPreviewFile(filepath)) {
+        return codeResult;
+      }
+      let language = codeResult.language;
       language ??= "text";
       return { isCodeFile: true, language };
     }
--- a/frontend/src/components/workspace/messages/markdown-link.tsx
+++ b/frontend/src/components/workspace/messages/markdown-link.tsx
@@ -16,7 +16,7 @@ import { CitationLink } from "../citations/citation-link";
  * branch below and ``resolveArtifactURL``) — relative ``/…`` URLs are
  * inherently safe and pass through ``URL.protocol === ""``.
  */
-const SAFE_HREF_PROTOCOLS = ["http:", "https:"] as const;
+const SAFE_HREF_PROTOCOLS = ["http:", "https:", "mailto:", "tel:"] as const;
 
 export function isSafeHref(href: string | undefined): boolean {
   if (typeof href !== "string" || href.length === 0) {
@@ -57,22 +57,14 @@ export function createMarkdownLinkComponent(threadId?: string) {
     href,
     ...props
   }: AnchorHTMLAttributes<HTMLAnchorElement>) {
-    if (typeof props.children === "string") {
-      const match = /^citation:(.+)$/.exec(props.children);
-      if (match) {
-        const [, text] = match;
-        return (
-          <CitationLink {...props} href={href}>
-            {text}
-          </CitationLink>
-        );
-      }
-    }
     // Reject unsafe schemes up front so a prompt-injected / pasted href can
-    // never reach the rendered anchor. Keep the visible label so the user
-    // can still see what the link claimed to point at.
+    // never reach the rendered anchor — including through the citation
+    // branch (which renders <a href={href}> directly). Check before the
+    // citation block so prompt-injected [citation:x](javascript:...) is
+    // blocked. Keep the visible label so the user can still see what the
+    // link claimed to point at.
     if (href !== undefined && !isSafeHref(href)) {
-      const { className, children, ...rest } = props;
+      const { className, children, target, rel, ...rest } = props;
       return (
         <span
           {...rest}
@@ -86,6 +78,18 @@ export function createMarkdownLinkComponent(threadId?: string) {
           {children}
         </span>
       );
+    }
+    // Safe-href check passed — citation links now route through CitationLink.
+    if (typeof props.children === "string") {
+      const match = /^citation:(.+)$/.exec(props.children);
+      if (match) {
+        const [, text] = match;
+        return (
+          <CitationLink {...props} href={href}>
+            {text}
+          </CitationLink>
+        );
+      }
     }
     if (threadId && href?.startsWith("/mnt/")) {
       return (
--- a/frontend/src/components/workspace/messages/markdown-link.tsx
+++ b/frontend/src/components/workspace/messages/markdown-link.tsx
@@ -13,8 +13,9 @@ import { CitationLink } from "../citations/citation-link";
  * browser happily executes the payload in the chat surface where
  * sessionStorage / CSRF cookies are reachable. We accept ``http(s)``
  * plus the non-executing ``mailto:`` / ``tel:`` schemes; same-origin
- * relative paths (``/…``) and in-document anchors (``#…``) are allowed
- * by the explicit prefix checks in ``isSafeHref`` before URL parsing.
+ * absolute paths (``/…``), scheme-less relative references (``report.md``,
+ * ``./…``, ``../…``), and in-document anchors (``#…``) are also allowed
+ * — they all resolve under the current HTTP(S) origin.
  */
 const SAFE_HREF_PROTOCOLS = ["http:", "https:", "mailto:", "tel:"] as const;
 
@@ -22,16 +23,24 @@ export function isSafeHref(href: string | undefined): boolean {
   if (typeof href !== "string" || href.length === 0) {
     return false;
   }
-  // Allow relative same-origin paths (no protocol — e.g. "/workspace/foo").
-  if (href.startsWith("/") && !href.startsWith("//")) {
-    return true;
-  }
-  // Allow same-document anchors (e.g. "#section").
+  // Same-document anchors (e.g. "#section").
   if (href.startsWith("#")) {
     return true;
   }
+  // Protocol-relative URLs (//evil.com/path) would inherit the current
+  // page's scheme and navigate to an unknown host. Also reject
+  // backslash-normalised variants (\\evil.com → //evil.com) because
+  // browsers treat backslash as a path separator in URLs.
+  if (/^(\/\/|\\\\)/.test(href)) {
+    return false;
+  }
+  // Parse the href so we can inspect its scheme. Items without an explicit
+  // protocol (report.md, ./report.md, ../assets/chart.png, /workspace/foo)
+  // resolve relative to the dummy HTTPS base, producing an https: URL.
+  // Explicitly-schemed URLs (https://…, mailto:, javascript:, data:, …)
+  // keep their native scheme, which we then check against the allowlist.
   try {
-    const parsed = new URL(href);
+    const parsed = new URL(href, "https://dummy.example/");
     return (SAFE_HREF_PROTOCOLS as ReadonlyArray<string>).includes(
       parsed.protocol,
     );
--- a/frontend/src/components/workspace/messages/markdown-link.tsx
+++ b/frontend/src/components/workspace/messages/markdown-link.tsx
@@ -5,8 +5,46 @@ import { cn } from "@/lib/utils";
 
 import { CitationLink } from "../citations/citation-link";
 
+/**
+ * Schemes we are willing to render as a navigable ``<a href=...>``.
+ *
+ * Anything else (``javascript:``, ``data:text/html``, ``vbscript:``,
+ * ``file:``, …) is blocked because once it lands in a real anchor the
+ * browser happily executes the payload in the chat surface where
+ * sessionStorage / CSRF cookies are reachable. We accept only ``http`` /
+ * ``https`` plus same-origin paths (which are governed by the artifact
+ * branch below and ``resolveArtifactURL``) — relative ``/…`` URLs are
+ * inherently safe and pass through ``URL.protocol === ""``.
+ */
+const SAFE_HREF_PROTOCOLS = ["http:", "https:"] as const;
+
+export function isSafeHref(href: string | undefined): boolean {
+  if (typeof href !== "string" || href.length === 0) {
+    return false;
+  }
+  // Allow relative same-origin paths (no protocol — e.g. "/workspace/foo").
+  if (href.startsWith("/") && !href.startsWith("//")) {
+    return true;
+  }
+  // Allow same-document anchors (e.g. "#section").
+  if (href.startsWith("#")) {
+    return true;
+  }
+  try {
+    const parsed = new URL(href);
+    return (SAFE_HREF_PROTOCOLS as ReadonlyArray<string>).includes(
+      parsed.protocol,
+    );
+  } catch {
+    return false;
+  }
+}
+
 function isExternalUrl(href: string | undefined): boolean {
-  return !!href && /^https?:\/\//.test(href);
+  if (typeof href !== "string") {
+    return false;
+  }
+  return /^https?:\/\//.test(href);
 }
 
 /**
@@ -29,6 +67,25 @@ export function createMarkdownLinkComponent(threadId?: string) {
           </CitationLink>
         );
       }
+    }
+    // Reject unsafe schemes up front so a prompt-injected / pasted href can
+    // never reach the rendered anchor. Keep the visible label so the user
+    // can still see what the link claimed to point at.
+    if (href !== undefined && !isSafeHref(href)) {
+      const { className, children, ...rest } = props;
+      return (
+        <span
+          {...rest}
+          className={cn(
+            "text-muted-foreground cursor-not-allowed underline decoration-dotted underline-offset-2",
+            className,
+          )}
+          aria-label="Unsafe link omitted"
+          title={`Unsafe link scheme in ${href}`}
+        >
+          {children}
+        </span>
+      );
     }
     if (threadId && href?.startsWith("/mnt/")) {
       return (
--- a/frontend/src/components/workspace/messages/markdown-link.tsx
+++ b/frontend/src/components/workspace/messages/markdown-link.tsx
@@ -11,10 +11,10 @@ import { CitationLink } from "../citations/citation-link";
  * Anything else (``javascript:``, ``data:text/html``, ``vbscript:``,
  * ``file:``, …) is blocked because once it lands in a real anchor the
  * browser happily executes the payload in the chat surface where
- * sessionStorage / CSRF cookies are reachable. We accept only ``http`` /
- * ``https`` plus same-origin paths (which are governed by the artifact
- * branch below and ``resolveArtifactURL``) — relative ``/…`` URLs are
- * inherently safe and pass through ``URL.protocol === ""``.
+ * sessionStorage / CSRF cookies are reachable. We accept ``http(s)``
+ * plus the non-executing ``mailto:`` / ``tel:`` schemes; same-origin
+ * relative paths (``/…``) and in-document anchors (``#…``) are allowed
+ * by the explicit prefix checks in ``isSafeHref`` before URL parsing.
  */
 const SAFE_HREF_PROTOCOLS = ["http:", "https:", "mailto:", "tel:"] as const;
 
@@ -64,10 +64,12 @@ export function createMarkdownLinkComponent(threadId?: string) {
     // blocked. Keep the visible label so the user can still see what the
     // link claimed to point at.
     if (href !== undefined && !isSafeHref(href)) {
-      const { className, children, target, rel, ...rest } = props;
+      // Intentionally no {...props} spread: react-markdown props like `node`
+      // (and anchor-only attributes such as target/rel) are not valid on a
+      // <span> and would trigger React DOM warnings.
+      const { className, children } = props;
       return (
         <span
-          {...rest}
           className={cn(
             "text-muted-foreground cursor-not-allowed underline decoration-dotted underline-offset-2",
             className,
--- a/frontend/src/core/settings/local.ts
+++ b/frontend/src/core/settings/local.ts
@@ -33,7 +33,7 @@ function isBrowser(): boolean {
  * settings panel. This wrapper traps every storage exception so callers can
  * always fall back to a sane default.
  */
-const safeLocalStorage = {
+export const safeLocalStorage = {
   getItem(key: string): string | null {
     if (!isBrowser()) return null;
     try {
--- a/frontend/src/core/settings/local.ts
+++ b/frontend/src/core/settings/local.ts
@@ -23,6 +23,45 @@ function isBrowser(): boolean {
   return typeof window !== "undefined";
 }
 
+/**
+ * Best-effort localStorage facade.
+ *
+ * Safari private mode, Firefox strict containers, some embedded WebViews, and
+ * quotas already filled by sibling tabs throw ``SecurityError`` or
+ * ``QuotaExceededError`` from ``getItem``/``setItem``. Without a guard those
+ * exceptions bubble into React render handlers and break the composer /
+ * settings panel. This wrapper traps every storage exception so callers can
+ * always fall back to a sane default.
+ */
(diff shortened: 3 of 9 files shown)


nankingjing:
Follow-up commits pushed on top of `fix/task-59-frontend-batch`. Both review gaps addressed; everything stays inside `frontend/`.

## 1. `useThreadHistory` fetch guard — `e8a29da8`

`test(frontend): pin useThreadHistory fetch guard against silent 5xx/404 parsing`

- NEW `frontend/tests/unit/core/threads/hooks.test.ts:1-78` — rstest test using `rs.stubGlobal("fetch", …)` to drive the wrapper import (`@/core/api/fetcher`) that the hook calls. Mirrors the same `.then((res) => { if (!res.ok) throw new Error(\`run messages fetch failed: ${res.status} ${res.statusText}\`) })` shape from `frontend/src/core/threads/hooks.ts:1669-1677`.
- `frontend/tests/unit/core/threads/hooks.test.ts:20-50` — 500 path: the `json()` spy throws if called, so the only way to flip `rejects.toThrow` to a resolve is to drop the guard.
- `frontend/tests/unit/core/threads/hooks.test.ts:52-78` — 404 path: same shape, asserts `run messages fetch failed: 404 Not Found` so the 404 consumer-visible failure mode is also pinned.

## 2. Artifact iframe sandbox — `dca9190f`

`fix(frontend): keep artifact preview iframe same-origin and drop the no-op sandbox`

- `frontend/src/components/workspace/artifacts/artifact-file-detail.tsx:360-365` — dropped the no-op `sandbox=""` attribute on the `ArtifactFileDetail` browser-preview iframe; the default already denies everything.
- `frontend/src/components/workspace/artifacts/artifact-file-detail.tsx:525-539` — `ArtifactFilePreview` html iframe: `sandbox=""` → `sandbox="allow-same-origin"`, with the comment updated to explain that blob URLs are origin-tied to the embedding page and that cross-origin scripts / popups / top navigation / forms are still denied by default.
- `frontend/tests/e2e/artifact-preview.spec.ts:114` — e2e sandbox-attribute assertion updated from `""` to `"allow-same-origin"`.

## Verification

- `pnpm exec rstest run` — 70 files / 618 tests pass (incl. the new `tests/unit/core/threads/hooks.test.ts`).
- `pnpm exec tsc --noEmit` — clean.
- `pnpm exec eslint . --ext .ts,.tsx` — clean.
- No backend / Python files touched.

frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
```diff
@@ -528,7 +528,12 @@ export function ArtifactFilePreview({
         ref={iframeRef}
         className="size-full"
         title="Artifact preview"
-        sandbox="allow-scripts allow-forms"
+        // Artifact HTML is untrusted; stay same-origin so the iframe is not
+        // promoted to an opaque origin (which would make blob URLs a unique
+        // origin and break inline scripts / forms in the rendered preview)
+        // while still denying cross-origin scripts, popups, top navigation,
+        // and form submission by default.
+        sandbox="allow-same-origin"
```
willem-bd: ⚠️ This sandbox change breaks HTML artifact scroll restoration, and the comment is backwards.

**Regression:** `appendHtmlPreviewScrollRestoration` (still called at line 496) injects a `<script>` into the blob HTML (`preview.ts:216-253`) that does `window.parent.postMessage(...)` to save/restore scroll position, and the parent listens for those messages (lines 460-488). With `allow-scripts` removed, that injected script never runs, so the iframe never emits `save`/`restore-request` events and scroll position is no longer preserved across re-renders. No test covers this - the e2e only asserts the `sandbox` attribute value, and the only scroll test is for *markdown* artifacts.

**No security gain:** the prior config (`allow-scripts allow-forms`, no `allow-same-origin`) was already the safe way to run sandboxed scripts - opaque origin can't reach `parent.document`/cookies, and `postMessage(..., "*")` works fine from it. So this trades a working feature for no real security benefit unless scroll restoration is reimplemented parent-side (which `allow-same-origin` would actually permit via `contentDocument` access).

**Misleading comment:** the rationale says staying same-origin prevents an opaque origin that would "break inline scripts / forms," but the new sandbox *denies* both scripts and forms. It should say they're intentionally disabled (matching the PR description) and explain why `allow-same-origin` is still needed.

frontend/src/components/workspace/messages/markdown-link.tsx
```diff
@@ -30,6 +68,25 @@ export function createMarkdownLinkComponent(threadId?: string) {
         );
       }
     }
+    // Reject unsafe schemes up front so a prompt-injected / pasted href can
+    // never reach the rendered anchor. Keep the visible label so the user
+    // can still see what the link claimed to point at.
+    if (href !== undefined && !isSafeHref(href)) {
```
willem-bd, line 75: 🟠 The `isSafeHref` guard runs *after* the citation branch (line 61), and `CitationLink` renders `<a href={href}>` directly (`citation-link.tsx:34`). So a prompt-injected `[citation:x](javascript:alert(1))` matches `^citation:(.+)$` on the label, skips this guard, and renders a clickable `javascript:` anchor - exactly the threat this hardening targets.

Consider moving the `isSafeHref` check above the citation branch, or validating `href` inside `CitationLink`.

frontend/src/components/workspace/messages/markdown-link.tsx
```diff
@@ -5,8 +5,46 @@ import { cn } from "@/lib/utils";
...
+ *
+ * Anything else (``javascript:``, ``data:text/html``, ``vbscript:``,
+ * ``file:``, …) is blocked because once it lands in a real anchor the
+ * browser happily executes the payload in the chat surface where
+ * sessionStorage / CSRF cookies are reachable. We accept only ``http`` /
+ * ``https`` plus same-origin paths (which are governed by the artifact
+ * branch below and ``resolveArtifactURL``) — relative ``/…`` URLs are
+ * inherently safe and pass through ``URL.protocol === ""``.
+ */
+const SAFE_HREF_PROTOCOLS = ["http:", "https:"] as const;
```
willem-bd: 🟡 Heads up: this allowlist also rejects `mailto:` and `tel:` schemes, so legitimate `[contact](mailto:foo@bar.com)` links now render as disabled spans. If that's intentional, fine - otherwise consider adding them. The unit test doesn't cover `mailto:`/`tel:`, so the behavior change is invisible.

frontend/src/components/workspace/messages/markdown-link.tsx
```diff
@@ -30,6 +68,25 @@ export function createMarkdownLinkComponent(threadId?: string) {
         );
       }
     }
+    // Reject unsafe schemes up front so a prompt-injected / pasted href can
+    // never reach the rendered anchor. Keep the visible label so the user
+    // can still see what the link claimed to point at.
+    if (href !== undefined && !isSafeHref(href)) {
+      const { className, children, ...rest } = props;
```
willem-bd: 🟡 Minor: this destructures only `{ className, children, ...rest }`, leaving `target`/`rel` (valid `AnchorHTMLAttributes`) in `rest` spread onto a `<span>`, which triggers React DOM warnings. The external branch below extracts `target`/`rel` explicitly; this branch should too. (`node` from react-markdown is a pre-existing spread issue, not new.)

frontend/src/core/threads/hooks.ts
```diff
@@ -1666,6 +1666,15 @@ export function useThreadHistory(
           },
           credentials: "include",
         }).then((res) => {
+          // 5xx/404 responses would otherwise be parsed as JSON and the real
+          // status swallowed. Surface the failure to TanStack Query so
```
willem-bd: 🟠 The guard throws on `!res.ok`, but the surrounding `catch (err) { console.error(err); }` (line 1720) swallows it, so `loadMessages()` resolves normally and the effect's `.catch(() => toast.error(...))` never fires. This is also a manual `useCallback`, not a TanStack Query, so `isError` does **not** flip - contrary to the comment added here ("Surface the failure to TanStack Query so `isError` flips").

The real effect is only a cleaner `console.error` and avoiding `res.json()` on HTML error bodies. To actually surface failures to the user, rethrow past the inner catch (or remove it) so the toast fires.

frontend/tests/unit/core/threads/hooks.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    stub.mockResolvedValue({
+      ok: false,
+      status: 500,
+      statusText: "Internal Server Error",
+      json: () => {
+        throw new Error("res.json() must not be called when res.ok is false");
+      },
+    });
+
+    // Mirror the .then guard from `useThreadHistory` so dropping the
```
willem-bd: 🟡 This test reconstructs the `.then` guard inline and asserts it rejects - it never renders `useThreadHistory`. So removing the guard from `hooks.ts` would **not** fail this test (the test has its own copy of the guard). The comment here claiming it pins the contract ("dropping the `!res.ok` throw causes this expectation to flip ... and the test to fail loudly") is inaccurate; it only verifies the `fetch` wrapper passes `ok`/`status`/`statusText` through.

Consider an integration test that renders the hook with a stubbed `fetch` returning 500 and asserts the error surfaces.

frontend/src/core/settings/local.ts
```diff
@@ -23,6 +23,45 @@ function isBrowser(): boolean {
...
+ * Best-effort localStorage facade.
+ *
+ * Safari private mode, Firefox strict containers, some embedded WebViews, and
+ * quotas already filled by sibling tabs throw ``SecurityError`` or
+ * ``QuotaExceededError`` from ``getItem``/``setItem``. Without a guard those
+ * exceptions bubble into React render handlers and break the composer /
+ * settings panel. This wrapper traps every storage exception so callers can
+ * always fall back to a sane default.
+ */
+const safeLocalStorage = {
```
willem-bd: 🟡 Only `core/settings/local.ts` is wrapped. `core/agents/feature-cache.ts:21,33` and `app/workspace/agents/new/page.tsx:133,137` still call `window.localStorage` directly and can throw into render handlers. Fine if intentionally scoped to settings, but the "best-effort storage" robustness is inconsistent across the app.

willem-bd (commented):
Reviewed the security-hardening changes. The direction is sound (allowlist for hrefs, no untrusted script execution in artifact iframes, wrapping localStorage, surfacing fetch errors), but I'm flagging one functional regression and a few correctness/comment issues:

- 🔴 `sandbox="allow-same-origin"` (no `allow-scripts`) breaks the app's own HTML artifact scroll-restoration script - the injected `<script>` can no longer run.
- 🟠 Citation links bypass the new `isSafeHref` guard.
- 🟠 The thread-history `!res.ok` throw is swallowed by the existing inner catch, so it isn't actually surfaced to the UI / TanStack Query as the comment claims.
- 🟡 The new `hooks.test.ts` is a strawman (doesn't render the hook).
- 🟡 `mailto:`/`tel:` links are now rejected; `safeLocalStorage` only covers `local.ts`.

Details inline. Given the `needs-validation` label, the scroll-restoration regression in particular is worth verifying on a real HTML artifact.

willem-bd (commented):
Thanks @nankingjing. One net-new test-coverage gap beyond the prior review.

[P2] markdown-link unit test only exercises `isSafeHref` in isolation, not the rendered component

- Location: `frontend/tests/unit/components/workspace/messages/markdown-link.test.ts:1-21`
- Problem: The new test imports and calls `isSafeHref` directly but never renders the `MarkdownLink` component. If the `isSafeHref` guard at `markdown-link.tsx:74` is removed or its condition is inverted (`!isSafeHref` → `isSafeHref`), every existing assertion still passes because the function itself is unchanged - yet the component would render a clickable `javascript:` anchor, silently reintroducing the XSS this PR is meant to prevent.
- Evidence: The test file imports only `{ isSafeHref }` and asserts on its boolean return. No test renders `createMarkdownLinkComponent()` output with a `javascript:` href and asserts the DOM contains a `<span>` rather than an `<a>`.
- Suggested fix: Add a render-level case, e.g. `render(<MarkdownLink href="javascript:alert(1)">x</MarkdownLink>)` and assert `container.querySelector("a")` is null and `container.querySelector("span")` is not null. Mirror it with a positive case (`https://example.com`) asserting an `<a>` is rendered.
- Test: The render-level test itself - it fails if the guard is removed or the condition flipped.

nankingjing:
Restored `allow-scripts` on the artifact iframe — turns out it was needed for scroll-restore. Moved the `isSafeHref` guard above the citation block so malicious citation links cannot bypass the check. Added `throw err` in the hooks error handler so TanStack Query sees the failure.

frontend/src/components/workspace/messages/markdown-link.tsx
```diff
@@ -19,6 +57,29 @@ export function createMarkdownLinkComponent(threadId?: string) {
     href,
     ...props
   }: AnchorHTMLAttributes<HTMLAnchorElement>) {
+    // Reject unsafe schemes up front so a prompt-injected / pasted href can
+    // never reach the rendered anchor — including through the citation
+    // branch (which renders <a href={href}> directly). Check before the
+    // citation block so prompt-injected [citation:x](javascript:...) is
+    // blocked. Keep the visible label so the user can still see what the
+    // link claimed to point at.
+    if (href !== undefined && !isSafeHref(href)) {
```
willem-bd, line 75: This guard is good, but it only covers `createMarkdownLinkComponent` (used by message content + message-list-item). Markdown **artifacts** render through a separate, unguarded component: `artifact-file-detail.tsx` renders `.md` previews via `<SafeStreamdown components={{ a: ArtifactLink }}>` (artifact-file-detail.tsx:518), and `ArtifactLink` (`citations/artifact-link.tsx`) renders `<a {...rest} />` with no scheme check.

A prompt-injected `[x](javascript:...)` in a `.md` artifact would then execute in the main document, where sessionStorage / CSRF cookies are reachable - exactly what this guard prevents elsewhere. Consider applying the exported `isSafeHref` at the top of `ArtifactLink` (same span fallback), or having `ArtifactLink` delegate to this shared component.

frontend/src/core/threads/hooks.ts
```diff
@@ -1710,6 +1719,7 @@ export function useThreadHistory(
       } while (pendingLoadRef.current);
     } catch (err) {
       console.error(err);
+      throw err;
```
willem-bd: The rethrow does enable the auto-load toast at `loadMessages().catch(() => toast.error(...))` (hooks.ts:1750) - that's the real improvement here. But two issues:

1. The comment on the `!res.ok` block above says this surfaces the failure "to TanStack Query so `isError` flips on the caller side." `loadMessages` isn't a queryFn (it uses local refs + `setMessageRows`), and the hook's return object (hooks.ts:1772) has no `isError`/`error` field - so `isError` never flips anywhere. The comment is misleading.

2. `loadMessages` is also exposed as `loadMore` and called as a bare `loadMore?.()` **without** `.catch()` from `LoadMoreHistoryIndicator` (message-list.tsx:133/147, wired via `loadMore={loadMoreHistory}` at :708). Before this change the catch swallowed the error so it never rejected; now a 5xx on the scroll/manual path produces an unhandled promise rejection (and no toast on that path), and the `loadMore?: () => void` signature understates the new contract.

Suggest catching + toasting inside `loadMessages` and not rethrowing (keeps the toast, removes the unhandled-rejection surface), or adding `.catch()` at both call sites. If callers need to observe failures, expose an explicit `error`/`isError` field rather than throwing through a `() => void` callback.

frontend/src/core/threads/hooks.ts
```diff
@@ -1666,6 +1666,15 @@ export function useThreadHistory(
           },
           credentials: "include",
         }).then((res) => {
+          // 5xx/404 responses would otherwise be parsed as JSON and the real
+          // status swallowed. Surface the failure to TanStack Query so
+          // `isError` flips on the caller side and we stop pretending the
+          // thread has an empty history.
+          if (!res.ok) {
```
willem-bd: This new error-surfacing path (the `!res.ok` throw here + the `throw err` rethrow below) doesn't have test coverage. Worth adding a test asserting a 5xx / 404 surfaces the failure to the caller and doesn't leave `loading` stuck `true`.

frontend/tests/e2e/artifact-preview.spec.ts
```diff
@@ -109,6 +109,9 @@ test.describe("Artifact preview stability", () => {
     await expect(
       artifactsPanel.locator('iframe[title="Artifact preview"]'),
     ).toBeVisible();
+    await expect(
+      artifactsPanel.locator('iframe[title="Artifact preview"]'),
+    ).toHaveAttribute("sandbox", "allow-scripts allow-forms");
```
willem-bd, line 116: Good to assert the sandbox on the HTML preview iframe. The PR also adds `sandbox=""` to the other artifact iframe - the `urlOfArtifact` one at artifact-file-detail.tsx:363 - but there's no assertion covering it. A parallel `toHaveAttribute("sandbox", "")` would catch a future regression (e.g. someone adding `allow-scripts`) on that iframe too.

willem-bd (commented):
Review findings (4 inline comments). Details in each thread.

WillemJiang:
@nankingjing, please take a look at the failed tests.

WillemJiang:
@nankingjing  Please resolve the conflicts.

frontend/src/components/workspace/artifacts/artifact-file-detail.tsx
nankingjing: Confirmed and fixed — `allow-scripts allow-forms` is restored (the opaque-origin config was indeed already the safe one), and the comment now explains that scripts are needed for the injected scroll-restoration postMessage bridge while `allow-same-origin` is deliberately omitted. The e2e asserts the restored sandbox value.

frontend/src/components/workspace/messages/markdown-link.tsx
nankingjing: Fixed — the `isSafeHref` guard now runs before the citation branch, so `[citation:x](javascript:...)` renders the disabled span. Added a render-level regression test ("blocks unsafe hrefs before the citation branch") that renders the component with a citation-labelled `javascript:` href and asserts no anchor is produced.

frontend/src/components/workspace/messages/markdown-link.tsx
nankingjing: Added `mailto:` and `tel:` to the allowlist, updated the stale docstring that still claimed http/https-only, and added unit cases covering both schemes so the behavior is pinned.

frontend/src/components/workspace/messages/markdown-link.tsx
nankingjing: Fixed more thoroughly: the span fallback no longer spreads `...rest` at all, so anchor-only attributes (`href`/`target`/`rel`) and react-markdown's `node` prop never reach the `<span>`. The new render-level test actually caught `href` leaking onto the span in `ArtifactLink` via the spread — both fallbacks now only take `className`/`children`.

frontend/src/core/threads/hooks.ts
nankingjing: You were right that the throw was swallowed and the TanStack comment was wrong. This is now moot: upstream #4065 rewrote `useThreadHistory` as a real TanStack `useInfiniteQuery` whose queryFn throws on `!response.ok` (via `readResponseErrorMessage`) and toasts on `historyQuery.error` — exactly what this change was groping toward. After rebasing, this PR no longer touches `core/threads/hooks.ts`.

frontend/tests/unit/core/threads/hooks.test.ts
nankingjing: Agreed it was a strawman — the test re-implemented the guard instead of exercising it. The file was removed during the rebase: the manual `.then` guard it mirrored no longer exists after upstream #4065 replaced `loadMessages` with a TanStack `useInfiniteQuery`.

frontend/src/core/settings/local.ts
nankingjing: Made it consistent: `safeLocalStorage` is now exported and the agent-create save-hint effect (`app/workspace/agents/new/page.tsx`) uses it — that was the only remaining unguarded call site. `core/agents/feature-cache.ts` already wraps its `window.localStorage` access in try/catch internally, so it was left as is.

frontend/src/components/workspace/messages/markdown-link.tsx
nankingjing: Fixed — `ArtifactLink` now applies the shared `isSafeHref` guard up front and renders the same disabled-span fallback, so `javascript:`/`data:` hrefs in `.md` artifact previews can't reach a real anchor in the main document. Added render-level tests for both the unsafe-span and safe-anchor paths (which also caught the href leaking onto the fallback span through the rest spread — fixed).

frontend/src/core/threads/hooks.ts
nankingjing: Both points were valid — and both are now moot after the rebase: upstream #4065 replaced `loadMessages` with `useInfiniteQuery`, so `loadMore` is `historyQuery.fetchNextPage` (which never throws through the callback — TanStack captures failures in query state) and errors surface through `historyQuery.error` + toast. This PR no longer touches `core/threads/hooks.ts`.

frontend/src/core/threads/hooks.ts
nankingjing: After rebasing onto main, the error path here belongs to upstream #4065's `useInfiniteQuery` queryFn (throws on `!response.ok`, toasts via `historyQuery.error`, and `loading` derives from `isLoading`/`isFetchingNextPage` so it can't stick). Our duplicate guard and its test were dropped rather than shadow-testing upstream's implementation from this PR.

frontend/tests/e2e/artifact-preview.spec.ts
nankingjing: Done — added an e2e that opens a PDF `write_file` artifact and asserts `sandbox=""` on the `urlOfArtifact` iframe. Writing it surfaced that non-code browser-previewable `write_file` artifacts were forced into the code editor (iframe never rendered — the CI failure you saw); fixed so they render in the sandboxed iframe. Full Playwright suite passes locally (95/95).

nankingjing:
@WillemJiang Conflicts resolved and the failed test fixed — branch rebased onto latest `main` and force-pushed.

**Rebase / conflict resolution.** The only overlap was `frontend/src/core/threads/hooks.ts`: #4065 rewrote `useThreadHistory` as a TanStack `useInfiniteQuery` whose queryFn already throws on `!response.ok` and toasts the failure — i.e. upstream now implements what this PR attempted there. I resolved the conflict by keeping upstream's implementation; this PR no longer touches `hooks.ts`, and the old guard-mirroring unit test was removed with it (PR description updated).

**Failed E2E test.** `renders sandboxed iframe for a browser-previewable non-code file` failed because non-code browser-previewable `write_file` artifacts (e.g. PDF) were forced into the code editor, so the `urlOfArtifact` iframe never rendered. Fixed the component so those files render in the sandboxed iframe, and aligned the mocked route with `urlOfArtifact`'s non-mock pattern.

**Remaining review items from @willem-bd** are addressed (inline replies on each thread), including the P2 note about the `isSafeHref`-only unit test: `MarkdownLink` and `ArtifactLink` now have render-level tests asserting an unsafe href produces a disabled `<span>` and never an `<a>` — which incidentally caught (and fixed) the unsafe href leaking onto `ArtifactLink`'s fallback span via the rest spread. The safe localStorage facade is now also used by the agent save-hint effect for consistency.

**Validation on the rebased branch:** `pnpm check` (eslint + tsc) clean, `pnpm test` 653 passed, `pnpm format` clean, and the full local Playwright e2e suite 95/95 — including the previously failing PDF sandbox test.

fancyboi999 (changes requested):

[P2] Preserve safe scheme-less relative Markdown links

- Location: `frontend/src/components/workspace/messages/markdown-link.tsx:23-40`
- Problem: `isSafeHref()` now rejects every relative URL that does not start with `/` or `#`. Valid Markdown links such as `report.md`, `./report.md`, and `../assets/chart.png` therefore render as disabled spans in both chat messages and Markdown artifact previews, regressing links that were navigable before this diff.
- Evidence: `new URL("report.md")` throws when no base URL is supplied, so the catch returns `false`; the new test explicitly locks `relative/path` to `false`. These hrefs are scheme-less and resolve under the current HTTP(S) origin, so rejecting them does not block an executable scheme. `ArtifactLink` imports the same guard, which extends the regression to `.md` artifacts.
- Suggested fix: accept scheme-less relative references explicitly, or parse them against a fixed HTTP(S) base before applying the protocol allowlist. Keep protocol-relative `//...` URLs and explicit non-allowlisted schemes rejected.
- Test: render both `MarkdownLink` and `ArtifactLink` with `report.md`, `./report.md`, and `../report.md` and assert that they produce anchors; retain the current `javascript:`, `data:`, and protocol-relative disabled-span cases.


Reviewed head SHA: `2aa447e041b22c34b1a8f7fb50949f7df74bf5b6`

nankingjing:
@fancyboi999 thanks for catching the scheme-less relative URL regression in the last review.

**Fix applied** in 95aa048f82d19efe57067eb95b0c0fe0afa868d1:

- **isSafeHref rewrite**: parse all hrefs against a dummy https://dummy.example/ base so scheme-less relative references (report.md, ./report.md, ../assets/chart.png, /workspace/foo) produce an https: URL and pass the protocol allowlist. Explicitly-schemed URLs (https://…, mailto:, javascript:, data:, …) keep their native scheme and are checked against the same allowlist as before. Protocol-relative URLs (//evil.com) and backslash-normalised variants (\evil.com → //evil.com) are still rejected before parsing — the browser treats backslash as a path separator, so double-backslash would be normalised to // and change origin.
- **Tests**: added isSafeHref cases for report.md, ./report.md, ../report.md (assert true); added render-level assertions in MarkdownLink and ArtifactLink that a scheme-less relative href actually produces an <a> element; flipped the old relative/path → false expectation to true; added \evil.com → false for the backslash protocol-relative case.
- **Revert-verify**: reverting the URL-base approach (back to new URL(href) without a base) flips the report.md/./report.md/../report.md expectations from true to false, and the render-level expect(html).toContain("<a") assertions fail because the component falls back to <span>.

**Gate results on the fix commit**:

| Gate | Result |
|------|--------|
| pnpm check (eslint + tsc) | clean |
| pnpm test | **656/656 passed** (73 files, +3 relative-URL tests) |
| pnpm format (prettier --check) | clean |

**Additional audit finding (documented, not a fix needed at this time)**: subtask-card.tsx:199 renders task.prompt through <SafeStreamdown components={{ a: CitationLink }}>. CitationLink renders <a href={href}> without its own scheme guard — but this is already protected by streamdown's bundled rehype-harden plugin (v1.1.7), which blocks javascript:/data:/file:/vbscript: hrefs and clears them to "" before they reach the component. The protection is correct at runtime; the PR's caller-side guards (MarkdownLink, ArtifactLink) are defence-in-depth.

WillemJiang (approved):

PR #4118, opened 2026-07-12
fix(checkpoint): persist run duration in checkpoints for history reads

Vanzeren:
## Why                                                                                               
                                                                                                        
   Thread history currently has to correlate checkpoint messages with run records and message events at 
 read time to recover each assistant turn’s duration.                                                   
                                                                                                        
   That fallback has two problems:                                                                      
                                                                                                        
   - It adds run-store and event-store queries to every history read.                                   
   - Event lookup is capped at 1,000 messages, so older turns in long-running threads can silently lose 
 their duration.                                                                                        
                                                                                                        
   This change makes turn duration part of the durable checkpointed conversation state. History reads   
 can return duration and run association directly from the checkpoint while retaining the existing      
 correlation path for legacy threads.                                                                   
                                                                                                        
##  What changed

* Successful runs now persist their `run_id -> turn_duration` mapping in checkpoint metadata.
* Duration persistence appends a metadata-only checkpoint that:

  * creates a fresh checkpoint instead of overwriting the latest row;
  * retains the previous checkpoint as its parent;
  * inherits the parent checkpoint's channel versions;
  * passes an empty `new_versions` mapping, so the `messages` channel is neither version-bumped nor rewritten;
  * preserves unrelated channel values and metadata.
* Duration checkpoint writes re-read the latest checkpoint identity before committing and retry if an intervening graph, title, or goal checkpoint advances the lineage.
* Same-thread graph streaming and duration persistence use the same checkpoint lock, preventing a duration update from being based on a mid-stream checkpoint.
* Concurrent duration updates for the same thread are serialized so one completed run cannot overwrite another run's duration.
* Thread history uses each human message's `additional_kwargs.run_id` as the turn boundary and associates subsequent AI and tool messages with that run.
* History reads inject `turn_duration` from checkpoint metadata into the corresponding serialized AI messages without querying the run and event stores for migrated data.
* Legacy checkpoints without persisted duration metadata continue to use the existing run/event correlation fallback. The computed durations are then lazily backfilled into a metadata-only checkpoint so subsequent history reads use the fast path.
* Duration-only checkpoints are ignored when locating the base checkpoint for message regeneration.
* Malformed legacy `run_durations` or `additional_kwargs` values are handled defensively instead of breaking the history response.
                                                                   
                                                                                                        
   ## Surface area                                                                                      
                                                                                                        
 (say what it buys us)                                                                                  

## Screenshots / Recording

## Bug fix verification                                                                              
                                                                                                        
   Regression coverage:                                                                                 
                                                                                                        
   - `backend/tests/test_run_duration_checkpoint.py`                                                    
     - duration remains attached to its AI message through later checkpoints;                           
     - concurrent duration updates preserve both turns;                                                 
     - duration checkpoints preserve parent lineage;                                                    
     - duration writes cannot race with an active graph stream.                                         
   -                                                                                                    
 `backend/tests/test_threads_router.py::test_get_thread_history_associates_tool_messages_from_checkpoin 
 t_turn`                                                                                                
     - AI and tool messages receive the correct owning `run_id`.                                        
                                                                                                        
   Did it go red on `main` and green on this branch? **No — the new checkpoint-duration behavior does   
 not exist on `main`, so the regression tests were instead observed red against the pre-fix feature     
 implementation and green after the fixes.**                                                            
                                                                                                        
   ## Validation                                                                                        
                                                                                                        
   Focused backend regression suite:                                                                    
                                                                                                        
   ```bash                                                                                              
   cd backend                                                                                           
   uv run pytest \                                                                                      
     tests/test_run_duration_checkpoint.py \                                                            
     tests/test_threads_router.py \                                                                     
     tests/test_goal_worker.py \                                                                        
     tests/test_run_worker_rollback.py \                                                                
     -q                                                                                                 
 ```                                                                                                    
                                                                                                        
 Result:                                                                                                
                                                                                                        
 ```text                                                                                                
   109 passed, 1 warning in 1.52s                                                                       
 ```                                                                                                    
                                                                                                        
 The warning is the existing FastAPI/Starlette httpx deprecation warning.                               
                                                                                                        
 Focused lint:                                                                                          
                                                                                                        
 ```bash                                                                                                
   cd backend                                                                                           
   uv run ruff check \                                                                                  
     app/gateway/routers/threads.py \                                                                   
     packages/harness/deerflow/runtime/runs/worker.py \                                                 
     tests/test_threads_router.py \                                                                     
     tests/test_run_duration_checkpoint.py                                                              
 ```                                                                                                    
                                                                                                        
 Result:                                                                                                
                                                                                                        
 ```text                                                                                                
   OK                                                                                                   
 ```                                                                                                    
                                                                                                        
 Blocking-I/O scan:                                                                                     
                                                                                                        
 ```bash                                                                                                
   uv run --project backend python scripts/scan_changed_blocking_io.py --base main                      
 ```                                                                                                    
                                                                                                        
 Result:                                                                                                
                                                                                                        
 ```text                                                                                                
   No blocking-IO candidates introduced by this change. 
```

## AI assistance

 ***Tool(s) used***: OpenAI Codex                                                                             
                                                                                                        
 ***How you used it***: Used Codex to review the initial implementation, reproduce the checkpoint association 
 and concurrency failures, implement the fixes with regression tests, validate the affected backend     
 paths, and consolidate/rebase the result into one commit. The implementation was additionally reviewed 
 for checkpoint lineage, saver semantics, concurrent graph writes, and legacy-history compatibility. 

### CODE REVIEW

File: backend/app/gateway/routers/threads.py [hunk 1]
```diff
@@ -1089,56 +1089,81 @@ async def get_thread_history(thread_id: str, body: ThreadHistoryRequest, request
...
-                                        msg["run_id"] = run_id
-                                        if msg.get("type") == "ai" and run_id in run_durations:
-                                            if "additional_kwargs" not in msg:
-                                                msg["additional_kwargs"] = {}
-                                            msg["additional_kwargs"]["turn_duration"] = run_durations[run_id]
+                        # Human messages carry run_id in additional_kwargs; use
+                        # them as turn boundaries for all following AI/tool
+                        # messages. New checkpoints persist turn_duration on AI
+                        # messages directly; legacy metadata maps remain readable.
+                        ckpt_run_durations = metadata.get("run_durations")
```
willem-bd: 🟠 `metadata.get("run_durations")` is read here (and `run_durations` is stripped from `user_meta` at line 1180), but **nothing writes that key**. `_persist_run_duration` writes `turn_duration` onto AI messages' `additional_kwargs` and sets `metadata["writes"]` (`worker.py:1211`) - it never sets `metadata["run_durations"]`. A repo-wide grep confirms no writer exists, so `ckpt_run_durations` is always `{}` and the block at lines 1114-1119 (`current_turn_run_id in ckpt_run_durations`) never fires. The comment "legacy metadata maps remain readable" is inaccurate.

The feature still works because `additional_kwargs["turn_duration"]` survives `serialize_channel_values_for_api` and is already on the messages; the fallback correctly skips when present. So this is dead code, not a functional bug - but it's a writer/reader contract mismatch. Either have `_persist_run_duration` also write `metadata["run_durations"] = {run_id: duration}` (then this lookup path works and you can drop the message scanning), or remove this block and the `run_durations` entry in the `user_meta` strip.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -448,39 +448,40 @@ def _get_goal_evaluator_model() -> Any:
 
         async def _stream_once(input_payload: Any, stream_config: RunnableConfig) -> None:
             nonlocal llm_error_fallback_message
-            if len(lg_modes) == 1 and not stream_subgraphs:
-                # Single mode, no subgraphs: astream yields raw chunks
-                single_mode = lg_modes[0]
-                async for chunk in agent.astream(input_payload, config=stream_config, stream_mode=single_mode):
+            async with goal_thread_lock(thread_id):
```
willem-bd: 🟠 Wrapping the **entire** `agent.astream(...)` in `goal_thread_lock(thread_id)` is a significant behavioral change - previously streaming didn't hold this lock. Since it's a non-reentrant per-thread `asyncio.Lock`:

- It **blocks same-thread goal commands** (`set_goal`/`clear_goal`/goal reads in `client.py:481,492`) for the whole stream duration - potentially minutes on a long run.
- It can **deadlock if any graph node or tool acquires `goal_thread_lock(thread_id)` during streaming**. The goal-continuation logic acquires it but runs *after* `_stream_once` returns (sequential, not nested), so that's fine - but the test uses a `DummyAgent` that never acquires the lock internally, so a real graph node doing so isn't covered.
- The lock is **scoped per event loop** (`goal.py:66`); if the worker and client sync-bridge run in different loops, they get different locks and don't serialize at all.

The stream-wide lock does prevent a concurrent run's duration write from building on a mid-stream checkpoint. But if `run_manager` already serializes same-thread runs (default `multitask_strategy="reject"`), the narrower lock you already have at `_persist_run_duration` (line 1162) may suffice. Worth validating the deadlock/latency assumptions per the `needs-validation` label.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -1130,6 +1150,86 @@ def _title_generation_state(channel_values: dict[str, Any], graph_input: Any | N
...
+    duration_seconds: int,
+) -> None:
+    """Persist *duration_seconds* on the AI messages owned by *run_id*."""
+    ckpt_config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
+    async with goal_thread_lock(thread_id):
+        ckpt_tuple = await _call_checkpointer_method(checkpointer, "aget_tuple", "get_tuple", ckpt_config)
+        if ckpt_tuple is None:
+            return
+
+        checkpoint = copy.deepcopy(getattr(ckpt_tuple, "checkpoint", {}) or {})
```
willem-bd: 🟡 `copy.deepcopy(checkpoint)` here and `copy.deepcopy(messages)` (line 1169) run on **every** successful run, on the post-stream hot path. The PR's motivation is long-running threads, where `messages` can be large and numerous (big tool results) - this is O(n) in total message size per run. `_persist_goal_evaluation` uses the same pattern but only during goal continuations; this runs unconditionally on success. Consider a shallower copy or copying only the AI messages being mutated.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -615,6 +616,25 @@ async def _stream_once(input_payload: Any, stream_config: RunnableConfig) -> Non
...
 
+        # Persist run duration to checkpoint metadata so history reads
+        # don't need to correlate runs and events.
+        if checkpointer is not None and record.status == RunStatus.success:
+            try:
+                from datetime import datetime
+
+                created = datetime.fromisoformat(record.created_at.replace("Z", "+00:00"))
+                updated = datetime.fromisoformat(record.updated_at.replace("Z", "+00:00"))
+                duration = int((updated - created).total_seconds())
```
willem-bd: 🟡 `duration = int((updated - created).total_seconds())` uses `record.created_at` -> `updated_at`, i.e. wall-clock end-to-end (including any queue wait), not pure execution time. `int()` truncates, so `if duration > 0` skips sub-second runs entirely (a 0.9s turn gets no duration). Both are defensible but should be documented as the intended semantic of `turn_duration`.

Nit: `from datetime import datetime` (line 623) is imported inline - move it to the module imports; it's not a hot-path lazy import.

willem-bd (commented):
Reviewed the checkpoint-duration change. Well-structured with good regression tests (survival through later checkpoints, concurrent updates, parent lineage, stream-vs-duration serialization). The checkpoint write protocol (version bump, parent lineage, `new_versions`) is correct and consistent with `_persist_goal_evaluation`.

Inline findings:
- 🟠 `metadata["run_durations"]` is read but never written - dead code; writer/reader mismatch (feature still works via `additional_kwargs["turn_duration"]`).
- 🟠 Wrapping the entire stream in `goal_thread_lock` is a coarse change - validate deadlock-with-graph-nodes and goal-command latency.
- 🟡 Deep-copies the whole checkpoint+messages on every successful run (perf on long threads).
- 🟡 Test seeds `turn_duration` but doesn't assert it reaches the response.
- 🟡 Duration is wall-clock incl. queue wait; sub-second runs skipped.
- Nit: inline `datetime` import; dense fallback comprehension.

Main thing to resolve before merge: the `run_durations` writer/reader mismatch. Second: validate the stream-wide lock doesn't deadlock or regress goal-command latency (the `needs-validation` label covers this).

backend/tests/test_threads_router.py
```diff
@@ -568,6 +568,31 @@ async def _seed() -> None:
         assert _ISO_TIMESTAMP_RE.match(entry["created_at"]), entry
 
 
+def test_get_thread_history_associates_tool_messages_from_checkpoint_turn() -> None:
```
willem-bd, line 574: 🟡 This test seeds `additional_kwargs={"turn_duration": 4}` but only asserts `run_id` association - it never checks that `turn_duration` actually reaches the response. Given the `run_durations` metadata path is dead (see the threads.py:1096 comment), an end-to-end assertion that `turn_duration` is present on the AI messages in the response would close the loop and guard against a future serialization change stripping it.

backend/app/gateway/routers/threads.py
```diff
@@ -1089,56 +1089,81 @@ async def get_thread_history(thread_id: str, body: ThreadHistoryRequest, request
...
+                                    additional_kwargs = {}
+                                    msg["additional_kwargs"] = additional_kwargs
+                                additional_kwargs.setdefault("turn_duration", ckpt_run_durations[current_turn_run_id])
+
+                        # Fallback: for messages still missing turn_duration
+                        # (old threads before _persist_run_duration existed),
+                        # correlate via event_store + run_mgr for exact per-message
+                        # mapping. The 1000-event window is acceptable here because
+                        # this path only runs for legacy data.
+                        if any(msg.get("type") == "ai" and not (isinstance(msg.get("additional_kwargs"), dict) and "turn_duration" in msg["additional_kwargs"]) for msg in serialized_msgs):
```
willem-bd: Nit: this fallback guard is a dense one-liner. Extracting a small `_ai_message_lacks_duration(msg)` helper would make it readable and reusable.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -1136,6 +1183,90 @@ def _title_generation_state(channel_values: dict[str, Any], graph_input: Any | N
...
+        checkpoint_ns = _checkpoint_namespace(ckpt_tuple)
+        parent_checkpoint_id = _checkpoint_identity(ckpt_tuple, checkpoint)
+        write_config = {
+            "configurable": {
+                "thread_id": thread_id,
+                "checkpoint_ns": checkpoint_ns,
+                "checkpoint_id": parent_checkpoint_id,
+            }
+        }
+        await _call_checkpointer_method(
```
willem-bd: Each successful run now appends a checkpoint row **plus a full `messages` channel blob** - DB-backed savers (PostgresSaver / SqliteSaver v4) persist the entire messages list per bumped version (per the comment in `_ensure_interrupted_title`). That's a read-optimization paid for at write time and in storage; worth noting for production Postgres threads.

Concrete downstream effect: `_find_base_checkpoint_before_human` walks checkpoints via `alist(limit=REGENERATE_HISTORY_SCAN_LIMIT)` where the limit is `200` (`thread_runs.py:33`). Previously ~1 checkpoint/turn -> ~200 turns covered; now ~2/turn -> **~100 turns**. Regenerating an older message in a 100-200-turn thread can now fail with "Could not locate target user message."

Options: raise the limit, filter out `metadata.writes.runtime_run_duration` checkpoints from that scan, or persist duration in checkpoint `metadata` instead of mutating `messages` (avoids the channel-version bump and blob write entirely).

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -1136,6 +1183,90 @@ def _title_generation_state(channel_values: dict[str, Any], graph_input: Any | N
...
+    *,
+    checkpointer: Any,
+    thread_id: str,
+    run_id: str,
+    duration_seconds: int,
+) -> None:
+    """Persist *duration_seconds* on the AI messages owned by *run_id*."""
+    ckpt_config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
+    async with _checkpoint_thread_lock(thread_id):
+        ckpt_tuple = await _call_checkpointer_method(checkpointer, "aget_tuple", "get_tuple", ckpt_config)
```
willem-bd: This read-modify-write holds `_checkpoint_thread_lock`, but the goal-eval and `_ensure_interrupted_title` checkpoint writes hold `goal_thread_lock` **instead** - the two lock domains don't overlap. If a goal/title write lands between this `aget_tuple` and the `aput` below, the duration checkpoint sets `parent = <stale id>` and the intervening write is orphaned out of the latest lineage (latest pointer jumps back, dropping the goal/title update).

`_ensure_interrupted_title` defends against exactly this with a 3-retry optimistic re-check (`_checkpoint_identity` before vs. after re-reading). `_persist_run_duration` doesn't re-verify identity before writing. Practical probability is low (runs/thread are largely serialized by `RunManager`, and goal writes + duration write are sequential *within* a run), but this PR is `risk:high` and concurrency-focused - so the missing re-check is a real gap. Recommend mirroring the title path's optimistic retry here.

backend/app/gateway/routers/threads.py
```diff
@@ -1089,56 +1094,71 @@ async def get_thread_history(thread_id: str, body: ThreadHistoryRequest, request
...
+                                            run_id = additional_kwargs.get("run_id")
+                                            if isinstance(run_id, str) and run_id:
+                                                current_turn_run_id = run_id
+                                        continue
+
+                                    if msg.get("type") in {"ai", "tool"}:
+                                        msg_id = msg.get("id")
+                                        run_id = msg_to_run.get(msg_id) or current_turn_run_id
+                                        if run_id:
+                                            if "run_id" not in msg:
```
willem-bd: Subtle behavior change vs. the original correlation path. The original did `msg["run_id"] = run_id` unconditionally, so the **exact** event-store `msg_to_run` mapping (per-message) overrode the turn-boundary mapping. The new first pass sets `run_id` from the human message's `additional_kwargs.run_id` first, and this `if "run_id" not in msg` guard means the fallback won't overwrite it with the exact mapping.

In practice the two usually agree, but for legacy threads whose human messages carry `run_id`, the exact per-message mapping no longer wins. Also, **no test exercises this fallback branch** - `test_get_thread_history_associates_tool_messages_from_checkpoint_turn` seeds checkpoints that already have `turn_duration`, so it only hits the first pass. Suggest adding a legacy case (AI messages without `turn_duration` -> durations injected via the event-store path) to lock in the intended precedence.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -615,6 +643,25 @@ async def _stream_once(input_payload: Any, stream_config: RunnableConfig) -> Non
             except Exception:
                 logger.debug("Failed to sync title for thread %s (non-fatal)", thread_id)
 
+        # Persist run duration to checkpoint metadata so history reads
+        # don't need to correlate runs and events.
+        if checkpointer is not None and record.status == RunStatus.success:
+            try:
+                created = datetime.fromisoformat(record.created_at.replace("Z", "+00:00"))
```
willem-bd, line 656: Minor: this duration is computed from the in-memory `record.updated_at`, captured *before* `set_finalizing(False)` (line 679) and the `on_run_completed` hook. Legacy `compute_run_durations` reads `updated_at` from the DB at history time (post-quiescence), so the persisted value can be a second or two shy of what legacy would have computed. The comment says "match legacy history semantics" - it's close but not an exact match. Already within the "slightly overshoot" tolerance noted in the legacy code, so acceptable; flagging only for fidelity.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -62,6 +67,28 @@
 
 logger = logging.getLogger(__name__)
 
+_checkpoint_locks_guard = threading.Lock()
+_checkpoint_locks_by_loop: weakref.WeakKeyDictionary[asyncio.AbstractEventLoop, dict[str, asyncio.Lock]] = weakref.WeakKeyDictionary()
```
willem-bd, line 72: `_checkpoint_locks_by_loop` evicts by event-loop GC, but the inner `dict[str, asyncio.Lock]` grows unbounded for the loop's lifetime (one Lock per `thread_id` ever seen, never evicted). This mirrors the existing `goal_thread_lock` shape, so it's consistent - flagging only because it's now a second such cache. Per-entry size is negligible; only a long-running production process accumulates. Low severity.

willem-bd (commented):
Reviewed the checkpoint-duration change. Overall it's a solid, well-tested refactor that removes a real read-path pathology (per-read run/event queries + the 1000-event cap silently dropping durations on long threads). Duration math matches `compute_run_durations`, parent lineage is correct, and the concurrency model is coherent.

Inline comments below flag the higher-impact items: checkpoint/storage amplification (which also halves regenerate scan coverage on long threads) and a lost-update window vs. the goal/title checkpoint writers. The rest are minor.

One scope note (not inline since the file isn't in this diff): `thread_runs.py::list_thread_messages` still calls `run_mgr.list_by_thread` + `compute_run_durations` on every request, so the run-store-query-on-every-read cost is removed from `/history` only, not `/messages`. Not a bug - just calibrating the PR's framing.

rayhpeng:
Overall this is a solid, careful change — it fixes a real bug (the read path's
`list_messages(limit=1000)` silently drops durations on long threads), and the
concurrency story is right: the per-(loop, thread) lock plus re-`aget_tuple`
*inside* the lock stops one run from clobbering another's duration, and the
`_YieldingSaver` test pins that race well. A few things I'd want addressed
before merge:

**1. Fallback silently downgrades `run_id` precision for legacy data** (`get_thread_history`)
The fast path now `setdefault`s every AI/tool message's `run_id` to the
turn-boundary id, and the fallback then guards with `if "run_id" not in msg` —
so the event store's *exact* `msg_to_run` mapping can no longer override it.
The old code was `msg_to_run.get(msg_id) or current_turn_run_id` (exact-first).
For legacy threads where one turn spans multiple runs, `run_id` regresses from
exact → boundary. `turn_duration` stays correct (it's computed from the local
exact id), only the `run_id` field changes — but it's a silent behavior change.
Suggest letting the exact mapping win, or skipping the fast-path pre-fill when
the fallback is going to run.

**2. Legacy threads are pinned to the slow path indefinitely** (`get_thread_history`)
`any(_ai_message_lacks_duration(...))` re-runs the full runs + events scan on
*every* read if even one AI message lacks a duration, and old turns are never
backfilled (`_persist_run_duration` only writes for a *new* successful run's own
messages). So any thread with pre-existing history keeps paying the correlation
cost forever and never converges onto the fast path. Consider backfilling the
fallback-computed durations into the checkpoint once (one-time migration), or
calling out in the PR description that existing threads stay on the old path.

**3. Confirm the storage tradeoff on SQL backends** (`_persist_run_duration`)
Each successful run writes an extra checkpoint carrying a full copy of the
messages list. The read win is backend-dependent and worth being explicit about:
near-zero on the `memory` backend (the two old queries were in-process dict scans
anyway), but real on MySQL/Postgres — it removes two genuine round-trips plus a
≤1000-row event scan. The catch is the *write* cost is highest on exactly those
SQL backends too: a full-messages checkpoint INSERT per turn. It's a sound
"write-to-save-reads" trade for a read-heavy history panel, I'd just like to
confirm the long-thread storage growth was measured.

**4. Test gap** — the fallback (legacy → event/run correlation) has no
integration coverage. `test_get_thread_history_associates_tool_messages_from_checkpoint_turn`
seeds `turn_duration` on every message, so only the fast path is exercised; a
case with duration-less AI messages would lock the legacy path (and the #1
interaction).

None of these are blockers — #1 and #3 are the ones I'd want a response on
before merge. Nice work overall.

Vanzeren:
- **Checkpoint concurrency:** duration writes re-read the latest checkpoint identity before          
 committing and retry when an intervening title/goal checkpoint advanced the lineage. The duration      
 writer remains serialized with same-thread graph streaming through the dedicated checkpoint lock. It   
 intentionally does not hold `goal_thread_lock` for the duration of a stream, so same-thread goal       
 commands remain independent and graph code can still acquire the goal lock without self-deadlock.      
                                                                                                        
- **Duration semantics:** `turn_duration` remains the non-negative integer `RunRecord` lifetime in   
 seconds, including admission delay. Successful sub-second runs persist `0` rather than being omitted.  
                                                                                                        
   Added regression coverage for metadata-only persistence, unchanged messages channel version,         
 concurrent duration updates, title-write interleaving, history response injection, legacy exact        
 `run_id` precedence plus one-time backfill, duration-only checkpoint filtering during regenerate,      
 zero-duration runs, and graph access to `goal_thread_lock` during streaming.                           
                                                                                                        
   Validation:                                                                                          
   - `cd backend && make lint`                                                                          
   - `cd backend && make test` — `7182 passed, 22 skipped`                                              
   - Focused duration/history/regenerate suite — `109 passed`                                           
   - Changed-delta blocking-I/O scan — no candidates introduced

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -1142,6 +1189,82 @@ def _title_generation_state(channel_values: dict[str, Any], graph_input: Any | N
...
+                }
+            }
+            await _call_checkpointer_method(
+                checkpointer,
+                "aput",
+                "put",
+                write_config,
+                checkpoint,
+                metadata,
+                {},
```
willem-bd, line 1258: PR description says duration checkpoints *bump and publish the `messages` channel version*, but `new_versions` is `{}` here and `channel_versions["messages"]` is inherited unchanged from the parent. The code is actually metadata-only (correct — the parent's messages blob is referenced by version, confirmed for both InMemorySaver and DB-backed savers), which the docstring above also states. Could you update the description to match? As written it implies a messages-channel bump that doesn't happen.

backend/app/gateway/routers/threads.py
```diff
@@ -1089,56 +1111,71 @@ async def get_thread_history(thread_id: str, body: ThreadHistoryRequest, request
...
+                                    run_id = msg_to_run.get(msg.get("id")) or current_turn_run_id
                                     if run_id:
                                         msg["run_id"] = run_id
-                                        if msg.get("type") == "ai" and run_id in run_durations:
-                                            if "additional_kwargs" not in msg:
-                                                msg["additional_kwargs"] = {}
-                                            msg["additional_kwargs"]["turn_duration"] = run_durations[run_id]
+                                        _set_message_turn_duration(msg, run_id, run_durations)
+
+                                await _persist_run_durations(
```
willem-bd: This `get_thread_history` endpoint (`threads.read`) performs a checkpoint write as a lazy-migration side effect. It's non-fatal and one-time per legacy thread, but two things worth a look:
(1) a read path mutating durable state is a semantic surprise — a short comment here flagging the write-on-read would help;
(2) `_persist_run_durations` acquires `_checkpoint_thread_lock`, so the first history read of a legacy thread during an active same-thread stream blocks until the stream releases the lock.
Migrated threads hit the fast path and are unaffected.

backend/app/gateway/routers/threads.py
```diff
@@ -1089,56 +1111,71 @@ async def get_thread_history(thread_id: str, body: ThreadHistoryRequest, request
...
+                            msg.setdefault("run_id", current_turn_run_id)
+                            _set_message_turn_duration(msg, current_turn_run_id, checkpoint_run_durations)
+
+                        # Legacy checkpoints without duration metadata are
+                        # correlated once via event-store + run-manager, then
+                        # upgraded by a metadata-only checkpoint write.
+                        if any(_ai_message_lacks_duration(msg) for msg in serialized_msgs):
+                            from app.gateway.deps import get_run_event_store, get_run_manager
+                            from app.gateway.routers.thread_runs import compute_run_durations
+                            from deerflow.runtime.runs.worker import _persist_run_durations
```
willem-bd: Importing `_persist_run_durations` (underscore-prefixed) couples the gateway router to harness internals and will break silently if the worker refactors this. Consider exposing a public wrapper (e.g. `persist_run_durations`) for this cross-module call.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -1142,6 +1189,82 @@ def _title_generation_state(channel_values: dict[str, Any], graph_input: Any | N
...
+
+            checkpoint = dict(getattr(ckpt_tuple, "checkpoint", {}) or {})
+            metadata = dict(getattr(ckpt_tuple, "metadata", {}) or {})
+            raw_run_durations = metadata.get("run_durations")
+            run_durations = {key: value for key, value in raw_run_durations.items() if isinstance(key, str) and key and isinstance(value, int) and not isinstance(value, bool)} if isinstance(raw_run_durations, dict) else {}
+            changed_durations = {run_id: duration for run_id, duration in updates.items() if run_durations.get(run_id) != duration}
+            if not changed_durations:
+                return False
+
+            run_durations.update(changed_durations)
```
willem-bd, line 1229: `run_durations` is merged cumulatively, so the latest checkpoint's `run_durations` grows to O(runs-in-thread); the legacy backfill path also persists all runs from `list_by_thread` at once. For long-lived threads this is a growing per-write and per-checkpoint cost. Not blocking, but worth a pruning strategy or storing only the current run's duration (the history fast path only reads the latest checkpoint anyway).

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -1142,6 +1189,82 @@ def _title_generation_state(channel_values: dict[str, Any], graph_input: Any | N
...
 
 
+async def _persist_run_durations(
+    *,
+    checkpointer: Any,
+    thread_id: str,
+    durations: dict[str, int],
+) -> bool:
+    """Merge validated run durations into a metadata-only checkpoint."""
+    updates = {run_id: max(0, duration_seconds) for run_id, duration_seconds in durations.items() if isinstance(run_id, str) and run_id and isinstance(duration_seconds, int) and not isinstance(duration_seconds, bool)}
```
willem-bd: The `isinstance(x, int) and not isinstance(x, bool)` duration-validation guard is repeated here, again at L1213 for `raw_run_durations`, and in `threads.py::_checkpoint_run_durations`. A shared `_valid_duration_entry` helper would reduce drift risk.

backend/app/gateway/routers/threads.py
```diff
@@ -1152,7 +1189,7 @@ async def get_thread_history(thread_id: str, body: ThreadHistoryRequest, request
             next_tasks = [t.name for t in tasks_raw if hasattr(t, "name")]
 
             # Strip LangGraph internal keys from metadata
-            user_meta = {k: v for k, v in metadata.items() if k not in ("created_at", "updated_at", "step", "source", "writes", "parents")}
+            user_meta = {k: v for k, v in metadata.items() if k not in ("created_at", "updated_at", "step", "source", "writes", "parents", "run_durations")}
```
willem-bd, line 1203: Unlike the regenerate path (which filters duration-only checkpoints via `_is_duration_only_checkpoint`), this history loop iterates *all* checkpoints, so each successful run adds a phantom history step (bumped `step`, no messages except when latest). This matches the existing `_ensure_interrupted_title` pattern so the frontend must already tolerate it — but this PR raises the cadence (one per successful run vs. one per interrupted run), so worth a conscious confirmation.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -31,6 +31,13 @@
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/api/threads", tags=["runs"])
 REGENERATE_HISTORY_SCAN_LIMIT = 200
+REGENERATE_HISTORY_RAW_SCAN_LIMIT = REGENERATE_HISTORY_SCAN_LIMIT * 2
```
willem-bd, line 38: Doubling the raw scan assumes duration-only checkpoints are ~50% of history (one per successful run). That holds in steady state, but a burst of duration-only checkpoints without intervening graph checkpoints could starve the 200-effective budget. `test_find_base_checkpoint_ignores_duration_only_checkpoints` pushes exactly 200 duration checkpoints — right at the boundary. Low real-world risk; just flagging the assumption.

WillemJiang (approved):

Issue #4044, opened 2026-07-10
fix(security): html-escape fact content fields before json.dumps in MEMORY_UPDATE_PROMPT


sontianye:
## Context

PR #4028 applied `html.escape()` to fact `content` and `category` in `_build_staleness_section` and `_build_consolidation_section` (the *action* sections that instruct the model to delete or merge facts).

The always-on `MEMORY_UPDATE_PROMPT` path formats the full memory state via `json.dumps(current_memory)` wrapped in `<current_memory>...</current_memory>`. `json.dumps` escapes `"` and `\` but not `<`, `>`, or `&`, so a fact `content` containing `</current_memory><evil>...</evil>` reaches the model verbatim in this block.

## Why deferred from #4028

This path is lower-risk than the action sections (read-only context, not delete/merge instructions). Blanket-escaping the full JSON blob would corrupt its structure.

## Correct fix

Escape `content` (and other user-influenced fields) at fact-content insert time, before they are stored — then `json.dumps` serializes already-safe values. Alternatively, apply `html.escape()` to each fact's `content` field immediately before passing to `json.dumps`.

## References

- PR #4028 — code comment at `updater.py` (`current_memory=json.dumps(...)`)

PR #4122, opened 2026-07-12
feat(memory): pluggable memory abstraction with self-contained DeerMem backend

qin-chenghan:
## Why

The current memory system is tightly coupled to DeerFlow internals — `MemoryUpdater` imports `create_chat_model`, `get_paths`, `inject_langfuse_metadata`, and `resolve_runtime_user_id` directly. This makes it impossible to swap the memory backend without touching DeerFlow source code.

This PR introduces a **pluggable memory abstraction** with a self-contained default backend (DeerMem), so that:

1. **Swap backend = drop a folder + change one config line.** A new `backends/<name>/` directory exposing `MANAGER_CLASS` is auto-discovered. No code changes elsewhere.
2. **The default backend (DeerMem) is self-contained.** It owns its LLM factory (`core/llm.py` via `init_chat_model`), path resolution (`core/paths.py`), and config (`DeerMemConfig`) — zero `from deerflow` imports inside the backend package (except the ABC contract line).
3. **Tool-driven memory mode (#4023) is preserved and adapted.** All 4 memory tools (`memory_search`, `memory_add`, `memory_update`, `memory_delete`) now route through the `MemoryManager` ABC instead of hard-importing DeerMem functions. `hasattr` capability probing means non-DeerMem backends don't crash — they return a JSON error.

### New: MemoryManager ABC + pluggable backend factory (`memory/manager.py`)

- **9-method ABC**: `add` / `add_nowait` / `get_context` / `search` / `get_memory` / `delete_memory` / `clear_memory` / `import_memory` / `export_memory`
- **Singleton factory** (`get_memory_manager()`): resolves `config.manager_class` → discovers backend → injects `backend_config` + host-default hooks
- **Drop-in backend discovery**: scans `backends/<name>/` for `MANAGER_CLASS` attribute; supports registered short names (`deermem`, `noop`) and dotted paths (`pkg.mod:Cls`)
- **Host-default hooks**: `tracing_callback` (Langfuse) and `should_keep_hidden_message` (human-input filter) injected by the factory at the host seam, not imported by the backend

### New: DeerMem default backend (`backends/deermem/`)

- **Self-contained package**: `deermem/config.py` (DeerMemConfig), `deermem/core/` (llm, paths, storage, queue, updater, prompt, message_processing)
- **No DeerFlow imports** inside the backend package (except the ABC contract). LLM via `langchain.chat_models.init_chat_model`, paths via inlined `safe_user_id` + `AGENT_NAME_PATTERN`, config via `DeerMemConfig.from_backend_config()`
- **DeerMem-internal capabilities** (`create_fact` / `delete_fact` / `update_fact` / `reload_memory` / `warm`) exposed via `hasattr` probing — not on the ABC

### New: Noop backend (`backends/noop/`)

- Empty 9-method implementation — proves pluggability end-to-end
- Doubles as the **template** for new backends (annotated with step-by-step instructions)

### Adapted: Tool-driven memory mode (preserves #4023)

- `memory.mode: "tool"` still registers `memory_search` / `memory_add` / `memory_update` / `memory_delete`
- All 4 tools now route through `get_memory_manager()` ABC methods instead of direct DeerMem imports
- `create_fact` / `update_fact` / `delete_fact` reached via `getattr` + `callable` check — absent backend returns JSON error, not crash

### Adapted: All call sites routed through ABC

| Call site | Before | After |
|-----------|--------|-------|
| `MemoryMiddleware.after_agent` | `MemoryUpdateQueue.add()` | `get_memory_manager().add()` |
| `summarization_hook` | `MemoryUpdateQueue.add_nowait()` | `get_memory_manager().add_nowait()` |
| `lead_agent/prompt.py` | `format_memory_for_injection(get_memory_data())` | `get_memory_manager().get_context()` |
| `gateway/routers/memory.py` | `get_memory_data()` / `clear_memory_data()` / `import_memory_data()` | `manager.get_memory()` / `clear_memory()` / `import_memory()` |
| `client.py` | `create_memory_fact()` / `delete_memory_fact()` / `update_memory_fact()` | `manager.create_fact()` / `delete_fact()` / `update_fact()` (via `hasattr`) |
| `factory.py` + `agent.py` | (new in #4023) direct imports | `should_use_memory_tools()` + `get_memory_tools()` (unchanged) |

### Config changes (`config.example.yaml`)

```yaml
memory:
  enabled: true
  mode: middleware          # middleware | tool (from #4023)
  injection_enabled: true
  manager_class: deermem    # NEW: backend selector
  backend_config:           # NEW: backend-private config (was flat memory.* fields)
    storage_path: ""
    debounce_seconds: 30
    model:
      # provider: openai
      # model: gpt-4o-mini
    max_facts: 100
    fact_confidence_threshold: 0.7
    max_injection_tokens: 2000
    # ... (all DeerMem-private fields moved here)
```

### Tests

- `test_memory_manager_pluggable.py` — ABC contract + factory + backend discovery
- `test_deermem_self_contained.py` — DeerMem config/llm/paths independence
- Existing memory tests adapted to pluggable layout

## Architecture

```text
                    ┌─── memory.mode ───┐
                    │                    │
              "middleware"            "tool"
                    │                    │
         ┌──────────▼──────────┐   ┌────▼──────────────┐
         │ MemoryMiddleware    │   │ MemoryToolSet     │
         │ (after_agent hook)  │   │ (tools registered │
         │ queue → debounce    │   │  to model)        │
         │ → LLM summarize     │   │                   │
         └──────────┬──────────┘   └────────┬──────────┘
                    │                       │
         ┌──────────▼───────────────────────▼──────────┐
         │          MemoryManager ABC (9 methods)      │
         │          get_memory_manager() singleton     │
         └──────────┬──────────────────────────────────┘
                    │
         ┌──────────▼──────────────────────────────────┐
         │       backends/<name>/ (drop-in)            │
         │  ┌─────────────┐  ┌─────────────┐           │
         │  │  deermem/   │  │   noop/     │  ...      │
         │  │  (default)  │  │  (template) │           │
         │  └──────┬──────┘  └─────────────┘           │
         └─────────┼───────────────────────────────────┘
                   │
      DeerMem owns: storage / queue / updater / llm / paths / config
      (zero `from deerflow` imports inside backend package)
```

## Surface area

- [x] **Backend API** - gateway memory router adapted to ABC (response shapes unchanged)
- [x] **Agents / LangGraph** - agent factory + lead_agent wiring through MemoryManager
- [x] **Default behavior change** - memory config structure changes (`memory.*` → `memory.backend_config.*`), but `manager_class: deermem` defaults make behavior identical

### Backend

```bash
cd backend && make lint
cd backend && uv run pytest tests/test_memory_manager_pluggable.py tests/test_deermem_self_contained.py
cd backend && uv run pytest tests/test_memory_tools.py tests/test_memory_search.py
cd backend && uv run pytest tests/test_memory_updater.py tests/test_memory_staleness_review.py
```

### Config example

```bash
uv run --project backend python - <<'PY'
from pathlib import Path
import yaml

with Path('config.example.yaml').open() as f:
    data = yaml.safe_load(f)

assert data['memory']['manager_class'] == 'deermem'
assert data['memory']['mode'] == 'middleware'
print('manager_class:', data['memory']['manager_class'])
print('mode:', data['memory']['mode'])
PY
```

### Results

- backend lint passed
- pluggable + self-contained tests passing
- tool-mode tests passing (adapted to ABC routing)
- existing memory regression tests passing
- config.example.yaml parses with correct defaults

## AI assistance

**Tool(s) used:** Claude (Anthropic)

**How you used it:** AI assisted with the pluggable architecture design, backend discovery mechanism, and self-contained DeerMem refactoring. Human reviewers reviewed every line, ran validation, and refined the final design.


WillemJiang:
@qin-chenghan Thanks for your contribution. Please resolve the conflict with the main branch and fix the lint error & unit test errors.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -775,14 +624,8 @@ def _prepare_update_prompt(
...
 
-        # HTML-escape user-influenced string values before embedding the memory
-        # state as a JSON blob inside <current_memory>...</current_memory>, so a
-        # fact/summary containing </current_memory> cannot break out of the block
-        # (prompt injection, #4044). Escaping values — not the serialized blob —
-        # keeps the JSON well-formed because json.dumps re-quotes safe values.
-        # The unescaped current_memory is returned unchanged for the apply path.
         prompt = MEMORY_UPDATE_PROMPT.format(
-            current_memory=json.dumps(_escape_memory_for_prompt(current_memory), indent=2, ensure_ascii=False),
+            current_memory=json.dumps(current_memory, indent=2, ensure_ascii=False),
```
willem-bd: RED **Security regression - reopens #4044.**

The `current_memory` blob is no longer HTML-escaped before being embedded in the `<current_memory>{current_memory}</current_memory>` block of `MEMORY_UPDATE_PROMPT` (verified: the template at `core/prompt.py:26-28` still uses that block). The pre-refactor code called `json.dumps(_escape_memory_for_prompt(current_memory), ...)`, which escaped every string leaf so a fact `content` like `</current_memory><evil>...` couldn't break out of the block.

`_escape_memory_for_prompt` was deleted during the move. I confirmed it was present at the merge base (`8be7411d`, line 661/785) and is gone here, so this is a true regression. Please restore it and this call site. Note escaping *is* preserved in `_build_consolidation_section` and `_format_fact_line`, so this looks like an accidental drop rather than a design choice.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -591,11 +375,11 @@ def _build_staleness_section(
     lines: list[str] = []
     for fact in stale_candidates:
         fid = fact.get("id", "?")
-        cat = html.escape(str(fact.get("category", "context")).strip() or "context")
+        cat = str(fact.get("category", "context")).strip() or "context"
         conf = fact.get("confidence", 0.0)
         created_raw = fact.get("createdAt", "")
         created_short = created_raw[:10] if isinstance(created_raw, str) and len(created_raw) >= 10 else created_raw
-        content = html.escape(str(fact.get("content", "")))
+        content = str(fact.get("content", ""))
```
willem-bd: RED **Security regression - reopens #4028.**

`content` (and `cat` on line 378) are no longer `html.escape`d before being rendered into the `<stale_facts>...</stale_facts>` block of `STALENESS_REVIEW_PROMPT`. Pre-refactor these were `html.escape(str(...))` (present at the merge base `8be7411d`, lines 594/598). A fact whose `content` contains `</stale_facts>` or prompt-control text could influence the KEEP/REMOVE decision or break the block.

Restore the `html.escape` on both `cat` and `content` here.

config.example.yaml
```diff
@@ -1383,87 +1383,94 @@ summarization:
...
-  consolidation_enabled: false
-  consolidation_min_facts: 8
-  consolidation_max_groups_per_cycle: 3
-  consolidation_max_sources: 8
+  # Backend-private config (a dict), passed verbatim to the backend __init__.
+  # Each backend self-interprets it (DeerMem parses it into DeerMemConfig).
+  backend_config:
+    storage_path: ""            # empty = deer-flow base_dir (factory injects absolute runtime_home); absolute path opts out of per-user isolation
+    debounce_seconds: 30 # Wait time before processing queued updates
+    model:                      # LLM for memory extraction; omit all fields = no extraction (non-LLM ops still work; an update raises)
```
willem-bd, line 1497: RED **Default-behavior regression - contradicts the PR's "behavior identical" claim.**

Shipping `model:` with all fields commented means `build_llm()` returns `None` (`core/llm.py`), so `MemoryUpdater._llm` is `None` and every memory update raises `RuntimeError("...no LLM is configured...")` (`core/updater.py:714`), which is caught and logged as `Memory update failed` -> `return False`. Under the shipped default config, memory extraction silently does nothing.

Pre-refactor, `model_name: null` resolved to `create_chat_model(name=None)`, which fell back to `config.models[0].name` (the app's default model) - so memory worked out of the box. Upgraders with `model_name: null` or `model_name: <name>` also silently lose extraction, because the old field is no longer in `MemoryConfig` and `load_memory_config_from_dict` drops unknown keys.

Options: (a) have the factory inject the host's default model into `backend_config.model` when empty (mirrors how `tracing_callback` is injected); (b) ship a working `model:` example here; (c) at minimum, emit a loud startup warning when DeerMem is active with no model. And please correct the "behavior identical" note in the PR description.

backend/app/gateway/routers/memory.py
```diff
@@ -115,27 +125,10 @@ class FactPatchRequest(BaseModel):
...
-        description="Token ceiling for guaranteed-category facts (displaces regular lines in the common case; additive only when guaranteed alone overflows max_injection_tokens)",
-    )
-    staleness_review_enabled: bool = Field(..., description="Whether staleness review is enabled for aged facts")
-    staleness_age_days: int = Field(..., description="Facts older than this many days are candidates for staleness review")
-    staleness_min_candidates: int = Field(..., description="Minimum stale facts required to trigger a review cycle")
-    staleness_max_removals_per_cycle: int = Field(..., description="Maximum number of facts staleness review can remove per cycle")
-    staleness_protected_categories: list[str] = Field(..., description="Fact categories exempt from staleness review")
+    enabled: bool = Field(..., description="Whether the memory mechanism is enabled (call-site gate).")
+    injection_enabled: bool = Field(..., description="Whether memory is injected into the system prompt (call-site gate).")
+    manager_class: str = Field(..., description="Active memory backend selector (backend name or dotted path).")
```
willem-bd, line 133: ORANGE **Breaking API response shape change** (the PR description says "response shapes unchanged").

`MemoryConfigResponse` now exposes only `enabled` / `injection_enabled` / `manager_class` / `backend_config`; the previous `storage_path`, `debounce_seconds`, `max_facts`, `fact_confidence_threshold`, `max_injection_tokens`, `token_counting`, `guaranteed_*`, `staleness_*` fields are gone. Same for `client.get_memory_config()`. Any frontend settings page or API consumer reading those flat fields breaks - and the `needs-validation` + `area:frontend` labels suggest the frontend hasn't been updated. The `frontend/.../datetime.ts` guard against `""` `lastUpdated` also hints at response-shape drift from non-DeerMem backends. Please audit the frontend consumer and update the PR description.

backend/packages/harness/deerflow/agents/memory/manager.py
```diff
@@ -0,0 +1,386 @@
...
+    # importable; if it is not there is no further fallback, so surface a clear
+    # error rather than a raw import traceback (mirrors the defensive style of
+    # get_memory_storage's fallback, which has a concrete default to fall to).
+    logger.warning("Falling back to DeerMem memory backend (manager_class=%r unresolved)", manager_class)
+    try:
+        from deerflow.agents.memory.backends.deermem.deer_mem import DeerMem
+    except Exception:  # noqa: BLE001
+        logger.exception("DeerMem default backend failed to import; no memory backend available")
+        raise RuntimeError("DeerMem default memory backend could not be imported") from None
+    return DeerMem
```
willem-bd: YELLOW If an explicit `manager_class` fails to resolve (typo, import error, missing attr), `_resolve_manager_class` logs a warning and silently falls back to DeerMem.

For a *memory* backend, silently switching to a different storage implementation can route writes to the wrong store - the user thinks they're on backend X, but data lands in DeerMem's file store. An explicit `manager_class` that fails to resolve should raise (or at least refuse to serve) rather than quietly fall back. Fail-fast is safer than a half-working wrong backend. (Falling back only makes sense when `manager_class` is unset/default.)

backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -70,13 +71,14 @@ def memory_search_tool(
...
+        results = get_memory_manager().search(
             query,
-            category=category,
-            limit=limit,
-            agent_name=agent_name,
+            top_k=limit,
             user_id=user_id,
+            agent_name=agent_name,
         )
+        if category is not None:
```
willem-bd: YELLOW **Behavior regression in tool-mode search.**

`manager.search(query, top_k=limit)` slices to `limit` *before* the category filter on line 80. The old `search_memory_facts` filtered by category first, then sliced. So a category-scoped search can now return fewer than `limit` results even when more matching facts exist just below the `top_k` cutoff.

Fix: push `category` into the ABC `search` signature (so the backend filters before slicing), or over-fetch then slice here.

willem-bd (commented):
Reviewed the pluggable-memory abstraction. The ABC/factory design and the self-contained DeerMem package are well executed, and the `noop` template is a great drop-in example. Requesting changes - a few items below are blockers.

**First: this branch is behind `main` by 4 commits and currently `CONFLICTING`/`DIRTY`.** Notably it's missing #4119 (`feb28707`, the `_escape_summary` summary-escaping fix). A rebase is needed before merge; see the rebase caution at the end.

**Blockers introduced by this PR's own changes:**

1. **Two prompt-injection escaping regressions** - the refactor drops `_escape_memory_for_prompt` (reopens #4044) and the staleness-section `html.escape` (reopens #4028). I verified both were present at the merge base (`8be7411d`) and are gone in the head, so this is a true regression, not a stale-branch artifact. The `<current_memory>` / `<stale_facts>` template blocks are unchanged, so the injection vectors are real. Escaping *is* preserved in `_build_consolidation_section` and `_format_fact_line`, so these look like accidental drops during the move. (Inline comments on each.)
2. **Breaks memory extraction under the shipped default config** - `model:` empty -> `build_llm()` returns `None` -> every update raises (caught + logged, silently no-ops). Pre-refactor, `model_name: null` used the app default model. This contradicts the PR's "behavior identical" claim and silently breaks extraction for new users and upgraders. (Inline comment on `config.example.yaml`.)
3. **Changes the config-response API shape** despite the "response shapes unchanged" note - `MemoryConfigResponse` lost the flat DeerMem fields; affects the frontend memory-settings consumer. (Inline comment on the router.)

**Lower-priority (inline):** silent DeerMem fallback on a bad `manager_class`; `memory_search` filtering by category *after* `top_k`.

**Rebase caution (re: #4119):** the head branched before #4119, so `core/prompt.py`'s `format_memory_for_injection` currently interpolates the Work/Personal/Current Focus/Recent/Earlier/Background summaries raw (no `_escape_summary`). When you rebase onto current main, #4119's `_escape_summary` lands in the old `memory/prompt.py` path that this PR renames - expect a conflict in `core/prompt.py`, and please make sure the summary escaping survives the rename rather than getting dropped. (Not attached inline because that region isn't in the diff yet - it's a pending conflict.)

**Minor (no inline thread):** the `request_trace_context(...)` `with` block was dropped from `_do_update_memory_sync`, so log records on the Timer/executor thread no longer carry the request trace id (it now only reaches `tracing_callback`); the langfuse `assistant_id`/`environment` kwargs were dropped from `_host_default_tracing_callback`; `delete_memory`/`export_memory` are ABC methods that `raise NotImplementedError` (consider a 501 capability-probe, or omit from the ABC until implemented); `client.py` fact-CRUD doesn't pass `user_id` (pre-existing, but `get_memory` does scope by user).

Happy to help verify the escaping restoration or the default-model fallback.

pclin1212:
Consider abstracting the common processes of memoryManager into a base class to improve accessibility and location.

backend/app/gateway/routers/memory.py
qin-chenghan: This is an intended design change, not a regression. With pluggable backends, DeerMem-private fields (`storage_path`, `debounce_seconds`, `max_facts`, etc.) no longer belong on the shared `MemoryConfigResponse` — different backends have different config schemas. The response now exposes `manager_class` + `backend_config` (a dict) for backend-private config.

We audited the frontend: `MemorySettingsPage` and `core/memory/` only consume the memory data API (`/api/memory` returning `UserMemory` with `user`/`history`/`facts`), not the config API (`/api/memory/config`). No frontend code reads `storage_path`, `debounce_seconds`, or any of the removed flat fields. So no frontend breakage in practice.

We will update the PR description to accurately reflect this: "Memory data API (`/api/memory`) response shape unchanged. Memory config API (`/api/memory/config`) response shape changed from flat fields to `manager_class` + `backend_config` dict — frontend does not consume this endpoint."

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -193,6 +77,10 @@ def set_memory_config(config: MemoryConfig) -> None:
...
 
 def load_memory_config_from_dict(config_dict: dict) -> None:
-    """Load memory configuration from a dictionary."""
+    """Load memory configuration from a dictionary.
+
+    DeerMem-private keys (from a not-yet-migrated config) are ignored -- they
+    must live under ``backend_config`` to reach the backend.
+    """
     global _memory_config
     _memory_config = MemoryConfig(**config_dict)
```
willem-bd, line 179: Existing `config.yaml` files carry DeerMem-private fields at the **top level** (`memory.storage_path`, `memory.max_facts`, `memory.debounce_seconds`, `memory.model_name`, `memory.token_counting`, staleness/consolidation ...). `MemoryConfig` is a bare `BaseModel` (pydantic default `extra='ignore'`), so `MemoryConfig(**config_dict)` silently drops every one of them -- a user who set `max_facts: 50` or `token_counting: "char"` silently reverts to defaults with no warning.

The PR body's "manager_class: deermem defaults make behavior identical" holds only for users who never customized those fields. `lead_agent/prompt.py` now reads `max_injection_tokens` / `token_counting` from `DeerMemConfig` (`backend_config`), so a customized top-level value is lost without a trace.

Suggest: emit a warning on unrecognized legacy keys, auto-migrate the known legacy keys into `backend_config`, or set `model_config = ConfigDict(extra='forbid')` with a migration-pointing error. At minimum a CHANGELOG / RELEASE_NOTES migration note.

backend/app/gateway/routers/memory.py
```diff
@@ -115,27 +127,11 @@ class FactPatchRequest(BaseModel):
...
-    staleness_age_days: int = Field(..., description="Facts older than this many days are candidates for staleness review")
-    staleness_min_candidates: int = Field(..., description="Minimum stale facts required to trigger a review cycle")
-    staleness_max_removals_per_cycle: int = Field(..., description="Maximum number of facts staleness review can remove per cycle")
-    staleness_protected_categories: list[str] = Field(..., description="Fact categories exempt from staleness review")
+    enabled: bool = Field(..., description="Whether the memory mechanism is enabled (call-site gate).")
+    mode: Literal["middleware", "tool"] = Field(..., description="Memory operation mode: 'middleware' (passive per-turn LLM summarization) or 'tool' (model calls memory tools directly). Mechanism-level, applies to any backend.")
+    injection_enabled: bool = Field(..., description="Whether memory is injected into the system prompt (call-site gate).")
+    manager_class: str = Field(..., description="Active memory backend selector (backend name or dotted path).")
+    backend_config: dict = Field(..., description="Backend-private config (self-interpreted by the backend).")
```
willem-bd, line 134: This response shape is a breaking change vs. `main`: the flat fields (`storage_path`, `max_facts`, `debounce_seconds`, `token_counting`, `guaranteed_*`, `staleness_*`) are gone, replaced by `{enabled, mode, injection_enabled, manager_class, backend_config}`. The same shape change applies to `client.get_memory_config()`.

I verified the frontend does not consume these fields (no `.ts`/`.tsx` references), so there's no frontend runtime break. But the PR body's surface-area note "response shapes unchanged" is inaccurate for the config endpoint (it's true only for the memory-*data* responses). Any external API/SDK client reading the old flat fields breaks -- worth calling out explicitly as a breaking API change (CHANGELOG / versioning).

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
```diff
@@ -189,43 +162,36 @@ def save(self, memory_data: dict[str, Any], agent_name: str | None = None, *, us
...
+        module = importlib.import_module(module_path)
+        storage_class = getattr(module, class_name)
+
+        # Validate that the configured storage is a MemoryStorage implementation
+        if not isinstance(storage_class, type):
+            raise TypeError(f"Configured memory storage '{storage_class_path}' is not a class: {storage_class!r}")
+        if not issubclass(storage_class, MemoryStorage):
+            raise TypeError(f"Configured memory storage '{storage_class_path}' is not a subclass of MemoryStorage")
+
+        return storage_class(config)
```
willem-bd, line 193: Two breaking changes for anyone with a custom `storage_class`:

1. The old default path `deerflow.agents.memory.storage.FileMemoryStorage` no longer exists (module moved here). Configs referencing it fail `importlib` and silently fall back to `FileMemoryStorage` (logged error).
2. `create_storage` now calls `storage_class(config)` (one arg); pre-existing custom `MemoryStorage` subclasses had a no-arg `__init__` -> `TypeError` -> silent fallback to `FileMemoryStorage`.

Both degrade to `FileMemoryStorage` (logged), so it won't crash, but a custom-storage user would see their backend silently replaced. Worth a migration note (update path + accept `config` in `__init__`).

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/paths.py
```diff
@@ -0,0 +1,95 @@
...
+) -> Path:
+    """Resolve the memory file path under DeerMem's own data root.
+
+    ``config.storage_path`` (absolute or relative) is the root; per-user /
+    per-agent / legacy layouts live under it. Empty -> default root
+    (``$DEERMEM_DATA_DIR`` / ``~/.deermem/``). The host (deer-flow factory)
+    injects an absolute base_dir as ``storage_path`` so memory lands at
+    ``{base_dir}/users/{user_id}/memory.json`` (CWD-independent).
+    """
+    root = Path(config.storage_path) if config.storage_path else _default_root()
```
willem-bd, line 83: Relative-path semantics shifted from `main`. Old: a relative `storage_path` was resolved against `base_dir`, and only used for the legacy no-`user_id` case (ignored for per-user). New: `root = Path(config.storage_path)` is **CWD-relative** and used for ALL cases including per-user.

Low practical impact because the factory injects an absolute `runtime_home()` when `storage_path` is empty (so default-config users are unaffected -- I verified `runtime_home()` == `base_dir`, so per-user paths are preserved). But a user who sets a relative `backend_config.storage_path` now gets CWD-dependent behavior, which differs from the old `base_dir`-relative resolution and is fragile. Consider resolving relative paths against `runtime_home()`, or document the CWD dependence loudly.

backend/app/gateway/app.py
```diff
@@ -195,24 +195,26 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     # When memory.token_counting is "char", token counting never touches
```
willem-bd: This comment describes the old lifespan-level check (`if memory.token_counting == "char": skip`). The check has moved into `DeerMem.warm()` (which re-checks `token_counting == "char"` and returns early). The char fast-path is preserved -- just no longer here. Update the comment to point at `manager.warm` so a reader finds the skip logic.

backend/packages/harness/deerflow/agents/memory/tools.py
```diff
@@ -112,23 +113,32 @@ def memory_add_tool(
...
+        updated_memory = create(
             normalized_content,
             category=category,
             confidence=confidence,
             agent_name=agent_name,
             user_id=user_id,
         )
-        fact_id = created_fact["id"]
-        if all(fact.get("id") != fact_id for fact in updated_memory.get("facts", [])):
+        fact_id = next(
```
willem-bd: The tool re-derives `fact_id` by content-key matching instead of receiving it from `create_fact`. It works given the prior duplicate check, but couples the tool to the backend's content normalization (`_memory_content_key`). A future backend that normalizes content differently would make this re-derivation return `None` and falsely report "Fact was not stored because memory.max_facts kept higher-confidence facts". Consider having the `create_fact` capability return the created fact/id directly.

(Side nit just above: the empty-content branch returns `json.dumps({"error": "content"})` -- terse vs the descriptive `{"error": "Duplicate fact"}`; `{"error": "empty content"}` reads better.)

willem-bd (commented):
Reviewed PR head `bc8bf0d4`.

Strong architecture: clean 9-method ABC, fail-fast `manager_class` resolution (correct for persistent state), double-checked-locking singleton, and host-default hooks injected at the factory seam so the backend stays portable. The noop backend doubles as a well-documented template.

Verified the highest-risk concern and it holds up: per-user memory paths are NOT orphaned. `safe_user_id` (`deermem/core/paths.py`) is byte-identical to host `make_safe_user_id` (same regexes, sha256, 16-hex len); `runtime_home()` == `base_dir` (host `_default_local_base_dir` returns `runtime_home()`); agent-name lowercasing matches. So `{runtime_home}/users/{uid}/memory.json` is unchanged. `warm()` also preserves the `token_counting=="char"` fast-path, and no stale imports of the moved modules remain outside `backends/deermem/`.

Inline findings below. The blocking one is the silent drop of legacy top-level `memory.*` fields -- for a `risk:high` PR touching persistent state, silently reverting customized settings on upgrade is the kind of regression that erodes trust. The response-shape and custom-storage changes are breaking and worth documenting explicitly. The rest are nits.

Two nits not posted inline: (a) `reload_memory` falls back to `get_memory` (silent) when unsupported while `create/delete/update_fact` raise 501 -- defensible but worth a comment explaining the asymmetry; (b) no regression test pins the per-user memory path across the abstraction -- given `risk:high` + persistent state, that lock-in would be valuable.

qin-chenghan:
@WillemJiang  All review points from the second round (head `bc8bf0d4`) have been addressed in commit `54d3747a`. Summary:

**Blocking:**
- Legacy field silent drop: `load_memory_config_from_dict` now auto-migrates pre-abstraction top-level `memory.*` fields (storage_path, max_facts, debounce_seconds, model_name, token_counting, staleness_*, consolidation_*) into `backend_config` with a warning. `model_name` maps to `backend_config.model.model`. Unknown keys are warned and ignored.

**Breaking changes (documented):**
- API response shape: CHANGELOG.md documents the `/memory/config` + `client.get_memory_config()` shape change (flat → `backend_config` dict). Frontend audited -- does not consume these fields.
- Custom `storage_class`: CHANGELOG.md documents the import path move and `__init__(config)` signature change.

**Nits fixed:**
- Relative `storage_path` now resolves against `runtime_home()` (base_dir-relative, CWD-independent) at the factory layer; `paths.py` stays portable.
- `app.py` comment updated to point at `manager.warm()` for the `token_counting == "char"` fast-path.
- `memory_add` tool uses the `fact_id` returned directly by `create_fact` (now returns `(memory_data, fact_id)`) instead of re-deriving via content-key matching. `{"error":"content"}` → `{"error":"empty content"}`.
- Comment added explaining `reload_memory` silent fallback vs fact-CRUD 501 asymmetry (read-only degrade vs write fail-loud).

**Extra:**
- Regression test pinning the per-user memory path (`{storage_path}/users/{safe_user_id}/memory.json` == host `make_safe_user_id`) across the abstraction.

Lint + 129 memory tests passed. Please re-review.

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -193,6 +112,50 @@ def set_memory_config(config: MemoryConfig) -> None:
...
+            if key == "model_name":
+                # old top-level model_name -> backend_config.model.model
+                model_cfg = dict(backend_config.get("model") or {})
+                if "model" not in model_cfg:
+                    model_cfg["model"] = value
+                    backend_config["model"] = model_cfg
+                    migrated.append(f"{key} -> backend_config.model.model")
+            elif key not in backend_config:
+                # don't override an explicit backend_config value
+                backend_config[key] = value
```
willem-bd, line 165: 🟥 `storage_path` semantics changed (file → directory); legacy migration silently relocates per-user memory

Pre-abstraction, `storage_path` was a *file path* in several cases: an absolute path was the shared memory file (opting out of per-user isolation), and a relative path with no `user_id` was the global file under `base_dir`. In the new `memory_file_path` (`backends/deermem/deermem/core/paths.py`), `storage_path` is **always a root directory** — per-user memory always lands at `root/users/{uid}/memory.json`.

The old `config.example.yaml` default was `storage_path: memory.json`. This migration carries that value **verbatim** into `backend_config.storage_path`, and the factory then resolves the relative `memory.json` against `runtime_home()` → `{base_dir}/memory.json` (now treated as a directory). For an upgrader keeping the old default, per-user memory moves:

- before: `{base_dir}/users/{uid}/memory.json`
- after:  `{base_dir}/memory.json/users/{uid}/memory.json`

…so existing per-user memory is **orphaned** (invisible at the new path). If a global `{base_dir}/memory.json` file already exists, `FileMemoryStorage.save`'s `mkdir(parents=True, exist_ok=True)` raises `NotADirectoryError` (caught as `OSError`, logged, returns `False`) and **every memory write silently fails**.

This isn't called out in the CHANGELOG (which covers the config-shape and `storage_class` changes but not the `storage_path` semantic flip), and the new `config.example.yaml` comment `absolute path opts out of per-user isolation` is now stale (absolute paths no longer opt out — they're just a root dir with per-user isolation under it).

Suggested: detect a legacy `storage_path` that looks like a file (e.g. ends in `.json`) during migration and clear/warn; or error in the factory when the resolved `storage_path` is an existing file; and document the semantic change in the CHANGELOG + fix the config comment. Also worth adding a test pinning the **empty-default** path (factory-injected `runtime_home()`), since `test_per_user_memory_path_matches_host_safe_user_id` only covers an explicit `storage_path`.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -689,22 +475,120 @@ def _escape_memory_for_prompt(memory: Any) -> Any:
...
+            {
+                "id": fact_id,
+                "content": normalized_content,
+                "category": normalized_category,
+                "confidence": validated_confidence,
+                "createdAt": now,
+                "source": "manual",
+            }
+        )
+        updated_memory["facts"] = facts
```
willem-bd: 🟧 `create_memory_fact` no longer enforces `max_facts`

The old `create_memory_fact_with_created_fact` applied `_trim_facts_to_max(facts)` after appending; the new instance method sets `updated_memory["facts"] = facts` with no trim. So `POST /memory/facts`, the `memory_add` tool, and `client.create_memory_fact` now always store the fact, growing memory beyond `max_facts`. The cap is enforced only on the LLM `_apply_updates` path (the inline trim ~line 1008). `max_facts`'s description ("Maximum number of facts to store") no longer holds for the manual path. If uncapped manual facts are intended, the field description should say so; otherwise restore the trim here.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -1088,12 +1005,19 @@ def _apply_updates(
                 if fact_key is not None:
                     existing_fact_keys.add(fact_key)
 
-        current_memory["facts"] = _trim_facts_to_max(current_memory["facts"])
+        # Enforce max facts limit
+        if len(current_memory["facts"]) > config.max_facts:
+            # Sort by confidence and keep top ones
+            current_memory["facts"] = sorted(
+                current_memory["facts"],
+                key=lambda f: f.get("confidence", 0),
```
willem-bd: 🟧 max_facts trim lost `_coerce_source_confidence` robustness

The old `_trim_facts_to_max` sorted by `_coerce_source_confidence` (clamps to [0,1], defaults non-finite/missing to 0.5). This inline trim uses raw `f.get("confidence", 0)`. For a fact with a **string** confidence (which can enter via `import_memory`/legacy files — import bypasses `_validate_confidence`), `sorted` comparing `str` vs `float` raises `TypeError`, caught by the outer `except Exception` → the whole memory update silently fails (returns `False`, logged) for that user until the bad data is fixed. `NaN` confidence also yields non-deterministic ordering. The search path still uses `_coerce_source_confidence` (safe), but the deleted test `test_search_memory_facts_sort_survives_non_float_stored_confidence` wasn't re-added against `DeerMem.search`, so the regression guard is gone. Reuse `_coerce_source_confidence` here (and re-add the non-float-confidence test against `DeerMem.search`).

backend/packages/harness/deerflow/agents/memory/manager.py
```diff
@@ -0,0 +1,442 @@
...
+    name mismatch that previously made memory LLM tracing silently TypeError
+    is bridged here, at the host seam, so the portable package is untouched.
+    """
+    from deerflow.tracing import inject_langfuse_metadata
+
+    inject_langfuse_metadata(
+        invoke_config,
+        thread_id=thread_id,
+        user_id=user_id,
+        assistant_id="lead-agent",
```
willem-bd: 🟨 `assistant_id` for memory-update tracing changed `memory_agent` → `lead-agent`

Pre-abstraction, the memory-update LLM call passed `assistant_id="memory_agent"` to `inject_langfuse_metadata`. The host default here passes `assistant_id="lead-agent"`, which changes langfuse attribution for memory-update LLM calls and contradicts the PR's "behaviour matches the pre-abstraction code" note. If this is a copy-paste from the lead-agent injection path, restore `"memory_agent"`; if intentional, note it.

willem-bd (commented):
Reviewing the pluggable memory abstraction. The architecture, DI, portability story, and test discipline are excellent. Posting inline findings, ordered by severity:

1. 🟥 `storage_path` semantic change + legacy migration silently relocates/breaks per-user memory on upgrade (HIGH)
2. 🟧 manual `create_memory_fact` no longer enforces `max_facts`
3. 🟧 `max_facts` trim lost `_coerce_source_confidence` (can TypeError on imported non-float confidence)
4. 🟨 memory-tracing `assistant_id` changed `memory_agent` → `lead-agent`
5. 🟨 duplicated `human_input_response` parsing (drift risk)
6. 🟦 nit: empty-string legacy `storage_path` triggers a spurious migration warning

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/message_processing.py
```diff
@@ -55,8 +54,54 @@ def extract_message_text(message: Any) -> str:
...
 
 
-def filter_messages_for_memory(messages: list[Any]) -> list[Any]:
-    """Keep only user inputs and final assistant responses for memory updates."""
+def _non_empty_str(value: object) -> str | None:
+    """Return ``value`` if it is a non-empty (stripped) string, else None."""
+    return value if isinstance(value, str) and value.strip() else None
+
+
+def _is_human_clarification_response(additional_kwargs: Any) -> bool:
```
willem-bd, line 62: 🟨 Duplicated `human_input_response` parsing — drift risk

`_is_human_clarification_response` is a host-agnostic structural re-implementation of `read_human_input_response`, with a "must keep in sync" note. A format change on one side silently changes which hidden messages reach memory. The host-injected `should_keep_hidden_message` is already the sole source of truth in production; consider reducing this standalone mirror to a minimal fallback (or extracting the parsing to one shared place) so the two can't diverge.

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -193,6 +112,50 @@ def set_memory_config(config: MemoryConfig) -> None:
...
     global _memory_config
+    config_dict = dict(config_dict or {})
+    backend_config = dict(config_dict.get("backend_config") or {})
+    migrated: list[str] = []
+    for key in list(config_dict.keys()):
+        if key in _SHARED_FIELDS:
+            continue
+        if key in _LEGACY_DEERMEM_FIELDS:
+            value = config_dict.pop(key)
+            if value is None:
```
willem-bd: 🟦 nit: empty-string legacy values still trigger a migration warning

`storage_path: ""` (the documented default) is falsy-but-not-`None`, so it passes the `if value is None` skip and emits a "Migrated legacy fields" warning even though nothing meaningful migrated. Consider `if not value: continue` (or an explicit `value == ""` check) to silence the noise for default values.

lllyfff:
@WillemJiang All points from the inline review have been addressed in `629af1c3`. Summary:                                         
                                                                                                                        
  Blocking:                                                                                                             
  - `max_facts` cap restored on manual fact creation. `create_memory_fact` now trims via `_trim_facts_to_max`           
  (highest-confidence wins; confidence read through `_coerce_source_confidence` so `null`/non-float legacy facts never  
  crash the sort). Returns `(memory_data, fact_id)` with `fact_id=None` when the cap evicts the just-added fact; the    
  `memory_add` tool, `client.create_memory_fact`, and `POST /memory/facts` all surface it (tool -> JSON error, client ->
  `ValueError`, router -> 409). Restores the cap + post-trim existence check from upstream's                            
  `create_memory_fact_with_created_fact`.                                                                               
  - `storage_path` file->dir flip guarded on upgrade: the factory raises `ValueError` when `storage_path` resolves to an
  existing file; legacy migration drops `.json`-style values with a warning instead of silently relocating per-user     
  memory.                                                                                                               
                                                                                                                        
  Breaking changes (documented):                                                                                        
  - `storage_path` semantic change (pre-abstraction FILE -> root DIRECTORY; per-user memory at                          
  `{storage_path}/users/{uid}/memory.json`) in `CHANGELOG.md`; `config.example.yaml` comment updated.                   
                                                                                                                        
  Nits fixed:                                                                                                           
  - memory-tracing `assistant_id` restored to `"memory_agent"` (+ `run_name`) - the `lead-agent` change was an          
  unintended regression.                                                                                                
  - `_is_human_clarification_response` mirror added to `message_processing`, pinned to the host's                       
  `read_human_input_response` via an 11-case cross-check test (closes the duplicated-parser drift risk).                
  - empty-string legacy `storage_path` no longer triggers a spurious migration warning (`if value is None or value ==   
  ""` - explicit `False` for `consolidation_enabled` etc. still migrates).                                              
                                                                                                                        
  Extra:                                                                                                                
  - Regression tests: per-user memory path (`{storage_path}/users/{safe_user_id}/memory.json` == host                   
  `make_safe_user_id`), `_trim_facts_to_max` non-float coercion, create-fact cap-eviction `None` sentinel, search       
  surviving non-float confidence, `_is_human_clarification_response` ↔ host cross-check, factory raising on file        
  `storage_path`, migration dropping `.json` legacy values, empty `storage_path` injecting `runtime_home`, tool         
  reporting not-stored on cap eviction.                                                                                 
                                                                                                                        
  ---                                                                                                                   
                                                                                                                        
  **Part 2 — additionally discovered in a separate pass (commit `cce3302d`)**                                           
                                                                                                                        
  Beyond the inline review, a separate self-audit found and fixed the following (not raised previously):                
                                                                                                                        
  Blocking:                                                                                                             
  - `create_storage` is now fail-fast: a misspelled/unimportable `storage_class` raises `ValueError` instead of silently
  falling back to `FileMemoryStorage` (memory is persistent state - a wrong store is a data-integrity footgun); mirrors 
  the `manager_class` policy.                                                                                           
                                                                                                                        
  Nits / hardening:                                                                                                     
  - `noop` template `create_fact` signature corrected: positional `content` + `tuple[dict, str|None]` return (the       
  `memory_add` tool passes `content` positionally; gateway/client/tools all tuple-unpack) - a backend copied from the   
  old keyword-only/`dict` template would have 500'd on fact-CRUD. `delete_fact`/`update_fact` templates left (callers   
  compatible).                                                                                                          
  - `build_llm` graceful degrade: `init_chat_model` wrapped in try/except, degrades to `None` + WARNING on failure      
  (mirrors `_host_default_llm`) - a misconfigured explicit model no longer crashes app startup; non-LLM memory ops still
  work, an update raises at runtime with the error logged.                                                              
  - `DeerMemConfig.from_backend_config` now warns on unknown `backend_config` keys (mirrors the host layer's            
  `load_memory_config_from_dict`) so a typo like `storage_pat` doesn't silently fall back to the default.               
                                                                                                                        
  Verified, not bugs (no change):                                                                                       
  - `search()` category docstring is accurate - `category` filters inside the comprehension before sort + `[:top_k]`, so
  other-category facts cannot starve a category-scoped search (the proposed split-filter refactor is a no-op).          
  - `mkdir` `NotADirectoryError` debuggability: the upgrade path is already guarded by the factory's `is_file()` check +
  `.json` migration; residual error-message polish deferred.                                                            
                                                                                                                        
  Extra:                                                                                                                
  - Tests: `create_storage` raising on bad `storage_class` (3 cases, rewritten from the old fallback tests), `build_llm`
  zero-config-None + degrade-on-failure, `from_backend_config` warn/silent on unknown/known keys.                       
                                                                                                                        
  `make lint` green; 413 memory tests passed (3 known module-skips). Please re-review.

WillemJiang:
@qin-chenghan, please resolve the conflict with the main branch.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
qin-chenghan: Resolved. `_escape_memory_for_prompt` restored in `ef2c3349` (updater.py:463). The `current_memory` blob is escaped again at the call site (updater.py:690): `json.dumps(_escape_memory_for_prompt(current_memory), ...)`.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
qin-chenghan: Resolved. `html.escape` restored on both `cat` and `content` in the staleness section (`ef2c3349`, updater.py:394/398). The consolidation section was also escaped for consistency (updater.py:456/459).

config.example.yaml
qin-chenghan: Resolved. Factory now injects the host's default chat model into `backend_config.host_llm` when no explicit model is configured (`ef2c3349`, manager.py:343-359, `_host_default_llm()`). Zero-config memory extraction works out of the box again. PR description corrected.

backend/app/gateway/routers/memory.py
qin-chenghan: Resolved. The response shape change is intentional and now documented as a breaking change in CHANGELOG (`54d3747a` / `629af1c3`). Verified the frontend does not consume these fields. PR description updated.

backend/packages/harness/deerflow/agents/memory/manager.py
qin-chenghan: Resolved. `_resolve_manager_class` now raises `ValueError` on an unresolved explicit `manager_class` instead of silently falling back to DeerMem (`ef2c3349`, manager.py:278). Only the default (unset) `manager_class` resolves to DeerMem.

backend/packages/harness/deerflow/agents/memory/tools.py
qin-chenghan: Resolved. `category` is now on the ABC `search` signature and the backend filters BEFORE the `top_k` slice (`ef2c3349`, deer_mem.py:192-194). The docstring explicitly documents the ordering (deer_mem.py:183).

backend/packages/harness/deerflow/config/memory_config.py
qin-chenghan: Resolved. `_LEGACY_DEERMEM_FIELDS` auto-migration added in `54d3747a` (memory_config.py:27-166). Known legacy fields are migrated into `backend_config` with a warning. CHANGELOG documents the migration.

backend/app/gateway/routers/memory.py
qin-chenghan: Resolved. CHANGELOG explicitly calls out the breaking API change. PR description corrected. Frontend verified to not consume the old flat fields.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/storage.py
qin-chenghan: Resolved. CHANGELOG documents both breaking changes (old path + `__init__` signature). `create_storage` was also changed to fail-fast (`cce3302d`, storage.py:193) - a broken `storage_class` now raises `ValueError` instead of silently falling back.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/paths.py
qin-chenghan: Resolved. The factory injects `runtime_home()` (=`base_dir`) when `storage_path` is empty, so default-config users get the same behavior. For explicit relative paths, the CWD dependence is documented in `paths.py:77` and `config.example.yaml` comments.

backend/app/gateway/app.py
qin-chenghan: Resolved. Comment updated in `4f884567` (app.py:206-208) to point at `manager.warm` instead of describing the old `if memory.token_counting == 'char'` check.

backend/packages/harness/deerflow/agents/memory/tools.py
qin-chenghan: Resolved. `create_fact` now returns `tuple[dict, str | None]` directly (`54d3747a`, deer_mem.py:273), so the tool receives `fact_id` from the return value instead of re-deriving it via content-key matching. The empty-content error message was also improved.

backend/packages/harness/deerflow/config/memory_config.py
qin-chenghan: Resolved. Legacy migration now detects `storage_path` values ending in `.json` and drops them with a warning (`629af1c3`, memory_config.py:145-157). CHANGELOG documents the semantic change. `config.example.yaml` comment updated.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
qin-chenghan: Resolved. `_trim_facts_to_max` is now called after `create_memory_fact` appends (`629af1c3`, updater.py:570). Both the CRUD path and the LLM extraction path enforce `max_facts` via the same shared helper.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
qin-chenghan: Resolved. `_trim_facts_to_max` (updater.py:73-86) now uses `_coerce_source_confidence` for sorting, shared by both the CRUD path (L570) and the LLM extraction path (L1036). String/NaN confidence values are safely coerced to 0.5.

backend/packages/harness/deerflow/agents/memory/manager.py
qin-chenghan: Resolved. `assistant_id="memory_agent"` restored (`ef2c3349`, manager.py:322). The copy-paste from the lead-agent injection path was corrected.

backend/packages/harness/deerflow/config/memory_config.py
qin-chenghan: Resolved. Migration loop now skips empty strings (`629af1c3`, memory_config.py:136): `if value is None or value == "": continue`. Default-value configs no longer emit spurious migration warnings.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/message_processing.py
qin-chenghan: Resolved. An 11-case cross-check test (`test_is_human_clarification_response_matches_host_read` in `test_deermem_self_contained.py:311`) now pins the standalone mirror against the host's `read_human_input_response` so the two cannot silently diverge. Production still uses the host-injected `should_keep_hidden_message` as the sole source of truth; the mirror only runs as a standalone fallback (tests, no-host scenarios).

WillemJiang (approved):

PR #4124, opened 2026-07-12
fix(mcp): invalidate tools cache on config content + path, not just newer mtime

DaoyuanLi2816:
## Summary

`packages/harness/deerflow/mcp/cache.py` invalidates the cached MCP tools using **only** the `extensions_config.json` mtime with a strict `>` comparison, and stores no resolved config path. That is strictly weaker than the sibling cache for the *other* runtime-editable config file — `config/app_config.py::get_app_config()`, which reloads when the resolved path **or** a `(mtime, size, sha256)` content signature differs, built precisely because "mtime can remain stale ... on object-store or network mounts". This PR brings MCP cache invalidation to parity with that pattern.

It also folds in a one-line incidental `AGENTS.md` doc-sync (see §4).

## 1. The bug: mtime-only, strict `>`, no path

`_is_cache_stale()` records only `_config_mtime` at init and returns stale only when `current_mtime > _config_mtime`. Three real edit patterns defeat it:

| # | Scenario | Why the strict `>` (or absence of a path) misses it |
|---|----------|------------------------------------------------------|
| 1 | Content changes, mtime unchanged | same-second edit, or an object-store / network mount that does not bump mtime → `current == recorded` → `>` never fires |
| 2 | Content changes, mtime moves backward | `git checkout`, `cp -p` / backup restore, `tar`/`rsync` preserving timestamps → `current < recorded` → `>` never fires |
| 3 | Resolved path switches to a different file with mtime ≤ recorded | no path is tracked at all → structurally invisible |

## 2. Impact: multi-worker / stale-mtime deployments

`PUT /api/mcp/config` (`app/gateway/routers/mcp.py`) calls `reset_mcp_tools_cache()`, but that only resets the cache **in the worker that served the request**. Other uvicorn/gunicorn workers and the LangGraph-embedded runtime rely solely on this mtime check to notice the change, so under any of the three patterns above, adding or removing an MCP server has no effect until a restart. The module's own docstring states the mtime check exists so "changes made through the Gateway API are reflected in the Gateway-embedded LangGraph runtime" — the three cases above silently break that contract.

## 3. The fix: resolved path + content signature (mirrors `app_config`)

Record the resolved config path **and** a `(mtime, size, sha256)` signature at `initialize_mcp_tools()` time; in `_is_cache_stale()` recompute both and return stale when the resolved path **or** the signature differs (`!=`, not `>`). This mirrors `app_config`'s `_get_config_signature()` / `get_app_config()` so the two runtime-editable config files now share one content-based staleness signal.

- The per-call `stat` was already being done, so adding `size` is free, and the small-file sha256 matches the cost `app_config` already pays on every request.
- The existing "config missing / not yet initialized → not stale" no-op behavior is preserved (the extensions config is optional; `resolve_config_path()` may return `None`).
- `POST /api/mcp/cache/reset` and the `PUT /api/mcp/config` reset path keep working — `reset_mcp_tools_cache()` now clears the path + signature.

**Tests** — new `backend/tests/test_mcp_cache.py` (snapshots and restores the `deerflow.mcp.cache` module globals so state does not leak between tests):

| Test | Asserts |
|------|---------|
| `test_same_mtime_content_change_is_stale` | failure mode 1 → `True` |
| `test_backward_mtime_content_change_is_stale` | failure mode 2 → `True` |
| `test_config_path_switch_is_stale` | failure mode 3 → `True` |
| `test_unchanged_file_is_not_stale` | no needless reinit → `False` |
| `test_forward_edit_is_stale` | genuine forward edit still → `True` |
| `test_initialize_records_path_and_signature` | init records path + full signature |
| `test_not_stale_before_initialization` | uninitialized cache → `False` |

The three failure-mode tests fail on the pre-fix code (`_is_cache_stale()` returns `False`) and pass afterwards.

## 4. Incidental doc-sync (`backend/AGENTS.md`)

While updating the MCP cache-invalidation description, I also corrected the "Config Hot-Reload Boundary" restart-required field list, which was missing `scheduler` and `run_ownership` — both present in `packages/harness/deerflow/config/reload_boundary.py::STARTUP_ONLY_FIELDS`. The prose now matches the code.

## Notes

Reproduced all three failure modes against `main` (each `_is_cache_stale()` returns `False`) and confirmed each returns `True` after the fix, with the unchanged-file and forward-edit sanity cases behaving correctly; also confirmed `app_config`'s signature detects the same-mtime content change (sanity that the direction is right). `ruff check` / `ruff format --check` are clean (line-length 240). The §4 change is docs-only; the code change is behavior-preserving for the config-missing and cache-reset paths.

### DIFF
--- a/backend/packages/harness/deerflow/mcp/cache.py
+++ b/backend/packages/harness/deerflow/mcp/cache.py
@@ -46,6 +46,12 @@ def _get_config_signature(config_path: Path) -> _ConfigSignature | None:
     except OSError:
         return None
 
+    # Always hash the full file here rather than short-circuiting when
+    # mtime/size already match a previously recorded signature: swapping in a
+    # different MCP server config of identical byte length within the same
+    # second leaves mtime *and* size unchanged, so only the sha256 catches
+    # that swap. Skipping the hash on an mtime/size match would reopen the
+    # narrow gap this signature was built to close.
     digest = hashlib.sha256()
     try:
         with config_path.open("rb") as f:
@@ -84,7 +90,14 @@ def _is_cache_stale() -> bool:
 
     # Preserve the original "config missing / not yet recorded" behavior: if
     # there was no readable config when the cache was populated, or there is
-    # none now, do not invalidate.
+    # none now, do not invalidate. This also covers the config being deleted
+    # entirely after a successful init (current_signature flips to None): the
+    # cache intentionally keeps serving its last-known-good MCP tools rather
+    # than invalidating into an unconfigured state, matching the pre-fix
+    # mtime-only contract (which also returned False once the file could no
+    # longer be stat-ed). Treat this as a deliberate fail-soft choice, not an
+    # oversight — a future change that wants "config deleted" to tear down
+    # MCP tools needs its own explicit signal here, not an inferred one.
     if _config_signature is None or current_signature is None:
         return False
 
--- a/backend/packages/harness/deerflow/mcp/cache.py
+++ b/backend/packages/harness/deerflow/mcp/cache.py
@@ -1,8 +1,9 @@
 """Cache for MCP tools to avoid repeated loading."""
 
 import asyncio
+import hashlib
 import logging
-import os
+from pathlib import Path
 
 from langchain_core.tools import BaseTool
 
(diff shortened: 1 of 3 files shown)


backend/packages/harness/deerflow/mcp/cache.py
```diff
@@ -11,43 +12,88 @@
...
-    from deerflow.config.extensions_config import ExtensionsConfig
+    try:
+        stat_result = config_path.stat()
+    except OSError:
+        return None
+
+    digest = hashlib.sha256()
+    try:
+        with config_path.open("rb") as f:
+            for chunk in iter(lambda: f.read(1024 * 1024), b""):
```
willem-bd, line 58: Minor note for future maintainers (not a blocker): since the signature already carries `mtime` and `size`, a future "optimize the hot path" pass might be tempted to skip the hash when both match the recorded values. Flagging here why that would reintroduce a (narrow) hole: swapping one MCP server for another of identical byte length on a same-second edit keeps `mtime` **and** `size` equal, so only the `sha256` catches it. The full-hash-on-every-call matches `app_config` and the file is small — this is the right call, just recording the reasoning so it doesn't get "optimized" away.

backend/packages/harness/deerflow/mcp/cache.py
```diff
@@ -11,43 +12,88 @@
...
 
-    current_mtime = _get_config_mtime()
+    current_path, current_signature = _current_config_state()
 
-    # If we couldn't get mtime before or now, assume not stale
-    if _config_mtime is None or current_mtime is None:
+    # Preserve the original "config missing / not yet recorded" behavior: if
+    # there was no readable config when the cache was populated, or there is
+    # none now, do not invalidate.
+    if _config_signature is None or current_signature is None:
```
willem-bd, line 101: Latent edge, preserved from the pre-fix code by design — surfacing for awareness, not asking for a change here: if the extensions config is *deleted* after init, `current_signature` becomes `None` and this returns `False` (not stale), so cached MCP tools keep serving even though the config is gone. This matches `app_config`'s `None`-handling and the old mtime-only behavior, so it's not a regression — but it's a subtle enough corner that it may warrant a one-line code comment if the contract is intentional.

willem-bd (commented):
Reviewed against the PR head (`a61eb6c`). Overall this is a solid, well-scoped fix — the new `_get_config_signature` faithfully mirrors `app_config.py:523-538`, the three failure modes (same-mtime / backward-mtime / path-switch) are real and now correctly detected, the "config missing → not stale" no-op and `reset_mcp_tools_cache()` paths are preserved, logging is improved (lazy `%`-format), and the regression tests are thorough (the per-test `asyncio.Lock` reset is a nice touch). The §4 `backend/AGENTS.md` doc-sync is accurate — verified that both `scheduler` and `run_ownership` are in `STARTUP_ONLY_FIELDS` on `main`.

The two inline notes below are minor awareness points, not blockers — recommend approve.

backend/packages/harness/deerflow/mcp/cache.py
```diff
@@ -11,43 +12,88 @@
...
+    from deerflow.config.extensions_config import ExtensionsConfig
 
-def _get_config_mtime() -> float | None:
-    """Get the modification time of the extensions config file.
+    return ExtensionsConfig.resolve_config_path()
 
-    Returns:
-        The modification time as a float, or None if the file doesn't exist.
+
+def _get_config_signature(config_path: Path) -> _ConfigSignature | None:
```
willem-bd, line 36: This `_get_config_signature` (plus the `_ConfigSignature` alias at line 24) is now a verbatim duplicate of `deerflow/config/app_config.py::_get_config_signature` / `_ConfigSignature`. Mirroring rather than refactoring is the right call for this focused behavior fix - but worth a follow-up to extract both into a small shared helper (e.g. under `deerflow/config/`) so the two runtime-editable config files don't drift apart over time. Non-blocking.

backend/packages/harness/deerflow/mcp/cache.py
```diff
@@ -11,43 +12,88 @@
...
     Returns:
         True if the cache should be invalidated, False otherwise.
     """
-    global _config_mtime
-
     if not _cache_initialized:
         return False  # Not initialized yet, not stale
 
-    current_mtime = _get_config_mtime()
+    current_path, current_signature = _current_config_state()
```
willem-bd, line 89: Pre-existing, out of scope, but flagging since this PR is about cache correctness: `_is_cache_stale()` - and the `reset_mcp_tools_cache()` it triggers - run from `get_cached_mcp_tools()` *outside* `_initialization_lock`, so a concurrent caller can reset / close sessions while another is mid-`initialize_mcp_tools()`. This PR widens that unlocked window slightly (a full file read + sha256 here vs. the old bare `stat()`). `close_all_sync()` is built for concurrent teardown and the file is small, so likely fine in practice - no change requested, just raising it in case you want to harden the call site separately.

willem-bd (commented):
Reviewed the MCP cache-invalidation fix - a clean, well-tested change that correctly mirrors `app_config.py`'s content-signature pattern, with the three failure modes and the config-missing / cache-reset paths all preserved. Two non-blocking observations inline.

WillemJiang:
@DaoyuanLi2816 Please take a look at the review comments.

DaoyuanLi2816:
Thanks @WillemJiang for the nudge, and @willem-bd for the two thorough passes — replying to each point.

**1. Hot-path hash short-circuit risk (`_get_config_signature`)**

Agreed this is worth guarding, not just noting. Pushed `2b15451`:
- Added a comment directly above the hash loop spelling out why the sha256 has to run unconditionally: an equal-length server-name swap inside the same second leaves mtime *and* size unchanged, so skipping the hash on an mtime/size match would silently reopen that gap.
- Added `test_same_mtime_same_size_swap_is_stale` — a sharper variant of the existing same-mtime test. The existing one (`test_same_mtime_content_change_is_stale`) happens to also change file size, so it wouldn't actually catch a future mtime+size shortcut; this one holds both constant via an equal-length key swap (`srv1` -> `srv9`) and asserts on the unchanged size too.
- Verified the new test isn't a rubber stamp: temporarily patched in the exact shortcut you described (skip to `return False` when `current_signature[:2] == _config_signature[:2]`) and confirmed the test fails against it, then reverted and confirmed it passes against real `main`.

**2. Deleted-config edge case (`_is_cache_stale`)**

Confirmed the contract is intentional and expanded the code comment to say so explicitly, plus why: preserved verbatim from the pre-fix mtime-only code (`_config_mtime is None or current_mtime is None: return False`), so keeping the cache serving on a deleted config isn't new behavior introduced by this fix. Added `test_config_deleted_after_init_is_not_stale` to pin it — same discrimination check as above (temporarily removed the None-guard, confirmed the test then fails with `_is_cache_stale()` returning `True`, reverted, confirmed it passes).

One layer deeper, though, and distinct from what you described: this only holds when `ExtensionsConfig.resolve_config_path()` resolves through the no-override search path. If the path came from an explicit `config_path` arg or `DEER_FLOW_EXTENSIONS_CONFIG_PATH` (which is exactly how Docker dev/prod point at it per AGENTS.md), `resolve_config_path()` itself raises `FileNotFoundError` on a deleted file rather than returning `None` — so `_is_cache_stale()` (and therefore `get_cached_mcp_tools()`) can raise before ever reaching the None-check we're discussing. Repro'd this directly against both `main` and this branch. It's pre-existing either way — `_get_config_mtime()` made the identical unguarded `resolve_config_path()` call before this PR, and `extensions_config.py` isn't touched by this diff — so I left it alone here rather than scope-creep a review-response commit into a second fix. Happy to file it separately if useful; flagging for visibility in the meantime.

**3. Duplicate `_get_config_signature` / `_ConfigSignature` vs. `app_config.py`**

Agreed on both counts — it is a verbatim duplicate, and mirroring rather than a shared helper was the right call for this focused fix. Leaving the extraction as the follow-up you suggested rather than pulling it into this branch.

**4. `_is_cache_stale()` running outside `_initialization_lock`**

Agreed this is real and pre-existing, and per your note not something this PR should take on. No change made here.

**Verification**: `test_mcp_cache.py` is 9/9 green (7 existing + 2 new), ruff check + format clean on both touched files. Pushed as `2b15451`.

WillemJiang (approved):

PR #4126, opened 2026-07-12
feat(frontend): show real project version on About page

zhfeng:
## Summary

The Settings ▸ About panel heading was a hardcoded `About DeerFlow 2.0` while the real project version is `2.1.0` (kept in lockstep across `backend/pyproject.toml`, `frontend/package.json`, and `Chart.yaml` by `scripts/verify_versions.sh`, gated on `v*` tag releases). With the nightly workflow shipping unreleased `main` daily, a hardcoded version drifts further from reality every day.

This wires the About heading to the real version, nightly-aware - matching the Helm chart's nightly scheme:

| Build | About heading |
| --- | --- |
| Release / local dev | `About DeerFlow 2.1.0` |
| Nightly | `About DeerFlow 2.1.0-nightly.20260712-abc1234` |

## Changes

- **`frontend/src/version.ts`** (new) - `APP_VERSION = process.env.NEXT_PUBLIC_APP_VERSION || pkg.version`. The `||` (not `??`) is deliberate: the Dockerfile sets the env to `""` for release/local builds, and that empty string must fall through to `package.json`'s version.
- **`about-content.ts`** - heading interpolates `APP_VERSION`. The "DeerFlow 1.0 and 2.0" milestone copy in acknowledgments is left as-is (it refers to product generations, not the patch version).
- **`frontend/Dockerfile`** - `ARG APP_VERSION` + `ENV NEXT_PUBLIC_APP_VERSION` in the builder stage so nightly CI can stamp the version.
- **`.github/workflows/nightly.yaml`** - compute the nightly version string **once** in the `prepare` job (`<base>-nightly.<date>-<sha>`, mirroring the chart scheme) and feed it to both the frontend image build-arg and the chart publish. `publish-chart` was refactored to consume the shared output so the displayed version and the chart version can't drift apart.
- **`container.yaml` (release)** - untouched. Releases rely on the `package.json` fallback, which `verify-versions` guarantees equals the tag.
- Tests (`version.test.ts`, `about-content.test.ts`) + a `RELEASING.md` note.

## Out of scope

The landing-page strings `Get Started with 2.0` (`hero.tsx`) and `Whats New in DeerFlow 2.0` (`whats-new-section.tsx`) are **milestone/generation labels**, not build version stamps - their content is the 2.0 launch showcase, so parameterizing them to `APP_VERSION` would make the titles lie (a 2.1.0 release would read "What's New in DeerFlow 2.1.0" over 2.0-era content). Left as-is after discussion.

## Verification

- `pnpm test` - 618/618 pass (5 new).
- `pnpm check` (eslint + tsc) - clean.
- pre-commit hooks (eslint + prettier) pass.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -42,12 +42,30 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       date: ${{ steps.date.outputs.date }}
+      nightly_version: ${{ steps.nightly_version.outputs.nightly_version }}
     steps:
       - name: Set nightly date (UTC)
         id: date
         # Shared by build-images (image tag) and publish-chart (chart version)
         # so the two stay in sync for a given run.
         run: echo "date=$(date -u +%Y%m%d)" >> "$GITHUB_OUTPUT"
+      - name: Checkout repository
+        # Needed to read Chart.yaml's base version for the nightly version
+        # string computed below.
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
+      - name: Compute nightly version
+        id: nightly_version
+        # Single source of truth for the nightly version string:
+        # <base>-nightly.<YYYYMMDD>-<short_sha> (mirrors the chart's nightly
+        # scheme). build-images injects it into the frontend image (About-page
+        # version) and publish-chart stamps it on Chart.yaml, so the version
+        # users see and the chart version can't drift apart.
+        run: |
+          BASE=$(grep -m1 '^version:' deploy/helm/deer-flow/Chart.yaml | awk '{print $2}')
+          SHORT_SHA="${GITHUB_SHA::7}"
+          NIGHTLY="${BASE}-nightly.${{ steps.date.outputs.date }}-${SHORT_SHA}"
+          echo "nightly_version=${NIGHTLY}" >> "$GITHUB_OUTPUT"
+          echo "Nightly version: ${NIGHTLY}"
 
   validate-chart:
     if: github.repository == 'bytedance/deer-flow'
@@ -132,7 +150,13 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          build-args: ${{ matrix.build-args }}
+          # APP_VERSION is consumed only by the frontend Dockerfile (it stamps
+          # the About-page version); backend/provisioner Dockerfiles don't
+          # declare it. Passed via build-arg because build-push-action doesn't
+          # forward host env into the BuildKit build.
+          build-args: |
+            ${{ matrix.build-args }}
+            ${{ matrix.component == 'frontend' && format('APP_VERSION={0}', needs.prepare.outputs.nightly_version) || '' }}
       - name: Generate artifact attestation
         uses: actions/attest-build-provenance@e8998f949152b193b063cb0ec769d69d929409be #v2.4.0
         with:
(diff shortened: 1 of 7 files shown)


willem-bd (commented):
Reviewed the version wiring across all three build paths (local dev, tag-driven release, nightly) and verified the claims against `main`:

- `frontend/package.json` = `2.1.0`, `Chart.yaml` version/appVersion = `2.1.0` ✅
- Release `container.yaml` passes no `APP_VERSION` → relies on the `ARG APP_VERSION=""` default → `"" || pkg.version` → `2.1.0` ✅
- `verify_versions.sh` gates all four sources against the tag ✅
- `tsconfig.json` has `resolveJsonModule: true`, so `import pkg from "../package.json"` builds clean ✅
- `rs.resetModules()` matches the established Rstest idiom ✅

The core mechanism (`||` not `??`, build-arg override, single source of truth computed once in `prepare`) is sound, and the `||` vs `??` rationale is correctly documented at both the code and the Dockerfile. The nightly string `2.1.0-nightly.<date>-<sha>` is valid semver and identical to what `publish-chart` already produced — this PR just moves its computation, so there's no chart regression.

Two inline notes below (one real suggestion, one nit). One minor non-code observation: the `area:sandbox` / `risk:high` labels don't match the actual scope — this touches frontend version display + nightly CI versioning; no sandbox code is involved.

Overall: looks good. The one thing worth doing before merge is the `BASE`-empty guard below, since a malformed nightly version would now propagate to both the chart and the UI.

frontend/tests/unit/components/workspace/settings/about-content.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  // generations and must NOT be parameterized.
+  expect(aboutMarkdown).toContain("DeerFlow 1.0 and 2.0");
+});
+
+test("aboutMarkdown heading reflects the package version when env is unset", async () => {
+  delete process.env.NEXT_PUBLIC_APP_VERSION;
+  const { aboutMarkdown } =
+    await import("@/components/workspace/settings/about-content");
+  // The stale hardcoded "2.0" must be gone - the heading uses APP_VERSION.
+  expect(aboutMarkdown).not.toContain("About DeerFlow 2.0]");
```
willem-bd: This assertion is negative-only: it verifies the stale `2.0` literal is gone, but would still pass if `APP_VERSION` interpolated to an empty string (`About DeerFlow ]`) or `undefined`. Consider also asserting the positive value, mirroring `version.test.ts`:

```ts
test("aboutMarkdown heading reflects the package version when env is unset", async () => {
  delete process.env.NEXT_PUBLIC_APP_VERSION;
  const { APP_VERSION } = await import("@/version");
  const { aboutMarkdown } =
    await import("@/components/workspace/settings/about-content");
  expect(aboutMarkdown).toContain(`[About DeerFlow ${APP_VERSION}]`);
});
```

Both modules are re-evaluated after `resetModules()` + the env change, so `APP_VERSION` reflects the deleted-env fallback (`pkg.version`) and ties the heading to the real value — catching interpolation regressions, not just removal of the old literal.

.github/workflows/nightly.yaml
```diff
@@ -132,7 +150,13 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          build-args: ${{ matrix.build-args }}
+          # APP_VERSION is consumed only by the frontend Dockerfile (it stamps
+          # the About-page version); backend/provisioner Dockerfiles don't
+          # declare it. Passed via build-arg because build-push-action doesn't
+          # forward host env into the BuildKit build.
+          build-args: |
```
willem-bd, line 164: Nit: for the frontend matrix entry, `matrix.build-args` is `""`, so this block emits a leading blank line; for backend/provisioner the `|| ''` emits a trailing blank line. `docker/build-push-action` skips blank lines, so it's safe in practice — just slightly noisy. Fine to leave as-is.

willem-bd (commented):
Reviewed the diff against the PR head (`a31acbbb`). Clean, well-scoped change — the `||`-vs-`??` fallback is correct and well-tested, and consolidating the nightly version string into the `prepare` job output (so the About-page version and the chart version can't drift) is a nice refactor. No blocking issues.

Two inline notes below: one test-quality suggestion and one CI watch-item. (The `prepare`-job checkout tradeoff is fine as-is — not flagging it inline.)

.github/workflows/nightly.yaml
```diff
@@ -42,12 +42,30 @@ jobs:
...
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
+      - name: Compute nightly version
+        id: nightly_version
+        # Single source of truth for the nightly version string:
+        # <base>-nightly.<YYYYMMDD>-<short_sha> (mirrors the chart's nightly
+        # scheme). build-images injects it into the frontend image (About-page
+        # version) and publish-chart stamps it on Chart.yaml, so the version
+        # users see and the chart version can't drift apart.
+        run: |
+          BASE=$(grep -m1 '^version:' deploy/helm/deer-flow/Chart.yaml | awk '{print $2}')
```
willem-bd, line 65: If `BASE` ever comes back empty (e.g. `Chart.yaml` format drift), `NIGHTLY` silently becomes `-nightly.<date>-<sha>` — invalid semver — and that now flows into **both** the chart publish *and* the frontend About-page version.

The same `grep | awk` was already in `publish-chart` before this PR, so this isn't a regression for the chart path — but the blast radius now includes the UI. A cheap hardening, mirroring the `set -eu` already used in `publish-chart`:

```yaml
set -euo pipefail
BASE=$(grep -m1 '^version:' deploy/helm/deer-flow/Chart.yaml | awk '{print $2}')
test -n "$BASE" || { echo "::error::empty base version from Chart.yaml"; exit 1; }
```

`pipefail` matters here — without it the pipeline's exit code is `awk`'s, so `set -e` alone won't catch a `grep` miss.

willem-bd (commented):
One additional `nightly.yaml` note (companion to the review above): a hardening suggestion for the `prepare` step's version computation.

frontend/tests/unit/components/workspace/settings/about-content.test.ts
zhfeng: Done in 655e736a — the unset-env test now imports `APP_VERSION` and asserts the heading contains `` `[About DeerFlow ${APP_VERSION}]` ``, mirroring `version.test.ts`. Dropped the negative `2.0]` check since the positive assertion subsumes it (only one `About DeerFlow` occurrence in the heading).

.github/workflows/nightly.yaml
zhfeng: Done in f96f6d40 - added `set -euo pipefail` + a `test -n "$BASE"` guard to the prepare step, mirroring publish-chart. Good catch on `pipefail` - without it the pipeline exit code is awk's, so `set -e` alone would have missed a `grep` miss.

.github/workflows/nightly.yaml
zhfeng: Leaving as-is per your call. Confirmed `docker/build-push-action` skips the blank lines, so the noise is cosmetic only.

WillemJiang (approved):

Issue #4063, opened 2026-07-11
[RFC] Pluggable fine-grained authorization system


zhfeng:
Opening this as a dedicated RFC issue (per @WillemJiang's request on #3462 for an RFC before implementation) so the design has its own thread for feedback.

- **Responds to** #3462 (fine-grained user permission management).
- **Consolidates** @hata33's Plan A / Plan B proposals (#3462 comments; related gradient proposal in #4032) and the maintainer's guidance to integrate tool/resource authorization with the existing `GuardrailProvider`.
- **Fills a roadmap gap**: RBAC is a top-priority unowned item under "Security and Permission Strengthening" in #1669, with no current implementation.

The same content was posted as a comment on #3462; this issue is the canonical home for discussion. @hata33 has endorsed the direction (see #3462 comments) and offered to collaborate on Phase 1.

---

# RFC: Pluggable Fine-Grained Authorization

**Status:** Draft for feedback (responds to [#3462](https://github.com/bytedance/deer-flow/issues/3462)).
**Affects:** `backend/packages/harness/deerflow/authz/` (new), `backend/app/gateway/authz.py`, `backend/packages/harness/deerflow/guardrails/`, `backend/packages/harness/deerflow/agents/lead_agent/agent.py`, `backend/packages/harness/deerflow/client.py`, `backend/packages/harness/deerflow/subagents/executor.py`, `backend/app/gateway/services.py`, `config.example.yaml`.

> Maintainer guidance from the issue ([@WillemJiang](https://github.com/bytedance/deer-flow/issues/3462)):
> *"在开始实现之前，我们需要梳理相关的资源与权限的管理映射关系。具体工具调用和资源访问这块可以结合现有的 GuardrailProvider 来进行整合。另外建议在实现之前，做一个 RFC 的设计，方便大家的提反馈建议。"*
>
> This RFC does exactly that: (1) maps the resource × permission space, (2) integrates with `GuardrailProvider`, (3) is an RFC for feedback before any implementation.

## TL;DR

DeerFlow's authorization today is **authentication + ownership**: every authenticated user gets every API permission (`authz.py:144` hands all users `_ALL_PERMISSIONS`), and the only real authorization beyond login is thread ownership (`owner_check`) and a handful of hard-coded `require_admin_user()` routes. There is **no resource-level authorization** — no role can be told "you may not call `write_file`", "you may not use model X", or "you may not run sandbox code".

[...]

Issue #3462, opened 2026-06-09
[feat] 支持细粒度用户权限管理


wkk0611:
### Before you start

- [x] I searched [existing issues](https://github.com/bytedance/deer-flow/issues?q=is%3Aissue) and this is not a duplicate.

### Problem / motivation

DeerFlow 2.0 目前只有 admin 和 user 两种角色，但实际上两者权限完全相同（authz.py 中所有已认证用户都获得全部权限）。

    企业落地场景需要：
    - 自定义角色（访客、普通用户、管理员等）
    - 按角色配置功能权限（可用模型、功能模块、资源配额等）
    - 用户启用/禁用、批量管理

    请问官方有计划支持更完善的权限管理吗？

### Proposed solution

无

### Affected area(s)

Not sure

WillemJiang:
企业级的用户管理系统是比较复杂的，后续我们会抽象一下通用的接口方便大家二次开发。
由于人手关系，我们不会专门投入人力做细粒度的权限管理。

hata33:
@WillemJiang 您提到「抽象通用接口方便二次开发」,我想接手实现资源级权限这一层。核心问题是**粒度做到多深**,列两个方案,想听您定夺。

要补的空缺:现有 `authz.py` 只管 API 级(`threads:read`),但企业最常问的是"某个角色能不能用 `write_file` / 某个 MCP server / 某个模型"——资源级目前没有。

## 方案 A:最小过滤 hook(轻量,可扩展)

不内置 RBAC。只暴露一个可插拔点,让企业接自己的权限体系:

```python
class ResourcePermissionProvider(Protocol):
    def filter_tools(self, role: str, tools: list) -> list: ...
    # skills / mcp / models 同形
```

仓库里给一个读 `config.yaml` 的**示例实现**(白/黑名单),方便用户 fork 改。bytedance 不背完整权限模型的长期维护成本。

- **优**:改动面小,契合"抽象接口"的原意,企业自由度高
- **劣**:用户开箱不能用,得自己写 provider;示例实现覆盖不了复杂场景

## 方案 B:内置资源级 RBAC(完整,开箱即用)

config 里直接配角色 → 资源白/黑名单,覆盖 tools / skills / mcp / models / sandbox 五类:

```yaml
permissions:
  roles:
    guest:
      tools: {allow: [web_search]}
      sandbox: {enabled: false}
    admin: {}  # 全开
  default_role: user
```

[...]

Issue #1669, opened 2026-03-31
RoadMap of DeerFlow Q2


WillemJiang:
## Core Stability 🔥🔥🔥🔥🔥
Objective: Ensure users can use the product stably and smoothly, and continuously fix issues submitted by users

- [ ] Regular release plan 
- [ ] Package extensions
- [ ] Sophisticated unit tests / integration tests / e2e tests
- [ ] Benchmark

## Performance and Experience Optimization 🔥🔥🔥🔥
Goal: Ensure users experience no waiting or lag
- [ ] Support concurrent user requests. Continue to improve the relevant interfaces of `/api/langgraph-compat` and support capabilities such as `stream join`.
   - #1851 
- [ ] FrontEnd performance improvement

## Channel and Integration Expansion 🔥🔥🔥
Goal: Make Agent ubiquitous
- [ ] Implement the built-in Cron scheduler
  - #1092
- [ ] LLM Provider extensions
  -  #1821
  -  #1860 
- [ ] new channels (Matrix, DingTalk, JIRA, WeCom) 
   - #1869
   - #1806
   - #1802
   - #1514
   - #1390 

## Security and Permission Strengthening 🔥🔥🔥🔥
Objective: Meet enterprise-level security compliance requirements
- [ ] Implement Role-Based Access Control (RBAC) 
  - #1721 
- [ ] Improve the sandbox security mechanism
  - #1808
  - #1881
- [ ] MCP and Tools calling auditing
  - #1240

## Hierarchical Memory Architecture 🔥🔥🔥🔥🔥
Objective: Enable the Agent to remember long-term conversations and become smarter as the conversation progresses

RFCs:
- #1590 
- #1620

## Skills and Extensions 🔥🔥🔥🔥🔥
Objective: Provide a way for the user to extend the functions of DeerFlow to solve their daily problem

- [ ] auditing customer skills 
- [ ] Skills Hub?
- RFCs
  - #1865 
- Skills implementations:
  - #1861   

## Documentation and Developer Experience 🔥🔥🔥🔥
Objective: Enable new users to get started within 30 minutes and reduce the number of issues by 40%

- [ ] Document website
- [ ] User guide
- [ ] Video Tutorials

fanmindnn:
尽快啊大佬

LittleChenLiya:
越来越好了，美滋滋

MarkHoch:
[...]

PR #4127, opened 2026-07-12
feat(authz): add pluggable AuthorizationProvider protocol and config scaffolding (Phase 0)

hata33:
## Summary

Phase 0 of the RFC in #4063 — **scaffolding only, zero behavior change** (`enabled: false` by default).

Adds a new `deerflow/authz/` package as a sibling to `deerflow/guardrails/`, containing the `AuthorizationProvider` Protocol and the adapter that bridges it to the existing `GuardrailMiddleware`. No runtime behavior changes — Phase 1 will wire it into the agent assembly pipeline.

## Why

#4063 proposes a pluggable `AuthorizationProvider` as the policy brain for resource-level authorization, enforced at two layers from one policy: assembly-time capability filtering (Layer 1) and run-time execution deny via `GuardrailMiddleware` (Layer 2). This PR delivers the Phase 0 foundations so that Phase 1 can build on them.

The `AuthorizationProvider` is deliberately a **sibling** to `GuardrailProvider`, not an extension — PR #3665 explicitly scoped guardrails to execution-time checks only ("不新增 policy engine、RBAC 系统"), deferring the RBAC brain that this protocol provides.

## What changed

**New files:**

| File | Purpose |
|------|---------|
| `deerflow/authz/provider.py` | `AuthorizationProvider` Protocol (`@runtime_checkable`) + `Principal`/`AuthzRequest`/`AuthzDecision`/`AuthzReason` dataclasses |
| `deerflow/authz/adapter.py` | `GuardrailAuthorizationAdapter`: maps `GuardrailRequest` → `AuthzRequest`, converts `AuthzDecision` → `GuardrailDecision`, so existing `GuardrailMiddleware` can enforce authz decisions without a new middleware |
| `deerflow/authz/__init__.py` | Public exports |
| `deerflow/config/authorization_config.py` | `AuthorizationConfig` + singleton load/reset (mirrors `GuardrailsConfig`) |
| `tests/test_authorization_provider.py` | 33 tests |

**Modified files:**

| File | Change |
|------|--------|
| `config/app_config.py` | +3 lines: import, field (`authorization: AuthorizationConfig`), and `load_authorization_config_from_dict` call |
| `config.example.yaml` | Added `authorization` section (`enabled: false`); bumped `config_version` to 25 |

## What is NOT in this PR (deferred to Phase 1)

- Layer 1 filter wiring (`agent.py`, `executor.py`, `client.py`)
- Layer 2 auto-wire in `_build_runtime_middlewares`
- `Principal` builder in `services.py`
- Built-in `RbacAuthorizationProvider`
- `Principal.is_internal` population (correct signal `auth_source` lives on `request.state`, not `GuardrailRequest`)

These are Phase 1 per the RFC and will be a separate PR.

## Design notes

- **Behavior identical at `enabled: false`**: the `AuthorizationConfig` defaults to disabled. No code path reads it yet (Phase 1 will). Existing behavior is 100% preserved.
- **Protocol shape mirrors `GuardrailProvider`**: `@runtime_checkable`, sync + async, `name` attribute, no base class. Class-path loaded via `resolve_variable()` (same as guardrails/models/sandbox).
- **`filter_resources` is a required method**: all providers must implement it. Providers without a static role→resource map should delegate to `authorize()` per item in their own implementation. The built-in RBAC provider (Phase 1) will override it for O(1) filtering. A provider omitting `filter_resources` fails `isinstance(..., AuthorizationProvider)` (verified by `test_provider_without_filter_resources_is_not_provider`).
- **`resource`/`action`/`target` are free-form strings** (not enums) so new resource types need no schema change.
- **`Principal.is_internal` is not populated by the adapter**: the correct signal (`auth_source == AUTH_SOURCE_INTERNAL`) lives on `request.state`, not on `GuardrailRequest`. Phase 1 will thread it into run context. See `.. note::` in the adapter docstring.
- **Adapter propagates provider exceptions**: fail-closed semantics are applied by `GuardrailMiddleware` (based on `fail_closed`), not the adapter. Both sync and async exception propagation are pinned by tests.

## Validation

```bash
# New tests — 33 passed
PYTHONPATH=. uv run pytest tests/test_authorization_provider.py -v
# → 33 passed

# Existing guardrail tests — 41 passed (no regression)
PYTHONPATH=. uv run pytest tests/test_guardrail_middleware.py -q
# → 41 passed

# Lint + format
uv run ruff check packages/harness/deerflow/authz/ packages/harness/deerflow/config/authorization_config.py tests/test_authorization_provider.py
# → All checks passed!
uv run ruff format --check packages/harness/deerflow/authz/ packages/harness/deerflow/config/authorization_config.py tests/test_authorization_provider.py
# → 5 files already formatted
```

## Related

- RFC: #4063
- Original issue: #3462
- Prior art: #3665 (guardrail scope statement), #1669 (RBAC roadmap item)

### DIFF
--- a/backend/packages/harness/deerflow/authz/adapter.py
+++ b/backend/packages/harness/deerflow/authz/adapter.py
@@ -1,0 +1,101 @@
+"""Adapter that presents an AuthorizationProvider as a GuardrailProvider.
+
+This lets the existing :class:`~deerflow.guardrails.middleware.GuardrailMiddleware`
+enforce :class:`~deerflow.authz.provider.AuthorizationProvider` decisions at
+tool-call time — no new middleware class required (see RFC §6.1).
+
+The adapter maps :class:`~deerflow.guardrails.provider.GuardrailRequest`
+fields to :class:`~deerflow.authz.provider.AuthzRequest` fields, calls the
+authorization provider, and converts the :class:`~deerflow.authz.provider.AuthzDecision`
+back to a :class:`~deerflow.guardrails.provider.GuardrailDecision`.
+"""
+
+from __future__ import annotations
+
+from deerflow.authz.provider import AuthorizationProvider, AuthzDecision, AuthzRequest, Principal
+from deerflow.guardrails.provider import GuardrailDecision, GuardrailReason, GuardrailRequest
+
+
+class GuardrailAuthorizationAdapter:
+    """Adapt an :class:`AuthorizationProvider` to the ``GuardrailProvider`` Protocol.
+
+    ``resource_type`` and ``action`` default to ``"tool"`` / ``"call"``,
+    which is correct for the tool-execution path. A different resource/action
+    pair can be injected if the adapter is reused outside the tool path.
+
+    .. note::
+
+        ``Principal.is_internal`` is not populated by this adapter — the
+        correct signal (``auth_source == AUTH_SOURCE_INTERNAL``) lives on
+        ``request.state``, not on :class:`GuardrailRequest`. The field
+        retains its dataclass default (``False``) until Phase 1 threads it
+        into run context, so both layers derive it from one source.
+    """
+
+    name = "authorization"
+
+    def __init__(
+        self,
+        provider: AuthorizationProvider,
+        *,
+        resource_type: str = "tool",
+        action: str = "call",
+    ) -> None:
+        self._provider = provider
+        self._resource_type = resource_type
+        self._action = action
+
+    def _to_authz(self, gr: GuardrailRequest) -> AuthzRequest:
+        """Map a guardrail request to an authorization request."""
+        return AuthzRequest(
+            principal=Principal(
+                user_id=gr.user_id,
+                role=gr.user_role,
+                oauth_provider=gr.oauth_provider,
+                oauth_id=gr.oauth_id,
+            ),
+            resource=self._resource_type,
+            action=self._action,
+            target=gr.tool_name,
+            context={
+                "thread_id": gr.thread_id,
+                "run_id": gr.run_id,
+                "tool_call_id": gr.tool_call_id,
+                "tool_input": gr.tool_input,
+                "is_subagent": gr.is_subagent,
+                "agent_id": gr.agent_id,
+                "timestamp": gr.timestamp,
+            },
+        )
+
+    @staticmethod
+    def _to_guardrail(d: AuthzDecision) -> GuardrailDecision:
+        """Convert an authorization decision to a guardrail decision."""
+        return GuardrailDecision(
+            allow=d.allow,
+            reasons=[GuardrailReason(code=r.code, message=r.message) for r in d.reasons],
+            policy_id=d.policy_id,
+            metadata=d.metadata,
+        )
+
+    def evaluate(self, request: GuardrailRequest) -> GuardrailDecision:
+        """Synchronous evaluation: delegate to ``provider.authorize``.
+
+        Provider exceptions are intentionally allowed to propagate. The
+        adapter is consumed by :class:`~deerflow.guardrails.middleware.GuardrailMiddleware`,
+        whose ``wrap_tool_call`` / ``awrap_tool_call`` already applies
+        fail-closed semantics based on its ``fail_closed`` parameter
+        (backed by ``AuthorizationConfig.fail_closed``). Catching exceptions
+        here would duplicate that logic and risk divergent behavior between
+        the two layers.
+        """
+        decision = self._provider.authorize(self._to_authz(request))
+        return self._to_guardrail(decision)
+
+    async def aevaluate(self, request: GuardrailRequest) -> GuardrailDecision:
+        """Async evaluation: delegate to ``provider.aauthorize``.
+
+        See :meth:`evaluate` for exception-propagation rationale.
+        """
+        decision = await self._provider.aauthorize(self._to_authz(request))
+        return self._to_guardrail(decision)
--- a/backend/packages/harness/deerflow/authz/provider.py
+++ b/backend/packages/harness/deerflow/authz/provider.py
@@ -1,0 +1,121 @@
+"""AuthorizationProvider protocol and data structures for fine-grained resource authorization.
+
+This is the policy brain for resource-level authorization (RBAC and beyond),
+deliberately kept as a sibling to :mod:`deerflow.guardrails` rather than folded
+into it. PR #3665 (which added ``user_role``/``user_id`` to
+``GuardrailRequest``) explicitly scoped guardrails to *execution-time* checks
+only — *"保持 Guardrail 的职责边界不变：不新增 policy engine、RBAC 系统、
+governance 子系统"*. This module is the RBAC brain that #3665 deferred.
+
+The provider is enforced at **two layers** from one policy:
+
+1. **Assembly-time capability filter** — removes tools a role can never use
+   *before* they are bound to the agent, so the model never sees them and
+   ``tool_search`` can never promote them back (fail-closed).
+2. **Run-time execution deny** — reuses :class:`~deerflow.guardrails.middleware.GuardrailMiddleware`
+   via a thin adapter (see :mod:`deerflow.authz.adapter`), catching dynamic
+   resources and argument-based restrictions.
+
+See ``docs/plans/2026-07-10-pluggable-authorization-rfc.md`` (issue #4063) for
+the full design rationale.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any, Protocol, runtime_checkable
+
+
+@dataclass
+class Principal:
+    """The actor. Built once per run from ``request.state.user`` + context.
+
+    Identity fields mirror what ``inject_authenticated_user_context``
+    (``app/gateway/services.py``) already stamps into the run context, so the
+    provider sees a single resolved actor rather than re-deriving it per call.
+    """
+
+    user_id: str | None = None
+    role: str | None = None
+    oauth_provider: str | None = None
+    oauth_id: str | None = None
+    channel_user_id: str | None = None
+    is_internal: bool = False
+    attributes: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class AuthzRequest:
+    """Context passed to the provider for each authorization check."""
+
+    principal: Principal
+    resource: str
+    """Resource type, e.g. ``"tool"``, ``"model"``, ``"skill"``, ``"sandbox"``, ``"mcp_server"``, ``"route"``."""
+
+    action: str
+    """Action on the resource, e.g. ``"call"``, ``"list"``, ``"use"``, ``"activate"``, ``"execute"``, ``"read"``, ``"write"``."""
+
+    target: str
+    """Resource identifier: tool name, model name, skill name, ``"route:threads:read"``, etc."""
+
+    context: dict[str, Any] = field(default_factory=dict)
+    """Additional context: ``thread_id``, ``run_id``, ``tool_call_id``, ``tool_input``, ``is_subagent``, etc."""
+
+
+@dataclass
+class AuthzReason:
+    """Structured reason for an allow/deny decision."""
+
+    code: str
+    message: str = ""
+
+
+@dataclass
+class AuthzDecision:
+    """Provider's allow/deny verdict."""
+
+    allow: bool
+    reasons: list[AuthzReason] = field(default_factory=list)
+    policy_id: str | None = None
+    metadata: dict[str, Any] = field(default_factory=dict)
+
+
+@runtime_checkable
+class AuthorizationProvider(Protocol):
+    """Contract for pluggable fine-grained authorization.
+
+    Any class with these methods works - no base class required.
+    Providers are loaded by class path via ``resolve_variable()``, the same
+    mechanism DeerFlow uses for models, tools, sandbox, and guardrails.
+
+    ``resource``, ``action``, and ``target`` are free-form strings (not an
+    enum) so new resource types and provider-specific resources need no schema
+    change. The built-in RBAC provider interprets them; custom providers
+    define their own.
+    """
+
+    name: str
+
+    def authorize(self, request: AuthzRequest) -> AuthzDecision:
+        """Per-call decision. Feeds Layer 2 (execution) and route checks."""
+        ...
+
+    async def aauthorize(self, request: AuthzRequest) -> AuthzDecision:
+        """Async variant."""
+        ...
+
+    def filter_resources(
+        self,
+        principal: Principal,
+        resource_type: str,
+        candidates: list[str],
+    ) -> list[str]:
+        """Layer 1: batch visibility filter at assembly time.
+
+        Returns the subset of *candidates* the principal is allowed to see.
+        This is a required method — providers that do not have a static
+        role→resource map should delegate to :meth:`authorize` per item and
+        return only the allowed subset. Providers with a static map can
+        override this for O(1) filtering and fail-closed visibility.
+        """
+        ...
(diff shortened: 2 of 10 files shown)


willem-bd (commented):
Thanks @hata33. I found one issue that should be addressed before this is ready.

[P2] `filter_resources` docstring promises a default implementation that does not exist
- Location: `backend/packages/harness/deerflow/authz/provider.py:223-237` (the `filter_resources` method on the `AuthorizationProvider` Protocol)
- Problem: The docstring states "The default implementation delegates to `authorize` per item (correct but slow)." A `typing.Protocol` method whose body is `...` provides no implementation. A provider that trusts this docstring and omits `filter_resources` will silently fail `isinstance(provider, AuthorizationProvider)` (because `@runtime_checkable` requires the method to exist), and a provider that explicitly inherits from the Protocol will get `None` back from the inherited `...` body instead of a filtered list. Either failure mode is fail-open for a Layer 1 capability filter, contradicting the `fail_closed: true` default.
- Evidence: Verified empirically - a class with `name`/`authorize`/`aauthorize` but no `filter_resources` fails `isinstance(..., AuthorizationProvider)`; a class that inherits from the Protocol without overriding `filter_resources` gets `None` returned, not a delegated `authorize` call. The `...` body is a no-op, not a delegation. The three test providers all define `filter_resources` explicitly, so the tests never exercise the "default" path the docstring describes.
- Suggested fix: Either (a) make the docstring accurate - e.g. "Implementations should delegate to `authorize()` per item when they do not have a static role->resource map" - and mark the method as required by the Protocol, or (b) if a default is genuinely desired, provide one via a non-Protocol base class (e.g. an `AuthorizationProviderBase` ABC with a real `filter_resources` body that calls `self.authorize(...)` per candidate and returns the allowed subset), and have the Protocol describe the interface separately.
- Test: Add a test that a provider defining only `name`/`authorize`/`aauthorize` either (a) is not considered an `AuthorizationProvider` (if `filter_resources` is required) or (b) gets correct per-item delegation behavior from the default (if a default implementation is provided). Today no test covers this gap.

hata33:
@willem-bd Good catch — verified empirically that you're right on both counts:

- `_MinimalProvider` (no `filter_resources`) → `isinstance(..., AuthorizationProvider)` returns `False`
- `_InheritingProvider` (inherits Protocol, no override) → `filter_resources` returns `None`

Went with fix (a): the docstring now accurately states `filter_resources` is a **required** method, and providers without a static role→resource map should delegate to `authorize()` per item in their own implementation. Added `test_provider_without_filter_resources_is_not_provider` to assert that a provider omitting `filter_resources` fails the `isinstance` check.

Pushed in `c00ca617`.

zhfeng (commented):
## Review: Phase 0 scaffolding

Solid Phase 0 scaffold. The package layout (sibling to `guardrails/`, not folded in), the `@runtime_checkable` Protocol mirroring `GuardrailProvider`, and the config-singleton pattern mirroring `GuardrailsConfig` are all the right calls. Dataclass and adapter-mapping tests are thorough.

A few contract issues worth resolving **before Phase 1 builds on them**, since Phase 1 will inherit these shapes as load-bearing. Nothing here breaks at runtime in Phase 0 (`enabled: false`), so non-blocking.

---

### Medium

**1. `filter_resources` "default implementation" does not exist - a `Protocol` can't provide one.**

`provider.py:107-121` declares `filter_resources` with a `...` body and a docstring stating *"The default implementation delegates to `authorize()` per item (correct but slow)."* The PR description also claims *"providers that only implement `authorize()` still work for Layer 1."* Neither is true. A `Protocol` body of `...` provides no implementation, and `@runtime_checkable` requires **all** declared members. Verified empirically:

```python
class OnlyAuthorize:           # has name + authorize, omits filter_resources
    ...
isinstance(OnlyAuthorize(), AuthorizationProvider)  -> False   # runtime_checkable requires ALL members
OnlyAuthorize().filter_resources(["a"])             -> AttributeError  # no fallback exists
```

So a provider that omits `filter_resources` (a) **fails** the `isinstance` conformance check and (b) has no per-item fallback to land on. The built-in RBAC provider will override it, so this only bites custom providers - but the contract is being set now. Resolve before Phase 1: ship a `BaseAuthorizationProvider` ABC/mixin that supplies the default, **or** drop the claim and require all providers to implement it, **or** move the per-item fallback into the Layer 1 caller (and fix the docstring).

**2. Adapter's `is_internal` is derived from the wrong signal and is effectively always `False`.**

`adapter.py:48` sets `is_internal=(gr.user_role == "internal")`. But `inject_authenticated_user_context` (`app/gateway/services.py:296-311`) makes `user_role == "internal"` essentially unreachable:

- Pure internal caller (no owner): `user_role` is **popped** → `None` (lines 301-303).
- Internal-with-owner (the IM-bridge case): `user_role` = the **owner's** `system_role` (line 308), e.g. `"user"`/`"admin"`.
- Regular user: `user_role` = `"user"`/`"admin"` (line 316).

So `Principal.is_internal` via the adapter path is `False` for the very internal callers it's meant to identify. The correct signal is `auth_source == AUTH_SOURCE_INTERNAL` (`services.py:671`), which lives on `request.state`, **not** on `GuardrailRequest` - the adapter simply can't compute this from what it's given. Phase 1 should thread `is_internal` (or `auth_source`) into the run context the same way `user_role` is, so Layer 1 (the `Principal` builder in `services.py`) and Layer 2 (the adapter) derive it from one source - otherwise the two layers will disagree about who the actor is.

**3. The one behavior-affecting change in this PR is untested, and the test that claims to cover it doesn't.**

The wiring line `load_authorization_config_from_dict(config.authorization.model_dump())` lives in `_apply_singleton_configs`, which is called **only** by `AppConfig.from_file()` (`app_config.py:331`; `test_runtime_lifecycle_e2e.py:227` even documents this). But `test_app_config_load_propagates_to_singleton` calls `AppConfig.model_validate(data)` - which runs the `_build_name_indexes` after-validator but **not** `_apply_singleton_configs` - and only asserts `validated.authorization.enabled` (the pydantic field), never `get_authorization_config().enabled`. Deleting the wiring line would leave all 29 tests green. Add a test that drives `from_file` (or calls `_apply_singleton_configs` directly) and asserts the singleton was populated.

---

### Low

**4. Dead code: `_REASON_ALLOW` / `_REASON_DENY`.** Defined at `adapter.py:20-21` with a comment about giving downstream consumers "consistent codes," but `_to_guardrail` forwards the provider's reason codes verbatim - the constants are never referenced. Either remove them, or actually normalize (e.g., synthesize a reason when the provider returns none).

**5. No conformance test that the adapter is a `GuardrailProvider`.** The entire premise of the adapter is "existing `GuardrailMiddleware` can enforce authz decisions without a new middleware." That rests on `isinstance(GuardrailAuthorizationAdapter(...), GuardrailProvider)` - untested. One assertion would lock it in.

**6. Adapter drops `agent_id` and `timestamp` from `GuardrailRequest`.** They're not mapped into `AuthzRequest.context`. `agent_id` (the `passport`) in particular could be relevant to an authz decision (which agent/subagent is calling). Worth including, or explicitly documenting as intentionally omitted.

---

### Nits

- `config.example.yaml`: the block is entirely commented out. Consider an uncommented `authorization:\n  enabled: false` so the key is discoverable in a real config. Acceptable for Phase 0 either way.
- Singleton mutation isn't thread-safe, but this mirrors the existing `GuardrailsConfig`/`checkpointer` pattern, so it's not a regression - just noting it's now one more module global in that family.
- `filter_resources` takes no `action` param. Fine for action-agnostic *visibility* filtering, but worth confirming the RFC intends visibility = "can see/call" only (no `list`-vs-`execute` distinction at assembly time).

hata33:
@zhfeng Thanks for the thorough review. All items addressed in `41d1d43e`:

**Medium**

1. **`filter_resources` default implementation** — Went with "drop the claim and require all providers to implement it." Docstring now states it is a required method; `test_provider_without_filter_resources_is_not_provider` asserts a provider omitting it fails `isinstance`.

2. **`is_internal` derived from wrong signal** — The adapter no longer sets `is_internal` at all (removed the `gr.user_role == "internal"` line). `Principal` retains its dataclass default (`False`). The docstring `.. note::` documents why (correct signal is `auth_source` on `request.state`, not on `GuardrailRequest`) and that Phase 1 will thread it into run context. This way the adapter makes no claim about `is_internal` — Phase 1's Principal builder will be the single source.

3. **Singleton wiring untested** — `test_app_config_load_propagates_to_singleton` now calls `AppConfig._apply_singleton_configs(validated, acp_agents={})` directly and asserts `get_authorization_config().enabled is True`. Deleting the wiring line fails this test.

**Low**

4. **Dead code** — Removed `_REASON_ALLOW` / `_REASON_DENY`.
5. **Adapter conformance** — Added `test_adapter_is_guardrail_provider`: `isinstance(GuardrailAuthorizationAdapter(...), GuardrailProvider)`.
6. **Dropped `agent_id` / `timestamp`** — Both now mapped into `AuthzRequest.context`. Test covers them.

**Nits**

- `config.example.yaml`: changed to uncommented `authorization:\n  enabled: false`.
- `filter_resources` without `action` param: confirmed this is intentional. Layer 1 is visibility-only ("what can this role ever see?"), Layer 2 (`authorize` with `action`) handles list-vs-call-vs-execute distinctions. Matches RFC §4's two-layer definition.

31 tests pass, lint/format clean.

config.example.yaml
```diff
@@ -1848,6 +1848,16 @@ run_ownership:
...
 #       key: value
 
+# ============================================================================
+# Authorization Configuration
+# ============================================================================
+# Fine-grained resource authorization (RBAC and beyond). Disabled by default;
+# every authenticated user has access to all resources. The schema and built-in
+# RBAC provider will be introduced in subsequent phases.
+# See RFC: https://github.com/bytedance/deer-flow/issues/4063
+authorization:
```
willem-bd, line 1879: **🟡 `config_version` not bumped despite a schema change.** Adding a new top-level `authorization` section (and a new `AuthorizationConfig` field on `AppConfig`) is a config-schema change. The project convention is near-universal - every recent feature PR bumps `config_version` (e.g. #4023, #3986, #3601, #3898, #3902). This PR leaves it at 22, so `make doctor` won't nudge upgraders and `make config-upgrade` won't merge this section into existing `config.yaml` files.

Defaults are safe (`enabled: false`), so this is observability/onboarding rather than a functional bug - but please bump to follow the pattern. Note #4143 also targets 23, so whichever merges second should go to 24.

backend/packages/harness/deerflow/authz/provider.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def authorize(self, request: AuthzRequest) -> AuthzDecision:
+        """Per-call decision. Feeds Layer 2 (execution) and route checks."""
+        ...
+
+    async def aauthorize(self, request: AuthzRequest) -> AuthzDecision:
+        """Async variant."""
+        ...
+
+    def filter_resources(
```
willem-bd, line 107: **🟡 Description contradicts the Protocol re: `filter_resources` default impl.** The PR description says `filter_resources` *"has a default impl path: providers that only implement `authorize()` still work for Layer 1 (per-item delegation)."* But this Protocol declares it with no real body (just `...`), and `test_provider_without_filter_resources_is_not_provider` explicitly asserts a provider omitting it is **not** an `AuthorizationProvider`. So there is no default-implementation path - a provider with only `authorize()`/`aauthorize` fails `isinstance` and would raise `AttributeError` at Layer 1.

Either ship a base class with a default `filter_resources` (delegating to `authorize()` per item), or correct the description to state the method is mandatory. A Phase 1 implementer reading the description could ship a provider that silently fails Layer 1.

backend/packages/harness/deerflow/authz/adapter.py
```diff
@@ -0,0 +1,89 @@
...
+    def _to_guardrail(d: AuthzDecision) -> GuardrailDecision:
+        """Convert an authorization decision to a guardrail decision."""
+        return GuardrailDecision(
+            allow=d.allow,
+            reasons=[GuardrailReason(code=r.code, message=r.message) for r in d.reasons],
+            policy_id=d.policy_id,
+            metadata=d.metadata,
+        )
+
+    def evaluate(self, request: GuardrailRequest) -> GuardrailDecision:
```
willem-bd, line 81: **🟡 `fail_closed` is not enforced here.** `AuthorizationConfig.fail_closed` is documented as *"Block access if the provider errors or identity is unresolved"* and the adapter docstring even references it - but `evaluate`/`aevaluate` call `self._provider.authorize(...)` with no try/except and never consult `fail_closed`. Since this adapter *is* the Layer 2 enforcement point, fail-closed semantics won't actually hold unless Phase 1 adds exception handling here.

The adapter isn't wired in Phase 0 (so latent), but it's shipped code. Suggest a `# TODO(phase1): consult fail_closed on provider error` marker, or a test pinning the current propagates-exception behavior so the fail-closed contract isn't forgotten when wiring lands.

config.example.yaml
```diff
@@ -1848,6 +1848,16 @@ run_ownership:
 #     config:
 #       key: value
 
+# ============================================================================
+# Authorization Configuration
+# ============================================================================
+# Fine-grained resource authorization (RBAC and beyond). Disabled by default;
+# every authenticated user has access to all resources. The schema and built-in
+# RBAC provider will be introduced in subsequent phases.
```
willem-bd: **🟢 Minor - comment is slightly inaccurate.** This says *"The schema and built-in RBAC provider will be introduced in subsequent phases,"* but `AuthorizationConfig` (the schema) **is** introduced in this PR - only the runtime wiring and built-in provider are deferred. Reword to e.g. *"Schema is present but inert; the built-in RBAC provider and runtime wiring arrive in subsequent phases."*

hata33:
Updated in `0b227372`:

- **config_version**: bumped to 25 (rebased on latest `upstream/main`, which is now at 24).
- **fail-closed description**: corrected. The adapter intentionally propagates provider exceptions; `GuardrailMiddleware`'s `wrap_tool_call`/`awrap_tool_call` applies `fail_closed` semantics. Updated both `evaluate`/`aevaluate` docstrings and test comments to reflect this. Added `test_aevaluate_propagates_provider_exception` for the async path.

33 tests pass, lint/format clean.

zhfeng:
@hata33 please sync the version to 25 in `deploy/helm/deer-flow/values.yaml` and `deploy/helm/deer-flow/README.md` as well

hata33:
@zhfeng Done — synced `config_version: 25` in both `deploy/helm/deer-flow/values.yaml` and `deploy/helm/deer-flow/README.md` (`d2473936`).

backend/packages/harness/deerflow/authz/provider.py
```diff
@@ -0,0 +1,121 @@
...
+The provider is enforced at **two layers** from one policy:
+
+1. **Assembly-time capability filter** — removes tools a role can never use
+   *before* they are bound to the agent, so the model never sees them and
+   ``tool_search`` can never promote them back (fail-closed).
+2. **Run-time execution deny** — reuses :class:`~deerflow.guardrails.middleware.GuardrailMiddleware`
+   via a thin adapter (see :mod:`deerflow.authz.adapter`), catching dynamic
+   resources and argument-based restrictions.
+
+See ``docs/plans/2026-07-10-pluggable-authorization-rfc.md`` (issue #4063) for
```
WillemJiang, line 19: @hata33  Please add the RFC document to the PR for us to track the design

hata33:
@WillemJiang Added the RFC document at `docs/plans/2026-07-10-pluggable-authorization-rfc.md` (`fb46c888`). It is the full text from #4063 authored by @zhfeng, included here for design tracking as the implementation progresses through phases.

WillemJiang (approved):

Issue #4121, opened 2026-07-12
[enhancement] GitHub channel silently drops concurrent comments while an autonomous run is in flight (multitask_strategy=reject)


zhfeng:
## Problem

When a GitHub webhook arrives for a `(repo, number, agent)` thread that already has an active run, the run is rejected and the triggering comment is **silently dropped** - the human sees no reaction, no comment, no acknowledgement, and reasonably concludes the bot ignored them.

## Current behavior

The GitHub chat path is `fire_and_forget` with `multitask_strategy: "reject"` (`app/gateway/github/run_policy.py`, `app/channels/manager.py`):

1. First comment -> `client.runs.create(thread_id, …)` succeeds -> autonomous run starts (legitimately long: clone, edit, test, push, comment).
2. A second comment arrives on the same thread while the first run is still active -> `runs.create` raises `ConflictError`.
3. The manager catches it (`_is_thread_busy_error`) and calls `_send_error(msg, THREAD_BUSY_MESSAGE)`.
4. `GitHubChannel.send` is **log-only by design** (agents post via `gh` themselves), so `THREAD_BUSY_MESSAGE` goes nowhere visible.
5. **Net: the second comment is silently dropped.** No reaction, no comment, no queue signal.

This collision window is ~6 minutes per typical coding run, wide enough to fire routinely:
- Human realizes 30s after their initial comment that they need to clarify.
- Multiple reviewers comment at once on a PR the bot is still working on.
- A bot reviewer (Copilot, CodeRabbit) submits a `pull_request_review` while our coder bot is mid-PR.

#4003 (merged) made the "one active run per thread" invariant DB-enforced via the `uq_runs_thread_active` partial unique index, so the `ConflictError` is now deterministic and cross-pod-safe (good - no more two-concurrent-runs race). But it does nothing about the silent drop itself - the second comment still vanishes.

## Proposed direction

A per-thread follow-up buffer + watcher, layered on the existing `StreamBridge.END_SENTINEL` completion signal:

- **On `ConflictError`**: append the comment to a per-`(thread_id)` buffer (dedup by `comment_id`), react `eyes` on the comment via the installation token ("queued; will pick up after the current run"). Hard cap (e.g. 20); overflow reacts `confused` + WARNING log as a visible overload signal.
[...]

PR #4131, opened 2026-07-12
fix(channels): drop redundant GitHub review-comment webhook fan-out

DaoyuanLi2816:
## Summary

GitHub fires a `pull_request_review_comment` webhook for *every* inline comment attached to a `pull_request_review` submission, **in addition to** the single `pull_request_review` event for the review as a whole. A bot reviewer like CodeRabbit routinely leaves 20-30 inline comments per review, so that's 20-30 near-duplicate webhook deliveries for content a properly-configured binding can recover from the parent `pull_request_review` event via `gh api`.

This filters those companion deliveries in `fanout_event()`, per matched agent binding.

## Background — relationship to #4121

This is "Slice 1" of the two independently-shippable pieces #4121 scopes out:

> **Fanout filter (independent, ship first)**: suppress `pull_request_review_comment` events that carry a `pull_request_review_id` (and no `in_reply_to_id`) - the parent `pull_request_review` event already covers them [...] This eliminates the 20-30 comment "CodeRabbit storm" as a buffer concern before the buffer is even built.

**Scope note:** this PR is a standalone, narrower improvement, *not* a fix for the silent-drop behavior #4121 reports. #4121 is fundamentally about the GitHub channel silently dropping a comment when it arrives while a `multitask_strategy="reject"` run is already active on that thread; addressing that needs the per-thread follow-up buffer + watcher #4121 also describes, which is a separate, larger change (in progress separately). This PR is only the smaller, standalone half: it reduces redundant webhook noise (and the wasted per-agent-binding iteration that noise causes today) regardless of whether the buffer ever lands. Referencing #4121 for context — this PR by itself leaves that issue open.

## The redundant fan-out

Concretely, for a PR review with N inline comments, GitHub delivers:

- **1x** `pull_request_review` — the review as a whole (state, top-level body, and the agent can pull every inline comment off it via `gh api repos/{owner}/{repo}/pulls/{number}/reviews/{review_id}/comments`).
- **Nx** `pull_request_review_comment` — one per inline comment, each carrying `pull_request_review_id` pointing back at the review above.

Every review-thread comment (`GET /repos/{owner}/{repo}/pulls/comments`, the shape the webhook `comment` object mirrors) carries `pull_request_review_id`. The one field that distinguishes "fresh companion comment from the review" versus "a later reply inside that comment's thread" is `in_reply_to_id`: absent on the former, set on the latter. A reply is a genuine new interaction (a human — or another bot — responding after the fact) and must still fire; the companion comments are not.

## The fix

This has gone through two review rounds and the design changed shape each time. This section describes what's actually shipped now, not the original proposal.

**Round 1 (as originally opened):** an unconditional, global early return in `fanout_event()` — before the registry lookup even ran — suppressed every companion comment. willem-bd/zhfeng requested changes: a binding that subscribes to `pull_request_review_comment` *alone* (no `pull_request_review` trigger) never receives the parent review event at all, so unconditionally dropping its companion comments was a silent, total loss of the review's inline content for it, not noise reduction.

**Round 2 (commit b3791b33):** the filter became a **per-binding gate**, applied inside the per-agent loop (after the registry lookup), next to the existing self-event gate. `_is_redundant_review_comment(payload)` still decides whether the *event* has the companion shape (`pull_request_review_id` set, `in_reply_to_id` absent). Whether that shape is safe to *drop* is decided per matched binding: only suppressed for a binding that is *also* registered for `pull_request_review` on this same repo — reusing the same `lookup_agents` registry resolution, so it doesn't duplicate any trigger-matching logic and stays correctly scoped per repo. A `pull_request_review_comment`-only binding is never suppressed. willem-bd approved this as "the right design" for the delivery gap.

**This round (require_mention gap — willem-bd, Medium):** the round-2 gate's premise — "a binding registered for `pull_request_review` on this repo has an independent path to the review" — only holds if that trigger's own gates would actually let the review event through. If the binding set `require_mention: true` on its `pull_request_review` trigger, that review event can be silently dropped by its own mention check against `review["body"]` (the review's top-level summary) — a field this `pull_request_review_comment` payload never carries, so there is no way to verify from a comment delivery whether that check would pass. A human `@mention` living only inside one inline comment (not the review summary) would then be lost twice over: the review event filtered as `no_mention`, and the one inline comment that actually carries the mention dropped here as "redundant" — the same silent-loss shape as the original bug, through a narrower path. The gate now additionally requires the paired trigger's resolved `require_mention` to be `False` before treating it as coverage. This trades a small amount of residual redundancy (an extra companion delivery on occasions when the review would have fired anyway) for zero silent loss — the same trade the round-2 gate already made at a coarser grain. It intentionally does not replay `allow_authors` or an `actions` whitelist that might also gate the paired trigger; those remain out of scope for this fix, same as `self_event`.

Two smaller things landed alongside it:
- The redundant-comment skip reason now prefers the companion's own trigger verdict when that verdict is *also* a skip (e.g. its own `require_mention` independently fails), instead of always reporting the generic `redundant_review_comment` label — a debugging-precision improvement with no behavior change (willem-bd, Minor).
- `_pr_review_prompt` now explicitly instructs the agent to fetch a review's inline comments via `gh api .../reviews/{review_id}/comments`. The redundancy gate's suppression is only genuinely redundant *if* the agent actually recovers that content from the parent review event — nothing previously told it to (zhfeng).

## Tests

`backend/tests/test_github_dispatcher.py` — the "Redundant review-comment fan-out suppression" block now has 9 tests:

- `test_review_comment_companion_to_review_is_suppressed` — a genuinely redundant companion (dual-subscribed, unconditional review trigger) is suppressed. The original fix's core value.
- `test_review_comment_only_binding_not_suppressed` — a review-comment-only binding still fires (round-2 regression fix).
- `test_review_comment_not_suppressed_when_review_trigger_is_on_different_repo` — coverage is scoped per repo, not per agent.
- `test_review_comment_reply_within_thread_still_fires` / `test_review_comment_without_review_id_still_fires` / `test_issue_comment_unaffected_by_review_comment_filter` — shape/event-type guards, no regression.
- `test_review_comment_not_suppressed_when_review_trigger_requires_mention` — **this round's regression test**: dual-subscribed binding, `require_mention: true` on the `pull_request_review` trigger, mention present only in the inline comment. Confirmed failing (silently suppressed) against the round-2 code via patch-file revert; passes with the fix.
- `test_review_comment_gate_is_independent_per_agent_in_same_call` — two bindings, one covered/suppressed and one not, resolved correctly within a single `fanout_event` call.
- `test_review_comment_redundant_skip_prefers_own_trigger_reason_when_it_also_fails` — the skip-reason-precedence regression test.

`backend/tests/test_github_prompts.py` — two new tests lock the inline-comment fetch hint on `_pr_review_prompt` and its guard against rendering a broken `gh api` path when `review.id` is missing.

Fail-before/pass-after re-verified this round by isolating the source changes (`dispatcher.py`, `prompts.py`) into patch files and reverse-applying them with the new tests left in place: `test_review_comment_not_suppressed_when_review_trigger_requires_mention`, `test_review_comment_redundant_skip_prefers_own_trigger_reason_when_it_also_fails`, and `test_pull_request_review_prompt_instructs_fetching_inline_comments` fail on the prior code; every other test — including the two new ones added this round that don't hinge on the fix itself, and all pre-existing tests — still passes, confirming those pin genuinely pre-existing behavior rather than being coincidentally broken. Re-applying the patches turns all of them green.

Full suite: `46 passed` across `test_github_dispatcher.py` (31 tests) + `test_github_prompts.py` (15 tests), run via:

```
PYTHONPATH="<repo>/backend/packages/harness;<repo>/backend" python -m pytest backend/tests/test_github_dispatcher.py backend/tests/test_github_prompts.py -v
```

`ruff check` and `ruff format --check` are both clean on all four changed files.

### DIFF
--- a/backend/app/gateway/github/dispatcher.py
+++ b/backend/app/gateway/github/dispatcher.py
@@ -33,7 +33,7 @@ from app.gateway.github.identity import extract_target, resolve_thread_id
 from app.gateway.github.prompts import build_prompt
 from app.gateway.github.registry import build_github_agent_registry, lookup_agents
 from app.gateway.github.triggers import event_should_fire
-from deerflow.config.agents_config import GitHubAgentConfig
+from deerflow.config.agents_config import GitHubAgentConfig, GitHubTriggerConfig
 
 logger = logging.getLogger(__name__)
 
@@ -127,12 +127,40 @@ def _is_redundant_review_comment(payload: dict[str, Any]) -> bool:
     IMPORTANT: a ``True`` result is NOT by itself a "safe to drop" signal —
     see the per-binding gate in :func:`fanout_event`, which only suppresses
     a companion comment for a binding that *also* has its own
-    ``pull_request_review`` trigger on the same repo (and therefore an
-    independent path to the review). A binding that subscribes to
-    ``pull_request_review_comment`` alone never receives the parent review
-    event, so unconditionally dropping its companion comments would be a
-    silent, total loss of the review's inline content for it — not noise
-    reduction (PR #4131 review feedback from willem-bd / zhfeng).
+    ``pull_request_review`` trigger on the same repo AND whose *resolved*
+    trigger does not itself require a mention (see the ``require_mention``
+    gap note below) — i.e. an unconditional, independent path to the
+    review. A binding that subscribes to ``pull_request_review_comment``
+    alone never receives the parent review event, so unconditionally
+    dropping its companion comments would be a silent, total loss of the
+    review's inline content for it — not noise reduction (PR #4131 review
+    feedback from willem-bd / zhfeng).
+
+    ``require_mention`` gap (PR #4131 review, Medium finding, willem-bd —
+    second round, against the per-binding gate above): a dual-subscribed
+    binding's ``pull_request_review`` trigger is only a *guaranteed*
+    independent path when that trigger does not itself gate on
+    ``require_mention``. If it does, the paired review event can be
+    silently dropped by :func:`app.gateway.github.triggers.event_should_fire`'s
+    own mention check against ``review["body"]`` — the review's
+    *top-level* summary, which this ``pull_request_review_comment``
+    payload never carries (there is no way to see, from a comment
+    delivery, what the sibling review's own summary said). A human
+    ``@mention`` that lives only inside one inline comment — not the
+    review summary — would otherwise be lost twice over: the review event
+    is filtered out (``no_mention``) *and* the one inline comment that
+    actually carries the mention is dropped here as "redundant", via a
+    narrower path than the original bug. The per-binding gate therefore
+    additionally requires ``require_mention`` to be false on the paired
+    trigger before treating it as coverage. This trades a small amount of
+    residual redundancy (an extra companion delivery on occasions when the
+    review's own summary happened to carry the mention too, or
+    ``allow_authors``/self-event would have let the review through anyway)
+    for zero silent loss — the same trade the original fix already made at
+    a coarser grain. It deliberately does not attempt to replay
+    ``allow_authors`` or an ``actions`` whitelist that might also be
+    configured on the paired trigger; those are accepted as out of scope
+    for this narrower fix, same as ``self_event``.
 
     Residual caveat (PR #4131 review, Concern 3, zhfeng): GitHub documents
     ``pull_request_review_id`` on the review-comment schema as nullable
@@ -145,7 +173,11 @@ def _is_redundant_review_comment(payload: dict[str, Any]) -> bool:
     own ``pull_request_review`` trigger could still lose such a comment
     under the per-binding gate. Confirmed via a real/documented webhook
     payload capture before ruling this out; treat it as an open,
-    low-probability risk rather than a settled non-issue.
+    low-probability risk rather than a settled non-issue. (Distinct from
+    the ``require_mention`` gap above: this caveat questions whether the
+    paired event fires *at all* for a given delivery shape; the
+    ``require_mention`` gap is about a paired event that fires but is then
+    filtered by its own trigger config, which the gate now accounts for.)
     """
     comment = payload.get("comment")
     if not isinstance(comment, dict):
@@ -215,18 +247,25 @@ async def fanout_event(
     #    shape of review fan-out is a property of the event and computed
     #    once here, but whether that fan-out is safe to DROP is a property
     #    of the individual binding: only a binding that also has its own
-    #    ``pull_request_review`` trigger on this repo has an independent
-    #    path to the review's content, so the actual suppression decision
-    #    is made per-agent below (next to the self-event gate), not here.
-    #    ``covered_by_review_trigger`` reuses :func:`lookup_agents` against
+    #    ``pull_request_review`` trigger on this repo, WITHOUT that trigger
+    #    itself requiring a mention, has a *guaranteed* independent path to
+    #    the review's content — so the actual suppression decision is made
+    #    per-agent below (next to the self-event gate), not here.
+    #    ``review_trigger_by_binding`` reuses :func:`lookup_agents` against
     #    the registry we just built — a binding only appears in the
     #    ``(repo, "pull_request_review")`` slot if it explicitly lists that
     #    event under its own ``triggers:`` (opt-in per binding, see
     #    ``triggers.py``) — so this does not duplicate any trigger-matching
     #    logic, and it stays correctly scoped to this repo (an agent with a
-    #    second binding on a *different* repo does not count).
+    #    second binding on a *different* repo does not count). It maps to
+    #    the binding's *resolved* :class:`GitHubTriggerConfig` (not just
+    #    membership) so the per-agent gate below can also check
+    #    ``require_mention`` — see the ``require_mention`` gap note on
+    #    :func:`_is_redundant_review_comment` for why a mention-gated review
+    #    trigger cannot be trusted as coverage (PR #4131 review, Medium
+    #    finding, willem-bd).
     is_redundant_review_comment = event == "pull_request_review_comment" and _is_redundant_review_comment(payload)
-    covered_by_review_trigger: set[tuple[str, str]] = {(m.user_id, m.agent.name) for m in lookup_agents(registry, repo, "pull_request_review")} if is_redundant_review_comment else set()
+    review_trigger_by_binding: dict[tuple[str, str], GitHubTriggerConfig] = {(m.user_id, m.agent.name): m.trigger for m in lookup_agents(registry, repo, "pull_request_review")} if is_redundant_review_comment else {}
 
     for match in matches:
         agent = match.agent
@@ -251,29 +290,11 @@ async def fanout_event(
             skipped.append({"agent": agent.name, "reason": "self_event"})
             continue
 
-        # 5. Redundant review-comment gate, per binding (PR #4131 review —
-        #    willem-bd / zhfeng). Only suppress THIS binding's companion
-        #    comment when it is also registered for ``pull_request_review``
-        #    on this repo — i.e. it has its own path to the review content
-        #    that makes the companion comment genuinely redundant *for it*.
-        #    A binding registered for ``pull_request_review_comment`` alone
-        #    never receives the parent review event at all, so it still
-        #    fires here even though the payload has the fan-out shape.
-        if is_redundant_review_comment and (match.user_id, agent.name) in covered_by_review_trigger:
-            logger.info(
-                "github_fanout: agent=%s skipped (reason=redundant_review_comment, repo=%s#%s, delivery=%s)",
-                agent.name,
-                repo,
-                number,
-                delivery_id,
-            )
-            skipped.append({"agent": agent.name, "reason": "redundant_review_comment"})
-            continue
-
-        # 6. Trigger filter.
-        # ``default_mention_login`` mirrors the precedence used by
-        # ``_is_self_event`` above, then extended with the operator
-        # default from ``channels.github.default_mention_login``:
+        # 5. Trigger filter — computed once, up front, rather than right
+        #    before its own gate further below. ``default_mention_login``
+        #    mirrors the precedence used by ``_is_self_event`` above, then
+        #    extended with the operator default from
+        #    ``channels.github.default_mention_login``:
         #
         #   1. ``trigger.mention_login`` — per-event override (handled
         #      inside ``event_should_fire``; we only pass the fallback).
@@ -299,9 +320,61 @@ async def fanout_event(
         # string. ``operator_default_mention_login`` is a plain function
         # argument (not a validated model field), so it is normalized here
         # explicitly.
+        #
+        # Computing this here (rather than at its old location right before
+        # its own gate) lets the redundant-review-comment gate just below
+        # also consult the verdict for a more precise skip reason, instead
+        # of always reporting ``redundant_review_comment`` even when this
+        # binding's own trigger would have skipped the event anyway for an
+        # unrelated reason (PR #4131 review, Minor finding, willem-bd).
+        # ``event_should_fire`` is a pure function of its arguments, so
+        # computing it once here and reusing it below is not a behavior
+        # change from calling it at the old step 6 location.
         operator_default = (operator_default_mention_login or "").strip() or None
         default_mention_login = github.bot_login or operator_default or agent.name
         fire, reason = event_should_fire(event, payload, trigger, default_mention_login)
+
+        # 6. Redundant review-comment gate, per binding (PR #4131 review —
+        #    willem-bd / zhfeng). Only suppress THIS binding's companion
+        #    comment when it is also registered for ``pull_request_review``
+        #    on this repo AND that trigger's own ``require_mention`` is not
+        #    set — i.e. it has an unconditional, independent path to the
+        #    review content that makes the companion comment genuinely
+        #    redundant *for it*. A binding registered for
+        #    ``pull_request_review_comment`` alone never receives the
+        #    parent review event at all, so it still fires here even though
+        #    the payload has the fan-out shape. A binding whose review
+        #    trigger DOES require a mention is also not treated as covered:
+        #    this ``pull_request_review_comment`` payload never carries the
+        #    review's own top-level body, so there is no way to verify from
+        #    here whether that mention check would actually pass for the
+        #    paired review event — see the ``require_mention`` gap note on
+        #    :func:`_is_redundant_review_comment`.
+        #
+        #    When this binding IS suppressed as redundant, the skip reason
+        #    prefers this binding's own trigger verdict (``reason`` from
+        #    step 5 above) over the generic ``redundant_review_comment``
+        #    label whenever that verdict is ALSO a skip — e.g. this
+        #    companion's own ``pull_request_review_comment`` trigger
+        #    separately requires a mention it doesn't have. The event is
+        #    skipped either way; this only makes the logged reason more
+        #    useful for operator debugging (PR #4131 review, Minor finding,
+        #    willem-bd).
+        review_trigger = review_trigger_by_binding.get((match.user_id, agent.name))
+        if is_redundant_review_comment and review_trigger is not None and not review_trigger.require_mention:
+            skip_reason = reason if not fire else "redundant_review_comment"
+            logger.info(
+                "github_fanout: agent=%s skipped (reason=%s, repo=%s#%s, delivery=%s)",
+                agent.name,
+                skip_reason,
+                repo,
+                number,
+                delivery_id,
+            )
+            skipped.append({"agent": agent.name, "reason": skip_reason})
+            continue
+
+        # 7. Apply the trigger filter's verdict from step 5.
         if not fire:
             logger.info(
                 "github_fanout: agent=%s skipped (reason=%s)",
@@ -311,7 +384,7 @@ async def fanout_event(
             skipped.append({"agent": agent.name, "reason": reason})
             continue
 
-        # 7. Build prompt + publish inbound message onto the bus.
+        # 8. Build prompt + publish inbound message onto the bus.
         prompt = build_prompt(event, payload)
         thread_id = resolve_thread_id(repo, number, agent.name)
 
--- a/backend/app/gateway/github/dispatcher.py
+++ b/backend/app/gateway/github/dispatcher.py
@@ -7,9 +7,9 @@ first-class :class:`Channel` (see ``app/channels/github.py``):
     POST /api/webhooks/github
         → verify HMAC (route)
         → :func:`fanout_event` (this module)
-            • drop redundant review-comment webhook noise
-            • filter bots
             • look up bound agents
+            • filter bots
+            • drop redundant review-comment webhook noise, per binding
             • apply per-binding trigger filter
             • publish one :class:`InboundMessage` per surviving agent
         → ChannelManager picks it up off the bus
@@ -102,17 +102,15 @@ def _is_self_event(
 
 
 def _is_redundant_review_comment(payload: dict[str, Any]) -> bool:
-    """Return True if this ``pull_request_review_comment`` is fan-out noise
-    from a ``pull_request_review`` submission the agent already saw.
+    """Return True if this ``pull_request_review_comment`` has the *shape*
+    of fan-out noise from a ``pull_request_review`` submission: a companion
+    inline comment that the review event already covers.
 
     GitHub fires one ``pull_request_review_comment`` webhook per inline
     comment attached to a review submission, ON TOP OF the single
     ``pull_request_review`` event for the review as a whole. A bot
     reviewer (CodeRabbit routinely posts 20-30 inline comments per review)
-    therefore floods the webhook with 20-30 near-duplicate deliveries that
-    carry nothing the agent doesn't already have: it fetches every inline
-    comment itself (via ``gh api``) when it processes the parent
-    ``pull_request_review`` event.
+    therefore floods the webhook with 20-30 near-duplicate deliveries.
 
     Each such companion comment carries ``pull_request_review_id`` (the id
     of the review it belongs to) and — the discriminator — no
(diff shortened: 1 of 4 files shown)


willem-bd (commented):

[P2] Unconditional review-comment filter silently drops the submission batch for `pull_request_review_comment`-only bindings

- Location: `backend/app/gateway/github/dispatcher.py:182-189` (early return in `fanout_event`), predicate at `:104` (`_is_redundant_review_comment`)
- Problem: The filter suppresses every companion `pull_request_review_comment` (`pull_request_review_id` set, no `in_reply_to_id`) before the per-binding registry lookup. Its premise - "an event no binding should ever need to see" - holds only for bindings that also subscribe to `pull_request_review`, because the inline comments are not carried in the review event's payload; the agent fetches them via `gh api` when it processes the parent `pull_request_review` event. The trigger model lets a binding declare `pull_request_review_comment` alone, and for such a binding the agent never receives `pull_request_review`, never fetches the review's inline comments, and the suppressed companion events were its only delivery of the submission-batch comments. Those comments are then silently lost (later replies, which carry `in_reply_to_id`, still fire - so the binding sees replies to comments it never received).
- Evidence: `build_prompt` / `app/gateway/github/prompts.py` does not fetch inline comments - the fetch happens inside the agent run off the `pull_request_review` event (the PR's own rationale states this). The filter returns at `dispatcher.py:189` before registry resolution, so it applies even to `pull_request_review_comment`-only bindings; `test_review_comment_companion_to_review_is_suppressed` configures exactly such a binding and asserts suppression, encoding the assumption rather than guarding against it.
- Suggested fix: Confirm no supported binding subscribes to `pull_request_review_comment` without `pull_request_review`. If such bindings are supported, either document the limitation in `backend/docs/GITHUB_AGENTS.md` (a `pull_request_review_comment`-only binding will not see review-submission comments) or make the filter per-binding - only suppress for matched bindings that also subscribe to `pull_request_review` (this trades the up-front perf win for correctness, since it must run after the registry lookup). At minimum, state the assumption next to `_is_redundant_review_comment`.
- Test: add a `pull_request_review_comment`-only binding case that asserts whether submission-batch comments are delivered, or explicitly document that they are not; the current test only asserts suppression.

zhfeng:
# Review: PR #4131 - drop redundant GitHub review-comment webhook fan-out

**Verdict: Request changes.** The problem is real and the discriminator is the right GitHub semantic, but the fix makes a global, non-configurable suppression of an event operators explicitly opt into, on an unverified assumption about agent behavior. That's a content-regression risk for exactly the config the PR targets. It should be opt-in per binding, or conditional on the agent also binding `pull_request_review`.

## What's good

- **Discriminator is correct.** `pull_request_review_id` set + `in_reply_to_id` absent is the right way to tell a review-companion comment from a genuine in-thread reply. The three-branch test matrix (`test_review_comment_*`) plus the `issue_comment` regression guard are well chosen.
- **Placement and shape are clean.** Filtering once up front (before the registry lookup) avoids the wasted per-binding iteration, and the `skipped: [{"reason": "redundant_review_comment"}]` early-return mirrors the existing `no_target` convention (`dispatcher.py:140`). Logging includes repo/number/delivery_id. ruff is clean.
- **Scope discipline.** The PR is honestly scoped as the narrower half of #4121 and doesn't pretend to fix the silent-drop. The fail-before/pass-after methodology (reverse-applying the source patch) is rigorous.

## Concern 1 (High) - the core justification isn't grounded in the code, and the suppression can't be recovered

The PR's entire rationale rests on this claim (PR body + #4121):

> content the agent already has in full - it fetches every inline comment itself via `gh api` when it processes the parent `pull_request_review` event.

Nothing in the dispatcher or any built-in skill enforces that:

- `_pr_review_prompt` (`prompts.py:127-149`) renders **only** the review's state, author, and top-level body. It does not include the inline comments, and it does **not** tell the agent to fetch them - it only mentions `gh pr comment` for *posting*.
- No public skill or backend tool fetches PR review comments (grepped `skills/` and `backend/`; the only references are dispatcher/triggers/prompts/identity/webhook-router).
- The one real agent on disk (`coding-llm-gateway`) doesn't bind either review event at all; its reviewer counterpart binds `pull_request: opened`.

**Failure scenario.** An operator runs an automated reviewer with `pull_request_review_comment: {require_mention: false}` (the natural config for a bot that wants to see every inline comment - and exactly the config in the PR's own tests). CodeRabbit submits a review with 30 inline comments + a summary body.

- *Before this PR:* the agent gets 30 `pull_request_review_comment` events, each carrying the specific comment, `file:line`, and `diff_hunk` (`_pr_review_comment_prompt`, `prompts.py:101-124`). The agent sees all 30.
- *After this PR:* all 30 are suppressed. The agent gets one `pull_request_review` event with state + summary body only. Unless that agent's SOUL.md happens to proactively `gh api …/reviews/{id}/comments`, **it never sees the 30 comments.** And note `require_mention: false` is precisely the config where the storm produces 30 real runs (under the default `require_mention: true`, CodeRabbit's comments don't @mention, so they're already skipped cheaply - the storm is 30 skips, not 30 runs; the filter's value is concentrated in the `false` case, which is also the case it regresses).

The filter also makes **no check that the matched agent is also bound to `pull_request_review`**. An agent bound *only* to `pull_request_review_comment` (e.g., to react to each inline comment individually) loses every companion comment with no recovery path - there is no review event for it at all.

## Concern 2 (Medium) - this conflicts with the codebase's own opt-in-per-binding contract

`triggers.py:6-11` is explicit:

> **Events are opt-in per binding.** If an event name does not appear as a key in the binding's `triggers:` mapping, the agent is **not registered** for that event … The agent's `config.yaml` is the single source of truth for "which events do I care about?".

An operator who writes `pull_request_review_comment: {require_mention: false}` has explicitly opted into that event. This PR silently overrides that opt-in via a payload-shape heuristic the operator can't see or configure. That's a real tension with the documented contract, and it's the root cause of Concern 1: the dispatcher is second-guessing an explicit binding based on an assumption about what the agent *would* do with the companion event.

**Recommended fix - make it opt-in per binding** (one of):

- **(a) A per-trigger flag**, e.g. `pull_request_review_comment: {require_mention: false, suppress_review_companion: true}`. Operators who know their agent fetches comments on the review event enable it; everyone else gets today's behavior. This aligns with the opt-in philosophy and the field-merge machinery in `_resolved_trigger` (`triggers.py:97-128`) already supports adding such a field cleanly.
- **(b) Conditional suppression**: only drop the companion comment for an agent that is *also* bound to `pull_request_review` for the same repo. This requires moving the check from the pre-registry early-return into the per-agent loop (next to the self-event gate, `dispatcher.py:174`), so it can see `match.agent`'s full binding set. It guarantees the agent actually receives the covering review event. Slightly more code, but it makes the "agent already has it" claim structurally true instead of asserted.
- **(c)** If you keep the global drop, at minimum document the assumption in `triggers.py`/`prompts.py` and have `_pr_review_prompt` actually instruct the agent to fetch inline comments - otherwise the prompt and the filter are working at cross purposes.

(a) is the smallest change that resolves both concerns; (b) is the most defensible.

## Concern 3 (Low) - residual risk on the standalone-comment edge

Test 3 (`test_review_comment_without_review_id_still_fires`) proves the *null* `pull_request_review_id` case fires. It does **not** prove that every standalone comment (no accompanying `pull_request_review` event) has a null id. The GitHub UI "Add single comment" path in particular has historically been murky about whether it creates a backing review object and sets `pull_request_review_id`. If any no-review comment ever carries a non-null `pull_request_review_id` and a null `in_reply_to_id`, this filter drops it with no review event to recover it from - a silent loss, the same shape of bug #4121 is about. Worth a sentence in the docstring acknowledging the assumption, and ideally a live payload capture from a real "Add single comment" to confirm before this ships under `risk:high`.

## Minor

- **Observability:** the early return reports `matched_agents: []` even when agents *would* have matched. The `skipped` reason disambiguates for smee redelivery, but an operator skimming the response sees "no agents interested" rather than "filtered as redundant." Acceptable, just note it.
- The renumbered step comments (2->3, 3->4, etc.) are correct and keep the narrative consistent.

## Recommendation

Don't merge as-is. The discriminator and tests are good; the policy is too blunt. Pick (a) or (b) from Concern 2 - either makes the suppression respect the operator's explicit binding and removes the unverified-assumption regression. If the team's reviewer agents genuinely do fetch inline comments on `pull_request_review`, that should be encoded (prompt instruction +/or skill), not assumed - and Concern 3's single-comment case should be confirmed against a real payload before this lands.

---

*Review posted via Claude Code. Reviewed against `main` + the PR diff; file:line references point at the current `backend/app/gateway/github/` sources.*

DaoyuanLi2816:
Thanks both for the thorough reviews — you're right, and I've pushed a fix implementing route (b) (zhfeng's "most defensible" option).

**What changed**

The suppression check moved from the pre-registry early-return into the per-agent loop. It's still keyed off the same `_is_redundant_review_comment` payload-shape predicate (`pull_request_review_id` set, `in_reply_to_id` absent), but the *decision to suppress* is now per-binding: a matched binding's companion comment is only dropped when that same binding also has its own `pull_request_review` trigger configured for the same repo. That query reuses the existing `lookup_agents(registry, repo, "pull_request_review")` against the registry already built for the current dispatch — no re-walking of `agent.github.bindings` and no new trigger-matching logic — and it's naturally scoped per-repo, since `lookup_agents` is called with the same `repo` the event fired on (an agent with a second binding on a different repo doesn't count).

Concretely: `is_redundant_review_comment` is still computed once per delivery (a property of the payload), and `covered_by_review_trigger = {(user_id, agent_name), ...}` is precomputed once too — but the actual suppression only happens inside the per-agent loop, next to the self-event gate, checking `(match.user_id, agent.name) in covered_by_review_trigger`. A `pull_request_review_comment`-only binding is never in that set, so it now always fires, matching the "opt-in per binding" contract `triggers.py` documents.

One side effect worth flagging: since suppression is no longer a pre-registry early return, a suppressed agent now shows up in `matched_agents` (with `skipped: [{"agent": ..., "reason": "redundant_review_comment"}]`) instead of the old `matched_agents: []`. zhfeng, this directly resolves your "Minor" observability note too — an operator skimming the response now sees the agent actually matched and was filtered, not "no agents interested."

**Tests**

- `test_review_comment_companion_to_review_is_suppressed` — updated: the binding is now dual-subscribed (`pull_request_review` + `pull_request_review_comment`), preserving the original fix's value — suppression still fires when the binding has its own path to the review.
- `test_review_comment_only_binding_not_suppressed` — new: the exact scenario from zhfeng's Concern 1 (a `pull_request_review_comment: {require_mention: false}` binding with no `pull_request_review` trigger). Confirmed this fails on the as-reviewed code (`fired_agents == []` instead of `["reviewer"]`) and passes after the fix.
- `test_review_comment_not_suppressed_when_review_trigger_is_on_different_repo` — new: an agent with two bindings (`pull_request_review` on repo X, `pull_request_review_comment`-only on repo Y) must still fire on Y — locks in that the gate is scoped per-repo, not per-agent.
- The other 3 pre-existing filter tests (reply-within-thread, no-review-id, issue_comment-unaffected) are untouched and still green.

Fail-before/pass-after was verified by isolating the dispatcher.py diff into a patch file and reverse-applying it against the final test file: with only the source change reverted, exactly 3 tests fail (the updated suppression test's shape assertions, plus both new regression tests), and all other 25 stay green. Re-applying the patch turns all 28 green. Full file: `28 passed` (25 pre-existing + 3 net new — one updated in place, two added).

**Concern 3 (standalone-comment edge)**

I checked GitHub's REST docs for the review-comment schema: `pull_request_review_id` is documented as "integer or null" — confirming the field is nullable by design, i.e. some review comments genuinely lack a backing review. I could not find documentation (and don't have a live payload capture) that definitively confirms or rules out whether the "Add single comment" UI action can produce a comment with `pull_request_review_id` set + `in_reply_to_id` absent *without* a companion `pull_request_review` event also firing. I'm not asserting certainty either way — I've added a docstring caveat on `_is_redundant_review_comment` spelling this out explicitly.

Worth noting what route (b) does and doesn't buy here: a `pull_request_review_comment`-only binding is now completely immune to this edge case regardless of the answer (it's never in `covered_by_review_trigger`, so it always fires). The residual risk is narrower than before — it only applies to a binding that is also subscribed to `pull_request_review` on the same repo, and only if a genuinely reviewless single-comment submission with this exact field shape turns out to exist. I'd treat that as an open, low-probability follow-up (a live payload capture, as suggested) rather than a blocker for this PR.

Thanks again for catching this — happy to adjust further if either of you sees a gap.

willem-bd (commented):
## Review summary

This is a clean revision - the per-binding gate (response to the first round's "Request changes") is the right design and resolves the silent-drop regression for review-comment-only bindings. The gate key is correct against the data model: I verified `GitHubAgentMatch.user_id` is the operator bucket (registry.py:45-67) and single-binding-per-repo is enforced (registry.py:56-58, 175-178), so `(user_id, agent.name)` within a repo uniquely identifies a binding - the two `lookup_agents` calls are both repo-scoped, so membership correctly means "this same binding also listens for the parent review event here." The extra lookup is a dict-get over the cached registry, negligible and only on the redundant path. Test coverage is strong (6 tests incl. the regression and multi-repo scoping), and the fail-before/pass-after methodology is exactly right.

### Findings

1. **(Medium) `require_mention` conditional-delivery gap** - the gate's premise breaks when the review event itself would be filtered. A dual-subscribed binding with `require_mention: true` on `pull_request_review` can lose a review whose mention lives only in an inline comment (review summary has no mention -> `no_mention` skip; the companion comment -> suppressed here). Not documented; the existing "Residual caveat" covers a different edge. Details + options inline on the gate.
2. **(Low) PR body is stale** - "## The fix" still shows the old global early-return (`return {"matched_agents": [], ...}`, no `agent` key) and "## Tests" says 4 tests asserting `matched_agents == []`. The shipped code is a per-binding gate recording `{"agent": ..., "reason": ...}` with `matched_agents == ["reviewer"]`, and 6 tests. The two extra tests + changed assertions were the point of the revision; please update the body to match (the module docstring at the top of the file *was* updated correctly).
3. **(Minor) Skip-reason precedence** - the gate runs before the trigger filter (step 5 vs 6), so a companion comment that would also fail `require_mention` is reported as `redundant_review_comment` rather than `no_mention`. Functionally equivalent; slightly less precise for debugging. Acceptable if deliberate.

backend/app/gateway/github/dispatcher.py
```diff
@@ -180,7 +251,26 @@ async def fanout_event(
...
-        # 4. Trigger filter.
+        # 5. Redundant review-comment gate, per binding (PR #4131 review —
+        #    willem-bd / zhfeng). Only suppress THIS binding's companion
+        #    comment when it is also registered for ``pull_request_review``
+        #    on this repo — i.e. it has its own path to the review content
+        #    that makes the companion comment genuinely redundant *for it*.
+        #    A binding registered for ``pull_request_review_comment`` alone
+        #    never receives the parent review event at all, so it still
+        #    fires here even though the payload has the fan-out shape.
+        if is_redundant_review_comment and (match.user_id, agent.name) in covered_by_review_trigger:
```
willem-bd: The gate's premise - "this binding has a `pull_request_review` trigger, so the review event is an adequate substitute for the companion comment" - is **conditional**: the review event still has to pass its own trigger filter to actually deliver. There's a config combination where it doesn't, and the content is lost:

- Dual-subscribed binding with `require_mention: true` on the `pull_request_review` trigger.
- A review whose summary does **not** `@mention` the bot, but one of its inline comments does.
- `pull_request_review` event: `_comment_body` scans only `payload["review"]["body"]` (the summary - `triggers.py:76-78`), finds no mention -> skipped (`no_mention`). The inline comment bodies are not in that payload.
- Companion `pull_request_review_comment` event: this gate sees dual-subscription -> suppressed (`redundant_review_comment`).
- **Net: total loss of the review for this binding, despite a legitimate mention in an inline comment.**

The `self_event` gate doesn't have this problem because it runs *before* this gate; `require_mention` runs *after* (step 6), so it isn't naturally covered.

Options:
- (a) Document as a known limitation (the "Residual caveat" paragraph at line 137 is the natural home) and advise against `require_mention: true` on `pull_request_review` when dual-subscribing.
- (b) Make the gate conservative: skip suppression when the binding's `pull_request_review` trigger has `require_mention` true (doesn't fully close it - `self_event`/`allow_authors` also gate delivery - but covers the common case).
- (c) Add a test pinning the current (lossy) behavior so the limitation is explicit.

Low probability in practice (requires explicit `require_mention: true` on a review trigger), but it's a content-loss path in the same class as the Add-single-comment caveat already documented.

backend/app/gateway/github/dispatcher.py
```diff
@@ -100,6 +101,58 @@ def _is_self_event(
...
+    see the per-binding gate in :func:`fanout_event`, which only suppresses
+    a companion comment for a binding that *also* has its own
+    ``pull_request_review`` trigger on the same repo (and therefore an
+    independent path to the review). A binding that subscribes to
+    ``pull_request_review_comment`` alone never receives the parent review
+    event, so unconditionally dropping its companion comments would be a
+    silent, total loss of the review's inline content for it — not noise
+    reduction (PR #4131 review feedback from willem-bd / zhfeng).
+
+    Residual caveat (PR #4131 review, Concern 3, zhfeng): GitHub documents
```
willem-bd, line 165: This "Residual caveat" paragraph is the natural place to also document the `require_mention` interaction (see the gate at line 262): a dual-subscribed binding with `require_mention: true` on its `pull_request_review` trigger can lose a review whose mention lives only in an inline comment, because the review event is filtered out by `require_mention` (the summary body has no mention) while the companion comment is suppressed here. Same "open, low-probability risk" framing as the Add-single-comment caveat above.

backend/tests/test_github_dispatcher.py
```diff
@@ -1034,3 +1034,318 @@ async def test_coder_and_reviewer_on_same_pr_get_distinct_threads(base_dir: Path
...
+# for it in the first place -- suppressing the companion comment too would
+# silently drop the entire review submission's inline content for that
+# binding, with no recovery path. The filter is now applied per matched
+# binding: it only suppresses when that SAME binding also has its own
+# `pull_request_review` trigger configured for this repo.
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+async def test_review_comment_companion_to_review_is_suppressed(base_dir: Path) -> None:
```
willem-bd, line 1070: Two tests that would strengthen this block:

1. **`require_mention` interaction** - the gap noted on the gate (`dispatcher.py:262`): a dual-subscribed binding with `require_mention: true` on `pull_request_review`, where the review summary lacks a mention but the inline comment has one. Today this loses the review entirely; a test (either pinning the current lossy behavior, or driving a fix) would make the limitation explicit.

2. **Multi-agent independence** - on one companion comment, agent A (dual-subscribed) is suppressed while agent B (review-comment-only) fires in the same `fanout_event` call. The current tests cover each binding in isolation; a combined test confirms the per-binding decision is truly independent within a single call (the `covered_by_review_trigger` set + per-match check should make this work, but it's untested).

zhfeng (commented):
**Concern 1 (still open): the "agent already has the inline comments" premise is unenforced - fix it in `_pr_review_prompt` or document it**

Route (b) closes the *delivery* gap I raised - a dual-subscribed binding now receives the `pull_request_review` event, and a `pull_request_review_comment`-only binding is no longer suppressed. But it doesn't close the *content* gap that was the actual root of Concern 1. The filter still drops the events that carry each inline comment's `file:line` + `diff_hunk` + body (that's exactly what `_pr_review_comment_prompt` renders, `prompts.py:101-124`), and nothing guarantees the agent recovers that content from the parent review event.

I checked the PR branch:

- `_pr_review_prompt` (`prompts.py:127-149`) renders only the review's state, author, and top-level body. It does **not** include the inline comments, and it does **not** tell the agent to fetch them - the only `gh` reference is `gh pr comment` for *posting* (`prompts.py:146`).
- No public skill or backend tool fetches `…/reviews/{review_id}/comments` (grep of `backend/` and `skills/` is clean).

So the PR body's rationale - "content the agent already has in full - it fetches every inline comment itself via `gh api` when it processes the parent `pull_request_review` event" - holds **only if** the agent's SOUL.md happens to proactively call `gh api repos/{owner}/{repo}/pulls/{number}/reviews/{review_id}/comments`. Nothing in the codebase enforces or even suggests that. The author's "structurally true instead of asserted" framing covers *delivery* of the review event, not *receipt* of the inline comments - those are not in the review payload.

For a dual-subscribed binding whose agent doesn't do that fetch, suppressing the companion comments is a silent, total loss of the review's inline content - the same shape of bug #4121 is about, just gated on the agent's SOUL.md rather than on a `reject` run.

**Fix (one of):**

- **In the prompt (preferred):** have `_pr_review_prompt` instruct the agent to fetch the inline comments itself - e.g. append:

  > This review's inline comments are not included in this message. Before deciding what to do, fetch them with `gh api repos/{owner}/{repo}/pulls/{number}/reviews/{review_id}/comments`.

  This makes the premise true by construction for every agent, and aligns the prompt with the filter instead of leaving them at cross purposes. (Option (c) from my first review.)

- **Or at minimum, document the contract:** in `backend/docs/GITHUB_AGENTS.md` and/or the `_is_redundant_review_comment` docstring, state that a binding subscribing to both `pull_request_review` and `pull_request_review_comment` must have a SOUL.md that fetches review comments on the review event - otherwise the companion comments are dropped with no recovery path.

Without one of these, the filter and the prompt are working against each other: the filter suppresses the only events that carry the inline content, and the prompt never tells the agent to go get it.

DaoyuanLi2816:
Thanks both for the thorough follow-up review, @willem-bd and @zhfeng.

**require_mention gap (willem-bd, Medium) — fixed.** Went with option (b) from your inline comment: the per-binding gate now also requires the paired `pull_request_review` trigger's *resolved* `require_mention` to be `False` before treating a binding as covered. Since a `pull_request_review_comment` payload never carries the review's own top-level body, there's no way to verify from a comment delivery whether that trigger's own mention check would actually pass — so a mention-gated review trigger is no longer trusted as an independent path, and the companion fires on its own terms instead. This trades a bit of residual redundancy (an extra companion delivery on occasions when the review would have fired anyway) for zero silent loss, the same trade the gate already made at a coarser grain. Doesn't replay `allow_authors`/an `actions` whitelist on the paired trigger — out of scope here, same as `self_event`, and called out explicitly in the docstring rather than left implicit.

New test `test_review_comment_not_suppressed_when_review_trigger_requires_mention` reproduces your exact scenario: dual-subscribed binding, `require_mention: true` on `pull_request_review`, mention present only in the inline comment (not the review summary). Confirmed it fails (the companion is silently suppressed) against the prior code via patch-file revert, and passes with the fix. Also added `test_review_comment_gate_is_independent_per_agent_in_same_call` per your test-file suggestion — for transparency, that one passes on both the prior code and the fix (the per-binding independence already held; this just makes it an explicit, locked test rather than an implicit property).

**Skip-reason precedence (willem-bd, Minor) — fixed rather than left as-is.** Hoisted the trigger filter's `(fire, reason)` computation up before the redundancy gate (it's a pure function of its arguments, so this isn't a behavior change from where it was called before) and the redundancy gate now prefers that verdict as the skip reason whenever it's *also* a skip — e.g. a companion whose own `require_mention` independently fails now reports that instead of the generic `redundant_review_comment`. Still skipped either way; this only makes the logged reason more useful for debugging. `test_review_comment_redundant_skip_prefers_own_trigger_reason_when_it_also_fails` pins it and fails on the prior code (which always reported `redundant_review_comment`).

**PR body (willem-bd, Low) — updated.** It now describes the actual per-binding gate plus this round's `require_mention` refinement, walks through all three review rounds so the design history is legible, and the Tests section reflects the current counts.

**Content-gap concern (zhfeng) — addressed with your preferred fix.** `_pr_review_prompt` now explicitly instructs the agent to fetch a review's inline comments via `gh api repos/{owner}/{repo}/pulls/{number}/reviews/{review_id}/comments` before deciding what to do, guarded so the instruction never renders when `review.id` (or the PR number) is missing. That makes the "agent recovers the inline content from the parent review event" premise true by construction instead of resting on an unenforced assumption about the agent's SOUL.md. Locked with two new tests in `test_github_prompts.py` (one for the fetch-hint text, one for the missing-id guard).

All four changed files (`dispatcher.py`, `prompts.py`, and both test files) are `ruff check` + `ruff format --check` clean. Full suite: 46 passed (`test_github_dispatcher.py`: 31, `test_github_prompts.py`: 15).

WillemJiang (approved):

PR #3280, opened 2026-05-27
fix(chat): preserve messages after summarization

ggnnggez:
## Summary

Fixed #2965 

- Preserve visible chat history when summarization resets the live message stream.
- Match DeerFlow summarization update keys by the stable SummarizationMiddleware.before_model suffix instead of one exact class name.
- Add frontend coverage for hidden control-message dedupe and DeerFlow summarization update keys.
- Add a backend regression guard for the frontend/backend summarization update-key contract.

## Root Cause

The backend emits summarization updates under DeerFlowSummarizationMiddleware.before_model, while the frontend only listened for SummarizationMiddleware.before_model. As a result, the frontend missed the summarization reset update and did not move visible messages into local history before RemoveMessage(__remove_all__) cleared the live stream.

## Validation
```
pnpm test tests/unit/core/threads/message-merge.test.ts
pnpm lint
uv run pytest tests/test_lead_agent_model_resolution.py -k summarization_middleware_preserves_frontend_update_key_contract
uv run pytest tests/test_lead_agent_model_resolution.py -k "create_summarization_middleware"
commit hook: frontend check and backend lint passed
```

Note: pnpm typecheck was previously blocked by stale Next generated route types under .next/types versus .next/dev/types, unrelated to this change.

PR #4132, opened 2026-07-12
chore(threads): extract messageIdentity to shared helper

AnandSundar:
## Summary

Extracts `messageIdentity` from `hooks.ts` into a shared module at `frontend/src/core/threads/message-identity.ts`, so the namespacing logic (`tool:<tool_call_id>` vs `message:<id>`) lives in one place instead of inline.

## What changed

- **New:** `frontend/src/core/threads/message-identity.ts` — the `messageIdentity` helper, lifted from `hooks.ts` byte-identically.
- **Modified:** `frontend/src/core/threads/hooks.ts` — replaces the inline definition with `import { messageIdentity } from "./message-identity";`. All ~18 existing call sites continue to work unchanged.
- **New:** `frontend/tests/unit/core/threads/message-identity.test.ts` — 6 unit tests covering AI message, tool message, tool-with-id precedence, empty input, and the dedup-helper namespace behavior.
- **Removed:** `frontend/src/core/threads/history-utils.ts` and `frontend/tests/unit/core/threads/history-utils.test.ts` — the speculative helpers (`deduplicateHistoryMessages`, `adjustHistoryIndex`) are out of scope for this PR; they would be deferred until a future consumer reconciles with the stronger `dedupeMessagesByIdentity` (`hooks.ts`).

## Why

The same namespacing logic was previously defined inline and was being re-implemented in ad-hoc form across `hooks.ts`. Centralizing it removes the drift risk and makes the helper available to any future thread utility (e.g., a future dedup helper that needs to match `dedupeMessagesByIdentity`'s semantics).


willem-bd (commented):
## Review summary

The two helpers here are well-tested and well-documented, but I think this needs another pass before merging:

1. **Not wired in.** Neither `deduplicateHistoryMessages` nor `adjustHistoryIndex` is imported anywhere in `src/`. The PR title advertises a runtime fix that the body acknowledges already landed on `main` via the `useThreadHistory` refactor — so as written this ships speculative utilities with no consumer.
2. **Redundant with stronger existing helpers.** `dedupeMessagesByIdentity` (`hooks.ts:157`) and `findLatestUnloadedRunIndex` + the recompute at `hooks.ts:1745` already cover both concerns more robustly (namespaced identities, last-occurrence-wins, hidden-message handling, ground-truth recompute).
3. **Namespace-collision bug** in `deduplicateHistoryMessages` — `tool_call_id` and `id` share one `Set`, so a message whose `id` equals a tool's `tool_call_id` is wrongly dropped. The existing `messageIdentity` namespaces these deliberately. A test below codifies the buggy behavior.
4. **`adjustHistoryIndex` is inverted** relative to the actual `desc` (newest-first) runs ordering — `newRunsLength - 1` points at the oldest run, not the newest.

Recommendation: either wire at least one helper into `useThreadHistory` and demonstrate the fix (after fixing the namespace + ordering issues), or reframe as a pure utility PR with an honest title and the bug fixed + the inverse test added. Inline details below.

frontend/src/core/threads/history-utils.ts
```diff
@@ -0,0 +1,49 @@
+import type { Message } from "@langchain/langgraph-sdk";
+
+/**
+ * Deduplicate incoming messages against an existing history.
+ * A message is considered a duplicate if its `id` or `tool_call_id`
+ * (for tool messages) already appears in the existing list.
+ */
+export function deduplicateHistoryMessages(
```
willem-bd: Neither `deduplicateHistoryMessages` nor `adjustHistoryIndex` is imported anywhere in `src/` (grep for both names and the `history-utils` module path returns zero hits), so this PR ships unused utilities. The title implies a runtime fix ("refresh run list after each turn so scroll-to-load history works"), but the body notes that fix already landed on `main` via the `useThreadHistory` refactor.

For this dedup concern specifically, `hooks.ts` already has the stronger `dedupeMessagesByIdentity` (`hooks.ts:157`): it namespaces identities (`tool:${tool_call_id}` vs `message:${id}`), keeps the *last* occurrence so refreshed content replaces stale, and handles hidden/UI-control messages. This helper is a weaker set-difference on the same idea. Unless one of these is wired into `useThreadHistory` to demonstrate the user-facing improvement, I'd hold off landing speculative utilities.

frontend/src/core/threads/history-utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  existing: Message[],
+  incoming: Message[],
+): Message[] {
+  const existingIds = new Set(
+    existing
+      .map((m) => ("tool_call_id" in m ? m.tool_call_id : m.id))
```
willem-bd: The `existingIds` set collapses two distinct identity namespaces into one. The existing `messageIdentity` (`hooks.ts:143`) deliberately separates them as `tool:${tool_call_id}` vs `message:${id}` — exactly to avoid the collision this creates: an incoming AI message whose `id` happens to equal an existing tool message's `tool_call_id` is wrongly dropped here (and vice versa). The test "deduplicates against tool_call_id from existing messages" below actually exercises this and asserts the incorrect result.

Secondary issue — **asymmetry:** for an existing tool message only `tool_call_id` is added to the set (its `id` is dropped), but the incoming-side check tests both `m.id` and `m.tool_call_id`. So a tool message re-arriving with a new `tool_call_id` but the same `id` won't be deduped.

If the helper is kept, namespace the keys (`tool:` / `message:`) like `messageIdentity`, and add a test for the inverse case (AI `id` == a tool `tool_call_id` should NOT be dropped).

frontend/src/core/threads/history-utils.ts
```diff
@@ -0,0 +1,49 @@
...
+ * - If no new runs were added, return `currentIndex` unchanged.
+ */
+export function adjustHistoryIndex(
+  currentIndex: number,
+  prevRunsLength: number,
+  newRunsLength: number,
+): number {
+  const added = newRunsLength - prevRunsLength;
+  if (added <= 0) return currentIndex;
+  if (currentIndex < 0) return newRunsLength - 1;
```
willem-bd: This reset (and the `currentIndex + added` shift below) assumes new runs are *appended* and that "the last run" is the newest. But runs are fetched newest-first (`sortBy: "updated_at", sortOrder: "desc"` — `hooks.ts:1906`, `hooks.ts:2087`), so new runs land at index 0 and `newRunsLength - 1` points at the *oldest* run — the opposite of "scroll up to load new runs." The shift branch has the same inversion (prepended runs should move toward 0, not add).

Also redundant on `main`: `useThreadHistory` already recomputes `indexRef.current` from ground truth via `findLatestUnloadedRunIndex` on every `runs.data` change (`hooks.ts:1745`), so no shift arithmetic is needed. If this gets wired in later, reconcile it with `desc` ordering first.

frontend/tests/unit/core/threads/history-utils.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    {
+      type: "tool",
+      id: "t0",
+      tool_call_id: "tc-x",
+      content: "result",
+      name: "tool",
+    } as unknown as Message,
+  ];
+  // Incoming AI message references the same id — should be filtered
+  const incoming: Message[] = [{ type: "ai", id: "tc-x", content: "response" }];
```
willem-bd: This test locks in the namespace-collision behavior as expected: an incoming **AI** message with `id: "tc-x"` is dropped because an existing **tool** message has `tool_call_id: "tc-x"`. Those are different identity namespaces in the existing `messageIdentity` (`message:tc-x` vs `tool:tc-x`), so the AI message should be kept. Once the helper namespaces its keys, this assertion should flip to `toHaveLength(1)`, and it'd be worth adding the symmetric case (incoming tool whose `tool_call_id` equals an existing AI `id` should also be kept).

AnandSundar:
Thanks @willem-bd for the careful review and for catching the namespace bug — that's exactly the kind of subtle regression that would have been painful to debug downstream.

I went with your Path A: this stays a pure utility PR (no `useThreadHistory` changes). All four points addressed:

1. **Not wired in.** Reframed the title as `chore(threads):` and dropped the runtime-fix framing. `useThreadHistory` on main already handles run-list refresh and dedup via `findLatestUnloadedRunIndex` + `dedupeMessagesByIdentity`, so wiring the helpers in here would either re-implement existing logic or step on the stronger semantics your helpers provide.
2. **Not duplicating your stronger helpers.** Lifted the `messageIdentity` function out of `hooks.ts` into a shared module (`frontend/src/core/threads/message-identity.ts`); both `hooks.ts` and the new `deduplicateHistoryMessages` now import from there. The dedup helper uses the same namespaced `tool:`/`message:` keys your `dedupeMessagesByIdentity` uses.
3. **Namespace fix.** `deduplicateHistoryMessages` no longer collapses tool and message identities into a single set; it keys both sides via `messageIdentity` so cross-namespace collisions are impossible by construction.
4. **Ordering fix.** Runs are newest-first (`updated_at desc`), so new runs prepend at index 0 and the previous newest shifts down by `added`. `adjustHistoryIndex`'s reset branch now returns `added` (the previous newest's new index), with a fallback to 0 for the empty-start edge case. The shift branch (`currentIndex + added`) coincidentally works under both appended and prepended ordering — kept as-is. The doc comment now states the desc-ordering assumption explicitly.

The test that codified the buggy namespace behavior is flipped, and a symmetric inverse case (existing AI `id` colliding with incoming tool `tool_call_id`) is added. Replies inline below.

frontend/src/core/threads/history-utils.ts
AnandSundar: Confirmed — neither helper is wired in. Following your recommendation, this stays a pure utility PR (Path A).

The original title implied a runtime fix; reframed to `chore(threads): add deduplicateHistoryMessages and adjustHistoryIndex utilities`. The `useThreadHistory` refactor already covers the runtime concern on main via `findLatestUnloadedRunIndex` + `dedupeMessagesByIdentity` (the latter also handles last-occurrence-wins and hidden-message preservation, which a straight re-wire of `deduplicateHistoryMessages` would lose), so wiring in here would duplicate logic.

frontend/src/core/threads/history-utils.ts
AnandSundar: Good catch. Fixed by mirroring your `messageIdentity`'s `tool:`/`message:` keys in the dedup helper — and lifting `messageIdentity` out of `hooks.ts` into a shared module so both call sites import the same function.

Both sides now namespace consistently: existing tool contributes only `tool:<tool_call_id>`, existing non-tool contributes only `message:<id>`, and incoming checks the same namespaced key. The asymmetry you flagged (existing tool's `id` dropped from the set, incoming checks `id` against `tool_call_id`-keyed entries) is gone — incoming AI matches `message:<id>`, incoming tool matches `tool:<tool_call_id>`, and cross-namespace collisions cannot happen.

The "should be filtered" test was flipped to assert the AI message is kept, and a symmetric inverse case was added: existing AI with `id: "tc-x"`, incoming tool with `tool_call_id: "tc-x"` → also kept.

frontend/src/core/threads/history-utils.ts
AnandSundar: Confirmed — runs are newest-first (`sortBy: "updated_at", sortOrder: "desc"`), so new runs land at index 0 and the user's previous-newest shifts down by `added`.

Fix in `adjustHistoryIndex`'s reset branch: it now returns `added` (the previous newest's new index) instead of `newRunsLength - 1` (which would have been the oldest run), with a fallback to 0 for the empty-start edge case (`prevRunsLength === 0`). The shift branch (`currentIndex + added`) coincidentally works under both appended and prepended ordering — kept as-is. The doc comment now states the desc-ordering assumption explicitly.

Three affected tests updated: the all-loaded-reset case (now returns `added`), the single-prepend case (now returns 1), and the empty-start case (now returns 0 instead of 2).

frontend/tests/unit/core/threads/history-utils.test.ts
AnandSundar: Flipped. The test now asserts the AI message is kept (because `message:tc-x` ≠ `tool:tc-x` under namespaced keys), which is the correct behavior given the helper's intent.

A symmetric inverse case is also added: existing AI with `id: "tc-x"`, incoming tool with `tool_call_id: "tc-x"` → kept, since `message:tc-x` ≠ `tool:tc-x` in that direction too.

frontend/src/core/threads/history-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Deduplicate incoming messages against an existing history.
+ *
+ * Each message is keyed by a stable per-message identity (see
+ * `messageIdentity`), which namespaces tool messages under `tool:<id>` and
+ * other messages under `message:<id>` so an AI message whose `id` happens to
+ * match a tool's `tool_call_id` is not incorrectly dropped.
+ *
+ * Messages without a usable identity are passed through unchanged.
+ */
+export function deduplicateHistoryMessages(
```
willem-bd: This helper and `adjustHistoryIndex` below currently have no consumer outside their own tests. The PR description notes the original runtime goal is already covered on `main` by `useThreadHistory`'s `findLatestUnloadedRunIndex` + `dedupeMessagesByIdentity`, so these land as speculative utilities.

Two concerns if they ship unwired:

- **Weaker than the live helper.** `deduplicateHistoryMessages` is a strictly weaker twin of `dedupeMessagesByIdentity` in `hooks.ts` — it drops on first identity match and has no last-occurrence-wins or hidden-message preservation. A future maintainer reaching for "dedupe messages" could grab this one and silently lose UI-display semantics. If kept, consider cross-referencing the stronger helper in the JSDoc, or unifying.
- **Existing-vs-incoming only.** It dedupes `incoming` against `existing` but not within `incoming` itself — duplicates inside `incoming` both pass through. Worth stating explicitly in the JSDoc so a consumer doesn't assume full dedup.

Either wire at least one into a real consumer in this PR, or defer these until a use case exists (the `messageIdentity` extraction is the valuable, mergeable part on its own).

willem-bd (commented):
Reviewed the helpers. The `messageIdentity` extraction is a clean, low-risk refactor (centralizes ~18 call sites, byte-identical, live path unchanged) — that part is solid.

Two inline notes below on the two new utilities (`deduplicateHistoryMessages`, `adjustHistoryIndex`): both are currently unwired and the dedupe helper is a weaker twin of the existing `dedupeMessagesByIdentity`. Suggest either wiring one into a real consumer here or deferring the utilities and merging just the extraction.

frontend/src/core/threads/history-utils.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Compute the new history-loading index when the runs list grows.
+ *
+ * Assumes the runs list is ordered **newest-first** (descending by
+ * `updated_at`), so newly-fetched runs are *prepended* at index 0 rather than
+ * appended at the end.
+ *
+ * - `added <= 0` (no new runs): return `currentIndex` unchanged.
+ * - `currentIndex < 0` (all previously-known runs were loaded):
```
willem-bd: The `-1` sentinel is overloaded relative to the live index model this is meant to sit alongside. `findLatestUnloadedRunIndex` in `hooks.ts` returns `-1` to mean "all runs loaded," but here `-1` is treated as "focal point = previous newest run" — the reset branch returns `added` (the old-newest's new position) rather than `0` (the first actually-unloaded new run).

The function is self-consistent and tested against its own contract, so this isn't a bug — but since the stated intent is to complement `findLatestUnloadedRunIndex`, a future consumer could wire them together expecting matching `-1` semantics and be surprised. A JSDoc note clarifying that `currentIndex` here is a focal/scroll index (not the output of `findLatestUnloadedRunIndex`) would close that gap.

AnandSundar:
@willem-bd Thanks for the round-2 read and for being explicit about the weak-twin concern — that's the kind of framing I was missing. Following your recommendation: both helpers (deduplicateHistoryMessages, adjustHistoryIndex) are out of this PR's scope; only the messageIdentity extraction remains. Re-titled to chore(threads): extract messageIdentity to shared helper. If a future consumer ever needs a dedup helper, the right starting point is reconciling it with dedupeMessagesByIdentity (last-occurrence-wins, hidden-message preservation), not reviving the weaker twin. Replies inline below.

frontend/src/core/threads/history-utils.ts
AnandSundar: Accepted — both points land. `deduplicateHistoryMessages` was a strictly weaker twin of `dedupeMessagesByIdentity`, and the existing-vs-incoming-only scope was a real maintenance footgun if a future consumer reached for it expecting full dedup. With the helper dropped from this PR, the weak-twin concern is moot. If a consumer needs this later, the right starting point is to first reconcile semantics with the stronger helper, not to revive this one.

frontend/src/core/threads/history-utils.ts
AnandSundar: Acknowledged — the `-1` sentinel collision was real (`findLatestUnloadedRunIndex`'s `-1` = "all runs loaded"; here `-1` = "focal point = previous newest"). With `adjustHistoryIndex` no longer in this PR, the JSDoc question is moot, but the underlying naming-collision risk is worth flagging if/when these helpers are reintroduced together. Thanks for catching it.

WillemJiang:
The issue is addressed by https://github.com/bytedance/deer-flow/pull/3280

PR #4133, opened 2026-07-12
fix(channels): buffer and drain GitHub comments queued during a busy run

DaoyuanLi2816:
## Summary

Addresses the silent loss of GitHub follow-up comments described in #4121. When a comment arrives while its thread already has an active run, the channel manager buffers it and starts a coalesced follow-up turn after the active run reaches the stream end sentinel.

The buffer is per-thread, deduplicated, capped, and drained in FIFO batches. Failed drains requeue their batch, watcher tasks are cancelled during shutdown, and comment text is escaped before it is placed in the XML-like wrapper. GitHub opts in through `ChannelRunPolicy`; other channels keep their existing behavior.

The buffer remains process-local, so multi-worker coordination is outside this change.

## Tests

- affected channel, lifespan, and tracing suites: 323 passed
- full backend suite on Windows: 9,011 passed, 90 skipped; 70 unrelated platform-dependent failures in POSIX permission, symlink, shell, and path tests
- Ruff check and format check on the changed files


willem-bd (commented):
Thanks @DaoyuanLi2816. I found a few issues that should be addressed before this is ready.

[P2] Drain loses the buffered batch if a pre-`runs.create` step raises

- Location: `backend/app/channels/manager.py:1162-1228` (`_drain_followups_for_thread`) - pop at `:1189`, resolve/apply at `:1193-1194`, requeue only in the `runs.create` `except` at `:1226`
- Problem: The drain pops the batch (`_pop_followup_batch`, line 1189) before `_resolve_run_params` (1193) and `_apply_channel_policy` (1194), and only the `runs.create` call (1212) is wrapped in the `try` that requeues (1226). If `_resolve_run_params` or `_apply_channel_policy` raises - e.g. the agent config was removed mid-run, or channel-policy credential injection fails - the already-popped entries are not requeued and are silently lost, contradicting the "requeued (not lost)" property.
- Evidence: lines 1189-1212 pop, then run resolve+apply uncovered, then `try` only around `client.runs.create`. The `except` at 1214-1227 requeues solely on a `runs.create` failure. A pre-create exception escapes `_drain_followups_for_thread` after the pop, so the batch vanishes from `_followup_buffers` with no follow-up run created.
- Suggested fix: Requeue on any failure in the drain body. Minimal: wrap `_resolve_run_params` / `_apply_channel_policy` / `runs.create` in one `try` whose `except` calls `_requeue_followups(thread_id, entries)`, or defer `_pop_followup_batch` until after resolve+apply succeed.
- Test: a drain where `_resolve_run_params` (or `_apply_channel_policy`) raises, asserting the batch is still in `_followup_buffers[thread_id]`. The existing requeue tests cover only `runs.create` `ConflictError` and `RuntimeError`, not pre-create failures.

[P2] Follow-up watcher tasks are not cancelled on `manager.stop()`

- Location: `backend/app/channels/manager.py:1118` (`asyncio.create_task(self._watch_run_and_drain_followups(...))` in `_maybe_spawn_followup_watcher`, tracked only via `add_done_callback`) and `:1433` (`stop()` cancels only `self._task`)
- Problem: Watcher tasks are fire-and-forget - created at line 1118 with `add_done_callback(self._log_task_error)` but not stored. `stop()` (1433) cancels only `self._task` (the dispatch loop). An in-flight watcher awaiting `stream_bridge.subscribe(run_id)` survives `stop()`; when the run later ends it calls `_drain_followups_for_thread` -> `_get_client()` -> `client.runs.create()` against a stopped manager, firing a follow-up run into a shutting-down app. The watcher also never checks `self._running`. (In the normal lifespan path the event-loop shutdown cancels pending tasks, mitigating this; it bites on any `stop()` that does not tear down the loop, e.g. manager restart or test teardown.)
- Evidence: `_maybe_spawn_followup_watcher` (1090) creates the task at 1118 with only a done-callback - no task set/list. `stop()` (1433-1444) cancels `self._task` only. `_watch_run_and_drain_followups` (1121) has no `self._running` guard before draining.
- Suggested fix: Track watcher tasks in a set (e.g. `self._watcher_tasks`), discard on done-callback, and cancel+await them in `stop()`; and/or guard the drain with `if not self._running: return` (requeuing the popped batch first, per the fix above). Mirrors how `_dispatch_loop` is tracked and cancelled in `stop()`.
- Test: spawn a watcher (run created, subscribe pending), call `manager.stop()`, assert the watcher task is cancelled/awaited and no `runs.create` fires after stop.

zhfeng (approved):
Approving - the core mechanism is sound. I verified the production-path assumption the unit tests can't cover: a fire-and-forget `client.runs.create()` actually emits `END_SENTINEL` to the `StreamBridge` the watcher subscribes to (`runs.create` returns a `Run` TypedDict with `run_id`; `start_run` launches `run_agent` as a background task whose `finally` block calls `bridge.publish_end(run_id)`), and both the memory and Redis backends deliver `END_SENTINEL` to late subscribers, so the subscribe-after-create timing is safe. The two P2s from @willem-bd's earlier review (drain losing the batch on a pre-`runs.create` failure; watchers not cancelled on `stop()`) are correctly fixed with tests.

**Before merge: please rebase onto `main`.** The branch is ~176 commits behind and several touched files have diverged (`app.py`, `manager.py`, `backend/AGENTS.md`, and `test_gateway_lifespan_shutdown.py` — the last will conflict directly, since `main` has since added the memory-flush/warm-up tests that this branch's version predates; the `fake_start(_startup_config, **_kwargs)` change is still needed on `main`). The regression numbers in the PR body were run against the stale base, so please rebase, re-run the full suite, and re-post the before/after against the rebased head.

**Follow-up worth filing (PR or issue):** the buffer knobs are all hardcoded today — `FOLLOWUP_BUFFER_MAX_PER_THREAD` (20), `FOLLOWUP_DRAIN_BATCH_SIZE` (10) are module constants in `manager.py`, and `buffer_followups_on_busy` is a code-level `ChannelRunPolicy` field flipped per-channel in `register_policy()`, not operator-configurable (unlike `recursion_limit`, which is per-agent). That's fine for the GitHub MVP, but it'd be good to make these configurable in a follow-up — e.g. lift them onto `ChannelRunPolicy` sourced from channel config, so a high-volume repo can raise the cap without a code change.

Two other things I noticed while reviewing against #4121, not blocking this merge:
- The buffer dedupes by `delivery_id` (delivery-level — the same layer as #4104) rather than the `comment_id` the issue specified (a comment-level complement to #4104). Same-comment / different-delivery events can therefore be processed twice. The comment id is available in the webhook payload but isn't threaded into `metadata["github"]` — worth reconciling with the issue's layered-dedup intent, either by threading a comment node id through or documenting the delivery-level choice explicitly.
- The `eyes`/`confused` reactions are deferred (knowingly). That's the half of #4121 that actually addresses "the human concludes the bot ignored them," so #4121 should stay open against that follow-up.

DaoyuanLi2816:
Rebased onto current `main`. The affected suites pass 323/323; the full backend run reached 9,011 passed and 90 skipped, with 70 unrelated Windows-only failures in POSIX permission, symlink, shell, and path tests. Ruff is clean.

WillemJiang (approved):

PR #4140, opened 2026-07-13
fix(agents): stop persisting base64 image data in checkpoint state

qin-chenghan:
## Problem

Relates to #4138

The `viewed_images` state field stored full base64-encoded image data. Since LangGraph's checkpointer saves a complete state snapshot at every super-step, this base64 payload was duplicated across every subsequent checkpoint — `O(n × steps)` growth. A single 1MB image viewed early in a conversation would be re-stored in every checkpoint for the rest of the session.

Additionally, `ViewImageMiddleware` injected a `HumanMessage` containing the same base64 data into `messages`, so the image bytes were duplicated **twice** in state: once in `viewed_images`, once in `messages`.

## Solution

Store only lightweight metadata (`mime_type`, `size`, `actual_path`) in the `viewed_images` state field. The `ViewImageMiddleware` reads the image file from disk on-demand in `before_model` and encodes it as base64 for the model call.

The base64 data still appears in the transient `HumanMessage` for the model call, so it's persisted in the checkpoint for **that step**. But the second copy (in `viewed_images`) is eliminated, roughly halving the image bytes in state. Full elimination of the persistent HumanMessage would require checkpoint retention policies, deferred.

### Defense in depth (review feedback addressed)

- **`actual_path`** is server-set by `view_image_tool` after virtual-root validation; client input cannot reach it. Documented in the read helper.
- **`size` mismatch** between recorded value and current disk size causes the middleware to skip the image (TOCTOU guard).
- **`_MAX_IMAGE_BYTES`** re-checked at read time in case the file grew past the cap between view and inject.
- **`abefore_model`** runs the blocking 20MB read + base64 via `asyncio.to_thread`.

## Changes

- **`thread_state.py`**: `ViewedImageData` TypedDict changed from `{base64, mime_type}` to `{mime_type, size, actual_path}`
- **`view_image_tool.py`**: No longer base64-encodes image data; stores only metadata in state. Removed unused `base64` import. Added TOCTOU guard for stat-vs-read size mismatch.
- **`view_image_middleware.py`**: Extracted `_read_image_as_data_url` helper. `_create_image_details_message` reads image files from disk on-demand using `actual_path`, encodes base64 temporarily for the model, and skips files that fail the size-mismatch or size-cap checks. `abefore_model` runs the blocking work via `asyncio.to_thread`.
- **Tests**: All tests updated to use the new metadata-only format. Added coverage for OSError during read, file-changed-since-view (TOCTOU), and size-exceeds-cap branches.

## What this does NOT fix (follow-up items from #4138)

- The injected `HumanMessage` (with base64 content) still appears in the checkpoint for the step where it was injected. Full dedup requires checkpoint retention policies.
- Large tool result deduplication
- Checkpoint pruning / TTL
- Storage abstraction for multi-gateway instances (per @WillemJiang's comment — this PR uses real filesystem `actual_path`; a follow-up would lift that to S3 / content-addressed storage so multi-instance deployments don't break)

## Testing

- [x] `make lint` passes (ruff check + ruff format, CI-equivalent)
- [x] All tests in `test_view_image_middleware.py`, `test_view_image_tool.py`, `test_sandbox_middleware.py`, `test_thread_state_reducers.py`, `test_create_deerflow_agent.py` pass
- [x] CI green: backend-blocking-io, backend-unit-tests, lint-backend, lint-frontend, frontend-unit-tests, replay e2e layers 1+2

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
@@ -1,6 +1,8 @@
 """Middleware for injecting image details into conversation before LLM call."""
 
+import base64
 import logging
+from pathlib import Path
 from typing import override
 
 from langchain.agents.middleware import AgentMiddleware
@@ -94,6 +96,12 @@ class ViewImageMiddleware(AgentMiddleware[ViewImageMiddlewareState]):
     def _create_image_details_message(self, state: ViewImageMiddlewareState) -> list[str | dict]:
         """Create a formatted message with all viewed image details.
 
+        Reads image files from disk on-demand and encodes them as base64
+        for the model. The base64 data is NOT persisted in state -- only
+        lightweight metadata (path, mime_type, size) is stored in
+        ``viewed_images``, avoiding large duplicate payloads across every
+        checkpoint (see #4138).
+
         Args:
             state: Current state containing viewed_images
 
@@ -110,19 +118,29 @@ class ViewImageMiddleware(AgentMiddleware[ViewImageMiddlewareState]):
 
         for image_path, image_data in viewed_images.items():
             mime_type = image_data.get("mime_type", "unknown")
-            base64_data = image_data.get("base64", "")
+            actual_path = image_data.get("actual_path", "")
 
             # Add text description
             content_blocks.append({"type": "text", "text": f"\n- **{image_path}** ({mime_type})"})
 
-            # Add the actual image data so LLM can "see" it
-            if base64_data:
-                content_blocks.append(
-                    {
-                        "type": "image_url",
-                        "image_url": {"url": f"data:{mime_type};base64,{base64_data}"},
-                    }
-                )
+            # Read the image file on-demand and encode as base64 for the model
+            if actual_path:
+                try:
+                    file_path = Path(actual_path)
+                    if file_path.exists() and file_path.is_file():
+                        with open(file_path, "rb") as f:
+                            image_bytes = f.read()
+                        base64_data = base64.b64encode(image_bytes).decode("utf-8")
+                        content_blocks.append(
+                            {
+                                "type": "image_url",
+                                "image_url": {"url": f"data:{mime_type};base64,{base64_data}"},
+                            }
+                        )
+                    else:
+                        content_blocks.append({"type": "text", "text": f"  (file no longer available: {actual_path})"})
+                except OSError as e:
+                    content_blocks.append({"type": "text", "text": f"  (error reading file: {e})"})
 
         return content_blocks
 
--- a/backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
@@ -1,5 +1,6 @@
 """Middleware for injecting image details into conversation before LLM call."""
 
+import asyncio
 import base64
 import logging
 from pathlib import Path
@@ -12,6 +13,11 @@ from langgraph.runtime import Runtime
 from deerflow.agents.thread_state import ThreadState
 
 logger = logging.getLogger(__name__)
+
+# Mirror the tool-side size cap as a defense-in-depth check. The tool
+# enforces this at write time; the middleware re-checks at read time in
+# case the file grew on disk between view and injection.
+_MAX_IMAGE_BYTES = 20 * 1024 * 1024
 
 
 class ViewImageMiddlewareState(ThreadState):
@@ -93,6 +99,34 @@ class ViewImageMiddleware(AgentMiddleware[ViewImageMiddlewareState]):
         # Check if all tool calls have been completed
         return tool_call_ids.issubset(completed_tool_ids)
 
+    @staticmethod
+    def _read_image_as_data_url(actual_path: str, mime_type: str, expected_size: int) -> str | None:
+        """Read image file and return a `data:` URL, or None on failure.
+
+        Trust assumption: ``actual_path`` is set by ``view_image_tool``
+        (server-side, validated against the allowed virtual roots at write
+        time) and held in LangGraph-controlled state. Client input cannot
+        reach this field, so the read scope is trusted. We still re-check
+        size at read time to defend against TOCTOU growth and skip files
+        exceeding ``_MAX_IMAGE_BYTES``.
+        """
+        try:
+            file_path = Path(actual_path)
+            if not file_path.exists() or not file_path.is_file():
+                return None
+            current_size = file_path.stat().st_size
+            if current_size != expected_size:
+                # File changed between view and inject - skip.
+                return None
+            if current_size > _MAX_IMAGE_BYTES:
+                return None
+            with open(file_path, "rb") as f:
+                image_bytes = f.read()
+            base64_data = base64.b64encode(image_bytes).decode("utf-8")
+            return f"data:{mime_type};base64,{base64_data}"
+        except OSError:
+            return None
+
     def _create_image_details_message(self, state: ViewImageMiddlewareState) -> list[str | dict]:
         """Create a formatted message with all viewed image details.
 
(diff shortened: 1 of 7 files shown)


backend/packages/harness/deerflow/tools/builtins/view_image_tool.py
```diff
@@ -151,11 +150,18 @@ def view_image_tool(
             update={"messages": [ToolMessage(f"Error: Image contents are {detected_mime_type}, but file extension indicates {expected_mime_type}", tool_call_id=tool_call_id)]},
         )
     mime_type = detected_mime_type
-    image_base64 = base64.b64encode(image_data).decode("utf-8")
-
-    # Update viewed_images in state
-    # The merge_viewed_images reducer will handle merging with existing images
-    new_viewed_images = {image_path: {"base64": image_base64, "mime_type": mime_type}}
+    image_size = len(image_data)
```
willem-bd: Two cleanup nits in this region:

1. The comment `# Read image file and convert to base64` (~15 lines above, above the `with open(...)` read) is now stale -- this PR removed the base64 conversion that justified it, so the read is now only for magic-byte validation. Suggest updating it, e.g. `# Read image file to validate contents`.

2. `image_size` is already computed earlier via `path.stat().st_size` (for the `_MAX_IMAGE_BYTES` check), so this `len(image_data)` reassignment is redundant -- the two are equal after a successful read. Reuse the earlier value.

backend/packages/harness/deerflow/agents/thread_state.py
```diff
@@ -18,8 +18,17 @@ class ThreadDataState(TypedDict):
...
+    """Metadata for a viewed image file.
+
+    Only lightweight metadata is persisted in checkpoint state; the actual
+    image bytes are read on-demand from disk when the model needs them.
+    This avoids duplicating large base64 payloads across every checkpoint
+    (see #4138).
+    """
+
     mime_type: str
+    size: int
```
willem-bd, line 30: `size` is persisted in checkpoint state but never read -- `ViewImageMiddleware` only consumes `mime_type` and `actual_path`. For a PR whose goal is *reducing* checkpoint bloat, storing unused metadata is a (tiny) step backward.

Either drop it, or put it to use: the middleware re-reads the file with no size guard, so a file that grew between view and injection would be read and base64-encoded uncapped. Re-checking `size` (or `path.stat().st_size`) against `_MAX_IMAGE_BYTES` before `open()` would be a sensible defense-in-depth use.

backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
```diff
@@ -110,19 +118,29 @@ def _create_image_details_message(self, state: ViewImageMiddlewareState) -> list
 
         for image_path, image_data in viewed_images.items():
             mime_type = image_data.get("mime_type", "unknown")
-            base64_data = image_data.get("base64", "")
+            actual_path = image_data.get("actual_path", "")
```
willem-bd, line 155: `actual_path` is a real server filesystem path now persisted in checkpoint state (previously only the virtual path was in state). The tool validates it's under the allowed virtual roots before storing it, but the middleware `open()`s it below with no scope re-check.

Since `actual_path` is set by trusted tool code and state is server-controlled the practical risk is low -- but for a `risk:high` change it's worth either re-validating the path scope here, or adding a one-line comment documenting the trust assumption (no client-settable path can reach this field).

backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
```diff
@@ -110,19 +118,29 @@ def _create_image_details_message(self, state: ViewImageMiddlewareState) -> list
...
-                        "type": "image_url",
-                        "image_url": {"url": f"data:{mime_type};base64,{base64_data}"},
-                    }
-                )
+            # Read the image file on-demand and encode as base64 for the model
+            if actual_path:
+                try:
+                    file_path = Path(actual_path)
+                    if file_path.exists() and file_path.is_file():
+                        with open(file_path, "rb") as f:
```
willem-bd: This synchronous `read()` + `base64.b64encode()` on up to 20 MB also runs through `abefore_model` -> `_inject_image_message`, i.e. on the async event-loop path, where it blocks the loop for the full read+encode duration.

Not a merge blocker (once per image view), but if it matters for large images, `abefore_model` could dispatch the read via `asyncio.to_thread`.

willem-bd (commented):
Reviewed PR head `a08d6b7f`.

The core fix is correct and well-scoped: removing base64 from `viewed_images` eliminates the avoidable `O(steps)` checkpoint duplication, while the remaining duplication in `messages` is inherent to the checkpoint model and correctly deferred. Backward compat is solid -- verified `viewed_images` has no frontend/API consumer, the reducer is a plain dict merge, and `.get("actual_path", "")` degrades old checkpoints gracefully. Tests genuinely exercise the on-demand read path via the new `_make_viewed_image` helper.

One accuracy note on the PR description: the Solution says the base64 "lives only in the transient `HumanMessage` for that step, not in persistent state," but that `HumanMessage` is appended to `messages` and is therefore re-snapshotted into every subsequent checkpoint. The real win is removing the *second* copy (the `viewed_images` field), not eliminating persistence -- worth rewording so future readers understand the residual duplication.

Inline nits/suggestions below. None are merge blockers; the stale-comment fix and the `size` field cleanup are quick wins for this change set.

backend/tests/test_view_image_middleware.py
```diff
@@ -197,26 +210,31 @@ def test_builds_blocks_for_multiple_images(self):
...
+                    "size": 0,
+                    "actual_path": str(tmp_path / "nonexistent.png"),
+                },
             }
         }
         blocks = mw._create_image_details_message(state)
-        # header + description only (no image_url since base64 is empty)
-        assert len(blocks) == 2
+        # header + description + error text (file no longer available)
+        assert len(blocks) == 3
```
willem-bd, line 230: This covers the "file no longer available" branch. The sibling `except OSError` branch in `_create_image_details_message` is uncovered -- a test that forces a read error (e.g. monkeypatch `builtins.open` to raise `OSError`) would close the gap. Low priority since the branch is simple.

WillemJiang:
@qin-chenghan, we may hit the issue if the gateway has multiple instances. It's better that the file path supports the S3 storage.

qin-chenghan:
@willem-bd thanks for the thorough review — all five nits addressed in 730f3184:

1. **Stale comment** — reworded to .
2. **__TEXT	__DATA	__OBJC	others	dec	hex field** — kept (as you suggested) and put to use as a TOCTOU + cap guard in the middleware. Files whose current size differs from the recorded __TEXT	__DATA	__OBJC	others	dec	hex, or that exceed , are skipped.
3. ** trust** — added a docstring on the new  helper documenting that this field is server-set after virtual-root validation and not client-settable.
4. **Async event loop** —  now runs the read+encode through , separate path from the sync .
5. **OSError branch** — added  covering the missing-except branch (plus two more for TOCTOU and size-cap).

PR description also rewritten to make the residual duplication explicit (the HumanMessage still carries base64 for one step — the win is removing the *second* copy in ).

qin-chenghan:
@willem-bd thanks for the thorough review — all five nits addressed in 730f3184:

1. **Stale comment** — reworded to `# Read image file to validate contents`.
2. **`size` field** — kept (as you suggested) and put to use as a TOCTOU + cap guard in the middleware. Files whose current size differs from the recorded `size`, or that exceed `_MAX_IMAGE_BYTES`, are skipped.
3. **`actual_path` trust** — added a docstring on the new `_read_image_as_data_url` helper documenting that this field is server-set after virtual-root validation and not client-settable.
4. **Async event loop** — `abefore_model` now runs the read+encode through `asyncio.to_thread`, separate path from the sync `before_model`.
5. **OSError branch** — added `test_omits_image_url_when_read_raises_oserror` covering the missing-except branch (plus two more for TOCTOU and size-cap).

PR description also rewritten to make the residual duplication explicit (the HumanMessage still carries base64 for one step — the win is removing the *second* copy in `viewed_images`).

qin-chenghan:
@WillemJiang good point — agreed this PR only solves the duplication half; the storage layer is a separate concern. With `actual_path` as a real filesystem path today, multi-instance deployments would break the middleware's on-demand read.

Logged as follow-up work (not addressed here): lift `actual_path` to a content-addressed ref (sha256 + bucket key) backed by S3 / shared storage so all gateway instances can resolve the same image. That change would touch this PR's contract — would prefer to land this checkpoint-bloat fix first, then design the storage abstraction in a separate PR.

Happy to open an issue tracking the storage-abstraction work if helpful.

WillemJiang (approved):

WillemJiang:
> @WillemJiang good point — agreed this PR only solves the duplication half; the storage layer is a separate concern. With `actual_path` as a real filesystem path today, multi-instance deployments would break the middleware's on-demand read.
> 
> Logged as follow-up work (not addressed here): lift `actual_path` to a content-addressed ref (sha256 + bucket key) backed by S3 / shared storage so all gateway instances can resolve the same image. That change would touch this PR's contract — would prefer to land this checkpoint-bloat fix first, then design the storage abstraction in a separate PR.
> 
> Happy to open an issue tracking the storage-abstraction work if helpful.

@qin-chenghan, please create an issue to track it.

PR #4143, opened 2026-07-13
feat(memory): LLM-assigned per-fact expected_valid_days and staleFactsToExtend

sontianye:
## Why

This is a re-open of this same PR. The original branch was force-pushed clean by accident (reset to plain `main` with no diff), which made GitHub auto-close it; the design itself was never lost. This re-open re-ports the feature onto the current `main`, which has since merged the pluggable-memory refactor **#4122**.

**What happened with #4122:** that refactor moved the memory staleness logic from the single-file `agents/memory/updater.py` into the self-contained DeerMem backend (`backends/deermem/deermem/core/`), introduced a `MemoryManager` abstraction with pluggable backends, and switched `MemoryUpdater` to dependency injection `(config, storage, llm)` with staleness config now living on `DeerMemConfig` (parsed from `backend_config`), not the host `MemoryConfig`. The refactor was based on the pre-feature code snapshot, so it carried the **old global-threshold staleness** version and the per-fact lifetime work from the original #4143 was not present. This PR re-applies that work against the new module layout and DI shape. There is no conflict to resolve manually — the code is written fresh against the new structure.

The staleness review system (#3860, KEEP/REMOVE pass over aged facts) solves *silent staleness* for globally-old facts, but all facts share one `staleness_age_days` threshold, which is too coarse:

- "Working on a memory leak in the auth service" should expire in days, not months.
- "Native language: English" should survive years without review.
- There is no way for the LLM to recalibrate a window it set too conservatively at creation time.

This PR makes fact lifetimes fully agentic: the model sets a per-fact review window at creation and can recalibrate it later during staleness review.

## What changed

**New field `expected_valid_days` on facts (creation)**
When storing a new fact, the LLM may supply an `expected_valid_days` integer. The prompt exposes five tiers (≤14 d transient → >365 d very stable). The value is capped at write time by `staleness_age_days × staleness_max_lifetime_multiplier` (default 20.0 → 1800 d ≈ 5 years; range 1.0–100.0) so the model cannot assign an initial lifetime so long the fact is never re-evaluated. The default 20.0 makes the "> 365 d very stable" tier achievable out of the box — with the earlier 3.0 default a 3650-day assignment was silently clamped to 270 d.

**New operation `staleFactsToExtend` (review)**
During the staleness review pass, the LLM can emit extension entries for facts it decides to KEEP but whose window seems miscalibrated:
```json
{ "id": "fact_id", "extend_by_days": 365, "reason": "core skill, originally mis-tagged as volatile" }
```
Extension formula: `min(days_since_created + extend_by_days, staleness_max_extension_days)`. Extensions use an **absolute** ceiling (`staleness_max_extension_days`, default 3650 d ≈ 10 years; range 90–36500) rather than the creation multiplier — they are deliberate review decisions that must be able to advance the window beyond the initial cap, but the absolute bound prevents `timedelta` overflow (a model-supplied `extend_by_days` of 10⁹ previously crashed every later candidate-selection pass with `OverflowError`) and guards against LLM misfire.

**Staleness review annotation updated**
Each stale-fact entry now shows `valid:Nd` (the stored review window) instead of the global age threshold, so the LLM can calibrate conservatism per fact. `_build_staleness_section` takes `config` again (was temporarily `(candidates, age_days: int)` during the #4122 migration).

**Invariant correctness (carried over from the original review)**
| # | Bug | Fix |
|---|---|---|
| 1 | Read-time cap in `_effective_fact_staleness_age` made extensions permanently ineffective for old facts | Removed read-time cap; cap applied at write-time only |
| 2 | `extendable_ids` was computed from the trimmed removal set; proposed-removal facts could escape to extensions when the per-cycle cap reduced actual deletions | Hoist `proposed_remove_ids` (full proposed set) and use it for exclusion |
| 3 | `extend_by` checked `> 0` before `int()` coercion, so a fractional `0.9` passed the float check then truncated to a silent zero-delta extension | Coerce to `int` before the `> 0` guard |
| 4 | `.days` truncation lost up to ~1 day of precision | `int(total_seconds() // 86400)` |
| 5 | staleness-section `html.escape` used `quote=True`, emitting `&quot;` inconsistent with the `prompt.py` `quote=False` convention | `quote=False` — only `<`, `>`, `&` break element-text structure |

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [x] **Agents / LangGraph** — DeerMem backend memory updater, memory-update prompt and staleness-review prompt
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [x] **Default behavior change** — facts now receive a `valid:Nd` annotation in the staleness-review section; new `expected_valid_days` field appears on newly stored facts; two new `backend_config` fields (`staleness_max_lifetime_multiplier`, `staleness_max_extension_days`)

`config_version` is **not** bumped: the new fields live under `backend_config` (a passthrough dict each backend self-interprets), and #4122 set the precedent of adding `backend_config` fields without a version bump. The version check (`scripts/check_config_version.sh`) only verifies the Helm chart version ≥ example version (both 26) and "config_version gates no runtime behavior; it only drives the outdated-warning."

## Screenshots / Recording

N/A — backend-only change with no UI surface.

## Validation

```
# ruff: All checks passed
# memory tests: 195 passed, 1 skipped
# full suite: 890 passed (1 pre-existing unrelated failure in test_checkpointer
#   caused by a stale local config.yaml v15 selecting sqlite checkpointer —
#   fails on plain main too, not introduced by this PR)
```

`test_memory_staleness_review.py` was module-level skipped by #4122 ("full unit-test migration is a follow-up"). This PR performs that migration: DI construction via `(DeerMemConfig, _FakeStorage)`, `_build_staleness_section` back to the `(candidates, config)` signature, plus new coverage for per-fact selection, EXTEND with the absolute cap, the overflow next-cycle regression, the proposed-removal-not-extendable case, fractional `extend_by` skipping, and the creation-time cap. 67 tests in that file, all green.

## AI assistance

**How you used it:** Pair-programming — I designed and implemented the staleness review system (#3860) and prior memory features (guaranteed injection #3592, consolidation #3996, html-escape #4028). Claude Code assisted with the per-fact `expected_valid_days` and `staleFactsToExtend` additions, the invariant fixes, and re-porting the work onto the #4122 backend layout + migrating the skipped tests.


config.example.yaml
```diff
@@ -15,7 +15,7 @@
 # ============================================================================
 # Bump this number when the config schema changes.
 # Run `make config-upgrade` to merge new fields into your local config.yaml.
-config_version: 22
+config_version: 23
```
willem-bd: **🔴 Blocking — CI will fail.** The `chart.yaml` workflow runs `validate-chart` on PRs that touch `config.example.yaml`, and its `config_version drift check` (`scripts/check_config_version.sh`) requires the chart's embedded version to be `>=` this one. This PR bumps the example to 23 but leaves:
- `deploy/helm/deer-flow/values.yaml:224` at `22`
- `deploy/helm/deer-flow/README.md:106` at `22`

So the check fails: `chart config_version (22) is behind config.example.yaml (23).` (`nightly.yaml` runs the same check.) Bump both to 23.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -1009,48 +1065,105 @@ def _apply_updates(
...
+                            extend_by = ext.get("extend_by_days")
+                            if isinstance(extend_by, int) and not isinstance(extend_by, bool) and extend_by > 0:
+                                created = _parse_fact_datetime(fact.get("createdAt", ""))
+                                if created is None:
+                                    # Unreachable: _select_stale_candidates already
+                                    # excludes facts with unparseable createdAt.
+                                    updated_facts.append(fact)
+                                    continue
+                                days_since = (now_utc - created).days
+                                new_evd = days_since + extend_by
```
willem-bd: **🟡 Unbounded extension is a footgun.** Extensions bypass the creation cap with *no* upper bound — `extend_by_days: 9999` is accepted (and `test_extension_not_capped_by_multiplier` asserts exactly that), setting `expected_valid_days` to ~10000 days (~27 years). The creation path is capped precisely to prevent indefinite deferral, yet a single LLM misfire on a "this looks stable" judgment can permanently bury a fact that later changes.

Consider a safety net, e.g. a configurable `staleness_max_extension_days` (or a multiple of the creation cap), so deliberate review decisions can still be generous but not unbounded. (Also: the extension records the recalibration only via `logger.info`; the fact itself carries no `lastReviewedAt`/`updatedAt`, so there's no on-fact audit trail of when/why a window was moved.)

backend/packages/harness/deerflow/agents/memory/prompt.py
```diff
@@ -75,6 +75,14 @@
...
 - Include proper nouns (company names, project names, technology names)
 - Preserve technical terminology and version numbers
+- Fact lifetime (``expected_valid_days``, optional integer):
+  How many days before this fact should be reviewed for possible removal.
+  The system schedules review automatically; omit when uncertain.
+  * ≤ 14: highly transient — active bugs, immediate tasks, today's focus
+  * 15–60: short-term — current experiments, in-progress side projects, near-term goals
+  * 60–180: medium-term — current role, active tech stack, ongoing preferences
+  * 180–365: stable — professional background, established working patterns
+  * > 365: very stable — core skills, native language, personality traits
```
willem-bd: **🟡 Prompt guidance contradicts the default cap.** The prompt invites the LLM to assign `> 365` days for "very stable" facts, but the default creation cap is `staleness_age_days × staleness_max_lifetime_multiplier = 90 × 3.0 = 270`, so a `3650`-day assignment is silently clamped to `270` (in `_apply_updates`). The LLM gets no feedback that its intent was overridden.

This undercuts the PR's own motivation ("Native language: English should survive years without review") — such a fact now needs several review cycles to grow its window. Either raise the default multiplier, or note the cap in the prompt so the model isn't surprised when its 3650-day value becomes 270.

backend/packages/harness/deerflow/agents/memory/prompt.py
```diff
@@ -175,6 +185,13 @@
 The reason should cite what signal in the conversation (or absence thereof)
 supports the removal.
 
+Optionally, for facts you KEEP and wish to recalibrate, add them to
```
willem-bd: **🟡 Header undersells EXTEND.** The section header above says `For each fact, decide KEEP or REMOVE:`, but EXTEND is introduced only here, further down. `backend/AGENTS.md` and `config.example.yaml` describe the decision as "KEEP, REMOVE, or EXTEND." Suggest changing the header to `decide KEEP, REMOVE, or EXTEND` so all three options are visible up front.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -1009,48 +1065,105 @@ def _apply_updates(
...
+                        if ext is not None:
+                            extend_by = ext.get("extend_by_days")
+                            if isinstance(extend_by, int) and not isinstance(extend_by, bool) and extend_by > 0:
+                                created = _parse_fact_datetime(fact.get("createdAt", ""))
+                                if created is None:
+                                    # Unreachable: _select_stale_candidates already
+                                    # excludes facts with unparseable createdAt.
+                                    updated_facts.append(fact)
+                                    continue
+                                days_since = (now_utc - created).days
```
willem-bd: **🟢 Nit — truncation.** `.days` truncates toward zero, so the realized "next review in N days" lands up to ~1 day short of the requested `extend_by`. Negligible for large extensions; for completeness, `int((now_utc - created).total_seconds() // 86400)` would be marginally more precise. Not blocking.

backend/tests/test_memory_staleness_review.py
```diff
@@ -719,3 +724,500 @@ def test_staleness_section_omitted_when_disabled(self):
...
+                staleness_max_lifetime_multiplier=10.0,
+            ),
+        ):
+            result = updater._apply_updates(current_memory, update_data)
+
+        assert len(result["facts"]) == 1
+        fact = result["facts"][0]
+        assert "expected_valid_days" in fact
+        # days_since ≈ 100; new_evd = 100 + 365 (extensions are uncapped)
+        assert fact["expected_valid_days"] == 100 + 365
```
willem-bd: **🟢 Nit — latent timing flakiness.** This asserts `expected_valid_days == 100 + 365`, where `100` derives from `(now - created).days` and `created` was set to `now - 100 days` moments earlier in `_make_fact`. If the test crosses a UTC midnight boundary between `_make_fact` and `_apply_updates` (a millisecond window), `days_since` flips to `101` and the assertion fails. Extremely unlikely, but asserting a small range (or freezing the clock) would be more robust.

backend/packages/harness/deerflow/agents/memory/updater.py
```diff
@@ -1009,48 +1065,105 @@ def _apply_updates(
...
+        _has_staleness_ops = (isinstance(stale_removals, list) and stale_removals) or (
+            isinstance(stale_extensions, list) and stale_extensions
+        )
+        if _has_staleness_ops:
             # Guard against legacy / hand-edited facts that predate the id
             # field: an aged, non-protected fact with no "id" is a valid
             # staleness candidate but has no id to intersect against, so skip
             # it here instead of raising KeyError (id-less facts can never be
             # targeted by the id-based removal set anyway).
             candidate_ids = {f["id"] for f in _select_stale_candidates(current_memory, config) if f.get("id") is not None}
```
willem-bd, line 971: **🟢 Nit — redundant candidate scan.** `_select_stale_candidates` is already called in `_prepare_update_prompt` to build the staleness section, and is called again here to compute `candidate_ids`. The recompute is *correct* (regular `factsToRemove` may have mutated `current_memory` in between), just an extra O(n) pass on large fact stores. Could thread the candidate set through or cache. Not blocking.

backend/packages/harness/deerflow/agents/memory/updater.py
fancyboi999: [P2] Bound the lifetime before persisting it

`extend_by_days` accepts any positive Python integer, and this stores `days_since + extend_by` without a representable upper bound. On the next memory update, `_select_stale_candidates` passes that value to `timedelta(days=effective_age)`, which raises `OverflowError` for sufficiently large values.

I reproduced this on `0d043616` with `extend_by_days=10**20`: the first pass stored `100000000000000000100`; the next candidate selection failed with `OverflowError: Python int too large to convert to C int`. Since this value comes from model output, please reject or clamp it to a representable policy bound before persistence and add a next-cycle regression test. Otherwise one malformed review response can make later memory updates fail deterministically.

fancyboi999 (changes requested):
One P2 correctness issue: an unbounded model-provided lifetime can be persisted and then deterministically crash every later staleness-selection pass. The inline comment includes a two-cycle reproduction.

Separately, current CI also shows backend formatting drift and the required Helm/README config-version bump; those are already explicit in the failing checks, so I did not duplicate them as review findings.

config.example.yaml
sontianye: Fixed in the follow-up commit — both `deploy/helm/deer-flow/values.yaml` and `deploy/helm/deer-flow/README.md` are now at `config_version: 23`. Thanks for catching this.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Good point on both counts — fixed in the follow-up commit.

**Absolute ceiling via `staleness_max_extension_days`**
Added a new config field (`default=3650`, range 90–36500) applied at write time:
```python
new_evd = min(days_since + int(extend_by), config.staleness_max_extension_days)
```
An absolute ceiling rather than the multiplier because extensions must be able to advance the window *beyond* the initial creation cap — that's their whole purpose. A relative cap (e.g. `staleness_age_days × multiplier`) would reintroduce the exact bug this PR fixed: for a fact already older than the cap, any extension result ≤ cap collapses back and the fact is immediately stale again next cycle.

**Audit trail**
You're right that `lastReviewedAt` / `updatedAt` on the fact itself would be a cleaner audit trail than log lines. That's a useful follow-up, but it changes the fact schema in a broader way than fits here. The `logger.info` line at least makes the recalibration visible in structured logs. Happy to open a follow-up issue if that's helpful.

backend/packages/harness/deerflow/agents/memory/prompt.py
sontianye: Agreed — the silent clamp was a real coherence problem. Fixed in the follow-up commit in two ways:

1. **Raised the default multiplier from 3.0 → 20.0** (90 × 20 = 1800 d ≈ 5 years; range widened to 1.0–100.0). The "> 365 d: very stable" tier is now achievable with the default config without needing multiple review cycles to escape the cap.
2. **Added a note to the prompt** after the tier list: *"Assign the value that semantically fits; values above the server-configured ceiling are silently reduced on storage."* — so the model is aware the cap exists even if it can't know the exact configured value.

backend/packages/harness/deerflow/agents/memory/prompt.py
sontianye: Fixed — changed to `For each fact, decide KEEP, REMOVE, or EXTEND:`.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Fixed — changed to `int((now_utc - created).total_seconds() // 86400)`.

backend/tests/test_memory_staleness_review.py
sontianye: Fixed — replaced the exact equality check with `abs(fact["expected_valid_days"] - (100 + 365)) <= 1` to tolerate a ±1 day difference at UTC midnight boundaries.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Good observation — the recompute is intentional here. The regular `factsToRemove` block runs *before* the staleness block and can shrink `current_memory["facts"]` in the same `_apply_updates` call, so a `candidate_ids` computed from the pre-removal memory would include IDs that no longer exist. Threading the set through would require returning it from `_prepare_update_prompt` (which has a different caller contract) and add non-trivial wiring. Happy to leave it as a known optimization opportunity with a comment if you'd prefer that over the current state, but agree it's not blocking.

backend/packages/harness/deerflow/agents/memory/updater.py
sontianye: Confirmed — reproduced the same `OverflowError` on the original commit. Fixed in the follow-up commit with a new `staleness_max_extension_days` config field (default 3650 d, range 90–36500) applied at write time:

```python
new_evd = min(days_since + int(extend_by), config.staleness_max_extension_days)
```

This keeps every stored value well within `timedelta.max.days` regardless of what the model returns. Also added `test_huge_extend_by_days_does_not_overflow_next_cycle` as the next-cycle regression test you requested — it stores a fact extended by `10**9` days and then runs `_select_stale_candidates` to verify no `OverflowError` is raised. Thanks for the clear reproduction steps.

WillemJiang:
@sontianye, please fix the CI test errors.

sontianye:
@WillemJiang I looked into the two failing frontend checks. Both `frontend-unit-tests` and `lint-frontend` fail at the **"Getting action download info"** setup step — before any code is checked out — with `Service Unavailable` (GitHub retried 3× then gave up). The runner never reached our code; in the same run all 6 backend jobs went green, and this PR doesn't touch `frontend/` at all, so the failures are a transient GitHub Actions platform issue, not something in this change.

I pushed an empty commit (`5f50933`) to retrigger them. The new run is currently showing `action_required` — it looks like the workflow needs a maintainer approval to run for this fork branch. Could you re-run / approve the frontend jobs when you have a moment? I don't have admin rights on the repo to trigger it myself.

I also updated the PR description to match the latest commit: the extension path now has an absolute ceiling (`staleness_max_extension_days`, default 3650 d) and the creation multiplier default is 20.0 (cap ≈ 1800 d), so the "> 365 d very stable" prompt tier is achievable out of the box.

WillemJiang:
@sontianye  there are some unit test errors around tests/test_memory_staleness_review.py.  Please take a look at it.
BTW, we just merged the Memory Manager refactoring PR https://github.com/bytedance/deer-flow/pull/4122 into the main branch; you may need to reapply it to the latest branch to resolve the conflicts.

sontianye:
Re-opening this PR - a recap of what happened and what changed since the last review.

**Why it closed:** the original branch was force-pushed to plain `main` by accident (no diff), so GitHub auto-closed it. The design was never lost. This re-open re-ports the same feature onto current `main`.

**The #4122 factor:** since the last review, #4122 (pluggable-memory refactor) landed on `main`. It moved the staleness logic into the self-contained DeerMem backend (`backends/deermem/deermem/core/`), introduced a `MemoryManager` abstraction, and switched `MemoryUpdater` to DI `(config, storage, llm)` with staleness config now on `DeerMemConfig` (parsed from `backend_config`) rather than the host `MemoryConfig`. The refactor was based on the pre-feature snapshot, so it carried the old global-threshold staleness and the per-fact lifetime work was absent. This PR re-applies that work against the new layout - written fresh, no manual conflict resolution needed.

**All prior review feedback is addressed in the port** (same fixes, new file locations):
- `staleness_max_extension_days` absolute ceiling (default 3650 d) - fixes the unbounded-extension footgun and the P2 `OverflowError` crash from `extend_by_days=10**9` (@fancyboi999). Includes the next-cycle regression test.
- `staleness_max_lifetime_multiplier` default raised 3.0 -> 20.0 so the "> 365 d very stable" prompt tier is achievable; prompt notes the server ceiling.
- Staleness header now `KEEP, REMOVE, or EXTEND`.
- `.days` -> `total_seconds() // 86400`; flaky timing assertion -> `abs(...) <= 1`.
- Bonus (self-audit): `extend_by` coerced to `int` before the `> 0` guard (a fractional 0.9 would have silently written a zero-delta extension); staleness-section `html.escape` switched to `quote=False` to match the `prompt.py` convention.

`config_version` is intentionally **not** bumped - the new fields are under `backend_config` (a passthrough dict), and #4122 itself added `backend_config` fields without a bump.

Also migrated `test_memory_staleness_review.py`, which #4122 had module-level skipped ("full unit-test migration is a follow-up") - that follow-up is done here (67 tests, DI-based).

@willem-bd @fancyboi999 could you take another look when you have a moment?

WillemJiang (approved):

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -357,16 +393,38 @@ def _parse_fact_datetime(raw: str) -> datetime | None:
...
+    *write time* (when a fact is first created) so the review window is bounded
+    from the start.  Re-applying it here would prevent lifetime-extension
+    operations from ever moving the review window beyond that initial cap,
+    defeating the purpose of ``staleFactsToExtend``.  Falls back to the global
+    ``staleness_age_days`` for facts that pre-date this feature or where the
+    LLM did not provide an estimate.
+    """
+    raw = fact.get("expected_valid_days")
+    if isinstance(raw, (int, float)) and not isinstance(raw, bool) and raw > 0:
+        return int(raw)
```
willem-bd, line 410: Nit (defense-in-depth, low priority): this returns the raw stored `expected_valid_days` with no upper bound. The overflow rationale for `staleness_max_extension_days` holds for every value *this code writes* — creation (`min(evd, staleness_age_days * multiplier)` ≤ 365*100 = 36500) and extension (`min(..., staleness_max_extension_days)` ≤ 36500) are both bounded, and `expected_valid_days` is a new field so no legacy facts carry it. The only remaining vector is a hand-edited / corrupted `memory.json` with a huge value, which would still raise `OverflowError` in `timedelta(days=...)` at the next `_select_stale_candidates` call.

A one-line read-side clamp to a hard safe constant — e.g. `return min(int(raw), _MAX_SAFE_EVD_DAYS)` with `_MAX_SAFE_EVD_DAYS = 10**6` (well under `timedelta.max` ~999_999_999 days, above any legit value ≤ 36500) — would make the "prevents timedelta overflow" guarantee hold unconditionally without affecting any legitimate value. Low priority since the write paths are already fully bounded, but cheap insurance.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/prompt.py
```diff
@@ -155,27 +166,37 @@
...
 The reason should cite what signal in the conversation (or absence thereof)
 supports the removal.
 
-Be conservative — when in doubt, KEEP. Removing a valid fact is worse than
+Optionally, for facts you KEEP and wish to recalibrate, add them to
+"staleFactsToExtend" with the number of days from now before the next review:
+{{"id": "fact_id", "extend_by_days": 365, "reason": "brief explanation"}}
+Use this when the current window seems miscalibrated - e.g. a core skill marked
+``valid:30d`` that is clearly stable, or a goal nearing completion whose window
+should shrink. Omit facts whose current window already seems appropriate.
```
willem-bd, line 197: Minor wording nit: the EXTEND path can only *grow* the stored `expected_valid_days` — `extend_by_days` must be > 0, and a stale candidate already has `days_since >= effective_age`, so `new_evd = days_since + extend_by_days` strictly increases. What actually shrinks for a "goal nearing completion" is the *remaining time to next review* (a small `extend_by_days` schedules a near-term re-review), which is almost certainly the intent here.

The current phrasing ("whose window should shrink") could be read as reducing the stored value, which isn't possible via EXTEND. Rephrasing to something like "schedule a sooner re-review" would match the actual mechanism. Not a behavior bug.

willem-bd (commented):
Reviewed the full diff against current `main`. Overall this is a well-designed, carefully-tested change: the two-ceiling model (multiplier creation cap vs. absolute extension cap) is sound, the guardrail invariants are correctly preserved through the #4122 re-port — the hoisted `proposed_remove_ids` and write-time-only cap are exactly right — and prompt-injection escaping stays correct (`quote=False` is appropriate for element-text position inside `<stale_facts>`).

Verified against the head commit: `_apply_updates` receives normalized data (`_parse_memory_update_response` → `_normalize_memory_update_data` → apply), the config-forwarding set and `DeerMemConfig` are updated consistently, and the new prompt has no stray `{age_days}` placeholder. Every invariant fix has a dedicated regression test.

Approve. Two minor inline nits below, neither a blocker. The read-side clamp is the more substantive one (would make the overflow guarantee unconditional) but is low priority since all write paths are already bounded. I also noted in review that `_select_stale_candidates` runs twice per update (prompt-build + apply) — negligible for ≤ max_facts and not worth caching, so I didn't flag it inline.

PR #4146, opened 2026-07-13
fix(models): scope the OpenAI-compat rules to BaseChatOpenAI, not a class-path allowlist

AnnaSuSu:
## Why

`_OPENAI_COMPAT_USE_PATHS` reads like "the OpenAI-compatible family". It holds two of its eight members:

```python
_OPENAI_COMPAT_USE_PATHS = (
    "langchain_openai:ChatOpenAI",
    "deerflow.models.patched_openai:PatchedChatOpenAI",
)
```

The harness ships six more `BaseChatOpenAI` subclasses — `VllmChatModel`, `MindIEChatModel`, `PatchedChatDeepSeek`, `PatchedChatMiMo`, `PatchedChatStepFun`, `PatchedChatMiniMax` — and three functions gate on that tuple as if it were the family.

#4102 fixed the third one, `_apply_stream_chunk_timeout_default`, and its docstring states the rule:

> We gate on `issubclass(model_class, BaseChatOpenAI)` **rather than an explicit class-path allowlist** so any OpenAI-compatible subclass inherits the default automatically … the original fix (#3195) matched only `ChatOpenAI` / `PatchedChatOpenAI`, so those subclasses kept langchain-openai's aggressive built-in chunk-gap timeout and — worse — **silently discarded** a user's explicit `stream_chunk_timeout`.

The other two consumers still use the allowlist, and they fail the same way.

**1. `_normalize_openai_base_url` — the endpoint is silently dropped.** For the five subclasses that inherit the endpoint only as `openai_api_base` (alias `base_url`), a config that says `api_base` is never normalized. `ModelConfig` is `extra="allow"`, so it reaches the constructor, which diverts it into `model_kwargs` — spread into every `Completions.create()` call. Driving the real factory on `main`:

```
model class            endpoint honored?   api_base in model_kwargs?
PatchedChatMiniMax     False               True     -> actually talks to: None
PatchedChatStepFun     False               True     -> actually talks to: None
PatchedChatMiMo        False               True     -> actually talks to: None
VllmChatModel          False               True     -> actually talks to: None
MindIEChatModel        False               True     -> actually talks to: None

UserWarning: api_base is not default parameter. api_base was transferred to model_kwargs.
```

So the client silently falls back to the default OpenAI endpoint instead of the user's gateway, *and* every request then dies with an opaque `unexpected keyword argument 'api_base'`. That is verbatim the failure `_normalize_openai_base_url`'s own docstring says it exists to prevent.

This is not a hypothetical typo. `config.example.yaml` documents `api_base` for the Doubao and Kimi blocks (`PatchedChatDeepSeek`, lines 91 and 282) and `base_url` for the MiniMax / StepFun / MiMo / vLLM / MindIE blocks around them. Copying an endpoint line between neighbouring examples is exactly the mistake the docstring anticipates — and the guard does not cover the classes you land on.

**2. `_warn_unknown_model_settings` — the diagnostic is off for the same classes.** Its docstring says it is "scoped to the OpenAI-compatible family (`_OPENAI_COMPAT_USE_PATHS`) — that is where the `model_kwargs` divert-and-crash behavior occurs". The divert is implemented in `BaseChatOpenAI`, so it occurs in *all* subclasses; the tuple is not the family. A typo'd key on those five is silent:

```
ChatOpenAI             DeerFlow typo warning? YES
PatchedChatMiniMax     DeerFlow typo warning? NO   <-- silent
VllmChatModel          DeerFlow typo warning? NO   <-- silent
```

The warning that exists to surface failure 1 is disabled for precisely the classes that suffer it.

## What changed

Both consumers now gate on the class, matching #4102:

```python
if not issubclass(model_class, BaseChatOpenAI) or _declares_api_base(model_class):
    return
```

**`_declares_api_base` is load-bearing — a bare `issubclass` gate would be a regression.** `PatchedChatDeepSeek` is the one subclass that declares `api_base` as its own constructor field (inherited from `langchain_deepseek:ChatDeepSeek`). There the key is canonical, not a typo, and rewriting it to `base_url` would break every Doubao / Kimi config in `config.example.yaml`. The predicate is therefore "`BaseChatOpenAI` subclass **that does not declare `api_base` itself**".

`_warn_unknown_model_settings` needs no such carve-out: it reads the class's own field/alias set, so `api_base` is already valid for `PatchedChatDeepSeek`.

**`_enable_stream_usage_by_default` removed (addressing review).** It read the same tuple. As @willem-bd noted, it is provably redundant: `create_chat_model` has a broader fallback a few lines down that sets `stream_usage=True` for any class declaring the field, so all eight classes get it regardless of the tuple. Verified a no-op — the two `stream_usage` tests stay green on `main` with the helper still present. The helper and its now-unused `_OPENAI_COMPAT_USE_PATHS` tuple are deleted, so all three tuple consumers are now resolved: two re-gated on `issubclass`, one removed.

## Surface area

The change is confined to the model factory (`backend/packages/harness/deerflow/models/factory.py`), which is none of the boxes above verbatim — so I have ticked the one that is materially true rather than the one that is structurally closest. It **is** a default-behaviour change, and reviewers should weigh it as one: a config that sets `api_base` on `VllmChatModel` / `MindIEChatModel` / `PatchedChatMiMo` / `PatchedChatStepFun` / `PatchedChatMiniMax` today silently yields a client pointed at the **default OpenAI endpoint**; after this PR it is pointed at the endpoint the user configured. Any deployment that is currently "working" on one of those five is not reaching its configured gateway at all — every request raises `unexpected keyword argument 'api_base'` — so there is no working configuration whose behaviour this alters. Configs that already use `base_url` (all such blocks in `config.example.yaml`) are byte-for-byte unaffected.

No documentation change needed: `backend/AGENTS.md` describes the model factory but makes no claim about which classes these two helpers cover.

## Screenshots / Recording

N/A — no user-visible UI surface. The behavioural evidence is the `main`-vs-branch factory transcript under **Why** and **Validation**.

## Bug fix verification

**Tests:** `backend/tests/test_model_factory.py`

**Red on `main`, green on this branch** — reverting only `factory.py` and keeping the new tests:

```
FAILED test_api_base_normalized_for_all_openai_subclasses[VllmChatModel]
FAILED test_api_base_normalized_for_all_openai_subclasses[MindIEChatModel]
FAILED test_api_base_normalized_for_all_openai_subclasses[PatchedChatMiMo]
FAILED test_api_base_normalized_for_all_openai_subclasses[PatchedChatStepFun]
FAILED test_api_base_normalized_for_all_openai_subclasses[PatchedChatMiniMax]
FAILED test_unknown_config_key_warns_for_all_openai_subclasses[  ...same five...  ]
FAILED test_api_base_reaches_real_minimax_constructor_as_base_url
11 failed, 66 passed
```

With `factory.py` restored: `77 passed`.

**Per-clause red check.** This change both widens (five classes join) and narrows (`PatchedChatDeepSeek` must be excluded), so the narrowing clause needs its own anchor. Deleting only `_declares_api_base` from the gate:

```
1 failed, 76 passed
FAILED test_api_base_preserved_for_provider_that_declares_it
```

Exactly one test fails, and it is the one that exists for that clause. No test in the new set stays green under every mutation.

`test_api_base_preserved_for_provider_that_declares_it` is green on `main` as well — it does not guard the bug, it guards the fix from over-widening, which is what that mutation demonstrates.

**Four pre-existing tests were updated, and they were not guarding anything.** `test_api_base_normalized_to_base_url_for_chatopenai`, `test_base_url_takes_precedence_when_both_set`, `test_api_base_normalized_for_patched_chatopenai` and `test_api_base_dropped_when_openai_api_base_field_name_set` each name a *ChatOpenAI* model but hand the factory a `FakeChatModel` stub, which is a plain `BaseChatModel`. They passed only because the old gate matched the config **string** and never looked at the class. They now use `_capturing_class(ChatOpenAI, …)` — the real-subclass stub #4102 introduced for exactly this reason — so they test what their names claim. Same for `test_no_unknown_key_warning_for_non_openai_class`, which asserted a negative against a stub that could never have tripped the gate; it also gained an `api_base` value so its normalization-skip assertion has real input (`api_base` passed through verbatim, never rewritten to `base_url`).

**Two more tests updated when the helper was removed.** `test_openai_compatible_provider_passes_base_url` and `test_openai_compatible_provider_enables_stream_usage_for_openai_api_base` asserted `stream_usage=True` against the same `FakeChatModel` stub, and passed only via the deleted helper's use-path match. They now use `_capturing_class(ChatOpenAI, …)` so they exercise the class-field fallback that actually provides the default — green on `main` (helper present) and on this branch (helper gone), which is what makes the removal a verified no-op.

## Validation

- `cd backend && make lint` — `ruff check`: all checks passed; `ruff format --check`: 836 files already formatted.
- `cd backend && make test` — **8 failed, 7283 passed, 26 skipped**. The same 8 fail on a clean `origin/main` (**8 failed, 7272 passed**): pre-existing `web_fetch` cases (`test_browserless_client`, `test_crawl4ai_tools`, `test_fastcrw_tools`) that need outbound DNS. Failure sets are byte-identical; this branch adds none and adds 11 passing.
- End-to-end through the real `create_chat_model` with the real provider classes, nothing stubbed — the `main` transcript is quoted under **Why**. On this branch all eight classes report `endpoint honored? True` with no `api_base` in `model_kwargs`, `PatchedChatDeepSeek` still receives its native `api_base` untouched, and the typo warning fires for all five previously-silent classes.
- Added a `CHANGELOG.md` `Unreleased` → `Fixed` entry for the `api_base` behavior change on the five subclasses (per review — this is ticked as a default-behavior change).

## Related

- #4102 — same root cause, fixed in `_apply_stream_chunk_timeout_default`; its docstring states the `issubclass`-over-allowlist rule this PR applies to the two remaining consumers.
- #3790 — introduced `_normalize_openai_base_url` and the unknown-key warning for `ChatOpenAI`. This extends both to the rest of the family rather than changing their behaviour for the two classes already covered.
- #3652 (open) — adds `RvllmChatModel`, another OpenAI-compatible subclass. It is a data point for gating on `issubclass` rather than appending two more strings to the tuple: a class-path allowlist has to be updated by every new provider, and silently misbehaves when it isn't. No file overlap with this PR.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Claude Code assisted with enumerating the three consumers of the class-path allowlist, driving each provider class through the real factory to establish which ones actually drop the endpoint (the `stream_usage` consumer looked affected on inspection and turned out to be backstopped — that only surfaced by running it), computing the delta set in both directions, and writing the regression tests. The contributor reviewed every line and takes responsibility for it.


backend/tests/test_model_factory.py
```diff
@@ -1427,11 +1442,104 @@ def test_no_unknown_key_warning_for_non_openai_class(monkeypatch, caplog):
...
+    _patch_factory(monkeypatch, cfg, model_class=_capturing_class(ChatAnthropic, captured))
 
-    FakeChatModel.captured_kwargs = {}
     with caplog.at_level(logging.WARNING, logger=factory_module.__name__):
         factory_module.create_chat_model(name="anthropic")
 
     assert not any("not recognized parameters" in rec.message for rec in caplog.records)
+    # api_base normalization is likewise scoped to the OpenAI family: a non-BaseChatOpenAI
+    # provider must never have its keys rewritten.
+    assert "base_url" not in captured
```
willem-bd, line 1450: This assertion confirms `base_url` isn't spuriously added, but because the config sets `frequency_penalty` (not `api_base`), it doesn't actually exercise the normalization-skip path - normalization only emits `base_url` when `api_base` is present in config. So the case "a non-`BaseChatOpenAI` class with `api_base` set is left untouched" isn't directly covered here.

The gate (`if not issubclass(model_class, BaseChatOpenAI) ...: return`) is correct by inspection, but to close the matrix consider adding `api_base="http://x/v1"` to this `ChatAnthropic` config and asserting `captured.get("api_base") == "http://x/v1"` (and still no `base_url`).

backend/packages/harness/deerflow/models/factory.py
```diff
@@ -265,7 +283,7 @@ def create_chat_model(name: str | None = None, thinking_enabled: bool = False, *
 
     # Normalize the api_base -> base_url alias FIRST, so the downstream OpenAI-compatible
     # heuristics (stream_usage / stream_chunk_timeout) see the canonical endpoint key.
-    _normalize_openai_base_url(model_config.use, model_settings_from_config)
+    _normalize_openai_base_url(model_class, model_settings_from_config)
     _enable_stream_usage_by_default(model_config.use, model_settings_from_config)
```
willem-bd: You flagged `_enable_stream_usage_by_default` as deliberately out of scope in the PR body - recording the follow-up nudge here. It's now provably redundant: the unconditional fallback a few lines below (`if "stream_usage" in getattr(model_class, "model_fields", {}): model_settings_from_config["stream_usage"] = True`) sets `stream_usage=True` for every class that declares the field, including the two stock paths, so this helper's tuple gate (and its "only when `base_url`/`openai_api_base` is set" condition) no longer gates anything real.

Not a blocker for this PR - just worth a follow-up to either remove the helper, or fold the `base_url`-gated behavior into the fallback if that gating was actually intended (today the fallback is broader and overrides it).

backend/packages/harness/deerflow/models/factory.py
```diff
@@ -56,20 +58,35 @@ def _enable_stream_usage_by_default(model_use_path: str, model_settings_from_con
...
+    error (and the endpoint override is silently dropped). Rename it here so the model works as the
+    user intended.
+
+    Gated on ``issubclass(model_class, BaseChatOpenAI)`` rather than a class-path allowlist, so any
+    OpenAI-compatible subclass is covered automatically — the divert-and-crash behaviour is a
+    property of the base class, not of the two paths that used to be listed. Classes that declare
+    ``api_base`` themselves are skipped: there the key is canonical, not a typo.
     """
-    if model_use_path not in _OPENAI_COMPAT_USE_PATHS:
+    if not issubclass(model_class, BaseChatOpenAI) or _declares_api_base(model_class):
```
willem-bd, line 63: This gate is the line that widens the behavior: `api_base` on `VllmChatModel` / `MindIEChatModel` / `PatchedChatMiMo` / `PatchedChatStepFun` / `PatchedChatMiniMax` goes from crashing (`unexpected keyword argument 'api_base'`, endpoint dropped) to being honored. The PR body's "no working config is affected" argument is sound, but since this is ticked as a Default behavior change, a one-line CHANGELOG / release-note entry would help operators of those five classes recognize the fix if they were hitting the old crash. The diff only touches `factory.py` + tests today.

willem-bd (commented):
Review findings (3 inline comments). All low/nit - the fix itself is correct and well-tested.

AnnaSuSu:
Thanks for the review — all three addressed in the latest commit.

**1. `_enable_stream_usage_by_default` (redundant helper).** You're right, and removing it turned out to be more instructive than expected. The helper and its now-unused `_OPENAI_COMPAT_USE_PATHS` tuple are deleted; the class-field fallback below already sets `stream_usage=True` for every `BaseChatOpenAI` subclass, so all three tuple consumers are now resolved (two re-gated on `issubclass`, one removed).

The catch: two existing tests (`test_openai_compatible_provider_passes_base_url`, `test_openai_compatible_provider_enables_stream_usage_for_openai_api_base`) asserted `stream_usage=True` against a `FakeChatModel` stub that doesn't declare the field — they were green only via the helper's use-path *string* match, exactly the same stub problem this PR already fixed for the four normalization tests. Pointed them at a real `ChatOpenAI` capturing class. That also gives a clean no-op proof: both stay green on `main` with the helper still present *and* on this branch with it gone.

**2. Test-matrix gap on `test_no_unknown_key_warning_for_non_openai_class`.** Fixed as suggested — added `api_base="http://x/v1"` to the `ChatAnthropic` config and `assert captured.get("api_base") == "http://x/v1"`. It now actually exercises the normalization-skip path instead of just asserting its absence. (This one is matrix coverage, not a red-on-`main` case: a non-`BaseChatOpenAI` provider's `api_base` is left untouched on `main` too — `_warn_unknown_model_settings` returns early for it — so the test guards against a future regression that would start rewriting non-family keys, and is green on both.)

**3. CHANGELOG / release note.** Added an `Unreleased` → `Fixed` entry naming the five affected subclasses and the old crash, so operators hitting the old `unexpected keyword argument 'api_base'` can recognize the fix.

WillemJiang (approved):

PR #4147, opened 2026-07-13
fix(frontend): show branch action only for completed turns

### LINKED ISSUE #4145: [bug] 页面上分叉按钮在一个run内有多余显示，导致多余显示处点击后报错409
### Before you start

### Problem summary

页面上分叉按钮在一个run内有多余显示，导致多余显示处点击后报错409

### Affected area(s)

Frontend (UI / Next.js)

### What happened?

问题现象: 智能体执行一个任务，该任务生成了to-dos计划列表。执行完成后观察页面发现：

- 页面上展示了执行步骤区域1（工具调用类的消息）

- 然后输出了文本内容正文区域2，而正文下方有复制按钮和分叉按钮出现

- 然后智能体又显示了执行步骤区域3（因为todo列表中还有未完成的任务）

- 当我点击正文区域2下方的分叉按钮时，收到报错：`{"detail":"This turn can no longer be branched from."}`

Image

### Expected behavior

分叉按钮应该只在运行的最终消息上显示，而不是在运行过程中的中间消息上显示。

### Steps to reproduce

使用pro或者ultra模式执行一个较复杂的任务

### Relevant logs

```shell
无
```

Docker (make docker-start)

### Operating system

Windows

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Support bundle summary

_No response_

### Additional context

_No response_

Huixin615:
Fixes #4145

## Why

Assistant text can be emitted before later tool or todo processing within the
same run. The frontend displayed a branch action for every non-streaming
assistant text group, while the backend only accepts a completed assistant turn.
Clicking the intermediate action therefore returned 409.

## What changed

- Show the branch action only on the final AI-bearing assistant group of each
  completed visible human turn.
- Keep historical completed turns branchable.
- Hide the action for the active turn while it is still loading.
- Keep copy and regenerate behavior unchanged.

## Surface area

- [x] **Frontend UI**
- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**

## Screenshots / Recording

Attach before/after screenshots showing that the intermediate response no longer
has a branch action while the final response still has one.

## Bug fix verification

- Test path: `frontend/tests/e2e/branch-thread.spec.ts`
- Red on main and green on this branch: yes
- Before the fix, the intermediate assistant response exposed one branch button;
  after the fix it exposes none, while historical and final turns remain
  branchable.

## Validation

- `pnpm format`
- `pnpm lint`
- `pnpm typecheck`
- `BETTER_AUTH_SECRET=local-dev-secret pnpm build`
- `pnpm test` — 623 passed
- `pnpm test:e2e -- branch-thread.spec.ts` — 1 passed
- Full E2E — 93 passed; unrelated `artifact-stream-state.spec.ts` mock-stream
  test failed and reproduced in isolation
- `git diff --check`

### DIFF
--- a/frontend/src/core/messages/utils.ts
+++ b/frontend/src/core/messages/utils.ts
@@ -157,10 +157,12 @@ export function getMessageGroups(messages: Message[]): MessageGroup[] {
 export function getBranchableAssistantGroupIds(
   groups: MessageGroup[],
   isCurrentTurnLoading: boolean,
-) {
+): Set<string> {
   // Hidden messages were already removed by getMessageGroups, matching the
   // backend's branch checkpoint visibility rules. Within each visible human
-  // turn, only the final AI-bearing group can be a completed branch target.
+  // turn, branching is exposed only when the final AI-bearing group is a
+  // terminal assistant text group. Processing, present-files, and subagent
+  // groups do not render assistant actions.
   const branchableGroupIds = new Set<string>();
   let lastAIGroup: MessageGroup | null = null;
 
--- a/frontend/src/core/messages/utils.ts
+++ b/frontend/src/core/messages/utils.ts
@@ -154,6 +154,41 @@ export function getMessageGroups(messages: Message[]): MessageGroup[] {
   return groups;
 }
 
+export function getBranchableAssistantGroupIds(
+  groups: MessageGroup[],
+  isCurrentTurnLoading: boolean,
+) {
+  // Hidden messages were already removed by getMessageGroups, matching the
+  // backend's branch checkpoint visibility rules. Within each visible human
+  // turn, only the final AI-bearing group can be a completed branch target.
+  const branchableGroupIds = new Set<string>();
+  let lastAIGroup: MessageGroup | null = null;
+
+  const completeTurn = () => {
+    if (lastAIGroup?.type === "assistant" && lastAIGroup.id) {
+      branchableGroupIds.add(lastAIGroup.id);
+    }
+    lastAIGroup = null;
+  };
+
+  for (const group of groups) {
+    if (group.type === "human") {
+      completeTurn();
+      continue;
+    }
+
+    if (group.messages.some((message) => message.type === "ai")) {
+      lastAIGroup = group;
+    }
+  }
+
+  if (!isCurrentTurnLoading) {
+    completeTurn();
+  }
+
+  return branchableGroupIds;
+}
+
 export function groupMessages<T>(
   messages: Message[],
   mapper: (group: MessageGroup) => T,
--- a/frontend/tests/unit/core/messages/utils.test.ts
+++ b/frontend/tests/unit/core/messages/utils.test.ts
@@ -121,6 +121,14 @@ describe("branchable assistant groups", () => {
       "ai-history",
     ]);
   });
+
+  test("does not expose a completed turn that ends in processing", () => {
+    const groups = getMessageGroups(messages.slice(0, -1));
+
+    expect([...getBranchableAssistantGroupIds(groups, false)]).toEqual([
+      "ai-history",
+    ]);
+  });
 });
 
 test("reasoning + content (no tool calls) yields a single assistant bubble, not a duplicate processing group", () => {
--- a/frontend/tests/unit/core/messages/utils.test.ts
+++ b/frontend/tests/unit/core/messages/utils.test.ts
@@ -5,6 +5,7 @@ import {
   extractContentFromMessage,
   extractTextFromMessage,
   extractReasoningContentFromMessage,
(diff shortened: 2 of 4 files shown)


frontend/src/core/messages/utils.ts
```diff
@@ -154,6 +154,41 @@ export function getMessageGroups(messages: Message[]): MessageGroup[] {
...
+  isCurrentTurnLoading: boolean,
+) {
+  // Hidden messages were already removed by getMessageGroups, matching the
+  // backend's branch checkpoint visibility rules. Within each visible human
+  // turn, only the final AI-bearing group can be a completed branch target.
+  const branchableGroupIds = new Set<string>();
+  let lastAIGroup: MessageGroup | null = null;
+
+  const completeTurn = () => {
+    if (lastAIGroup?.type === "assistant" && lastAIGroup.id) {
```
willem-bd, line 170: The comment above says "only the final **AI-bearing** group can be a completed branch target," but this guard restricts to `type === "assistant"` specifically. `assistant:present-files` and `assistant:subagent` groups also contain AI messages, so a turn whose final message is a present-files or subagent step (no trailing text answer) won't be marked branchable here.

Pre-fix, the branch button rendered on any group with an `ai` action target, including those, so this is a (minor) behavior narrowing, not just a comment mismatch. If the backend accepts branching from those messages, the button is now gone; if it doesn't, this is strictly an improvement. Could you confirm the exclusion is intended, and tighten the comment to "final terminal `assistant` group"?

willem-bd (commented):
Review notes (all non-blocking). The fix is correct and well-tested; the inline comments are minor clarifications and a suggested test hardening.

frontend/src/core/messages/utils.ts
```diff
@@ -154,6 +154,41 @@ export function getMessageGroups(messages: Message[]): MessageGroup[] {
   return groups;
 }
 
+export function getBranchableAssistantGroupIds(
+  groups: MessageGroup[],
+  isCurrentTurnLoading: boolean,
+) {
```
willem-bd: Minor: the function's return type isn't annotated on the signature (it's inferred from `new Set<string>()` inside). Sibling `getMessageGroups(): MessageGroup[]` annotates its return type. Adding `): Set<string> {` here would match house style and make the `.has()` contract explicit at the call site.

frontend/tests/unit/core/messages/utils.test.ts
```diff
@@ -82,6 +83,46 @@ test("aggregates token usage messages once per assistant turn", () => {
...
+  test("keeps historical turns branchable and selects only the final AI group in the current completed turn", () => {
+    const groups = getMessageGroups(messages);
+
+    expect([...getBranchableAssistantGroupIds(groups, false)]).toEqual([
+      "ai-history",
+      "ai-final",
+    ]);
+  });
+
+  test("does not expose the current turn while it is still loading", () => {
```
willem-bd, line 117: The loading case covers the in-progress turn. It would also lock in the `!isCurrentTurnLoading`-but-no-final-bubble branch of `completeTurn`: a completed turn that ends in an `assistant:processing` group (e.g. tool calls with no final text answer) should yield **no** branchable id for that turn, while earlier completed turns remain branchable. A small additional fixture asserting that would guard the "incomplete completed turn" path.

frontend/src/core/messages/utils.ts
Huixin615: > The comment above says "only the final **AI-bearing** group can be a completed branch target," but this guard restricts to `type === "assistant"` specifically. `assistant:present-files` and `assistant:subagent` groups also contain AI messages, so a turn whose final message is a present-files or subagent step (no trailing text answer) won't be marked branchable here.
> 
> Pre-fix, the branch button rendered on any group with an `ai` action target, including those, so this is a (minor) behavior narrowing, not just a comment mismatch. If the backend accepts branching from those messages, the button is now gone; if it doesn't, this is strictly an improvement. Could you confirm the exclusion is intended, and tighten the comment to "final terminal `assistant` group"?

Thanks for your review! Confirmed—the exclusion is intentional. Before this fix, renderAssistantActions was already invoked only for group.type === "assistant"; present-files and subagent groups use separate render paths and did not expose branch actions. The backend accepts AI message IDs generically, but this PR does not narrow the existing UI surface. I tightened the comment to explicitly require a terminal assistant text group and to call out processing, present-files, and subagent groups.

WillemJiang (approved):

PR #4148, opened 2026-07-13
feat(checkpoint): retention policy to prevent DB bloat (#4138)

nankingjing:
Add configurable checkpoint retention to prevent O(N^2) storage growth from LangGraph's full-state-per-checkpoint persistence.

LangGraph persists full cumulative messages in every checkpoint. Over long conversations this causes quadratic storage growth (issue #4138). This PR adds a configurable retention policy that automatically prunes old checkpoints.

## Changes

- **CheckpointRetentionConfig**: New configuration model with these fields:
  - `checkpoint_retention.enabled`: gate the background pruning task
  - `checkpoint_retention.max_checkpoints_per_thread`: keep N most recent checkpoints per thread
  - `checkpoint_retention.max_age_hours`: remove checkpoints older than this many hours
  - `checkpoint_retention.min_interval_seconds`: throttle pruning frequency

- **retention.py**: Pruning logic that:
  - Queries all thread_ids from the checkpoints table
  - For each thread, determines which checkpoints to keep based on policy
  - Always preserves the first (base) checkpoint for replayability
  - Supports dry_run mode for auditing without deletion
  - Batch-deletes in groups of 200 to avoid large IN clauses

- **Gateway lifecycle**: Background asyncio task started during lifespan
  - Sleeps for min_interval_seconds between pruning cycles
  - Cancelled gracefully on Gateway shutdown
  - Logs pruning results (count of pruned checkpoints, threads affected)

- **Config example**: Commented-out config block in config.example.yaml

- **Tests**: 14 tests covering:
  - Base checkpoint preservation
  - Max checkpoints per thread limit
  - Max age filtering
  - Multiple thread pruning
  - Dry run (non-destructive)
  - Empty table no-op
  - Concurrent write safety
  - Missing timestamp handling
  - Single checkpoint threads

## How it works

When enabled in config.yaml, a background task wakes up every `min_interval_seconds` and prunes checkpoints that exceed the retention limits. The pruning is safe against concurrent writes because it only deletes rows that existed at query time.

Default configuration (disabled):
```yaml
database:
  checkpoint_retention:
    enabled: false
    max_checkpoints_per_thread: 50
    max_age_hours: 168
    min_interval_seconds: 300
```

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/frontend/src/components/workspace/messages/message-list.tsx
+++ b/frontend/src/components/workspace/messages/message-list.tsx
@@ -47,6 +47,7 @@ import {
   extractTextFromMessage,
   getAssistantTurnCopyData,
   getAssistantTurnUsageMessages,
+  getBranchableAssistantGroupIds,
   getMessageGroups,
   getStreamingMessageLookup,
   hasContent,
@@ -404,6 +405,10 @@ export function MessageList({
     }
     return null;
   }, [groupedMessages, thread.isLoading]);
+  const branchableAssistantGroupIds = useMemo(
+    () => getBranchableAssistantGroupIds(groupedMessages, thread.isLoading),
+    [groupedMessages, thread.isLoading],
+  );
 
   const clearSelectionToolbar = useCallback(() => {
     setSelectionToolbar(null);
@@ -544,6 +549,7 @@ export function MessageList({
     (
       messages: Message[],
       isStreaming: boolean,
+      enableBranchForTurn: boolean,
       enableRegenerateForTurn: boolean,
     ) => {
       const clipboardData = getAssistantTurnCopyData(messages, { isStreaming });
@@ -561,36 +567,41 @@ export function MessageList({
       return (
         <div className="mt-2 flex justify-start gap-1 opacity-0 transition-opacity delay-200 duration-300 group-hover/assistant-turn:opacity-100">
           {clipboardData && <CopyButton clipboardData={clipboardData} />}
-          {!isStreaming && actionTarget?.id && onBranchTurn && (
-            <Tooltip content={t.common.branch}>
-              <Button
-                aria-label={t.common.branch}
-                size="icon-sm"
-                type="button"
-                variant="ghost"
-                disabled={!canBranch || branchingMessageId === actionTarget.id}
-                onClick={() => {
-                  const targetId = actionTarget.id;
-                  if (!targetId) {
-                    return;
+          {enableBranchForTurn &&
+            !isStreaming &&
+            actionTarget?.id &&
+            onBranchTurn && (
+              <Tooltip content={t.common.branch}>
+                <Button
+                  aria-label={t.common.branch}
+                  size="icon-sm"
+                  type="button"
+                  variant="ghost"
+                  disabled={
+                    !canBranch || branchingMessageId === actionTarget.id
                   }
-                  setBranchingMessageId(targetId);
-                  void Promise.resolve(
-                    onBranchTurn(targetId, assistantMessageIds),
-                  ).finally(() => {
-                    setBranchingMessageId(null);
-                  });
-                }}
-              >
-                <GitBranchPlusIcon
-                  className={cn(
-                    "size-4",
-                    branchingMessageId === actionTarget.id && "animate-pulse",
-                  )}
-                />
-              </Button>
-            </Tooltip>
-          )}
+                  onClick={() => {
+                    const targetId = actionTarget.id;
+                    if (!targetId) {
+                      return;
+                    }
+                    setBranchingMessageId(targetId);
+                    void Promise.resolve(
+                      onBranchTurn(targetId, assistantMessageIds),
+                    ).finally(() => {
+                      setBranchingMessageId(null);
+                    });
+                  }}
+                >
+                  <GitBranchPlusIcon
+                    className={cn(
+                      "size-4",
+                      branchingMessageId === actionTarget.id && "animate-pulse",
+                    )}
+                  />
+                </Button>
+              </Tooltip>
+            )}
           {enableRegenerateForTurn &&
             actionTarget?.id &&
             onRegenerateMessage && (
@@ -782,6 +793,8 @@ export function MessageList({
                         group.messages,
                         streamingMessages,
                       ),
+                      group.id !== undefined &&
+                        branchableAssistantGroupIds.has(group.id),
                       group.id === latestAssistantGroupId,
                     )}
                 </div>
--- a/frontend/src/core/messages/utils.ts
+++ b/frontend/src/core/messages/utils.ts
@@ -154,6 +154,43 @@ export function getMessageGroups(messages: Message[]): MessageGroup[] {
   return groups;
 }
 
+export function getBranchableAssistantGroupIds(
+  groups: MessageGroup[],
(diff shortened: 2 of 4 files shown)


fancyboi999 (changes requested):
The retention key is incomplete: LangGraph checkpoints are scoped by `(thread_id, checkpoint_ns, checkpoint_id)`, but this code groups, orders, preserves a single "base", and deletes using only `thread_id` plus `checkpoint_id`.

That mixes the root graph and every subgraph namespace into one retention sequence. A minimal case with one thread, two namespaces, three checkpoints in each, and `max_checkpoints_per_thread=2` deletes three rows and leaves only one checkpoint in the subgraph namespace. Each namespace should instead retain its own base and recent checkpoints. This can destroy valid subgraph/branch history even though the configured per-thread limit has not been exceeded within that namespace.

Please enumerate and prune `(thread_id, checkpoint_ns)` independently, include `checkpoint_ns` in the delete predicate, and cover multiple namespaces in the test schema. The current tests include the column in the primary key but never insert a non-empty namespace, so they do not exercise the real LangGraph key.

nankingjing:
New feature PR: checkpoint retention policy to prevent O(N²) database bloat. 20 tests covering pruning logic, config, and the background retention loop. Brief review appreciated.

nankingjing:
New feature PR: checkpoint retention policy to prevent DB bloat. 20 tests covering pruning logic, config, and background retention loop. Brief review appreciated when you have a chance.

nankingjing (commented):
Thanks for the thorough review @fancyboi999.

You're absolutely right — the old code enumerated by thread_id only, pooling root and subgraph namespaces into one retention sequence. The fix:

1. `prune_checkpoints` now queries `SELECT DISTINCT thread_id, checkpoint_ns` and iterates over each `(thread_id, checkpoint_ns)` pair independently.
2. `_prune_thread` → `_prune_namespace`: filters SELECT by both `thread_id` AND `checkpoint_ns`, determines base/recent within that namespace scope only.
3. DELETE predicate now includes `checkpoint_ns = :ns` so cross-namespace contamination is impossible.

The reviewer's minimal case (1 thread, 2 namespaces, 3 cps each, max=2) is now covered by `test_namespaces_pruned_independently` — each namespace retains its own 3 checkpoints, zero deletions. 4 additional namespace tests cover independent base preservation, DELETE predicate isolation, multi-thread × multi-ns, and dry_run per-ns counting.

All 19 tests pass (14 original + 5 new namespace-scoped).

willem-bd (commented):
Posting the two critical findings from review as inline comments:

1. **`max_age_hours` is a silent no-op in production** - the `checkpoint` BLOB is msgpack (`JsonPlusSerializer`), but `_parse_checkpoint_ts` parses it as JSON, so every real checkpoint fails to parse and the age rule never fires. Tests pass only because they fabricate JSON fixtures.
2. **The pruning SELECT loads the full `checkpoint` BLOB for every checkpoint in a namespace into memory** (`fetchall`), which can OOM the gateway on exactly the bloated threads this PR targets.

Both are addressable by deriving ordering + age from `checkpoint_id` (LangGraph `uuid6`, chronologically sortable) instead of parsing the blob. The count-based retention and overall structure are sound.

backend/packages/harness/deerflow/runtime/checkpointer/retention.py
```diff
@@ -0,0 +1,258 @@
...
+
+    LangGraph stores ``checkpoint['ts']`` as an ISO-8601 string.
+    Returns a timezone-aware UTC datetime or ``None`` if parsing fails.
+    """
+    if checkpoint_raw is None:
+        return None
+    try:
+        if isinstance(checkpoint_raw, bytes):
+            checkpoint_raw = checkpoint_raw.decode("utf-8")
+        data = json.loads(checkpoint_raw)
```
willem-bd: The `checkpoint` column is serialized by LangGraph's `JsonPlusSerializer`, which is **ormsgpack, not JSON** (`langgraph/checkpoint/serde/jsonplus.py`; the sqlite saver round-trips it via `serde.loads_typed((type, checkpoint))`, never `json.loads`). So `json.loads(checkpoint_raw)` raises `JSONDecodeError` on every real checkpoint row -> `_parse_checkpoint_ts` returns `None` for all of them -> the age guard below (`if ts is not None and ts < stale_before`) never discards anything.

**Impact:** `max_age_hours` is a silent no-op in production. The count rule still works, but age-based pruning never fires - so a long-lived thread with fewer than `max_checkpoints_per_thread` checkpoints is never age-pruned, which is exactly the low-volume/stale case the knob exists for. Fail-safe direction is good (None -> keep, no data loss), but the feature is dead.

The tests miss this because `_checkpoint_json()` writes `json.dumps(...)`; `test_old_checkpoints_removed` asserts age-based deletion and passes - only against JSON fixtures.

**Fix:** deserialize via the proper serde (`JsonPlusSerializer().loads_typed((row.type, blob))["ts"]`, where the companion `type` column holds the serde module name); or derive ordering + age from `checkpoint_id` itself, which LangGraph generates as `uuid6` (`langgraph/pregel/_checkpoint.py:19`) - it sorts chronologically and `UUID(hex=cid).time` gives an absolute timestamp, so no blob load is needed at all (also fixes the memory issue on the SELECT below). An integration test writing real checkpoints through `AsyncSqliteSaver` before pruning would catch this.

backend/packages/harness/deerflow/runtime/checkpointer/retention.py
```diff
@@ -0,0 +1,258 @@
...
+    """Prune checkpoints for a single (thread_id, checkpoint_ns) namespace.
+
+    Returns the number of checkpoints deleted (or would-be-deleted).
+    """
+    # Fetch all checkpoints for this namespace, ordered by checkpoint_id for
+    # stable ordering. We also fetch the raw checkpoint JSON so we can
+    # extract timestamps.
+    rows = await session.execute(
+        text(
+            "SELECT checkpoint_id, checkpoint "
```
willem-bd: This materializes the **full `checkpoint` BLOB for every checkpoint in the namespace** into Python (`entries = rows.fetchall()` below), purely to read a timestamp. Given the PR's premise is that these blobs are large and numerous, the first sweep of an already-bloated thread (hundreds of checkpoints x multi-MB cumulative state) can pull gigabytes into the gateway process and OOM it - the cure loads the disease into RAM.

Today it's pure waste (the timestamp isn't even parsed correctly - see comment on the `json.loads` line above). Even after fixing that, prefer to compute keep/delete from `checkpoint_id` alone: base = `MIN(checkpoint_id)`, recent-N = `ORDER BY checkpoint_id DESC LIMIT :n`, and parse the v6 timestamp from `checkpoint_id` for age filtering - all pure SQL / no blob load. Then drop the `checkpoint` column from this SELECT entirely.

WillemJiang:
@nankingjing  Please fix the lint error.

0xzkslr-ai:
Thanks for the fix! I'll wait for this PR to land, since it resolves the issue I reported.

WillemJiang:
@nankingjing, please verify the pre-commit setting.

WillemJiang:
@nankingjing, please check if your force push selected the wrong option.

PR #4149, opened 2026-07-13
feat(extensions): config-declared middleware loading (#3923)

### LINKED ISSUE #3923: [feat]  Extension points for downstream integrators: config-declared middlewares, SSE stream wrapper, and run-level model override hooks

nankingjing:
Add config-declared extension middlewares: downstream integrators can register custom `AgentMiddleware` subclasses via `extensions_config.json` without forking the agent assembly code.

## Changes
- **ExtensionsConfig**: new `middlewares` field (list of `module:Attribute` colon-paths) plus an `enabled` gate (default `true`)
- **factory.py**: `load_extension_middlewares()` resolves each path via `resolve_class(path, AgentMiddleware)` — the same loader/validation used by the repo's other extension points — and instantiates it
- **lead_agent/agent.py**: wires the loaded middlewares into `build_middlewares(custom_middlewares=...)` in both agent branches (inserted before the terminal-response/safety/clarification tail)
- **Error handling (fail closed)**: only an absent declared module (the underlying `ModuleNotFoundError` names the declared module or one of its parent packages) is logged at INFO and skipped (optional-dependency pattern). Malformed paths, missing attributes, transitive import failures inside an installed module, non-`AgentMiddleware` classes, constructor failures, and duplicate paths abort agent construction with an actionable error — declared middlewares can be deterministic guards (domain allowlists, tool-call gating), so starting without one would silently change production behavior
- **Docs**: `config.example.yaml` + `extensions_config.example.json` document the feature; AGENTS.md documents the trust model

## Where to declare
In production `AppConfig.from_file()` populates the `extensions` section from `extensions_config.json` (an `extensions:` key in `config.yaml` is not read), so middlewares are declared there:

```json
{
  "mcpServers": { "...": {} },
  "skills": {},
  "enabled": true,
  "middlewares": [
    "my_package.middleware:MyCustomMiddleware"
  ]
}
```

## Security / trust model
Middleware entries are imported and instantiated in the agent process — operator-declared code by design, the same trust model as `mcpServers` commands and `mcpInterceptors` in the same file. The Gateway config API (`PUT /api/mcp/config`) is admin-gated and rewrites only `mcpServers`/`skills`, preserving `middlewares` verbatim, so the field is not writable via any API. Documented in `factory.py`, `extensions_config.py`, and AGENTS.md.

## Tests
22 tests in `backend/tests/test_extensions_config.py`: deserialization, valid/multiple paths, the `enabled` gate, the fail-closed error matrix (absent optional package skipped; transitive import failure, malformed path, missing attribute, non-middleware class, constructor failure, and duplicate path all raise), and real `ExtensionsConfig.from_file` / `AppConfig.from_file` integration tests pinning that extensions come from `extensions_config.json`.

Closes #3923


fancyboi999 (changes requested):
This head cannot import `AppConfig`: `app_config.py` imports `deerflow.config.tool_streaming_config.ToolStreamingConfig`, but `tool_streaming_config.py` is not present anywhere in this PR tree. A clean `from deerflow.config.app_config import AppConfig` fails immediately with `ModuleNotFoundError`, so Gateway startup and backend test collection are blocked before the extension middleware code can run.

Please remove the unrelated `tool_streaming` import/field from this PR, or include the missing module and its intended integration in the same self-contained change.

nankingjing:
Thanks for catching this, @fancyboi999!

The stray `ToolStreamingConfig` import and `tool_streaming` field were accidentally leaked from another branch. They are unrelated to the config-declared middleware feature in this PR.

Fix pushed in 65d0163e — both lines removed from `app_config.py`:
- `from deerflow.config.tool_streaming_config import ToolStreamingConfig` (import)
- `tool_streaming: ToolStreamingConfig = Field(...)` (field)

Verified:
- `from deerflow.config.app_config import AppConfig` imports cleanly
- All 24 `test_extensions_config` tests pass
- Full test suite shows no regressions (the single failure is the pre-existing Windows file-permission test in `test_channel_runtime_config_store`)

Ready for re-review.

WillemJiang:
Please fix the lint error.

backend/packages/harness/deerflow/agents/factory.py
```diff
@@ -328,6 +328,125 @@ def _assemble_from_features(
...
+
+        instances.append(instance)
+
+    if instances:
+        logger.info("Loaded %d extension middleware(s) from config: %s", len(instances), [type(m).__name__ for m in instances])
+
+    return instances
+
+
+def resolve_extension_hooks(app_config) -> tuple[object | None, object | None]:
```
willem-bd: **Dead code - `sse_wrapper` / `run_model_override` are never wired up.**

`resolve_extension_hooks()` is defined here and exercised by 6 tests, but it's never called from any production path (a repo-wide grep on the PR branch shows the only references are inside this function and the tests). The two `ExtensionsConfig` fields are parsed and validated but have zero runtime effect - an operator setting `extensions.sse_wrapper` will see nothing happen, with no warning.

Either wire these hooks into the SSE streaming / model-resolution paths in this PR, or drop the fields + this function + its tests and ship only the `middlewares` feature. Shipping dormant config that *looks* active is a footgun.

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -141,6 +147,18 @@ class ExtensionsConfig(BaseModel):
         default_factory=dict,
         description="Map of skill name to state configuration",
     )
+    middlewares: list[str] = Field(
+        default_factory=list,
+        description="List of dotted paths to AgentMiddleware subclasses to insert into the agent middleware chain.",
```
willem-bd: **Path format + `enabled` footgun.**

- "Dotted paths" is the wrong term here and in the PR body. `resolve_variable` requires colon form (`module:Attribute` via `rsplit(":", 1)`); a no-colon path like `my_package.middleware.MyCustomMiddleware` is silently skipped - exactly the case `test_invalid_path_format_skipped` covers. So the PR's own config example wouldn't load. Recommend `module:Attribute` and colon-form examples.
- The PR body shows `extensions.enabled: false`, but there's no `enabled` field. With `extra="allow"` the key is silently ignored, so an operator who sets it *thinking it disables extensions* still gets middlewares loaded. Add a real `enabled` gate (and honor it in `load_extension_middlewares`) or drop `enabled` from the docs.
- No example was added to `config.example.yaml` or `extensions_config.example.json`, so operators can't discover the feature from the templates.

backend/packages/harness/deerflow/agents/factory.py
```diff
@@ -328,6 +328,125 @@ def _assemble_from_features(
...
+        if not issubclass(cls, AgentMiddleware):
+            logger.error(
+                "Extension middleware %r resolved to %s which is not an AgentMiddleware subclass — skipped.",
+                path,
+                cls.__name__,
+            )
+            continue
+
+        try:
+            instance = cls()
```
willem-bd, line 462: **Security note: arbitrary class instantiation from config.**

`cls()` imports and instantiates any operator-declared class. This is fine if config is strictly operator-controlled on disk, but AGENTS.md notes runtime config files "may be edited at runtime via the Gateway API" - so any caller who can write to the extensions config now has a path to arbitrary code execution in the agent process. Acceptable for an extension mechanism, but worth (a) explicitly documenting the trust assumption here / in `AGENTS.md`, and (b) confirming the config-edit API is authenticated and not reachable from chat-side / untrusted callers.

backend/tests/test_extensions_config.py
```diff
@@ -0,0 +1,326 @@
...
+# cls() without async_setup/on_tool_start/on_chat_model_start kwargs, so we
+# monkey-patch AgentMiddleware.__init_subclass__ only for this module.
+_original_init_subclass = getattr(AgentMiddleware, "__init_subclass__", None)
+
+
+def _noop_init_subclass(cls, **kwargs):
+    pass
+
+
+AgentMiddleware.__init_subclass__ = classmethod(_noop_init_subclass)  # type: ignore[assignment]
```
willem-bd: **Global monkey-patch leaks across the whole test session.**

This runs at module import and permanently disables `AgentMiddleware.__init_subclass__` for every subclass for the rest of the test session - not just this module - and `_original_init_subclass` is never restored. That breaks test isolation and can silently mask hook-registration side effects other tests rely on. Prefer an `autouse` `monkeypatch` fixture or `unittest.mock.patch` so it's scoped + restored per test.

backend/tests/test_extensions_config.py
```diff
@@ -0,0 +1,326 @@
...
+        )
+        app_cfg = MagicMock(extensions=cfg)
+        result = load_extension_middlewares(app_cfg)
+        assert result == []
+
+    def test_instantiation_failure_skipped(self):
+        """If cls() raises, log and skip."""
+        from deerflow.agents.factory import load_extension_middlewares
+
+        class _BrokenMiddleware(AgentMiddleware):
```
willem-bd: **This test passes for the wrong reason.**

`_BrokenMiddleware` is defined inside the test method body, so it's a function-local class - *not* an attribute of the `test_extensions_config` module. `resolve_variable("test_extensions_config:_BrokenMiddleware")` does `getattr(module, "_BrokenMiddleware")`, raises `AttributeError` -> `ImportError`, which is caught by the `except ImportError` branch (logged at INFO, skipped) **before** `cls()` is ever reached.

So the instantiation-failure code path (`instance = cls()` -> `except Exception`) is never exercised here. To actually test it, make `_BrokenMiddleware` module-level (resolvable) or patch `resolve_variable` to return a class whose `__init__` raises.

backend/tests/test_extensions_config.py
```diff
@@ -0,0 +1,326 @@
...
+
+# ---------------------------------------------------------------------------
+# AppConfig integration
+# ---------------------------------------------------------------------------
+
+
+class TestAppConfigExtensionsMerge:
+    """Verify that YAML-declared extensions merge with JSON-loaded ones."""
+
+    def test_yaml_extensions_override_json(self):
```
willem-bd: **No-op test (and dead-code coverage).**

This test's body is literally `pass` with a comment "we can't fully test from_file" - it asserts nothing. Combined with the 6 `TestResolveExtensionHooks` tests, which cover `resolve_extension_hooks` (see the dead-code note on `factory.py`), ~7 of the 24 tests either validate nothing or validate a function that's never called in production. Worth trimming so the "24 tests" claim reflects meaningful coverage.

config.example.yaml
```diff
@@ -1561,6 +1561,14 @@ skill_evolution:
 database:
   backend: sqlite
   sqlite_dir: .deer-flow/data
+  # Checkpoint retention prevents O(N^2) storage bloat from LangGraph's
+  # full-state-per-checkpoint persistence. When enabled, a background task
+  # prunes old checkpoints per thread, keeping only the most recent N.
+  # checkpoint_retention:
```
willem-bd: **Out of scope and unimplemented.**

This commented-out `checkpoint_retention` block under `database:` has no corresponding implementation - a repo-wide grep finds no code consuming `checkpoint_retention` / `max_checkpoints_per_thread` / `max_age_hours` / `min_interval_seconds`. It's also unrelated to this PR's middleware-loading purpose. Recommend dropping it (or shipping it in its own PR alongside the actual pruning task).

willem-bd (commented):
Code review - **request changes**. The core feature (config-declared `middlewares`) is correctly wired: `load_extension_middlewares` -> `build_middlewares(custom_middlewares=...)` -> inserted before `ClarificationMiddleware` in both agent branches. Defensive error handling (optional-dep INFO skip, ERROR+traceback otherwise, `issubclass` validation) is solid.

Top asks before merge:
1. **Dead config** - `sse_wrapper` / `run_model_override` + `resolve_extension_hooks()` are never called in production; wire up or remove.
2. **`test_instantiation_failure_skipped`** passes for the wrong reason (function-local class is unresolvable).
3. **Global `AgentMiddleware.__init_subclass__` monkey-patch** at import leaks across the whole test session.
4. **Out-of-scope `checkpoint_retention`** block in `config.example.yaml` has no implementation.
5. **Config docs** - "dotted paths" should be colon-form `module:Attr`; `enabled: false` is silently ignored (no such field); no example added to the templates.

7 inline comments follow. Details in each.

WillemJiang:
@nankingjing  Please fix the lint error.

backend/packages/harness/deerflow/agents/factory.py
nankingjing: Done in fc65793 — removed `sse_wrapper` / `run_model_override`, `resolve_extension_hooks()`, and their 6 tests. The PR now ships only the `middlewares` feature. Agreed on the footgun: if the SSE/model-override hooks are wanted later they should come in their own PR wired end-to-end.

backend/packages/harness/deerflow/config/extensions_config.py
nankingjing: All three addressed in fc65793:

- **Path format**: docs, docstrings, field descriptions, and the PR body now consistently say `module:Attribute` colon-form (matching `resolve_variable`'s `rsplit(":", 1)`); the misleading "dotted paths" wording is gone.
- **`enabled`**: added a real `enabled: bool = True` field on `ExtensionsConfig`, honored by `load_extension_middlewares()` (covered by `test_disabled_returns_empty`).
- **Templates**: `extensions_config.example.json` now carries a `middlewares` key, and `config.example.yaml` gained a commented section documenting the feature. While doing this I found that `AppConfig.from_file()` overwrites any YAML `extensions:` section with `ExtensionsConfig.from_file()` (i.e. `extensions_config.json`), so declaring middlewares in `config.yaml` would silently do nothing — the docs now direct operators to `extensions_config.json`, and a new integration test (`test_app_config_loads_extensions_from_json_not_yaml`) pins that behavior.

backend/packages/harness/deerflow/agents/factory.py
nankingjing: Verified and documented in fc65793:

- **Auth**: the only Gateway write path for this file is `PUT /api/mcp/config` (`app/gateway/routers/mcp.py`), which requires an admin user via `require_admin_user`, and it only rewrites `mcpServers`/`skills` — every other top-level key (including `middlewares` and `mcpInterceptors`) is preserved verbatim from the on-disk file, never taken from the request body. The skills router likewise writes only `mcpServers` + `skills`. So `middlewares` is not writable via any API; changing it requires filesystem access, the same bar as the pre-existing `mcpInterceptors` (arbitrary import) and `mcpServers.command` (arbitrary process execution) in the same file.
- **Docs**: the trust assumption is now stated in the `load_extension_middlewares` docstring, the `ExtensionsConfig` docstring, the root `AGENTS.md` runtime-config paragraph, the `backend/AGENTS.md` extensions_config.json schema, and the `config.example.yaml` comment block.

I considered an allowlist, but for a downstream-integrator extension point the classes live in the integrator's own packages, and the repo's existing pattern (`mcpInterceptors`, `sandbox.use`, config-declared tools — all `resolve_variable` under operator trust) is unrestricted import. Happy to add a namespace restriction if you'd prefer it anyway.

backend/tests/test_extensions_config.py
nankingjing: Fixed in db1be47 — the module-level patch was replaced by an `autouse` pytest fixture that saves the original `AgentMiddleware.__init_subclass__`, installs the no-op for the duration of each test in this module, and restores the original afterwards, so nothing leaks into other test modules.

backend/tests/test_extensions_config.py
nankingjing: Fixed in db1be47 — `_BrokenMiddleware` is now module-level (with a docstring explaining why), so `resolve_variable` resolves it and `cls()` raising `RuntimeError("boom")` actually exercises the `except Exception` instantiation branch.

backend/tests/test_extensions_config.py
nankingjing: Fixed across db1be47 + fc65793 — the `pass` test is gone, the 6 `resolve_extension_hooks` tests were removed along with the dead code, and the class was replaced by `TestExtensionsJsonIntegration` with real assertions: `ExtensionsConfig.from_file` parsing middlewares from an actual JSON file, and an `AppConfig.from_file` integration test pinning that extensions come from `extensions_config.json` (an `extensions:` section in `config.yaml` is ignored). The suite is now 19 tests, all asserting behavior.

config.example.yaml
nankingjing: Removed in db1be47 — the commented `checkpoint_retention` block was out of scope and unimplemented; `config.example.yaml` no longer carries it.

nankingjing:
@fancyboi999 @WillemJiang — all review feedback is addressed as of fc65793:

- **Import blocker**: the stray `ToolStreamingConfig` import/field were removed earlier (197a784); `from deerflow.config.app_config import AppConfig` imports cleanly and backend test collection works on the current head.
- **Lint**: `ruff check` and `ruff format --check` pass on the full backend locally; `lint-backend` was already green on the previous head (db1be47) and is running on this one.
- **willem-bd's 7 inline comments**: each answered inline — dead `sse_wrapper`/`run_model_override`/`resolve_extension_hooks()` removed (+ their 6 tests), a real `enabled` gate added and honored, `module:Attribute` colon-form documented everywhere, templates updated (`extensions_config.example.json` gets `middlewares`, `config.example.yaml` documents the feature), the `AgentMiddleware.__init_subclass__` monkey-patch is now a scoped autouse fixture, `_BrokenMiddleware` is module-level so the instantiation-failure branch is genuinely exercised, the no-op test was replaced with real integration tests, and the out-of-scope `checkpoint_retention` block is gone.
- **Correctness find along the way**: `AppConfig.from_file()` populates `extensions` from `extensions_config.json` and ignores any `extensions:` key in `config.yaml`, so docs and tests now target the JSON file (pinned by `test_app_config_loads_extensions_from_json_not_yaml`).
- **Security** (willem-bd's note): trust model documented in the factory/config docstrings, AGENTS.md, and config.example.yaml; verified `PUT /api/mcp/config` is admin-gated and only rewrites `mcpServers`/`skills` — `middlewares` is never writable via the API.

Local verification: 19/19 `test_extensions_config.py`, plus the adjacent app-config/reflection/factory/lead-agent suites (106 tests) all pass; full-backend ruff check + format clean.

@fancyboi999 could you take another look when you get a chance? Thanks!

nankingjing:
Note: the CI workflows for the new head (fc65793) are queued as **awaiting maintainer approval** (fork PR workflow gating), so the checks won't show until someone approves the run. Local verification in the meantime: 19/19 `test_extensions_config.py`, the adjacent app-config/reflection/factory/lead-agent suites (106 tests) green, and full-backend `ruff check` + `ruff format --check` clean.

fancyboi999 (changes requested):

[P1] Do not classify every resolver `ImportError` as an absent optional package

- Location: `backend/packages/harness/deerflow/agents/factory.py:393-398`
- Problem: `resolve_variable()` deliberately wraps malformed paths, missing attributes, and import failures from inside an installed module as `ImportError`. This handler treats all of them as "module not found", logs only at INFO, and skips the middleware. A typo or broken dependency can therefore leave the agent running without a configured domain guard or tool-governance middleware.
- Evidence: `deerflow/reflection/resolvers.py` raises `ImportError` for a missing `module:Attribute` separator and a missing attribute, and re-raises an installed module's internal import failure as `ImportError("Error importing module ...")`. All three land in this broad branch even though only direct absence of the declared optional package is a legal skip. The PR's target use case includes deterministic guards and tool-call gating, so failing open here can change production behavior rather than merely disable observability.
- Suggested fix: skip only when the direct missing module is the declared optional package (for example, inspect the underlying `ModuleNotFoundError.name` against the declared module root). Let malformed paths, missing attributes, transitive import failures, invalid classes, and instantiation failures abort agent construction with the existing actionable error rather than starting without the configured middleware. Reusing `resolve_class(path, AgentMiddleware)` will also keep class validation aligned with the repository's other extension points.
- Test: cover (1) an absent declared top-level package, which may be skipped, (2) an installed module that raises an internal/transitive `ImportError`, (3) a missing attribute or malformed path, and (4) a constructor failure; cases 2-4 must not produce a runnable agent with the middleware silently absent.



Reviewed head SHA: `e76b647208145d6f520953cac74b32629555c09f`

nankingjing:
@fancyboi999 Thanks — agreed on the P1, and it's fixed at the new head `e3bd052`.

**What changed** (`backend/packages/harness/deerflow/agents/factory.py`):
- The broad `except ImportError` optional-dependency branch is gone. The loader now resolves via `resolve_class(path, AgentMiddleware)` (aligning class validation with `sandbox.use`, skills storage, and config-declared tools), and a new `_declared_module_missing()` helper walks the exception chain to the underlying `ModuleNotFoundError` and compares its `.name` against the declared module path.
- Only a genuinely absent declared module (the `ModuleNotFoundError` names the declared module or one of its parent packages) is an INFO skip. Everything else fails agent construction with the resolver's actionable error: malformed paths (missing `:`), missing attributes, an installed module raising an internal/transitive `ImportError`, non-`AgentMiddleware` classes (`ValueError` from `resolve_class`), and constructor failures (wrapped in a `RuntimeError` carrying the path). A duplicate declared path also raises up front with a clear message instead of surfacing later as LangChain's bare "remove duplicate middleware instances" assertion.

**Tests** — the four cases you listed are pinned in `backend/tests/test_extensions_config.py` (plus two more):
1. absent declared top-level package → skipped (`test_module_not_found_skipped`, `test_mixed_valid_and_missing_optional`)
2. installed module with an internal/transitive import failure → raises (`test_transitive_import_failure_raises`, uses a real on-disk module whose import fails on a missing dep)
3. malformed path / missing attribute → raises (`test_malformed_path_raises`, `test_missing_attribute_raises`)
4. constructor failure → raises with the path and original cause (`test_instantiation_failure_raises`)
plus `test_non_middleware_class_raises` and `test_duplicate_path_raises`. Revert-verified: with the old loader restored, all the new fail-closed tests fail.

Docs updated to match (fail-closed semantics in the `load_extension_middlewares` docstring, `config.example.yaml`, and the `backend/AGENTS.md` schema entry — which now also references the Custom-middlewares slot by name instead of a hardcoded number, so it stays correct against the renumbered middleware chain on current main).

Local gates on `e3bd052`: full-backend `ruff format --check` + `ruff check` clean; 22/22 `test_extensions_config.py`; adjacent app-config/reflection/lead-agent suites (87 tests) and the client/durable-context suites (178 tests) all green. Note: the push was a force-push (commit-message metadata cleanup on the branch's base commit — file trees of all previous commits are byte-identical, the only content change is the new fix commit), and fork CI will need a maintainer approval to run again.

nankingjing:

Conflicts resolved:
- `lead_agent/agent.py`: combined upstream's `normalize_middleware_state_schemas(...)` wrapper + `authorization_provider` with the branch's `custom_middlewares=extension_middlewares` wiring (both call sites).
- `config.example.yaml`: kept upstream's `database` block (pool_recycle/command_timeout/checkpoint_channel_mode); dropped the stale checkpoint-retention comment (superseded by main's own retention work).
- `backend/AGENTS.md` / `extensions_config.py`: merged the `middlewares` doc entry with upstream's config.yaml-override wording; kept both `enabled` and `middlewares` fields.
- Updated `test_app_config_loads_extensions_from_json_not_yaml` to match upstream's documented yaml-wins-per-field override semantics (added a JSON-source-of-truth test for the yaml-silent case).

Checks: `ruff format --check` + `ruff check` clean; `pytest tests/test_extensions_config.py` 23/23 passed. New head: d7ad37ba.

willem-bd (commented):
## Verdict: Request changes (blocker)

Thanks for the thorough error-handling design and docs. The feature can't ship as wired, though - it double-loads every configured middleware against an existing loader already on `main`.

### Blocker
`main` already implements this feature: `agents/middlewares/configured_extensions.py::load_configured_extension_middlewares(app_config)` reads the same `app_config.extensions.middlewares`, resolves via the same `resolve_class(path, AgentMiddleware)`, instantiates with `cls()`, and is already wired into `build_middlewares` (unchanged by this PR). This PR adds a parallel `load_extension_middlewares` and feeds it via `custom_middlewares=`, so both loaders run on the same config -> each middleware is appended twice. Per this PR's own comment (`factory.py:133`), LangChain's `create_agent` rejects duplicate middleware names with a bare `AssertionError`, so agent construction crashes for any non-empty `middlewares`. The existing loader also doesn't check `enabled`, so `enabled: false` does **not** disable middlewares - the gate is ineffective.

**Suggested fix:** consolidate the PR's genuine improvements (enabled gate, optional-dep skip, duplicate-path detection, ctor-error wrapping, path trimming) into the existing `load_configured_extension_middlewares`, and drop the new `load_extension_middlewares` + `custom_middlewares=` wiring.

### Genuine improvements worth preserving
- Optional-dependency skip (absent declared module -> INFO + skip). `_declared_module_missing`'s `__cause__` walk is correct - verified `resolve_variable` chains every re-raise with `from err`, and the `name` vs `module_path` check correctly distinguishes a missing declared module (skip) from transitive / missing-attr / malformed / non-middleware / ctor failures (raise).
- Duplicate-path detection, whitespace trimming, `RuntimeError` wrapping for constructor failures.

### Security - verified sound
Confirmed `backend/app/gateway/routers/mcp.py::_apply_mcp_config_update` only rewrites `mcpServers`/`skills` and preserves all other top-level keys verbatim, so `PUT /api/mcp/config` cannot write `middlewares`. Trust model is sound.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -613,6 +613,13 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
 
     enabled_skills = _load_enabled_available_skills(available_skills, app_config=resolved_app_config, user_id=resolved_user_id)
 
+    # Load config-declared extension middlewares (issue #3923).
+    # Resolved at agent-build time so downstream integrators can register
+    # custom middlewares without forking the assembly code.
+    from deerflow.agents.factory import load_extension_middlewares
+
+    extension_middlewares = load_extension_middlewares(resolved_app_config)
```
willem-bd, line 633: 🔴 **Blocker - double-load.** This loads `extensions.middlewares` via the new `load_extension_middlewares`, but `build_middlewares` (which this is fed into via `custom_middlewares=`) **already** calls `load_configured_extension_middlewares(resolved_app_config)` on the same `extensions.middlewares` field (~line 441, unchanged from `main`).

So each configured middleware is instantiated and appended **twice** to the chain. Your own comment in `factory.py:133` notes LangChain's `create_agent` rejects duplicate middleware names with a bare `AssertionError` - two instances of the same class share the class name -> agent construction crashes for any non-empty `middlewares` config.

Corollary: the existing `load_configured_extension_middlewares` doesn't check `enabled`, so setting `enabled: false` still loads middlewares via the old path - the gate is ineffective.

**Fix:** move the improvements (enabled gate, optional-dep skip, dedup, ctor wrapping) into the existing `load_configured_extension_middlewares` and drop this parallel loader + the `custom_middlewares=` wiring.

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -145,6 +163,10 @@ class ExtensionsConfig(BaseModel):
         default_factory=dict,
         description="Map of skill name to state configuration",
     )
+    middlewares: list[str] = Field(
```
willem-bd, line 166: 🟠 **Duplicate field.** `middlewares` is declared twice in this model - here and at line 153. Pydantic v2 uses the last definition, so the first `Field` (description/metadata) is dead. This looks like an edit meant to *update* the existing field's description rather than add a second one. Remove one (keep the existing slot, update its description).

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -130,8 +130,26 @@ class SkillStateConfig(BaseModel):
...
+    at agent-build time via :func:`deerflow.reflection.resolvers.resolve_class`.
+    Set ``enabled: false`` to disable extension middleware loading entirely.
+
+    Security: middleware entries are imported and instantiated in-process,
+    so they are operator-declared code by design (same trust model as
+    ``mcpServers`` commands and ``mcpInterceptors``).  Keep
+    ``extensions_config.json`` writable only by trusted operators.
+    """
 
+    enabled: bool = Field(
```
willem-bd, line 149: 🟡 **Naming/scope concern.** A top-level `ExtensionsConfig.enabled` that gates **only** middleware loading (not `mcpServers`/`skills`) is surprising - an operator setting `enabled: false` would reasonably expect all extensions disabled. If it's middleware-specific, consider `middlewares_enabled`; if it should gate all extension loading, wire it into the MCP/skills loaders too. (Moot if the blocker is fixed by consolidating into the existing loader, but the naming still applies.)

backend/tests/test_extensions_config.py
```diff
@@ -0,0 +1,398 @@
...
+        assert len(result) == 1
+        assert isinstance(result[0], _ValidMiddleware)
+
+
+# ---------------------------------------------------------------------------
+# AppConfig integration
+# ---------------------------------------------------------------------------
+
+
+class TestExtensionsJsonIntegration:
```
willem-bd, line 320: **Test gap.** These integration tests cover config *loading* (`AppConfig.from_file`), but none exercise `_make_lead_agent` / `build_middlewares` end-to-end to assert the chain contains **exactly one** instance per configured path. That's precisely where the double-load blocker lives, which is why it wasn't caught. An end-to-end test (configure one middleware -> build the agent -> assert a single instance in the chain) would prevent regressions here.

extensions_config.example.json
```diff
@@ -3,6 +3,7 @@
   "mcpInterceptors": [
     "my_package.mcp.auth:build_auth_interceptor"
   ],
+  "middlewares": [],
```
willem-bd, line 6: Duplicate JSON key — `"middlewares"` is already declared at line 2 (carried over from `main`), and this PR adds a second occurrence here. RFC 8259 parsers take the last value, so behavior is unchanged while both are `[]`, but a future editor updating only one will silently leave the other stale and operators reading the template won't know which is canonical. Remove one — keeping the line 2 slot is the more discoverable position.

willem-bd (commented):
One new nit from re-reading the template file end-to-end: `extensions_config.example.json` now declares `"middlewares"` twice (line 2, carried over from `main`, and the new line 6 added by this PR). JSON parsers take the last value per RFC 8259 so it isn't behaviorally broken, but it's confusing for operators reading the template — and a future edit touching only one of the two will silently leave the other stale. Drop one (the top-of-file slot at line 2 is the more discoverable position; or consolidate all extension keys together). Everything else from my prior pass (double-load blocker on `agent.py:633`, duplicate `middlewares` field on `ExtensionsConfig`, `enabled` scope, end-to-end test gap) is unchanged at this head and still stands.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
ShenAC-SAC: Revalidated this blocker against current head `5097fb028b29898dac0250582be7c926aaa95a4b`. It is still present in both the normal and bootstrap agent paths.

I added assembly regressions that retain the real `_make_lead_agent()`, `build_middlewares()`, and LangChain `create_agent()` (using a fake model, without making model calls). For each branch:
- One valid configured middleware with `enabled=true` fails compilation with `AssertionError: Please remove duplicate middleware instances.`
- The same config with `enabled=false` compiles, but still constructs the supposedly disabled middleware once.
- A missing optional module is skipped by the new loader, then raises `ImportError` when `build_middlewares()` invokes the existing loader.

All six review regressions fail, while the 64 existing tests in `test_extensions_config.py` and `test_lead_agent_model_resolution.py` pass. The new loader's helper behavior therefore does not establish the final graph's behavior.

Suggested concrete fix:
1. Move the intended enabled gate, path normalization/deduplication, error classification, and constructor diagnostics into `agents/middlewares/configured_extensions.py::load_configured_extension_middlewares()`.
2. Remove this second loading call and both new `custom_middlewares=extension_middlewares` arguments. Keep `custom_middlewares` available for its existing programmatic callers; configured middleware should enter the chain only through the existing configured-middleware slot.
3. Add graph-construction regressions for normal/bootstrap agents asserting exactly one constructor invocation when enabled, zero when disabled, and the intended behavior for missing optional modules. Preserve fail-closed handling for malformed paths, missing attributes, transitive import errors, and constructor failures.
4. Cover the embedded-client and subagent consumers of that shared loader so they honor the same configuration behavior.

This consolidates loading at the existing shared boundary and fixes all three observed failures without introducing a separate extension path. Validation was local graph assembly against the remote head; no external model/service calls or full backend suite were run.

nankingjing:
Closing this, because `main` already ships the feature and this branch cannot assemble as wired.

**Already on `main`.** `backend/packages/harness/deerflow/agents/middlewares/configured_extensions.py` defines `load_configured_extension_middlewares(app_config)`, reading the same `app_config.extensions.middlewares`, resolving classes the same way, and it is already wired into the middleware chain and covered by `backend/tests/test_configured_extensions.py`. This PR adds a second, parallel loader fed through `custom_middlewares=`, so both run over the same config and every middleware is appended twice.

**The blocker is current, not historical.** It was raised in review on 2026-07-26 and revalidated against head `5097fb028b29898dac0250582be7c926aaa95a4b` on 2026-09-23, with assembly-level regressions that exercise the real `_make_lead_agent()`, `build_middlewares()` and LangChain `create_agent()`:

- one valid configured middleware with `enabled=true` fails compilation with `AssertionError: Please remove duplicate middleware instances.`;
- the same config with `enabled=false` compiles but still constructs the supposedly disabled middleware;
- a missing optional module is skipped by the new loader and then raises `ImportError` when the existing `build_middlewares` reaches it.

The branch has also sat unmergeable since the July rebase.

**What is worth keeping.** The review called out real improvements here that would be better folded into the existing loader than carried by a second one: the `enabled` gate, skipping genuinely-absent optional dependencies, duplicate-path detection, constructor-error wrapping, and path trimming. The 398 lines of config-loading tests in `backend/tests/test_extensions_config.py` are also worth carrying over, extended with the end-to-end assertion the review asked for (one middleware configured in, exactly one instance in the built chain). Separately, `extensions_config.example.json` declares `"middlewares"` twice — harmless to a parser, but worth dropping one occurrence.

Closing to stop this sitting as a stale request-changes. The branch and commits stay on the fork, so a consolidated follow-up can take the pieces listed above without re-deriving them.

PR #4150, opened 2026-07-13
feat(tools): real-time streaming tool output

nankingjing:
Stream tool output in real-time like Claude Code — currently users only see tool results when the entire call completes.

### Backend
- **ToolStreamingMiddleware**: wraps async tool calls to emit `tool_output_chunk` custom stream events (start, error, final completion signal); tools can emit intermediate partial chunks themselves via `langgraph.config.get_stream_writer()`
- **ToolStreamingConfig**: single `enabled` flag (default false)
- **worker.py**: always includes `custom` stream mode so `get_stream_writer()` works (also used by LLM retry hints, SafetyFinishReasonMiddleware, and the task tool)
- Positioned **inner** of ToolErrorHandlingMiddleware so raw tool exceptions reach the streaming error-chunk path first; `GraphBubbleUp` control-flow signals are re-raised untouched; the ordering is pinned by a regression test

### Frontend
- **ToolStreamingProvider** + `useToolStreaming` / `useToolCallStream` hooks: streaming state keyed by `tool_call_id`; partial chunks accumulate; a final or error chunk tears the entry down (the canonical ToolMessage carries the full result), so the map only ever holds actively-streaming calls
- **SubtaskCard**: renders live tool output scoped to the card's own active tool calls
- **hooks.ts**: maps `tool_output_chunk` custom stream events via `toolStreamUpdateFromEvent`
- **backend/AGENTS.md**: middleware-chain section documents the new middleware

## Design
- **Disabled by default** — fully backward compatible
- **Graceful degradation** — if the stream writer is unavailable the middleware passes through silently; a frontend without streaming events simply never populates the store
- **Error resilience** — a tool exception emits an error chunk before re-raising for ToolErrorHandlingMiddleware (outer) to convert to an error ToolMessage
- **No double payload** — the final chunk carries only a short first-line preview (max 200 chars) as a completion signal; the full result arrives via the canonical ToolMessage

## Tests
- Backend 29 (test_tool_streaming_middleware.py): chunk builders, content extraction, enable/disable toggle, stream-writer fallback, start/final/error chunk emission, writer exception handling, dict-format regression, and a graph-level regression driving the middleware through a real compiled LangGraph `astream(stream_mode="custom")` channel (no writer mock); plus test_tool_streaming_is_inner_of_tool_error_handling in test_tool_error_handling_middleware.py pinning the middleware ordering
- Frontend 11 (tool-streaming.test.ts): exercises the exported production reducer and event mapper directly — guard direction (final accepted, late partial dropped), partial accumulation, per-tool-call scoping (P1-2), teardown on final/error, provider rendering


backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
```diff
@@ -0,0 +1,213 @@
...
+        writer = _get_stream_writer()
+        if writer is None:
+            # Custom stream mode not active — silently pass through.
+            # This is not an error: the middleware is enabled but the run
+            # may not have been configured with stream_mode="custom".
+            return await handler(request)
+
+        # Emit start-of-execution chunk so the frontend knows a tool is running.
+        try:
+            writer((TOOL_OUTPUT_CHUNK_EVENT, _build_start_chunk(tool_call_id, tool_name)))
```
fancyboi999: [P1] Emit the custom-event object directly. The run worker publishes custom chunks verbatim, so this tuple serializes to `["tool_output_chunk", { ... }]`; the frontend `onCustomEvent` handler only reads a top-level `event.type`, making every new branch unreachable. Existing retry/safety writers call `writer({"type": ...})`. Please use that contract and add a graph-to-SSE regression test rather than another writer mock.

frontend/src/components/workspace/messages/subtask-card.tsx
```diff
@@ -233,6 +235,30 @@ export function SubtaskCard({
               />
             );
           })}
+          {/* Streaming tool output: show real-time output from the currently executing tool */}
+          {task.status === "in_progress" && streamingOutput && (
```
fancyboi999: [P1] Scope output to the owning task/tool call. `ToolStreamingProvider` stores one page-global output, this card checks only `task.status`, and `hooks.ts` discards `tool_call_id`; with two parallel in-progress subtasks every card renders the same tool output. Key the state by run/task/tool-call identity and render only the matching card, with a parallel-subtask regression test.

fancyboi999 (changes requested):
This head has two independent correctness blockers in the end-to-end custom-event path and parallel-task rendering. The focused middleware tests pass because they stop at a mocked writer; they do not exercise worker serialization, SSE delivery, or multiple in-progress subtasks. The implementation also emits only start/final lifecycle events after the handler returns: no existing producer applies `min_chunk_size` / `max_buffer_seconds` or forwards actual incremental tool output, so the claimed real-time behavior still needs one real producer and an integration test.

WillemJiang:
@nankingjing, please set up a pre-commit hook in your development environment to avoid the lint error before committing code.  It's quite simple: just run 'make install' from the repo's root directory.

nankingjing:
Done — pre-commit hooks installed via `pre-commit install`, thanks for the tip.

nankingjing:
Addressed both P1 review items from @fancyboi999:

**P1-1: Dict format for custom events** — The original code passed writer((event_name, data)) which serialised to a JSON array ["tool_output_chunk", {...}]. Now emit the dict directly: writer(_build_start_chunk(...)). Added TestWriterReceivesDictFormat regression tests (4 cases).

**P1-2: Scoped output per tool_call_id** — Keyed ToolStreamingProvider state by tool_call_id (Record<string, ToolStreamOutput>), filter SubtaskCard to active call ids, added useToolCallStream. Added tool-streaming.test.tsx with unit tests.

Lint (ruff + eslint + prettier) passes locally.

frontend/src/core/tasks/tool-streaming.tsx
```diff
@@ -0,0 +1,106 @@
...
+          }
+          const next = { ...prev.outputs };
+          delete next[toolCallId];
+          return { outputs: next };
+        }
+        // Prevent a stale final chunk from overwriting a newer partial chunk
+        // from a subsequent tool call of the same subagent (late-arriving
+        // start/final sequence could race).
+        const existing = prev.outputs[toolCallId];
+        if (existing && !output.isPartial && existing.isPartial) {
```
willem-bd: 🔴 **Blocker — the final (complete) tool output never renders.**

This guard is inverted. The backend emits exactly two chunks per call (`tool_streaming_middleware.py`):

| chunk | `is_partial` | `chunk` |
|---|---|---|
| start | `True` | `""` |
| final | `False` | full output |

Trace through this provider:
1. `start` → stored as `{isPartial: true, text: ""}` (first call, `existing` is undefined → guard skipped).
2. `final` → `existing.isPartial` is `true`, `!output.isPartial` is `true` → **the final chunk is rejected**, `outputs[id]` stays `{isPartial: true, text: ""}` forever.

So in `SubtaskCard`, `entry.output.text` is always `""` → it renders the `<Shimmer>` branch permanently, the spinner spins forever, and the actual tool output is never shown. `useIsToolStreaming(id)` also stays `true` indefinitely.

The comment's rationale ("stale final from a subsequent tool call") can't occur — `tool_call_id`s are UUIDs, a subsequent call has a different id. The real risk is the opposite: a late *partial* clobbering a completed *final*. Flip the condition:

```ts
if (existing && output.isPartial && !existing.isPartial) {
  return prev;
}
```

The test `stale final chunk does not overwrite a newer partial chunk` encodes the wrong model (it treats the normal partial→final order as a stale-final race) and needs inverting too. The unit tests all pass today despite this because they don't drive a real start→final sequence through provider state.

Separately: entries are never removed (the `null` path in `updateToolStream` is never invoked from `hooks.ts`), so the `outputs` map grows for the whole thread — add an explicit teardown on `is_final`/error.

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
```diff
@@ -0,0 +1,210 @@
...
+
+        # Emit start-of-execution chunk so the frontend knows a tool is running.
+        try:
+            writer(_build_start_chunk(tool_call_id, tool_name))
+        except Exception:
+            logger.debug("Failed to emit tool start chunk for %s/%s", tool_name, tool_call_id, exc_info=True)
+
+        try:
+            result = await handler(request)
+        except Exception as exc:
```
willem-bd, line 211: 🟠 **This `except` path is effectively dead, and it mishandles `GraphBubbleUp`.**

`ToolStreamingMiddleware` sits **outer** of `ToolErrorHandlingMiddleware` (`tool_error_handling_middleware.py:247-260`, first-in-list = outermost). `ToolErrorHandlingMiddleware.awrap_tool_call` catches **all** `Exception`s (except `GraphBubbleUp`) and converts them to error `ToolMessage`s, returning normally.

So for a real tool failure, `ToolErrorHandling` (inner) catches first → `ToolStreaming` (outer) sees a normal return, never entering this `except`. Consequences:

1. The `error: True` chunk (`_build_error_chunk`) is never emitted for tool errors — the error text instead leaks out as a **final** chunk (extracted from the error `ToolMessage` via `_extract_content`), with no error flag. The frontend's `isError` path is never exercised.
2. The inline comment here ("ToolErrorHandlingMiddleware (inner) will catch this and convert it…") is backwards — `ToolErrorHandling` already caught it *before* this `except` could fire, because it's inner.
3. `test_emits_error_chunk_on_exception` / `test_error_chunk_truncated` pass only because they call `awrap_tool_call` directly with a raising handler, bypassing `ToolErrorHandling`.
4. **`GraphBubbleUp` is treated as an error.** `ToolErrorHandling` deliberately re-raises `GraphBubbleUp` (LangGraph interrupt/pause/resume control-flow). This bare `except Exception` then catches it, emits a spurious error chunk (`error_text` falls back to `"GraphBubbleUp"`), and re-raises. Add `except GraphBubbleUp: raise` before the generic handler, as `ToolErrorHandling` does.

If the intent is for `ToolStreaming` to see real errors, it must sit **inner** of `ToolErrorHandling`. As built, the error-chunk feature is non-functional.

frontend/src/core/threads/hooks.ts
```diff
@@ -1048,6 +1050,25 @@ export function useThreadStream({
...
+          tool_call_id: string;
+          tool_name: string;
+          chunk: string;
+          is_partial: boolean;
+          is_final: boolean;
+          error?: boolean;
+        };
+        updateToolStream(e.tool_call_id, {
+          toolName: e.tool_name,
+          text: e.chunk,
```
willem-bd: 🟠 **Partial chunks are overwritten, not accumulated.**

`ToolStreamOutput.text` is documented as *"Accumulated output text so far"* (`types.ts:10`), but this sets `text` to the latest chunk only — each partial replaces the previous. This is invisible today because the backend's `ToolStreamingMiddleware` emits only start + final, but its docstring advertises that tools can emit *"intermediate (optional)"* partial chunks via `get_stream_writer()`. If any tool does, partial output would render chunk-by-chunk overwrite instead of growing.

Either accumulate (in `updateToolStream`'s reducer: when both existing and incoming are `isPartial`, append `e.chunk` to `existing.text`) or drop the "intermediate chunks" claim from the middleware docs.

backend/packages/harness/deerflow/config/tool_streaming_config.py
```diff
@@ -0,0 +1,29 @@
...
+    streaming (e.g. bash) can emit intermediate chunks via
+    ``langgraph.config.get_stream_writer()`` during execution so the frontend
+    renders partial results in real-time instead of waiting for the full result.
+    """
+
+    enabled: bool = Field(
+        default=False,
+        description="Enable real-time tool output streaming via custom stream events.",
+    )
+    min_chunk_size: int = Field(
```
willem-bd: 🟡 **`min_chunk_size` / `max_buffer_seconds` are dead config.**

`ToolStreamingMiddleware.__init__` stores both (`tool_streaming_middleware.py:165-166`) but never reads them — the middleware emits exactly two unbuffered chunks (start + final) with no coalescing. The descriptions promise behavior ("coalesced to prevent UI jank", "buffer streaming output before emitting a chunk") that isn't implemented. Either implement the buffering (also what would make partial streaming useful) or drop the fields so the config doesn't promise behavior that doesn't exist.

frontend/src/core/tasks/types.ts
```diff
@@ -30,4 +41,6 @@ export interface Subtask {
    * badge can show "capped" without parsing result text (#3875 Phase 2).
    */
   stopReason?: string;
+  /** Live streaming tool output from ``tool_output_chunk`` custom events. */
+  streamingTool?: ToolStreamOutput;
```
willem-bd: 🟡 **`Subtask.streamingTool` is dead — nothing writes it.**

This field is added here, threaded into `subtaskChanged` (`subtask-update.ts`), and covered by two new tests, but nothing ever populates `streamingTool` on a `Subtask`. The live streaming state actually lives in `ToolStreamingProvider.outputs` (keyed by `tool_call_id`) and is read via `useToolStreaming()` in `SubtaskCard`. The field, the change-detection line, and its tests are vestigial — remove, or wire it up if it was intended as the transport.

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -425,6 +425,12 @@ def _continuation_runnable_config() -> RunnableConfig:
         if not lg_modes:
             lg_modes = ["values"]
 
+        # Always include "custom" so middlewares can emit custom stream events
+        # (e.g. ToolStreamingMiddleware).  Including it adds zero overhead when
+        # no middleware writes to the custom channel — LangGraph only emits
+        # events when get_stream_writer() is called.
+        lg_modes.append("custom")
```
willem-bd: 🟡 **`"custom"` is appended unconditionally for an opt-in feature.**

This runs for every run regardless of `tool_streaming.enabled`. The "zero overhead" claim is approximate — it adds the custom-channel writer/forwarding to all runs globally. Consider gating on `app_config.tool_streaming.enabled`.

Also: is `_continuation_runnable_config` the sole stream-mode source? If the initial run config is built elsewhere without `"custom"`, the first step wouldn't forward custom events while continuations do — worth confirming the initial path also gets it (or that this covers it).

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
```diff
@@ -0,0 +1,210 @@
...
+            try:
+                writer(_build_error_chunk(tool_call_id, tool_name, error_text))
+            except Exception:
+                logger.debug("Failed to emit tool error chunk for %s/%s", tool_name, tool_call_id, exc_info=True)
+            raise
+
+        # Emit final chunk with the complete tool output.
+        content = _extract_content(result)
+        try:
+            writer(_build_final_chunk(tool_call_id, tool_name, content))
```
willem-bd, line 229: 🟡 **Full tool output is sent twice.**

The final chunk ships the entire tool output via the custom channel, and the same content also arrives as the canonical `ToolMessage` in the graph state update. For large outputs (big file reads, long command logs) this doubles the payload per tool call when streaming is enabled. Consider emitting only a completion signal (or a short preview) on the final chunk, since the full result already arrives through the state.

willem-bd (commented):
Code review of this PR. Verified file contents/line numbers against the PR head commit (`f2572be`). Summary of findings — details in the inline comments:

- 🔴 **Blocker**: the "stale final chunk" guard in `tool-streaming.tsx` is inverted, so the final (complete) tool output never renders — the UI shows a perpetual shimmer. The feature's core purpose doesn't work when enabled.
- 🟠 The error-chunk path in `ToolStreamingMiddleware` is effectively dead (it sits outer of `ToolErrorHandlingMiddleware`, which catches tool exceptions first) and mishandles `GraphBubbleUp`. Tests pass only because they bypass the real middleware chain.
- 🟠 Partial chunks are overwritten instead of accumulated (`hooks.ts`), so the advertised intermediate streaming wouldn't actually grow.
- 🟡 `min_chunk_size`/`max_buffer_seconds` config and `Subtask.streamingTool` are dead code.
- 🟡 `"custom"` stream mode is appended unconditionally for an opt-in feature; full tool output is sent twice.

Positives: clean module structure following existing middleware conventions, default-off/graceful degradation design, good attention to the dict-not-tuple serialization contract and parallel-subtask scoping. Recommend requesting changes given the blocker.

nankingjing:
Addressed the remaining blocker from @willem-bd's review:

**Fix: middleware ordering (error-chunk path was dead)**
The middleware ordering in  had ToolStreaming OUTER of ToolErrorHandling — this meant ToolErrorHandling (inner) caught exceptions first, converted them to error ToolMessages, and ToolStreaming never saw raw exceptions. Swapped the order so ToolStreaming is now INNER: the actual tool raises → ToolStreaming catches, emits error chunk, re-raises → ToolErrorHandling catches and converts.

**Fix: frontend test encoded inverted guard**
The  test used the old inverted guard ( — rejecting final when partial exists). Rewrote it to verify the correct guard ( — rejecting late partial when final already exists). Renamed to  to match rstest include pattern.

**Worker comment update**
Updated the  comment to document all  consumers (ToolStreamingMiddleware, LLMErrorHandlingMiddleware, SafetyFinishReasonMiddleware, task tool) — custom mode stays unconditional.

**Test results**
- Backend: 27 tool-streaming tests pass, 33 tool-error-handling tests pass (including ordering guards), 10 reload-boundary tests pass
- Frontend: 627/627 tests pass (including updated tool-streaming tests)
- Lint: ruff + eslint + prettier all pass

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
nankingjing: Fixed in commit 41b1abee. The middleware now calls writer(dict) directly instead of writer((event_name, data)). Regression tests in TestWriterReceivesDictFormat (4 cases) verify every chunk builder returns a dict with top-level type field, and the actual writer invocation passes a dict not a tuple.

frontend/src/components/workspace/messages/subtask-card.tsx
nankingjing: Fixed in commit 41b1abee. ToolStreamingProvider keys state by tool_call_id (Record<string, ToolStreamOutput>). SubtaskCard filters streamingEntries to only tool_call_ids present in the task's active latestMessage.tool_calls. useToolCallStream(toolCallId) provides per-call access. Unit tests in tool-streaming.test.ts verify two independent tool_call_ids never share output.

frontend/src/core/tasks/tool-streaming.tsx
nankingjing: Fixed. The guard was inverted (rejecting final chunks instead of rejecting late partials). Now correctly drops a late partial when a final already exists: `if (existing and output.isPartial and not existing.isPartial)`. The test was also updated — it previously encoded the old inverted logic and tested the wrong scenario. It now verifies: normal start-to-final flow accepted, late partial after final dropped. Also renamed from .test.tsx to .test.ts to match rstest include pattern.

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
nankingjing: Fixed. The middleware ordering in `_build_runtime_middlewares` was swapped: ToolErrorHandling is now outer and ToolStreaming is inner. This means the actual tool raises -> ToolStreaming catches, emits error chunk, re-raises -> ToolErrorHandling catches and converts to error ToolMessage. GraphBubbleUp is caught separately and re-raised without emitting an error chunk. The docstring in tool_streaming_middleware.py already described the correct architecture; now the code matches it.

frontend/src/core/threads/hooks.ts
nankingjing: Fixed in commit 400d8166. The updateToolStream reducer now accumulates partial chunks: when both existing and incoming are isPartial, the text is appended (`existing.text + output.text`) instead of overwritten. Intermediate partials now grow the preview instead of flashing the last fragment.

backend/packages/harness/deerflow/config/tool_streaming_config.py
nankingjing: Fixed in commit 14bd3388. min_chunk_size and max_buffer_seconds fields are removed from ToolStreamingConfig. The middleware no longer stores or reads them.

frontend/src/core/tasks/types.ts
nankingjing: Fixed in commit 14bd3388. The Subtask.streamingTool field and its change-detection in subtask-update.ts are removed. The live streaming state lives in ToolStreamingProvider.outputs.

backend/packages/harness/deerflow/runtime/runs/worker.py
nankingjing: Updated the comment in worker.py to document all get_stream_writer() consumers (ToolStreamingMiddleware, LLMErrorHandlingMiddleware for retry hints, SafetyFinishReasonMiddleware, and the task tool). The custom mode stays unconditional because these other middlewares and tools also depend on it. The overhead when no one writes is genuinely zero — LangGraph only allocates the writer context when get_stream_writer() is called.

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
nankingjing: Fixed in commit 400d8166. The final chunk now sends only a short preview (first line, max 200 chars) as a completion signal rather than the full tool output. The complete result still arrives via the canonical ToolMessage in the message stream. This avoids the double payload while still letting the frontend know streaming has ended for the tool_call_id.

nankingjing:
@fancyboi999 all P1 items and willem-bd blockers should be resolved. Requesting re-review.

nankingjing:
Pushed 0fc51a0 addressing the remaining unresolved review items found in a self-audit of head 8c1f903:

**1. Teardown on final/error (remainder of @willem-bd's blocker comment).** The `outputs` map never removed entries — the `null` path in `updateToolStream` was unreachable from `hooks.ts`, so the map grew for the lifetime of the thread. `tool_output_chunk` events with `is_final` (success or error) now tear the entry down via `toolStreamUpdateFromEvent`; the store only ever holds actively-streaming calls. The full result still arrives via the canonical ToolMessage.

**2. Frontend regression tests now exercise production code.** The previous `tool-streaming.test.ts` tested an inline re-implementation of the reducer: re-introducing the inverted final-chunk guard or deleting partial accumulation left all 627 tests green (verified by revert-testing). The provider's reducer is now the exported `applyToolStreamUpdate` and the tests import it directly. Re-verified: guard inversion now fails 2 tests, removing accumulation fails 1, removing teardown fails 2.

**3. Backend ordering pinned.** No test asserted ToolStreaming sits inner of ToolErrorHandling — reverting commit 8c1f903's ordering fix passed the full suite. Added `test_tool_streaming_is_inner_of_tool_error_handling` (fails on the swapped order, verified).

**4. Graph-level regression test (@fancyboi999's P1-1 ask, "rather than another writer mock").** `TestGraphToCustomStream` drives the middleware inside a real compiled LangGraph and consumes `astream(stream_mode="custom")` — the exact channel `worker.py` serialises to SSE verbatim — asserting start/final and start/error lifecycles arrive as dicts with top-level `type`.

**5. Docs.** `backend/AGENTS.md` middleware-chain section now documents ToolStreamingMiddleware (placement, error contract, GraphBubbleUp, degradation). PR body updated to match the final diff.

**Local gates (Windows, conda py3.12 backend / pnpm frontend):**
- Backend: `ruff format --check` + `ruff check` clean on all touched files; `test_tool_streaming_middleware.py` + `test_tool_error_handling_middleware.py`: **63 passed** (29 + 34)
- Frontend: `pnpm check` (eslint + tsc) clean; `pnpm format` (prettier) clean; `pnpm test`: **634 passed, 0 failed, 0 skipped** — `tool-streaming.test.ts` executes 11 tests (verified by running the file standalone)
- Merged against current `main` (c9b6131) locally: no conflicts, both backend suites pass on the merge result
- Byte-scan of all changed files: LF-only, UTF-8 clean

frontend/src/core/tasks/tool-streaming.tsx
nankingjing: Follow-up in 0fc51a0: the teardown half of this comment is now implemented — final and error chunks remove the entry (toolStreamUpdateFromEvent returns null for is_final), so the outputs map only holds actively-streaming calls and useIsToolStreaming(id) goes false at completion. The guard/accumulation tests also now exercise the exported production reducer (applyToolStreamUpdate) instead of an inline copy; re-inverting the guard fails 2 tests (verified by revert-testing).

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
nankingjing: Follow-up in 0fc51a0: the ordering is now pinned by test_tool_streaming_is_inner_of_tool_error_handling in test_tool_error_handling_middleware.py — it builds the runtime chain with tool_streaming.enabled=True and asserts ToolErrorHandling sits at a lower index (outer). Verified the test fails when the append order is swapped back.

frontend/src/core/threads/hooks.ts
nankingjing: Follow-up in 0fc51a0: accumulation is now covered by a test against the exported production reducer (applyToolStreamUpdate) — deleting the accumulation branch fails the test (verified by revert-testing). Previously the test file exercised an inline re-implementation, so this fix had no coverage.

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
nankingjing: Follow-up in 0fc51a0: added TestGraphToCustomStream — it drives the middleware inside a real compiled LangGraph and consumes astream(stream_mode="custom") (the channel worker.py publishes to SSE verbatim), asserting start/final and start/error lifecycle chunks arrive as dicts with top-level type. This covers the graph-to-stream leg without a writer mock; the SSE bridge itself forwards custom chunks unmodified.

fancyboi999 (changes requested):
Thanks @nankingjing. The current head fixes the previously reported event shape, reducer, teardown, and middleware-ordering issues, but the subagent-to-parent custom-event path is still disconnected.


Reviewed head SHA: `0fc51a05ef412dfb5e02d9e1c0c0b9b0d610092f`

[P1] Forward subagent custom chunks into the parent run stream

- Location: `backend/packages/harness/deerflow/subagents/executor.py:805`, with the middleware installed from `backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py:263-266`
- Problem: `ToolStreamingMiddleware` is installed in the shared subagent middleware chain, but `SubagentExecutor` runs that graph with `stream_mode="values"` only. The middleware's own contract says `get_stream_writer()` is available only with the custom mode, so subagent tool calls take its pass-through path and emit no `tool_output_chunk`. Even if the subgraph produced custom chunks, the executor currently consumes only state snapshots and `task_tool` forwards only the polled `ai_messages` as `task_running`. No subagent chunk reaches the parent Gateway SSE stream.
- Evidence: the new graph-level test calls the middleware in a standalone graph with `stream_mode="custom"`; it does not drive the real `task` / `SubagentExecutor` path. The frontend's only renderer is `SubtaskCard`, which filters entries by tool-call IDs from the subtask's `latestMessage`. Lead-agent lifecycle events therefore cannot substitute for the missing subagent events, and enabling the feature leaves those cards with no streaming output.
- Suggested fix: add an explicit subagent custom-event bridge. Consume `values` and `custom` from the subagent graph, carry `tool_output_chunk` across the isolated executor boundary, and publish it through the parent task tool's `StreamWriter` without waiting for terminal completion. Preserve `tool_call_id` (and preferably `task_id`) so the existing card scoping remains deterministic.
- Test: run a real parent `task` call whose subagent invokes an async tool that emits an intermediate chunk, then assert the parent run's custom/SSE stream receives `tool_output_chunk` before the canonical ToolMessage. The test must use `SubagentExecutor`, not a standalone one-node graph.

nankingjing:
Thanks for the review @fancyboi999! Addressed the CHANGES_REQUESTED feedback.

**Root cause**: SubagentExecutor used `stream_mode="values"` only, so `ToolStreamingMiddleware` (installed via the shared `_build_runtime_middlewares` chain) received `None` from `get_stream_writer()` inside the subagent context — a silent pass-through. Subagent `tool_output_chunk` events never reached the parent SSE stream.

**Fix** (3 files + tests):

1. `executor.py` `SubagentResult` — added `tool_output_chunks: list[dict[str, Any]]` field
2. `executor.py` `_aexecute` — changed to `stream_mode=["values", "custom"]`; the async for loop now decomposes `(mode, chunk)` tuples, routing custom chunks into result.tool_output_chunks while continuing to handle values chunks as before
3. `task_tool.py` polling loop — forwards new `tool_output_chunks` from the `SubagentResult` to the parent `StreamWriter` after each poll, tracked via `forwarded_chunk_count`

**Tests added**:
- `test_aexecute_accumulates_tool_output_chunks_from_custom_stream` — verifies tuple decomposition when mock agent yields mixed tuple/chunk items
- `test_aexecute_ignores_invalid_tuple_items` — verifies non-"custom" tuples fall through correctly
- `test_task_tool_polling_forwards_tool_output_chunks_to_parent_writer` — verifies chunks forwarded in order (task_started < chunks < task_completed)
- `test_task_tool_polling_skips_forward_when_result_lacks_tool_output_chunks` — legacy compat (no attr crash)

**Gate results**: ruff format --check + ruff check pass; pytest passes for test_subagent_executor (65 tests), test_task_tool_core_logic (40 tests), test_tool_streaming_middleware (29 tests).

**Revert verification**: removing the tuple-handling logic with `stream_mode=["values", "custom"]` produces `AttributeError: tuple object has no attribute get` — confirmed the original code fails when LangGraph yields tuples.

Only backend changed — frontend untouched.

fancyboi999 (changes requested):
Thanks @nankingjing. The subagent custom-event bridge is now connected, but the current head still has three blockers against fresh `main`.


Reviewed head SHA: `acd9ac5b491926a0872fd2b658d68919871bc47f`

[P1] Rebase the subagent test changes onto current `main`

- Location: `backend/tests/test_subagent_executor.py`
- Problem: the current head no longer merges cleanly with `origin/main`, so this PR cannot enter the normal validation or merge path as reviewed.
- Evidence: `git merge-tree --write-tree origin/main HEAD` reports a content conflict in `backend/tests/test_subagent_executor.py`; current `main` changed the same executor test scaffolding in #4268.
- Suggested fix: rebase onto current `main` and preserve both #4268's cancellation/cache coverage and this PR's mixed custom-stream coverage instead of resolving by dropping either side.
- Test: rerun the complete changed backend suites and the frontend gates after the rebase, then confirm `git merge-tree --write-tree origin/main HEAD` succeeds.

[P2] Advance the forwarding cursor by the captured snapshot, not the live list length

- Location: `backend/packages/harness/deerflow/tools/builtins/task_tool.py:439-443`
- Problem: the polling thread slices `tool_output_chunks`, forwards that snapshot, and then sets the cursor from a second live `len()` read. If the subagent appends a chunk while `writer()` is forwarding the snapshot, the new chunk was not forwarded but the cursor advances past it, so it is permanently dropped. Losing a final chunk also leaves the frontend's per-call streaming entry uncleared.
- Evidence: a controlled PR-head `task_tool` run started with only the `start` chunk and appended `final` from the writer callback. The result stored `['start', 'final']`, but the parent writer received only `['start']`; the newly added happy-path test starts with both chunks already present and cannot exercise this interleaving.
- Suggested fix: take one stable snapshot and advance by `len(chunks)` (or expose a locked drain operation on `SubagentResult`) so items appended after the slice remain pending for the next poll.
- Test: block or hook the writer after the first captured chunk, append another chunk from the result producer, then assert the next poll forwards it exactly once and in order.

[P2] Add the opt-in setting to the supported config upgrade path

- Location: `backend/packages/harness/deerflow/config/app_config.py:184`; missing from `config.example.yaml` and user-facing configuration docs
- Problem: `tool_streaming` defaults to disabled, but the PR does not expose the new section in `config.example.yaml` or bump `config_version`. A normal installation copied from the example therefore has no discoverable way to enable the feature, and `make config-upgrade` cannot merge the new field into existing configurations.
- Evidence: the repository's config contract says every schema change must bump `config_version`, and the upgrader merges fields from `config.example.yaml`; this diff changes only the Pydantic schema and internal `backend/AGENTS.md`.
- Suggested fix: add the documented `tool_streaming.enabled` section to `config.example.yaml`, document the user-facing behavior, bump the config version and its Helm mirror per `scripts/check_config_version.sh`, and keep the default off.
- Test: run the config-version consistency check and an upgrade fixture proving an older config gains `tool_streaming.enabled: false` without changing existing values.

nankingjing:
@fancyboi999 All three blockers addressed in 7ae466d:

**[P1] Rebase onto current main** — rebased onto a028dfd5 (current origin/main). Conflict in test_subagent_executor.py from #4268 resolved by keeping the `user_id` param signature. `git merge-tree --write-tree origin/main HEAD` succeeds cleanly.

**[P2] Forwarding cursor race** — now advances by `len(chunks)` (the stable snapshot) instead of a second live `len(result.tool_output_chunks)`. Items appended during `writer()` forwarding stay pending for the next poll.

**[P2] Config upgrade path** — `tool_streaming.enabled` section added to `config.example.yaml`, `config_version` bumped 26→27 in both `config.example.yaml` and `deploy/helm/deer-flow/values.yaml`. `scripts/check_config_version.sh` passes.

Backend tests: 124 passed, 1 skipped. Ready for re-review.

nankingjing:
Rebased onto current main (735f67a5). Merge conflicts resolved — the branch now contains 8 commits on top of latest main. All changes preserved: tool-streaming middleware, subagent executor propagation, cursor race fix, config upgrade path, frontend components. CI should re-run on the new head.

backend/packages/harness/deerflow/tools/builtins/task_tool.py
```diff
@@ -448,6 +449,16 @@ async def task_tool(
                     error=error,
                 )
 
+            # Forward any new tool_output_chunks from the subagent to the
+            # parent stream writer so custom events (emitted by middlewares
+            # like ToolStreamingMiddleware) propagate out of the subagent
+            # context to the lead agent's SSE stream (#4150).
+            if hasattr(result, "tool_output_chunks"):
+                chunks = result.tool_output_chunks[forwarded_chunk_count:]
```
LittleChenLiya: 这里从后台子线程正在追加的 `tool_output_chunks` 列表直接切片/遍历，但该列表由 `SubagentExecutor` 所在线程写入，当前轮询线程同时读取，没有通过 `SubagentResult._state_lock` 取一致快照。具体竞态是：轮询线程先切出旧列表（最后一个 chunk 此时尚未追加）；随后后台线程追加最后 chunk 并把状态设为 `COMPLETED/FAILED`；轮询线程接着读取到终态，马上 cleanup/return，于是这个尾块永久丢失，前端也收不到对应的 final/error teardown。建议让 chunk append 与终态更新共用锁，并提供一次性 snapshot/drain 方法，同时返回待转发 chunks 与状态；再用“切片后追加最后 chunk 并进入终态”的交错回归测试覆盖这个边界。

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
```diff
@@ -0,0 +1,233 @@
...
+    """Emit ``tool_output_chunk`` custom stream events during tool execution.
+
+    Config-driven: ``tool_streaming.enabled`` in config.yaml (default: ``false``).
+    When disabled the middleware is a pure pass-through with zero overhead.
+
+    Lifecycle chunks:
+      - **start**: emitted before the handler is called (``is_partial=True``,
+        ``is_final=False``, empty ``chunk``).  Signals the frontend to show a
+        loading indicator for this tool.
+      - **intermediate** (optional): emitted by tools that call
```
LittleChenLiya: 当前实现仍没有真正的“中间输出”生产者：仓库中 `tool_output_chunk` 只由本 middleware 产生，而这里自身只发空的 start 和完成时的 final；前端又在 final 到达时立即 teardown。因此即使开启配置，用户实际只能看到工具名/转圈，执行过程中的 stdout/文件读取内容不会增量出现，与 PR 标题和开头“real-time like Claude Code”的表述不一致。建议至少接入一个真实异步工具的 partial 生产路径，并做 parent task → subagent → SSE → 前端的时序测试（断言 canonical ToolMessage 之前收到非空 partial）；否则应把功能和文档明确收窄为“工具执行状态提示”。

nankingjing:
@LittleChenLiya — both points addressed on the current head (`191eed59`).

**1. The tail-chunk race — fixed along the lines you described.**

`SubagentResult.append_tool_output_chunk` and `SubagentResult.drain_tool_output_chunks` both take `SubagentResult._state_lock`, and the polling loop in `task_tool.py` drains pending chunks *and* snapshots terminal status in a single locked call — so a chunk appended concurrently with a `COMPLETED`/`FAILED` transition can no longer be sliced away and dropped.

Regression coverage: `test_tool_output_chunks_drain_with_terminal_status` in `tests/test_subagent_executor.py` runs exactly your interleaving — append the final chunk, then `try_set_terminal(COMPLETED)`, then drain — and asserts that the chunk and the terminal status come back together, with a second drain returning empty. `test_task_tool_polling_forwards_tool_output_chunks_to_parent_writer` covers the forwarding path out of the polling loop.

**2. "No real intermediate producer" — you were right; I took your second option and narrowed it.**

You offered two ways out: wire up a real partial producer, or narrow the feature and the docs to tool-execution status. I narrowed it.

`get_stream_writer()` is not called anywhere under `tools/` — the only chunks that exist are the three the middleware builds: `_build_start_chunk` (empty `chunk`, `is_partial=True`), `_build_final_chunk` (first line only, capped at 200 chars) and `_build_error_chunk`. Enabled today, this reports *that* a tool started and *how* it ended; it does not stream stdout or file contents.

What changed for that:
- Title: `feat(tools): stream tool-call lifecycle events to the frontend` (was "real-time streaming tool output")
- Body: new opening paragraph plus a scope note up front — built-in tools emit lifecycle status only, the final chunk is a ≤200-char preview, and the full result still arrives via the canonical `ToolMessage`
- `backend/AGENTS.md` already described it as lifecycle-only; the PR body now matches that wording instead of contradicting it

A genuine partial producer for built-in async tools — plus the parent task → subagent → SSE → frontend timing test you asked for, asserting a non-empty partial arrives before the canonical `ToolMessage` — is tracked as follow-up rather than implied here.

**On CI:** every workflow on this head is `completed / action_required`, so nothing has run since 2026-07-25 (`76208c0f`, where Lint Check, Unit Tests and Frontend Unit Tests were red). On `191eed59` I re-ran the backend tests that failed in that July run — they pass — and `ruff check backend` is clean. I could not re-run the frontend suite locally (this checkout's pnpm install is incomplete), so I am not claiming the frontend job is green. A maintainer approving the workflow run would settle both.

nankingjing:
Correction to the last paragraph of my previous comment, which is now out of date.

I wrote there that I could not re-run the frontend suite because this checkout's `pnpm install` is incomplete. That was accurate at the time: the install was truncated — `@rstest/core`'s `bin/` directory was empty and `react-refresh` was missing its `runtime.js`/`index.js`/`cjs/` contents. Plain `pnpm install` reported "Already up to date" and repaired nothing; `pnpm install --force` fixed it.

With the toolchain working, I ran the frontend suite on the current head, `191eed591f856788745b36bfe0527ce2b304b81d` (`frontend/`, via its own `test` script → `rstest`):

```
status        pass
testFiles     227
tests         1949
passed        1949
failed        0
skipped       0
exit code     0
```

The two frontend test files this PR touches pass as part of that run — 13 tests across `tests/unit/core/threads/stream-options.test.ts` (modified, +5) and `tests/unit/core/tasks/tool-streaming.test.ts` (new, 192 lines).

So: the frontend suite is **green locally on `191eed59`**. That is a local run on Windows, not CI. Every workflow on this head is still `completed / action_required`, so the CI job itself has not executed and I am not claiming it is green — a maintainer approving the queued runs would settle that. This note corrects only my earlier claim; nothing in the PR's code or body depends on it.

nankingjing:
Rebased onto current `main` — the branch is no longer conflicting.

**What was stale.** The head was `191eed59`, last rebased onto `735f67a5` (2026-07-25). `main` has since moved to `2c43767b`, and `mergeable` had gone back to `false` with `mergeable_state: dirty`.

**The conflict.** Exactly one file, `frontend/src/components/workspace/messages/subtask-card.tsx`, and only in the import block. `main` added `resolveRenderedSubtask` from `@/core/tasks/subtask-render` (the `useSubtask(id)` → `resolveRenderedSubtask(useSubtask(id), fallbackTask)` change); this branch adds the `useToolStreaming` and `ToolStreamOutput` imports. Both are needed, so both are kept, ordered by module path to match the file's existing group.

Every other file auto-merged.

**Head is now `002a9246`, 10 commits on top of `2c43767b`.** The net diff is unchanged — 24 files, +1629/−25, byte-for-byte the same patch body as before the rebase; only the hunk `@@` line numbers shift, because `main` grew underneath. Nothing from this branch was dropped or altered in resolution.

**Gates run locally on the rebased head:**

Backend (`df312`, Python 3.12, `PYTHONPATH` pinned to the rebased checkout so the editable `deerflow` install could not shadow it):

```
ruff format --check .   1617 files already formatted
ruff check .            All checks passed!
pytest                  tests/test_tool_streaming_middleware.py
                        tests/test_tool_error_handling_middleware.py
                        tests/test_subagent_executor.py
                        tests/test_task_tool_core_logic.py
                        tests/test_gateway_services.py
                        tests/test_run_worker_rollback.py
                        → 723 passed
scripts/check_config_version.sh → OK, chart (47) current with config.example.yaml (47)
```

Frontend (`frontend/`, node 22 / pnpm 10.26.2):

```
rstest run              → 232 files, 2004 passed, 0 failed, 0 skipped
eslint . --ext .ts,.tsx → clean
tsc --noEmit            → clean
prettier --check .      → All matched files use Prettier code style
```

`main` added `frontend/tests/unit/components/workspace/messages/subtask-card.test.ts` and `subtask-card.dom.test.tsx` while this branch was open — they cover the exact file that conflicted. Run directly on the rebased head: **6 tests, 0 failed.** That is a decent check that the resolution kept both features working rather than just splicing imports.

**On CI:** the 8 workflow runs created for `002a9246` are all `completed / action_required` again, so nothing has executed and I am not claiming any job is green. These are local runs. A maintainer approving the queued runs would settle it — and would also settle `Lint Check`, `Unit Tests` and `Frontend Unit Tests`, which were the three that failed in the last run that actually executed (`76208c0f`, 2026-07-25).

backend/packages/harness/deerflow/agents/middlewares/tool_streaming_middleware.py
```diff
@@ -0,0 +1,233 @@
...
+        writer = _get_stream_writer()
+        if writer is None:
+            # Custom stream mode not active — silently pass through.
+            # This is not an error: the middleware is enabled but the run
+            # may not have been configured with stream_mode="custom".
+            return await handler(request)
+
+        # Emit start-of-execution chunk so the frontend knows a tool is running.
+        try:
+            writer(_build_start_chunk(tool_call_id, tool_name))
```
willem-bd: [P2] Emit lifecycle payloads through `aemit_custom_event`, not the stream writer alone. The harness custom-event invariant requires both paths: the writer feeds Gateway/Web consumers, while `adispatch_custom_event` feeds LangChain callback consumers such as `astream_events(version="v2")`. Calling `writer(...)` directly here (and for error/final below) makes these lifecycle events invisible to callback/event-stream users. The new graph integration test only exercises `astream(stream_mode="custom")`, so it cannot catch that missing half.

frontend/src/core/threads/hooks.ts
```diff
@@ -2108,6 +2114,16 @@ export function useThreadStream({
         return;
       }
 
+      if (eventType === "tool_output_chunk") {
```
willem-bd: [P2] Clear active tool-stream entries when a run ends without a final chunk. A start event inserts provider state, but only `is_final` removes that individual entry. `GraphBubbleUp`, cancellation, transport errors, and replay gaps can terminate after start without final; the replay-gap branch resets messages/tasks but not tool streams, and neither `onError` nor `onFinish` clears them. Because the provider persists for the thread, affected cards can retain a stale spinner until remount. Add an explicit clear-all/reset path for stream termination and replay-gap recovery.

willem-bd (commented):
Static review of the current PR head.

PR #4151, opened 2026-07-13
feat(firecrawl): add crawl, map, interact, and structured extract tools

nankingjing:
## Summary

Expand the Firecrawl community integration from 2 tools (search + fetch) to 6 tools, giving agents richer web capabilities.

### New tools

| Tool | Method | Purpose |
|------|--------|---------|
| `web_map` | `client.map_url()` | Discover all URLs on a website |
| `web_crawl` | `client.crawl_url()` | Deep-crawl with configurable depth/limit/subdomains |
| `web_interact` | `client.interact()` | Browser-based page interaction (click, type, navigate) |
| `structured_extract` | `client.extract()` | LLM-powered structured data extraction with optional JSON Schema |

### web_fetch fix

- Removed hardcoded 4096-char limit → configurable `max_chars` (default: 16384)
- Added configurable `formats` param (default: `[markdown]`)

### Tests

14 tests covering all tools: config passthrough, success paths, error handling, and edge cases (empty sites, schema vs no-schema extraction).

### Configuration

All new tools accept the same config pattern as existing Firecrawl tools:

```yaml
tools:
  - name: web_crawl
    group: web
    use: deerflow.community.firecrawl.tools:web_crawl_tool
    max_depth: 2
    limit: 10
    allow_subdomains: true
    # api_key: $FIRECRAWL_API_KEY

  - name: web_map
    group: web
    use: deerflow.community.firecrawl.tools:web_map_tool

  - name: structured_extract
    group: web
    use: deerflow.community.firecrawl.tools:structured_extract_tool

  - name: web_interact
    group: web
    use: deerflow.community.firecrawl.tools:web_interact_tool
````

### DIFF
--- a/backend/packages/harness/deerflow/community/firecrawl/tools.py
+++ b/backend/packages/harness/deerflow/community/firecrawl/tools.py
@@ -1,17 +1,18 @@
 import json
 
-from firecrawl import FirecrawlApp
+from firecrawl import Firecrawl
+from firecrawl.v2.types import ScrapeOptions
 from langchain.tools import tool
 
 from deerflow.config import get_app_config
 
 
-def _get_firecrawl_client(tool_name: str = "web_search") -> FirecrawlApp:
+def _get_firecrawl_client(tool_name: str = "web_search") -> Firecrawl:
     config = get_app_config().get_tool_config(tool_name)
     api_key = None
     if config is not None and "api_key" in config.model_extra:
         api_key = config.model_extra.get("api_key")
-    return FirecrawlApp(api_key=api_key)  # type: ignore[arg-type]
+    return Firecrawl(api_key=api_key)  # type: ignore[arg-type]
 
 
 @tool("web_search", parse_docstring=True)
@@ -58,8 +59,15 @@ def web_fetch_tool(url: str) -> str:
         url: The URL to fetch the contents of.
     """
     try:
+        config = get_app_config().get_tool_config("web_fetch")
+        max_chars = 16384
+        formats = ["markdown"]
+        if config is not None:
+            max_chars = config.model_extra.get("max_chars", max_chars)
+            formats = config.model_extra.get("formats", formats)
+
         client = _get_firecrawl_client("web_fetch")
-        result = client.scrape(url, formats=["markdown"])
+        result = client.scrape(url, formats=formats)
 
         markdown_content = result.markdown or ""
         metadata = result.metadata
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: backend/packages/harness/deerflow/community/firecrawl/tools.py [hunk 1]
```diff
@@ -70,4 +78,175 @@ def web_fetch_tool(url: str) -> str:
...
+        limit = 100
+        ignore_sitemap = False
+        include_subdomains = True
+        if config is not None:
+            limit = config.model_extra.get("limit", limit)
+            ignore_sitemap = config.model_extra.get("ignore_sitemap", ignore_sitemap)
+            include_subdomains = config.model_extra.get("include_subdomains", include_subdomains)
+
+        client = _get_firecrawl_client("web_map")
+        result = client.map_url(
```
fancyboi999: `FirecrawlApp` in the locked 4.23.0 SDK has no `map_url` method, so every `web_map` invocation is caught and returned as an error. The same contract mismatch occurs at `crawl_url`, while `interact` receives unsupported `actions`/`wait_for` kwargs. Please implement these calls against the installed `map`, `crawl`, and `interact` signatures rather than extending the mocks around a nonexistent API.

fancyboi999 (changes requested):
The new Firecrawl tools are wired to API methods that do not exist in the repository's locked `firecrawl-py==4.23.0`. A real `FirecrawlApp` instance exposes `map(...)` and `crawl(...)`, not `map_url(...)` / `crawl_url(...)`; its `interact(job_id, code=None, *, prompt=..., language=..., timeout=..., origin=...)` also does not accept `actions` or `wait_for`. As written, `web_map`, `web_crawl`, and `web_interact` deterministically return `Error: ...` before making a valid request. The tests miss this because they mock the client with the same nonexistent contract. Please target the installed SDK surface and add at least one contract test against a real `FirecrawlApp` instance/signature.

Also note that this PR currently includes all 12 files from #4150 before its two Firecrawl files, so the fix should be rebased onto `main` or explicitly kept as a dependent stack.

WillemJiang:
@nankingjing Please fix the lint errors.

WillemJiang (approved):

backend/packages/harness/deerflow/community/firecrawl/tools.py
```diff
@@ -70,4 +78,220 @@ def web_fetch_tool(url: str) -> str:
...
+            include_subdomains=include_subdomains,
+            sitemap="skip" if ignore_sitemap else None,
+        )
+
+        urls = getattr(result, "links", None) or []
+        normalized = {
+            "urls": urls,
+            "total_count": len(urls),
+        }
+        return json.dumps(normalized, indent=2, ensure_ascii=False)
```
willem-bd, line 115: 🔴 **Blocking - `web_map` errors on every non-empty site.**

`client.map(...)` returns a `MapData` whose `.links` is `List[LinkResult]` (Pydantic models), **not** `List[str]`. Passing that list straight to `json.dumps({"urls": urls, ...})` raises `TypeError: Object of type LinkResult is not JSON serializable`, which the broad `except` turns into a returned `"Error: ..."` string.

Verified against the installed SDK (firecrawl-py 4.23.0):
```
>>> json.dumps({'urls': [LinkResult(url='https://example.com')]})
TypeError: Object of type LinkResult is not JSON serializable
```

The unit test passes only because it mocks `mock_result.links` as plain strings (`["https://example.com", ...]`), which doesn't match the real return shape.

Fix - extract the URL from each result:
```python
urls = [getattr(l, "url", None) for l in (result.links or [])]
```
(or `model_dump()` each if you want title/description too.)

backend/packages/harness/deerflow/community/firecrawl/tools.py
```diff
@@ -70,4 +78,220 @@ def web_fetch_tool(url: str) -> str:
...
+
+        pages = []
+        for page in result.data or []:
+            page_url = ""
+            page_md = ""
+            if isinstance(page, dict):
+                page_url = page.get("url", "") or ""
+                page_md = page.get("markdown", "") or ""
+            else:
+                page_url = getattr(page, "url", None) or ""
```
willem-bd, line 156: 🟠 **`page.url` is always empty here.**

`CrawlJob.data` is `List[Document]`, and `Document` has no top-level `url` attribute (verified: `hasattr(doc, "url") is False`). The URL lives in `doc.metadata.url`, so `getattr(page, "url", None)` returns `None` -> `""` for every page. Markdown survives, but you can't tell pages apart.

The test misses this because it sets `mock_doc1.url = "..."` on a `MagicMock`, which accepts any attribute.

Fix:
```python
md = getattr(page, "metadata", None)
page_url = (getattr(md, "url", None) if md else None) or ""
```

backend/packages/harness/deerflow/community/firecrawl/tools.py
```diff
@@ -70,4 +78,220 @@ def web_fetch_tool(url: str) -> str:
...
+            elif act_type == "write":
+                parsed_actions.append({"type": "write", "selector": act["selector"], "text": act["text"]})
+            elif act_type == "press":
+                parsed_actions.append({"type": "press", "key": act["key"]})
+            elif act_type == "scroll":
+                parsed_actions.append({"type": "scroll", "direction": act.get("direction", "down"), "amount": act.get("amount", 300)})
+            elif act_type == "screenshot":
+                parsed_actions.append({"type": "screenshot", "full_page": act.get("full_page", False)})
+            elif act_type == "execute_javascript":
+                parsed_actions.append({"type": "executeJavascript", "code": act["code"]})
```
willem-bd, line 226: 🔴 **Blocking - the `execute_javascript` action fails at runtime.**

`ExecuteJavascriptAction` requires a `script: str` field - there is no `code` field. Pydantic v2 drops the unknown `code` (extra ignored) and `script` is required, so constructing `ScrapeOptions(actions=[{"type": "executeJavascript", "code": ...}])` raises `ValidationError` -> the whole `scrape()` call fails -> returns `"Error: ..."`. Since `actions` is a list, one bad action fails the entire interaction.

Verified against the real SDK (firecrawl-py 4.23.0) by constructing `ScrapeOptions` with each of the PR's action dicts: only `executeJavascript` fails (13 validation errors); the rest coerce.

Fix: use `"script"` instead of `"code"`, and update the docstring at line 183 which advertises `{"type": "execute_javascript", "code": "..."}`.

backend/packages/harness/deerflow/community/firecrawl/tools.py
```diff
@@ -70,4 +78,220 @@ def web_fetch_tool(url: str) -> str:
...
+        # Build structured action objects
+        parsed_actions = []
+        for act in action_dicts:
+            act_type = act.get("type", "")
+            if act_type == "wait":
+                parsed_actions.append({"type": "wait", "milliseconds": act.get("milliseconds", 1000)})
+            elif act_type == "click":
+                parsed_actions.append({"type": "click", "selector": act["selector"]})
+            elif act_type == "write":
+                parsed_actions.append({"type": "write", "selector": act["selector"], "text": act["text"]})
```
willem-bd, line 218: 🟠 **`selector` (and `amount`/`selector` for scroll/scrape) is silently dropped by the SDK's action models.**

Verified via real `ScrapeOptions` coercion of the PR's action dicts:
- `WriteAction` has only `type` + `text` - no `selector`, so `{"type": "write", "selector": "#search", "text": "hello"}` types into the default focused element, **not** the field the agent named.
- `ScrollAction` has `type`/`direction`/`selector` - no `amount` (dropped).
- `ScrapeAction` has only `type` - no `selector` (dropped).

These don't raise, so the interaction "succeeds" while ignoring the agent's intent. The docstring advertises all three fields.

Fix: align the docstring / implemented fields with the SDK's action models (e.g. document the click-then-write pattern for typing into a specific field).

backend/packages/harness/deerflow/community/firecrawl/tools.py
```diff
@@ -1,17 +1,18 @@
 import json
 
-from firecrawl import FirecrawlApp
+from firecrawl import Firecrawl
```
willem-bd, line 3: 🔴 **Blocking - dependency floor not bumped.**

`Firecrawl` and `firecrawl.v2.types.ScrapeOptions` are v2-only APIs (added in firecrawl-py 2.0.0), but `backend/packages/harness/pyproject.toml` still pins `firecrawl-py>=1.15.0` and this PR doesn't touch it. A resolve to any 1.x breaks this import at tool-load time.

The `fastcrw/` sibling still uses the v1 `FirecrawlApp`, so both APIs coexist - the floor must satisfy v2.

Fix: bump to `firecrawl-py>=2.0.0` (or higher to match the v2 shapes used). The `TestRealFirecrawlAppContract` suite will guard against drift once the floor is correct.

backend/packages/harness/deerflow/community/firecrawl/tools.py
```diff
@@ -70,4 +78,220 @@ def web_fetch_tool(url: str) -> str:
...
+        allow_subdomains = True
+        scrape_formats = ["markdown"]
+        if config is not None:
+            max_discovery_depth = config.model_extra.get("max_depth", max_discovery_depth)
+            limit = config.model_extra.get("limit", limit)
+            allow_subdomains = config.model_extra.get("allow_subdomains", allow_subdomains)
+            scrape_formats = config.model_extra.get("scrape_formats", scrape_formats)
+
+        client = _get_firecrawl_client("web_crawl")
+        result = client.crawl(
```
willem-bd, line 140: 🟠 **Security - the new remote-content tools bypass `ToolResultSanitizationMiddleware`.**

`tool_result_sanitization_middleware.py` neutralizes forged framework tags (e.g. `<system-reminder>`) in remote-content tool results via a name-based allowlist currently `{web_fetch, web_search, image_search, web_capture}`. `web_crawl`, `web_map`, `web_interact`, and `structured_extract` all return attacker-controlled remote content (crawled pages, mapped URLs, extracted data, interacted-page markdown) but aren't in that set, so forged tags in a crawled/extracted page reach the model unneutralized - re-opening the injection vector the middleware was built to close (`web_capture` was added for exactly this reason).

Fix: add the four tool names to `_REMOTE_CONTENT_TOOL_NAMES`.

backend/packages/harness/deerflow/community/firecrawl/tools.py
```diff
@@ -70,4 +78,220 @@ def web_fetch_tool(url: str) -> str:
...
+
+        trimmed_urls = urls[:max_urls]
+        kwargs = {
+            "urls": trimmed_urls,
+            "prompt": prompt,
+        }
+        if schema is not None:
+            kwargs["schema"] = schema
+
+        result = client.extract(**kwargs)
```
willem-bd, line 285: Heads-up - the v2 SDK's `extract()` is deprecated. Its docstring says: *"The extract endpoint is in maintenance mode and its use is discouraged."* Adding a brand-new tool on a deprecated endpoint means it may break when Firecrawl removes it. Worth a note in the docstring, or reconsider whether this tool should ship yet.

willem-bd (commented):
Reviewed against the PR diff and cross-checked behavior against the installed **firecrawl-py 4.23.0** SDK (this checkout is on a different branch, so I validated the SDK return shapes/types directly rather than running the PR's tests).

**Verdict: request changes.** Clean migration to the v2 client and a nice `TestRealFirecrawlAppContract` idea, but several new tools are broken against the real SDK and the unit tests mask it by mocking return values with shapes that don't match what Firecrawl actually returns (e.g. `links` as `List[str]` instead of `List[LinkResult]`, `mock_doc.url` on a `MagicMock` instead of a real `Document`).

**Highest-leverage follow-up:** extend the contract tests to assert real SDK return-type shapes, not just method signatures - that would have caught findings #1, #2, #4, and #5.

**Additional minor notes (no clean diff anchor):**
- No `timeout` on `crawl()`/`extract()` - both block-and-poll with `timeout=None` (no overall cap); a stuck crawl/extract hangs the agent. `web_interact` correctly exposes `timeout`; crawl/extract should too. Note the unit mismatch: `web_interact`'s `timeout` is ms (30000) while crawl/extract `timeout` is seconds.
- `web_map`/`web_crawl` default `include_subdomains=True`/`allow_subdomains=True`, expanding scope/cost/data surface by default - consider `False`.
- `web_fetch`/`web_crawl` with `formats` excluding `markdown` yields an empty body (markdown is the only field read) - no guard.
- Broad `except Exception: return f"Error: {str(e)}"` turns the above into opaque strings (pre-existing pattern, consistent).

PR #4163, opened 2026-07-14
fix(runs): stamp elapsed duration once per run across message APIs

fancyboi999:
Related to #4152. Complements #4348, which owns the frontend run-duration presentation and wording.

## Why

`turn_duration` is a run-level wall-clock duration (`updated_at - created_at`), but the backend message APIs stamped the same value onto every AI message produced by a run. Consumers could therefore render duplicate duration metadata inside one answer.

The frontend display and the "Completed in" wording are already handled by #4348. This PR is intentionally backend-only: it makes the API payload match the run-scoped contract.

## What changed

- Stamp each run's `turn_duration` only on that run's final visible AI message.
- Apply the same helper to all three read paths:
  - `GET /api/threads/{thread_id}/messages`
  - `GET /api/threads/{thread_id}/runs/{run_id}/messages`
  - `POST /api/threads/{thread_id}/history`
- Preserve the existing middleware-caller exclusion on event-store messages.
- On `/history`, only run the legacy event-store/run-manager correlation when checkpoint metadata is missing a referenced run duration.
- Keep the existing field name and placement (`additional_kwargs.turn_duration`) for compatibility.

No frontend files are changed by this PR. The frontend implementation from #4348 remains unchanged.

## Surface area

- [x] **Backend API** - endpoint / SSE event / request-response shape under `backend/app`
- [ ] **Agents / LangGraph** - agent node, graph wiring, `langgraph.json`, or prompt change
- [ ] **Dependencies** - new/upgraded dependency
- [x] **Default behavior change** - repeated per-message duration metadata becomes once per run

## Screenshots / Recording

Not applicable: the final diff is backend-only. Frontend display coverage belongs to #4348.

## Bug fix verification

- Test paths:
  - `backend/tests/test_thread_run_messages_pagination.py::test_list_run_messages_injects_turn_duration_only_on_last_ai`
  - `backend/tests/test_thread_run_messages_pagination.py::test_list_run_messages_turn_duration_skips_middleware_tail`
  - `backend/tests/test_thread_run_messages_pagination.py::test_list_thread_messages_injects_turn_duration_once_per_run`
  - `backend/tests/test_thread_run_messages_pagination.py::test_list_thread_messages_turn_duration_skips_middleware_tail`
  - `backend/tests/test_threads_router.py::test_get_thread_history_injects_turn_duration_once_per_run`
  - `backend/tests/test_threads_router.py::test_get_thread_history_fast_path_skips_runs_already_in_checkpoint_metadata`
- Red on `main`, green on this branch: yes. Current `main` stamps the duration on every AI message; the updated assertions require exactly one stamped AI message per run.

## Validation

- `uv run pytest tests/test_thread_run_messages_pagination.py tests/test_threads_router.py -q` - 90 passed, 1 warning.
- `make lint` - Ruff check passed; 1,021 files already formatted.
- Real Gateway startup - `GET /health` returned HTTP 200, and all three affected routes were present in OpenAPI.
- GitHub Actions on `db223562` - 10 successful checks, 2 workflow-defined skips, 0 failures.
- Full local backend suite - 10,190 passed, 61 skipped, 4 failed. Three failures were caused by local `config.yaml` overrides and passed with isolated config; the remaining failure is the CI-skipped live LLM test where the current model returned text instead of calling `ls`. No changed-file test failed.

## AI assistance

**Tool(s) used:** Claude Code and Codex

**How you used it:** implemented and reviewed the backend fix, resolved the main-branch conflict, updated regression tests, and ran local/runtime/CI verification. I reviewed the final four-file diff and the PR scope.


willem-bd (commented):
Thanks @fancyboi999. I found one issue that should be addressed before this is ready.

[P2] `get_thread_history` still stamps `turn_duration` on every AI message, so the duplicate-badge bug persists on page reload

- Location: `backend/app/gateway/routers/threads.py:1138-1141` (in `get_thread_history`, not changed by this PR)
- Problem: The shared `stamp_turn_duration_on_last_ai` helper is applied to `list_thread_messages` and `list_run_messages`, but a third injection point in `get_thread_history` stamps `turn_duration` on every AI message of the run with no last-AI filter and no middleware skip. This is the exact symptom #4152 reports.
- Evidence: `threads.py:1138` runs `if msg.get("type") == "ai" and run_id in run_durations:` and stamps unconditionally - no reversed/last-AI check, no middleware skip, unlike the helper. The frontend `useStream` hook is configured with `fetchStateHistory: { limit: 1 }` (`frontend/src/core/threads/hooks.ts:896`), so on mount/reload the SDK calls `POST /threads/{thread_id}/history` → `get_thread_history`; the response's `values.messages` carry `turn_duration` on every AI message and flow into `persistedMessages` → `mergeMessages`, where `dedupeMessagesByIdentity` preserves `turn_duration` per-identity from any source without stripping it. So on a multi-AI-message run, every non-last AI message retains its duplicate stamp on reload - just labeled "Worked for" instead of "Thought for". The PR's reload validation appears to hit `GET /api/threads/{id}/messages` (which is fixed) rather than this history path.
- Suggested fix: replace the inline loop at `threads.py:1138-1141` with the shared helper: `stamp_turn_duration_on_last_ai(serialized_msgs, run_durations)`. `serialized_msgs` already carries `run_id` (set at line 1137). This also brings the middleware-skip behavior to this endpoint.
- Test: add a test for `POST /api/threads/{thread_id}/history` asserting only the last non-middleware AI message per run carries `turn_duration` in a multi-AI-message run, mirroring `test_list_thread_messages_injects_turn_duration_once_per_run`.

fancyboi999:
Confirmed and fixed in 5cff7d2c — `get_thread_history` was a third injection point stamping `turn_duration` on every AI message of the run, exactly as described.

One adjustment to the suggested one-liner: `stamp_turn_duration_on_last_ai` only understood event-store rows, where the message payload is wrapped in a `content` dict (`{"run_id": ..., "content": {"type": "ai", ...}}`). The `/history` path carries flat serialized checkpoint messages (`{"run_id": ..., "type": "ai", ...}`), so a direct swap would have silently stamped nothing. The helper now resolves the payload from either shape and is reused in `get_thread_history`; the inline loop keeps only the `run_id` attachment.

- Test: `test_get_thread_history_injects_turn_duration_once_per_run` drives `POST /api/threads/{thread_id}/history` with a multi-AI-message run and asserts only the last AI message per run carries `turn_duration` (fails on the previous code with the first AI message stamped). The middleware-skip half is not testable on this path — checkpoint messages don't carry the event-store `metadata.caller`, and middleware messages (e.g. title generation) don't enter graph state, so the skip is inert here by construction.
- Real-data validation: ran this branch's gateway against a copy of a local dev DB; for a run with 4 AI messages, `POST /threads/{id}/history` now returns `turn_duration` only on the final AI message (previously all four were stamped).

frontend/src/components/ai-elements/reasoning.tsx
```diff
@@ -170,7 +170,10 @@ const defaultGetThinkingMessage = (
   if (duration === undefined) {
     return <span>Thought for a few seconds</span>;
```
willem-bd, line 171: **Copy inconsistency (minor):** This `duration === undefined` branch still reads *"Thought for a few seconds"*, while the numbered branch just below now reads *"Worked for {duration} seconds"*. Since `turn_duration` is wall-clock elapsed work either way, the two completed-state strings now disagree on framing. Low impact (this path only fires when a completed run has no `turn_duration`), but for consistency consider *"Worked for a few seconds"* here as well.

frontend/src/components/ai-elements/reasoning.tsx
```diff
@@ -170,7 +170,10 @@ const defaultGetThinkingMessage = (
...
   }
-  return <span>Thought for {duration} seconds</span>;
+  // `duration` is the run's wall-clock elapsed time (backend `turn_duration`),
+  // which includes tool execution — "Thought for" would mislabel it as model
+  // thinking latency (#4152).
+  return <span>Worked for {duration} seconds</span>;
```
willem-bd: **Convention note:** `frontend/AGENTS.md` marks `ai-elements/` as registry-generated ("don't manually edit these"), and this line edits that file directly. The PR-body justification holds up - git history shows `reasoning.tsx` has carried repo-specific behavior (`onTurnDurationChange`, `startTimeProp`, duration logic, its own unit test) since #3663, so it's effectively a forked file already. Flagging so a maintainer explicitly blesses direct edits here (the convention has been bent across several merged PRs), or tracks a follow-up to relocate the file out of `ai-elements/`. Not a blocker.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -52,6 +52,34 @@ def compute_run_durations(runs) -> dict[str, int]:
...
+    """
+    stamped: set[str] = set()
+    for msg in reversed(messages):
+        rid = msg.get("run_id")
+        if not rid or rid in stamped or rid not in run_durations:
+            continue
+        content = msg.get("content")
+        payload = content if isinstance(content, dict) else msg
+        metadata = msg.get("metadata") or {}
+        is_middleware = str(metadata.get("caller", "")).startswith("middleware:")
```
willem-bd, line 110: **Middleware skip is shape-asymmetric (minor):** `metadata = msg.get("metadata")` reads row-level metadata on event-store rows (so `caller: "middleware:..."` title messages are skipped on the live feed), but serialized checkpoint messages on the `/history` path don't carry `metadata.caller`, so the skip is effectively inert there - a middleware message persisted into the message channel could grab the badge on reload. In practice title generation writes thread metadata, not the message channel, so this is likely moot, but it's an implicit assumption worth a one-line note in the docstring.

backend/tests/test_thread_run_messages_pagination.py
```diff
@@ -320,16 +321,59 @@ def test_list_thread_messages_injects_turn_duration():
...
+        response = client.get("/api/threads/thread-1/runs/run-1/messages")
+
+    assert response.status_code == 200
+    data = response.json()["data"]
+
+    assert data[0]["content"]["additional_kwargs"]["turn_duration"] == 5
+    assert "turn_duration" not in data[1]["content"].get("additional_kwargs", {})
+
+
+def test_list_thread_messages_injects_turn_duration_once_per_run():
```
willem-bd, line 345: **Test gap (minor):** middleware-skip is asserted on `list_run_messages` (`test_list_run_messages_turn_duration_skips_middleware_tail`) but not on the `list_thread_messages` feed. `list_thread_messages` newly gained the middleware-skip via this shared helper (it previously stamped *all* AI messages, middleware included) - that's the intended fix, but a sibling test asserting the skip on this path would lock the contract, since the behavior change is specific to the feed.

willem-bd (commented):
Reviewing #4163 (fix #4152). The core fix is correct and well-tested: the shared `stamp_turn_duration_on_last_ai` helper cleanly dedupes three stamping sites, the reverse-iteration + `stamped` set correctly picks each run's last non-middleware AI message, and the dual-shape handling (`content` dict vs flat row) is right. The "Worked for X seconds" relabel is a truthful framing of wall-clock elapsed (tools included). Tests are flipped to pin the new once-per-run contract across `list_run_messages`, `list_thread_messages`, and `/history`.

The inline nits below are all minor. One adjacent note (not inline - it's pre-existing unchanged code outside the diff hunks): the `/history` `event_store.list_messages(thread_id, limit=1000)` (flagged by an existing `FIXME` in the file) silently drops durations for messages older than the cap on long threads. Out of scope for this PR, but worth a follow-up.

frontend/src/components/ai-elements/reasoning.tsx
```diff
@@ -170,7 +170,10 @@ const defaultGetThinkingMessage = (
   if (duration === undefined) {
     return <span>Thought for a few seconds</span>;
   }
-  return <span>Thought for {duration} seconds</span>;
+  // `duration` is the run's wall-clock elapsed time (backend `turn_duration`),
```
willem-bd: The new comment says `duration` is "the run's wall-clock elapsed time (backend `turn_duration`)", but that's only one of two sources. In `message-list-item.tsx` the value reaching here is `turnDuration = rawTurnDuration ?? cachedDuration`, where `cachedDuration` is populated from the `Reasoning` component's own timer via `onTurnDurationChange={handleDurationChange}` (wired at both Reasoning render sites). So `duration` can also be a **client-side streaming timer** - a different metric than the backend run wall-clock.

This matters for the #4152 goal. The backend now stamps `turn_duration` once per run (final answer only), but the client-side cache is an independent source this PR doesn't touch. An intermediate AI message (e.g. a tool-calling turn that renders `Reasoning` and streams) can acquire a `cachedDuration` and render a "Worked for X seconds" badge on its own, re-introducing multiple badges per run. The Reasoning block renders whenever `!!reasoningContent || wasLoading || turnDuration !== undefined`, so a cached duration alone is enough to surface one.

The live check verified one badge per run - but if those runs were single-AI-message (no tools), the multi-tool case is unverified. Could you exercise a run with several AI messages before the final answer (tool calls in between) and confirm only one badge appears? If extras show up, the once-per-run dedup needs to extend to `clientTurnDurations`, or the cache should be scoped to the final message.

frontend/src/components/ai-elements/reasoning.tsx
willem-bd: Heads-up on file ownership: `frontend/AGENTS.md` explicitly lists `ai-elements/` as registry-generated ("don't manually edit these", L100), and this PR edits `reasoning.tsx` directly. I checked for an automated sync and found none in `package.json` or `scripts/` (only `sync_labels.py`, unrelated), so the clobber risk today is limited to a future manual `shadcn add` / Vercel AI SDK registry pull - low, but such a pull would silently revert just this wording line, leaving the frontend copy inconsistent with the backend's "elapsed work" intent.

The file already carries other repo-specific behavior (`startTimeProp`, `onTurnDurationChange`, `hasContent`), so the convention was already bent. To honor it cleanly, the repo-specific copy could live in a repo-owned wrapper that composes the stock `Reasoning`; otherwise just confirm a maintainer is OK with the override. Not a blocker given there's no automated sync today.

willem-bd (commented):
Reviewing the #4152 fix. The backend change is correct, clean, and well-tested - it genuinely resolves both contract violations (run-scoped data shown per-message; elapsed time labeled as thinking latency) and tightens a latent over-stamping bug in `list_run_messages` (the old `reversed()` loop never broke, so it stamped every non-middleware AI message). The three stamping sites are consolidated into one helper that handles both message shapes and the `/history` `current_turn_run_id` fallback. Tests are meaningful (the `list_run_messages` test uses a real 2-AI-message run and asserts the earlier AI is *not* stamped).

Two follow-ups in the inline comments:
1. (Medium) The frontend `duration` prop has a second, client-side source (`cachedDuration` via `onTurnDurationChange` in `message-list-item.tsx`) that the backend fix doesn't touch - it could badge intermediate AI messages in multi-tool runs. Needs a multi-tool live verification.
2. (Low) The wording change is in a registry-generated "don't edit" file; no auto-sync found, so clobber risk is manual-only, but worth acknowledging.

Verdict: approve with the multi-tool live check requested in comment 1.

WillemJiang:
@fancyboi999 , please resolve the conflicts of main branch.

frontend/src/components/ai-elements/reasoning.tsx
fancyboi999: Fixed in 67981a2b — both branches now read "Worked for a few seconds" / "Worked for {duration} seconds". Updated the two existing unit tests plus added a third asserting both branches share the same framing.

frontend/src/components/ai-elements/reasoning.tsx
fancyboi999: Agreed this needs a maintainer call, not something I should decide unilaterally. For transparency: this round added two more small edits to this same file (the copy-text fix in 67981a2b, and removing the now-dead `onTurnDurationChange` prop wiring at the two call sites in `message-list-item.tsx` — the prop itself stays on `Reasoning` since nothing else references it and it's a generically useful capability, so that specific edit was in the caller, not here) on top of what the original PR already touched. Both are minimal continuations of the existing repo-specific behavior already flagged (`startTimeProp`, `onTurnDurationChange`, duration logic since #3663), not a new pattern.

I didn't extract a repo-owned wrapper around the stock `Reasoning` — that's a real option but a bigger structural change than this bug fix calls for, and per your note it's not a blocker. Happy to file a follow-up issue for it, or do it in this PR if you'd rather land it now than defer.

backend/app/gateway/routers/thread_runs.py
fancyboi999: Fixed in 7e5a4367 — added a docstring note explaining the skip is only effective on the event-store shape, and why that's not a gap in practice (middleware writes go to thread metadata, not the `messages` channel, so no middleware message ever reaches a checkpoint's `messages` list for `/history` to replay).

backend/tests/test_thread_run_messages_pagination.py
fancyboi999: Fixed in 7e5a4367 — added `test_list_thread_messages_turn_duration_skips_middleware_tail`, mirroring the existing `list_run_messages` test on the `list_thread_messages` feed.

frontend/src/components/ai-elements/reasoning.tsx
fancyboi999: Traced this end to end. The literal scenario ("a tool-calling turn that renders `Reasoning`") can't actually happen: `getMessageGroups` (`frontend/src/core/messages/utils.ts`) routes any AI message with `hasToolCalls(message) === true` into an `"assistant:processing"` group, rendered by `MessageGroup`/`ChainOfThought` (`message-group.tsx`) — which never imports `Reasoning`/`turnDuration` at all. Only a message with content and *no* tool calls (`becomesAssistantBubble`) reaches `MessageContent_`'s `<Reasoning>`. `message-group.test.ts` already locks this routing in (`"renders assistant text attached to a tool-calling processing message"`).

But you're right that the underlying mechanism is broken, just through a different door: a run can produce **multiple standalone content-only AI messages** with no tool calls between them (e.g. a subagent handoff emitting its own content message before the lead agent's final answer) — each becomes its *own* `"assistant"` bubble group. `turnStartTime`/`isLoading` are only passed non-null to the *last* group (`message-list.tsx`), so when a second such bubble appears, the first one's `isLoading` flips `true → false` on that same render — not because the run finished, but because it got superseded. Its `Reasoning` instance still fires the demotion as a legitimate stream-end, computes `Math.floor((Date.now() - startTime) / 1000)` (the time until the *second* message appeared, not the run's total elapsed time), and `onTurnDurationChange` caches that wrong, premature number into `clientTurnDurations` keyed by its own message id — surfacing a "Worked for X seconds" badge with a number that's both duplicated *and* incorrect.

Fixed in b13f0c16 by removing the `onTurnDurationChange` wiring at both `Reasoning` call sites in `message-list-item.tsx`. `cachedDuration` now only ever mirrors a backend-confirmed `turn_duration` (the other, untouched effect here) — by construction it can no longer hold a client-measured value, so a superseded message can't display a stale/wrong duration regardless of how many standalone content messages a run produces.

Verification, and where it falls short of what you asked:
- Manually drove a real single-AI-message run against a fresh `make dev` instance (real LLM, real streaming): live "Thinking... (Ns)" ticking during the stream, then exactly one "Worked for 9 seconds" badge once the run completed — confirms the base case still works and the copy is consistent.
- I could **not** reproduce the literal multi-AI-message-before-final-answer case live — this local dev instance has no sandbox/tool backend configured, so the model has no bash/tool access and never emits a multi-message run to click through. I did not stand up a sandbox just for this check.
- No interactive component-test harness exists in this repo for `Reasoning` (`reasoning-trigger.test.ts` only uses `renderToStaticMarkup`, which never runs `useEffect`, so it can't exercise the streaming→duration transition this bug lives in) or the demotion path in `message-list.tsx`, and I didn't think adding one was in scope for this fix.

So the fix is verified by construction (the only path that ever wrote a non-backend value into the cache is now gone) plus the live single-message check above, not by a live repro of the multi-message case. If you'd rather see that live before this merges, let me know and I'll set up a local sandbox to force it.

frontend/src/components/ai-elements/reasoning.tsx
fancyboi999: Agreed this needs a maintainer call, not something I should decide unilaterally. For transparency: this round added two more small edits to this same file (the copy-text fix in 67981a2b, and removing the now-dead `onTurnDurationChange` prop wiring at the two call sites in `message-list-item.tsx` — the prop itself stays on `Reasoning` since nothing else references it and it's a generically useful capability, so that specific edit was in the caller, not here) on top of what the original PR already touched. Both are minimal continuations of the existing repo-specific behavior already flagged (`startTimeProp`, `onTurnDurationChange`, duration logic since #3663), not a new pattern.

I didn't extract a repo-owned wrapper around the stock `Reasoning` — that's a real option but a bigger structural change than this bug fix calls for, and per your note it's not a blocker. Happy to file a follow-up issue for it, or do it in this PR if you'd rather land it now than defer.

WillemJiang:
@fancyboi999  please resolve the conflicts with the main branch.

fancyboi999:
@WillemJiang conflicts are resolved against the current main branch.

- Kept main's frontend run-level duration UI from #4348 and removed the superseded frontend changes from this PR.
- The remaining diff is backend-only: stamp `turn_duration` once per run across thread messages, per-run messages, and history replay.
- Verified locally: 90 targeted tests passed, backend lint passed, and a real Gateway startup returned HTTP 200 from `/health` with all three affected routes present in OpenAPI.

Full local backend suite result: 10,190 passed, 61 skipped, 4 failed. Three failures came from local `config.yaml` overrides and passed under isolated config; the remaining failure is the CI-skipped live LLM test where the current model returned text instead of calling `ls`. No changed-file test failed.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -78,6 +78,41 @@ def compute_run_durations(runs) -> dict[str, int]:
     return durations
 
 
+def stamp_turn_duration_on_last_ai(messages, run_durations: dict[str, int]) -> None:
+    """Attach each run's elapsed seconds to that run's final visible AI message only.
+
+    ``turn_duration`` is the run's wall-clock lifetime (``compute_run_durations``),
+    not model thinking time — it belongs to the run, not to individual messages.
+    Stamping every AI message made the UI repeat the same number once per
+    message and let tool-wait time read as thinking latency (#4152).
```
willem-bd, line 87: **Scope/body mismatch (minor):** With the `reasoning.tsx` changes reverted in 53e68dfe + db223562c, this PR now ships only the backend half of #4152. The once-per-run stamping here fixes the 'repeatedly inside one answer' symptom, but the UI still renders the stamped value as "Thought for X seconds" — so the issue's 'X far larger than any plausible thinking time' complaint (X includes tool execution and queueing, not just model thinking) is unchanged on its own.

The PR body still advertises the "Worked for X seconds" label change and lists `frontend/tests/unit/core/reasoning-trigger.test.ts` as a test path, neither of which is in the diff anymore; the 'Frontend UI' surface checkbox is also still ticked. Worth either re-adding the frontend piece (pending the ai-elements ownership call from the prior round) or trimming the body so this doesn't land with an inaccurate description of what it actually changes.

willem-bd (commented):
Re-reviewing the post-force-push head (db223562). All 6 prior findings are resolved: the 4 frontend items are moot (reasoning.tsx changes reverted in 53e68dfe / db223562c) and the 2 backend items (middleware-skip docstring, missing `list_thread_messages` middleware-skip test) were addressed in 7e5a4367. The `stamp_turn_duration_on_last_ai` helper and the `/history` `missing_run_ids` gate are byte-for-byte identical to the prior reviewed head (7e5a4367) — only line numbers shifted from the main merge. The core fix (once-per-run stamping across all three message APIs) is correct and well-tested. One new scope observation inline.

backend/app/gateway/routers/thread_runs.py
fancyboi999: Agreed. I updated the PR body to match the post-#4348 scope: it now describes the backend-only once-per-run stamping across all three message APIs, marks only Backend API and Default behavior change, removes the reverted frontend test, screenshot, and wording claims, and points to #4348 as the owner of frontend presentation. No code change was needed for this review item.

WillemJiang (approved):

PR #4164, opened 2026-07-14
fix(#4095): prevent eval fixture skills from stripping task tool from lead agent

nankingjing:
## Problem
Issue #4095: The lead agent's `task` subagent tool was silently stripped on every run because eval fixture skills from `skills/public/skill-reviewer/evals/fixtures/` were loaded as real enabled skills. Their restrictive `allowed-tools` declarations (e.g., `[bash]` or `[]`) got unioned into the tool-policy filter, clamping the entire lead-agent toolset.

## Scope after rebase onto current main
This PR originally shipped three fixes. #4098 (65afc9b1, merged 2026-07-16) restructured tool policy — `allowed-tools` now applies only to *actively activated* skills via `SkillToolPolicyMiddleware`, and `task` is deliberately not framework-exempt so a restricted skill cannot delegate around its policy. That supersedes former Fix 2 (tool-policy loader returning `[]` for `available_skills=None`) and contradicts former Fix 3 (`task` in `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`), so both were **dropped in the rebase**. What remains:

**Fix: exclude orphaned `evals/` and `fixtures/` support trees from skill discovery** (`skill_storage.py`, `local_skill_storage.py`, `user_scoped_skill_storage.py`): all four `os.walk` sites prune a directory named `evals` or `fixtures` **only when it is not itself a skill package** (no `SKILL.md` directly inside — shared predicate `is_skill_support_data_dir()`), alongside the existing dotfile filter. This is defense-in-depth on top of main's package-boundary rule: it keeps *orphaned* fixture trees (no parent `SKILL.md` boundary) out of the registry, so they can never appear in the catalog, be activated, or contribute `allowed-tools`. Per review feedback, `evals` and `fixtures` are **not** reserved skill names: a package whose own directory carries `SKILL.md` is still discovered in every walk (public, user-custom, legacy) across both storage backends.

## Tests
- `test_load_skills_skips_evals_and_fixtures_directories` — fixture `SKILL.md` under `evals/fixtures/` is not discovered (LocalSkillStorage)
- `test_skill_packages_named_evals_or_fixtures_are_discovered` (LocalSkillStorage + UserScopedSkillStorage) and `test_legacy_skill_package_named_evals_is_discovered` — `evals`/`fixtures` are not reserved names: packages carrying their own `SKILL.md` are discovered in the public/custom/legacy walks, while nested support data below a package boundary stays excluded (review feedback)
- `test_eval_fixture_skills_cannot_clamp_tool_policy` — end-to-end through the **real** loader and **real** `tool_policy` functions (no mocks): a restrictive orphaned fixture is not discovered, `allowed_tool_names_for_skills` stays allow-all, and `task` survives `filter_tools_by_skill_allowed_tools`
- `test_public_orphan_eval_fixture_dirs_are_not_registered`, `test_user_custom_eval_fixture_dirs_are_not_registered`, `test_legacy_global_custom_eval_fixture_dirs_are_not_registered` — cover the three `UserScopedSkillStorage` walk sites this PR changes
- `test_passive_eval_fixture_like_skill_cannot_strip_task_from_lead_tools` — a merely-enabled restrictive skill leaves `task` in the model request (middleware, real filter path)
- `test_always_available_builtin_tool_names_membership_and_filter_behavior` — direct pin on `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`: discovery tools survive a restrictive filter; `task` is deliberately not exempt (#4098)
- `test_load_enabled_available_skills_real_path_none_returns_full_catalog` — exercises the **real** `_load_enabled_available_skills`: `available_skills=None` returns the full enabled catalog (skill discovery keeps working for the default agent), a whitelist filters by name

All affected suites pass locally (`test_skills_loader`, `test_user_scoped_skill_storage`, `test_skill_tool_policy_middleware`, `test_lead_agent_skills`, `test_local_skill_storage_write`, `test_skill_policy_deferred_discovery`, `test_slash_skills`, `test_subagent_skills_config`, `test_memory_tools`, `test_skill_storage_lifecycle`, `test_skills_installer` — 301 passed), plus `ruff check` / `ruff format --check` clean. Docs: one-line note in `backend/AGENTS.md` (skills Loading) per the documentation update policy.


willem-bd (commented):

[P2] Missing regression tests for the two security-sensitive tool-policy changes (Fix 2 and Fix 3)

- Location: `backend/packages/harness/deerflow/agents/lead_agent/agent.py:422-423` (Fix 2: `if available_skills is None: return []`) and `backend/packages/harness/deerflow/skills/tool_policy.py:16` (Fix 3: `task` added to `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`)
- Problem: The PR adds exactly one new test (`test_load_skills_skips_evals_and_fixtures_directories`) covering only Fix 1. The two highest-risk changes - Fix 2 (default agent no longer clamped by passive skills' `allowed-tools`) and Fix 3 (`task` unconditionally preserved through allowed-tools filtering) - have no regression test. The CI rollup only shows Triage + CLA; the full lint/unit/E2E workflows did not run, so this logic ships unverified.
- Evidence: Existing lead-agent tests mock `_load_enabled_skills_for_tool_policy` (e.g. `test_lead_agent_skills.py:238,284,324`), so the real `available_skills is None -> return []` gate at `agent.py:422` is never exercised - reverting Fix 2 to `return skills` would fail no test. For Fix 3, no test asserts `task` survives filtering when a whitelisted skill declares restrictive `allowed-tools`; `test_make_lead_agent_filters_tools_from_available_skills` doesn't include `task` in its input tools or assertions, and `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES` has no direct test.
- Suggested fix: add (a) a unit test calling the real `_load_enabled_skills_for_tool_policy(available_skills=None, ...)` and asserting it returns `[]`, plus a complementary case with `available_skills={"some-skill"}` that still loads and filters; and (b) a test calling `filter_tools_by_skill_allowed_tools` with a `task`-named tool plus a skill declaring `allowed-tools: [bash]` and `always_allowed_tool_names=ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`, asserting `task` is retained.
- Test: `assert _load_enabled_skills_for_tool_policy(None, app_config=cfg, user_id=...) == []`; `assert "task" in {t.name for t in filter_tools_by_skill_allowed_tools([NamedTool("task"), NamedTool("bash")], [_make_skill("restrict", ["bash"])], always_allowed_tool_names=ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES)}`.

nankingjing:
@willem-bd Thanks for the review — both points are now addressed, with one important adaptation.

**Conflicts / rebase.** Rebased onto current main. While doing so I found that #4098 (65afc9b1, merged after your review) restructured exactly the area your P2 covers: `allowed-tools` is now applied at runtime by `SkillToolPolicyMiddleware` only to *actively activated* skills, `_load_enabled_skills_for_tool_policy` became `_load_enabled_available_skills` (feeding skill discovery, not tool authority), and the backend docs now state that `task` is deliberately **not** framework-exempt ("a restricted skill cannot delegate around its policy", with `test_slash_activated_skill_filters_first_model_call_and_task` pinning it). So I dropped former Fix 2 and Fix 3 in the rebase: keeping Fix 2's `return []` would now empty the default agent's skill-search catalog, and keeping Fix 3's `task` exemption would contradict #4098's documented security decision and break its tests. The PR now carries only the discovery-exclusion fix (`evals/` + `fixtures/` filtered in all four walk sites), which remains useful defense-in-depth on top of main's package-boundary rule for orphaned fixture trees.

**Regression tests (your (a)/(b), adapted to the post-#4098 shape).**
- (a) real-path loader test: `test_load_enabled_available_skills_real_path_none_returns_full_catalog` calls the real `_load_enabled_available_skills` (not monkeypatched like the `make_lead_agent` tests): `available_skills=None` returns the full enabled catalog, `{"beta"}` filters by name — guarding against reintroducing the `return []` variant.
- (b) task-survival + `ALWAYS_AVAILABLE_BUILTIN_TOOL_NAMES`: `test_eval_fixture_skills_cannot_clamp_tool_policy` runs end-to-end through the real loader and real `filter_tools_by_skill_allowed_tools` with a `task` tool and a fixture declaring `allowed-tools: [bash]`, asserting `task` is retained; `test_passive_eval_fixture_like_skill_cannot_strip_task_from_lead_tools` asserts the same at the middleware layer; and `test_always_available_builtin_tool_names_membership_and_filter_behavior` directly pins the constant's membership and filter behavior (discovery tools survive, `task` intentionally does not, per #4098).
- Added three `UserScopedSkillStorage` tests covering the public / user-custom / legacy walk sites this PR changes (previously only the `LocalSkillStorage` walk was tested). All four new discovery tests fail when the exclusion is reverted (verified locally).

Affected suites pass locally (240 tests across the skills/tool-policy/lead-agent files) and ruff check + format are clean. PR description updated to reflect the reduced scope.

fancyboi999 (changes requested):


Reviewed head SHA: `264ef75e3b1cb4ee7db8ce9ad96b02951f4e2af6`

### [P2] Do not reserve `evals` and `fixtures` as invalid skill names

- Location: `backend/packages/harness/deerflow/skills/storage/local_skill_storage.py:83` and `backend/packages/harness/deerflow/skills/storage/user_scoped_skill_storage.py:268,279,295`
- Problem: filtering these directory names at every `os.walk` level silently hides legitimate public, user-custom, and legacy skills whose package name is exactly `evals` or `fixtures`.
- Evidence: skill names are otherwise allowed by `SkillStorage.validate_skill_name()` whenever they match the hyphen-case pattern, and neither name is reserved. Because the filter runs in the category root before checking the child for `SKILL.md`, a valid `public/evals/SKILL.md` is never visited. Current main already stops descending as soon as it finds a package-level `SKILL.md`, which prevents the bundled `skill-reviewer/evals/fixtures/**` support tree from being registered without globally banning these names.
- Suggested fix: keep the package-boundary rule and either remove this global name filter or scope any orphan-fixture exclusion to a known support-data path. Do not make arbitrary package names unavailable across all storage backends.
- Test: add public and custom packages at `evals/SKILL.md` and `fixtures/SKILL.md` and assert both storage implementations discover them, while nested fixtures below an existing skill package remain excluded.

nankingjing:
@fancyboi999 Thanks — confirmed and fixed in 1c5cb855. I reproduced the exact failure mode you described: at the previous head, a legitimate `public/evals/SKILL.md` or `custom/fixtures/SKILL.md` package (both names pass `validate_skill_name()`) was silently invisible to both storage backends, and a user could even create the skill via `write_custom_skill("fixtures", ...)` and then not see it in listings.

**Fix (per your suggested direction).** The package-boundary rule is untouched. The flat name filter is replaced by a shared predicate `is_skill_support_data_dir(parent, name)` in `skill_storage.py`, used at all four walk sites: a directory named `evals`/`fixtures` is pruned **only when it is not itself a skill package** (no `SKILL.md` directly inside). So:
- `public/evals/SKILL.md`, `custom/fixtures/SKILL.md`, legacy `custom/evals/SKILL.md` — discovered again (no globally reserved names, any storage backend);
- orphaned support trees like `public/evals/fixtures/orphan/SKILL.md` (the issue #4095 shape, where `evals/` has no `SKILL.md` of its own) — still never registered;
- nested fixtures below an existing package boundary — still excluded by the existing `dir_names.clear()` rule.

**Tests (your suggested test, both storage implementations).** `test_skill_packages_named_evals_or_fixtures_are_discovered` (LocalSkillStorage: public `evals` + custom `fixtures` discovered, nested support data below the `evals` package still excluded; UserScopedSkillStorage: public `evals` + user-custom `fixtures`) and `test_legacy_skill_package_named_evals_is_discovered` (legacy global-custom walk). Revert-verified in both directions: with the old name-only filter the three new discovery tests fail; with no filter at all the five orphan-exclusion tests fail.

**Local gates at 1c5cb855** (conda py3.12, Windows, `core.autocrlf=false`): `ruff format --check` clean (870 files), `ruff check` clean, and 301 tests passed across `test_skills_loader`, `test_user_scoped_skill_storage`, `test_skill_tool_policy_middleware`, `test_lead_agent_skills`, `test_local_skill_storage_write`, `test_skill_policy_deferred_discovery`, `test_slash_skills`, `test_subagent_skills_config`, `test_memory_tools`, `test_skill_storage_lifecycle`, `test_skills_installer`. Also added the one-line discovery note to `backend/AGENTS.md` (skills Loading) per the documentation update policy. PR description updated to match.

nankingjing:
Follow-up: main moved again while the review fix above was in flight (#4264 restructured the same `backend/AGENTS.md` skills section and left the PR unmergeable), so the branch is rebased onto current main `c9b6131f`. The review-fix commit `1c5cb855` referenced above is now `f148c36d` (same content; only the `backend/AGENTS.md` hunk was re-merged around #4264's new "External reload" bullet).

Gates re-run at `f148c36d` on top of `c9b6131f`: `ruff format --check` and `ruff check` clean; 313 tests passed across the 11 suites listed above plus `test_skills_reload.py` (new on main from #4264). The reserved-name reproduction confirms `public/evals/SKILL.md` and `custom/fixtures/SKILL.md` packages are discovered while orphaned `evals/fixtures` support trees stay excluded.

fancyboi999 (changes requested):
Thanks @nankingjing. The direct `evals` / `fixtures` package-name cases from the previous review are fixed, but the current predicate still reserves those names as namespace segments.


Reviewed head SHA: `f148c36df5804894e3ed2ff8ac2df4a45d2bbbe0`

[P2] Do not prune legitimate skill namespaces by directory name

- Location: `backend/packages/harness/deerflow/skills/storage/skill_storage.py:19-30` and the four discovery walks that call `is_skill_support_data_dir()`
- Problem: a supported namespace directory has no `SKILL.md` of its own, so the new predicate still prunes every namespace segment named `evals` or `fixtures`. A valid skill at `public/fixtures/team-helper/SKILL.md`, or the equivalent user-custom/legacy path, is therefore unreachable even though recursive namespace discovery is an explicit loader contract.
- Evidence: each `os.walk` filters `dir_names` before descending. For `public/fixtures/team-helper/SKILL.md`, `is_skill_support_data_dir(public, "fixtures")` returns true solely because `public/fixtures/SKILL.md` is absent, so `team-helper` is never visited. The adjacent loader comment says namespace directories without `SKILL.md` must recurse and gives `public/team/helper` as the supported shape; the current tests cover direct packages named `evals` / `fixtures`, not those names as namespace segments.
- Suggested fix: avoid a global name-based support-data heuristic. The bundled `skill-reviewer/evals/fixtures/**` tree is already excluded by the existing package-boundary rule once `skill-reviewer/SKILL.md` is found. If orphan support trees need an additional defense, scope it to an explicit known support-data root or metadata that distinguishes support data from a normal namespace; absence of a direct `SKILL.md` cannot make that distinction.
- Test: add public, user-custom, and legacy skills below `evals/<skill>/SKILL.md` and `fixtures/<namespace>/<skill>/SKILL.md` and assert they are discovered, while the fixture tree below an actual skill package remains excluded by the package boundary.

nankingjing:
@fancyboi999 Good catch — you're right, the predicate was pruning namespace directories named evals/fixtures even when they contain actual skill packages (e.g., `public/fixtures/team-helper/SKILL.md`).

**Fixed in 49e5ebaa:** `is_skill_support_data_dir` now accepts `category_root` and uses three checks:
1. No own `SKILL.md`
2. No immediate-child `SKILL.md` (single-level namespace detection — this is the fix)
3. Skill package in ancestor chain (defense-in-depth)

All four walk-site callers pass their category root. New tests cover namespace dirs in both `LocalSkillStorage` (`public/fixtures/team-helper`) and `UserScopedSkillStorage` (public, user-custom, legacy walks — all three walk sites). Orphan-exclusion tests updated to two-level patterns (`evals/fixtures/stale`) that remain correctly excluded.

**Gates (conda df312, ruff 0.15.21):** ruff format + check clean, **120 tests passed** across 5 suites (test_skills_loader, test_user_scoped_skill_storage, test_skill_tool_policy_middleware, test_lead_agent_skills, test_skills_reload). Revert-verified: removing the namespace-child check fails the 3 new namespace tests.

Ready for re-review.

fancyboi999 (changes requested):
Thanks @nankingjing. The direct-package and single-level namespace cases are fixed, but the current heuristic still reserves these names for deeper supported namespaces.


Reviewed head SHA: `49e5ebaaa45dedd2fbae8ca4058bb9cf21bd6d29`

[P2] Preserve arbitrary-depth skill namespaces named `evals` or `fixtures`

- Location: `backend/packages/harness/deerflow/skills/storage/skill_storage.py:42-51`
- Problem: the new exception only checks whether an immediate child contains `SKILL.md`. A valid recursive namespace such as `public/fixtures/team/deep-helper/SKILL.md` has no direct child package under `fixtures`, so the loader still prunes the whole namespace and silently hides the skill. The loader contract explicitly supports recursive namespace directories, not only one level.
- Evidence: running the PR-head `LocalSkillStorage` against `public/fixtures/team/deep-helper/SKILL.md` plus the control `public/team/deep-helper-control/SKILL.md` returned only `deep-helper-control`. The new tests cover `fixtures/team-helper/SKILL.md`, but the previous review's requested `fixtures/<namespace>/<skill>/SKILL.md` shape is still absent.
- Suggested fix: do not infer support data from a fixed namespace depth. The existing package-boundary rule already excludes fixtures below a real skill package; if orphan support trees need extra handling, scope it to an explicit known support-data root or use a rule that cannot conflate an arbitrary recursive namespace with fixture data.
- Test: add public, user-custom, and legacy cases with at least two namespace levels below `evals` / `fixtures`, and assert both storage implementations discover them while nested fixtures below an actual skill package remain excluded.

nankingjing:
@fancyboi999 Addressed in 418b991:

**[P2] Arbitrary-depth namespace** — replaced the immediate-child SKILL.md check with a child-name check: if any immediate child directory name is NOT `evals` or `fixtures`, the directory is treated as a legitimate namespace and recursed into. This handles `public/fixtures/team/deep-helper/SKILL.md` (two levels below `fixtures`, where `team` is not a support-data name) while still pruning orphan support-data trees like `public/evals/fixtures/restrictive/` (where every immediate child of both `evals` and `fixtures` is itself named `evals` or `fixtures`).

Tests added for all three discovery walks (LocalSkillStorage public, UserScopedSkillStorage public + user-custom + legacy fallback). Backend: 53 passed (test_skills_loader) + 43 passed (test_skill_tool_policy_middleware). Ready for re-review.

nankingjing:
All three review rounds from July have answers in the threads above, and the branch has since been rebased onto current `main` and extended with the remaining `evals`/`fixtures` handling:

- `b4c52b7c` support recursive namespace dirs named `evals/fixtures`
- `1a3915cc` stop pruning namespace directories named `evals/`
- `f0f2aed0` do not reserve `evals/fixtures` as skill names in discovery walks
- `8e7adc16` pin the #4095 regression coverage across the storage walks and tool policy
- `1ce150cc` prune eval fixtures from the integrations walk and port the doc note

Head is `1ce150cc`: 8 files, +416/-8, `mergeable: true` against `main` (the `blocked` state is the outstanding review decision, not a conflict). 238 of those lines are tests across `test_skills_loader.py`, `test_user_scoped_skill_storage.py`, `test_skill_tool_policy_middleware.py` and `test_lead_agent_skills.py`.

@fancyboi999 @willem-bd could this get another pass when you have a moment? Happy to split the doc/AGENTS.md note out if it is easier to land separately.

nankingjing:
Adding the one thing missing from the note above: CI on this head has never run — the workflows are gated behind maintainer approval.

All six runs on `1ce150cc` are `completed / action_required`:

- Lint Check
- Unit Tests
- Backend Blocking IO
- Frontend Unit Tests
- Replay E2E (front-back contract)

That is fork-PR workflow gating, not a result — none of those jobs has executed even once. `Triage` is the only workflow that shows a real conclusion (`success`), because it isn't gated. So the checks you'd want to see before merging will not appear on their own; someone with write access needs to hit "Approve and run workflows" on one of them.

Local verification on `1ce150cc` in the meantime:

- `pytest tests/test_skills_loader.py tests/test_user_scoped_skill_storage.py tests/test_skill_tool_policy_middleware.py tests/test_lead_agent_skills.py` — **138 passed**
- `ruff check .` — clean; `ruff format --check .` — 1593 files already formatted

Still happy to split the `AGENTS.md` note out if that makes it easier to land.

nankingjing:
Status update — this one has been quiet since 2026-07-18, so here is where it stands.

**All four review rounds are addressed.** @fancyboi999 asked for changes four times between 2026-07-16 and 2026-07-18, each time narrowing the same point: the `evals` / `fixtures` reserved-name predicate, first for direct package names, then for single-level namespace segments, then at arbitrary depth. Each round was fixed; the last landed in `418b991`. @willem-bd's [P2] on the same predicate (missing regression tests for Fix 2 / Fix 3) was answered earlier. There has been no review activity on the PR since.

If any of those points is still open from your side, tell me which one and I'll take it.

**Branch state.** `mergeable: true`, no conflicts against `main`.

**One thing worth knowing about the green check:** CI has never run on the current head. There are six check suites on `1ce150cc` — one is `completed/success` (`pr-labels`, `reviewing`, `issue-triage`), and **five are `completed/action_required` with zero runs**. `action_required` means the workflow is queued waiting for a maintainer to approve it; it is not a failure, and it is not a pass. The green tick on the PR page comes from `pr-labels`, which only labels.

So the "success" here is not a test result. Approving the queued runs would settle Lint, Unit Tests and Frontend in one pass.

PR #4167, opened 2026-07-14
fix(skills): handle non-string frontmatter keys

DaoyuanLi2816:
## Summary

YAML frontmatter can contain non-string mapping keys. Mixed string and numeric keys previously reached `sorted()` in both review and install validation and raised `TypeError` instead of returning a normal validation finding.

`split_skill_markdown`, the shared parser for both paths, now converts mapping keys to strings once. Review and install validation therefore report the malformed fields consistently without crashing.

## Tests

The focused review and install-validation suites pass (40 tests); ruff check and format are clean.

### DIFF
--- a/backend/tests/test_skill_review_core.py
+++ b/backend/tests/test_skill_review_core.py
@@ -56,6 +56,22 @@ def test_review_core_reports_missing_description_blocker(tmp_path):
     assert any(f["rule_id"] == "structure.missing-description" for f in facts["findings"])
 
 
+def test_review_core_reports_non_string_frontmatter_key_as_unknown_field(tmp_path):
+    _write(
+        tmp_path / "SKILL.md",
+        "---\nname: demo-skill\ndescription: Demo skill. Invoke when testing review.\n42: stray-value\nunexpected-field: another-value\n---\n\n# Demo\n\nFollow the steps and stop.\n",
+    )
+
+    facts = analyze_skill_package(LocalDirectoryReader(tmp_path).read())
+
+    assert facts["summary"]["blockers"] == 0
+    finding = next(f for f in facts["findings"] if f["rule_id"] == "structure.unknown-frontmatter-field")
+    assert finding["severity"] == "warning"
+    assert finding["evidence"] == ["42", "unexpected-field"]
+    assert "42" in finding["message"]
+    assert "unexpected-field" in finding["message"]
+
+
 def test_resource_graph_reports_unreferenced_resource(tmp_path):
     _write(tmp_path / "SKILL.md", _valid_skill())
     _write(tmp_path / "references" / "unused.md", "# Unused\n")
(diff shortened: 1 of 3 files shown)


backend/packages/harness/deerflow/skills/review/analyzer.py
```diff
@@ -160,7 +160,13 @@ def _analyze_skill_md(content: str, *, profile: ProfileName, findings: list[dict
...
 
     metadata = parts.metadata
-    unexpected = sorted(set(metadata) - ALLOWED_FRONTMATTER_PROPERTIES)
+    # YAML permits non-string top-level keys (e.g. a bare `42:` or `true:`), so a
+    # malformed-but-technically-valid frontmatter block can hand back a dict with
+    # mixed str/non-str keys. Coerce to strings before the set/sort below so such a
+    # key is still reported as an unknown frontmatter field instead of crashing
+    # `sorted()` on an unorderable mix of types (mirrors the same `str(k) for k in
+    # metadata` guard used for untrusted mapping keys in persistence/thread_meta/sql.py).
+    unexpected = sorted({str(key) for key in metadata} - ALLOWED_FRONTMATTER_PROPERTIES)
```
willem-bd: This fix is correct for the review path, but the identical crash still exists one layer up in the install-time validator, which this PR doesn't touch:

```python
# deerflow/skills/validation.py:36-38
unexpected_keys = set(frontmatter.keys()) - ALLOWED_FRONTMATTER_PROPERTIES
if unexpected_keys:
    return False, f"Unexpected key(s) in SKILL.md frontmatter: {', '.join(sorted(unexpected_keys))}", None
```

`split_skill_markdown` (`frontmatter.py:57`) only rejects *non-dict* metadata, so a dict with a non-string key still reaches this `sorted()` and raises the same `TypeError: '<' not supported between instances of 'str' and 'int'` on a mixed int/str set (verified locally with `sorted({42, 'unexpected-field'})`). That validator is the skill-**install** path (`skill_storage.py:75`, `local_skill_storage.py:181`), so installing a skill with the same malformed frontmatter tracebacks instead of returning a clean `(False, "Unexpected key(s)...", None)` error.

Since `frontmatter.py` is the shared schema source for the runtime parser, install validator, and review core, the cleaner root-cause fix is to coerce keys once in `split_skill_markdown` — e.g. `metadata = {str(k): v for k, v in metadata.items()}` — which covers all three consumers and makes the value match the already-declared `metadata: dict[str, Any]` annotation (which non-string keys currently violate). At minimum, mirror the same `str(k)` coercion into `validation.py:36` so the install path is consistent with this one.

backend/tests/test_skill_review_core.py
```diff
@@ -56,6 +56,22 @@ def test_review_core_reports_missing_description_blocker(tmp_path):
     assert any(f["rule_id"] == "structure.missing-description" for f in facts["findings"])
 
 
+def test_review_core_reports_non_string_frontmatter_key_as_unknown_field(tmp_path):
```
willem-bd, line 59: Good coverage of the review path (direct + CLI). One gap: there's no test exercising non-string keys through `_validate_skill_frontmatter` — exactly the sibling path that still crashes (see the comment on `analyzer.py`). A regression test in `tests/test_skills_validation.py` (which already covers that validator) feeding a `42:` + unknown-field frontmatter would both prove that crash and guard the follow-up fix.

WillemJiang:
@DaoyuanLi2816, please take a look at the latest review comments.

backend/packages/harness/deerflow/skills/review/analyzer.py
DaoyuanLi2816: Addressed by coercing keys in `split_skill_markdown`, the shared parser used by both review and install validation.

backend/tests/test_skill_review_core.py
DaoyuanLi2816: Added install-path regression coverage in `test_skills_validation.py`; the focused validation and review suites now pass.

WillemJiang (approved):

WillemJiang (approved):

PR #4168, opened 2026-07-14
fix(sandbox): scope e2b grep() glob filter to its directory prefix

DaoyuanLi2816:
## Summary

`E2BSandbox.grep()`'s `glob` filter silently drops directory scoping. Given a
caller-supplied pattern like `"src/*.js"`, it reduces the pattern to just its
basename before building the `grep` invocation:

```python
if glob is not None:
    include_pattern = glob.split("/")[-1] or glob
    flags.append(f"--include={include_pattern}")
```

`"src/*.js"` becomes `--include=*.js`. GNU grep's `--include=PATTERN` matches
by **basename only, at any depth** -- it has no concept of a directory
prefix. So a caller scoping their search to `src/` gets matches from every
directory in the sandbox tree that happens to contain a same-extension file,
not just `src/`.

Verified empirically against a real GNU grep binary (3.0), not a mock. Given
this tree:

```
workspace/
  src/app.js          <- intended match
  other_dir/unrelated.js   <- must NOT match when scoped to src/*.js
```

```
$ grep -r -n -H -I -i --include=*.js -E -- needle workspace
workspace/other_dir/unrelated.js:1:console.log('needle in other_dir...');
workspace/src/app.js:1:console.log('needle in src');
```

Both files come back -- the `other_dir` match leaks through even though the
caller asked to scope the search to `src/`. Confirmed `--include="src/*.js"`
(i.e. just not stripping the prefix) doesn't fix it either -- GNU grep
matches zero files, since the flag's matcher is basename-only and never sees
the directory component.

The sibling `glob()` method in the same file already handles this correctly:
it resolves candidate paths against the search root and filters through
`path_matches()` (`deerflow/sandbox/search.py`), which is directory-aware
(`PurePosixPath.match()` respects a pattern's leading directory segments).
`grep()` never used that helper for its `glob` argument -- it only ever
built a `--include=` flag.

## Fix

Keep the existing `--include=<basename>` flag as a coarse pre-filter (it's a
superset of the true match set -- any file `path_matches()` would accept
also satisfies the basename pattern -- so it only reduces what grep has to
search, never drops a valid hit). Then post-filter grep's raw output through
`path_matches()`, the same helper `glob()` already uses, to enforce the
caller's actual directory scope:

```python
if glob is not None:
    if file_path != root and not file_path.startswith(root_prefix):
        continue
    rel_path = file_path[len(root):].lstrip("/")
    if not rel_path or not path_matches(glob, rel_path):
        continue
```

This mirrors `glob()`'s own root/rel_path/`path_matches()` logic instead of
inventing new matching code. A plain non-scoped pattern (no `/`, e.g.
`"*.py"`) is unaffected -- `path_matches()` matches those at any depth, same
as before.

## Testing

Added to `backend/tests/test_e2b_sandbox_provider.py` (the tests here mock
the e2b SDK's `commands.run`, so they supply the raw stdout a real broadened
`--include=*.js` grep would return -- verified against the matrix above --
and assert `E2BSandbox.grep()` narrows it to the caller's real scope):

- `test_grep_scoped_glob_excludes_unrelated_directory_matches` -- regression:
  `glob="src/*.js"` must exclude a same-extension match from a sibling
  directory.
- `test_grep_plain_glob_matches_files_in_any_directory` -- no regression:
  a plain `glob="*.py"` (no directory component) still matches at any depth.
- `test_grep_scoped_glob_still_passes_coarse_include_flag` -- the coarse
  `--include=*.js` pre-filter is still emitted (perf: narrows what grep has
  to search) even though it can't express directory scoping by itself.
- `test_grep_without_glob_is_unaffected` -- omitting `glob` entirely returns
  every match with no path-based post-filtering.

Fail-before/pass-after: reverted just the source fix via a patch file (kept
the new tests in place), confirmed `test_grep_scoped_glob_excludes_unrelated_directory_matches`
fails against the pre-fix code (the other three pass in both states, since
they cover behavior the bug never touched), then re-applied the fix and
confirmed all four pass.

`cd backend && PYTHONPATH=packages/harness python -m pytest tests/test_e2b_sandbox_provider.py -v`
-- 50 passed (46 pre-existing + 4 new).

`ruff check` / `ruff format --check` (line-length 240) -- clean on both
changed files.

### DIFF
--- a/backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py
+++ b/backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py
@@ -428,6 +428,13 @@ class E2BSandbox(Sandbox):
         else:
             flags.append("-E")
         if glob is not None:
+            # ``grep --include`` only matches by basename, at any depth -- it
+            # cannot express a directory-scoping prefix like ``src/`` in
+            # ``src/*.js``. Pass just the basename portion as a coarse
+            # pre-filter (a superset of the true match set: every file
+            # ``path_matches`` can accept also satisfies this basename
+            # pattern) and enforce the real directory scope below via
+            # ``path_matches``, the same helper ``glob()`` uses.
             include_pattern = glob.split("/")[-1] or glob
             flags.append(f"--include={include_pattern}")
 
@@ -448,6 +455,9 @@ class E2BSandbox(Sandbox):
                 logger.error("Failed to grep in e2b sandbox: %s", e)
                 return [], False
 
+        root = resolved.rstrip("/") or "/"
+        root_prefix = root if root == "/" else f"{root}/"
+
         matches: list[GrepMatch] = []
         truncated = False
         for raw in output.splitlines():
@@ -461,6 +471,14 @@ class E2BSandbox(Sandbox):
                 continue
             if should_ignore_path(file_path):
                 continue
+            if glob is not None:
+                # Restrict to the caller's real directory scope -- the
+                # ``--include`` flag above only pre-filtered by basename.
+                if file_path != root and not file_path.startswith(root_prefix):
+                    continue
+                rel_path = file_path[len(root) :].lstrip("/")
+                if not rel_path or not path_matches(glob, rel_path):
+                    continue
             matches.append(
                 GrepMatch(
                     path=file_path,
--- a/backend/tests/test_e2b_sandbox_provider.py
+++ b/backend/tests/test_e2b_sandbox_provider.py
@@ -838,3 +838,78 @@ def test_sync_outputs_to_host_skips_oversize_files(monkeypatch, tmp_path):
     assert files.read_calls == [], "oversize files must be skipped without invoking download_file"
     host_target = Paths(base_dir=tmp_path).thread_dir("t1", user_id="u1") / "user-data" / "outputs" / "huge.bin"
     assert not host_target.exists(), "no oversize artefact must be written to host"
+
+
+# ──────────────────────────────────────────────────────────────────────────────
+# grep() directory-scoped glob filtering
+#
+# Real GNU grep's ``--include=PATTERN`` matches by basename only, at any
+# depth -- it cannot express the directory-scoping portion of a pattern like
+# ``src/*.js``. These tests can't invoke a real grep binary (the command
+# runs remotely inside the e2b VM via the mocked ``client.commands.run``), so
+# they instead supply the raw stdout a real broadened ``--include=*.js``
+# grep would actually return (matches from every directory, not just the
+# intended one) and assert ``E2BSandbox.grep`` narrows it down to the
+# caller's real directory scope via ``path_matches`` -- the same helper
+# ``glob()`` already uses -- exactly as verified empirically against a real
+# GNU grep binary during development of this fix.
+# ──────────────────────────────────────────────────────────────────────────────
+
+
+def test_grep_scoped_glob_excludes_unrelated_directory_matches():
+    """Regression: grep(glob="src/*.js") must not leak matches from sibling
+    directories that merely share the file extension."""
+    raw_stdout = "/home/user/workspace/other_dir/unrelated.js:1:console.log('needle in other_dir');\n/home/user/workspace/src/app.js:1:console.log('needle in src');\n"
+    client = FakeClient(commands=FakeCommandsAPI([SimpleNamespace(stdout=raw_stdout, stderr="", exit_code=0)]))
+    sb = _make_sandbox(client)
+
+    matches, truncated = sb.grep("/mnt/user-data/workspace", "needle", glob="src/*.js")
+
+    paths = [m.path for m in matches]
+    assert paths == ["/home/user/workspace/src/app.js"]
+    assert "/home/user/workspace/other_dir/unrelated.js" not in paths
+    assert truncated is False
+
+
+def test_grep_plain_glob_matches_files_in_any_directory():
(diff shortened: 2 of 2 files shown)


backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py
```diff
@@ -448,6 +455,9 @@ def grep(
                 logger.error("Failed to grep in e2b sandbox: %s", e)
                 return [], False
 
+        root = resolved.rstrip("/") or "/"
```
willem-bd, line 458: This root/root_prefix/rel_path scope-check is now duplicated verbatim with `glob()` (same file, lines 387-397). If the two drift, `grep` and `glob` will silently disagree on what "under the search root" means.

Consider a small helper, e.g. `_rel_path_under_root(path, root) -> str | None` (returns the rel_path when `path` is under `root`, else `None`) that both methods call. As a bonus, note the per-line ordering already differs: `glob()` checks the root-prefix *before* `should_ignore_path`, while `grep()` checks `should_ignore_path` first (line 472) then this block. A shared helper would align them.

backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py
```diff
@@ -428,6 +428,13 @@ def grep(
...
         if glob is not None:
+            # ``grep --include`` only matches by basename, at any depth -- it
+            # cannot express a directory-scoping prefix like ``src/`` in
+            # ``src/*.js``. Pass just the basename portion as a coarse
+            # pre-filter (a superset of the true match set: every file
+            # ``path_matches`` can accept also satisfies this basename
+            # pattern) and enforce the real directory scope below via
+            # ``path_matches``, the same helper ``glob()`` uses.
             include_pattern = glob.split("/")[-1] or glob
             flags.append(f"--include={include_pattern}")
```
willem-bd, line 439: Minor / pre-existing: `--include={include_pattern}` is emitted unquoted (it's joined into `cmd` via `" ".join(flags)` and run through the shell - the `| head` pipe confirms a shell). `*.js` only survives because bash leaves a no-match glob literal; if `nullglob` were ever set the flag would vanish, and if the VM's CWD contained a file named like `--include=*.js` the shell would expand it. Not introduced by this PR (and the new test asserts the unquoted form), but `shlex.quote(include_pattern)` would harden it.

backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox.py
```diff
@@ -461,6 +471,14 @@ def grep(
                 continue
             if should_ignore_path(file_path):
                 continue
+            if glob is not None:
+                # Restrict to the caller's real directory scope -- the
+                # ``--include`` flag above only pre-filtered by basename.
+                if file_path != root and not file_path.startswith(root_prefix):
+                    continue
+                rel_path = file_path[len(root) :].lstrip("/")
+                if not rel_path or not path_matches(glob, rel_path):
```
willem-bd, line 480: Heads-up on `path_matches` semantics now that `grep` routes through it: `PurePosixPath.match` treats `**` as a single path segment, not recursive. So `glob="src/**/*.js"` matches only `src/<one>/*.js`, not arbitrary depth under `src/` (and not `src/app.js` directly either).

Before this fix, `grep(glob="src/**/*.js")` matched *all* `.js` files anywhere (no scoping); now it's `src/<one>/*.js`. The new behavior is more correct and matches what `glob()` already does, so it's a consistency win - just flagging in case `**` patterns are expected to be common, since the result may surprise callers.

willem-bd (commented):
Reviewed `E2BSandbox.grep()` directory-scoping fix. Real bug, correct fix - the superset argument (`--include=<basename>` can only ever be a superset of what `path_matches` accepts, so the post-filter only ever *drops* out-of-scope hits) holds, and mirroring `glob()`'s proven root/rel_path/`path_matches` logic is the right call. Tests are well-designed - they feed the raw stdout a real broadened `--include=*.js` grep would return and assert the narrowing. Fail-before/pass-after verification is documented.

Inline notes below are all minor / follow-up: duplicated scope-check logic with `glob()` (suggest a shared helper), the unquoted `--include` flag (pre-existing hardening), and a `**`-pattern caveat in `path_matches` (pre-existing, now consistent with `glob()`).

Net: approve. The duplication is the only one I'd act on, and it's not a blocker.

WillemJiang:
@DaoyuanLi2816, please check out the latest review comments.

DaoyuanLi2816:
@WillemJiang thanks for the nudge -- had seen willem-bd's review, replying properly now. All three are marked non-blocking/heads-up rather than required changes, so noting where each stands:

- **Shared `_rel_path_under_root` helper for `grep()`/`glob()`:** agreed this would be worth doing so the two can't silently drift, but it means touching the working, tested scope-check in both methods for a pure dedup -- happy to file that as a small follow-up PR rather than widen this one's diff.
- **Unquoted `--include={include_pattern}`:** you noted this is pre-existing and not introduced here, and the new test asserts the current unquoted form -- leaving as-is, agreed a `shlex.quote` hardening pass is a separate concern from this fix.
- **`PurePosixPath.match` and `**` semantics:** agreed this is a real behavior change (unscoped -> single-segment-scoped), but you noted it's more correct and consistent with `glob()`'s existing behavior, so no action needed there -- just good to have it on record in case `**` patterns turn out to be common enough to want real recursive-glob support later.

WillemJiang (approved):

PR #4170, opened 2026-07-14
fix(workspace-changes): classify a symlink replacing a file distinctly from deleted

DaoyuanLi2816:
## Summary

`workspace_changes` (added in #3945) captures pre/post-run snapshots of the agent's workspace/outputs directories so a run's file changes can be audited. `scan_workspace_roots()` silently omits every symlinked path from a snapshot entirely, so when an agent run replaces a tracked file with a symlink (e.g. `rm config.txt && ln -s /some/other/path config.txt`), the resulting change is reported as plain `"deleted"` — hiding that the path is still alive on disk, now as a symlink that can point anywhere on the host, including outside the workspace root. That is exactly the kind of action a file-change audit exists to surface.

## The gap

- `scanner.py::scan_workspace_roots()` had `if host_file.is_symlink() or not host_file.is_file(): continue` — a symlinked path is skipped entirely, not even recorded as a metadata-only stub the way binary/large/sensitive-looking files already are (see `_snapshot_file`'s `sensitive` branch, a few lines below).
- Because the symlink never enters the "after" snapshot at all, `diff.py::_status()` sees `before_file` set and `after_file=None`, and unconditionally returns `"deleted"` — with no way to distinguish "genuinely deleted" from "replaced by something the scanner chose not to snapshot".
- Net effect: an agent run doing `rm config.txt && ln -s /etc/shadow config.txt` (or any other host path) inside the audited workspace/outputs roots is reported to the user as a plain deletion, actively hiding that a symlink — a potentially security-relevant detail, since it can point anywhere — now occupies that path.

## The fix

- `scanner.py`: add `_snapshot_symlink()`, mirroring the existing binary/large/sensitive metadata-only-stub pattern. A symlink is `lstat()`'d and never followed (its target may be arbitrary, so nothing ever `stat()`s/`read_bytes()`s through it) and recorded with `symlink=True` and `symlink_target` from `os.readlink()`, instead of being skipped.
- `types.py`: add `"symlink_created"` to `WorkspaceChangeStatus` and `"symlink"` to `DiffUnavailableReason`; add `symlink` / `symlink_target` to `FileSnapshot`, `symlink` / `symlink_target_before` / `symlink_target_after` to `WorkspaceFileChange`, and a `symlink_created` counter to `WorkspaceChangeSummary` (folded into `has_changes()` too).
- `diff.py::_status()`: a symlink newly occupying a path that was not already a symlink — brand new or replacing a prior file — now reports `"symlink_created"` instead of falling through to `"created"`/`"deleted"`. A symlink genuinely removed with nothing replacing it is unchanged: still `"deleted"`.
- `recorder.py` / `api.py`: keep the human-readable changed-file count and the empty-summary shape in sync with the new counter.

## Testing

- Added to `backend/tests/test_workspace_changes.py`: `test_compare_snapshots_classifies_symlink_replacing_file_as_symlink_created`, `test_scan_workspace_roots_captures_symlinks_as_metadata_only_stubs`, and `test_compare_snapshots_reports_removed_symlink_without_replacement_as_deleted` (the last one pins the scope boundary: a symlink genuinely removed with nothing replacing it stays `"deleted"`).
- Real symlink creation needs elevated privilege on stock Windows (no Developer Mode/admin on this dev machine), so the new tests probe support via a `symlink_support` fixture and skip gracefully where it's unavailable, rather than failing the run — the same pattern CPython's own test suite uses for this exact platform gap. They create real symlinks and run unconditionally on Linux/CI.
- Verification method: copied the three dependency-free `workspace_changes` modules (`types.py`/`scanner.py`/`diff.py`, unmodified) onto a real POSIX filesystem under WSL and drove the actual `scan_workspace_roots()`/`compare_snapshots()` functions against a real symlink pointing outside the scanned root. This confirmed `main`'s current behavior reports `status="deleted"` while `os.readlink()`/`os.path.realpath()` independently show the path alive on disk as a symlink, and confirmed this fix reports `status="symlink_created"` with `symlink_target` populated to the real target path instead.
- Fail-before/pass-after: reverse-applying just this PR's source-file diff (`git apply -R`, not `git stash`) and re-running the WSL driver reproduces the `"deleted"` bug; re-applying the same patch restores `"symlink_created"`.
- Ran the full `backend/tests/test_workspace_changes.py`: 17 passed, 3 skipped (the new symlink tests, gracefully degrading on this platform's missing symlink privilege) — no regressions among the existing tests.
- `ruff check` / `ruff format --check` (line length 240) clean on all changed files.

## Frontend follow-up (not included here)

This PR is backend-only by design. The frontend's `workspace-changes` module has a closed `WorkspaceChangeStatus` union (`frontend/src/core/workspace-changes/types.ts`) and a `statusRank` / status-icon / status-label switch (`frontend/src/core/workspace-changes/summary.ts`, `frontend/src/components/workspace/changes/workspace-change-panel.tsx`) that only handle `created`/`modified`/`deleted`. A `symlink_created` file currently falls through those to the generic "modified" rank/icon/label. Extending those to render the new status distinctly (and surface `symlink_target`) is a natural follow-up, left out of this change since it is backend-correctness-focused.

### DIFF
--- a/frontend/src/core/workspace-changes/summary.ts
+++ b/frontend/src/core/workspace-changes/summary.ts
@@ -43,9 +43,16 @@ export function getWorkspaceChangeLineClass(
 }
 
 export function sortWorkspaceChanges(files: WorkspaceFileChange[]) {
+  // `symlink_created` shares modified's rank: it is a distinct change kind
+  // (a symlink replacing a file) that otherwise renders like a modification
+  // (see StatusIcon / statusLabel). This `satisfies Record<...>` guard forces
+  // every WorkspaceChangeStatus variant to have a rank, so adding a new status
+  // without updating this table is a compile error instead of a silent NaN
+  // from an unranked lookup (which broke Array#sort's ordering contract).
   const statusRank = {
     created: 0,
     modified: 1,
+    symlink_created: 1,
     deleted: 2,
   } satisfies Record<WorkspaceFileChange["status"], number>;
 
--- a/frontend/src/core/workspace-changes/summary.ts
+++ b/frontend/src/core/workspace-changes/summary.ts
@@ -8,7 +8,12 @@ export type WorkspaceChangeLineClass =
   | "meta";
 
 export function getChangedFileCount(summary: WorkspaceChangeSummary) {
-  return summary.created + summary.modified + summary.deleted;
+  return (
+    summary.created +
+    summary.modified +
+    summary.deleted +
+    summary.symlink_created
+  );
 }
 
 export function getWorkspaceChangeBadgeLabel(summary: WorkspaceChangeSummary) {
--- a/frontend/src/core/workspace-changes/types.ts
+++ b/frontend/src/core/workspace-changes/types.ts
@@ -10,6 +10,7 @@ export interface WorkspaceChangeSummary {
   created: number;
   modified: number;
   deleted: number;
+  symlink_created: number;
   additions: number;
   deletions: number;
   truncated: boolean;
--- a/frontend/src/core/workspace-changes/types.ts
+++ b/frontend/src/core/workspace-changes/types.ts
@@ -1,10 +1,15 @@
-export type WorkspaceChangeStatus = "created" | "modified" | "deleted";
+export type WorkspaceChangeStatus =
+  | "created"
+  | "modified"
+  | "deleted"
+  | "symlink_created";
 
 export type DiffUnavailableReason =
   | "binary"
   | "large"
   | "sensitive"
-  | "truncated";
+  | "truncated"
+  | "symlink";
 
 export interface WorkspaceChangeSummary {
   created: number;
@@ -31,6 +36,9 @@ export interface WorkspaceFileChange {
   diff_unavailable_reason: DiffUnavailableReason | null;
   additions: number;
   deletions: number;
+  symlink: boolean;
+  symlink_target_before: string | null;
+  symlink_target_after: string | null;
 }
 
 export interface WorkspaceChangesResponse {
--- a/backend/packages/harness/deerflow/workspace_changes/api.py
+++ b/backend/packages/harness/deerflow/workspace_changes/api.py
@@ -8,6 +8,7 @@ EMPTY_SUMMARY = {
     "created": 0,
     "modified": 0,
     "deleted": 0,
+    "symlink_created": 0,
     "additions": 0,
     "deletions": 0,
     "truncated": False,
--- a/backend/packages/harness/deerflow/workspace_changes/diff.py
+++ b/backend/packages/harness/deerflow/workspace_changes/diff.py
@@ -23,7 +23,7 @@ def compare_snapshots(
     resolved_limits = limits or WorkspaceChangeLimits()
     all_paths = sorted(set(before.files) | set(after.files))
     changes: list[WorkspaceFileChange] = []
-    created = modified = deleted = additions = deletions = 0
+    created = modified = deleted = symlink_created = additions = deletions = 0
     total_diff_bytes = 0
     truncated = before.truncated or after.truncated
 
@@ -38,6 +38,8 @@ def compare_snapshots(
             created += 1
         elif status == "modified":
             modified += 1
+        elif status == "symlink_created":
+            symlink_created += 1
         else:
             deleted += 1
 
@@ -73,6 +75,9 @@ def compare_snapshots(
                     diff_unavailable_reason=reason,
                     additions=line_additions,
                     deletions=line_deletions,
+                    symlink=bool((after_file or before_file).symlink if (after_file or before_file) else False),
+                    symlink_target_before=before_file.symlink_target if before_file else None,
+                    symlink_target_after=after_file.symlink_target if after_file else None,
                 )
             )
         else:
@@ -83,6 +88,7 @@ def compare_snapshots(
             created=created,
             modified=modified,
             deleted=deleted,
+            symlink_created=symlink_created,
             additions=additions,
             deletions=deletions,
             truncated=truncated,
@@ -107,6 +113,15 @@ def _status(
     before_file: FileSnapshot | None,
     after_file: FileSnapshot | None,
 ) -> WorkspaceChangeStatus:
+    # A symlink now occupying a path that was not already a symlink is always
+    # surfaced distinctly - whether it is brand new (before_file is None) or it
+    # just replaced a regular file (before_file is None => "deleted" would
+    # otherwise be reported even though the path is still alive on disk, just
+    # as a symlink that may point anywhere on the host).
+    before_was_symlink = before_file is not None and before_file.symlink
+    after_is_symlink = after_file is not None and after_file.symlink
+    if after_is_symlink and not before_was_symlink:
+        return "symlink_created"
     if before_file is None:
         return "created"
     if after_file is None:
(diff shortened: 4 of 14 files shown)


willem-bd (commented):

[P2] Workspace-changes badge hidden entirely when `symlink_created` is the only change type

- Location: `frontend/src/core/workspace-changes/summary.ts:11` (`getChangedFileCount`), consumed by `frontend/src/components/workspace/changes/workspace-change-badge.tsx:42-43`
- Problem: `getChangedFileCount` sums only `created + modified + deleted` and does not include `symlink_created`. `WorkspaceChangeBadge` returns `null` when `count === 0`, so a run whose only file change is a symlink replacing a file - the exact security-relevant scenario this PR targets (`rm config.txt && ln -s /elsewhere config.txt`) - will have its workspace-changes badge hidden entirely. The user sees no indication any file changed.
- Evidence: `summary.ts:11` returns `summary.created + summary.modified + summary.deleted` (no `symlink_created`); `workspace-change-badge.tsx:42-43` does `if (count === 0) return null;`. The backend `WorkspaceChangeSummary` now emits `symlink_created` (and `has_changes()` includes it), but the frontend `WorkspaceChangeSummary` interface (`types.ts:10-12`) only declares `created/modified/deleted`. This is a regression introduced by this diff: before it, file->symlink was reported as `"deleted"` (counted by `getChangedFileCount`, badge visible); after it, `"symlink_created"` is not counted, so the badge disappears for symlink-only runs - the opposite of the PR's goal of surfacing the change. The PR note that `symlink_created` "falls through to modified rank/icon/label" covers the per-file `StatusIcon`/`statusLabel` rendering (which does fall through to "modified") but not `getChangedFileCount`, which gates whether the badge renders at all.
- Suggested fix: include `symlink_created` in the frontend count (one line), either in this PR or as a blocking follow-up:
  - `summary.ts`: `return summary.created + summary.modified + summary.deleted + summary.symlink_created;`
  - `types.ts`: add `symlink_created: number;` to the `WorkspaceChangeSummary` interface.
  (The per-file `StatusIcon`/`statusLabel` "modified" fallthrough can stay a non-blocking follow-up since it still renders something visible once the badge shows.)
- Test: add a frontend unit test asserting `getChangedFileCount` returns `1` for a summary with only `symlink_created: 1`; or a backend/E2E test asserting the badge renders when the only change is a symlink replacing a file.

DaoyuanLi2816:
Thanks @willem-bd, good catch -- fixed in 51d605fe.

Went with your suggested fix exactly: added `symlink_created: number` to the frontend `WorkspaceChangeSummary` interface (`types.ts`) and included it in `getChangedFileCount` (`summary.ts`), so a symlink-only run no longer collapses to a count of 0 and the badge renders as it should.

Added a regression test asserting `getChangedFileCount` returns `1` for a summary with only `symlink_created: 1`. Reverting just the `summary.ts` change reproduces the exact bug you found (count of `0` instead of `1`); with the fix, it passes.

Left the per-file `StatusIcon`/`statusLabel` "modified" fallthrough alone per your note that it can stay a non-blocking follow-up.

`pnpm test`, `pnpm check` (eslint + tsc), and `pnpm format` are all clean.

frontend/src/core/workspace-changes/summary.ts
```diff
@@ -8,7 +8,12 @@ export type WorkspaceChangeLineClass =
   | "meta";
 
 export function getChangedFileCount(summary: WorkspaceChangeSummary) {
-  return summary.created + summary.modified + summary.deleted;
+  return (
+    summary.created +
+    summary.modified +
+    summary.deleted +
+    summary.symlink_created
```
willem-bd, line 15: Updating `getChangedFileCount` here was the right call - without it, a run whose only change was a symlink replacing a file reported 0 changes and hid the badge entirely. But the sibling `sortWorkspaceChanges` in this same file wasn't updated, and it breaks on the new status:

```ts
const statusRank = { created: 0, modified: 1, deleted: 2 } satisfies Record<WorkspaceFileChange["status"], number>;
// ...
const rankDiff = statusRank[left.status] - statusRank[right.status];
```

`statusRank["symlink_created"]` is `undefined`, so the comparator returns `NaN` (verified locally), which violates `Array#sort`'s consistency contract - any list containing a `symlink_created` entry ends up in an unspecified order, *not* the "modified" rank the PR description suggests. (`StatusIcon` / `statusLabel` in `workspace-change-panel.tsx` are fine because they have default branches; the sort has none.)

Root cause: `WorkspaceChangeStatus` in `types.ts` was never widened to include `"symlink_created"`, so the `satisfies Record<WorkspaceFileChange["status"], number>` guard - which would otherwise have demanded a `symlink_created` key here - silently passes, and `pnpm check` stays green despite the runtime bug. Adding `"symlink_created"` to that union plus a rank entry fixes the sort and re-arms the compile-time guard.

frontend/src/core/workspace-changes/types.ts
```diff
@@ -10,6 +10,7 @@ export interface WorkspaceChangeSummary {
   created: number;
   modified: number;
   deleted: number;
+  symlink_created: number;
```
willem-bd, line 18: Adding `symlink_created` to the summary here keeps the count honest, but the per-file type contract wasn't widened to match what the backend now emits:
- `status: "symlink_created"` on `WorkspaceFileChange` (`diff.py::_status`)
- `diff_unavailable_reason: "symlink"` on symlink stubs (`scanner.py::_snapshot_symlink`)
- `symlink` / `symlink_target_before` / `symlink_target_after` on `WorkspaceFileChange`

None of these reached the frontend types - `WorkspaceChangeStatus`, `WorkspaceFileChange`, and `DiffUnavailableReason` here all still omit the new values. Consequences:
- `sortWorkspaceChanges` returns `NaN` for symlink entries (see the comment on `summary.ts`) - a real UI bug.
- `unavailableLabel` (`workspace-change-panel.tsx`) has no `"symlink"` branch, so a symlink change's `diff_unavailable_reason: "symlink"` falls through to the generic "diff unavailable" label.
- `pnpm check` can't catch any of it, because the closed unions make the drift invisible to the compiler.

Widening all three unions in this file (plus their `statusRank` / `unavailableLabel` consumers) would close the contract and turn `pnpm check` into the guard it was designed to be. Worth doing in this PR rather than as a follow-up, since this PR already ships the matching frontend count change.

WillemJiang:
@DaoyuanLi2816 please fix the e2e test.

DaoyuanLi2816:
@WillemJiang sorry for the slow follow-up here -- this was fixed in e024de58, along with a related gap it uncovered:

- The failing `workspace-changes.spec.ts` e2e case: the mock summary predated the `symlink_created` field, so `getChangedFileCount` computed `1 + 1 + 0 + undefined = NaN` and the badge rendered "Edited NaN files" instead of "Edited 2 files". Updated the mock to match the real backend contract.
- While tracing that, found `sortWorkspaceChanges`'s `statusRank` also had no entry for `symlink_created`, so any comparison involving a symlink-created file evaluated to `NaN` and violated `Array#sort`'s ordering contract. Widened `WorkspaceChangeStatus`/`DiffUnavailableReason` and gave `symlink_created` a rank alongside `modified`, so the `satisfies Record<...>` guard can catch a future unranked status again.

`pnpm test` (626 tests), `pnpm check`, and `pnpm format` are clean, and the previously-failing e2e spec plus the full e2e suite (94 tests) pass.

WillemJiang (approved):

Issue #2539, opened 2026-04-26
[harness][sandbox] AIO sandbox deterministic IDs use only 32 bits


hetaoBackend:
## Problem

The AIO sandbox provider derives deterministic sandbox IDs from only the first 8 hex characters of a SHA-256 hash of `thread_id`. That gives roughly 32 bits of namespace. Collisions are plausible at scale and can cause one thread to reuse or collide with another thread's sandbox identity.

## Impact

A collision can attach a run to the wrong persistent sandbox, mixing files, process state, or credentials between threads.

## Suggested Fix

Use a much longer stable identifier, such as 128+ bits from the hash, a namespace UUID, or HMAC-based ID. Store and verify original thread metadata before reusing an existing sandbox.

## Tests

- Unit test that generated IDs include sufficient entropy/length.
- Reuse path verifies sandbox metadata matches the requested thread before attaching.

## References

- `backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py:238`
- `backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py:461`

PR #4171, opened 2026-07-14
fix(sandbox): widen boxlite/aio_sandbox tenant hash and verify identity on reclaim

DaoyuanLi2816:
## Summary

- widen BoxLite and AIO deterministic tenant IDs from 8 to 16 hex characters
- track the full `(user_id, thread_id)` identity for active and warm entries
- fail closed on any identity mismatch without evicting the rightful tenant's warm sandbox
- guard reclaim and health-check races against same-ID entry replacement
- reuse an already-registered BoxLite instance for a late same-tenant race while closing the duplicate instance

Older 8-character containers are not reused after rollout and remain eligible for normal orphan cleanup. Startup-adopted warm entries have unknown identity until their first successful reclaim; subsequent reclaims enforce the recorded identity.

## Tests

- complete BoxLite and AIO provider modules (71 passed)
- targeted same-tenant and cross-tenant collision regressions
- Ruff check and format check on the changed files

### DIFF
--- a/backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
+++ b/backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
@@ -85,6 +85,19 @@ class SandboxBeingDestroyedError(RuntimeError):
         self.sandbox_id = sandbox_id
 
 
+class SandboxIdentityCollisionError(RuntimeError):
+    """A deterministic ID is already tracked for a different user/thread."""
+
+    def __init__(
+        self,
+        sandbox_id: str,
+        stored_key: tuple[str, str] | None,
+        requested_key: tuple[str, str],
+    ) -> None:
+        super().__init__(f"sandbox ID collision for {sandbox_id}: tracked identity is {stored_key!r}, requested identity is {requested_key!r}")
+        self.sandbox_id = sandbox_id
+
+
 def _lock_file_exclusive(lock_file) -> None:
     if fcntl is not None:
         fcntl.flock(lock_file, fcntl.LOCK_EX)
@@ -181,6 +194,8 @@ class AioSandboxProvider(WarmPoolLifecycleMixin[SandboxInfo], SandboxProvider):
         # Containers here can be reclaimed quickly (no cold-start) or destroyed
         # when replicas capacity is exhausted.
         self._warm_pool: dict[str, tuple[SandboxInfo, float]] = {}
+        self._active_sandbox_identity: dict[str, tuple[str, str] | None] = {}
+        self._warm_pool_identity: dict[str, tuple[str, str] | None] = {}
         # sandbox_id -> when reconciliation first saw it running with no lease.
         # Gates adoption behind a recovery grace (see _adoptable_after_grace).
         self._unowned_since: dict[str, float] = {}
@@ -698,6 +713,7 @@ class AioSandboxProvider(WarmPoolLifecycleMixin[SandboxInfo], SandboxProvider):
                     logger.debug("Deferring container %s during reconciliation: this instance is tearing it down", info.sandbox_id)
                     continue
                 self._warm_pool[info.sandbox_id] = (info, current_time)
+                self._warm_pool_identity[info.sandbox_id] = None
             self._unowned_since.pop(info.sandbox_id, None)
             adopted += 1
             logger.info(f"Adopted container {info.sandbox_id} into warm pool (age: {age:.0f}s)")
@@ -726,8 +742,41 @@ class AioSandboxProvider(WarmPoolLifecycleMixin[SandboxInfo], SandboxProvider):
 
         Includes user_id so a previously-created default-bucket sandbox cannot be
         reused for an auth/channel run that should mount a user-scoped bucket.
+
+        During a mixed-version rollout, older 8-character containers are not
+        reused under the new 16-character identity. They remain eligible for
+        normal orphan cleanup while the first new-version acquire cold-starts.
         """
-        return hashlib.sha256(f"{user_id}:{thread_id}".encode()).hexdigest()[:8]
+        return hashlib.sha256(f"{user_id}:{thread_id}".encode()).hexdigest()[:16]
+
+    def _assert_active_identity_available_locked(
+        self,
+        sandbox_id: str,
+        requested_key: tuple[str, str],
+    ) -> None:
+        """Fail closed if an active truncated ID belongs to another identity."""
+        if sandbox_id not in self._sandboxes and sandbox_id not in self._sandbox_infos:
+            return
+
+        stored_key = self._active_sandbox_identity.get(sandbox_id)
+        if stored_key is None:
+            matching_keys = [key for key, mapped_id in self._thread_sandboxes.items() if mapped_id == sandbox_id]
+            if len(matching_keys) == 1:
+                stored_key = matching_keys[0]
+        if stored_key != requested_key:
+            raise SandboxIdentityCollisionError(sandbox_id, stored_key, requested_key)
+
+    def _assert_warm_identity_available_locked(
+        self,
+        sandbox_id: str,
+        requested_key: tuple[str, str],
+    ) -> None:
+        """Fail closed if a warm ID changed tenants during an acquire."""
+        if sandbox_id not in self._warm_pool:
+            return
+        stored_key = self._warm_pool_identity.get(sandbox_id)
+        if stored_key is not None and stored_key != requested_key:
+            raise SandboxIdentityCollisionError(sandbox_id, stored_key, requested_key)
 
     # ── Mount helpers ────────────────────────────────────────────────────
 
@@ -956,8 +1005,10 @@ class AioSandboxProvider(WarmPoolLifecycleMixin[SandboxInfo], SandboxProvider):
 
             sandbox = self._sandboxes.pop(sandbox_id, None)
             self._sandbox_infos.pop(sandbox_id, None)
+            self._active_sandbox_identity.pop(sandbox_id, None)
             self._last_activity.pop(sandbox_id, None)
             self._warm_pool.pop(sandbox_id, None)
+            self._warm_pool_identity.pop(sandbox_id, None)
             self._acquire_epoch.pop(sandbox_id, None)
             for key, mapped_id in list(self._thread_sandboxes.items()):
                 if mapped_id == sandbox_id:
@@ -1197,6 +1248,7 @@ class AioSandboxProvider(WarmPoolLifecycleMixin[SandboxInfo], SandboxProvider):
         with self._lock:
             if sandbox_id not in self._warm_pool:
                 return None
+            self._assert_warm_identity_available_locked(sandbox_id, key)
             if self._being_torn_down_locally(sandbox_id):
                 # The entry deliberately stays in `_warm_pool` for the whole stop
                 # (so a refused claim does not lose the container), so pool
@@ -1237,13 +1289,16 @@ class AioSandboxProvider(WarmPoolLifecycleMixin[SandboxInfo], SandboxProvider):
                 # that is already stopped.
                 logger.info("Warm-pool sandbox %s was claimed for teardown while publishing ownership; not reclaiming it", sandbox_id)
                 return None
+            self._assert_warm_identity_available_locked(sandbox_id, key)
             warm_item = self._warm_pool.pop(sandbox_id, None)
             if warm_item is None:
                 return None
+            self._warm_pool_identity.pop(sandbox_id, None)
             info, _ = warm_item
             sandbox = AioSandbox(id=sandbox_id, base_url=info.sandbox_url)
             self._sandboxes[sandbox_id] = sandbox
             self._sandbox_infos[sandbox_id] = info
+            self._active_sandbox_identity[sandbox_id] = key
             self._last_activity[sandbox_id] = time.time()
             self._thread_sandboxes[key] = sandbox_id
 
@@ -1272,6 +1327,7 @@ class AioSandboxProvider(WarmPoolLifecycleMixin[SandboxInfo], SandboxProvider):
                 prevent. The window is a peer's in-flight container stop, so the
                 thread's next turn discovers nothing and cold-starts cleanly.
         """
+        key = self._thread_key(thread_id, user_id)
         with self._lock:
             if self._being_torn_down_locally(info.sandbox_id):
                 # Discovery is the fall-through once the caches miss, so it is
@@ -1279,9 +1335,10 @@ class AioSandboxProvider(WarmPoolLifecycleMixin[SandboxInfo], SandboxProvider):
                 # only refuse this once the reaper's `del:` claim has landed;
                 # until then it succeeds against our own lease.
                 raise SandboxBeingDestroyedError(info.sandbox_id)
+            self._assert_active_identity_available_locked(info.sandbox_id, key)
+            self._assert_warm_identity_available_locked(info.sandbox_id, key)
 
         sandbox = AioSandbox(id=info.sandbox_id, base_url=info.sandbox_url)
-        key = self._thread_key(thread_id, user_id)
         # Ownership first, so a failure cannot leave a tracked-but-unowned sandbox.
         # There is no container to roll back (we did not create it), but the
         # host-side HTTP client constructed above is ours and must not leak —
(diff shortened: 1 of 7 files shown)


backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -490,10 +570,15 @@ def _reclaim_warm_pool(self, sandbox_id: str) -> str | None:
...
+                return None  # Raced with another thread
+            if self._refuse_if_identity_mismatch(sandbox_id, expected_key):
+                return None  # A different tenant's release() swapped in a colliding entry
             warm_entry = self._warm_pool.pop(sandbox_id, None)
             if warm_entry is None:
                 return None  # Raced with another thread
             self._skip_health_check_warm_ids.discard(sandbox_id)
+            self._warm_identity_store().pop(sandbox_id, None)
             box, _ = warm_entry
             self._boxes[sandbox_id] = box
```
willem-bd, line 606: This promote is also where a *persistent* collision leaks, despite the identity check above. The check correctly refuses tenant B reclaiming tenant A's warm box (it logs "Refusing to reclaim..."), but it does not protect the case where A reclaims its OWN warm box while a colliding tenant B is still active:

1. `A.acquire` -> box A active (id X); `A.release` -> box A warm (id X, identity=A)
2. `B.acquire` -> reclaim refused (A!=B), so B creates a FRESH box and registers `_boxes[X] = box B` - but the create-fallthrough doesn't evict A's lingering warm entry, so `_warm_pool[X]` (box A, identity=A) still coexists with active `_boxes[X]` (box B)
3. `A.acquire` again (B still active) -> reclaim: id X in warm, identity A==A -> MATCHES -> pops box A and runs `self._boxes[sandbox_id] = box` HERE, overwriting box B
4. B's next `get(X)` returns box A -> cross-tenant leak

I confirmed this empirically with a probe (A.acquire -> A.release -> B.acquire -> A.acquire-while-B-active -> get(sid_b)): `box_b_now is box_b` is False - B observes A's box, including A's session marker.

Production impact: none - this needs a real 64-bit collision, which the widened hash makes infeasible. But the existing forced-collision test only covers steps 1-2 and stops, so the leaking path (3-4) is untested and could regress silently (especially if the width is ever weakened again). Suggest extending that test through steps 3-4, and/or evicting the mismatched warm entry in the create-fallthrough so a rightful-owner reclaim can't clobber a colliding tenant's active box.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
```diff
@@ -569,10 +604,15 @@ def _reclaim_warm_pool_sandbox(
...
+                return None
+            if self._refuse_if_identity_mismatch(sandbox_id, key):
+                return None  # A different tenant's release() swapped in a colliding entry
             warm_item = self._warm_pool.pop(sandbox_id, None)
             if warm_item is None:
                 return None
             info, _ = warm_item
+            self._warm_identity_store().pop(sandbox_id, None)
             sandbox = AioSandbox(id=sandbox_id, base_url=info.sandbox_url)
             self._sandboxes[sandbox_id] = sandbox
```
willem-bd, line 1300: Same residual as Boxlite (see the inline comment on boxlite/provider.py's `_reclaim_warm_pool` promote for the full trace + empirical proof): AIO's create-fallthrough (`_register_created_sandbox` -> `self._sandboxes[sandbox_id] = sandbox` here at the promote) likewise doesn't evict a colliding tenant's lingering warm entry, so a rightful-owner reclaim at this line overwrites the colliding tenant's active sandbox. Production-moot (needs a real 64-bit collision) with the same test-coverage gap; AIO's structure is parallel, so the same forced-collision-test extension applies.

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -410,36 +464,57 @@ def release(self, sandbox_id: str) -> None:
...
             if self._shutdown_called:
                 close_box = box
                 self._skip_health_check_warm_ids.discard(sandbox_id)
+                self._warm_identity_store().pop(sandbox_id, None)
             else:
                 self._warm_pool[sandbox_id] = (box, time.time())
                 self._skip_health_check_warm_ids.add(sandbox_id)
+                # Record the real identity this box served so a later reclaim
+                # attempt by a *different* (user_id, thread_id) that happens to
+                # hash to the same sandbox_id can be refused (#3936 follow-up).
```
willem-bd: Minor: this comment cites `(#3936 follow-up)`, but the PR body says "Closes #2539" and discusses #2539/#2576/#3951 - #3936 isn't mentioned anywhere in the description. A reader tracing the issue hits a dead-end. Either update the comment to the correct issue number or reference #3936 in the PR body. (Same comment appears in aio_sandbox_provider.py at the release identity-store write.)

backend/tests/test_boxlite_provider.py
```diff
@@ -1044,3 +1051,164 @@ def test_shutdown_stops_idle_reaper_and_destroys_all_boxes(monkeypatch):
...
+# truncated SHA-256 to 8 hex chars (32 bits), so two distinct (user_id,
+# thread_id) pairs could collide (~77k-pair birthday bound) and, if tenant A's
+# box was still warm, tenant B's acquire() silently received A's live box.
+# Fixed by (a) widening the truncation to 16 hex chars (64 bits, matching
+# community.e2b_sandbox's _stable_seed) and (b) storing the real identity that
+# released each warm-pool entry and refusing to hand it back to a mismatched
+# caller, independent of hash uniqueness.
+
+
+def _find_hash_collision(hex_len: int, max_iterations: int = 500_000) -> tuple[tuple[str, str], tuple[str, str]]:
```
willem-bd: Nit: this brute-forces ~65k+ iterations (the 8-char birthday bound) and is called in 4 tests across the two provider test files (width-isolation + end-to-end, per provider). Consider a session-scoped fixture that caches the discovered colliding pair so it's computed once per test run instead of re-brute-forcing per test. Not blocking.

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -232,6 +245,41 @@ def _lock_for_sandbox(self, sandbox_id: str) -> threading.Lock:
                 self._acquire_locks[sandbox_id] = lock
             return lock
 
+    def _warm_identity_store(self) -> dict[str, tuple[str, str] | None]:
```
willem-bd: Nit: the lazy `getattr`/attach here exists only to support test doubles built via `__new__` that skip `__init__`, but the AIO test already sets `_warm_pool_identity = {}` explicitly (and production always runs `__init__`). The lazy path is therefore barely exercised - it's indirection for a test-only concern. If the doubles always set the attribute, this could be a plain instance attribute and the accessor dropped. Not blocking.

willem-bd (commented):
Reviewed against PR head (`1bf7d77`) and the current codebase; ran both suites against the PR's code (73 passed) and probed the residual-collision path empirically (probe since removed, working tree verified clean).

**Verdict: approve.** The 64-bit widening closes the real production bug (32-bit ~77k-pair birthday bound), and the identity check is a sound reclaim-path backstop that correctly handles the release/acquire inter-check race (re-checked immediately before promotion). `_thread_key` ordering is consistent between acquire and release; `reset()` (Boxlite) correctly preserves identity; AIO has no `reset()` so no parity gap.

The substantive inline comment (on both providers) documents a residual *persistent*-collision leak that is production-moot (needs a real 64-bit collision) but currently untested - the main actionable is extending the forced-collision test through A.re-acquire-while-B-active so the boundary is documented and regression-safe. The rest are minor nits.

- **`area:docs` label looks wrong** - this is a security code fix (2 provider files + 2 test files + doc), not docs-only. Same mislabel seen on PR #4025; should be the sandbox/security area.

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -490,10 +570,15 @@ def _reclaim_warm_pool(self, sandbox_id: str) -> str | None:
 
         # Promote from warm pool to active
         with self._lock:
+            if sandbox_id not in self._warm_pool:
+                return None  # Raced with another thread
+            if self._refuse_if_identity_mismatch(sandbox_id, expected_key):
```
willem-bd: This re-check (and the matching one in the skip-health-check fast path above) correctly guards every *promotion* - the moment a warm box is handed back to a caller. The one path that isn't guarded is the **health-check-failure eviction** a few lines up: `box` is read under CHECK 1's lock, the lock is dropped for the `execute_command("echo ok")` health check, and on failure the code does `warm_entry = self._warm_pool.pop(sandbox_id)` then `_destroy_warm_entry(sandbox_id, warm_entry[0], ...)` without verifying it's still the same `box` that actually failed.

Under a forced/simulated collision plus a concurrent `release()` from the colliding tenant, that window can swap a *different* (healthy) box into `warm[sandbox_id]`, and this eviction path then destroys the wrong tenant's box while leaving the failed one untracked. AIO closes exactly this gap - its failure path calls `_drop_unhealthy_sandbox(..., expected_info=info)`, and `_remove_tracked_sandbox` refuses to drop if the tracked entry was swapped:

```python
if expected_info is not None and active_info is not expected_info and warm_info is not expected_info:
    return None, None, False   # "tracked info changed after health check"
```

So AIO's promotion *and* eviction paths are both swap-safe, while Boxlite's promotion paths now match AIO's rigor but its eviction path doesn't - a defense-in-depth asymmetry. Suggested: in the failure path, re-fetch under lock and verify `warm_entry[0] is box` before popping/destroying; if swapped, leave the new entry and return `None`. Low severity in practice (needs an infeasible 64-bit collision + race + health failure), but the PR's stated goal is symmetric collision-resistance across both providers, and a forced-collision + failing-health-check test would pin this (the current forced-collision test uses a healthy box, so it doesn't exercise the eviction path).

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
```diff
@@ -296,10 +307,17 @@ def _thread_key(thread_id: str, user_id: str) -> tuple[str, str]:
...
+        user-scoped bucket. 16 hex chars (64 bits) matches the width used by
+        ``community.e2b_sandbox``'s equivalent ``_stable_seed``, which makes a
+        collision between two distinct (user_id, thread_id) pairs
+        computationally infeasible. That said, this is a probabilistic
+        guarantee, not a proof, so ``_reclaim_warm_pool_sandbox`` additionally
+        verifies the real identity recorded for a warm-pool entry before
+        handing it back — see that method's docstring.
         """
-        return hashlib.sha256(f"{user_id}:{thread_id}".encode()).hexdigest()[:8]
+        return hashlib.sha256(f"{user_id}:{thread_id}".encode()).hexdigest()[:16]
```
willem-bd, line 750: Widening 8 -> 16 chars is the right call and matches `e2b_sandbox._stable_seed`. One deploy consideration worth a changelog/AGENTS.md note: this changes the id for every `(user_id, thread_id)`, so on a rolling deploy, boxes/containers adopted from a pre-deploy process (named with the old 8-char ids) enter the warm pool under their *old* ids. The new process computes 16-char ids, so those entries are never matched and get reaped by idle timeout instead of being reused - a transient cold-start blip for in-flight sessions on cutover.

This is self-healing and carries no cross-tenant risk (a 16-char lookup can't match an 8-char key, and the new identity check would refuse even a spurious match) - just a brief spike in fresh box/container creation post-deploy that might surprise an operator if it isn't called out.

backend/packages/harness/deerflow/community/boxlite/provider.py
DaoyuanLi2816: Reworked this on current main. A mismatched warm box is destroyed before fresh creation, active identities are tracked separately, and an A re-acquire while colliding tenant B is active now fails closed. The full A-release-B-A sequence is covered for BoxLite.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
DaoyuanLi2816: Applied the same invariant to AIO: mismatched warm containers are destroyed before creation, active identities cannot be overwritten, and the full forced-collision sequence is covered.

backend/packages/harness/deerflow/community/boxlite/provider.py
DaoyuanLi2816: The stale issue reference is gone in the rewritten implementation.

backend/tests/test_boxlite_provider.py
DaoyuanLi2816: Replaced the repeated brute-force search with one fixed, known pair that collides at the legacy 8-character width.

backend/packages/harness/deerflow/community/boxlite/provider.py
DaoyuanLi2816: Removed the lazy accessor. Both identity maps are ordinary instance attributes, and the test doubles initialize them explicitly.

backend/packages/harness/deerflow/community/boxlite/provider.py
DaoyuanLi2816: The BoxLite failure path now re-checks that the warm entry is the same object that failed health checking before it pops or destroys anything. A regression test swaps in another tenant entry during the failed check and verifies it remains intact.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
DaoyuanLi2816: Documented the mixed-version cold-start behavior in both provider ID docstrings and in the PR rollout note.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
```diff
@@ -726,8 +742,41 @@ def _deterministic_sandbox_id(thread_id: str, user_id: str) -> str:
...
+    def _assert_warm_identity_available_locked(
+        self,
+        sandbox_id: str,
+        requested_key: tuple[str, str],
+    ) -> None:
+        """Fail closed if a warm ID changed tenants during an acquire."""
+        if sandbox_id not in self._warm_pool:
+            return
+        stored_key = self._warm_pool_identity.get(sandbox_id)
+        if stored_key is not None and stored_key != requested_key:
```
willem-bd, line 779: Adopted/orphaned warm entries are stamped with identity `None` on adoption (`_reconcile_orphans` ~line 716 here, and `_adopt_existing_boxes` ~line 362 in boxlite), and this guard's `stored_key is not None` clause means a `None`-identity entry **never raises**. So an entry adopted from a running orphan (e.g. after a process restart) isn't identity-protected until the rightful tenant reclaims-and-releases it once and a real key gets stamped.

Exploiting it still requires a 16-char collision (~2^-64) on an orphaned box, so this is a precision/documentation gap rather than a live hole -- but since the PR frames identity tracking as the *enforcement*, a one-line comment here (and on boxlite's warm check) noting "adopted entries are identity-unknown until first reclaim" would keep that claim honest. Same pattern applies to boxlite's inline `if stored_key is not None and stored_key != expected_key` checks in `_reclaim_warm_pool`.

backend/packages/harness/deerflow/community/boxlite/provider.py
```diff
@@ -358,17 +384,39 @@ def acquire(self, thread_id: str | None = None, *, user_id: str | None = None) -
...
             with self._lock:
-                self._boxes[box.id] = box
-                self._thread_boxes[key] = box.id
+                if box.id in self._boxes:
+                    conflict = self._active_box_identity.get(box.id)
+                else:
+                    self._boxes[box.id] = box
+                    self._active_box_identity[box.id] = key
+                    self._thread_boxes[key] = box.id
+            if conflict is not _NO_ACTIVE_IDENTITY:
```
willem-bd, line 415: This branch raises even when `conflict == key` (same identity) -- unlike AIO's `_assert_active_identity_available_locked`, which narrows with `stored_key != requested_key`.

It's currently **unreachable**: `_lock_for_sandbox(sandbox_id)` serializes same-id acquires for the whole `acquire`, so a same-identity race can't reach `_create_box` (the earlier lock block returns the existing box first). But it's a latent foot-gun -- if that per-sandbox lock ever stops fully serializing the create, a legitimate same-identity acquire would spuriously close its freshly-created box and raise. Consider `if conflict is not _NO_ACTIVE_IDENTITY and conflict != key:` to mirror AIO and make the intent explicit.

backend/packages/harness/deerflow/community/aio_sandbox/aio_sandbox_provider.py
DaoyuanLi2816: Added the one-line note to both AIO and BoxLite warm identity checks, and clarified the limitation in the PR description.

backend/packages/harness/deerflow/community/boxlite/provider.py
DaoyuanLi2816: Handled the same-identity case without raising, while still closing the duplicate box and restoring the thread mapping. A regression test covers the late-registration path.

WillemJiang (approved):

PR #4174, opened 2026-07-14
feat(uploads): lazy-load historical files via list_uploaded_files tool

Leeminjing:
﻿## Summary

Replace per-turn injection of all historical upload metadata with on-demand discovery via a new `list_uploaded_files` built-in tool, following the same deferred-discovery pattern used by skills — at the idea level: stop injecting metadata every turn, let the model query a tool on demand.

## Before / After

| Approach | Call chain | API cost | Token cost |
| --- | --- | --- | --- |
| Before: inject all every turn | `model → read_file` | 1 fewer call | Every turn carries all historical file metadata + outlines; grows with file count × conversation length |
| After: lazy-load on demand | `model → list_uploaded_files → model → read_file` | 1 extra call when discovering unknown files | No historical files in prompt; file list only enters context as tool result when needed |

- **Before**: saves one call, but makes the file index a fixed per-turn input cost.
- **After**: occasionally costs one extra call, but makes the file index an on-demand cost.

## What Changed

- `<uploaded_files>` → `<current_uploads>` (current-run new files only, no historical section)
- New `list_uploaded_files` built-in tool with `include_outline: bool | list[str]` parameter
- Outline extraction shared via `deerflow/utils/file_outline.py`
- System prompt updated with lazy-loading guidance
- `UploadsMiddleware.before_agent()` simplified — historical file scan removed
- `InputSanitizationMiddleware` scans only original user content (`ORIGINAL_USER_CONTENT_KEY`); both `<uploaded_files>` and `<current_uploads>` in blocked tags prevents user forgery while server-injected blocks survive untouched
- Memory pipeline (deermem) regexes support both `<uploaded_files>` and `<current_uploads>` with backreference precision
- `list_uploaded_files` gated to lead agent only (not subagents) — subagents have independent ThreadState and cannot exclude current-run files
- IM channel fix (part 1): `_human_input_message()` now passes `additional_kwargs.files` so IM uploads are correctly excluded by `list_uploaded_files`
- IM channel fix (part 2): manual `<uploaded_files>` prepend and `_format_uploaded_files_block()` removed; `UploadsMiddleware` is the sole upload-context producer for both IM and web

## Files

| File | Δ |
|------|---|
| `deerflow/utils/file_outline.py` | +162 (new) |
| `deerflow/tools/builtins/list_uploaded_files_tool.py` | +220 (new) |
| `tests/test_list_uploaded_files_tool.py` | +487 (new) |
| `tests/test_input_sanitization_middleware.py` | +247 |
| `deerflow/agents/middlewares/uploads_middleware.py` | +111/-244 |
| `deerflow/agents/middlewares/input_sanitization_middleware.py` | +86/-6 |
| `deerflow/utils/file_conversion.py` | +6/-112 |
| `app/channels/manager.py` | +9/-3 |
| `deerflow/tools/tools.py` | +8/-1 |
| `deerflow/tools/builtins/task_tool.py` | +6 |
| `deerflow/agents/lead_agent/prompt.py` | +4/-2 |
| `deerflow/tools/builtins/__init__.py` | +2 |
| `deerflow/agents/memory/.../message_processing.py` | +2/-2 |
| `deerflow/agents/memory/.../prompt.py` | +1/-1 |
| `deerflow/agents/memory/.../updater.py` | +1/-1 |
| `tests/test_uploads_middleware_core_logic.py` | +75/-72 |
| `tests/test_tool_error_handling_middleware.py` | +9/-9 |

## Test Results

- list_uploaded_files tool: 25 passed (incl. integration + IM regression tests)
- Uploads middleware: 48 passed
- Input sanitization: 146 passed
- Affected files total: 219 passed
- Full suite: 7480 passed, 144 failed (pre-existing Windows failures unrelated to this PR)

## Known follow-ups (post-merge)

- #4212 — Clean up backward-compat `<uploaded_files>` handling in deermem, denylist, and frontend (injection side unified in this PR)
- #4213 — Deduplicate extension-summary helpers
- #4214 — Blocked: subagent access to `list_uploaded_files` requires state propagation fix first


backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
```diff
@@ -0,0 +1,216 @@
...
+    uploads_dir = paths.sandbox_uploads_dir(thread_id, user_id=user_id)
+
+    if not uploads_dir.exists():
+        return {"files": [], "message": "No uploads directory for this thread."}
+
+    # Resolve the set of filenames uploaded in the current run so we can exclude them.
+    current_run_filenames: set[str] = set()
+    try:
+        state = runtime.state
+        uploaded = state.get("uploaded_files") if isinstance(state, dict) else getattr(state, "uploaded_files", None)
```
fancyboi999, line 92: [P2] Do not treat persisted `uploaded_files` state as current-run files

`ThreadState.uploaded_files` is a plain LastValue field, and `UploadsMiddleware.before_agent()` returns `None` when the next turn has no new upload, so this list still contains the previous turn's files. Reading it here as `current_run_filenames` therefore excludes exactly the file that just became historical: upload `report.pdf` in turn 1, ask `what did I upload?` in turn 2, and `list_uploaded_files` returns `No historical uploaded files in this thread.` Derive the exclusion set from the current last human message/run-local metadata (or explicitly clear the state at each run) and add a two-turn regression test; the existing test only injects an already-curated state list and misses the persistence boundary.

fancyboi999 (changes requested):
One deterministic cross-turn regression needs to be fixed before this lazy-discovery path is reliable.

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
Leeminjing: Thanks for catching this.
Fixed in f7b0606.
uploaded_files is now explicitly cleared when the current turn has no new uploads, so files from the previous turn are correctly treated as historical files. I also added a two-turn regression test covering this persistence boundary.

WillemJiang:
@Leeminjing  Thanks for your contribution. Please fix the CI test errors.

Leeminjing:
@WillemJiang I’ve fixed the reported issues and updated the branch. Could you please help move the approval process forward?

WillemJiang:
@Leeminjing, please run 'make install' from the root directory to install the pre-commit handler to avoid the lint error next time.

backend/packages/harness/deerflow/agents/middlewares/uploads_middleware.py
```diff
@@ -167,88 +106,52 @@ def _format_file_entry(self, file: dict, lines: list[str]) -> None:
...
-            omitted_historical_files: Older historical files omitted from the prompt context.
+            files: Files uploaded in the current message.
+            omitted_files: Files omitted from the prompt context (over cap).
 
         Returns:
-            Formatted string inside <uploaded_files> tags.
+            Formatted string inside <current_uploads> tags.
         """
-        lines = ["<uploaded_files>"]
+        lines = ["<current_uploads>"]
```
willem-bd, line 135: The `<uploaded_files>` -> `<current_uploads>` rename wasn't propagated to the memory pipeline, which still strips the old tag:

- `memory/message_processing.py:11,75` - `_UPLOAD_BLOCK_RE` matches `<uploaded_files>`; `filter_messages_for_memory` does `if "<uploaded_files>" in content_str:` then strips via the regex.
- `memory/prompt.py:808` - `re.sub(r"<uploaded_files>...</uploaded_files>", "", content)`.
- `memory/updater.py:501` - `r"|<uploaded_files>"` in a regex alternation.

Because `filter_messages_for_memory` checks for the old name, a message carrying `<current_uploads>` falls to the `else` and is appended **unstripped** - so the file-list block (filenames, sizes, outlines) is fed to the memory-extraction LLM as conversation content, polluting long-term memory with file-metadata "facts". The `skip_next_ai` path (uploads-only messages) also no longer fires.

This is silent - the test suite passes because the memory tests don't flow real `<current_uploads>` blocks through `filter_messages_for_memory`. A shared constant for the tag name (used here, the memory strippers, and the sanitization sets) would prevent this class of rename-incompleteness.

backend/tests/test_input_sanitization_middleware.py
```diff
@@ -247,6 +247,10 @@ def test_neutralize_untrusted_tags_covers_framework_structured_tags(tag):
     "skill",
     "skill_content",
     "user_request",
+    # UploadsMiddleware instruction block — wraps trusted server-generated file
+    # metadata, not user-controlled content. Renamed from uploaded_files in the
+    # lazy-loading PR (#4174).
+    "current_uploads",
```
willem-bd: The rename moved this tag from `_BLOCKED_TAG_NAMES` (where `uploaded_files` still sits, now stale - no longer emitted) to `_EXEMPT_BLOCK_TAGS`. So a user typing `<current_uploads>` in their input is no longer HTML-escaped - they can forge a structured uploads block indistinguishable from the server's. The old blocked `<uploaded_files>` closed that.

The exemption is also load-bearing the other way: `InputSanitizationMiddleware._process_request` runs `_check_user_content` on the *full message text* at model-call time (input_sanitization_middleware.py:286-296), which includes the block `UploadsMiddleware.before_agent` prepended - so blocking `current_uploads` would HTML-escape the server's own block. The comment's rationale ("trusted server-generated metadata") addresses server trust, but the blocklist's job is preventing user forgery, which the exemption now skips.

Cleaner fix: block `current_uploads` (restore forgery prevention) and have `InputSanitizationMiddleware` operate on the original user content (`ORIGINAL_USER_CONTENT_KEY`) rather than the full prepended message, so the server's trusted block isn't escaped. At minimum, remove the stale `uploaded_files` from `_BLOCKED_TAG_NAMES` so the classification isn't half-and-half.

willem-bd (commented):
Reviewed the uploads lazy-loading change. The design is sound - on-demand discovery via `list_uploaded_files` removes the per-turn file-index cost (which grew with file count x conversation length) at the price of one extra call when discovery is needed. Outline extraction is properly deduplicated into `file_outline.py`, the tool is well-tested (20 tests incl. the cross-turn state-clearing regression), and per-user isolation in the tool is correct.

Inline notes below:
1. (High, blocker) The `<uploaded_files>` -> `<current_uploads>` rename wasn't propagated to the memory pipeline - the file-list block now leaks into memory extraction.
2. (Medium) The tag moved from `_BLOCKED_TAG_NAMES` (blocked) to `_EXEMPT_BLOCK_TAGS` (exempt) - an inconsistency and a user-forgery surface worth resolving.
3. (Nit) `ponytail:` stray word in the re-export comment.
4. (Nit) Dead branch in `_format_omitted_summary`.

#1 is a blocker - the memory files need updating (ideally via a shared tag-name constant) before merge. #2 is a secondary prompt-injection consideration.

backend/packages/harness/deerflow/utils/file_conversion.py
```diff
@@ -177,121 +176,18 @@ async def convert_file_to_markdown(file_path: Path) -> Path | None:
...
-#      headers like **2023** **2022** **2021** while allowing non-ASCII titles such as
-#      **1** **概述** or accented words (negative lookahead instead of [A-Za-z])
-#   4. At most two additional blocks (four total) with [^*]+ (no * inside) to keep
-#      the regex linear and avoid ReDoS on attacker-controlled content
-_SPLIT_BOLD_HEADING_RE = re.compile(r"^\*\*[\dA-Z][\d\.]*\*\*\s+\*\*(?!\d[\d\s.,\-–—/:()%]*\*\*)[^*]+\*\*(?:\s+\*\*[^*]+\*\*){0,2}\s*$")
-
-# Maximum number of outline entries injected into the agent context.
-# Keeps prompt size bounded even for very long documents.
-MAX_OUTLINE_ENTRIES = 50
+# ponytail: backward-compat re-exports — outline extraction moved to file_outline.py.
```
willem-bd: Nit: the `ponytail:` prefix here looks like a stray word - did you mean `Note:` or just `Backward-compat re-exports`?

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
```diff
@@ -0,0 +1,216 @@
...
+
+def _extension_label(file_path: Path) -> str:
+    suffix = file_path.suffix.lower()
+    return suffix or "(no extension)"
+
+
+def _format_omitted_summary(omitted: list[str], total_omitted: int) -> str:
+    counts = Counter(_extension_label(Path(f)) for f in omitted)
+    parts = [f"{count} {ext}" for ext, count in sorted(counts.items())]
+    if total_omitted > len(omitted):
```
willem-bd: Nit: `if total_omitted > len(omitted)` is never true here. `omitted_paths` is built from *all* `candidates[max_results:]`, so `len(omitted) == total_omitted` always, and the `... (N total)` suffix never renders. Either drop the branch or, if you want to cap the counted extensions, truncate `omitted_paths` first.

willem-bd (commented):
## Review summary

**Request changes.** The lazy-loading design is sound and the per-turn token savings are real; the new tool is well-tested (including a good cross-turn state-clear regression test). But renaming `<uploaded_files>` -> `<current_uploads>` leaves two other pipelines keyed off the old tag, producing a memory-pollution regression (high) and a channel-path inconsistency (medium). Details in the inline comments.

**Verified on this branch** (commit `3dcd0bb`):
- Middleware ordering: `UploadsMiddleware` (Layer-2 `before_agent`) prepends the block *before* `InputSanitizationMiddleware` (Layer-1 `wrap_model_call`) processes the whole message. So making `<current_uploads>` *exempt* (not blocked) in `input_sanitization_middleware.py` is correct - blocking would HTML-escape the legit block. Good call.
- Memory regression: ran `filter_messages_for_memory` directly - `<uploaded_files>` blocks are stripped/skipped, `<current_uploads>` blocks pass through to the memory LLM verbatim.

**Strengths:** clean `file_outline.py` extraction; the `uploaded_files` clear-on-non-upload-turn fix is right; `current_uploads` exempt classification is architecturally correct; net per-turn perf win.

**Minor nits (not blocking):**
- `file_conversion.py`: the `# ponytail: backward-compat re-exports` comment - "ponytail" reads like a stray token; "backward-compat" is the intent.
- `list_uploaded_files` is in `BUILTIN_TOOLS` only, not `SUBAGENT_TOOLS` (by design - lead discovers, subagents work via `read_file`/`grep`). Fine, but confirm the lead passes discovered paths when delegating file analysis to a subagent.
- The `@tool` wrapper + `Runtime` injection isn't exercised end-to-end (tests use `_list_uploaded_files_impl` directly). Low risk.

backend/packages/harness/deerflow/agents/middlewares/uploads_middleware.py
willem-bd, line 135: 🟠 Memory pipeline doesn't strip `<current_uploads>` (regression)

This rename from `<uploaded_files>` -> `<current_uploads>` breaks the memory-stripping pipeline, which still keys off the old tag. None of these were updated:
- `agents/memory/message_processing.py` - `if "<uploaded_files>" in content_str` + `_UPLOAD_BLOCK_RE = re.compile(r"<uploaded_files>...</uploaded_files>")`
- `agents/memory/prompt.py:808` - `re.sub(r"<uploaded_files>...", ...)`
- `agents/memory/updater.py:_UPLOAD_SENTENCE_RE` - matches `<uploaded_files>` when scrubbing memory facts

I verified the impact on this branch by calling `filter_messages_for_memory` directly:
- `<uploaded_files>` block -> stripped to empty, the upload-only turn is **dropped** from memory input (as designed).
- `<current_uploads>` block -> **passed through verbatim** to the memory-updating LLM, including filenames, `/mnt/user-data/uploads/...` paths, and outline/preview heading text.

So upload-only turns are no longer skipped, and outline prose can leak into long-term memory - exactly the ephemeral-file-info persistence those regexes exist to prevent. No test covers it (no memory test files in this PR).

Fix: match both tags, e.g. `r"</?(?:uploaded_files|current_uploads)>"` in `_UPLOAD_BLOCK_RE`, the literal `"<uploaded_files>" in content_str` check, `memory/prompt.py:808`, and `_UPLOAD_SENTENCE_RE`. The channel path still uses `<uploaded_files>` (see the prompt comment), so both must remain matched.

backend/packages/harness/deerflow/agents/lead_agent/prompt.py
```diff
@@ -584,14 +584,16 @@ def _build_subagent_section(
...
+- Historical uploads: `/mnt/user-data/uploads` - Files from earlier turns. Use `list_uploaded_files` to discover which historical files exist. If you know the filename, access it directly with `read_file` or `grep`.
 - User workspace: `/mnt/user-data/workspace` - Working directory for temporary files
 - Output files: `/mnt/user-data/outputs` - Final deliverables must be saved here
 
 **File Management:**
-- Uploaded files are automatically listed in the <uploaded_files> section before each request
+- Newly uploaded files in this run are listed in the `<current_uploads>` section before your first response
 - Use `read_file` tool to read uploaded files using their paths from the list
 - For PDF, PPT, Excel, and Word files, converted Markdown versions (*.md) are available alongside originals
+- Files uploaded in previous turns are NOT automatically listed. Use `list_uploaded_files` to discover them on demand — it returns filenames, sizes, and optionally document outlines
```
willem-bd, line 626: 🟡 IM channel path still emits the old `<uploaded_files>` tag

`backend/app/channels/manager.py:1588` (`_format_uploaded_files_block`) still prepends an `<uploaded_files>` block to Feishu/Slack/Telegram/Discord/DingTalk messages and isn't in this PR's changed files. After this PR the chat UI uses `<current_uploads>` + the `list_uploaded_files` lazy model, but IM-channel runs still inject all of a message's uploads directly as `<uploaded_files>` - a tag this updated prompt no longer documents, and a path that doesn't get the historical/current split.

Functionally channels still work (the agent can read the block), but for a `needs-validation` PR this is a real contract divergence: an agent on a channel run sees neither `<current_uploads>` nor the `list_uploaded_files` flow the prompt now describes. Worth either aligning `channels/manager.py` to `<current_uploads>` (and deciding whether channels get the lazy split), or documenting why channels intentionally keep the legacy tag.

backend/packages/harness/deerflow/agents/middlewares/uploads_middleware.py
```diff
@@ -327,73 +220,34 @@ def before_agent(self, state: UploadsMiddlewareState, runtime: Runtime) -> dict
...
-        )
-        for file in historical_files:
-            file_path = file.pop("_host_path")
-            file.pop("_mtime", None)
-            outline, preview = _extract_outline_for_file(file_path)
-            file["outline"] = outline
-            file["outline_preview"] = preview
-
-        # Attach outlines to new files as well
+        if not new_files:
```
willem-bd, line 233: 🟡 Stale `uploaded_files` on the early-return paths

Good that the non-upload case clears `uploaded_files` here. But the two earlier `return None` paths in `before_agent` (`if not messages` and `if not isinstance(last_message, HumanMessage)`) don't clear it. `uploaded_files` is checkpointed state, so if `before_agent` is re-entered on such a path (e.g. a regeneration whose last message is an AI message), the previous turn's `uploaded_files` persists and `list_uploaded_files` then over-excludes files that have since become historical - the tool reads `state["uploaded_files"]` as "current run" and excludes those filenames.

Narrow, but the new exclusion semantics make `uploaded_files` accuracy load-bearing across turns in a way it wasn't before. Consider clearing `uploaded_files` on those early-return paths too (or having the tool treat a stale-looking `uploaded_files` defensively).

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
willem-bd: 🟢 Dead branch: `total_omitted > len(omitted)` is always false here

The caller passes `omitted_paths = [p.name for _, p in candidates[max_results:]]` - i.e. *all* omitted filenames - so `total_omitted` (passed as `total_count - max_results`) always equals `len(omitted)`. The `"... ({total_omitted} total)"` suffix never fires in practice; `test_with_total` only hits it by calling the helper directly with a truncated sample.

Either drop the branch, or have the caller pass a sampled subset (e.g. first ~50 omitted names) so the summary stays bounded for threads with thousands of uploads.

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
```diff
@@ -0,0 +1,216 @@
...
+    else:
+        outline_for_all = False
+        outline_filenames = set(include_outline)
+
+    # Collect historical files (sorted by mtime descending).
+    # Skip .md files that are conversion artifacts (have a same-stem non-.md sibling).
+    candidates: list[tuple[float, Path]] = []
+    try:
+        # First pass: collect all non-staging filenames so we can detect conversion artifacts
+        all_names: set[str] = {entry.name for entry in os.scandir(uploads_dir) if entry.is_file() and not is_upload_staging_file(entry.name)}
```
willem-bd: 🟢 Minor perf: two `os.scandir` passes + unbounded outline reads

Two passes here (first builds `all_names` to detect conversion artifacts, then iterates again). Fine for typical threads, but could be one pass collecting entries then filtering in memory.

Separately, with `include_outline=True`, `extract_outline_for_file` reads each `.md` file to EOF (the outline loop only breaks at the 50-entry heading cap, not at EOF), for up to `_MAX_MAX_RESULTS=100` files - real IO for large converted PDFs. On-demand only, so acceptable; just flagging. Capping bytes read per outline (e.g. first N KB) would bound it.

backend/packages/harness/deerflow/agents/lead_agent/prompt.py
Leeminjing: Thanks for catching this.
I left this as-is deliberately for two reasons:
1. Layer boundary. This PR touches packages/harness/ (the agent framework). channels/manager.py is in app/ (the application layer), and the harness→app import firewall (test_harness_boundary.py) means these layers are changed in separate PRs by design.
2. Porting the lazy-loading model to channels isn't a tag rename, it's a separate design question: should channel runs get <current_uploads> + list_uploaded_files, or keep the simpler all-in-one injection? That decision belongs in its own PR.

backend/packages/harness/deerflow/agents/middlewares/uploads_middleware.py
Leeminjing: Fixed in the latest push (6da25d88). All three sites now match <(?:uploaded_files|current_uploads)> — message_processing.py, prompt.py:808, and updater.py. The literal string check in filter_messages_for_memory now also checks for both tags.

backend/packages/harness/deerflow/agents/middlewares/uploads_middleware.py
Leeminjing: Fixed. Both early-return paths now return {"uploaded_files": []} instead of None. Same rationale as the no-new-files path — UploadsMiddleware owns uploaded_files state and must clear it on every exit that doesn't produce a current-turn value.

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
Leeminjing: Dropped the branch, the total_omitted parameter, and the corresponding test. For now the helper gets the full list — threads with thousands of uploads are unlikely in practice, and if it becomes a problem, sampling the first N omitted names can be added then.

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
Leeminjing: Acknowledged. Both are fine for typical usage, as you noted. If perf becomes an issue for threads with many large files, the two-pass scan can be folded into one and outline reads can be byte-capped. Leaving as-is for now.

backend/tests/test_input_sanitization_middleware.py
Leeminjing: Went with the cleaner fix rather than the minimum:

- current_uploads is in _BLOCKED_TAG_NAMES — user-forged tags are HTML-escaped. 
- _process_request now reads ORIGINAL_USER_CONTENT_KEY first. When present (set by UploadsMiddleware on upload turns), only the user's original text is scanned — the server's prepended <current_uploads> block is never checked against the denylist. When absent (plain text turns without uploads), falls back to full-content scanning, which is safe since no server-injected blocks exist in those messages. 
- Stale uploaded_files removed from _BLOCKED_TAG_NAMES. 
- Added tests: forged <current_uploads> escaped, server block preserved.

backend/packages/harness/deerflow/utils/file_conversion.py
Leeminjing: Removed. It was a marker left by the Ponytail plugin I use.

Leeminjing:
Thanks for the thorough review — I’ve addressed the comments in the latest push.

Leeminjing:
Thanks for the reminder. I’ve run make install from the repository root and installed the pre-commit hooks.

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
```diff
@@ -0,0 +1,214 @@
...
+    uploads_dir = paths.sandbox_uploads_dir(thread_id, user_id=user_id)
+
+    if not uploads_dir.exists():
+        return {"files": [], "message": "No uploads directory for this thread."}
+
+    # Resolve the set of filenames uploaded in the current run so we can exclude them.
+    current_run_filenames: set[str] = set()
+    try:
+        state = runtime.state
+        uploaded = state.get("uploaded_files") if isinstance(state, dict) else getattr(state, "uploaded_files", None)
```
willem-bd, line 92: The current-run exclusion depends on `runtime.state["uploaded_files"]` being populated by `UploadsMiddleware.before_agent` (which returns `{"uploaded_files": [...]}` or `{"uploaded_files": []}`). The tests cover the logic by passing `state_uploaded` through a mocked runtime, but none confirm that LangGraph actually surfaces the `before_agent` state write to `runtime.state` **inside a tool call in the same turn**.

If `runtime.state` is a pre-`before_agent` snapshot (or doesn't expose middleware-written keys), `current_run_filenames` stays empty and the tool fails open -- listing current-run files alongside historical ones, so they appear in both `<current_uploads>` and this tool's result. The `try/except: pass` at line 86 makes that failure silent.

This is the core `needs-validation` risk on the PR. A real-path smoke test (upload a file, same turn call `list_uploaded_files`, assert the file is excluded) would gate it. (High)

backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
```diff
@@ -297,10 +297,33 @@ def _process_request(self, request: ModelRequest) -> ModelRequest:
...
+            # those: no server-injected <current_uploads> block exists to
+            # accidentally escape.
+            preserved_kwargs = dict(msg.additional_kwargs or {})
+            original_user_content = preserved_kwargs.get(ORIGINAL_USER_CONTENT_KEY)
+            if isinstance(original_user_content, str) and original_user_content:
+                processed_user = _check_user_content(original_user_content)
+                if processed_user != original_user_content:
+                    # Replace only the user's text suffix within the full
+                    # content — server-prepended blocks stay untouched.
+                    idx = text_content.rfind(original_user_content)
```
willem-bd, line 325: The `rfind`-based suffix replacement assumes `original_user_content` is a clean suffix of `text_content`, but the two are produced by different helpers that disagree on list (multimodal) content:

- `original_user_content` is set via `message_content_to_text`, which **filters empty parts** and accepts any dict with a `text` key.
- `text_content` comes from `_extract_text_from_content`, which only accepts `type=="text"` blocks and does **not** filter empties.

For string content (the common path) this is fine, but a list message with an interspersed empty text block makes `rfind` return -1 -> the `else` fallback runs `_check_user_content(text_content)` on the **full** content, which HTML-escapes the server's `<current_uploads>` block (degraded to the old behavior; not a security regression since user forgeries are still escaped in that branch).

Two asks:
1. Add a multimodal (list-content) test through this path.
2. Add an integrated test for the actual attack scenario -- `ORIGINAL_USER_CONTENT_KEY` set **and** user text containing a forged `<current_uploads>...</current_uploads>` tag -- asserting the forged tag is escaped **and** the server block is preserved. `test_escapes_user_forged_current_uploads_tag` only tests the `_check_user_content` primitive, and `test_server_current_uploads_block_not_escaped` only uses clean user text; their intersection is untested. (Medium)

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/message_processing.py
```diff
@@ -7,7 +7,7 @@
 from copy import copy
 from typing import Any
 
-_UPLOAD_BLOCK_RE = re.compile(r"<uploaded_files>[\s\S]*?</uploaded_files>\n*", re.IGNORECASE)
+_UPLOAD_BLOCK_RE = re.compile(r"<(?:uploaded_files|current_uploads)>[\s\S]*?</(?:uploaded_files|current_uploads)>\n*", re.IGNORECASE)
```
willem-bd: Minor: the alternation allows mismatched open/close pairs, e.g. `<uploaded_files>...</current_uploads>`. Harmless in practice (the server emits matching pairs and the match is non-greedy), but a backreference would be precise: `r"<(?P<tag>uploaded_files|current_uploads)>[\s\S]*?</(?P=tag)>\n*"`. Same pattern is duplicated in `prompt.py` and `updater.py`. (Low)

backend/packages/harness/deerflow/utils/file_conversion.py
```diff
@@ -177,121 +176,18 @@ async def convert_file_to_markdown(file_path: Path) -> Path | None:
...
-_SPLIT_BOLD_HEADING_RE = re.compile(r"^\*\*[\dA-Z][\d\.]*\*\*\s+\*\*(?!\d[\d\s.,\-–—/:()%]*\*\*)[^*]+\*\*(?:\s+\*\*[^*]+\*\*){0,2}\s*$")
-
-# Maximum number of outline entries injected into the agent context.
-# Keeps prompt size bounded even for very long documents.
-MAX_OUTLINE_ENTRIES = 50
+# Backward-compat re-exports — outline extraction moved to file_outline.py.
+# Keep these names available under the original module so existing callers
+# (tests/test_uploads_middleware_core_logic.py, tests/test_file_conversion.py)
+# don't have to update their imports.
+from deerflow.utils.file_outline import (  # noqa: E402, F401
```
willem-bd: Nit: the backward-compat re-export sits mid-file (hence `# noqa: E402`). Since `file_outline` has no circular dependency on `file_conversion`, this could move to the top-level imports. Not blocking -- the `# noqa` is honest about it. (Low)

willem-bd (commented):
Reviewed against head 52727da. Overall this is a solid, well-motivated change -- moving the per-turn historical scan + outline extraction to on-demand is a real token-cost win for long conversations, the cross-turn state-clear fix is correctly identified and tested, and the security model is sound (user-forged <current_uploads> tags are escaped in every branch; no regression). Left 4 inline comments.

A few items that don't sit on diff lines, noting here:

- **IM channels still emit `<uploaded_files>`** (`backend/app/channels/manager.py:773`, not touched by this PR). It works -- `<uploaded_files>` stays in the blocklist and the updated deermem stripper handles both tags -- but two tag names for the same "uploads context" concept is a maintenance smell; worth a follow-up to unify. Note also that IM channels still inject all historical files every turn, so the token savings are web-frontend-only.
- **Duplicated extension-summary helpers**: `_format_omitted_summary` (this tool, takes `list[str]`) and `_format_omitted_file_types` (middleware, takes `list[dict]`) are near-duplicate logic -- could share a primitive. Minor.
- **Subagent access**: `list_uploaded_files` lands in the lead/builtin toolset. Worth confirming whether delegated subagents should also get it (a subagent working on historical files currently can't discover them).

Recommendation: **approve with conditions** -- gate on the `runtime.state` visibility validation and the integrated forged-tag test (see inline comments); the rest is follow-up polish.

Leeminjing:
Thanks again for the exceptionally thorough review.All items addressed in the latest push (d8d3f63, now rebased on main):

Inline comments:

1. (High) runtime.state visibility — Added TestMiddlewareToolStateBridge: UploadsMiddleware.before_agent() writes state, _list_uploaded_files_impl() reads it — verifies format compatibility between middleware write and tool read. The try/except: pass is now logger.warning(..., exc_info=True) so silent degradation is observable.
2. (Medium) multimodal fallback + forged-tag test — Instead of adding a per-block sanitization path in the rfind-failure branch, I fixed the root cause: _extract_text_from_content now skips empty text blocks, matching message_content_to_text's behaviour. This eliminates the text extraction disagreement that caused rfind to fail in the first place. The rfind-failure fallback still exists as a safety net with logger.warning. Added both requested tests: test_multimodal_list_content_forged_tags_escaped and test_forged_current_uploads_escaped_server_block_preserved (with exact unescaped-tag count assertions).
3. (Low) regex backreference — Applied (?P<tag>)(?P=tag) to message_processing.py and prompt.py (block-level regexes). Left updater.py as alternation — it's _UPLOAD_SENTENCE_RE, a sentence-level pattern matching individual tag appearances (not open/close pairs), so backreference doesn't apply cleanly there.
4. (Low) re-export placement — Moved to top of file_conversion.py, dropped # noqa: E402.

Non-diff observations:

- IM channels → #4212 (discussion issue, not pre-committed to implementation)
- Duplicate helpers → #4213
- Subagent access → #4214

Test results: 213 passed across the three affected test files, ruff clean.

backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
```diff
@@ -73,13 +73,13 @@
         "critical_reminders",
         "response_style",
         "citations",
+        "current_uploads",
```
willem-bd, line 83: Dropping `uploaded_files` from the denylist while adding `current_uploads` leaves the old tag name un-neutralized in user input. The denylist's own docstring says it must cover "the framework's authority blocks as a class," and deermem still processes `<uploaded_files>` for backward-compat with persisted history - so a user forging `<uploaded_files>` mimics historical server context and is no longer escaped (the system prompt treats structured tags as trusted).

Consider keeping **both** tags blocked; the cost is negligible and it closes the old-tag forgery path consistently.

backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
```diff
@@ -297,10 +300,44 @@ def _process_request(self, request: ModelRequest) -> ModelRequest:
...
+                    idx = text_content.rfind(original_user_content)
+                    if idx >= 0:
+                        processed = text_content[:idx] + processed_user
+                    else:
+                        # _extract_text_from_content and message_content_to_text
+                        # disagreed on text extraction — rfind failed.
+                        # text_blocks is None for string content; if we get here
+                        # with string content, log the unexpected path and fall
+                        # back.  Server-injected blocks are safe to escape here
+                        # because string content means no <current_uploads> was
```
willem-bd: This comment is inaccurate. `UploadsMiddleware` does **not** convert string content to list content when it injects the block - for string content it keeps it a string and prepends inline (`uploads_middleware.py`: `updated_content = f"{files_message}\n\n{original_content}"`). It only produces list content when the input was already a list.

The fallback is safe for strings for a *different* reason: `message_content_to_text(str)` returns the string verbatim, so `original_user_content` is a suffix of `text_content` and `rfind` succeeds - the fallback isn't reached. Worth correcting so the next maintainer doesn't reason from the wrong premise.

backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
```diff
@@ -297,10 +300,44 @@ def _process_request(self, request: ModelRequest) -> ModelRequest:
...
+                        # prepended (UploadsMiddleware converts to list content
+                        # when it injects the block).
+                        logger.warning(
+                            "rfind failed with original_user_content set; falling back to full-content sanitization",
+                        )
+                        processed = _check_user_content(text_content)
+                else:
+                    processed = text_content  # no change needed
+            else:
+                processed = _check_user_content(text_content)  # fallback
```
willem-bd, line 382: This fallback (and the `rfind < 0` branch above) runs `_check_user_content(text_content)` on the **full** content, which escapes the server's `<current_uploads>` block (it's in the denylist). Two reachable triggers:

- **Empty user text on an upload** (e.g. attach a PDF with no message): `ORIGINAL_USER_CONTENT_KEY` is set to `""` (falsy) -> skips the `if` -> fallback -> server block escaped.
- **Multimodal extraction disagreement**: `_extract_text_from_content` only keeps `{"type":"text"}` dict blocks, but `message_content_to_text` (which set `original_user_content`) also includes raw-string list items and `.get("text")` on non-text blocks. Content like `["raw string", {type:image, text:"alt"}]` makes the two strings differ -> `rfind` returns -1 -> fallback -> server block escaped.

Not a regression vs `main` (main always escaped the server block), but it contradicts the PR's "server-injected blocks survive untouched" guarantee. The fallback is "escape everything," so no forgery survives - the impact is UX (server block neutralized), not security. Could tighten by skipping the server portion when the key is present-but-empty on an upload turn.

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
```diff
@@ -0,0 +1,220 @@
...
+    else:
+        outline_for_all = False
+        outline_filenames = set(include_outline)
+
+    # Collect historical files (sorted by mtime descending).
+    # Skip .md files that are conversion artifacts (have a same-stem non-.md sibling).
+    candidates: list[tuple[float, Path]] = []
+    try:
+        # First pass: collect all non-staging filenames so we can detect conversion artifacts
+        all_names: set[str] = {entry.name for entry in os.scandir(uploads_dir) if entry.is_file() and not is_upload_staging_file(entry.name)}
```
willem-bd: Minor: `os.scandir(uploads_dir)` is called twice here (once to build `all_names`, once to iterate on line 120). Uploads dirs are small so this is negligible, but a single `entries = list(os.scandir(uploads_dir))` reused for both would be cleaner and avoid a second syscall.

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
```diff
@@ -0,0 +1,220 @@
...
+            if not entry.is_file():
+                continue
+            if is_upload_staging_file(entry.name):
+                continue
+            if entry.name in current_run_filenames:
+                continue
+            # Skip .md files that are conversion artifacts of another file
+            if entry.name.endswith(".md"):
+                stem = entry.name[:-3]  # remove ".md"
+                non_md_siblings = {n for n in all_names if n != entry.name and Path(n).stem == stem}
```
willem-bd, line 132: Edge case: if a user uploads both `report.pdf` **and** their own `report.md`, this sibling check suppresses the `.md` as a "conversion artifact" - the user's manually-uploaded markdown file is hidden from the listing. The `Path(n).stem == stem` check is also broad (e.g. `report.tar.gz` vs `report.tar.md`). Probably acceptable for the MVP, but worth a note / future follow-up if users upload mixed `.md` + convertible pairs.

willem-bd (commented):
Review of the lazy-load historical files change. The core mechanism is sound - verified that `uploaded_files` uses LangGraph's default replace reducer (no annotation on `UploadsMiddlewareState`), so the cross-turn clearing via `return {"uploaded_files": []}` works correctly, and the middleware->tool state-visibility contract holds. Strong test coverage (incl. the bridge + cross-turn regression tests).

Inline comments below: one minor security-consistency note on the denylist, one inaccurate fallback comment, one correctness edge case in the fallback path (non-regressing), and two minor nits on the new tool. None are blocking.

Leeminjing:
Thanks again for the careful fourth pass.All five addressed across 4ccc6ae, fae1a1c, and 9edd613:

1. Denylist — uploaded_files re-added. Both tags blocked equally.

2. Comment — Corrected. The real reason string fallback is unreachable is that message_content_to_text(str) is a no-op, not list conversion.

3. Two reachable edge cases — Split the rfind-failure fallback into two paths:

- Distinguishable (isinstance(content, list) and len(content) >= 2): UploadsMiddleware guarantees content[0] is the server block. Only content[1:] is sanitized — traversing each user block by type (str / type:"text" dict / other like image) and applying neutralize_untrusted_tags. Server block untouched. Test: test_rfind_failure_distinguishable_blocks_server_survives.
- Indistinguishable (non-list or single element): degrade to full sanitization — safe, user forgery never leaks. logger.warning on both. Test: test_rfind_failure_indistinguishable_degrade_to_full_sanitization.

Note on the threshold: the condition uses len(content) not text_blocks — text_blocks only includes type:"text" dicts and misses raw strings or non-standard dict blocks (the {type:image, text:"alt"} example), which would cause those blocks to fall into the wrong branch.

4. os.scandir — Single call.

5. .md sibling — Comment documenting the known limitation. A proper fix needs cross-turn source tracking metadata (was this .md user-uploaded or conversion-produced?), which is out of scope for this PR.

Empty-text upload — Now a separate branch: when ORIGINAL_USER_CONTENT_KEY is present but empty string, skip sanitization entirely — server block survives.

Tests: 216 passed across the three affected files, ruff clean, branch up to date with main.

willem-bd (commented):
Review from a focused pass on the lazy-loading change. Overall the design (deferred discovery mirroring the skills pattern) and test coverage are strong — details inline. Main item to resolve before merge: confirm the `runtime.state` visibility assumption with a real-runtime test (comment on `list_uploaded_files_tool.py:90`).

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
```diff
@@ -0,0 +1,221 @@
...
+    paths = _paths or get_paths()
+    uploads_dir = paths.sandbox_uploads_dir(thread_id, user_id=user_id)
+
+    if not uploads_dir.exists():
+        return {"files": [], "message": "No uploads directory for this thread."}
+
+    # Resolve the set of filenames uploaded in the current run so we can exclude them.
+    current_run_filenames: set[str] = set()
+    try:
+        state = runtime.state
```
willem-bd, line 91: **`runtime.state` visibility is the load-bearing assumption here.** This read assumes `UploadsMiddleware.before_agent()`'s `{"uploaded_files": [...]}` state write is visible to the tool at call time (LangGraph applies the middleware state update before the agent node runs tools). `Runtime = ToolRuntime[dict[str, Any], ThreadState]`, so this *should* hold — but it's exactly the surface the PR's `needs-validation` label points at.

The `try/except` + warning is good defensive coding: if the read fails, current-run files may get duplicated in the output (appearing here *and* in `<current_uploads>`), which is a UX nuisance rather than a correctness/security bug.

The bridge test in `test_list_uploaded_files_tool.py` only verifies **format** compatibility — it manually sets `rt.state = {...}` rather than running the real middleware→tool chain through `ToolRuntime`. Consider adding a test that exercises actual LangGraph state propagation (middleware write → tool read in one run) to confirm the exclusion works end-to-end before merge.

backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py
```diff
@@ -297,10 +301,85 @@ def _process_request(self, request: ModelRequest) -> ModelRequest:
...
+                                elif isinstance(block, dict) and block.get("type") == "text" and isinstance(block.get("text"), str):
+                                    sanitized = neutralize_untrusted_tags(block["text"])
+                                    if sanitized != block["text"]:
+                                        new_content.append({**block, "text": sanitized})
+                                    else:
+                                        new_content.append(block)
+                                else:
+                                    new_content.append(block)
+                            # Guard ORIGINAL_USER_CONTENT_KEY (shared tail)
+                            if not isinstance(original_user_content, str) and ORIGINAL_USER_CONTENT_KEY in preserved_kwargs:
```
willem-bd: This guard is unreachable. We're inside `if isinstance(original_user_content, str) and original_user_content:` → `if processed_user != original_user_content:` → `else:` (rfind failed) → `if isinstance(content, list) and len(content) >= 2:`, so `original_user_content` is guaranteed to be a non-empty `str`. `not isinstance(original_user_content, str)` is therefore always `False` and the assignment never runs.

Leaving `ORIGINAL_USER_CONTENT_KEY` unchanged here is actually *correct* (it should hold the user's original text, not the sanitized rebuild), so this looks like dead code copied from the lower repair block. Suggest deleting the `if` (and the `# Guard ...` comment above it), or replacing it with a one-line note explaining why the key is intentionally left as-is.

backend/packages/harness/deerflow/tools/tools.py
```diff
@@ -16,6 +16,7 @@
     present_file_tool,
     ask_clarification_tool,
     review_skill_package,
+    list_uploaded_files,
```
willem-bd: Adding this to `BUILTIN_TOOLS` makes `list_uploaded_files` available to **every** agent, including subagents — `get_available_tools` does `builtin_tools = BUILTIN_TOOLS.copy()` unconditionally and only *extends* with `SUBAGENT_TOOLS` when `subagent_enabled`. I don't see a downstream filter stripping builtins for subagents in this path.

This seems to contradict the PR body's follow-up #4214 ("Consider adding `list_uploaded_files` to subagent tools"), which reads as if subagents don't have it yet. Could you clarify? Either subagents *do* get it (then #4214 is stale and the description should say so), or they *shouldn't* (then this belongs outside `BUILTIN_TOOLS`, gated like `task_tool`).

backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py
```diff
@@ -0,0 +1,221 @@
...
+
+    total_count = len(candidates)
+    truncated = total_count > max_results
+    visible = candidates[:max_results]
+    omitted_paths = [p.name for _, p in candidates[max_results:]]
+
+    files: list[dict] = []
+    for _, file_path in visible:
+        filename = file_path.name
+        stat = file_path.stat()
```
willem-bd: Minor: `stat()` is called twice per visible file — once in the `os.scandir` collection loop (`entry.stat()` at line 134, for `st_mtime`) and again here (for `st_size`). `DirEntry.stat()` results are cheap to carry through; storing `(mtime, path, stat_result)` in `candidates` avoids the second syscall per file and closes the small TOCTOU window between the two calls.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/message_processing.py
```diff
@@ -130,7 +130,7 @@ def filter_messages_for_memory(messages: list[Any], *, should_keep_hidden_messag
                 if not keep:
                     continue
             content_str = extract_message_text(msg)
-            if "<uploaded_files>" in content_str:
+            if "<uploaded_files>" in content_str or "<current_uploads>" in content_str:
```
willem-bd: Minor inconsistency: this `in` pre-check is case-sensitive, but the `_UPLOAD_BLOCK_RE` strip just below it is `re.IGNORECASE`. So `<CURRENT_UPLOADS>` (any non-lowercase casing) would be stripped by the regex but wouldn't enter this branch. The middleware always emits lowercase, so it's fine in practice — but the guard and the regex disagree on case. Either drop the pre-check (let the regex decide) or make the `in` check case-insensitive for consistency.

Leeminjing:
Thanks for the thorough review. All five items addressed in the latest push:

1. Dead code — Removed the unreachable ORIGINAL_USER_CONTENT_KEY guard.
2.  3. Subagent gating — list_uploaded_files moved out of BUILTIN_TOOLS; get_available_tools() takes include_upload_tool (default True), task_tool.py passes False. Subagents no longer see the tool.
3. State visibility — Added integration test through real create_agent graph (no mocked runtime.state). Test passes.
4. Double stat() — Cached st_size in candidates tuple.
5. Case inconsistency — content_str.lower() for the pre-check.

Also fixed the IM channel bug fancyboi999 found: _human_input_message() now passes additional_kwargs.files so IM uploads are correctly excluded by list_uploaded_files. Tag unification to <current_uploads> is left for a separate PR.

Tests: 219 across affected files, 7480 full suite.

fancyboi999 (changes requested):
Thanks @Leeminjing. One current-head correctness regression remains.


Reviewed base SHA: `b3a0dac8ad191fbfa248c0a1cecc8f2deeaa1c7d`
Reviewed head SHA: `8d86dbf6de768765f717346549660655704e80ef`

### [P2] Avoid injecting the same IM upload metadata twice

Location: `backend/app/channels/manager.py:1593-1594`

This path first prepends `_format_uploaded_files_block(uploaded)` to `msg.text`, then passes the same `uploaded` records through `files=`. `UploadsMiddleware.before_agent()` consumes that metadata and prepends a second `<current_uploads>` block to content that already contains the legacy `<uploaded_files>` block. Every inbound IM attachment therefore reaches the model twice, under two authority tags; the original-content sanitization path preserves both server-side prefixes, so it does not collapse them later.

Please make the channel path use one upload-context producer: either stop manually prepending the legacy block when `files` is handed to the middleware, or keep the legacy path without passing the same records through `files`. Add an integration test that builds the channel `HumanMessage`, runs it through `UploadsMiddleware`, and asserts that one attachment produces exactly one upload block and one filename occurrence.

Leeminjing:
@fancyboi999 Thanks for catching this.

Went with option A — removed the manual <uploaded_files> prepend and the now-dead _format_uploaded_files_block() function. UploadsMiddleware is the sole upload-context producer for both IM and web paths.

Added test_channel_message_single_upload_block_via_middleware — constructs a channel-style HumanMessage (content == original_text, no ORIGINAL_USER_CONTENT_KEY, files in additional_kwargs), runs through UploadsMiddleware, asserts exactly one <current_uploads> block, no <uploaded_files> block, and ORIGINAL_USER_CONTENT_KEY backfill by the middleware.

fancyboi999 (changes requested):
Thanks @Leeminjing. I found one issue that should be addressed before this is ready.


Reviewed base SHA: `7df44f586cd5fbbb8200849cf480a047422b4515`
Reviewed head SHA: `953f57144545f0465f2e1fa1788e46c854521225`

### [P2] Remove the temporary issue-body artifact

- Location: `tmp_issue_body.txt:1`
- Problem: the current head adds a repository-root scratch file containing a copy of issue #4212's body. It is not used by DeerFlow, tests, or documentation, so merging this head would permanently ship a one-off maintainer artifact.
- Evidence: commit `953f57144545` is titled as an issue-body update, but its only repository change is adding `tmp_issue_body.txt`; the issue body itself is already maintained on GitHub.
- Suggested fix: remove `tmp_issue_body.txt` from the branch and keep issue-body edits on issue #4212 rather than in the source tree.
- Test: verify `gh pr diff 4174 --name-only` no longer lists `tmp_issue_body.txt`; the existing upload/channel regression checks should remain unchanged.

Leeminjing:
@fancyboi999 Thanks for catching this — removed tmp_issue_body.txt in eb7bbbd. The commit only removes the temporary scratch file; no functional code or regression tests were changed.

fancyboi999 (changes requested):
Thanks @Leeminjing. I found one security boundary issue that should be addressed before this is ready.


Reviewed head SHA: `eb7bbbd5f1fd3f7b43ada6f49dc739e725dc3062`

[P1] Neutralize upload-derived fields before preserving the trusted wrapper

- Location: `backend/packages/harness/deerflow/agents/middlewares/input_sanitization_middleware.py:304-321`; `backend/packages/harness/deerflow/agents/middlewares/uploads_middleware.py:78-100`
- Problem: The new path excludes the entire server-prepended `<current_uploads>` prefix from blocked-tag neutralization, but that prefix contains user-controlled filenames, paths, document outline titles, and preview text interpolated verbatim. A crafted upload can therefore place blocked authority tags into model context outside the user-input boundary.
- Evidence: `normalize_filename()` only enforces basename, backslash, and byte-length rules, so it does not remove angle brackets; outline titles and previews are read directly from uploaded document content. On `main`, `_check_user_content(text_content)` scans the complete message and neutralizes blocked tags. This diff instead keeps `text_content[:idx]` untouched and sanitizes only `original_user_content`, so tags embedded in those upload-derived fields bypass the guardrail.
- Suggested fix: Preserve only the framework-owned outer `<current_uploads>` pair. Neutralize every user-derived value before formatting the block, or sanitize the block interior while retaining the outer wrapper; apply the same rule to filename/path, outline titles, and preview lines.
- Test: Add middleware-to-model regressions with blocked tags in both an upload filename and a converted Markdown heading/preview. Assert that exactly one framework-owned `<current_uploads>` pair remains unescaped and every upload-derived blocked tag is neutralized.

Leeminjing:
@fancyboi999 Thanks for catching this.

Fixed. _format_file_entry() now runs neutralize_untrusted_tags() on all four user-derived values — filename, path, outline titles, and preview text. Outer <current_uploads> wrapper untouched.

Two regression tests:
- Filename containing <system-reminder> → escaped to &lt; form
- Markdown heading ## Section <system>evil</system> → escaped, wrapper count == 1

fancyboi999 (changes requested):
Thanks @Leeminjing. The latest fix closes the four per-file fields from the prior review, but one user-derived branch still bypasses the same boundary.


Reviewed head SHA: `10d668968a5621e21963f3a819e3570f9589ce1d`

[P1] Neutralize the omitted-file extension summary too

- Location: `backend/packages/harness/deerflow/agents/middlewares/uploads_middleware.py:35-38`, `:142-144`, and `:193-196`
- Problem: files after the 10-item context cap bypass `_format_file_entry()`. Their extensions are derived from the user-controlled filename and interpolated verbatim into the trusted `<current_uploads>` wrapper, so the latest neutralization pass does not cover every upload-derived value.
- Evidence: `_files_from_kwargs()` preserves a basename such as `eleven.<system>` and stores `Path(filename).suffix`; `_format_omitted_file_types()` then emits `1 .<system>` without `neutralize_untrusted_tags()`. With 11 uploads, that blocked tag lands inside the framework-owned wrapper even though the selected ten filenames, paths, outlines, and previews are now escaped.
- Suggested fix: neutralize each extension label, or neutralize the complete omitted-type summary before appending it. Keep the outer `<current_uploads>` pair unchanged.
- Test: build a context with more than `_MAX_FILES_PER_CONTEXT_SECTION` files and put a blocked tag in an omitted file's suffix. Assert there is exactly one unescaped framework wrapper and the omitted summary contains only the escaped tag.

Leeminjing:
@fancyboi999 Thanks again for catching this.

Fixed. _extension_label() now neutralizes the extension before returning — one line, covers both the extension field and the Path(filename).suffix fallback, so _format_omitted_file_types() never sees a raw tag.

Added test_neutralizes_blocked_tags_in_omitted_extension_label.

fancyboi999 (changes requested):
Thanks @Leeminjing. I found one remaining security boundary issue that should be addressed before this is ready.


Reviewed head SHA: `91284f543cf9ecb1c67442361b9716e48d7df25d`

[P1] Neutralize historical upload metadata before returning it to the model

- Location: `backend/packages/harness/deerflow/tools/builtins/list_uploaded_files_tool.py:145-177`
- Problem: The current-upload path now neutralizes user-derived filenames, paths, outlines, previews, and omitted extensions, but the new historical-upload tool returns the same attacker-controlled fields verbatim. Its result becomes a model-visible `ToolMessage`, so a previously uploaded Markdown heading or filename can still inject a blocked framework tag through `list_uploaded_files`.
- Evidence: On this head, a historical `notes.md` whose heading is `## <system-reminder>INJECTED</system-reminder>` is returned by `_list_uploaded_files_impl(include_outline=True)` with that raw title. Passing the resulting ToolMessage through the real `ToolResultSanitizationMiddleware` leaves it unchanged because that middleware only sanitizes `web_fetch`, `web_search`, `image_search`, and `web_capture`; `list_uploaded_files` is not covered. This bypasses the same boundary the latest `UploadsMiddleware` fixes closed for `<current_uploads>`.
- Suggested fix: Apply `neutralize_untrusted_tags()` to every model-visible user-derived value produced by `list_uploaded_files`: filename, virtual path, extension/omitted summary, outline titles, and preview lines. Keep the returned structure and safe numeric fields unchanged.
- Test: Add a real tool-result regression with blocked tags and boundary markers in a historical filename, Markdown heading/preview, and an omitted extension. Assert the resulting ToolMessage contains only neutralized forms and no raw authority tag.

Leeminjing:
Thanks @fancyboi999 . Fixed.

Fix: Applied neutralize_untrusted_tags() to all 6 user-derived fields in _list_uploaded_files_impl(): filename, path, extension, outline[].title, outline_preview[*], plus the omitted summary via _extension_label().

Approach: Neutralized at the data-producer layer (inside the tool), same pattern as the <current_uploads> fixes. Did NOT add list_uploaded_files to _REMOTE_CONTENT_TOOL_NAMES — that set's contract is "remote-content tools" (web_fetch/web_search/etc.), and adding a local-file-metadata tool would conflate categories while leaving the _extension_label inconsistency between this file and uploads_middleware.py unresolved.

Tests: 9 new regression tests (in TestListUploadedFilesNeutralization + standalone):
- blocked tags in filename, extension, outline title, preview text, and omitted extension summary
- boundary markers (--- BEGIN/END USER INPUT ---) in a filename
- safe fields unchanged (size, line number, total_count, JSON round-trip)
- full ToolMessage serialization path with both blocked tags and boundary markers in outline titles

4 filename-extension tests are @pytest.mark.skipif(sys.platform == "win32") because <> are invalid NT path characters. They run on Linux CI where these bytes are valid. All other tests run cross-platform. 34/34 pass on Linux (Docker).

The </system-reminder> closing-tag case contains a literal / which is a path separator on every OS — it can never appear in a single filename component. It IS exercised through .md content (outline titles / preview text), which the outline and preview tests cover.

Leeminjing:
@fancyboi999   @willem-bd   Could you please take another look? I’ve resolved the merge conflict, completed an additional security review, and summarized the results below.`

Merge conflict resolution & security hardening

Round 1 — Merge conflict with upstream main

Detected: message_processing.py conflicted between PR #4174 (dual-tag <uploaded_files>|<current_uploads> backreference) and upstream PR #4286 (YAML-externalized signal patterns).

Resolved:
- Kept our _UPLOAD_BLOCK_RE with named backreference supporting both tags
- Accepted upstream YAML externalization (load_patterns(), message_patterns/*.yaml, prompts/*.yaml)
- Dropped hardcoded _CORRECTION_PATTERNS / _REINFORCEMENT_PATTERNS tuples
- Verified 4 auto-merged files (deer_mem.py, updater.py, prompt.py, config.py) for call-chain consistency

Testing gap found & fixed: test_memory_upload_filtering.py only tested <uploaded_files> blocks. Production emits <current_uploads>. Added 2 test variants for the production tag path.

Round 2 — Trail of Bits differential security review

Detected (6 findings, all LOW/INFORMATIONAL):
1. list_uploaded_files follows symlinks via entry.is_file() — could enumerate outside uploads dir
2. No structural assertion that every string field in tool result is neutralized
3. Omitted-file summary string constructed from individually-neutralized parts but not itself neutralized
4. .md conversion-artifact heuristic may hide legitimate user uploads (known limitation, not fixed)
5. WeChat AES key resolution scans unstructured metadata (pre-existing, not fixed)
6. ViewImageMiddleware TOCTOU gap (accepted as designed)

Fixed (Findings 1–3):
- Added and not entry.is_symlink() guard in _list_uploaded_files_impl() scan loop
- Added test_all_string_fields_in_result_are_neutralized — walks the entire result dict tree, asserts no raw blocked tag in any str leaf
- Wrapped _format_omitted_file_types() and _format_omitted_summary() return values in neutralize_untrusted_tags()
- Added denylist maintenance doc comment on _BLOCKED_TAG_NAMES
- Added test_list_uploaded_files_skips_symlinks

Round 3 — Semgrep static analysis
Passed. No findings were detected in PR-changed files.

WillemJiang:
@Leeminjing, please take a look at the failed unit tests.

Leeminjing:
@WillemJiang Thanks for catching the CI failure. Apologies for missing this — I updated the get_available_tools() call in task_tool.py to pass include_upload_tool=False but forgot to update the 5 existing mock assertions in test_task_tool_core_logic.py to expect the new kwarg.

Fixed in ceff6ce — each assert_called_once_with now includes include_upload_tool=False.

Also ran the full backend suite in Docker with a bind mount of the live working tree — 8778 passed, 3 remaining failures are purely Docker dev environment quirks (hostname mapping, UV_EXTRAS resolution) that don't exist on the CI runner. Zero regressions from this fix.

Could you please re-run the checks and continue the review process when convenient？

WillemJiang (approved):

PR #4175, opened 2026-07-14
ci(helm): publish chart to charts/ namespace prefix on GHCR

zhfeng:
## What

Publishes the Helm chart to `ghcr.io/<owner>/charts/deer-flow` instead of the bare `ghcr.io/<owner>/deer-flow` package, by adding a `charts/` prefix to the `helm push` target in both `chart.yaml` (release) and `nightly.yaml`. Also refreshes the chart release docs.

## Why

The chart previously landed at the bare `deer-flow` GHCR package, which is ambiguous next to the three image packages (`deer-flow-{backend,frontend,provisioner}`) and permanently claims the bare `deer-flow` name. Namespacing it under `charts/` distinguishes it **without renaming the chart**: `Chart.yaml` `name` stays `deer-flow`, so dir = chart name = in-cluster resource/selector names, and no `nameOverride` hack (with its immutable-selector footgun) is needed.

The chart is new in 2.1.0 (chart infra landed in #3987, after v2.0.0), so **2.1.0 is the first chart release** - the cheapest moment to make this change. Early nightly builds remain at the legacy non-prefixed `ghcr.io/<owner>/deer-flow`.

GHCR package names are independent of repos (the existing image packages have no matching repos either), so no `bytedance/charts` repo is needed - the `charts/deer-flow` package is auto-created on first push and links to `bytedance/deer-flow` via the `GITHUB_TOKEN`.

## Changes

- `.github/workflows/chart.yaml`, `nightly.yaml`: `helm push … oci://ghcr.io/$owner` -> `…/$owner/charts` (helm appends the chart name `deer-flow` -> `charts/deer-flow`). Header comments updated.
- `RELEASING.md`: retarget the chart install URL to `charts/deer-flow` and the illustrative version `2.2.0` -> `2.1.0`.
- `deploy/helm/deer-flow/README.md`: install URL + a note that the chart is new in 2.1.0; replace the removed `scripts/build-and-push.sh` with raw `docker build`/`push` commands (contexts/args match `container.yaml`); helm prerequisite `3` -> `3+`.
- `deploy/helm/deer-flow/templates/NOTES.txt`: empty-registry warning points at the README section instead of the removed script.

`Chart.yaml` and `values.yaml` are unchanged.

## Verification

- `helm lint` + `helm template` pass; `check_config_version.sh` passes (24==24).
- Both workflow files parse as valid YAML.
- No `build-and-push.sh` references remain; all three Dockerfiles referenced by the new build commands exist.
- First publish to `charts/deer-flow` happens on the next nightly run - a safe canary before the `v2.1.0` tag.

## Notes for release

The new `charts/deer-flow` package defaults to **private** on first publish (same as the prior chart package). Flip it public once in its GHCR settings for unauthenticated `helm pull`, or use an image pull secret.

### DIFF
--- a/.github/workflows/chart.yaml
+++ b/.github/workflows/chart.yaml
@@ -9,7 +9,7 @@ name: Publish Helm Chart
 # stale chart fails the PR, not the release.
 #
 # Users then install with:
-#   helm install deer-flow oci://ghcr.io/${{ owner }}/deer-flow --version <ver>
+#   helm install deer-flow oci://ghcr.io/${{ owner }}/charts/deer-flow --version <ver>
 
 on:
   push:
@@ -83,5 +83,5 @@ jobs:
         run: |
           for pkg in ./packages/*.tgz; do
             echo "--- pushing $pkg"
-            helm push "$pkg" oci://ghcr.io/${{ github.repository_owner }}
+            helm push "$pkg" oci://ghcr.io/${{ github.repository_owner }}/charts
           done
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -33,40 +33,40 @@ distinguishes it from a release.
 ## Helper scripts
 
 - `scripts/bump_version.sh <version>` — set all four fields at once, then
-  self-verify. Tolerates a leading `v` (e.g. `v2.2.0`).
+  self-verify. Tolerates a leading `v` (e.g. `v2.1.0`).
   ```bash
-  scripts/bump_version.sh 2.2.0
+  scripts/bump_version.sh 2.1.0
   ```
 - `scripts/verify_versions.sh [version]` — check that all sources agree. With
   no argument it requires mutual equality; with an argument it requires every
   source to equal it. Exits non-zero on mismatch. Run it locally before tagging
   to catch drift early:
   ```bash
-  scripts/verify_versions.sh 2.2.0
+  scripts/verify_versions.sh 2.1.0
   ```
 
 ## Release procedure
 
 1. **Bump the version** across all sources:
    ```bash
-   scripts/bump_version.sh 2.2.0
+   scripts/bump_version.sh 2.1.0
    ```
 2. **Update `CHANGELOG.md`**: rename the `## [Unreleased]` section to
-   `## [2.2.0] — YYYY-MM-DD` (note the em dash `—`), and add a link reference
+   `## [2.1.0] — YYYY-MM-DD` (note the em dash `—`), and add a link reference
    at the bottom of the file:
    ```
-   [2.2.0]: https://github.com/bytedance/deer-flow/releases/tag/v2.2.0
+   [2.1.0]: https://github.com/bytedance/deer-flow/releases/tag/v2.1.0
    ```
    Start a fresh `## [Unreleased]` section above it for the next cycle.
 3. **Commit** the version + changelog changes:
    ```bash
    git add -A
-   git commit -m "release: v2.2.0"
+   git commit -m "release: v2.1.0"
    ```
 4. **Tag and push**:
    ```bash
-   git tag v2.2.0
-   git push origin v2.2.0
+   git tag v2.1.0
+   git push origin v2.1.0
    ```
    Pushing the tag triggers the publishing workflows (below).
 
@@ -78,7 +78,7 @@ distinguishes it from a release.
 - `.github/workflows/chart.yaml` — packages the Helm chart and pushes it as an
   OCI artifact to `ghcr.io`. Users install with:
   ```bash
-  helm install deer-flow oci://ghcr.io/<owner>/deer-flow --version 2.2.0
+  helm install deer-flow oci://ghcr.io/<owner>/charts/deer-flow --version 2.1.0
   ```
 
 ## Nightly builds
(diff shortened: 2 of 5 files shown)


deploy/helm/deer-flow/README.md
```diff
@@ -11,7 +11,7 @@ Kubernetes resources. No existing repo files are modified.
 ## Prerequisites
 
 - A Kubernetes cluster (Docker Desktop K8s, OrbStack, kind, k3d, or a real cluster).
-- `kubectl` + `helm` 3 installed.
+- `kubectl` + `helm` 3+ installed.
```
fancyboi999: [P2] Require Helm 3.8+ for the documented OCI flow

This prerequisite currently says every Helm 3 release works, but OCI support is only enabled by default starting in 3.8.0; 3.0-3.7 require `HELM_EXPERIMENTAL_OCI=1` and have version-specific behavior. As written, a user with (for example) Helm 3.7 satisfies the prerequisite and then the `helm install oci://...` command below fails in the default setup. Since the README does not document the experimental path, make this `helm 3.8+` (or explicitly document the older setup). Helm reference: https://helm.sh/docs/v3/topics/registries/#oci-support-prior-to-v380

fancyboi999 (changes requested):
One installation-compatibility issue needs to be corrected before the new GHCR instructions are reliable.

.github/workflows/chart.yaml
```diff
@@ -83,5 +83,5 @@ jobs:
         run: |
           for pkg in ./packages/*.tgz; do
             echo "--- pushing $pkg"
-            helm push "$pkg" oci://ghcr.io/${{ github.repository_owner }}
+            helm push "$pkg" oci://ghcr.io/${{ github.repository_owner }}/charts
```
willem-bd, line 86: Moving the push target to `charts/` is correct - helm appends the chart name `deer-flow` from `Chart.yaml` -> `charts/deer-flow`, and this job already has `packages: write` (chart.yaml:67-69), so the namespaced package auto-creates on first push with no new perms. Two follow-up considerations on the path change:

- **Orphaned bare `deer-flow` package.** Early nightly charts already pushed to the bare `ghcr.io/<owner>/deer-flow`; after this PR that package receives no new versions and goes stale. Anyone pulling from the bare path - especially without a pinned `--version` - silently gets the last old nightly. The chart is new in 2.1.0 so the audience is small, but consider deleting or deprecating the old bare package and calling out the path change in the 2.1.0 release notes. The PR body notes early nightlies "remain at the legacy" path but doesn't address cleanup.

- **First-publish verification is the nightly canary, not pre-merge.** There's no dry-run of `helm push` to the new namespace in CI, so the first proof the `charts/` push works is the next nightly run (the chosen canary before `v2.1.0`). Reasonable and acknowledged; risk is low since `packages: write` is already proven for the bare package and namespacing doesn't change the auth model.

RELEASING.md
```diff
@@ -122,20 +122,20 @@ When it fails, the job annotation names the offending file and suggests the
 fix:
 
 ```
-::error::frontend/package.json is '2.1.0' but expected '2.2.0'.
-Tip: run scripts/bump_version.sh 2.2.0 to align all sources.
+::error::frontend/package.json is '2.0.0' but expected '2.1.0'.
```
willem-bd, line 125: Minor: the repo is currently at `2.1.0` (both `frontend/package.json` and `Chart.yaml` read `2.1.0`), so this example's "actual" value (`2.0.0`) now sits below the real current version - it illustrates a pre-bump state that no longer matches the tree. The original (`is '2.1.0' but expected '2.2.0'`) tracked the real repo state more closely. Purely illustrative and harmless, but optional to realign the "actual" with the current dev version.

willem-bd (commented):
Reviewed the chart `charts/` namespace change. Verified against the repo: `helm push .../charts` correctly yields `charts/deer-flow` (chart name `deer-flow` from `Chart.yaml`), both push jobs already have `packages: write` (no new perms needed), the README's local `docker build` commands match `container.yaml` contexts/args, image names match `values.yaml` defaults (`gatewayImage`/`frontendImage`/`provisionerImage`), `build-and-push.sh` is genuinely gone so dropping its references is correct, and the PR head leaves no stale bare-path or build-and-push references. The `helm 3` -> `3+` bump is a real improvement.

Approve - CI/docs-only, low risk, conventional `charts/<name>` layout. Inline notes below: an orphaned-bare-package cleanup follow-up + the nightly-canary verification caveat on the push target, a minor version-example realignment in RELEASING.md, and a `3.8+` precision nit on the helm prerequisite.

deploy/helm/deer-flow/README.md
willem-bd: Nit: OCI registry support (`helm push oci://...`, `helm install oci://...`) stabilized in Helm 3.8 - it was experimental (and required `HELM_EXPERIMENTAL_OCI=1`) before 3.8. `3+` is loose enough that a 3.0-3.7 install would satisfy this prerequisite but fail at push/install time. `3.8+` would be precise. Optional.

zhfeng: Done — bumped the prerequisite to `helm 3.8+` and noted that 3.0–3.7 need `HELM_EXPERIMENTAL_OCI=1` (9e63793).

.github/workflows/chart.yaml
zhfeng: - **Orphaned package:** added a post-release note to `RELEASING.md` to delete/revoke the legacy bare `ghcr.io/<owner>/deer-flow` package once nothing pulls from it (9e63793). I'll also surface the path change in the 2.1.0 release notes.
- **Canary:** agreed — the next nightly run is the first real proof, and risk is low since `packages: write` is unchanged by namespacing. Acknowledged, no change.

RELEASING.md
zhfeng: Kept `2.1.0` intentionally — it's the next *unreleased* version (first chart release; the tree is pre-bumped but there's no `v2.1.0` tag yet), so the runbook targets the release that's about to ship. The mismatch `is '2.0.0' but expected '2.1.0'` illustrates forgetting to bump from the previous 2.0.0.

You're right that the tree is already at 2.1.0; since 2.1.0 is still unreleased I kept it as the forward-looking target rather than jumping to 2.2.0. Happy to revert to `2.2.0` if you'd prefer the examples to always read one release ahead.

deploy/helm/deer-flow/README.md
zhfeng: Fixed in the same edit — `3.8+` with the `HELM_EXPERIMENTAL_OCI=1` note (9e63793).

WillemJiang (approved):

PR #4177, opened 2026-07-14
docs(api): document stateless /api/runs/stream endpoint

yym36991:
Record that clients can start a conversation without pre-creating a thread,
    and that Gateway returns thread_id and run_id via the Content-Location header.

    Co-authored-by: yym36991@gmail.com

Fixes #

## Surface area

## AI assistance

**Tool(s) used:** 

**How you used it:** 

### DIFF
--- a/backend/docs/API.md
+++ b/backend/docs/API.md
@@ -11,6 +11,11 @@ DeerFlow backend exposes two sets of APIs:
 
 All APIs are accessed through the Nginx reverse proxy at port 2026.
 
+For agent conversations, clients can either pre-create a thread
+(`POST /api/langgraph/threads`) or start immediately with the stateless stream
+endpoint (`POST /api/langgraph/runs/stream`). The latter auto-creates a thread
+and returns `thread_id` and `run_id` in the response `Content-Location` header.
+
 ## LangGraph-compatible API
 
 Base URL: `/api/langgraph`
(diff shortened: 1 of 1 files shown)


backend/docs/API.md
```diff
@@ -629,10 +719,19 @@ from langgraph_sdk import get_client
 
 client = get_client(url="http://localhost:2026/api/langgraph")
 
-# Create thread
-thread = await client.threads.create()
+# Option A: stateless stream — no thread pre-creation
+# Gateway auto-creates a thread; SDK reads thread_id/run_id from Content-Location.
+async for event in client.runs.stream(
```
fancyboi999, line 733: [P2] Capture the generated IDs via `on_run_created`

The Python SDK does not expose `Content-Location` metadata through the yielded SSE events or persist it automatically. With the repo's locked `langgraph-sdk` 0.3.13, header parsing is only attached when `on_run_created` is provided. As written, this example discards the generated `thread_id`, so callers cannot perform the documented next-turn continuation. Please pass a callback (for example, update a metadata dict) and show persisting its `thread_id` before the follow-up call.

fancyboi999 (changes requested):
The stateless endpoint contract matches the Gateway implementation, but the Python example needs to expose the generated identifiers before the documented continuation flow is usable.

backend/docs/API.md
yym36991: Updated the Python example to capture thread_id/run_id via on_run_created and added the follow-up call with config.configurable.thread_id, aligned with the existing my local TS/cURL examples.

WillemJiang (approved):

Issue #4107, opened 2026-07-12
[bug] bash bypasses the disabled-skill gate: cat/grep read a disabled skill's files


AnnaSuSu:
## Problem summary

The disabled-skill gate is enforced only by the structured read tools (`read_file`, `ls`, and — with #4096 — `glob`/`grep`). The `bash` tool has no enabled-state gate, so `cat` / `grep -r` / `find -exec cat` read a **disabled** skill's files whenever bash is executable.

## Affected area(s)

Skills, Sandbox

## What happens

`_is_disabled_skill_path()` (`backend/packages/harness/deerflow/sandbox/tools.py`) is referenced only from the structured read tools. `bash_tool` never consults it, and nothing else filters `/mnt/skills` for bash: the sandbox mounts skills **per category** (`skills/public` → `/mnt/skills/public`), not per skill, so a disabled skill's files are present inside the sandbox regardless of its enabled state.

Verified against `main` with a real `extensions_config.json` (`secret-skill: {enabled: false}`), a real `LocalSandbox` with the `/mnt/skills` path mapping, and the real tools:

```
read_file: Error: Skill 'secret-skill' is disabled. Access to its files is blocked. Enable the skill...
bash cat : SECRET_PROCEDURE = step-1-step-2
bash grep: /mnt/skills/public/secret-skill/SKILL.md:SECRET_PROCEDURE = step-1-step-2
```

`read_file` refuses; `bash` serves the same bytes.

### When is this reachable?

`is_host_bash_allowed()` (`sandbox/security.py`) decides whether bash runs at all:

| Sandbox provider | bash executable? | Bypass reachable? |
|---|---|---|
| `LocalSandboxProvider` (default) | No — host bash disabled by default | **No** |
| `LocalSandboxProvider` + `sandbox.allow_host_bash: true` | Yes | **Yes** (repro above) |
| `AioSandboxProvider` / Boxlite / E2B | Yes — non-local providers are allowed unconditionally | **Yes** — `_get_skills_mounts()` bind-mounts the whole `public` / `custom` / `legacy` category dirs into the container |

So the gap does not affect the out-of-the-box local setup, but it does affect the containerized setups that are recommended for real use — the ones where bash is the primary execution path.

## Expected behavior

Disabling a skill should mean the agent cannot read its files, independent of which tool it reaches for. Today "disabled" is enforced at the tool layer for four tools and not enforced at all for the one tool that can read arbitrary paths.

## Steps to reproduce

[...]

Issue #4030, opened 2026-07-10
feat(provisioner): define the skills-PVC layout contract — three-way skills mount is unsupported in PVC mode (follow-up to #3977 / #3985)


fancyboi999:
## TL;DR

**EN** — After #3985, the provisioner's **hostPath** mode correctly implements the three-way skills layout (`/mnt/skills/{public,custom,legacy}`), but **PVC** mode (`SKILLS_PVC_NAME`) explicitly falls back to a single flat `/mnt/skills` mount with a runtime warning ("three-way skills layout is not supported in PVC mode yet"). Since the harness computes skill container paths unconditionally as `/mnt/skills/{category}/…`, the original #3977 failure mode (`read_file` 404 for per-user and legacy custom skills) still exists in PVC mode — now loudly warned instead of silent, but unresolved. The root blocker is that **the skills-PVC internal layout has never been a defined contract**. This issue tracks defining that contract and implementing the three-way mount for PVC mode.

**中文** — #3985 之后 hostPath 模式已正确实现三路挂载，但 PVC 模式显式回退为单一扁平 `/mnt/skills` 挂载（带运行时 warning）。harness 侧 `Skill.get_container_path()` 无条件产出 `/mnt/skills/{category}/…`，所以 #3977 的 404 在 PVC 模式下依然存在——从"静默失败"变成了"显式警告"，但没有解决。根因是 skills PVC 的内部布局从未被定义为正式契约。本 issue 跟踪：定义该契约 + 在 PVC 模式实现三路挂载。

## Current state (verified on `main` @ 90976426)

- `docker/provisioner/app.py` `_build_volumes()`: when `SKILLS_PVC_NAME` is set, it logs
  `"SKILLS_PVC_NAME is set — three-way skills layout is not supported in PVC mode yet; falling back to single /mnt/skills mount"`
  and mounts the PVC root (optionally scoped by `SKILLS_PVC_SUBPATH_TEMPLATE`) at `/mnt/skills`.
- `backend/packages/harness/deerflow/skills/types.py` `Skill.get_container_path()`: always returns `{base}/{category}/{skill_path}` — the harness has no knowledge of the provisioner's mount mode, so the model is always told to `read_file` from `/mnt/skills/{public,custom,legacy}/…`.
[...]

PR #4178, opened 2026-07-14
fix(sandbox): enforce disabled skills in filesystem views

fancyboi999:
Refs #4107

## Why

Disabled skills were filtered from prompts and skill-aware tools, but sandbox providers still exposed the raw skill directories. An agent could therefore bypass the disabled-skill gate with shell or structured file reads.

The issue discussion converged on stable, enabled-only filesystem projections. @AnnaSuSu also identified the atomic-rename hazard: storage writes replace the source inode, so a hardlink projection must rebuild after every content mutation, not only after enable/disable changes.

## What changed

- Materialize enabled-only views at `skills_view/public` and per-user `skills_view/{custom,legacy}` roots. Files use hardlinks with a copy fallback; category root inodes stay stable for live bind mounts.
- Rebuild under in-process and cross-process locks after writes, supporting-file removals, archive installs, deletes, and public/per-user toggles. Public config RMW reloads disk state inside the lock so stale worker caches cannot overwrite another worker's update.
- Ensure the shared public projection at Gateway boot without scanning historical users; per-user projections self-heal lazily at Local/AIO acquire, including cached sandbox reuse. Public signature/load/staging/storage-initialization failures clear the view and manifest under the same cross-process lock and do not abort Gateway startup.
- Fold legacy visibility and nested-skill boundaries into the projection instead of duplicating provider-side visibility rules.
- Point Local, AIO/Docker, and provisioner hostPath mounts at the stable projections. New E2B sandboxes receive an enabled-only snapshot and conflicting custom `/mnt/skills` uploads are rejected.
- Offload projection-producing Gateway mutations from the async event loop.

Acquire self-healing is **O(files + directories) metadata stat work**, not O(1), and does not read file contents. This corrects the shorthand in the earlier issue comment. On the repository's 104-file public skill tree, a local run measured a 3.25 ms median, 4.02 ms P95, and 4.88 ms max across 50 unchanged acquire checks; cold materialization was about 403 ms.

PVC-backed skill materialization remains unchanged. Existing E2B sandboxes keep their creation-time snapshot; only newly created E2B sandboxes receive the projection.

## Surface area

## Screenshots / Recording

Not applicable; this changes backend filesystem visibility and sandbox mounts.

## Bug fix verification

- Test paths that reproduce the bug and edge cases:
  - `backend/tests/test_skill_projection.py`
  - `backend/tests/test_three_way_skills_mount_e2e.py`
  - `backend/tests/test_skill_manage_tool.py`
  - `backend/tests/test_skills_custom_router.py`
  - provider/provisioner tests under `backend/tests/`
- Red on `main`, green here: yes. Initial tests failed because the projection module did not exist; provider tests then exposed raw mounts. Additional red/green tests cover atomic rename, create/install/remove, nested skills, stale worker caches, concurrent writes/toggles, public-only boot ensure, lazy per-user repair, fail-closed signature/staging/storage-factory errors, and serialization between rebuild and failure cleanup.

## Validation

Merged the latest `main` into the branch (picked up unrelated commits landed since this PR opened; no conflicts).

Static:
- `uv run ruff check app packages tests` — passed
- `uv run ruff format --check app packages tests` — passed (856 files already formatted)
- `uv run python ../scripts/scan_changed_blocking_io.py --base origin/main` — 0 candidates

Follow-up review after merging the latest `main`:
- Final targeted projection/Gateway/provider matrix: `137 passed`; independent verifier reran all 19 projection tests and the concurrency regression 10/10 times.
- `uv run ruff check --no-cache app packages tests` — passed; `uv run ruff format --check --no-cache app packages tests` — 862 files formatted.
- `scan_changed_blocking_io.py --base origin/main` — 0 candidates.
- Real Gateway startup plus `GET /health` — HTTP 200 with `{"status":"healthy","service":"deer-flow-gateway"}`, followed by clean shutdown.
- Independent full backend run: `7724 passed, 25 skipped, 4 failed, 12 warnings`; all four failures reproduced as local config or live-model variability outside this diff (SQLite-vs-memory config, two custom-subagent list assertions, and a live model response without the expected tool call).

Cross-worker enabled-state follow-up (`f58b1187`):
- Added a red/green regression that leaves one worker's singleton enabled while disk state disables both CUSTOM and LEGACY skills; the first ensure removes both projections and writes a manifest, and the second ensure keeps the manifest byte-for-byte stable.
- Related projection/storage/router/client/provider matrix: `366 passed`; projection file: `30 passed`; blocking-I/O suite: `39 passed`; ruff check/format clean across 862 files; changed blocking-I/O scan: 0 candidates.
- A fresh independent verifier reproduced the boundary with two real processes and temporary files: both disabled projections disappeared after the first and second ensure, both manifests stayed stable, and unrelated public/user skills remained visible.
- Unfiltered full backend verification: `7746 passed, 11 skipped, 8 failed, 12 warnings`. Three failures reproduced from local config overrides (SQLite checkpointer plus two custom-subagent-name assertions); five were live LLM/tool-call failures with `127.0.0.1:7890` proxy TLS resets. No failure touched the projection/storage path.
- Real Gateway startup and `GET /health` returned 200, followed by clean channel, heartbeat, persistence, and application shutdown.

Full backend suite (`uv run pytest`, excluding `test_client_live`): `7537 passed` on the first pass. A second full run surfaced network-flaky failures in the live-LLM `test_client_e2e.py` / `test_create_deerflow_agent_live.py` suites (confirmed transient: `test_basic_chat` failed in the full run, passed immediately after in isolation), plus two failures pre-existing and unrelated to this diff — `test_checkpointer.py::test_returns_in_memory_saver_when_not_configured` (untouched code path, no relation to skills) and `test_subagent_prompt_security.py` (asserts on a hardcoded subagent-name list that doesn't account for custom agents defined in a local `config.yaml`, unrelated to this change).

Three real issues were found and fixed during this final review pass, all by running the actual test suite (twice caught only by CI, not locally) rather than re-reading the diff:
1. **Boot projection availability and scalability**: boot originally rebuilt every known user sequentially, making readiness O(users), and storage-factory errors could escape before fail-closed cleanup. Boot now ensures only the shared public projection; user scopes repair lazily on acquire. Public initialization failures clear view and manifest under the projection lock, return without aborting lifespan, and never emit a misleading success log. Regression tests cover lazy user repair, factory failure, and deterministic rebuild/cleanup interleaving.
2. **`test_skills_router_authz.py::test_enable_toggle_allowed_for_admin`** monkeypatched a `get_extensions_config` import that this branch's earlier commits already removed from the skills router (superseded by the fresh-disk-read-under-lock in `_persist_shared_skill_state`, which is the actual point of the stale-worker-cache fix). Updated the test to match.
3. **`acquire()` projection-ensure not best-effort**: `LocalSandboxProvider`/`AioSandboxProvider` both called `_ensure_skills_projection()` directly inside `acquire()`, outside any try/except — unlike every other skill-mount setup path in these providers, which has always caught exceptions and logged rather than failing acquire outright (e.g. when `config.yaml` can't be resolved). CI's environment has no `config.yaml`, which exposed this; my local runs didn't catch it because a real `config.yaml` was present, masking the failure. `_ensure_skills_projection` now swallows its own exceptions and returns `None`; callers already tolerate that by skipping the skill-specific mounts for that acquire (made explicitly None-safe where they weren't).

Real end-to-end verification (no mocks) beyond the test suite: wrote a script exercising the actual `UserScopedSkillStorage` + `LocalSandboxProvider` + `LocalSandbox` against the running Gateway's real data directory with `allow_host_bash: true`, driving genuine subprocess `bash cat` calls against a real created skill:
- enabled skill → `cat` succeeds, content visible
- disabled skill, **same sandbox instance, no recreation** → `cat` fails (`No such file or directory`) — the exact bug #4107 reports
- re-enabled, same sandbox → `cat` succeeds again, confirming mid-thread enable stays instant
- deleted skill → gone from the projection too

Also manually drove the real browser UI (register/login, real chat, real streaming) against this branch's `make dev` to confirm no regressions in ordinary chat/skill-authoring flows; the model itself declined to touch `/mnt/skills/custom` paths from inside a chat turn as an unrelated model-policy caution (not a technical gap — the tools and mounts work, as the script above proves), so the disabled/enabled bash-visibility assertions were made via the script instead of the chat transcript.

Not run against a real E2B account, Docker/AIO container, or Kubernetes hostPath Pod. Helm rendering was not run because the local environment has no Helm CLI. PVC dynamic projection and hot-updating existing E2B sandboxes are explicitly outside this PR.

## AI assistance

**Tool(s) used:** Claude Code, Codex

**How you used it:** Claude Code captured the issue discussion and created the branch. Codex implemented the change with red/green tests, ran the verification matrix, and used an independent review agent to identify and fix mutation, concurrency, fail-closed, recovery, and async-I/O gaps. Claude Code did a final independent re-review after rebasing onto latest `main`: read every changed file in the diff, found and fixed the two issues described above, and ran the real end-to-end sandbox verification described above.


AnnaSuSu:
Reviewed the projection layer as promised. Checked out the branch and ran the code rather than reading it only — repro scripts and numbers below.

**The design is right and the two hazards I raised are genuinely closed.** Specifically, I verified:

- **Trigger completeness.** Every content mutation in the storage layer is inside the rebuild hook: `write_custom_skill` (both `local_skill_storage.py:106` and `user_scoped_skill_storage.py:357`), `remove_custom_skill_file` (`:114`), `_commit_skill_install` (`:207`, which the user-scoped install also routes through), `delete_custom_skill` (`:230`), `set_skill_enabled_state` (`user_scoped_skill_storage.py:155`), plus the public RMW in the router (`routers/skills.py:114`) and the client (`client.py:1188`). I went looking for a mutation primitive outside the hook and didn't find one — the rollback route writes through `write_custom_skill`, so it's covered transitively. This is the part I was most worried about and it holds.
- **The stale-worker-cache claim.** `ExtensionsConfig.from_file()` reads the file on every call — no memoization — so `_source_signature` and `_load_public_skills` both see disk, not a worker's cached singleton. The claim in the PR body is accurate.
- **Blocking IO.** `ensure_skill_projections` is a full-tree walk, so I expected to find it on an event loop. It isn't: AIO offloads it in `_acquire_internal_async` (`aio_sandbox_provider.py:782`) while the bare call at `:758` is on the sync `_acquire_internal`; E2B's `acquire_async` wraps the whole sync `acquire` in `to_thread`; Local doesn't override the base `acquire_async`, which offloads. Clean.
- **The legacy fold.** `user_should_see_legacy_skills` has no callers left — the rule now lives only in `_iter_skill_files`'s `if not user_custom_exists`, and the providers unconditionally mount a possibly-empty legacy root. That's the collapse to one site I asked for. (It is now dead code but still exported from `skills/storage/__init__.py:146` — worth deleting, or the next provider will call it and re-fork the rule.)

Three findings. The first two are the same root cause and I think #1 is a blocker; #3 is a scaling issue that your own benchmark can't see.

---

## 1. Rebuilding blanks the live view, so enabled skills transiently vanish

`_replace_category` (`projection.py:119-136`) clears the category root **before** staging:

```python
_clear_category(root)          # the live mount is now EMPTY
with tempfile.TemporaryDirectory(...) as staging_dir:
    for relative_path, skill in desired.items():
        _stage_skill(...)      # ... for the whole duration of this
    for path in staging.iterdir():
        path.replace(root / path.name)
```

and `skill_projection_mutation` (`:388-399`) clears the roots *before* yielding to the mutation, so the window covers the mutation **and** the rebuild.

Because the category root inode is stable and bind-mounted into live sandboxes — the property the whole design is built on — every concurrent reader sees that empty window. An enabled, untouched skill disappears from `/mnt/skills` and comes back.

**Repro A** — user has two enabled custom skills, `alpha` and `beta`. A live sandbox `cat`s `alpha/SKILL.md` in a loop. The agent edits **`beta`** via one ordinary `skill_manage_tool` write:

```
write_custom_skill('beta') took 38.2 ms
reads of alpha/SKILL.md during the window: ok=159  ENOENT=50
```

`alpha` was never touched. It vanished for 50 consecutive reads.

**Repro B** — the repo's real public tree (23 skills / 104 files, the same tree you benchmarked). A live sandbox reads `deep-research/SKILL.md`; an admin disables the unrelated `skill-creator` via the router's RMW:

```
disable('skill-creator') took 76.9 ms end to end
reads of deep-research/SKILL.md: ok=170  ENOENT=98
observed blank window on the enabled, untouched skill: ~66.0 ms
```

Disabling one public skill blanks the public view — all 22 others — for **every running sandbox** of **every user**. That's on a same-filesystem hardlink farm. Where `_link_or_copy` hits `EXDEV` and falls back to `shutil.copy2` (a separate mount for `.deer-flow/` is the normal Docker/K8s shape), the window becomes the full copy cost — your own cold-materialization figure, ~403 ms.

On `main` this can't happen: the mount is the raw category dir and the write is a single atomic rename inside `beta/`, so `alpha` is never touched. This is a new failure mode introduced by the projection.

The comment at `:122-124` says the clear-first order is deliberate — *"Clear before staging so allocation/copy failures cannot leave a stale enabled view behind."* That argument is sound for a **disable or delete**, where a skill must lose visibility and a crash mid-rebuild would otherwise leave it readable by bash (and the #3889/#4096 tool gates don't cover bash, which is the entire point of #4107 — so the gate underneath genuinely doesn't backstop this one).

But the same blanket clear is applied to **`write_custom_skill` and archive install**, where *nothing loses visibility*. There is no fail-closed argument there at all — it's pure cost, and it's the path the agent itself hits on every skill edit. Repro A is exactly that case.

Suggestion, keeping fail-closed intact:
- In `_replace_category`, delete the entries that must **disappear** first (the security-critical delta), then stage and per-entry `replace()` the rest. Per-entry rename is atomic, so enabled skills stay continuously visible and a staging failure still can't resurrect a disabled skill.
- In `skill_projection_mutation`, drop the pre-clear for purely additive/content mutations and rely on the manifest unlink (already there) plus rebuild-after. Keep the pre-clear only where visibility is being removed.

## 2. Same window, but it also lands on the agent's own read-back

Worth calling out separately because it's the case your integration tests cover. The tests assert *agent edits a skill → reads back new content*, which passes: the rebuild completes under the lock before `write_custom_skill` returns, so a **serial** read-back is correct.

What they don't cover is a *concurrent* reader — a second thread of the same user, or any other user during a public toggle. That's Repro A/B. A test that pins this would be: hold a reader on skill X, mutate skill Y, assert X was never unreadable.

## 3. `ensure_skill_projections` holds a global exclusive lock on the per-run hot path

`ensure_skill_projections` (`projection.py:338`) takes the cross-process public lock **unconditionally**, before it knows whether anything is stale:

```python
with _projection_lock(paths.public.parent):
    manifest = _read_manifest(paths.public.parent)
    signature = _source_signature(storage, "public")   # full scandir of the public tree
    if ...stale...: _rebuild_public_locked(...)
```

And it is on the path of **every acquire**, before the cache hit — `local_sandbox_provider.py:363` runs ahead of the LRU lookup at `:366`, and `aio_sandbox_provider.py:758` runs ahead of `_reuse_in_process_sandbox`. So every agent turn, for every user, takes one global exclusive lock and stats the whole public tree, even though ~always nothing has changed.

Your benchmark measures this correctly but serially. I reproduce your number exactly — single no-op check, 104-file tree: **median 3.19 ms, max 4.88 ms** (you reported 3.25 / 4.88). Then 16 concurrent acquires, **16 distinct users**, nothing stale:

```
wall clock          : 50.6 ms
per-acquire median  : 27.8 ms
per-acquire max     : 48.9 ms
perfectly parallel  : ~3.2 ms
serialization factor: 15.8x
```

Fully serialized — the no-op self-heal check is a global mutex. And it compounds with #1: during a public rebuild the lock is held for the whole re-stage, so every acquire in every worker queues behind it *while* the view is empty.

Suggestion: make the steady-state path lock-free. Read the manifest and compute the signature **outside** the lock; only if they disagree take the lock and re-check under it before rebuilding (standard double-checked locking). The signature reads the source tree and needs no lock; the manifest is already written by atomic rename, so a torn read isn't possible. That turns the common case into N parallel stat walks and keeps the lock for actual rebuilds.

---

## Question, not a finding

The PR body says existing E2B sandboxes keep their creation-time snapshot, so in a live E2B VM a disabled skill stays readable by `bash cat` until the VM is recreated. That is the exact bypass #4107 describes, narrowed to existing sandboxes. I'm fine with it being out of scope — E2B has no shared host mount, and you said so explicitly. But: should #4107 stay open to track it, or does it get closed with a documented residual gap? As the issue author I'd rather it be recorded than quietly inherited.

Also, the AI-assistance checkbox (*"I've read and understand every line of this change"*) is unchecked. Probably just a draft not filled in — flagging it because it's the one box a reviewer can't check for you.

Repro scripts for A/B/#3 are standalone and I'll paste them into a gist / attach on request.

---

**TL;DR（中文）**：投影层的方向和实现都对，我提的两个坑（rename 脱钩、legacy 三处重复）确实都堵上了——触发点我逐个数过 storage 层所有变异原语，没有漏网的；`ExtensionsConfig.from_file()` 确实每次读盘，跨 worker 陈旧缓存那条声明成立；`ensure_skill_projections` 也没有跑在 event loop 上（三个 provider 都正确 offload）。三个发现：**(1) 重建是「先清空后 staging」，而分类根 inode 是稳定的、被活沙箱 bind-mount 着——于是重建期间整个视图是空的**：agent 改 skill B，另一个线程读**没被碰过的、启用中的** skill A 会连续 50 次 ENOENT；管理员禁用一个 public skill，**所有用户**的 public 视图空窗 ~66ms（跨文件系统走 copy 回退时是你自己测的 ~403ms）。清空优先对 disable/delete 是有意义的 fail-closed（bash 不被 #3889 工具门禁覆盖），但对 write/install 没有任何 fail-closed 意义——纯成本。建议：`_replace_category` 先删「该消失的」，再 stage + 逐条原子 replace。(2) 集成测试只覆盖串行读回，并发读者没测。(3) `ensure_skill_projections` 在每次 acquire（缓存命中之前）无条件抢全局跨进程排他锁再算全树签名：单次 3.19ms 中位数（和你报的 3.25ms 一致），但 16 个不同用户并发时 **15.8× 完全串行化**。建议把稳态检查挪到锁外，双重检查后再进锁重建。另：`user_should_see_legacy_skills` 已成死代码但仍在 `__all__` 导出，建议删掉，否则下一个 provider 又会调它。最后一个问题：E2B 存量沙箱仍可 `bash cat` 到已禁用的 skill——#4107 是继续开着追踪，还是关闭并记录这个残留缺口？

fancyboi999:
This is ready for review — moved out of draft. Per #4107, @AnnaSuSu, this is the single-author implementation covering both hazards you flagged: the atomic-rename staleness issue (trigger completeness at the storage layer, verified by red/green tests for edit/create/install) and the nondeterministic-disable timing issue (stable projection roots, so disable takes effect on the next content rebuild rather than waiting for LRU eviction or warm-pool churn).

Since opening the draft I rebased onto the latest `main` and did a final independent pass: read through the full diff again, and found + fixed two real issues by actually running the suite rather than re-reading code — a Gateway-boot availability gap (one broken user directory could abort startup for everyone; now isolated per scope) and a stale test from an earlier commit on this branch. Full details and a real (non-mocked) end-to-end verification — a script driving actual subprocess `bash cat` calls through `LocalSandboxProvider` against a live-created skill, disabling it in the same sandbox instance and confirming the read fails without recreation — are in the updated PR description.

@WillemJiang tagging you too in case you'd like a look, given the earlier PVC/provisioner discussion in #4030 that this converges with.

docker/provisioner/app.py
```diff
@@ -299,30 +299,29 @@ def _build_volumes(
...
+            "skills_view",
             "custom",
         )
         volumes.append(
             k8s_client.V1Volume(
                 name="skills-custom",
                 host_path=k8s_client.V1HostPathVolumeSource(
                     path=user_custom_path,
-                    type="DirectoryOrCreate",
+                    type="Directory",
```
willem-bd, line 538: All three skills volumes are now `type="Directory"` (custom changed from `DirectoryOrCreate`). Since the projection dir is materialized by the **gateway**, not the provisioner, a missing `skills_view/.../custom` now fails Pod creation rather than silently mounting empty.

Two interactions worth confirming:
- **Boot-failed scope**: boot rebuild now fails-isolated per user, so a user whose boot rebuild failed may have no `skills_view/custom` dir. `Directory` (strict) fails their Pod, where `DirectoryOrCreate` would have mounted empty. The AIO path calls `_ensure_skills_projection(user_id)` before creating the sandbox, which should create the dir first — can you confirm that ordering holds for the provisioner path too (the gateway always materializes the user projection on the node before the provisioner mounts it)?
- **Multi-node hostPath**: hostPath is node-local; if the gateway materializes on node A and the Pod lands on node B, this turns a silent empty-mount into a hard Pod-creation failure. Fine if single-node shared-`base_dir` is the assumed topology — just want it confirmed or documented.

backend/packages/harness/deerflow/skills/projection.py
```diff
@@ -0,0 +1,444 @@
...
+            relative = relative_root / entry.name
+            metadata = entry.stat(follow_symlinks=False)
+            if entry.is_symlink():
+                kind = "link"
+            elif entry.is_dir(follow_symlinks=False):
+                kind = "dir"
+                child_dirs.append((Path(entry.path), relative))
+            else:
+                kind = "file"
+            digest.update((f"{label}:{relative.as_posix()}:{kind}:{metadata.st_ino}:{metadata.st_mode}:{metadata.st_size}:{metadata.st_mtime_ns}\0").encode())
```
willem-bd, line 246: The source signature hashes metadata (`st_ino/st_mode/st_size/st_mtime_ns`) per entry, not contents. Runtime writes are covered (the mutation path rebuilds under lock, and atomic-rename changes the inode), but an **external** edit that preserves size+mtime+inode would be invisible to `ensure_skill_projections` and leave the projection stale until the next explicit rebuild.

Low probability, but for a security boundary it's worth a one-line comment here noting the trade-off (metadata-only for speed vs. content-aware hashing) so a future maintainer doesn't assume byte-level integrity.

backend/packages/harness/deerflow/sandbox/local/local_sandbox_provider.py
```diff
@@ -340,13 +368,17 @@ def acquire(self, thread_id: str | None = None, *, user_id: str | None = None) -
...
                 if self._generic_sandbox is None:
-                    self._generic_sandbox = LocalSandbox("local", path_mappings=list(self._path_mappings))
+                    mappings = list(self._path_mappings)
+                    self._append_public_skill_mapping(mappings, skill_projection)
+                    self._generic_sandbox = LocalSandbox("local", path_mappings=mappings)
                     _singleton = self._generic_sandbox
                 return self._generic_sandbox.id
 
         effective_user_id = self._effective_acquire_user_id(user_id)
+        skill_projection = self._ensure_skills_projection(effective_user_id)
```
willem-bd, line 393: This runs on every acquire (including cache hits) to self-heal — fine at the measured ~3-4 ms metadata walk. The spike case: if the manifest is stale here (e.g. another worker just mutated skills), this acquire triggers the ~400 ms cold materialization **under the projection lock**, serializing all concurrent acquires + mutations for that user.

Acceptable for an editing-frequency event, but a short comment noting "intentional self-heal; worst case ~400 ms rebuild under lock if a concurrent edit left the manifest stale" would set expectations for anyone profiling acquire latency.

backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox_provider.py
```diff
@@ -670,11 +670,28 @@ def _bootstrap_sandbox_paths(self, client: E2BClientSandbox) -> None:
...
-        if not mounts:
-            return
-        for mount in mounts:
+    def _skill_projection_mounts(self, user_id: str) -> list[tuple[Path, str, bool]]:
+        from deerflow.skills.projection import ensure_skill_projections
+        from deerflow.skills.storage import get_or_new_user_skill_storage
+
+        config = get_app_config()
+        storage = get_or_new_user_skill_storage(user_id, app_config=config)
+        projection = ensure_skill_projections(storage)
```
willem-bd: Unlike Local/AIO's `_ensure_skills_projection` (which swallows and returns `None`), this calls `ensure_skill_projections` with no guard. A raise propagates out of `_apply_mounts` before the configured-mounts loop runs, so a projection failure drops **configured mounts too** (caught only by `_create_sandbox`'s outer warning).

Still fail-closed for skills, but the blast radius is wider and the pattern inconsistent. Consider wrapping this the way Local/AIO do and uploading only the skill mounts that succeeded.

backend/packages/harness/deerflow/skills/projection.py
```diff
@@ -0,0 +1,444 @@
...
+    else:
+        raise ValueError(f"Unknown skill projection scope: {scope}")
+
+    with _projection_lock(scope_root):
+        for category_root in category_roots:
+            _clear_category(category_root)
+        _manifest_path(scope_root).unlink(missing_ok=True)
+        try:
+            yield
+        except Exception:
```
willem-bd, line 523: Minor: this `try/except: raise` is a no-op — it re-raises without acting. The "keep the view empty" intent is already satisfied by the `_clear_category` + manifest `unlink` before `yield`, so the `except` block can be dropped (just `yield` / `else: rebuild()`) for clarity.

(While here: `rebuild_all_skill_projections` reaches into the private `config.paths._validate_user_id`; a small public helper would be cleaner than importing across module privacy.)

willem-bd (commented):
Reviewed the full diff and verified the load-bearing security invariants against the actual code:

- `_iter_skill_files` only yields `LEGACY` when `not user_custom_exists` -> a user with custom skills gets an empty `projection.legacy`, so mounting it unconditionally is safe (no legacy leak).
- The disable path (`_persist_shared_skill_state` / `set_skill_enabled_state`) clears the projection **before** the state write and rebuilds after -> no window where a just-disabled skill stays readable. Fail-closed.
- On projection failure, providers skip skill mounts entirely rather than degrading to the raw source tree.

The locking (in-process RLock + cross-process flock), self-healing manifest/signature check, and per-scope boot isolation are well-considered and well-tested (fail-closed on staging/signature failure, concurrent writes/toggles, archive-install projection, the exact #4107 mid-thread enable/disable scenario).

The inline comments are refinements + one confirmation, not blockers. The main one is **#1** (provisioner `Directory` hostPath): confirm the gateway always materializes the user projection on the node before the provisioner mounts it — especially for a boot-failed scope or multi-node hostPath, where the old `DirectoryOrCreate` would have silently mounted empty and the new `Directory` fails the Pod. **#4** (E2B error-handling consistency) is a cheap correctness/blast-radius win if you want to take it now.

Minor tracked-cleanup nit (not posted inline): `_build_volumes` / `_build_volume_mounts` retain `include_legacy_skills` only to `del` it for request backward-compat — fine during rollout, worth removing from the request schema + signature later.

**Triage:** Approve pending confirmation on #1.

docker/provisioner/app.py
fancyboi999: Confirmed by tracing the call path, not just inspecting in isolation:

`AioSandboxProvider._acquire_internal` / `_acquire_internal_async` call `self._ensure_skills_projection(user_id)` as their **first statement**, before the fast-path cache check, warm-pool reclaim, and `_discover_or_create_with_lock(_async)` / `_create_sandbox(_async)` — which is what eventually calls `self._backend.create(...)`. `self._backend` is `RemoteSandboxBackend` in provisioner/K8s mode, so every path that reaches the provisioner's `create_sandbox` HTTP call goes through this ensure call first, unconditionally. There's no branch that reaches `_backend.create()` without first going through `_acquire_internal`.

On the directory-existence side: even when the rebuild inside `ensure_skill_projections` fails (boot-failed scope, or any runtime error), `_clear_projection_scope` → `_clear_category` calls `root.mkdir(parents=True, exist_ok=True)` **before** clearing contents — so the category root directory itself always exists (possibly empty) once the ensure call reaches that point, satisfying hostPath `type: Directory`. Combined with the acquire-time self-heal (this PR's design) retrying on every acquire, a boot-failed scope gets a real chance to repair itself before the provisioner is ever called for that user.

So: single-node or NFS-shared `DEER_FLOW_HOST_BASE_DIR` between gateway and K8s node was already this feature's assumption before this PR (the old code mounted `users/{uid}/skills/custom` as hostPath from the same env var) — this PR doesn't introduce that assumption, it just extends it to `skills_view`. What *does* change is that violating it now fails loudly (`Directory` → Pod creation error) instead of silently mounting empty (`DirectoryOrCreate`), which is the right trade-off for a security boundary but is worth being explicit about. Added a note to `docker/provisioner/README.md` (a5f04600) making the single-node/shared-storage requirement and the `Directory`-vs-`DirectoryOrCreate` behavior change explicit, rather than leaving it implicit.

backend/packages/harness/deerflow/skills/projection.py
fancyboi999: Fixed in a5f04600 — added a docstring on `_update_tree_digest` noting the metadata-only-hash trade-off and why runtime writes through this codebase are covered regardless.

backend/packages/harness/deerflow/sandbox/local/local_sandbox_provider.py
fancyboi999: Fixed in a5f04600 — added a comment noting the cost profile (cheap on a fresh manifest, ~400ms rebuild under the cross-process lock on stale/drift, serializing concurrent acquires+mutations for that user).

backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox_provider.py
fancyboi999: Fixed in 88ca8c6e — `_skill_projection_mounts` now catches and returns `[]` instead of raising, matching the Local/AIO pattern. A skills-projection failure no longer drops the operator's configured mounts as collateral. Added `test_skill_projection_mounts_swallows_projection_failure` and an end-to-end `test_apply_mounts_keeps_configured_mounts_when_projection_fails` covering both the unit-level swallow and that a configured mount still uploads when the projection call fails.

backend/packages/harness/deerflow/skills/projection.py
fancyboi999: Fixed in a5f04600 — dropped the no-op `except Exception: raise`; a raise from the mutation now propagates past the `yield` naturally, with the view already left cleared beforehand.

On the `_validate_user_id` private-import note: agreed it's not ideal, but it's consistent with an existing pattern — `user_scoped_skill_storage.py` already reaches across the same boundary the same way, predating this PR. Adding a public helper would mean touching that pre-existing call site too, which felt like scope creep for this PR itself; happy to file it as a follow-up if you'd rather it not sit as-is.

backend/packages/harness/deerflow/skills/projection.py
```diff
@@ -0,0 +1,451 @@
...
+        rebuild_skill_projections(public_storage, include_user=False)
+    except Exception:
+        logger.warning("Failed to rebuild the public skill projection during boot; it stays empty until a sandbox acquire self-heals it", exc_info=True)
+
+    users_dir = get_paths().base_dir / "users"
+    if not users_dir.is_dir():
+        return 0
+
+    rebuilt = 0
+    for user_dir in sorted(users_dir.iterdir(), key=lambda path: path.name):
```
willem-bd: This rebuilds every known user's projection sequentially at boot (~400 ms cold each, per the PR's measurement). For many-user deployments that's an O(N) block on Gateway readiness - and it's a pure optimization, since acquire already self-heals each scope on first use. Consider lazy/per-scope first-acquire or bounding/parallelizing the loop so a large tenant list can't stall startup. The per-scope failure isolation is good; this is only about the serial loop length.

backend/packages/harness/deerflow/skills/projection.py
```diff
@@ -0,0 +1,451 @@
...
+            if fcntl is not None:
+                fcntl.flock(lock_file, fcntl.LOCK_UN)
+            else:  # pragma: no cover - Windows
+                lock_file.seek(0)
+                msvcrt.locking(lock_file.fileno(), msvcrt.LK_UNLCK, 1)
+
+
+def _link_or_copy(source: str, target: str, *, follow_symlinks: bool = True) -> str:
+    try:
+        os.link(source, target, follow_symlinks=follow_symlinks)
```
willem-bd, line 101: Heads-up: projection entries are hardlinks to the source skill files. LocalSandbox.execute_command (host bash) does not enforce PathMapping.read_only - only the structured write_file/read_file tools do. So under allow_host_bash: true, an in-place write through the projection (e.g. echo >> /mnt/skills/public/X/SKILL.md) mutates the shared inode and edits the SOURCE skill. Not a regression (old direct mount had the same write-through), and deletion via the projection is safer (unlink only drops the projection entry). But the coupling is non-obvious - a one-line note here ('projection entries are hardlinks to source; read-only is the mount's job, not the link's') would help future readers.

backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox_provider.py
```diff
@@ -684,6 +714,12 @@ def _apply_mounts(self, client: E2BClientSandbox) -> None:
                 container_path = (mount.get("container_path", "") or "").rstrip("/")
                 read_only = bool(mount.get("read_only", False))
 
+            if container_path == skills_root or container_path.startswith(skills_root + "/"):
```
willem-bd, line 1551: Behavior change worth a release-notes line: any operator-configured E2B mount targeting /mnt/skills or beneath is now silently skipped (only a WARNING). Operators who previously mounted extra content under /mnt/skills/* via E2B config will see it disappear on upgrade. The rejection is correct (prevents shadowing the enabled-only projection), just worth calling out for upgraders alongside the DEER_FLOW_HOST_SKILLS_PATH/SKILLS_HOST_PATH removal.

willem-bd (commented):
Overall this is a strong fix for a real sandbox boundary bypass (#4107). The concurrency model (cross-process flock + per-path in-process RLock, scope locks acquired sequentially and never nested), fail-closed error handling, atomic-rename rebuild, stale-worker-cache disk-reload-under-lock, and per-scope boot isolation are all correctly handled. The test matrix - including the real end-to-end bash-visibility repro - is thorough.

Verified the subtle points: isinstance(storage, LocalSkillStorage) still fires for UserScopedSkillStorage (it subclasses it), so the immediate public-projection rebuild on toggle works; projection failures don't drop unrelated mounts; removed DEER_FLOW_HOST_SKILLS_PATH/SKILLS_HOST_PATH have no lingering references.

Two operational items before merge (expanded inline): boot-time O(users) rebuild, and a release-notes note for the E2B mount-dropping behavior change.

backend/packages/harness/deerflow/skills/projection.py
fancyboi999: Handled in 1c7e67d1. Added an explicit note at the hardlink helper that hardlinks share the source inode and provide no write isolation, so any read-only guarantee must come from the consuming sandbox or mount.

backend/packages/harness/deerflow/skills/projection.py
fancyboi999: Handled in 1c7e67d1. Gateway boot now ensures only the shared public projection and never enumerates historical users; per-user views repair lazily on Local/AIO acquire. Public storage/signature/staging failures clear the view and manifest under the projection lock without aborting lifespan. Added red/green tests for lazy repair, factory failure, and deterministic rebuild/cleanup interleaving; the concurrency case passed 10/10 in independent verification.

backend/packages/harness/deerflow/community/e2b_sandbox/e2b_sandbox_provider.py
fancyboi999: Handled in 1c7e67d1. Added an Unreleased breaking-change entry covering both removed host-path variables and the E2B behavior: configured mounts at /mnt/skills or descendants are skipped with a warning and operators must move extra content to another container path.

docker/provisioner/app.py
```diff
@@ -360,6 +359,7 @@ def _build_volume_mounts(
     scope that mount with ``SKILLS_PVC_SUBPATH_TEMPLATE``.
     """
     mounts: list[k8s_client.V1VolumeMount] = []
+    del include_legacy_skills  # retained for request compatibility
 
     if SKILLS_PVC_NAME:
```
willem-bd, line 644: **PVC-backed deployments still expose disabled skills (scope caveat)**

In PVC mode (`SKILLS_PVC_NAME` set), this branch mounts the operator-provisioned PVC root at `/mnt/skills` read-only, and the gateway does not populate that PVC with enabled-only projections. So any deployment using `SKILLS_PVC_NAME` still has the exact #4107 exposure — disabled skills on the PVC remain readable via shell/file tools.

The PR notes dynamic PVC materialization is tracked separately, but given this is a `risk:high` security fix, could the residual exposure be called out more loudly (e.g., a CHANGELOG/release-note line that the fix applies to hostPath/Local/AIO/new-E2B only)? Not a regression — PVC mode was always raw — but operators may assume the projection fix covers them.

backend/packages/harness/deerflow/skills/projection.py
```diff
@@ -0,0 +1,435 @@
...
+    return paths
+
+
+def ensure_skill_projections(storage: SkillStorage) -> SkillProjectionPaths:
+    """Repair stale projection scopes, otherwise leave their inodes untouched."""
+    paths = get_skill_projection_paths(storage)
+    with _projection_lock(paths.public.parent):
+        try:
+            manifest = _read_manifest(paths.public.parent)
+            signature = _source_signature(storage, "public")
```
willem-bd: **Global public lock serializes every acquire (scalability)**

`_source_signature(storage, "public")` — a stat walk of the public tree plus an `extensions_config.json` disk read/parse — runs *inside* `_projection_lock(paths.public.parent)`, which is a single in-process `RLock` shared by all users. Since every user acquire checks the public scope, all acquires serialize for the duration of this read-only work. The 3.25 ms median is fine at low concurrency, but this is a global chokepoint under load.

Consider computing the signature *outside* the lock and taking the lock only to compare-and-rebuild. The rebuild's existing `before`/`after` signature check already guards the TOCTOU, so this would shrink common-case lock hold time to a manifest read + compare.

backend/packages/harness/deerflow/skills/projection.py
```diff
@@ -0,0 +1,435 @@
...
+    host_root = storage.get_skills_root_path()
+    if scope == "public":
+        _update_tree_digest(digest, host_root / SkillCategory.PUBLIC.value, "public")
+        state = {"extensions": _extensions_state()}
+    elif scope == "user":
+        user_custom_root = storage.get_user_custom_root()
+        _update_tree_digest(digest, user_custom_root, "custom")
+        _update_tree_digest(digest, host_root / SkillCategory.CUSTOM.value, "legacy")
+        state = {
+            "extensions": _extensions_state(),
```
willem-bd, line 273: **User-scope signature includes unrelated public state (minor inefficiency)**

The `"user"` scope signature folds in `_extensions_state()` (global public enable states) even though user custom/legacy projections don't depend on public state. A public-skill toggle therefore changes every user's signature and triggers a ~400 ms user-projection rebuild on each user's next acquire.

Dropping `_extensions_state()` from the user signature (keeping `storage._read_skill_states()`) would avoid spurious user rebuilds on public toggles.

backend/packages/harness/deerflow/skills/projection.py
```diff
@@ -0,0 +1,435 @@
...
+
+def _load_public_skills(storage: SkillStorage, *, enabled_only: bool) -> list[Skill]:
+    from deerflow.config.extensions_config import ExtensionsConfig
+
+    public_root = storage.get_skills_root_path() / SkillCategory.PUBLIC.value
+    if not public_root.is_dir():
+        return []
+    extensions = ExtensionsConfig.from_file()
+    skills: list[Skill] = []
+    for current_root, dir_names, file_names in os.walk(public_root, followlinks=True):
```
willem-bd, line 316: **`_load_public_skills` diverges from the runtime loader's package boundary (minor consistency)**

This walk registers *every* `SKILL.md` it finds — there's no `dir_names.clear()` after a hit, so it keeps descending. The runtime `_iter_skill_files`/`load_skills` treats a `SKILL.md` directory as a package boundary and stops descending, so nested `SKILL.md` files are supporting data, not separate skills.

For enabled nested skills the projection still lands at the correct path (the full `relative_path` carries the nesting), but a nested skill whose name is set to `enabled: false` in `extensions_config` is removed from the sandbox filesystem while the runtime still loads it as part of the parent package — a visibility inconsistency. Mirroring the runtime rule (`dir_names.clear()` once a skill is registered) keeps the projection and runtime in agreement.

fancyboi999:
Follow-up to @AnnaSuSu's live-view repro: fixed in f4bd0d99.

The projection no longer clears a category before staging. It now builds the complete desired tree first, removes only entries that must disappear, and atomically replaces staged files into the stable live root. A content write to `beta` therefore keeps unrelated `alpha` continuously visible. Disable/delete paths preserve fail-closed behavior by hiding only the affected package before the source/state mutation.

The final pass also covered edge cases the first implementation missed:

- namespaced packages use `Skill.relative_path`, not the frontmatter name as a filesystem path;
- every same-named public namespace path is hidden before a global disable;
- a drifted namespace symlink is unlinked rather than followed;
- mutation, targeted-removal, staging, and rebuild failures clear the whole affected scope and manifest;
- incremental reconcile handles file/directory/symlink type changes without blanking unrelated skills.

Validation on the pushed commit: projection suite `28 passed`; combined projection/provider/router/client/storage matrix `385 passed`; blocking-I/O regression suite `39 passed`; full ruff check/format clean; changed-line blocking-I/O scan found 0 candidates; real Gateway startup and `GET /health` returned 200 followed by clean shutdown. A fresh independent verifier reran the original namespace/fail-closed/symlink/type-change repros and reported no blocking findings.

backend/packages/harness/deerflow/skills/projection.py
fancyboi999: I traced this through `UserScopedSkillStorage.load_skills()` before changing it. CUSTOM/LEGACY visibility is computed as the per-user state **AND** `extensions_config.is_skill_enabled(...)`; therefore their projection does depend on global extensions state. Dropping `_extensions_state()` would let a global disable leave an existing user projection incorrectly fresh and readable.

I kept the dependency and added an invariant comment in f4bd0d99. Your performance concern is still valid: because the config is keyed only by skill name/category-agnostic state, a purely public toggle conservatively invalidates user manifests too. Filtering that precisely would require discovering/parsing the relevant CUSTOM/LEGACY names during the signature check, adding I/O to every user acquire. This commit chooses correctness over that over-invalidation rather than applying the unsafe direct deletion.

backend/packages/harness/deerflow/skills/projection.py
fancyboi999: Handled in f4bd0d99. The public fresh-path manifest/signature check now runs outside the global projection lock. Only stale/error paths acquire the lock, then re-check freshness before rebuilding, so concurrent stale acquires still coalesce to one rebuild.

Regression coverage includes concurrent fresh acquires meeting at a barrier (which deadlocked under the old serialized path), eight concurrent stale acquires rebuilding exactly once, and a manifest-change-during-signature-scan TOCTOU case.

docker/provisioner/app.py
fancyboi999: Handled in f4bd0d99. The Unreleased CHANGELOG entry now states explicitly that PVC-backed provisioner deployments still mount the operator-supplied snapshot directly, so disabled-skill filesystem isolation does not apply in PVC mode until dynamic PVC materialization exists. This keeps the residual #4107 exposure visible to operators instead of implying the hostPath/Local/AIO/new-E2B coverage also applies to PVC.

backend/packages/harness/deerflow/skills/projection.py
fancyboi999: Handled in f4bd0d99. `_load_public_skills` now clears `dir_names` after finding a package `SKILL.md`, matching the runtime loader. Nested `SKILL.md` files remain supporting data inside the enabled parent package and are no longer independently filtered by their frontmatter name. The regression test now asserts that exact package-boundary behavior.

backend/packages/harness/deerflow/skills/projection.py
fancyboi999: Follow-up in `f58b1187`: the global extensions state must remain in the user signature because CUSTOM/LEGACY visibility is the intersection of per-user state and that global state. Removing it would still make a global disable incorrectly fresh.

While validating that invariant across workers, I found and fixed a separate stale-cache gap: `UserScopedSkillStorage.load_skills()` performed its CUSTOM/LEGACY merge through the process singleton even though the signature read disk. A worker with an old singleton could therefore rebuild an enabled view from stale state and write a manifest for the new disabled signature. The user-scoped merge now re-reads `ExtensionsConfig.from_file()`, matching the base loader's cross-process behavior.

The regression deliberately leaves the singleton enabled while fresh disk state disables both a CUSTOM and a LEGACY skill. It asserts the first ensure removes each target and writes the manifest, then the second ensure leaves the target absent and the manifest byte-for-byte stable. A fresh independent verifier also reproduced the boundary with two real processes and temporary files; unrelated public/user skills stayed visible.

Validation: projection file `30 passed`; related projection/storage/router/client/provider matrix `366 passed`; blocking-I/O `39 passed`; ruff check/format clean across 862 files; real Gateway `/health` returned 200 and shut down cleanly.

AnnaSuSu:
Re-ran my repros against `f58b1187` rather than reading the diff — same method as last time, plus a red check on `1c7e67d1` (the commit before `f4bd0d99`) so a green result can't just mean my script broke.

**#1 and #2 are closed.** Confirmed, and the fail-closed property survived the change:

*Repro A* — two enabled custom skills, live reader on `alpha/SKILL.md`, one ordinary `write_custom_skill('beta')`:

```
1c7e67d1 (before): ok=3429  ENOENT=933   blank window 3.5 ms
f58b1187 (after) : ok=3398  ENOENT=0     blank window: none
```

*Repro B* — the real 23-skill / 104-file public tree, real router RMW (`_persist_shared_skill_state`), real on-disk extensions config (no `from_file` patching), reader on `deep-research/SKILL.md`, admin disables the unrelated `skill-creator`:

```
1c7e67d1 (before): ok=2373  ENOENT=26535  blank window 129.9 ms
f58b1187 (after) : ok=4752  ENOENT=0      blank window: none
skill-creator actually hidden after disable: True in both  <- fail-closed intact
```

The delete-what-must-disappear-then-stage-and-replace order does exactly what it needed to. Nice.

---

**#3: I have to retract my own recommendation. It was wrong, and acting on it made this slower.**

The lock genuinely is off the steady-state public path — you implemented what I asked for. But what I asked for was based on a mechanism I never verified, and it doesn't hold. Sweep of N concurrent no-op acquires, N distinct users, nothing stale:

```
                 1c7e67d1          f58b1187
single             2.56 ms           2.71 ms
N= 8      wall    25.9 ms           41.0 ms     (1.58x)
N=16      wall    49.2 ms           85.4 ms     (1.74x)
N=32      wall   101.9 ms          159.0 ms     (1.56x)
N=64      wall   190.7 ms          369.1 ms     (1.94x)
per-acquire      ~3.0-3.2 ms       ~5.0-5.8 ms
```

7 trials at N=16, tight enough to rule out noise: old 48.4 ms (47–50), new 76.6 ms (71–80).

The load-bearing column is `per-acquire`: it's **flat in both versions**. Neither one parallelizes at all. My "15.8x serialization factor" was real, but I attributed it to the lock without testing that attribution. It's the GIL. Replacing `_projection_lock` with a no-op:

- on `f58b1187`: 32.1x → 29.6x — the lock accounts for essentially nothing, because it's already off the path
- on `1c7e67d1`: 9.2x → **26.6x**, i.e. *worse* — the old global lock was incidentally acting as a **throttle**, admitting one thread at a time to GIL-bound work and thereby avoiding convoy. Removing it bought no parallelism (there was none available) and lost the throttle.

Why there was never parallelism to win: profiling the steady-state check (200 calls), `_source_signature` is 92.6% of it, and the GIL-*releasing* syscalls (`posix.scandir` 0.158 s + `DirEntry.stat` 0.097 s tottime out of 1.008 s) are only ~25%. Amdahl caps any possible win from parallelizing at ~1.33x even with a perfect lock-free path. And on Linux, where stat is cheaper than on APFS, the GIL-bound share is *higher*, so there's even less to win — my macOS numbers are conservative toward your change, not against it.

The real chokepoint is that an O(files) walk runs on **every acquire at all** — not where the lock sits. Two things follow, both your call:

- The double-checked-locking refactor doesn't pay for itself on these numbers. Keeping it is defensible (it's correct, and it's not the reason for the regression), but it isn't buying the scalability I claimed it would.
- If you want the win: make the steady-state check cheap or skip it. Every in-codebase mutation already rebuilds under the lock, so the acquire-time walk is purely a repair path for *out-of-band* drift — rare. A short TTL or a generation counter would make the common case ~free and keep the self-heal. Separately, `_update_tree_digest` constructs a `Path` per entry (`__truediv__`/`as_posix`/`_load_parts` ≈ 75% of single-call CPU); a plain-string version produces a byte-identical digest (`210b6b7a4278a4eb…`) and cuts single-call 2.46 → 1.72 ms. It does *not* fix the concurrent case (67.9 vs 69.5 ms at N=16) — worth it only as a cheap constant-factor win, not as an answer to #3.

**Environment, because this one is platform-sensitive:** macOS 15 / APFS, Python 3.12.9, 8 cores (4 performance). I could not cross-check on Linux — no Docker on this machine. The GIL argument is CPython-wide and platform-independent in direction; the convoy *magnitude* is FS/OS dependent, so please re-run the sweep on CI before you act on the ratios. I'd rather hand you a number you can re-measure than a conclusion you have to take on trust — mine is exactly what went wrong the first time.

---

**Two things still open from my last pass:**

1. `user_should_see_legacy_skills` still has zero callers repo-wide and is still exported in `__all__` (`skills/storage/__init__.py:146` / `:200`). Its docstring still says it "centralizes that contract" for the mount-vs-discovery rule — but the rule now lives only in `_iter_skill_files`'s `if not user_custom_exists`. That's what makes it a trap rather than harmless dead weight: it advertises itself as the canonical entry point for an invariant it no longer implements, which is precisely how the rule gets re-forked.

2. **`Fixes #4107`.** The PVC residual made it into the CHANGELOG — good. But the *existing-E2B-sandbox* residual (a disabled skill stays `bash cat`-readable until the VM is recreated) isn't recorded anywhere, and the question I asked when I opened this review is still unanswered: does #4107 stay open to track the residuals, or get closed with them documented? As it stands `Fixes #4107` auto-closes my issue on merge with two live exposure paths inherited silently. I'm fine with either resolution — I'm not fine with it being implicit. Happy to keep #4107 open myself and scope it down to "E2B existing sandboxes + PVC mode" if you'd rather this PR close clean.

Repro scripts for A/B and the #3 sweep are standalone; say the word and I'll attach them.

---

**TL;DR（中文）**：**#1（我标的 blocker）和 #2 确认关闭**——对着 `f58b1187` 重跑，并对 `1c7e67d1` 做了红检确保不是脚本坏了：Repro A ENOENT 933→0，Repro B（真实 104 文件树 + 真实 router RMW + 真实磁盘配置，无 mock）ENOENT 26535→0、空窗 129.9ms→无，且 fail-closed 仍成立。**#3 我要收回自己的建议**：锁确实按我说的挪出了稳态路径，但我当初把 15.8x 归因于锁**从没验证过这个归因**，是错的。并发扫描显示两个版本的 per-acquire 成本都是**平的**——谁都没有并行：把 `_projection_lock` 换成 no-op，新代码 32.1x→29.6x（锁本就不在路径上），旧代码 9.2x→**26.6x 反而更差**（旧的全局锁其实在**当节流阀**，一次放一个线程进 GIL 密集区，避免了 convoy）。结果是 N≥8 时 wall clock 反而慢 1.5–1.9 倍（N=64: 190.7ms→369.1ms）。profile 显示 `_source_signature` 占 92.6%，而释放 GIL 的系统调用只占 ~25%——按 Amdahl，就算完美并行上限也只有 ~1.33x；Linux 上 stat 更快、GIL 占比更高，可赢的更少，所以我 macOS 的数字对你的改动是**偏保守**的。真正的瓶颈是**每次 acquire 都做 O(files) 全树走**，不是锁在哪。建议（你定）：稳态检查加 TTL 或 generation 计数器——所有代码内变异本来就在锁下重建了，acquire 时的走查只是防**外部漂移**的修复路径，很罕见。另 `_update_tree_digest` 每个条目造一个 `Path`（占单次 CPU 约 75%），改纯字符串可得**逐字节相同**的 digest、单次 2.46→1.72ms，但救不了并发（67.9 vs 69.5ms），只算常数优化。**环境说明**：macOS/APFS + Python 3.12.9 + 8 核，**没有 Linux 交叉验证（本机没装 Docker）**，convoy 幅度与文件系统强相关，动手前请在 CI 上重跑。**两件还开着的**：(1) `user_should_see_legacy_skills` 仍零调用者、仍在 `__all__`，且 docstring 仍自称 "centralizes that contract"——规则已经只在 `_iter_skill_files` 里了，它把自己宣传成权威入口才是真正的陷阱；(2) `Fixes #4107` 会在 merge 时自动关掉我的 issue，而 PVC 残留进了 CHANGELOG、**存量 E2B 沙箱的残留没有任何记录**，我两次问的"继续开着还是关掉+记录"仍没答案。两种处理我都接受，但不接受它是隐式发生的。

fancyboi999:
Followed up on the remaining points in `27358bd5`:

- Removed `user_should_see_legacy_skills` entirely, including its public export and now-unused `SkillCategory` import. The projection is now the only owner of the legacy visibility rule; a repo-wide search has no remaining references to the old helper.
- Made both residual paths explicit in `README.md` and `CHANGELOG.md`: existing E2B sandboxes retain their creation-time snapshot until recreation, and PVC-backed provisioner deployments retain the operator-supplied snapshot. I also changed the PR body from `Fixes #4107` to `Refs #4107`, so the issue stays open rather than being auto-closed while those paths remain.

I am not adding a TTL or generation cache in this PR. The current acquire-time self-heal is correct and its O(files + directories) cost is disclosed; a TTL would introduce a stale window for out-of-band drift. The reported convoy magnitude is also macOS/APFS-specific and was not cross-checked on Linux. That performance redesign needs its own cross-platform benchmark and security trade-off, rather than changing the projection's freshness contract here.

Validation on the current head:

- Related projection/storage/router/provider matrix: `423 passed` locally.
- Ruff lint and format checks passed; changed blocking-I/O scan reported 0 candidates.
- All executed GitHub checks passed on `27358bd5`; conditional triage/publish/version jobs were skipped.

WillemJiang:
@fancyboi999  please resolve the conflict with the main branch.

backend/app/gateway/routers/skills.py
```diff
@@ -107,6 +107,32 @@ def _skill_to_response(skill: Skill) -> SkillResponse:
...
+    from deerflow.skills.storage.local_skill_storage import LocalSkillStorage
+
+    removal_names = (skill_name,) if not enabled else ()
+    projection_update = skill_projection_mutation(storage, "public", remove_names=removal_names) if rebuild_public_projection and isinstance(storage, LocalSkillStorage) else nullcontext()
+    with projection_update:
+        # The projection lock is cross-process, but the singleton cache is not.
+        # Reload the latest disk state inside the lock before this RMW.
+        extensions_config = ExtensionsConfig.from_file(config_path) if config_path.exists() else ExtensionsConfig(mcp_servers={}, skills={})
+        extensions_config.skills[skill_name] = SkillStateConfig(enabled=enabled)
+        config_data = extensions_config.model_dump(by_alias=True, exclude_unset=True)
```
willem-bd: **Regression from the merge resolution: `exclude_unset=True` silently drops the `skills` key when the source config has no `skills` field, so a public-skill toggle doesn't persist.**

Reproduced against the actual `ExtensionsConfig` model on this HEAD. Operator config with only `mcpServers` (a common shape — MCP configured, no skill ever toggled):

```
before: {"mcpServers": {"weather": {...}}}             # no "skills" key
PUT /api/skills/deep-research  body={"enabled": false}
after:  keys written = ["mcpServers"]                 # "skills" missing
reload: is_skill_enabled("deep-research", "public") -> True   # disable lost
```

Why: `ExtensionsConfig.from_file()` goes through `model_validate()`, so `model_fields_set` only contains fields present in the source JSON. `extensions_config.skills[skill_name] = ...` mutates the default `{}` dict in place — it does NOT add `skills` to `model_fields_set` (Pydantic v2 doesn't track dict-item mutations, and `validate_assignment` is off here). `model_dump(by_alias=True, exclude_unset=True)` therefore omits `skills`, and the toggle is not written to disk.

Security implication: the projection rebuild reads enabled state via `ExtensionsConfig.from_file()` inside `_load_public_skills`, so the just-disabled public skill stays visible in the sandbox filesystem — i.e. the exact #4107 bypass this PR is fixing, for the common "config with no `skills` key" shape.

The pre-merge code (`config_data = {"mcpServers": {...}, "skills": {...}}` reconstructed manually) always wrote both keys. The merge resolution switched to `model_dump(by_alias=True, exclude_unset=True)` and introduced this. `test_enable_toggle_allowed_for_admin` doesn't catch it because it seeds the config with `"skills": {}` (putting `skills` in `model_fields_set`); a test whose seed omits the `skills` key would.

Fix options:
- Minimal: call `extensions_config.to_file_dict()` (which already does `model_dump(by_alias=True)` without `exclude_unset`) — restores the pre-merge "always write known keys" behavior. Cost: default-shape fields like `middlewares: []` get serialized into operator configs that didn't have them.
- Preserve the merge's apparent intent (don't add unset optional fields) while still writing the mutated key: keep `exclude_unset=True` and explicitly re-add `skills` afterward, e.g. `config_data["skills"] = {name: s.model_dump() for name, s in extensions_config.skills.items()}`.

The first is the smallest delta; the second preserves the intent.

willem-bd (commented):
Scoped this pass to the new merge commit `9b995b7a` (resolving the main conflict) since the projection layer itself has already been reviewed end-to-end across the prior rounds and the inline findings there were all addressed.

The merge only touches one PR-owned file substantively — `backend/app/gateway/routers/skills.py` — and introduces one real regression in the conflict resolution, flagged inline: `model_dump(by_alias=True, exclude_unset=True)` silently drops the `skills` key when the operator's `extensions_config.json` has no `skills` field, so a public-skill disable doesn't persist and the projection keeps the skill visible. Reproduced against the actual `ExtensionsConfig` model on this HEAD. The existing `test_enable_toggle_allowed_for_admin` doesn't catch it because it seeds the config with `"skills": {}`.

The other merge-induced change in the same file (offloading `get_custom_skill_history`'s filesystem reads to `asyncio.to_thread`) looks correct.

Minor aside (not blocking, didn't anchor inline because the lines aren't in any PR hunk): `rollback_custom_skill` still runs its read-side filesystem calls (`custom_skill_exists`, `read_history`, `validate_skill_markdown_content`, `get_custom_skill_file`, `skill_file.read_text`) directly on the event loop even though the same function now offloads the write-side calls and `get_custom_skill_history` offloads the same read calls. The changed-lines blocking-IO scan won't flag this, but the route is now visibly inconsistent with its neighbors.

backend/app/gateway/routers/skills.py
```diff
@@ -483,20 +511,39 @@ def _load_storage_and_skills() -> tuple[SkillStorage, list[Skill]]:
...
-            # the whole read→write window, so it stays serialized against the MCP
-            # router even if this request is cancelled mid-write.
-            await asyncio.to_thread(_write_extensions_skill_state, skill_name, body.enabled)
+            config_path = ExtensionsConfig.resolve_config_path()
+            if config_path is None:
+                config_path = Path.cwd().parent / "extensions_config.json"
+                logger.info(f"No existing extensions config found. Creating new config at: {config_path}")
+
+            await asyncio.to_thread(
+                _persist_shared_skill_state,
```
willem-bd: **This swap is where the earlier line-130 regression stays live.** The merge-base called `_write_extensions_skill_state` (the corrected helper from `main`: `to_file_dict()` instead of `exclude_unset=True`, snapshots the singleton, holds the shared `extensions_config_write_lock`). The merge resolution replaced **both** call sites (here and the fallback at line 540) with this branch's `_persist_shared_skill_state` to preserve the `skill_projection_mutation` lock, and in doing so re-opened two things:

1. The toggle-persistence bug: `_persist_shared_skill_state` still ends with `extensions_config.model_dump(by_alias=True, exclude_unset=True)` (line 130). When the operator config has no `skills` key, the toggle is silently dropped on write — the exact #4107 bypass for the common "MCP configured, no skill ever toggled" config shape.
2. The cross-router RMW race with `mcp.py`: `extensions_config_write_lock` was imported (line 15) for this, but `_persist_shared_skill_state` never acquires it, so a skill toggle and an MCP-config update can still interleave their read-modify-writes and drop each other's change. The lock is only taken by `_write_extensions_skill_state`, which is now dead code (defined at line 451, never called).

The tests imported from `main` (`backend/tests/blocking_io/test_skills_update_router.py`) were written against `_write_extensions_skill_state`'s semantics and fail against this path: `test_update_skill_writes_from_snapshot_without_mutating_singleton` expects `existing-skill` (held in the patched singleton) to survive the write and `middlewares` to appear in the file — both fail because `_persist_shared_skill_state` rebuilds from `from_file()` and uses `exclude_unset`. `test_skill_and_mcp_config_writes_are_serialized` likewise fails because no shared lock is held.

Cleanest reconciliation: fold `skill_projection_mutation` (the one thing this branch's helper does that `main`'s doesn't) into `_write_extensions_skill_state`, restore the two call sites to use it, and delete `_persist_shared_skill_state`.

willem-bd (commented):
Re-reviewed the new merge head `e5b784bf`. The earlier line-130 finding (public-skill toggle not persisting when the source config lacks a `skills` key) is **still live and now sharper**: GitHub's base...head diff shows the merge resolution actively swapped both `update_skill` call sites from `main`'s corrected `_write_extensions_skill_state` back to this branch's buggy `_persist_shared_skill_state` (to preserve the projection lock). The corrected helper is now dead code in the file, the `extensions_config_write_lock` it carries is imported but never acquired on the live path, and the regression tests imported from `main` will fail against the current routing. One inline follow-up at the swap site. No other new issues in the merged delta.

WillemJiang:
@fancyboi999, could you take a look at the failed tests and resolve the conflicts with the main branch?

backend/app/gateway/routers/skills.py
fancyboi999: Fixed in fe9cceea. The live route now goes through the unified `_write_extensions_skill_state` helper and serializes with `to_file_dict()`, so a source config without a `skills` key persists the toggle (including the full known top-level shape). I added missing-key regression coverage for both the Gateway route and `DeerFlowClient.update_skill`. A real Gateway run with an input containing only `mcpServers` returned `enabled: false`, wrote `skills.deep-research.enabled=false`, and removed the skill from the public projection.

backend/app/gateway/routers/skills.py
fancyboi999: Fixed in fe9cceea with the suggested reconciliation: `skill_projection_mutation` is now folded into `_write_extensions_skill_state`, both call sites use that helper again, and `_persist_shared_skill_state` is removed. The worker owns the full critical section in projection-lock -> `extensions_config_write_lock` order, re-reads existing disk state under those locks, writes via `to_file_dict()`, reloads config, and rebuilds the projection before returning. Verification: 62 blocking-I/O tests passed, the related skills/sandbox matrix passed 588 tests (11 skipped), and the full non-live backend suite passed 10,272 tests (61 skipped).

fancyboi999:
@WillemJiang conflicts are resolved on current `main` in `87fe9bb3`. The merge keeps the projection lock and adopts `main`'s atomic extensions-config writer; the missing-`skills` persistence regression is covered, and all 14 review threads are resolved.

Current verification: 12 focused router/atomic-write tests and a 591-test projection/provider/router matrix passed; backend lint/format and the blocking-I/O scan are clean; a real Gateway `PUT /api/skills/deep-research` followed by `GET /api/skills` persisted the disable. GitHub CI is green and the PR is mergeable now. Could you please take another look?

willem-bd (commented):
Re-checked the latest merge head `87fe9bb3` (the only delta since the last-reviewed `fe9cceea` - a merge of `main`). It touches one PR-owned source file, `backend/app/gateway/routers/skills.py`, and the conflict resolution is clean and beneficial:

- `_write_extensions_skill_state` now writes via main's `atomic_write_extensions_config(config_path, config_data)` (temp + replace) instead of a bare `json.dump` - consistent with the rest of the codebase and atomic.
- The earlier correctness fix is preserved: it still serializes with `extensions_config.to_file_dict()` (not `exclude_unset=True`), so a source config lacking a `skills` key no longer drops the field on a public-skill toggle.
- No swap back to `_persist_shared_skill_state` - both toggle call sites still route through the unified `_write_extensions_skill_state`.
- The dropped `import json` is safe - no remaining `json.*` usage in the file. `atomic_write_extensions_config(path: Path, data: dict)` signature matches the call.

All prior inline findings are addressed by the fix commits (a5f04600, 88ca8c6e, 1c7e67d1, f4bd0d99, f58b1187, fe9cceea), which are ancestors of this head. No new issues from the merge.

WillemJiang (approved):

PR #4181, opened 2026-07-14
fix(memory): flush memory queue on graceful shutdown to prevent loss

### LINKED ISSUE #4180: [bug] Gateway 优雅关闭时 memory queue 未 flush，pending 记忆更新静默丢失
### Before you start

### Problem summary

Gateway 的 `lifespan` 优雅关闭流程不排空 `MemoryUpdateQueue`。该队列纯内存（`queue.py:41`）、debounce Timer 是 daemon 线程（`queue.py:174`），而 `flush()` 在生产代码中从未被调用——因此默认 30s debounce 窗口内已入队、Timer 尚未触发的 memory 更新，在重启 / 滚动更新 / SIGTERM / Ctrl+C 时静默丢失，无错误、无日志痕迹。

### Affected area(s)

### What happened?

`MemoryUpdateQueue`（`packages/harness/deerflow/agents/memory/queue.py`）是 memory 模块的入队缓冲层：`MemoryMiddleware.after_agent` 在每次 agent run 结束后把"用户输入 + 最终 AI 回复"入队，队列默认 debounce 30s（`memory.debounce_seconds`）后再批量做一次 LLM 记忆提取并写 `memory.json`。这个队列有两个关键属性导致 shutdown 丢失：

- **纯内存**：`self._queue: list[ConversationContext] = []`（`queue.py:41`），无任何持久化。
- **daemon Timer**：debounce 用 `threading.Timer` 实现，`self._timer.daemon = True`（`queue.py:174`），进程退出时直接被杀，不会跑完。

而 Gateway 的优雅关闭流程（`app/gateway/app.py::lifespan`，`yield` 之后）只做了三件事——close OIDC、`stop_channel_service`、stop scheduled task service——**没有任何代码排空 memory queue**。因此任何在 debounce 窗口内入队、Timer 尚未触发的更新，都会在进程退出（重启 / 滚动更新 / SIGTERM / Ctrl+C）时静默丢失，无错误、无日志痕迹。

证据链：

1. **`flush()` / `flush_nowait()` 在生产代码中从未被调用。** 全仓 grep（排除测试）只有 `summarization_hook.py` 的 `add_nowait()` 和 `memory_middleware.py` 的 `add()` 命中 `get_memory_queue()`；`MemoryUpdateQueue.flush` 只出现在 `tests/test_memory_queue.py`。其余 `.flush()` 命中的都是无关对象（`RunJournal`、`_SubagentEventBuffer`、文件句柄）。
2. **`lifespan` 关闭段不 flush。** `app/gateway/app.py` 在 `yield` 之后无 `get_memory_queue().flush()` 调用。
3. **`flush_nowait` 自己的 docstring 承认会丢**："queued messages may be lost if the process exits before _process_queue completes"（`queue.py:251-253`）。但可靠的同步 `flush()`（`queue.py:245-255`，取消 Timer 并在调用线程同步跑 `_process_queue`）存在却无人调用。
4. **`memory_flush_hook` 不能弥补。** 它（`summarization_hook.py:27`）调用 `queue.add_nowait()`（入队立即处理），不是 `flush()`；且触发于 summarization 前（防止消息被压缩掉前先抽取记忆），与 shutdown 路径无关。
5. **`updater.py` 的 atexit 也不 flush。** 它只 `shutdown(wait=False)` 了 LLM 调用的线程池（`updater.py:40-44`），不排空队列。
6. **默认配置即触发。** `config.example.yaml`：`memory.enabled: true`、`memory.debounce_seconds: 30`。

**影响**：默认部署即受影响。IM channel 场景最严重——每条入站消息都触发一次入队，30s 窗口内重启就丢最近一条记忆；用户表现为 agent "失忆"，且无任何日志可追溯。K8s 滚动更新逐个重启 pod 时，每个 pod 都可能丢其 debounce 窗口内的更新。

### Expected behavior

Gateway 优雅关闭时应排空 in-memory `MemoryUpdateQueue`，使 debounce 窗口内已入队、Timer 尚未触发的更新在 worker 退出前完成持久化（LLM 提取 + 写 `memory.json`），从而在重启 / 滚动更新 / SIGTERM 后存活。

具体要求：

- 在 `lifespan` 的 shutdown 段（`yield` 之后）调用队列的同步排空（队列已有同步 `flush()`：取消 Timer 并在调用线程跑 `_process_queue`），而不是 `flush_nowait()`（后者只派生一个 daemon Timer，进程退出照样被杀，不解决问题）。建议新增带硬超时封装 `flush_sync(timeout)`（daemon 线程 + `Event.wait`）供 shutdown 调用，见 Additional context。
- 排空须带**硬超时**：每条更新要做一次 LLM 调用（数秒级），不能无限期阻塞 worker 退出；超时后允许丢弃未完成部分（严格优于现状的"全丢"）。
- 排空应在 channel service / scheduler 停掉**之后**进行，避免排空期间又有新更新入队。
- 现有 `_SHUTDOWN_HOOK_TIMEOUT_SECONDS = 5.0` 对单条 LLM 调用太短，需要独立的更长预算。

### Steps to reproduce

1.  放大 debounce 窗口便于从容操作：
修改config.yaml中的memory.debounce_seconds为300

2. `make dev`，浏览器 `http://localhost:2026` 新建对话，发一条含明确事实的消息：
```
请记住我改用 Python 3.12 了，这是复现 bug A1 的标记事实。
```

3. 在 Timer 到期前（300s 内）停掉 Gateway：前台 `make dev` 按 `Ctrl+C`

4. 检查 memory.json：
```
ls backend/.deer-flow/users/*/memory.json 2>/dev/null || echo "无 memory.json"
```

5. 重启 `make dev`，新建一个对话（new thread），问「我之前告诉过你我用什么 Python 版本？」 —— agent 答不上来。
必须用新对话：同对话里 checkpointer 保留了聊天历史，agent 能从历史读到"Python 3.12"，会误判为已修复。本 bug 影响的是跨对话的长期记忆（memory.json），不是同 对话历史。

6. 重复步骤 1-2 但**不停进程**，等 debounce 到期（`logs/gateway.log` 出现 `Processing 1 queued memory updates` → `Memory saved to .../memory.json`），再查 memory.json，标记**会**写入。证明写入逻辑本身没问题，只是 shutdown 缺了排空。

### Relevant logs

```shell
$ ls backend/.deer-flow/users/*/memory.json 2>/dev/null || echo "无 memory.json"
>>> 无 memory.json → 入队了但 shutdown 没 flush，memory.json 没生成

# gateway.log
2026-07-14 20:21:12 - httpx - INFO - HTTP Request: POST https://api.deepseek.com/v1/chat/completions "HTTP/1.1 200 OK"
2026-07-14 20:21:17 - deerflow.agents.middlewares.token_usage_middleware - INFO - LLM token usage: input=7469 output=309 total=7778 input_token_details={'cache_read': 7424} output_token_details={'reasoning': 258}
2026-07-14 20:21:17 - deerflow.agents.memory.queue - INFO - Memory update queued for thread 7dec8788-2563-453f-98ee-6f9e313ecf78, queue size: 1
2026-07-14 20:21:17 - deerflow.runtime.runs.manager - INFO - Run 6c785873-6f91-4ea2-b2bf-a6122bde88d6 -> success
INFO:     ::1:0 - "POST /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/history HTTP/1.1" 200 OK
INFO:     ::1:0 - "POST /api/threads/search HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/token-usage HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78 HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/messages/page HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/runs/6c785873-6f91-4ea2-b2bf-a6122bde88d6/workspace-changes?include_files=true&include_diff=false HTTP/1.1" 200 OK
2026-07-14 20:21:18 - httpx - INFO - HTTP Request: POST https://api.deepseek.com/v1/chat/completions "HTTP/1.1 200 OK"
INFO:     ::1:0 - "POST /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/suggestions HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/1e23ce1a-ed9b-4f0a-8ad9-da3560bd8ae7/token-usage HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/1e23ce1a-ed9b-4f0a-8ad9-da3560bd8ae7/messages/page HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/1e23ce1a-ed9b-4f0a-8ad9-da3560bd8ae7/messages/page HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/skills HTTP/1.1" 200 OK
INFO:     ::1:0 - "POST /api/threads/search HTTP/1.1" 200 OK
INFO:     ::1:0 - "POST /api/threads/1e23ce1a-ed9b-4f0a-8ad9-da3560bd8ae7/history HTTP/1.1" 200 OK
INFO:     ::1:0 - "POST /api/threads/1e23ce1a-ed9b-4f0a-8ad9-da3560bd8ae7/history HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/1e23ce1a-ed9b-4f0a-8ad9-da3560bd8ae7 HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/token-usage HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/messages/page HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/messages/page HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/skills HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/uploads/limits HTTP/1.1" 200 OK
INFO:     ::1:0 - "POST /api/threads/search HTTP/1.1" 200 OK
INFO:     ::1:0 - "POST /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/history HTTP/1.1" 200 OK
INFO:     ::1:0 - "POST /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78/history HTTP/1.1" 200 OK
INFO:     ::1:0 - "GET /api/threads/7dec8788-2563-453f-98ee-6f9e313ecf78 HTTP/1.1" 200 OK
2026-07-14 20:26:17 - deerflow.agents.memory.queue - INFO - Processing 1 queued memory updates
2026-07-14 20:26:17 - deerflow.agents.memory.queue - INFO - Updating memory for thread 7dec8788-2563-453f-98ee-6f9e313ecf78
2026-07-14 20:26:18 - httpx - INFO - HTTP Request: POST https://api.deepseek.com/v1/chat/completions "HTTP/1.1 200 OK"
2026-07-14 20:26:20 - deerflow.agents.memory.storage - INFO - Memory saved to /home/tnc/Documents/deer-flow/backend/.deer-flow/users/fb0de27f-6e56-4e18-9a6a-e9ff1ebd675e/memory.json
2026-07-14 20:26:20 - deerflow.agents.memory.queue - INFO - Memory updated successfully for thread 7dec8788-2563-453f-98ee-6f9e313ecf78
```

Local (make dev)

### Operating system

Linux

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Support bundle summary

_No response_

### Additional context

先输入“请记住我改用 Python 3.12 了，这是复现 bug A1 的标记事实。”

Image
马上关闭程序，然后重启（make dev），开启新对话问“我之前告诉过你我用什么 Python 版本？”

Image
回答不出。

TuNaiChao:
Closes bytedance/deer-flow#4180

> **Rework note.** The prior revisions targeted the pre-`#4122` memory module (top-level `MemoryUpdateQueue` / `get_memory_queue()` singleton). `#4122` (pluggable `MemoryManager` abstraction) merged to `main` and architecturally invalidated that implementation — the queue is now an internal detail of the DeerMem backend. This force-push re-applies the **same reviewer-approved drain logic** on top of the new abstraction: the shutdown flush is a `MemoryManager` ABC method instead of host code reaching into the queue. The bug is unchanged and still present on `main` (the DeerMem queue is still an in-memory buffer drained by a daemon `Timer` that is killed on process exit, and the Gateway lifespan never drained it). Thanks @willem-bd for the three review passes — every finding is addressed in the changes below.

## Why

`MemoryUpdateQueue` (`backends/deermem/deermem/core/queue.py`) is the memory module's debounce buffer: `MemoryMiddleware.after_agent` enqueues each completed turn, and a `threading.Timer` fires `debounce_seconds` later (default 30s) to batch-extract memories via one LLM call and persist `memory.json`.

Two properties make it lossy on shutdown:
- **Pure in-memory** — `self._queue: list[ConversationContext] = []`, no persistence.
- **Daemon timer** — `self._timer.daemon = True`; process exit kills it mid-window.

Any update enqueued within the debounce window when the worker exits — restart, rolling deploy, SIGTERM, Ctrl+C — is silently lost. IM channels are hit hardest (every inbound message enqueues); K8s rolling updates lose each pod's in-flight window.

## What changed

**`MemoryManager` ABC (`agents/memory/manager.py`)** — new abstract method `shutdown_flush(timeout) -> bool`. Every backend now implements a bounded graceful-shutdown drain. The host owns the lifespan budget and the K8s-grace relationship, so the contract takes the timeout as a parameter rather than each backend reading its own.

**DeerMem (`backends/deermem/`)** — `DeerMem.shutdown_flush` delegates to the queue's `flush_sync`. The full reviewer-approved logic lives in `core/queue.py`:
- `flush_sync(timeout)`: runs `flush()` on a daemon thread, waits via `threading.Event.wait(timeout)`, returns whether the drain genuinely finished.
- **In-flight worker race** (`_processing_thread` + step 1): if the debounce Timer already fired, an `_process_queue` worker is mid-LLM-call holding contexts it already pulled out (`pending_count == 0`, `is_processing == True`). `flush_sync` stashes the active worker thread and **joins it first** (bounded by the remaining budget) instead of reporting a false-positive `True`.
- **Failed-flush result**: `success` is set on the happy path only, so `flush_sync` returns `False` when `flush()` raises (`return bool(success) and not self.is_processing`).
- **Drain skips inter-item sleep**: `_process_queue` / `flush` take `skip_inter_item_delay`; the `flush_sync` path sets it so the bounded budget goes to LLM calls, not `time.sleep(0.5)`.
- **Per-item succeeded/failed count**: `_process_queue` now logs `Memory update batch done: %d succeeded, %d failed` so an operator can distinguish "drained" (queue emptied) from "saved" (every extraction persisted) — per-item `update_memory` failures are swallowed inside the loop.

**noop backend** — `shutdown_flush` is a clean no-op `return True` (nothing is ever queued).

**Gateway lifespan (`app/gateway/app.py`)** — calls `get_memory_manager().shutdown_flush(timeout)` via `asyncio.to_thread`, **after** channels and the scheduler stop, wrapped in `try/except` so shutdown can never block on it.
- **No host-level `pending_count`/`is_processing` guard.** The backend short-circuits on a truly idle buffer (returns `True` immediately), so calling it unconditionally is cheap, and keeping the in-flight race entirely inside the backend means the host **cannot "forget" that case** the way a `pending_count > 0`-only guard would (eliminated structurally rather than patched).
- Timeout read from config; result logged (`INFO` `Memory queue flush completed within %.1fs` on success, `WARNING` `Memory queue flush did not finish within %.1fs; remaining updates may be lost` on timeout/failure).

**Config (`config/memory_config.py` + `config.example.yaml`)** — new host-shared `memory.shutdown_flush_timeout_seconds` (default 30, 1–300). Host-shared (not `DeerMemConfig`) because the lifespan budget and the K8s-grace relationship are host concerns; every backend benefits, and the host reads it without coupling to a specific backend. `config_version` 25 → 26.

**API surface (`routers/memory.py` + `client.py`)** — `shutdown_flush_timeout_seconds` exposed on `MemoryConfigResponse` (`/memory/config`, `/memory/status`) alongside the other shared fields, and mirrored on the embedded client.

**Helm (`deploy/helm/deer-flow/`) — the P0 operational fix.** The default Gateway deployment set neither `terminationGracePeriodSeconds` nor a `preStop` hook, so K8s defaulted to a 30s grace — shorter than the shutdown work (channel stop ~5s + memory drain default 30s) — and SIGKILLed the drain mid-flight, silently re-introducing the loss this PR fixes. The bare-process SIGTERM in the prior validation couldn't surface this.
- gateway pod: `terminationGracePeriodSeconds` (default 45, configurable via `gateway.terminationGracePeriodSeconds`).
- gateway container: `preStop` sleep (default 5, `0`/unset disables) so the Service/ingress deregisters before SIGTERM begins the drain.
- `values.yaml` + README: both configurable; README documents that the grace period must track `memory.shutdown_flush_timeout_seconds`.

### Why a daemon thread, not `asyncio.to_thread(flush) + wait_for`
`flush()` does a synchronous LLM call that `wait_for` can't interrupt — on timeout the await cancels but the default-executor thread keeps running, and the executor's `shutdown(wait=True)` blocks worker exit. `flush_sync` instead runs `flush()` on its own daemon thread and blocks only on `Event.wait(timeout)`; the unbounded LLM call lives on the daemon thread, killed on process exit. Best-effort: on timeout, the unfinished tail is dropped (same failure direction as today, scoped to the tail). Note: steps (1) and (3) share the same `deadline`, so the timeout must cover both a slow in-flight worker and the remaining queue.

## Tests
- `backend/tests/test_memory_queue.py` — 7 new: empty-queue short-circuit, normal drain, timeout hard-stop, in-flight worker unfinished → `False`, in-flight finished in budget → `True`, flush raises → `False`, drain skips inter-item sleep.
- `backend/tests/test_gateway_lifespan_shutdown.py` — 3 new: shutdown calls `shutdown_flush(configured_timeout)` and logs `INFO` "completed"; `False` return logs `WARNING` (and **not** "completed", via caplog); `memory.enabled=False` skips the drain.
- `backend/tests/test_memory_manager_pluggable.py` — 3 new ABC-contract tests: `shutdown_flush` is on the ABC + noop returns `True`; DeerMem delegates to `queue.flush_sync(timeout)`; DeerMem end-to-end drains a pending batch.

## Validation
```
cd backend
uv run ruff check .          # All checks passed
uv run ruff format --check . # already formatted
uv run pytest tests/test_memory_queue.py -q                       # 21 passed
uv run pytest tests/test_gateway_lifespan_shutdown.py -q          # 5 passed
uv run pytest tests/test_memory_manager_pluggable.py -q           # 16 passed
uv run pytest tests/test_client.py::TestGatewayConformance -q     # passed
uv run pytest tests/blocking_io/ -q                               # 36 passed (no event-loop blocking introduced)
uv run pytest tests/test_harness_boundary.py -q                   # passed (no app.* imports in deerflow)
python ../scripts/scan_changed_blocking_io.py                    # No blocking-IO candidates introduced
```
Helm chart verified by render (not just inspection):
```
helm lint deploy/helm/deer-flow                  # 1 chart(s) linted, 0 chart(s) failed
helm template x deploy/helm/deer-flow            # terminationGracePeriodSeconds: 45 + preStop ["sleep","5"] render at the correct spec/container scope
helm template x deploy/helm/deer-flow --set gateway.terminationGracePeriodSeconds=90 --set gateway.preStopSleepSeconds=10   # overrides apply
helm template x deploy/helm/deer-flow --set gateway.preStopSleepSeconds=0      # preStop block correctly omitted
```
## End-to-end verification

Beyond the unit tests (which mock the LLM), the full SIGTERM path was verified end-to-end on a real Gateway against `main` (bug) vs this branch (fix), using identical conditions: isolated Gateway (`auth` disabled, `database: memory`, `memory.debounce_seconds: 300`), DeepSeek model, a single turn that states a memorable fact, then SIGTERM well inside the 300s debounce window.

| | `main` (bug) | this branch (fix) |
|---|---|---|
| shutdown log | `Waiting for application shutdown.` → `Application shutdown complete.` (no drain) | `Processing 1 queued memory updates` → `Invoking memory-update LLM` → `Memory saved` → `Memory queue flush completed within 30.0s` |
| `memory.json` | **absent** — the queued update is lost | **created** with the extracted fact |

`main` (bug) — memory enqueued, then SIGTERM, no drain, lost:
```
Memory update queued for thread …, queue size: 1
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
# memory.json never written
```

This branch (fix) — same enqueue + SIGTERM, the shutdown drain runs and persists:
```
Memory update queued for thread …, queue size: 1
INFO:     Waiting for application shutdown.
Processing 1 queued memory updates
Updating memory for thread … (trace_id=None)
Invoking memory-update LLM (thread=… trace_id=None)
Memory saved to …/users/default/memory.json
Memory updated successfully for thread …
Memory update batch done: 1 succeeded, 0 failed
Memory queue flush completed within 30.0s
INFO:     Application shutdown complete.
```

Extracted (real LLM, not mocked) into `memory.json`:
```json
"workContext": { "summary": "Uses Python 3.14 for all development and debugging work." },
"facts": [{ "content": "Actively uses Python 3.14 for all development and debugging tasks.", "category": "preference" }]
```

**Why this is conclusive:** with `debounce_seconds: 300`, the normal Timer fires ~5 minutes after enqueue. The drain on this branch ran ~1s after enqueue and persisted within ~3s — so it can only be the `shutdown_flush` path, not the debounce Timer. On `main` the same update is silently lost because the Gateway lifespan never drains the queue and the daemon Timer is killed on process exit.

## AI assistance

**How you used it:** Used Claude Code to re-apply the reviewer-approved shutdown-flush design onto the post-`#4122` `MemoryManager` abstraction (after #4122 invalidated the prior top-level-queue implementation), implement the ABC method + both backends + the Helm grace-period fix + the config/API surface, write the regression tests, and verify via the test suite and a real `helm template` render. I've read and take responsibility for every line.

### DIFF
--- a/backend/app/gateway/app.py
+++ b/backend/app/gateway/app.py
@@ -301,6 +301,43 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
             except Exception:
                 logger.exception("Failed to stop scheduled task service")
 
+        # Drain the memory backend's pending-update buffer before the worker
+        # exits (best-effort, bounded). IM channels and the scheduler are
+        # already stopped above, so no new IM/scheduler updates arrive during
+        # the drain; the LangGraph runtime / in-flight HTTP requests can still
+        # complete memory enqueues in a narrow window, but anything added after
+        # the drain copies the buffer only resets the debounce Timer
+        # (best-effort, same as today).
+        #
+        # No host-level pending/processing guard: ``shutdown_flush``
+        # short-circuits on a truly idle buffer (returns True immediately), so
+        # calling it unconditionally is cheap and keeps the in-flight-worker
+        # race entirely inside the backend (where the buffer lives) -- the host
+        # cannot "forget" that case the way a ``pending_count > 0``-only guard
+        # would (review #6 on the original PR).
+        #
+        # K8s caveat: ``shutdown_flush_timeout_seconds`` must fit inside the
+        # pod's ``terminationGracePeriodSeconds`` (channel stop + this drain +
+        # buffer), set on the gateway Helm deployment -- or K8s SIGKILLs the
+        # drain mid-flight and the loss this is fixing is silently re-introduced.
+        try:
+            app_cfg = get_app_config()
+            if app_cfg.memory.enabled:
+                from deerflow.agents.memory import get_memory_manager
+
+                manager = get_memory_manager()
+                flush_timeout = app_cfg.memory.shutdown_flush_timeout_seconds
+                completed = await asyncio.to_thread(manager.shutdown_flush, flush_timeout)
+                if completed:
+                    logger.info("Memory queue flush completed within %.1fs", flush_timeout)
+                else:
+                    logger.warning(
+                        "Memory queue flush did not finish within %.1fs; remaining updates may be lost",
+                        flush_timeout,
+                    )
+        except Exception:
+            logger.exception("Failed to flush memory queue on shutdown")
+
     logger.info("Shutting down API Gateway")
 
 
--- a/backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/queue.py
+++ b/backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/queue.py
@@ -47,6 +47,11 @@ class MemoryUpdateQueue:
         self._lock = threading.Lock()
         self._timer: threading.Timer | None = None
         self._processing = False
+        # Thread currently running ``_process_queue`` (None when idle). ``flush_sync``
+        # joins an in-flight worker instead of reporting a false-positive "completed"
+        # while contexts it already pulled out of the queue are still being processed
+        # (and would be lost on exit). See ``flush_sync`` step (1).
+        self._processing_thread: threading.Thread | None = None
         self._reprocess_pending = False
 
     @staticmethod
@@ -172,8 +177,15 @@ class MemoryUpdateQueue:
         self._timer.daemon = True
         self._timer.start()
 
-    def _process_queue(self) -> None:
-        """Process all queued conversation contexts."""
+    def _process_queue(self, *, skip_inter_item_delay: bool = False) -> None:
+        """Process all queued conversation contexts.
+
+        Args:
+            skip_inter_item_delay: When set, skip the inter-item rate-limit
+                ``time.sleep``. Intended for the shutdown-drain path
+                (:meth:`flush_sync`), which races a bounded timeout and should
+                not waste budget sleeping between items.
+        """
         with self._lock:
             if self._processing:
                 # Another worker is already draining the queue. Instead of
@@ -188,12 +200,15 @@ class MemoryUpdateQueue:
                 return
 
             self._processing = True
+            self._processing_thread = threading.current_thread()
             contexts_to_process = self._queue.copy()
             self._queue.clear()
             self._timer = None
 
         logger.info("Processing %d queued memory updates", len(contexts_to_process))
 
+        succeeded = 0
+        failed = 0
         try:
             for context in contexts_to_process:
                 try:
@@ -208,35 +223,122 @@ class MemoryUpdateQueue:
                         trace_id=context.trace_id,
                     )
                     if success:
+                        succeeded += 1
                         logger.info("Memory updated successfully for thread %s (trace_id=%s)", context.thread_id, context.trace_id)
                     else:
+                        failed += 1
                         logger.warning("Memory update skipped/failed for thread %s (trace_id=%s)", context.thread_id, context.trace_id)
                 except Exception as e:
+                    failed += 1
                     logger.error("Error updating memory for thread %s (trace_id=%s): %s", context.thread_id, context.trace_id, e)
 
-                # Small delay between updates to avoid rate limiting
-                if len(contexts_to_process) > 1:
+                # Small delay between updates to avoid rate limiting.
+                # Skipped on the shutdown-drain path, which races a bounded
+                # timeout and should spend that budget on LLM calls, not on
+                # sleeping between items.
+                if not skip_inter_item_delay and len(contexts_to_process) > 1:
                     time.sleep(0.5)
-
         finally:
+            # Summary count disambiguates "drained" (queue emptied) from "saved"
+            # (every extraction persisted): per-item ``update_memory`` failures are
+            # swallowed above, so without this an operator debugging missing
+            # memories would see only the happy-path "Processing N" line.
+            if succeeded or failed:
+                logger.info("Memory update batch done: %d succeeded, %d failed", succeeded, failed)
             with self._lock:
                 self._processing = False
+                self._processing_thread = None
                 if self._reprocess_pending:
                     self._reprocess_pending = False
                     if self._queue:
                         self._schedule_timer(0)
 
-    def flush(self) -> None:
+    def flush(self, *, skip_inter_item_delay: bool = False) -> None:
         """Force immediate processing of the queue.
 
         This is useful for testing or graceful shutdown.
+
+        Args:
+            skip_inter_item_delay: Forwarded to :meth:`_process_queue`; skip the
+                inter-item rate-limit sleep. Intended for the shutdown-drain
+                path (:meth:`flush_sync`).
         """
         with self._lock:
             if self._timer is not None:
                 self._timer.cancel()
                 self._timer = None
 
-        self._process_queue()
+        self._process_queue(skip_inter_item_delay=skip_inter_item_delay)
+
+    def flush_sync(self, timeout: float) -> bool:
+        """Best-effort synchronous flush bounded by ``timeout`` seconds.
+
+        Unlike :meth:`flush_nowait` (which only schedules a daemon timer that
+        is killed on process exit), this runs :meth:`flush` on a daemon thread
+        and waits up to ``timeout`` seconds for it to finish. Intended for
+        graceful shutdown: without it, any updates enqueued since the last
+        timer fire are lost on restart / rolling deploy / SIGTERM, because the
+        queue is pure in-memory and the debounce Timer is a daemon thread.
+
+        The drain accounts for two races a naive ``flush()`` would miss:
+
+        - **In-flight worker.** If the debounce Timer already fired, an
+          ``_process_queue`` worker is mid-LLM-call holding contexts it already
+          pulled out of the queue (``_processing=True``, queue empty). ``flush``
+          alone would see ``_processing=True``, no-op, and report success while
+          that worker is still running and likely killed on exit. So we join
+          the in-flight worker first (bounded by the remaining budget).
+        - **Failed flush.** ``flush`` makes a synchronous LLM call that can
+          raise; success is tracked on the happy path only, so the return value
+          matches the docstring's "completed".
+
+        Note: steps (1) and (3) share the same ``deadline`` budget. A slow
+        in-flight worker can consume most/all of it, leaving step (3) to no-op;
+        ``timeout`` must therefore cover both a slow in-flight worker *and* the
+        remaining queue (best-effort: any tail not drained in budget is dropped,
+        same failure direction as no flush, scoped to the tail).
+
+        Returns ``True`` only if the drain genuinely finished (queue empty, no
+        worker still running, flush did not raise) within ``timeout``.
+        """
+        deadline = time.monotonic() + timeout
+
+        # (1) Wait for an in-flight _process_queue first (bounded). Otherwise
+        # flush() would see _processing=True, no-op, and we would report
+        # success while that worker is still mid-LLM-call on a daemon thread
+        # that exit will kill — losing the contexts it already pulled out.
+        with self._lock:
+            in_flight = self._processing_thread
+        if in_flight is not None:
+            in_flight.join(timeout=max(0.0, deadline - time.monotonic()))
+
+        # (2) Genuine idle: nothing pending and no worker still running.
+        if self.pending_count == 0 and not self.is_processing:
+            return True
+
+        # (3) Drain the queue on a daemon thread so the timeout is a real hard
+        # stop: flush() makes a synchronous LLM call that cannot be
+        # interrupted, so we wait on Event.wait, not on Thread.join.
+        success = False
+        done = threading.Event()
+
+        def _run() -> None:
+            nonlocal success
+            try:
+                self.flush(skip_inter_item_delay=True)
+                success = True
+            except Exception:
+                logger.exception("Memory queue flush failed during shutdown drain")
+            finally:
+                done.set()
+
+        worker = threading.Thread(target=_run, name="memory-shutdown-flush", daemon=True)
+        worker.start()
+        finished = done.wait(timeout=max(0.0, deadline - time.monotonic()))
+        if not finished:
+            return False
+        # flush() returned; only report success if no worker raced back in.
+        return bool(success) and not self.is_processing
 
     def flush_nowait(self) -> None:
         """Start queue processing immediately in a background thread."""
@@ -256,6 +358,7 @@ class MemoryUpdateQueue:
                 self._timer = None
             self._queue.clear()
             self._processing = False
+            self._processing_thread = None
             self._reprocess_pending = False
 
     @property
--- a/backend/packages/harness/deerflow/config/memory_config.py
+++ b/backend/packages/harness/deerflow/config/memory_config.py
@@ -4,7 +4,8 @@ DeerMem-private fields live in ``backends/deermem/config.py`` (``DeerMemConfig``
 reached via ``backend_config`` (a dict the factory passes to the backend's
 ``__init__``). This module holds ONLY the host-shared fields every backend /
 call site / factory reads: ``enabled`` / ``injection_enabled`` /
-``manager_class`` / ``backend_config``. Keeping the shared schema slim is what
+``shutdown_flush_timeout_seconds`` / ``manager_class`` / ``backend_config``.
+Keeping the shared schema slim is what
 makes backends swappable and portable (DeerMem's knobs do not leak onto the
 shared contract).
 """
@@ -17,7 +18,7 @@ from pydantic import BaseModel, Field
 logger = logging.getLogger(__name__)
 
 # Host-shared MemoryConfig fields (read by every backend / call site / factory).
-_SHARED_FIELDS = frozenset({"enabled", "mode", "injection_enabled", "manager_class", "backend_config"})
+_SHARED_FIELDS = frozenset({"enabled", "mode", "injection_enabled", "shutdown_flush_timeout_seconds", "manager_class", "backend_config"})
 
 # DeerMem-private fields that used to live at the top level of `memory:` in
 # config.yaml (pre-abstraction). On load they are auto-migrated into
@@ -65,6 +66,23 @@ class MemoryConfig(BaseModel):
     injection_enabled: bool = Field(
         default=True,
         description="Whether to inject memory into the system prompt (call-site gate).",
+    )
+    shutdown_flush_timeout_seconds: float = Field(
+        default=30.0,
+        ge=1.0,
+        le=300.0,
+        description=(
+            "Hard time budget (seconds) for draining the memory backend's "
+            "pending-update buffer during Gateway graceful shutdown. The drain "
+            "makes one LLM call per pending item, so large IM batches may need "
+            "a higher value. Must fit inside the pod's K8s "
+            "terminationGracePeriodSeconds (together with channel/scheduler "
+            "stop) or K8s SIGKILLs the drain mid-flight. The drain runs on a "
+            "daemon thread, so on timeout the process proceeds to exit and any "
+            "unfinished tail is dropped (same failure direction as no flush, "
+            "scoped to the tail). Host-shared (not backend-private): the host "
+            "owns the lifespan budget and the K8s grace relationship."
+        ),
     )
     manager_class: str = Field(
         default="deermem",
--- a/backend/tests/test_gateway_lifespan_shutdown.py
+++ b/backend/tests/test_gateway_lifespan_shutdown.py
@@ -10,6 +10,7 @@ signal-reentrancy deadlock described in
 from __future__ import annotations
 
 import asyncio
+import logging
 from contextlib import asynccontextmanager
 from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock, patch
@@ -35,6 +36,9 @@ async def _run_lifespan_with_hanging_stop() -> float:
     app = FastAPI()
     startup_config = MagicMock()
     startup_config.log_level = "INFO"
+    # Keep this test focused on the channel-hang timing: skip the memory drain.
+    startup_config.memory.enabled = False
+    startup_config.memory.shutdown_flush_timeout_seconds = 5.0
     fake_service = MagicMock()
     fake_service.get_status = MagicMock(return_value={})
 
@@ -79,7 +83,7 @@ async def _run_lifespan_with_upload_staging_cleanup():
     from app.gateway.app import lifespan
 
     app = FastAPI()
-    startup_config = SimpleNamespace(log_level="INFO", memory=SimpleNamespace(token_counting="char"))
+    startup_config = SimpleNamespace(log_level="INFO", memory=SimpleNamespace(token_counting="char", enabled=False, shutdown_flush_timeout_seconds=30.0))
     fake_service = MagicMock()
     fake_service.get_status = MagicMock(return_value={})
     cleanup_upload_staging_files = MagicMock(return_value=2)
@@ -110,3 +114,78 @@ def test_lifespan_sweeps_upload_staging_files_on_startup():
     cleanup_upload_staging_files.assert_called_once_with()
     close_oidc_service.assert_awaited_once()
     stop_channel_service.assert_awaited_once()
+
+
+async def _run_lifespan_with_memory_flush(*, enabled: bool, flush_return: bool) -> MagicMock:
+    """Drive lifespan with a spied memory manager.shutdown_flush.
+
+    Returns the manager mock so the caller can assert the shutdown flush was
+    reached (and with what timeout). The host calls ``shutdown_flush``
+    unconditionally when memory is enabled -- there is no host-level
+    ``pending_count/is_processing`` gate, because the backend short-circuits on
+    an idle buffer and keeping the in-flight race inside the backend means the
+    host cannot "forget" it (review #6 on the original PR).
+    """
+    from app.gateway.app import lifespan
+
+    app = FastAPI()
+    startup_config = SimpleNamespace(
+        log_level="INFO",
+        memory=SimpleNamespace(
+            token_counting="char",
+            enabled=enabled,
+            shutdown_flush_timeout_seconds=5.0,
+        ),
+    )
+    fake_service = MagicMock()
+    fake_service.get_status = MagicMock(return_value={})
+    close_oidc_service = AsyncMock()
+    stop_channel_service = AsyncMock()
+
+    async def fake_start(_startup_config):
+        return fake_service
+
+    manager = MagicMock()
+    manager.shutdown_flush.return_value = flush_return
+
+    with (
+        patch("app.gateway.app.get_app_config", return_value=startup_config),
+        patch("app.gateway.app.get_gateway_config", return_value=MagicMock(host="x", port=0)),
+        patch("app.gateway.app.langgraph_runtime", _noop_langgraph_runtime),
+        patch("app.gateway.app.auth.close_oidc_service", close_oidc_service),
+        patch("app.channels.service.start_channel_service", side_effect=fake_start),
+        patch("app.channels.service.stop_channel_service", stop_channel_service),
+        patch("deerflow.agents.memory.get_memory_manager", return_value=manager),
+    ):
+        async with lifespan(app):
+            pass
+
+    return manager
+
+
+def test_lifespan_drains_memory_on_shutdown_with_configured_timeout(caplog) -> None:
+    """When memory is enabled, shutdown calls manager.shutdown_flush with the
+    configured timeout (asserts the timeout is forwarded, review #3) and logs
+    'completed' at INFO when the drain finishes."""
+    caplog.set_level(logging.INFO, logger="app.gateway.app")
+    manager = asyncio.run(_run_lifespan_with_memory_flush(enabled=True, flush_return=True))
+    manager.shutdown_flush.assert_called_once_with(5.0)
+    assert any(r.levelno == logging.INFO and "flush completed" in r.message for r in caplog.records)
+
+
+def test_lifespan_warns_when_memory_flush_does_not_finish(caplog) -> None:
+    """A False return (timeout/failure) is the path operators actually see when
+    K8s SIGKILLs the drain; the host must log a WARNING (not 'completed'), so
+    the loss risk is visible (review #3 False-branch coverage; review #2/#4
+    failed-flush semantics)."""
+    caplog.set_level(logging.WARNING, logger="app.gateway.app")
+    manager = asyncio.run(_run_lifespan_with_memory_flush(enabled=True, flush_return=False))
+    manager.shutdown_flush.assert_called_once_with(5.0)
+    assert any(r.levelno == logging.WARNING and "did not finish" in r.message for r in caplog.records)
+    assert not any("flush completed" in r.message for r in caplog.records)
+
+
+def test_lifespan_skips_memory_flush_when_disabled() -> None:
+    """memory.enabled=False skips the drain entirely."""
+    manager = asyncio.run(_run_lifespan_with_memory_flush(enabled=False, flush_return=True))
+    manager.shutdown_flush.assert_not_called()
--- a/backend/tests/test_memory_queue.py
+++ b/backend/tests/test_memory_queue.py
@@ -247,3 +247,159 @@ def test_process_queue_forwards_trace_id_to_updater() -> None:
         user_id=None,
         trace_id="trace-memory-1",
     )
+
+
+# ---------------------------------------------------------------------------
+# shutdown_flush / flush_sync (graceful-shutdown drain) — review carry-overs.
+# The queue is a daemon-timer + in-memory buffer, so anything pending at
+# process exit is lost. flush_sync drains it within a hard timeout, joining an
+# in-flight worker first so contexts a debounce Timer already pulled out of the
+# queue are not lost either.
+# ---------------------------------------------------------------------------
+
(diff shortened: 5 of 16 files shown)


backend/packages/harness/deerflow/agents/memory/queue.py
```diff
@@ -254,6 +254,38 @@ def flush(self) -> None:
...
+            try:
+                self.flush()
+            except Exception:
+                logger.exception("Memory queue flush failed during shutdown drain")
+            finally:
+                done.set()
+
+        worker = threading.Thread(target=_run, name="memory-shutdown-flush", daemon=True)
+        worker.start()
+        return done.wait(timeout)
```
willem-bd: **`True` here can mean "nothing was drained"** - the in-flight `_process_queue` isn't waited on.

If the debounce Timer already fired, `_process_queue` set `_processing=True`, copied `_queue` into `contexts_to_process`, and cleared `_queue` (queue.py:197-198). Now either `pending_count == 0` short-circuits at line 273, or `flush()` -> `_process_queue()` sees `_processing=True`, sets `_reprocess_pending=True`, and returns immediately (queue.py:184-191). Either way `flush_sync` returns `True` while the in-flight worker is still mid-LLM-call on a daemon Timer thread - likely killed on exit, losing the contexts it already pulled out. The lifespan then logs `"Memory queue flush completed"` (false-positive).

I know joining in-flight `_process_queue` is flagged as follow-up; for the MVP consider checking `is_processing` here and joining the processing thread up to the remaining budget (needs `threading.current_thread()` stashed in `_process_queue`). The three new tests all start from an idle queue - a non-idle/in-flight case would cover this.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/queue.py
```diff
@@ -254,6 +254,38 @@ def flush(self) -> None:
...
+            return True
+        done = threading.Event()
+
+        def _run() -> None:
+            try:
+                self.flush()
+            except Exception:
+                logger.exception("Memory queue flush failed during shutdown drain")
+            finally:
+                done.set()
```
willem-bd, line 333: `done.set()` in `finally` fires whether or not `flush()` raised, so `flush_sync` returns `True` for a *failed* flush too. The lifespan then logs `"Memory queue flush completed"` (info) alongside the exception (error) - conflicting signals.

Consider tracking success on the happy path only (e.g. a flag set before `done.set()`) so the return matches the docstring's "completed" and the caller can distinguish.

backend/app/gateway/app.py
```diff
@@ -57,6 +57,15 @@
...
 _SHUTDOWN_HOOK_TIMEOUT_SECONDS = 5.0
 
+# Budget for draining the in-memory MemoryUpdateQueue on graceful shutdown.
+# Each pending item performs one LLM call (memory extraction), which alone can
+# take several seconds — the 5s channel-shutdown budget above is too tight.
+# The drain runs on a daemon thread (see MemoryUpdateQueue.flush_sync) so this
+# is a real hard stop: on timeout the process proceeds to exit and any
+# unfinished update is dropped, same as the prior no-flush behaviour but scoped
+# to the tail of a long queue rather than the whole queue.
+_MEMORY_FLUSH_TIMEOUT_SECONDS = 30.0
```
willem-bd: This is a module constant, so it can't be tuned per deployment. With `debounce_seconds: 30` (default) and each item doing one LLM call + a `time.sleep(0.5)` between items (queue.py:234-235), 30s covers only ~5 items - and IM-heavy deployments (the hardest-hit case per the PR description) can exceed that in a single batch.

Worth exposing via `config.yaml` next to `memory.debounce_seconds`, given the `risk:high` label.

backend/app/gateway/app.py
```diff
@@ -297,6 +306,30 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
             except Exception:
                 logger.exception("Failed to stop scheduled task service")
 
+        # Drain the in-memory MemoryUpdateQueue before the worker exits. The
+        # queue debounces (default 30s) and processes on a daemon Timer, so
+        # without an explicit flush any updates enqueued since the last timer
+        # fire are lost on restart / rolling deploy / SIGTERM. Channels and the
+        # scheduler are already stopped above, so no new updates can arrive
```
willem-bd: Minor: this reads as absolute, but it only covers IM channels + the scheduler. The LangGraph runtime / in-flight HTTP requests can still complete `MemoryMiddleware.after_agent` during the drain - an item enqueued after `flush()` copies the queue gets a fresh daemon Timer that's killed on exit. Narrow window, but maybe soften to "no new IM/scheduler updates"?

willem-bd (commented):
Reviewed the memory-queue shutdown flush. Real bug, and the daemon-thread + `Event.wait(timeout)` design is the right call - the rationale for why `asyncio.to_thread(flush) + wait_for` can't hard-cancel the synchronous LLM call is correct.

Inline comments cover: (1) the in-flight `_process_queue` race that makes `"Memory queue flush completed"` false-positive, (2) `flush_sync` returning `True` on a failed flush, (3) the 30s budget not being configurable, (4) a comment that slightly overstates "no new updates can arrive".

One more, not anchored because it's pre-existing unchanged code: the inter-item `time.sleep(0.5)` in `_process_queue` (queue.py:234-235) is pure overhead during a shutdown drain that's racing the 30s budget - skipping it on the `flush_sync` path would let the budget cover more items.

Net: safe to merge for the common case (queued-but-undrained items at shutdown). The in-flight race (#1) is the one I'd want addressed or accurately logged before this ships under `risk:high`.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/queue.py
TuNaiChao: Thanks — fixed in [1ad248ea](https://github.com/bytedance/deer-flow/commit/1ad248ea). done.set() now only signals "stop waiting"; a separate success flag is set on the happy path (after flush() returns without raising), so flush_sync returns False when flush() raises and the lifespan no longer logs "completed" alongside the exception.

Covered by test_flush_sync_returns_false_when_flush_raises, which patches flush to raise and asserts the return is False.

backend/app/gateway/app.py
```diff
@@ -297,6 +297,38 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
...
+        # flush_sync runs flush() on a daemon thread so the timeout is a real
+        # hard stop (see MemoryUpdateQueue.flush_sync). The timeout budget is
+        # configurable via memory.shutdown_flush_timeout_seconds because each
+        # pending item performs one LLM call and large IM batches may need more
+        # than the default.
+        try:
+            from deerflow.agents.memory.queue import get_memory_queue
+
+            queue = get_memory_queue()
+            if queue.pending_count > 0:
```
willem-bd: This guard skips `flush_sync` when `pending_count == 0` - which is exactly the in-flight-worker case that `flush_sync` step (1) + `_processing_thread` were added to handle.

When the debounce Timer has already fired, `_process_queue` has done `contexts_to_process = self._queue.copy(); self._queue.clear()` and is mid-LLM-call: at that moment `pending_count == 0` and `is_processing == True`. This guard is False, so `flush_sync` is never called, the in-flight daemon worker is never joined, and it's killed on exit - losing the contexts it already pulled out.

`flush_sync`'s own docstring calls this out as the target race ("If the debounce Timer already fired, an `_process_queue` worker is mid-LLM-call holding contexts it already pulled out of the queue... So we join the in-flight worker first") - but that join only runs when `pending_count > 0` AND a worker is in flight. The pure in-flight case (`pending_count == 0`) - the documented target - is missed, so `_processing_thread` / step (1) is effectively dead for that scenario in production.

Fix: broaden the guard (and log) to also cover an in-flight worker:

    if queue.pending_count > 0 or queue.is_processing:
        flush_timeout = get_app_config().memory.shutdown_flush_timeout_seconds
        logger.info("Flushing %d pending memory update(s) on shutdown (in_flight=%s)",
                    queue.pending_count, queue.is_processing)
        ...

`is_processing` is already a public property, and `flush_sync` short-circuits on truly-idle (step 2), so this is safe.

backend/tests/test_memory_queue.py
```diff
@@ -455,3 +455,177 @@ def test_process_queue_restores_outer_contextvar_after_return(self) -> None:
...
+
+    thread = threading.Thread(target=_inflight, name="fake-inflight-worker", daemon=True)
+    thread.start()
+    # Wait until the fake worker has claimed _processing.
+    while not queue.is_processing:
+        time.sleep(0.005)
+    return thread
+
+
+def test_flush_sync_waits_for_inflight_worker_and_returns_false_if_unfinished() -> None:
```
willem-bd, line 341: This test proves `flush_sync` handles the in-flight-worker case correctly (returns False when the worker doesn't finish in budget) - but it calls `queue.flush_sync(...)` directly, bypassing the lifespan guard in `app/gateway/app.py`.

That guard (`if queue.pending_count > 0`) skips `flush_sync` when the queue is empty - i.e. the in-flight-only case this test sets up (`pending_count == 0`, `is_processing == True`). So the regression this test pins at the `flush_sync` level isn't actually exercised through the lifespan in production. A lifespan-level test (drive the shutdown path with an in-flight worker and `pending_count == 0`) would close the gap and catch the guard fix above if it regresses.

willem-bd (commented):
Reviewed against PR head (`4597551`) and the current codebase; ran the PR's tests against the PR's code (25 passed - 18 existing + 7 new). **Verdict: approve with one fix.**

`flush_sync`'s design is sound (traced all branches): the daemon-thread + `Event.wait` gives a true hard timeout for an uninterruptible sync LLM call, step (1) joins an in-flight worker with the remaining budget, `return bool(success) and not self.is_processing` correctly avoids false-success, and `skip_inter_item_delay` cleanly separates the drain path from the normal debounce path without changing defaults. The `asyncio.to_thread + wait_for` alternative was correctly ruled out. `config_version` bump is non-breaking (stale version only warns; the new field has a default). Memory-disabled path is a clean no-op.

The one substantive finding (inline on `app/gateway/app.py:318`): the lifespan guard `if queue.pending_count > 0` skips `flush_sync` when the queue is empty - precisely the in-flight-worker case (`pending_count == 0`, `is_processing == True`) that `_processing_thread` / `flush_sync` step (1) were added to handle. Fix: broaden the guard to `pending_count > 0 or is_processing`. A second comment on `test_memory_queue.py` notes the matching coverage gap (the in-flight test calls `flush_sync` directly, bypassing the lifespan guard).

- Minor: the PR body is stale - it says "three regression tests" / "21 passed" but the diff adds 7 tests (25 passed). The 4 extra tests reference "review comment #1/#2/#5" from a prior revision; the Testing/Validation sections weren't updated.

TuNaiChao:
> Reviewed against PR head (`4597551`) and the current codebase; ran the PR's tests against the PR's code (25 passed - 18 existing + 7 new). **Verdict: approve with one fix.**
> 
> `flush_sync`'s design is sound (traced all branches): the daemon-thread + `Event.wait` gives a true hard timeout for an uninterruptible sync LLM call, step (1) joins an in-flight worker with the remaining budget, `return bool(success) and not self.is_processing` correctly avoids false-success, and `skip_inter_item_delay` cleanly separates the drain path from the normal debounce path without changing defaults. The `asyncio.to_thread + wait_for` alternative was correctly ruled out. `config_version` bump is non-breaking (stale version only warns; the new field has a default). Memory-disabled path is a clean no-op.
> 
> The one substantive finding (inline on `app/gateway/app.py:318`): the lifespan guard `if queue.pending_count > 0` skips `flush_sync` when the queue is empty - precisely the in-flight-worker case (`pending_count == 0`, `is_processing == True`) that `_processing_thread` / `flush_sync` step (1) were added to handle. Fix: broaden the guard to `pending_count > 0 or is_processing`. A second comment on `test_memory_queue.py` notes the matching coverage gap (the in-flight test calls `flush_sync` directly, bypassing the lifespan guard).
> 
> * Minor: the PR body is stale - it says "three regression tests" / "21 passed" but the diff adds 7 tests (25 passed). The 4 extra tests reference "review comment #1/#2/#5" from a prior revision; the Testing/Validation sections weren't updated.

Thanks @willem-bd for the review — both the guard and the matching lifespan coverage are fixed in cf55fa26; mind taking another look when you have a moment?

WillemJiang:
@TuNaiChao  There are major refactorings on the memory side https://github.com/bytedance/deer-flow/pull/4122; you may need to reapply the patch to main once it's done.

TuNaiChao:
> @TuNaiChao There are major refactorings on the memory side #4122; you may need to reapply the patch to main once it's done.

Thanks for the heads-up! I'll rebase onto the latest main and rework the shutdown flush against the new MemoryManager abstraction once #4122 lands. Will update here.

backend/app/gateway/app.py
```diff
@@ -297,6 +297,50 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
...
+            # handle. flush_sync short-circuits on a truly idle queue, so
+            # calling it here is safe.
+            if queue.pending_count > 0 or queue.is_processing:
+                flush_timeout = get_app_config().memory.shutdown_flush_timeout_seconds
+                logger.info(
+                    "Flushing %d pending memory update(s) on shutdown (in_flight=%s)",
+                    queue.pending_count,
+                    queue.is_processing,
+                )
+                completed = await asyncio.to_thread(queue.flush_sync, flush_timeout)
```
willem-bd: **K8s grace period will SIGKILL this drain in the default Helm deployment.** The default `shutdown_flush_timeout_seconds` (30s) is consumed *after* the 5s channel stop (`_SHUTDOWN_HOOK_TIMEOUT_SECONDS`) and before uvicorn drains in-flight HTTP - already ~35s of shutdown work. But the Gateway Helm deployment sets neither `terminationGracePeriodSeconds` nor a `preStop` hook (`deploy/helm/deer-flow/templates/gateway-deployment.yaml`), so K8s defaults to a 30s grace period and SIGKILLs the pod mid-drain. With default Helm settings and any non-trivial batch, this silently re-introduces the exact loss the PR is fixing. The e2e validation in the PR body used a bare-process SIGTERM (no K8s grace budget), so it won't surface this. Worth at least one of: set `terminationGracePeriodSeconds` on the gateway pod (flush_timeout + channel_timeout + buffer), add a `preStop` hook, or document that operators must raise the grace period if they raise the timeout.

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -36,6 +36,20 @@ class MemoryConfig(BaseModel):
         le=300,
         description="Seconds to wait before processing queued updates (debounce)",
     )
+    shutdown_flush_timeout_seconds: float = Field(
```
willem-bd: Sibling `debounce_seconds` is exposed to the frontend via `MemoryConfigResponse` (`backend/app/gateway/routers/memory.py`, the `/memory/config` + `/memory/status` endpoints), but this new field isn't. If the settings UI lets operators tune `debounce_seconds`, the shutdown budget is a natural companion - especially given the K8s grace-period interaction. No functional break (the response is built field-by-field, so omitting it won't error), but worth a deliberate decision: is this intentionally config-only, or should it join `debounce_seconds` on the API surface?

backend/tests/test_gateway_lifespan_shutdown.py
```diff
@@ -110,3 +110,60 @@ def test_lifespan_sweeps_upload_staging_files_on_startup():
...
+    """The shutdown guard must call flush_sync when the queue is empty but a
+    worker is in flight (pending_count == 0, is_processing == True) — the
+    documented target race. Gating on ``pending_count > 0`` alone skips
+    flush_sync there, so the in-flight worker is never joined and is killed on
+    exit, losing the contexts it already pulled out (review follow-up)."""
+    mock_queue = asyncio.run(_run_lifespan_with_inflight_memory_worker())
+
+    # The guard reached flush_sync despite pending_count == 0, because an
+    # in-flight worker was running.
+    mock_queue.flush_sync.assert_called_once()
```
willem-bd: This proves the guard reaches `flush_sync` with `pending_count==0, is_processing==True` (the documented target race) - good. Two small gaps: it asserts `assert_called_once()` but not that the timeout was forwarded (`assert_called_once_with(5.0)`), and it doesn't cover the result branching (`True` -> `"Memory queue flush completed"` vs `False` -> `"did not finish within..."`). The `False` path is the one operators will actually see in logs when the drain is SIGKILLed (see the app.py comment), so a lifespan-level `flush_sync.return_value = False` variant would be the higher-value addition.

backend/app/gateway/app.py
```diff
@@ -297,6 +297,50 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
...
+                if completed:
+                    logger.info("Memory queue flush completed")
```
willem-bd: `completed=True` means the drain *finished* (queue empty, no worker running, no raise) - not that every memory was persisted. Per-item `update_memory` failures are swallowed inside `_process_queue` (`logger.error`/`warning`, then `continue`), so this log can read as success while individual extractions failed. Consistent with existing `_process_queue` semantics, but for an operator debugging missing memories, a debug-level succeeded/failed count alongside this line would disambiguate "drained" from "saved".

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/queue.py
```diff
@@ -230,29 +243,106 @@ def _process_queue(self) -> None:
...
+        deadline = time.monotonic() + timeout
+
+        # (1) Wait for an in-flight _process_queue first (bounded). Otherwise
+        # flush() would see _processing=True, no-op, and we would report
+        # success while that worker is still mid-LLM-call on a daemon thread
+        # that exit will kill — losing the contexts it already pulled out.
+        with self._lock:
+            in_flight = self._processing_thread
+        if in_flight is not None:
+            in_flight.join(timeout=max(0.0, deadline - time.monotonic()))
```
willem-bd, line 313: Minor: the join in step (1) and the drain in step (3) draw from the same `deadline`. If the in-flight worker is processing a large batch, the join can consume most/all of the budget; step (3) then no-ops (`_processing=True` -> `_reprocess_pending`) and any items added during that run are lost. This is the documented best-effort tail, but a one-line note that the budget is shared (so `shutdown_flush_timeout_seconds` must cover both a slow in-flight worker *and* the remaining queue) would help operators size it.

TuNaiChao:
Reworked onto MemoryManager.shutdown_flush (36918b90) — please re-review.

WillemJiang (approved):

PR #4184, opened 2026-07-14
feat(agents): let custom agents opt out of memory via config.yaml

### LINKED ISSUE #3626: [feat] 允许关闭Custom Agent记忆
### Before you start

### Problem / motivation

如题。

有些场景下，Custom Agent就是无情的执行机器，他不需要有自己的记忆。

特指： Per-user lead-agent 形态的 custom agent（users/{user_id}/agents/{name}/config.yaml + SOUL.md）

### Proposed solution

在Custom Agent 的config.yml 文件里允许配置是否开启记忆。

### Affected area(s)

### Alternatives considered

_No response_

### Additional context

_No response_

he-yufeng:
Fixes #3626

## Why

Issue #3626 (maintainer-filed) asks for a way to turn a custom agent's memory off from its own config. Some custom agents — the per-user lead agents at `users/{user_id}/agents/{name}/` — are meant to be stateless execution machines: they shouldn't accumulate memory of their own, even on a deployment that has memory enabled globally. Today the only lever is the global `memory.enabled`, which is all-or-nothing for the whole deployment. Scope was aligned up front on #3626 before writing code.

## What changed

A custom agent can now opt out of memory by hand-authoring a `memory:` block in its `config.yaml`:

```yaml
# users/{user_id}/agents/stateless-worker/config.yaml
name: stateless-worker
model: fast-model
memory:
  enabled: false   # this agent runs without memory
```

`AgentConfig.memory` defaults to `None`, so every existing agent keeps inheriting the global memory config and nothing changes for them.

The override is deliberately narrow: an agent can only turn memory *off* — it can't re-enable memory that the operator disabled globally, so the global `memory.enabled` stays the master kill switch (e.g. a deployment that disables memory for compliance shouldn't be overridable per-agent). When an agent opts out it gets no memory at all: no capture, no `memory_*` tools, and no `<memory>` block injected into its prompt.

## Surface area

- [x] **Agents / LangGraph** — custom-agent config schema + the memory gating in the lead-agent factory

Not a default-behavior change: existing agents are untouched unless they explicitly add the block.

## Implementation notes (for reviewers)

Memory isn't gated in one place — it's read at four: `MemoryMiddleware` capture, the `memory_*` tools, the `<memory>` injection in `DynamicContextMiddleware`, and the summarization memory-flush hook. Instead of threading a per-agent flag through each of those, `apply_agent_memory_override` folds the opt-out into the effective `AppConfig` once, right after `load_agent_config`. Everything downstream keeps reading `app_config.memory` and just sees the resolved value, so there's no new flag to plumb and no gate that can drift. `disabled_memory_config` clears both the write side (`enabled`) and the read side (`injection_enabled`) so a disabled agent neither records nor recalls memory.

Like the existing `github:` block, `memory` is a hand-authored field kept out of `MANAGED_AGENT_CONFIG_FIELDS`, so `preserve_non_managed_fields` carries it through `update_agent` / `PATCH /api/agents/{name}` rewrites instead of dropping it.

## Validation

From `backend/`:

- `ruff format --check` and `ruff check` on the changed files — both clean.
- `pytest` on the new `tests/test_agent_memory_toggle.py` (15 tests) plus the touched-surface suites (agent-config, lead-agent factory + skills, update-agent + e2e, memory prompt-injection / tools / queue / consolidation, durable-context, client, subagent-description) — all green, ~400 tests total. I did not run the full `make test` suite locally, but the areas this change touches are covered above.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Used it to map the memory-gating call sites, work out the "fold the override into the effective AppConfig once" approach, and write the implementation, tests, and docs. I reviewed the design and every line, confirmed the four gates it covers, and ran the checks above.


backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -455,6 +480,14 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
 
     agent_config = load_agent_config(agent_name) if not is_bootstrap else None
     available_skills = _available_skill_names(agent_config, is_bootstrap)
+
+    # Fold the custom agent's memory opt-out (issue #3626) into the effective app
+    # config before anything reads it, so a custom agent with
+    # ``memory: {enabled: false}`` gets no memory capture, tools, or injection.
+    # A no-op for every other agent (returns the same object), so behavior is
+    # unchanged unless an agent explicitly disables memory.
+    resolved_app_config = apply_agent_memory_override(resolved_app_config, agent_config)
```
willem-bd, line 960: This override only takes effect on the `make_lead_agent` -> `_make_lead_agent` (LangGraph runtime) path. The embedded `DeerFlowClient` (client.py:289) calls `build_middlewares` directly with `self._app_config` (the global `get_app_config()`, client.py:173) and never calls `load_agent_config`, so it neither reads nor honors a `memory:` block.

That's a pre-existing limitation - DeerFlowClient doesn't load any per-agent `config.yaml` (not `memory`, `model`, or `tool_groups`) - not a regression, but it means an operator who sets `memory: {enabled: false}` gets the opt-out only when the agent runs through the runtime. Worth a one-line doc note clarifying the scope, or a follow-up that mirrors `load_agent_config` + `apply_agent_memory_override` into the embedded path if it's expected to honor it.

backend/tests/test_agent_memory_toggle.py
```diff
@@ -0,0 +1,191 @@
...
+# apply_agent_memory_override — the fold into the effective app config
+# --------------------------------------------------------------------------- #
+def _app_config(*, enabled: bool = True, injection: bool = True, mode: str = "middleware") -> AppConfig:
+    return AppConfig(
+        sandbox=SandboxConfig(use="deerflow.sandbox.local:LocalSandboxProvider"),
+        memory=MemoryConfig(enabled=enabled, injection_enabled=injection, mode=mode),
+    )
+
+
+def test_override_disables_memory_for_opted_out_agent() -> None:
```
willem-bd, line 140: The override tests here verify `apply_agent_memory_override` / `disabled_memory_config` in isolation, but none drive `_make_lead_agent` / `build_middlewares` end-to-end to assert a disabled-memory agent actually gets no MemoryMiddleware, no memory tools, no `<memory>` injection, and no flush hook.

For a `risk:high` change, the four gates are currently held together by convention (all read the passed `app_config`) rather than by a test, so a future refactor could silently break a gate without anything going red. An integration test through `build_middlewares` (or `_make_lead_agent`) with a disabled-memory agent config would lock in the wiring.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -81,6 +81,31 @@ def _append_memory_tools_without_name_conflicts(tools: list) -> None:
...
+
+    When a custom agent sets ``memory: {enabled: false}`` in its ``config.yaml``,
+    return a shallow copy of ``app_config`` whose ``memory`` section is switched
+    off (see :func:`deerflow.config.memory_config.disabled_memory_config`).
+    Otherwise return ``app_config`` unchanged — the *same* object, no copy — so
+    every existing agent (and any agent that leaves memory on) is unaffected.
+
+    Resolving the override once, here, means the four downstream memory gates all
+    read the already-resolved ``app_config.memory`` and need no per-gate flag:
+    MemoryMiddleware capture, the ``memory_*`` tools, the ``<memory>`` injection
```
willem-bd, line 216: The "fold once" design works only while every memory gate reads the *passed* `app_config.memory`. Worth calling out the invariant here: a future memory gate that reads the global `get_memory_config()` / `get_app_config()` directly would bypass this opt-out.

The flush hook is the existing edge case - `memory_flush_hook` reads the global `get_memory_config()` internally (summarization_hook.py:14), and is only safe because `create_summarization_middleware` skips *registering* it when `app_config.memory.enabled` is false (summarization_middleware.py:495). A one-line note ("new gates must read the passed `app_config.memory`, not the global accessors") would protect this for the next contributor.

backend/packages/harness/deerflow/config/agents_config.py
```diff
@@ -156,6 +156,36 @@ def validate_agent_name(name: str | None) -> str | None:
...
+    The override is deliberately *narrow*: an agent may only turn memory
+    **off**. It cannot force memory **on** when the operator disabled it
+    globally in the main ``config.yaml`` (``memory.enabled: false``) — the
+    global switch stays the master kill switch (e.g. a deployment that disables
+    memory for compliance). The resolution rule lives in
+    ``deerflow.config.memory_config.disabled_memory_config`` and the call site
+    is ``deerflow.agents.lead_agent.agent.apply_agent_memory_override``.
+    """
+
+    enabled: bool = Field(
```
willem-bd, line 177: `AgentMemoryConfig` uses Pydantic's default `extra="ignore"`, so `memory: {enabled: true, debounce_seconds: 60}` silently drops `debounce_seconds`. Harmless in practice (`enabled: true` is an identity no-op; `enabled: false` is fully off so other fields are irrelevant), but could surprise a user expecting per-agent tuning.

Either document that only `enabled` is honored, or set `extra="forbid"` to fail fast on a typo'd / unrecognized field.

willem-bd (commented):
Reviewed the per-agent memory opt-out. Verified the override (applied at agent.py:489) actually flows into all four claimed memory gates via `resolved_app_config`:

1. MemoryMiddleware capture - `after_agent` no-ops on `if not config.enabled` (memory_middleware.py:65), constructed with the overridden `memory_config`.
2. `memory_*` tools - `should_use_memory_tools(resolved_app_config.memory)` (agent.py:578/644).
3. `<memory>` injection - `DynamicContextMiddleware` reads `app_config.memory.injection_enabled` (dynamic_context_middleware.py:158).
4. Summarization flush hook - `create_summarization_middleware` skips *registering* `memory_flush_hook` when `app_config.memory.enabled` is false (summarization_middleware.py:495).

The "can-only-disable" semantics is correctly enforced, and the identity (no-copy) path means no overhead for the common case. Good unit coverage of the override logic, plus the AGENTS.md doc update.

Inline notes below are all follow-up: the DeerFlowClient embedded path isn't covered (pre-existing - it doesn't load agent configs), no integration test through the factory, an invariant note for future memory gates, and a nit on `AgentMemoryConfig` silently ignoring extra fields.

Net: approve. The DeerFlowClient scope (#1) is the main thing to clarify; an integration test (#2) would lock in the wiring for a `risk:high` change.

he-yufeng:
Thanks for the thorough pass. On the main point (#1): the embedded `DeerFlowClient` path is out of scope here on purpose — it builds middlewares from the global `get_app_config()` and never loads per-agent configs, so there's no agent-level memory opt-out for it to honor yet; wiring that is a separate change from the factory opt-out this PR adds. On #2, happy to add a factory-level integration test that asserts a disabled-memory agent actually gets no `MemoryMiddleware` / memory tools / `<memory>` injection — that's the right thing to lock in for a `risk:high` change; I'll push it as a follow-up commit here rather than leaving it to a separate PR. #3 (the invariant note at agent.py:95) and #4 (`AgentMemoryConfig` `extra="ignore"`) are both worth doing — I'll fold the invariant comment in with the test commit, and can switch the config to `extra="forbid"` if you'd prefer a typo in `memory:` to fail loudly rather than silently drop.

he-yufeng:
Pushed 089b4fef with both. One clarification from writing the test: a disabled-memory agent doesn't get *no* MemoryMiddleware — the chain still appends it, but with the resolved disabled config, and `MemoryMiddleware` self-gates (`if not config.enabled: return None`), so it's present-but-inert. That's the existing pattern (the middleware always takes a `memory_config` and short-circuits), so the opt-out rides it rather than changing chain assembly. The integration test asserts exactly that fold-through: the `MemoryMiddleware` instance carries `enabled=True` for a normal agent and `enabled=False` for an opted-out one, plus `injection_enabled=False` on the read side. Invariant comment for #3 is in the same commit; on #4 I left `extra="ignore"` for now but can switch `AgentMemoryConfig` to `extra="forbid"` in a follow-up if you'd rather a typo under `memory:` fail loudly.

fancyboi999 (changes requested):
Thanks @he-yufeng. I found one current-main compatibility issue that should be fixed before this is ready.

[P1] Update the preservation test for the current MemoryConfig schema

- Location: `backend/tests/test_agent_memory_toggle.py:119-124`
- Problem: this new test constructs and reads `MemoryConfig.debounce_seconds`, but latest `main` no longer exposes that field. Once the PR is tested against the current base, `test_disabled_memory_config_turns_off_both_sides` raises `AttributeError`, so the backend unit-test gate cannot pass.
- Evidence: current `origin/main` (`693507870cae26dadf3af487f811eb2bcfe18f87`) moved DeerMem-only settings such as `debounce_seconds` under `backend_config`; the host-shared `MemoryConfig` now contains `enabled`, `mode`, `injection_enabled`, `shutdown_flush_timeout_seconds`, `manager_class`, and `backend_config`. Constructing it with `debounce_seconds=42` ignores that extra input, and the subsequent `off.debounce_seconds` access fails. This head is 37 commits behind current main and its new head exposes only triage/CLA checks, so the old green run does not cover the current schema.
- Suggested fix: rebase on current `main` and use a field that still belongs to the shared schema (for example `shutdown_flush_timeout_seconds`) or a value inside `backend_config` to verify that `disabled_memory_config()` preserves unrelated settings.
- Test: run `cd backend && uv run pytest tests/test_agent_memory_toggle.py -v`, then the normal backend unit-test workflow against the rebased head.


Reviewed head SHA: `089b4fef4776f0c0bd429313b57e61b93cc8023b`

WillemJiang:
@he-yufeng please take a look at the failed unit test.

willem-bd:
One more nit I couldn't line-anchor because the line isn't in this PR's diff — flagging here.

**`agent.py:355-356`** — the `logger.warning("memory.mode is 'tool' but memory.enabled is false; memory tools will not be registered.")` now fires on **every run** of an agent that opts out of memory when the global `mode` is `"tool"`. `disabled_memory_config` preserves `mode` (it only flips `enabled` + `injection_enabled`), so after the fold `resolved_app_config.memory.mode == "tool"` and `not enabled` are both true here, and the message reads like a misconfiguration when it's actually an intentional per-agent opt-out.

Could gate it on the override not having been applied — e.g. only warn when the agent didn't supply a `memory:` block (`agent_config.memory is None`) — so the signal still fires for a genuinely misconfigured global `config.yaml` but not for an explicit opt-out. Minor, but it'll be noisy in logs for any stateless agent deployed alongside a global tool-mode memory config.

he-yufeng:
@WillemJiang The failed unit test was the upstream `debounce_seconds` -> `shutdown_flush_timeout_seconds` rename landing under this branch; the merges above (through a204b73e) picked it up and the toggle tests are green again (`uv run pytest tests/test_agent_memory_toggle.py`: 18 passed).

@willem-bd Done in 09e69ecf. The fold already returns a copy exactly when it applies an opt-out, so `_make_lead_agent` forwards that as a `memory_opt_out` flag into `build_middlewares` and the warning is gated on it. A genuinely misconfigured global `config.yaml` still warns on every other path, bootstrap and the embedded client included, since neither loads an agent config. Two tests pin both sides.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -530,6 +567,19 @@ def _make_lead_agent(config: RunnableConfig, *, app_config: AppConfig):
...
+    # config before anything reads it, so a custom agent with
+    # ``memory: {enabled: false}`` gets no memory capture, tools, or injection.
+    # A no-op for every other agent (returns the same object), so behavior is
+    # unchanged unless an agent explicitly disables memory.
+    resolved_app_config = apply_agent_memory_override(resolved_app_config, agent_config)
+    # The fold returns a copy exactly when it applied an opt-out (same object
+    # otherwise), so identity tells build_middlewares to skip the mode/enabled
+    # mismatch warning — that signal is for a misconfigured global config.yaml,
+    # not for a deliberate per-agent opt-out.
+    memory_opt_out = resolved_app_config is not app_config
```
willem-bd: The `resolved_app_config is not app_config` identity check works today, but it silently couples this site to `apply_agent_memory_override`'s "return a copy exactly when an opt-out was applied" contract. A future refactor that switches the helper to always return a defensive `model_copy` (a reasonable immutability-minded change) would flip this flag unconditionally `True` and silence the mode/enabled mismatch warning on every `_make_lead_agent` invocation — including genuine global `mode: tool + enabled: false` misconfigs.

The existing warning tests (`test_mode_tool_warning_fires_on_global_misconfig` / `..._skipped_for_agent_opt_out`) call `build_middlewares` directly with an explicit `memory_opt_out=` argument, so they don't exercise this inference and wouldn't catch such a regression.

Either option closes the gap:
- Compute the flag directly from `agent_config.memory` (mirrors the helper's own predicate, decouples from object identity): `memory_opt_out = bool(agent_config and agent_config.memory and not agent_config.memory.enabled)`
- Or add a `_make_lead_agent` / `make_lead_agent`-level test that asserts the warning is suppressed for a memory-opted-out agent and still fires for a genuine global misconfig, locking the inference end-to-end.

Minor — the contract is documented in the helper's docstring and the current behavior is correct; flagging only because this is the one path through the `risk:high` change that isn't covered by a test.

willem-bd (commented):
Follow-up on the new `memory_opt_out` wiring (09e69ecf). One inline note on a coverage gap in the identity-check inference; the four-gate fold-through and warning suppression logic itself look correct.

he-yufeng:
Good catch on the identity coupling. Fixed in dc0cadf6: the flag no longer comes from `resolved_app_config is not app_config` at all. A small `agent_memory_opt_out(agent_config)` helper now reads the condition straight from the agent's own `memory` field (same predicate the fold uses), so a future always-copy refactor of `apply_agent_memory_override` cannot flip the warning behavior. The fold keeps its copy-only-when-applied contract unchanged.

Two regression tests added: the helper answers True only for an explicit `memory: {enabled: false}` (True/True/None cases), and a memory-on agent keeps the flag off across a fold run. `test_agent_memory_toggle.py` 20/20 green locally.

he-yufeng:
@fancyboi999 the P1 is addressed in dc0cadf6: the preservation test no longer touches `MemoryConfig.debounce_seconds` (it now only exercises fields current main still exposes), and CI is green. Could you take another look when you have a moment?

he-yufeng:
Rebuilt the branch as 9a1cc6cd. An earlier merge on my side had swept macOS iCloud sync duplicates (the " 2"-suffixed files) into the history, so the PR diff showed 95 junk files on top of the actual change. The branch is now the feature and nothing else: 4 files, 430 lines, current with main.

Content is unchanged in intent, ported onto current main with two adjustments for how upstream moved: `allowed_subagents` is left intact (my branch predated it), and the old `backend/AGENTS.md` bullet was dropped because upstream restructured that doc and the section it referenced no longer exists. The opt-out mechanism itself is identical: `AgentConfig.memory` (default None, inherits global), `apply_agent_memory_override` folded once after `load_agent_config`, and the mode/enabled warning suppressed only for the intentional opt-out via `agent_memory_opt_out`.

Verification: `test_agent_memory_toggle.py` 20/20, plus the focused set around it (lead-agent model resolution, client, agents/memory config paths): 275 passed, 6 skipped.

he-yufeng:
@fancyboi999 The preservation test on the rebuilt head (9a1cc6cd) already uses `shutdown_flush_timeout_seconds` instead of `debounce_seconds`, matching the current `MemoryConfig` schema you pointed at, and the suite passes against current main: `uv run pytest tests/test_agent_memory_toggle.py` 20/20, plus the surrounding lead-agent/config/client set (275 passed, 6 skipped). The branch was rebuilt from current main yesterday, so the old "37 commits behind" state is gone too.

he-yufeng:
Tidying my open queue to stay under the new two-open-PR limit, and this one has been quiet for five weeks. The change is still valid against main as far as I can tell; happy to reopen when a slot frees or when there is reviewer appetite for it. Thanks @willem-bd for the careful passes on it.

PR #4185, opened 2026-07-14
feat(omnigent-mcp): FastMCP stdio server wrapping DeerFlow REST API for Omnigent/polly

lcquintela:
## Summary

- Adds `omnigent-mcp/server.py` — a Python MCP stdio server that exposes DeerFlow's REST API as three MCP tools (`deerflow_run`, `deerflow_create_thread`, `deerflow_health`) so Omnigent/polly can call DeerFlow directly without bespoke HTTP plumbing
- Handles session-cookie auth (POST form-encoded to `/api/v1/auth/login/local`), CSRF Double Submit Cookie pattern (`X-CSRF-Token` header), lazy client init, and automatic re-auth on 401/403
- Includes `omnigent-mcp/pyproject.toml` for dependency pinning and `omnigent-mcp/README.md` documenting setup, required env vars, and Omnigent stdio config snippet

## Tools exposed

| Tool | Endpoint | Notes |
|------|----------|-------|
| `deerflow_run(prompt, thread_id?)` | `POST /api/runs/wait` or `POST /api/threads/{id}/runs/wait` | Waits up to 300 s; returns last assistant message |
| `deerflow_create_thread()` | `POST /api/threads` | Returns `thread_id` for multi-turn use |
| `deerflow_health()` | `GET /health` | No auth; returns health JSON |

## Test plan

- [ ] `cd omnigent-mcp && python3 -c "from server import mcp; print('ok')"` passes
- [ ] Set `DEERFLOW_EMAIL` / `DEERFLOW_PASSWORD`, run `python3 server.py`, connect an MCP client and call `deerflow_health` — should return `{"status":"ok"}`
- [ ] Call `deerflow_run` with a simple prompt while DeerFlow is running — should return assistant response text
- [ ] Unset credentials, call any tool — should return a clear error string (no crash)

### DIFF
--- a/omnigent-mcp/pyproject.toml
+++ b/omnigent-mcp/pyproject.toml
@@ -1,0 +1,10 @@
+[project]
+name = "deerflow-mcp"
+version = "0.1.0"
+description = "MCP server wrapping DeerFlow REST API for Omnigent/polly"
+requires-python = ">=3.11"
+dependencies = ["mcp>=1.0.0", "httpx>=0.28.0"]
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
--- a/omnigent-mcp/server.py
+++ b/omnigent-mcp/server.py
@@ -1,0 +1,174 @@
+"""FastMCP stdio server wrapping DeerFlow REST API."""
+
+import asyncio
+import json
+import os
+from typing import Optional
+
+import httpx
+from mcp.server.fastmcp import FastMCP
+
+mcp = FastMCP("deerflow")
+
+DEERFLOW_URL = os.environ.get("DEERFLOW_URL", "http://localhost:2026")
+DEERFLOW_EMAIL = os.environ.get("DEERFLOW_EMAIL", "")
+DEERFLOW_PASSWORD = os.environ.get("DEERFLOW_PASSWORD", "")
+
+_client: Optional[httpx.AsyncClient] = None
+_csrf_token: str = ""
+
+
+def _make_client() -> httpx.AsyncClient:
+    return httpx.AsyncClient(
+        base_url=DEERFLOW_URL,
+        timeout=310.0,
+        follow_redirects=True,
+    )
+
+
+async def _login(client: httpx.AsyncClient) -> None:
+    global _csrf_token
+    resp = await client.post(
+        "/api/v1/auth/login/local",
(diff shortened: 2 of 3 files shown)


omnigent-mcp/server.py
```diff
@@ -0,0 +1,174 @@
...
+    messages = []
+    if isinstance(data, list):
+        messages = data
+    elif isinstance(data, dict):
+        messages = data.get("messages", [])
+
+    for msg in reversed(messages):
+        role = msg.get("role", "")
+        content = msg.get("content", "")
+        if role == "assistant":
```
fancyboi999, line 85: [P1] Match the Gateway's actual wait-response message schema

Both `/api/runs/wait` endpoints return `serialize_channel_values_for_api(...)`; their LangChain message dumps identify assistant messages as `type: "ai"`, not `role: "assistant"`. This branch therefore never matches a normal successful response and falls through to `json.dumps(data)`, so `deerflow_run` returns the entire state instead of the promised final answer. A response-shaped case such as `{"messages":[{"type":"ai","content":"expected answer"}]}` reproduces this. Please recognize the actual wire schema and add a regression test against it.

omnigent-mcp/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
+name = "deerflow-mcp"
+version = "0.1.0"
+description = "MCP server wrapping DeerFlow REST API for Omnigent/polly"
+requires-python = ">=3.11"
+dependencies = ["mcp>=1.0.0", "httpx>=0.28.0"]
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
```
fancyboi999, line 10: [P2] Make the documented `uv sync` path installable

This selects Hatchling as the build backend, but the project only contains a top-level `server.py` while the normalized project name expects a `deerflow_mcp` package, and no wheel file-selection rule is configured. As submitted, both `uv sync` and `uv build` fail with Hatchling's `Unable to determine which files to ship inside the wheel` error. Please either configure the intended package/build target or explicitly make this a non-package uv project, then cover the documented setup command.

fancyboi999 (changes requested):
Two reproducible blockers remain in the primary install/run path: the documented uv setup cannot build the project, and successful run responses are parsed against the wrong message schema. Details are inline.

Issue #4186, opened 2026-07-14
[RFC] DeerFlow Agentic Browser Control


18062706139fcz:
# DeerFlow Agentic Browser Control RFC

## Recommendation

DeerFlow should treat browser control as a first-class capability instead of stopping at read-only web fetching or one-off screenshots.

The recommended design is:

1. The backend provides a thread-scoped, reusable Playwright browser session.
2. The agent uses structured browser tools to navigate, observe, click, type, capture screenshots, and close the session.
3. The frontend provides a right-side Browser panel in the Workspace. It streams high-resolution live frames over WebSocket and sends user input back to the same thread browser session.
4. Browser action screenshots are observable progress feedback only. They must not appear in the Workspace Changes code diff.
5. The web version should not pursue native Chromium embedding or a full-screen iframe. v1 should use pixel streaming plus input event replay, which is the practical boundary for a controllable, deployable, and auditable web app.

This RFC is not trying to redesign a browser. It captures the direction that has already emerged so future discussions do not keep bouncing between screenshots, iframes, Electron, and Manus/Codex-style experiences.

## Background

DeerFlow is a general-purpose Super Agent Harness. Research, content production, form workflows, authenticated page reading, and dynamic web app investigation all need browser control.

The existing read-only tools do not cover these scenarios:

1. `web_fetch` can read static or server-visible content, but it cannot log in, click, type, or complete multi-step flows.
2. `web_capture` can take screenshots, but it renders a stateless page load and does not represent the real session the agent is operating.
3. Traditional tool results return state only to the model. Users cannot see what the agent is doing on the page in real time.
4. If browser progress screenshots appear in Workspace Changes, they pollute the code review view.

Users expect an experience closer to Manus or Codex-style products:

1. The agent can proactively open and operate web pages.
2. The user can observe the browser workflow.
3. The user can take over the browser when needed.
4. The image should be sharp, and scrolling plus keyboard input should feel close to a real browser.
5. The Browser panel should fit into the Workspace instead of disrupting the main conversation flow.

### Insufficient Observability

If the system only returns a text snapshot, the user cannot tell whether the agent opened the right page, got stuck on a login screen, or clicked the wrong area.

### Web Constraints Are Easy to Underestimate

A web app cannot embed a native Chromium View the way an Electron desktop app can.

Embedding the remote page directly in an iframe is also not a general solution:

1. Many sites set `X-Frame-Options` or CSP `frame-ancestors`, which prevents embedding.
[...]

PR #4187, opened 2026-07-14
feat(browser): add agentic browser control

18062706139fcz:
Refs #4186

## Why

DeerFlow currently has useful read-only web tools, but they do not cover browser-native workflows: authenticated pages, JavaScript-heavy apps, multi-step form flows, OAuth popups, or cases where the user needs to see and take over what the agent is doing.

This PR adds agentic browser control as a first-class thread capability so the agent can operate a real browser session while the user can observe or take over that same session from the Workspace. The design follows the RFC direction: Playwright-backed thread sessions, a WebSocket live channel, and a Workspace Browser panel instead of iframe or native Chromium embedding.

## What changed

- Added a thread-scoped Playwright browser automation package with `browser_navigate`, `browser_snapshot`, `browser_click`, `browser_type`, `browser_get_text`, `browser_back`, `browser_screenshot`, and `browser_close` tools.
- Added Gateway browser APIs: a REST navigation endpoint for URL-bar navigation and a WebSocket stream for live frames plus click, wheel, keyboard, navigation, and tab input events.
- Added the Workspace Browser panel with a persistent trigger, URL bar, back/forward controls, Live mode, fullscreen, close, frame rendering, input forwarding, reconnect handling, and URL synchronization.
- Added per-step browser screenshots as progress feedback while keeping automatic frames out of Workspace Changes via the hidden `.browser-frames` path.
- Kept browser tooling available to the default Lead Agent instead of accidentally narrowing it out through skill allowlists.
- Added backend tests for browser sessions/tools, Gateway browser routes, workspace-change filtering, and Lead Agent tool availability.
- Documented the new browser capability and configuration surface in README/config/backend guidance.

## Scenarios

- An agent opens a public website, clicks through dynamic UI, fills inputs, and extracts page text from the same stateful browser session.
- A user watches browser progress in the chat stream and opens the Browser panel to inspect the live page.
- A user takes over the agent's browser session through the URL bar, mouse, keyboard, wheel, back, and forward controls.
- A login or OAuth flow opens a popup/new tab and remains visible and controllable.
- Browser progress screenshots provide observability without polluting the code diff in Workspace Changes.

## Surface area

- [x] **Frontend UI** — Workspace Browser panel, trigger, live controls, input forwarding
- [x] **Backend API** — browser REST navigation endpoint and WebSocket live stream
- [x] **Agents / LangGraph** — browser tool registration and default Lead Agent tool availability
- [x] **Dependencies** — optional Playwright dependency for browser automation
- [x] **Default behavior change** — browser tools become available to the default Lead Agent; Live mode is the browser panel default

## Screenshots / Recording

Not attached in this PR. The Browser panel entry point is the Workspace header browser trigger, and the panel opens on the right side of the chat workspace.

## Validation

- `git diff --cached --check`
- `cd backend && uv run pytest tests/test_browser_automation.py tests/test_browser_router.py tests/test_workspace_changes.py tests/test_lead_agent_skills.py` — 73 passed
- `cd frontend && pnpm check` — not completed locally because the current shell is on Node.js v20.19.5, while this pnpm version requires Node.js v22.13+ and fails on missing `node:sqlite`.

image

image


backend/app/gateway/routers/browser.py
```diff
@@ -0,0 +1,365 @@
...
+    down, up, wheel, key, text, navigate) that drive the live page.
+    """
+    user = await _authenticate_ws(websocket)
+    if user is None:
+        await websocket.close(code=4401)
+        return
+
+    thread_store = getattr(websocket.app.state, "thread_store", None)
+    if thread_store is not None:
+        allowed = await thread_store.check_access(thread_id, str(user.id), require_existing=False)
```
fancyboi999: [P1] Require an existing owned thread and close its browser session on deletion\n\n`require_existing=False` deliberately allows a missing metadata row. The thread delete endpoint removes that row but never closes the browser session, while `BrowserSessionManager` retains sessions keyed only by `thread_id`. After the owner deletes a thread, another authenticated user who knows that ID can pass this check and reuse the old page/context, including cookies. A minimal probe returns both `missing-row-access=True` and `retained-session-reused=True`. Use strict ownership for the WS and REST mutation paths and tear down the session as part of thread deletion.

backend/packages/harness/deerflow/community/browser_automation/session.py
```diff
@@ -0,0 +1,749 @@
...
+
+    async def _rebind_screencast_safe(self) -> None:
+        try:
+            await self._rebind_screencast()
+        except Exception as exc:
+            logger.debug("browser live screencast rebind failed: %s", exc)
+
+    async def _navigate(self, url: str) -> PageSnapshot:
+        page = await self._ensure_page()
+        await page.goto(url, wait_until="domcontentloaded")
```
fancyboi999, line 468: [P1] Enforce the URL policy at the browser request boundary\n\nThe explicit navigate URL is checked before this call, but Playwright follows redirects automatically, and later clicks, popups, iframes, and subresource requests bypass that check entirely. A public URL can redirect to `http://169.254.169.254/...`; the resulting page is accepted and can then be read through `browser_get_text` or screenshots. The current tests only reject literal private initial URLs. Install context-level request/navigation interception and validate every redirect/destination (including resolved IP) before the request is sent.

frontend/src/components/workspace/browser-view/browser-view-panel.tsx
```diff
@@ -0,0 +1,486 @@
...
+          return;
+        }
+        // Forward printable single chars as text; named keys as key presses.
+        if (event.key.length === 1 && !event.metaKey && !event.ctrlKey) {
+          sendInput({ type: "text", text: event.key });
+          event.preventDefault();
+          event.stopPropagation();
+        } else if (
+          [
+            "Enter",
```
fancyboi999: [P2] Do not forward keys while an IME composition is active\n\nDuring Chinese/Japanese input, Enter confirms the current candidate with `nativeEvent.isComposing=true`. This handler still sends a remote `Enter`, which can submit the focused page form before composition finishes. Reuse the existing `isIMEComposing` helper or track composition state before forwarding printable and named keys, and add a regression for composing Enter.

fancyboi999 (changes requested):
Requesting changes for the inline security/input findings plus one ingress blocker: the standard same-origin client targets `/api/threads/{id}/browser/stream`, but both Nginx thread locations omit `Upgrade` / `Connection` forwarding. Those regex locations win over the frontend `/` block that has WebSocket headers, so `make dev` and Docker route Live as ordinary HTTP instead of upgrading it. Add a browser-stream location or WebSocket headers/timeouts to the thread API location and cover the Nginx entry path.\n\n

backend/app/gateway/routers/browser.py
```diff
@@ -0,0 +1,370 @@
...
+        # SSRF / URL validation failure.
+        raise HTTPException(status_code=400, detail=str(exc)) from exc
+    except Exception as exc:
+        logger.exception("Browser navigate failed: thread_id=%s url=%s err=%s", thread_id, url, exc)
+        raise HTTPException(status_code=502, detail=f"Browser navigation failed: {exc}") from exc
+
+    return BrowserNavigateResponse(**result)
+
+
+async def _authenticate_ws(websocket: WebSocket):
```
willem-bd, line 115: **P1 - WebSocket stream has no `Origin` validation (cross-origin WS CSRF)**

`_authenticate_ws` replicates cookie->user auth because WS upgrades bypass `AuthMiddleware`. But `CSRFMiddleware` is *also* a `BaseHTTPMiddleware` (`csrf_middleware.py:184`), so the WS bypasses it too - there is no `Origin`/CORS check here. The repo already has `get_configured_cors_origins()` / `_request_origin()` helpers.

In an auth-enabled multi-user deployment, a cross-origin page can open this WS riding the victim's cookie (browsers send cookies on WS upgrades), then both observe live frames and drive the victim's authenticated browser session (navigate/click/type). Gated by browser-tools-enabled and knowing a valid `thread_id`, but thread IDs are often enumerable/leaked.

Fix: reject the upgrade when `Origin` isn't in the configured allowed set (reuse `get_configured_cors_origins()`; fall back to same-origin vs `Host`). Standard WS-CSRF mitigation, cheap to add.

backend/app/gateway/routers/browser.py
```diff
@@ -0,0 +1,370 @@
...
+    captured via CDP screencast. Client → server: input events (click, move,
+    down, up, wheel, key, text, navigate) that drive the live page.
+    """
+    user = await _authenticate_ws(websocket)
+    if user is None:
+        await websocket.close(code=4401)
+        return
+
+    thread_store = getattr(websocket.app.state, "thread_store", None)
+    if thread_store is not None:
```
willem-bd: **P2 - ownership check fails open when `thread_store` is absent**

`if thread_store is not None:` skips the ownership check entirely if `thread_store` ever resolves to `None` (init failure, a test app, a future mount path) - then any authenticated caller can attach to any `thread_id`'s retained browser session (cookies, logged-in pages). `thread_store` is set at startup (`deps.py:286`), so this is unlikely in prod, but for a surface that drives a real browser, fail-*closed* (deny 4404 when the store can't be resolved) is the safer pattern.

backend/packages/harness/deerflow/community/browser_automation/session.py
```diff
@@ -0,0 +1,804 @@
...
+        """Abort any request whose URL fails the SSRF guard.
+
+        Runs at the context level so it covers the top navigation, every
+        redirect hop, popups/new tabs, iframes, and subresource fetches — the
+        paths a one-time initial-URL check cannot see. A public URL that
+        redirects to ``http://169.254.169.254/...`` is aborted before the
+        response is exposed through snapshots/text. Skipped for CDP-attached
+        real Chrome, which owns its own browsing context.
+        """
+        if self._url_guard is None or self._context is None or self._request_guard_bound or self._cdp_url:
```
willem-bd, line 435: **P2 - CDP mode bypasses the SSRF request guard**

The guard returns early when `self._cdp_url` is set, so redirect/subresource requests to private/metadata hosts are *not* aborted for CDP-attached real Chrome. That's intentional (Chrome owns its context) and `cdp_url` is documented local/trusted-only - but consider at least a `logger.warning` when a CDP session is created so operators can see the SSRF guard is off for that session. The headless path's guard is solid and well-tested.

backend/packages/harness/deerflow/community/browser_automation/session.py
```diff
@@ -0,0 +1,804 @@
...
+        await self._emit_live_frame()
+        self._schedule_settle_live_frames()
+
+    async def _stop_screencast(self, on_frame: Callable[[str], None] | None = None) -> None:
+        if on_frame is not None and self._on_frame is not on_frame:
+            return
+        self._on_frame = None
+        self._manual_live_frame_pending = False
+        self._settle_live_frames_pending = False
+        cdp = self._cdp
```
willem-bd: **P3 - dead CDP screencast teardown**

`self._cdp` is never assigned anywhere (`__init__`/`_close` set it to `None`; `_start_screencast` was rewritten to on-demand screenshots and no longer binds a CDP session), so `cdp = self._cdp` is always `None` and this `Page.stopScreencast`/`detach()` block is unreachable. Either remove the vestigial CDP-screencast state or restore it intentionally - the `_start_screencast` docstring ("This used to attach Chrome's CDP screencast…") suggests it's historical.

backend/packages/harness/deerflow/community/browser_automation/session.py
```diff
@@ -0,0 +1,804 @@
...
+        self._loop: _PlaywrightLoopThread | None = None
+        self._sessions: dict[str, BrowserSession] = {}
+        self._lock = threading.Lock()
+
+    def _ensure_loop(self) -> _PlaywrightLoopThread:
+        if self._loop is None:
+            self._loop = _PlaywrightLoopThread()
+        return self._loop
+
+    def get_session(
```
willem-bd, line 868: **P2 - session config is "first tool to run wins"**

`get_session` returns the existing session and ignores `headless/viewport/timeout/cdp_url` for any later caller. But `_resolve_session` (`tools.py`) reads *each tool's own* config name (`browser_click` vs `browser_navigate`…), and the README says "each tool reads its own config name, so repeat the keys." Result: whichever browser tool initializes the session first bakes in its config - a `headless: false` set only on `browser_navigate` is silently ignored if another tool (or the WS, or `navigate_and_capture`) runs first with defaults. (SSRF policy is consistent - `validate_browser_url` always reads `browser_navigate`'s `allow_private_addresses`.)

Fix: derive session-creation config from a single canonical source (e.g. always `browser_navigate`, or a group-level `browser` config) so launch config is deterministic regardless of call order.

backend/packages/harness/deerflow/community/browser_automation/session.py
```diff
@@ -0,0 +1,804 @@
...
+        await self._loop.run(self._stop_screencast(on_frame))
+
+    async def dispatch_input(self, event: dict) -> None:
+        await self._loop.run(self._dispatch_input(event))
+
+    async def close(self) -> None:
+        await self._loop.run(self._close())
+
+
+class BrowserSessionManager:
```
willem-bd, line 834: **P2 - no eviction / idle timeout for per-thread Chromium processes**

Sessions are keyed by `thread_id` and only closed by `browser_close`, thread deletion, or app shutdown (`app.py` lifespan). A long-running multi-user gateway accumulates one headless Chromium per thread that ever used browser tools - a real memory/FD leak. Consider an idle-timeout sweep (close sessions unused for N minutes) or a max-sessions cap, mirroring warm-pool eviction patterns elsewhere in the codebase.

frontend/src/components/workspace/browser-view/use-browser-stream.ts
```diff
@@ -0,0 +1,231 @@
...
+      const delay = Math.min(
+        RECONNECT_BASE_DELAY_MS * 2 ** connectionAttempt,
+        RECONNECT_MAX_DELAY_MS,
+      );
+      reconnectTimer = window.setTimeout(() => {
+        setConnectionAttempt((attempt) => attempt + 1);
+      }, delay);
+    };
+
+    socket.onopen = () => {
```
willem-bd, line 145: **P2 - reconnect counter never resets on open -> permanent stall after 6 reconnects**

`socket.onopen` sets `status="open"` but never `setConnectionAttempt(0)`; the counter only resets when `enabled` flips false. Since `connectionAttempt` is in the connect-effect deps and capped at `RECONNECT_MAX_ATTEMPTS=6`, after 6 cumulative reconnect cycles over the panel's lifetime - even across many successful connections - `scheduleReconnect` returns early forever and Live goes dead until the panel is toggled off/on.

Fix: `setConnectionAttempt(0)` in `onopen`.

frontend/src/components/workspace/browser-view/keyboard.ts
```diff
@@ -0,0 +1,55 @@
...
+  ctrlKey: boolean;
+}
+
+/**
+ * Decide how a keydown maps to a remote browser input, or ``null`` to ignore it.
+ *
+ * Pure so the forwarding policy (including the IME-composition guard that must
+ * swallow a composing Enter) can be unit-tested without a DOM.
+ */
+export function decideBrowserKeyInput(
```
willem-bd, line 36: **P2 - the promised unit test for `decideBrowserKeyInput` is missing**

The docstring above explicitly says this is "Pure so the forwarding policy (including the IME-composition guard) can be unit-tested without a DOM" - but no test was added. This function is security-relevant (decides which keystrokes reach the remote browser, incl. the IME-composition guard that must swallow a composing Enter). Please add the test the file's own comment promises.

willem-bd (commented):
Reviewed the diff against the head commit. Posting line-anchored findings inline; higher-level notes below.

**Frontend is untested and unverified.** `pnpm check` was not run (per the PR body), so lint/typecheck/build status for ~1,000 new frontend lines is unknown. No tests for the `browser-view` components or the `useBrowserStream` reconnect/state machine (non-trivial: `urlEditingRef`, `liveFallback`, `navSpinnerTimer`). See the inline note on `keyboard.ts`.

**Strengths.** SSRF posture on the headless path is solid: initial-URL check + a context-level request guard that aborts redirect/subresource hops to private hosts, with a real-Chromium `importorskip` integration test. Loop-affinity pattern matches the BoxLite provider. `_redact_url` log hygiene is a nice touch. Harness boundary clean (no `app.*` import).

**Lead-agent behavior change (has a test).** `agent.py` now sets `skills_for_tool_policy = []` for the default lead agent, so installed skills stay discoverable via `describe_skill` but no longer globally clamp the configured toolset (keeps `browser_navigate`/`bash` available). Sound and consistent with #4164/#4095; pinned by `test_default_lead_agent_does_not_apply_installed_skill_allowlists`. Worth flagging in the PR description as an intentional default-behavior change.

**Scope.** The `recent-chat-list.tsx` dedup and `sidecar-panel.tsx` close-button changes are unrelated to browser control - consider splitting out to shrink this `risk:high` diff's blast radius.

Priorities: (1) WS `Origin` validation, (2) reconnect-counter reset, (3) fail-closed `thread_store`, (4) canonical session-config source, (5) session eviction, (6) frontend tests + `pnpm check`.

WillemJiang:
@18062706139fcz  Please resolve the conflicts of main.

WillemJiang (dismissed):

fancyboi999 (changes requested):
Thanks @18062706139fcz. I found two current-head issues that should be addressed before this is ready.

[P1] Route Live control to the worker that owns the browser session

- Location: `backend/packages/harness/deerflow/community/browser_automation/session.py:747-868`, `backend/app/gateway/routers/browser.py:241-248`
- Problem: browser sessions are process-local, so a supported multi-worker gateway cannot guarantee that an agent tool call, the REST navigation endpoint, and the Live WebSocket reach the same Chromium session. The panel can open a blank second browser while the agent drives another process's browser, making observation and takeover unreliable.
- Evidence: `BrowserSessionManager` explicitly keeps a process-local module singleton keyed only inside that worker. Both tools and the WebSocket independently call `get_browser_session_manager().get_session(thread_id)`; there is no shared ownership or cross-worker forwarding. DeerFlow's production command supports `GATEWAY_WORKERS > 1`, and ordinary uvicorn worker dispatch provides no thread-id affinity.
- Suggested fix: add cross-worker browser-session ownership and route/forward Live, navigation, and tool operations to the owning worker, or move browser automation behind a shared service. If that is outside this MVP, fail closed when browser automation is enabled with more than one gateway worker instead of presenting a Live session that is not the agent's.
- Test: run two gateway workers, create/navigate the session through a run handled by worker A, connect the browser WebSocket through worker B, and assert the stream observes and controls the same page and URL.

[P2] Do not evict a session while it is still in use

- Location: `backend/packages/harness/deerflow/community/browser_automation/session.py:783-836`, `backend/app/gateway/routers/browser.py:241-248`
- Problem: the new idle/LRU eviction can close a browser that still has an active Live WebSocket or an in-flight browser operation.
- Evidence: last-use time is refreshed only by `BrowserSessionManager.get_session()`. The WebSocket calls it once, then uses the returned `session` directly for the lifetime of the connection. After 30 minutes, any other thread's `get_session()` can classify that actively streamed session as idle; the 32-session cap can likewise evict it as LRU. `_schedule_close()` then closes the underlying page/browser without any active-reference check.
- Suggested fix: pin/reference-count sessions for the duration of each WebSocket and browser operation, refresh activity on real session work, and evict only unpinned sessions. Release the pin in a `finally` block.
- Test: keep a stream pinned past a short test idle timeout (and separately exceed a small session cap), request another session, and assert the pinned session is not closed until the stream releases it.

Reviewed head SHA: `84290ae626537fa872b544c4ce9baa2a2be0bb4b`

backend/packages/harness/deerflow/community/browser_automation/session.py
```diff
@@ -0,0 +1,875 @@
...
+        # may call _ensure_page (which routes through _set_active_page); without
+        # this flag that would schedule another rebind and recurse.
+        self._screencast_binding = False
+        self._last_manual_live_frame_at = 0.0
+        self._settle_live_frames_pending = False
+        self._input_live_frame_generation = 0
+        self._input_live_frame_pending = False
+        self._page_listener_bound = False
+
+    async def _ensure_page(self) -> Page:
```
willem-bd, line 345: **`_ensure_page` rebuild race (narrow).** This method isn't guarded by a per-session lock. Two concurrent callers on the private loop (e.g. an agent tool call + a WS input dispatch during live takeover) that both find `_page` closed/`None` can each pass the `if self._page is not None and not self._page.is_closed()` check and each `launch()` / `new_context()` / `new_page()`, orphaning browser/context objects. Single-loop cooperative scheduling confines the race to the await points in the rebuild branch, so it's narrow - but an `async with self._ensure_lock:` around `_ensure_page` would close it cheaply.

backend/packages/harness/deerflow/community/browser_automation/session.py
```diff
@@ -0,0 +1,875 @@
...
+        """Abort any request whose URL fails the SSRF guard.
+
+        Runs at the context level so it covers the top navigation, every
+        redirect hop, popups/new tabs, iframes, and subresource fetches — the
+        paths a one-time initial-URL check cannot see. A public URL that
+        redirects to ``http://169.254.169.254/...`` is aborted before the
+        response is exposed through snapshots/text. Skipped for CDP-attached
+        real Chrome, which owns its own browsing context.
+        """
+        if self._url_guard is None or self._context is None or self._request_guard_bound or self._cdp_url:
```
willem-bd, line 435: **CDP sessions silently disable the SSRF request guard.** When `cdp_url` is set, `_install_request_guard` returns early here, so redirects / popups / iframes / subresources to private or cloud-metadata hosts are **not** aborted for that session (the per-session `logger.warning` is the only signal). It's documented as "local/trusted only" in the README, but nothing programmatically prevents enabling `cdp_url` on a shared multi-user gateway, where one user's agent could be redirected at internal infra. For a `risk:high` feature, consider refusing `cdp_url` unless `allow_private_addresses` is also explicitly set (or gated behind a separate trust flag) - at minimum a startup-level warning harder to miss than this per-session log line.

backend/app/gateway/routers/browser.py
```diff
@@ -0,0 +1,418 @@
...
+    target_host = _first_header_value(websocket.headers.get("x-forwarded-host")) or websocket.headers.get("host")
+    if target_host:
+        normalized_host = normalized.split("://", 1)[-1]
+        if normalized_host == target_host.strip().lower():
+            return True
+    return False
+
+
+@router.websocket("/threads/{thread_id}/browser/stream")
+async def browser_stream(websocket: WebSocket, thread_id: str) -> None:
```
willem-bd, line 179: **No integration test for this coroutine's gating sequence.** `test_browser_router.py` covers the *pure helpers* (`_ws_origin_allowed`, `_should_apply_browser_seed`, `validate_browser_url`), but none of the close-code branches here are exercised end-to-end: `4401` (unauthenticated), `4403` (cross-origin), `4404` (thread_store missing / non-owned / tools-disabled), `4501` (Playwright missing). Since the WS path **manually replicates** AuthMiddleware / CSRFMiddleware - the part most likely to drift - a `TestClient` / `httpx-ws` integration test asserting each close code would materially de-risk this. The underlying `check_access` is already tested, so this is a wiring-coverage gap, not a known defect.

backend/app/gateway/routers/browser.py
```diff
@@ -0,0 +1,418 @@
...
+        # Fail closed: the live stream drives a real browser (cookies,
+        # logged-in pages), so if the ownership store can't be resolved we must
+        # deny rather than let any authenticated caller attach to any thread's
+        # retained session.
+        await websocket.close(code=4404)
+        return
+    # Strict ownership: require an existing owned thread. A permissive check
+    # would let any authenticated caller attach to a deleted thread's id and
+    # reuse the retained page/context.
+    allowed = await thread_store.check_access(thread_id, str(user.id), require_existing=True)
```
willem-bd: **`user_id is None` threads are reachable by any authenticated user via the live WS.** `check_access(..., require_existing=True)` still returns `True` when the row exists but `row.user_id is None` (shared / legacy pre-auth threads). This matches the existing REST `require_permission` semantics (documented as intentional), so it's not a new hole - but the WS path drives a real browser carrying the user's cookies, so the blast radius is larger than a read. Flagging for the high-risk sign-off: confirm whether legacy `user_id=None` threads realistically exist on multi-user deployments, and whether the live WS should additionally require a non-null owner.

willem-bd (commented):
Code review pass on the agentic browser-control feature. Architecture and the security design (defense-in-depth SSRF, WS auth/CSRF/ownership replication, opt-in gating, lifecycle teardown) are sound and well-documented. Leaving the actionable refinements as inline comments below - none are hard blockers, but given `risk:high` / `needs-validation` I'd want explicit sign-off on the CDP-SSRF trade-off and a green CI `pnpm check` before merge.

Two non-inline notes:

- **Frontend validation gap.** `pnpm check` (lint + typecheck) was not run locally (Node v20 vs required v22). The new TS modules aren't type-checked locally; please confirm CI is green before merge, since `needs-validation` is set.
- **E2E.** I'd hold merge until at least one real-path E2E (browser open -> navigate -> click -> live takeover) is confirmed on a deployed stack.

fancyboi999 (changes requested):
Thanks @18062706139fcz. I found three current-head issues that should be addressed before this is ready.

[P1] Make the optional browser capability authoritative end to end

- Location: `README.md:775-786`, `frontend/src/app/workspace/chats/[thread_id]/page.tsx:264,294`, `scripts/detect_uv_extras.py`
- Problem: the browser is documented and packaged as optional, but ordinary startup does not preserve the extra and the frontend exposes the Browser control even when the backend cannot provide it. A default install therefore shows a control that reconnects to a 404/501-only surface, while the documented install command is undone by the next `make dev`/Docker dependency sync.
- Evidence: the example browser tools are commented out and the root backend project has no `browser` proxy extra. `make dev` runs `uv sync --all-packages` with flags from `detect_uv_extras.py`, which never detects an enabled browser tool. In an isolated current-head environment, `uv sync --all-packages --extra browser` made `playwright` importable, then the same ordinary sync without the extra removed it. Independently, the saved-thread page renders `BrowserTrigger` solely from `!isNewThread`; `/api/features` still reports only `agents_api`, despite the router deliberately returning 404 when browser tooling is disabled.
- Suggested fix: choose one capability model and wire it through the whole application. Either ship Playwright/browser support as a normal full-app dependency, or keep it optional and add a root extra plus startup-extra detection/documentation, fail startup when an enabled browser config lacks its runtime, expose one backend browser-availability flag, and gate both `BrowserTrigger` and `ChatBox` from that flag. The harness package can remain optional either way.
- Test: cover a default config/install (dependency absent, feature false, no Browser control or WS attempts) and an enabled config across a repeated dependency sync/startup (dependency preserved, feature true, Browser control reaches a real navigate/stream smoke).

[P1] Give Live leases hard admission and explicit viewer ownership

- Location: `backend/packages/harness/deerflow/community/browser_automation/session.py:857-946`, `backend/app/gateway/routers/browser.py:268-278`
- Problem: pinning fixes eviction of active sessions, but it turns `max_sessions` into an unbounded soft limit. Every Live WebSocket pins one session for its lifetime; when all entries are pinned, a new thread is inserted and `_collect_evictable_locked()` just breaks with the registry over capacity. An authenticated user can keep opening owned threads and retain an unbounded number of Chromium processes. The same lease model also has no multi-viewer rule: `BrowserSession` stores one `_on_frame`, so a second Live connection silently replaces the first connection's frame callback while both sockets remain open and can still send input.
- Evidence: a current-head runtime probe with `BrowserSessionManager(max_sessions=1)` acquired two pinned thread sessions and reported `live_sessions=2`. The WebSocket path holds `manager.acquire_session(...)` until disconnect, and `_start_screencast()` overwrites `_on_frame` after `_stop_screencast()` instead of rejecting the new viewer or registering another subscriber.
- Suggested fix: make admission atomic and hard: before creating a new session, evict an unpinned entry or reject with a retryable capacity error when none exists. Then define Live ownership explicitly: either fan frames out to a bounded subscriber set or close/reject the displaced connection so no accepted socket is left frozen and input-active. This keeps the prior no-evict-while-in-use invariant without losing the resource bound.
- Test: fill a size-1 manager with a pinned Live lease, assert a second thread cannot increase the registry/process count, then release and assert admission succeeds. Also connect two clients to one thread and assert the documented fan-out or takeover behavior, including cleanup after each disconnect.

[P2] Redact browser URLs before logging navigation failures

- Location: `backend/app/gateway/routers/browser.py:102-110`
- Problem: the failure path logs the complete user-supplied URL. Browser/OAuth flows commonly place authorization codes, reset tokens, signed query parameters, and fragments in URLs; a timeout or Playwright error copies those credentials into persistent Gateway logs.
- Evidence: the router passes raw `url` to `logger.exception`, while the new session layer already has `_redact_url()` specifically to drop query and fragment data for the same browser surface. The API also returns the raw exception text, which Playwright may include the navigated URL in.
- Suggested fix: move/reuse a shared URL redactor at the router boundary and return a generic 502 detail while retaining only redacted diagnostics server-side.
- Test: force navigation to fail for a URL containing `?code=secret#token`; assert neither captured logs nor the HTTP response contains the query or fragment, while scheme/host/path remain useful in the log.

Reviewed base SHA: `0f088033fed8abc193f907002b281dc57ab3845a`
Reviewed head SHA: `9a2037aa5e38786f9d533473522486a54c0615eb`

WillemJiang (approved):

PR #4198, opened 2026-07-15
fix(DB): check scheduled_task_runs independently in 0003 migration guard

DaoyuanLi2816:
## Status

Closed to reduce the outstanding PR queue and focus on issue-backed changes that already have reviewer involvement.


DaoyuanLi2816:
Closing this to reduce the outstanding PR queue and focus on the issue-backed changes that already have reviewer involvement.

PR #4200, opened 2026-07-15
fix(blocking-io): trace self/cls attribute chains and local aliases in the call graph

DaoyuanLi2816:
## Summary

`backend/tests/support/detectors/blocking_io_static.py` builds a same-file call
graph so it can tell whether a blocking IO call in a sync helper is reachable
from async code. `_record_call_ref` only ever recorded an edge for two call
shapes:

- a bare-name call (`flush()`)
- a literal single-hop `self.`/`cls.` attribute call (`self.flush()`)

Any other receiver shape fell through the `"." not in call_name` fallback and
was silently **dropped from the graph** — including:

- a deeper `self.`/`cls.` attribute chain: `self.store.flush()`
- a local variable holding a `self.`/`cls.` attribute: `store = self.store; store.flush()`
- a parameter used directly as a receiver: `def helper(self, store): store.flush()`

So a real blocking call reached only through one of those shapes never showed
up as `ASYNC_REACHABLE_SAME_FILE`, even though it's genuinely reachable from
async code in the same file.

This repo's own docs (`backend/AGENTS.md`) already disclose the detector's
accepted **false-positive** tradeoff ("duplicate helper names in one file can
conservatively over-report async reachability"). This bug is the opposite,
more dangerous, **false-negative** direction — a real blocking call silently
missing from the safety net.

I confirmed the gap directly: 3 structurally identical async-calls-a-blocking-`open()`
snippets differing only in the call shape used to reach the blocking method
gave 1/3 findings before this fix (only the literal `self.flush()` shape was
caught) and 3/3 after.

## Fix

Extend `_record_call_ref` to also trace, **within the same function only**:

- multi-hop `self.`/`cls.`-rooted attribute chains (any depth), and
- a local variable or parameter traced back to a `self.`/`cls.` attribute or
  used directly as a receiver (seeded from the function's own parameters,
  extended by simple `x = self.attr` / `x = cls.attr` / `x = <already-traced-name>`
  assignments)

Both are resolved through the **same** bare-method-name fallback the detector
already uses for receivers it cannot resolve to a name at all (`functions_by_name`
lookup), so this adds no new *kind* of false-positive risk beyond what the tool
already accepts for duplicate helper names in one file — it just applies that
same existing, documented tradeoff to a couple more call shapes instead of
silently dropping them.

Deliberately out of scope (documented in `backend/AGENTS.md`): cross-function
or cross-module alias/type inference, `with ... as` aliasing, walrus (`:=`)
assignment, and `for` loop targets. These remain unreported false negatives,
same as before.

Ran the fixed scanner against the actual codebase before/after — the finding
set is byte-for-byte identical (41 findings both times), so this is a
low-blast-radius, purely additive fix with no practical effect on current
`make detect-blocking-io` output.

## Test plan

- [x] Added regression tests reproducing all 3 previously-missed call shapes
      (`self.store.flush()`, local-variable alias, parameter-used-directly),
      plus a scope-boundary test proving an unrelated local variable (not a
      parameter, not traced from `self.`/`cls.`) is still **not** traced —
      guarding against reintroducing broad false positives.
- [x] Confirmed fail-before/pass-after: reverted just the source fix (kept the
      new tests) and reran — 3 of the 4 new tests failed with the exact gap
      described above; reapplied the fix and all 4 passed.
- [x] `pytest backend/tests/test_detect_blocking_io_static.py backend/tests/test_scan_changed_blocking_io.py backend/tests/test_detector_repo_root.py backend/tests/blocking_io/test_gate_smoke.py` — 36 passed. (1 pre-existing failure in `test_json_output_uses_concise_review_record_schema` is an unrelated Windows path-separator comparison issue in the test itself, reproducible identically on `main` before this change.)
- [x] `ruff check --line-length 240` and `ruff format --check --line-length 240` clean on both changed files, plus a full `ruff check .` over `backend/`.
- [x] Updated `backend/AGENTS.md`'s description of the detector to document the new same-function attribute-chain/alias tracing alongside the existing false-positive disclosure.

### DIFF
--- a/backend/tests/support/detectors/blocking_io_static.py
+++ b/backend/tests/support/detectors/blocking_io_static.py
@@ -293,6 +293,27 @@ def dotted_name(node: ast.AST | None) -> str | None:
     return None
 
 
+def _simple_receiver_name(node: ast.AST | None) -> str | None:
+    """Like `dotted_name`, but only for Name/Attribute chains.
+
+    `dotted_name` intentionally unwraps `ast.Call` and `ast.Subscript` to build a
+    symbolic name for blocking-call pattern matching (`visit_Call` /
+    `_blocking_rule` / `_sync_http_client_factory_base`). Reusing that for
+    receiver/alias tracking is wrong: it would make a Call or Subscript result
+    inherit its base's alias-worthiness (e.g. treating `factory()` as if it
+    were the traced name `factory`), which this restricted extractor refuses
+    to do by simply not recognizing those node shapes at all.
+    """
+    if isinstance(node, ast.Name):
+        return node.id
+    if isinstance(node, ast.Attribute):
+        parent = _simple_receiver_name(node.value)
+        if parent:
+            return f"{parent}.{node.attr}"
+        return node.attr
+    return None
+
+
 def relative_to_repo(path: Path, repo_root: Path = REPO_ROOT) -> str:
     try:
         return path.resolve().relative_to(repo_root.resolve()).as_posix()
@@ -400,6 +421,33 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
                 else:
                     current_clients[name] = previous
 
+    def visit_If(self, node: ast.If) -> None:
+        # `ast.If` is the only branching construct given isolated, merged alias
+        # state: `body` and `orelse` are mutually exclusive at runtime, so an
+        # alias added in one must not leak into the other, but a name aliased in
+        # either branch might still be aliased after the `if` (a conservative
+        # may-alias join) -- and the result must not depend on which branch is
+        # textually `body` vs `orelse`. `ast.Try`/`ast.Match` have different,
+        # more complex control-flow semantics (exception edges, multiple
+        # mutually exclusive case bodies) and are deliberately out of scope
+        # here; they keep the prior unisolated `generic_visit` behavior.
+        self.visit(node.test)
+        if not self.local_receiver_alias_stack:
+            for statement in node.body:
+                self.visit(statement)
+            for statement in node.orelse:
+                self.visit(statement)
+            return
+        before = set(self.current_local_receiver_aliases)
+        for statement in node.body:
+            self.visit(statement)
+        after_body = set(self.current_local_receiver_aliases)
+        self.local_receiver_alias_stack[-1] = set(before)
+        for statement in node.orelse:
+            self.visit(statement)
+        after_orelse = self.current_local_receiver_aliases
+        self.local_receiver_alias_stack[-1] = after_body | after_orelse
+
     def visit_Call(self, node: ast.Call) -> None:
         current = self.current_context
         call_name = self._canonical_name(dotted_name(node.func))
@@ -416,15 +464,44 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
         self.functions_by_name[node.name].append(qualname)
         if class_name is not None:
             self.class_methods[class_name].add(node.name)
+
+        # Decorators, parameter defaults/annotations, the return annotation, and
+        # any PEP 695 type-parameter bounds all run at definition time in the
+        # ENCLOSING scope, not inside the function body -- visit them before
+        # pushing this function's own context. Otherwise a call/receiver there
+        # (e.g. a default value referencing an outer variable that happens to
+        # share a parameter's name) gets misattributed to the function being
+        # defined instead of to whatever actually executes it.
+        for decorator in node.decorator_list:
+            self.visit(decorator)
+        self._visit_definition_time_arguments(node.args)
+        if node.returns is not None:
+            self.visit(node.returns)
+        for type_param in node.type_params:
+            bound = getattr(type_param, "bound", None)
+            if bound is not None:
+                self.visit(bound)
+
         self.function_stack.append(context)
         self.sync_http_client_stack.append({})
         self.path_like_name_stack.append(set(_path_like_argument_names(node.args, self._canonical_name)))
         self.local_receiver_alias_stack.append(set(_all_argument_names(node.args)))
-        self.generic_visit(node)
+        for statement in node.body:
+            self.visit(statement)
         self.local_receiver_alias_stack.pop()
         self.path_like_name_stack.pop()
         self.sync_http_client_stack.pop()
         self.function_stack.pop()
+
+    def _visit_definition_time_arguments(self, arguments: ast.arguments) -> None:
+        for default in arguments.defaults:
+            self.visit(default)
+        for default in arguments.kw_defaults:
+            if default is not None:
+                self.visit(default)
+        for argument in _iter_arguments(arguments):
+            if argument.annotation is not None:
+                self.visit(argument.annotation)
 
     def _canonical_name(self, name: str | None) -> str | None:
         if name is None:
@@ -442,7 +519,7 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
             return
         if not isinstance(node.func, ast.Attribute):
             return
-        receiver = dotted_name(node.func.value)
+        receiver = _simple_receiver_name(node.func.value)
         if receiver in {"self", "cls"}:
             self.call_refs[current.qualname].append(_CallRef(node.func.attr, current.class_name, self_method=True))
             return
@@ -557,17 +634,27 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
                 current_clients[name] = client_base
 
     def _record_local_receiver_alias_targets(self, value: ast.AST, targets: Iterable[ast.AST]) -> None:
+        # Every assignment to a name previously in `current_local_receiver_aliases`
+        # must resolve that name's traceability from scratch -- not only add new
+        # traceable names -- or a stale alias from an earlier, unrelated value
+        # would keep exposing same-named blocking methods (dead code below this
+        # point) forever. A non-traceable value (an unrecognized shape, or a
+        # Call/Subscript result -- see `_simple_receiver_name`) therefore kills
+        # the name instead of leaving it untouched.
         if not self.local_receiver_alias_stack:
             return
-        dotted = dotted_name(value)
-        if dotted is None:
-            return
         current_aliases = self.current_local_receiver_aliases
-        root = dotted.split(".", 1)[0]
-        if root not in {"self", "cls"} and root not in current_aliases:
+        assigned_names = [name for target in targets for name in _iter_assigned_names(target)]
+        if not assigned_names:
             return
-        for target in targets:
-            current_aliases.update(_iter_assigned_names(target))
+        dotted = _simple_receiver_name(value)
+        root = dotted.split(".", 1)[0] if dotted is not None else None
+        is_traceable = root is not None and (root in {"self", "cls"} or root in current_aliases)
+        for name in assigned_names:
+            if is_traceable:
+                current_aliases.add(name)
+            else:
+                current_aliases.discard(name)
 
     def _sync_http_client_factory_base(self, node: ast.AST) -> str | None:
         if not isinstance(node, ast.Call):
--- a/backend/tests/support/detectors/blocking_io_static.py
+++ b/backend/tests/support/detectors/blocking_io_static.py
@@ -302,15 +302,20 @@ def _simple_receiver_name(node: ast.AST | None) -> str | None:
     receiver/alias tracking is wrong: it would make a Call or Subscript result
     inherit its base's alias-worthiness (e.g. treating `factory()` as if it
     were the traced name `factory`), which this restricted extractor refuses
-    to do by simply not recognizing those node shapes at all.
+    to do by simply not recognizing those node shapes at all -- including when
+    one is buried further down the chain (`factory().client`,
+    `clients[0].client`): an unsupported node anywhere in the chain makes the
+    whole receiver None, it never falls back to just the trailing attribute
+    name, or that trailing name alone could still collide with an unrelated
+    traced parameter or local alias of the same name.
     """
     if isinstance(node, ast.Name):
         return node.id
     if isinstance(node, ast.Attribute):
         parent = _simple_receiver_name(node.value)
-        if parent:
-            return f"{parent}.{node.attr}"
-        return node.attr
+        if parent is None:
+            return None
+        return f"{parent}.{node.attr}"
     return None
 
 
@@ -335,6 +340,7 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
         self.relative_path = relative_path
         self.source_lines = source_lines
         self.import_aliases: dict[str, str] = {}
+        self.postponed_annotations = False
         self.class_stack: list[str] = []
         self.function_stack: list[_FunctionContext] = []
         self.module_context = _FunctionContext("<module>", None, False)
@@ -370,6 +376,15 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
     def visit_ImportFrom(self, node: ast.ImportFrom) -> None:
         if node.module is None:
             return
+        if node.module == "__future__" and any(alias.name == "annotations" for alias in node.names):
+            # `from __future__ import annotations` (PEP 563) makes CPython skip
+            # evaluating parameter/return annotations at runtime entirely --
+            # they are kept as unevaluated strings -- so a call written in one
+            # never actually runs at definition time in this module, in either
+            # scope. This is always visited before any function def that could
+            # use it: the statement is required to appear before any other
+            # code in the file (`ast.parse` itself rejects it elsewhere).
+            self.postponed_annotations = True
         for alias in node.names:
             local_name = alias.asname or alias.name
             self.import_aliases[local_name] = f"{node.module}.{alias.name}"
@@ -388,16 +403,28 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
         self._visit_function(node, is_async=True)
 
     def visit_Assign(self, node: ast.Assign) -> None:
+        # Visit the RHS before recording anything about the assignment's own
+        # target(s): Python evaluates the value before binding it, so a call
+        # in the RHS (e.g. `client = client.flush()`) must see the receiver-
+        # alias state as it stood immediately BEFORE this assignment, not
+        # after. Updating/killing the target's alias first would make the
+        # target's own old alias disappear before the RHS that still runs
+        # under it gets a chance to be resolved.
+        self.visit(node.value)
         self._record_sync_http_client_targets(node.value, node.targets)
         self._record_local_receiver_alias_targets(node.value, node.targets)
-        self.generic_visit(node)
+        for target in node.targets:
+            self.visit(target)
 
     def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
         self._record_path_like_annotation(node.annotation, [node.target])
         if node.value is not None:
+            # Same RHS-before-target-update ordering as visit_Assign above.
+            self.visit(node.value)
             self._record_sync_http_client_targets(node.value, [node.target])
             self._record_local_receiver_alias_targets(node.value, [node.target])
-        self.generic_visit(node)
+        self.visit(node.target)
+        self.visit(node.annotation)
 
     def visit_With(self, node: ast.With) -> None:
         temporary_clients: dict[str, str | None] = {}
@@ -456,6 +483,32 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
             self._record_blocking_candidate(node, call_name, current)
         self.generic_visit(node)
 
+    def visit_Lambda(self, node: ast.Lambda) -> None:
+        # A lambda's parameter defaults run eagerly when the lambda object
+        # itself is created (same timing as a regular function's defaults),
+        # but its body only runs later, whenever/if the lambda is actually
+        # called -- possibly in a completely different scope, possibly never.
+        # This applies everywhere `self.visit` can reach a Lambda, including
+        # (via `_visit_definition_time_arguments` below) inside another
+        # function's own definition-time expressions, so `node.body` must
+        # never be visited from here.
+        for default in node.args.defaults:
+            self.visit(default)
+        for default in node.args.kw_defaults:
+            if default is not None:
+                self.visit(default)
+
+    def visit_GeneratorExp(self, node: ast.GeneratorExp) -> None:
+        # Only the outermost `for`'s iterable is evaluated eagerly, to build
+        # the generator object; the element expression, any `if` filters, and
+        # any additional `for` clauses live inside the generator's own frame
+        # and only run once it is iterated -- which may never happen. List/
+        # set/dict comprehensions differ (their implicit scope runs
+        # immediately as part of building the result) so they are left fully
+        # eager and are not given a matching override.
+        if node.generators:
+            self.visit(node.generators[0].iter)
+
     def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef, *, is_async: bool) -> None:
         qualname = ".".join((*self.class_stack, node.name)) if self.class_stack else node.name
         class_name = self.class_stack[-1] if self.class_stack else None
@@ -465,22 +518,29 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
         if class_name is not None:
             self.class_methods[class_name].add(node.name)
 
-        # Decorators, parameter defaults/annotations, the return annotation, and
-        # any PEP 695 type-parameter bounds all run at definition time in the
+        # Decorators and parameter defaults run at definition time in the
         # ENCLOSING scope, not inside the function body -- visit them before
         # pushing this function's own context. Otherwise a call/receiver there
         # (e.g. a default value referencing an outer variable that happens to
         # share a parameter's name) gets misattributed to the function being
-        # defined instead of to whatever actually executes it.
+        # defined instead of to whatever actually executes it. Parameter and
+        # return annotations get the same enclosing-scope treatment unless
+        # this module postpones annotation evaluation (`from __future__
+        # import annotations`), in which case they never execute at runtime at
+        # all, in either scope, so they are skipped entirely. PEP 695
+        # type-parameter bounds (`node.type_params`) are not visited at all,
+        # in either scope: CPython evaluates each one lazily, in its own
+        # hidden function, only when something like `T.__bound__` is actually
+        # accessed -- never as part of running this `def` statement.
+        # `visit_Lambda`/`visit_GeneratorExp` apply the same eager-vs-lazy
+        # split within whatever we do visit here, so a lambda body or
+        # generator element nested inside a decorator/default/annotation is
+        # not treated as running right now either.
         for decorator in node.decorator_list:
             self.visit(decorator)
         self._visit_definition_time_arguments(node.args)
-        if node.returns is not None:
+        if node.returns is not None and not self.postponed_annotations:
             self.visit(node.returns)
-        for type_param in node.type_params:
-            bound = getattr(type_param, "bound", None)
-            if bound is not None:
-                self.visit(bound)
 
         self.function_stack.append(context)
         self.sync_http_client_stack.append({})
@@ -499,6 +559,8 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
         for default in arguments.kw_defaults:
             if default is not None:
                 self.visit(default)
+        if self.postponed_annotations:
+            return
         for argument in _iter_arguments(arguments):
             if argument.annotation is not None:
                 self.visit(argument.annotation)
--- a/backend/tests/support/detectors/blocking_io_static.py
+++ b/backend/tests/support/detectors/blocking_io_static.py
@@ -325,6 +325,7 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
         self.functions_by_name: dict[str, list[str]] = defaultdict(list)
         self.call_refs: dict[str, list[_CallRef]] = defaultdict(list)
         self.path_like_name_stack: list[set[str]] = []
+        self.local_receiver_alias_stack: list[set[str]] = []
         self.potential_findings: list[_PotentialFinding] = []
 
     @property
@@ -367,12 +368,14 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
 
     def visit_Assign(self, node: ast.Assign) -> None:
         self._record_sync_http_client_targets(node.value, node.targets)
+        self._record_local_receiver_alias_targets(node.value, node.targets)
         self.generic_visit(node)
 
     def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
         self._record_path_like_annotation(node.annotation, [node.target])
         if node.value is not None:
             self._record_sync_http_client_targets(node.value, [node.target])
+            self._record_local_receiver_alias_targets(node.value, [node.target])
         self.generic_visit(node)
 
     def visit_With(self, node: ast.With) -> None:
@@ -416,7 +419,9 @@ class BlockingIOStaticVisitor(ast.NodeVisitor):
         self.function_stack.append(context)
         self.sync_http_client_stack.append({})
         self.path_like_name_stack.append(set(_path_like_argument_names(node.args, self._canonical_name)))
+        self.local_receiver_alias_stack.append(set(_all_argument_names(node.args)))
         self.generic_visit(node)
+        self.local_receiver_alias_stack.pop()
         self.path_like_name_stack.pop()
         self.sync_http_client_stack.pop()
         self.function_stack.pop()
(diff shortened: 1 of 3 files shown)


ShenAC-SAC (commented):
@DaoyuanLi2816 Thank you for the clear write-up, focused regression tests, and careful base/head comparison. The direction makes sense, and closing these false negatives is valuable. I found a few edge cases in the new alias tracking that appear to extend beyond the stated scope:

- **Call and subscript results are treated as aliases.** Both `_record_call_ref` and `_record_local_receiver_alias_targets` rely on `dotted_name()`, which intentionally unwraps `ast.Call` and `ast.Subscript`. As a result, `factory().flush()`, `client = factory(); client.flush()`, and `client = clients[0]; client.flush()` are treated like calls on a traced parameter or attribute. In base/head probes, base reported nothing while this branch marked an unrelated same-file blocking `flush()` as `ASYNC_REACHABLE_SAME_FILE`. This goes beyond the documented simple `Name`/`Attribute` aliases and effectively infers call/subscript result types. Could we use a receiver-specific extractor that accepts only the explicitly supported AST shapes and add negative tests for call/subscript results?

- **Alias state is never invalidated.** The alias set only grows, so after `client = NonBlockingClient()`, a parameter or prior `self.client` alias remains traceable. A later `client.flush()` can still expose unrelated same-name blocking helpers. Mutually exclusive branches also leak state into each other, and reversing branch order can change the detector output. Could we model assignment kills plus a conservative branch-state merge, or explicitly define and test a consistent may-alias policy?

- **Definition-time expressions are visited under the new function's parameter aliases.** `_visit_function` pushes the function state before `generic_visit(node)`, but defaults, decorators, and annotations execute in the enclosing scope. For example, with `receiver = NonBlocking()` followed by `async def route(receiver=receiver.flush()): ...`, this branch links the global call to the async function and can report an unrelated blocking `flush()`; base reports nothing. Could we visit definition-time expressions in the enclosing context and push the new function state only while visiting its body?

Local validation is otherwise clean: the targeted suite passes (37 tests), the full backend Ruff check/format passes, and the default base/head finding lists are still identical at 41/41. The cases above are not covered by that final-count comparison because they change call-graph semantics without affecting today's 41 candidate lines.

Thanks again — happy to re-check once this is updated.

DaoyuanLi2816:
Thanks for the thorough review and the concrete base/head repros — replying to each point. Pushed as `428660d`.

**1. Call and subscript results treated as aliases**

Confirmed and fixed. `_record_call_ref` and `_record_local_receiver_alias_targets` now use a new restricted extractor, `_simple_receiver_name()`, that only recognizes `Name`/`Attribute` chains instead of reusing `dotted_name()` (which intentionally unwraps `Call`/`Subscript` for the blocking-call symbolic-name matching it's actually used for elsewhere — `visit_Call` / `_blocking_rule` / `_sync_http_client_factory_base` — and is otherwise untouched). Added negative tests for all three shapes you listed: `factory().flush()`, `client = factory(); client.flush()`, and `client = clients[0]; client.flush()`.

**2. Alias state never invalidated + branch leak**

Both fixed, separately:
- `_record_local_receiver_alias_targets` now resolves traceability from scratch on every assignment to a previously-known name and kills it (`.discard()`) when the new value isn't traceable, instead of only ever adding. Added a kill-then-reference regression test (`client = self.store`, then `client = NonBlockingClient()`, then `client.flush()`).
- Added a `visit_If` override that snapshots the alias set before the branch, visits `body`, resets to the snapshot, visits `orelse`, then unions both exit states as the state after the whole `if` — a conservative may-alias join that no longer depends on which branch is textually first. Verified order-independence directly: the same two statements swapped between `if`/`else` now produce identical (and correctly empty) output on both sides, where before they diverged. Also added a test confirming the union half of this actually carries forward (a name aliased in only one branch is still treated as possibly aliased after the `if`).

Deliberately scoped this to `ast.If` only, per your note — `ast.Try`/`ast.Match` have different, more complex control-flow semantics (exception edges, multiple mutually-exclusive case bodies) and keep the previous unisolated traversal. Calling that out as an explicit scope boundary rather than a silent gap.

**3. Definition-time expressions under the new function's context**

Confirmed and fixed with your exact repro (`receiver = NonBlocking()` / `async def route(receiver=receiver.flush())`). `_visit_function` now visits `decorator_list`, parameter defaults/annotations, and the return annotation *before* pushing the function's own context (and its parameter-seeded alias set) onto the stacks, so they resolve in the enclosing scope instead.

One addition beyond what was flagged: Python 3.12 (this repo's floor) added a `type_params` field to `FunctionDef`/`AsyncFunctionDef` for PEP 695 generics, which `generic_visit` was also walking under the new function's context. Moved that to the enclosing scope too for the same reason, with its own regression test — no real impact today (the codebase doesn't use PEP 695 generic syntax anywhere), but it's the same bug shape and `generic_visit` was covering it, so the rewrite shouldn't silently drop it.

Also added a test proving the fix doesn't just make default-time findings disappear: a never-called nested function's default-value blocking call now correctly attaches to the actual enclosing async function that defines it, rather than to the never-reached nested function (which would otherwise silently swallow the finding).

**Verification**

- All three of your repros reproduced first against the unfixed branch (standalone scans, before touching source), then confirmed resolved after.
- TDD: added 10 new tests; proved fail-before/pass-after via patch-file revert of just the source fix (9/10 fail without it — the 10th is a supplementary positive test for the union-merge that happens to hold under both old and new code for that specific input, not a discriminator, flagging that for transparency rather than presenting it as something it isn't) — then reapplied and confirmed all pass.
- Targeted suite (`test_detect_blocking_io_static.py` + `test_scan_changed_blocking_io.py` + `test_detector_repo_root.py` + `blocking_io/test_gate_smoke.py`): 47/47 pass (37 original + 10 new).
- Also ran the full backend suite on both the pre-fix and post-fix commit to rule out any wider regression: identical 64 pre-existing failures on both (all Windows-environment artifacts in unrelated modules — symlinks, POSIX shell metacharacter handling, etc. — verified unrelated to this file), plus the 10 new passing tests on top.
- Real scanner against the actual backend tree: still 41/41, byte-identical JSON before/after — these were latent risks in shapes the current codebase doesn't happen to contain, not active miscounts, consistent with what you noted about the final-count comparison not covering them.
- `ruff check` / `ruff format --check` clean on the whole backend.
- Updated `backend/AGENTS.md`'s description of this detector to match the corrected scope (restricted alias extraction, kill-on-reassignment, the `ast.If`-only branch merge, and enclosing-scope definition-time expressions).

Thanks again for catching these — appreciate the precision on where the line actually was.

backend/tests/support/detectors/blocking_io_static.py
```diff
@@ -367,12 +389,14 @@ def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
 
     def visit_Assign(self, node: ast.Assign) -> None:
         self._record_sync_http_client_targets(node.value, node.targets)
+        self._record_local_receiver_alias_targets(node.value, node.targets)
```
ShenAC-SAC, line 426: IMPORTANT: This updates or kills the target alias before the assignment RHS is visited. With `client = self.store; client = client.flush()`, this method discards `client` because the second RHS is an `ast.Call`; `generic_visit()` then scans `client.flush()` after the old alias has disappeared, so the current head returns no finding. Python evaluates the RHS before assigning the target. Please visit the RHS with the pre-assignment alias state, then apply the target update afterward for both `Assign` and `AnnAssign`, and add a regression test for this sequence.

ShenAC-SAC (commented):
@DaoyuanLi2816, thank you for the detailed response and the substantial follow-up. I reran the original repros and confirmed that the exact cases from the first review are now fixed. I also appreciate the added regression tests and the documentation update.

I found three remaining correctness issues in the follow-up:

- Alias state is updated before the assignment RHS is visited, which introduces a false negative.
- The restricted receiver extractor still truncates Call/Subscript-rooted attribute chains into a traceable local name.
- The definition-time traversal recursively treats lazy subexpressions as eagerly executed.

I left concrete repros and requested actions inline below. Thanks again for working through these carefully.

backend/tests/support/detectors/blocking_io_static.py
```diff
@@ -293,6 +293,27 @@ def dotted_name(node: ast.AST | None) -> str | None:
...
+    were the traced name `factory`), which this restricted extractor refuses
+    to do by simply not recognizing those node shapes at all.
+    """
+    if isinstance(node, ast.Name):
+        return node.id
+    if isinstance(node, ast.Attribute):
+        parent = _simple_receiver_name(node.value)
+        if parent:
+            return f"{parent}.{node.attr}"
+        return node.attr
```
ShenAC-SAC: IMPORTANT: Falling back to `node.attr` when the parent is unsupported means this helper still accepts mixed chains rooted in a Call or Subscript. Both `factory().client` and `clients[0].client` collapse to `"client"`; if `client` is also a parameter or local alias, `factory().client.flush()` is incorrectly linked to a same-file `Store.flush`. Please return `None` when the recursive parent is `None` so the entire receiver must be a Name/Attribute-only chain, and add nested Call/Subscript attribute tests.

backend/tests/support/detectors/blocking_io_static.py
```diff
@@ -413,14 +464,45 @@ def _visit_function(self, node: ast.FunctionDef | ast.AsyncFunctionDef, *, is_as
         self.functions_by_name[node.name].append(qualname)
         if class_name is not None:
             self.class_methods[class_name].add(node.name)
+
+        # Decorators, parameter defaults/annotations, the return annotation, and
+        # any PEP 695 type-parameter bounds all run at definition time in the
+        # ENCLOSING scope, not inside the function body -- visit them before
```
ShenAC-SAC, line 559: IMPORTANT: Not every subtree visited below executes at function-definition time. On Python 3.12, annotations under `from __future__ import annotations` are deferred, PEP 695 bounds are evaluated only when attributes such as `T.__bound__` are accessed, and lambda or generator bodies embedded in defaults do not execute when the default object is created. For a nested function inside an async function, the current head newly reports each of these as `DIRECT_ASYNC` even though runtime instrumentation shows no call at definition. Please traverse only the eagerly evaluated portions in the enclosing context, replace the PEP 695 positive test, and add negative coverage for postponed annotations and lazy lambda/generator bodies.

DaoyuanLi2816:
Thanks for the follow-up round and the precise repros again — all three fixed. Pushed as `f6c35944`.

**1. Alias state updated before the assignment RHS is visited**

Confirmed with your exact repro. `visit_Assign` called `_record_local_receiver_alias_targets` (which can now `.discard()` a name) before `generic_visit` ever reached `node.value`, so with `client = self.store; client = client.flush()`, the second statement killed `client`'s alias *before* its own RHS (`client.flush()`) was analyzed — Python evaluates the RHS before binding the target, so the call should have still seen `client` aliased to `self.store`. `visit_Assign`/`visit_AnnAssign` now explicitly visit `node.value` first, then run the sync-http-client/alias target updates, then visit the target(s) — no more `generic_visit(node)` for either. Added a regression test for exactly this sequence, plus a second one through `AnnAssign` since it shares the same ordering fix independently.

**2. Restricted receiver extractor still truncates Call/Subscript-rooted chains**

You were right that `_simple_receiver_name` didn't fully close this — the `ast.Attribute` branch still fell back to `return node.attr` whenever the recursive parent came back `None` (unsupported node), instead of propagating the `None`. So `factory().client` and `clients[0].client` both collapsed to plain `"client"`, and when `"client"` also happened to be a traced parameter, `factory().client.flush()` / `clients[0].client.flush()` incorrectly linked to `Store.flush`. Changed the fallback to `if parent is None: return None`, so any unsupported node anywhere in the chain makes the whole receiver unresolved rather than a truncated suffix. Added both nested Call- and Subscript-rooted attribute tests from your report.

**3. Definition-time traversal treats lazy subexpressions as eager**

Confirmed — the enclosing-scope walk added last round visited every subtree under decorators/defaults/annotations/type_params uniformly via `self.visit`, with no distinction between what's actually eager and what only runs later (or never). Fixed in three parts:
- Added `visit_Lambda` (visits only `args.defaults`/`kw_defaults`, never `body`) and `visit_GeneratorExp` (visits only `generators[0].iter`, never `elt`/`ifs`/later `for` clauses) overrides, since neither a lambda body nor a generator's element runs just because the lambda/generator object is created. These are unconditional overrides (not special-cased to definition-time only), so the same eager/lazy split applies everywhere `self.visit` can reach one of these nodes.
- Added a `postponed_annotations` flag, set from `visit_ImportFrom` when it sees `from __future__ import annotations`; parameter and return annotations are skipped entirely (in either scope) once it's set, since CPython never evaluates them at runtime at all in that case.
- Removed the `type_params` bound traversal instead of moving it to the enclosing scope. You're right that PEP 695 bounds are always lazy (CPython compiles each into its own hidden function, called only if something like `T.__bound__` is actually accessed) — visiting them anywhere as if they run "now" was the wrong model regardless of which scope. Replaced the old positive test (which asserted the bound *was* captured) with a negative one, and added negative coverage for postponed annotations and lazy lambda/generator bodies, plus two contrast tests proving the still-genuinely-eager parts (a lambda's own parameter default, a generator's outermost iterable) are still attributed correctly — so this isn't just a blanket suppression.

**Verification**

- All three of your repros confirmed against the unfixed branch first (each failing test reproduced the exact gap you described), then against the fix.
- TDD: 9 new tests + 1 replaced (the PEP 695 test flips from positive to negative). Fail-before/pass-after confirmed by running the new/changed tests against the pre-fix commit (8 of 10 failed for the described reasons; the 2 "still treats ... as eager" contrast tests already passed pre-fix, as expected, since they pin behavior that was never broken).
- Targeted suite (`test_detect_blocking_io_static.py` + `test_scan_changed_blocking_io.py` + `test_detector_repo_root.py` + `blocking_io/test_gate_smoke.py`): 55/56 pass, the 1 failure being the same pre-existing Windows path-separator comparison issue in `test_json_output_uses_concise_review_record_schema` noted last round (reproducible identically on `main`, unrelated to this file).
- Real scanner against the actual backend tree, before/after this commit: still 41/41, byte-identical JSON — these were latent risks in shapes the current codebase doesn't happen to contain, not active miscounts.
- `ruff check .` / `ruff format --check .` clean across the whole backend.
- Updated `backend/AGENTS.md`'s description of the detector to match: RHS-before-target-update ordering, the tightened restricted-extractor behavior for deeper Call/Subscript-rooted chains, postponed-annotation handling, the dropped (not relocated) PEP 695 bound traversal, and the eager/lazy lambda-vs-generator split.

fancyboi999 (changes requested):
[P2] Preserve eager traversal when lazy expressions are consumed immediately

`backend/tests/support/detectors/blocking_io_static.py:486-510`

The unconditional `visit_Lambda` and `visit_GeneratorExp` overrides fix the false positives for a lambda/generator that is merely created, but they also suppress bodies that do execute synchronously in the enclosing async function. For example:

```python
import os

async def route():
    first = (lambda: os.listdir("."))()
    second = list(os.listdir(path) for path in ["."])
    return first, second
```

Both `os.listdir` calls run before `route()` can return control to the event loop. Using the detector's real `scan_source()` API, `origin/main` reports each one as `os.listdir / DIRECT_ASYNC`, while the current head returns an empty list for both. This means the change removes real blocking-I/O findings along with the intended definition-time false positives.

Please scope the lazy traversal to cases where the expression is not executed, and explicitly preserve traversal for immediate execution/consumption contexts (at minimum an immediately invoked lambda and a generator passed to an eager consumer such as `list`). Add positive regression tests for those two cases alongside the existing negative tests for a bare lambda and bare generator.

DaoyuanLi2816:
Follow-up round: found two more real gaps in last round's lazy-suppression logic. Pushed as `f28e9fc`.

**Root cause**

`visit_Lambda`/`visit_GeneratorExp` from last round were unconditional overrides, so they suppressed traversal everywhere they applied — including two shapes where the body/elements genuinely run right away, not lazily:

- `(lambda: os.listdir("."))()` — the lambda is invoked at the exact expression that defines it, so its body executes eagerly, right then.
- `list(os.listdir(path) for path in ["."])` — `list(...)` fully iterates the generator immediately to build the result, unlike an unconsumed generator sitting in a variable.

Confirmed both against `origin/main` (reports both as `os.listdir / DIRECT_ASYNC`) and the pre-fix head (returns `[]` for both) before implementing the fix, then re-confirmed after.

**Fix**

`visit_Call` now marks two specific shapes by node identity, before `generic_visit` runs:
- a `Lambda` used as its own `func` (i.e. immediately invoked), and
- a `GeneratorExp` passed directly as the sole argument to a known eager-consuming builtin (`list`, `set`, `tuple`, `frozenset`, `dict`, `sorted` — a new `EAGER_ITERABLE_CONSUMER_NAMES` constant).

`visit_Lambda`/`visit_GeneratorExp` check that marker by `id()`; on a match they fall through to a full `generic_visit` (body/elements included) instead of the defaults-only / outermost-iterable-only lazy walk. Everything that's still genuinely lazy — merely creating a lambda/generator, assigning it to a variable, passing it as a callback, returning it, or invoking a lambda later *through* a stored variable rather than at its own literal definition site — is unaffected and stays suppressed, matching what round 2 established.

Deliberately excluded from the eager builtin set (documented in a comment): reducer builtins that also consume eagerly (`sum`, `any`, `all`, `min`, `max`) and builtins that wrap the generator in *another* lazy iterator without consuming it (`map`, `filter`, `zip`, `enumerate`) — e.g. `list(map(str, (x for x in gen)))` has the `GeneratorExp` nested one level inside `map(...)`, not as `list`'s direct argument, so it's a separate, still-unrecognized case rather than a false fix.

**Tests**

Added 9 new tests to `test_detect_blocking_io_static.py`:
- `test_scan_file_treats_immediately_invoked_lambda_as_eager` — the IIFE lambda case.
- `test_scan_file_treats_generator_passed_to_list_as_eager` — the `list(generator)` case.
- `test_scan_file_treats_reviewer_repro_both_eager_shapes_together` — both shapes together, each producing its own finding.
- `test_scan_file_treats_generator_passed_to_other_eager_consumers_as_eager` — exercises `set`/`tuple`/`sorted`/`frozenset`/`dict` too, so the rest of `EAGER_ITERABLE_CONSUMER_NAMES` has real coverage instead of shipping unverified.
- `test_scan_file_treats_immediately_invoked_lambda_in_definition_time_default_as_eager_in_enclosing_scope` — an IIFE lambda used as a parameter default, proving this fix composes correctly with the enclosing-scope attribution from the last two rounds (eager *and* attributed to the enclosing function, not the one being defined).
- `test_scan_file_still_treats_bare_uninvoked_lambda_as_lazy` / `test_scan_file_still_treats_bare_unconsumed_generator_as_lazy` — plain-body-level negative regressions (lambda/generator merely created, never invoked/consumed).
- `test_scan_file_still_treats_lambda_invoked_through_stored_variable_as_lazy` — explicit scope-boundary test: storing a lambda then calling it through the variable is a separate `Call` node (`func` is a `Name`, not a `Lambda`), so it's intentionally still not recognized — no cross-variable dataflow tracking, same conservative boundary already used for receiver aliasing.
- `test_scan_file_still_treats_generator_wrapped_in_non_eager_builtin_as_lazy` — explicit scope-boundary test for `map(...)`.

Fail-before/pass-after: the 5 new positive (eager) tests all failed against the unfixed commit (reverted just the source file via a patch, kept the new tests, confirmed each fails for the stated reason, then reapplied). The negative/boundary tests pass unchanged on both sides, as expected.

**Verification**

- Targeted suite (`test_detect_blocking_io_static.py` + `test_scan_changed_blocking_io.py` + `test_detector_repo_root.py`): 60/62 pass; the 2 failures are the same pre-existing, environment-specific issues already on `main` — the Windows path-separator comparison in `test_json_output_uses_concise_review_record_schema` (noted in round 2), and `test_pep695_type_parameter_bound_is_never_treated_as_definition_time_eager` failing on this machine's Python 3.10 (PEP 695 `def f[T: ...]` syntax needs 3.12+) — confirmed both fail identically with or without this fix, so neither is a regression.
- All 4 pre-existing lambda/generator tests from last round (bare lambda body, lambda's own default, bare generator element, generator's outer iterable) still pass unchanged.
- Real scanner against the actual backend tree, before/after this commit: 47/47, byte-identical JSON — no real IIFE-lambda/list-generator shapes exist in the current codebase today, consistent with the "latent risk, not active miscount" pattern from earlier rounds.
- `ruff check .` / `ruff format --check .`: clean across the whole backend (863 files).
- Updated `backend/AGENTS.md`'s description of the lambda/generator eager/lazy split to describe both the general everywhere-not-just-definition-time scope and the two eager exceptions, with the same documented non-goals (stored-then-called lambda, generator wrapped in another lazy iterator).

The over-correction was a real gap in last round's fix, and I've tried to nail the boundary precisely this time rather than trade one partial fix for another.

fancyboi999 (changes requested):
Thanks @DaoyuanLi2816. The prior eager-consumption finding is only fixed for a short list of literal AST shapes; the new global lambda/generator overrides still remove real blocking-I/O findings from ordinary async code.

[P2] Re-scope lazy-expression traversal instead of encoding a partial Python execution model

- Location: `backend/tests/support/detectors/blocking_io_static.py:545-585`
- Problem: `visit_Lambda` and `visit_GeneratorExp` now suppress their bodies everywhere unless the parent call matches one of two special cases. A lambda stored in a local and called on the next line executes synchronously, and reducers such as `sum` consume a generator immediately, but the current head reports neither blocking call. The PR even pins the first false negative in `test_scan_file_still_treats_lambda_invoked_through_stored_variable_as_lazy` and explicitly excludes the eager reducers from `EAGER_ITERABLE_CONSUMER_NAMES`.
- Evidence: against head `f28e9fcb3f74446d1ed0ca525ac86f6f212f813b`, the real `scan_source()` API returns `[]` for both `callback = lambda: os.listdir("."); callback()` and `sum(len(os.listdir(path)) for path in ["."])`. The same probe on current `origin/main` reports two `DIRECT_ASYNC` findings, while a CPython runtime control invokes `os.listdir` in both cases (`runtime_events == [".", "."]`). The PR's focused suite is green (`62 passed`) because it asserts the stored-lambda miss as intended behavior.
- Suggested fix: stop growing an eager-consumer/name-shape allowlist. The false-positive that motivated these overrides was definition-time traversal inside defaults/decorators; handle the eager/lazy split in that narrow definition-time walker and leave ordinary async-function body traversal conservative unless the analyzer gains a coherent local dataflow model that can prove later invocation/consumption. This keeps the fix inside the original invariant instead of reimplementing Python evaluation semantics across the whole AST visitor.
- Test: add detector-vs-runtime regressions for a stored-then-called lambda and at least one eagerly reduced generator (`sum` is sufficient), and confirm the current head no longer loses findings that `origin/main` emits.

Reviewed base SHA: `a028dfd5fb70bd6e26c7dbf9e89543c7c006f9a2`
Reviewed head SHA: `f28e9fcb3f74446d1ed0ca525ac86f6f212f813b`

DaoyuanLi2816:
Going back through this file's actual traversal logic and this PR's own history surfaced a deeper problem than another missed edge case — worth explaining in full rather than just patching in two more cases.

**Confirmed both cases reproduce.** Turned `callback = lambda: os.listdir("."); callback()` and `sum(len(os.listdir(path)) for path in ["."])` directly into the two new/changed tests below, then verified each independently against the pre-fix source by reverting just the source file via a patch (keeping the tests) and re-running: both produced `[]`, matching what the previous head actually returns. Reapplied the fix and confirmed both now report the expected `os.listdir`/`DIRECT_ASYNC` finding.

**Root cause, and why the right fix here is narrower, not broader**

Tracing where `visit_Lambda`/`visit_GeneratorExp` actually came from: this PR's original scope was call-graph/receiver-alias tracing for `self`/`cls` chains. Lambda/generator handling was never the point — it entered only because the enclosing-scope fix for decorators/parameter-defaults/annotations used a plain `self.visit(...)` walk that, as a side effect, also recursed into any lambda body or generator element nested inside a default value, treating it as if it ran at the outer `def` statement's definition time. That's real and narrow: a lambda used as a default value's body doesn't execute when the default object is constructed.

The fix for that got implemented as unconditional `visit_Lambda`/`visit_GeneratorExp` overrides on the whole visitor, so they intercepted *every* lambda/generator anywhere in the traversal, not just the definition-time walk that actually motivated them. That silently flipped the default for ordinary function-body code from "scan everything" (this file's actual behavior everywhere else, and `origin/main`'s behavior before any of this) to "assume lazy unless proven otherwise" — and last round's fix for that (id-keyed marks for exactly two eager shapes, plus `EAGER_ITERABLE_CONSUMER_NAMES` excluding `sum`/`any`/`all`/`min`/`max` because they weren't the two shapes the first fix covered) was a second special case sitting on top of the first, not a boundary. This doesn't stop on its own: it's the same shape-matching problem one layer down, and `sum`/`any`/`all`/`min`/`max` consume their generator argument exactly as eagerly as `list`/`set`/etc. — there's no principled reason they were excluded, it was just that they weren't covered by the first fix yet.

I considered generalizing the alias-tracking machinery already in this file (`local_receiver_alias_stack`) to also track "this name holds a Lambda/GeneratorExp value" through simple local reassignment, which would directly catch the stored-then-called case. I decided against it: it would still miss a lambda stored in a list/dict, passed into another function and called there, returned and called by the caller, etc. — fixing just these two cases without a real dataflow model would just relocate the same unbounded problem into a "smarter" mechanism instead of removing it.

**Fix**

Replaced both the id-marker sets and `EAGER_ITERABLE_CONSUMER_NAMES` with one boolean, `_in_definition_time_expression`, set only while `_visit_function` is walking another function's own decorators/parameter-defaults/annotations/return-annotation, and unset again before that function's body is visited. `visit_Lambda`/`visit_GeneratorExp` apply their lazy (defaults-only / outermost-iterable-only) walk *only* while that flag is set. Everywhere else — module level, class bodies, ordinary function/method bodies — they now fall through to a plain `generic_visit`, unconditionally, with no shape-matching at all. This is `origin/main`'s original behavior for these two node types; the eager/lazy distinction now only exists in the one place it was ever actually needed.

Net effect: a lambda's body or a generator's element is scanned in ordinary code regardless of how it's invoked or consumed — immediately, through a stored variable, via a reducer, wrapped in `map`, or never at all. None of that is special-cased anymore, so there's nothing left to keep growing.

**What this gives up, on purpose**

The one thing the removed machinery could do that the new design can't: an immediately-invoked lambda or an eagerly-consumed generator used *as* another function's own decorator/default/annotation value (e.g. `def helper(flag=(lambda: os.listdir("."))())`) is no longer recognized as eager there — `test_scan_file_does_not_treat_immediately_invoked_lambda_in_definition_time_default_as_eager` now asserts `[]`, flipped from asserting a finding. I think this is the right trade: it's a narrow, definition-time-only gap (nobody writes an IIFE as a parameter default in practice — you'd just call the function directly), and closing it would mean re-introducing exactly the kind of shape-specific exception inside the definition-time walker that got us here. Documented as an explicit limitation in `backend/AGENTS.md` rather than left implicit.

**Tests**

Flipped the four existing "still lazy" tests that were pinning the over-suppression as intended behavior — bare uninvoked lambda, lambda called through a stored variable, bare unconsumed generator, generator wrapped in `map` — to assert a finding instead, and renamed each to say `_as_eager` (`test_scan_file_treats_lambda_invoked_through_stored_variable_as_eager`, etc.). Added `test_scan_file_treats_generator_reduced_by_sum_as_eager` covering the `sum()` reduction case, and extended the eager-consumer test to cover `sum`/`any`/`all`/`min`/`max` alongside the original `list`/`set`/`tuple`/`sorted`/`frozenset`/`dict` in one pass (`test_scan_file_treats_generator_consumed_by_any_builtin_as_eager`). Flipped the one definition-time IIFE-as-default test to a documented negative as described above.

Fail-before/pass-after: reverted just the source fix via a patch file (keeping the new/changed tests), confirmed all 7 changed/new assertions fail against the pre-fix source for the stated reasons, reapplied, confirmed all pass.

**Verification**

- Targeted suite (`test_detect_blocking_io_static.py` + `test_scan_changed_blocking_io.py` + `test_detector_repo_root.py` + `blocking_io/test_gate_smoke.py`): 65/66 pass; the 1 failure is the same pre-existing Windows path-separator comparison in `test_json_output_uses_concise_review_record_schema` noted in earlier rounds, unrelated to this file.
- Full backend suite run on both the pre-fix and post-fix commit: identical 64 pre-existing failures on both, confirmed by diffing the two sorted failure-name lists directly rather than just comparing counts — the diff contains exactly the 7 tests this change touches, nothing else.
- Real scanner against the actual backend tree, before/after: still 41/41, byte-identical JSON — no real instance of these shapes exists in the current codebase today.
- `ruff check .` / `ruff format --check .`: clean across the whole backend (863 files).
- Updated `backend/AGENTS.md` to describe the new scope: lambda/generator laziness applies only inside another function's definition-time expressions, is absolute there (no eager exceptions), and ordinary code scans both unconditionally with no consumption-semantics inference attempted.

Pushed as `1af42f21`.

WillemJiang (approved):

willem-bd (commented):
Reviewed the call-graph / alias-tracing changes by hand against `main`. The core logic is correct; two non-blocking inline nits below, plus a few general notes.

**Verified correct**
- `_simple_receiver_name` refuses the whole chain (returns `None`) when any node isn't Name/Attribute, so `factory().flush()`, `clients[0].flush()`, and `factory().client.flush()` all collapse to unresolvable instead of a trailing-name collision. Sound FP guard.
- `visit_If` may-alias join is order-independent and conservative (union), with no cross-branch leakage; composes correctly with nested `if` and nested `def`.
- RHS-then-target-update ordering in `visit_Assign`/`visit_AnnAssign` matches Python's evaluate-then-bind.
- Definition-time scope (decorators/defaults/annotations visited in the enclosing frame) fixes attribution and prevents the param-seeded alias set from leaking into default evaluation.

**Inline nits**
- `visit_AnnAssign` doesn't honor `postponed_annotations` for its annotation (inconsistent with the function-signature path this PR added).
- `after_orelse` is a live reference rather than a snapshot copy like `after_body`.

**Other**
- Confirmed the three earlier review comments (RHS-before-target ordering, `_simple_receiver_name` chain refusal, and the definition-time eager-vs-lazy traversal incl. PEP 563/695 + lambda/generator bodies) are all addressed in this head.
- Tests are exhaustive — arguably to a fault: `..._reduced_by_sum_as_eager` and `..._passed_to_list_as_eager` are strict subsets of `..._consumed_by_any_builtin_as_eager` and could be folded.
- The AGENTS.md addition is accurate but very dense for an informational, non-CI tool; some of the lambda/generator/PEP 695 nuance might fit better as code comments than the orientation doc.

Overall: solid, low-risk, purely-additive change. Approve with nits.

backend/tests/support/detectors/blocking_io_static.py
```diff
@@ -366,14 +414,28 @@ def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
...
     def visit_AnnAssign(self, node: ast.AnnAssign) -> None:
         self._record_path_like_annotation(node.annotation, [node.target])
         if node.value is not None:
+            # Same RHS-before-target-update ordering as visit_Assign above.
+            self.visit(node.value)
             self._record_sync_http_client_targets(node.value, [node.target])
-        self.generic_visit(node)
+            self._record_local_receiver_alias_targets(node.value, [node.target])
+        self.visit(node.target)
+        self.visit(node.annotation)
```
willem-bd, line 438: Nit (consistency): `_visit_function` skips annotations under `from __future__ import annotations` (`postponed_annotations`, see L581/L602), but this `visit_AnnAssign` visits `node.annotation` unconditionally. With PEP 563 active a variable annotation like `x: time.sleep(1) = None` is never evaluated at runtime (stored as a string), yet this scan would record it as a blocking candidate in the enclosing scope — a latent FP.

Pre-existing in spirit (the old `generic_visit` walked the annotation too) and negligible in practice since annotations are almost never calls, but since this PR introduced postponed-awareness for function signatures, mirroring it here keeps the two paths consistent:

```python
if not self.postponed_annotations:
    self.visit(node.annotation)
```

backend/tests/support/detectors/blocking_io_static.py
```diff
@@ -397,6 +459,33 @@ def visit_With(self, node: ast.With) -> None:
...
+            return
+        before = set(self.current_local_receiver_aliases)
+        for statement in node.body:
+            self.visit(statement)
+        after_body = set(self.current_local_receiver_aliases)
+        self.local_receiver_alias_stack[-1] = set(before)
+        for statement in node.orelse:
+            self.visit(statement)
+        after_orelse = self.current_local_receiver_aliases
+        self.local_receiver_alias_stack[-1] = after_body | after_orelse
```
willem-bd, line 487: Nit: `after_orelse = self.current_local_receiver_aliases` returns the *live* top-of-stack set object, which is then consumed by the union on the next line before the stack entry is replaced. Correct as written — but `after_body` just above is captured as `set(...)` (a copy). Making `after_orelse` a copy too would make the snapshot-vs-reference intent explicit and avoid a footgun if a later edit mutated it after the union:

```python
after_orelse = set(self.current_local_receiver_aliases)
```

PR #4203, opened 2026-07-15
feat(authz): propagate trusted authorization principal context

hata33:
## Why

Phase 0 (#4127) introduced the pluggable authorization protocol and configuration, but runtime authorization still lacked a trusted, consistent `Principal` across Gateway, lead-agent tools, guardrails, and subagents.

Without a server-owned identity chain, downstream authorization providers could not safely distinguish authenticated users, internal callers, channel identities, roles, and provider-specific attributes.

## Root cause

Runtime identity fields were assembled independently at several call sites, and some authorization-relevant values were not propagated through the Gateway trust boundary. There was no pure principal builder that normalized all seven principal fields, and subagent/guardrail paths did not consistently inherit the trusted parent identity.

## What changed

- Added `build_principal_from_context()` in `authz/principal.py` — the single Principal builder that normalizes all 7 fields (user_id, role, oauth_provider, oauth_id, channel_user_id, is_internal, attributes).
- Added `normalize_authz_attributes()` — the single normalization helper shared by the builder, middleware, executor, and task_tool. Ensures consistent `TypeError` for non-Mapping values across all consumption boundaries.
- Made `is_internal` and `authz_attributes` server-owned: Gateway clears client-forged values from both `config["context"]` and `config["configurable"]`, then writes the authoritative `is_internal` from `request.state.auth_source`.
- `channel_user_id` accepted only from internally authenticated callers' top-level `body.context`; cleared from free-form `body.config` sections.
- Propagated `is_internal` and copied `authz_attributes` through task_tool → SubagentExecutor → subagent runtime context. `is_internal` written unconditionally (including `False`).
- Extended `GuardrailRequest` with `channel_user_id`, `is_internal`, `authz_attributes`. Adapter now uses `build_principal_from_context()` (not hand-constructed Principal).
- Added regression coverage for session/internal callers, spoofing attempts, guardrail attribution, subagent inheritance, and real `start_run` anti-forgery.

This PR intentionally does not add Layer 1/Layer 2 filtering, an RBAC provider, provider factories, or execution wiring beyond trusted principal propagation.

## Surface area

- [x] **Backend API** — Gateway run-context trust boundary; no request/response schema change.
- [x] **Agents / LangGraph** — trusted identity propagation through tools, guardrails, and subagents.
- [ ] **Frontend UI**
- [x] **Sandbox** — principal construction uses trusted runtime context.
- [ ] **Skills**
- [ ] **Dependencies**
- [x] **Default behavior change** — Invalid/forged authorization context (`is_internal`, `authz_attributes`, `channel_user_id`) is rejected or cleared at the Gateway boundary. Legitimate callers (session, internal, auth-disabled, scheduler) are unaffected.
- [x] **Docs / tests / CI only** — RFC and implementation guidance updated alongside code.

## Validation

- `cd backend && PYTHONPATH=. uv run pytest tests/test_authorization_principal.py tests/test_authorization_provider.py tests/test_channel_user_id_env.py tests/test_gateway_services.py tests/test_guardrail_middleware.py tests/test_subagent_executor.py tests/test_task_tool_core_logic.py -q --tb=short` -> **323 passed**
- `cd backend && uv run ruff check .` -> **All checks passed**
- `cd backend && uv run ruff format --check .` -> **865 files already formatted**
- Cold-start import smoke: all 5 modified modules import without circular dependency at module level.
- `git diff --check upstream/main...HEAD` -> clean
- CI: all checks pass (backend lint, backend unit tests).

## Pre-landing review

No blocking findings. Security-sensitive context keys are stripped and re-injected at the Gateway trust boundary, and downstream paths consume only the trusted runtime context. Two observable behavior changes in `inject_authenticated_user_context` (raises `TypeError` on non-dict context; writes `is_internal` before early returns) are defensible and tested.

## AI assistance

**Tool used:** OpenAI Codex + ZCode agent

**How it was used:** Codex helped inspect the existing runtime identity flow, implement the trusted principal chain, add regression tests, review the final diff, and prepare this PR. ZCode agent assisted with iterative review-response implementation. The resulting changes were reviewed and validated with the commands above.

- [x] I have read and understand every line of this change and take responsibility for it.

## Documentation

- `backend/AGENTS.md`: links the authorization RFC and cumulative implementation notes so every follow-up authz phase starts from the established trust-boundary decisions.
- RFC, implementation notes, and Phase 1A plans are synchronized with the shipped trusted-principal behavior.

## Related

- Part of #4063
- Builds on #4127


backend/packages/harness/deerflow/authz/principal.py
```diff
@@ -0,0 +1,52 @@
...
+    if resolved_role is None or resolved_role == "":
+        resolved_role = default_role
+
+    raw_attributes = context.get("authz_attributes")
+    if raw_attributes is None:
+        attributes: dict[str, Any] = {}
+    elif isinstance(raw_attributes, Mapping):
+        attributes = dict(raw_attributes)
+    else:
+        raise TypeError(f"authz_attributes must be a Mapping, got {type(raw_attributes).__name__}")
```
willem-bd: This `None -> {} / Mapping -> dict() / else -> TypeError("authz_attributes must be a Mapping...")` normalization block is duplicated verbatim in three other sites added by this PR: `guardrails/middleware.py` (`_build_request`), `subagents/executor.py` (`__init__`), and `tools/builtins/task_tool.py`.

The PR's stated invariant is that all in-process consumption boundaries uniformly raise `TypeError` for non-Mapping `authz_attributes`, but that invariant is now enforced by copy-paste across 4 sites rather than by the builder. A 5th site added in a later phase that silently coerces to `{}` would break it with no test catching the drift.

Consider extracting a small `_normalize_authz_attributes(raw) -> dict` helper used here (inside the builder) and by the three propagation sites, so the builder is the single source of truth for the normalization, not just for `Principal` construction.

backend/tests/test_gateway_services.py
```diff
@@ -1346,3 +1357,160 @@ def test_strip_internal_context_keys_scrubs_config_smuggled_non_interactive():
...
+    return SimpleNamespace(
+        state=SimpleNamespace(
+            auth_source=auth_source,
+            user=SimpleNamespace(id=user_id, system_role=system_role) if user_id else None,
+        ),
+    )
+
+
+def _assemble_authz_run_config(request_config: dict, request, *, body_context: dict | None = None):
+    """Replay the real start_run config sequence for authz identity tests."""
```
willem-bd, line 1463: This helper hand-replays `build_run_config -> merge_run_context_overrides -> strip_internal_context_keys -> inject_authenticated_user_context`. The internal-caller spoofing path IS exercised through real `start_run` (`test_start_run_stamps_internal_owner_guardrail_attribution`), but the session/non-internal anti-forgery assertions are replay-only -- if `start_run`'s ordering or gating drifts, these tests stay green.

Suggestion: add at least one session-caller assertion through the real `start_run` (e.g. assert `context["is_internal"] is False` and that `channel_user_id`/`authz_attributes` cannot be injected via `body.config`). The replay is fine as additional unit coverage.

Minor: `is_internal = request.state.auth_source == "internal"` (line 1386) hardcodes the string `"internal"` instead of importing `AUTH_SOURCE_INTERNAL`, duplicating the source-of-truth constant the production code uses.

backend/AGENTS.md
```diff
@@ -237,6 +237,10 @@ Lead-agent middlewares are assembled in strict order across three functions: the
...
+
+Before changing a later authorization phase, read the [authorization RFC](../docs/plans/2026-07-10-pluggable-authorization-rfc.md) and its [implementation notes](../docs/plans/2026-07-10-pluggable-authorization-implementation-notes.md). The notes are the cumulative handoff record for merged PR behavior, reviewer feedback, trust-boundary decisions, deferred scope, and required regression coverage.
```
willem-bd, line 242: The new implementation-notes doc (`2026-07-10-pluggable-authorization-implementation-notes.md`) and both phase plan docs are written entirely in Chinese, while `backend/AGENTS.md` (which now links them as required reading) and the RFC itself are in English. An English-only reader following this link lands on Chinese prose -- and these notes are positioned as the cumulative handoff record every follow-up authz phase must read.

Not blocking. Worth a consistency pass: either add a short English summary/TOC at the top of the notes, or note the language in the AGENTS.md link so readers know what to expect.

willem-bd (commented):
Reviewed the trusted-principal chain end-to-end against the head commit. The trust-boundary logic is sound -- server-owned keys stripped from both config sections, `is_internal` derived only from `auth_source`, `channel_user_id` honored only for internally-authenticated callers. I verified that IM channels (internal-token path) and the scheduler path both produce `auth_source == AUTH_SOURCE_INTERNAL`, so `channel_user_id`/`is_internal` still propagate for real IM traffic -- no regression for the `DEERFLOW_CHANNEL_USER_ID` sandbox feature. No correctness bug found in the security model; the inline comments are maintainability / test-hardening suggestions.

One item I couldn't anchor inline (it's unchanged code sitting between two diff hunks): `merge_run_context_overrides` has a pre-existing broken docstring fragment around `_CONTEXT_INTERNAL_CALLER_KEYS; those keys are dropped from client requests.` -- it reads like the second half of a deleted sentence. Now that this PR removed the adjacent `channel_user_id` forwarding, anyone tracing `channel_user_id`'s flow lands on that confusing docstring; worth tidying while in the area.

hata33:
@willem-bd Thanks for the review — all three items addressed in `b411bddd`:

**1. Normalization helper extracted.** Added `normalize_authz_attributes(raw) -> dict` in `authz/principal.py`. The builder, `GuardrailMiddleware._build_request`, `SubagentExecutor.__init__`, and `task_tool` now all call it instead of copy-pasting the None/Mapping/else block. Exported from `authz/__init__.py`.

**2. Real `start_run` anti-forgery test + constant fix.**
- Added `test_start_run_session_caller_anti_forgery`: exercises the real `start_run` path (not a replay) with a session caller that forges `is_internal=True`, `authz_attributes`, and `channel_user_id` via `body.config`. Asserts all three are stripped/overwritten.
- Replaced hardcoded `"internal"` strings in `_assemble_authz_run_config` and all `_make_request_with_auth_source("internal")` calls with the imported `AUTH_SOURCE_INTERNAL` constant.

**3. English TOC added to implementation notes.** Added a `Language: Chinese` HTML comment and an English section summary at the top of `docs/plans/2026-07-10-pluggable-authorization-implementation-notes.md` so English readers can navigate before landing on Chinese prose.

310 tests pass, lint/format clean.

backend/packages/harness/deerflow/authz/adapter.py
```diff
@@ -38,22 +44,31 @@ def __init__(
...
         action: str = "call",
     ) -> None:
         self._provider = provider
+        self._default_role = default_role
         self._resource_type = resource_type
         self._action = action
 
     def _to_authz(self, gr: GuardrailRequest) -> AuthzRequest:
         """Map a guardrail request to an authorization request."""
+        principal = build_principal_from_context(
```
willem-bd, line 58: Finding 1 - this is currently dormant end-to-end.

`build_principal_from_context` is called only here (in `GuardrailAuthorizationAdapter._to_authz`), and the adapter is instantiated only in tests - no production code wires it. `authz_attributes` is also always `{}` at runtime because the Gateway always clears it (no Phase 1A producer). So the authorization enforcement path (Layer 1 + Layer 2) doesn't run yet; the immediate live payoff is the `channel_user_id` trust tightening, and `is_internal` is computed/propagated as metadata only.

This matches the PR's stated scope ("intentionally does not add... execution wiring"), so not a defect - flagging so reviewers know CI is the real gate (the author's full-suite run didn't complete) and that no authorization decision is made yet.

backend/packages/harness/deerflow/guardrails/middleware.py
```diff
@@ -40,6 +40,8 @@ def _resolve_context(request: ToolCallRequest) -> dict:
         return context if isinstance(context, dict) else {}
 
     def _build_request(self, request: ToolCallRequest, context: dict) -> GuardrailRequest:
+        from deerflow.authz.principal import normalize_authz_attributes
```
willem-bd: Finding 2 - lazy import repeated in 3 call sites.

`normalize_authz_attributes` is imported lazily inside method bodies in 3 places (here in `GuardrailMiddleware._build_request`, `SubagentExecutor.__init__`, and `task_tool`), while `adapter.py` imports `build_principal_from_context` at module level without issue. If there's a genuine import cycle, a one-line comment explaining it would help; otherwise these can be hoisted to module-level imports. Function-body imports are a smell and add a small per-call cost.

backend/packages/harness/deerflow/authz/adapter.py
```diff
@@ -38,22 +44,31 @@ def __init__(
         self,
         provider: AuthorizationProvider,
         *,
+        default_role: str = "user",
```
willem-bd, line 47: Finding 3 - `default_role="user"` is a Phase 1B placeholder.

The default matches `AuthorizationConfig.default_role` (also `"user"`), so it's safe while the adapter is dormant. When Phase 1B wires the adapter into execution, pass `AuthorizationConfig.default_role` explicitly rather than relying on this default - the docstring already notes the requirement. Not blocking.

backend/app/gateway/services.py
```diff
@@ -298,14 +302,39 @@ def inject_authenticated_user_context(
...
+    Values copied through the free-form RunnableConfig are always cleared.
     """
 
+    # --- Server-owned authorization identity fields ---
+    # Clear any client-forged values from both config sections, then write the
+    # authoritative is_internal. This runs before ALL early returns so that
+    # even user_id-is-None paths get a defined is_internal value.
+    runtime_context = config.setdefault("context", {})
+    if not isinstance(runtime_context, dict):
+        raise TypeError("run context must be a mapping")
```
willem-bd, line 325: Finding 4 - behavior tightening in `inject_authenticated_user_context`.

Two observable changes here: (1) it now raises `TypeError` on a non-dict `config["context"]` instead of silently returning; (2) it always creates `config["context"]` and writes `is_internal` even on the `user_id is None` early-return path that previously returned immediately. Both are defensible (a corrupted context should fail loud) and are tested, but they are behavior changes - worth a release note if any internal caller relied on the old lenient/early-return behavior.

hata33:
@willem-bd Thanks for the second-round review. Responses:

**Finding 1 (dormant path):** Confirmed — the adapter is instantiated only in tests; no production code wires it yet. This matches the PR scope. The immediate live payoff is the `channel_user_id` trust tightening and `is_internal` propagation as metadata. CI (backend unit tests + lint) is the real gate.

**Finding 2 (lazy imports):** Fixed in `88a5c393`. All three lazy imports moved to module level — `normalize_authz_attributes` is now imported at the top of `middleware.py`, `executor.py`, and `task_tool.py`. Verified with a cold-start import smoke test (all 5 modules import without circular dependency).

**Finding 3 (`default_role` placeholder):** Confirmed. Phase 1B wiring will pass `AuthorizationConfig.default_role` explicitly. The docstring already notes this requirement.

**Finding 4 (behavior changes):** Confirmed both changes are defensible and tested:
- `TypeError` on non-dict context: the only production caller is `start_run`, and `build_run_config` already raises `ValueError` on non-Mapping context before reaching `inject_authenticated_user_context`.
- `is_internal` written before early returns: `user_id=None` only occurs in direct test/defensive paths; all production callers (session, internal, auth-disabled, scheduler) construct a user with a valid ID.

PR body updated: 323 tests, CI status, and "Default behavior change" checkbox are now accurate.

hata33:
Resolving all review threads with a summary of actions taken:

**Round 1 threads:**
- **Normalization helper (R1):** Done in `b411bddd` — extracted `normalize_authz_attributes()` shared by all 4 sites.
- **Replay vs real start_run (R1):** Done in `b411bddd` — added `test_start_run_session_caller_anti_forgery` through the real `start_run` path. Replaced hardcoded `"internal"` with `AUTH_SOURCE_INTERNAL`.
- **Chinese docs (R1):** Done in `b411bddd` — English TOC and language note added to implementation notes.

**Round 2 threads:**
- **Dormant path (R2-F1):** Confirmed — no code change needed.
- **Lazy imports (R2-F2):** Done in `88a5c393` — moved to module level, verified with cold-start smoke test.
- **default_role placeholder (R2-F3):** Confirmed — Phase 1B will pass explicit config value.
- **Behavior changes (R2-F4):** Confirmed — both defensible and tested, details in prior comment.

WillemJiang (approved):

Issue #4205, opened 2026-07-15
Frontend Streamdown renders <memory> tag as HTML, causing React console error


qin-chenghan:
## Bug Description

When the model's output contains ```<memory>``` tags (which are part of the memory system prompt format), the frontend's `ClipboardSafeStreamdown` component (in `src/components/ai-elements/streamdown.tsx:62`) tries to render them as HTML elements. React logs a console error:

> The tag `<memory>` is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.

## Reproduction

1. Start DeerFlow with memory enabled.
2. Enter a conversation with the agent.
3. Open the browser console (F12).
4. The error appears whenever the model's output contains `<memory>` tags (e.g. in tool call results, system prompt introspection, or memory context rendering).

## Impact

- Console error noise (functional impact is minimal — the output still renders correctly).
- Adds confusion for users debugging other issues.

## Suggested Fix

In `src/components/ai-elements/streamdown.tsx`, escape or strip unknown HTML tags (like `<memory>`, `</memory>`, `<system-reminder>`, etc.) before passing content to the streamdown renderer, or add a allowlist of known React components / HTML tags.

## Environment

- DeerFlow HEAD: `b68e1c68`
- Frontend: Next.js 16.2.6 via `pnpm dev --port 2026`
- Browser: Observed in Chrome DevTools console

Issue #4204, opened 2026-07-15
Memory config singleton desync: AppConfig auto-reloads but _memory_config does not follow


qin-chenghan:
## Bug Description

When the user changes `memory.mode` in `config.yaml` (e.g. from `"middleware"` to `"tool"`), the `AppConfig` singleton correctly detects the file change and reloads, but `_memory_config` (the singleton used by `get_memory_config()` / `should_use_memory_tools()`) retains the old value.

This means `mode: tool` never actually takes effect at runtime — the agent never receives `memory_search`/`memory_add` etc. tools, and `should_use_memory_tools()` returns `False` despite the config file being correct.

## Root Cause

`AppConfig` (in `app_config.py`) has a file-signature-based auto-reload mechanism (`_app_config_signature`). When it detects a change, `get_app_config()` calls `_load_and_cache_app_config()` which calls `AppConfig.from_file()`, which calls `_apply_singleton_configs()` → `load_memory_config_from_dict()`.

However, `get_memory_config()` returns `_memory_config` directly with no signature check. If the user's code calls `get_memory_config()` **before** `get_app_config()` triggers the reload, the stale singleton is returned. Conversely, if the user calls `get_app_config()` first (which triggers reload), the `_memory_config` IS updated, but only because it happens to be called during the same process.

The real issue: the two singletons are not synchronized. `_memory_config` should check the AppConfig's signature or be invalidated when the app config reloads.

## Reproduction

1. Set `memory.mode: middleware` in `backend/config.yaml`, start the server.
2. Change to `memory.mode: tool` in the config file while server is running.
3. Call `get_memory_config().mode` — returns `"middleware"` (stale).
4. Call `get_app_config().memory.mode` — returns `"tool"` (correct).

## Expected Behavior

`get_memory_config()` should always reflect the current config file state, either by:
- Adding a signature check to `get_memory_config()` similar to `get_app_config()`, or
- Invalidating `_memory_config` when AppConfig reloads, or
- Making `get_memory_config()` delegate to `get_app_config().memory` (avoiding the double-singleton entirely).

## Environment

- DeerFlow HEAD: `b68e1c68`
- Memory backend: DeerMem
- Occurred on `feat/memory-fts5-retrieval` branch (but the bug is in base DeerFlow code, not branch-specific)

PR #4207, opened 2026-07-15
fix(config): sync _memory_config with AppConfig auto-reload + fix(frontend): strip <memory> tags in Streamdown

qin-chenghan:
Two small fixes:

## #4204 — Memory config singleton desync

`get_memory_config()` now calls `get_app_config()` before returning the cached `_memory_config` singleton. `get_app_config()` auto-reloads when the config file signature changes, and its `_apply_singleton_configs()` calls `load_memory_config_from_dict()` to populate `_memory_config`. Without this side effect, changing `memory.mode` in `config.yaml` (e.g. from `"middleware"` to `"tool"`) would never take effect at runtime because `_memory_config` retained the stale value.

## #4205 — Frontend <memory> tag console error

The `ClipboardSafeStreamdown` component renders `<memory>` and `</memory>` as raw HTML tags, causing React to log "unrecognized tag" errors. Strip them before passing content to the renderer.

## Testing

- `cd backend && uvx ruff check` — passes
- The change is minimal and scoped: 5 lines in memory_config.py (local import to avoid circular dependency), 7 lines in streamdown.tsx (strip regex).

### DIFF
--- a/backend/packages/harness/deerflow/config/memory_config.py
+++ b/backend/packages/harness/deerflow/config/memory_config.py
@@ -102,6 +102,14 @@ _memory_config: MemoryConfig = MemoryConfig()
 
 def get_memory_config() -> MemoryConfig:
     """Get the current memory configuration."""
+    # Ensure AppConfig signature is fresh before returning _memory_config.
+    # get_app_config() auto-reloads on file change and calls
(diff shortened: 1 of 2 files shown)


qin-chenghan:
Splitting into two separate PRs per review — backend fix and frontend fix are independent changes.

PR #4208, opened 2026-07-15
fix(config): sync _memory_config with AppConfig auto-reload

### LINKED ISSUE #4204: Memory config singleton desync: AppConfig auto-reloads but _memory_config does not follow

qin-chenghan:
When `memory.mode` is changed in `config.yaml` (e.g. from `"middleware"` to `"tool"`), `get_memory_config()` returns the stale cached value because `_memory_config` is a singleton with no invalidation mechanism.

`AppConfig` already has a file-signature-based auto-reload (`get_app_config()`), and its `_apply_singleton_configs()` calls `load_memory_config_from_dict()` to populate `_memory_config`. The fix is to call `get_app_config()` before returning `_memory_config` — this triggers the auto-reload if the config file has changed.

```python
def get_memory_config() -> MemoryConfig:
    from deerflow.config.app_config import get_app_config as _get_app_config
    _get_app_config()  # triggers auto-reload + _memory_config update
    return _memory_config
```

Closes #4204


backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -102,6 +102,14 @@ def should_use_memory_tools(config: MemoryConfig) -> bool:
 
 def get_memory_config() -> MemoryConfig:
     """Get the current memory configuration."""
+    # Ensure AppConfig signature is fresh before returning _memory_config.
```
willem-bd: **Missing test for the actual fix.** The existing `get_memory_config().enabled` assertions in `tests/test_app_config_reload.py` (lines ~396, 410) are always preceded by an explicit `get_app_config()` call (lines 393, 407) -- and that call is what triggers the reload + `_apply_singleton_configs()` -> `load_memory_config_from_dict()` sync. So that test passes **without** this change.

What this PR actually adds -- `get_memory_config()` self-syncing with **no** prior `get_app_config()` (the background memory path in `updater`/`queue`/`storage`, which never calls `get_app_config()` directly) -- is untested.

Per `AGENTS.md` (TDD mandatory for backend), suggest a case: write `memory.enabled: false`, `reset_app_config()`, call `get_memory_config()` directly (no `get_app_config()`), assert `enabled is False`; then mtime-bump + rewrite to `true`, call `get_memory_config()` again, assert `True`.

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -102,6 +102,14 @@ def should_use_memory_tools(config: MemoryConfig) -> bool:
...
+    # get_app_config() auto-reloads on file change and calls
+    # load_memory_config_from_dict() during _apply_singleton_configs(),
+    # so this side effect keeps _memory_config in sync.
+    # Local import to avoid circular dependency (app_config -> memory_config).
+    from deerflow.config.app_config import get_app_config as _get_app_config
+
+    _get_app_config()
```
willem-bd: **Performance: every call now reads + hashes the whole `config.yaml`.** `get_app_config()` computes `_get_config_signature()` on **every** invocation -- a `stat()` plus a full read of `config.yaml` (1 MB chunks) to SHA256 it -- even when nothing changed (the mtime is computed but only used for logging, not as a fast-path skip; verified at base `app_config.py:583-586`).

`get_memory_config()` is called several times per turn (memory middleware, summarization hook, lead-agent prompt build) and repeatedly per memory update cycle (`updater`/`queue`/`storage`). For a small config this is sub-ms, but it's real synchronous disk I/O -- including in async request paths -- where this was previously a pure `return _memory_config`.

Worth at least an mtime+size fast-path in `get_app_config()` so the full hash is skipped on the unchanged hot path; that would also benefit existing callers.

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -102,6 +102,14 @@ def should_use_memory_tools(config: MemoryConfig) -> bool:
 
 def get_memory_config() -> MemoryConfig:
     """Get the current memory configuration."""
+    # Ensure AppConfig signature is fresh before returning _memory_config.
+    # get_app_config() auto-reloads on file change and calls
+    # load_memory_config_from_dict() during _apply_singleton_configs(),
+    # so this side effect keeps _memory_config in sync.
```
willem-bd: **Scope: the same stale-on-reload bug exists in the sibling singletons.** `get_title_config`, `get_summarization_config`, `get_subagents_app_config`, `get_tool_search_config`, and `get_guardrails_config` all just `return _<x>_config` -- they're refreshed only inside `_apply_singleton_configs()`, so they go stale whenever a caller hits them without first going through `get_app_config()`.

This PR fixes only `memory`. If the bug is worth fixing (#4204 says yes), the others exhibit identical symptoms when their sections are edited at runtime. Either fix them the same way, or add a one-line note on the PR explaining why memory alone is in scope (e.g. `memory.mode` is the field actually toggled via the runtime config-edit API).

willem-bd (commented):
Review of `get_memory_config()` -> `get_app_config()` sync fix.

Correctness verified: the sync chain works (`get_memory_config` -> `get_app_config` -> on signature change `_load_and_cache_app_config` -> `AppConfig.from_file` -> `_apply_singleton_configs` -> `load_memory_config_from_dict`), with no infinite recursion (the reload path writes `_memory_config` directly, never reads it back) and the function-local import correctly breaks the `app_config -> memory_config` cycle. The comment is accurate and there's precedent for the local-import pattern (`checkpointer_config.ensure_config_loaded`).

Three inline findings below: a missing test for the actual self-sync scenario, a per-call file-read/hash cost, and a scope/consistency note.

backend/packages/harness/deerflow/config/memory_config.py
qin-chenghan: Not addressed in this PR. The per-call SHA256 digest is the existing behaviour of `get_app_config()` itself, shared by all callers of that function. An mtime+size fast-path in `get_app_config()` would benefit all consumers and is a separate optimisation concern. The per-call overhead is negligible for a small config.yaml (< 1 ms).

backend/packages/harness/deerflow/config/memory_config.py
qin-chenghan: Acknowledged. This PR is scoped to the memory config desync reported in #4204. The sibling singletons (`get_title_config`, `get_summarization_config`, etc.) exhibit the same pattern and should be fixed in a follow-up PR once the pattern is validated here.

backend/packages/harness/deerflow/config/memory_config.py
qin-chenghan: Addressed by the latest commit: the _app_config is not None guard prevents the side-effect issue (tests that never call get_app_config() won't trigger a reload), and try/except FileNotFoundError handles the no-config-file case silently. A dedicated self-sync regression test (test_get_memory_config_follows_config_file_reload) is already in he-yufeng's #4216. Keeping this PR focused on the core fix.

qin-chenghan:
Hi @WillemJiang — UT fix is in on commit 2b29f324 (the FileNotFoundError + _app_config-is-not-None guard). Local `make lint` + 117 relevant tests pass; the one deselected test (`test_returns_in_memory_saver_when_not_configured`) is pre-existing on main and unrelated. CI is blocked on `action_required` for fork approval — could you re-run?

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -101,7 +101,30 @@ def should_use_memory_tools(config: MemoryConfig) -> bool:
...
+    as a side effect of the first access to ``get_memory_config()``, which would
+    break callers that expect module-level defaults (e.g. unit tests).
+    """
+    # Lazy import: app_config imports this module, so a top-level import cycles.
+    from .app_config import _app_config, get_app_config
+
+    if _app_config is not None:
+        try:
+            get_app_config()
+        except FileNotFoundError:
```
willem-bd: **Risk: broken-config edits now crash per-turn hot paths.** `get_app_config()` can raise more than `FileNotFoundError`: `yaml.YAMLError` (broken YAML mid-edit), `pydantic.ValidationError` (schema violation after edit), or `ValueError` (`resolve_env_variables` on a missing `$ENV_VAR`).

Previously `get_memory_config()` never raised — it returned the last good cached singleton, so an in-flight turn completed fine even if the operator saved a malformed config. After this change, the same mid-edit failure propagates out of every per-turn call site (`memory_middleware`, `summarization_hook`, `lead_agent/prompt.py`), turning a recoverable degraded state into a crash.

Consider broadening the catch so a transiently-broken file falls back to the existing `_memory_config` (preserving the old "last good value" semantics), e.g. `except (FileNotFoundError, ValueError, yaml.YAMLError, ValidationError): pass`. Worth an explicit decision rather than the `FileNotFoundError`-only choice.

backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -101,7 +101,30 @@ def should_use_memory_tools(config: MemoryConfig) -> bool:
...
+    of returning the in-memory singleton.  This avoids picking up a config file
+    as a side effect of the first access to ``get_memory_config()``, which would
+    break callers that expect module-level defaults (e.g. unit tests).
+    """
+    # Lazy import: app_config imports this module, so a top-level import cycles.
+    from .app_config import _app_config, get_app_config
+
+    if _app_config is not None:
+        try:
+            get_app_config()
```
willem-bd, line 145: **Missing regression test for the isolated path.** The existing coverage in `tests/test_app_config_reload.py:393-410` calls `get_app_config()` *before* `get_memory_config()`, so it exercises the already-working path and would **not** catch #4204.

A test pinning this fix should call `get_memory_config()` in isolation — after mutating `memory.mode`/`enabled` in the file, with **no** intervening `get_app_config()` call — and assert the new value is reflected. Without it, this fix can silently regress.

backend/packages/harness/deerflow/config/memory_config.py
qin-chenghan: Broadened to `except Exception` with `logger.warning(exc_info=True)` — any transient config error (YAMLError, ValidationError, ValueError, or anything else) now falls back to the last-good singleton. No new imports needed. Added `test_get_memory_config_falls_back_on_broken_config` to pin the behavior.

backend/packages/harness/deerflow/config/memory_config.py
qin-chenghan: Added `test_get_memory_config_self_syncs_without_prior_get_app_config` — calls `get_memory_config()` in isolation after mutating the file, with no intervening `get_app_config()` call. Asserts the new value is reflected. All 16 tests pass.

qin-chenghan:
3d7d0dea addresses both review findings:

1. Broken-config crash — broadened except FileNotFoundError to except Exception (logger.warning with exc_info=True), so transiently broken config.yaml falls back to the last-good singleton.

2. Missing regression test — added test_get_memory_config_self_syncs_without_prior_get_app_config (isolated get_memory_config() after file mutation) and test_get_memory_config_falls_back_on_broken_config. All 16 tests pass, make lint passes.

@WillemJiang  PTAL when you get a chance — CI re-trigger is in progress.

WillemJiang (approved):

PR #4209, opened 2026-07-15
fix(frontend): strip <memory> tags in Streamdown to avoid React console error

### LINKED ISSUE #4205: Frontend Streamdown renders <memory> tag as HTML, causing React console error

qin-chenghan:
The `ClipboardSafeStreamdown` component passes `<memory>` and `</memory>` tags through to the renderer, which React interprets as unknown HTML elements and logs a console error:

> The tag `<memory>` is unrecognized in this browser.

Strip this system-level tag from the content before passing it to the Streamdown component. The tag is only used in the memory system prompt format and has no semantic meaning in rendered output.

Closes #4205

### DIFF
--- a/frontend/src/components/ai-elements/streamdown.tsx
+++ b/frontend/src/components/ai-elements/streamdown.tsx
@@ -3,6 +3,7 @@
 import { Component, type ComponentProps, type ReactNode } from "react";
 import { Streamdown } from "streamdown";
 
+import { stripLeakedSystemTags } from "@/core/streamdown/preprocess";
 import { installClipboardFallback } from "@/core/clipboard";
 
 export type ClipboardSafeStreamdownProps = ComponentProps<typeof Streamdown>;
@@ -57,9 +58,15 @@ export function ClipboardSafeStreamdown({
   children,
   ...props
 }: ClipboardSafeStreamdownProps) {
+  // Strip leaked system-internal tags (<memory>, <system-reminder>, etc.)
+  // that would cause React to log "unrecognized tag" console errors when
+  // the markdown renderer passes them through as raw HTML.
+  const sanitizedChildren =
+    typeof children === "string" ? stripLeakedSystemTags(children) : children;
+
   return (
-    <StreamdownFallbackBoundary raw={children}>
-      <Streamdown {...props}>{children}</Streamdown>
+    <StreamdownFallbackBoundary raw={sanitizedChildren}>
+      <Streamdown {...props}>{sanitizedChildren}</Streamdown>
     </StreamdownFallbackBoundary>
   );
 }
--- a/frontend/src/core/streamdown/preprocess.ts
+++ b/frontend/src/core/streamdown/preprocess.ts
@@ -357,12 +357,12 @@ export function stripLeakedSystemTags(markdown: string): string {
     .map((line) => {
       const fenceMatch = FENCE_MARKER_RE.exec(line);
       if (fenceMatch) {
-        const marker = fenceMatch[1];
+        const marker = fenceMatch[1]!;
         if (fenceMarker === null) {
           // Opening a fenced code block
           fenceMarker = marker;
         } else if (
-          marker[0] === fenceMarker[0] &&
+          marker.startsWith(fenceMarker.charAt(0)) &&
           marker.length >= fenceMarker.length
         ) {
           // Closing fence: same character and at least as long as opener
--- a/frontend/src/core/streamdown/preprocess.ts
+++ b/frontend/src/core/streamdown/preprocess.ts
@@ -326,6 +326,10 @@ const _INTERNAL_TAG_RE = new RegExp(
   "g",
 );
 
+// Regex matching the start/end of a fenced code block (3+ backticks or tildes).
+// Captures the marker string so we can compare character and length.
+const FENCE_MARKER_RE = /^ {0,3}(`{3,}|~{3,})/;
+
 /**
  * Strip leaked system-internal HTML tags from markdown content.
  *
@@ -339,18 +343,36 @@ const _INTERNAL_TAG_RE = new RegExp(
  * Code-aware: tags inside fenced code blocks (````` `````) and indented code
  * blocks (4-space indent) are left untouched, so user-written meta-discussions
  * about the memory system are not silently stripped.
+ *
+ * Fence tracking is marker-aware (tracking the opening character and run
+ * length) so that a tilde-fenced block containing a shorter backtick run, or a
+ * 4-backtick block containing a 3-backtick run, does not prematurely close the
+ * fence.
  */
 export function stripLeakedSystemTags(markdown: string): string {
   const lines = markdown.split("\n");
-  let insideFence = false;
+  let fenceMarker: string | null = null;
 
   return lines
     .map((line) => {
-      if (CODE_FENCE_RE.test(line)) {
-        insideFence = !insideFence;
+      const fenceMatch = FENCE_MARKER_RE.exec(line);
+      if (fenceMatch) {
+        const marker = fenceMatch[1];
+        if (fenceMarker === null) {
+          // Opening a fenced code block
+          fenceMarker = marker;
+        } else if (
+          marker[0] === fenceMarker[0] &&
+          marker.length >= fenceMarker.length
+        ) {
+          // Closing fence: same character and at least as long as opener
+          fenceMarker = null;
+        }
+        // Otherwise: different fence type or shorter run inside a fence
+        // (e.g. ``` inside ~~~~, or `` inside `````) — stay inside.
         return line;
       }
-      if (insideFence || INDENTED_CODE_RE.test(line)) {
+      if (fenceMarker !== null || INDENTED_CODE_RE.test(line)) {
         return line;
       }
       return line.replace(_INTERNAL_TAG_RE, "");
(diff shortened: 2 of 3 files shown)


frontend/src/components/ai-elements/streamdown.tsx
```diff
@@ -57,9 +57,17 @@ export function ClipboardSafeStreamdown({
   children,
   ...props
 }: ClipboardSafeStreamdownProps) {
+  // Strip system-level memory tags (<memory>, </memory>) that would cause
+  // React to log "unrecognized tag" console errors when the markdown
+  // renderer passes them through as raw HTML.
+  const sanitizedChildren =
+    typeof children === "string"
+      ? children.replace(/<\/?memory>/g, "")
```
willem-bd: This strips only `<memory>`/`</memory>`, but a leaked dynamic-context reminder is a whole `<system-reminder><memory>…</memory><current_date>…</current_date></system-reminder>` block (see `backend/.../middlewares/dynamic_context_middleware.py`). If `<memory>` reached the renderer, the sibling `<system-reminder>` and `<current_date>` tags almost certainly did too — and each independently triggers the same "unrecognized tag" warning, so the console error would persist for them.

`frontend/src/core/messages/utils.ts` already keeps `INTERNAL_MARKER_TAGS = ["uploaded_files","slash_skill_activation","system-reminder","memory","current_date"]` for this exact defense-in-depth case. Deriving the strip from that list (instead of hardcoding `"memory"`) would cover all leaked internal tags and prevent the two from drifting apart.

Minor: the pattern misses `<memory/>` (self-closing), `<memory attr="x">` (attributed), and case variants — fine today since the backend emits none of these, but worth a one-line comment if this is meant as a backstop.

willem-bd (commented):
Fixes a real React "unrecognized tag" console warning from `<memory>` tags reaching Streamdown. The change is small and low-risk, correctly guards non-string children, and applies the strip to both the fallback `raw` and the rendered `<Streamdown>` path. Placement in `ClipboardSafeStreamdown` is well-chosen — it covers `MessageResponse`/`ReasoningContent`/`SafeStreamdown`, which the `safe-children` hook alone would not.

Main concerns (details inline):
- **One-tag fix** — sibling `<system-reminder>`/`<current_date>` from the same leaked block would still warn; `INTERNAL_MARKER_TAGS` already enumerates them.
- **Code-blind strip** (no fence/code guard like `preprocess.ts`) — can strip `<memory>` a user legitimately posts in code or a meta-discussion.
- **No test**, despite the repo's TDD convention.

Recommend a code-aware `stripLeakedSystemTags` in `preprocess.ts` covering all `INTERNAL_MARKER_TAGS` + unit tests, and a follow-up to find why a `hide_from_ui` reminder reaches the renderer in the first place.

frontend/src/components/ai-elements/streamdown.tsx
```diff
@@ -57,9 +57,17 @@ export function ClipboardSafeStreamdown({
   children,
   ...props
 }: ClipboardSafeStreamdownProps) {
+  // Strip system-level memory tags (<memory>, </memory>) that would cause
+  // React to log "unrecognized tag" console errors when the markdown
+  // renderer passes them through as raw HTML.
+  const sanitizedChildren =
```
willem-bd, line 64: This blanket `.replace` runs across the whole string, including inside fenced/indented code blocks and inline code. The functions in `preprocess.ts` deliberately track `CODE_FENCE_RE`/`INDENTED_CODE_RE` to avoid rewriting `>`-patterns inside code. A user posting `<memory>…</memory>` in a code block, or in a meta-discussion about the memory system, would have their tags silently stripped — `utils.ts` (around the `INTERNAL_MARKER_TAGS` / `stripInternalMarkers` block) explicitly anticipates users typing `<memory>` in the UI.

Suggestions:
- Gate the strip the same way `preprocess.ts` does (skip fenced/indented-code lines), or better, extract it into a named, code-aware fn in `preprocess.ts` (e.g. `stripLeakedSystemTags`) so it composes with `capMarkdownNesting`/`normalizeStreamdownMathMarkdown` and reuses the existing fence helpers.
- Hoist the regex to a module-level constant (`const MEMORY_TAG_RE = …`), matching `preprocess.ts` (`CODE_FENCE_RE`, …) and `utils.ts` (`INTERNAL_MARKER_RE`); the inline literal also rebuilds a `RegExp` on every render.

frontend/src/components/ai-elements/streamdown.tsx
```diff
@@ -57,9 +57,17 @@ export function ClipboardSafeStreamdown({
   children,
   ...props
 }: ClipboardSafeStreamdownProps) {
+  // Strip system-level memory tags (<memory>, </memory>) that would cause
```
willem-bd: No test for this new behavior. `AGENTS.md` (Cross-Cutting Conventions: "features and bug fixes ship with tests") plus the thorough `frontend/tests/unit/core/streamdown/preprocess.test.ts` suggest it should be covered. Worth locking in: tags stripped, inner content preserved, multiple occurrences, non-string passthrough — and the no-op-inside-code case if a code-aware variant is adopted. (This is also an argument for extracting to `preprocess.ts`, where it's trivially unit-testable.)

frontend/src/components/ai-elements/streamdown.tsx
qin-chenghan: Addressed in the latest commit: stripLeakedSystemTags() now covers all INTERNAL_MARKER_TAGS (<memory>, <system-reminder>, <current_date>, <uploaded_files>, <slash_skill_activation>) via a single regex built from the shared tag list, so they won't drift apart.

frontend/src/components/ai-elements/streamdown.tsx
qin-chenghan: Addressed in the latest commit: 15 unit tests added in preprocess.test.ts covering tag stripping, code-block preservation, self-closing/attributed tags, multiple occurrences, mixed tags, empty/plain text, and multiple fences.

frontend/src/components/ai-elements/streamdown.tsx
qin-chenghan: Addressed in the latest commit: the strip is now code-aware via stripLeakedSystemTags() in preprocess.ts, which skips fenced code blocks and indented code blocks (4-space indent) using the same CODE_FENCE_RE/INDENTED_CODE_RE pattern as the other functions in that module. The regex is also hoisted to a module-level constant (_INTERNAL_TAG_RE).

qin-chenghan:
Hi @WillemJiang — all three review notes addressed in commit 80edc450: strip now covers all INTERNAL_MARKER_TAGS (not just <memory>), is code-aware (skips fenced + indented code blocks), and comes with 15 unit tests. eslint + prettier + 40 tests all pass locally. CI is blocked on action_required for fork approval — could you re-run?

fancyboi999 (changes requested):
Thanks @qin-chenghan. I found one issue that should be addressed before this is ready.

[P2] Track the actual fence delimiter before treating content as code

- Location: `frontend/src/core/streamdown/preprocess.ts:349-350`
- Problem: The boolean fence tracker exits a code block on any line that starts with either three backticks or three tildes. Valid Markdown can contain a different fence marker, or a shorter run of the same marker, inside the code block; tags after that line are then silently removed from literal code.
- Evidence: DeerFlow uses marked 16.2.1, whose fence rule binds the closing run to the opening delimiter. For example, in `~~~~md\n```\n<memory>literal</memory>\n~~~~`, marked keeps the backtick line and `<memory>` text inside the tilde-fenced block, but `stripLeakedSystemTags()` flips `insideFence` off at the backticks and changes the code to `literal`. A four-backtick fence containing a triple-backtick example fails the same way. The added tests only exercise matching triple-backtick fences.
- Suggested fix: Track the opening marker character and run length, and only close on the same marker with a run at least as long, or preserve code ranges from the Markdown lexer.
- Test: Add tilde-fenced content containing a backtick fence and a four-backtick block containing triple backticks; assert internal tags remain literal, while tags after the real closing fence are still stripped.

Reviewed head SHA: `80edc4505052fc9055b3307cf8f5d8d3177d3480`

WillemJiang (approved):

Issue #4123, opened 2026-07-12
Inbound message dedup is per-pod; webhook redeliveries routed to a different replica can double-execute under HA (follow-up to #4104)


DaoyuanLi2816:
## Context

Following up on @zhfeng's review note in #4104. That PR routes GitHub webhook redeliveries through the manager-level inbound dedup (originally added for the IM channels in #3584), so a GitHub "Redeliver" / retry-on-timeout no longer re-runs the agent within a single gateway process. On approving it, @zhfeng flagged that this dedup no longer holds once the gateway runs with replicas ≥ 2 for HA:

> it remind me that if the gateway is deployed as replica >= 2 in a k8s cluster for example and want to support HA. I think it could still be an issue if the redelivery message route to a different pod even after the TTL 10min expired. It needs to investigate this case in the future. There is a PR which could resolve it partially. So it maybe similar improvement for the dedupe. — #4003

He's right, and this is the agreed follow-up. The observation generalizes beyond GitHub: the dedup lives in `ChannelManager` and is shared by every channel, so any channel whose provider redelivers events is affected, not just GitHub.

## The gap

The inbound dedup is entirely **per-process / in-memory**. In `backend/app/channels/manager.py`:

- The seen-set is `self._recent_inbound_events: OrderedDict[tuple[str, str, str, str], float]` — an in-process `OrderedDict` mapping the dedupe key to a `time.monotonic()` timestamp. Nothing is shared across processes; the stored value is itself a per-process monotonic-clock reading, so it isn't even comparable between pods.
- The dedupe key (`_inbound_dedupe_key`) is the 4-tuple `(channel_name, workspace_id, chat_id, message_id)`, where `message_id` is read from `event_id` / `message_id` / `msg_id` in metadata (GitHub supplies the `X-GitHub-Delivery` GUID via #4104). It fails open (no dedup) when either `message_id` or `workspace_id` is absent.
- TTL is `INBOUND_DEDUPE_TTL_SECONDS = 10 * 60` — the "10min" @zhfeng referenced — with lazy front-eviction in `_is_duplicate_inbound` and an `INBOUND_DEDUPE_MAX_ENTRIES = 4096` overflow cap.

Under replicas ≥ 2 behind a load balancer, a provider redelivery of the same event can land on a **different** pod than the original:

1. Pod A handles delivery `d`, records key `k` in *its own* `_recent_inbound_events`, and runs the agent — with real side effects (a PR comment, an outbound reply).
[...]

PR #4210, opened 2026-07-15
feat(channels): share inbound webhook dedupe across pods via Postgres

### LINKED ISSUE #4120: [feat] Inbound webhook dedupe is per-pod; cross-pod post-completion redeliveries still duplicate
## Context

- PR #4104 routes GitHub webhook redeliveries through `ChannelManager._is_duplicate_inbound`, bringing GitHub to parity with the other IM channels (Slack/Telegram/Discord/WeChat/WeCom/DingTalk), which already keyed dedupe on a top-level `metadata["message_id"]` + a workspace id.
- #4003 closed the cross-process *run-creation* race via the `uq_runs_thread_active` partial unique index on `(thread_id) WHERE status IN ('pending','running')`, enforced by Postgres and shared across pods.

## The remaining gap

`_recent_inbound_events` is an in-memory `OrderedDict` per `ChannelManager` instance (`app/channels/manager.py:841`), TTL 10 min (`INBOUND_DEDUPE_TTL_SECONDS = 600`). It is **not shared across pods**. Combined with #4003, webhook-redelivery coverage today is:

| Redelivery window | Cross-pod protected? | By what |
|---|---|---|
| While first run is active (pending/running) | ✅ Yes | #4003 `uq_runs_thread_active` + GitHub's `multitask_strategy: "reject"` → `ConflictError` dropped gracefully |
| After first run completed (delayed auto-retry, manual "Redeliver") | ❌ No | Only per-pod/10-min dedupe; a different pod or >10 min → second run → duplicate side effect (e.g. a duplicate PR comment) |
| After pod restart/rollout | ❌ No | In-memory table wiped |

This affects **every channel**, not just GitHub — GitHub just produces the most visible duplicate (a second PR comment posted via `gh` mid-run).

## Proposed fix

Swap the backing of `_recent_inbound_events` (already isolated behind `_is_duplicate_inbound` / `_release_inbound_dedupe_key`) for a shared store, behind a config seam (mirror `memory.storage_class`). Two options:

1. **Redis** — `SET <key> 1 EX 600 NX` on receipt (returns 1 = new, 0 = duplicate), `DEL` on the failure-release path. Reuses the existing `redis.asyncio.Redis` client + `DEER_FLOW_STREAM_BRIDGE_REDIS_URL` from `runtime/stream_bridge/redis.py`. Native TTL removes the manual `OrderedDict` expiry sweep. This is a single idempotency-key primitive, not a stream.
2. **Postgres** — a `webhook_deliveries(dedupe_key text PK, first_seen timestamptz default now())` table, `INSERT ... ON CONFLICT (dedupe_key) DO NOTHING` (0 rows inserted = duplicate), plus a periodic `DELETE WHERE first_seen < now() - interval '10 min'`. Zero new infra — Postgres is already required for multi-worker (#3960) and is consistent with #4003's DB-enforced style. Cost: a write per webhook (low volume) + a cleanup job.

Either way, **fail open** to the current per-pod behavior if the store is unavailable — failing closed would 503 the webhook and trigger more retries.

## Scope note

A 10-min shared TTL is the right ceiling: it kills automatic / double-click duplicates cross-pod while still letting an intentional late manual "Redeliver" re-run. A permanent no-expiry idempotency log is a different, heavier tradeoff (storage growth + wrong UX for intentional re-runs) and is out of scope.

Relates to: #3948 (closed; multi-worker run correctness), #4003 (merged; in-flight run atomicity), #4104 (open; per-pod GitHub dedupe).

ech0hol:
Fixes #4120

## Why

- **Trigger**: #4104 added per-pod inbound webhook dedupe (it fills `metadata["message_id"]` and routes through `ChannelManager._inbound_dedupe_key` to drop provider redeliveries). But the store is in-process, so it cannot see a redelivery that lands on another replica. #4123 (now closed) was opened as the explicit follow-up to #4104 for exactly this gap; this PR delivers it under #4120.
- **Pain addressed**: In a multi-replica deployment the same inbound webhook (GitHub delivery redelivery, Slack/Telegram/Feishu/DingTalk retry, etc.) can trigger **two agent runs / two final answers**, doubling cost and sending duplicate responses to users. This PR makes dedupe state **shared across pods** so a redelivery is still dropped as a duplicate no matter which replica receives it.

## What changed

- New `InboundDedupeStore` protocol with two implementations (`backend/app/channels/dedupe_store.py`):
  - `MemoryInboundDedupeStore` — in-process, preserves the exact pre-#4120 per-pod behavior (backward compatible).
  - `PostgresInboundDedupeStore` — shared across pods via the application DB.
- **Atomic dedup**: `INSERT ... ON CONFLICT (channel, workspace_id, chat_id, message_id) DO NOTHING RETURNING channel`. The returned row distinguishes *insert succeeded* (new delivery → proceed) from *conflict* (duplicate → drop). Conflict detection uses the returned row rather than `rowcount`, for driver independence.
- **Lazy TTL cleanup** of rows older than `INBOUND_DEDUPE_TTL_SECONDS` (default 10 min) runs in the **same transaction** as the insert — no background task, no separate sweep.
- **Fail-open**: any DB error makes `try_record` return *allow* (proceed) so a storage outage never drops a webhook or returns 5xx. (This was the key correctness property — a bug where DB errors dropped **all** inbound webhooks was caught and fixed during review; covered by `test_postgres_try_record_fail_open_on_exception` and `test_postgres_try_record_fail_open_when_no_session_factory`.)
- **Config** (`dedupe_storage` backend `auto` / `memory` / `postgres` in `config.example.yaml` + `app_config.py`):
  - `auto` selects Postgres **whenever `database.backend=postgres`** — this covers the common multi-replica K8s case where each pod runs `GATEWAY_WORKERS=1` but all pods share one Postgres DB. (Earlier gating on `multi_worker` was removed because it silently disabled cross-pod dedupe for that exact deployment.)
  - Explicit `postgres` on a non-Postgres DB logs a **WARNING** and falls back to the memory store, so misconfiguration is visible rather than silent.
- **Migration `0005_webhook_dedupe`** (idempotent): creates `webhook_deliveries` with a composite PK mirroring `_inbound_dedupe_key`. Bootstrap `create_all` parity is preserved (empty-DB path still provisions the table).

## Surface area

- [x] **Backend API** — new `dedupe_storage` config + `InboundDedupeStore` injection into `ChannelManager` / `ChannelService`
- [x] **Default behavior change** — Backward compatible: `auto` resolves to the memory store for non-Postgres DBs, so existing single-pod behavior is unchanged; cross-pod sharing only activates on a Postgres DB

## Bug fix verification

This PR is a new feature, not a bug fix, so the failing-test gate does not apply. (It does, however, fix a latent fail-open bug introduced during development where a DB error caused `try_record` to drop all webhooks — encoded by the two `fail_open` tests above.)

## Validation

Backend (Linux dev container, `docker exec deer-flow-gateway` → `cd /app/backend`):

- `make lint` (ruff check + `ruff format --check`): **pass**
- `make test`:
  - `tests/test_inbound_dedupe.py` — store + factory unit tests (incl. atomic ON CONFLICT, lazy cleanup, fail-open, auto/postgres resolution): **pass**
  - `tests/test_migration_0005_webhook_dedupe.py` — `0005` creates `webhook_deliveries` with exactly `{channel, workspace_id, chat_id, message_id, first_seen}` and is idempotent via `bootstrap_schema`: **pass**
  - `tests/test_multi_pod_inbound_dedupe.py` — **real Postgres cross-pod integration**: **2 passed** (two `PostgresInboundDedupeStore` / two `ChannelManager` instances share one table; first delivery proceeds, redelivery to the "other pod" is deduped)
  - Upstream/downstream HEAD-bump tests updated to `0005` (`test_persistence_bootstrap*.py`, `test_migration_0004_run_ownership_dedupe.py`, `test_channels.py`, `test_github_dispatcher.py`): **pass**
- Cross-pod evidence command:
  ```
  DEDUPE_TEST_POSTGRES_URL='postgresql+asyncpg://postgres:postgres@host.docker.internal:5432/deer' \
    uv run pytest tests/test_multi_pod_inbound_dedupe.py -v
  # => 2 passed
  ```


ech0hol:
Thanks for the review. Two things came up while rebasing onto latest `main`, and here's what I did about each — with the actual code and test runs so you can see it's not hand-waving.

**The migration conflict (the real one).** This branch added `0005_webhook_dedupe` right after `0004_run_ownership`, but `main` had already merged `0005_run_stop_reason` as a sibling of the same parent. Two children, one parent means alembic sees two heads and every migration test dies with "Multiple head revisions are present." The fix was to renumber ours to `0006` and hang it off the new head:

```python
# .../migrations/versions/0006_webhook_dedupe.py
revision = "0006_webhook_dedupe"
down_revision = "0005_run_stop_reason"
```

So the chain is now `0004_run_ownership → 0005_run_stop_reason → 0006_webhook_dedupe`. Four test files still had unresolved `<<<<<<<` markers from the rebase (one side said `0005_run_stop_reason`, the other `0005_webhook_dedupe`); I resolved all of them to `0006`. Minor trap: the markers were in column 0, so the resolved `assert` needed 8 spaces re-indented or the `try` block won't parse. Tests:

```
pytest tests/test_persistence_bootstrap.py tests/test_persistence_bootstrap_concurrency.py \
       tests/test_migration_0004_run_ownership_dedupe.py tests/test_migration_0006_webhook_dedupe.py
→ 36 passed
```

**Copilot's comments.** Honestly, the 4 comments just quoted the code back (e.g. "here is `make_inbound_dedupe_store`…") with no blocking concern, so there was nothing broken to fix. But I tightened three spots so the next reader doesn't have to guess the intent:

- *`dedupe_storage_config.py` wording.* The doc said the store "uses Postgres when multi-worker/Postgres is configured" — but the selector only checks `database.backend`, not worker count. Clarified to "whenever `database.backend='postgres'`". Docs-only; the runtime already did this.
- *`manager.py` None-check.* Changed `inbound_dedupe_store or MemoryInboundDedupeStore()` to `inbound_dedupe_store if inbound_dedupe_store is not None else MemoryInboundDedupeStore()`. Same behavior (a store is never falsy), just unambiguous about "inject a store, else default".
- *`dedupe_store.py` cleanup scope.* The TTL sweep (`DELETE … WHERE first_seen < now() - make_interval(…)`) used to run on every call, even on duplicate redeliveries that never inserted a row. Now it's inside `if inserted:`, so the sweep only fires when we actually wrote a new row. A redelivery still returns "duplicate, drop" exactly as before — no change to dedup correctness.

```python
inserted = result.fetchone() is not None
if inserted:
    await session.execute(
        text("DELETE FROM webhook_deliveries WHERE first_seen < now() - make_interval(secs => :ttl)"),
        {"ttl": INBOUND_DEDUPE_TTL_SECONDS},
    )
```

Tests for those three: `pytest tests/test_inbound_dedupe.py tests/test_migration_0006_webhook_dedupe.py → 16 passed`. The lazy-cleanup assertion still holds because the fake session returns a truthy `fetchone` for a new key, so `inserted=True` and the `DELETE` still runs. Cross-pod integration (`test_multi_pod_inbound_dedupe.py`) is 2 passed against a real Postgres (`DEDUPE_TEST_POSTGRES_URL` set), 2 skipped otherwise.

Net: one commit on top of current `main`, `ruff format --check` + `ruff check` clean (verified in the Linux container — ruff collapses a multi-line ternary differently than the Windows IDE, so I confirmed it there), migration chain linear, all bootstrap/migration tests green. Happy to revert any of the three cosmetic tweaks if you'd rather keep the original phrasing.

fancyboi999 (changes requested):
Thanks @ech0hol. I found two issues that should be addressed before this is ready.

[P1] Reclaim an expired key before treating the delivery as a duplicate

- Location: `backend/app/channels/dedupe_store.py:113-132`
- Problem: A delivery whose row is older than the 10-minute TTL can still be dropped indefinitely, so an intentional late GitHub "Redeliver" may never start a new run.
- Evidence: The insert checks the primary-key conflict before any expiry cleanup. On the exact same key, `ON CONFLICT DO NOTHING` returns no row, `inserted` is false, and the cleanup is skipped because it only runs inside `if inserted`. In a quiet deployment, nothing removes that expired row, so every later retry follows the same path. This violates #4120's explicit 10-minute ceiling.
- Suggested fix: Make expiry reclamation part of the atomic acquire operation, for example with a conditional `ON CONFLICT ... DO UPDATE` that refreshes and returns the row only when the stored `first_seen` is older than the TTL. Keep the broader cleanup as amortized housekeeping.
- Test: Against real Postgres, seed the same key with `first_seen` older than the TTL and assert the next `try_record` returns `False` and refreshes the timestamp; also assert a live row still returns `True` under concurrent callers.

[P2] Bump the repository config version for the new schema field

- Location: `config.example.yaml:13-18`
- Problem: This PR adds `AppConfig.dedupe_storage` but leaves `config_version` at 26, so existing version-26 installations are reported current and `make config-upgrade` exits without considering the new schema.
- Evidence: `config.example.yaml` says to bump the number when the schema changes, and `AppConfig._check_config_version` plus `scripts/config-upgrade.sh` gate their warning/upgrade path solely on that comparison. The repository check also requires the Helm embedded config and its README example to match the example version.
- Suggested fix: Increment `config.example.yaml`'s version and update `deploy/helm/deer-flow/values.yaml` plus `deploy/helm/deer-flow/README.md` to the same value. Keep the new default documented consistently in those generated/operator-facing surfaces.
- Test: Run `scripts/check_config_version.sh` and the config-version/upgrade tests, and assert a version-26 config is detected as outdated after this change.

Reviewed head SHA: `14a9e0f51d91f07fb1dcee35cffc5a9c8b520b7e`

zhfeng (commented):
## Review: share inbound webhook dedupe across pods via Postgres

Independent review - I verified both issues fancyboi999 raised against the actual code/repo state and traced the dispatch/release flow.

**Verdict:** Solid design and test coverage, but **one blocking correctness bug** in the Postgres store (the TTL ceiling is violated), plus a config-version convention miss.

### What's done well
- **Fail-open semantics are correct and well-tested.** `try_record` returns `False` (proceed) on any DB exception or missing session factory, so a storage outage never drops a webhook or 5xxs the provider. The two `fail_open` tests pin this.
- **Atomic dedup via `INSERT ... ON CONFLICT DO NOTHING RETURNING`** is the right primitive, and using `RETURNING`/`fetchone()` rather than `rowcount` for driver-independence is correct.
- **`auto` resolving to Postgres whenever `database.backend=postgres`** (independent of `GATEWAY_WORKERS`) is the right call - the common K8s case is N pods × 1 worker sharing one DB, and gating on in-pod worker count would have silently disabled cross-pod dedupe there. Good that the earlier `multi_worker` gating was removed.
- **Migration numbering is consistent.** Main's head is `0005_run_stop_reason`; this adds `0006_webhook_dedupe` with `down_revision="0005_run_stop_reason"`, and all bootstrap/regression tests reference `0006`. (PR body and Copilot review still say `0005` - stale prose only.) `create_all` parity + the `has_table` idempotency guard are correctly handled.
- **Release semantics preserved.** `_release_inbound_dedupe_key` is called only in the `except Exception` branch of `_handle_message`, so a redelivery can recover; on success the key stays until TTL. Matches pre-PR behavior.

### 🔴 P1 (blocking): Postgres store violates the 10-minute TTL ceiling

`backend/app/channels/dedupe_store.py`, `PostgresInboundDedupeStore.try_record` (~lines 113–132).

The lazy cleanup is gated behind a fresh insert, and the conflict check runs *before* any expiry reclamation:

```python
result = await session.execute(text("INSERT ... ON CONFLICT (...) DO NOTHING RETURNING channel"), ...)
inserted = result.fetchone() is not None
if inserted:                     # <-- cleanup only on a fresh insert
    await session.execute(text("DELETE FROM webhook_deliveries WHERE first_seen < now() - make_interval(secs => :ttl)"), ...)
return not inserted
```

**Failure scenario:**
1. `T0` - delivery with key `K` arrives → INSERT succeeds (`first_seen=T0`), cleanup runs (nothing to delete), returns `False` (proceed). Agent runs.
2. `T0 + 15min` - provider redelivers the same `K` (e.g. a manual GitHub **Redeliver**, which the PR body explicitly names as a target). The row still exists but is *expired* (`first_seen` > 10 min old). `ON CONFLICT DO NOTHING` fires, returns no row → `inserted=False` → cleanup is **skipped** → returns `True` → **dropped as a duplicate**.
3. The expired row is never reclaimed by this redelivery - only by some *other* new key triggering a cleanup. In a quiet deployment `K` persists indefinitely, so **every** later redelivery of `K` is dropped forever.

This diverges from `MemoryInboundDedupeStore`, which evicts expired entries from the front before the membership check and therefore correctly lets a post-TTL redelivery proceed. The Postgres path (the default for `auto` + Postgres, i.e. the exact multi-replica case this PR targets) is stricter-than-TTL - a regression versus the in-process behavior it's meant to generalize.

**Fix** (agreeing with fancyboi999): make expiry reclamation part of the atomic acquire, e.g.:

```sql
INSERT INTO webhook_deliveries (channel, workspace_id, chat_id, message_id, first_seen)
VALUES (:c, :w, :ch, :m, now())
ON CONFLICT (channel, workspace_id, chat_id, message_id)
DO UPDATE SET first_seen = now()
WHERE webhook_deliveries.first_seen < now() - make_interval(secs => :ttl)
RETURNING channel
```

- No conflict → row returned → proceed.
- Conflict + **expired** row → `DO UPDATE` refreshes `first_seen`, row returned → proceed (treated as new). ✅ honors TTL ceiling.
- Conflict + **live** row → `WHERE` fails, no update, no row → `inserted=False` → drop. ✅ still dedupes within TTL.

Atomicity across pods is preserved: two pods racing on the same expired key - one wins the insert/update and proceeds, the other sees a now-fresh row and is dropped. Keep the broad `DELETE` as amortized housekeeping for keys never re-accessed.

**Test to add** (against real Postgres, like `test_multi_pod_inbound_dedupe.py`): seed a row with `first_seen` older than TTL, assert the next `try_record` for the same key returns `False` and refreshes `first_seen`; separately assert a live row still returns `True` under concurrent callers.

### 🟡 P2 (should fix): `config_version` not bumped for the new `dedupe_storage` field

`config.example.yaml:18` still says `config_version: 26` (unchanged from `main`), and `deploy/helm/deer-flow/values.yaml:243` is also still `26`, despite the new `AppConfig.dedupe_storage` schema field. The example file's own header says *"Bump this number when the config schema changes."*

Nuance on severity: this does **not** break CI. `scripts/check_config_version.sh` only enforces `chart >= example` (both 26 → passes), and `test_config_version.py` only tests the user-vs-example outdated-warning logic - neither enforces that a schema change bumps the version, and neither checks the README. So this is a **convention/upgrade-UX issue**, not a build break: existing v26 installs won't get an "outdated config" warning and `make config-upgrade` won't surface the new `dedupe_storage` field. Per the repo convention, bump to `27` in `config.example.yaml`, `deploy/helm/deer-flow/values.yaml`, and the chart `README.md` example.

### Minor notes
- **Oversized keys → noisy fail-open.** `message_id` is `VARCHAR(1024)`, `workspace_id`/`chat_id` `VARCHAR(512)`. A provider id exceeding the length makes the INSERT raise → fail-open (proceeds, but dedupe is silently off for that key) and logs an exception on *every* such delivery. Low risk in practice; if you want belt-and-suspenders consider `TEXT` or truncation. The NUL-in-TEXT comment in the migration/model is slightly misleading - Postgres `text`/`varchar` can't hold NUL regardless of column type - but harmless since provider ids won't contain NUL.
- **Dispatch-loop latency.** `_is_duplicate_inbound` is now `await`ed in the serialized `_dispatch_loop`, so every inbound message does a synchronous Postgres round-trip (INSERT + DELETE) before dispatch. This is correct (dedupe must precede dispatch to be race-free) and the op is fast, but worth being aware of for very high-volume channels. The memory path is unaffected.
- **Stale test comment.** `test_factory_auto_with_postgres_and_multi_worker_resolves_postgres_store` says "a single worker is already deduped by the memory store" - slightly misleading given multi-pod × 1-worker isn't fully deduped by memory. The sibling `..._regardless_of_workers` test documents the real behavior; consider aligning the comment.

### Summary
Architecture and the fail-open/atomic-insert design are good and well-tested. The one change I'd hold the PR for is **P1**: the Postgres store doesn't reclaim expired keys on conflict, so it drops post-TTL redeliveries (notably manual GitHub Redeliver) - potentially indefinitely in quiet deployments - breaking the TTL ceiling the memory store honors. P2 is a quick version bump.

---

_Review posted by Zheng Feng via Claude Code. Reviewed head SHA: `14a9e0f5`._

ech0hol:
@zhfeng  @fancyboi999,Thanks both for taking the time to review this — and for verifying the two issues independently. I'm glad your conclusions landed on the same spot (P1 + P2), which makes the fix straightforward. Both are now addressed on the branch, and I've added test coverage + a real-Postgres integration run to back it up. Details below.

## P1 — Postgres store was violating the 10-minute TTL ceiling 🔴

You're right that the original `try_record` (the version you reviewed at `14a9e0f5`) gated the TTL sweep behind `if inserted:`, so a redelivery of an *expired-but-unreleased* row hit `ON CONFLICT DO NOTHING`, returned no row, skipped the sweep, and was dropped forever. That's exactly the regression versus the in-memory store you both flagged, and a manual "Redeliver" of an old webhook would never get through in a quiet deployment.

My first attempt fixed it with a two-step approach: delete this key's expired row, then `INSERT … ON CONFLICT DO NOTHING`. @zhfeng correctly caught that this still has a TOCTOU race — between the `DELETE` and the `INSERT`, another pod can insert the same expired key, and both end up treating it as new (double-dispatch). So I dropped that approach in favour of a single atomic conditional upsert:

```147:153:backend/app/channels/dedupe_store.py
                            "INSERT INTO webhook_deliveries "
                            "(channel, workspace_id, chat_id, message_id, first_seen) "
                            "VALUES (:c, :w, :ch, :m, now()) "
                            "ON CONFLICT (channel, workspace_id, chat_id, message_id) "
                            "DO UPDATE SET first_seen = now() "
                            "WHERE webhook_deliveries.first_seen < now() - make_interval(secs => :ttl) "
                            "RETURNING channel"
```

This collapses acquire + expiry reclamation into one row-locked statement, so there's no window for a race:

| Case | Behaviour | Result |
|------|-----------|--------|
| No conflict (new key) | row inserted, `RETURNING channel` yields a row | admit (proceed) |
| Conflict + row **expired** | `DO UPDATE` refreshes `first_seen`, row is returned | re-admit (proceed) — honours the TTL ceiling |
| Conflict + row **still live** | `WHERE` fails, no update, no row returned | drop as duplicate (still dedupes within TTL) |

Because it's a single `ON CONFLICT DO UPDATE`, two pods racing on the same expired key can't both proceed: one wins the update and proceeds, the other sees the now-fresh row and is dropped. The broad `DELETE` sweep stays, but only on the admit path as amortized housekeeping for keys that are never re-touched:

```173:177:backend/app/channels/dedupe_store.py
                    if inserted:
                        await session.execute(
                            text("DELETE FROM webhook_deliveries WHERE first_seen < now() - make_interval(secs => :ttl)"),
                            {"ttl": INBOUND_DEDUPE_TTL_SECONDS},
                        )
```

`return not inserted` is unchanged — `True` = duplicate/drop, `False` = proceed — and the fail-open path is untouched (any DB error → `return False`, never silently drops a webhook).

---

## P2 — `config_version` not bumped for the new `dedupe_storage` field 🟡

Agreed. I bumped it to `27` in all three places so existing v26 installs get the "outdated config" warning and `make config-upgrade` surfaces the new field:

| File | Line | Value |
|------|------|-------|
| `config.example.yaml` | 18 | `config_version: 27` |
| `deploy/helm/deer-flow/values.yaml` | 236 | `config_version: 27` |
| `deploy/helm/deer-flow/README.md` | 127 | `config_version: 27` |

`scripts/check_config_version.sh` passes (`27 >= 27`).

---

## Test report

All checks run inside the Linux dev container (Python 3.12), not the host IDE:

```bash
# format + lint
cd /app/backend && uv run ruff format app/channels/dedupe_store.py tests/test_inbound_dedupe.py \
  && uv run ruff check app/channels/dedupe_store.py tests/test_inbound_dedupe.py

# unit (mocked session)
uv run pytest tests/test_inbound_dedupe.py -q

# real Postgres cross-pod integration
DEDUPE_TEST_POSTGRES_URL=postgresql+asyncpg://test:test@localhost:5432/dedupe \
  uv run pytest tests/test_multi_pod_inbound_dedupe.py -q

# config version consistency
bash ../scripts/check_config_version.sh
```

| Check | Result |
|-------|--------|
| `ruff format` | 2 files left unchanged |
| `ruff check` | All checks passed |
| `tests/test_inbound_dedupe.py` | **17 passed** |
| `tests/test_multi_pod_inbound_dedupe.py` (real Postgres) | **3 passed** |
| `scripts/check_config_version.sh` | passes (`27 >= 27`) |

The key cases:

- `test_postgres_try_record_reclaims_expired_unreleased_row` (mock) — an expired-row conflict takes the `DO UPDATE … WHERE first_seen < TTL` branch, `fetchone()` returns a row → admitted; asserts the SQL carries `DO UPDATE SET first_seen = now()` and the TTL condition.
- `test_postgres_try_record_alive_row_still_deduped_without_cleanup` (mock) — a live-row conflict returns no row → dropped, and the sweep is skipped (single execute).
- `test_expired_unreleased_row_is_reclaimed_on_next_redelivery` (real Postgres) — seeds `first_seen = now() - (TTL + 1s)` for a key, then asserts the next `try_record` returns `False` and re-checks that `(now() - first_seen) < ttl` (i.e. `first_seen` was refreshed). This exercises the P1 TTL-ceiling behaviour end-to-end.

---

The fix is committed on the branch (latest commit `6e82520b`). Would you mind re-reviewing when you get a chance?

WillemJiang:
@ech0hol  please resolve the conflicts with the main branch.

ech0hol:
@WillemJiang Thanks for the review — I've resolved the merge conflicts with `main` and fixed the CI gaps (the cross-pod dedupe integration tests now actually run in CI via the Postgres service mapping), so all checks should be green now.

WillemJiang (approved):

Issue #4043, opened 2026-07-10
Subagent internal messages leak into the parent run's `messages` SSE stream (raw tool JSON in chat + subtask cards flash failed)


CodexterLi:
## Summary

While a `task` subagent is running, the subagent's **internal** LLM tokens and tool results leak into the **parent** run's `messages` SSE stream. This produces two user-visible bugs in the workspace UI:

1. **Raw tool-result JSON appears in the parent chat** (e.g. a full-screen `web_search` JSON blob rendered as markdown) while a subtask is executing. The blobs disappear on page refresh because they never enter the parent thread's persisted state — they only exist in the live stream.
2. **Subtask cards flash `failed` → `completed`**: the leaked messages arrive *after* the `assistant:subagent` group, so that group is no longer the last one, `groupIsLoading` goes `false`, and `derivePendingSubtaskStatus` falls through to `"failed"` for every pending card. Minutes later the `task` ToolMessage arrives with `subagent_status: "completed"` and the cards flip green.

Verified at commit `c0b917cc` (2026-07-10).

## Root cause

`SubagentExecutor._aexecute` (`backend/packages/harness/deerflow/subagents/executor.py`) calls `agent.astream(...)` with a **fresh** `run_config`, but LangChain propagates the parent's callback context through contextvars — so the parent run's `StreamMessagesHandler` (langgraph `pregel/_messages.py`) is still attached.

That handler's namespace check does not exclude a nested-but-independent graph: the subagent is its own root graph, so its `langgraph_checkpoint_ns` splits to `len(ns) == 0` and every event passes the `if not self.subgraphs and len(ns) > 0: return` guard as if it were a parent root-level event. Both emission paths leak:

- **chat-model path** (`on_chat_model_start` / `on_llm_new_token`) → subagent LLM tokens (metadata carries `tags` incl. `subagent:<name>`);
- **chain path** (`on_chain_start` / `on_chain_end`) → subagent ToolMessages — this is the raw JSON in the chat. Note the chain-path metadata does **not** carry `tags`, so consumers cannot filter these events downstream.

The worker (`runtime/runs/worker.py`, `_stream_once`) then publishes every chunk to the bridge unfiltered.

Your own frontend comment already describes the symptom — `core/messages/utils.ts`, orphan tool-message fallback: *"LangGraph `messages-tuple` can emit tool-result events out of order or replay them from subagent state…"* — but the fallback attaches the leaked message to the most recent group, which renders it as assistant markdown.

[...]

PR #4215, opened 2026-07-15
fix(subagents): preserve parent checkpoint namespace

ggnnggez:
Refs #4043

## Why

LangGraph expects nested graphs to inherit checkpoint coordinates from the parent runtime. Its official examples pass `thread_id` to the parent and invoke child graphs without rebuilding the child config:

- [Subgraphs](https://docs.langchain.com/oss/python/langgraph/use-subgraphs)
- [Subgraph outputs](https://docs.langchain.com/oss/python/langgraph/streaming#subgraph-outputs)
- [Upstream config contract tests](https://github.com/langchain-ai/langgraph/blob/1.2.9/libs/langgraph/tests/test_utils.py#L509-L595)

`SubagentExecutor` instead passed the parent `thread_id` explicitly to the child. LangGraph 1.2.6+ treats that as a new root checkpoint lineage, drops the inherited `checkpoint_ns`, and can route child AI/tool messages into the parent stream.

## What changed

- Let subagent graphs inherit checkpoint coordinates from the parent runtime.
- Continue passing the business `thread_id` through `runtime.context`.
- Keep the invocation-contract test active on all supported versions; run the production-shaped parent-stream regression on LangGraph 1.2.6+.
- Document the checkpoint-lineage requirement.

## Surface area

- [x] **Agents / LangGraph**

## Screenshots / Recording

Not applicable.

## Bug fix verification

- Test: `backend/tests/test_subagent_executor.py::TestSubagentCheckpointLineage`
- Red/green: yes. Restoring the old child config on LangGraph 1.2.9 leaks all child sentinels into the parent stream; this branch excludes them.

## Validation

- `cd backend && make lint`
- Locked LangGraph 1.1.9 executor, tracing, context, and checkpointer tests: 11 passed, 1 skipped
- LangGraph 1.2.9 regression tests: 2 passed
- The full executor test file still encounters an unrelated existing fixture error: the mocked skill storage lacks `get_or_new_user_skill_storage`.

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Investigated the upstream contract, implemented the minimal fix, and added TDD regression coverage.


backend/tests/test_subagent_executor.py
```diff
@@ -2540,6 +2541,182 @@ async def astream(self, state, *, config, context, stream_mode):  # noqa: ARG002
...
+
+        parent_builder = StateGraph(MessagesState)
+        parent_builder.add_node("delegate", delegate)
+        parent_builder.add_edge(START, "delegate")
+        parent_builder.add_edge("delegate", END)
+        parent_graph = parent_builder.compile()
+
+        streamed_messages = [
+            message
+            async for message, _metadata in parent_graph.astream(
```
willem-bd, line 2721: ⚠️ Heads-up: this "stream isolation" test is inert at the repo's currently locked `langgraph` version, so it gives no regression protection today.

`uv.lock` resolves `langgraph` to **1.1.9** (pin is `>=1.1.9`), but the leak this guards against is **1.2.6+-specific** (validated on 1.2.9 per the PR description). I restored the old `run_config["configurable"] = {"thread_id": ...}` line and re-ran `TestSubagentCheckpointLineage` on 1.1.9: this test still **passes** with the buggy code. Only `test_aexecute_leaves_checkpoint_coordinates_to_parent_context` actually fails red/green at the current pin.

So a future maintainer who reverts the fix on 1.1.9 will see all-green here. To fix the false confidence, either:
- add a `skipif(langgraph < 1.2.6, reason="leak only manifests on >=1.2.6")` so the test can't silently pass-both-ways; or
- bump the pin to ≥1.2.6 in this PR so the guard is live and the fix is actually exercised in CI.

backend/tests/test_subagent_executor.py
```diff
@@ -2540,6 +2541,182 @@ async def astream(self, state, *, config, context, stream_mode):  # noqa: ARG002
...
+                        id="parent-final-sentinel",
+                    )
+                ]
+            }
+
+        parent_builder = StateGraph(MessagesState)
+        parent_builder.add_node("delegate", delegate)
+        parent_builder.add_edge(START, "delegate")
+        parent_builder.add_edge("delegate", END)
+        parent_graph = parent_builder.compile()
```
willem-bd: The parent graph is compiled with **no checkpointer**, but the production lead-agent graph always runs with one (Postgres/SQLite). Since the bug is fundamentally about a dropped `checkpoint_ns` (a checkpointer concept), a checkpointer-less parent may not traverse the leaking code path at all — which is likely why the test passes both ways on 1.1.9 (see the other comment).

To make this a faithful reproduction once you're on ≥1.2.6, consider `parent_builder.compile(checkpointer=MemorySaver())`. Separately, calling the sync `executor.execute(...)` from inside an async node (line 2682) blocks the event loop for the child run, and the real dispatch path is async via the task tool — so this also diverges from production.

backend/packages/harness/deerflow/subagents/executor.py
```diff
@@ -728,7 +728,11 @@ async def _aexecute(self, task: str, result_holder: SubagentResult | None = None
             collector_caller = f"subagent:{self.config.name}"
             collector = SubagentTokenCollector(caller=collector_caller)
 
-            # Build config with thread_id for sandbox access and recursion limit
+            # Do not put checkpoint coordinates (thread_id/checkpoint_ns/etc.)
+            # in the child config. LangGraph inherits those coordinates from
+            # the ambient parent run so this execution keeps its subgraph
+            # namespace. Business consumers receive thread_id via ``context``
```
willem-bd, line 734: The change itself is correct. I traced every `thread_id` consumer and they all still resolve it — most read `runtime.context` first (kept at line 774), with `sandbox/tools.py`, `sandbox/middleware.py`, `memory_middleware`, `task_tool`, `skill_manage_tool`, and `present_file_tool` all falling back to `config["configurable"]` only *after* `context`.

One exception worth confirming: `invoke_acp_agent_tool.py:173` reads `thread_id` from `config["configurable"]` **only** (no `context` fallback). After this change it relies entirely on LangGraph inheriting the parent's `configurable` — which is exactly the inheritance this fix preserves, so it's fine in the normal parent→child dispatch path. Just flagging it as the single consumer whose resolution mechanism shifted; worth a mental note if `_aexecute` is ever invoked outside an ambient parent config.

(Also confirmed `inject_langfuse_metadata` only writes `config["metadata"]`, never `configurable`, so it can't reintroduce a checkpoint coordinate and defeat the fix.)

willem-bd (commented):
Reviewed the checkpoint-namespace fix. The core change is correct and minimal — I traced every `thread_id` consumer and they all still resolve it, and confirmed `inject_langfuse_metadata` can't reintroduce a checkpoint coordinate. Detailed findings are inline. The main one is test fidelity: the "stream isolation" regression test is inert at the repo's currently locked `langgraph` (1.1.9) — it passes both with and without the fix — and uses a checkpointer-less parent that doesn't match production. Recommend a version gate or a `langgraph` bump so the guard is actually live.

backend/tests/test_subagent_executor.py
ggnnggez: Addressed in 3aedf5a9. The parent test graph now uses `MemorySaver`, and the async parent node dispatches through the public `execute_async()` path and polls without blocking the event loop. The production-shaped test passes on LangGraph 1.2.9 and fails when the old child `configurable.thread_id` override is restored.

WillemJiang (approved):

PR #4216, opened 2026-07-15
fix(memory): follow config reload in get_memory_config

### LINKED ISSUE #4204: Memory config singleton desync: AppConfig auto-reloads but _memory_config does not follow

he-yufeng:
## Summary

`get_memory_config()` returned the `_memory_config` singleton directly, but that singleton is only refreshed as a *side effect* of `get_app_config()` reloading (via `_apply_singleton_configs` → `load_memory_config_from_dict`). Any reader that reaches memory config without first going through `get_app_config()` — most importantly the agent factory deciding whether to bind the memory tools (`feat.memory_config or get_memory_config()` in `agents/factory.py`) — sees a stale `memory.mode` after a `config.yaml` edit, even though `memory.*` is documented as hot-reloadable in `backend/AGENTS.md`. The agent keeps the old mode (memory tools bound / not bound) until some unrelated `get_app_config()` call happens to refresh the singleton.

## Fix

Trigger the same signature-checked reload inside `get_memory_config()` so the singleton follows the config file, mirroring how the other config getters already resolve through `get_app_config()` (same shape as the skills-prompt fix in #4160). When no config file is on disk — tests, or a minimal deployment running purely on defaults / `set_memory_config` — `get_app_config()` raises `FileNotFoundError`; that is swallowed and we fall back to the in-memory singleton, so the accessor is never more fragile than before.

## Test

`test_get_memory_config_follows_config_file_reload` writes a config with `memory.mode: middleware`, reads it back through `get_memory_config()`, edits the file to `mode: tool` (bumping mtime), then asserts `get_memory_config().mode == "tool"` without an intervening `get_app_config()` call. It fails on `main` (`AssertionError: 'middleware' == 'tool'`) and passes with this change. `test_app_config_reload.py` (15) and `test_memory_manager_pluggable.py` (16 — covers the no-config-file fallback and `set_memory_config` injection) stay green; `ruff format --check` / `ruff check` are clean.

## Relation to #4208

#4208 targets the same issue with the same `get_app_config()` lever. This PR additionally guards the no-config-file path so `get_memory_config()` stays a total function: the backend suite and CI run without a `config.yaml` (`make test` is `pytest tests/` with none generated), so an unconditional `get_app_config()` makes `get_memory_config()` raise `FileNotFoundError`, which is what breaks `test_memory_manager_pluggable.py` (autouse fixture) and `test_create_deerflow_agent.py` (factory reaches memory config via `feat.memory_config or get_memory_config()`). The `try/except FileNotFoundError` fallback here keeps those green, and the change ships with the regression test above per the backend TDD guidance in `AGENTS.md`. Happy to defer or fold these two pieces into #4208 if the maintainers prefer a single PR.

Fixes #4204

### DIFF
--- a/backend/packages/harness/deerflow/config/memory_config.py
+++ b/backend/packages/harness/deerflow/config/memory_config.py
@@ -119,7 +119,28 @@ _memory_config: MemoryConfig = MemoryConfig()
 
 
 def get_memory_config() -> MemoryConfig:
-    """Get the current memory configuration."""
+    """Get the current memory configuration.
+
+    ``_memory_config`` is only refreshed as a side effect of ``get_app_config()``
+    reloading (via ``_apply_singleton_configs`` -> ``load_memory_config_from_dict``).
+    A reader that reaches memory config without going through ``get_app_config()``
+    first -- e.g. the agent factory deciding whether to bind the memory tools --
+    would otherwise see a stale ``memory.mode`` after a ``config.yaml`` edit, even
+    though ``memory.*`` is documented as hot-reloadable. Trigger the same
+    signature-checked reload here so the singleton follows the config file.
+
+    If no config file is on disk (tests, or a minimal deployment running purely
+    on defaults / ``set_memory_config``), there is nothing to reload from, so we
+    keep the pre-existing behavior of returning the in-memory singleton instead of
+    becoming more fragile than before.
+    """
+    # Lazy import: app_config imports this module, so a top-level import cycles.
+    from .app_config import get_app_config
+
+    try:
+        get_app_config()
+    except FileNotFoundError:
+        pass
     return _memory_config
 
 
--- a/backend/packages/harness/deerflow/config/memory_config.py
+++ b/backend/packages/harness/deerflow/config/memory_config.py
@@ -132,7 +132,10 @@ def get_memory_config() -> MemoryConfig:
     If no config file is on disk (tests, or a minimal deployment running purely
     on defaults / ``set_memory_config``), there is nothing to reload from, so we
     keep the pre-existing behavior of returning the in-memory singleton instead of
-    becoming more fragile than before.
+    becoming more fragile than before. The catch is deliberately narrow: a config
+    file that exists but fails to load (malformed YAML, failed validation) is a
+    real error and is surfaced here just as it is at every other unguarded
+    ``get_app_config()`` call site, leaving the last-good singleton untouched.
     """
     # Lazy import: app_config imports this module, so a top-level import cycles.
     from .app_config import get_app_config
--- a/backend/tests/test_app_config_reload.py
+++ b/backend/tests/test_app_config_reload.py
@@ -530,3 +530,34 @@ def test_get_app_config_does_not_mutate_singletons_when_reload_validation_fails(
         assert get_store() is initial_store
     finally:
         _reset_config_singletons()
+
+
+def test_get_memory_config_follows_config_file_reload(tmp_path, monkeypatch):
+    """get_memory_config() picks up a config.yaml memory.mode edit on its own (#4204).
+
+    ``_memory_config`` used to refresh only as a side effect of ``get_app_config()``,
+    so a reader reaching memory config directly (e.g. the agent factory deciding
+    whether to bind the memory tools) saw a stale mode after a config edit even
+    though ``memory.*`` is documented as hot-reloadable.
+    """
+    config_path = tmp_path / "config.yaml"
+    extensions_path = tmp_path / "extensions_config.json"
+    _write_extensions_config(extensions_path)
+    _write_config_with_sections(config_path, {"memory": {"enabled": True, "mode": "middleware"}})
+
+    monkeypatch.setenv("DEER_FLOW_CONFIG_PATH", str(config_path))
+    monkeypatch.setenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH", str(extensions_path))
+    _reset_config_singletons()
+
+    try:
+        assert get_memory_config().mode == "middleware"
+
+        _write_config_with_sections(config_path, {"memory": {"enabled": True, "mode": "tool"}})
+        next_mtime = config_path.stat().st_mtime + 5
+        os.utime(config_path, (next_mtime, next_mtime))
+
+        # No explicit get_app_config() call: get_memory_config() must reflect the
+        # new mode on its own.
+        assert get_memory_config().mode == "tool"
+    finally:
+        _reset_config_singletons()
--- a/backend/tests/test_app_config_reload.py
+++ b/backend/tests/test_app_config_reload.py
@@ -561,3 +561,44 @@ def test_get_memory_config_follows_config_file_reload(tmp_path, monkeypatch):
         assert get_memory_config().mode == "tool"
     finally:
         _reset_config_singletons()
+
+
+def test_get_memory_config_surfaces_malformed_config_edit(tmp_path, monkeypatch):
+    """A malformed config edit surfaces through get_memory_config() rather than
+    being swallowed.
+
+    The ``except FileNotFoundError`` guard is deliberately narrow: it covers only
+    the genuinely-absent file (tests, or a defaults-only deployment) where there
+    is nothing to reload from. A file that exists but fails validation is a real
+    error and propagates -- consistent with every other unguarded
+    ``get_app_config()`` caller -- and the last-good singleton is left intact
+    rather than cleared or half-applied.
+    """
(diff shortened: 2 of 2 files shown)


backend/packages/harness/deerflow/config/memory_config.py
```diff
@@ -119,7 +119,28 @@ def should_use_memory_tools(config: MemoryConfig) -> bool:
...
+    on defaults / ``set_memory_config``), there is nothing to reload from, so we
+    keep the pre-existing behavior of returning the in-memory singleton instead of
+    becoming more fragile than before.
+    """
+    # Lazy import: app_config imports this module, so a top-level import cycles.
+    from .app_config import get_app_config
+
+    try:
+        get_app_config()
+    except FileNotFoundError:
```
willem-bd, line 145: The `except FileNotFoundError` is deliberately narrow, but `get_app_config()` can raise other exceptions during a reload - `ValidationError` (pydantic), `yaml.YAMLError`, or errors from `ExtensionsConfig.from_file()` / `_check_config_version`. Before this change `get_memory_config()` was a pure accessor that never raised; now, if `config.yaml` is edited to be invalid mid-run, those errors propagate out - including to the factory's `feat.memory_config or get_memory_config()` (`agents/factory.py:249`).

The docstring's "never more fragile than before" (line 135) holds for the no-config-file case but is overstated for the malformed-edit case. This is arguably acceptable (fail-fast is consistent with `get_app_config()` raising elsewhere), but it's a real contract change worth pinning down - either:
- widen the `except` to also swallow reload-validation errors (the existing `test_get_app_config_does_not_mutate_singletons_when_reload_validation_fails` confirms singletons keep their last-good value, but the exception still propagates today), or
- keep the narrow `except` and tighten the docstring wording so the malformed-edit behavior isn't implied to be a no-op.

backend/tests/test_app_config_reload.py
```diff
@@ -530,3 +530,34 @@ def test_get_app_config_does_not_mutate_singletons_when_reload_validation_fails(
         assert get_store() is initial_store
     finally:
         _reset_config_singletons()
+
+
+def test_get_memory_config_follows_config_file_reload(tmp_path, monkeypatch):
```
willem-bd, line 535: Solid regression test for the staleness bug. One gap: nothing pins the malformed-edit behavior - what `get_memory_config()` does when a config edit fails validation mid-reload (raise vs. fall back to the singleton). Since that's the main contract change in this PR (see the comment on `memory_config.py`), a sibling test asserting the intended behavior would both document the decision and guard against regression - e.g. write a valid config, overwrite it with malformed YAML, and assert the chosen outcome.

willem-bd (commented):
Automated code review. The fix is correct and mirrors the established `checkpointer_config.ensure_config_loaded()` pattern (lazy import + `try/except FileNotFoundError`); I verified the reload chain (`get_app_config` -> `_apply_singleton_configs` -> `load_memory_config_from_dict`) refreshes `_memory_config`, and the regression test faithfully reproduces the staleness.

Two inline notes below: (1) the narrow `except FileNotFoundError` means `get_memory_config()` now propagates non-FNF reload errors (validation/YAML) - a behavior change vs. the previous pure-accessor contract, worth deciding/documenting; (2) a sibling test for the malformed-edit case.

Non-inline follow-up: several sibling getters - `get_guardrails_config`, `get_tool_search_config`, `get_subagents_app_config`, `get_summarization_config` - still just `return _*_config` without triggering a reload, so they have the same latent staleness bug this PR fixes for memory. Worth a follow-up (or folding into #4208) so the pattern is applied uniformly.

(Minor, not a blocker: `get_app_config()` recomputes a SHA256 digest of the whole config file on every call, so `get_memory_config()` went from O(1) to O(file size) per call. This is consistent with the other reload-aware getters and negligible in practice - and the unconditional `get_app_config()` call here is the right choice for hot-reload; do NOT copy `ensure_config_loaded()`'s short-circuit guard, which is a one-time bootstrap with no hot-reload.)

backend/packages/harness/deerflow/config/memory_config.py
he-yufeng: You're right to flag it, and the narrow catch is intentional. `FileNotFoundError` is the one "nothing to reload from" case (tests, or a defaults-only deploy running on `set_memory_config`). A file that exists but fails to load — malformed YAML, a `ValidationError`, a version/extensions error — is a real error, and I'd rather surface it here exactly as every other unguarded `get_app_config()` caller does than have `get_memory_config()` quietly serve stale config while the rest of the app is already erroring on the same edit. Pushed c844ac9f to pin that: a malformed edit raises through `get_memory_config()` and leaves the last-good singleton intact, plus a note on the guard documenting the contract.

backend/tests/test_app_config_reload.py
he-yufeng: Added in c844ac9f — `test_get_memory_config_surfaces_malformed_config_edit` writes an invalid config that also flips `memory.mode`, then asserts `get_memory_config()` raises `ValidationError` while the singleton stays at the last-good `"middleware"`. So a half-applied reload (leaking `"tool"` into the singleton) or a swallowed error would both fail the test.

he-yufeng:
Good catch on the sibling getters — `get_summarization_config` / `get_subagents_app_config` at least share the documented hot-reload contract (`summarization.*` / `subagents.*` are in the reload boundary), so they have the same latent staleness. I'd rather not widen this PR past the memory fix in #4204, but I'll do a focused follow-up that applies the same reload-follow pattern uniformly, after checking which of the four are actually hot-reloadable vs. startup-only so I don't add a reload to a getter that shouldn't have one.

WillemJiang (approved):

he-yufeng:
Follow-up on the sibling-getter staleness note from the review. I went to write the promised focused PR against current main and the gap has closed itself since:

- `_apply_singleton_configs` refreshes both singletons on every `get_app_config()` reload (app_config.py:457/460), so they track config.yaml edits as long as any code path touches `get_app_config()`.
- `get_summarization_config()` has no production callers left; middlewares read `app_config.summarization` off the per-run resolved config instead (tool_error_handling_middleware.py:427/463, summarization_middleware.py:761).
- `get_subagents_app_config()` remains only as the `app_config=None` fallback in the subagents registry, and every production caller I traced (apply_prompt_template at agent.py:854/936 and client.py:368, task_tool.py:265) passes a resolved config in. The bare-singleton path is only reachable from tests.

So there is nothing to fix without inventing a caller that doesn't exist. Leaving the getters as-is; if a future change adds a production read of the bare singleton, it should come with the reload trigger then.

PR #3553, opened 2026-06-13
fix(agents): sync agent_name across context/configurable and reject empty soul (#3549)

Huixin615:
Fixes #3549

## Why

Reported in #3549: creating a custom agent through the Agent creation chat appeared to succeed in the UI, but the resulting `SOUL.md` was empty and ended up in the wrong directory. Tracing the user's logs showed two independent root causes stacked on the same flow, so the agent was unusable and — worse — the global default `SOUL.md` could be silently overwritten with empty content on every failed creation.

## What changed

From a caller's perspective:

- Creating a custom agent now writes `SOUL.md` to `users/{user_id}/agents/{name}/` consistently (instead of occasionally landing in the global `.deer-flow/SOUL.md`).
- If the model invokes `setup_agent` with an empty or whitespace-only `soul`, the tool now returns an explicit error message instead of silently persisting an empty file and reporting success. The model can then retry; the global default `SOUL.md` is no longer at risk of being clobbered.
- No public API shape, default model, or user-facing setting changes.

Internals:

1. `build_run_config` (`backend/app/gateway/services.py`) now mirrors `agent_name` into **both** `config["configurable"]` and `config["context"]` when an `assistant_id` is injected, aligning with the dual-write pattern already used by `merge_run_context_overrides` and `runtime/runs/naming.py`. This is required because, since LangGraph ≥1.1.9, `ToolRuntime.context` no longer falls back to `configurable` — and `setup_agent` reads `runtime.context` exclusively, so a single-container write left it observing `agent_name=None`.
2. `setup_agent` (`backend/packages/harness/deerflow/tools/builtins/setup_agent_tool.py`) rejects empty / whitespace-only `soul` before any filesystem operation, so an empty payload can never overwrite an existing per-agent or global `SOUL.md`.

## Surface area

- [ ] **Frontend UI**
- [x] **Backend API** — `build_run_config` behavior in `backend/app/gateway/services.py`
- [x] **Agents / LangGraph** — `setup_agent` tool guard
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [ ] **Default behavior change** — bug fix only; restores intended behavior
- [ ] **Docs / tests / CI only**

## Screenshots / Recording

N/A — backend-only fix, no UI change.

## Bug fix verification

Failing tests that reproduce both root causes before the fix:

[...]

PR #4219, opened 2026-07-15
fix(agents): refuse empty SOUL.md updates in update_agent

AnnaSuSu:
## Why

`setup_agent` refuses empty / whitespace-only `soul` before touching the
filesystem (#3553 / #3549): writing a blank SOUL.md would report success for
an unusable agent and could clobber existing content.

`update_agent` is the sibling write path for the same file. It accepted the
same empty input, reported **success**, and replaced a working `SOUL.md` with
whitespace — so the next turn loads an empty personality.

This is the same guard on the update side, not a new product rule.

## What changed

`tools/builtins/update_agent_tool.py` — before any staging/rename work, if
`soul is not None and not soul.strip()`, return an error ToolMessage and leave
the existing SOUL.md / config.yaml untouched. Message mirrors setup_agent
("soul content is empty; refusing to update…").

## Surface area

- [x] **Agents / LangGraph** — `update_agent` built-in tool under `packages/harness/deerflow/tools`
- [x] **Docs / tests / CI only** — `tests/test_update_agent_tool.py`

## Bug fix verification

- Test paths:
  - `tests/test_update_agent_tool.py::test_update_agent_rejects_empty_soul_and_does_not_overwrite`
  - `tests/test_update_agent_tool.py::test_update_agent_rejects_whitespace_only_soul_and_does_not_overwrite`
- Red on `main` / green on this branch? **yes** — both fail on unfixed main with
  `"Agent 'test-agent' updated successfully. Changed: soul..."` and a wiped SOUL;
  both pass after the guard (error status, original soul preserved).
- Red check was in place: only `update_agent_tool.py` swapped to `origin/main`
  (confirmed no empty-soul guard string), tests held fixed, then restored.
- Non-empty soul updates and description-only updates still succeed (existing
  suite + E2E).

## Validation

```
cd backend
ruff check packages/harness/deerflow/tools/builtins/update_agent_tool.py tests/test_update_agent_tool.py
# All checks passed!
ruff format --check packages/harness/deerflow/tools/builtins/update_agent_tool.py tests/test_update_agent_tool.py
# 2 files already formatted
PYTHONPATH=packages/harness:. .venv/bin/python -m pytest tests/test_update_agent_tool.py -q
# 25 passed
PYTHONPATH=packages/harness:. .venv/bin/python -m pytest \
  tests/test_update_agent_tool.py tests/test_setup_agent_tool.py tests/test_update_agent_e2e_user_isolation.py -q
# 38 passed
```

E2E through real tools (paths mocked to a temp dir, same harness as unit tests):

| call | result |
|---|---|
| `update_agent(soul="   ")` | error; SOUL stays `"good soul"` |
| `update_agent(soul="new good soul")` | success; SOUL replaced |
| `update_agent(description=...)` only | success; SOUL unchanged |
| `setup_agent(soul="   ")` | still errors (sibling unchanged) |

## AI assistance

**Tool(s) used:** Grok Build (xAI)

**How you used it:** Used an AI coding assistant to confirm the setup_agent / update_agent sibling gap against #3553, draft the mirrored guard and regression tests, and run red→green / red-check / adjacent suite verification.

### DIFF
--- a/backend/packages/harness/deerflow/tools/builtins/update_agent_tool.py
+++ b/backend/packages/harness/deerflow/tools/builtins/update_agent_tool.py
@@ -150,7 +150,7 @@ def update_agent(
     # otherwise a custom agent can report success while wiping a working
     # SOUL.md and leaving the next turn with an empty personality.
     if soul is not None and not soul.strip():
-        return _err("soul content is empty; refusing to update agent with an empty SOUL.md")
+        return _err("soul content is empty; refusing to update agent with an empty SOUL.md. Omit the soul field if you do not want to change it.")
 
     try:
         agent_name = validate_agent_name(agent_name_raw)
--- a/backend/packages/harness/deerflow/tools/builtins/update_agent_tool.py
+++ b/backend/packages/harness/deerflow/tools/builtins/update_agent_tool.py
@@ -145,6 +145,13 @@ def update_agent(
     if soul is None and description is None and skills is None and tool_groups is None and model is None:
         return _err('No fields provided. Pass at least one of: soul, description, skills, tool_groups, model. Omit unchanged fields instead of passing null-like strings such as "null", "none", or "undefined".')
 
+    # Reject empty / whitespace-only soul before touching the filesystem.
+    # setup_agent already refuses this (#3553 / #3549); update_agent must too,
+    # otherwise a custom agent can report success while wiping a working
+    # SOUL.md and leaving the next turn with an empty personality.
+    if soul is not None and not soul.strip():
+        return _err("soul content is empty; refusing to update agent with an empty SOUL.md")
+
     try:
         agent_name = validate_agent_name(agent_name_raw)
     except ValueError as e:
--- a/backend/tests/test_update_agent_tool.py
+++ b/backend/tests/test_update_agent_tool.py
@@ -262,6 +262,36 @@ def test_update_agent_updates_soul_only(tmp_path, patched_paths):
     assert "soul" in result.update["messages"][0].content
 
 
+def test_update_agent_rejects_empty_soul_and_does_not_overwrite(tmp_path, patched_paths):
+    """Mirror setup_agent's empty-SOUL guard (#3553 / #3549).
+
+    setup_agent refuses empty/whitespace soul before touching the filesystem.
+    update_agent previously accepted the same input and reported success while
+    writing a blank SOUL.md, wiping a working agent personality.
+    """
+    agent_dir = _seed_agent(tmp_path, description="keep me", soul="original soul")
+
+    result = update_agent.func(runtime=_runtime(), soul="")
+
+    msg = result.update["messages"][0]
+    assert "soul content is empty" in msg.content
+    assert msg.status == "error"
+    assert (agent_dir / "SOUL.md").read_text() == "original soul"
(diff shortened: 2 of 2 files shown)


backend/packages/harness/deerflow/tools/builtins/update_agent_tool.py
```diff
@@ -145,6 +145,13 @@ def _err(message: str) -> Command:
     if soul is None and description is None and skills is None and tool_groups is None and model is None:
         return _err('No fields provided. Pass at least one of: soul, description, skills, tool_groups, model. Omit unchanged fields instead of passing null-like strings such as "null", "none", or "undefined".')
 
+    # Reject empty / whitespace-only soul before touching the filesystem.
+    # setup_agent already refuses this (#3553 / #3549); update_agent must too,
+    # otherwise a custom agent can report success while wiping a working
+    # SOUL.md and leaving the next turn with an empty personality.
+    if soul is not None and not soul.strip():
```
willem-bd, line 152: **Nit: guard ordering vs. name validation.** This check runs before `validate_agent_name` (line 156) and the `if not agent_name` check, so a call with no `agent_name` in context *and* an empty soul returns "soul content is empty" instead of the more fundamental "update_agent is only available inside a custom agent's chat".

No correctness or security impact - neither path touches the filesystem, and the guard doesn't use `agent_name` so there's no path-traversal surface - just a less-precise error in an edge case. It mirrors `setup_agent`'s early-guard style so it's fine as-is; flagging only in case you revisit: name-validation-first would give sharper errors.

backend/packages/harness/deerflow/tools/builtins/update_agent_tool.py
```diff
@@ -145,6 +145,13 @@ def _err(message: str) -> Command:
...
+        return _err("soul content is empty; refusing to update agent with an empty SOUL.md")
```
willem-bd: **Nit: message could guide the retry.** The "No fields provided" error above (line 146) explicitly tells the model to *omit* unchanged fields rather than pass null-like strings. This message doesn't say what to do instead, so the model may retry with another empty-ish value.

A trailing "Omit the `soul` field if you do not want to change it." would help it self-correct in one step and match the sibling message's helpfulness. (Minor: the PR body says the message "mirrors setup_agent", but setup_agent says "create" vs. "update" here - the create/update difference is correct, just the body's phrasing is slightly imprecise.)

willem-bd (commented):
Focused review of the empty-soul guard. The fix is correct - the `is not None` distinction for an optional field is exactly right, and the atomic reject + no-partial-write placement are sound. Two cosmetic nits inline; neither blocks merge.

AnnaSuSu:
Thanks for the review. Nit #2 applied in 50682bac — the empty-soul error now ends with "Omit the soul field if you do not want to change it." so the model self-corrects in one step instead of retrying with another empty-ish value; both regression tests now assert that guidance.

Left nit #1 (guard ordering) as-is per your note that it mirrors setup_agent's early-guard style. And you're right that the message says "update" vs setup_agent's "create" — the create/update wording is intentional; only the PR-body phrase "mirrors setup_agent" was slightly loose.

WillemJiang (approved):

AnnaQiQii:
goodgood

PR #4220, opened 2026-07-15
feat(skills): gate slash runs on skill-declared required-outputs

oThTJx:
﻿## Summary

- Add optional `required-outputs` skill frontmatter so slash-activated skills can declare deliverable basenames under thread `outputs/`.
- Introduce `SlashSkillDeliverableMiddleware` (TerminalResponse / `#4027` style): packaging reminder + `jump_to=model`, then `deerflow_error_fallback` so lead runs cannot end as silent success without the declared package (`#4176` class of failure).
- Tighten `{skill}.json` validation beyond a forged shell (`source` + `schema_version` + `generated_at`, plus known article-packager field shapes) and re-check in `after_agent` as a last-chance denial.

This ports the harness architecture from three commits on a staging branch (`aa91d7fb`, `296d3c76`, `b45ef79e`), squashed and trimmed for upstream: vendor-specific `content-*` skill trees / `package_review.py` path-rewrite patches are intentionally omitted because those packages are not in this repo. Skills that opt into `required-outputs` get the completion contract; others are unchanged.

## Motivation

Skill docs can say "run `package_*.py`", but models still hand-write JSON shells. Without a lead-run completion contract, the host sees `success` while the artifact is missing or invalid. Declaring outputs in frontmatter keeps the gate skill-driven rather than hardcoding a product skill list in the agent factory.

## Test plan

- [x] `pytest backend/tests/test_required_outputs.py backend/tests/test_slash_skill_deliverable_middleware.py` (18 passed)
- [ ] Slash skill **without** `required-outputs` still ends normally
- [ ] Slash skill **with** `required-outputs` and missing file → recovery reminder → then error fallback
- [ ] Valid packaged `{skill}.json` (`source` / `schema_version` / `generated_at`) allows terminal success
- [ ] Hand-written shell missing packager fields is rejected (including `after_agent` finalization)


backend/packages/harness/deerflow/agents/middlewares/slash_skill_deliverable_middleware.py
```diff
@@ -0,0 +1,454 @@
...
+        logger.debug("SlashSkillDeliverableMiddleware: skill registry lookup failed", exc_info=True)
+    return ()
+
+
+def output_is_satisfied(outputs_path: Path, skill_name: str, filename: str) -> bool:
+    """Return whether one required output basename is present and minimally valid."""
+    if "/" in filename or "\\" in filename or Path(filename).name != filename:
+        return False
+    path = outputs_path / filename
+    if not path.is_file():
```
fancyboi999, line 215: [P2] Reject stale outputs from earlier runs

`outputs_path` is stable per thread, but this only checks current existence/shape. I reproduced a new run returning `None` here when a valid file from an earlier run was left in the directory and the new run wrote nothing, so the silent-success contract is still bypassed. Record a fingerprint at activation/before_agent and require this run to create/change the output (or use run-scoped deliverables), with a two-run regression test.

backend/packages/harness/deerflow/agents/middlewares/slash_skill_deliverable_middleware.py
```diff
@@ -0,0 +1,454 @@
...
+        return False
+    if not filename.endswith(".json"):
+        return True
+    try:
+        payload = json.loads(path.read_text(encoding="utf-8"))
```
fancyboi999, line 220: [P2] Offload output validation from async hooks

Both `aafter_model` and `aafter_agent` call the synchronous validator directly. The repo's diff scanner flags `Path.is_file` and this `read_text` as async-reachable blocking I/O, and the strict Blockbuster gate raises `BlockingError: Blocking call to os.stat` on the real `aafter_model` path. Offload the whole validation pass with `asyncio.to_thread` in async hooks and add a `tests/blocking_io/` runtime anchor.

fancyboi999 (changes requested):
Two correctness issues reproduced on this head; details inline.

WillemJiang:
@oThTJx  You need to run the 'make install' from the root directory to install the per-commit to avoid the lint error. Please click the CLA assistant button to sign the CLA before we start the release.

PR #4222, opened 2026-07-16
fix(channels): accept leading @mentions before /connect bind codes

AnnaSuSu:
## Summary

Group Feishu/DingTalk messages often look like `@_user_1 /connect <code>` (the platform requires @bot to deliver the message). `extract_connect_code` only accepted a bare leading `/connect`, so those binds silently fell through to chat. Slack and Discord already strip bot mentions before calling the shared helper.

## Approach

- In shared `extract_connect_code`, skip leading mention tokens (`@…`, `<@…>`).
- Bare `/connect <code>` unchanged.
- Mid-sentence `/connect` still does not bind.

## Test plan

- [x] `@_user_1 /connect code` / `@bot /connect` / `<@id> /connect` extract the code
- [x] mid-sentence and non-connect mentions stay non-binding
- [x] `tests/test_additional_channel_connections.py` + connect-related suite green
- [x] ruff clean

## AI assistance

How you used it: Used Claude Code to compare Slack/Discord mention stripping with Feishu/DingTalk connect parsing and draft the shared `extract_connect_code` change plus regression tests; I verified the bug by driving `extract_connect_code` / `_pending_connect_code` on `@_user_1 /connect <code>` (returned `None` before, code after) and re-ran the channel connection tests and ruff on the touched files.

I've read and understand every line of this change and take responsibility for it — it's not unreviewed AI output.

### DIFF
--- a/backend/app/channels/commands.py
+++ b/backend/app/channels/commands.py
@@ -21,14 +21,42 @@ KNOWN_CHANNEL_COMMANDS: frozenset[str] = frozenset(
 )
 
 
+def _is_leading_mention_token(token: str) -> bool:
+    """Return whether *token* looks like a platform bot/user mention.
+
+    Group chats often require ``@bot`` before the message is delivered. Slack
+    and Discord strip those tokens before connect parsing; Feishu / DingTalk
+    leave them in the text (``@_user_1``, ``@bot``, ``<@id>``). Treat them as
+    transport noise only when they lead the message so
+    ``@bot /connect <code>`` still binds.
+    """
+    if not token:
+        return False
+    # Slack / Discord style: <@U123> or <@!U123> or <@U123|name>
+    if token.startswith("<@") and token.endswith(">"):
+        return True
+    # Feishu / DingTalk / generic: @_user_1, @bot, @nickname
+    if token.startswith("@") and len(token) > 1:
+        return True
+    return False
+
+
 def extract_connect_code(text: str) -> str | None:
-    """Extract the one-time channel binding code from a connect command."""
+    """Extract the one-time channel binding code from a connect command.
+
+    Accepts a leading platform mention so group ``@bot /connect <code>``
+    messages bind the same way as bare ``/connect <code>`` (Slack/Discord
+    already strip mentions before calling this helper).
+    """
     parts = text.strip().split()
-    if len(parts) < 2:
+    index = 0
+    while index < len(parts) and _is_leading_mention_token(parts[index]):
+        index += 1
+    if index + 1 >= len(parts):
         return None
-    command = parts[0].lower()
+    command = parts[index].lower()
     if command in {"/connect", "connect"}:
-        return parts[1]
+        return parts[index + 1]
     return None
 
 
--- a/backend/tests/test_additional_channel_connections.py
+++ b/backend/tests/test_additional_channel_connections.py
@@ -29,11 +29,35 @@ def test_pending_connect_code_extracts_code_when_connections_configured():
     assert channel._pending_connect_code("hello world") is None
 
 
+def test_pending_connect_code_accepts_leading_platform_mentions():
+    """Group chats prefix @bot; Feishu/DingTalk leave that noise in the text.
+
+    Slack/Discord strip mentions before this helper; the shared parser must
+    still accept the unstripped form so @bot /connect <code> binds.
+    """
+    from app.channels.commands import extract_connect_code
+
+    assert extract_connect_code("@_user_1 /connect abc123") == "abc123"
+    assert extract_connect_code("@bot  /connect code-xyz") == "code-xyz"
+    assert extract_connect_code("<@U123ABC> /connect slackish") == "slackish"
+    assert extract_connect_code("@bot @_user_2 /connect multi") == "multi"
+    # Mentions without a connect command stay non-binding.
+    assert extract_connect_code("@_user_1 hello") is None
+    # Connect not at the command position after mentions does not bind.
+    assert extract_connect_code("please @bot /connect leaked") is None
+    # Mid-sentence /connect must not bind (unchanged).
+    assert extract_connect_code("hi /connect abc123") is None
+
+    channel = _StubChannel(name="stub", bus=MessageBus(), config={"connection_repo": object()})
+    assert channel._pending_connect_code("@_user_1 /connect via-base") == "via-base"
+
+
 def test_pending_connect_code_is_none_when_connections_disabled():
     # With no connection repo, binding is not configured and connect codes are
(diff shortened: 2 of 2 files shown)


backend/app/channels/commands.py
```diff
@@ -21,14 +21,42 @@
...
+    # Slack / Discord style: <@U123> or <@!U123> or <@U123|name>
+    if token.startswith("<@") and token.endswith(">"):
+        return True
+    # Feishu / DingTalk / generic: @_user_1, @bot, @nickname
+    if token.startswith("@") and len(token) > 1:
+        return True
+    return False
+
+
 def extract_connect_code(text: str) -> str | None:
```
willem-bd, line 44: Scope note (not a blocker for this PR): this fixes `/connect`, but the same `@bot` prefix affects general command dispatch too. `is_known_channel_command` below requires a bare leading `/` (`text.startswith("/")`), so in a Feishu/DingTalk group chat `@bot /connect <code>` now binds while `@bot /goal` / `@bot /status` still fall through to CHAT. If group chats require `@bot` to deliver, the mention-prefix problem spans the whole command surface — worth a follow-up so this fix doesn't read as arbitrarily narrow.

backend/tests/test_additional_channel_connections.py
```diff
@@ -29,11 +29,35 @@ def test_pending_connect_code_extracts_code_when_connections_configured():
...
+    """Group chats prefix @bot; Feishu/DingTalk leave that noise in the text.
+
+    Slack/Discord strip mentions before this helper; the shared parser must
+    still accept the unstripped form so @bot /connect <code> binds.
+    """
+    from app.channels.commands import extract_connect_code
+
+    assert extract_connect_code("@_user_1 /connect abc123") == "abc123"
+    assert extract_connect_code("@bot  /connect code-xyz") == "code-xyz"
+    assert extract_connect_code("<@U123ABC> /connect slackish") == "slackish"
```
willem-bd: The `<@…>` branch docstring (`commands.py:35`) lists `<@!U123>` (ping) and `<@U123|name>` (name) variants, but only the plain `<@U123ABC>` form is exercised here. Worth pinning those two as well, plus a couple of uncovered edge cases: case-insensitive command (`@bot /Connect code` — handled by `.lower()` but unasserted) and empty/mention-only input (`""`, `@bot` → `None`).

backend/tests/test_additional_channel_connections.py
```diff
@@ -29,11 +29,35 @@ def test_pending_connect_code_extracts_code_when_connections_configured():
     assert channel._pending_connect_code("hello world") is None
 
 
+def test_pending_connect_code_accepts_leading_platform_mentions():
+    """Group chats prefix @bot; Feishu/DingTalk leave that noise in the text.
+
+    Slack/Discord strip mentions before this helper; the shared parser must
+    still accept the unstripped form so @bot /connect <code> binds.
+    """
+    from app.channels.commands import extract_connect_code
```
willem-bd: Nit: the rest of the file imports at module top. Moving `from app.channels.commands import extract_connect_code` up with the others would match the file's convention.

willem-bd (commented):
Overall: clean, correct, well-tested fix. The logic checks out (boundary guard `index + 1 >= len(parts)`, multi-mention `while` loop, mid-sentence `/connect` still non-binding) and the security model is preserved — the connect code binds the sender's identity (`sender_staff_id`/`user_id`), not the mentioned user, so skipping a leading `@mention` can't enable impersonation. Left a few inline notes; none are blockers.

backend/tests/test_additional_channel_connections.py
AnnaSuSu: Done in 9f04b40b — moved `extract_connect_code` up to the module-top imports with the others.

backend/tests/test_additional_channel_connections.py
AnnaSuSu: Added in 9f04b40b. Pinned the two mention forms the `_is_leading_mention_token` docstring lists — `<@!U123ABC>` (ping) and `<@U123ABC|alice>` (name) — plus the edges you called out: `@bot /Connect …` for the `.lower()` case-insensitivity, and `@bot` / `""` returning `None` for the mention-only/empty boundary.

Red-checked each against pre-fix `commands.py`: the three variant asserts and the case one go red without the fix; the two boundary asserts are green regardless, so they're documented edges rather than fix guards.

backend/app/channels/commands.py
AnnaSuSu: You're right, and it's a real pre-existing gap. Confirmed: Feishu keeps the `@`mention in the text (`feishu.py` ~1001, "Include both normal text and @ mentions"), `_pending_connect_code` runs before classification so `@bot /connect <code>` now binds, but `is_known_channel_command` still requires a bare leading `/`, so `@bot /goal` / `@bot /status` stay CHAT in Feishu/DingTalk group chats.

A correct general fix touches two sites, not one: classification (`is_known_channel_command`, shared by feishu/dingtalk/discord/slack/wechat/wecom) **and** the parser `_handle_command` (`manager.py:1847`), which also assumes `raw_text.startswith("/")` — flipping only classification would mark `@bot /goal` as COMMAND and then bounce it as "unknown". That's the whole-command-surface part you flagged, so I'd rather land it as a focused follow-up than widen this PR.

I kept `/connect` here on purpose: it runs *before* command classification, in the identity-bootstrap path (consumed ahead of `allowed_users`), so it's fixable in isolation without touching dispatch. Follow-up for the general command surface incoming.

AnnaSuSu: Correction to my "follow-up incoming" line above — after digging in, the general command surface isn't the clean sibling I implied, so I'm going to leave it out of this bug-fix track rather than over-promise.

Slack/Discord already handle `@bot /cmd` correctly: `_strip_leading_slack_bot_mention` strips **only the bot's own** leading mention (identity-aware via a resolved `bot_user_id`) and deliberately keeps a leading *non-bot* mention like `<@someone> /help <@bot>` as CHAT (`test_app_mention_preserves_leading_non_bot_mention_when_bot_id_unknown`). Matching that for Feishu/DingTalk needs bot-identity resolution those adapters don't currently have — Feishu keeps only the `at` element's display text, not its id — which is feature-adjacent rather than a bug fix. And a blanket "strip any leading mention" in the shared classifier regresses exactly that Slack non-bot-mention case.

So this PR stays scoped to the `/connect` bind path (which is safe to strip leniently because it binds the sender's identity, as you noted). Happy to revisit the command surface as a separate, properly-scoped change if you think it's worth the bot-identity plumbing.

WillemJiang (approved):

PR #4225, opened 2026-07-16
fix(memory): consolidated facts inherit expected_valid_days from sources

sontianye:
## Why

This PR closes a seam between two memory features that were both contributed to this repo but never wired together, because of the order in which they landed and a module-level refactor that sat between them.

**Timeline (all three are prior contributions to this repo):**
- **#3996** (merged 2026-07-10) added memory consolidation - synthesising fragmented facts into one richer fact.
- **#4122** (merged 2026-07-15) refactored the memory module into a pluggable `MemoryManager` abstraction with a self-contained DeerMem backend, moving the staleness + consolidation logic from the single-file `agents/memory/updater.py` into `backends/deermem/deermem/core/`.
- **#4143** (merged 2026-07-16) added per-fact `expected_valid_days` (LLM-assigned review window at creation) and `staleFactsToExtend` (recalibration at review time).

`#3996` was written and merged **before** `expected_valid_days` existed as a concept, so its consolidated `new_fact` naturally carried no lifetime field. `#4122` then ported consolidation's logic verbatim into the new backend layout - it was a structural move, not a behavioural one, so it preserved the no-evd behaviour. `#4143` added per-fact lifetimes to the `newFacts` ingestion path and the staleness review path, but consolidation is a *third* code path that produces a new fact, and at port time onto the post-`#4122` layout the consolidation path was not revisited for the new lifetime semantics. The result: two correct, individually-reviewed features that do not talk to each other at the one place they intersect - the consolidated fact.

**The concrete bug this produces:** a consolidated fact inherits the newest source's `createdAt` (so its staleness clock reflects the underlying information, not synthesis time - by `#3996`'s own design) but **no** `expected_valid_days`. At read time `_effective_fact_staleness_age` therefore falls back to the global `staleness_age_days` (default 90). So if the LLM merges several 200-day-old stable facts that each carry `expected_valid_days: 3650` (a 5-year window), the merged fact:

1. lands with `createdAt` = 200 days ago (newest source),
2. has no `expected_valid_days`, so its review window collapses to 90 days,
3. is **immediately a staleness candidate again on the next cycle**, because 200 > 90.

The merge silently discards the lifetime signal of every source fact and contradicts consolidation's own premise - these were judged stable, related facts worth synthesising, yet the synthesis forces them back into review within a quarter. Worse, this is silent: there is no log, no error, the merged fact just re-enters the candidate set.

## What changed

The consolidated `new_fact` now inherits `expected_valid_days` set so it is re-reviewed at the **earliest source review deadline**:

```python
source_deadlines = [
    dt + timedelta(days=_effective_fact_staleness_age(fact_index[sid], config))
    for sid, dt in zip(source_ids, _source_dts)
]
earliest_deadline = min(source_deadlines)
days_until_earliest = int((earliest_deadline - _newest_dt).total_seconds() // 86400)
inherited_evd = max(days_until_earliest, 1)  # overdue source -> review next cycle
new_fact["expected_valid_days"] = min(inherited_evd, creation_cap)
```

- **Earliest source review deadline, not the longest source window.** A merge combines details from every source, so a volatile sub-detail (e.g. `evd=7`) must not inherit a stable source's 3650-day window and escape staleness review for years - staleness KEEP/REMOVE is the *only* path that re-validates a merged fact, so biasing toward the soonest deadline keeps uncertain merges re-checked sooner. (An earlier iteration of this PR took `max(source_evds)`; review feedback from @fancyboi999 / @willem-bd pointed out that a volatile sub-detail could then be buried for years, which is the same class of bug this PR fixes - so the direction was reversed to the earliest deadline.)
- **Every source participates, including legacy facts.** `_effective_fact_staleness_age` returns the stored `expected_valid_days` when present and falls back to the global `staleness_age_days` otherwise, so a legacy source's default 90-day window is not silently swallowed by a long-lived sibling. As a consequence the merged fact now **always** carries `expected_valid_days` (there is no omit branch) - the all-legacy case derives a window from the global fallback instead of leaving the field absent.
- **Overdue sources clamp to a minimal positive window.** A source already past its deadline yields `inherited_evd = 1` (review next cycle) rather than the global fallback, which would defer an overdue review.
- **Capped at the creation-time `staleness_max_lifetime_multiplier`**, identical to the `newFacts` path, so consolidation cannot defer first review indefinitely. The cap is hoisted out of the per-merge loop since it is a config constant.

This mirrors how `createdAt` and `sourceError` are already derived from the sources on the consolidated fact, so lifetime inheritance completes the "the merged fact carries the provenance of its sources" pattern rather than introducing a new one.

**Why not let the LLM assign the merged lifetime?** The consolidation prompt does not surface source `expected_valid_days` to the model (it shows `[id | confidence] content`, not `valid:Nd`), so asking the model for a merged `expected_valid_days` would be a guess without signal. Source inheritance is deterministic, always available, and uses the real lifetime signal each source fact already carries. Surfacing source evd to the LLM and accepting an explicit override is a reasonable future extension, but it belongs in a consolidation-prompt change, not this fix.

**DRY refactor:** the evd type rule was previously inlined in four places - `_normalize_memory_update_fact`, `_effective_fact_staleness_age`, the `newFacts` creation cap, and now consolidation. It is extracted into one helper, `_read_expected_valid_days(fact) -> int | None`, and all four call sites use it. The rule now coerces to `int` *before* the `> 0` check (so a fractional `0.5` does not leak as `0`) and rejects non-finite floats (`NaN`, `+/-inf`) before the coercion - `int(nan)` raises `ValueError` and `int(inf)` raises `OverflowError`, and Python's JSON decoder accepts them as floats, so one malformed `expected_valid_days` in a hand-edited `memory.json` could otherwise abort staleness selection or consolidation.

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [x] **Agents / LangGraph** - DeerMem backend consolidation apply path + shared evd helper
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [x] **Default behavior change**:
  - Consolidated facts now **always** carry `expected_valid_days` (inherited from sources via the earliest-deadline rule), where they previously omitted it. The next-cycle staleness eligibility of a merged stable fact no longer spuriously re-enters review, while a merge involving a volatile or legacy source re-enters review at that source's sooner deadline instead of inheriting a long window.
  - **Not strictly additive for pre-`#4143` facts.** A legacy source (no `expected_valid_days`) merged with a long-lived source previously let the long-lived window govern (via the omitted-field -> global-fallback path); now the legacy source's 90-day fallback deadline participates and can drag the merged window sooner. This is the intended fix (a legacy detail should not be buried by a stable sibling), but the "cannot regress facts that predate `#4143`" property of an earlier iteration no longer holds.
  - A hand-edited float `expected_valid_days` in `(0, 1)` (e.g. `0.5`) now falls back to the global `staleness_age_days` at read time instead of being treated as `0` (immediately stale). This is a correct fix - a 0-day lifetime was never intended - surfaced by routing `_effective_fact_staleness_age` through the shared helper that coerces before the positivity guard.

No config schema change and no `config_version` bump: the fix reuses the existing `staleness_max_lifetime_multiplier` field already on `DeerMemConfig` (`#4143`); it introduces no new config. `backend/AGENTS.md` consolidation description is updated to mention the inherited lifetime.

## Bug fix verification

- Test that reproduces the bug: `tests/test_memory_consolidation.py::TestReviewerFindings::test_consolidation_preserves_stable_lifetime_across_cycle` - merges two 200-day-old `evd=3650` sources and asserts the merged fact does **not** re-enter `_select_stale_candidates` on the next cycle.
- Did it go red on `main` and green on this branch? **Yes** - on `main` the merged fact has no `expected_valid_days`, falls back to the 90-day global age, and is selected as stale (test fails); on this branch it inherits 1800 (capped) and is excluded (test passes).
- The earliest-deadline direction (vs an earlier `max` iteration) is locked by `test_consolidated_evd_volatile_source_governs_earliest_deadline` (volatile `evd=7` + stable `evd=3650` -> merged `evd=7`, immediately stale next cycle) and `test_consolidated_evd_legacy_source_not_swallowed_by_stable_sibling` (legacy 90-day fallback governs over a 3650 sibling).
- The non-finite / coercion-order fixes are covered at three layers: `TestReadExpectedValidDays` (helper), `TestEffectiveFactStalenessAge` (read path), and `test_consolidation_with_non_finite_source_evd_does_not_raise[nan|inf|-inf]` (end-to-end consolidation).

## Validation

```
# ruff: All checks passed
# memory suite: 219 passed, 1 skipped (incl. consolidation evd cases, the
#   non-finite/parametrized end-to-end, and TestReadExpectedValidDays units)
```

## AI assistance

**How you used it:** Pair-programming - I authored consolidation (#3996), the pluggable-memory port that surfaced this seam, and per-fact `expected_valid_days` (#4143). Claude Code assisted with the inheritance implementation, the `_read_expected_valid_days` DRY extraction across the four call sites, and the test coverage.


fancyboi999 (changes requested):
Thanks @sontianye. I found a couple of issues that should be addressed before this is ready.

[P2] Preserve the earliest review deadline when consolidating facts

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:1298-1300`
- Problem: Taking the maximum source lifetime lets one stable source postpone review of every volatile detail retained in the consolidated fact.
- Evidence: The new `test_consolidated_evd_takes_max_of_mixed_sources` merges a 3650-day fact with a 7-day fact and deliberately stores 1800 days. Consolidation keeps the combined content and the newest source `createdAt`, so the 7-day detail can remain injected for roughly five years without entering staleness review.
- Suggested fix: Preserve the earliest absolute source review deadline (`createdAt + expected_valid_days`) relative to the merged fact timestamp, or avoid consolidating source details whose review windows cannot safely be represented by one fact.
- Test: Cover mixed 7-day/3650-day sources with both equal and different `createdAt` values, and assert the merged fact becomes eligible at the earliest source deadline.

[P2] Validate the coerced day count instead of the raw float

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:404-406`
- Problem: A value between 0 and 1 passes `raw > 0` and is then truncated to `0`, despite the helper contract promising a positive integer. `_effective_fact_staleness_age()` treats that zero as an explicit lifetime instead of falling back to the configured age, so an existing float-valued fact becomes immediately stale.
- Evidence: The previous normalization converted first and only kept the value when the resulting integer was positive. The shared helper reverses that order, changing behavior for values such as `0.5`.
- Suggested fix: Convert to `int` first, then return it only when the converted value is greater than zero.
- Test: Assert `0.5` returns `None` and that `_effective_fact_staleness_age()` falls back to `staleness_age_days` for that stored value.

Reviewed head SHA: `510e68bccdb710e683941bad74763f4b0f0dbf8e`

willem-bd (commented):
Reviewing against head `510e68b`. Clean, well-motivated, strictly-additive fix with a genuine red/green repro, and the DRY extraction is behavior-preserving on the production path (newFacts are already coerced to `int` by `_normalize_memory_update_fact` before `_apply_updates` reads them, so widening the type rule there is a no-op in prod). Overall positive — the three inline notes below are one design consideration, one minor consistency nit, and one latent test-fragility item; none are blocking.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -1267,6 +1283,22 @@ def _apply_updates(
...
+                    # aspect of the user, so the merged fact should not be re-
+                    # reviewed sooner than its most-stable source.  Capped at the
+                    # creation-time multiplier (hoisted above the loop) like any new
+                    # fact so consolidation cannot defer first review indefinitely.
+                    # Omitted when no source carries a valid evd (legacy facts fall
+                    # back to the global age at read time, matching pre-feature
+                    # behaviour).
+                    source_evds = [v for v in (_read_expected_valid_days(fact_index[sid]) for sid in source_ids) if v is not None]
+                    if source_evds:
+                        new_fact["expected_valid_days"] = min(max(source_evds), creation_cap)
```
willem-bd: Taking `max(source_evds)` is the conservative direction and the PR body justifies it well, but flagging the one edge case worth being deliberate about: if a genuinely volatile fact (e.g. `evd=7`) is merged into a stable one (`evd=3650`), the volatile sub-information inherits the 3650-day window and won't be re-reviewed for years — and the staleness KEEP/REMOVE pass can't catch it, because at that window it never becomes a candidate.

Since consolidation only fires when the LLM judges the facts to describe the same aspect (and confidence is capped at source-max), `max` is defensible. Calling it out only as the non-obvious choice here; `min` (re-review at the soonest any source would have) would be the alternative if you'd rather bias toward re-validating uncertain merges sooner.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -1179,6 +1191,10 @@ def _apply_updates(
                 fact_index = {f.get("id"): f for f in current_memory.get("facts", []) if isinstance(f, dict)}
                 max_groups = config.consolidation_max_groups_per_cycle
                 max_sources = config.consolidation_max_sources
+                # Creation-time lifetime cap shared with the newFacts path: an
+                # inherited expected_valid_days is clamped so a merge of long-lived
+                # sources cannot defer first review indefinitely.
+                creation_cap = int(config.staleness_age_days * config.staleness_max_lifetime_multiplier)
```
willem-bd, line 1235: Minor consistency nit (not blocking): `creation_cap` is hoisted above the loop here, but the `newFacts` path recomputes the same expression inside its loop a few dozen lines up. Functionally identical since `config` is loop-invariant — hoisting it in `newFacts` too would just make the two fact-creation paths read identically.

backend/tests/test_memory_consolidation.py
```diff
@@ -895,6 +895,238 @@ def test_consolidated_created_at_tracks_newest_source(self):
...
+            consolidation_enabled=True,
+            consolidation_min_facts=2,
+            consolidation_max_groups_per_cycle=3,
+            consolidation_max_sources=8,
+            staleness_age_days=90,
+            staleness_max_lifetime_multiplier=20.0,
+            staleness_min_candidates=1,
+        )
+        # Two 200-day-old stable facts (evd 5 years) in the same category.
+        old_date = "2025-01-01T00:00:00Z"
```
willem-bd: Latent time-sensitivity: this hardcodes `createdAt` to `2025-01-01` against an inherited 1800-day window. The "NOT yet stale" assertion holds today, but will flip around **late 2029** once the fact exceeds 1800 days old — a silent future break with no obvious cause.

Consider deriving `createdAt` relative to `now` (e.g. `now - 200 days`) so the repro stays valid regardless of when it runs.

sontianye:
@fancyboi999 thanks - both points were valid and are fixed in `9a5d1896`.

**[P2] Earliest source deadline instead of max source lifetime**
You're right that taking `max(source_evds)` lets a stable source postpone review of every volatile detail retained in the merged fact, and that staleness KEEP/REMOVE is the only path that re-validates a merged fact - so a volatile sub-detail could escape review for years. Changed to your first suggestion: the merged fact is re-reviewed at the **earliest source review deadline**, `min(createdAt + expected_valid_days)` across sources, expressed relative to the merged fact's `createdAt` (the newest source's):
```python
earliest_deadline = min(dt + timedelta(days=evd) for dt, evd in source_deadline_pairs)
days_until_earliest = int((earliest_deadline - _newest_dt).total_seconds() // 86400)
inherited_evd = max(days_until_earliest, 1)  # overdue source -> review next cycle
new_fact["expected_valid_days"] = min(inherited_evd, creation_cap)
```
A source already past its deadline yields a minimal positive window (review next cycle) rather than the global fallback, which would defer an overdue review. Covered by `test_consolidated_evd_volatile_source_governs_earliest_deadline` (volatile evd=7 merged with stable evd=3650, both 100 days old -> merged evd=7, immediately a staleness candidate next cycle) and `test_consolidated_evd_overdue_source_clamps_to_minimal_window` (old source deadline before merged createdAt -> clamped to 1), plus the equal/different `createdAt` cases you asked for.

**[P2] Coerce before the positivity guard**
Confirmed: the shared helper did `raw > 0` then `int(raw)`, so `0.5` passed the float check and truncated to `0`, violating the "positive int or None" contract. Fixed to `int()` first, then `> 0` - matching the original `_normalize_memory_update_fact` order. `_read_expected_valid_days({"expected_valid_days": 0.5})` now returns `None`, and `_effective_fact_staleness_age` falls back to `staleness_age_days` for it (regression tests in `TestReadExpectedValidDays` / `TestEffectiveFactStalenessAge`).

One small correction on the downstream effect: `_effective_fact_staleness_age` would not actually treat `0` as an explicit lifetime - `0 > 0` is false, so it falls back to the global age regardless. The real problem was the helper returning a non-positive `0` (a contract violation that could leak as dirty `expected_valid_days` on write), not an immediate-stale read. Fixed either way.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
sontianye: Addressed - switched from `max(source_evds)` to the earliest source review deadline (`min(createdAt + expected_valid_days)`), per the concern you and @fancyboi999 both raised: a volatile sub-detail must not inherit a stable source's long window when staleness KEEP/REMOVE is the only re-validation path. See the PR comment for the full rationale.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
sontianye: Done - hoisted `creation_cap` above the `newFacts` loop too, so both fact-creation sites read identically.

backend/tests/test_memory_consolidation.py
sontianye: Good catch - switched those tests to a `_days_ago(n)` helper that derives `createdAt` relative to `now`, so the repro stays valid regardless of when it runs (no more 2029 flip).

fancyboi999 (changes requested):
Thanks @sontianye. I found one remaining lifetime propagation issue that should be addressed before this is ready.


Reviewed head SHA: `9a5d1896f067bc16d0ed7e553fd38356e76b9619`

[P2] Include legacy-source fallback deadlines in mixed consolidations

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:1313-1325`
- Problem: Sources without `expected_valid_days` are omitted from `source_deadline_pairs`. When a legacy source is consolidated with a long-lived source, the merged fact inherits only the long-lived deadline and can postpone review of the legacy detail far beyond its configured global window.
- Evidence: `_effective_fact_staleness_age()` treats a missing EVD as `staleness_age_days`, but the new comprehension filters that source out whenever `_read_expected_valid_days()` returns `None`. If a same-age legacy source should be reviewed after the default 90 days and its sibling carries EVD 3650, this branch derives the merged lifetime solely from 3650 (then caps it, typically at 1800). The all-legacy test does not cover this mixed case.
- Suggested fix: Compute an absolute deadline for every source using its effective lifetime, including the configured fallback for missing/invalid EVD, then preserve the earliest deadline relative to `_newest_dt` with the existing positive/cap bounds.
- Test: Cover mixed legacy/stable sources with equal and different `createdAt` values, and assert the merged fact becomes eligible at the legacy source's global fallback deadline when that is earliest.

sontianye:
@fancyboi999 good catch - fixed in `327e8121`.

**[P2] Legacy sources now participate in the earliest-deadline computation**
You're right that filtering sources without `expected_valid_days` out of `source_deadline_pairs` let a legacy source's default 90-day window be silently swallowed by a long-lived sibling. The fix reuses `_effective_fact_staleness_age(fact, config)` per source - which returns the stored `expected_valid_days` when present and falls back to `staleness_age_days` otherwise - so every source contributes its deadline:
```python
source_deadlines = [
    dt + timedelta(days=_effective_fact_staleness_age(fact_index[sid], config))
    for sid, dt in zip(source_ids, _source_dts)
]
earliest_deadline = min(source_deadlines)
```
Reusing `_effective_fact_staleness_age` keeps the "missing evd -> global fallback" rule in one place (the same function the read path uses), rather than re-implementing it in the consolidation branch. As a side effect the merged fact now always carries an `expected_valid_days` (no omit branch) - the all-legacy case derives a window from the global fallback instead of leaving the field absent, which is more honest and re-enters review at the same point (age > 90) either way.

Covered by:
- `test_consolidated_evd_all_legacy_sources_uses_global_fallback_deadline` (two 100-day-old legacy facts -> evd 90, immediately stale)
- `test_consolidated_evd_legacy_source_not_swallowed_by_stable_sibling` (mixed legacy/stable with both equal and different `createdAt`; legacy 50 days old -> evd 90 governs over the stable 3650; legacy 200 days old with a fresh stable sibling -> overdue, clamped to 1)

fancyboi999 (changes requested):
Thanks @sontianye. The legacy-source deadline fix closes the prior review, but the shared lifetime reader introduces one malformed-persistence regression.


Reviewed head SHA: `327e812128bdbc36a8bc6c5ef764d3094def2690`

[P2] Reject non-finite lifetimes before integer coercion

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:394-412`
- Problem: `_read_expected_valid_days()` now backs the persisted-fact read path, but `int(raw)` raises for `NaN` and infinity. One malformed `expected_valid_days` in `memory.json` can therefore abort staleness selection or consolidation instead of falling back to the configured global lifetime.
- Evidence: Python's JSON decoder accepts `NaN` as a float by default, and `int(float("nan"))` raises `ValueError`. On current `main`, `_effective_fact_staleness_age()` checks `raw > 0` first, so `NaN` falls back to `staleness_age_days`; this diff routes the same stored value through the new unconditional integer conversion. The helper explicitly treats hand-edited persisted floats as supported input, so this is part of its stated boundary rather than an unreachable model response.
- Suggested fix: require `math.isfinite(float(raw))` before coercing numeric values, then keep the existing positivity check and fallback behavior.
- Test: cover `NaN` and positive/negative infinity in `_read_expected_valid_days()` and drive at least the `_effective_fact_staleness_age()` persisted-fact path to assert it returns the global fallback without raising.

sontianye:
@fancyboi999 confirmed - fixed in `b0356ff0`.

**[P2] Reject non-finite lifetimes before int coercion**
You're right that routing the persisted-fact read path through `int(raw)` introduces a regression: `int(nan)` raises `ValueError` and `int(inf)` / `int(-inf)` raise `OverflowError`, and Python's JSON decoder accepts `NaN` / `Infinity` as floats, so one malformed `expected_valid_days` in a hand-edited `memory.json` could abort staleness selection or consolidation. Added `math.isfinite(float(raw))` before the coercion:
```python
raw = fact.get("expected_valid_days")
if isinstance(raw, (int, float)) and not isinstance(raw, bool) and math.isfinite(float(raw)):
    evd = int(raw)  # coerce before the positivity guard
    if evd > 0:
        return evd
return None
```
One refinement to the framing: on `main`, `_effective_fact_staleness_age` checked `raw > 0` first, so `NaN` fell back safely (`nan > 0` is false) - but `inf` did **not** (`inf > 0` is true, so `main` also crashed on `inf`). Since this helper is now the persisted-fact read path for both staleness and consolidation, closing it here also closes that pre-existing `inf` crash across all four call sites. Normal int/float values (including large ones) are unaffected - `isfinite` is a no-op for them.

Tests (all three layers, all three non-finite values):
- `TestReadExpectedValidDays.test_rejects_non_finite_values` - `NaN`, `+inf`, `-inf` return `None` (not raise).
- `TestEffectiveFactStalenessAge.test_falls_back_for_non_finite_values` - the persisted-fact read path returns the global age for each, no raise.
- `test_consolidation_with_non_finite_source_evd_does_not_raise[nan|inf|-inf]` - parametrized end-to-end: a non-finite-evd source merged with a stable source does not abort consolidation; the bad source's effective lifetime falls back to the global 90 and its deadline participates in the earliest-deadline computation.

willem-bd (commented):
Code review of the consolidation `expected_valid_days` inheritance, verified against head `b0356ff0`.

Note: the PR was force-pushed since the initial review. The new commit `b0356ff0` also fixed a NaN/Inf crash in `_read_expected_valid_days` (`int(nan)` -> ValueError, `int(inf)` -> OverflowError on a hand-edited memory.json) that would have been a finding against the earlier head — that's now resolved, plus the consolidation path was refactored to use `_effective_fact_staleness_age` (legacy sources now participate via the global fallback).

Net: the implementation, tests, and `AGENTS.md` are internally consistent and the design is sound. The main blocker is that the PR **body is stale** — it describes a "max source window / omit-when-no-evd / strictly additive" design, while the shipped code takes the earliest source deadline and always sets the field. Details inline.

Minor items not anchored inline:
- `creation_cap == 0` edge: if `staleness_age_days * staleness_max_lifetime_multiplier < 1`, `int(...)` is `0` and `min(inherited_evd, 0)` writes `expected_valid_days: 0` (read-time treats 0 as absent via the helper, so benign). Pre-existing in the `newFacts` path too; a `max(..., 1)` on the cap would make both sites obviously safe. Not blocking.
- The earlier "no mixed evd/no-evd test" gap is now addressed by `test_consolidated_evd_legacy_source_not_swallowed_by_stable_sibling`.

Recommend reconciling the body and tidying the one test docstring; then this is good to go.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -1267,6 +1295,41 @@ def _apply_updates(
...
+                    # read-time fallback), so a legacy source's default 90-day window
+                    # is not silently swallowed by a long-lived sibling. The deadline
+                    # is expressed relative to the merged fact's createdAt (the newest
+                    # source's), so a source already past its deadline yields a
+                    # minimal positive window (review next cycle) rather than the
+                    # global fallback, which would otherwise defer an overdue review.
+                    # Capped at the creation-time multiplier (hoisted above the loop)
+                    # like any new fact so consolidation cannot defer first review
+                    # indefinitely.
+                    source_deadlines = [dt + timedelta(days=_effective_fact_staleness_age(fact_index[sid], config)) for sid, dt in zip(source_ids, _source_dts)]
```
willem-bd: **PR body describes a different design than what ships here.** The body's "What changed" says it inherits the **max** source window (`new_fact["expected_valid_days"] = min(max(source_evds), creation_cap)`) and is "Omitted when no source carries a valid evd ... strictly additive and cannot regress facts that predate #4143." The shipped code does neither:

1. It takes the **earliest** source review deadline (`min(dt + effective_lifetime)`), not the max. For a volatile (evd=7) + stable (evd=3650) merge this inherits 7, not 3650/1800 — the body's stated rationale ("should not be re-reviewed *sooner* than its most-stable source") is the opposite of the shipped behavior, and `test_consolidated_evd_volatile_source_governs_earliest_deadline` asserts 7.
2. It uses `_effective_fact_staleness_age`, so legacy/no-evd sources **participate** via the global `staleness_age_days` fallback and the field is **always set**, not omitted. A legacy source merged with a stable one now drags the merged window to the legacy source's (possibly overdue) global-fallback deadline (see `test_consolidated_evd_legacy_source_not_swallowed_by_stable_sibling`), which contradicts the body's "cannot regress facts that predate #4143" claim.

The shipped design is defensible (the code comment and tests make the case that a soonest-expiring/volatile sub-detail shouldn't be buried), but the body needs to be reconciled — a reviewer reading only the body would approve a different design and miss that the "strictly additive / cannot regress" safety claim no longer holds.

backend/tests/test_memory_consolidation.py
```diff
@@ -895,6 +902,478 @@ def test_consolidated_created_at_tracks_newest_source(self):
...
+            max_facts=100,
+            consolidation_enabled=True,
+            consolidation_min_facts=2,
+            consolidation_max_groups_per_cycle=3,
+            consolidation_max_sources=8,
+            staleness_age_days=90,
+            staleness_max_lifetime_multiplier=20.0,
+        )
+        # Both sources 3 days old. Volatile source (evd=7): deadline in 4 days
+        # (3 + 7 = 10 days from createdAt, minus 3 already elapsed = 7? No: the
```
willem-bd: This working is confusing to read — "deadline in 4 days (3 + 7 = 10 ... minus 3 already elapsed = 7? No:" leaves the abandoned calculation inline. The value (7) is correct, but the whole point of the "earliest deadline relative to merged createdAt" design is that elapsed-since-creation doesn't factor in: `days_until = (createdAt + 7) - createdAt = 7` regardless of age. Suggest collapsing to a single clear line so future readers aren't misled into thinking elapsed days matter.

backend/tests/test_memory_consolidation.py
```diff
@@ -895,6 +902,478 @@ def test_consolidated_created_at_tracks_newest_source(self):
...
+        # earliest deadline (10 + 3650) - merged createdAt (10 days ago) = 3650, clamped to 270
+        assert merged[0]["expected_valid_days"] == 270
+
+    def test_consolidated_evd_all_legacy_sources_uses_global_fallback_deadline(self):
+        """When no source carries expected_valid_days, every source's effective
+        lifetime is the global staleness_age_days (matching the read-time
+        fallback). The merged fact's deadline is derived from that fallback, not
+        omitted - so a merge of aged legacy facts still re-enters review rather
+        than silently inheriting nothing."""
+        from deerflow.agents.memory.backends.deermem.deermem.core.updater import _select_stale_candidates
```
willem-bd: This import is duplicated inline inside three test methods (here, and at lines 1105 and 1333). The staleness test module imports `_select_stale_candidates` at the top; if this module doesn't, hoist it to module level so the dependency is declared once rather than re-imported per test.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -405,10 +429,8 @@ def _effective_fact_staleness_age(fact: dict[str, Any], config: Any) -> int:
     ``staleness_age_days`` for facts that pre-date this feature or where the
     LLM did not provide an estimate.
     """
-    raw = fact.get("expected_valid_days")
-    if isinstance(raw, (int, float)) and not isinstance(raw, bool) and raw > 0:
-        return int(raw)
-    return config.staleness_age_days
+    evd = _read_expected_valid_days(fact)
```
willem-bd, line 454: Subtle behavior change worth surfacing in the body: switching `_effective_fact_staleness_age` to the shared helper changes its handling of a float `expected_valid_days` in `(0, 1)` (e.g. a hand-edited `0.5`). The old code checked `raw > 0` first and returned `int(0.5)` = `0` (immediately stale); the helper coerces first, so `0.5` -> `0` -> rejected -> falls back to the global age. It's a correct fix (a 0-day lifetime was never intended) and is covered by `test_falls_back_for_fractional_below_one`, but the body frames the extraction as purely DRY and doesn't list it under "Default behavior change."

sontianye:
@willem-bd thanks - all four points addressed in `8ea4a5ac` + a body rewrite.

**PR body reconciled.** The body now describes the shipped design (earliest source review deadline, always-set field) instead of the earlier "max source window / omit-when-no-evd / strictly additive" iteration. I kept an explicit note that the "cannot regress facts that predate #4143" property of that earlier iteration no longer holds, since a legacy source's 90-day fallback now participates and can drag the merged window sooner - which is the intended fix, but worth saying plainly. The (0, 1) float behavior change (0.5 -> fallback to global age instead of immediately-stale 0) is listed under Default behavior change too, attributed to routing `_effective_fact_staleness_age` through the shared helper that coerces before the positivity guard.

**Test docstring tidied.** `test_consolidated_evd_volatile_source_with_equal_created_at_future_deadline` no longer carries the abandoned calculation - collapsed to one line stating the window is relative to the merged `createdAt` regardless of the source's current age.

**`_select_stale_candidates` import hoisted** to the module-level import block; the three inline re-imports are removed.

On the `creation_cap == 0` edge: left as-is for now since it's pre-existing in the newFacts path, benign (the helper treats 0 as absent at read time), and only reachable with `staleness_age_days * staleness_max_lifetime_multiplier < 1` which the validated config ranges forbid (`staleness_age_days ≥ 30`, multiplier ≥ 1.0). Happy to add a `max(..., 1)` floor at both sites in a follow-up if you'd rather have it belt-and-braces, but didn't want to expand scope here.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -393,6 +391,32 @@ def _parse_fact_datetime(raw: str) -> datetime | None:
...
+    ``OverflowError`` - because Python's JSON decoder accepts them as floats by
+    default and they can appear in a hand-edited ``memory.json``; rejecting them
+    keeps a single malformed field from aborting staleness selection or
+    consolidation.  Hand-edited files and pre-feature facts may also carry
+    floats or lack the field entirely; returning ``None`` lets callers fall back
+    to the global age or omit the field rather than silently writing a
+    zero/negative/non-finite lifetime.
+    """
+    raw = fact.get("expected_valid_days")
+    if isinstance(raw, (int, float)) and not isinstance(raw, bool) and math.isfinite(float(raw)):
```
willem-bd: ⚠️ This `float(raw)` call raises `OverflowError` when `raw` is a huge `int`, which slips through the robustness goal this helper was added to enforce.

`math.isfinite(float(raw))` is evaluated left-to-right, so for a huge int the `float(raw)` coercion raises **before** `math.isfinite` is ever called:

```python
>>> _read_expected_valid_days({"expected_valid_days": 10**400})
OverflowError: int too large to convert to float
```

Python's JSON decoder parses an integer literal with no decimal point as an arbitrary-precision `int` (unlike `1e400`, which decodes to `inf` and is correctly rejected here). So a hand-edited `memory.json` carrying `"expected_valid_days": 1000000000000` - exactly the malformed-field scenario the docstring calls out ("keeps a single malformed field from aborting staleness selection or consolidation") - aborts the entire consolidation batch. I confirmed this end-to-end against this branch's `updater.py`: a source with `expected_valid_days: 10**400` raises `OverflowError` and aborts the merge (and would likewise abort `_effective_fact_staleness_age` and the newFacts cap, since all four call sites route through this helper).

The float cases (`NaN` / `±inf` / `1e400->inf`) are handled correctly; only the int case is missed.

Suggested fix - branch on type so an `int` never passes through `float()`:

```python
raw = fact.get("expected_valid_days")
if isinstance(raw, bool):
    return None
if isinstance(raw, int):
    evd = raw
elif isinstance(raw, float) and math.isfinite(raw):
    evd = int(raw)
else:
    return None
return evd if evd > 0 else None
```

Severity is low (the input is pathological - no LLM would emit a 10^400-day lifetime), but it's the same class of hand-edited-file robustness this PR targets, and the fix is trivial. At minimum the docstring's "keeps a single malformed field from aborting…" claim should be narrowed if this isn't addressed.

willem-bd (commented):
## Review: consolidated facts inherit `expected_valid_days` from sources

**Verdict:** The core fix is correct and well-tested. The seam between consolidation (#3996/#4122) and per-fact `expected_valid_days` (#4143) is real - a merged fact previously inherited the newest source's `createdAt` but no `expected_valid_days`, collapsing to the 90-day global fallback and re-entering staleness review immediately. The earliest-source-deadline inheritance (with overdue clamp + creation cap) is the right semantic, and the reversal from an earlier `max(source_evds)` is well-justified.

**Verification (against head `8ea4a5ac`):**
- ✅ All 132 tests in `test_memory_consolidation.py` + `test_memory_staleness_review.py` pass on the branch.
- ✅ All 12 new evd tests **fail on `origin/main`** (`KeyError: 'expected_valid_days'`) and pass on the branch - confirms the bug and the fix.
- ✅ Traced the inheritance math: the merged staleness deadline (`createdAt + inherited_evd`) equals the earliest source deadline in the normal case and re-reviews next cycle when a source is overdue. Imports/scope all check out; the DRY refactor is complete (no leftover inline evd checks).

**Actionable finding (inline below):** `_read_expected_valid_days` raises `OverflowError` on a huge-int `expected_valid_days` - `float(raw)` overflows before the `math.isfinite` guard. Same class of hand-edited-file robustness this PR targets; trivial fix.

**Non-blocking observations:**
- The newFacts cap now accepts a **float** `expected_valid_days` (previously `isinstance(evd, int)` only, so an LLM-emitted `365.0` was silently dropped). This is more correct, but it's a behavior broadening worth a line in the PR body's "Default behavior change" section.
- Under degenerate config (`staleness_age_days=0` or multiplier `0`), `creation_cap` can be `0` and a merged fact lands with `expected_valid_days: 0` (reads back as `None` -> global fallback). Pre-existing pattern in the newFacts path, not a regression - flagging only.

Nice work on the test stability (`_days_ago` relative to `now()` with the rationale comment) and the end-to-end regression repro.

sontianye:
@willem-bd good catch - fixed in `96ab14ed`, with one extension beyond your suggested fix.

**The `float(raw)` OverflowError on huge int.** Confirmed: `10**400` raised in `float(raw)` before `math.isfinite` ran, because Python's JSON decoder keeps an integer literal with no decimal point as an arbitrary-precision `int`. Switched to your type-branch shape so an `int` never passes through `float()`:
```python
raw = fact.get("expected_valid_days")
if isinstance(raw, bool):
    return None
if isinstance(raw, int):
    evd = raw
elif isinstance(raw, float) and math.isfinite(raw):
    evd = int(raw)
else:
    return None
if 0 < evd <= timedelta.max.days:
    return evd
return None
```

**One extension.** Your suggested fix closes the `float()` crash, but a huge int below the float limit yet above `timedelta.max.days` (999999999) - e.g. `10**12` - would pass the helper and raise `OverflowError` downstream in `timedelta(days=evd)` during `_select_stale_candidates` / consolidation. That's the same crash fancyboi999 flagged for `extend_by_days`, just reached via a stored evd. So I also cap the returned int at `timedelta.max.days`; both branches share the `0 < evd <= timedelta.max.days` check. Normal values are unaffected (the config ceiling `staleness_max_extension_days` tops out at 36500).

Tests at all three layers: helper (`test_rejects_huge_int_above_timedelta_max`, incl. the `timedelta.max.days` boundary itself accepted), read path (`test_falls_back_for_huge_int_above_timedelta_max`), and end-to-end consolidation (`test_consolidation_with_huge_int_source_evd_does_not_raise[1e400|1e12|1e9]`).

On your two non-blocking notes: the float-evd broadening on the newFacts cap is already covered by the body's "Default behavior change" (the `(0,1)` float entry attributes it to routing through the shared helper); the degenerate-config `creation_cap == 0` is the pre-existing pattern we discussed - still left as-is for the same reasons.

fancyboi999 (changes requested):
Thanks @sontianye. The huge-integer fix closes the `float(raw)` failure, but its accepted upper boundary still overflows the new consolidation deadline calculation.


Reviewed head SHA: `96ab14ed14b9c3aa1dae5aa8b34970797c4a4bd8`

[P2] Bound datetime arithmetic, not only `timedelta` construction

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:417-429` and `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:1333`
- Problem: `_read_expected_valid_days()` explicitly accepts `timedelta.max.days`, but that only proves `timedelta(days=evd)` can be constructed. Adding that duration to an ordinary fact timestamp exceeds `datetime.max`, so one accepted persisted value still aborts the consolidation path this PR adds.
- Evidence: on this exact head, I drove `_apply_updates()` with two consolidation sources, one carrying `expected_valid_days=timedelta.max.days`. It raised `OverflowError: date value out of range` at `dt + timedelta(...)`. The new helper test asserts this boundary is valid, while the added end-to-end tests cover only `10**9` and larger values that the helper rejects. Current `origin/main` has no source-deadline calculation in consolidation, so this failure belongs to the new path rather than the pre-existing staleness read.
- Suggested fix: reject or safely fall back for values that cannot participate in the actual `datetime` arithmetic, or guard the source-deadline calculation and use the configured global lifetime for an unrepresentable deadline. The supported configuration ceiling is far below either boundary.
- Test: change the boundary expectation and drive consolidation with `timedelta.max.days` itself, asserting the merge completes with the global fallback; keep `timedelta.max.days + 1` and the arbitrary-precision integer cases.

sontianye:
@fancyboi999 you're right, and the fix is in `08823c51`. Reproduced your case exactly: a source with `expected_valid_days=timedelta.max.days` raised `OverflowError: date value out of range` at `dt + timedelta(...)` in the consolidation deadline calc - capping at `timedelta.max.days` only proved the `timedelta` could be constructed, not that adding it to a real timestamp stayed in range.

That cap was the third patch in a row chasing the next overflow boundary (`int(nan)` -> `isfinite`; `float(10**400)` -> type-branch; `timedelta(10**12)` -> `timedelta.max.days`; then `dt + timedelta` overflow). Each one just moved the crash one step downstream. This change moves the guard to where the magnitude actually matters and uses try/except as the terminal boundary, so there is no further concrete upper limit to be wrong about.

**Design change:** the helper's job is now type/positivity validation only - it returns any positive int (huge ints included, not routed through `float()`). Datetime-range validation belongs at the arithmetic site, because whether a given evd overflows depends on the `datetime` it's added to, not on the value alone.

```python
def _safe_add_days(dt, days):
    try:
        return dt + timedelta(days=days)
    except (OverflowError, ValueError):
        return None
```

- `_select_stale_candidates` uses `_safe_add_days(now, -effective_age)`; a `None` means the window is unrepresentably large, so the fact cannot yet be stale and is skipped (not selected).
- Consolidation computes each source's deadline via `_safe_add_days`; a source whose deadline overflows falls back to the global `staleness_age_days` deadline (same treatment as a legacy no-evd source), so one malformed field cannot abort the merge.

Tests:
- `TestSafeAddDays` - normal/negative shifts pass; `10**400`/`10**12`/`10**9` return `None`; and `timedelta.max.days` itself (the exact value you reproduced) returns `None` rather than raising.
- `TestSelectStaleCandidates.test_huge_evd_does_not_abort_selection` - a fact with a huge evd is skipped, not selected, and selection does not raise.
- `test_consolidation_with_huge_int_source_evd_does_not_raise` now parametrized over `[1e400, 1e12, 1e9, timedelta.max.days]` - the last being the value that constructs a valid `timedelta` but overflows datetime arithmetic.

Normal values are unaffected (the config ceiling `staleness_max_extension_days` tops out at 36500, far below any overflow boundary).

WillemJiang (approved):

PR #4230, opened 2026-07-16
fix: harden Postgres async engine with pool_recycle and command_timeout to stop stale-connection 504s

### LINKED ISSUE #3882: [bug] nginx频繁报超时504错误
### Before you start

### Problem summary

worker=1的情况下，频繁触发nginx超时，而gateway服务并没用报错

### Affected area(s)

### What happened?

1. 使用make up部署。
2. 使用postgre数据库存储用户等数据。
3. 然后就有很大的频率触发504超时错误，主要的接口是：/api/v1/auth/register，/api/v1/auth/login/local，GET /api/agents/my-agent-name

不知道什么情况下触发的，无法定位错误原因。

### Expected behavior

不应该超时。

### Steps to reproduce

1. 正常配置
2. 使用postgre数据库存储用户数据等。
3. make up

### Relevant logs

```shell
## nginx log
172.18.0.1 - - [29/Jun/2026:10:30:04 +0000] "POST /api/langgraph/threads/2816129c-6bf0-48f1-a113-a3eb70621d2c/runs/stream HTTP/1.1" 200 1077437 "http://localhost:2027/workspace/chats/2816129c-6bf0-48f1-a113-a3eb70621d2c" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code/1.124.2 Chrome/148.0.7778.97 Electron/42.2.0 Safari/537.36"
172.18.0.1 - - [29/Jun/2026:10:30:04 +0000] "POST /api/langgraph/threads/2816129c-6bf0-48f1-a113-a3eb70621d2c/history HTTP/1.1" 200 27750 "http://localhost:2027/workspace/chats/2816129c-6bf0-48f1-a113-a3eb70621d2c" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code/1.124.2 Chrome/148.0.7778.97 Electron/42.2.0 Safari/537.36"
172.18.0.1 - - [29/Jun/2026:10:30:04 +0000] "GET /api/threads/2816129c-6bf0-48f1-a113-a3eb70621d2c/token-usage HTTP/1.1" 200 261 "http://localhost:2027/workspace/chats/2816129c-6bf0-48f1-a113-a3eb70621d2c" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code/1.124.2 Chrome/148.0.7778.97 Electron/42.2.0 Safari/537.36"
172.18.0.1 - - [29/Jun/2026:10:30:04 +0000] "POST /api/langgraph/threads/search HTTP/1.1" 200 490 "http://localhost:2027/workspace/chats/2816129c-6bf0-48f1-a113-a3eb70621d2c" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code/1.124.2 Chrome/148.0.7778.97 Electron/42.2.0 Safari/537.36"
172.18.0.1 - - [29/Jun/2026:10:30:06 +0000] "POST /api/threads/2816129c-6bf0-48f1-a113-a3eb70621d2c/suggestions HTTP/1.1" 200 149"http://localhost:2027/workspace/chats/2816129c-6bf0-48f1-a113-a3eb70621d2c" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code/1.124.2 Chrome/148.0.7778.97 Electron/42.2.0 Safari/537.36"
2026/06/30 02:02:20 [error] 8#8: *71 upstream timed out (110: Operation timed out) while reading response header from upstream, client: 172.18.0.1, server: _, request: "GET /api/skills HTTP/1.1", upstream: "http://172.18.0.3:8001/api/skills", host: "localhost:2027", referrer: "http://localhost:2027/workspace/chats/2816129c-6bf0-48f1-a113-a3eb70621d2c"
172.18.0.1 - - [30/Jun/2026:02:02:20 +0000] "GET /api/skills HTTP/1.1" 504 569 "http://localhost:2027/workspace/chats/2816129c-6bf0-48f1-a113-a3eb70621d2c" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code/1.124.2 Chrome/148.0.7778.97 Electron/42.2.0 Safari/537.36"
2026/06/30 02:52:31 [error] 8#8: *74 upstream timed out (110: Operation timed out) while reading response header from upstream, client: 192.168.0.17, server: _, request: "POST /api/v1/auth/login/local HTTP/1.1", upstream: "http://172.18.0.3:8001/api/v1/auth/login/local", host: "10.7.0.204:2026"
192.168.0.17 - - [30/Jun/2026:02:52:31 +0000] "POST /api/v1/auth/login/local HTTP/1.1" 504 167 "-" "okhttp/3.14.9"
192.168.0.17 - - [30/Jun/2026:02:52:31 +0000] "POST /api/v1/auth/login/local HTTP/1.1" 200 41 "-" "okhttp/3.14.9"
192.168.0.17 - - [30/Jun/2026:02:52:31 +0000] "GET /api/agents/czgj-parent-agent-333 HTTP/1.1" 200 4129 "-" "okhttp/3.14.9"
2026/06/30 02:53:31 [error] 8#8: *74 upstream timed out (110: Operation timed out) while reading response header from upstream, client: 192.168.0.17, server: _, request: "POST /api/v1/auth/register HTTP/1.1", upstream: "http://172.18.0.3:8001/api/v1/auth/register", host: "10.7.0.204:2026"


## gateway log

[2026-06-30 10:52:31 CST] [TraceId:LppzxyI0] [uvicorn.access] [INFO] 172.18.0.4:58120 - "POST /api/v1/auth/login/local HTTP/1.1" 200
[2026-06-30 10:52:31 CST] [TraceId:4XIYsUP9] [uvicorn.access] [INFO] 172.18.0.4:58134 - "GET /api/agents/czgj-parent-agent-333 HTTP/1.1" 200
[2026-06-30 10:53:31 CST] [TraceId:hZDyppuK] [uvicorn.access] [INFO] 172.18.0.4:36988 - "POST /api/v1/auth/login/local HTTP/1.1" 200
[2026-06-30 10:53:31 CST] [TraceId:qiaNDJ8o] [uvicorn.access] [INFO] 172.18.0.4:36990 - "POST /api/agents HTTP/1.1" 409
[2026-06-30 10:53:31 CST] [TraceId:IpSbugPG] [uvicorn.access] [INFO] 172.18.0.4:36998 - "POST /api/v1/auth/login/local HTTP/1.1" 200
[2026-06-30 10:53:31 CST] [TraceId:dIlCrLXt] [app.gateway.routers.threads] [INFO] Thread created: ecd19343-ab86-402b-a0c2-5b9081123521
[2026-06-30 10:53:31 CST] [TraceId:dIlCrLXt] [uvicorn.access] [INFO] 172.18.0.4:37010 - "POST /api/threads HTTP/1.1" 200
[2026-06-30 10:53:31 CST] [TraceId:Yo5o21_3] [deerflow.runtime.runs.manager] [INFO] Run created: run_id=daef066f-49b7-4ced-bd00-2baeb7070e5f thread_id=ecd19343-ab86-402b-a0c2-5b90811c6103
[2026-06-30 10:53:31 CST] [TraceId:Yo5o21_3] [deerflow.runtime.runs.worker] [INFO] Run daef066f-49b7-4ced-bd00-2baeb7070e5f: 'events' stream_mode not supported in gateway (requires astream_events + checkpoint callbacks). Skipping.
[2026-06-30 10:53:31 CST] [TraceId:Yo5o21_3] [uvicorn.access] [INFO] 172.18.0.4:37012 - "POST /api/threads/ecd19343-ab86-402b-a0c2-5b90811c6103/runs/stream HTTP/1.1" 200
[2026-06-30 10:53:31 CST] [TraceId:Yo5o21_3] [deerflow.runtime.runs.manager] [INFO] Run daef066f-49b7-4ced-bd00-2baeb7070e5f -> running
[2026-06-30 10:53:31 CST] [TraceId:Yo5o21_3] [deerflow.agents.lead_agent.agent] [INFO] Create Agent(czgj-parent-agent-333) -> thinking_enabled: True, reasoning_effort: minimal, model_name: doubao-seed-1-0-lite, is_plan_mode: True, subagent_enabled: False, max_concurrent_subagents: 3
[2026-06-30 10:53:31 CST] [TraceId:Yo5o21_3] [deerflow.tools.tools] [INFO] Including view_image_tool for model 'doubao-seed-2-0-lite' (supports_vision=True)
[2026-06-30 10:53:31 CST] [TraceId:Yo5o21_3] [deerflow.tools.tools] [INFO] Total tools loaded: 8, built-in tools: 3, MCP tools: 0, ACP tools: 0
```

Other

### Operating system

Linux

### Platform details

_No response_

### Python version

_No response_

### Node.js version

_No response_

### pnpm version

_No response_

### uv version

_No response_

### Git state

_No response_

### Additional context

_No response_

mvanhorn:
Fixes #3882

## Why

The reported 504s came from a stale half-open Postgres connection hanging at SQLAlchemy pool checkout. The async engine sets `pool_pre_ping=True` but has no `pool_recycle` and no asyncpg `command_timeout`/keepalive, so a connection killed server-side can wedge a request until timeout.

## What changed

The Postgres async engine is hardened with `pool_recycle` and an asyncpg `command_timeout` via a shared helper applied at both `create_async_engine` call sites, so stale connections are recycled instead of hanging.

## Surface area

- [ ] **Frontend UI**
- [x] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**

## Screenshots / Recording

N/A - backend connection-pool hardening.

## Bug fix verification

- Test path that reproduces the bug: added a test asserting the engine is configured with `pool_recycle` and `command_timeout` at both call sites.
- If a red test wasn't cheap to write: a true live-repro needs a killed Postgres connection under load; the added test pins the engine configuration that prevents the wedge.

## Validation

Ran the backend engine/persistence tests covering the new configuration.

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Implemented the fix from a written plan (root cause pinned to the engine config); I reviewed the diff before submission.


fancyboi999 (changes requested):
Thanks @mvanhorn. I found one issue that should be addressed before this is ready.


Reviewed base SHA: `b3a0dac8ad191fbfa248c0a1cecc8f2deeaa1c7d`
Reviewed head SHA: `e7b19240c556eae5edb30e62beca964201457c1d`

### [P1] Keep the database timeout inside the proxy deadline

- Location: `backend/packages/harness/deerflow/persistence/engine.py:22`
- Problem: `POSTGRES_COMMAND_TIMEOUT_SECONDS` is fixed at 60 seconds, which is the same deadline nginx applies to ordinary `/api/*` routes. A stale checkout or pre-ping that reaches this timeout still loses the race to nginx, so clients continue to receive the 504 this PR is meant to prevent. The constant also imposes an unconfigurable 60-second cap on every app-ORM command, including legitimate long-running operations.
- Evidence: the Docker nginx locations for `/api/agents`, `/api/skills`, and the `/api/` catch-all do not override `proxy_read_timeout`, so nginx uses its 60-second default. The issue evidence records `rt=60.000` and `rt=60.034` 504s. The locked asyncpg 0.31.0 source defines `command_timeout` as the default timeout for operations on the connection, so setting it to the same 60-second boundary leaves no time for Gateway to turn the database timeout into an upstream response.
- Suggested fix: make the app-ORM command timeout a `database` setting and choose a default comfortably below the proxy deadline (for example 30 seconds), while keeping the recycle policy. Preserve an explicit opt-out or larger override for deployments that intentionally run longer database commands.
- Test: add a regression that stalls the asyncpg ping/command past the configured database deadline and asserts checkout/request handling terminates before nginx's 60-second boundary; also cover the config override so a legitimate longer timeout is not silently clamped.

mvanhorn:
Fixed in 3087049. The app-ORM command timeout is now a `database.command_timeout` setting defaulting to 30s (comfortably under nginx's 60s deadline), with `null` to disable it for intentionally long commands, and the recycle policy is unchanged. Added a regression that stalls the asyncpg command past the configured deadline and a case asserting a larger override isn't clamped.

backend/packages/harness/deerflow/config/database_config.py
```diff
@@ -63,6 +63,11 @@ class DatabaseConfig(BaseModel):
         default=5,
         description="Connection pool size for the app ORM engine (postgres only).",
     )
+    command_timeout: float | None = Field(
+        default=30,
```
willem-bd, line 72: Finding 1 — two sources of truth for "30".

This `30` duplicates `POSTGRES_COMMAND_TIMEOUT_SECONDS = 30` in `persistence/engine.py`. The two must stay in sync: `init_engine_from_config` passes `config.command_timeout` while direct `init_engine` callers default to the module constant, so bumping one without the other makes the two call paths diverge.

A one-line comment cross-referencing the constant is the pragmatic fix (deriving the default from the constant would create a config -> persistence import cycle). Not a blocker.

backend/packages/harness/deerflow/config/database_config.py
```diff
@@ -63,6 +63,11 @@ class DatabaseConfig(BaseModel):
...
+        gt=0,
```
willem-bd, line 73: Finding 2 — `gt=0` rejects `0`.

`gt=0` rejects `command_timeout: 0` with a validation error, so `null` is the only way to disable the timeout. That's fine and documented in the field description - flagging because a user who tries `0` to "turn it off" gets a validation error rather than a disable.

If `0` should mean "disabled", switch to `ge=0` and treat `0` as no-timeout in `_postgres_engine_kwargs`; otherwise the current behavior is acceptable.

backend/packages/harness/deerflow/persistence/engine.py
```diff
@@ -16,12 +16,33 @@
 
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, async_sessionmaker, create_async_engine
 
+# Recycle pooled Postgres connections before stale idle sockets can hang
+# pool_pre_ping, and keep the default asyncpg command deadline inside nginx's
+# 60-second proxy timeout (#3882).
+POSTGRES_POOL_RECYCLE_SECONDS = 300
+POSTGRES_COMMAND_TIMEOUT_SECONDS = 30
```
willem-bd, line 22: Finding 3 — worst-case latency is a slow success, not a 504.

`command_timeout` also bounds `pool_pre_ping`'s `SELECT 1`, so a request landing on a wedged half-open connection now fails the ping after ~30s, recycles, and retries - a slow success instead of a 504 (the right trade-off).

If the perceived latency of that first dead-connection request matters for the Gateway UX, a lower default (e.g. 5-10s) recovers faster while still clearing nginx's 60s deadline. Configurable either way; not a blocker.

backend/tests/test_persistence_engine_postgres_config.py
```diff
@@ -0,0 +1,222 @@
...
+def test_postgres_engine_kwargs_allow_command_timeout_opt_out() -> None:
+    config = DatabaseConfig(command_timeout=None)
+    kwargs = engine_mod._postgres_engine_kwargs(echo=False, pool_size=5, command_timeout=config.command_timeout)
+
+    assert config.command_timeout is None
+    assert kwargs["connect_args"] == {}
+
+
+@pytest.mark.asyncio
+async def test_configured_command_timeout_ends_stalled_command_before_proxy_deadline() -> None:
```
willem-bd: Finding 4 — this test simulates asyncpg's command_timeout, not the real driver.

This simulates asyncpg's `command_timeout` with `asyncio.timeout` on a fake `_StalledAsyncpgEngine`, so it validates the config -> `connect_args` plumbing and the conceptual timeout behavior, not real asyncpg driver behavior.

The PR description acknowledges a true repro needs a killed connection under load, so this is a reasonable pinned-config test - just calibrating expectations when reading it.

fancyboi999 (changes requested):
Thanks @mvanhorn. The configurability fixes my earlier finding, but I need to correct one premise from that review: DeerFlow's checked-in nginx timeout is 600 seconds, not 60 seconds.


Reviewed base SHA: `10890e10a83c85bff73a85456614082952774992`
Reviewed head SHA: `3087049362e7be1a6a4858f2413f636f158db084`

[P2] Do not couple the database timeout to a nonexistent 60-second DeerFlow proxy deadline

- Location: `backend/packages/harness/deerflow/persistence/engine.py:19-23`, `config.example.yaml:1649-1651`, and `backend/tests/test_persistence_engine_postgres_config.py:17-34`
- Problem: the new comments, example, and test present 60 seconds as DeerFlow's nginx deadline and make the 30-second command timeout look like a repository invariant. All four relevant timeout blocks in the checked-in Docker and local nginx configs are actually 600 seconds. Because asyncpg's `command_timeout` bounds every operation on the app-ORM connection, not only `pool_pre_ping`, operators following this guidance can abort legitimate PostgreSQL work much earlier than the bundled proxy requires.
- Evidence: current `origin/main` sets `proxy_connect_timeout`, `proxy_send_timeout`, and `proxy_read_timeout` to `600s` in both `docker/nginx/nginx.conf` and `docker/nginx/nginx.local.conf`, including the catch-all API locations. The 60-second observations in #3882 describe that deployment's effective outer boundary; the test's `NGINX_PROXY_TIMEOUT_SECONDS = 60` is an unrelated hard-coded value and cannot verify repository behavior.
- Suggested fix: document `database.command_timeout` as deployment-specific and tell operators to keep it below their actual outer ingress/reverse-proxy deadline. If 30 seconds remains the protective default, explicitly describe the all-command compatibility tradeoff and the `null` opt-out instead of attributing it to DeerFlow nginx.
- Test: remove the invented 60-second proxy relationship from the unit test and retain focused assertions for the default, override, opt-out, and `connect_args` plumbing. If a repository proxy invariant is intentionally tested, derive it from the checked-in 600-second config rather than another test constant.

mvanhorn:
Thanks for the correction — you're right the nginx timeout is 600s. Fixed in d558146: the DB command timeout now stands on its own (bounds stalled ORM queries independently), and I dropped the references to the nonexistent 60s proxy deadline in the code, config, and tests.

willem-bd (commented):
Two more non-blocking nits from my review pass. (Skipping the 'two sources of truth for 30' note -- I already flagged that on database_config.py:67.)

backend/packages/harness/deerflow/config/database_config.py
```diff
@@ -63,6 +63,11 @@ class DatabaseConfig(BaseModel):
...
+        description="Timeout in seconds for app ORM PostgreSQL commands. Set to null to disable the command timeout.",
```
willem-bd, line 74: Default behavior change: the `command_timeout=30` default caps the app ORM at 30s per command where it was previously unbounded. A deployment with a legitimately long query (e.g. a large run-events fetch, bulk write) that used to succeed will now raise `asyncio.TimeoutError` / `QueryCanceledError`. The PR's "Default behavior change" checkbox is unchecked and the field description doesn't flag the cap, so it's easy for an operator to miss.

Suggestion: check the box, note the cap in the description (e.g. "defaults to 30s; previously unbounded"), and add a CHANGELOG line. The field is already tunable (null / raise), so this is just signaling -- not a blocker.

config.example.yaml
```diff
@@ -1646,6 +1646,9 @@ skill_evolution:
 database:
   backend: sqlite
   sqlite_dir: .deer-flow/data
+  # App ORM PostgreSQL command timeout. Set to null to disable it or raise it
```
willem-bd, line 1667: `null` reopens the stale-connection 504 wedge this PR fixes. With `command_timeout` removed, `pool_pre_ping`'s `SELECT 1` on a half-open socket can hang indefinitely -- `pool_recycle=300` only lowers the *probability* of handing out a stale connection, it doesn't bound the hang itself. The comment frames `null` as "for intentionally long commands" but doesn't warn it disables the wedge protection.

Suggestion: add a one-line warning here, e.g. "# Note: null re-enables the stale-connection hang this guards against; prefer raising the value." Not a blocker.

fancyboi999 (changes requested):
Thanks @mvanhorn. The earlier timeout findings are fixed, but the current-main merge exposes one remaining issue that should be addressed before this is ready.


Reviewed base SHA: `0cd55067f39823e7f0950498b5197a8195512481`
Reviewed head SHA: `d5581460cea4d48543deb94b180942ed880d24cb`

[P2] Advance the config version relative to current main

- Location: `config.example.yaml:18,1651`; `deploy/helm/deer-flow/values.yaml:236,261`; `deploy/helm/deer-flow/README.md:127,140`
- Problem: this PR adds the user-facing `database.command_timeout` schema and changes the app ORM default from unbounded to 30 seconds, but current `main` is already `config_version: 27` and the clean merge remains at 27. An installation already upgraded to v27 is therefore reported as current, so `make config-upgrade` never exposes the new timeout key or its opt-out even though the new default takes effect.
- Evidence: I merged current `origin/main` (`0cd55067`) into this head in an isolated worktree; all three config mirrors remained at 27. Running the real `scripts/config-upgrade.sh` against a v27 Postgres config printed `config.yaml is already up to date (version 27)` and left `database.command_timeout` absent. The focused config/engine suites still passed, and `validate-chart` only checks mirror parity, so the existing green gates do not catch the missing relative version increment.
- Suggested fix: rebase or merge current main, bump `config.example.yaml` and both Helm mirrors to the next version, and keep `command_timeout` as a real defaulted key so the recursive upgrader adds it. Document that the 30-second default was previously unbounded and that `null` disables the stale-connection guard.
- Test: start from a v27 config without `database.command_timeout`, run the real upgrade path, and assert it advances the version and adds `command_timeout: 30` without changing existing database values; retain the Helm parity check.

WillemJiang:
@mvanhorn , thanks for your contribution. Please take a look at the latest review comments.

willem-bd (commented):
Posting four review findings as inline comments (full review posted in the review thread). Finding 1 is the only actionable-required item (config_version bump); 2–4 are suggestions. Note: Finding 1's numbers are updated from my written review — `config_version` has since progressed to `28` on main (this branch merged it in), so the bump is now `28 → 29`, not `26 → 27`.

config.example.yaml
```diff
@@ -1662,6 +1662,9 @@ skill_evolution:
...
+  # for intentionally long commands.
+  command_timeout: 30
```
willem-bd, line 1669: **Finding 1 — `config_version` should be bumped for this schema addition.**

Per `backend/AGENTS.md`: *"When changing the config schema, bump `config_version` in `config.example.yaml`."* Adding `database.command_timeout` is a new schema field, but this PR's diff doesn't touch `config_version` (currently `28` at line 18 of this file, after the recent main merge). An existing deployment already on `28` that picks up this change won't get a staleness warning, so they won't be prompted to run `make config-upgrade` to discover the new tunable. Bumping `28 → 29` here keeps the convention honest.

(The field has a default of `30`, so nothing breaks either way — this is purely about operator discoverability.)

backend/packages/harness/deerflow/persistence/engine.py
```diff
@@ -16,12 +16,32 @@
...
 def _json_serializer(obj: object) -> str:
     """JSON serializer with ensure_ascii=False for Chinese character support."""
     return json.dumps(obj, ensure_ascii=False)
 
 
+def _postgres_engine_kwargs(*, echo: bool, pool_size: int, command_timeout: float | None = POSTGRES_COMMAND_TIMEOUT_SECONDS) -> dict[str, object]:
+    """Build the shared SQLAlchemy engine options for PostgreSQL."""
+    connect_args = {}
+    if command_timeout is not None:
+        connect_args["command_timeout"] = command_timeout
```
willem-bd, line 40: **Finding 2 — consider adding asyncpg TCP keepalive args for parity with the checkpointer pool.**

The checkpointer's psycopg pool (`runtime/checkpointer/async_provider.py::_build_postgres_pool`) already sets `keepalives=1, keepalives_idle=60, keepalives_interval=10, keepalives_count=6`, which proactively detects half-open sockets. `pool_recycle=300` narrows the stale-connection window and `command_timeout=30` bounds the hang, but neither *detects* a half-open socket the way keepalives do — the `pool_pre_ping` SELECT itself can hang on a dead socket until `command_timeout` fires (up to 30s of latency before recovery).

asyncpg accepts the same keys via `connect_args`, e.g. merge in `keepalives=1, keepalives_idle=60, keepalives_interval=10, keepalives_count=6` alongside `command_timeout`. This would bring the app engine to parity with the already-hardened checkpointer and more directly address the *half-open* symptom described in #3882.

backend/packages/harness/deerflow/config/database_config.py
willem-bd, line 74: **Finding 3 — worth flagging in the description that this also bounds alembic migrations.**

`bootstrap_schema(_engine, ...)` runs migrations on this hardened app engine, so the 30s per-statement ceiling applies to DDL/data migrations during bootstrap, not just runtime ORM queries. The current migration chain is DDL-only and completes well under 30s, but a future data-backfill revision could trip the ceiling and fail startup with a confusing timeout.

Operators can set `null` to disable, but a one-line note here (e.g. *"raise or disable when running long backfill migrations"*) would save them the surprise. The `null` opt-out is the right escape hatch — this is just about surfacing it at the point of configuration.

backend/packages/harness/deerflow/persistence/engine.py
```diff
@@ -16,12 +16,32 @@
 
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, async_sessionmaker, create_async_engine
 
+# Recycle pooled Postgres connections before stale idle sockets can hang
+# pool_pre_ping. The command timeout bounds stalled ORM queries independently.
+POSTGRES_POOL_RECYCLE_SECONDS = 300
```
willem-bd, line 21: **Finding 4 — `pool_recycle` is hardcoded while `command_timeout` and `pool_size` are both configurable.**

Operators behind an aggressive DB-side idle timeout (e.g. RDS `idle_in_transaction_session_timeout`) or a cloud LB idle timeout shorter than 300s can't tune recycle without editing code. `pool_pre_ping` is a backstop, but exposing `pool_recycle` as a sibling `DatabaseConfig` field (default `300`, matching the pattern of `command_timeout`/`pool_size`) would cost little and let operators align recycle with their environment's idle cutoff.

mvanhorn:
Good finding. pool_recycle is configurable in 9ebcfb0, plumbed the same way as command_timeout and pool_size (config, example config, helm values), default still 300s.

WillemJiang (approved):

mvanhorn:
Thanks for taking it, @WillemJiang - pool_recycle plus command_timeout should keep the async Postgres engine from wedging.

PR #4234, opened 2026-07-16
fix(feishu): check response.success() on card/reaction SDK calls

DaoyuanLi2816:
## Problem

`_reply_card`, `_create_card`, `_update_card`, and `_add_reaction` in
`backend/app/channels/feishu.py` all call the `lark-oapi` SDK and only ever
use the response on the happy path — none of them check
`response.success()`.

`lark-oapi` signals a *business-level* failure (expired/invalid card,
permission error, rate limit, etc.) by returning a response object whose
`.success()` is `False` — the SDK call itself does not raise. So today, a
rejected card send/update looks identical to a successful one to every
caller.

This is not a hypothetical: this same file already gets this right elsewhere.
`_upload_image`, `_upload_file`, and `_receive_single_file` all check
`response.success()` before trusting the response and raise/log accordingly.
The card/reaction helpers next to them just never adopted that pattern.

The gap is most exposed on `_update_card`: Feishu is
`supports_streaming=True`, so a single conversation issues many
`_update_card` patches while the agent is producing output — each one is a
chance to silently drop an update the user never sees.

It also makes existing, currently-unreachable logic in
`_send_card_message` reachable for the first time. That method already
wraps `_update_card` in a `try/except` that:
- re-raises on a non-final message, so `send()`'s `_send_with_retry` retries, and
- on a final message, falls back to replying with a brand-new card instead of losing the final answer.

Both branches were dead code in practice, since `_update_card` had no way to
raise on a business failure — it just returned `None` either way.

## Fix

Add `response.success()` checks to all four methods, mirroring the
established pattern in this same file:

- `_reply_card` / `_create_card` / `_update_card`: raise `RuntimeError` with
  the response `code`/`msg`, mirroring `_upload_image`/`_upload_file`. This
  makes the retry-on-non-final and fallback-on-final logic in
  `_send_card_message` reachable.
- `_add_reaction`: log a warning with `code`/`msg`/`log_id` instead of
  raising, mirroring `_receive_single_file`. A failed reaction is
  fire-and-forget; raising here would otherwise trigger a wasteful full
  resend of the card content that already succeeded.

## Tests

Added `TestFeishuCardSuccessChecks` in `backend/tests/test_channels.py`:

- One test per method (`_reply_card`, `_create_card`, `_update_card`,
  `_add_reaction`) that mocks the underlying SDK client call to return a
  response object shaped like a real business failure (`code=99991400`,
  `success()` returns `False`, no exception raised by the call itself) and
  asserts the new behavior (raise, or warning log without raising).
- Two end-to-end tests driving `_send_card_message`/`send()` directly to
  confirm the previously-dead fallback-to-new-card (final message) and
  retry (non-final message) paths actually engage now that `_update_card`
  can raise.

Confirmed all 4 unit tests fail against the pre-fix code (`DID NOT RAISE`),
confirmed the existing Feishu suite (`test_channels.py`,
`test_feishu_parser.py`, `test_additional_channel_connections.py`, 297
tests total) passes unchanged after the fix — the existing SDK-response
mocks in this suite don't set `.success()` explicitly and remain truthy by
default (`MagicMock().success()`), so none of them exercised this path
before.

`ruff check` / `ruff format --check` clean on both changed files.

### DIFF
--- a/backend/app/channels/feishu.py
+++ b/backend/app/channels/feishu.py
@@ -517,7 +517,7 @@ class FeishuChannel(Channel):
         request = self._ReplyMessageRequest.builder().message_id(message_id).request_body(self._ReplyMessageRequestBody.builder().msg_type("interactive").content(content).build()).build()
         response = await asyncio.to_thread(self._api_client.im.v1.message.reply, request)
         if not response.success():
-            raise RuntimeError(f"Feishu card reply failed: code={response.code}, msg={response.msg}")
+            raise RuntimeError(f"Feishu card reply failed: code={response.code}, msg={response.msg}, log_id={response.get_log_id()}")
         response_data = getattr(response, "data", None)
         return getattr(response_data, "message_id", None)
 
@@ -530,7 +530,7 @@ class FeishuChannel(Channel):
         request = self._CreateMessageRequest.builder().receive_id_type("chat_id").request_body(self._CreateMessageRequestBody.builder().receive_id(chat_id).msg_type("interactive").content(content).build()).build()
         response = await asyncio.to_thread(self._api_client.im.v1.message.create, request)
         if not response.success():
-            raise RuntimeError(f"Feishu card creation failed: code={response.code}, msg={response.msg}")
+            raise RuntimeError(f"Feishu card creation failed: code={response.code}, msg={response.msg}, log_id={response.get_log_id()}")
 
     async def _update_card(self, message_id: str, text: str) -> None:
         """Patch an existing card message in place."""
@@ -541,7 +541,7 @@ class FeishuChannel(Channel):
         request = self._PatchMessageRequest.builder().message_id(message_id).request_body(self._PatchMessageRequestBody.builder().content(content).build()).build()
         response = await asyncio.to_thread(self._api_client.im.v1.message.patch, request)
         if not response.success():
-            raise RuntimeError(f"Feishu card update failed: code={response.code}, msg={response.msg}")
+            raise RuntimeError(f"Feishu card update failed: code={response.code}, msg={response.msg}, log_id={response.get_log_id()}")
 
     def _track_background_task(self, task: asyncio.Task, *, name: str, msg_id: str) -> None:
         """Keep a strong reference to fire-and-forget tasks and surface errors."""
--- a/backend/app/channels/feishu.py
+++ b/backend/app/channels/feishu.py
@@ -493,7 +493,17 @@ class FeishuChannel(Channel):
             return
         try:
             request = self._CreateMessageReactionRequest.builder().message_id(message_id).request_body(self._CreateMessageReactionRequestBody.builder().reaction_type(self._Emoji.builder().emoji_type(emoji_type).build()).build()).build()
-            await asyncio.to_thread(self._api_client.im.v1.message_reaction.create, request)
+            response = await asyncio.to_thread(self._api_client.im.v1.message_reaction.create, request)
+            if not response.success():
+                logger.warning(
+                    "[Feishu] reaction '%s' add failed for message %s: code=%s, msg=%s, log_id=%s",
+                    emoji_type,
+                    message_id,
+                    response.code,
+                    response.msg,
+                    response.get_log_id(),
+                )
+                return
             logger.info("[Feishu] reaction '%s' added to message %s", emoji_type, message_id)
         except Exception:
             logger.exception("[Feishu] failed to add reaction '%s' to message %s", emoji_type, message_id)
@@ -506,6 +516,8 @@ class FeishuChannel(Channel):
         content = self._build_card_content(text)
         request = self._ReplyMessageRequest.builder().message_id(message_id).request_body(self._ReplyMessageRequestBody.builder().msg_type("interactive").content(content).build()).build()
         response = await asyncio.to_thread(self._api_client.im.v1.message.reply, request)
+        if not response.success():
+            raise RuntimeError(f"Feishu card reply failed: code={response.code}, msg={response.msg}")
         response_data = getattr(response, "data", None)
         return getattr(response_data, "message_id", None)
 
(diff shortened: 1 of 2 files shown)


backend/app/channels/feishu.py
```diff
@@ -504,6 +514,8 @@ async def _reply_card(self, message_id: str, text: str) -> str | None:
         content = self._build_card_content(text)
         request = self._ReplyMessageRequest.builder().message_id(message_id).request_body(self._ReplyMessageRequestBody.builder().msg_type("interactive").content(content).build()).build()
         response = await asyncio.to_thread(self._api_client.im.v1.message.reply, request)
+        if not response.success():
+            raise RuntimeError(f"Feishu card reply failed: code={response.code}, msg={response.msg}")
```
willem-bd: Optional nit (non-blocking): this matches the `_upload_image`/`_upload_file` format exactly, which is great for consistency. One small upgrade worth considering — include `response.get_log_id()` (e.g. `..., log_id={response.get_log_id()}`) so that when retries exhaust and the error propagates, the logs carry a Feishu support-traceable id. The upload siblings omit it, so keeping the current form is also fine. Applies the same way to `_create_card` and `_update_card`.

backend/app/channels/feishu.py
```diff
@@ -514,7 +526,9 @@ async def _create_card(self, chat_id: str, text: str) -> None:
 
         content = self._build_card_content(text)
         request = self._CreateMessageRequest.builder().receive_id_type("chat_id").request_body(self._CreateMessageRequestBody.builder().receive_id(chat_id).msg_type("interactive").content(content).build()).build()
-        await asyncio.to_thread(self._api_client.im.v1.message.create, request)
+        response = await asyncio.to_thread(self._api_client.im.v1.message.create, request)
+        if not response.success():
+            raise RuntimeError(f"Feishu card creation failed: code={response.code}, msg={response.msg}")
```
willem-bd: Follow-up scope (out of this PR): the sibling `send_file` (around lines 313-318) makes the same unchecked `message.reply` / `message.create` calls for file attachments and returns `True` on a silent business-level failure — the same class of bug this PR fixes for cards. Its upload deps (`_upload_image`/`_upload_file`) already raise, but the actual file-message send does not check `.success()`. A small follow-up PR giving `send_file` the same treatment would close the last gap.

backend/tests/test_channels.py
```diff
@@ -4924,6 +4924,223 @@ async def go():
         _run(go())
 
 
+class TestFeishuCardSuccessChecks:
```
willem-bd, line 5458: Minor coverage gap: the `_update_card` failure path has full e2e coverage (final-fallback + retry through `send()`), but `_create_card`'s failure on the no-`thread_ts` path (`feishu.py:663`, unwrapped) is only covered by the direct unit test. An e2e test mirroring `test_send_retries_after_update_card_business_failure_then_succeeds` for the no-source path would lock in that retry path too. Low priority since propagation is trivial.

willem-bd (commented):
@DaoyuanLi2816  Overall this is an approve — the raise/log split, the retry/fallback reachability, and the test coverage are all solid. The three inline notes below are optional/non-blocking follow-ups.

WillemJiang:
@DaoyuanLi2816, could you take a look at the latest review comments?

DaoyuanLi2816:
Thanks for the review, @willem-bd — and thanks for the nudge, @WillemJiang. Addressed all three inline points in 2468bf31 (pushed to this branch):

1. **log_id** ("Optional nit (non-blocking): ... One small upgrade worth considering — include `response.get_log_id()` ... Applies the same way to `_create_card` and `_update_card`."): done — `_reply_card`/`_create_card`/`_update_card` now include `log_id={response.get_log_id()}` in the raised `RuntimeError`, matching `_add_reaction`/`_receive_single_file`'s existing warning-log format in this file. Updated the three corresponding unit tests to assert the log_id surfaces in the error.

2. **`send_file` follow-up scope** ("Follow-up scope (out of this PR): the sibling `send_file` (around lines 313-318) makes the same unchecked `message.reply` / `message.create` calls ... A small follow-up PR giving `send_file` the same treatment would close the last gap."): agreed, kept out of this PR as you suggested — filed as a separate follow-up: #4335. Same `response.success()` check applied to `send_file`'s reply/create calls, with a discriminating test (verified it actually fails against the pre-fix code, not just green against the fix) covering both the `thread_ts` and no-`thread_ts` paths plus a success control case.

3. **`_create_card` no-`thread_ts` retry coverage** ("Minor coverage gap: ... An e2e test mirroring `test_send_retries_after_update_card_business_failure_then_succeeds` for the no-source path would lock in that retry path too."): done — added `test_send_retries_after_create_card_business_failure_then_succeeds`, mirroring the `_update_card` e2e test, confirming `_send_with_retry` engages through `send()` for the no-`thread_ts` path too. No source change was needed here since the retry path was already reachable through `send()`'s existing wrapper — this just closes the coverage gap you flagged.

All changes ruff format/lint clean and full `test_channels.py` suite green (260 passed) on the updated branch.

WillemJiang:
@DaoyuanLi2816, please resolve the conflicts with the main branch.

DaoyuanLi2816:
@WillemJiang, thanks for the ping — done. Rebased `fix/feishu-card-success-check` onto current `main` and force-pushed.

The conflict was entirely in `backend/tests/test_channels.py`: this PR's `TestFeishuCardSuccessChecks` class and #4335's `TestFeishuSendFileSuccessChecks` class (merged to main after this PR was opened) were both inserted at the same spot in the file — right after `TestFeishuChannel`, before `TestWeComChannel` — so git couldn't tell which insertion went where. Not a real logic conflict, just two independent additions landing in the same place; resolved by keeping both classes intact, back to back. `backend/app/channels/feishu.py` auto-merged with no manual changes needed.

Verified after rebase: full `test_channels.py` suite green (281 passed), `ruff format --check` and `ruff check` clean on both touched files. `gh pr view` now reports `mergeable: MERGEABLE`.

WillemJiang (approved):

PR #4237, opened 2026-07-16
fix(tui): make ctrl+u clear the whole composer, not just left of cursor

DaoyuanLi2816:
## Problem

`DeerFlowTUI.BINDINGS` maps `ctrl+u` to `clear_composer` (intended to wipe the
whole composer), but Textual's built-in `Input` widget already has its own
`ctrl+u -> delete_left_all` binding. A focused widget's own (non-priority)
bindings are checked before the App's non-priority bindings, so while the
composer has focus — the normal state while typing — `Input`'s binding always
won the race and `action_clear_composer` never actually fired. In practice,
Ctrl+U only deleted text left of the cursor instead of clearing the composer.

## Fix

Mark the binding `priority=True`, matching the pattern already used for
`ctrl+c`, `up`/`down`, `tab`, `escape`, and `enter` in the same `BINDINGS`
list. Priority bindings on the App are checked before the event is even
forwarded to the focused widget, so `clear_composer` now wins outright.

This interacts with the existing `check_action` modal-overlay gate. Textual's
own modal-safety truncation (`Screen._modal_binding_chain`) only applies to
the *non*-priority key-resolution pass — the priority pass walks the full
chain up to the App regardless of whether a modal screen (e.g. the `/model`
or `/threads` picker) is on top. So a bare `priority=True` would let Ctrl+U
reach past an open overlay and silently clear the (hidden) composer's text
while the user is interacting with the picker. `check_action` now includes
`clear_composer` in the same modal-gated set as the other priority bindings
(`nav_up`/`nav_down`/`palette_complete`/`escape`), so it's suppressed while
`len(self.screen_stack) > 1`, consistent with the existing "never steal keys
from a modal overlay or the composer" invariant.

## Tests

- `test_ctrl_u_clears_full_composer_not_just_left_of_cursor` (`test_tui_app.py`):
  types text, moves the cursor left of the end, presses Ctrl+U, and asserts
  the composer is fully empty rather than only missing the text left of the
  cursor.
- `test_ctrl_u_does_not_leak_into_composer_while_modal_open` (`test_tui_overlays.py`):
  opens the model picker, presses Ctrl+U, and asserts the composer text and
  the open overlay are both undisturbed.

Verified both halves of the fix are independently load-bearing: with only
`priority=True` and no `check_action` change, the first test passes but the
second fails (composer is wiped while the picker is open); with the full fix,
both pass. Reverting the whole change reproduces the original bug (only text
left of the cursor is removed). Full TUI suite (`tests/test_tui_*.py`, 151
tests) passes with no regressions.

`ruff check` / `ruff format --check` clean.

### DIFF
--- a/backend/packages/harness/deerflow/tui/app.py
+++ b/backend/packages/harness/deerflow/tui/app.py
@@ -144,7 +144,12 @@ class DeerFlowTUI(App):
     BINDINGS = [
         Binding("ctrl+c", "interrupt", "Interrupt / Quit", priority=True, show=True),
         Binding("ctrl+l", "redraw", "Redraw", show=False),
-        Binding("ctrl+u", "clear_composer", "Clear input", show=False),
+        # priority=True: Input's own BINDINGS already map ctrl+u to delete_left_all,
+        # which otherwise wins outright while the composer has focus (a focused
+        # widget's own bindings are checked before the App's non-priority ones).
+        # check_action gates this the same way as the other priority bindings below
+        # so it doesn't reach into the composer while a modal overlay is open.
+        Binding("ctrl+u", "clear_composer", "Clear input", show=False, priority=True),
         # Up/Down drive the palette when it's open, otherwise input history.
         # Tab/Enter/Esc only act when the palette is open. check_action gates all
         # of these so they never steal keys from a modal overlay or the composer.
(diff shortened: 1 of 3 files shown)


DaoyuanLi2816:
Closing this one on reflection — it's a minor keybinding nicety, not worth taking up review time for. Will focus on more substantive things going forward.

PR #4238, opened 2026-07-16
fix(harness): add timeout to invoke_acp_agent to prevent indefinite hangs

DaoyuanLi2816:
## Problem

`invoke_acp_agent` (`backend/packages/harness/deerflow/tools/builtins/invoke_acp_agent_tool.py`) has no timeout anywhere in its call path, and `ACPAgentConfig` (`backend/packages/harness/deerflow/config/acp_config.py`) has no timeout field either.

If the ACP agent subprocess answers `initialize`/`new_session` correctly but then hangs inside `prompt()` (e.g. it wedges on some internal step and never sends a `session/prompt` response), the `await conn.prompt(...)` call never returns. That blocks the tool call — and therefore the whole agent turn — indefinitely, with the child process left running.

This is the same class of gap MCP servers already guard against: `extensions_config.json -> mcpServers.<server>.tool_call_timeout` bounds individual stdio MCP tool calls (`backend/packages/harness/deerflow/mcp/tools.py`, via `read_timeout_seconds` passed to `session.call_tool`). ACP agent invocations had no equivalent.

## Fix

- Added `ACPAgentConfig.timeout_seconds` (default `1800`, `ge=1`), mirroring the shape/default of `subagents.timeout_seconds` (`backend/packages/harness/deerflow/config/subagents_config.py`) — both invoke a delegated, potentially long-running agent process, so both need a sane default backstop rather than an unbounded wait.
- Wrapped the `conn.prompt(...)` call in `asyncio.wait_for(..., timeout=agent_config.timeout_seconds)`. On `TimeoutError`, the tool logs the timeout and returns a clear, actionable error message instead of hanging.
- Exiting the `async with spawn_agent_process(...)` block (by returning, same as any other exit path) is what triggers the ACP library's own subprocess cleanup in `spawn_stdio_transport` (graceful stdin close, then terminate, then kill, each with a bounded grace period) — so the hung subprocess is actually terminated, not just abandoned.
- Documented the new field in `config.example.yaml` and bumped `config_version`.

## Testing

Added `test_invoke_acp_agent_times_out_and_kills_hung_subprocess` — spawns a real (not mocked) ACP agent subprocess that answers `initialize`/`new_session` correctly and then hangs forever on `prompt()`, drives it through the real, unmodified `invoke_acp_agent` tool with a short configured `timeout_seconds`, and asserts:
- the tool call returns a clear timeout error within the configured timeout instead of hanging;
- the underlying subprocess is actually terminated afterward (`proc.returncode is not None`), via a spy on `spawn_agent_process` that captures the real `asyncio.subprocess.Process` while leaving all spawn/handshake/cleanup behavior untouched.

The test wraps the call in a generous `asyncio.wait_for(..., timeout=20)` purely as a test-level safety net, so a regression fails the suite instead of hanging it. Verified this test actually discriminates: reverting just the source fix (keeping the test) and re-running it fails via that safety net after ~20s; reapplying the fix passes again in ~4s for the full file.

Also added config-level tests for the new field's default, override, and `ge=1` validation in `test_acp_config.py`.

`cd backend && make test` (targeted: full `test_invoke_acp_agent_tool.py` + `test_acp_config.py`, plus `test_app_config_reload.py`/`test_authorization_provider.py`/`test_lead_agent_prompt.py`/`test_mcp_routing_prompt.py`/`test_tool_deduplication.py` for config-schema regressions) all pass; `make lint` / `ruff format --check` clean.

### DIFF
--- a/backend/packages/harness/deerflow/config/acp_config.py
+++ b/backend/packages/harness/deerflow/config/acp_config.py
@@ -24,6 +24,17 @@ class ACPAgentConfig(BaseModel):
             "are denied — the agent must be configured to operate without requesting permissions."
         ),
     )
+    timeout_seconds: int = Field(
+        default=1800,
+        ge=1,
+        description=(
+            "Maximum time in seconds to wait for the agent to respond to a single invoke_acp_agent "
+            "call before the invocation is aborted and the subprocess is terminated. Mirrors "
+            "subagents.timeout_seconds (default: 1800 = 30 minutes) — without this backstop, an ACP "
+            "agent subprocess that hangs after initialize/new_session blocks the tool call, and "
+            "therefore the whole agent turn, indefinitely."
+        ),
+    )
 
 
 _acp_agents: dict[str, ACPAgentConfig] = {}
--- a/backend/packages/harness/deerflow/tools/builtins/invoke_acp_agent_tool.py
+++ b/backend/packages/harness/deerflow/tools/builtins/invoke_acp_agent_tool.py
@@ -1,5 +1,6 @@
 """Built-in tool for invoking external ACP-compatible agents."""
 
+import asyncio
 import logging
 import os
 import shutil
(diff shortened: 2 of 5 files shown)


WillemJiang:
@DaoyuanLi2816  please fix CI failed test.

DaoyuanLi2816:
@WillemJiang re: "please fix CI failed test" — dug into the `validate-chart` failure, here's what was actually going on and what I did.

**Actual failure** (`scripts/check_config_version.sh`, called from `validate-chart`):
```
config.example.yaml config_version=27
chart values.yaml     config_version=26
##[error]chart config_version (26) is behind config.example.yaml (27). Bump 'config_version' in deploy/helm/deer-flow/values.yaml (and the README example) to 27.
```
So yes, a `config_version` mismatch between `config.example.yaml` and the Helm chart's `deploy/helm/deer-flow/values.yaml`, confirmed straight from the log.

**Why it happened**: this branch had drifted ~67 commits behind `main`. In the meantime, several unrelated PRs bumped `config_version` further (up to 29) on both `config.example.yaml` and the chart in lockstep, so current `main` is actually internally consistent — this branch was just stuck on its old base with the older, now-mismatched numbers, and had a real conflict with `main` on that exact line (confirmed via a local merge test, and `mergeable: CONFLICTING` on the PR at the time).

**Fix**: rebased onto current `main` (one conflict, on the `config_version` line, resolved by keeping main's value since it already supersedes this branch's bump) and force-pushed. `config.example.yaml`, the chart's `values.yaml`, and the chart README are now all at `config_version: 29`.

**Verification**: no new unit test for this (it's a version-sync fix, not a behavior change), but: `scripts/check_config_version.sh` passes locally now, both YAML files parse cleanly, and the full existing suites for `test_config_version.py` / `test_acp_config.py` / `test_invoke_acp_agent_tool.py` / `test_app_config_reload.py` (65 tests) still pass.

One flag: the workflow runs on the new commit are currently sitting at `action_required` — could you approve them when you get a chance so `validate-chart` actually executes and goes green?

WillemJiang (approved):

PR #4239, opened 2026-07-16
fix(browserless): surface target-page error status in web_fetch_tool

DaoyuanLi2816:
## Problem

Browserless returns HTTP 200 for the render request itself even when the
target page responded with a 4xx/5xx, or served an anti-bot block page.
The real target status is exposed via the `X-Response-Code` /
`X-Response-Status` response headers.

`capture_screenshot` (and `web_capture_tool` on top of it) already handles
this correctly: the headers are threaded into `BrowserlessScreenshotResult`,
and `_target_status_warning()` turns a non-2xx/3xx target status into a
visible `(warning: target page responded ...)` suffix on the tool result.

`fetch_html` does not follow the same pattern. It reads the same two
headers, but only `logger.debug`s them and then returns the raw
`resp.text` unconditionally whenever the Browserless render request itself
was HTTP 200 — regardless of what the target page actually returned. As a
result, `web_fetch_tool` silently hands the model a 404 page or an
anti-bot block page as if it were a normal, successful fetch, with no
warning or error indication at all.

## Fix

Apply the same pattern `capture_screenshot`/`web_capture_tool` already
establish, to `fetch_html`/`web_fetch_tool`:

- `fetch_html` now returns a new `BrowserlessFetchResult` frozen dataclass
  (`html`, `target_status_code`, `target_status`) on success instead of a
  bare `str`, mirroring `BrowserlessScreenshotResult`. Error paths
  (non-200 render response, empty response, timeout, request failure) are
  unchanged and still return a plain `"Error: ..."` string.
- `_target_status_warning()` is widened to accept either result type
  (`BrowserlessScreenshotResult | BrowserlessFetchResult`), since both
  expose the same `target_status_code` / `target_status` fields.
- `web_fetch_tool` checks `isinstance(result, str)` for the error case
  (same shape as `web_capture_tool`), then appends
  `_target_status_warning(result)` to the extracted markdown when the
  target page's real status is an error/block.

A legitimate 200-target-page fetch is unaffected: no warning is appended,
and the extracted content is unchanged.

## Testing

Added/updated coverage in `backend/tests/test_browserless_client.py`:

- `test_fetch_html_success` / `test_fetch_html_with_*` updated for the new
  `BrowserlessFetchResult` return shape.
- `test_fetch_html_surfaces_target_status_headers` — `fetch_html` carries
  the target-page status headers on its result.
- `test_web_fetch_tool_warns_on_target_error_status` — `web_fetch_tool`
  now surfaces the warning for a 404 target page.
- `test_web_fetch_tool_no_warning_for_normal_target_status` — regression
  guard: a legitimate 200-target fetch is unaffected.
- `test_web_fetch_and_web_capture_tools_agree_on_target_error_warning` —
  drives both `web_fetch_tool` and `web_capture_tool` with identical
  target-error headers and asserts they now produce the same warning.

Full `backend/tests/test_browserless_client.py` suite: 30 passed.

### Before / after

With the source change reverted (tests unchanged), the suite fails to even
collect, since `BrowserlessFetchResult` does not exist yet on the
unfixed `fetch_html`:

```
ImportError: cannot import name 'BrowserlessFetchResult' from
'deerflow.community.browserless.browserless_client'
```

With the fix restored, all 30 tests pass.

## Before submitting

- [x] Ran `ruff check` / `ruff format --check` on all changed files — clean.
- [x] Ran the full `backend/tests/test_browserless_client.py` suite — 30 passed.
- [x] Ran the full backend test suite before/after this change (excluding
      pre-existing, unrelated collection gaps in this environment) —
      identical failure/error signature on both, only delta is the new
      tests added here passing.

### DIFF
--- a/backend/packages/harness/deerflow/community/browserless/browserless_client.py
+++ b/backend/packages/harness/deerflow/community/browserless/browserless_client.py
@@ -47,8 +47,55 @@ class BrowserlessClient:
         wait_for_selector_timeout_ms: int = 5000,
         reject_resource_types: list[str] | None = None,
         reject_request_pattern: list[str] | None = None,
-    ) -> BrowserlessFetchResult | str:
+    ) -> str:
         """Fetch the rendered HTML of a page using Browserless.
+
+        Public string contract: this always returns either the rendered HTML or
+        an "Error: ..." string, never a richer result object. Callers that also
+        need the target page's real status (Browserless returns HTTP 200 for the
+        render request itself even when the target page responded with a 4xx/5xx
+        or an anti-bot block page) should use fetch_html_with_status() instead.
+
+        Args:
+            url: The URL to fetch.
+            wait_for_event: Wait for a page event (e.g. "networkidle", "load").
+            wait_for_timeout_ms: Extra wait after page load.
+            wait_for_selector: CSS selector to wait for.
+            wait_for_selector_timeout_ms: Timeout for selector wait.
+            reject_resource_types: Resource types to block (e.g. ["image"]).
+            reject_request_pattern: URL patterns to block.
+
+        Returns:
+            Rendered HTML content, or an "Error: ..." string on failure.
+        """
+        result = await self.fetch_html_with_status(
+            url=url,
+            wait_for_event=wait_for_event,
+            wait_for_timeout_ms=wait_for_timeout_ms,
+            wait_for_selector=wait_for_selector,
+            wait_for_selector_timeout_ms=wait_for_selector_timeout_ms,
+            reject_resource_types=reject_resource_types,
+            reject_request_pattern=reject_request_pattern,
+        )
+        return result.html if isinstance(result, BrowserlessFetchResult) else result
+
+    async def fetch_html_with_status(
+        self,
+        url: str,
+        wait_for_event: str = "",
+        wait_for_timeout_ms: int = 0,
+        wait_for_selector: str = "",
+        wait_for_selector_timeout_ms: int = 5000,
+        reject_resource_types: list[str] | None = None,
+        reject_request_pattern: list[str] | None = None,
+    ) -> BrowserlessFetchResult | str:
+        """Fetch the rendered HTML of a page using Browserless, with target status.
+
+        Same request/response handling as fetch_html(), except a successful fetch
+        returns a BrowserlessFetchResult carrying the target page's real status
+        headers instead of a bare string, so a caller can tell a genuine 200 apart
+        from a render-succeeded-but-target-errored (or anti-bot blocked) response.
+        Use fetch_html() instead when only the HTML/error string is needed.
 
         Only sends accepted parameters for the current Browserless API version.
         Sets a default navigation timeout (30s) via query param.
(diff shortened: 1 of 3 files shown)


fancyboi999 (changes requested):

[P2] Preserve the exported `fetch_html()` string contract

- Location: `backend/packages/harness/deerflow/community/browserless/browserless_client.py:50`
- Problem: successful `BrowserlessClient.fetch_html()` calls now return `BrowserlessFetchResult` instead of the rendered HTML string. `BrowserlessClient` is explicitly re-exported from `deerflow.community.browserless.__all__`, so harness consumers that call string methods, compare the result, or pass it directly to a parser will fail after upgrading even though their fetch succeeded.
- Evidence: current main declares `fetch_html(...) -> str`, its tests assert string results, and `docs/superpowers/plans/2026-06-29-community-browserless-web-capture.md` explicitly says to preserve existing `fetch_html()` behavior. The new result is not a `str` subclass and there is no compatibility path or opt-in flag.
- Suggested fix: keep `fetch_html()` returning the HTML/error string and add a separate status-aware method for `web_fetch_tool`, or otherwise make the richer result an explicit opt-in while retaining the existing default contract.
- Test: retain a direct public-client regression asserting a successful `fetch_html()` result is a `str`, alongside the new tool-level test proving target status still reaches the warning.


Reviewed head SHA: `b0ab7519c6437b4968fced94f16974ee9c028bd7`

DaoyuanLi2816:
Follow-up: `fetch_html()` was returning `BrowserlessFetchResult` instead of `str`, breaking the original string contract for any caller relying on the `__all__` re-export.

Fixed in d71421e: `fetch_html()` now keeps its original contract exactly - it always returns either the rendered HTML `str` or an `"Error: ..."` `str`, never `BrowserlessFetchResult`. The richer, status-aware result moves to a new `fetch_html_with_status()` method instead. `web_fetch_tool` now calls `fetch_html_with_status()`, so it still gets the target-page status headers and still surfaces the `(warning: target page responded ...)` text for a render-succeeded-but-target-errored page. Every other consumer of `BrowserlessClient.fetch_html()` - including anything outside this repo that imports it via the `__all__` re-export - keeps getting a plain `str` back, exactly like before this PR.

Test changes in `backend/tests/test_browserless_client.py`:
- `test_fetch_html_success` now asserts `isinstance(result, str)` instead of `BrowserlessFetchResult` (the direct public-client regression test for this).
- New `test_fetch_html_returns_plain_string_even_when_target_page_errored` reproduces exactly the case this bug lived in: a successful render (200) against a target page that itself returned 404. `fetch_html()` still returns the plain HTML string, not the result object. Both of these fail against the previous commit (`b0ab751`) and pass against the current one.
- `test_fetch_html_with_status_surfaces_target_status_headers` keeps the status-aware coverage, retargeted onto the new `fetch_html_with_status()` method.
- The existing `web_fetch_tool` tests (success/error/exception/private-address/warning cases) are retargeted to mock `fetch_html_with_status` instead of `fetch_html`, since that's what the tool calls now.

Full `backend/tests/test_browserless_client.py` suite: 31 passed. `ruff check` / `ruff format --check` clean on all three changed files. Also ran the full backend suite (7529 passed, 64 pre-existing failures unrelated to this change - Windows/POSIX-path/symlink/chmod environment gaps, none touching `browserless`, confirmed by grepping the failure list) to make sure nothing else regressed.

WillemJiang (approved):

PR #4240, opened 2026-07-16
fix(mcp): use threading.Lock for OAuth token refresh to avoid cross-thread deadlock

DaoyuanLi2816:
## Summary

`OAuthTokenManager` (`backend/packages/harness/deerflow/mcp/oauth.py`) creates one `asyncio.Lock()` per OAuth-protected MCP server, once, in `__init__`, and keeps it for the process lifetime. That lock is safe as long as every caller shares one event loop — but the embedded/TUI sync tool-call path does not:

```
DeerFlowClient.stream()                     # sync generator, calls agent.stream() (sync)
  -> LangGraph ToolNode._func                # sync entrypoint for the compiled graph
    -> get_executor_for_config(...)          # a ThreadPoolExecutor, one worker thread per concurrent tool call
      -> deerflow.tools.sync.make_sync_tool_wrapper's run_coroutine(...)
        -> asyncio.run(coro(...))            # a BRAND NEW event loop, per call, per thread
          -> OAuthTokenManager.get_authorization_header(server_name)
```

Every concurrent MCP tool call in a turn gets its own fresh event loop on its own OS thread (`asyncio.run()` inside `make_sync_tool_wrapper.run_coroutine`, used for every async-only tool including MCP tools — see `mcp/tools.py:707` / `tools/tools.py:41`). If two of those calls hit the same OAuth-protected server at the same time (including the very first cold-start token fetch, not just a later refresh), they contend on the *same* `asyncio.Lock` instance from *two different event loops*.

`asyncio.Lock` is documented as safe only within a single event loop. Concretely, on CPython (`asyncio/locks.py`):
- The fast, uncontended `acquire()` path never touches `self._loop`, so the first caller doesn't bind the lock to any loop.
- A second, contending caller does call `self._get_loop()`, which lazily binds the lock to *that* caller's loop and creates a waiter `Future` on it.
- When the first caller's `_fetch_token()` finishes and `release()` runs `_wake_up_first()`, it calls `waiter_future.set_result(...)` — but this executes on the *first* caller's thread, not the thread that owns the waiter's loop, and without `call_soon_threadsafe`. The owning loop's selector is never woken, so that caller's `asyncio.run()` never returns. **No exception, no log line — a silent, permanent hang** that freezes the entire agent turn (`stream()`'s generator just stops producing events).
- With a *third* concurrent caller, the failure mode flips: it calls `_get_loop()`, sees the lock is already bound to a different loop, and raises `RuntimeError: <Lock object at ...> is bound to a different event loop` synchronously.

I reproduced both outcomes directly against `OAuthTokenManager` (3 threads, each running its own `asyncio.run()`, all targeting the same server): one thread hung forever (had to be bounded by a join timeout), and a second thread raised the `RuntimeError` above — both against unmodified `oauth.py`.

This is specific to the embedded/TUI sync path. Gateway's `ToolNode._afunc` (async graph execution, single event loop, `await asyncio.gather(...)`) never crosses threads or loops and is unaffected.

## Fix

Replace the per-server `asyncio.Lock` with a plain `threading.Lock`, which has no event-loop affinity. To keep it awaitable without blocking the caller's event loop, it's acquired via `await asyncio.to_thread(lock.acquire)` (a real OS-level blocking wait runs on a worker thread; `asyncio.to_thread`/`run_in_executor` already handles the cross-thread-to-event-loop handoff correctly, via `call_soon_threadsafe`, under the hood) and released synchronously in a `finally: lock.release()` (release never blocks, so no need to offload it).

This preserves the original correctness the lock provided — only one concurrent caller per server actually performs the HTTP token fetch; the rest see the freshly cached token once they get the lock — while removing the cross-loop hazard entirely, regardless of how many event loops or threads call in.

## Tests

Added `test_get_authorization_header_concurrent_threads_no_deadlock` to `backend/tests/test_mcp_oauth.py`: three threads, each with its own event loop (`asyncio.run`), call `get_authorization_header` concurrently for the same (initially uncached) server. One thread is synchronized (via an event set from inside the mocked token-fetch call) to be holding the lock while the other two contend, so the race is deterministic rather than probabilistic. The test:
- joins all threads with a bounded timeout, then asserts none are still alive (a regression back to `asyncio.Lock` hangs at least one thread — this is what actually catches the bug, quickly, instead of hanging the suite)
- asserts none of the threads raised (catches the `RuntimeError` variant)
- asserts exactly one real token fetch happened (de-duplication is preserved)

Verified fail-before / pass-after directly: with the old `asyncio.Lock` implementation this exact test fails in ~6.6s with `AssertionError: deadlock: thread(s) still blocked after bounded timeout: ['waiter-1']`; with the fix it passes in ~1.5s together with the rest of the file.

```
cd backend
PYTHONPATH=packages/harness python -m pytest tests/test_mcp_oauth.py -v
# 6 passed
```

## Validation

```
cd backend
ruff check packages/harness/deerflow/mcp/oauth.py tests/test_mcp_oauth.py   # All checks passed!
ruff format --check packages/harness/deerflow/mcp/oauth.py tests/test_mcp_oauth.py   # 2 files already formatted
PYTHONPATH=packages/harness python -m pytest tests/test_mcp_oauth.py -v   # 6 passed
```

### DIFF
--- a/backend/packages/harness/deerflow/mcp/oauth.py
+++ b/backend/packages/harness/deerflow/mcp/oauth.py
@@ -70,7 +70,34 @@ class OAuthTokenManager:
         # keeps the de-duplication behavior of the old `async with lock:` (only one
         # concurrent caller per server actually fetches a token) while remaining
         # safe when callers are on different event loops/threads.
-        await asyncio.to_thread(lock.acquire)
+        #
+        # The acquisition itself runs as an explicit Task, shielded from this
+        # coroutine's own cancellation. A bare `await asyncio.to_thread(lock.acquire)`
+        # cannot be safely cancelled: once the executor thread has started running
+        # lock.acquire(), Python has no way to stop it, so a cancellation delivered
+        # at that await would still let the thread go on to acquire the lock later
+        # (whenever the current holder releases it) with this coroutine already
+        # gone and nobody left to call release() -- the lock would stay locked
+        # forever and every later call for this server would block permanently at
+        # this same line. Shielding the acquisition task means a cancelled caller
+        # can instead wait for that (unstoppable) acquisition to actually land and
+        # release the lock immediately, rather than leaking ownership of it.
+        acquire_task = asyncio.create_task(asyncio.to_thread(lock.acquire), name=f"oauth-lock-acquire:{server_name}")
+        try:
+            await asyncio.shield(acquire_task)
+        except asyncio.CancelledError:
+            # Keep waiting -- shielded on every retry -- until the acquisition
+            # actually finishes, even if this coroutine is cancelled again while
+            # cleaning up: the underlying thread cannot be interrupted, so this is
+            # the only way to learn when the lock becomes ours and release it
+            # right away instead of leaving it locked forever.
+            while not acquire_task.done():
+                try:
+                    await asyncio.shield(acquire_task)
+                except asyncio.CancelledError:
+                    continue
+            lock.release()
+            raise
         try:
             token = self._tokens.get(server_name)
             if token and not self._is_expiring(token, oauth):
--- a/backend/packages/harness/deerflow/mcp/oauth.py
+++ b/backend/packages/harness/deerflow/mcp/oauth.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import threading
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from typing import Any
@@ -28,7 +29,17 @@ class OAuthTokenManager:
     def __init__(self, oauth_by_server: dict[str, McpOAuthConfig]):
         self._oauth_by_server = oauth_by_server
         self._tokens: dict[str, _OAuthToken] = {}
-        self._locks: dict[str, asyncio.Lock] = {name: asyncio.Lock() for name in oauth_by_server}
+        # A plain threading.Lock, not asyncio.Lock: the embedded/TUI sync tool-call
+        # path (DeerFlowClient.stream() -> LangGraph ToolNode._func -> a
+        # ThreadPoolExecutor -> deerflow.tools.sync.make_sync_tool_wrapper's
+        # per-call asyncio.run()) invokes get_authorization_header from a fresh
+        # event loop on a fresh OS thread for every concurrent tool call. An
+        # asyncio.Lock binds to whichever loop first contends on it; a second
+        # caller's release/wake-up crossing loops without call_soon_threadsafe
+        # either deadlocks silently or raises "bound to a different event loop".
+        # threading.Lock has no loop affinity, so it is safe to share across
+        # however many event loops/threads call into the same server's lock.
+        self._locks: dict[str, threading.Lock] = {name: threading.Lock() for name in oauth_by_server}
 
     @classmethod
     def from_extensions_config(cls, extensions_config: ExtensionsConfig) -> OAuthTokenManager:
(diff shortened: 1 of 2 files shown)


fancyboi999 (changes requested):

[P1] Make the off-thread lock acquisition cancellation-safe

- Location: `backend/packages/harness/deerflow/mcp/oauth.py:73`
- Problem: cancelling a caller while it is waiting in `asyncio.to_thread(lock.acquire)` does not stop the executor thread. After the current holder releases, that abandoned thread can still acquire the `threading.Lock`; because cancellation exits at the `await` before execution enters the following `try/finally`, nothing releases it. Every later OAuth token refresh for that server can then block permanently.
- Evidence: the new `try` starts only after the acquisition await. A bounded standard-library reproduction of this exact sequence (hold lock -> start waiter -> cancel waiter -> release holder) leaves `lock.locked()` true and a subsequent non-blocking acquire fails. The old `asyncio.Lock` waiter path removed cancelled waiters instead of leaking ownership, so this is introduced by the replacement.
- Suggested fix: keep an explicit acquisition task shielded from outer cancellation and, if the caller is cancelled, wait for that acquisition to finish and release the lock before re-raising; or use another cross-loop serialization design with equivalent cancellation cleanup. The release must cover cancellation that lands while acquisition is still queued, not only cancellation after line 73 returns.
- Test: hold the per-server lock, start and cancel a waiting authorization-header call, release the holder, then assert a third call completes within a short bound and still performs exactly one token fetch.


Reviewed head SHA: `82d0edc1561c8fe243e39a4d6e3947cba7a8c8ab`

DaoyuanLi2816:
Follow-up: found a cancellation-safety gap in the OAuth lock-acquisition fix.

Reproduced with a bounded standalone repro (hold the lock, start a waiter, cancel it once its executor thread is genuinely blocked in `lock.acquire()`, release the holder): the abandoned thread does go on to acquire the lock, `lock.locked()` stays `True` forever afterward, and a subsequent call hangs. Root cause: the `try/finally` at line 73 only starts after the acquisition `await` returns, so a `CancelledError` delivered while that `await` is suspended skips the `finally` entirely, even though the underlying OS thread is still going to acquire the lock later and nothing is left to release it. It's also worth noting this is precisely the failure mode the *old* `asyncio.Lock` didn't have: a cancelled waiter on an `asyncio.Lock` removes itself from the internal waiter queue instead of phantom-acquiring, so this was specifically introduced by the `threading.Lock` + `asyncio.to_thread` replacement.

Fix (in the new commit): the acquisition now runs as its own `asyncio.create_task(asyncio.to_thread(lock.acquire))`, awaited via `asyncio.shield()`. Shielding means cancelling the caller no longer cancels the acquisition task itself -- the caller still observes `CancelledError` promptly, but the handler for it now (re-)waits on the still-shielded task until it actually finishes (tolerating being cancelled again mid-cleanup by just retrying the wait), calls `lock.release()` the instant it lands, and only then re-raises. That covers all three cases: cancellation before the task is even scheduled, while it's queued behind another holder, or after it's already been silently granted.

Test added: `test_get_authorization_header_cancelled_while_waiting_does_not_leak_lock` in `backend/tests/test_mcp_oauth.py` -- hold the per-server lock, start a second caller, cancel it once it's genuinely blocked in its executor thread, release the original holder, then assert a third caller completes inside a bounded `asyncio.wait_for` and still performs exactly one token fetch. Confirmed red against the pre-fix code (the third call's `wait_for` times out at 5s because the lock is left permanently locked -- bounded so the failure is fast instead of hanging the run) and green after the fix. Full `test_mcp_oauth.py` (7 tests) is green, including `test_get_authorization_header_concurrent_threads_no_deadlock`, so the original cross-thread fix this PR is about is unaffected.

WillemJiang (approved):

PR #4193, opened 2026-07-15
fix(middleware): sanitize invalid tool call arguments

Huixin615:
Fixes #4172

## Why

Strict OpenAI-compatible providers reject the next model request when a malformed tool call is replayed from conversation history.

PR #4008 normalized empty tool-call names, but `AIMessage.invalid_tool_calls` could still contain malformed JSON arguments. LangChain serializes those arguments back into the next request unchanged, causing errors such as:

`Extra data: line 1 column 77 (char 76)`

This terminates the agent instead of allowing it to recover from the invalid tool call.

## What changed

- Sanitize model-bound tool-call names and arguments before provider serialization.
- Normalize empty, missing, or invalid tool names to `unknown_tool`.
- Preserve valid JSON-object argument strings.
- Encode dictionary arguments as JSON.
- Replace malformed, missing, or non-object arguments with `{}`.
- Apply sanitization only to a copied `AIMessage`, preserving the original checkpoint history for diagnostics.
- Add regression coverage for the exact mixed tool-call payload reported in #4172 and related raw-provider edge cases.
- Update the backend middleware documentation.

## Surface area

- [ ] **Frontend UI** — page / component / setting / interaction under `frontend/`
- [ ] **Backend API** — endpoint / SSE event / request-response shape under `backend/app`
- [x] **Agents / LangGraph** — agent node, graph wiring, `langgraph.json`, or prompt change
- [ ] **Sandbox** — `docker/` or sandboxed execution
- [ ] **Skills** — change under `skills/`
- [ ] **Dependencies** — new/upgraded entry in `backend/pyproject.toml` or `frontend/package.json`
- [x] **Default behavior change** — malformed tool calls now recover instead of causing a provider 400
- [ ] **Docs / tests / CI only** — no runtime behavior change

## Screenshots / Recording

N/A — backend middleware fix with no UI changes.

## Bug fix verification

- Test path: `backend/tests/test_dangling_tool_call_middleware.py`
- Did it go red on `main` and green on this branch? **Yes**
  - Before the fix, the #4172 fixture reproduced `JSONDecodeError: Extra data: line 1 column 77 (char 76)`.
  - After the fix, all 45 focused middleware tests pass.
- The regression test also verifies that the original `AIMessage.invalid_tool_calls` payload remains unchanged.

## Validation

- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_dangling_tool_call_middleware.py -q`
  - `45 passed`
[...]

PR #4008, opened 2026-07-09
fix: recover from empty tool call names

Huixin615:
Fixes #4007

## Why

When a model emits a structured tool call with an empty tool name, DeerFlow can produce or carry a matching ToolMessage but still replay the original assistant tool call with `function.name == ""` into the next OpenAI-compatible request.

Strict providers reject that malformed message history with a 400 error, which aborts the agent run after the invalid tool call path.

## What changed

- Empty, whitespace-only, missing, or non-string tool-call names are normalized to `unknown_tool`.
- `DanglingToolCallMiddleware` now sanitizes both structured `AIMessage.tool_calls` and raw provider payloads in `additional_kwargs["tool_calls"]`.
- Existing error ToolMessages for malformed empty-name calls are kept paired with their assistant tool calls and normalized when needed.
- Added regression coverage that asserts the OpenAI-serialized payload no longer contains an empty `function.name`.
- Documented the middleware recovery behavior in `backend/AGENTS.md`.

## Surface area

- [ ] **Frontend UI** — page / component / setting / interaction under `frontend/`
- [ ] **Backend API** — endpoint / SSE event / request-response shape under `backend/app`
- [x] **Agents / LangGraph** — agent node, graph wiring, `langgraph.json`, or prompt change
- [ ] **Sandbox** — `docker/` or sandboxed execution
- [ ] **Skills** — change under `skills/`
- [ ] **Dependencies** — new/upgraded entry in `backend/pyproject.toml` or `frontend/package.json` (say what it buys us)
- [x] **Default behavior change** — changes existing behavior without the user opting in (default model, default setting, data shape)
- [ ] **Docs / tests / CI only** — no runtime behavior change

## Screenshots / Recording

N/A — backend agent message-history recovery only.

## Bug fix verification

Test path that reproduces the bug:
- `backend/tests/test_dangling_tool_call_middleware.py`

Did it go red on `main` and green on this branch?
- The added tests encode the malformed empty-name tool-call history that previously serialized as `function.name == ""`; they pass on this branch after sanitization.

## Validation

- `cd backend && UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/test_dangling_tool_call_middleware.py`
- `cd backend && UV_CACHE_DIR=/tmp/uv-cache uv run ruff check packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py tests/test_dangling_tool_call_middleware.py`
[...]

Issue #2894, opened 2026-05-11
[runtime] Write_file tool is brittle for large Markdown payloads due to JSON argument escaping


ggnnggez:
# Summary

The `write_file` tool is brittle for large Markdown payloads because the entire file body is passed as a JSON string inside the tool-call arguments.

When the generated content contains inline JSON, quotes, code fences, or other JSON-significant characters, the model must escape the whole payload perfectly. A single unescaped quote can make the function-call arguments invalid JSON. In that case, LangChain records the call in `AIMessage.invalid_tool_calls`, the tool never executes, and the intended file is not written.

This is related to, but separate from, invalid tool-call pairing recovery. Even if invalid calls are handled safely, the current `write_file` protocol still makes large report/artifact writing unnecessarily fragile.

## Error Shape

A representative failure looks like this:

```python
AIMessage(
    content="",
    tool_calls=[],
    invalid_tool_calls=[
        {
            "type": "invalid_tool_call",
            "id": "write_file:36",
            "name": "write_file",
            "args": '{"description":"write report","path":"/mnt/user-data/outputs/report.md","content":"... `{"thread_data": {**paths}}` ..."}',
            "error": "Failed to parse tool arguments: ...",
        }
    ],
)
```

The `content` field is intended to be Markdown, but because it is embedded inside the JSON tool arguments, raw quotes in snippets such as:

```text
`{"thread_data": {**paths}}`
```

can break the entire function-call arguments object if they are not escaped.

## Root Cause

`write_file` currently expects the full file content as a JSON string argument:

```python
def write_file_tool(
    runtime: Runtime,
    description: str,
    path: str,
    content: str,
    append: bool = False,
) -> str:
```

For long Markdown reports, `content` often includes:

- inline JSON snippets
- fenced code blocks
- quotes and backslashes
- tables and other Markdown syntax
- multilingual text

This pushes a large, high-entropy text payload through a channel that requires strict JSON escaping.

## Expected Behavior

Writing a large Markdown report should not depend on the model perfectly escaping a large JSON string.

If the agent needs to save an artifact, the system should provide a more robust pathway where small metadata such as `path` goes through tool-call JSON, while the large Markdown body is captured through a safer channel.

## Suggested Fixes

Short-term:

1. Add regression tests for `write_file` calls whose `content` contains inline JSON, code fences, quotes, and long Markdown.
[...]

PR #4246, opened 2026-07-16
fix(middleware): recover malformed tool-call ids in dangling repair

AnnaSuSu:
## Why

`DanglingToolCallMiddleware` normalizes malformed tool-call **names**
(`unknown_tool`, #4008) and **arguments** (`{}`, #4193) so strict
OpenAI-compatible providers do not reject the next request. The **id** is the
third field of that same recovery contract and is not normalized.

A provider that emits `"id": ""` — or omits it — parses into a *well-formed*
`tool_calls` entry with `id=''`/`None` (not `invalid_tool_calls`), so it reaches
the middleware through the normal path. Three sites then interact:

- `tool_call_ids` is built with `if tc_id:`, so an empty id never enters the pairing set
- the orphan pass drops any `ToolMessage` whose `tool_call_id` is not in that set
- the placeholder pass skips the call with `if not tc_id: continue`

So an empty-id call is forwarded to the provider verbatim (HTTP 400), and if its
tool **already ran**, the middleware **drops the real result** as an orphan while
the assistant message keeps `id=""`. For the dangling case
`_build_patched_messages` returns `None` — "nothing to patch". This is the same
recovery the sibling fields already get, not a new product rule.

**Scope note on evidence:** the contract gap and both failure modes are
reproduced below against the real provider parse and serializer. What I have
*not* measured is how often providers emit an empty id in live traffic — the
empty-name sibling (#4008) is presumably more common. The argument here is the
asymmetry plus the dropped result, not frequency.

## What changed

`agents/middlewares/dangling_tool_call_middleware.py` — a `_normalize_tool_call_ids`
pass now runs first, replacing malformed ids with stable position-derived synthetic
ids and re-pointing each already-paired `ToolMessage` at its call's new id in
document order. The existing pairing / orphan / placeholder logic is unchanged — it
simply no longer sees a malformed id.

Two deliberate boundaries:

- **Only the view that is actually serialized is relabelled.** `_convert_message_to_dict`
  serializes `tool_calls + invalid_tool_calls` and reaches for the raw
  `additional_kwargs` payload only in its `elif` — i.e. never while *either* structured
  view is non-empty. Relabelling a shadowed raw view would invent a *second* id for a
  call the provider never sees, and that id's placeholder would reach the wire as a tool
  result with no matching `tool_call`. The ids also cannot simply be copied across, since
  a partially-parsed turn splits calls between `tool_calls` and `invalid_tool_calls` and
  breaks positional alignment. `invalid_tool_calls` **is** relabelled — it is serialized
  alongside `tool_calls`. (Corrected after @willem-bd's review: the first revision gated
  on `tool_calls` alone, so a malformed turn carrying both an `invalid_tool_calls` and a
  raw view — the realistic #2894 shape — put an orphan tool result on the wire.)
- **Valid ids are left byte-for-byte alone.** They are matched verbatim against
  `ToolMessage.tool_call_id`, so normalizing (e.g. stripping) one would break
  pairing that works today.

**How a recovered result finds its call** (reworked twice after @fancyboi999's reviews —
the first revision paired on the original id alone; the second still defaulted to the
first eligible sibling. Both were wrong). A malformed id cannot identify its own result:
every one of them is equally empty, so keying on it is a global FIFO over the transcript.
The rule is therefore that a result claims a call **only when the choice is forced**, and
is otherwise left malformed for the existing orphan pass to drop:

- **A result may only claim an unanswered call from the most recent AIMessage.** The pass
  walks messages once in document order, so a result can only answer the turn that issued
  it. Otherwise an earlier *dangling* call consumes a later turn's result — serving the
  real result to the wrong call while the call that actually ran gets the placeholder —
  and an orphan whose AIMessage is already gone gets adopted rather than dropped.
- **Within that turn, a usable `ToolMessage.name` narrows the candidates.** Either side
  may legitimately lack a name (that is what the empty-name sibling recovery is for), so
  only two usable names that differ rule a pairing out. One candidate left → claim it.
- **Several candidates → position decides, but only while the turn is fully answered.**
  Identical parallel calls (two `bash`) are distinguishable by nothing else, and order
  here is a construction guarantee rather than an assumption about the provider:
  LangGraph's `ToolNode` builds the results with `asyncio.gather` / `executor.map` over
  `tool_calls`, both of which yield in input order however the tools interleave. A
  **missing** result means a call was interrupted — this middleware's own trigger — so
  the survivors can no longer be trusted to line up, and nothing is claimed.
- **An unattributable or ambiguous result is dropped, not repurposed.** Dropping is
  exactly what it gets on `main` today; handing it to an arbitrary call would invent a
  pairing and silently corrupt the transcript.

**Behaviour worth calling out:** a real result *can* now be dropped rather than
misattributed — two malformed parallel calls with a single nameless result yield two
placeholders and no `REAL RESULT`. That is deliberate: nothing ties the result to either
call, and a model told a call was interrupted retries it, whereas a misattributed result
is silently wrong.

`test_tool_call_id_none_orphan_is_dropped` and the existing orphan-drop / reused-id /
grouping behavior are unaffected.

## Surface area

- [x] **Agents / LangGraph** — `DanglingToolCallMiddleware` under `packages/harness/deerflow/agents/middlewares`
- [x] **Docs / tests / CI only** — `tests/test_dangling_tool_call_middleware.py`

## Bug fix verification

- Test path: `tests/test_dangling_tool_call_middleware.py::TestMalformedToolCallIdRecovery` (18 tests)
- Red on `main` / green on this branch? **yes** — `18 failed` on unfixed main, `18
  passed` after the fix (source tree confirmed reverted: 0 hits for the new code while
  running them). `test_valid_ids_are_left_byte_for_byte_unchanged` used to be the one
  green-by-construction item here; after @willem-bd's review it mixes a malformed
  sibling into the transcript so patching actually happens, which turned it into a real
  red→green anchor instead of one asserting through the `patched is None` no-op path.
  (Counts are pytest items throughout — one test is parametrized over `""` / `"   "` /
  `None`.)
- Red check was in place: only the middleware swapped to `origin/main`, tests held
  fixed, then restored (`shasum`-verified both ways). Each clause of the fix was
  additionally deleted on its own and every one has a failing test:

  | clause deleted | test that goes red |
  |---|---|
  | `invalid_tool_calls` source | `test_empty_id_invalid_tool_call_is_normalized` |
  | raw payload source | `test_empty_id_raw_provider_tool_call_is_normalized` |
  | ToolMessage re-pointing pass | `..._keeps_its_paired_tool_result`, `..._pair_in_order` |
  | raw-only-when-unread condition | `test_only_the_serialized_view_of_a_call_gets_a_recovered_id` |
  | turn-scoped claim (`open_calls` reset per AIMessage) | `test_dangling_call_does_not_consume_a_later_turns_result` |
  | name disambiguation (`_names_can_pair`) | `test_sibling_call_does_not_consume_its_neighbours_result`, `test_result_naming_no_call_is_dropped_rather_than_misattributed` |
  | raw gate's `not invalid` clause | `test_raw_view_shadowed_by_invalid_calls_gets_no_placeholder` |
  | ambiguity guard (`candidates > 1` → no claim) | `..._nameless_result_is_dropped_when_several_siblings_are_eligible`, `..._identical_parallel_calls_drop_a_lone_ambiguous_result` |
  | positional alignment condition | `test_identical_parallel_calls_pair_with_their_results_in_order` |
  | (fix made to strip valid ids) | `test_valid_ids_are_left_byte_for_byte_unchanged` |

  One clause had no failing test — an `isinstance(original, str)` guard I had
  written. It was removed from the fix rather than left unguarded: keeping it meant a
  `None`-id call could not reclaim its own `None`-id result, which is the very bug
  this PR fixes. `test_none_id_call_reclaims_its_own_none_id_result` now pins that.

## Validation

```
cd backend
make lint                # All checks passed! / 869 files already formatted
make test-blocking-io    # 39 passed
PYTHONPATH=. uv run pytest tests/test_dangling_tool_call_middleware.py -q   # 63 passed

# full suite vs clean main, same command, same worktree
PYTHONPATH=. uv run pytest tests/ -q -p no:randomly
# this branch: 8 failed, 7680 passed, 28 skipped
# origin/main: 8 failed, 7662 passed, 28 skipped
# failure sets identical verbatim; +18 passed = the 18 new tests.
# new failures introduced: 0
```

The 8 pre-existing failures are the network-dependent
`browserless` / `crawl4ai` / `fastcrw` web-fetch tests; they fail identically on a
clean `origin/main`.

E2E through the real async model-call path — real provider frame → real
`langchain_openai` parse → real `awrap_model_call` → real OpenAI serializer:

| | `main` | this branch |
|---|---|---|
| assistant `tool_call.id` on the wire | `''` → strict provider 400 | `deerflow_synthetic_tool_call_1_call_0` |
| the tool's real result | **dropped** (`1 orphan(s) dropped, 0 placeholder(s) injected`) | preserved, paired to the recovered id |

## AI assistance

**Tool(s) used:** Claude Code (Anthropic)

**How you used it:** Used an AI coding assistant to confirm the name/args/id sibling
gap against the middleware's own recovery contract, reproduce both failure modes
through the real provider parse + serializer, draft the normalization pass and
regression tests, and run red→green / per-clause red-check / full-suite-vs-main
verification.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
@@ -101,10 +101,10 @@ def _claim_synthetic_id(open_calls: list[dict], result: ToolMessage) -> str | No
     for the orphan pass to drop, which is what an unattributable result gets today —
     better than repurposing it as some other call's answer.
     """
-    for entry in open_calls:
+    for position, entry in enumerate(open_calls):
         if entry["original"] != result.tool_call_id or not _names_can_pair(entry["name"], result.name):
             continue
-        open_calls.remove(entry)
+        del open_calls[position]
         return entry["synthetic"]
     return None
 
@@ -359,13 +359,17 @@ class DanglingToolCallMiddleware(AgentMiddleware[AgentState]):
                 additional_kwargs = getattr(msg, "additional_kwargs", None) or {}
                 raw_tool_calls = additional_kwargs.get("tool_calls")
 
+                invalid = getattr(msg, "invalid_tool_calls", None) or []
                 sources: list[tuple[str, list, str]] = [
                     ("call", structured, "tool_calls"),
-                    ("invalid", getattr(msg, "invalid_tool_calls", None) or [], "invalid_tool_calls"),
+                    ("invalid", invalid, "invalid_tool_calls"),
                 ]
-                # Mirror _message_tool_calls: the raw payload is a fallback view of the same
-                # calls, so relabel it only when it is the view actually read and serialized.
-                if not structured and isinstance(raw_tool_calls, list):
+                # The raw payload is a fallback view of the same calls, so relabel it only when
+                # it is the view actually serialized: the OpenAI serializer reaches for it only
+                # once *both* structured views are empty. Minting an id for a shadowed raw view
+                # would owe a placeholder to a call the provider never sees, putting an orphan
+                # tool result on the wire.
+                if not structured and not invalid and isinstance(raw_tool_calls, list):
                     sources.append(("raw", raw_tool_calls, "additional_kwargs"))
 
                 for source, tool_calls, field in sources:
--- a/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
@@ -48,28 +48,65 @@ def _valid_tool_call_id(tool_call_id: object) -> bool:
     return isinstance(tool_call_id, str) and bool(tool_call_id.strip())
 
 
-def _relabel_tool_call_ids(tool_calls: list, msg_index: int, source: str) -> tuple[list, list[tuple[object, str]], bool]:
+def _tool_call_name(tool_call: dict) -> object:
+    """Return a call's declared name, mirroring _message_tool_calls' raw-payload fallback."""
+    name = tool_call.get("name")
+    if _valid_tool_name(name):
+        return name
+    function = tool_call.get("function")
+    return function.get("name") if isinstance(function, dict) else name
+
+
+def _names_can_pair(call_name: object, result_name: object) -> bool:
+    """Whether a result's name contradicts a call's name.
+
+    Either side may legitimately be missing (the empty-name sibling recovery exists
+    for exactly that), and a missing name cannot contradict anything — only two
+    usable names that differ rule the pairing out.
+    """
+    if not _valid_tool_name(call_name) or not _valid_tool_name(result_name):
+        return True
+    return call_name.strip() == result_name.strip()
+
+
+def _relabel_tool_call_ids(tool_calls: list, msg_index: int, source: str) -> tuple[list, list[dict], bool]:
     """Replace malformed ids in one tool-call list with stable synthetic ids.
 
     The id is derived from the call's position so both the pairing pass and the
     model-bound message agree on it without threading state between them.
 
-    Returns the rewritten list, the ``(original_id, synthetic_id)`` pairs whose
-    existing ToolMessages must follow their call, and whether anything changed.
+    Returns the rewritten list, one ``{original, synthetic, name}`` claim entry per
+    relabelled call, and whether anything changed.
     """
     relabeled: list = []
-    assigned: list[tuple[object, str]] = []
+    assigned: list[dict] = []
     changed = False
     for position, tool_call in enumerate(tool_calls):
         if not isinstance(tool_call, dict) or _valid_tool_call_id(tool_call.get("id")):
             relabeled.append(tool_call)
             continue
-        original = tool_call.get("id")
         synthetic = f"{_SYNTHETIC_TOOL_CALL_ID_PREFIX}{msg_index}_{source}_{position}"
         relabeled.append({**tool_call, "id": synthetic})
         changed = True
-        assigned.append((original, synthetic))
+        assigned.append({"original": tool_call.get("id"), "synthetic": synthetic, "name": _tool_call_name(tool_call)})
     return relabeled, assigned, changed
+
+
+def _claim_synthetic_id(open_calls: list[dict], result: ToolMessage) -> str | None:
+    """Consume the open malformed call that ``result`` answers, returning its new id.
+
+    Malformed originals are all equally empty, so they cannot identify their own
+    result; ``open_calls`` is already scoped to the issuing turn, and the name rules
+    out the wrong sibling within it. Returning ``None`` leaves the result malformed
+    for the orphan pass to drop, which is what an unattributable result gets today —
+    better than repurposing it as some other call's answer.
+    """
+    for entry in open_calls:
+        if entry["original"] != result.tool_call_id or not _names_can_pair(entry["name"], result.name):
+            continue
+        open_calls.remove(entry)
+        return entry["synthetic"]
+    return None
 
 
 def _normalize_tool_name(name: object) -> str:
@@ -308,49 +345,51 @@ class DanglingToolCallMiddleware(AgentMiddleware[AgentState]):
         logic treat the call like any other, mirroring the empty-name recovery.
         """
         rewritten: dict[int, object] = {}
-        synthetic_by_original: dict[object, deque[str]] = defaultdict(deque)
+        # Malformed calls from the most recent AIMessage that are still unanswered.
+        # Walking in document order and resetting here is what scopes a result to the
+        # turn that issued it: a result never answers a call from an earlier turn, so
+        # an earlier dangling call must not consume a later turn's result.
+        open_calls: list[dict] = []
 
         for index, msg in enumerate(messages):
-            if getattr(msg, "type", None) != "ai":
+            if getattr(msg, "type", None) == "ai":
+                update: dict = {}
+                assigned: list[dict] = []
+                structured = getattr(msg, "tool_calls", None) or []
+                additional_kwargs = getattr(msg, "additional_kwargs", None) or {}
+                raw_tool_calls = additional_kwargs.get("tool_calls")
+
+                sources: list[tuple[str, list, str]] = [
+                    ("call", structured, "tool_calls"),
+                    ("invalid", getattr(msg, "invalid_tool_calls", None) or [], "invalid_tool_calls"),
+                ]
+                # Mirror _message_tool_calls: the raw payload is a fallback view of the same
+                # calls, so relabel it only when it is the view actually read and serialized.
+                if not structured and isinstance(raw_tool_calls, list):
+                    sources.append(("raw", raw_tool_calls, "additional_kwargs"))
+
+                for source, tool_calls, field in sources:
+                    relabeled, source_assigned, changed = _relabel_tool_call_ids(tool_calls, index, source)
+                    assigned.extend(source_assigned)
+                    if not changed:
+                        continue
+                    update[field] = {**additional_kwargs, "tool_calls": relabeled} if field == "additional_kwargs" else relabeled
+
+                open_calls = assigned
+                if update:
+                    rewritten[index] = msg.model_copy(update=update)
                 continue
 
-            update: dict = {}
-            structured = getattr(msg, "tool_calls", None) or []
-            additional_kwargs = getattr(msg, "additional_kwargs", None) or {}
-            raw_tool_calls = additional_kwargs.get("tool_calls")
-
-            sources: list[tuple[str, list, str]] = [
-                ("call", structured, "tool_calls"),
-                ("invalid", getattr(msg, "invalid_tool_calls", None) or [], "invalid_tool_calls"),
-            ]
-            # Mirror _message_tool_calls: the raw payload is a fallback view of the same
-            # calls, so relabel it only when it is the view actually read and serialized.
-            if not structured and isinstance(raw_tool_calls, list):
-                sources.append(("raw", raw_tool_calls, "additional_kwargs"))
-
-            for source, tool_calls, field in sources:
-                relabeled, assigned, changed = _relabel_tool_call_ids(tool_calls, index, source)
-                for original, synthetic in assigned:
-                    synthetic_by_original[original].append(synthetic)
-                if not changed:
-                    continue
-                update[field] = {**additional_kwargs, "tool_calls": relabeled} if field == "additional_kwargs" else relabeled
-
-            if update:
-                rewritten[index] = msg.model_copy(update=update)
+            # Re-point an already-paired result at its call's new id so the pairing
+            # below keeps it instead of dropping it as an orphan.
+            if not isinstance(msg, ToolMessage) or _valid_tool_call_id(msg.tool_call_id):
+                continue
+            synthetic = _claim_synthetic_id(open_calls, msg)
+            if synthetic is not None:
+                rewritten[index] = msg.model_copy(update={"tool_call_id": synthetic})
 
         if not rewritten:
             return messages
-
-        # Re-point each already-paired result at its call's new id, in document order,
-        # so the pairing below keeps it instead of dropping it as an orphan.
-        for index, msg in enumerate(messages):
-            if not isinstance(msg, ToolMessage) or _valid_tool_call_id(msg.tool_call_id):
-                continue
-            queue = synthetic_by_original.get(msg.tool_call_id)
-            if queue:
-                rewritten[index] = msg.model_copy(update={"tool_call_id": queue.popleft()})
-
         return [rewritten.get(index, msg) for index, msg in enumerate(messages)]
 
     def _build_patched_messages(self, messages: list) -> list | None:
--- a/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
@@ -113,9 +113,12 @@ def _claim_synthetic_id(open_calls: list[dict], result: ToolMessage, positional:
     * one compatible call — its name, or being the turn's only call, identifies it;
     * several compatible calls — position identifies them, but only while ``positional``
       holds, i.e. every open call in the turn has a result. Identical parallel calls
-      (two ``bash``) are distinguishable by nothing else, and a provider returns their
-      results in call order. A *missing* result means a call was interrupted — this
-      middleware's own trigger — and the survivors can no longer be trusted to line up.
+      (two ``bash``) are distinguishable by nothing else, and order here is a
+      construction guarantee rather than an assumption about the provider: LangGraph's
+      ``ToolNode`` builds the results with ``asyncio.gather`` / ``executor.map`` over
+      ``tool_calls``, both of which yield in input order however the tools interleave.
+      A *missing* result means a call was interrupted — this middleware's own trigger —
+      so the surviving results can no longer be trusted to line up with the calls.
 
     Returning ``None`` leaves the result malformed for the orphan pass to drop, which is
     what an unattributable result gets today — better than inventing a pairing.
--- a/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
@@ -37,10 +37,39 @@ logger = logging.getLogger(__name__)
 _MAX_RECOVERY_ERROR_DETAIL_LEN = 500
 _UNKNOWN_TOOL_NAME = "unknown_tool"
 _EMPTY_TOOL_NAME_ERROR = "Tool call could not be executed because its name was missing or empty."
+_SYNTHETIC_TOOL_CALL_ID_PREFIX = "deerflow_synthetic_tool_call_"
 
 
 def _valid_tool_name(name: object) -> bool:
     return isinstance(name, str) and bool(name.strip())
+
+
+def _valid_tool_call_id(tool_call_id: object) -> bool:
+    return isinstance(tool_call_id, str) and bool(tool_call_id.strip())
+
+
+def _relabel_tool_call_ids(tool_calls: list, msg_index: int, source: str) -> tuple[list, list[tuple[object, str]], bool]:
+    """Replace malformed ids in one tool-call list with stable synthetic ids.
+
+    The id is derived from the call's position so both the pairing pass and the
+    model-bound message agree on it without threading state between them.
+
+    Returns the rewritten list, the ``(original_id, synthetic_id)`` pairs whose
+    existing ToolMessages must follow their call, and whether anything changed.
+    """
+    relabeled: list = []
+    assigned: list[tuple[object, str]] = []
+    changed = False
+    for position, tool_call in enumerate(tool_calls):
+        if not isinstance(tool_call, dict) or _valid_tool_call_id(tool_call.get("id")):
+            relabeled.append(tool_call)
+            continue
+        original = tool_call.get("id")
+        synthetic = f"{_SYNTHETIC_TOOL_CALL_ID_PREFIX}{msg_index}_{source}_{position}"
+        relabeled.append({**tool_call, "id": synthetic})
+        changed = True
+        assigned.append((original, synthetic))
+    return relabeled, assigned, changed
 
 
 def _normalize_tool_name(name: object) -> str:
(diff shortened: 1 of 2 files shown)


fancyboi999 (changes requested):
Thanks @AnnaSuSu. I found one ordering bug in the malformed-ID recovery path.


Reviewed head SHA: `d0dcc42f0fe60ecd27685e7ae75ef43c52f383d9`

[P2] Do not let a later result satisfy an earlier dangling call

- Location: `backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py:310-352`
- Problem: all malformed calls are queued by their original empty/`None` ID before any `ToolMessage` is reassigned. A real result from a later turn therefore consumes the first synthetic ID even when that first call was dangling, silently attaching the result to the wrong call and leaving the actual call with a placeholder.
- Evidence: for `[AI(call id="", no result), AI(call id=""), ToolMessage(id="", "REAL RESULT")]`, `synthetic_by_original[""]` contains the first and second call IDs, while the lone result at lines 347-352 is relabeled to the first. The grouping pass then emits `REAL RESULT` after the first AI message and injects an error for the second. The current multi-empty-ID test only covers the aligned case where every call has a result.
- Suggested fix: associate malformed results within their causal AI/tool segment (or with the nearest eligible preceding malformed call), rather than using one global FIFO keyed only by the invalid original ID. If an association is ambiguous, do not repurpose a later result as an older call's result.
- Test: add a two-turn regression where the first empty-ID call is dangling and the second has a real empty-ID result; assert the first gets the placeholder and the real result remains paired with the second. Also cover an orphan empty-ID `ToolMessage` before a later malformed call.

AnnaSuSu:
Thanks @fancyboi999 — you're right, and the first revision was wrong. Reproduced your case verbatim:

```
in:  [AI(search id=""), AI(write_file id=""), ToolMessage(id="", "REAL RESULT")]
out: AI(search)     → ToolMessage(name='write_file') content='REAL RESULT'   ← wrong call
     AI(write_file) → ToolMessage(content='[Tool call was interrupted...]')  ← the one that ran
```

One detail worse than your report: the misattributed `ToolMessage` carries `name='write_file'` too, so the model sees `search` answered with `write_file`'s output — not just a misordering.

Your second case is the one that really needed catching, because **it was a regression I introduced**: on `main` that stale orphan is dropped (`1 orphan(s) dropped`); my pass let a later malformed call adopt it. My PR broke this middleware's own existing orphan-drop contract.

**A third instance your review didn't cover.** Same bug class, but *within a single AIMessage* between parallel calls — which is this middleware's own trigger (interruption during parallel tool execution):

```
in:  [AI(tool_calls=[search id="", write_file id=""]), ToolMessage(id="", name="write_file")]
out: write_file's real result → attached to search
```

Segment scoping alone doesn't fix this one — both calls are in the same segment. So the fix is two clauses:

- **Turn-scoped claim** — one walk in document order; `open_calls` holds only the most recent AIMessage's unanswered calls. A result can only answer the turn that issued it. (Your case 1; case 2 falls out for free — an orphan before any call finds `open_calls` empty and drops as before.)
- **Name disambiguation** — within a turn, a usable `ToolMessage.name` rules out the wrong sibling. Either side may legitimately lack a name (that's what the empty-name recovery is for), so only two usable names that differ rule a pairing out.
- **Unattributable → dropped, not repurposed** — matches no open call, stays malformed, existing orphan pass drops it. That's what it already gets on `main`; handing it to an arbitrary call would invent a pairing. (Your "if ambiguous, do not repurpose".)

**On the per-clause red check**, worth flagging: my first two-turn regression used *different* tool names, so the name clause rescued it and the turn-scoping clause had **zero** failing tests when deleted — it was pinning the wrong clause. Rewritten as a same-name retry (`bash` → `bash`, where only document order can discriminate), which is also the more realistic shape. Both clauses now have a failing test:

| clause deleted | test that goes red |
|---|---|
| turn-scoped claim | `test_dangling_call_does_not_consume_a_later_turns_result` |
| name disambiguation | `test_sibling_call_does_not_consume_its_neighbours_result`, `test_result_naming_no_call_is_dropped_rather_than_misattributed` |

4 new tests; `13 failed, 1 passed` on unfixed `main`, all 14 green after. Full suite `8 failed, 7676 passed` — same pre-existing network-dependent failures as `main`, 0 new. PR body updated; the old claim that a result "is reclaimed only by a call that actually carried that id" was exactly the false premise you caught.

backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
```diff
@@ -267,19 +333,80 @@ def _sanitize_ai_message_tool_calls(msg):
...
+                additional_kwargs = getattr(msg, "additional_kwargs", None) or {}
+                raw_tool_calls = additional_kwargs.get("tool_calls")
+
+                sources: list[tuple[str, list, str]] = [
+                    ("call", structured, "tool_calls"),
+                    ("invalid", getattr(msg, "invalid_tool_calls", None) or [], "invalid_tool_calls"),
+                ]
+                # Mirror _message_tool_calls: the raw payload is a fallback view of the same
+                # calls, so relabel it only when it is the view actually read and serialized.
+                if not structured and isinstance(raw_tool_calls, list):
```
willem-bd: Pre-existing latent edge (not a regression, but worth a one-line note so a future reader doesn't assume one-call-per-view): when `tool_calls` is empty *and* both `invalid_tool_calls` and raw `additional_kwargs.tool_calls` are present with malformed ids, both views get relabelled here with *different* synthetic ids (`..._invalid_0` vs `..._raw_0`). The paired `ToolMessage` claims the first view (invalid); the raw view then looks dangling and gets a placeholder, so the model sees a duplicate call. This mirrors `_message_tool_calls`, which already reads both views on `main` and produces duplicate calls when ids are valid - so it's consistent, not new - but a comment that `invalid` + `raw` can be two views of one turn would help. Out of scope to fix in this PR.

backend/tests/test_dangling_tool_call_middleware.py
```diff
@@ -727,6 +727,263 @@ def test_invalid_tool_call_already_responded_is_sanitized_without_placeholder(se
...
+                response_metadata={},
+            )
+        ]
+
+        patched = mw._build_patched_messages(msgs)
+
+        assert patched is not None
+        recovered_id = patched[0].tool_calls[0]["id"]
+        assert recovered_id
+        assert _convert_message_to_dict(patched[0])["tool_calls"][0]["id"] == recovered_id
```
willem-bd: This asserts `payload["tool_calls"][0]["id"] == recovered_id` but doesn't lock the count. The whole 'only the serialized view is relabelled' design rests on the OpenAI serializer ignoring raw `additional_kwargs.tool_calls` when structured `tool_calls` coexists (line 856 keeps the raw `id=""`). If that serializer behavior ever changes, the raw `id=""` would slip onto the wire as a second `tool_calls` entry and cause the 400 this PR exists to prevent - and this test wouldn't catch it. A one-line `assert len(_convert_message_to_dict(patched[0])["tool_calls"]) == 1` would pin the invariant the test is named after.

backend/tests/test_dangling_tool_call_middleware.py
```diff
@@ -727,6 +727,263 @@ def test_invalid_tool_call_already_responded_is_sanitized_without_placeholder(se
...
+        """
+        mw = DanglingToolCallMiddleware()
+        msgs = [
+            _ai_with_tool_calls([_tc("bash", " call_1 ")]),
+            ToolMessage(content="result", tool_call_id=" call_1 ", name="bash"),
+        ]
+
+        patched = mw._build_patched_messages(msgs)
+
+        assert patched is None or patched[0].tool_calls[0]["id"] == " call_1 "
```
willem-bd: This asserts the byte-for-byte guarantee only via the no-op path: the transcript is fully responded, so `_build_patched_messages` returns `None` and `patched is None` short-circuits. It does catch stripping indirectly (a stripped AIMessage id would orphan the ToolMessage, making `patched` non-`None` and failing the equality) - but that's subtle. A stronger version mixes a valid whitespace id with a malformed one so patching actually happens, then asserts the valid id is unchanged in the patched output, e.g. `_ai_with_tool_calls([_tc("bash", " call_1 "), _tc("ls", "")])` plus a result for ` call_1 `.

backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
```diff
@@ -37,12 +37,78 @@
...
+    Malformed originals are all equally empty, so they cannot identify their own
+    result; ``open_calls`` is already scoped to the issuing turn, and the name rules
+    out the wrong sibling within it. Returning ``None`` leaves the result malformed
+    for the orphan pass to drop, which is what an unattributable result gets today —
+    better than repurposing it as some other call's answer.
+    """
+    for entry in open_calls:
+        if entry["original"] != result.tool_call_id or not _names_can_pair(entry["name"], result.name):
+            continue
+        open_calls.remove(entry)
```
willem-bd: Minor: `list.remove(entry)` does an equality scan and is only safe here because synthetic ids are unique (no two entries compare equal). Iterating by index and `del open_calls[i]` once a match is found would make the 'remove exactly this entry' intent obvious and not rely on uniqueness for correctness. Pure style nit.

willem-bd (commented):
Review nits (non-blocking) from code review. Overall the fix is correct and well-tested; these are minor robustness/style suggestions. Findings #2 and #3 are the ones most worth a follow-up since they harden invariants the design depends on.

fancyboi999 (changes requested):
Thanks @AnnaSuSu. The previous cross-turn finding is fixed on this head, and the named-sibling cases now preserve the real result. One ambiguous same-turn path still misattributes a result.


Reviewed base SHA: `ae223199fd8670b8f132210fb1038590b2befdda`
Reviewed head SHA: `703bdc6a7df76faec189517bcf62ddb61fff74f2`

[P2] Do not claim a nameless result when multiple malformed calls are eligible

- Location: `backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py:95-108`
- Problem: `_claim_synthetic_id()` still consumes the first open call whenever either side lacks a usable name. With two malformed parallel calls and one nameless `ToolMessage`, there is no evidence identifying which call produced it, but the result is nevertheless attached to the first call and the other receives a placeholder.
- Evidence: for `AI(tool_calls=[search(id=""), write_file(id="")])` followed by `ToolMessage(tool_call_id="", name=None, content="REAL RESULT")`, `_names_can_pair()` returns `True` for the first entry because the result name is missing, so `REAL RESULT` is emitted as the `search` result. `ToolMessage.name` is optional, the new `None`-ID regression already constructs a result without a name, and this PR's own docstring says either side may legitimately lack one. This therefore preserves the same silent transcript corruption the follow-up is meant to eliminate, only inside one turn.
- Suggested fix: determine all compatible open calls before consuming one. A usable result name may narrow the set, but only claim the result when the remaining candidate is unique; otherwise leave it malformed so the existing orphan pass drops it rather than inventing a pairing.
- Test: add two empty-ID parallel calls plus a single nameless empty-ID result and assert it is not assigned to the first sibling (with no unique attribution, it should be dropped and both calls should receive placeholders). Keep coverage that a nameless result is preserved when there is only one eligible call and that distinct named siblings still pair correctly.

backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
AnnaSuSu: Thanks — you found the right code and the right mechanism, but the severity is worse than "latent, not new", and it's mine. I ran your case both ways before answering.

**Where you're right:** the *valid*-id shape is pre-existing. `main` and this branch emit byte-identical output for `[AI(invalid id="call_1", raw id="call_1"), ToolMessage("call_1")]` — one serialized call, two results (a duplicate *result*, strictly). That's `_message_tool_calls` gating raw on `not tool_calls` while the serializer gates it on `not tool_calls and not invalid_tool_calls`. This PR doesn't touch that, and I'd rather not widen a malformed-id fix into a valid-id behaviour change — so: agreed, out of scope.

**Where it's a regression I introduced:** the *malformed*-id shape, which is the one this PR creates. `_convert_message_to_dict` reaches for the raw view only in its `elif`, so raw is never serialized while `invalid_tool_calls` is non-empty. My gate only checked `not structured`, so the shadowed raw view got `..._raw_0` and was owed a placeholder for a call the provider never sees:

```
# this branch, BEFORE the fix — the realistic #2894 shape:
#   [AI(tool_calls=[], invalid_tool_calls=[write_file id=""], additional_kwargs.tool_calls=[id=""]),
#    ToolMessage(id="", "REAL RESULT")]
  assistant tool_call  id='deerflow_synthetic_tool_call_0_invalid_0'
  tool     result      id='deerflow_synthetic_tool_call_0_raw_0'      <-- no matching tool_call
  tool     result      id='deerflow_synthetic_tool_call_0_invalid_0'
```

That orphan tool result *is* the HTTP 400 this PR exists to prevent, and `main` does not produce it — it drops the result and sends `id=''` instead. Two different failures, so "consistent with main" doesn't hold on this path. Fixed rather than commented.

The gate is now `not structured and not invalid`, i.e. the serializer's actual rule — which is what my own comment already claimed ("relabel it only when it is the view actually read and serialized") while the code did something looser. This is deliberately stricter than `_message_tool_calls`, which is harmless: the shadowed raw entry keeps its empty id and `_build_patched_messages` skips it at `if not tc_id: continue`.

The regression test asserts on the serialized wire, since that is where a strict provider rejects. It is also the only thing guarding the clause — deleting just `not invalid` and running the file gives `1 failed, 59 passed`, the 1 being the new test.

E2E through the real `wrap_model_call`, same input:

| | `main` (7df44f58) | this branch |
|---|---|---|
| assistant tool_call id | `''` | `..._invalid_0` |
| real result | **dropped** | preserved, paired |
| verdict | provider rejects (400) | valid request |

backend/tests/test_dangling_tool_call_middleware.py
AnnaSuSu: Added `assert len(_convert_message_to_dict(patched[0])["tool_calls"]) == 1`.

And your predicted failure is real rather than hypothetical — I monkeypatched a serializer that reads raw *alongside* structured (exactly the drift you describe) and re-ran the test:

```
>       assert len(payload["tool_calls"]) == 1
E       AssertionError: assert 2 == 1
E        +  where 2 = len([{...'id': 'deerflow_synthetic_tool_call_0_call_0'...}, {...'id': ''...}])
```

So the `id=""` does ride onto the wire as a second `tool_calls` entry the moment that `elif` stops shadowing raw, and this assert is what catches it. Good catch — the test was named after an invariant it wasn't actually pinning.

backend/tests/test_dangling_tool_call_middleware.py
AnnaSuSu: Confirmed, and rewritten your way. `patched is None` is `True` for that transcript, so `assert patched is None or ...` never evaluated its right-hand side — the guarantee was only ever asserted through the no-op path.

It now mixes a valid whitespace id with a malformed sibling, as you suggested, so patching actually happens and the valid id is checked in the patched output:

```python
msgs = [
    _ai_with_tool_calls([_tc("bash", " call_1 "), _tc("ls", "")]),
    ToolMessage(content="result", tool_call_id=" call_1 ", name="bash"),
]
...
assert patched is not None
kept_id, recovered_id = (tc["id"] for tc in patched[0].tool_calls)
assert kept_id == " call_1 "
```

Verified it's a live guard rather than just a better-looking one: mutating `_valid_tool_call_id` to reject non-stripped ids makes it fail (`1 failed, 59 passed`).

backend/packages/harness/deerflow/agents/middlewares/dangling_tool_call_middleware.py
AnnaSuSu: Right, and your reasoning checks out — `(msg_index, source, position)` makes the synthetic ids unique within one `open_calls`, so no two entries can compare equal and `remove()` cannot take the wrong one. Safe today, but only by relying on that.

Switched to `enumerate` + `del open_calls[position]` so "remove exactly this entry" no longer depends on the uniqueness argument.

AnnaSuSu:
Thanks @willem-bd — all four addressed in dd6fdfb8. Flagging one thing at the top level, because it changes what this push is: **finding #1 was not a nit, and not pre-existing. It was a regression this PR introduced, so it's a behaviour fix rather than the comment you suggested.**

`_convert_message_to_dict` reads the raw `additional_kwargs` view only in its `elif`, so raw is never serialized while `invalid_tool_calls` is non-empty. My id-recovery gate checked `not structured` and ignored `invalid`, so on the realistic #2894 shape (parse failure in `invalid_tool_calls`, raw payload still in `additional_kwargs` — two views of one call) the shadowed raw view got its own synthetic id and was owed a placeholder for a call the provider never sees. Result: an **orphan tool result on the wire — the exact HTTP 400 this PR exists to prevent.** `main` doesn't do that; it drops the result and sends `id=''`. So the two failures aren't the same shape, and I couldn't take "consistent, not new" for this path.

Your *valid*-id precedent does hold — `main` and this branch are byte-identical there (duplicate result, from `_message_tool_calls` disagreeing with the serializer about raw). Untouched here, still out of scope.

Gate is now `not structured and not invalid` — the serializer's actual rule, and what my own comment already claimed while the code did something looser.

Verification for this push:
- New regression test asserts on the **serialized wire**, and is the only thing guarding the clause: deleting just `not invalid` → `1 failed, 59 passed`.
- Findings #2/#3 confirmed live rather than cosmetic: the count canary fires under a drifted serializer (`assert 2 == 1`, second entry `id=''`); the byte-for-byte test really was short-circuiting on `patched is None`.
- Full suite vs clean `7df44f58`, baseline taken in place: **identical failure sets** (8 pre-existing `web_fetch` failures both sides), **0 new failures**.
- E2E through the real `wrap_model_call` on the #2894 shape: `main` → `id=''` on the wire + real result silently dropped (400); this branch → valid paired request, result preserved.
- `cd backend && make format && make lint` clean.

@fancyboi999 the turn-scoping fix for your P2 is unchanged in 703bdc6a; this push only narrows which view gets an id.

AnnaSuSu:
Thanks @fancyboi999 — confirmed, reproduced, and fixed in 577ad486. You're right that this is the same corruption as the cross-turn case, just inside one turn:

```
# reviewed head — AI(tool_calls=[search(id=""), write_file(id="")]) + ToolMessage(id="", name=None, "REAL RESULT")
  search      <- 'REAL RESULT'                     <- guessed, on no evidence
  write_file  <- '[Tool call was interrupted...]'  <- possibly the one that actually ran
```

**One deviation from your suggested fix, with evidence.** I implemented "determine all compatible open calls first" literally — `only claim the result when the remaining candidate is unique` — and measured it before keeping it:

| turn | your literal rule |
|---|---|
| `[search, write_file]` + one nameless result | dropped ✅ (what you asked for) |
| `[bash, ls]` + two results | pairs correctly ✅ (the "distinct named siblings" coverage you asked me to keep) |
| **`[bash, bash]` + two `bash` results** | **both real results dropped** ❌ |

Identical parallel calls are a normal shape, and this PR's premise is a provider that omits ids — so both calls are malformed and both results are named `bash`. The candidate set is then never unique, so every result is judged unattributable and the model is told both calls were interrupted while their output is discarded. **That rule also passes all 60 existing tests**, so the loss was invisible.

So I kept your principle — only take a *forced* choice — and graded the evidence instead:

- the result's name narrows the candidates;
- **one** candidate → claim it (the name, or being the turn's only call, identifies it);
- **several** candidates → position decides, but *only* while every open call in the turn has a result. Order here is a construction guarantee rather than an assumption about the provider: LangGraph's `ToolNode` builds these results with `asyncio.gather` (async) / `executor.map` (sync) over `tool_calls`, and both yield in **input** order however the tools interleave — so a fully answered turn lines up by construction. A **missing** result means a call was interrupted, which is this middleware's own trigger, and the survivors can no longer be trusted to line up. Your case lands here: two calls, one result → not forced → dropped.
- otherwise → `None`, and the existing orphan pass drops it.

This satisfies every case in your review, including both coverages you asked me to keep, and additionally protects the same-name shape.

Tests added (all three of your asks plus the regression your rule would have hidden):
- `test_nameless_result_is_dropped_when_several_siblings_are_eligible` — your exact case; result dropped, both calls get placeholders.
- `test_identical_parallel_calls_pair_with_their_results_in_order` — pins the case above.
- `test_identical_parallel_calls_drop_a_lone_ambiguous_result` — two `bash` calls, one result → still dropped, so the positional tie-break can't decay into a first-sibling default.

Per-clause red check, each new clause deleted on its own:

```
revert to first-eligible claim   -> 2 failed (nameless-ambiguous, lone-ambiguous), 61 passed
drop the positional exemption    -> 1 failed (identical-parallel-in-order),        62 passed   <- your literal rule
force positional = True          -> 2 failed (nameless-ambiguous, lone-ambiguous), 61 passed
```

Every clause is guarded by exactly one test, and the existing pairing tests still drive their own logic (`_none_id_call_reclaims_its_own_none_id_result` → single eligible call → still claims; `_sibling_call_does_not_consume_its_neighbours_result` → name narrows to one → still claims).

Full suite vs clean baseline, taken in place: `8 failed, 7680 passed` vs `8 failed, 7662 passed` — failure sets byte-identical (pre-existing `web_fetch` cases), **0 new failures**. `make format` / `make lint` clean.

E2E through the real `wrap_model_call`:
```
[search, write_file] + nameless result -> both get placeholders, REAL RESULT attached to neither
[bash, bash] + RESULT A, RESULT B      -> both preserved, paired in call order
```



---

*Edited to correct the justification for the positional tie-break: I originally wrote that the **provider** returns parallel results in call order. That was an assumption about a third party and you would have been right to reject it — providers guarantee no such thing. The real basis is stronger and local: `ToolNode` constructs the results in `tool_calls` order itself (`asyncio.gather` / `executor.map` both preserve input order), so it holds regardless of provider behaviour or which tool finishes first. Code and test docstrings corrected in b0b93467 (comments only, no behaviour change).*

WillemJiang (approved):

PR #1127, opened 2026-03-13
feat: add multi-tenant authentication and user isolation

knight0940:
This PR implements multi-tenant authentication with user data isolation, allowing DeerFlow to support multiple users while maintaining backward compatibility for single-tenant deployments. See #1062

Features:
- JWT-based authentication with register/login/logout endpoints
- Per-user memory storage and isolation
- Per-user thread directory isolation for workspace/uploads/outputs
- Quota management system (threads, sandboxes, storage)
- User store with JSON file backend and atomic writes
- Sandbox state store with file-based persistence
- Verification scripts for testing user isolation

Security:
- PBKDF2-HMAC-SHA256 password hashing with 100k iterations
- Constant-time password comparison to prevent timing attacks
- Optional authentication for backward compatibility (default user)

Configuration:
- New config sections: multi_tenant, state_store, user_quotas
- Environment variable fallback for JWT secret
- Default user ID for unauthenticated requests

Backward Compatibility:
- Unauthenticated requests use user_id="default"
- Single-tenant deployments work without changes
- Optional authentication via get_optional_user()

PR #4249, opened 2026-07-17
feat(extensions): first-class optional hook for identity/memory providers (reference: Lithtrix)

learningloons-hash:
## Summary

This pattern already works on stock DeerFlow today, no code changes required: a `skills/public/<name>/SKILL.md` file plus an `mcpServers` entry in `extensions_config.json` is enough for an external identity/memory provider to prefetch context at session start and persist findings at session end. We've been running it against Lithtrix (`lithtrix-mcp@0.20.2`) for a few days — [writeup + the exact 3 files](https://docs.lithtrix.ai/integrations/deerflow), [community post](https://github.com/bytedance/deer-flow/discussions/4183).

This PR proposes making that pattern **first-class**: an explicit, optional `identityHooks` block so the harness fires session-start calls itself instead of relying on the agent to follow a skill's instructions. It's not a prerequisite for the pattern to work — it's a smaller, more reliable version of something already possible.

This is a **hook interface**, not a Lithtrix-specific hard dependency.

## Motivation

DeerFlow research runs are instance-local. Run N+1 doesn't inherit vetted findings from Run N across restarts or frameworks unless something outside DeerFlow carries that context forward. We measured this directly: a warm run (memory/commons prefetched before research starts) finished 48.2% faster and made 62.5% fewer search calls than a cold run, with 5 commons sources reused and directly counted. Full method and what we're *not* claiming (this isn't a full DeerFlow-UI benchmark) are in the linked writeup.

DeerFlow should not become a shared-state service. This PR exposes a minimal hook so external identity providers can plug in — Lithtrix is the reference implementation, not the only one it should support.

## What's included

- Typed `identityHooks` config block on `ExtensionsConfig` (opt-in, absent by default — zero behavior change for existing installs)
- New `IdentityHooksMiddleware` — fires configured `sessionStart` MCP tool calls once per thread; failures are non-fatal
- v1 scope is `sessionStart` only; `sessionEnd` (write-path) deferred to a follow-up once argument templating for write tools is designed
- Docs: en/zh "Optional identity hooks" subsection + cross-link from `MCP_SERVER.md`
- Reference integration example (Lithtrix): `skills/public/use-lithtrix/SKILL.md` + `extensions_config.json` snippet
- New unit tests + one existing test file updated for the new middleware count
- Full test suite passes locally, no regressions
- Live-tested against the real Lithtrix API (not mocked) — session-start hooks fire correctly, failure path is non-fatal, no metered billing or reputation side effects

## Non-goals

- No Lithtrix SDK in DeerFlow core
- No per-sub-agent passports
- No claims about multi-user isolation (that's #1127's territory)
- No `skills.<name>.path` config key — filesystem discovery under `skills/public/` only
- No requirement to enable `identityHooks` for Lithtrix — the config + skill path alone is sufficient today

## Evidence

- Writeup + the 3 files: https://docs.lithtrix.ai/integrations/deerflow
- Prior community post: https://github.com/bytedance/deer-flow/discussions/4183

### DIFF
--- a/backend/docs/MCP_SERVER.md
+++ b/backend/docs/MCP_SERVER.md
@@ -14,6 +14,10 @@ DeerFlow supports configurable MCP servers and skills to extend its capabilities
 3. Configure each server’s command, arguments, and environment variables as needed.
 4. Restart the application to load and register MCP tools.
 
+## Optional identity hooks
+
+For external identity or memory providers, DeerFlow can optionally fire MCP tool calls at thread start via an `identityHooks` block in `extensions_config.json`. See the frontend docs subsection **Optional identity hooks** in `frontend/src/content/en/harness/mcp.mdx` for schema, Lithtrix reference integration (`skills/public/use-lithtrix/SKILL.md`, `docs/examples/lithtrix/`), and v1 scope (`sessionStart` only; failures are non-fatal).
+
 ## Routing Hints
 
 Use `routing` when an MCP server should be preferred for specific requests, such
--- a/backend/packages/harness/deerflow/agents/middlewares/identity_hooks_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/identity_hooks_middleware.py
@@ -1,0 +1,137 @@
+"""Optional identity-provider MCP hooks at thread/session start.
+
+Fires configured ``sessionStart`` tool calls against an external MCP server
+(e.g. Lithtrix) without importing provider code into DeerFlow core. Failures
+are logged and ignored so research is not blocked by substrate outages.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from datetime import timedelta
+from typing import Any, override
+
+from langchain.agents import AgentState
+from langchain.agents.middleware import AgentMiddleware
+from langgraph.config import get_config
+from langgraph.runtime import Runtime
+
+from deerflow.agents.middlewares._bounded_dict import BoundedDict
+from deerflow.config.extensions_config import ExtensionsConfig, get_extensions_config
+from deerflow.mcp.client import build_server_params
+from deerflow.mcp.session_pool import get_session_pool
+from deerflow.runtime.user_context import get_effective_user_id
+
+logger = logging.getLogger(__name__)
+
+
+def _resolve_thread_id(runtime: Runtime) -> str | None:
+    context = runtime.context or {}
+    thread_id = context.get("thread_id")
+    if thread_id is not None:
+        return str(thread_id)
+    try:
+        config = get_config()
+        thread_id = config.get("configurable", {}).get("thread_id")
+        return str(thread_id) if thread_id is not None else None
+    except RuntimeError:
+        return None
+
+
+async def run_session_start_hooks(
+    *,
+    config: ExtensionsConfig,
+    thread_id: str,
+    user_id: str,
+) -> None:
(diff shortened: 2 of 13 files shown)


fancyboi999 (changes requested):
Thanks @learningloons-hash. I found two issues that should be addressed before this is ready.

[P1] Feed the hook results into the agent context

- Location: `backend/packages/harness/deerflow/agents/middlewares/identity_hooks_middleware.py:88`
- Problem: every `session.call_tool(...)` return value is discarded, `run_session_start_hooks()` returns `None`, and both agent hooks return no state update. The advertised prefetch therefore makes remote calls but none of the fetched memory or commons data can affect the following model request.
- Evidence: this is the only call site in the new middleware; there is no `ToolMessage`, hidden context message, state field, or prompt projection carrying the result. The tests only assert that `call_tool` was awaited, so they pass even though the feature is behaviorally inert.
- Suggested fix: normalize the MCP result into a bounded, tag-neutralized data block and project it into the next model request through a durable state/message path. Keep remote content outside system-role authority.
- Test: return a distinctive value from the fake MCP session and assert that the next model-bound message/state contains that value after sanitization and size limiting.

[P1] Make once-per-thread idempotency survive run construction and exclude subagents

- Location: `backend/packages/harness/deerflow/agents/middlewares/identity_hooks_middleware.py:102` and `backend/packages/harness/deerflow/agents/middlewares/tool_error_handling_middleware.py:188`
- Problem: `_fired_threads` belongs to one middleware instance, but Gateway `run_agent()` calls the agent factory for every run, which rebuilds the middleware chain. The same thread therefore fires the hooks again on every user turn. Adding the middleware to `_build_runtime_middlewares()` also installs a fresh copy in every subagent, despite the PR's thread-start scope and stated non-goal for per-subagent identity.
- Evidence: `runtime/runs/worker.py` constructs `agent = agent_factory(...)` per run, and both lead and subagent builders use the shared runtime middleware factory. A `BoundedDict` on the new instance cannot provide cross-run or cross-agent idempotency.
- Suggested fix: attach this hook only to the lead chain and persist the fired marker in checkpointed thread state or another durable user/thread-scoped store. Mark success only after the configured calls complete according to the intended retry policy.
- Test: create two separate lead-agent/middleware instances for the same thread and assert one logical session-start execution, then run a subagent on that thread and assert it does not invoke the hooks.


Reviewed head SHA: `38b66b6cc6ea1bbf18a8aca4949aa8f79bed19d9`

learningloons-hash:
Hi @fancyboi999, really appreciate you taking the time on this, and sorry
for the slow turnaround.

I addressed the issues that you highlighted:

*Hook results discarded* — fixed. run_session_start_hooks() now extracts
and sanitizes the MCP result (reusing neutralize_untrusted_tags), caps it
at 16k/tool and 32k total, and injects it as a hidden, non-system-role
HumanMessage so it's model-visible but not authoritative. Added a test
asserting a canary value survives the full path into result["messages"].

*Idempotency + subagent leak* — fixed. The fired-flag now lives in
checkpointed thread state instead of an instance attribute, so it survives
the per-run middleware rebuild. And the middleware is now gated behind the
same include_uploads flag UploadsMiddleware already uses for lead-only
scoping, so subagents never get it. Added a test confirming it's present in
build_lead_runtime_middlewares and absent from
build_subagent_runtime_middlewares.

Pushed as da98b5b1. Full suite green. Let me know if anything still looks
off — thank you again for your feedback and advice!

Mark.

PR #4252, opened 2026-07-17
fix(skills): don't attach model tracing to the in-graph skill security scan

AnnaSuSu:
## Why

`agents/lead_agent/agent.py`'s module docstring states an `INVARIANT — tracing callback placement`: tracing callbacks are attached at the **graph invocation root**, so every in-graph `create_chat_model(...)` MUST pass `attach_tracing=False`. It spells out the cost of forgetting the flag:

> Forgetting that flag emits duplicate spans (one rooted at the graph, one at the model) AND prevents the Langfuse handler's `propagate_attributes` path from firing, so `session_id` / `user_id` never reach the trace.

It lists four current sites (bootstrap agent, default agent, summarization middleware, `TitleMiddleware`'s async path) and asks that any new in-graph call be added to the list.

`skills/security_scanner.py::scan_skill_content` is reachable in-graph through the `skill_manage` tool — `tools/tools.py` registers `skill_manage_tool` in `builtin_tools`, and `skill_manage_tool.py::_scan_or_raise` calls the scanner — but it builds its moderation model with `create_chat_model`'s default `attach_tracing=True`. On that path the model attaches tracing handlers on top of the ones the graph root already attached, which is exactly what the invariant forbids.

The same function is also reachable standalone (the Gateway skill routes in `app/gateway/routers/skills.py` and `skills/installer.py`), where `attach_tracing=True` is the correct wiring — per the Tracing System section of `backend/AGENTS.md`, "Standalone callers — anything that invokes a model outside such a graph (e.g. `MemoryUpdater`) — keep `create_chat_model`'s default `attach_tracing=True`". A single hard-coded value cannot satisfy both callers, so the flag becomes the caller's to set.

## What changed

`skills/security_scanner.py` — `scan_skill_content` takes an `attach_tracing: bool = True` parameter and forwards it to `create_chat_model`. `tools/skill_manage_tool.py` — `_scan_or_raise`, the single in-graph choke point, passes `attach_tracing=False`. Standalone callers keep the default, so their wiring is unchanged.

`agents/lead_agent/agent.py` — the INVARIANT list said "four current sites" and asks that new in-graph calls be added to it, so this records the fifth: `scan_skill_content` as dual-use, with `_scan_or_raise` named as the in-graph choke point that passes the flag. Without that, a future audit of the four listed sites would not find this one.

Enumerated rather than assumed, since the design above depends on `_scan_or_raise` being the *only* in-graph caller. `scan_skill_content` has four non-test call sites: `routers/skills.py:235`/`:326` and `installer.py:232` (standalone), `skill_manage_tool.py:67` (in-graph). The installer is the one that could have broken the claim — `skill_manage_tool` does import skill storage, whose `ainstall_skill_from_archive` reaches `installer.py:232` — but the tool only uses storage for write/delete/validate/append_history and never installs; install's only callers are the Gateway route and `client.py:1264`, both standalone.

The `ainvoke` config is deliberately untouched: on the in-graph path the ambient config already carries the root's callbacks and Langfuse metadata, and `ensure_config` merges it in, so the flag was the only defect there.

> An earlier revision of this PR derived the flag by probing `langgraph.config.get_config()` instead. Both work, but the explicit parameter matches the convention the INVARIANT declares, keeps that list auditable by grep rather than adding a self-detecting exception to it, and drops the latent assumption that a runnable config being present implies the root attached tracing. Thanks to @willem-bd for pushing on it.

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [x] **Agents / LangGraph** — `packages/harness/deerflow/agents/lead_agent/agent.py` (INVARIANT docstring only, no behavior)
- [ ] **Sandbox**
- [x] **Skills** — `packages/harness/deerflow/skills/security_scanner.py`, `packages/harness/deerflow/tools/skill_manage_tool.py`
- [ ] **Dependencies**
- [x] **Docs / tests / CI only** — `tests/test_security_scanner.py`, `tests/test_skill_manage_tool.py`

## Bug fix verification

- Test paths:
  - `tests/test_skill_manage_tool.py::test_scan_or_raise_does_not_attach_model_tracing` — drives the **real** `scan_skill_content` rather than stubbing it, pinning the flag all the way to the model factory.
  - `tests/test_security_scanner.py::test_scan_skill_content_forwards_attach_tracing_to_the_model`
  - `tests/test_security_scanner.py::test_scan_skill_content_attaches_model_tracing_by_default`
- Red on `main` / green on this branch? **yes** — on unfixed main the in-graph path never passes the kwarg, so the choke-point test fails; it passes after the change.
- Red check was **per-test**, one mutation per clause, and every new test has teeth:

  | mutation | result |
  |---|---|
  | `_scan_or_raise` stops passing the flag (reverts the actual fix) | `test_scan_or_raise_does_not_attach_model_tracing` **FAILED** |
  | `scan_skill_content` ignores the param, hard-codes `True` | that test **+** `test_..._forwards_attach_tracing_to_the_model` — **2 failed** |
  | over-narrowed fix: hard-code `attach_tracing=False` | `test_..._attaches_model_tracing_by_default` **FAILED** |

  The last one is the other direction of the change: narrowing the fix into an unconditional `False` would silently drop the standalone callers' spans, and that test goes red if it happens.
- The pre-existing `test_scan_skill_content_passes_run_name_to_model` — which asserts the exact `config == {"run_name": "security_agent"}` — still passes **unmodified**, because the invoke config is not touched.

## Validation

```
cd backend
make format   # All checks passed! / 870 files left unchanged
make lint     # All checks passed! / 870 files already formatted

PYTHONPATH=. uv run pytest tests/test_security_scanner.py tests/test_skill_manage_tool.py -q
# 28 passed

PYTHONPATH=. uv run pytest tests/ -q
# 8 failed, 7680 passed, 28 skipped in 121.03s
```

The 8 failures are pre-existing and unrelated (`test_browserless_client.py`, `test_crawl4ai_tools.py`, `test_fastcrw_tools.py` — `web_fetch` against a blocked metadata address in this environment). The identical 8 fail with this change reverted, so this branch adds **0** new failures.

E2E through a real compiled `StateGraph` → real `ToolNode` → real `skill_manage` tool → real `_scan_or_raise` → real `scan_skill_content`. Only the lead model (to avoid needing an API key) and `create_chat_model` (to observe the kwargs it is given) are stubbed:

| | this branch | fix reverted (main's shape) |
|---|---|---|
| in-graph: `attach_tracing` reaching the factory | `False` | **absent → factory default `True`** |
| tool result | `Created custom skill 'e2e-skill'.` | same |
| skill actually written to disk | yes | yes |

That run drives the choke-point claim rather than reading a call table: the in-graph tool really does reach the scanner through `_scan_or_raise`, and the fix really does fire there in production shape.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Used an AI coding assistant to check the in-graph `create_chat_model` sites against the tracing INVARIANT in `agents/lead_agent/agent.py`, enumerate `scan_skill_content`'s call sites, write the regression tests, and run the red→green, per-test red check, full-suite comparison against clean `main`, and the real-graph E2E.


backend/packages/harness/deerflow/skills/security_scanner.py
```diff
@@ -103,7 +123,8 @@ async def scan_skill_content(
     try:
         config = app_config or get_app_config()
         model_name = config.skill_evolution.moderation_model_name
-        model = create_chat_model(name=model_name, thinking_enabled=False, app_config=config) if model_name else create_chat_model(thinking_enabled=False, app_config=config)
+        attach_tracing = not _inherits_graph_tracing()
```
willem-bd: Design consideration (non-blocking, maintainer's call): the invariant in `agents/lead_agent/agent.py` says new in-graph `create_chat_model` sites should *pass* `attach_tracing=False` explicitly - the other four sites (bootstrap, default, summarization, TitleMiddleware) all do. This dynamic detection works and is arguably more future-proof for a dual-use function, but the explicit alternative would match the convention: add an `attach_tracing: bool = True` param to `scan_skill_content` and have `_scan_or_raise` (skill_manage_tool.py:67 - unambiguously in-graph, the single choke point) pass `False`. That covers all five internal call sites, Gateway/installer keep the default `True`, and it removes the latent assumption that 'runnable config present ⟺ graph root attached tracing' (which would silently drop all spans if a standalone caller ever wrapped its call in a runnable without attaching at that root - no such caller today). Either is fine; flagging for your call.

backend/packages/harness/deerflow/skills/security_scanner.py
```diff
@@ -16,6 +18,24 @@
 logger = logging.getLogger(__name__)
 
 
+def _inherits_graph_tracing() -> bool:
```
willem-bd: Doc sync: this adds a fifth in-graph `create_chat_model` site, but the INVARIANT docstring in `agents/lead_agent/agent.py:16-18` still lists four and instructs new sites to 'add to this list and pass the flag.' Per the repo's keep-docs-in-sync convention, the invariant should note this site and its dynamic detection - otherwise a future audit of the four listed sites won't find `scan_skill_content`, and a refactor could silently regress it. (Applies regardless of whether you keep the dynamic approach or switch to the explicit param - either way the invariant is currently incomplete.)

backend/packages/harness/deerflow/skills/security_scanner.py
```diff
@@ -103,7 +123,8 @@ async def scan_skill_content(
     try:
         config = app_config or get_app_config()
         model_name = config.skill_evolution.moderation_model_name
-        model = create_chat_model(name=model_name, thinking_enabled=False, app_config=config) if model_name else create_chat_model(thinking_enabled=False, app_config=config)
+        attach_tracing = not _inherits_graph_tracing()
+        model = create_chat_model(name=model_name, thinking_enabled=False, app_config=config, attach_tracing=attach_tracing) if model_name else create_chat_model(thinking_enabled=False, app_config=config, attach_tracing=attach_tracing)
```
willem-bd: Minor readability: this ternary was already long and grew with the added kwarg. Extracting the shared kwargs helps, e.g. `kwargs = {'thinking_enabled': False, 'app_config': config, 'attach_tracing': attach_tracing}` then `model = create_chat_model(name=model_name, **kwargs) if model_name else create_chat_model(**kwargs)`. Lint/format pass as-is, so purely optional. (Don't collapse to a single `create_chat_model(name=model_name, **kwargs)` - `model_name` could be a falsy non-None value like `''`, and the current guard is defensively correct.)

backend/packages/harness/deerflow/skills/security_scanner.py
```diff
@@ -16,6 +18,24 @@
...
+    and standalone (Gateway skill routes, ``skills/installer.py``), and the two
+    need opposite wiring: in-graph must not attach at the model — the graph root
+    already did, and double-attaching emits duplicate spans *and* blocks the
+    Langfuse handler's ``propagate_attributes`` path (see the tracing INVARIANT in
+    ``agents/lead_agent/agent.py``) — while a standalone caller has no root to
+    inherit from and keeps model-level attachment.
+    """
+    try:
+        get_config()
+    except RuntimeError:
```
willem-bd: Minor consistency note: this catches `RuntimeError` while `TitleMiddleware._get_runnable_config` (title_middleware.py:180) catches broad `Exception`. The `RuntimeError` choice is correct per `langgraph.config.get_config` (which raises exactly `RuntimeError` outside a runnable context) and arguably better - just flagging the inconsistency in case you want the two probes to match.

willem-bd (commented):
Review comments (non-blocking) from code review - overall approve. The fix is correct: I verified against the langgraph source that get_config() reads the same contextvar the test sets and that real graph/ToolNode execution populates, and that scan_skill_content was the only in-graph site missing the flag (goal.py/oneshot_llm/MemoryUpdater are all correctly standalone). Comment 1 is a design discussion (dynamic detection vs the invariant's explicit-flag convention); the rest are minor.

backend/packages/harness/deerflow/skills/security_scanner.py
AnnaSuSu: Switched to the explicit param — I think you're right, but let me show the check I ran first, because your suggestion rests on `_scan_or_raise` being the *only* in-graph call site and I didn't want to take that on faith.

Four non-test callers: `routers/skills.py:235`/`:326` and `installer.py:232` (standalone), `skill_manage_tool.py:67` (in-graph). The one that could have falsified it is the installer — `skill_manage_tool` *does* import skill storage, and storage's `ainstall_skill_from_archive` reaches `installer.py:232`. But the tool only ever uses storage for write/delete/validate/append_history; it never installs. Install's only callers are the Gateway route and `client.py:1264`, both standalone. So your claim holds, and both designs are correct today.

I switched anyway, for the reason your comment implies: the INVARIANT's value is that it's *auditable* — a list of sites you can grep for the flag. A fifth site that works by self-detection is a permanent exception to that. Explicit also puts the decision where the knowledge is (the caller knows whether it's in-graph; the callee can only guess), and it removes the `runnable config present ⟺ root attached tracing` biconditional you flagged. I did go looking for a live counterexample to that — `goal.py` looked closest, since AGENTS.md describes it as running "after the graph root's tracing scope has already closed" — but it's dispatched from the run worker, outside any runnable, so it's consistent with your "no such caller today".

The tradeoff I'm accepting: a future in-graph caller now has to remember the flag. That's exactly what the INVARIANT list in your next comment is for, so I've updated it.

One argument I'd expected to make and dropped after checking: I assumed pulling `langgraph.config` into `skills/` was a layering smell, but `skills/describe.py:20` already imports `langgraph.types`. It isn't one.

backend/packages/harness/deerflow/skills/security_scanner.py
AnnaSuSu: Fixed — and it's the sharper version of the point: the INVARIANT is the rule this PR cites as its authority, so adding a fifth site without listing it would be the same omission the PR exists to fix. Updated to five sites, naming `scan_skill_content` as dual-use and `_scan_or_raise` as the in-graph choke point that passes the flag.

backend/packages/harness/deerflow/skills/security_scanner.py
AnnaSuSu: Applied, in the shape you suggested. I kept the `if model_name` guard rather than collapsing to a single call, per your parenthetical.

backend/packages/harness/deerflow/skills/security_scanner.py
AnnaSuSu: Moot now — the probe is gone with the switch to the explicit param, so there's no catch left to be inconsistent with.

For the record, your reading was right: I checked the installed `langgraph.config.get_config` and it raises exactly `RuntimeError` outside a runnable context. The two probes also weren't quite the same shape — `TitleMiddleware` is building a config dict, where degrading to `parent = {}` is benign, whereas mine returned a bool that decided the wiring, so swallowing anything broader would have silently selected the wrong branch.

WillemJiang (approved):

PR #4255, opened 2026-07-17
feat(auth): add keep me signed in login option

### LINKED ISSUE #4194: [feat] 管理页面添加记住密码功能
### Before you start

### Problem / motivation

不用每次登陆都填写账号密码

### Proposed solution

添加记住密码勾选框

### Affected area(s)

Frontend (UI / Next.js)

### Alternatives considered

_No response_

### Additional context

_No response_

18062706139fcz:
## Why

Fixes #4194.

The request is to avoid typing credentials on every login. This should be implemented as session persistence, not as browser-side password storage.

## What

- Adds a `Keep me signed in` option to the login page.
- Sends `remember_me=true|false` through the existing `/api/v1/auth/login/local` form flow.
- Adds a centralized Gateway session cookie policy:
  - HTTPS / trusted forwarded HTTPS: persistent `Secure` cookies.
  - `localhost` / loopback HTTP: persistent cookies for local development.
  - public HTTP / temporary sandbox URLs: session cookies by default.
  - explicit operator opt-in can allow insecure persistent cookies.
- Mirrors the final `access_token` lifetime onto `csrf_token` so the cookie pair expires together.
- Stores only the remembered email in frontend `localStorage`; password and tokens are never stored there.
- Adds a small `HttpOnly` session-preference cookie so password change / admin setup / OIDC token re-issue preserves the user's remember choice.
- Adds the same keep-signed-in choice to first-admin setup.
- Clears `access_token`, `csrf_token`, and the session-preference cookie on logout.

## Implementation Flow

```mermaid
flowchart TD
    A["LoginPage"] --> B["Submit email/password"]
    A --> A1["Keep me signed in checkbox"]
    A1 --> B1["remember_me form field"]
    B --> C["/api/v1/auth/login/local"]
    B1 --> C
    C --> D["LocalAuthProvider verifies password"]
    D --> E["create_access_token"]
    E --> F["SessionCookiePolicy"]
    F --> G{"Deployment context"}
    G -->|HTTPS or trusted forwarded HTTPS| H["Persistent Secure cookie"]
    G -->|localhost HTTP| I["Persistent dev cookie"]
    G -->|public HTTP / sandbox URL| J["Session cookie"]
    H --> K["Set HttpOnly access_token"]
    I --> K
    J --> K
    K --> L["Stamp final max_age on request.state"]
    L --> M["CSRFMiddleware sets csrf_token with same max_age"]
    M --> N["AuthProvider / fetchWithAuth continue existing flow"]

    classDef added fill:#E8F5E9,stroke:#2E7D32,stroke-width:2px,color:#1B5E20;
    classDef existing fill:#F5F5F5,stroke:#BDBDBD,color:#424242;
    class A,B,C,D,E,K,N existing;
    class A1,B1,F,G,H,I,J,L,M added;
```

## Impact Surface

```mermaid
flowchart LR
    Change["Remember login change"] --> FE["Frontend auth UI"]
    Change --> GW["Gateway auth routes"]
    Change --> CSRF["CSRFMiddleware"]
    Change --> Deploy["Deployment modes"]
    Change --> Tests["Regression tests"]

    FE --> FE1["Login checkbox"]
    FE --> FE2["Remembered email only"]
    FE --> FE3["No password/token storage"]

    GW --> GW1["remember_me parsing"]
    GW --> GW2["SessionCookiePolicy"]
    GW --> GW3["Logout clears both cookies"]

    CSRF --> CSRF1["csrf_token lifetime mirrors access_token"]

    Deploy --> D1["HTTPS persistent"]
    Deploy --> D2["localhost persistent"]
    Deploy --> D3["public HTTP session-only"]

    Tests --> T1["Cookie policy branches"]
    Tests --> T2["CSRF lockstep"]
    Tests --> T3["Frontend storage helper"]

    classDef added fill:#E8F5E9,stroke:#2E7D32,stroke-width:2px,color:#1B5E20;
    classDef area fill:#E3F2FD,stroke:#1565C0,color:#0D47A1;
    classDef existing fill:#F5F5F5,stroke:#BDBDBD,color:#424242;
    class Change area;
    class FE,GW,CSRF,Deploy,Tests existing;
    class FE1,FE2,FE3,GW1,GW2,GW3,CSRF1,D1,D2,D3,T1,T2,T3 added;
```

## Validation

- `UV_PROJECT_ENVIRONMENT=/tmp/deerflow-issue-4194-venv PYTHONPATH=. uv run pytest tests/test_auth_type_system.py tests/test_csrf_middleware.py tests/test_oidc_auth.py -q`
  - `98 passed`
- `RUFF_CACHE_DIR=/tmp/deerflow-issue-4194-ruff-cache UV_PROJECT_ENVIRONMENT=/tmp/deerflow-issue-4194-venv uv run ruff format --check app/gateway/auth/session_cookie.py app/gateway/auth/session_cookie_state.py app/gateway/csrf_middleware.py app/gateway/routers/auth.py tests/test_auth_type_system.py`
- `RUFF_CACHE_DIR=/tmp/deerflow-issue-4194-ruff-cache UV_PROJECT_ENVIRONMENT=/tmp/deerflow-issue-4194-venv uv run ruff check app/gateway/auth/session_cookie.py app/gateway/auth/session_cookie_state.py app/gateway/csrf_middleware.py app/gateway/routers/auth.py tests/test_auth_type_system.py`
- `source ~/.nvm/nvm.sh && nvm use 22.21.1 && pnpm test tests/unit/core/auth/remember-login.test.ts`
  - `4 passed`
- `source ~/.nvm/nvm.sh && nvm use 22.21.1 && pnpm lint && pnpm exec tsc --noEmit --tsBuildInfoFile /tmp/deerflow-issue-4194-tsconfig.tsbuildinfo`
- Manual/browser check on a production-mode local frontend:
  - login page shows `Keep me signed in`.
  - `remember_me=true` produces persistent `access_token` and `csrf_token`.
  - `remember_me=false` produces session `access_token` and `csrf_token`.
  - First-admin setup with keep-signed-in disabled creates session cookies, and password change keeps them session-only.


backend/app/gateway/routers/auth.py
```diff
@@ -149,18 +151,9 @@ class MessageResponse(BaseModel):
 # ── Helpers ───────────────────────────────────────────────────────────────
 
 
-def _set_session_cookie(response: Response, token: str, request: Request) -> None:
+def _set_session_cookie(response: Response, token: str, request: Request, *, remember_me: bool = True) -> None:
```
willem-bd: This helper's `remember_me` defaults to `True`, so the three non-login call sites — password change (`change-password`), `initialize`, and the OIDC callback — all silently issue a *persistent* cookie.

A user who unchecked "Keep me signed in" at login (session cookie) gets upgraded to a 7-day persistent cookie the moment they change their password or finish admin setup, because the server keeps no record of the original preference.

Options:
- Thread `remember_me` through these re-issue endpoints (accept it on change-password / initialize, or read it from OIDC params), or
- Persist the preference server-side (user record / session) so re-issue honors the original choice.

At minimum this limitation should be called out in the plan / docs.

backend/app/gateway/auth/session_cookie.py
```diff
@@ -0,0 +1,81 @@
...
+        host = urlsplit(origin).hostname
+        if host:
+            return host.lower()
+
+    if request.url.hostname:
+        return request.url.hostname.lower()
+    return ""
+
+
+def is_local_browser_origin(request: Request) -> bool:
```
willem-bd, line 45: `is_local_browser_origin` derives the host via `_request_hostname` -> `_request_origin`, which unconditionally trusts `Forwarded` / `X-Forwarded-Host` (the same trust model `is_secure_request` already uses for `X-Forwarded-Proto`).

Behind the documented nginx deployment this is fine — nginx overwrites client-supplied forwarded headers — but a direct-to-Gateway public-HTTP deployment would let a client classify itself as `localhost` (e.g. `X-Forwarded-Host: localhost`) and obtain a persistent, non-`Secure` cookie over plain HTTP.

Incremental risk is low (self-inflicted persistence, not privilege escalation), but worth either confirming nginx strips client-supplied `X-Forwarded-*`, or gating forwarded-header trust behind an explicit trusted-proxy config. A one-line note in the README/AGENTS.md auth section would help operators running without the bundled nginx.

backend/app/gateway/routers/auth.py
```diff
@@ -552,17 +549,18 @@ async def close_oidc_service() -> None:
 def _set_csrf_cookie(response: Response, request: Request) -> None:
     """Set the CSRF double-submit cookie (needed for GET-based OIDC callback)."""
     csrf_token = generate_csrf_token()
-    is_https = is_secure_request(request)
+    policy = resolve_session_cookie_policy(request)
```
willem-bd: This `_set_csrf_cookie` path (used by the GET OIDC callback, since the CSRF middleware only sets the cookie on POST) is modified to mirror the session max_age, but no test exercises it — the lockstep tests all go through the middleware POST branch. Given OIDC is a `risk:high` auth surface, please add a test asserting `access_token`/`csrf_token` `max_age` and `secure` parity on the OIDC callback response.

Minor: `resolve_session_cookie_policy(request)` here is called only to get `policy.secure` (which equals `is_secure_request(request)`) and a fallback `policy.max_age` that's never used in practice — `_set_session_cookie` on the preceding line always stamps `request.state`. Could simplify to `is_secure_request(request)` + the stamped state. Related: the middleware sets `secure=is_https` (csrf_middleware.py:232) while `set_session_cookie` sets `secure=policy.secure` — equal today, but two sources of truth that could diverge if the policy changes.

backend/app/gateway/csrf_middleware.py
```diff
@@ -230,7 +236,7 @@ async def dispatch(self, request: Request, call_next: Callable[[Request], Awaita
                 # evicted when iOS Safari terminates a home-screen PWA while the persistent
                 # access_token survives — leaving the user "logged in" but unable to make
                 # any state-changing request (403 "CSRF token missing").
-                max_age=get_auth_config().token_expiry_days * 24 * 3600 if is_https else None,
+                max_age=session_max_age if session_cookie_issued else (get_auth_config().token_expiry_days * 24 * 3600 if is_https else None),
```
willem-bd: This nested ternary is the riskiest line in the PR, and the `session_cookie_issued=False` fallback branch (e.g. a *failed* login, where `_set_session_cookie` is never reached) isn't covered by any test. Add a failed-login case asserting the CSRF cookie still issues with the expected fallback `max_age`.

Readability: the double ternary is hard to scan. Consider extracting a small helper (e.g. `_csrf_max_age(request, state)`) for symmetry with `resolve_session_cookie_policy` — it also makes the fallback unit-testable in isolation.

backend/app/gateway/auth/session_cookie.py
```diff
@@ -0,0 +1,81 @@
...
+ALLOW_INSECURE_PERSISTENT_COOKIE_ENV = "DEER_FLOW_AUTH_ALLOW_INSECURE_PERSISTENT_COOKIE"
+
+
+@dataclass(frozen=True)
+class SessionCookiePolicy:
+    """Resolved cookie settings for a session-creating auth response."""
+
+    secure: bool
+    max_age: int | None
+    reason: str
```
willem-bd, line 31: `reason` is populated on every branch but never read or logged anywhere. Either emit it at debug level (genuinely useful for diagnosing why a sandbox deployment got session-only cookies) or drop it to keep the dataclass lean.

willem-bd (commented):
Reviewed the session-persistence change. The lockstep `request.state` design for `access_token`/`csrf_token` max_age is sound, and the credential posture (passwords/tokens never in `localStorage`, only email) is correct. Detailed findings are inline; the more actionable ones:

- Re-issue flows (password change, initialize, OIDC) default to `remember_me=True` and silently upgrade session-only users to persistent cookies.
- OIDC `_set_csrf_cookie` and the CSRF middleware fallback branch are modified but untested.
- `is_local_browser_origin` extends the existing unconditional forwarded-header trust to host detection — fine behind nginx, worth a note for direct-to-Gateway deployments.

Minor (no clean inline anchor): the remembered email in `localStorage` isn't cleared on logout, so it stays pre-filled on shared machines — acceptable for "remember email" intent, but worth a deliberate decision. The long single-line import (csrf_middleware.py:18) and ternary pass `ruff format` but hurt readability.

backend/app/gateway/auth/session_cookie.py
```diff
@@ -0,0 +1,106 @@
...
+        host = urlsplit(origin).hostname
+        if host:
+            return host.lower()
+
+    if request.url.hostname:
+        return request.url.hostname.lower()
+    return ""
+
+
+def is_local_browser_origin(request: Request) -> bool:
```
willem-bd, line 45: ⚠️ Security: this localhost classification trusts a client-controllable header, which bypasses the PR's "public HTTP -> session cookie" degradation.

`is_local_browser_origin` resolves the hostname via `_request_origin()` (csrf_middleware.py), which prefers `Forwarded` host / `X-Forwarded-Host` over the direct `Host` header. Neither is a forbidden browser header, so same-origin JS or any header-injecting MITM can send `X-Forwarded-Host: localhost` against a public-HTTP sandbox URL -> `localhost_persistent` -> a **persistent, non-`Secure` `access_token` cookie over plain HTTP**, exactly what the degradation is meant to prevent.

This works even behind the bundled nginx: `docker/nginx/nginx.conf` only sets `X-Forwarded-Proto` (and even that preserves a client-supplied `https` via the `map`); it does **not** set or strip `X-Forwarded-Host` or `Forwarded`, so they're forwarded verbatim. That contradicts the new docs ("the bundled nginx overwrites them" in backend/AGENTS.md, and the README implication that nginx prevents spoofing).

Impact is bounded (on public HTTP the JWT already travels in cleartext and is valid for `token_expiry_days` regardless of cookie persistence, so this is defense-in-depth, not fresh disclosure), but it does defeat the PR's stated mitigation.

Suggested fixes (any combination):
- Base `is_local_browser_origin` on the direct `Host` / `request.url.hostname` (browser-controlled; the `localhost:2026` dev case still matches since nginx passes `Host $http_host`) rather than the forwarded host.
- Have `nginx.conf` explicitly overwrite `X-Forwarded-Host`/`Forwarded`.
- Treat localhost persistence as an explicit operator opt-in (like `DEER_FLOW_AUTH_ALLOW_INSECURE_PERSISTENT_COOKIE`) instead of auto-detected from a forwarded header.
- Correct the README/AGENTS wording (nginx only normalizes `X-Forwarded-Proto`).

backend/app/gateway/routers/auth.py
```diff
@@ -398,7 +397,7 @@ async def change_password(request: Request, response: Response, body: ChangePass
 
     # Re-issue cookie with new token_version
     token = create_access_token(str(user.id), token_version=user.token_version)
-    _set_session_cookie(response, token, request)
+    _set_session_cookie(response, token, request, remember_me=body.remember_me)
```
willem-bd, line 401: Nit / latent fragility: this re-issues `access_token` with a new lifetime, but `change_password` is **not** in `_AUTH_EXEMPT_PATHS`, so `CSRFMiddleware`'s `_is_auth and POST` branch doesn't fire here and `csrf_token` is **not** re-issued. The lifetime stamped on `request.state` by `set_session_cookie` is therefore dead on this path, and `csrf_token` keeps its original login-time lifetime.

Today this is benign: the frontend change-password flow (`setup/page.tsx` `handleChangePassword`) doesn't send `remember_me`, so `body.remember_me=None` -> reads the preference cookie -> resolves to the same `remember`/scheme as login -> access lifetime unchanged -> csrf still matches. But there's no test guarding the lockstep across change-password, and the moment a `remember_me` toggle is added to a change-password UI (or the scheme shifts between login and password change), access/csrf lifetimes can diverge.

Consider re-issuing `csrf_token` here (call `_set_csrf_cookie`) or adding a test that asserts lockstep across change-password.

frontend/src/app/(auth)/setup/page.tsx
```diff
@@ -221,6 +223,23 @@ export default function SetupPage() {
...
+            <label className="text-muted-foreground flex items-start gap-2 text-sm">
+              <input
+                type="checkbox"
+                checked={rememberMe}
+                onChange={(e) => setRememberMe(e.currentTarget.checked)}
+                className="border-input mt-1 h-4 w-4 rounded"
+              />
+              <span>
+                <span className="text-foreground block font-medium">
+                  Keep me signed in
```
willem-bd: Nit: this label and its description are hardcoded English, while `login/page.tsx` correctly uses the new `t.login.rememberMe` / `t.login.rememberMeDescription` i18n keys (added to `en-US`/`zh-CN`/`types.ts`). The rest of the setup page doesn't use i18n either, so this is consistent with the existing surface rather than a regression -- but it means the new translations don't cover first-admin setup. Consider reusing the same keys here (or routing the setup page through i18n in a follow-up) so the `zh-CN` preference is honored on setup too.

backend/app/gateway/auth/session_cookie.py
18062706139fcz: Fixed in 54ed6702. Localhost persistence now reads only request.url.hostname (the direct Host) and ignores Forwarded/X-Forwarded-Host. Regression coverage includes both spoofed header forms on a public HTTP host, and README/AGENTS now describe the actual nginx forwarding behavior.

backend/app/gateway/routers/auth.py
18062706139fcz: Fixed in 54ed6702. change_password now calls _set_csrf_cookie immediately after stamping the access-cookie policy. The regression test flips a session-only login to remember_me=true during password change and asserts secure/max-age parity for access_token and csrf_token.

frontend/src/app/(auth)/setup/page.tsx
18062706139fcz: Fixed in 54ed6702. Login and first-admin setup now share RememberSessionOption, which renders the existing t.login.rememberMe keys. A focused unit test verifies the zh-CN copy.

willem-bd (commented):
Low-severity findings from a code review (3 inline comments). None block merge; all have safe fallbacks or are polish.

frontend/src/app/(auth)/setup/page.tsx
```diff
@@ -221,6 +224,10 @@ export default function SetupPage() {
                 minLength={8}
               />
             </div>
+            <RememberSessionOption
```
willem-bd, line 231: This `RememberSessionOption` is rendered only in the init-admin form. The change-password form (`handleChangePassword` -> `POST /api/v1/auth/change-password`) omits the checkbox and does not send `remember_me` in the body, so a user completing first-boot setup via the change-password screen cannot toggle persistence there.

The backend handles `remember_me=None` gracefully (falls back to the `deerflow_session_persistent` preference cookie, default `true`), so this is a UX gap, not a correctness bug. Consider adding the checkbox to the change-password form and sending `remember_me`, or a one-line note that change-password intentionally inherits the prior preference.

backend/app/gateway/routers/auth.py
```diff
@@ -295,6 +289,7 @@ async def login_local(
     request: Request,
     response: Response,
     form_data: OAuth2PasswordRequestForm = Depends(),
+    remember_me: bool = Form(default=True),
```
willem-bd, line 293: `login_local` now accepts `remember_me`, but the `register` endpoint does not - `RegisterRequest` has no such field and `register()` calls `_set_session_cookie(response, token, request)` without it. A newly registering user therefore always gets the preference-cookie default (`true`) -> a persistent cookie on HTTPS/localhost, with no user choice. The login page also hides the checkbox in register mode and does not send the field.

Reasonable default, but slightly inconsistent with the "user chooses" framing. Consider accepting `remember_me` on register too, or documenting the default.

backend/app/gateway/auth/session_cookie.py
```diff
@@ -0,0 +1,100 @@
...
+    """Return the direct request host without trusting forwarded host headers."""
+    if request.url.hostname:
+        return request.url.hostname.lower()
+    return ""
+
+
+def is_local_browser_origin(request: Request) -> bool:
+    """Return True for loopback browser origins where HTTP persistence is acceptable."""
+    host = _request_hostname(request)
+    return host in {"localhost", "127.0.0.1", "::1"} or host.endswith(".localhost")
```
willem-bd: `is_local_browser_origin` only matches `127.0.0.1` and `::1`, not the full loopback range `127.0.0.0/8`. A request to `http://127.1.2.3:2026` (valid loopback) would not be treated as local and would degrade to a session cookie.

Uncommon in practice; if you want to be precise, `ipaddress.ip_address(host).is_loopback` (guarded for non-IP hosts) covers the whole range. Very minor.

WillemJiang (approved):

PR #4259, opened 2026-07-17
fix(sandbox): sync DEERFLOW_USER_ID prefix ordering in bash_tool tests

wangzhengxi:
- Update test_channel_user_id_env.py assertions to match actual bash_tool behavior: DEERFLOW_USER_ID is only injected on the local-sandbox path, wrapping the already identity-prefixed command (appears first in the final command string).
- Add test coverage for local sandbox without channel_user_id to isolate the DEERFLOW_USER_ID prefix from the channel identity prefix.

Fixes #

## Surface area

## AI assistance

**Tool(s) used:** 

**How you used it:** 


wangzhengxi:
image


单元测试通过的文件

wangzhengxi:
我本地再重新把单元测试的文件都跑下。

fancyboi999 (changes requested):
Thanks @wangzhengxi. I found one issue that should be addressed before this is ready.

[P2] Keep the new user-id environment contract consistent across sandbox providers

- Location: `backend/packages/harness/deerflow/sandbox/tools.py:1731-1755`
- Problem: The PR introduces `DEERFLOW_USER_ID` as a new bash environment contract, but injects it only inside the `is_local_sandbox(runtime)` branch. Docker/AIO, E2B, and Boxlite executions take the non-local branch, so a skill or subprocess that works locally by reading this variable will silently lose it in deployed sandbox modes.
- Evidence: `resolve_runtime_user_id(runtime)` is computed for every invocation, but `_apply_user_env_prefix()` is called only for POSIX LocalSandbox. The new tests explicitly preserve the absence of `DEERFLOW_USER_ID` on the AIO path, while the helper itself claims the mechanism works uniformly for LocalSandbox and AioSandbox. This is a runtime behavior change, not merely a test-order correction.
- Suggested fix: If `DEERFLOW_USER_ID` is intended to be a supported contract, apply the shell-quoted prefix consistently to every POSIX sandbox execution path (or use an equivalent provider-safe environment channel) and document the contract. If that contract is not intended in this PR, remove the runtime injection and keep the change scoped to the actual test correction.
- Test: Add provider-parity coverage for LocalSandbox and at least the AIO/non-local command path, asserting the same authenticated user id is available without leaking into secret-bearing arguments; include a representative Docker/AIO smoke test if the contract is retained.

wangzhengxi:
@javenfang @

> Thanks @wangzhengxi. I found one issue that should be addressed before this is ready.
> 
> [P2] Keep the new user-id environment contract consistent across sandbox providers
> 
> * Location: `backend/packages/harness/deerflow/sandbox/tools.py:1731-1755`
> * Problem: The PR introduces `DEERFLOW_USER_ID` as a new bash environment contract, but injects it only inside the `is_local_sandbox(runtime)` branch. Docker/AIO, E2B, and Boxlite executions take the non-local branch, so a skill or subprocess that works locally by reading this variable will silently lose it in deployed sandbox modes.
> * Evidence: `resolve_runtime_user_id(runtime)` is computed for every invocation, but `_apply_user_env_prefix()` is called only for POSIX LocalSandbox. The new tests explicitly preserve the absence of `DEERFLOW_USER_ID` on the AIO path, while the helper itself claims the mechanism works uniformly for LocalSandbox and AioSandbox. This is a runtime behavior change, not merely a test-order correction.
> * Suggested fix: If `DEERFLOW_USER_ID` is intended to be a supported contract, apply the shell-quoted prefix consistently to every POSIX sandbox execution path (or use an equivalent provider-safe environment channel) and document the contract. If that contract is not intended in this PR, remove the runtime injection and keep the change scoped to the actual test correction.
> * Test: Add provider-parity coverage for LocalSandbox and at least the AIO/non-local command path, asserting the same authenticated user id is available without leaking into secret-bearing arguments; include a representative Docker/AIO smoke test if the contract is retained.

Thanks for the careful review! You're absolutely right — I missed the consistency across sandbox providers.

Here's my plan to fix it:

1. **Move `DEERFLOW_USER_ID` injection outside the `is_local_sandbox` branch**, so it applies uniformly to all sandbox providers (Local, AIO, Docker, E2B, Boxlite).
2. **Add provider-parity test** (`test_nonlocal_bash_forwards_user_id`) to verify non-local sandbox paths also receive the environment variable correctly.

Will push the fix shortly.

WillemJiang:
@wangzhengxi, did you have a chance to polish the code?

wangzhengxi:
> @wangzhengxi, did you have a chance to polish the code?

Apologies for the late reply, this slipped my mind. I'll get to it this weekend and clean it up.

PR #4260, opened 2026-07-17
feat(authz): add built-in RBAC provider and provider factory

hata33:
﻿## Summary

Phase 1A-2 of the RFC in #4063: **built-in RBAC provider + provider factory. No runtime behavior change** (`enabled: false` by default; no execution wiring).

Builds on Phase 1A-1 (PR #4203, trusted principal chain).

## Why

Phase 1A-1 established the trusted identity chain but had no policy engine. This PR adds the built-in `RbacAuthorizationProvider` and a unified `resolve_authorization_provider()` factory so that Phase 1B can wire a single provider instance into both Layer 1 (assembly filter) and Layer 2 (execution deny).

## What changed

**New files:**

| File | Purpose |
|------|---------|
| `deerflow/authz/rbac.py` | `RbacAuthorizationProvider`: config-compiled RBAC with deny-wins semantics |
| `deerflow/authz/runtime.py` | `resolve_authorization_provider(config)`: single factory entry point |
| `tests/test_rbac_authorization_provider.py` | 75 tests: allow/deny/wildcard/role/request-validation/immutability/filter-parity/malformed-config |
| `tests/test_authorization_runtime.py` | 15 tests: disabled/enabled/missing/invalid-path/Protocol/non-class/error-propagation/no-injection |

**Modified files:**

| File | Change |
|------|--------|
| `authz/__init__.py` | Export `RbacAuthorizationProvider` and `resolve_authorization_provider` |
| `docs/plans/2026-07-10-pluggable-authorization-implementation-notes.md` | Phase 1A-2 decision log and Phase 1B caveats |

## RBAC semantics

| Config | Behavior |
|--------|----------|
| `allow: *` or `allow: true` | Allow all candidates |
| `allow: [...]` | Only listed items |
| `allow: []` or `allow: false` | Deny all |
| `allow` missing | Default allow, deny still applies |
| `allow: null` | `ValueError` (not silent allow-all) |
| `deny: [...]` | Remove from allow set; **deny always wins** |
| `deny: null` | `ValueError` (not silent empty-deny) |
| Unknown/misspelled policy keys | `ValueError` at construction |
| Unknown/misspelled provider config keys | `ValueError` at construction |
| Reserved request aliases used as config keys | `ValueError` naming the required mapped key |
| Resource config missing | Unrestricted for that resource |
| Role missing/None | `ValueError` (not silent allow) |
| Role unknown | `ValueError` (not silent allow) |
| Invalid config types | `ValueError` at construction |

Resource names use explicit mapping (`tool -> tools`, `model -> models`, etc.); there is no pluralization guessing. Built-in policies are scoped by role, resource, and target. `AuthzRequest.action` is retained for protocol compatibility but is not a rule dimension in Phase 1A-2.

## Factory rules

- `enabled=false` -> `None` (no import attempted, even with a garbage class path)
- `enabled=true` with no provider -> `ValueError`
- Invalid class path, non-class target, or construction failure -> `ValueError` with path in message
- `resolve_variable(class_path, expected_type=type)`: only classes accepted
- `isinstance(instance, AuthorizationProvider)` check post-construction
- RBAC construction errors surface the class path and preserve `__cause__`
- No caching; no `fail_closed`/`default_role` injection

## What is NOT in this PR

- Layer 1 tool filtering (Phase 1B)
- Layer 2 middleware auto-wiring (Phase 1B)
- `config.example.yaml` RBAC example (Phase 1B; showing enabled RBAC before enforcement wiring would be misleading)
- `config_version` bump (no schema change)
- Lead/subagent/client modifications

## Validation

```bash
# New tests: 90 passed
cd backend && PYTHONPATH=. uv run pytest tests/test_rbac_authorization_provider.py tests/test_authorization_runtime.py -q
# 90 passed

# Related authorization/config/reflection tests: 161 passed
cd backend && PYTHONPATH=. uv run pytest tests/test_authorization_provider.py tests/test_authorization_principal.py tests/test_authorization_runtime.py tests/test_rbac_authorization_provider.py tests/test_auth_config.py tests/test_reflection_resolvers.py -q
# 161 passed

# Lint + format
cd backend && uv run ruff check packages/harness/deerflow/authz tests/test_rbac_authorization_provider.py tests/test_authorization_runtime.py
# All checks passed
cd backend && uv run ruff format --check packages/harness/deerflow/authz tests/test_rbac_authorization_provider.py tests/test_authorization_runtime.py
# All files formatted
```

## Related

- RFC: #4063
- Phase 0: #4127
- Phase 1A-1: #4203

### DIFF
--- a/backend/packages/harness/deerflow/authz/rbac.py
+++ b/backend/packages/harness/deerflow/authz/rbac.py
@@ -66,6 +66,10 @@ class RbacAuthorizationProvider:
     to ``{allow: ..., deny: [...]}`` policies. All validation happens at
     construction; the request path is pure membership checks.
 
+    Policies are scoped by role, resource, and target. ``AuthzRequest.action``
+    is accepted for protocol compatibility but is not a rule dimension in this
+    built-in provider.
+
     Example config::
 
         roles:
@@ -79,7 +83,11 @@ class RbacAuthorizationProvider:
 
     name = "rbac"
 
-    def __init__(self, *, roles: dict[str, Any], **_kwargs: Any) -> None:
+    def __init__(self, *, roles: dict[str, Any] | object = _ABSENT, **kwargs: Any) -> None:
+        if kwargs:
+            raise ValueError(f"unknown provider config keys {sorted(kwargs, key=repr)}; supported: ['roles']")
+        if roles is _ABSENT:
+            raise ValueError("missing required provider config key 'roles'")
         if not isinstance(roles, dict):
             raise ValueError(f"roles must be a dict, got {type(roles).__name__}")
 
--- a/backend/packages/harness/deerflow/authz/rbac.py
+++ b/backend/packages/harness/deerflow/authz/rbac.py
@@ -1,0 +1,219 @@
+"""Built-in RBAC authorization provider.
+
+Reads a role→resource policy from config and compiles it into immutable
+structures at construction time. Deny always wins over allow. Unknown or
+missing roles raise ``ValueError`` (not a silent allow) so that the execution
+layer's ``fail_closed`` can make the final decision.
+
+See ``docs/plans/2026-07-15-authz-phase1a-implementation-plan.md`` §3.3 for
+the full semantic table.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from deerflow.authz.provider import (
+    AuthzDecision,
+    AuthzReason,
+    AuthzRequest,
+    Principal,
+)
+
+# Explicit resource-type → config-key mapping. Prevents silent mis-lookup
+# when ``AuthzRequest.resource`` (singular, e.g. "tool") doesn't match the
+# config key (plural, e.g. "tools").
+_RESOURCE_POLICY_KEYS: dict[str, str] = {
+    "tool": "tools",
+    "model": "models",
+    "skill": "skills",
+    "sandbox": "sandbox",
+    "mcp_server": "mcp_servers",
+    "route": "routes",
+}
+
+_ALL = object()  # sentinel meaning "allow all candidates"
+_ABSENT = object()  # sentinel meaning "key not present in dict"
+
+
+# The only supported keys in a resource policy dict. Any other key (typos,
+# unknown fields) is rejected at construction to prevent silent mis-grants.
+_SUPPORTED_POLICY_KEYS: frozenset[str] = frozenset({"allow", "deny"})
+
+
+class _CompiledPolicy:
+    """Immutable, pre-validated policy for a single (role, resource_type) pair."""
+
+    __slots__ = ("allowed", "denied")
+
+    def __init__(self, *, allowed: frozenset[str] | object, denied: frozenset[str]):
+        self.allowed = allowed
+        self.denied = denied
+
+    def is_allowed(self, target: str) -> bool:
+        # Deny always wins.
+        if target in self.denied:
+            return False
+        if self.allowed is _ALL:
+            return True
+        return target in self.allowed
+
+
+class RbacAuthorizationProvider:
+    """Built-in role-based authorization provider.
+
+    Configured via ``roles`` mapping where each role maps resource-type keys
+    to ``{allow: ..., deny: [...]}`` policies. All validation happens at
+    construction; the request path is pure membership checks.
+
+    Example config::
+
+        roles:
+          admin:
+            tools: {allow: "*"}
+          user:
+            tools: {allow: "*", deny: ["update_agent"]}
+          guest:
+            tools: {allow: ["web_search", "read_file"]}
+    """
+
+    name = "rbac"
+
+    def __init__(self, *, roles: dict[str, Any], **_kwargs: Any) -> None:
+        if not isinstance(roles, dict):
+            raise ValueError(f"roles must be a dict, got {type(roles).__name__}")
+
+        # Compile all policies up front.
+        self._policies: dict[tuple[str, str], _CompiledPolicy] = {}
+        self._known_roles: frozenset[str] = frozenset(roles.keys())
+
+        for role_name, role_config in roles.items():
+            if not isinstance(role_name, str) or not role_name:
+                raise ValueError(f"role name must be a non-empty string, got {role_name!r}")
+            if not isinstance(role_config, dict):
+                raise ValueError(f"role '{role_name}' config must be a dict, got {type(role_config).__name__}")
+
+            for resource_key, resource_policy in role_config.items():
+                if not isinstance(resource_key, str) or not resource_key:
+                    raise ValueError(f"role '{role_name}' has invalid resource key {resource_key!r}")
+                if not isinstance(resource_policy, dict):
+                    raise ValueError(f"role '{role_name}' resource '{resource_key}' must be a dict, got {type(resource_policy).__name__}")
+
+                compiled = self._compile_resource_policy(role_name, resource_key, resource_policy)
+                self._policies[(role_name, resource_key)] = compiled
+
+    @staticmethod
+    def _compile_resource_policy(
+        role_name: str,
+        resource_key: str,
+        policy: dict[str, Any],
+    ) -> _CompiledPolicy:
+        """Validate and compile a single resource policy into immutable structures.
+
+        Distinguishes "key absent" (use default) from "key present but null"
+        (invalid — reject). Unknown keys (typos) are rejected to prevent
+        silent mis-grants.
+        """
+        # --- reject unknown keys (catch typos like "alow") ---
+        unknown_keys = set(policy.keys()) - _SUPPORTED_POLICY_KEYS
+        if unknown_keys:
+            raise ValueError(f"role '{role_name}' resource '{resource_key}': unknown policy keys {sorted(unknown_keys, key=repr)}; supported: {sorted(_SUPPORTED_POLICY_KEYS)}")
+
+        # --- allow ---
+        raw_allow = policy.get("allow", _ABSENT)
+        if raw_allow is _ABSENT:
+            allowed: frozenset[str] | object = _ALL  # missing allow = allow all (deny still applies)
+        elif raw_allow is None:
+            raise ValueError(f"role '{role_name}' resource '{resource_key}': allow must not be null; omit the key, or use '*' / bool / list")
+        elif raw_allow is True:
+            allowed = _ALL
+        elif raw_allow is False:
+            allowed = frozenset()  # allow: false = deny all
+        elif isinstance(raw_allow, str):
+            if raw_allow == "*":
+                allowed = _ALL
+            else:
+                raise ValueError(f"role '{role_name}' resource '{resource_key}': allow string must be '*', got {raw_allow!r}")
+        elif isinstance(raw_allow, (list, tuple)):
+            for item in raw_allow:
+                if not isinstance(item, str) or not item:
+                    raise ValueError(f"role '{role_name}' resource '{resource_key}': allow list contains non-string or empty item {item!r}")
+            allowed = frozenset(raw_allow)
+        else:
+            raise ValueError(f"role '{role_name}' resource '{resource_key}': allow must be '*', bool, or list of strings, got {type(raw_allow).__name__}")
+
+        # --- deny ---
+        raw_deny = policy.get("deny", _ABSENT)
+        if raw_deny is _ABSENT or raw_deny is None:
+            if raw_deny is None:
+                raise ValueError(f"role '{role_name}' resource '{resource_key}': deny must not be null; omit the key for no deny list")
+            denied: frozenset[str] = frozenset()
+        elif isinstance(raw_deny, (list, tuple)):
+            for item in raw_deny:
+                if not isinstance(item, str) or not item:
+                    raise ValueError(f"role '{role_name}' resource '{resource_key}': deny list contains non-string or empty item {item!r}")
+            denied = frozenset(raw_deny)
+        else:
+            raise ValueError(f"role '{role_name}' resource '{resource_key}': deny must be a list of strings, got {type(raw_deny).__name__}")
+
+        return _CompiledPolicy(allowed=allowed, denied=denied)
+
+    def _resolve_policy(self, principal: Principal, resource: str) -> _CompiledPolicy | None:
+        """Look up the compiled policy for (role, resource_type).
+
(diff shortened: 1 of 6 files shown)


fancyboi999 (changes requested):
Thanks @hata33. I found one issue that should be addressed before this is ready.

[P1] Reject malformed policy fields instead of silently opening access

- Location: `backend/packages/harness/deerflow/authz/rbac.py:107-130`
- Problem: `_compile_resource_policy()` uses `dict.get()` for both `allow` and `deny` and never rejects unknown policy keys. In YAML, `allow: null`, `deny: null`, or a typo such as `alow: [...]` is therefore treated like an omitted field: `allow` becomes `_ALL` and `deny` becomes empty. A malformed RBAC policy can silently grant unrestricted access instead of failing during provider construction.
- Evidence: The Phase 1A contract says only a missing `allow` defaults to allow-all and that illegal fields/types must raise `ValueError`. The current tests reject numeric invalid types and invalid list members, but do not cover explicit nulls or unknown keys, so this fail-open path remains green.
- Suggested fix: Validate that each resource policy contains only the supported `allow`/`deny` keys, use an explicit sentinel to distinguish an absent key from a present `None`, and reject explicit null or other unsupported values during construction.
- Test: Add construction tests for `allow: null`, `deny: null`, and unknown/misspelled keys, and assert the provider factory surfaces the class path while preserving the underlying validation error.

hata33:
@fancyboi999 Thanks — the P1 is fixed in `9224c3ec`. Additional hardening based on follow-up review:

**P1 (malformed policy fields):**
- `allow: null` and `deny: null` now raise `ValueError` (distinguished from absent key via `_ABSENT` sentinel).
- Unknown/misspelled keys (e.g. `alow`) raise `ValueError` at construction — validated against `_SUPPORTED_POLICY_KEYS`.
- 4 new construction tests: null allow, null deny, unknown key, unknown key alongside valid keys.

**Follow-up fixes:**
- `sorted(unknown_keys, key=repr)` — mixed-type keys (str + int) no longer trigger `TypeError`, always `ValueError`. Test: `test_mixed_type_unknown_keys_raises_value_error`.
- `resolve_variable(class_path, expected_type=type)` — non-class targets (functions, variables) rejected at resolution. Test: `test_non_class_target_raises`.
- Factory error propagation tests: `test_invalid_rbac_config_surfaces_class_path` + `test_invalid_rbac_config_preserves_cause` (class path in message, `__cause__` chained).
- Fixed typo in allow null error message.

**56 tests pass** (37 RBAC + 15 factory + 4 malformed-config). Lint/format clean.

willem-bd (commented):
Findings from re-review of Phase 1A-2 (built-in RBAC provider + factory). Overall clean and security-conscious; these are minor/Phase-1B notes.

backend/packages/harness/deerflow/authz/rbac.py
```diff
@@ -0,0 +1,219 @@
...
+            tools: {allow: "*"}
+          user:
+            tools: {allow: "*", deny: ["update_agent"]}
+          guest:
+            tools: {allow: ["web_search", "read_file"]}
+    """
+
+    name = "rbac"
+
+    def __init__(self, *, roles: dict[str, Any], **_kwargs: Any) -> None:
```
willem-bd: **`**_kwargs` silently absorbs unknown top-level provider-config keys.** Any key in `provider.config` other than `roles` is dropped by `**_kwargs`. A typo like `rolez:` produces a wrapped "missing required argument: roles" that does not name the offending key, and an extra key alongside a valid `roles` (e.g. `{roles: ..., bogus: true}`) is ignored without error. This is inconsistent with the reject-unknown-policy-keys stance (`_SUPPORTED_POLICY_KEYS`). For a security-sensitive config, consider rejecting unknown provider-config keys at construction (and add a test for an extra unknown top-level key).

backend/packages/harness/deerflow/authz/rbac.py
```diff
@@ -0,0 +1,219 @@
...
+                    message=f"role '{request.principal.role}' is denied '{request.target}' on resource '{request.resource}'",
+                )
+            ],
+            policy_id="rbac:deny",
+        )
+
+    async def aauthorize(self, request: AuthzRequest) -> AuthzDecision:
+        return self.authorize(request)
+
+    def filter_resources(
```
willem-bd, line 236: **`filter_resources` raises on an unknown/missing role - untested.** It calls `_resolve_policy`, which raises `ValueError` for an unknown/missing role, identical to `authorize`. For a batch *visibility* filter (e.g. a UI listing available tools), raising rather than returning `[]` surfaces as a 500 on the listing endpoint when a principal role is misconfigured. Consider whether the list path should degrade to deny-all instead, or at least document that it raises identically to `authorize`. No test currently covers `filter_resources` with an unknown/missing role.

backend/packages/harness/deerflow/authz/rbac.py
```diff
@@ -0,0 +1,219 @@
...
+        if role not in self._known_roles:
+            raise ValueError(f"Unknown role '{role}'; known roles: {sorted(self._known_roles)}")
+
+        resource_key = _RESOURCE_POLICY_KEYS.get(resource, resource)
+        return self._policies.get((role, resource_key))
+
+    def authorize(self, request: AuthzRequest) -> AuthzDecision:
+        """Evaluate a single authorization request."""
+        policy = self._resolve_policy(request.principal, request.resource)
+        if policy is None:
```
willem-bd, line 208: **Security note: a known role with no policy for a resource type is unrestricted, not fail-closed.** This `allow=True` / `policy_id="rbac:unrestricted"` path means an operator who configures RBAC for `tools` but forgets `models` / `mcp_servers` / `routes` grants unrestricted access to those. `fail_closed` only covers provider errors / unresolved identity, not missing policies. The behavior is documented in the PR table, but worth calling out in the Phase 1B `config.example.yaml` example / docs so operators know to enumerate every resource type they want restricted (or add an explicit `allow: []`).

backend/packages/harness/deerflow/authz/rbac.py
```diff
@@ -0,0 +1,219 @@
...
+        if role is None or role == "":
+            raise ValueError("Principal has no role; cannot evaluate RBAC policy")
+
+        if role not in self._known_roles:
+            raise ValueError(f"Unknown role '{role}'; known roles: {sorted(self._known_roles)}")
+
+        resource_key = _RESOURCE_POLICY_KEYS.get(resource, resource)
+        return self._policies.get((role, resource_key))
+
+    def authorize(self, request: AuthzRequest) -> AuthzDecision:
```
willem-bd, line 204: Two notes on `authorize`: (1) `request.action` is never read - decisions are role+resource+target only, so the built-in RBAC cannot express action-scoped rules (e.g. "allow listing a tool but not calling it"). Likely intentional for Phase 1A-2; a one-line note in the class docstring would set operator expectations. (2) `policy_id` is generic (`rbac:allow` / `rbac:deny` / `rbac:unrestricted`) - when Phase 1B wires audit logging, a more descriptive id (or `metadata` with role/resource/target) will aid traceability.

docs/plans/2026-07-10-pluggable-authorization-implementation-notes.md
```diff
@@ -200,6 +200,28 @@ Phase 1 最低验证要求：
...
+- **决策：** 未知角色和缺失角色抛 `ValueError`（不返回 allow），由执行层
+  根据 `fail_closed` 决定。
+- **决策：** 资源名使用显式映射（`tool → tools`，`model → models` 等），
+  不通过加 `s` 猜测。未知 resource 使用原名查找；未配置时视为"不受限"。
+- **决策：** `resolve_authorization_provider()` 是唯一 provider 解析入口。
+  disabled 时返回 `None`（不 import provider 模块）；enabled 但缺少 provider
+  时抛 `ValueError`。不缓存实例。不注入 `fail_closed` 或 `default_role`。
+- **决策：** 内置和自定义 provider 使用完全相同的 `resolve_variable` class-path
+  解析路径，无特殊分支。
+- **证据：** 48 tests passed（37 RBAC + 11 factory）。
```
willem-bd: The evidence count here is stale: "48 tests passed（37 RBAC + 11 factory）". The PR body and the follow-up comment say 56 (37 RBAC + 4 malformed-config + 15 factory). Update the count / split so the decision log matches the shipped tests.

fancyboi999 (changes requested):
Thanks @hata33. I found one issue that should be addressed before this is ready.

[P1] Reject reserved resource aliases that silently bypass their policy

- Location: `backend/packages/harness/deerflow/authz/rbac.py:104-111` and `backend/packages/harness/deerflow/authz/rbac.py:182-192`
- Problem: The constructor accepts `tool`, `model`, `skill`, `mcp_server`, and `route` as role resource keys, but `_resolve_policy()` always rewrites those request-side names to their plural config keys. A policy stored under one of these singular aliases can therefore never match. Because a missing policy returns `rbac:unrestricted`, an intuitive deny-all config such as `tool: {allow: []}` silently permits every tool.
- Evidence: On the current head, a real provider call with `roles={"user": {"tool": {"allow": []}}}` returns `allow=True` for `resource="tool", target="bash"`; changing only the key to `tools` returns `allow=False`. The current mapping tests cover the plural key and arbitrary custom resources, but not these reserved aliases. This is distinct from the documented choice to leave a genuinely absent resource policy unrestricted: here the operator supplied a policy, but the provider made it unreachable.
- Suggested fix: Reject config keys that collide with the request-side aliases in `_RESOURCE_POLICY_KEYS` and name the required mapped key in the error, or normalize them during construction while rejecting singular/plural duplicates. Keep truly custom resource keys supported.
- Test: Parameterize the reserved mappings and prove that a singular alias either fails construction or governs the corresponding request without falling through to unrestricted; also cover a config containing both alias forms.

Reviewed against fresh base `cd34a1a504ac41f3b581b02f1cd32a5917642b41` (merge base `9a4c72db99f6731608dfdadf9583d2de1a56d566`) at head `84f3113c6cfd1c4044fd59f1644e605445c10608`.

hata33:
﻿@fancyboi999 Thanks - fixed the reserved resource alias fail-open in `5415a0e3`.

**What changed**
- `RbacAuthorizationProvider` now derives reserved aliases from `_RESOURCE_POLICY_KEYS` during construction.
- Request-side aliases whose mapped config key differs (`tool`, `model`, `skill`, `mcp_server`, `route`) raise `ValueError` and name the required key (`tools`, `models`, `skills`, `mcp_servers`, `routes`).
- Same-name mappings such as `sandbox` remain valid, and custom resource keys remain supported.
- A config containing both `tool` and `tools` is rejected rather than allowing an unreachable policy to coexist.

**Regression coverage**
- Parameterized all five reserved alias mappings.
- Covered alias + mapped-key collision.
- Covered valid `sandbox` same-name mapping; the existing custom-resource test remains green.

**Validation**
- `66 passed` for the RBAC + factory suites.
- `137 passed` across authorization/provider/principal/config/reflection tests.
- Ruff check and format check pass.

backend/tests/test_authorization_runtime.py
```diff
@@ -0,0 +1,223 @@
...
+        try:
+            resolve_authorization_provider(config)
+        except ValueError as err:
+            assert err.__cause__ is not None
+            assert isinstance(err.__cause__, ValueError)
+        else:
+            pytest.fail("Expected ValueError for invalid RBAC config")
+
+
+class _IncompleteProvider:
```
willem-bd: **[Minor] `_IncompleteProvider` is defined but never used**

This helper (has `name`/`authorize`/`aauthorize` but no `filter_resources`, presumably to exercise the Protocol-conformance check) isn't referenced by any test - a grep finds only the class definition. Either add the intended test (construct it via `resolve_authorization_provider` and assert the `AuthorizationProvider Protocol` failure), or remove the class to keep the suite clean.

backend/packages/harness/deerflow/authz/rbac.py
```diff
@@ -0,0 +1,230 @@
...
+        policy = self._resolve_policy(request.principal, request.resource)
+        if policy is None:
+            # No policy for this role+resource → unrestricted.
+            return AuthzDecision(
+                allow=True,
+                reasons=[AuthzReason(code="authz.no_policy", message="no policy configured")],
+                policy_id="rbac:unrestricted",
+            )
+
+        if policy.is_allowed(request.target):
```
willem-bd: **[Minor] `request.target` is not validated**

`AuthzRequest.target` reaches `is_allowed` unchecked. Under a wildcard `allow` with no matching `deny`, `authorize(target=None)` / `authorize(target="")` returns *allow* (since `None not in self.denied` and `allowed is _ALL`); under a list `allow` it's denied. The inconsistency is benign today (callers always pass a real target), but a one-line guard in `authorize` (reject `None`/empty `target` with `ValueError`, matching the construction-time strictness) would make the safe behavior explicit and avoid a future surprise.

backend/tests/test_authorization_runtime.py
hata33: Fixed in `2c672ff3`.

Removed the unused `_IncompleteProvider`. Protocol rejection remains covered by `TestProtocolConformance.test_non_protocol_class_raises`, which resolves `builtins:dict` through the real factory and asserts the `AuthorizationProvider Protocol` failure.

backend/packages/harness/deerflow/authz/rbac.py
hata33: Fixed in `2c672ff3`.

`authorize()` now rejects `None` and empty targets with `ValueError` before either the unrestricted or wildcard-allow path can return a decision. The same request-boundary validation was applied to resource identifiers and `filter_resources()` candidates so the result cannot vary by policy shape. Regression coverage includes unrestricted, wildcard, allow-list, and sync/async paths.

backend/packages/harness/deerflow/authz/rbac.py
hata33: Documented and covered in `2c672ff3`.

`filter_resources()` intentionally keeps the same `ValueError` behavior as `authorize()` for missing or unknown roles; it does not silently convert identity/configuration errors to an empty result. Phase 1B's integration layer remains responsible for applying `fail_closed`. Added explicit regression tests for `None`, empty, and unknown roles, and recorded the contract in the cumulative implementation notes.

WillemJiang (approved):

PR #4264, opened 2026-07-17
fix(skills): reload mounted skills without restarting Gateway

### LINKED ISSUE #4006: [feat] Skill动态刷新
### Before you start

### Problem / motivation

我现在的项目是通过minio 挂载到K8s容器里面，每次添加skills，都需要重启，这样导致需要一直重启deerflow，之前的mcp ，tools，都有动态刷新接口，能否提供一个skills的刷新接口，刚才看了一下代码只有
POST /api/skills/install
安装 skill 后刷新当前用户 skill prompt cache
[skills.py (line 150)]

PUT /api/skills/custom/{skill_name}
编辑 custom skill 后刷新
[skills.py (line 225)]

DELETE /api/skills/custom/{skill_name}
删除 custom skill 后刷新
[skills.py (line 264)]

能否有一个全局刷custom 的接口

### Proposed solution

@router.post("/skills/reload")
async def reload_skills(request: Request):
    await require_admin_user(request, detail=_ADMIN_REQUIRED_DETAIL)
    await refresh_skills_system_prompt_cache_async()
    return {"success": True}

### Affected area(s)

Skills

### Alternatives considered

_No response_

### Additional context

_No response_

Huixin615:
Fixes #4006

## Why

Operators may manage DeerFlow Skills through externally mounted MinIO, NFS, or CSI directories in Kubernetes.

These filesystem changes bypass DeerFlow's install, edit, and delete APIs, so the existing lead-agent prompt caches are not invalidated. Newly added, updated, or removed skills therefore remain stale until the Gateway process is restarted.

`SkillStorage.load_skills()` already scans the filesystem on every call. The stale data comes from:

- The process-level enabled-skills cache.
- The per-user and per-config enabled-skills cache.
- The rendered skills prompt-section LRU cache.

This change adds an explicit, administrator-controlled reload endpoint so operators can make mounted-directory changes visible to subsequent runs without restarting Gateway.

## What changed

- Added a new admin-only endpoint:

  ```http
  POST /api/skills/reload
  ```

- The endpoint has no request body.

- The response explicitly declares that the reload scope is the current Gateway process:

  ```json
  {
    "success": true,
    "scope": "process",
    "message": "Skill caches invalidated; subsequent runs in this Gateway process will rescan the latest skills."
  }
  ```

- Reload invalidates:
  - The process-level enabled-skills cache.
  - Every `(app_config, user_id)` enabled-skills cache entry.
  - The rendered skills prompt-section LRU cache.

- The existing single-flight refresh worker performs the filesystem scan outside the ASGI event loop.

- Reload waits for the background scan using the existing bounded cache-refresh timeout.

- If the wait times out:
  - The API returns a generic server error.
  - Filesystem paths and internal exception details are not exposed.
  - The daemon refresh worker may continue and populate the process cache when it finishes.

- Per-user and per-config loads now capture the current refresh version before scanning.

- A scan only writes its result into the shared cache if the refresh version is still current.

- This prevents an in-flight scan that started before reload from repopulating stale cache entries after reload completes.

- A run that started before reload may continue using its existing skill snapshot.

- Subsequent runs rebuild their agent from the latest discoverable skills.

- Reload remains process-local:
  - Each Uvicorn worker must be targeted separately.
  - Each Kubernetes Gateway Pod must be targeted separately.
  - Requests sent repeatedly through a load-balanced Service do not guarantee that every process is refreshed.
  - The endpoint does not claim cluster-wide invalidation.

- Direct writes to mounted Skills directories remain an operator-controlled trust boundary because they bypass:
  - Install and edit validation.
  - SkillScan.
  - Change history.

- Updated:
  - `README.md`
  - `backend/docs/API.md`
  - `backend/AGENTS.md`

- Added regression coverage for:
  - Administrator authorization.
  - API response contract.
  - Generic error handling.
  - Skill addition, update, and deletion.
  - Process-wide invalidation of user and config caches.
  - Concurrent stale scan write-back.
  - Refresh timeout behavior.
  - Blocking filesystem I/O on the ASGI event loop.

## Surface area

- [x] **Backend API** - endpoint / SSE event / request-response shape under `backend/app`
- [x] **Agents / LangGraph** - agent node, graph wiring, `langgraph.json`, or prompt change
- [ ] **Default behavior change** - changes existing behavior without the user opting in

## Screenshots / Recording

Not applicable. This is a backend-only API and cache-lifecycle change with no frontend UI.

## Bug fix verification

Test paths that reproduce and verify the bug:

- `backend/tests/test_skills_reload.py`
- `backend/tests/test_skills_router_authz.py`
- `backend/tests/blocking_io/test_skills_reload.py`

Did the regression tests go red before the fix and green after the fix? **Yes.**

The following problems were reproduced before their fixes:

- An in-flight per-user scan could repopulate stale cache data after reload.
- The refresh wait had no timeout.
- The reload API and its authorization contract did not exist on `main`.

The tests now verify that:

- An old scan result may be returned to the run that started before reload, but it cannot be written back into the shared cache.
- The next run reloads the current filesystem state.
- A refresh wait that exceeds the configured timeout raises an error.
- The API converts unexpected refresh errors into a generic HTTP 500 response.
- Filesystem scanning does not execute on the ASGI event loop.

## Validation

The branch was rebased onto the latest `upstream/main` before final validation.

### Related tests

```bash
cd backend

uv run pytest \
  tests/test_skills_router_authz.py \
  tests/test_skills_reload.py \
  tests/test_lead_agent_prompt.py \
  tests/test_lead_agent_skills.py \
  tests/blocking_io/test_skills_reload.py \
  -q
```

Result:

```text
52 passed
```

Blocking I/O regression suite

```bash
cd backend
uv run pytest tests/blocking_io -q --tb=short
```

Result:

```text
40 passed
```

Ruff lint

```bash
cd backend
uv run ruff check app packages/harness/deerflow tests
```

Result:

```text
All checks passed!
```

Ruff format check

```bash
cd backend
uv run ruff format --check app packages/harness/deerflow tests
```

Result:

```text
864 files already formatted
```

Full backend test suite

```bash
cd backend
uv run pytest -q
```

Result:

```text
7775 passed, 25 skipped, 1 failed
```

The only full-suite failure was:

```text
tests/test_checkpointer.py::TestGetCheckpointer::test_returns_in_memory_saver_when_not_configured
```

This is a pre-existing local-environment failure that was reproduced before this change. The local, outdated `config.yaml` selects `SqliteSaver`, while the test expects `InMemorySaver`. It is unrelated to the Skills reload implementation.

The final diff check also passed:

```bash
git diff upstream/main...HEAD --check
```


fancyboi999 (changes requested):
Thanks @Huixin615. I found one issue that should be fixed before this is ready.

[P1] Do not report a failed filesystem refresh as successful

- Location: `backend/packages/harness/deerflow/agents/lead_agent/prompt.py:215`
- Impact: if the mounted skills tree is temporarily unreadable or the storage loader raises, `POST /api/skills/reload` still returns 200 while the process-level enabled-skills cache is replaced with an empty list. The operator is told the reload succeeded even though the exact MinIO/NFS/CSI failure this endpoint needs to surface occurred, and subsequent consumers of the global cache can silently lose every enabled skill.
- Evidence: `_refresh_enabled_skills_cache_worker()` catches every exception from `_load_enabled_skills_sync()`, substitutes `skills = []`, and sets the same completion event used here. The new wrapper treats only an event timeout as failure, so it cannot distinguish a successful scan from that exception path. On this head, forcing `_load_enabled_skills_sync()` to raise `PermissionError("mounted skills unavailable")` logged the exception but `refresh_skills_system_prompt_cache_async()` returned normally with `_enabled_skills_cache == []`.
- Suggested fix: carry a success/error result for each refresh generation (or use an equivalent future/condition), publish the new cache only after a successful scan, and propagate a failed generation to this waiter so the router returns its generic 500. Preserve the last known-good cache on scan failure while retaining the existing version/single-flight protection against stale write-back.
- Test: exercise the real worker path with `_load_enabled_skills_sync()` raising, assert the reload endpoint returns the generic 500, and assert the prior cached generation remains available. Keep the existing timeout and concurrent stale-scan tests as separate cases.

Huixin615:
> h

Thanks for your detailed review! I will fix it.

WillemJiang (approved):

PR #4266, opened 2026-07-17
fix(frontend): prevent streamed word animation replay

### LINKED ISSUE #2824: `bug(frontend): streaming word fade-in re-mounts already-rendered words, causing visible "swallow-text" effect`
昨天开始突然发现的前端的问题。

## Summary

The custom `rehypeSplitWordsIntoSpans` rehype plugin in `frontend/src/core/rehype/index.ts` wraps every non-CJK word inside `<p>`, `<h1..h6>`, `<li>`, `<strong>` blocks in a `<span class="animate-fade-in">`. The intent is to give streaming markdown a per-word fade-in animation (1.1s duration via the `--animate-fade-in` CSS variable in `globals.css`).

The bug: as new chunks arrive during streaming, `<Streamdown>` re-runs the rehype pipeline on the entire affected block. Because react-markdown produces these spans without stable React keys, React treats every word — including ones that were already rendered moments earlier — as a fresh mount. Each remount resets the span's `opacity` to `0` and re-runs the 1.1s fade-in. To the user this looks like text "appears, disappears, then reappears".

## Reproduction

Send any English markdown response of moderate length (e.g. `Write a 600-word technical primer in English on X. Use markdown headings, bullet lists, and bold key terms.`) and watch the response render. The flicker is most noticeable on common words (`the`, `a`, `of`, …) inside long paragraphs.

## Evidence

In a 25-second English streaming session captured via a `MutationObserver` watching `body` for `.animate-fade-in` mounts/unmounts:

| Metric | Value |
|---|---|
| Total span mounts | 2,121 |
| Mount-time `opacity` (all 2,121) | `0` (then fade to 1) |
| Animation duration (all 2,121) | `1.1s` (default) |
| Same content word re-mounted multiple times | `the`: 41× · `distance`: 7× · `velocity`: 7× · `Eth`: 8× · `time`: 10× · `zone`: 9× |

Every remount visually replays the 1.1s entry animation, which is what users perceive as the "swallow-text" flicker.

## Why the existing `rehype/index.ts` does not avoid this

The plugin is shape-only: it mutates the hast tree by replacing text nodes with `span` nodes, but does not assign stable `key` properties. React Reconciliation falls back to positional keys, but:

- `streamdown@1.4.0` re-builds the affected block on each chunk (incremental block re-diff)
- the new spans are positionally aligned with old ones, but the `key`-less identity check still drives a remount path under streamdown's block boundary
- so the fade restarts even for words whose text and position haven't changed

Note that `952059eb fix(ui): avoid over-segmenting cjk messages (#1726)` only addresses Chinese being over-split; it does not fix the English/markdown re-mount case described here.

## Proposed fix

`streamdown@2.3.0` introduced a first-class `animated` prop (`fadeIn` / `blurIn` / `slideUp`, configurable `duration` / `easing` / `sep`) that solves exactly this problem. From the streamdown 2.3.0 changelog:

> Per-block `prevContentLength` tracking — each `Block` component keeps a `useRef` with the content length from its previous render. Before each render the `animatePlugin.setPrevContentLength(n)` method is called so the rehype plugin can detect which text-node positions were already rendered. Characters whose cumulative hast-text offset falls below the previous raw-content length receive `--sd-duration:0ms`, making them appear in their final state instantly rather than re-animating.

Using this official animation:

- preserves the per-word fade-in feature
- delegates re-mount detection to streamdown (where it belongs — outside the rehype plugin pipeline)
- removes the custom `rehype-fold` plugin that has been superseded by upstream

PR coming from `noldus-cn-beijing/deerflow-noldus:fix-streaming-word-animation-remount`. The PR keeps the change minimal (no migration to streamdown 2.x's named-plugin `plugins={...}` API, no removal of `rehype-katex` / `rehype-raw` / etc — those are independent cleanup opportunities for follow-up PRs).

## Versions

- Reproduces on `bytedance/deer-flow@main` at the time of writing (commit `7caf03e9`).
- `frontend/package.json` pins `streamdown: "1.4.0"` (released 2025-10-10), which predates the official `animated` prop.

---

March-77:
Fixes #2824

## Why

Appending a streaming chunk reparsed the whole Markdown block. The custom rehype word wrapper recreated every span, so CSS fade-in replayed for text that was already visible and produced the reported swallow/flicker effect.

## What changed

- Upgrade Streamdown from 1.4.0 to 2.5.0 and load its animation stylesheet.
- Use Streamdown's incremental `animated` / `isAnimating` API so only newly arrived words animate.
- Remove the custom per-word rehype plugin and its global keyframes.
- Keep animation props reactive in the safe response wrapper and centralize the Streamdown 2.5 component-map type compatibility boundary.
- Add regression coverage for streaming and completed Markdown rendering.

## Surface area

- [x] **Frontend UI** - streamed Markdown messages and subtask content
- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [x] **Dependencies** - Streamdown 2.5 provides incremental animation bookkeeping that avoids replaying old words
- [x] **Default behavior change** - word animation no longer replays for previously rendered text

## Screenshots / Recording

The regression is encoded as server-rendered markup assertions: streaming content gets Streamdown's incremental animation attributes, while completed content does not. No static screenshot can demonstrate the prior re-mount flicker.

## Bug fix verification

- Test path that reproduces the bug: `frontend/tests/unit/components/workspace/messages/markdown-content.test.ts`
- Did it go red on `main` and green on this branch? yes

## Validation

- `cd frontend && corepack pnpm check`
- `cd frontend && corepack pnpm format`
- `cd frontend && corepack pnpm test` (647/647 passed)
- `cd frontend && SKIP_ENV_VALIDATION=1 corepack pnpm build`
- Playwright E2E could not start on Windows because the existing `webServer.command` invokes `./node_modules/.bin/next`, which `cmd.exe` cannot execute. The production build itself passed.

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Triaged the issue, inspected the rendering lifecycle, wrote the regression tests, implemented the fix, and ran the validation commands. The resulting diff and test output were reviewed before publishing.


fancyboi999 (changes requested):
Thanks @March-77. I found one issue that should be addressed before this is ready.

[P1] Preserve the rendering plugins during the Streamdown 2.5 upgrade

- Location: `frontend/package.json:89` and `frontend/src/core/streamdown/components.tsx:43-68`
- Problem: Streamdown 2.5 requires explicit named plugins for Mermaid rendering and Shiki highlighting, but this branch neither installs those plugin packages nor passes a `plugins` configuration through the shared wrappers. Merging this makes every Mermaid fence render as ordinary source code and removes syntax highlighting from completed code fences across chat, subtasks, artifacts, and the other `SafeStreamdown` call sites.
- Evidence: In the same browser probe, current `origin/main` rendered an accessible `Mermaid chart` and produced colored Shiki token spans for a completed TypeScript fence. This head rendered zero Mermaid charts, emitted a normal `data-language="mermaid"` code block instead, and gave the TypeScript body only `inherit` token styling. Streamdown 2.5's bundled usage contract shows `plugins={{ code, mermaid, ... }}`; this repository has no `@streamdown/code` / `@streamdown/mermaid` dependency or `plugins` prop. The test named `restores Streamdown highlighting after streaming finishes` currently checks only the code-block wrapper, so it stays green when highlighting is gone.
- Suggested fix: add the required Streamdown 2.5 code and Mermaid plugins, expose one stable shared plugin configuration from `core/streamdown`, and pass it through every shared Streamdown wrapper/call site that previously received the 1.x built-ins. Keep the existing raw-HTML and math policy intact rather than replacing it wholesale.
- Test: add a completed-code regression that asserts real highlighted token output, plus a browser/component regression that renders a `mermaid` fence and asserts the diagram/SVG (or its accessible `Mermaid chart`) rather than a plain code block.

March-77:
﻿Addressed the requested Streamdown 2.5 rendering regression in `bac9abcf`:

- registered the official `@streamdown/code` and `@streamdown/mermaid` plugins through one shared config used by safe wrappers and derived configs
- aligned Shiki to the plugin-compatible version and corrected the Tailwind `@source` paths
- strengthened regressions to verify actual Shiki token colors and accessible Mermaid SVG output

Validation:
- `pnpm format`
- `pnpm check`
- `pnpm test -- --run` (654 passed)
- `pnpm build`
- targeted Playwright Mermaid/Shiki test (passed)

frontend/src/core/streamdown/plugins.ts
```diff
@@ -17,23 +17,28 @@ const sharedRemarkPlugins = [
...
+export const streamdownRenderingPlugins = {
+  code,
+  mermaid,
+} satisfies NonNullable<StreamdownProps["plugins"]>;
+
 export const streamdownPlugins = {
+  plugins: streamdownRenderingPlugins,
   remarkPlugins: sharedRemarkPlugins,
   rehypePlugins: [
     rehypeRaw,
```
willem-bd, line 29: Streamdown 2.5 adds an internal `rehype-sanitize` step that the kept `rehypeRaw` now flows through.

The 2.5 upgrade brings `rehype-sanitize@6.0.0` into Streamdown's dependency tree (alongside the bumped `rehype-harden@1.1.8` - see `pnpm-lock.yaml`). This config keeps `rehypeRaw` for the main message path, so raw HTML in assistant markdown now also passes through Streamdown's internal sanitize step, which is stricter than the previous harden-only pipeline.

Net this is a *security improvement*, but it's a behavior change the PR description doesn't call out: trusted raw-HTML content (e.g. artifact previews, or any markdown relying on specific tags/attributes) could lose elements or attributes that the old path passed through.

Suggestion: manually verify a representative raw-HTML markdown sample (an artifact preview, inline HTML in a message) still renders as expected under 2.5, and consider noting the sanitize step in the PR body or a CHANGELOG line. Not a blocker if the check passes.

willem-bd (commented):
One non-blocking note from my review: the Streamdown 2.5 upgrade changes HTML sanitization behavior for the path that keeps rehypeRaw. Details in the inline comment on plugins.ts.

LittleChenLiya:
我本地验证了 PR #4266 对 issue #2824 的修复。

未应用该 patch 前，`main` 仍然使用 `streamdown@1.4.0`，并通过自定义 `rehypeSplitWordsIntoSpans` 插件给英文单词添加 `animate-fade-in` span，因此流式输出时已渲染过的英文单词仍可能重复播放淡入动画，出现吞字/闪烁效果。

在该 PR 分支上，我用包含 heading、bullet list、bold text 的长英文 Markdown 流式回答进行了手动验证：已出现的旧单词保持稳定，只有新流出的文字有入场动画，未再复现吞字/闪烁问题。

同时确认：
- 相关前端单测通过
- 前端 check/type/build 本地通过
- 浏览器手动验证通过，使用 DeepSeek API

结论：该 PR 看起来可以修复原问题。

WillemJiang (approved):

Issue #4189, opened 2026-07-15
Checkpoint storage still grows: residual base64 in messages, multi-instance storage, retention policy, large tool results


qin-chenghan:
## Background

#4138 reported that checkpoint storage grows rapidly because cumulative conversation state (messages, tool results, image data) is persisted repeatedly across every checkpoint.

PR #4140 addressed the first slice: removed base64 image data from the `viewed_images` state field. The middleware now reads image files from disk on-demand and encodes base64 only for the transient model call.

This issue tracks the remaining items from the original report that were not addressed by #4140.

### 1. Injected HumanMessage still carries base64 in checkpoint

`ViewImageMiddleware.before_model` injects a `HumanMessage` containing base64-encoded image data so the LLM can "see" the image. This message is appended to `messages` and therefore snapshotted into the checkpoint for that step. While #4140 eliminated the *second* copy (the `viewed_images` field), the first copy in `messages` remains.

**Possible fix:** Use `RemoveMessage` in `after_model` to delete the injected HumanMessage after the LLM call completes, so subsequent checkpoints no longer carry it. Need to verify that LangGraph's checkpoint lifecycle allows this (the intermediate checkpoint at the `model` node will still contain it, but later checkpoints won't).

### 2. Multi-gateway storage abstraction

`actual_path` in `ViewedImageData` is a real server filesystem path. In a multi-gateway deployment (Kubernetes, load-balanced), the file may not exist on the instance handling the request.

**Possible fix:** Lift `actual_path` to a content-addressed reference (e.g., `sha256:abc123` + bucket key) backed by S3 / MinIO / shared storage. All gateway instances can then resolve the same image. This is also a prerequisite for the on-demand read in `ViewImageMiddleware` to work reliably in distributed deployments.

### 3. Checkpoint retention policy

LangGraph's `AsyncSqliteSaver` / `AsyncPostgresSaver` have no built-in retention API. Old checkpoints accumulate indefinitely. Even with #4140's image fix, long conversations with many tool calls still produce O(steps × message_size) storage growth.

**Possible fix:** Add a configurable retention strategy at the DeerFlow layer:
- Maximum number of checkpoints per thread
- TTL-based cleanup (delete checkpoints older than N days)
- Background compaction job that merges old checkpoints

### 4. Large tool result deduplication

[...]

PR #4267, opened 2026-07-17
fix(backend): remove transient image context after model calls

March-77:
Addresses item 1 of #4189 (the remaining storage-abstraction, retention, and tool-result items stay open).

## Why

`ViewImageMiddleware.before_model` rebuilds image bytes on demand and appends the resulting base64 `HumanMessage` to reducer-backed message state. The model needs that message for one call, but the old implementation left it in every later checkpoint.

LangChain wires `before_model`, `model`, and `after_model` as separate graph nodes. With a checkpointer, each node is a separate super-step, so one injection is still present in two checkpoints for that call: the `before_model` node checkpoint and the `model` node checkpoint.

## What changed

- Give each middleware-injected image context message a reserved ID prefix plus a server-owned metadata marker.
- Strip that marker from untrusted Gateway input and require both identifiers for deduplication and cleanup, so a colliding client-selected ID is preserved.
- Emit targeted `RemoveMessage` updates from both `after_model` and `aafter_model` after the model consumes the image.
- Keep legacy content-marker deduplication for older checkpoints.
- Add hook-level reducer tests plus real `create_agent` lifecycle coverage proving the model sees the injected image message, a normalized client message with the reserved prefix survives, and returned graph state does not retain the transient payload.
- Document the exact checkpoint boundary.

## Residual checkpoint behavior

The `before_model` and `model` node checkpoints for the image-consuming call still contain the transient payload. `after_model` / `aafter_model` then emits `RemoveMessage`, so the `after_model` checkpoint and every subsequent checkpoint no longer retain it.

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [x] **Agents / LangGraph** - ViewImageMiddleware lifecycle, Gateway trust boundary, and message reducer updates
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [x] **Default behavior change** - transient image context is removed after each model call

## Screenshots / Recording

Not applicable; this changes persisted agent state rather than UI output.

## Bug fix verification

- Test paths: `backend/tests/test_view_image_middleware.py` and `backend/tests/test_gateway_services.py`
- Red/green coverage: cleanup previously removed any client message whose ID used the reserved prefix; the new Gateway-to-`create_agent` regression failed before the server-owned marker fix and passes afterward.

## Validation

- `uv run --no-sync pytest tests/test_view_image_middleware.py tests/test_gateway_services.py -q` (132 passed)
- `uv run --no-sync ruff check .`
- `uv run --no-sync ruff format --check .`

## AI assistance

**Tool(s) used:** Codex

**How you used it:** Triaged the checkpoint and trust-boundary issues, wrote red/green lifecycle tests, implemented targeted cleanup, inspected the final diff, and ran the listed validation.


fancyboi999 (changes requested):
Thanks @March-77. I found one issue that should be addressed before this is ready.

[P2] Do not delete client-owned messages by an untrusted ID prefix

- Location: `backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py:231`
- Problem: `_remove_image_context_messages()` deletes every `HumanMessage` whose ID starts with `view-image-context:`, even if that message came from the client and no image context was injected. The Gateway intentionally preserves client-supplied message IDs, so a colliding ID causes the user's input to disappear from the returned state and subsequent checkpoint history after the next model call.
- Evidence: Passing `{type: "human", id: "view-image-context:client-owned", content: "must survive"}` through `normalize_input()` preserves that ID. Invoking a real `create_agent(..., middleware=[ViewImageMiddleware()])` with the normalized input on this head returned only the AI response; the original human message had been removed. Current `main` does not perform this deletion.
- Suggested fix: make the cleanup marker server-owned rather than trusting the public ID namespace. For example, stamp an internal metadata marker that the Gateway strips from untrusted input, and require that marker when emitting `RemoveMessage`; alternatively, reserve and sanitize this ID prefix at the input boundary before relying on it.
- Test: normalize a client message with a colliding ID through the Gateway input path, run the graph with `ViewImageMiddleware`, and assert that the client message survives while a middleware-injected image context message is still removed.

March-77:
﻿Addressed the requested client-ID collision in `3f9a81bf`:

- transient view-image messages now carry a server-owned metadata marker in addition to the reserved ID prefix
- Gateway strips that marker from untrusted input while preserving client-selected IDs and unrelated metadata
- injection deduplication and cleanup require the trusted marker, so a client message whose ID starts with `view-image-context:` is retained
- added the requested Gateway `normalize_input()` -> real `create_agent()` regression proving the client message survives while the injected base64 context is removed

Validation:
- focused gateway/view-image suite: 132 passed
- full-repository Ruff check: passed
- full-repository Ruff format check: passed
- the complete backend suite exceeded the local 4-minute execution limit without emitting a failure; GitHub Actions will provide the full-suite result

backend/AGENTS.md
```diff
@@ -255,7 +255,7 @@ Before changing a later authorization phase, read the [authorization RFC](../doc
...
 22. **MemoryMiddleware** - Queues conversations for async memory update (filters to user + final AI responses)
-23. **ViewImageMiddleware** - *(optional, if the model supports vision)* Injects base64 image data before the LLM call
+23. **ViewImageMiddleware** - *(optional, if the model supports vision)* Injects an ID-marked, hidden HumanMessage with base64 image data before the LLM call, then emits `RemoveMessage` from `after_model` / `aafter_model` so later checkpoints do not retain that transient payload
```
willem-bd: Worth tightening this phrasing - the `risk:high` / `needs-validation` labels mean reviewers will rely on it.

In LangChain's `create_agent`, `before_model` / `model` / `after_model` are wired as **separate graph nodes** (see `langchain/agents/factory.py`: `add_node(f"{m.name}.before_model", ...)`, `add_node("model", ...)`, `add_node(f"{m.name}.after_model", ...)`). With a checkpointer, each is its own super-step with its own checkpoint, so per image injection the payload is still retained in **two** checkpoints - the `before_model` node and the `model` node - before `after_model`'s `RemoveMessage` cleans it up. The PR body's "the model-node checkpoint can still contain the payload" understates this: the `before_model` checkpoint does too.

Still a big win over the prior behavior (payload retained in *all* subsequent checkpoints), and every later node is clean. Suggest rewording to make the 2-checkpoint residual explicit, e.g.: "...emits `RemoveMessage` from `after_model` / `aafter_model` so checkpoints after the model call do not retain that transient payload (the `before_model` and `model` node checkpoints for that call still do)."

WillemJiang:
@March-77, could you take a look at the latest review comments?

backend/AGENTS.md
March-77: Addressed in `c89f7bab`. I updated both `ViewImageMiddleware` entries in `backend/AGENTS.md` and the PR body to state the exact three-node lifecycle: the transient payload is present in the `before_model` and `model` node checkpoints for that call, then `after_model` / `aafter_model` emits `RemoveMessage`, so the `after_model` checkpoint and every subsequent checkpoint are clean. I also updated the PR body's marker description to match the current ID-prefix + server-owned metadata implementation. Could you please take another look?

March-77:
@WillemJiang addressed in `c89f7bab`: the PR body and both relevant `backend/AGENTS.md` entries now explicitly state that the `before_model` and `model` node checkpoints for the call contain the payload, and that `after_model` cleanup makes its checkpoint and every subsequent checkpoint clean.

@willem-bd I also replied on the inline thread. Could you please re-review when convenient?

backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
```diff
@@ -33,13 +36,19 @@ class ViewImageMiddleware(AgentMiddleware[ViewImageMiddlewareState]):
...
     This enables the LLM to automatically receive and analyze images that were loaded via view_image tool,
     without requiring explicit user prompts to describe the images.
     """
 
     state_schema = ViewImageMiddlewareState
 
+    @staticmethod
+    def _is_image_context_message(message: object) -> bool:
+        """Return whether a message is trusted transient image context."""
+        return isinstance(message, HumanMessage) and bool(message.id) and message.id.startswith(_IMAGE_CONTEXT_MESSAGE_ID_PREFIX) and message.additional_kwargs.get(_IMAGE_CONTEXT_MESSAGE_MARKER_KEY) is True
```
willem-bd, line 50: The cleanup depends on `additional_kwargs["deerflow_view_image_context"]` surviving checkpoint serialization as a boolean so this `is True` check still matches in `after_model`.

The end-to-end tests in `test_view_image_middleware.py` use `create_agent(...)` with **no checkpointer**, so the marker is never serialized/deserialized - it stays an in-memory `bool` throughout. In production (Postgres saver) each of `before_model` / `model` / `after_model` is a separate super-step, so the marker must round-trip through the saver for `_is_image_context_message` to match in `after_model`. If a saver ever coerces `True` (e.g. to `"true"` / `1`), this returns `False`, `after_model` emits no `RemoveMessage`, and the base64 payload leaks into every subsequent checkpoint - the exact regression this PR fixes, but only visible under a real saver.

Given `risk:high` / `needs-validation`, could you add a test with a real saver (`MemorySaver` minimum, ideally `SqliteSaver`) that injects, crosses the super-step boundary, and asserts the marker is still `True` (and the message is removed) after deserialization?

backend/packages/harness/deerflow/agents/middlewares/view_image_middleware.py
```diff
@@ -204,13 +213,35 @@ def _should_inject_image_message(self, state: ViewImageMiddlewareState) -> bool:
...
+            additional_kwargs={
+                "hide_from_ui": True,
+                _IMAGE_CONTEXT_MESSAGE_MARKER_KEY: True,
+            },
+        )
+
+    @staticmethod
+    def _remove_image_context_messages(state: ViewImageMiddlewareState) -> dict | None:
+        """Remove transient image context messages after the model consumed them."""
+        removals = [RemoveMessage(id=msg.id) for msg in state.get("messages", []) if ViewImageMiddleware._is_image_context_message(msg)]
```
willem-bd, line 240: This only removes messages that carry the server-owned marker, so pre-upgrade threads whose existing image messages predate the marker (content `"Here are the images you've viewed"`, `hide_from_ui: True`, no marker) are **not** cleaned up here.

The content-string dedup in `_should_inject_image_message` correctly prevents *re*-injecting those legacy messages, so there's no duplication - but the old base64 payload lingers until summarization/compaction drops it. Reasonable forward-only trade-off; worth a one-line note in the `backend/AGENTS.md` residual-checkpoint paragraph so operators know existing threads won't shrink until summarization runs.

backend/tests/test_view_image_middleware.py
```diff
@@ -458,3 +478,140 @@ async def test_abefore_model_returns_none_when_no_injection(self):
...
+                    additional_kwargs={"hide_from_ui": True},
+                ),
+                AIMessage(content="response"),
+            ]
+        }
+
+        assert mw.after_model(state, _runtime()) is None
+
+    def test_graph_preserves_normalized_client_message_with_reserved_prefix(self, tmp_path):
+        from app.gateway.services import normalize_input
```
willem-bd, line 550: This test imports `app.gateway.services.normalize_input` (app layer) into what's otherwise a harness-focused middleware test file. The lazy import avoids a collection-time failure if `app` isn't on the path (e.g. running the harness package in isolation), which is defensive - but the test is really a Gateway <-> `create_agent` trust-boundary integration test, so it might sit more naturally in `test_gateway_services.py` next to the other `normalize_input` tests.

If kept here, consider hoisting the import to module top-level with the rest of the harness imports for consistency (or keep lazy with a one-line comment explaining why).

willem-bd (commented):
Posting three follow-up notes from review as inline comments. Design is sound (dual-identifier trust boundary, `RemoveMessage` cleanup, self-healing resume); these are coverage/migration/style considerations rather than correctness blockers.

WillemJiang (approved):

PR #4268, opened 2026-07-17
fix(workspace-changes): offload blocking filesystem IO in text-cache lifecycle

AnnaSuSu:
## Why

`workspace_changes/recorder.py`'s two async entry points — `capture_workspace_snapshot` and `record_workspace_changes`, both awaited by `runtime/runs/worker.py` on every agent run — already offload their scans via `asyncio.to_thread`, but ran the snapshot **text cache's whole lifecycle** on the event loop:

| Site | Blocking call | When |
|---|---|---|
| `capture_workspace_snapshot` | `build_thread_workspace_roots` → `os.path.abspath` | every run |
| `capture_workspace_snapshot` | `tempfile.mkdtemp` → `os.mkdir` | every run (`include_text=True`) |
| `capture_workspace_snapshot` | `shutil.rmtree` | scan-failure branch |
| `record_workspace_changes` | `build_thread_workspace_roots` → `os.path.abspath` | every run |
| `record_workspace_changes` | `shutil.rmtree` (in `finally`) | **every run, incl. abort paths** |

The last one is the notable one: `_cleanup_snapshot_text_cache` sits in a `finally`, so it is not an error path — every agent run removed up to `max_files` (200) cached text files on the loop.

`make detect-blocking-io` flagged the two `shutil.rmtree` calls as **HIGH** — the last 2 HIGH findings in the repo. After this change the scanner reports **zero HIGH**, which per `.agent/skills/blocking-io-guard/SKILL.md`'s batching policy ("do not start MEDIUM until every HIGH is dispositioned") also unblocks the MEDIUM rounds.

Same class as the already-merged #3457, #1917, #3084, and #3529; this is the runtime-side counterpart of #3529's inbound-ingestion fix and reuses its `_prepare_uploads_dir()` single-worker-hop shape.

## What changed

- `_prepare_capture()` runs roots resolution + `mkdtemp` in one worker thread, returning `(roots, text_cache_dir)`.
- The `_prepare_capture` handoff is cancellation-safe, including under **repeated** cancellation: `capture_workspace_snapshot` shields the prepare future, and on `CancelledError` hands the reclaim+remove to a task the caller cannot abandon, draining repeated cancellation until it completes before re-raising. Otherwise a cancel between `mkdtemp` and the coroutine receiving the path would orphan the dir — and a second cancel landing on the reclaim await would do so again (thanks @fancyboi999).
- `record_workspace_changes` offloads its own `build_thread_workspace_roots` call.
- Both `shutil.rmtree` sites now route through `_remove_text_cache_dir()`, offloaded via `asyncio.to_thread`.
- Cleanup stays **best-effort**: `_remove_text_cache_dir` swallows and logs, so a failing cleanup cannot replace the exception or result already in flight (SOP Step 2's requirement for `finally`/cleanup paths).

Externally observable behavior is unchanged: same cleanup timing, same `ignore_errors=True` semantics, same return values.

Two scoping notes worth surfacing for review:

1. **The roots resolution is not in the scanner's output**, but it is in this diff. `build_thread_workspace_roots` is a sync helper whose blocking call lives in `paths.sandbox_work_dir` (another file), so it falls in the documented `ASYNC_REACHABLE_SAME_FILE` blind spot. It blocks the same async path, and — concretely — the anchor cannot reach the `rmtree` without it: against pre-fix code the gate fires on `os.path.abspath` first.

2. **`asyncio.shield` — cleanup vs creation are different questions.** For *cleanup* (rmtree) no shield is needed: `asyncio.to_thread` submits immediately, so cancelling the future does not stop the running thread and the cache is still removed (measured under single and repeated cancellation — naive and sync versions all clean up). For the *creation* handoff it is needed: `_prepare_capture` builds the cache dir in the worker, so a cancel between `mkdtemp` and receiving the path would drop the path and orphan the dir. `capture_workspace_snapshot` shields the prepare future and, on cancellation, reclaims its result to remove the dir. Because that reclaim `await` is itself cancellable, the reclaim+remove is owned by a task the caller cannot abandon and repeated cancellation is drained until it finishes, then the cancellation is restored — a second `shield()` on the await alone would let a re-cancel skip the reclaim and orphan the dir (regressions: `test_capture_workspace_snapshot_cancelled_handoff_leaks_no_text_cache`, `test_capture_workspace_snapshot_repeated_cancellation_leaks_no_text_cache`). Note: the symmetric window during the *scan* await (`except Exception` does not catch `CancelledError`) is pre-existing — the branch's `except Exception` is byte-identical to `main`, which leaks the same way — and is left as a follow-up because fixing it correctly requires ordering the rmtree after the still-running scan worker.

## Surface area

- **Backend** — `packages/harness/deerflow/workspace_changes/recorder.py` (+62/−7) + new `tests/blocking_io` anchor + `backend/AGENTS.md` anchor-list entry. No API / schema change.

## Validation

- New `tests/blocking_io/test_workspace_changes_recorder.py` (4 tests), driving the **capture-failure branch**, the **record `finally`**, and the **cancelled `mkdtemp` handoff** under both single and repeated cancellation — not the happy path, which never reaches the rmtree.
- **Teeth verified per clause** under the strict Blockbuster gate — each offload reverted *alone*, confirming each test reddens on its own call rather than being rescued by a sibling:

  | Reverted clause | Test | Reddens on |
  |---|---|---|
  | `rmtree` offload | `test_capture_..._cleanup` | `capture_workspace_snapshot` → `_remove_text_cache_dir` → `BlockingError: os.path.samestat` |
  | `rmtree` offload | `test_record_..._cleanup` | `_remove_text_cache_dir` → `BlockingError: os.path.samestat` |
  | `_prepare_capture` offload | `test_capture_..._cleanup` | `_prepare_capture` → `BlockingError: os.path.abspath` |
  | `record` roots offload | `test_record_..._cleanup` | `build_thread_workspace_roots` → `BlockingError: os.path.abspath` |

  All GREEN after the fix. Blockbuster already has rules covering `rmtree`'s internals, so no new rule is needed (coverage opening, not the RULE route).
- Re-scan (Step 2): `make detect-blocking-io` → `recorder.py` findings **0**, HIGH remaining repo-wide **0** (was `2 HIGH / 20 MEDIUM / 16 LOW`).
- `make test-blocking-io` → **43 passed** (39 before + 4).
- Workspace-changes regression (`-k "workspace_change or recorder"`) → **31 passed, 3 skipped**.
- Full `make test` (round-1 baseline; round 2 re-ran the blocking-io gate + workspace-changes regression above, not the full suite) → **7761 passed, 8 failed, 28 skipped**. The 8 failures are **pre-existing and unrelated** — verified by re-running the same three files against unmodified `origin/main`: **8 failed / 52 passed on both**. 6 of the 8 share one mechanism (`Error: Refusing to fetch a private, loopback, or metadata address` — the SSRF guard in the web-fetch tools, environment/DNS dependent); none of `test_crawl4ai_tools.py`, `test_fastcrw_tools.py`, `test_browserless_client.py` reference `workspace_changes`.
- `uvx ruff check .` → All checks passed! · `uvx ruff format --check .` → 873 files already formatted.

## AI assistance

**How you used it:** AI ran `make detect-blocking-io`, routed the two HIGH findings per the `blocking-io-guard` SOP, implemented the `asyncio.to_thread` offloads, and wrote the anchor (teeth verified red→green per clause). After review feedback it reproduced the `mkdtemp`-handoff cancellation leak, added the shield+reclaim guard and its regression, and verified the pre-existing scan-await window against `origin/main`. In a second review round it reproduced the repeated-cancellation leak in the reclaim path, moved reclaim into a drained task the caller cannot abandon, and added the double-cancel regression. I reviewed the change.

### DIFF
--- a/backend/AGENTS.md
+++ b/backend/AGENTS.md
@@ -142,10 +142,13 @@ Blocking-IO runtime gate (`tests/blocking_io/`):
   `test_jsonl_run_event_store.py` (locks `JsonlRunEventStore`'s async
   API offloading its file IO via `asyncio.to_thread`);
   `test_uploads_middleware.py` (locks `UploadsMiddleware.abefore_agent`
-  offloading the uploads-directory scan off the event loop); and
+  offloading the uploads-directory scan off the event loop);
   `test_uploads_router.py` (locks Gateway upload/list/delete endpoints
   offloading upload directory creation, staged writes, chmod/cleanup,
-  directory scans/deletes, and remote sandbox sync off the event loop).
+  directory scans/deletes, and remote sandbox sync off the event loop); and
+  `test_workspace_changes_recorder.py` (locks the offload around the snapshot
+  text cache lifecycle — roots resolution, `mkdtemp`, and the `shutil.rmtree`
+  on both the capture-failure branch and `record_workspace_changes`' `finally`).
 - `test_gate_smoke.py` is a meta-test asserting the gate actually catches
   unoffloaded blocking IO and that the `@pytest.mark.allow_blocking_io`
   opt-out works.
--- a/backend/packages/harness/deerflow/workspace_changes/recorder.py
+++ b/backend/packages/harness/deerflow/workspace_changes/recorder.py
@@ -66,7 +66,22 @@ async def capture_workspace_snapshot(
     limits: WorkspaceChangeLimits | None = None,
     include_text: bool = True,
 ) -> WorkspaceSnapshot:
-    roots, text_cache_dir = await asyncio.to_thread(_prepare_capture, thread_id, user_id=user_id, include_text=include_text)
+    # `_prepare_capture` creates the text cache dir inside the worker, so the
+    # handoff must be cancellation-safe: if the run is cancelled after mkdtemp
+    # but before we receive the path, the shielded worker still finishes and we
+    # reclaim its result to remove the orphaned dir before re-raising.
+    prepare = asyncio.ensure_future(asyncio.to_thread(_prepare_capture, thread_id, user_id=user_id, include_text=include_text))
+    try:
+        roots, text_cache_dir = await asyncio.shield(prepare)
+    except asyncio.CancelledError:
+        orphaned = None
+        try:
+            _, orphaned = await asyncio.shield(prepare)
+        except Exception:
+            orphaned = None  # prepare failed before creating a dir; nothing to reclaim
+        if orphaned is not None:
+            await asyncio.shield(_remove_text_cache_dir(orphaned))
+        raise
     try:
         return await asyncio.to_thread(
             scan_workspace_roots,
--- a/backend/packages/harness/deerflow/workspace_changes/recorder.py
+++ b/backend/packages/harness/deerflow/workspace_changes/recorder.py
@@ -59,6 +59,21 @@ async def _remove_text_cache_dir(text_cache_dir: str | Path) -> None:
         logger.warning("Failed to remove workspace text cache %s", text_cache_dir, exc_info=True)
 
 
+async def _reclaim_prepare_and_cleanup(prepare: asyncio.Future[tuple[list[WorkspaceRoot], Path | None]]) -> None:
+    """Await a cancelled prepare handoff and remove any dir it created.
+
+    Owned by its own task so that caller cancellation during reclaim can interrupt
+    the *await* but never abandon a just-created text cache dir. Best-effort,
+    mirroring `_remove_text_cache_dir`.
+    """
+    try:
+        _, orphaned = await prepare
+    except Exception:
+        return  # prepare failed before creating a dir; nothing to reclaim
+    if orphaned is not None:
+        await _remove_text_cache_dir(orphaned)
+
+
 async def capture_workspace_snapshot(
     thread_id: str,
     *,
@@ -74,13 +89,18 @@ async def capture_workspace_snapshot(
     try:
         roots, text_cache_dir = await asyncio.shield(prepare)
     except asyncio.CancelledError:
-        orphaned = None
-        try:
-            _, orphaned = await asyncio.shield(prepare)
-        except Exception:
-            orphaned = None  # prepare failed before creating a dir; nothing to reclaim
-        if orphaned is not None:
-            await asyncio.shield(_remove_text_cache_dir(orphaned))
+        # `prepare` is shielded, so it keeps running and may still create the dir
+        # after this cancel. Own the reclaim+remove in a task the caller cannot
+        # abandon: a repeat cancel can interrupt our await but not the task, so we
+        # drain repeated cancellation until cleanup finishes, then restore it. A
+        # second `shield()` on the await alone would let a re-cancel skip the
+        # reclaim and orphan the dir.
+        cleanup = asyncio.ensure_future(_reclaim_prepare_and_cleanup(prepare))
+        while not cleanup.done():
+            try:
+                await asyncio.shield(cleanup)
+            except asyncio.CancelledError:
+                pass
         raise
     try:
         return await asyncio.to_thread(
--- a/backend/packages/harness/deerflow/workspace_changes/recorder.py
+++ b/backend/packages/harness/deerflow/workspace_changes/recorder.py
@@ -38,6 +38,27 @@ def build_thread_workspace_roots(thread_id: str, *, user_id: str | None = None) 
     ]
 
 
+def _prepare_capture(thread_id: str, *, user_id: str | None, include_text: bool) -> tuple[list[WorkspaceRoot], Path | None]:
+    # Worker thread: resolving the sandbox roots hits the filesystem, and mkdtemp
+    # creates the text cache directory — both blocking IO that must stay off the
+    # event loop.
+    roots = build_thread_workspace_roots(thread_id, user_id=user_id)
+    text_cache_dir = Path(tempfile.mkdtemp(prefix="deerflow-workspace-changes-")) if include_text else None
+    return roots, text_cache_dir
+
+
+async def _remove_text_cache_dir(text_cache_dir: str | Path) -> None:
+    """Remove a snapshot's text cache off the event loop.
+
+    Best-effort by contract: every caller is a failure or teardown path, so a
+    cleanup error must never replace the exception or result already in flight.
+    """
+    try:
+        await asyncio.to_thread(shutil.rmtree, text_cache_dir, ignore_errors=True)
+    except Exception:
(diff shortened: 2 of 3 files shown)


fancyboi999 (changes requested):
Thanks @AnnaSuSu. I found one issue that should be addressed before this is ready.

[P2] Clean up a cache created during cancelled preparation

- Location: `backend/packages/harness/deerflow/workspace_changes/recorder.py:69`
- Problem: `_prepare_capture()` now creates the temporary text-cache directory inside the worker used by `asyncio.to_thread()`. If the run is cancelled after `mkdtemp()` succeeds but before the worker result is handed back, `capture_workspace_snapshot()` exits without ever receiving `text_cache_dir`, while the worker continues and leaves the directory and its cached files behind. Repeated cancelled runs can therefore accumulate orphaned `deerflow-workspace-changes-*` trees in the system temp directory.
- Evidence: I synchronized `_prepare_capture()` immediately after its real `mkdtemp()`, cancelled `capture_workspace_snapshot()` while the worker was blocked, then released and waited for the worker. The task raised `CancelledError` and one cache directory remained (`leftovers=1`). This handoff window is introduced by the diff: on current `origin/main`, the directory is created synchronously before the scan's first await, so the coroutine already owns the path when cancellation can occur.
- Suggested fix: make the worker-result handoff cancellation-safe. For example, keep a task for the preparation future, shield the handoff, and on cancellation await the worker result and remove its cache before re-raising; alternatively, redesign the worker so it retains cleanup ownership until the coroutine explicitly accepts the result. `shield()` by itself is not sufficient because it would still leave the completed result unclaimed.
- Test: add a deterministic cancellation regression that blocks the worker after `mkdtemp()`, cancels the coroutine, releases the worker, and asserts that no `deerflow-workspace-changes-*` directory remains.

AnnaSuSu:
Thanks @fancyboi999 — confirmed both the mechanism and that it's introduced by this diff. Reproduced it (park the worker after `mkdtemp`, cancel, release → `leftovers=1`); on `origin/main` the dir is created synchronously before the first await, so the window really is new here.

Fixed with the shield+reclaim shape you suggested (`ac4c2be`): `capture_workspace_snapshot` now shields the prepare future, and on `CancelledError` re-awaits it to claim the dropped result and remove the dir before re-raising — so `shield()` isn't left with an unclaimed result. I kept `mkdtemp` in the worker because the Blockbuster gate flags `os.mkdir` from `deerflow` code (moving it back onto the loop reddens the existing anchor), so un-offloading wasn't an option.

Added a deterministic regression (`test_capture_workspace_snapshot_cancelled_handoff_leaks_no_text_cache`) that blocks the worker after `mkdtemp`, cancels, releases, and asserts no `deerflow-workspace-changes-*` remains — RED on the pre-fix code, green now. `make test-blocking-io` → 42 passed; `make detect-blocking-io` → recorder.py 0 findings, 0 HIGH.

One scope note: the symmetric window during the *scan* await (cancel there also leaks, since `except Exception` doesn't catch `CancelledError`) is pre-existing — the branch's `except Exception` is byte-identical to `main` and `main` leaks the same way. I left it out of this PR because fixing it correctly means ordering the rmtree after the still-running scan worker (otherwise cleanup races the worker still writing files); happy to do that as a follow-up.

fancyboi999 (changes requested):
Thanks @AnnaSuSu. The single-cancel leak is fixed, but the reclaim path is still cancellable and can leave the same directory behind.

[P2] Keep cleanup ownership across repeated cancellation

- Location: `backend/packages/harness/deerflow/workspace_changes/recorder.py:76-83`
- Problem: after the first `CancelledError`, the coroutine awaits the shielded preparation task again. A second cancellation at that await raises another `CancelledError` (which is not caught by `except Exception`) and exits before the worker result is reclaimed. The worker still completes and the `deerflow-workspace-changes-*` directory remains orphaned.
- Evidence: on head `ac4c2beeb61e6f1f9aa88d232329452c839913a5`, I parked `_prepare_capture()` after creating the real directory, cancelled the capture task once, yielded until it was waiting in the reclaim branch, cancelled it again, then released the worker. The final state was `done=True`, `cancelling=2`, and `exists=True`. The new single-cancel regression and the focused file still pass (`3 passed`), so they do not exercise this cleanup interruption.
- Suggested fix: keep cleanup ownership in a task/guard that caller cancellation cannot abandon until the preparation result is explicitly accepted, or otherwise drain repeated cancellation while reclaiming and restore the cancellation state only after cleanup completes. A second `shield()` alone does not make the caller's await non-cancellable.
- Test: extend the handoff regression to issue a second `task.cancel()` after the first cancellation has entered the reclaim await, then release the worker and assert no cache directory remains.

Reviewed head SHA: `ac4c2beeb61e6f1f9aa88d232329452c839913a5`

AnnaSuSu:
Thanks @fancyboi999 — you're right, and it reproduces. Round 1 closed the single cancel but left the reclaim path itself cancellable, which is the same leak one level up. Confirmed on `ac4c2bee`: park the worker after `mkdtemp`, cancel, let it reach the reclaim await, cancel again, release → the `deerflow-workspace-changes-*` dir is still there. The second `CancelledError` lands on the reclaim `await` and, being `BaseException`, slips past `except Exception`, so the reclaim is skipped while the shielded worker finishes — exactly your walk.

Fixed by moving reclaim ownership off the caller's await, per your option: `capture_workspace_snapshot` now hands the reclaim+remove to its own task and **drains repeated cancellation** until that task completes, then restores the cancellation:

```python
except asyncio.CancelledError:
    cleanup = asyncio.ensure_future(_reclaim_prepare_and_cleanup(prepare))
    while not cleanup.done():
        try:
            await asyncio.shield(cleanup)
        except asyncio.CancelledError:
            pass
    raise
```

A repeat cancel can interrupt our `await` but not the task, so the dir can't be abandoned; the loop exits only once cleanup has actually finished (no fire-and-forget task left pending, no reliance on the loop outliving our return). `_prepare_capture` is bounded, so `cleanup` always completes — the drain can't spin.

Added the regression you described — `test_capture_workspace_snapshot_repeated_cancellation_leaks_no_text_cache`: first cancel enters the reclaim await, second cancel lands on it, then the worker is released and it asserts no `deerflow-workspace-changes-*` remains. RED on the pre-fix HEAD (`leaked a text cache dir: [...]`), green now. The single-cancel regression stays green. `make test-blocking-io` → 43 passed.

The symmetric **scan-await** window is unchanged from round 1: still pre-existing (the branch's `except Exception` is byte-identical to `main`, which leaks the same way), and still better fixed as the follow-up that orders the rmtree after the running scan worker rather than folded in here.

WillemJiang (approved):

PR #4274, opened 2026-07-18
fix(channels): document GitHub dedupe TTL, tighten redelivery tests

DaoyuanLi2816:
## Summary

Follow-up to #4104 ("dedupe GitHub webhook redeliveries"), which merged with three review points from willem-bd explicitly left non-blocking-but-deferred rather than addressed:

> Left the TTL-documentation point and the two test-shape nits as-is for now since they're non-blocking per your review — happy to follow up separately if you'd like those tracked.

That follow-up was never actually filed. This PR delivers it.

## What's addressed

From willem-bd's review on #4104:

1. **TTL documentation** ([discussion_r3569700119](https://github.com/bytedance/deer-flow/pull/4104#discussion_r3569700119)) — `ChannelManager._recent_inbound_events`, the cache the whole inbound-redelivery dedupe is built on, is a process-local, in-memory-only `OrderedDict` that is never persisted to `ChannelStore`. `INBOUND_DEDUPE_TTL_SECONDS` (10 minutes) and `INBOUND_DEDUPE_MAX_ENTRIES` (4096) had no comment explaining any of this. Added one directly above the constants covering: what the TTL is, why 10 minutes is a reasonable window (it comfortably covers a provider's own immediate retry-on-timeout — e.g. GitHub's 10s webhook timeout retrying the same `X-GitHub-Delivery` GUID — without an unbounded in-memory ledger), and what happens at the boundary (a manual "Redeliver" clicked after the TTL has elapsed, or any redelivery after a Gateway restart, is **not** deduped — the agent runs again and may repeat a real side effect, e.g. a duplicate PR comment). Also added a short cross-reference from the GitHub dispatcher's fan-out comment, which previously implied unconditional redelivery coverage without this caveat.

2. **Test-shape nit, `test_channels.py`** ([discussion_r3569700125](https://github.com/bytedance/deer-flow/pull/4104#discussion_r3569700125)) — `test_github_redelivery_is_deduped_like_other_channels`'s `_gh` helper stamped a 2-part `f"{delivery}:{agent}"` dedupe id, while `fanout_event` emits a 3-part `f"{delivery_id}:{match.user_id}:{agent.name}"` id (per the cross-user fix from earlier in #4104's own review cycle). The helper's comment claimed to be "shaped exactly as fanout_event emits" but wasn't. Fixed the helper to build the real 3-part shape and added an assertion that a second user's same-named agent on the same delivery is not cross-deduped — a case the old 2-part shape couldn't even express.

3. **Test-shape nit, `test_github_dispatcher.py`** ([discussion_r3569700127](https://github.com/bytedance/deer-flow/pull/4104#discussion_r3569700127)) — `test_missing_delivery_header_leaves_dedupe_open` only asserted the dispatcher-layer `message_id` was `None`; it never locked in the actual manager-level consequence. Added `ChannelManager._is_duplicate_inbound` assertions across two separate header-less deliveries, so a future change that let a missing id fall through as a real (constant) key — which would silently collapse distinct header-less deliveries into "the same" message — is caught here too, not only by inspecting the raw id.

## Not in scope

#4104 also had a real cross-user silent-drop bug that willem-bd found and the contributor fixed inline during that same review (folding `match.user_id` into the dedupe id). That was fixed before merge and is unrelated to this PR — it's mentioned here only to distinguish it from the three items above, which were the ones actually left open.

## Tests

```
cd backend && PYTHONPATH=. pytest tests/test_github_dispatcher.py tests/test_channels.py tests/test_github_channel.py tests/test_github_registry.py -q
```

313 passed. `ruff check` / `ruff format --check` clean on all four touched files.

For both tightened tests, verified they actually discriminate (not just pass) by temporarily reintroducing the exact bug each targets and confirming the new assertion fails, then reverting:
- Reverted the `_gh` helper to the old 2-part id shape — the new cross-user assertion failed (a different user's message was wrongly flagged as a duplicate).
- Patched `_inbound_dedupe_key` to fall back to a constant `"no-id"` key instead of returning `None` for a missing id — the new manager-level assertion on the second header-less delivery failed, confirming it would catch that regression.

### DIFF
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -72,6 +72,24 @@ MESSAGE_STREAM_EVENTS = ("messages-tuple", "messages")
 THREAD_BUSY_MESSAGE = "This conversation is already processing another request. Please wait for it to finish and try again."
 BOUND_IDENTITY_REQUIRED_MESSAGE = "Connect this channel from DeerFlow Settings, complete the in-channel connect step, then send your message again."
 BOUND_IDENTITY_UNAVAILABLE_MESSAGE = "Channel connection verification is temporarily unavailable. Please try again later or contact the DeerFlow operator."
+# Inbound-redelivery dedup window. ``_recent_inbound_events`` (below) is a
+# process-local, in-memory-only OrderedDict — it is never persisted to
+# ``ChannelStore`` — so a recorded key survives only for this TTL (or until
+# evicted by the entry cap below) and is gone entirely across a Gateway
+# restart. 10 minutes comfortably covers a provider's own immediate
+# retry-on-timeout, the common duplicate-side-effect cause (e.g. GitHub
+# retrying the same ``X-GitHub-Delivery`` GUID after its 10s webhook
+# timeout), without keeping a growing in-memory ledger.
+#
+# At the boundary: a manual redelivery (e.g. GitHub's "Redeliver" button)
+# clicked *after* the TTL has elapsed, or any redelivery following a Gateway
+# restart, is no longer recognized as a duplicate — the key has already been
+# evicted, or never existed in the new process — so the agent runs again
+# and may repeat a real side effect (e.g. a duplicate PR comment on
+# GitHub). This is parity with every other IM channel's dedupe (same
+# mechanism, same TTL), not a channel-specific gap. True idempotency against
+# a late/manual redelivery would require persisting the dedupe key in
+# ``ChannelStore`` instead, which is not implemented here.
 INBOUND_DEDUPE_TTL_SECONDS = 10 * 60
 INBOUND_DEDUPE_MAX_ENTRIES = 4096
 # Only server-stable provider message ids: client-generated ids (client_msg_id,
--- a/backend/app/channels/manager.py
+++ b/backend/app/channels/manager.py
@@ -76,10 +76,22 @@ BOUND_IDENTITY_UNAVAILABLE_MESSAGE = "Channel connection verification is tempora
 # process-local, in-memory-only OrderedDict — it is never persisted to
 # ``ChannelStore`` — so a recorded key survives only for this TTL (or until
 # evicted by the entry cap below) and is gone entirely across a Gateway
-# restart. 10 minutes comfortably covers a provider's own immediate
-# retry-on-timeout, the common duplicate-side-effect cause (e.g. GitHub
-# retrying the same ``X-GitHub-Delivery`` GUID after its 10s webhook
-# timeout), without keeping a growing in-memory ledger.
+# restart. 10 minutes is a deliberately bounded window: long enough to
+# absorb a near-term redelivery of the same event — whether a provider's
+# own automatic retry (where the provider implements one) or an operator
+# explicitly triggering a resend — without keeping a growing in-memory
+# ledger.
+#
+# For GitHub specifically: GitHub does NOT automatically retry or redeliver
+# a failed delivery (non-2xx response, timeout, or connection error) — it
+# is simply recorded as failed. See GitHub's own documentation:
+# https://docs.github.com/en/webhooks/using-webhooks/handling-failed-webhook-deliveries.
+# Every redelivery of the same ``X-GitHub-Delivery`` GUID is therefore an
+# explicit action — the repo/App "Redeliver" button, the REST API, or an
+# operator's own scheduled recovery script polling the failed-deliveries
+# endpoint (the pattern GitHub's own docs recommend) — never an automatic
+# GitHub-side retry. This TTL exists to absorb exactly those explicit
+# near-term replays.
 #
 # At the boundary: a manual redelivery (e.g. GitHub's "Redeliver" button)
 # clicked *after* the TTL has elapsed, or any redelivery following a Gateway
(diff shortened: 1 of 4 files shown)


fancyboi999 (changes requested):
Thanks @DaoyuanLi2816. The tightened tests are sound, but the new TTL documentation relies on a GitHub retry behavior that GitHub explicitly says does not occur.

[P2] Do not document failed-webhook delivery as an automatic retry

- Location: `backend/app/channels/manager.py:75-83` and `backend/app/gateway/github/dispatcher.py:405-412`
- Problem: the new comments justify the 10-minute window with GitHub retrying the same delivery after its 10-second timeout and call that a common duplicate source. GitHub records a slow/failed delivery as failed but does not automatically redeliver it. This gives operators the wrong durability model and implies the in-memory TTL protects a failure mode it never sees unless an operator or external recovery process explicitly requests redelivery.
- Evidence: GitHub's current official [Handling failed webhook deliveries](https://docs.github.com/en/webhooks/using-webhooks/handling-failed-webhook-deliveries) documentation states that failed deliveries are not automatically redelivered; manual UI/API redelivery (or a separately implemented recovery script) is required. GitHub's [webhook best practices](https://docs.github.com/en/webhooks/using-webhooks/best-practices-for-using-webhooks) separately says the 10-second timeout marks failure and that a requested redelivery reuses `X-GitHub-Delivery`. The code's dedupe behavior is fine, but the new automatic-retry rationale is factually wrong.
- Suggested fix: describe the window in terms of explicit/manual/API redelivery and replay protection, and justify 10 minutes as the chosen bounded in-memory window without claiming GitHub performs an immediate retry. Keep the restart/TTL limitation and the same-GUID behavior for requested redelivery.
- Test: no runtime change is required; retain the current dedupe tests and add the authoritative GitHub redelivery documentation link beside the corrected operational note so this assumption does not drift again.

Reviewed head SHA: `1a5761f856199d3cf7cd8e7c2103577181b23eea`

WillemJiang (dismissed):

DaoyuanLi2816:
@WillemJiang — since your review comment flagged that issues were still unresolved, I went back and re-verified the webhook-retry claims in the code comments against GitHub's actual documented behavior rather than relying on priors.

**What I verified (fetched directly, not from memory):**

- ["Handling failed webhook deliveries"](https://docs.github.com/en/webhooks/using-webhooks/handling-failed-webhook-deliveries): a failed delivery (non-2xx response, timeout, or connection error) is simply recorded as failed. GitHub does not automatically redeliver it.
- ["Best practices for using webhooks"](https://docs.github.com/en/webhooks/using-webhooks/best-practices-for-using-webhooks): confirms the 10-second timeout marks a delivery as failed, and that a *requested* redelivery reuses the same `X-GitHub-Delivery` header.
- ["Automatically redelivering failed deliveries for a GitHub App webhook"](https://docs.github.com/en/webhooks/using-webhooks/automatically-redelivering-failed-deliveries-for-a-github-app-webhook) / [...for a repository webhook](https://docs.github.com/en/webhooks/using-webhooks/automatically-redelivering-failed-deliveries-for-a-repository-webhook): despite the word "automatically" in the title, both pages are guides for a *user-written* recovery script (GitHub's own example is a GitHub Actions workflow on a 6-hour cron, plus `workflow_dispatch` for manual runs) — not a GitHub-side feature.
- I also specifically checked whether this differs between GitHub Apps and repository webhooks, since our route accepts both (`receive_github_webhook` handles "GitHub App / repo webhook deliveries"). It doesn't — same behavior either way, per both dedicated docs pages above.

So the comments' claim that "10 minutes comfortably covers... GitHub retrying the same `X-GitHub-Delivery` GUID after its 10s webhook timeout" was factually wrong. Every GitHub redelivery — regardless of our TTL — is the result of an explicit action: the repo/App "Redeliver" button, the REST API, or an operator's own scheduled recovery script.

**Fix pushed** (`b0d2bd88`): reworded the `INBOUND_DEDUPE_TTL_SECONDS` comment in `manager.py` and the `dedupe_message_id` comment in `dispatcher.py` to describe the actual mechanism (explicit manual/API/scripted redelivery + replay protection) instead of an automatic-retry rationale, with the authoritative doc link inline next to the corrected note. I also reworded the two test docstrings that repeated the same "retry-on-timeout" framing (`test_channels.py::test_github_redelivery_is_deduped_like_other_channels`, `test_github_dispatcher.py::test_delivery_id_populates_inbound_dedupe_identity`) so the corrected mental model isn't sitting next to contradicting prose a few files over. No assertions changed, since the dedupe *behavior* itself was never in question. All four affected test files pass, ruff check/format are clean.

**One more thing found while verifying this, for full transparency:** the same misconception — in a more specific, unverified form ("GitHub retries 5xx deliveries with exponential backoff, up to ~5 attempts over ~8 hours") — was already present in `github_webhooks.py`'s route docstring, predating this PR. That figure doesn't come from anywhere in GitHub's docs; it reads like generic third-party webhook-retry advice (the kind written for services designing their *own* outbound retry policy) misattributed to GitHub. I didn't want to inflate this PR's diff with an unrelated file, so I filed that separately: #4289 (also comments/docstrings/log-message only, zero behavior change — the 503-vs-200 status code logic there was already correct regardless of the wrong rationale attached to it).

Let me know if I'm still missing something here — happy to keep digging.

WillemJiang (approved):

PR #4275, opened 2026-07-18
fix(config): close out #4124 review follow-ups (shared signature helper, resolve_config_path None contract)

DaoyuanLi2816:
## Summary

This closes out the two follow-ups noted in #4124 ("fix(mcp): invalidate tools cache on config content + path, not just newer mtime") that were explicitly deferred during review rather than fixed there.

## What was left open on #4124

**1. Duplicate `_get_config_signature` / `_ConfigSignature`**

willem-bd, reviewing `mcp/cache.py`:

> This `_get_config_signature` (plus the `_ConfigSignature` alias at line 24) is now a verbatim duplicate of `deerflow/config/app_config.py::_get_config_signature` / `_ConfigSignature`. Mirroring rather than refactoring is the right call for this focused behavior fix - but worth a follow-up to extract both into a small shared helper (e.g. under `deerflow/config/`) so the two runtime-editable config files don't drift apart over time. Non-blocking.

The response on #4124:

> Agreed on both counts — it is a verbatim duplicate, and mirroring rather than a shared helper was the right call for this focused fix. Leaving the extraction as the follow-up you suggested rather than pulling it into this branch.

**2. `resolve_config_path()` raises instead of returning `None` on a deleted explicit/env-var path**

Also from #4124, found one layer deeper while responding to a related-but-distinct comment about the deleted-config edge case:

> If the path came from an explicit `config_path` arg or `DEER_FLOW_EXTENSIONS_CONFIG_PATH` (which is exactly how Docker dev/prod point at it per AGENTS.md), `resolve_config_path()` itself raises `FileNotFoundError` on a deleted file rather than returning `None` — so `_is_cache_stale()` (and therefore `get_cached_mcp_tools()`) can raise before ever reaching the None-check we're discussing. Repro'd this directly against both `main` and this branch. ... Happy to file it separately if useful; flagging for visibility in the meantime.

Neither one got filed afterward. This PR is that follow-up.

### 1. Shared `deerflow.config.file_signature` helper

New module `backend/packages/harness/deerflow/config/file_signature.py` holds one `get_config_signature()` implementation (the `(mtime, size, sha256)` tuple, plus the `ConfigSignature` type alias). `config/app_config.py` and `mcp/cache.py` both now import it — aliased to their existing private names (`_get_config_signature` / `_ConfigSignature`) so every internal call site and existing test monkeypatch keeps working unchanged — instead of each maintaining its own verbatim copy. The "always hash the full file, don't short-circuit when mtime/size already match" reasoning (from willem-bd's other review comment on this PR) now lives in one place instead of only in `cache.py`'s copy.

### 2. `ExtensionsConfig.resolve_config_path()` returns `None` instead of raising

Reproduced the exact scenario described above: with `DEER_FLOW_EXTENSIONS_CONFIG_PATH` pointing at a file that existed and was then deleted, `resolve_config_path()` raised `FileNotFoundError`, and that propagated straight through `mcp/cache.py`'s `_resolve_config_path()` into `_is_cache_stale()` uncaught. Since Docker dev/prod always sets that env var (per AGENTS.md), this isn't just a cold edge case for CLI-style explicit paths — it's the normal resolution mode for that deployment target.

The method's own docstring already promised `None` ("Extensions config file if found, otherwise None... extensions are optional"), but only the no-override search branch honored it — the explicit-`config_path` and env-var branches raised instead. Both now log a warning and return `None`, matching the third branch and every observed caller (`app/gateway/routers/mcp.py`, `app/gateway/routers/skills.py`, `client.py`), which already defensively check `if config_path is None:` — none of them relied on the raise.

`AppConfig.resolve_config_path()` (for the mandatory `config.yaml`, in `config/app_config.py`) is untouched and still raises on purpose — that file isn't optional the way `extensions_config.json` is, so the two methods' contracts are meant to differ.

## Tests

- `tests/test_file_signature.py` (new) — unit tests for the shared helper, including a regression guard (`test_app_config_and_mcp_cache_share_the_same_implementation`) asserting both modules delegate to it, so it can't silently re-duplicate.
- `tests/test_runtime_paths.py` — `test_extensions_config_explicit_path_missing_file_returns_none` and `test_extensions_config_env_var_missing_file_returns_none`.
- `tests/test_mcp_cache.py` — `test_config_deleted_after_init_via_real_env_resolution_does_not_raise`, exercising the real resolver end-to-end (not monkeypatched) through `_is_cache_stale()`. Also updated the docstring on the neighboring `test_config_deleted_after_init_is_not_stale` (which deliberately monkeypatches the resolver to isolate a different, already-handled edge) since it previously described this as an unresolved issue "outside this module's scope."

All three new tests were verified fail-before/pass-after: reverted just the `resolve_config_path()` behavior change, confirmed all three fail with the same `FileNotFoundError`, then reapplied the fix and confirmed green.

## Validation

- New/touched tests: `pytest tests/test_file_signature.py tests/test_runtime_paths.py tests/test_mcp_cache.py` — 26 passed.
- Broader regression sweep across every module touching the changed files (`config/app_config`, all `mcp/*` tests, skills routers that call `ExtensionsConfig`, `client.py`, the harness-boundary test, the reload-boundary test, and the blocking-IO MCP router anchor): 600+ passed. 5 pre-existing failures in `test_mcp_session_pool.py` / `test_sandbox_tools_security.py` reproduce identically on unmodified `main` (Windows path-separator assertions, unrelated to this change) — 0 new failures introduced.
- `ruff check` and `ruff format --check` clean on all changed/added files.


fancyboi999 (changes requested):
Thanks @DaoyuanLi2816. Extracting the signature helper is clean, but changing every explicit-path resolution failure into an empty extensions config weakens a production configuration boundary.

[P1] Keep explicit missing extension configs fail-loud

- Location: `backend/packages/harness/deerflow/config/extensions_config.py:176-192`
- Problem: an explicit `config_path` or `DEER_FLOW_EXTENSIONS_CONFIG_PATH` is an operator assertion that a particular file must be used. Returning `None` makes `ExtensionsConfig.from_file()` silently construct `mcp_servers={}` and `skills={}` when that file is missing, so a bad Docker mount, typo, or deleted production config can start/continue with all configured extensions absent instead of surfacing the configuration failure. Optional default discovery is a different contract from an explicit override.
- Evidence: on head `9998aeeb7f8358f374d8791c6a0da4a9e43745dd`, I pointed `DEER_FLOW_EXTENSIONS_CONFIG_PATH` at a missing temporary path and called the real resolver plus `ExtensionsConfig.from_file()`; it logged a warning and returned `resolved=None mcp_servers={} skills={}`. `docker-compose-dev.yaml`, `docker-compose.yaml`, and `scripts/deploy.sh` all use this env-var path in normal deployments, so this is not limited to an optional local lookup. The focused tests pass (`26 passed`) because the new tests encode the weakened contract.
- Suggested fix: preserve `FileNotFoundError` for explicit argument/env-var paths and keep only fallback discovery optional. Solve the original cache-deletion edge inside `mcp/cache.py` by catching the missing explicit file while computing current cache state and retaining the last-known-good tools, without changing `ExtensionsConfig.resolve_config_path()`/`from_file()` for every caller.
- Test: restore explicit/env missing-path regressions that assert a loud failure, and separately keep an MCP-cache regression proving that deleting the previously loaded file does not drop the last-known-good cache or propagate through `get_cached_mcp_tools()`.

Reviewed head SHA: `9998aeeb7f8358f374d8791c6a0da4a9e43745dd`

willem-bd (commented):
Posting the low-severity (non-blocking) observations from review as inline notes for visibility. Overall the PR looks good — these are optional follow-ups, not blockers.

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -165,17 +165,31 @@ def resolve_config_path(cls, config_path: str | None = None) -> Path | None:
...
+                    "Extensions config file specified by param `config_path` not found at %s; treating extensions as unconfigured.",
+                    path,
+                )
+                return None
             return path
         elif os.getenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH"):
             path = Path(os.getenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH"))
             if not path.exists():
-                raise FileNotFoundError(f"Extensions config file specified by environment variable `DEER_FLOW_EXTENSIONS_CONFIG_PATH` not found at {path}")
+                logger.warning(
```
willem-bd: **Low-severity observation (non-blocking):** this warning fires on *every* MCP tools request, not once. `get_cached_mcp_tools()` → `_is_cache_stale()` → `_current_config_state()` → `resolve_config_path()` runs per request, and the env-var branch (the Docker dev/prod path) is taken whenever `DEER_FLOW_EXTENSIONS_CONFIG_PATH` is set — so a stale/deleted path logs this warning each call.

The fail-soft logic correctly returns "not stale" and keeps serving cached tools, so this is purely log noise in an uncommon state, not a correctness issue. If it gets noisy in practice, options: drop to `logger.debug` here, or cache the not-found result. The `warning` level is genuinely useful for diagnosing "why aren't my extensions loaded?", so it's a judgment call — flagging for visibility.

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -165,17 +165,31 @@ def resolve_config_path(cls, config_path: str | None = None) -> Path | None:
             4. Finally, search backend/repository-root defaults for monorepo compatibility.
 
         Returns:
-            Path to the extensions config file if found, otherwise None.
+            Path to the extensions config file if found, otherwise None. Every
```
willem-bd: **Low-severity observation (non-blocking):** the raise → `None` change is a public-API behavior change on `ExtensionsConfig.resolve_config_path()`. It aligns the explicit-arg and env-var branches with the already-documented `None` contract (this docstring) and the search branch, so it's defensible as a bug fix — but a short `CHANGELOG.md` entry would help anyone who happened to catch the previous `FileNotFoundError`. Noting since `CHANGELOG.md` isn't touched in this PR.

backend/packages/harness/deerflow/config/extensions_config.py
```diff
@@ -165,17 +165,31 @@ def resolve_config_path(cls, config_path: str | None = None) -> Path | None:
...
             if not path.exists():
-                raise FileNotFoundError(f"Extensions config file specified by param `config_path` not found at {path}")
+                logger.warning(
+                    "Extensions config file specified by param `config_path` not found at %s; treating extensions as unconfigured.",
+                    path,
+                )
+                return None
             return path
         elif os.getenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH"):
             path = Path(os.getenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH"))
```
willem-bd: **Pre-existing nit (optional, not introduced by this PR):** `os.getenv("DEER_FLOW_EXTENSIONS_CONFIG_PATH")` is evaluated twice in this branch — once in the `elif` on line 185 and again here. Could be hoisted to a local (`env_path = os.getenv(...)`). Just flagging since this branch is being touched; feel free to ignore.

DaoyuanLi2816:
@willem-bd — thanks for the three inline notes; addressing those plus a scope issue I found while reviewing this further.

`resolve_config_path()` now distinguishes **explicit** configuration from **fallback search** (pushed in 68f2d66).

**What changed** (`backend/packages/harness/deerflow/config/extensions_config.py`):
- An explicit `config_path` argument or `DEER_FLOW_EXTENSIONS_CONFIG_PATH` (the mode Docker dev/prod use) is an operator assertion that one specific file must be used. A missing file in either mode now raises `FileNotFoundError` again (restoring the pre-#4124 behavior) instead of silently building an empty `mcp_servers={}`/`skills={}` config — a bad Docker mount, typo, or deleted production config now surfaces loudly instead of quietly running with no extensions.
- Only the fallback *search* mode (no explicit path/env var; nothing found in the project root or legacy backend/repo-root locations) still returns `None`. That's the genuine "extensions were never configured" case the original #4124 fix was actually about, and it still needs to degrade silently — that part of the old fix was correct and is untouched.

**Why the MCP cache doesn't regress**: `get_cached_mcp_tools()` → `_is_cache_stale()` re-resolves the config path on every call (with no try/except around it) to detect a path/content change. If `resolve_config_path()` raised unconditionally for a previously-valid-but-now-deleted explicit/env-var path, that raise would propagate straight out of this hot per-request path. So the fail-soft handling now lives narrowly in `deerflow/mcp/cache.py::_resolve_config_path()`: it catches that specific `FileNotFoundError` locally and treats it as "unconfigured," so a config file disappearing mid-run still degrades the cache to "not stale" (serving last-known-good tools) instead of crashing every subsequent request.

I audited every other caller of `resolve_config_path()` / `from_file()` (`mcp/tools.py`, `app/gateway/routers/mcp.py`, `app/gateway/routers/skills.py`, `client.py`): they either already have broad exception handling that converts a raise into a proper error response, or explicitly check for `None` themselves for the search-mode contract (unaffected by this change). None of them silently swallow a raised `FileNotFoundError` into incorrect behavior.

**@willem-bd**, on your three inline notes:
- Per-request warning noise: resolved as a side effect of the above. `resolve_config_path()` no longer logs anything on this path — it raises once, loudly, for a real misconfiguration. The new cache-side catch logs at `debug`, not `warning`, so this doesn't reintroduce request-volume log spam.
- CHANGELOG entry: since the "raise → None" behavior you flagged never shipped in a release (this PR hasn't merged yet) and this follow-up restores the original raising contract for explicit paths, there's no net public-API change from the last released behavior left to document, so I skipped the entry.
- Double `os.getenv()` read: hoisted into a local (`env_path := os.getenv(...)`) since I was already touching this branch.

**Tests**: `backend/tests/test_runtime_paths.py` now has explicit-path-raises and env-var-raises resolver tests (verified fail-before/pass-after against the prior code), plus a dedicated search-mode "nothing found → `None`" regression test. `backend/tests/test_mcp_cache.py`'s existing `test_config_deleted_after_init_via_real_env_resolution_does_not_raise` continues to pass — I confirmed it would *not* (crashes with the raised `FileNotFoundError`) without the new cache-side catch, which is what proves that catch is load-bearing rather than incidental. `backend/AGENTS.md` updated in both places it documented the old contract.

WillemJiang (approved):

PR #4280, opened 2026-07-18
fix(wechat): validate timing configuration

18062706139fcz:
## Summary

- Reject non-positive and non-finite WeChat timing values.
- Fall back to defaults for invalid `polling_timeout`, `polling_retry_delay`, `qrcode_poll_interval`, and `qrcode_poll_timeout` values.
- Document the timing configuration contract and add regression coverage.

## Why

Values such as `0`, negative numbers, `NaN`, and infinities can produce a hot polling loop or invalid sleep/timeout behavior. The channel now accepts only positive finite seconds.

## Validation

- `PYTHONPATH=. uv run pytest tests/test_wechat_channel.py -q` — 34 passed
- `uv run ruff check app/channels/wechat.py tests/test_wechat_channel.py`
- `uv run ruff format --check app/channels/wechat.py tests/test_wechat_channel.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,7 +427,10 @@ channels:
     ilink_bot_id: $WECHAT_ILINK_BOT_ID
     qrcode_login_enabled: true      # optional: allow first-time QR bootstrap when bot_token is absent
     allowed_users: []               # empty = allow all
-    polling_timeout: 35
+    polling_timeout: 35             # timing values must be positive finite seconds
+    polling_retry_delay: 5
+    qrcode_poll_interval: 2
+    qrcode_poll_timeout: 180
     state_dir: ./.deer-flow/wechat/state
     max_inbound_image_bytes: 20971520
     max_outbound_image_bytes: 20971520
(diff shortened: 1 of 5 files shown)


fancyboi999 (changes requested):
Thanks @18062706139fcz. I found one edge case in the new validation contract that should be addressed before this is ready.


Reviewed head SHA: `504b30976d01ef82438e479b16e5c874abd9a863`

### [P2] Treat overflowing numeric config values as invalid

- Location: `backend/app/channels/wechat.py:1459-1462`
- Problem: The newly documented fallback contract still lets a valid YAML numeric value abort `WechatChannel` construction instead of falling back to the configured default. A sufficiently large integer makes `float(value)` raise `OverflowError`, which is not handled by the current exception clause.
- Evidence: On this exact head, parsing `polling_retry_delay:` followed by a 1000-digit positive integer with `yaml.safe_load()` produces a Python `int`; passing that config to `WechatChannel(...)` raises `OverflowError: int too large to convert to float`. The added regression covers zero, negatives, NaN, and infinities, but not conversion overflow.
- Suggested fix: Treat `OverflowError` like `TypeError` and `ValueError` in `_coerce_float` so values that cannot be represented as finite floats use the default.
- Test: Add an overflowing integer case (for example `10**1000`) to the timing-config regression and assert all four fields fall back without aborting channel initialization.

WillemJiang (approved):

PR #4281, opened 2026-07-18
fix(artifacts): serve inline binary artifacts via FileResponse for Range support

### LINKED ISSUE #3240: 音频进度条无法拖动
我拖动音频的进度条，松开鼠标后进度条回到了起点，音频从头开始播放。我尝试自己修改了代码，但是发现还是一样会回到起点。这是因为什么原因？如何进行优化？

DaoyuanLi2816:
Fixes #3240

## Why

Issue #3240 reports that dragging an audio artifact's progress bar resets
playback to the start instead of seeking. The report was thin (one line,
no repro steps), so before attempting a fix I traced the actual playback
path: the frontend has no custom audio player — `ArtifactFileDetail` embeds
previewable binary artifacts (audio/video/images) via a plain
`<iframe src={urlOfArtifact(...)}>` pointed straight at the backend
artifact endpoint, so the browser's native media engine owns seeking.

That pointed at the backend. `GET /api/threads/{thread_id}/artifacts/{path}`
had three response shapes: `FileResponse` for active-content/explicit
downloads, `PlainTextResponse` for text, and a plain buffered `Response`
for everything else (images, audio, video previewed inline). That last
branch read the whole file into memory and returned it with no
`Accept-Ranges` header, and Starlette's base `Response` has no Range
parsing at all — it always answers `200` with the full body regardless of
any `Range` header sent. When a browser seeks an `<audio>`/`<video>`
element backed by a remote URL, it issues a ranged `GET`; since this
endpoint never returns `206`, most engines fall back to restarting
playback from byte 0 — exactly the reported symptom.

## What changed

- Binary artifacts that are previewed inline (not downloaded, not
  active-content) are now served through `FileResponse` instead of being
  buffered into a plain `Response`, the same way the download/active-content
  branch already was. `Content-Disposition` stays `inline` for this branch
  (only download/active-content force `attachment`).
- This is a behavior-only change in a "kind" is returned from
  `_read_artifact_payload`'s existing 3-way plan (`"file"` / `"text"` /
  the new `"inline_file"`) — no new dependency, since `FileResponse`
  already handles `Range`/`If-Range` natively on the pinned Starlette
  version.
- No frontend change needed: the same `<iframe>`/URL now transparently
  gets Range support from the backend.

## Surface area

- [ ] **Frontend UI**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**

## Bug fix verification

  - `backend/tests/test_artifacts_router.py::test_get_artifact_binary_preview_is_inline_file_response`
  - `backend/tests/test_artifacts_router.py::test_get_artifact_binary_preview_supports_range_requests`
  - `backend/tests/blocking_io/test_artifacts_router.py::test_get_artifact_binary_does_not_block_event_loop` (updated for the new response shape)
- Did it go red on `main` and green on this branch? **Yes** — confirmed by
  temporarily reverting only `artifacts.py` on this branch and re-running:
  all three fail on the pre-fix code (no `FileResponse`, no
  `Accept-Ranges`, `200` instead of `206` for a ranged request) and pass
  once the fix is restored.

## Validation

```
cd backend
make lint   # ruff check . && ruff format --check .   -> clean on the changed files
make test   # full suite: 7556 passed (97 failed / 51 errors, all pre-existing
            # and unrelated to this change — sandbox/k8s-provisioner/skill-scan/
            # channel-SDK tests that need external services not present in this
            # environment; none touch artifacts.py or these test files)
```

Also verified empirically outside the app, against the exact pinned
`starlette==1.3.1` in `backend/uv.lock`: a bare `Response(content=..., ...)`
ignores a `Range: bytes=500000-` header entirely (always `200`, full body,
no `Accept-Ranges`), while `FileResponse` on the same version correctly
returns `206` with `Content-Range`/`Accept-Ranges` and only the requested
byte slice.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** Used it to investigate the report end-to-end: located
the frontend rendering path (ruled out a controlled/uncontrolled
`currentTime` bug — there is no custom player), located the backend
serving code, wrote a standalone script against the pinned Starlette
version to empirically confirm the Range-handling gap before writing any
fix, implemented the fix, wrote the regression tests, and verified
red-on-`main`/green-on-branch plus the full backend test suite.

### DIFF
--- a/backend/app/gateway/routers/artifacts.py
+++ b/backend/app/gateway/routers/artifacts.py
@@ -250,7 +250,6 @@ async def get_artifact(thread_id: str, path: str, request: Request, download: bo
         # reset playback to the start instead of jumping to the new position.
         return FileResponse(
             path=actual_path,
-            filename=actual_path.name,
             media_type=mime_type,
             headers={"Content-Disposition": _build_content_disposition("inline", actual_path.name)},
         )
--- a/backend/app/gateway/routers/artifacts.py
+++ b/backend/app/gateway/routers/artifacts.py
@@ -123,11 +123,14 @@ def _read_artifact_payload(actual_path: Path, path: str, download: bool) -> tupl
     """Worker-thread body for the regular branch of ``get_artifact``.
 
     Stat probes, MIME sniffing (``mimetypes`` lazily stats the system MIME
-    database on first use), and full-file reads are all blocking filesystem IO.
-    Returns a ``(kind, mime_type, payload)`` plan the handler turns into a
-    response on the loop: ``("file", mime, None)`` (let ``FileResponse`` stream
-    it), ``("text", mime, str)``, or ``("bytes", mime, bytes)``. Behavior/error
-    codes match the previous inline logic.
+    database on first use), and text reads are blocking filesystem IO. Returns
+    a ``(kind, mime_type, payload)`` plan the handler turns into a response on
+    the loop: ``("file", mime, None)`` (attachment / forced-download active
+    content, streamed by ``FileResponse``), ``("inline_file", mime, None)``
+    (inline binary preview — also streamed by ``FileResponse`` so the client
+    can issue byte-``Range`` requests, e.g. to seek within audio/video
+    artifacts instead of always replaying from byte 0), or ``("text", mime,
+    str)``. Behavior/error codes match the previous inline logic.
     """
     if not actual_path.exists():
         raise HTTPException(status_code=404, detail=f"Artifact not found: {path}")
@@ -141,7 +144,10 @@ def _read_artifact_payload(actual_path: Path, path: str, download: bool) -> tupl
         return ("text", mime_type, actual_path.read_text(encoding="utf-8"))
     if is_text_file_by_content(actual_path):
         return ("text", mime_type, actual_path.read_text(encoding="utf-8"))
-    return ("bytes", mime_type, actual_path.read_bytes())
+    # Binary inline preview (images, audio, video, PDFs, ...): stream via
+    # FileResponse instead of buffering the whole file in memory, so it also
+    # gets FileResponse's built-in byte-Range handling (see get_artifact).
+    return ("inline_file", mime_type, None)
 
 
 @router.get(
@@ -235,7 +241,21 @@ async def get_artifact(thread_id: str, path: str, request: Request, download: bo
         # execution in the application origin when users open generated artifacts.
         return FileResponse(path=actual_path, filename=actual_path.name, media_type=mime_type, headers=_build_attachment_headers(actual_path.name))
 
+    if kind == "inline_file":
+        # FileResponse (unlike a fully-buffered Response) honors byte-Range
+        # requests. Browsers issue these when seeking an <audio>/<video>
+        # element backed by a remote URL; serving the same bytes through a
+        # plain Response ignores Range headers and always replays from byte
+        # 0, which is why dragging an audio/video artifact's progress bar
+        # reset playback to the start instead of jumping to the new position.
+        return FileResponse(
+            path=actual_path,
+            filename=actual_path.name,
+            media_type=mime_type,
+            headers={"Content-Disposition": _build_content_disposition("inline", actual_path.name)},
+        )
+
     if kind == "text":
         return PlainTextResponse(content=payload, media_type=mime_type)
 
-    return Response(content=payload, media_type=mime_type, headers={"Content-Disposition": _build_content_disposition("inline", actual_path.name)})
+    raise AssertionError(f"Unhandled artifact response kind: {kind!r}")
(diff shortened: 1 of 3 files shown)


willem-bd (commented):
Reviewed against head `9000fe9f`. Clean fix - the root-cause trace (frontend `<iframe>` -> backend plain `Response` ignoring `Range` -> empirically confirmed against the pinned Starlette) is exactly right, and switching the inline binary branch to `FileResponse` is a real memory/latency win on top of fixing the seek bug.

Verified the security property directly: `_read_artifact_payload` routes `download or mime_type in ACTIVE_CONTENT_MIME_TYPES` to the `"file"` (attachment) branch *before* the text/binary checks, so HTML/XHTML/SVG still force download - no XSS regression. The existing `test_get_artifact_forces_download_for_active_content` already pins this. The new `test_get_artifact_binary_preview_supports_range_requests` is a strong regression test (1 MB all-bytes payload; asserts `200` + `accept-ranges` and `206` + `content-range` + correct slice for `Range: bytes=500000-`).

Two minor inline nits, neither blocking. Other minor notes not worth a line comment: (1) the new `raise AssertionError(...)` fallback is a good strictness improvement (`raise AssertionError` is NOT stripped by `python -O` - only the bare `assert` statement is), so the guard is safe in optimized runs; `RuntimeError` would be a touch more conventional for "unreachable" guards but it's a style toss-up. (2) The deferred `FileResponse` open slightly widens the TOCTOU window (file deleted between the worker-thread `exists()` check and ASGI send) from downloads to all inline binary previews, but the attachment branch already has this behavior, so it's an accepted trade-off, not a regression.

Approve.

backend/app/gateway/routers/artifacts.py
```diff
@@ -235,7 +241,21 @@ async def get_artifact(thread_id: str, path: str, request: Request, download: bo
...
+        # requests. Browsers issue these when seeking an <audio>/<video>
+        # element backed by a remote URL; serving the same bytes through a
+        # plain Response ignores Range headers and always replays from byte
+        # 0, which is why dragging an audio/video artifact's progress bar
+        # reset playback to the start instead of jumping to the new position.
+        return FileResponse(
+            path=actual_path,
+            filename=actual_path.name,
+            media_type=mime_type,
+            headers={"Content-Disposition": _build_content_disposition("inline", actual_path.name)},
```
willem-bd, line 254: Minor: the explicit `Content-Disposition: inline` header here makes `filename=actual_path.name` redundant - Starlette's `FileResponse` derives `content-disposition` from `filename` via `setdefault`, which the explicit header already satisfies (so `inline` wins, as the tests pin). Harmless today, and the `"file"` (attachment) branch above uses the same pattern, so no need to change it - just flagging: if a future Starlette version ever switched `setdefault` to a hard set, inline preview could silently flip to `attachment`. Could drop `filename=` when an explicit disposition is passed.

backend/tests/blocking_io/test_artifacts_router.py
```diff
@@ -1,11 +1,17 @@
...
+and extracts ``.skill`` archive members — all blocking filesystem IO. The
+handler offloads each branch's IO via ``asyncio.to_thread``; if any regresses
+back onto the event loop, the strict Blockbuster gate raises ``BlockingError``
+and these tests fail.
+
+Binary (non-text, non-active-content) artifacts are handled by returning a
+``FileResponse`` that defers its own file IO to ASGI response time (streamed
+via ``anyio.open_file`` when the response is actually sent, which this test
+never triggers) — so awaiting ``get_artifact`` itself for a binary artifact
+does zero filesystem IO and the gate has nothing to catch there either way.
```
willem-bd: This is slightly inaccurate and contradicts the first paragraph. `_read_artifact_payload` still runs `exists()`, `is_file()`, `mimetypes.guess_type`, and `is_text_file_by_content` (an 8 KB read) for a binary file - all filesystem IO, just offloaded to the worker thread via `asyncio.to_thread`. The Blockbuster gate has nothing to catch because that IO is off the event loop, not because there's zero IO. Suggest rewording to something like "does no full-file read - `FileResponse` streams the bytes at ASGI send time, off the loop" so it matches the first paragraph.

DaoyuanLi2816:
Thanks for the review @willem-bd -- addressed the two nits from your review in 2ae6d8d1:

- Dropped the redundant `filename=` kwarg on the `inline_file` `FileResponse` call, since `Content-Disposition` is already set explicitly there.
- Reworded the blocking-IO regression-anchor docstring's binary-artifact paragraph so it no longer claims "zero filesystem IO" -- `_read_artifact_payload` still runs the same probes for binary files, just offloaded via `asyncio.to_thread` like the text branch; only the full-file read is avoided.

`test_artifacts_router.py` + `blocking_io/test_artifacts_router.py` (19 tests) are green, ruff check/format clean.

WillemJiang (approved):

PR #4282, opened 2026-07-18
feat(frontend): restore per-thread composer drafts

18062706139fcz:
## Why

Unsent composer text currently lives only in React state. Reloading the page or moving between conversations can discard a partially written prompt, including a selected slash skill.

New-chat routes need additional handling because DeerFlow creates a fresh temporary runtime thread ID on every reload. Using that UUID as the storage key would make the saved draft unreachable. Custom-agent chats also need an explicit agent scope so drafts from two agents never share the same new-chat key.

## What

- Add a versioned, failure-tolerant composer draft storage helper backed by `sessionStorage`.
- Scope drafts by user, agent, and logical conversation ID.
- Use the stable logical ID `new` until a new conversation receives its real thread ID.
- Persist draft text and the selected slash-skill name after a short debounce.
- Flush the latest draft during page unload or conversation switches so fast navigation does not lose input.
- Clear the stored draft only after a send passes the existing in-flight guard.
- Restore enabled skills as chips; degrade missing or disabled skills to editable slash text.
- Keep attachments, quoted conversation context, voice state, and polish undo state out of browser storage.

## Change Scope

- `frontend/src/core/threads/composer-draft.ts`
  - Storage key construction, payload validation, read/write/clear behavior, and skill fallback.
- `frontend/src/components/workspace/input-box.tsx`
  - Draft hydration, debounced persistence, lifecycle flushes, and accepted-send cleanup.
- Main and custom-agent chat pages
  - Stable new-chat scope and explicit custom-agent draft scope.
- Tests and documentation
  - Unit coverage, Chromium E2E scenarios, README, and frontend architecture guidance.

No Gateway API, database schema, LangGraph state, attachment persistence, or cross-tab synchronization changes are included.

## Scenarios

- Reload an unfinished new chat and recover its text.
- Reload a draft with a selected slash skill and recover both the chip and text.
- Send the prompt and confirm the draft is not restored afterward.
- Move between custom agents without leaking one agent's new-chat draft into another.
- Disable or remove a previously selected skill and recover the full prompt as editable slash text.
- Continue typing normally when browser storage is unavailable or full.

## Validation

- `cd frontend && pnpm check`
- `cd frontend && pnpm test` (72 files, 649 tests)
- `cd frontend && playwright test tests/e2e/chat.spec.ts` (22 passed)
- `cd frontend && playwright test tests/e2e/agent-chat.spec.ts --grep "isolated"` (1 passed)
- `bits-code-guard` review completed; the custom-agent scope issue found during review was fixed, and the final review reported no remaining P0-P2 findings.


fancyboi999 (changes requested):
Thanks @18062706139fcz. I found two draft-lifecycle issues that should be addressed before this is ready.


Reviewed head SHA: `93c4207b8558f20c283191127181def03f646f09`

### [P2] Acquire `sessionStorage` inside the failure boundary

- Location: `frontend/src/components/workspace/input-box.tsx:649` (and the corresponding accesses at 598, 709, and 1025)
- Problem: The helpers catch failures from `getItem`/`setItem`/`removeItem`, but each call evaluates `window.sessionStorage` before entering that `try`. Browsers can throw `SecurityError` from the property getter itself when persistence is blocked, so the new effect throws and the composer does not render instead of degrading to in-memory typing as the PR promises.
- Evidence: A real Chromium probe replaced the `sessionStorage` getter with a standard `DOMException("Blocked", "SecurityError")`; `/workspace/chats/new` never rendered the textarea. This repository already handles the same getter-level failure correctly in `core/api/api-client.ts::clearReconnectRun` by acquiring the storage object inside its `try`.
- Suggested fix: Add one safe accessor that reads `window.sessionStorage` inside `try/catch` and returns `null` on failure, then make every draft read/write/clear/flush path tolerate `null` (or move storage acquisition into the helper boundary itself).
- Test: Add a component or browser regression where the `sessionStorage` getter throws and verify that the composer still renders, accepts edits, and sends normally.

### [P2] Cancel pending debounce writes when an attachment send is accepted

- Location: `frontend/src/components/workspace/input-box.tsx:708-711`
- Problem: `onSent` clears the stored draft at 1021-1026 but does not cancel the timeout already scheduled by this effect. The page deliberately returns the `sendMessage` promise for messages with attachments, so the input state and its old timeout remain alive while upload/streaming is in flight; the timeout can recreate the just-cleared draft. Reloading during that window restores text that has already passed the send guard.
- Evidence: In Chromium, I first persisted a draft, changed it to `Send this immediately`, attached a file, and delayed the upload response. After `onSent` had run but before upload completion, `sessionStorage` again contained `{"version":1,"text":"Send this immediately","skillName":null}`.
- Suggested fix: Keep the debounce timer (or a per-key write generation) in a ref and cancel/invalidate it before clearing an accepted submission. A stale callback must not be able to write after the submission tombstone for that key.
- Test: Add an E2E case with a delayed attachment upload: edit and send inside the debounce window, then assert the draft key stays absent and a reload does not restore the accepted text.

18062706139fcz:
Addressed the two requested draft-lifecycle fixes.

What changed:
- Added `getSessionComposerDraftStorage()` so `window.sessionStorage` is acquired inside a `try/catch`. All draft read/write/clear paths now accept `null` storage and degrade to in-memory typing when browser persistence is blocked.
- Added a debounce timer ref plus accepted-draft tombstone. When a send passes the existing in-flight guard, the pending draft write is cancelled and stale callbacks cannot recreate the just-cleared draft, including delayed attachment uploads.
- Kept the earlier user/agent/thread scoping and disabled-skill fallback behavior.

New coverage:
- Unit: missing storage and getter-level `SecurityError`.
- E2E: composer still sends when sessionStorage is blocked after load.
- E2E: delayed attachment upload does not rewrite an accepted draft from a stale debounce.

Validation run locally:
- `cd frontend && pnpm check`
- `cd frontend && pnpm test` (72 files, 651 tests)
- `cd frontend && playwright test tests/e2e/chat.spec.ts` (24 passed)
- `cd frontend && playwright test tests/e2e/agent-chat.spec.ts --grep "isolated"` (1 passed)
- `bits-code-guard` focused review on the review-fix diff: no remaining P0-P2 findings.

fancyboi999 (changes requested):
Thanks @18062706139fcz. The two earlier lifecycle cases are fixed on this head, but the new tombstone logic introduces one repeat-draft loss path that should be addressed before this is ready.


Reviewed head SHA: `5f2da43750608880ec684e063d5a24155ef5a4c7`

### [P2] Do not treat a later identical draft as the already-accepted submission

- Location: `frontend/src/components/workspace/input-box.tsx:736-738`
- Problem: `acceptedDraftRef` is retained after the composer clears and suppresses persistence by comparing text and skill values. If the user later types the same prompt with the same selected skill again, this effect returns before clearing the tombstone or updating `latestDraftRef`; no debounce is scheduled, and pagehide still has no new draft to flush. Reloading therefore loses a legitimate new draft solely because its content matches the last sent message.
- Evidence: On this exact head in real Chromium, I sent `Repeat this request`, waited for the textarea to clear, entered `Repeat this request` again without sending, and waited past the 300 ms debounce. `sessionStorage` remained empty through a 5-second poll. The two newly added lifecycle E2Es pass, as do all 651 frontend unit tests, so this transition is currently uncovered rather than caught by the suite.
- Suggested fix: Invalidate stale callbacks with a per-write generation/timer identity when `onSent` runs instead of keeping a content-valued tombstone. The invalidation should block callbacks created before acceptance while allowing any later edit event, including an identical value, to start a new draft generation.
- Test: Add an E2E regression that sends a prompt, waits for the controlled input to clear, types the exact same text again without sending, waits past the debounce, reloads, and verifies that the repeated draft is restored.

WillemJiang (approved):

PR #4284, opened 2026-07-18
fix(gateway): reject non-positive read limits

18062706139fcz:
## Summary

- Reject non-positive `limit` values on legacy thread-message reads.
- Apply the same lower-bound validation to run-event audit reads.
- Add endpoint-level regression tests and document the supported limit ranges.

## Root Cause

The affected Gateway read endpoints only declared upper bounds for `limit`, so FastAPI accepted zero and negative values and passed them into storage-dependent query behavior.

## Impact

Invalid limits now fail consistently with HTTP 422. Existing defaults and maximum limits are unchanged.

## Validation

- `backend/.venv/bin/pytest tests/test_thread_run_query_validation.py -q` after rebasing onto latest `origin/main`: 4 passed
- `backend/.venv/bin/ruff check .`: passed
- `backend/.venv/bin/ruff format --check .`: 877 files already formatted
- `git diff --check origin/main...HEAD`: passed
- Pre-rebase full backend suite: 7783 passed, 28 skipped

### DIFF
--- a/backend/app/gateway/routers/thread_runs.py
+++ b/backend/app/gateway/routers/thread_runs.py
@@ -708,9 +708,9 @@ async def stream_existing_run(
 async def list_thread_messages(
     thread_id: str,
     request: Request,
-    limit: int = Query(default=50, le=200),
-    before_seq: int | None = Query(default=None),
-    after_seq: int | None = Query(default=None),
+    limit: int = Query(default=50, ge=1, le=200),
+    before_seq: int | None = Query(default=None, ge=1),
+    after_seq: int | None = Query(default=None, ge=1),
 ) -> list[dict]:
     """Return displayable messages for a thread (across all runs), with feedback attached."""
     event_store = get_run_event_store(request)
@@ -910,8 +910,8 @@ async def list_run_messages(
     run_id: str,
     request: Request,
     limit: int = Query(default=50, le=200, ge=1),
-    before_seq: int | None = Query(default=None),
-    after_seq: int | None = Query(default=None),
+    before_seq: int | None = Query(default=None, ge=1),
+    after_seq: int | None = Query(default=None, ge=1),
 ) -> dict:
     """Return paginated messages for a specific run.
 
@@ -954,8 +954,8 @@ async def list_run_events(
     request: Request,
     event_types: str | None = Query(default=None),
     task_id: str | None = Query(default=None),
-    limit: int = Query(default=500, le=2000),
-    after_seq: int | None = Query(default=None),
+    limit: int = Query(default=500, ge=1, le=2000),
+    after_seq: int | None = Query(default=None, ge=1),
 ) -> list[dict]:
     """Return the full event stream for a run (debug/audit).
 
--- a/backend/tests/test_thread_run_query_validation.py
+++ b/backend/tests/test_thread_run_query_validation.py
@@ -1,0 +1,87 @@
+"""Query validation for thread message and run event read endpoints."""
+
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+from _router_auth_helpers import make_authed_test_app
+from fastapi.testclient import TestClient
+
+from app.gateway.routers import thread_runs
+
+
+def _make_app():
+    app = make_authed_test_app()
+    app.include_router(thread_runs.router)
+
+    event_store = MagicMock()
+    event_store.list_messages = AsyncMock(return_value=[])
+    event_store.list_messages_by_run = AsyncMock(return_value=[])
+    event_store.list_events = AsyncMock(return_value=[])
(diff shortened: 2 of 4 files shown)


backend/app/gateway/routers/thread_runs.py
```diff
@@ -708,7 +708,7 @@ async def stream_existing_run(
 async def list_thread_messages(
     thread_id: str,
     request: Request,
-    limit: int = Query(default=50, le=200),
+    limit: int = Query(default=50, ge=1, le=200),
```
willem-bd, line 711: Good catch — the substantive risk here is negative `limit`: `LIMIT -1` is treated as "no limit" in PostgreSQL and SQLite, so `limit=-1` would have flowed straight into `list_messages` and returned the thread's entire message table (unbounded read). `ge=1` closes that cleanly.

One behavior change to flag: `limit=0` previously returned `[]` and now returns 422. That's defensible (0 is nonsensical for a list endpoint) and the same gap is closed on `GET /{rid}/events` below, but worth confirming no frontend / IM-channel caller sends `limit=0` today.

backend/app/gateway/routers/thread_runs.py
```diff
@@ -708,7 +708,7 @@ async def stream_existing_run(
...
     before_seq: int | None = Query(default=None),
```
willem-bd: Follow-up (out of scope for this PR): the `before_seq`/`after_seq` cursors here, and `after_seq` on `GET /{rid}/events` (line 958), have no lower-bound validation, while `GET /../messages/page` validates `before_seq` with `ge=1`. Sequence values start at 1, so a non-positive cursor is suspect and gets passed straight to the store query. Adding `ge=1` to these cursors would make validation consistent across the read surface — happy to do that as a separate change.

backend/tests/test_thread_run_query_validation.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def _make_app():
+    app = make_authed_test_app()
+    app.include_router(thread_runs.router)
+
+    event_store = MagicMock()
+    event_store.list_messages = AsyncMock(return_value=[])
+    event_store.list_events = AsyncMock(return_value=[])
+    app.state.run_event_store = event_store
```
willem-bd, line 20: These store / run_manager mocks aren't actually hit on the rejection paths: FastAPI raises 422 during query validation, before the handler runs, so they only serve app construction here. The test is valid for what it asserts.

Optional strengthening: add a positive case (e.g. `limit=1` -> 200 with the mocked store returning `[]`) to exercise the store wiring and confirm valid limits still pass through. `_make_app()` could become a fixture if it gets reused.

backend/AGENTS.md
```diff
@@ -332,7 +332,7 @@ CORS is same-origin by default when requests enter through nginx on port 2026. S
...
+| **Thread Runs** (`/api/threads/{id}/runs`) | `POST /` - create background run; `POST /stream` - create + SSE stream; `POST /wait` - create + block; `POST /regenerate/prepare` - prepare clean input + checkpoint metadata for regenerating the latest assistant answer; `GET /` - list runs; `GET /{rid}` - run details; `POST /{rid}/cancel` - cancel; `GET /{rid}/join` - join SSE; `GET /{rid}/messages` - paginated per-run messages `{data, has_more}`; `GET /{rid}/events` - full event stream (`limit` 1..2000); `GET /{rid}/workspace-changes` - workspace/output file change summary and optional diffs; `GET /../messages` - legacy thread message array (`limit` 1..200); `GET /../messages/page` - backward thread-global `seq` history page with middleware/successful-regenerate filtering and page-run-scoped feedback enrichment; `GET /../token-usage` - aggregate tokens |
```
willem-bd: Nit: the inline range annotations are selective. `GET /{rid}/messages` (per-run) and `GET /../messages/page` also enforce `ge=1, le=200` but aren't annotated here. Either annotate every limit-bearing endpoint or drop the inline notes to keep the row scannable — just the two currently annotated feels arbitrary.

willem-bd (commented):
Reviewed against head `e8746600`. Clean, well-scoped fix — the two `limit` params touched here were the only ones in the gateway still missing `ge=1` (I grepped every `limit: int = Query(` under `backend/app/`; everything else already had it). The genuine bug being closed is negative `limit` → unbounded read (`LIMIT -1` = no limit in PG/SQLite). The test follows the existing `_router_auth_helpers` convention and asserts the right thing.

Inline notes are all nits / optional follow-ups — none block this. (One more minor, not worth a line comment: the new README sentence is accurate but is fairly endpoint-specific detail for the top-level routing paragraph; `backend/AGENTS.md` already captures the ranges, so the README line could arguably drop.)

WillemJiang (approved):

PR #4286, opened 2026-07-18
feat(memory): memory template externalization

### LINKED ISSUE #4285: [feat] Externalize memory extraction prompts and signal detection patterns to YAML
### Before you start

### Problem / motivation

Memory extraction prompts (MEMORY_UPDATE_PROMPT, STALENESS_REVIEW_PROMPT, etc.) and correction/reinforcement signal   
  detection patterns are currently hardcoded as Python string constants in the DeerMem backend. To customise prompt     
  wording, add a language's confirmation phrases, or adjust extraction guidelines, operators must edit Python code and  
  redeploy.                                                                                                             
                                                                                                                        
  This is particularly painful for:                                                                                     
  - Non-English deployments that need localised signal patterns (e.g. Korean confirmation phrases).                     
  - Operators who want to iterate on extraction prompt quality without touching code.                                   
  - Upstream prompt evolution: when the prompt content changes (e.g. `staleFactsToExtend` in #4143), downstream         
  customisation is a code merge, not a config update.

### Proposed solution

Externalize four memory extraction prompts and two signal-detection pattern sets to YAML files bundled in the DeerMem
  package:

  **Prompts** (under `core/prompts/`):
  - `memory_update.chat.yaml` — chat format (system/user message split), so the extraction LLM's static rules are
  prefix-cache friendly. Uses `.format()` syntax, same placeholders as before.
  - `staleness_review.yaml`, `consolidation.yaml`, `fact_extraction.yaml` — text format, loaded via the same
  `load_prompt()` function.

  **Signal patterns** (under `core/message_patterns/`):
  - `correction.yaml` / `reinforcement.yaml` — regex patterns with per-pattern flags (`ignorecase`). Loaded via
  `load_patterns()`. A `patterns_dir` config field lets operators point at a custom directory to override bundled
  patterns without code changes.

  **Key behaviours**:
  - **Zero-config = current behaviour**: bundled yaml files mirror the pre-externalization constants byte-for-byte (or,
  for memory_update, the same prompt content split into system/user messages).
  - **Detect functions backward-compatible**: `detect_correction` / `detect_reinforcement` gain a keyword-only
  `patterns=` parameter (None → bundled defaults). Existing callers are unchanged.
  - **Invalid YAML fails loudly**: malformed yaml raises `ValueError` with the file path — no silent fallback to empty
  prompts or disabled detection.
  - **No new dependencies**: uses pyyaml (already a dependency); `.format()` syntax (no Jinja2).

  `config.example.yaml` is updated with the new `patterns_dir` field under `memory.backend_config`.

### Affected area(s)

### Alternatives considered

_No response_

### Additional context

_No response_

lllyfff:
Fixes #4285

## Why

  Two categories of hardcoded content in the DeerMem backend force operators to edit Python code:

  - **Extraction prompts**: `MEMORY_UPDATE_PROMPT`, `STALENESS_REVIEW_PROMPT`, `CONSOLIDATION_PROMPT`,
  `FACT_EXTRACTION_PROMPT` are module-level string constants. Updating wording, guidelines, or the output JSON schema
  requires a code change and redeploy.
  - **Signal detection patterns**: correction/reinforcement regex patterns (e.g. "that's wrong", "不对", "perfect") are
  hardcoded tuples. Adding a language's confirmation phrases means editing Python, locating the right source file, and
  understanding the re.compile call pattern.

  During upstream prompt evolution (e.g. `staleFactsToExtend` in #4143), merging downstream prompt customisation is a
  code conflict, not a config update.

  ## What changed

  **Extraction prompts are now YAML files** under `core/prompts/`:
  - `memory_update.chat.yaml` — chat format (system/user message split), prefix-cache friendly.
  - `staleness_review.yaml`, `consolidation.yaml`, `fact_extraction.yaml` — text format.

  Loaded via `load_prompt(name)` and `load_prompt_messages(name, variables)`. Zero-config behaviour is unchanged:
  bundled yaml files mirror the former constants.

  **Signal detection patterns are now YAML files** under `core/message_patterns/`:
  - `correction.yaml` / `reinforcement.yaml` — per-pattern regex with optional `flags` (`ignorecase`).

  Loaded via `load_patterns(name)`. A new `patterns_dir` config field under `memory.backend_config` lets operators
  override the bundled patterns with a custom directory — add a language without touching code.

  **Key behaviours**:
  - `detect_correction` / `detect_reinforcement` gain a keyword-only `patterns=` parameter (None → bundled defaults).
  Existing callers are unchanged.
  - Invalid YAML raises `ValueError` with the file path (no silent fallback).
  - No new dependencies (pyyaml already present; `.format()` syntax, no Jinja2).

  ## Surface area

  - [ ] **Dependencies** — new/upgraded entry in `backend/pyproject.toml` or `frontend/package.json` (say what it buys
  us)
  - [ ] **Default behavior change** — changes existing behavior without the user opting in (default model, default
  setting, data shape)

  ## Validation

  ```bash

  - make lint (ruff check + ruff format --check) — passed
  - Memory test suite: 218 passed, 1 skipped

  AI assistance

  Tool(s) used: Claude Code

  How you used it: Spec-driven implementation — generated the loader functions, yaml files, deer_mem wiring, and tests
  from the plugin design documents. Reviewed every line and iterated on error handling and upstream compatibility under
  direction.

  - [x] I've read and understand every line of this change and take responsibility for it — it's not unreviewed AI
  output.


fancyboi999 (changes requested):
Thanks @lllyfff. The bundled assets build and the focused tests pass, but the operator-facing override path is not wired through the runtime and the new configuration can silently disable signal detection.


Reviewed head SHA: `87653288feab97c360304e7db049d81ef42563ac`

[P1] Wire prompt overrides into the actual DeerMem update path

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/prompt.py:40-132`; `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:713-762`
- Problem: `load_prompt()` / `load_prompt_messages()` expose `prompts_dir` and per-agent lookup, but `DeerMemConfig` has no `prompts_dir`, `MemoryUpdater` never receives one, and `_prepare_update_prompt()` calls the loader without it. The other three templates are loaded into module constants from the bundled directory at import time. Operators therefore cannot select a custom prompt directory through `memory.backend_config`, and the per-agent override exercised by `test_prompt_external.py` is unreachable from a real memory update. This leaves the main customization goal of #4285 non-functional outside direct helper calls or editing the installed package itself.
- Evidence: repository-wide search on this head finds `prompts_dir` only in `core/prompt.py` and `tests/test_prompt_external.py`; the production config contains only `patterns_dir`. A real update always calls `load_prompt_messages("memory_update", variables, agent_name=agent_name)` with the bundled base, while staleness/consolidation/fact-extraction use bundled module-level constants.
- Suggested fix: add and document a prompt-directory field in `DeerMemConfig`, pass it into `MemoryUpdater`, and resolve all four templates through the instance configuration. Preserve the agent-specific lookup for the active `agent_name`; avoid module-level constants for templates that are meant to be overridden.
- Test: construct DeerMem through the normal factory with a custom prompt directory and a recording model, run a default-agent and custom-agent update, and assert the actual `model.invoke()` messages contain the configured default/agent markers. Exercise staleness, consolidation, and fact extraction through their production callers too, not only `load_prompt()` directly.

[P2] Fail loud when an explicit pattern directory is incomplete

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/message_processing.py:41-56`
- Problem: when `patterns_dir` is explicitly configured but either YAML file is missing or unreadable, `load_patterns()` logs a warning, caches `[]`, and DeerMem starts with correction/reinforcement detection disabled. A typo, missing mount, or permission regression silently changes memory semantics for the lifetime of the process; this contradicts the issue's fail-loud configuration boundary.
- Evidence: the new test `test_load_patterns_missing_file_returns_empty` locks in this behavior. `DeerMem.__init__()` stores those empty lists and `_prepare_update()` then treats an explicit phrase such as `That's wrong` as neither correction nor reinforcement.
- Suggested fix: use bundled defaults only when `patterns_dir is None`. For an explicit directory, require both files to exist, be readable, and contain valid entries; raise a path-specific configuration error instead of caching an empty detector.
- Test: construct DeerMem with an explicit directory missing each file in turn and with an unreadable/invalid file, and assert startup fails loudly. Keep a zero-config test proving bundled patterns still load.

[P2] Version and document the new runtime configuration

- Location: `config.example.yaml:1561-1569`; `config.example.yaml:18`
- Problem: the PR adds `memory.backend_config.patterns_dir` to the schema but leaves `config_version` at 26 and does not update the Helm mirror or user/developer docs. Existing installations receive no outdated-config warning, and the repository's documented schema-change contract requires a version bump so `make config-upgrade` users can discover the new field.
- Evidence: `backend/AGENTS.md:269` explicitly requires bumping `config_version` for every config schema change. Both `config.example.yaml` and `deploy/helm/deer-flow/values.yaml` remain at 26, and no README/AGENTS/configuration doc is changed.
- Suggested fix: bump the example version and Helm embedded version together, update the Helm README mirror required by the config-version check, and document the effective prompt/pattern override paths and their failure behavior.
- Test: run `scripts/check_config_version.sh` and add/adjust an upgrade fixture proving a version-26 config gains the new default fields without changing existing values.

lllyfff:
@fancyboi999  Thanks for the thorough review ， all three addressed:

  **P1 (prompts_dir wiring):** Added `prompts_dir` to `DeerMemConfig`, threaded through `DeerMem` →
  `MemoryUpdater.__init__()` → `_prepare_update_prompt()`, and the two section builders (`_build_staleness_section`,
  `_build_consolidation_section`) now accept a keyword `prompts_dir=` and call `load_prompt()` instead of relying on the
  module-level shim constants. Also added `_PROMPT_CACHE` to `load_prompt()` so the per-cycle lookup is a dict hit, not
  a yaml re-read.

  **P2 (explicit patterns_dir must not silently degrade):** When `patterns_dir` is explicitly set and a YAML file is
  missing, `load_patterns()` now raises `FileNotFoundError`. Bundled defaults (`patterns_dir=None`) still log a warning
  and return `[]` (that's a packaging bug, not a configuration error).

  **P3 (config_version + Helm):** Bumped `config_version` 26 → 27 in `config.example.yaml`,
  `deploy/helm/deer-flow/values.yaml`, and `deploy/helm/deer-flow/README.md`. `scripts/check_config_version.sh` confirms
  parity.

  Tests: 224 passed (218 memory + 6 config_version). Lint clean.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/prompt.py
```diff
@@ -19,241 +23,124 @@
...
+    For the text form (single string), use :func:`load_prompt` instead.
+    """
+    base = Path(prompts_dir) if prompts_dir else _PROMPTS_DEFAULT_DIR
+    candidates: list[Path] = [base / f"{name}.chat.yaml"]
+    if agent_name:
+        candidates.insert(0, base / agent_name / f"{name}.chat.yaml")
+    for path in candidates:
+        if path.is_file():
+            try:
+                data = yaml.safe_load(path.read_text(encoding="utf-8"))
```
willem-bd, line 157: **`load_prompt_messages` is uncached - re-reads + reparses YAML on every memory update.**

`load_prompt` caches in `_PROMPT_CACHE` (line 81) and `load_patterns` caches in `_PATTERN_CACHE`, but `load_prompt_messages` has no cache: this `yaml.safe_load(path.read_text(...))` runs on every `_prepare_update_prompt` call (i.e. every debounced memory update via `_do_update_memory_sync`). The PR emphasizes the static system message being prefix-cache friendly, yet the file is re-parsed each time to produce it.

Caching the parsed `{role, content}` template (keyed `(name, agent, dir)`, same shape as the sibling cache) and re-rendering per call would make this consistent and remove the per-update file I/O + YAML parse. The system message has no variables so its rendered form is constant; only the user message needs per-call rendering.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
```diff
@@ -749,15 +755,17 @@ def _prepare_update_prompt(
...
-            consolidation_section=consolidation_section,
-        )
+        variables = {
+            "current_memory": json.dumps(_escape_memory_for_prompt(current_memory), indent=2, ensure_ascii=False),
+            "conversation": conversation_text,
+            "correction_hint": correction_hint,
+            "staleness_review_section": staleness_section,
+            "consolidation_section": consolidation_section,
+        }
+        prompt = load_prompt_messages("memory_update", variables, agent_name=agent_name, prompts_dir=self._prompts_dir)
```
willem-bd, line 771: **Behavior shift worth validating: `memory_update` was restructured from a single string to system+user messages.**

Previously the model was invoked with one `HumanMessage` containing the full prompt (instructions + `<current_memory>` + `<conversation>`); now it gets `SystemMessage` (instructions) + `HumanMessage` (memory + conversation). The system/user split is a deliberate improvement (static system = prefix-cache friendly, mirrors the lead agent) and the three injected text sections are unchanged, but the `memory_update` LLM input is *not* byte-identical to before - system-role framing can change extraction quality and authority handling.

The unit tests only assert structural properties (role split, byte-stability, placeholder substitution), not output fidelity. Worth a real-path check on extraction quality, since the "zero-config behaviour is unchanged" framing reads as same-LLM-input but only the three text sections actually mirror the old constants.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/message_processing.py
```diff
@@ -2,40 +2,87 @@
...
+        logger.warning("Signal patterns file not found (%s); %s detection disabled.", path, name)
+        _PATTERN_CACHE[cache_key] = []
+        return []
+
+    try:
+        with path.open(encoding="utf-8") as f:
+            data = yaml.safe_load(f) or []
+    except yaml.YAMLError as e:
+        raise ValueError(f"Invalid YAML in {path}: {e}") from e
+    except OSError as e:
```
willem-bd, line 59: **For an explicitly-set `patterns_dir`, the "fail loud on config error" design is inconsistently applied - several branches silently degrade instead.**

The function's stated intent is to fail loud for explicit dirs (`path.exists()` False -> `FileNotFoundError` above), but:
- This `except OSError` (e.g. permission denied on a file that *does* exist) returns `[]` and disables detection rather than raising - contradicting the missing-file-raises path right above it.
- Malformed entries are silently skipped (`else: continue` at line 75, the `not isinstance(pattern_text, str)` skip at 77) - an operator typo like `patten:` instead of `pattern:` silently weakens detection with no signal.
- Unknown flag names are silently dropped (the loop at line 79 only honors `ignorecase`; `flags: [multiline]` is ignored without a warning).
- A bad regex at `re.compile(...)` (line 82) raises `re.error` with no file path in the message - hard to locate.

Consider raising (with the file path) for explicit-dir errors and logging a warning for skipped entries/flags, reserving the silent-disable + `logger.warning` for the bundled-default path only (where a missing file is a packaging bug, not a config error).

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/prompts/consolidation.yaml
```diff
@@ -0,0 +1,25 @@
+format: text
```
willem-bd, line 1: **`format` and `version` are not read by either loader - they're documentary only.**

`load_prompt` reads `template`; `load_prompt_messages` reads `messages`. Neither checks `format`, so a file with `format: chat` but a `template` key would be loaded by `load_prompt` without complaint (and vice versa). Either validate these (e.g. reject a mismatched `format`) or drop them to avoid misleading operators who might assume `format` selects the loader. Same applies to the other three prompt YAMLs.

willem-bd (commented):
Review of the memory template/pattern externalization. **No correctness bugs** - the invoke path handles `list[BaseMessage]`, the escaping layer (`format_conversation_for_update` / `_escape_memory_for_prompt`) is preserved, and the YAML regex patterns compile identically to the former `_CORRECTION_PATTERNS`/`_REINFORCEMENT_PATTERNS` constants (verified several incl. the YAML `''` -> `'` single-quote escaping and `` staying literal in single-quoted strings). `.format()` does a single pass so user-controlled `conversation` values are inserted verbatim and not re-parsed - no format-string injection.

I confirmed `prompts_dir` + per-agent `agent_name` override is wired end-to-end (`DeerMem.__init__` -> `MemoryUpdater(prompts_dir=…)` -> `load_prompt_messages(agent_name=…)`), so I'm not re-raising the earlier review's "override path not wired" point.

Findings posted inline are improvement/risk notes. A few minor ones not worth a thread:
- `load_prompt("memory_update")` (text form) is exercised by `test_load_prompt_custom_prompts_dir` but unused in production - only the chat form is called. The test implies an API surface that isn't real.
- `config_version` 26->27 adds two `None`-defaulted fields; confirm `make config-upgrade` merges cleanly (the `patterns_dir: ""` / `prompts_dir: ""` example comments are accurate - empty string is falsy so falls back to bundled, consistent with the `if patterns_dir` / `if prompts_dir` checks).

Positives: good new test coverage (two files covering bundled defaults, caching, custom-dir + agent-override isolation, missing-file errors, the last-6 window, system-message byte-stability); patterns loaded once at construction; no new deps (`pyyaml>=6.0.3` already declared). Reviewed against head `ac0a709`.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/prompt.py
lllyfff: > load_prompt_messages is uncached - re-reads + reparses YAML on every memory update.                                 
                                                                                                                        
  Fixed in the latest push. Added `_CHAT_TEMPLATE_CACHE` (keyed `(name, agent, prompts_dir)` — same shape as            
  `_PROMPT_CACHE`) and a `_render_messages()` helper. The raw `{role, content}` templates are cached after the first    
  file read; on cache hit only `.format(**variables)` runs. Consistent with `load_prompt` and `load_patterns`.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py
lllyfff: > Behavior shift worth validating: memory_update was restructured from a single string to system+user messages.       
                                                                                                                        
  Acknowledged — the system/user split is deliberate, not an oversight.                                                 
                                                                                                                        
  The "zero-config behaviour is unchanged" framing in the PR description was imprecise: what we meant is that operators 
  don't need to change any config after upgrading (the bundled yamls produce the same prompt *content* as before, split 
  across two message roles). You're right that the LLM sees a different message structure — `SystemMessage` carries more
  authority with most models, which is the intended improvement (higher schema-compliance — important for memory        
  extraction which must produce valid JSON), and the static system message is prefix-cache friendly, mirroring the lead 
  agent's design.                                                                                                       
                                                                                                                        
  The three injected text sections (staleness_review, consolidation, fact_extraction) are still rendered as text and    
  remain byte-identical to the pre-externalization constants via the shim loaders. Only `memory_update` changed roles.  
                                                                                                                        
  Real-path extraction quality validation noted — we'll run extraction fidelity checks before merging. Unit tests only  
  verify structural correctness (placeholder substitution, byte-stability of the system message, and that the input list
  type flows through `model.invoke()`), not semantic quality.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/message_processing.py
lllyfff: > For an explicitly-set patterns_dir, the "fail loud on config error" design is inconsistently applied                
                                                                                                                        
  Fixed in the latest push. All four branches now behave differently for explicit vs. bundled defaults:                 
                                                                                                                        
  - **OSError (permission, etc.)**: re-raised when `patterns_dir` is explicit (`raise OSError(...) from e`). Degrades   
  only for bundled defaults (packaging issue, not a config error).                                                      
  - **Malformed entries** (wrong type, missing/empty `pattern` key): skipped with a `WARNING` that includes the entry   
  index and file path — no more silent drop.                                                                            
  - **Unknown flag names**: similarly logged with a `WARNING` (`"Ignoring unknown flag %r in entry %d of %s"`).         
  - **`re.error` at compile**: now raises `ValueError` with the file path, entry index, and offending pattern —         
  traceable to the exact line of the YAML file.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/prompts/consolidation.yaml
lllyfff: > `format` and `version` are not read by either loader — they're documentary only.                                    
                                                                                                                        
  Fixed in the latest push. Both loaders now validate `format`:                                                         
                                                                                                                        
  - `load_prompt` rejects `format: chat` (raises `ValueError` directing the operator to `load_prompt_messages`).        
  - `load_prompt_messages` rejects `format: text` (raises `ValueError` directing the operator to `load_prompt`).        
  - Absent `format` defaults to the expected value for each loader (`text` / `chat`).                                   
                                                                                                                        
  The `version` field is left documentary — there isn't a schema-version guard yet, but the YAML shape checks (missing  
  `template` / missing `messages` → `ValueError`) plus the `format` gate catch the realistic misuse cases.

fancyboi999 (changes requested):
Thanks @lllyfff. I found two issues that should be addressed before this is ready.


Reviewed base SHA: `d2f8f61e3aaa7be4b0fccb6ff9a4bff94d56da39`
Reviewed head SHA: `c900a4651a2ae518caaa8a35c365c1626ac189c6`

[P1] Preserve the custom-prompt contract through the real updater

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:444-472`, `:504-531`, `:742-769`, and `:877-919`
- Problem: the previous prompt-wiring finding is only partially fixed. The main `memory_update.chat.yaml` load receives the active `agent_name`, but `_build_staleness_section()` and `_build_consolidation_section()` receive only `prompts_dir`, so they always select the global template and ignore `{prompts_dir}/{agent_name}/...`. In addition, an invalid explicit prompt is accepted at `DeerMem` construction and its `ValueError`/`FileNotFoundError` is caught by the updater's generic exception handler, which drops that background memory update as `False`; the operator-facing fail-loud configuration boundary is therefore not preserved by the production path.
- Evidence: on this head, a direct `load_prompt("staleness_review", agent_name="researcher", ...)` selected an `AGENT-STALE` fixture, while `_prepare_update_prompt(..., agent_name="researcher")` rendered the global `GLOBAL-STALE` fixture. A second real-updater probe constructed `DeerMem` successfully with a chat template containing an unknown placeholder; `update_memory()` returned `False` without invoking the model. The new tests exercise the loaders directly and concatenate prepared messages, but do not construct the configured backend and assert the messages actually passed to `model.invoke()` for stale/consolidation branches.
- Suggested fix: thread the active `agent_name` through every section builder and `load_prompt()` call. Validate an explicit prompt package at backend construction (or propagate a typed prompt-configuration error outside the recoverable LLM/update catch) so missing, malformed, or unrenderable operator templates cannot turn into repeated dropped updates. Keep the fallback order consistent for the main and section templates.
- Test: construct `DeerMem` with global and per-agent marker templates plus a recording model, seed stale and consolidation candidates, and assert the actual `model.invoke()` messages use the active agent's templates. Cover missing files, malformed YAML, and invalid placeholders through this same runtime path and assert the configuration error is surfaced rather than converted to `False`.

[P2] Advance the config version after merging the other v27 schema change

- Location: `config.example.yaml:18` and `config.example.yaml:1561-1571`
- Problem: current `origin/main` is already `config_version: 27`, and this merge head also remains at 27 while adding `patterns_dir` and `prompts_dir`. An installation already upgraded to main's v27 schema is therefore considered current: `AppConfig._check_config_version()` emits no warning, and `scripts/config-upgrade.sh` exits at `user_version >= example_version`. The earlier 26 -> 27 bump was consumed by a different schema change when main was merged, so it no longer versions this PR's fields. The paths are also only commented examples, so the recursive upgrader would not add them even after a bump.
- Evidence: `git show origin/main:config.example.yaml` and the current head both report 27; an actual `_check_config_version({"config_version": 27}, ...)` probe emitted no outdated warning. `scripts/check_config_version.sh` passes because it checks Helm parity, not whether this PR advanced the schema relative to its base.
- Suggested fix: rebase/merge current main, advance the example and Helm mirrors to the next version, and decide whether the two defaults should be real nullable keys that `config-upgrade` can merge or a documented manual-only configuration. Update the backend/user configuration docs required by the repository policy with the effective path, per-agent fallback, restart/cache behavior, and failure semantics.
- Test: start from a v27 config without these fields and prove the current example reports it outdated and the documented upgrade path exposes the new configuration without changing existing values; keep the Helm parity check.

lllyfff:
@fancyboi999 Thanks for the thorough reviews — all findings addressed across the latest pushes:            
                                                                                                                        
  **P1 — prompts_dir + agent_name wired end-to-end:** Added `prompts_dir` to `DeerMemConfig`, threaded through `DeerMem`
  → `MemoryUpdater.__init__()` → `_prepare_update_prompt()`, and all three templates (`_build_staleness_section`,       
  `_build_consolidation_section`, `load_prompt_messages`) now receive both `prompts_dir=` and `agent_name=` so per-agent
  prompt overrides apply uniformly. Also added `_PROMPT_CACHE` to `load_prompt()` so repeated lookups are a dict hit,   
  not a yaml re-read.                                                                                                   
                                                                                                                        
  **P1 — Explicit prompts validated at construction:** When `DeerMemConfig.prompts_dir` is explicitly set,              
  `DeerMem.__init__` pre-loads all four prompt templates (staleness_review, consolidation, fact_extraction,             
  memory_update with dummy variables) at construction time. A missing file, malformed YAML, or invalid placeholder      
  raises immediately at startup — before the first memory update — instead of being caught by the updater's generic     
  `except Exception` and silently dropped as `False`.                                                                   
                                                                                                                        
  **P2 — patterns_dir fail-loud for explicit directories:** When `patterns_dir` is explicitly set: missing files raise  
  `FileNotFoundError`; unreadable files (OSError) re-raise; malformed YAML / non-list top-level / invalid regex raise   
  `ValueError` with the file path. Malformed entries, missing `pattern` keys, and unknown flag names are logged with a  
  `WARNING` (including entry index) instead of silently skipped. Bundled defaults (`patterns_dir=None`) retain the prior
  degrade behavior (warn + return `[]` — a packaging bug, not a config error).                                          
                                                                                                                        
  **Cache for `load_prompt_messages`:** Added `_CHAT_TEMPLATE_CACHE` (keyed `(name, agent, prompts_dir)`) and a         
  `_render_messages()` helper. The raw `{role, content}` templates are cached after the first file read; on cache hit   
  only `.format(**variables)` runs. Consistent with `load_prompt`'s `_PROMPT_CACHE` and `load_patterns`'s               
  `_PATTERN_CACHE`.                                                                                                     
                                                                                                                        
  **`format` field validated:** `load_prompt` rejects `format: chat` (→ use `load_prompt_messages`);                    
  `load_prompt_messages` rejects `format: text` (→ use `load_prompt`). Absent `format` defaults to the expected value   
  for each loader. Prevents operators from loading a chat yaml via the text loader (or vice versa) without a clear      
  error.                                                                                                                
                                                                                                                        
  **Config version:** Bumped `config_version` 26 → 27 → 28 (upstream main consumed the intermediate bump with a         
  different schema change). All three mirrors synced: `config.example.yaml`, `deploy/helm/deer-flow/values.yaml`,       
  `deploy/helm/deer-flow/README.md`. `scripts/check_config_version.sh` confirms parity.                                 
                                                                                                                        
  **Tests:** 170 passed (memory suite), lint clean on all changed files.

fancyboi999 (changes requested):
Thanks @lllyfff. The agent-specific routing and config-version fixes are present, but the explicit prompt package can still fail inside the recoverable update path and silently drop memory updates.


Reviewed base SHA: `d075be0277ea759623b3924a8032c2983f4d470a`
Reviewed head SHA: `d08173351fffca4f46e0270f1e7523931ca84f4b`

[P1] Keep prompt-configuration failures out of the recoverable update catch

- Location: `backend/packages/harness/deerflow/agents/memory/backends/deermem/deer_mem.py:70-83`; `backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/updater.py:444-473`, `:505-533`, and `:880-922`
- Problem: construction now opens the three text templates, but it never renders them, so an unknown placeholder is accepted until staleness or consolidation actually triggers. Construction also validates only the global files; a malformed `{prompts_dir}/{agent_name}/...` override is first loaded during that agent's update. Both errors are raised from `_prepare_update_prompt()` inside `_do_update_memory_sync_impl()` and are still caught by the generic `except Exception`, which returns `False`. An explicitly configured prompt package can therefore keep the Gateway healthy while silently discarding every affected memory update, contrary to #4285's fail-loud boundary and the previous review.
- Evidence: against this head in an isolated worktree, a complete prompt directory whose `staleness_review.yaml` added `{missing_placeholder}` constructed `DeerMem` successfully; a real stale-fact `update_memory()` then returned `False` without invoking the model. A second oracle put the bad placeholder only in `researcher/memory_update.chat.yaml`; construction again succeeded and `update_memory(..., agent_name="researcher")` returned `False` without a model call. The PR's focused suite still passed (`130 passed`) because it tests loader calls and rendered message structure, not these configured backend paths.
- Suggested fix: make explicit prompt-package validation render every template with its actual supported variable set, and keep configuration errors distinct from recoverable LLM/update failures. Agent-specific overrides cannot all be known at singleton construction, so validate them when the agent path is resolved and propagate a typed configuration error outside the broad update catch (or pre-validate every declared override directory). The invariant should be that a bad explicit template cannot be converted into a normal `False` update result.
- Test: construct `DeerMem` with an invalid placeholder in each text template and assert construction fails; then cover a malformed per-agent chat/text override through the real `update_memory(..., agent_name=...)` path and assert the configuration error escapes while `model.invoke()` is never called.

lllyfff:
Thanks @fancyboi999 — all three gaps addressed in the latest push:                                                    
                                                                                                                        
  **Text templates rendered at construction:** `DeerMem.__init__` now renders staleness_review, consolidation, and      
  fact_extraction with dummy variables (`.format(stale_facts="")`, `.format(consolidation_groups="" , max_groups=1)`,   
  `.format(message="")`) so an unknown placeholder in any of them raises immediately at startup — not during the first  
  triggered staleness cycle.                                                                                            
                                                                                                                        
  **Typed configuration error:** Defined `PromptConfigurationError(ValueError)` in `prompt.py`. All `load_prompt` /     
  `load_prompt_messages` / `_render_messages` errors (bad YAML, missing keys, invalid placeholders, format mismatches)  
  now raise `PromptConfigurationError` instead of bare `ValueError`.                                                    
                                                                                                                        
  **Configuration errors escape the recoverable catch:** Three catch sites in the update path now re-raise              
  `(PromptConfigurationError, FileNotFoundError, OSError)` before the generic `except Exception`:                       
  - `_do_update_memory_sync_impl` (the sync LLM call wrapper)                                                           
  - `update_memory`'s thread-pool executor path (`future.result()`)                                                     
  - `_process_queue`'s per-context catch in `MemoryUpdateQueue`                                                         
                                                                                                                        
  A bad explicit prompt (global *or* per-agent override) now propagates up through the worker thread rather than being  
  silently converted to `False`.                                                                                        
                                                                                                                        
  **Per-agent overrides cannot all be known at singleton construction** — acknowledged. The typed-error propagation     
  handles the runtime case: when a per-agent override is first resolved and fails, the `PromptConfigurationError`       
  escapes the queue's catch and surfaces to the worker thread's caller instead of being logged as a generic `Memory     
  update failed`.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/queue.py
```diff
@@ -228,6 +229,8 @@ def _process_queue(self, *, skip_inter_item_delay: bool = False) -> None:
                     else:
                         failed += 1
                         logger.warning("Memory update skipped/failed for thread %s (trace_id=%s)", context.thread_id, context.trace_id)
+                except (PromptConfigurationError, FileNotFoundError, OSError):
+                    raise
```
willem-bd: **Re-raising from the background queue worker aborts the batch and surfaces only via stderr.**

`_process_queue` is the direct target of a daemon `threading.Timer` (`_schedule_timer`, ~line 174) with no outer try/except wrapping it. So when one context hits this `raise`, the exception propagates out of the per-item loop, the `finally` resets `_processing`, and escapes into the Timer thread - hitting Python's default `sys.excepthook` (stderr traceback only, no structured `logger` call) and **dropping every remaining context in the batch** (they were already cleared from `self._queue` at ~line 206).

Reachable at runtime: the `__init__` validation in `deer_mem.py` runs without `agent_name`, so a malformed agent-specific override (`{prompts_dir}/{agent}/*.yaml`) isn't caught at startup and will hit this path on the first update for that agent - aborting its co-tenants' pending updates too, and repeating every subsequent batch.

Also, grouping `FileNotFoundError`/`OSError` with `PromptConfigurationError` treats potentially-transient FS errors (brief mount/permission hiccup) as permanent-and-fatal. Contrast `flush_sync`'s shutdown path (~line 333), which wraps the same call in `except Exception: logger.exception(...)` - so the two paths behave inconsistently for the same error.

Consider catching `PromptConfigurationError` per-item (log at ERROR, count as failed, `continue`) so one bad agent doesn't abort its co-tenants, and keeping transient `OSError`/`FileNotFoundError` in the swallow-and-retry bucket rather than the re-raise bucket.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deer_mem.py
```diff
@@ -56,11 +57,32 @@ def __init__(self, backend_config: dict[str, Any] | None = None) -> None:
...
+                "current_memory": "{}",
+                "conversation": "(validation)",
+                "correction_hint": "",
+                "staleness_review_section": "",
+                "consolidation_section": "",
+            }
+            load_prompt("staleness_review", prompts_dir=self._config.prompts_dir).format(stale_facts="")
+            load_prompt("consolidation", prompts_dir=self._config.prompts_dir).format(consolidation_groups="", max_groups=1)
+            load_prompt("fact_extraction", prompts_dir=self._config.prompts_dir).format(message="")
+            load_prompt_messages("memory_update", _dummy_vars, prompts_dir=self._config.prompts_dir)
```
willem-bd, line 86: **Agent-specific prompt overrides escape startup validation.**

This validation block runs every `load_prompt` / `load_prompt_messages` call without `agent_name`, so it only validates the default templates (`{prompts_dir}/{name}.yaml`). At runtime, `_build_staleness_section` / `_build_consolidation_section` / `load_prompt_messages` are all called with `agent_name=agent_name`, so a malformed `{prompts_dir}/{agent}/staleness_review.yaml` (e.g. `{staleFact}` instead of `{stale_facts}`) isn't caught here - it surfaces as a `PromptConfigurationError` mid-batch (feeding the queue re-raise flagged separately on `queue.py`).

The `__init__` comment ("surfaces at startup rather than as a silent dropped update") holds for the global `prompts_dir` defaults but overclaims for the agent-override case. Either validate lazily-once-per-agent (cached), or soften the comment to scope the claim to the default templates.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deer_mem.py
```diff
@@ -56,11 +57,32 @@ def __init__(self, backend_config: dict[str, Any] | None = None) -> None:
...
+            _dummy_vars = {
+                "current_memory": "{}",
+                "conversation": "(validation)",
+                "correction_hint": "",
+                "staleness_review_section": "",
+                "consolidation_section": "",
+            }
+            load_prompt("staleness_review", prompts_dir=self._config.prompts_dir).format(stale_facts="")
+            load_prompt("consolidation", prompts_dir=self._config.prompts_dir).format(consolidation_groups="", max_groups=1)
+            load_prompt("fact_extraction", prompts_dir=self._config.prompts_dir).format(message="")
```
willem-bd: **`fact_extraction` is validated and shipped but never rendered at runtime.**

`FACT_EXTRACTION_PROMPT` is defined only as a shim in `prompt.py` and is never imported or called anywhere else in the repo (pre-existing dead code - the constant was unused before this PR too). So this validation line, the bundled `fact_extraction.yaml`, and the shim alias all exercise a path no runtime caller hits.

Not a regression, but the PR adds a bundled file and a validation line for a dead prompt. Consider dropping the shim + file + validation, or wiring `load_prompt("fact_extraction")` into an actual extraction path.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deermem/core/queue.py
lllyfff: > Re-raising from the background queue worker aborts the batch and surfaces only via stderr.                          
                                                                                                                        
  Agreed — removed all three `(PromptConfigurationError, FileNotFoundError, OSError)` re-raise guards from              
  `_do_update_memory_sync_impl`, `update_memory`'s executor path, and `_process_queue`. The existing `except Exception: 
  logger.exception(...)` already logs prompt-config errors at ERROR level with a full traceback (and returns `False` so 
  the queue continues processing other contexts). One bad per-agent override no longer aborts its co-tenants.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deer_mem.py
lllyfff: > Agent-specific prompt overrides escape startup validation.                                                          
                                                                                                                        
  Softened the construction comment to accurately scope the claim: the validation block only covers *global* templates  
  (`{prompts_dir}/{name}.yaml`). Per-agent overrides (`{prompts_dir}/{agent}/*.yaml`) cannot be known at singleton      
  construction time — they are validated lazily at first use and logged at ERROR by the updater's `except Exception:    
  logger.exception(...)` handler, which returns `False` without aborting co-tenant updates.

backend/packages/harness/deerflow/agents/memory/backends/deermem/deer_mem.py
lllyfff: > fact_extraction is validated and shipped but never rendered at runtime.                                             
                                                                                                                        
  Dropped `load_prompt("fact_extraction", ...).format(message="")` from the construction validation block. The          
  `FACT_EXTRACTION_PROMPT` shim and bundled `fact_extraction.yaml` are left in place for backward compatibility, but the
  startup check no longer exercises a path with no runtime caller.

WillemJiang (approved):

PR #4287, opened 2026-07-18
fix(channels): key inbound dedupe on chat-scoped workspaces

AnnaSuSu:
## Why

`ChannelManager._inbound_dedupe_key` fails closed when it cannot find a
workspace/team/guild identifier, so the whole inbound dedupe path is skipped and
a provider redelivery runs the agent again (and repeats its real side effects).

The only writer of `msg.workspace_id` is `attach_connection_identity`, and it
assigns it **only when a persisted connection row is found**. `channel_connections.enabled`
defaults to `False`, so `_make_connection_repo` returns `None`, `attach_connection_identity`
returns the message untouched, and `msg.workspace_id` is never set. Slack, Discord,
WeCom and GitHub survive that because they also stamp `team_id` / `guild_id` /
`aibotid` / `workspace_id` on the message itself. Telegram, Feishu, WeChat and
DingTalk stamp none of them.

Net effect on a default configuration: **Telegram, Feishu, WeChat and DingTalk have
no inbound dedupe at all.** A Feishu/Telegram webhook retry or a WeChat redelivery
runs the agent twice and posts two answers.

The fail-closed itself is right, and the reason given in the comment is specific to
Slack ("Slack channel ids are not globally unique"). For the chat-scoped providers the
repo already declares the opposite: their adapters persist
`connection.workspace_id = chat_id` at bind time, and the GitHub dispatcher states the
model explicitly when it explains its own key —

> The repo is globally unique and always present — mirrors Telegram/WeChat keying the
> workspace on the chat id.
> — `app/gateway/github/dispatcher.py:434-436`

So the workspace identity of those chats is the chat id; the dedupe path just had no way
to see it until a connection existed.

## What changed

`ChannelManager._inbound_dedupe_key` resolves the workspace from two more sources:

- `metadata["conversation_id"]` — DingTalk stamps it on every inbound. This also covers
  **bound** DingTalk P2P, whose `connection.workspace_id` is deliberately `None`
  (`_connection_workspace_id` returns the conversation id for groups only).
- `msg.chat_id`, for `telegram` / `feishu` / `wechat` only
  (`CHAT_SCOPED_WORKSPACE_CHANNELS`), when nothing else is present.

Callers and adapters are unchanged. A message that already carries `workspace_id`
produces the same 4-tuple as before, so a bound and an unbound redelivery of the same
chat share one cache entry.

**Slack keeps failing closed** without a `team_id` — no chat-id fallback is added for it.
That fail-closed was requested in review (`test_inbound_dedupe_key_fails_closed_without_workspace`,
"willem #3") and this change does not relax it; the existing test still drives that path
and still passes.

## Surface area

- [ ] **Frontend UI**
- [ ] **Backend API**
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [x] **Default behavior change** — two of them.
      **(1)** Redeliveries on Telegram/Feishu/WeChat/DingTalk are now absorbed instead of
      re-running the agent. That is the fix, but it does change behavior for anyone who
      was (unknowingly) getting duplicate runs.
      **(2)** A transient failure now releases the inbound dedupe key at the three sites
      that handle their error in place instead of re-raising (the streaming path, and
      both thread-busy branches), so a provider redelivery can still recover instead of
      being dropped for the dedupe TTL. That release is keyed on the **swallow site, not
      the channel**, so it also closes the same pre-existing black hole on **WeCom** —
      which already had a dedupe key on `main` (`supports_streaming=True` plus `aibotid`)
      and is otherwise untouched by this PR. Rationale and the main-vs-branch measurement
      are in the round comment below.

## Bug fix verification

  - `backend/tests/test_channels.py::TestChannelManager::test_inbound_dedupe_key_uses_chat_id_for_chat_scoped_providers_when_unbound`
  - `backend/tests/test_channels.py::TestChannelManager::test_inbound_dedupe_key_uses_dingtalk_conversation_id_when_unbound`
  - `backend/tests/test_channels.py::TestChannelManager::test_inbound_dedupe_chat_scoped_fallback_does_not_collapse_distinct_chats`
  - `backend/tests/test_channels.py::TestChannelManager::test_dispatch_loop_dedupes_unbound_chat_scoped_redelivery`
- Did they go red on `main` and green on this branch? **yes** — verified by restoring
  only `app/channels/manager.py` to `origin/main` (tests untouched) and running each test
  individually. The dispatch-loop test is the end-to-end one: it publishes the same WeChat
  `message_id` twice through a real `ChannelManager` with no connection repo (the default
  configuration) and asserts the agent ran once; on `main` it runs twice.
- Both new clauses are anchored independently: deleting only the `conversation_id`
  fallback reds the DingTalk test alone; deleting only the chat-scoped fallback reds the
  other three. Neither is an unguarded addition.
- Reverse direction (`newly_missed`): `test_inbound_dedupe_chat_scoped_fallback_does_not_collapse_distinct_chats`
  asserts the same `message_id` in two different chats still yields two distinct keys, so
  the new fallback cannot cross-dedupe two conversations.

## Validation

```
cd backend
PYTHONPATH=. uv run pytest tests/test_channels.py -q     # 257 passed
PYTHONPATH=. uv run pytest tests/ -q                     # 8 failed, 7828 passed, 28 skipped
uvx ruff check . && uvx ruff format --check .            # clean
```

The 8 failures are pre-existing on `origin/main` (the `web_fetch` suites refuse
private-network addresses in this environment); the failure set is byte-for-byte identical
before and after this change, so this branch adds none.

## AI assistance

**Tool(s) used:** Grok (xAI, headless agent) and Claude Code.

**How you used it:** Grok ran the investigation and produced the change in an isolated
worktree — adapter enumeration, the failing tests, and the fix. Claude Code was then used
to review it: I re-ran the red check myself with the source rolled back to `origin/main`,
re-ran the per-clause deletion checks, re-ran the full suite against a clean baseline, and
re-verified the dispatcher quotation and the default value of `channel_connections.enabled`
rather than taking them from the tool's report. Neither tool committed or pushed.


backend/app/channels/manager.py
```diff
@@ -1170,7 +1175,13 @@ def _inbound_dedupe_key(msg: InboundMessage) -> tuple[str, str, str, str] | None
         # Fail closed: without a workspace/team/guild identifier we cannot tell two
         # workspaces apart (e.g. Slack channel ids are not globally unique), so
         # skip dedupe rather than risk collapsing distinct workspaces' messages.
-        workspace_id = msg.workspace_id or metadata.get("workspace_id") or metadata.get("team_id") or metadata.get("guild_id") or metadata.get("aibotid")
+        # conversation_id covers DingTalk (group + P2P) which stamps it on every
```
willem-bd: The DingTalk branch of this dedupe relies on `conversation_id` being identical across a provider **redelivery of the same message** - not just present on every inbound. That's almost certainly true (conversation ids are conversation-scoped, not delivery-scoped), but the comment asserts "stamps it on every inbound," which is a weaker guarantee than dedupe actually needs.

If you can cite redelivery-stability from the DingTalk docs, or add a DingTalk redelivery integration test mirroring the WeChat one in `test_dispatch_loop_dedupes_unbound_chat_scoped_redelivery`, the assumption becomes explicit rather than implied. Low priority - the conclusion is right either way.

backend/tests/test_channels.py
```diff
@@ -1058,6 +1058,160 @@ def test_inbound_dedupe_key_fails_closed_without_workspace(self):
...
+
+            async def capture_outbound(msg: OutboundMessage) -> None:
+                outbound_received.append(msg)
+
+            bus.subscribe_outbound(capture_outbound)
+            await manager.start()
+
+            def _wechat_inbound(message_id: str) -> InboundMessage:
+                return InboundMessage(
+                    channel_name="wechat",
```
willem-bd, line 1343: The docstring already documents the WeChat-only tradeoff (telegram/feishu key formation covered by the unit tests above) - fair given WeChat's non-streaming shape matches the `runs.wait` mock.

If you want to harden it cheaply later: parametrizing this dispatch test over `(wechat, telegram, feishu)` would also exercise the streaming-path dedupe end-to-end for telegram/feishu, catching any future regression there. Low priority - the shared `_inbound_dedupe_key` path is well unit-covered.

willem-bd (commented):
Verified every factual claim against head `827aa677` - all hold:

- `channel_connections.enabled` defaults to `False` -> `_connection_repo` is `None` -> `attach_connection_identity` returns the message untouched -> `msg.workspace_id` is never set. So on the default config, Telegram/Feishu/WeChat/DingTalk had no inbound dedupe at all.
- telegram/feishu/wechat persist `connection.workspace_id = chat_id` and call `attach_connection_identity(..., workspace_id=inbound.chat_id)`, so a *bound* redelivery ends up with `msg.workspace_id = chat_id` - matching the new unbound `chat_id` fallback. Bound and unbound redeliveries of the same chat share one cache entry.
- DingTalk stamps `conversation_id` on every inbound (group + P2P); P2P stores `connection.workspace_id = None`, so bound P2P correctly falls through to `conversation_id`.
- Slack stays fail-closed (no `chat_id` fallback); `test_inbound_dedupe_key_fails_closed_without_workspace` still drives that path.

The fix is correct and the tests are well-anchored: per-clause deletion reds only the relevant test, the `newly_missed` guard prevents cross-conversation collapse, and the dispatch-loop test is a genuine end-to-end proof on the default config. Two minor inline notes below - neither blocks.

One non-inline item: this changes default behavior on Telegram/Feishu/WeChat/DingTalk (redeliveries are now absorbed instead of double-running the agent). Worth a CHANGELOG entry - I didn't see one in the diff.

Assessment: approve.

backend/app/channels/manager.py
AnnaSuSu: You're right that "stamps it on every inbound" is the wrong guarantee to assert, and I'd rather not settle it with a docs citation — partly because I can't verify DingTalk's redelivery contract from here, and partly because **the integration test you suggest can't settle it either**: it would prove the manager dedupes *given* identical `conversation_id`s, not that DingTalk *sends* identical ones on redelivery. Both would be me asserting the same unverified thing in a new place.

What is verifiable is the failure direction, so I rewrote the comment around that instead. Both new fallbacks are appended **last** in the `or` chain, and the `chat_id` one is additionally gated on `not workspace_id`. They can only turn *no key* into *a key* — no message that already formed a key forms a different one. So if `conversation_id` were not reused across a redelivery, the result degrades to today's behaviour on `main` (no dedupe); it can never collapse two conversations, because `chat_id` and `message_id` stay in the tuple (locked by `test_inbound_dedupe_chat_scoped_fallback_does_not_collapse_distinct_chats`).

While checking this I enumerated the whole chain rather than just this clause, since the invariant you're applying is the repo's own — `manager.py:77-80`, "not guaranteed identical across a provider's own redelivery ... would miss exactly the retries we want to absorb". Result: the repo has **no** affirmative redelivery-stability statement for `team_id`, `guild_id`, or `aibotid` either. The only one anywhere is `X-GitHub-Delivery` ("reused verbatim when a delivery is retried", `gateway/github/dispatcher.py:401-422`). So the property you're asking about is one the three sources already on `main` share; I've left those alone rather than widen this PR, but flagging it so it's on the record rather than looking like an oversight.

backend/tests/test_channels.py
AnnaSuSu: Parametrized in `3218b56` over `(wechat, telegram, feishu)` — it asserts against `runs.stream` or `runs.wait` depending on `_channel_supports_streaming`.

One correction on the rationale, though: it doesn't exercise "streaming-path dedupe", because there isn't one. `_is_duplicate_inbound` is called in the consume loop at `manager.py:1142`, **before** `_handle_message` and before the streaming branch, so all channels share a single dedupe gate. The parametrization is still worth having as a regression guard on key formation, but it wasn't covering a second dedupe path.

Where the streaming/non-streaming split *does* matter is the **release** side — and chasing your suggestion is what surfaced it. Details in the top-level comment; short version is that `_release_inbound_dedupe_key` sits in `_handle_message`'s generic `except Exception`, and `_handle_streaming_chat` never re-raises, so on `main` a transient streaming failure black-holes the message id for the full dedupe TTL. This PR gave telegram/feishu a key for the first time, which newly exposed them to it.

AnnaSuSu:
Thanks for the verification pass — all three items addressed in `3218b56`. Replies to the two inline notes are in-thread; this covers the CHANGELOG and one thing your second note led me to that neither of us had flagged.

## CHANGELOG

Added, as a clause on the existing consolidated `**channels:**` bullet in `[Unreleased] → Fixed` (next to the "dedupe GitHub webhook redeliveries" clause) plus the `[#4287]` link ref. English only — `CHANGELOG_zh.md` has had no `[Unreleased]` section since its bulk creation, so it isn't a live mirror.

Worth saying: your periodic sweep (the #4270 pattern) would have picked this up anyway, so please drop my line if it collides — I'd rather it be recorded once than twice.

## The thing your second note surfaced

Chasing "parametrize over the streaming providers" turned up a real defect — not in key formation, but in key *release*.

`_release_inbound_dedupe_key` lives in `_handle_message`'s generic `except Exception` handler. Its docstring states the invariant: a transient failure must not "turn a recoverable error into a TTL-long black hole for the same message_id". But three sites handle their error in place and never re-raise, so that handler never runs:

| site | swallows |
|---|---|
| `_handle_streaming_chat` `except Exception` → `finally` | every stream error (generic + thread-busy) |
| fire-and-forget `runs.create` busy branch | `ConflictError` |
| non-streaming `runs.wait` busy branch | `ConflictError` |

Measured — transient failure, then the provider redelivers the same `message_id`:

| | `main` | this branch, before the fix |
|---|---|---|
| feishu (streaming) | key `None` → redelivery **reprocessed** (stream 1→2) | key formed → redelivery **absorbed** (1→1), full 10-min TTL |
| wechat (non-streaming) | reprocessed (1→2) | reprocessed (1→2) — release fires |

**The mechanism is pre-existing, not introduced here** — and I'd rather show that than assert it: on `main`, WeCom already has a key (`supports_streaming=True` *and* stamps `aibotid`) and exhibits the same 1→1 black hole. What this PR changed is the *reach*: giving telegram/feishu/wechat/dingtalk a key for the first time newly exposes them to it. Since my change is what exposed them, fixing it here rather than deferring.

Severity, stated accurately: the user still gets the error message on the first attempt. What breaks is *recovery* — the redelivery that would have retried is dropped for the TTL. Not silent message loss.

Fix is one release call per site. Note the invariant already had one anchor on `main` (`test_dispatch_loop_releases_dedupe_key_when_handling_fails`, the generic path); the three new tests extend it to the remaining sites, so it's 4 sites / 4 tests rather than a patch on the one you'd have hit.

## Verification

Per-clause red check, each site deleted individually with the others intact (isolation confirmed by symbol count + `ast.parse`, not by git state):

| deleted site | `…streaming_transient_failure…` | `…thread_busy…` | `…fire_and_forget_thread_busy…` |
|---|---|---|---|
| streaming `finally` | **failed** | passed | passed |
| fire-and-forget busy | passed | passed | **failed** |
| `runs.wait` busy | passed | **failed** | passed |

Clean diagonal — no test is green for another clause's reason.

One correction on my own method: the first attempt at the streaming row was invalid. Deleting only the call left an empty `if` body → `IndentationError` → all three tests failed, which *looked* like a stronger red than it was. Redone deleting the whole block, with `ast.parse` confirming the module still imports. Recording it because the bad result was the more flattering one.

At `3218b565`:

```
PYTHONPATH=. uv run pytest tests/test_channels.py -q   # 262 passed
uvx ruff check . && uvx ruff format --check .          # clean
PYTHONPATH=. uv run pytest tests/ -q                   # 8 failed, 7833 passed, 28 skipped
```

Clean-baseline comparison: reverted all three changed files to `origin/main`, verified the isolation, ran the full suite, restored and re-verified.

```
main:   8 failed, 7824 passed
branch: 8 failed, 7833 passed
sorted FAILED lines: identical
```

The 8 are the pre-existing `web_fetch` suites refusing private-network addresses in this environment. Pass delta +9 = 4 tests from the original round (vs bare `main`) + 5 this round.

## Boundary update

- **Covered:** inbound dedupe key formation for chat-scoped workspaces; release-on-transient-failure across all 4 swallow sites.
- **Known non-goal:** affirmative per-provider redelivery-stability proof for `team_id` / `guild_id` / `aibotid` / `conversation_id`. The repo asserts this only for `X-GitHub-Delivery`; the fallback chain is monotone so an unstable source degrades to no-dedupe rather than mis-dedupe. Happy to open a follow-up if you want those documented properly.

AnnaSuSu:
Correcting an under-disclosure in my previous comment — no code change, but the PR body now says it explicitly.

I used WeCom as *evidence* that the release gap predates this PR, and left the consequence implicit. Stating it plainly: **this PR also changes WeCom's behavior.** The release is keyed on the swallow site, not the channel, so the same three call sites that fix telegram/feishu/wechat/dingtalk also close WeCom's pre-existing black hole — and WeCom is otherwise not part of this change. A behavior change in a channel the PR description didn't name is exactly the kind of thing that should be in the description, so I've added it to the "Default behavior change" box rather than leaving it inferable from "one release call per site".

On why it isn't scoped to only the newly-exposed channels: narrowing it would mean a channel check inside a generic error handler, which encodes "these four channels have dedupe keys" in a place that has no business knowing that — and it would go stale the moment another provider starts stamping a workspace id. The invariant the release upholds ("a transient failure must not black-hole the message id for the TTL") is channel-agnostic, so the guard is too.

Happy to split the release fix into its own PR if you'd rather keep this one to key formation. My reasoning for keeping it here is that my change is what made the gap reachable for four channels, so shipping the exposure without the fix seemed worse than a slightly wider diff — but that's a judgement call and yours overrides it.

backend/app/channels/manager.py
```diff
@@ -1803,6 +1823,12 @@ async def _handle_streaming_chat(
                 len(artifacts),
                 stream_error,
             )
+            if stream_error is not None:
+                # This path swallows its own errors, so _handle_message's generic
+                # handler never runs and never releases the key. Mirror it here:
+                # the key was recorded on receipt, and without this a transient
+                # stream failure black-holes the message_id for the dedupe TTL.
+                self._release_inbound_dedupe_key(msg)
```
willem-bd: **Suggestion: move this release to after the `publish_outbound` call immediately below.**

Releasing the key here — before the `is_final=True` error reply is published — opens a small window where a provider redelivery of the same `message_id` can start a second run before this run's error reply lands, so the user could see partial-stream → run-2 output → run-1 error reply (out of order).

Not a correctness bug — the retry is the point of the release. But relocating this `if stream_error is not None:` block to just after `await self.bus.publish_outbound(...)` would guarantee the error reply is visible before a retry can begin. The window is tiny and provider redelivery isn't instantaneous, so this is polish, not a blocker.

AnnaSuSu: Addressed in 713ece5d. _handle_streaming_chat now awaits the final publish_outbound call before releasing the inbound dedupe key, so a redelivery cannot overtake the terminal reply. I also strengthened test_streaming_transient_failure_releases_dedupe_key to assert that the key is still present while the final outbound callback runs, while retaining the post-failure redelivery check. Validation: 6 targeted tests passed; Ruff check and format passed. Thanks!

WillemJiang (approved):

PR #2789, opened 2026-05-08
Fix duplicate gateway upload filenames

ZenCGL:
### Motivation
- Multipart uploads to the Gateway could silently overwrite earlier parts when multiple form fields used the same basename, losing data for the overwritten file.

### Description
- Use the shared `claim_unique_filename` helper to auto-rename duplicate filenames within a single `POST /api/threads/{thread_id}/uploads` request by adding `_N` suffixes and track them with a `seen_filenames` set so later parts do not truncate earlier ones in the same request (`app/gateway/routers/uploads.py`).
- Include `original_filename` in the upload response when a part was renamed so callers/agents can see the original basename (`app/gateway/routers/uploads.py`).
- Add a regression test `test_upload_files_auto_renames_duplicate_form_filenames` proving two form parts named `data.txt` are persisted as `data.txt` and `data_1.txt` with preserved contents (`tests/test_uploads_router.py`).
- Update backend docs to mention the new auto-rename behavior in uploads (`backend/README.md`, `backend/CLAUDE.md`).

### Testing
- Ran the uploads router unit tests: `cd backend && PYTHONPATH=. uv run pytest tests/test_uploads_router.py -q`, all tests passed (29 passed). 
- Ran the linter/formatter check: `cd backend && uv run ruff check app/gateway/routers/uploads.py tests/test_uploads_router.py` and fixed import ordering; final check passed.
- 
Fixes #2795 -

PR #1727, opened 2026-04-01
feat(uploads): add pymupdf4llm PDF converter with auto-fallback and async offload

syzhang622:
## Background

DeerFlow is a deep research agent — users routinely upload lengthy PDFs (financial reports, academic papers, regulatory filings, annual reports) and expect the agent to understand their structure. The current PDF conversion relies solely on MarkItDown, which produces flat plain text with zero heading recognition. Across every PDF we tested, MarkItDown identified 0 headings, meaning the agent has no sense of document structure, chapters, or sections.

This directly contributes to:
- **#1569** — large PDF uploads timing out (504) because the synchronous conversion blocks the async event loop for tens of seconds
- **#1647** — agent ignoring uploaded files, partly because the converted text lacks structure for the agent to navigate

## Root Cause

MarkItDown does not perform layout analysis on PDFs. It extracts raw text without identifying headings, table boundaries, or reading order. For non-PDF formats (DOCX, XLSX, PPTX, HTML) it works well, but PDF is a fundamentally harder format that requires dedicated parsing.

## Solution

Add `pymupdf4llm` as an optional, higher-quality PDF converter alongside the existing MarkItDown, with a three-mode configuration and automatic fallback.

- Only `.pdf` files are affected — all other formats continue to use MarkItDown unchanged
- `pymupdf4llm` is an optional dependency — if not installed, behavior is identical to before
- Large-file conversion is offloaded via `asyncio.to_thread()` to prevent event loop blocking

## Library Selection

We evaluated three tiers of PDF extraction libraries:

**Tier 1 — Legacy pure-Python parsers** (PyPDF2/pypdf, pdfminer.six, pdfplumber, Tabula, Camelot): All produce plain text without Markdown or heading structure. Being pure Python, they hold the GIL throughout execution, making `asyncio.to_thread()` ineffective for concurrency. Eliminated.

**Tier 2 — Modern ML-powered parsers** (Docling, Marker, MinerU, Unstructured): Higher quality, but impractical for DeerFlow's use case. Docling processes ~4 s/page, Marker ~11 s per test document — unacceptable for a real-time interactive agent. All require or strongly recommend GPU and carry heavy dependencies. Eliminated.

**Tier 3 — pymupdf4llm**: Built on MuPDF (C engine), it hits the right trade-off:

| Dimension | pymupdf4llm |
|---|---|
| Engine | MuPDF (C library) — releases GIL during I/O and parsing |
[...]

PR #4288, opened 2026-07-18
fix(uploads): claim the converted markdown filename before writing it

AnnaSuSu:
## Why

Uploads already guarantee that files in one request cannot silently destroy each other.
#2789 introduced that guarantee, and its comment still states it:

> Track filenames within this request so duplicate form parts do not silently truncate
> each other. Existing uploads keep the historical overwrite behavior for a single
> replacement upload.
> — `app/gateway/routers/uploads.py`

Every uploaded form part goes through `claim_unique_filename(...)`, which renames on
collision and records the name in `seen_filenames`.

The companion Markdown produced by auto-conversion does not. `convert_file_to_markdown`
writes to `file_path.with_suffix(".md")` unconditionally, and that name is neither claimed
nor recorded. Auto-conversion (#1727) predates the guarantee (#2789), so this path was
simply never brought under it.

Within a single upload request that means real, silent data loss:

- `notes.md` + `notes.docx` → converting the `.docx` overwrites the user's own `notes.md`.
- `a.docx` + `a.pdf` → both convert to `a.md`; the second silently replaces the first.
- `a.docx` + `a.md` (in that order) → the uploaded `a.md` overwrites the companion the
  response is still pointing at.

In each case the API response happily reports `markdown_file: "a.md"` for a document
whose conversion output is no longer there, and `written_paths` can contain the same
path twice.

Severity, stated honestly: `uploads.auto_convert_documents` is **off by default**
(#2332 made host-side conversion explicit opt-in), so only deployments that turned it on
are affected. The consequence for those deployments is lost file content with no error,
which is why this is worth fixing rather than documenting.

## What changed

Both call sites now claim the companion name **before** conversion writes it, using the
same `claim_unique_filename` / `seen` bookkeeping as the uploaded parts:

- `app/gateway/routers/uploads.py` — the gateway upload endpoint.
- `packages/harness/deerflow/client.py::DeerFlowClient.upload_files` — the SDK path, which
  has the identical concept (`claim_unique_filename` over `seen_names`) and the identical gap.

To let a caller say where the Markdown goes, `convert_file_to_markdown` takes an optional
`output_path`. When it is omitted the behavior is exactly as before
(`file_path.with_suffix(".md")`), so no other caller changes.

I kept the parameter rather than converting first and renaming afterwards: renaming after
the fact still lets the write land on the victim's path, so the clobber would happen and
then be papered over. Reserving the name up front is what makes the guarantee hold.

Out of scope, deliberately: cross-request overwrites. The comment above declares those
intentional ("Existing uploads keep the historical overwrite behavior"), and this change
does not touch them.

## Surface area

- [ ] **Frontend UI**
- [x] **Backend API** — `POST /threads/{id}/uploads` may now report `markdown_file: "a_1.md"`
      where it previously reported `a.md` and silently lost one of the two files.
- [ ] **Agents / LangGraph**
- [ ] **Sandbox**
- [ ] **Skills**
- [ ] **Dependencies**
- [ ] **Default behavior change** — `uploads.auto_convert_documents` is off by default;
      deployments that have not enabled it are unaffected.

## Bug fix verification

  - `backend/tests/test_uploads_router.py::test_upload_files_converted_markdown_does_not_overwrite_user_markdown`
  - `backend/tests/test_uploads_router.py::test_upload_files_two_convertibles_get_distinct_markdown_companions`
  - `backend/tests/test_uploads_router.py::test_upload_files_user_markdown_after_convertible_is_renamed_not_overwritten`
  - `backend/tests/test_client.py::TestUploads::test_upload_files_converted_markdown_uses_unique_names_on_stem_collision`
- Did they go red on `main` and green on this branch? **yes** — verified by restoring only
  the three source files to `origin/main` (tests untouched) and running each test
  individually. They drive the real upload path with real files on disk and assert the
  victim's content survives, not just that the names differ.
- The two call sites are anchored separately: restoring only `routers/uploads.py` reds the
  three router tests and leaves the client test green; restoring only `client.py` reds the
  client test and leaves the router tests green. Neither site is covered only by the other.

## Validation

```
cd backend
PYTHONPATH=. uv run pytest tests/test_uploads_router.py tests/test_client.py -q  # 190 passed
PYTHONPATH=. uv run pytest tests/ -q                    # 8 failed, 7828 passed, 28 skipped
uvx ruff check . && uvx ruff format --check .           # clean
```

The 8 failures are pre-existing on `origin/main` (the `web_fetch` suites refuse
private-network addresses in this environment); the failure set is byte-for-byte identical
before and after this change, so this branch adds none.

## AI assistance

**Tool(s) used:** Grok (xAI, headless agent) and Claude Code.

**How you used it:** Grok ran the investigation and produced the change in an isolated
worktree — the sibling-site enumeration that found the SDK path, the failing tests, and the
fix. Claude Code was then used to review it: I re-ran the red check myself with the sources
rolled back to `origin/main`, re-ran the per-call-site isolation checks, re-ran the full
suite against a clean baseline, and checked the history (#1727 / #2332 / #2789) rather than
taking the tool's framing. Neither tool committed or pushed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -241,6 +241,13 @@ This section accumulates work toward the **2.1.0** milestone
   blocking filesystem IO in artifact serving, gateway uploads, and the Discord
   channel; limit the uploaded-file context manifest; and live-tail malformed
   Redis reconnect ids. ([#3651], [#3551], [#3935], [#3927], [#3917], [#4012])
+- **uploads:** Claim the converted-Markdown companion filename before writing
+  it, so two convertible uploads sharing a stem (or a convertible plus a
+  same-stem `.md` upload) no longer silently clobber each other within one
+  request. When `uploads.auto_convert_documents` is on, the companion `.md` now
+  gets a unique name (e.g. `a_1.md`); `POST /threads/{id}/uploads` and
+  `DeerFlowClient.upload_files` both report the actual name in `markdown_file`.
+  ([#4288])
 - **config:** Coerce null object config sections to their defaults; honor the
   unified database configuration in the store and sync checkpointer; and have
   legacy DB backfill create missing `Index` objects on existing tables. ([#3573],
@@ -1087,3 +1094,4 @@ with **180 merged pull requests** since the first 2.0 milestone tag.
 [#4246]: https://github.com/bytedance/deer-flow/pull/4246
 [#4251]: https://github.com/bytedance/deer-flow/pull/4251
 [#4264]: https://github.com/bytedance/deer-flow/pull/4264
+[#4288]: https://github.com/bytedance/deer-flow/pull/4288
--- a/backend/app/gateway/routers/uploads.py
+++ b/backend/app/gateway/routers/uploads.py
@@ -379,7 +379,13 @@ async def upload_files(
 
             file_ext = file_path.suffix.lower()
             if auto_convert_documents and file_ext in CONVERTIBLE_EXTENSIONS:
-                md_path = await convert_file_to_markdown(file_path)
+                # Reserve the companion .md name in this request's seen set
+                # before writing so conversion cannot silently truncate another
+                # uploaded or derived file (same invariant as form-part dedupe).
+                provisional_md_name = Path(safe_filename).with_suffix(".md").name
+                unique_md_name = claim_unique_filename(provisional_md_name, seen_filenames)
+                md_output = file_path.with_name(unique_md_name)
+                md_path = await convert_file_to_markdown(file_path, output_path=md_output)
                 if md_path:
                     written_paths.append(md_path)
                     md_virtual_path = upload_virtual_path(md_path.name)
--- a/backend/app/gateway/routers/uploads.py
+++ b/backend/app/gateway/routers/uploads.py
@@ -397,6 +397,11 @@ async def upload_files(
                     file_info["markdown_path"] = str(sandbox_uploads / md_path.name)
                     file_info["markdown_virtual_path"] = md_virtual_path
                     file_info["markdown_artifact_url"] = upload_artifact_url(thread_id, md_path.name)
+                else:
+                    # Conversion failed and wrote nothing, so release the claim;
+                    # holding it would rename a later same-stem upload against
+                    # a name nothing occupies.
+                    seen_filenames.discard(unique_md_name)
 
             uploaded_files.append(file_info)
 
--- a/backend/packages/harness/deerflow/client.py
+++ b/backend/packages/harness/deerflow/client.py
@@ -1455,6 +1455,11 @@ class DeerFlowClient:
                         info["markdown_path"] = str(uploads_dir / md_path.name)
                         info["markdown_virtual_path"] = upload_virtual_path(md_path.name)
                         info["markdown_artifact_url"] = upload_artifact_url(thread_id, md_path.name)
+                    else:
+                        # Conversion failed and wrote nothing, so release the
+                        # claim; holding it would rename a later same-stem
+                        # upload against a name nothing occupies.
+                        seen_names.discard(unique_md_name)
 
                 uploaded_files.append(info)
         finally:
--- a/backend/packages/harness/deerflow/client.py
+++ b/backend/packages/harness/deerflow/client.py
@@ -1412,8 +1412,8 @@ class DeerFlowClient:
                 # creating a new ThreadPoolExecutor per converted file.
                 conversion_pool = concurrent.futures.ThreadPoolExecutor(max_workers=1)
 
-        def _convert_in_thread(path: Path):
-            return asyncio.run(convert_file_to_markdown(path))
+        def _convert_in_thread(path: Path, output_path: Path | None = None):
+            return asyncio.run(convert_file_to_markdown(path, output_path=output_path))
 
         try:
             for src_path, dest_name in resolved_files:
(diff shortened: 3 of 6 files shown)


willem-bd:
From the review — one non-inline follow-up: this PR is labeled `risk:high` and changes the `POST /threads/{id}/uploads` response shape. When `uploads.auto_convert_documents` is on and two uploads collide on a stem, the reported `markdown_file` can now be `a_1.md` where it was previously `a.md` (and one file was silently lost). There's no CHANGELOG entry in the diff; worth adding one under `## [Unreleased]` → `### Fixed`.

Suggested entry, matching the existing `- **scope:** … ([#PR])` format:

> - **uploads:** Claim the converted-Markdown companion filename before writing it, so two convertible uploads sharing a stem (or a convertible plus a same-stem `.md` upload) no longer silently clobber each other within one request. When `uploads.auto_convert_documents` is on, the companion `.md` now gets a unique name (e.g. `a_1.md`) instead of overwriting a same-request file; the API response reports the actual name. ([#4288])

plus the link reference at the bottom of the file (numerically ordered with the rest):

> [#4288]: https://github.com/bytedance/deer-flow/pull/4288

Scope label is your call — could also fold under the existing `**gateway:**` Fixed group, which already mentions gateway uploads. Happy to open a follow-up docs PR if you'd rather keep this branch focused on the fix itself.

AnnaSuSu:
Added in ebcf1406. Three deltas from your suggested text, plus one process note.

**1. Widened to both surfaces.** Your draft described the API response only, but the fix touches two call sites that each expose `markdown_file` — the gateway route and `DeerFlowClient.upload_files` (`client.py:1454`), which has the same pre-claim gap and is covered by its own tests in this PR. The entry names both so the SDK side isn't invisible in the release notes:

> - **uploads:** Claim the converted-Markdown companion filename before writing it, so two convertible uploads sharing a stem (or a convertible plus a same-stem `.md` upload) no longer silently clobber each other within one request. When `uploads.auto_convert_documents` is on, the companion `.md` now gets a unique name (e.g. `a_1.md`); `POST /threads/{id}/uploads` and `DeerFlowClient.upload_files` both report the actual name in `markdown_file`. ([#4288])

**2. Kept `**uploads:**` rather than folding into `**gateway:**`.** `**uploads:**` is already an established scope in `### Fixed` (two entries under 2.0.0), and the fix isn't gateway-only per point 1.

**3. Placed mid-section, after the `**gateway:**` entry, not at the tail** — see the process note.

**4. Skipped `CHANGELOG_zh.md`** — it's stale since 2026-06-26 and has no `[Unreleased]` section, so it isn't a live mirror. Say the word if you'd rather it be revived, but that felt like a separate change.

One small precision on the framing: the response *shape* is unchanged — same key set. What changes is the `markdown_file` *value* in the collision case (and that the second file now survives at all).

**Process note, so this doesn't get entered twice:** I went looking for the convention before writing this and it doesn't look like PRs normally carry their own entries — 3 of the last 40 merged PRs touched `CHANGELOG.md`, and there's no requirement in `CONTRIBUTING.md` or the PR template. The mechanism appears to be your periodic sweep: #4270 landed 34 PR references in one pass on 2026-07-18. So #4288 would very likely have been picked up anyway. Happy to have it in the branch either way — just flagging it so the next sweep skips #4288, and I placed the entry mid-section rather than at the end of `### Fixed` to reduce the chance of it conflicting with that sweep.

If you'd prefer the changelog stay sweep-owned and out of feature branches, I'll drop the commit — your call.

AnnaSuSu:
Follow-up on this branch — I found and fixed a regression that **this PR itself introduced**. Flagging it explicitly rather than letting it ride, since you're the one reviewing it.

## What I did first

The `risk:high` / `needs-validation` labels were never actually discharged. `needs-validation` is defined as "needs real-path validation", and every existing test in this PR mocks `convert_file_to_markdown` — so the real converter had never run against this change. I stood up the actual gateway (real `uvicorn`, `uploads.auto_convert_documents: true`) and uploaded genuine OOXML documents converted by markitdown 0.1.5.

Two things came out of it.

## 1. The bug is worse than the PR body describes

On `main`, uploading a real `report.xlsx` + `report.pptx` in one request doesn't only lose a file — **the response misreports the surviving one**:

| | `main` | this branch |
|---|---|---|
| `markdown_file` | `report.md`, `report.md` | `report.md`, `report_1.md` |
| `.md` files on disk | **1** | 2 |
| content of `report.md` | the **pptx**'s markdown | the xlsx's, correct |

So on `main` the xlsx's entry points at `report.md`, but that file holds the pptx's content. The API doesn't just drop a file, it hands back a name whose bytes belong to a different upload. I'll fold that into the PR body.

## 2. The regression my own fix introduced

Pre-claiming the companion name means a conversion that **fails** leaves the name reserved for the rest of the request:

| | `main` | this branch (before `d0185df`) |
|---|---|---|
| user's `notes.md` stored as | `notes.md` | **`notes_1.md`** |
| on disk | `notes.docx`, `notes.md` | `notes.docx`, `notes_1.md` — **no `notes.md`** |

The user's own filename got a `_1` suffix against a name nothing occupies. No bytes lost, but the response reports a name the user never chose, and `main` did neither of those things.

Reachability is low — markitdown is tolerant enough that corrupt `.xlsx`/`.docx`/`.pdf` inputs all fall back to plain text and *succeed*; forcing a genuine `None` took an unreadable file. Low reachability isn't a reason to ship it, so it's fixed in `d0185df`: discard the claim when conversion writes nothing, at both call sites. Three tests, each verified to fail individually when its clause is deleted — two gateway (victim = later `.md` upload; victim = next convertible's companion) and one client.

Worth recording why I missed it the first time: I had considered discard-on-failure and rejected it, on the reasoning that every new test should go red when rolled back against `main`. That criterion is for bug-fix tests. A regression guard defends against a fault the branch itself introduces, so it is green on `main` by definition — the rule I was applying structurally excludes exactly the tests that catch this class of problem.

## Verification

```
3 new tests                      -> pass; each red individually when its clause is deleted
tests/test_uploads_router.py
  + tests/test_client.py         -> 193 passed
full backend suite               -> 8 failed, 7831 passed, 28 skipped
same 8 against origin/main       -> 8 failed  => pre-existing (browserless/crawl4ai/fastcrw), unrelated
ruff check + format --check      -> clean
real path, fixed tree            -> collision: report.md + report_1.md, content correct
                                    failed convert: notes.md keeps its name, no notes_1.md
```

The CHANGELOG entry from `ebcf1406` still reads correctly — it describes the collision fix, which is unchanged. Happy to add a clause for the release behaviour if you'd rather it be spelled out.

WillemJiang (approved):

PR #4289, opened 2026-07-18
fix(gateway): correct GitHub auto-retry claim in webhook route docs

DaoyuanLi2816:
## Why

While investigating fancyboi999's review on #4274 (which flagged that a comment there incorrectly described GitHub as automatically retrying a failed webhook delivery after its 10-second timeout), I checked the rest of the codebase for the same misconception and found a more elaborate version of it here, predating #4274: `github_webhooks.py`'s route docstring claimed "GitHub retries 5xx deliveries with exponential backoff (up to ~5 attempts over ~8 hours)". That specific behavior doesn't exist - verified directly against GitHub's own documentation (cited in the diff), which states failed deliveries (5xx, timeout, or connection error alike) are simply recorded as failed with no automatic redelivery, for both repository and GitHub App webhooks. Every redelivery is explicit: the "Redeliver" button, the REST API, or an operator's own recovery script.

The actual 503-vs-200 status code choice in the route was already correct and needed no change - 503 is still the right response for a transient failure (it keeps the delivery recorded as failed so it can be recovered via one of the explicit paths above), and 200 is still right for permanent/non-retryable conditions. Only the stated *rationale* was factually wrong, and it was repeated in five places (docstring + four inline comments/log messages) plus mirrored in a test docstring/name and one `AGENTS.md` line.

## What changed

- `github_webhooks.py`'s route docstring and four inline comments/log messages no longer describe GitHub as auto-retrying on a timeout or on 5xx; they now state GitHub's actual behavior and cite the source doc.
- `test_github_webhooks.py`: renamed `test_dispatch_failure_returns_503_so_github_retries` to `test_dispatch_failure_returns_503_not_200` and corrected its docstring plus two nearby docstrings/comments with the same framing. No assertions changed - same status codes and response bodies are checked before and after.
- `AGENTS.md`: corrected the GitHub Webhooks router table row to match.

No runtime behavior changes anywhere in this PR.

## Surface area

- [ ] **Backend API** - endpoint / SSE event / request-response shape under `backend/app`
- [ ] **Agents / LangGraph** - agent node, graph wiring, `langgraph.json`, or prompt change
- [ ] **Dependencies** - new/upgraded entry in `backend/pyproject.toml` or `frontend/package.json` (say what it buys us)
- [ ] **Default behavior change** - changes existing behavior without the user opting in (default model, default setting, data shape)
- [x] **Docs / tests / CI only** - no runtime behavior change

## Validation

```
cd backend
uv run pytest tests/test_github_webhooks.py tests/test_github_dispatcher.py tests/test_github_channel.py tests/test_github_registry.py tests/test_channels.py
uv run ruff check app/gateway/routers/github_webhooks.py tests/test_github_webhooks.py
uv run ruff format --check app/gateway/routers/github_webhooks.py tests/test_github_webhooks.py
```

All 313 tests pass; ruff check and format --check are both clean.


fancyboi999 (changes requested):
Thanks @DaoyuanLi2816. I found one remaining documentation error that should be corrected before this is ready.

[P2] Do not describe a successful delivery record as unrecoverable

- Location: `backend/app/gateway/routers/github_webhooks.py:190-197,325-335` and `backend/tests/test_github_webhooks.py:548-559`
- Problem: the PR correctly removes the nonexistent automatic retry, but the replacement says that returning 200 drops the webhook "forever with no way to recover it" / "permanently drops" it. GitHub allows an operator to redeliver a recent webhook delivery from the UI or REST API without requiring that the original delivery be marked failed. The 200 path therefore hides the failure and prevents failure-status-driven recovery from finding it, but it does not make manual/API recovery impossible. Leaving the stronger wording in place gives operators another incorrect durability model in the same documentation this PR is meant to fix.
- Evidence: GitHub's [Redelivering webhooks](https://docs.github.com/en/webhooks/testing-and-troubleshooting-webhooks/redelivering-webhooks) documentation says an operator can select the recent delivery they want to redeliver, and the [repository webhook REST endpoint](https://docs.github.com/en/rest/repos/webhooks#redeliver-a-delivery-for-a-repository-webhook) accepts a delivery ID without a failed-status precondition. The failed-delivery recovery example filters failures by policy; that filter is not a platform restriction on redelivery.
- Suggested fix: describe 503 as preserving the truthful failed status so operators and scheduled recovery scripts can discover the transient failure. State that a mistaken 200 suppresses that failure signal and automatic recovery workflow, while manual/API redelivery remains possible for a recent delivery if an operator identifies it. Apply the correction consistently to the route docstring, the fan-out comment, and the test docstring.
- Test: no runtime assertion needs to change; keep the current behavior tests and add the general redelivery documentation link beside the corrected operational note so the distinction between discoverability and redelivery capability stays explicit.

Reviewed base SHA: `a028dfd5fb70bd6e26c7dbf9e89543c7c006f9a2`
Reviewed head SHA: `ac4e4ee9ff15dcad901907bdd52f308961f0e0d2`

WillemJiang (approved):

Issue #4291, opened 2026-07-19
[feat][RFC] Dual-mode checkpoint storage with LangGraph DeltaChannel


Vanzeren:
### Before you start

- [x] I searched [existing issues](https://github.com/bytedance/deer-flow/issues?q=is%3Aissue) and this is not a duplicate.

## Summary

Adopt LangGraph 1.2's `DeltaChannel` for the `messages` channel as an opt-in
checkpoint storage mode (`checkpoint_channel_mode: full | delta`, default
`full`), cutting checkpoint storage and write-side serialization from O(N²) to
O(N) in conversation turns. The feature ships with a dual-mode safety
framework (process-frozen mode, per-checkpoint metadata markers, fail-closed
compatibility gate), a single materialized-access choke point for all
checkpoint consumers, and mechanism-preserving rewrites of run rollback and
context compaction.

## Motivation

Default checkpointing serializes the full message list into every checkpoint,
so a thread's storage grows quadratically with turns. Upstream's reference
numbers for a 200-turn Deep Agents run: **5.3 GB → 129 MB (41×)** with
`DeltaChannel`. DeerFlow research/coding threads hit the same growth pattern —
they routinely run hundreds of turns with large tool outputs.

Delta mode stores only a sentinel per checkpoint plus the per-step writes;
reads reconstruct state by replaying ancestor writes through the reducer
(`snapshot_frequency=1000` bounds replay). All checkpointer backends
(memory/sqlite/postgres) serve both modes unchanged — the semantics live in
the compiled graph's channel table, not the saver.

## Goals

- O(N) checkpoint storage for `messages` in delta mode; zero data migration
  for existing threads (full → delta is smooth).
- No silent failure modes: every wrong-mode access becomes an explicit error.
- All checkpoint consumers (threads API, branching, regeneration, compaction,
  memory, goal, rollback, client) behave identically in both modes.

## Non-goals

- Performance benchmarking (upstream numbers above; local profiling is
  follow-up — `snapshot_frequency` stays at the upstream default).
- A delta → full conversion tool (the gate error message names it; building it
  is out of scope here).
- Delta channels for other append-only fields (`artifacts`, etc.) until their
  reducers are proven batch-invariant.

### 1. Mode selection and process freeze

`checkpoint_channel_mode` in `config.yaml` (default `full`, warn-once when
unset on an LLM-configured deployment). `make_lead_agent` resolves and freezes
the mode before compiling the graph
[...]

PR #4292, opened 2026-07-19
feat(checkpoint): dual-mode checkpoint storage with LangGraph DeltaChannel

Vanzeren:
## Why

DeerFlow's research/coding threads routinely run hundreds of turns, and the                          
   default checkpointer serializes the full message list into **every**                                 
   checkpoint — thread storage grows O(N²) with turns. Upstream's reference                             
   numbers for a 200-turn Deep Agents run: 5.3 GB baseline vs 129 MB with                               
   LangGraph 1.2 `DeltaChannel` (41×).                                                                  
                                                                                                        
   This PR adopts `DeltaChannel` for the `messages` channel as an **opt-in**                            
   storage mode, and — the bulk of the work — builds the dual-mode safety                               
   framework that keeps full and delta checkpoints coexisting in the same                               
   database without silent corruption.                                                                  
                                                                                                        
   Implements #4291 — design rationale, compatibility matrix, risk analysis, and                        
   open questions (default-flip criteria, delta→full conversion tool, next                              
   candidate fields) live there. 

## What changed                                                                                      
                                                                                                        
   - **New opt-in setting**: `checkpoint_channel_mode: full | delta` in                                 
     `config.yaml`, default `full` — **existing deployments see zero behavior                           
     change**. Delta mode stores only per-step deltas for `messages` (O(N)                              
     storage/serde instead of O(N²)).                                                                   
   - **Mode is restart-required**: the mode is frozen per process and compiled                          
     into the graph's channel schema; hot-switching raises                                              
     `CheckpointModeReconfigurationError`.                                                              
   - **Migration contract**: full → delta is seamless (existing checkpoints seed                        
     the delta channel transparently, no data migration). A full-mode process                           
     opening a delta-mode database now fails closed with                                                
     `CheckpointModeMismatchError` (HTTP 500) instead of silently reading                               
     empty/partial state.                                                                               
   - **User-visible behavior unchanged in both modes**: thread state/history,                           
     branching, regeneration, context compaction, memory, goal continuation, and                        
     cancel-with-rollback all behave identically — verified by dual-mode parity                         
     tests.                                                                                             
   - Latent bugs fixed along the way (each encoded as a regression test):                               
     checkpoint writers that severed parent lineage (breaks delta replay),                              
     rollback restoring via reducers instead of wholesale `Overwrite`, and an                           
     upstream `InMemorySaver` override dropping the first post-migration write.                         
                                                                                                        
   ## Surface area                                                                                      
                                                                                                        
   - [ ] **Frontend UI**                                                                                
   - [x] **Backend API** — threads router: new fail-closed 500 on mode-mismatch;                        
         no request/response shape changes                                                              
   - [x] **Agents / LangGraph** — lead-agent graph is compiled with a                                   
         mode-matched state schema (`messages` = `DeltaChannel` in delta mode);                         
         middleware state schemas normalized to match                                                   
   - [ ] **Sandbox**                                                                                    
   - [ ] **Skills**                                                                                     
   - [x] **Dependencies** — `langgraph>=1.2.0` (buys us `DeltaChannel` +                                
         `Overwrite`); `langgraph-checkpoint-sqlite/postgres >=3.1.0` (delta-aware                      
         savers); lockfile re-resolved the langchain 1.2.x subgraph                                     
   - [ ] **Default behavior change** — default stays `full`; delta is strictly                          
         opt-in                                                                                         
                                                                                                        
   ## Screenshots / Recording                                                                           
                                                                                                        
   N/A — no frontend changes.                                                                           
                                                                                                        
   ## Validation                                                                                        
                                                                                                        
   - `cd backend && make lint` — clean                                                                  
   - `cd backend && make test` — **277 passed** (pre-rebase full suite)                                 
   - Post-rebase onto latest `main`: focused checkpoint suite re-run —                                  
     `test_checkpoint_mode.py`, `test_checkpoint_state.py`,                                             
     `test_delta_channel_checkpointers.py`, `test_context_compaction.py`                                
     (**26 passed**, 4 postgres skips)                                                                  
   - Dual-mode e2e parity: `test_threads_checkpoint_mode.py`,                                           
     `test_gateway_checkpoint_mode.py` (state/history/branch/compress/regenerate                        
     on memory + sqlite, both modes)                                                                    
   - Rollback: `tests/test_run_worker_rollback.py` — capture, `Overwrite`                               
     restore, pending-write replay, both modes                                                          
   - Not run: local performance benchmark (explicitly out of scope; follow-up                           
     will profile storage/latency vs full mode before any default flip)                                 
                                                                                                        
   ## AI assistance                                                                                     
                                                                                                        
   **Tool(s) used:** oh-my-pi/codex (coding agent)                                                         
                                                                                                        
   **How you used it:** AI executed the 9-task implementation plan from                                 
   `spec.md` with TDD per task, drafted the AGENTS.md section and the RFC doc,                          
   and drove the review rounds. Architecture calls (dual-mode gate direction,                           
   mutation-graph + `Overwrite` writes, dropping the hand-rolled checkpoint                             
   writer) were decided and verified against LangGraph internals during the                             
   session; I reviewed the final diff and design trade-offs.                                            
                                                                                                        
 unreviewed AI output.


fancyboi999 (changes requested):
I found two blocking issues on `a8a800ad63e775791beabbba6e96cd7c5c0d29a3`.

### [P1] Complete the materialized-state boundary before enabling delta mode

Locations: `backend/packages/harness/deerflow/runtime/runs/worker.py:971-994` and `backend/packages/harness/deerflow/agents/factory.py:127-159`.

The PR's central invariant is that every checkpoint consumer goes through the mode-matched `CheckpointStateAccessor`, but two public/runtime paths still bypass it:

* `_prepare_goal_continuation_input()` reads `messages` from the raw `CheckpointTuple`. In delta mode that value is a `_DeltaSnapshot`/sentinel rather than a `list`, so `_read_checkpoint_messages()` returns `[]`. Against the exact PR tree, I created a delta thread containing a human and assistant message plus an active goal. `CheckpointStateAccessor.aget()` materialized both messages, while `_prepare_goal_continuation_input()` never called the evaluator and persisted `blocker=run_failed` with `stand_down_reason=no_durable_end_of_turn`. Thus opt-in delta mode disables goal continuation after an otherwise durable assistant turn.
* `create_deerflow_agent()` selects the delta/full schema but neither injects the mode marker nor performs the compatibility gate on graph operations. With one `InMemorySaver`, a delta graph persisted `u1/delta-ai`; a full graph created through this public factory then returned an empty state instead of raising, accepted another turn, and the subsequent delta read contained only `u2/full-ai`. The original turn was silently lost. This reproduces exactly the corruption that the PR says the fail-closed gate prevents.

Please route goal evaluation and every conversation-signature reread through a mode-matched accessor (retaining raw tuple reads only for tuple-only metadata such as pending writes), and make the public factory's persisted graph entry points enforce the same inject/gate contract. If the raw `CompiledStateGraph` API cannot enforce that invariant without a larger wrapper/API change, keep `checkpoint_channel_mode="delta"` plus a checkpointer out of this factory for this PR and scope it to the guarded application/client paths. Add real delta goal-continuation coverage and a mixed-mode public-factory regression that asserts an explicit mismatch before any state is read or written.

### [P2] Keep the documented default test installation collectible

Location: `backend/tests/test_delta_channel_checkpointers.py:32`.

This module imports `langgraph.checkpoint.postgres` unconditionally, but postgres remains an optional extra. In a fresh isolated environment using the repository's documented/default install (`uv sync --frozen --all-packages`, equivalent to the backend `make install` contract), collecting the focused suite fails immediately with `ModuleNotFoundError: No module named 'langgraph.checkpoint.postgres'`. The workflow added by this PR hides the regression by installing `--extra postgres`, while local `make install && make test` still fails before the memory/sqlite cases can run.

Please import the Postgres saver only inside the Postgres fixture path (and skip that parameter when the extra or `TEST_POSTGRES_URI` is absent), or make the dependency part of the normal dev group if every default test run is intentionally required to install it. Add a default-install collection check so CI covers the same command contributors are told to run.

Validation performed in a detached worktree with an isolated uv environment:

* Focused checkpoint/Gateway/rollback suite with the postgres extra: `305 passed, 4 skipped`.
* `ruff check .`: passed; `ruff format --check .`: 886 files already formatted.
* Default install focused collection: failed as described above.
* Independent LangGraph runtime probes reproduced both P1 failure modes above.
* The full backend suite was also started but not completed: a live `test_client_e2e` case hit an external LLM `APIConnectionError`/proxy reset and the long run was then interrupted. I am not treating that run as passing evidence or as a PR finding.

Vanzeren:
Addressed both review items in b703c7ac.

**1. Materialization boundary** — fixed at both surfaces:
- `runtime/runs/worker.py`: goal-continuation message reads now go through `CheckpointStateAccessor.aget` (new `_materialized_checkpoint_messages` helper); `checkpoint_ns: ""` preserves checkpoint identity. Raw `channel_values` reads remain only for tuple-level metadata (checkpoint id, `pending_writes`). Added `test_goal_worker_evaluates_materialized_messages_in_delta_mode` reproducing the reviewer's scenario: delta checkpoint + durable assistant turn → evaluator receives the full materialized list and continuation is returned. Counterfactual verified: the old raw-read logic yields `[]` on the same seeded delta thread (which produced `stand_down=no_durable_end_of_turn` and silently disabled goal continuation).
- `create_deerflow_agent`: `checkpoint_channel_mode="delta"` + `checkpointer` now raises `ValueError` at construction, before any state is read or written. Correct that this surface bypasses mode-marker injection and the fail-closed gate — it reproduced the mixed-mode corruption end to end. Applied the accepted scope cut: non-persistent factory graphs keep `delta` for middleware-level tests; every persisted path (`make_lead_agent`, `DeerFlowClient`, `inject_checkpointer_mode` on PostgreSQL) was already guarded. Regression test asserts the saver is untouched.

**2. Test collection / install policy** — `AsyncPostgresSaver` is now imported inside the saver fixture via `pytest.importorskip`, so `uv sync --group dev` (the documented contributor path) collects and runs the full suite without the `postgres` extra. Added a `default-install-collection` CI job (`uv sync --group dev` + `uv run pytest --collect-only -q`) to keep this honest; the main job keeps `--extra postgres` so the postgres contract cases still execute against the service.

Also fixed a pre-existing environment-dependent failure hit during local validation: `test_returns_in_memory_saver_when_not_configured` patched `get_app_config` at its definition module, but the checkpointer provider binds it directly at import time — a local `config.yaml` selecting a persistent `database` backend leaked in and the test returned `SqliteSaver`. It now patches the provider's use site.

Validation: `make lint` clean; full backend suite `7937 passed, 27 skipped`; the new delta goal-continuation test was red-checked against the old raw-read logic.

fancyboi999 (changes requested):
Thanks @Vanzeren. I found two issues that should be addressed before this is ready.

### [P1] Preserve extension-owned state when materializing branch writes

- Location: `backend/app/gateway/services.py:553-579` and `backend/app/gateway/routers/threads.py:664-679`
- Problem: the source branch snapshot is materialized with the assistant-specific graph, but `build_checkpoint_state_mutation_accessor()` always compiles a state-only graph from the base `ThreadState`. Any channel contributed by a custom agent or extension middleware is therefore unknown to the writer and is silently discarded. The same base-only mutation path also makes `POST /state` return 200 while ignoring writes to extension-owned channels.
- Evidence: against `b703c7ac`, I created a real `InMemorySaver` source thread whose effective state schema adds `extension_state`, then passed its materialized values through the PR's branch mutation graph. `messages` survived, but `extension_state` changed from `{"token": "must-survive"}` to `None`. Before this rewrite, the raw branch checkpoint clone preserved every persisted channel. This is a data-loss regression for the supported custom middleware state-schema extension point.
- Suggested fix: compile the no-op mutation graph with the thread's full effective state schema, including assistant/custom middleware channels, and use that schema consistently for branch and state-update operations. Do not silently accept and drop an unknown channel; if a caller truly cannot update one, reject it explicitly rather than returning a false-success 200.
- Test: add an endpoint-level regression with a custom `AgentMiddleware.state_schema`: write an extension-owned value, branch the thread, materialize the branch with the same assistant, and assert the value survives. Also update that field through `POST /state` and assert the persisted value changes (or that an intentionally unsupported field receives an explicit 4xx).

### [P2] Advance the config version for the new persisted setting

- Location: `config.example.yaml:18` and `config.example.yaml:1650-1657`
- Problem: this PR adds `database.checkpoint_channel_mode` to the configuration schema but leaves `config_version` at 26. Existing installations already on version 26 will not get the repository's outdated-config warning, so `make config-upgrade` is not discoverable for this new field. This directly violates the config-versioning rule in `backend/AGENTS.md`.
- Evidence: both current `origin/main` and `b703c7ac` declare `config_version: 26`, while only the PR adds `checkpoint_channel_mode`.
- Suggested fix: bump `config.example.yaml` to the next version (27 against the current base), resolving any version collision again if `main` moves before merge.
- Test: cover a version-26 user config against the updated example and assert it is reported as outdated and the upgrade path merges `database.checkpoint_channel_mode: full` without changing the existing database backend settings.

Validation on the current head used a detached worktree and isolated uv environment: default-install full test collection passed; the focused checkpoint/Gateway/rollback surface passed (`667 passed, 4 skipped`); `ruff check .` and `ruff format --check .` passed. The broader backend run excluding `tests/test_client_e2e.py` reached `7904 passed, 14 skipped`; its only three failures were the explicitly live `test_create_deerflow_agent_live.py` cases, all failing on an external proxy connection reset rather than assertions.

Vanzeren:
Addressed both items in 9297680e.

**1. Extension-owned state in materialized branch writes** — the mutation path now uses the thread's effective schema everywhere:
- `build_state_mutation_graph(as_node, mode, state_schema)` and `build_checkpoint_state_mutation_accessor(..., state_schema=...)` accept the schema explicitly; new helpers `graph_state_schema` / `graph_writable_channels` / `graph_reducer_channels` derive it from the compiled assistant graph (`StateGraph.schemas` first entry).
- Branch (`services.py`/`threads.py`): the branch write compiles the mutation graph from `graph_state_schema(source_accessor.graph)` — the same assistant-specific graph that materialized the source snapshot — so custom `AgentMiddleware.state_schema` channels survive. Counterfactual verified against your repro shape: base-schema mutation graph drops `extension_state` to `None`; effective-schema preserves it.
- `POST /state`: resolves the thread's `assistant_id` from the thread record, derives the effective schema via the read accessor, and compiles the mutation graph with it. Writes to fields outside the graph's channels are now rejected with **422** (listing the unknown fields) instead of a false-success 200. Reducer detection is dynamic (`BinaryOperatorAggregate` or `DeltaChannel`), so extension reducer channels get `Overwrite` replace semantics in both modes.
- Test: `test_branch_and_state_update_preserve_extension_owned_channels` — endpoint-level, real `AgentMiddleware.state_schema` via `create_agent`: extension value survives branching, updates through `POST /state` persist, and an unknown field returns 422.

**2. Config version** — `config_version` bumped 26 → 27 in `config.example.yaml`, plus the two other places the literal lives (Helm chart `values.yaml` embedded config block and its README example — `scripts/check_config_version.sh` drift check passes), and a support-bundle fixture. Test: `test_version_26_config_upgrades_to_checkpoint_channel_mode` drives a v26 user config against the real example file (asserts the outdated warning fires) and runs the real `scripts/config-upgrade.sh`, asserting `database.checkpoint_channel_mode: full` is merged while `backend`/`sqlite_dir` stay untouched.

Also updated the wholesale-replacement rule in `backend/AGENTS.md` to make the effective-schema requirement explicit for future mutation-graph consumers.

Validation: `make lint` clean; full backend suite `7939 passed, 27 skipped`; focused checkpoint/gateway/rollback surface green; drift check green.

fancyboi999 (changes requested):
Thanks @Vanzeren. I found one remaining issue that should be addressed before this is ready.

### [P1] Close the effective-schema boundary for reads and wholesale branch writes

- Location: `backend/app/gateway/routers/threads.py:665-684,931-955`
- Problem: the latest fix teaches the branch mutation graph about extension-owned channels, but it still sends every materialized channel except `messages` through its reducer again. A custom middleware reducer can therefore change an already-aggregated value while the branch is created. The same boundary is still missing on `GET /threads/{id}/state`: that route builds the accessor without resolving the thread's `assistant_id`, so a custom agent checkpoint is materialized with the default lead schema and extension channels disappear from the response.
- Evidence: on current head `9297680e`, a real LangGraph `Annotated[list[str], reducer]` channel materialized as `["merged", "payload"]` on the source and became `["merged", "merged", "payload"]` after the exact branch mutation path. Wrapping that channel in `Overwrite` preserved the value. Separately, reading the same checkpoint through the base `ThreadState` accessor returned no `extension_state` at all. The new endpoint test uses a `LastValue` extension (no reducer) and monkeypatches `build_checkpoint_state_accessor` to always return the custom graph, so it bypasses both production failures.
- Suggested fix: make thread metadata -> `assistant_id` -> effective graph/schema resolution a single reusable accessor boundary for state endpoints instead of resolving it ad hoc in selected handlers. When a materialized snapshot is copied to a branch, derive reducer channels from that effective mutation graph and wrap every copied reducer value in `Overwrite` (the same replace semantics already used by `POST /state`), rather than special-casing only `messages`.
- Test: use a real custom `AgentMiddleware.state_schema` with a non-identity reducer; assert `GET /state` returns the extension value and a branch preserves it byte-for-byte in both full and delta modes. Keep the unknown-field 422 regression as a separate assertion.

Current-head validation used a detached worktree and isolated uv environment. The checkpoint/thread/config/support focused suite passed (`92 passed`), Ruff check/format, config-version drift, and `git diff --check` passed. The two external runtime invariants above failed as described; the `Overwrite` control passed.

Vanzeren:
Addressed the remaining P1 in 097b8f29.

**1. Single resolution boundary** — new `build_thread_checkpoint_state_accessor` / `build_thread_checkpoint_state_mutation_accessor` in `services.py`: thread metadata -> `assistant_id` -> effective assistant graph, derived once and reused. `GET /threads/{id}/state` and `POST /state` both go through it now (no more ad hoc record/schema plumbing in the POST handler).

**2. Branch replace semantics** — the branch write derives reducer channels from the effective mutation graph (`graph_reducer_channels`: `BinaryOperatorAggregate` + `DeltaChannel`) and wraps **every** copied reducer value in `Overwrite`, not just `messages`. Already-aggregated values can no longer be re-merged by a custom reducer.

**Tests** — rewritten per your guidance: real `AgentMiddleware.state_schema` with a non-identity reducer (`operator.add` list channel), parametrized over full and delta modes. `GET /state` returns the extension value, `POST /state` replaces it, branch preserves it byte-for-byte; the unknown-field 422 is a separate test. Only `get_run_context` and `resolve_agent_factory` are stubbed (selectively: unresolved assistant -> default schema), so the production builders and the assistant-resolution path run for real. Mutation-checked: reverting either fix makes the new tests fail (`KeyError: ext_list` / non-`Overwrite` copy).

Validation: `make lint` clean; full backend suite `7941 passed, 27 skipped`.

backend/packages/harness/deerflow/runtime/checkpoint_state.py
```diff
@@ -0,0 +1,184 @@
...
+                break
+        return result
+
+    async def ahistory(self, config: dict[str, Any], *, limit: int | None = None) -> list[Any]:
+        prepared = self._prepare_config(config)
+        await aensure_checkpoint_mode_compatible(self.checkpointer, prepared, self.mode)
+        if limit is not None and limit <= 0:
+            return []
+        result = []
+        async for snapshot in self.graph.aget_state_history(prepared):
```
willem-bd: **P1 - perf regression: `limit` isn't pushed to SQL.**

`ahistory` (and the sync `history`) iterate `self.graph.aget_state_history(prepared)` and `break` at `limit` in Python, but never pass `limit` into the graph call. Verified against langgraph 1.2.9: `Pregel.aget_state_history(config, *, filter, before, limit)` accepts `limit` and threads it to `checkpointer.alist(..., limit=limit)`, which the sqlite/postgres savers push into the SQL `LIMIT` clause. With no `limit` passed, `alist(limit=None)` fetches **all** history rows and materializes them all (pregel's eager `list(...)` consume) before the caller discards all but N.

The old `checkpointer.alist(config, limit=N)` pushed `LIMIT N` to SQL. This regresses `get_thread_history`, branch scanning (`_BRANCH_HISTORY_SCAN_LIMIT`), and regenerate's `REGENERATE_HISTORY_RAW_SCAN_LIMIT` - the deep-history paths delta mode is meant to help.

Fix: `async for snapshot in self.graph.aget_state_history(prepared, limit=limit):` (keeping the `limit <= 0 -> []` guard above for the explicit-empty semantic).

backend/packages/harness/deerflow/runtime/checkpoint_mode.py
```diff
@@ -0,0 +1,76 @@
...
+def ensure_checkpoint_mode_compatible(checkpointer: Any, config: dict[str, Any], mode: CheckpointChannelMode) -> None:
+    if mode == "delta":
+        return
+    _raise_if_incompatible(checkpointer.get_tuple(config), mode)
+
+
+async def aensure_checkpoint_mode_compatible(checkpointer: Any, config: dict[str, Any], mode: CheckpointChannelMode) -> None:
+    if mode == "delta":
+        return
+    _raise_if_incompatible(await checkpointer.aget_tuple(config), mode)
```
willem-bd: **P2 - extra checkpoint round-trip on every read/write in the default (full) mode.**

`aensure_checkpoint_mode_compatible` does a standalone `await checkpointer.aget_tuple(config)` before every `aget` / `ahistory` / `aupdate`. In a deployment that never enables delta, no checkpoint ever carries the marker, so this is pure overhead - and `aget_state` already fetches the tuple internally, so each accessor read is now >=2 checkpoint fetches.

Since `aget_state` returns a `StateSnapshot` that carries `.metadata`, the marker check could be done on the returned snapshot (raise `CheckpointModeMismatchError` post-read) instead of a separate `aget_tuple`. Reading the blob is harmless - only silently *using* an empty/partial delta state is the danger - so failing after the read preserves fail-closed semantics while removing the extra round trip.

backend/app/gateway/routers/threads.py
```diff
@@ -775,57 +794,49 @@ async def patch_thread(thread_id: str, body: ThreadPatchRequest, request: Reques
...
+    )
 
-    # Derive accurate status from the checkpointer
-    config = {"configurable": {"thread_id": thread_id, "checkpoint_ns": ""}}
     try:
-        checkpoint_tuple = await checkpointer.aget_tuple(config)
+        snapshot = await accessor.aget(config)
+        snapshot_config = snapshot.config or {}
+        checkpoint_id = snapshot_config.get("configurable", {}).get("checkpoint_id")
+        checkpoint_tuple = await checkpointer.aget_tuple(snapshot_config) if checkpoint_id else None
```
willem-bd: **P2 - `get_thread` now does ~3 checkpoint fetches per request (was 1).**

This path runs: `accessor.aget` (compat-gate `aget_tuple` + `aget_state`'s own `aget_tuple` + materialize) **plus** this second `checkpointer.aget_tuple` just to read `pending_writes` for status derivation. The old code read everything from a single `aget_tuple`.

The `__error__` status signal could be derived from `snapshot.tasks` (Pregel tasks expose error state), which would drop this extra `aget_tuple`. Combined with folding the compat gate onto the snapshot (see checkpoint_mode.py), `get_thread` could return to a single checkpoint fetch in the common full-mode case.

backend/app/gateway/services.py
```diff
@@ -536,11 +541,148 @@ def build_run_config(
...
+    configurable["checkpoint_ns"] = ""
+    if checkpoint_id is not None:
+        configurable["checkpoint_id"] = checkpoint_id
+
+    if ctx.app_config is not None:
+        config.setdefault("context", {})["app_config"] = ctx.app_config
+    inject_checkpoint_mode(config, ctx.checkpoint_channel_mode)
+
+    agent_factory = resolve_agent_factory(assistant_id)
+    graph = agent_factory(config=config)
```
willem-bd: **P2 - every state-read endpoint now builds a full lead agent per request.**

`build_checkpoint_state_accessor` calls `agent_factory(config=config)` (= `make_lead_agent`, uncached - confirmed `resolve_agent_factory` returns it directly with no caching wrapper) on every `get_thread` / `get_thread_state` / `get_thread_history` / `update_thread_state` / branch / compaction. That instantiates the model client, loads skills, and constructs the full middleware stack on each GET - paths that previously only touched the checkpointer.

The agent is needed only to surface extension-middleware channels in the effective schema. Since mode is process-frozen, the effective schema per `(assistant_id, mode)` is stable; consider caching the compiled graph - or at least `graph_state_schema` / `graph_writable_channels` / `graph_reducer_channels` - per `(assistant_id, mode)` rather than rebuilding per request.

backend/packages/harness/deerflow/client.py
```diff
@@ -555,41 +566,43 @@ def list_threads(self, limit: int = 10) -> dict:
...
+        for snapshot in accessor.history(config):
+            values = dict(snapshot.values or {})
+            if "messages" in values:
+                values["messages"] = [self._serialize_message(message) if hasattr(message, "content") else message for message in values["messages"]]
+
+            snapshot_config = snapshot.config or {}
+            configurable = snapshot_config.get("configurable", {})
+            parent_config = snapshot.parent_config or {}
+            parent_configurable = parent_config.get("configurable", {})
+            raw_tuple = checkpointer.get_tuple(snapshot_config)
```
willem-bd: **P2 - `DeerFlowClient.get_thread` is now N+1.**

`accessor.history()` yields `StateSnapshot`s, which don't expose `pending_writes`, so this fetches `get_tuple` once per snapshot to recover them. The old `checkpointer.list()` yielded tuples that already carried `pending_writes` in a single walk - for a 200-checkpoint thread this is 200 extra lookups.

Options: a single `alist` pass for pending_writes alongside the history walk, or accept the loss of per-snapshot `pending_writes` in this client view (the gateway path doesn't need it).

backend/packages/harness/deerflow/checkpoint_patches.py
```diff
@@ -0,0 +1,58 @@
...
+    override collect the terminating checkpoint's writes *before* treating its
+    blob as the seed, which is the correct order. This patch delegates
+    InMemorySaver to the base implementation - one ``get_tuple`` per ancestor
+    instead of a single fused walk, which is fine for dict-backed storage.
+
+    Idempotent. Remove once LangGraph fixes the override upstream.
+    """
+    if getattr(InMemorySaver, _PATCH_FLAG, False):
+        return
+    InMemorySaver.get_delta_channel_history = _get_delta_channel_history_via_base  # type: ignore[method-assign]
```
willem-bd: **P3 - monkey-patching the upstream `InMemorySaver` override is fragile.**

Verified against langgraph 1.2.9: both `BaseCheckpointSaver.get_delta_channel_history` and the `InMemorySaver` override exist, and the override is a real, perf-tuned single parent-chain walk - this patch swaps it for the base's one-`get_tuple`-per-ancestor walk. The perf tradeoff is acceptable for dict-backed storage and the upstream bug (first post-migration write dropped) is real, but:

- The patch is global and unconditional - it mutates `InMemorySaver` even in full-mode-only processes (no-op there, but still mutates the upstream class at import time, including tests).
- No version guard: if LangGraph fixes the override in a `1.2.x` patch or changes the signature, this silently overrides the fix or breaks. Consider a `try/except` guard on the assignment and/or a pinned-version assertion, plus a link to an upstream issue so the 'remove once fixed' comment is actionable.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
```diff
@@ -443,8 +448,19 @@ def _load_enabled_available_skills(available_skills: set[str] | None, *, app_con
 def make_lead_agent(config: RunnableConfig):
     """LangGraph graph factory; keep the signature compatible with LangGraph Server."""
     runtime_config = _get_runtime_config(config)
+    trusted_config = config.get("configurable", {}) or {}
     runtime_app_config = runtime_config.get("app_config")
-    return _make_lead_agent(config, app_config=runtime_app_config or get_app_config())
+    if isinstance(runtime_app_config, AppConfig):
+        requested_mode = trusted_config.get(
```
willem-bd: **P3 - mode precedence is inverted for a trust boundary.**

`requested_mode` reads the client-visible `configurable[INTERNAL_CHECKPOINT_MODE_KEY]` *before* `app_config.database.checkpoint_channel_mode`. `build_run_config` strips that key on the gateway run path and `freeze_checkpoint_channel_mode` makes a mismatched value non-corrupting (it raises `CheckpointModeReconfigurationError` -> 500; a matching value is a no-op), so this isn't a corruption hole - but the precedence is backwards for server-owned state. If the LangGraph-runtime factory path (langgraph server invoking `make_lead_agent` directly with a client config) doesn't go through `build_run_config`'s strip, a client can trip a 500 by injecting a mismatched key.

Suggested: derive `requested_mode` solely from the frozen `app_config` / `_frozen_checkpoint_channel_mode` and ignore the config key on entry (it's only meaningful when set internally by `inject_checkpoint_mode`).

backend/packages/harness/deerflow/runtime/runs/worker.py
```diff
@@ -1055,65 +1113,101 @@ async def _persist(
...
+        # Unreachable in practice: a rollback point can only be captured
+        # through the bound accessor. Stay fail-closed.
+        logger.warning("Run %s rollback skipped: agent accessor unavailable", run_id)
+        return
+
+    mutation_graph = build_state_mutation_graph("rollback_restore", accessor.mode)
+    mutation_accessor = CheckpointStateAccessor.bind(mutation_graph, checkpointer, mode=accessor.mode)
+    restored_config = await mutation_accessor.aupdate(
+        rollback_point.config,
+        {"messages": Overwrite(list(rollback_point.messages))},
```
willem-bd, line 1213: **P3 - rollback restores only `messages`; non-message channel restoration relies on forking and isn't asserted.**

`_rollback_to_pre_run_checkpoint` overwrites only `messages` here and relies on `update_state`'s forking to inherit `sandbox` / `goal` / `todos` / `artifacts` / `delegations` from the pre-run parent checkpoint. I verified the forking semantics hold in langgraph 1.2.9 (`update_state` with `checkpoint_id` creates a child whose non-updated channels inherit from the parent), so this is *correct* - but it's the load-bearing assumption of the whole new rollback design, and the rollback tests only assert `messages` restoration + pending-write replay. A test that mutates a non-message channel pre-run and verifies it's restored post-rollback would lock this in.

(Also: no test currently pins checkpoint round-trip count or large-history + `limit`, so the P1/P2 perf regressions aren't caught - a counted-mock saver test would prevent silent reintroduction.)

willem-bd (commented):
Review pass - posting P1/P2/P3 findings as inline comments.

- **P1** (1): `ahistory` doesn't push `limit` to SQL - real perf regression with a one-line fix.
- **P2** (4): extra checkpoint round-trips in the default (full) mode on every read/write; `get_thread` triple-read; per-request full lead-agent build on state-read endpoints; `DeerFlowClient.get_thread` N+1.
- **P3** (3): fragile `InMemorySaver` monkey-patch; `make_lead_agent` mode-precedence is inverted for a trust boundary; rollback non-message restoration relies on forking and is untested.

Core dual-mode safety design is sound; these are perf/hardening items, none are delta-mode correctness bugs.

backend/AGENTS.md
```diff
@@ -674,6 +674,28 @@ This invokes `alembic revision --autogenerate` against the live ORM models. Revi
...
+
+**Where things live**:
+- `runtime/checkpoint_mode.py` — mode freeze, marker injection, delta detection, compatibility gate, both error types
+- `runtime/checkpoint_state.py` — `CheckpointStateAccessor`, `build_state_mutation_graph`, `RollbackPoint`
+- `checkpoint_patches.py` (package root) — saver patches: delta-history folding for `InMemorySaver` (delegating to the base walk), stable message IDs across materialization, upstream first-write drop fix
+- `agents/thread_state.py` — `ThreadState`/`DeltaThreadState`, `DELTA_MESSAGES_FIELD` (`DeltaChannel` with `snapshot_frequency=1000`), schema adaptation helpers
+- `runtime/context_compaction.py` — compaction via accessor + mutation graph (reference consumer)
+- Tests: `tests/test_checkpoint_mode.py` (freeze/detect/gate), `tests/test_checkpoint_state.py` (accessor/mutation graph), `tests/test_delta_channel_checkpointers.py` (saver parity), `tests/test_threads_checkpoint_mode.py`, `tests/test_gateway_checkpoint_mode.py` (dual-mode e2e parity), `tests/test_context_compaction.py` (mutation-graph write, no scheduling), `tests/test_run_worker_rollback.py`
```
willem-bd, line 732: **Referenced parity test files are not on this branch.**

`backend/AGENTS.md` lists `tests/test_threads_checkpoint_mode.py` and `tests/test_gateway_checkpoint_mode.py` as the dual-mode e2e parity tests, and the PR Validation section states they were run (state/history/branch/compress/regenerate, memory + sqlite, both modes). Neither file exists at the current head (`b5f68074`) - the only checkpoint-mode test shipped is `test_checkpoint_mode.py` (unit-level: freeze/detect/gate). So the parity coverage the docs and validation point at cannot be verified against shipped tests.

Either add the two files (if they were intended but missed from the commit) or update this list and the Validation section to match what is actually here.

willem-bd (commented):
Addendum to the earlier review pass - one additional finding not yet captured in the existing inline threads. The earlier perf-oriented threads remain open in the current head.

backend/packages/harness/deerflow/runtime/checkpoint_state.py
Vanzeren: Fixed in 40452830. `ahistory`/`history` now pass `limit` through to `aget_state_history(prepared, limit=limit)` (and the sync twin), so the savers push it into the SQL `LIMIT` clause; the `limit <= 0 -> []` guard stays for the explicit-empty semantic.

Locked in by `test_ahistory_pushes_limit_into_alist_and_reads_each_snapshot_once`: a counting saver asserts `alist` receives `limit=2` and zero `aget_tuple` calls in the read path. Mutation-verified (removing the kwarg fails the test).

backend/packages/harness/deerflow/runtime/checkpoint_mode.py
Vanzeren: Fixed in 40452830. The read-side gate is folded onto the returned snapshot: `aget`/`ahistory` check `CHECKPOINT_MODE_METADATA_KEY` on `snapshot.metadata` post-read and raise `CheckpointModeMismatchError` there — fail-closed semantics preserved (reading the blob is harmless, using it is not). Writes (`aupdate`/`update`) keep the pre-write `aget_tuple` since a write can't be un-applied.

`test_aget_fetches_the_checkpoint_exactly_once_in_full_mode` pins one `aget_tuple` per `aget` in full mode (mutation-verified: re-adding the standalone gate fails it); zero-limit history now costs zero fetches; delta-marked snapshots raise via `test_full_accessor_raises_when_the_returned_snapshot_is_delta_marked` with no pre-read fetch.

backend/app/gateway/routers/threads.py
Vanzeren: Partially addressed in 40452830 — down from 3 fetches to 2, but the second fetch is semantically required, not redundant.

The compat gate is folded onto the snapshot (see checkpoint_mode.py reply), so `accessor.aget` is back to one fetch. The raw `pending_writes` fetch on the resolved `snapshot.config` stays: I verified empirically against langgraph 1.2.9 that `__error__` writes attached to the latest checkpoint *after* completion (rollback reattachment, worker error fallback via `aput_writes`) never surface in `snapshot.tasks` — `get_state` only populates tasks for writes that were pending while a task was still scheduled. Repro: seed a completed thread, `aput_writes(latest_config, [("__error__", ...)])`, `aget_state` returns `tasks == []`. Deriving status from `snapshot.tasks` alone would report `idle` for a thread whose last run failed post-checkpoint. The existing `test_get_thread_status_uses_raw_pending_writes_for_materialized_checkpoint` pins this.

backend/app/gateway/services.py
Vanzeren: Fixed in 40452830. `build_checkpoint_state_accessor` now resolves the graph through `_state_accessor_graph`, a cache keyed by `(assistant_id, mode)`:

- Factory identity is re-validated on every call, so patched/plugged factories (tests, extensions) take effect immediately.
- Bounded at 64 distinct assistants (clears on overflow) — no unbounded growth from arbitrary assistant_id strings.
- Accessor operations (`aget_state`/`aupdate_state`) never execute graph nodes or middleware, so per-request variations (user, model, skills) cannot affect materialization semantics; the compiled graph is stable per `(assistant_id, mode)` exactly as you noted.

Per-request cost is now `inject_checkpoint_mode` + `CheckpointStateAccessor.bind` (both trivial); the model client / skill loading / middleware stack are built once per assistant.

backend/packages/harness/deerflow/client.py
Vanzeren: Fixed in 40452830. `get_thread` now does one `checkpointer.list(config)` walk up front to collect `pending_writes` per checkpoint id, then joins them onto the snapshots from `accessor.history()`. Two streaming walks total (the history walk + the raw tuple walk), zero per-snapshot `get_tuple` round-trips — for a 200-checkpoint thread that's 2 queries instead of 201. `TestThreadQueries::test_get_thread` updated to pin the single-walk contract.

backend/packages/harness/deerflow/checkpoint_patches.py
Vanzeren: Hardened in 40452830:

- **Stand-down guard**: `_upstream_override_present()` checks the override still exists on `InMemorySaver`; if upstream removes or refactors it (i.e. the fix landed), the patch returns without touching anything.
- **try/except** on the assignment (`AttributeError`/`TypeError`) with a logged skip.
- **Version guard**: validated against langgraph 1.2.9; any newer installed version logs a re-inspection warning instead of silently overriding a potential upstream fix.
- On the upstream-issue link: no upstream issue exists for this bug yet (searched langchain-ai/langgraph). Rather than fabricate a link, the docstring says so explicitly and instructs re-checking `InMemorySaver.get_delta_channel_history` on every langgraph upgrade. Filing an upstream issue with a minimal repro is a reasonable follow-up.
- Global application at import: intentional — every process that materializes delta threads (gateway, workers, in-process runtime, tests) must have the fix; in full-mode processes the patched path is never exercised.

Guard tests: `test_inmemory_delta_history_patch_is_active`, `..._stands_down_without_upstream_override`, `..._warns_on_unvalidated_langgraph` in `tests/test_delta_channel_checkpointers.py`.

backend/packages/harness/deerflow/agents/lead_agent/agent.py
Vanzeren: Fixed in 40452830. Precedence is now server-owned at the trust boundary:

- **First freeze** (process not yet frozen): `requested_mode` comes solely from the resolved `app_config`; a client-supplied `configurable` key is ignored — a fresh process can't be frozen to the wrong mode or crashed by a forged key.
- **Once frozen**: the internally injected key (run worker / gateway) or the app_config value must match the frozen mode; any mismatch raises `CheckpointModeReconfigurationError` (fail-closed 500) instead of silently diverging.

This is pinned by three tests: `test_direct_langgraph_request_cannot_select_delta_in_full_process` (client key ignored at first freeze), `test_gateway_runtime_app_config_can_supply_its_frozen_internal_mode` (reloaded config + internally injected key stays consistent), and `test_yaml_mode_change_is_rejected_when_graph_is_reconstructed` (config.yaml flip fails closed).

Note: I deliberately did *not* go further and ignore the key entirely once frozen (use the frozen value unconditionally) — that would mask a genuine operator `config.yaml` change, which the fail-closed raise is designed to surface. Post-freeze, a forged key can still trip a 500 on that request only (process stays frozen); that path is reachable only via direct LangGraph-SDK calls (post-auth, server-local), since the gateway run path strips the key in `build_run_config`.

backend/packages/harness/deerflow/runtime/runs/worker.py
Vanzeren: Both added in 40452830:

- **Non-message channel restoration**: `test_rollback_restores_non_message_channels_via_fork_inheritance` builds a graph with an `artifacts` reducer channel alongside `messages`, captures the rollback point, runs a cancelled turn (messages 2, artifacts `["a0","a1"]`), rolls back, and asserts the restored checkpoint has the pre-run messages *and* `artifacts == ["a0"]` — fork inheritance for sibling channels is now pinned, not assumed.
- **Counted round-trip guards**: a `_CountingSaver` (InMemorySaver subclass counting `aget_tuple` calls and `alist` limits) backs `test_ahistory_pushes_limit_into_alist_and_reads_each_snapshot_once` (limit reaches SQL, zero tuple fetches in the read path) and `test_aget_fetches_the_checkpoint_exactly_once_in_full_mode` (one fetch per read). Both mutation-verified: removing the limit kwarg or re-adding the standalone gate fails them.

backend/AGENTS.md
Vanzeren: Both files added in 40452830 with real coverage:

- `tests/test_threads_checkpoint_mode.py` — drives the full thread lifecycle (multi-turn writes, latest-state read, bounded history, branch-from-earlier-checkpoint, compaction-style `Overwrite` write, resume-after-compaction) through `CheckpointStateAccessor` against the production `get_thread_state_schema(mode)` schemas, on both `InMemorySaver` and `AsyncSqliteSaver`, and asserts the materialized state is identical across full and delta modes. Includes a storage-level guard: delta checkpoints must not carry a per-step serialized `messages` blob (`<=1` snapshot blob across 4 turns) while full mode blobs every step — this is the O(N²) → O(N) contract, pinned.
- `tests/test_gateway_checkpoint_mode.py` — drives `GET /threads/{id}`, `GET /state`, `POST /history` through the real route stack (factory-built graph, real `InMemorySaver`) once per mode and asserts the wire responses are identical modulo checkpoint ids/timestamps.

Full suite: 7974 passed, 28 skipped.

Vanzeren:
Review round 4 (perf + parity) addressed in 40452830 — all 8 inline threads plus the parity-files addendum have individual replies.

**Perf fixes**
- `ahistory`/`history` `limit` now pushed through Pregel into `checkpointer.alist` (SQL `LIMIT`) — no more full-history materialization on bounded reads.
- Read-side mode-compat gate folded onto the returned snapshot's metadata; full-mode `aget` is back to exactly one checkpoint fetch (pinned by a counting-saver test, mutation-verified). Writes keep the pre-write fetch (fail-closed).
- `get_thread`: 3 fetches → 2. The remaining raw `pending_writes` fetch on the resolved checkpoint is semantically required: post-checkpoint `__error__` writes (rollback reattachment, worker error fallback) never surface in `snapshot.tasks` (verified empirically on 1.2.9), so task-based status derivation would misreport failed threads as idle.
- State-read endpoints no longer build a full lead agent per request: accessor graphs are cached per `(assistant_id, mode)` with factory-identity revalidation.
- `DeerFlowClient.get_thread` N+1 eliminated: one `checkpointer.list` walk collects per-checkpoint `pending_writes`.

**Hardening**
- `InMemorySaver` delta-history patch: stands down if the upstream override disappears, try/except on assignment, warns when langgraph moves past the validated version (1.2.9). No upstream issue exists for the bug yet — the docstring says so and instructs re-checking on every upgrade instead of fabricating a link.
- `make_lead_agent` mode precedence: first freeze is owned by `app_config` (client-supplied configurable key ignored); once frozen, any mismatch (forged key or config.yaml flip) fails closed. Deliberately not ignoring the key post-freeze — that would mask genuine operator config changes.
- Rollback's load-bearing assumption (non-message channels restored via fork inheritance) is now pinned by a dedicated reducer-channel test.

**Parity coverage** — the two files referenced by AGENTS.md / the Validation section now exist with real coverage:
- `test_threads_checkpoint_mode.py`: lifecycle parity (writes/read/history/branch/compaction-Overwrite/resume) on memory + sqlite, plus a storage guard asserting delta checkpoints carry no per-step `messages` blob (the O(N²) → O(N) contract).
- `test_gateway_checkpoint_mode.py`: `GET /threads/{id}`, `/state`, `POST /history` wire responses identical across modes through the real route stack.

`make lint` clean; full suite: **7974 passed, 28 skipped**.

backend/packages/harness/deerflow/runtime/runs/worker.py
willem-bd: Thanks for adding `test_rollback_restores_non_message_channels_via_fork_inheritance` in 40452830. On the current head I think this still has a gap:

The test restores `artifacts`, which is a **base `ThreadState`** channel — so it doesn't exercise the case the `build_state_mutation_graph` docstring warns about: channels contributed by custom `AgentMiddleware.state_schema` (memory, todos, uploads, sandbox state, …). The rollback still calls `build_state_mutation_graph("rollback_restore", accessor.mode)` with **no `state_schema`** (`worker.py:1205`), falling back to base `ThreadState`, while `branch` and `update_thread_state` resolve and pass `graph_state_schema(...)`. If `create_checkpoint` only persists channels present in the mutation graph's schema, middleware channels would be dropped from the restored checkpoint.

Separately, `compact_thread` has the same shape: it calls `build_checkpoint_state_mutation_accessor(..., as_node="manual_compaction")` with no `state_schema`, unlike `update_thread_state` which goes through `build_thread_checkpoint_state_mutation_accessor` (resolves the effective schema). The old compaction cloned the whole checkpoint blob, so this would be a regression for middleware channels.

Could you pass `state_schema=graph_state_schema(accessor.graph)` on both paths (matching branch/update), and add a rollback + compaction test asserting a middleware-contributed channel survives?

backend/app/gateway/services.py
willem-bd: Thanks — the `_state_accessor_graph` cache addresses the per-request rebuild. Two residual concerns on the current head:

1. The cache key is `(assistant_id, mode)`, which excludes `app_config` version. A config reload that adds or changes middleware channels would serve a stale compiled graph (and thus a stale schema) until the 64-entry cache evicts it. Consider keying on a config version or clearing the cache on reload.

2. The graph-build failure mode remains: if `make_lead_agent` fails (bad model config, MCP server down, misconfigured skill), state reads now 500 where the old checkpointer-only read would have succeeded — and accessor construction sits outside the endpoint `try/except`, so it surfaces as a raw 500 with no sanitized detail. In `full` mode the graph isn't strictly needed (full checkpoints store all `channel_values`); could reads short-circuit to the checkpointer in full mode and reserve the accessor path for delta?

backend/packages/harness/deerflow/runtime/checkpoint_mode.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def state_snapshot_uses_delta(snapshot: Any) -> bool:
+    """Whether a materialized ``StateSnapshot`` originates from a delta checkpoint."""
+    if snapshot is None:
+        return False
+    return checkpoint_metadata_uses_delta(getattr(snapshot, "metadata", {}) or {})
+
+
+def raise_if_snapshot_incompatible(snapshot: Any, mode: CheckpointChannelMode) -> None:
```
willem-bd, line 80: **The fail-closed mismatch path is only exercised with mocks.** Every `CheckpointModeMismatchError` test uses a `MagicMock`/`AsyncMock` checkpointer or a `FakeGraph` with hand-set metadata. No test writes a delta-marker checkpoint to a real `AsyncSqliteSaver`/`AsyncPostgresSaver`, reopens it in full mode, and asserts the error fires on `aget`/`aupdate`/`ahistory` — and no gateway-layer test asserts HTTP 500 from the read endpoints when a full-mode app opens a delta thread (router tests use `_FakeStateAccessor`/`_RawStateAccessor` that bypass the real gate).

For a feature whose entire purpose is preventing silent state corruption, the corruption-prevention path is regression-unprotected at the integration boundary. Could you add an E2E test that seeds a delta checkpoint into a real saver, points a full-mode accessor at it, and asserts the mismatch error on each read endpoint?

backend/app/gateway/routers/threads.py
```diff
@@ -930,154 +963,124 @@ async def compact_thread(thread_id: str, body: ThreadCompactRequest, request: Re
...
+    """Get the latest materialized graph state for a thread."""
+    # Resolve through the thread's assistant so custom middleware channels
+    # appear in the response instead of being dropped by the default schema.
+    accessor, config = await build_thread_checkpoint_state_accessor(request, thread_id=thread_id)
     try:
-        checkpoint_tuple = await checkpointer.aget_tuple(config)
+        snapshot = await accessor.aget(config)
     except Exception:
         logger.exception("Failed to get state for thread %s", sanitize_log_param(thread_id))
         raise HTTPException(status_code=500, detail="Failed to get thread state")
```
willem-bd, line 1030: **`CheckpointModeMismatchError` is swallowed into a generic 500.** The accessor call is wrapped in `except Exception: raise HTTPException(500, "Failed to get thread state")`, so the specific mismatch type and its message ("Thread requires delta mode; materialize and convert its checkpoints before using full mode") land only in server logs. An operator seeing "Failed to get thread state" on every thread after a mode flip has no signal that the root cause is a checkpoint mode mismatch without grepping logs. The same applies to the other state endpoints ("Failed to get thread", "Failed to update thread state", …).

Could you add a specific `except CheckpointModeMismatchError as exc: raise HTTPException(409, detail=str(exc)) from exc` ahead of the generic handler (409 Conflict is more precise than 500 for "persisted data conflicts with the process's frozen mode"; 503 fits `CheckpointModeReconfigurationError`), and include `thread_id` in the message?

backend/uv.lock
```diff
@@ -4755,47 +4818,33 @@ wheels = [
...
-    { url = "https://files.pythonhosted.org/packages/c2/b6/b9afed2afadddaf5ebb2afa801abf4b0868f42f8539bfe4b071b5266c9fe/websockets-16.0-cp314-cp314t-win32.whl", hash = "sha256:5a4b4cc550cb665dd8a47f868c8d04c8230f857363ad3c9caf7a0c3bf8c61ca6", size = 178085, upload-time = "2026-01-10T09:23:33.816Z" },
-    { url = "https://files.pythonhosted.org/packages/9f/3e/28135a24e384493fa804216b79a6a6759a38cc4ff59118787b9fb693df93/websockets-16.0-cp314-cp314t-win_amd64.whl", hash = "sha256:b14dc141ed6d2dde437cddb216004bcac6a1df0935d79656387bd41632ba0bbd", size = 178531, upload-time = "2026-01-10T09:23:35.016Z" },
-    { url = "https://files.pythonhosted.org/packages/6f/28/258ebab549c2bf3e64d2b0217b973467394a9cea8c42f70418ca2c5d0d2e/websockets-16.0-py3-none-any.whl", hash = "sha256:1637db62fad1dc833276dded54215f2c7fa46912301a24bd94d45d46a011ceec", size = 171598, upload-time = "2026-01-10T09:23:45.395Z" },
+version = "15.0.1"
```
willem-bd: **`websockets` is downgraded 16.0 -> 15.0.1 (a major-version regression).** The `langgraph-sdk` bump (0.3.13 -> 0.4.2) pulls in `websockets` and the resolver downgrades the existing 16.0 to 15.0.1. This affects every IM integration that uses websockets — `discord.py`, `dingtalk-stream`, `python-telegram-bot` — and no CI test exercises any of them against the resolved version.

Could you confirm which package forces `<16`; if none do, pin `websockets>=16`. At minimum, run the Discord/DingTalk/Telegram channel tests against 15.0.1. (Also: `langchain` 1.2.15->1.3.14 and `langchain-core` 1.3.3->1.4.9 came along transitively, but `harness/pyproject.toml` still declares `langchain>=1.2.15` with no upper bound — the lower bound no longer reflects what was tested.)

WillemJiang:
@Vanzeren, please fix the lint error.

backend/packages/harness/deerflow/runtime/runs/worker.py
Vanzeren: Verified this against LangGraph's checkpoint write path, and the premise doesn't hold for these two call sites: `update_state` clones the parent's channel blobs into the forked checkpoint, so channels absent from the write are inherited regardless of the mutation graph's schema. Only channels *present in the write itself* are filtered by schema — and rollback writes only `messages`, compaction writes only `messages` + `summary_text` (all base channels). Teeth-check: the new rollback test passes on the pre-change code too, i.e. nothing was actually dropped.

What I changed in f845caf7:
- `worker.py` rollback now passes `graph_state_schema(accessor.graph)` — zero cost (the agent graph already exists there) and aligns with the documented `build_state_mutation_graph` contract.
- `compact_thread` deliberately stays on the base-schema builder: resolving the effective schema would compile the full lead graph per compact call (a new factory-failure surface) for zero behavioral gain.
- Added inheritance contract locks that pin the load-bearing LangGraph semantic for both paths: `test_rollback_preserves_middleware_contributed_channels` (parametrized full + delta) and `test_compact_thread_context_preserves_middleware_contributed_channels`. If fork cloning ever stops carrying unknown channels, these fail before production state is lost.

backend/app/gateway/services.py
Vanzeren: Both addressed in f845caf7:

1. **Cache staleness**: the cache entry now carries the `app_config` identity and revalidates it on every call. `get_app_config()` builds a new object on hot-reload, so a config.yaml edit that changes middleware channels misses the cache and rebuilds. The cached reference keeps the old config object alive, so id-reuse can't produce a false hit.

2. **Graph-build failure mode**: implemented as a degraded fallback rather than a permanent full-mode short-circuit. Reason: `snapshot.next`/tasks are not derivable from a raw checkpoint tuple — they come from the graph's channel→node trigger analysis — so a second permanent read path would either approximate `next` or silently diverge on interrupted threads. Instead, `build_checkpoint_state_accessor` now falls back to a `_RawCheckpointReadAccessor` (raw `aget_tuple`/`alist`) only when the factory build raises in full mode: values/metadata/ancestry are complete from the tuple, `next`/tasks degrade to empty, and the fail-closed delta gate still applies (a delta thread gets the same 409, never partial state). Delta mode has no fallback — materialization needs the channel table, so the factory error surfaces as-is.

Covered by `test_full_mode_state_reads_degrade_to_raw_checkpointer_when_factory_fails`.

backend/packages/harness/deerflow/runtime/checkpoint_mode.py
Vanzeren: Done in f845caf7, at both boundaries:

- `test_full_mode_accessor_rejects_real_delta_checkpoint_on_sqlite`: seeds a delta checkpoint through the real delta mutation graph into a real `AsyncSqliteSaver` (marker + LangGraph delta counters verified present), reopens the thread with a full-mode accessor, and asserts `CheckpointModeMismatchError` on `aget`, `aupdate`, and `ahistory`.
- `test_full_mode_gateway_rejects_delta_thread_with_409`: drives the real route stack (factory-built graph → `CheckpointStateAccessor` → router) against a seeded delta thread and asserts HTTP 409 with the cause and thread id on GET /state, POST /state, POST /history, and GET /thread.

backend/app/gateway/routers/threads.py
Vanzeren: Done in f845caf7. The threads router now maps the guard errors ahead of every generic handler via `_checkpoint_mode_http_error`: `CheckpointModeMismatchError` → 409 with `str(exc)` and the thread id in the detail, `CheckpointModeReconfigurationError` → 503. Wired across get/update state, history, get thread, compact, the branch history scan, and the branch write — including the accessor-construction sites that previously sat outside the try blocks. The gateway-layer test in the sibling thread asserts the 409 detail end-to-end.

backend/uv.lock
Vanzeren: Confirmed which package forces `<16`: it is `langgraph-sdk` 0.4.2 itself (`Requires-Dist: websockets<16,>=14`), and langgraph 1.2.9 requires `langgraph-sdk>=0.4.2,<0.5.0` — so no sdk version compatible with this PR relaxes the cap, and a plain `websockets>=16` pin would be unsatisfiable.

Resolution in f845caf7: `[tool.uv] override-dependencies = ["websockets==16.0"]`, restoring the exact pre-PR version. Grounds for overriding the upstream pin:
- websockets 16.0's only breaking change is requiring Python ≥3.10 (no API removals; `websockets.legacy` stays until 2030); this repo requires ≥3.12.
- The sdk's ws transport only imports `websockets.asyncio.client`/`websockets.exceptions`, both 16-compatible — and DeerFlow never uses that transport (httpx/SSE only), so even an sdk-side ws incompatibility would be unreachable.
- On the affected surface: `discord.py` doesn't depend on websockets (aiohttp), `python-telegram-bot` only via an unenabled extra, `dingtalk-stream` declares `>=11.0.2` and is satisfied; `tests/test_channels.py` passes on 16.0.

Also bumped the harness `langchain` lower bound to `>=1.3` to match what the lockfile resolves. Longer term the clean fix is upstream: ask langgraph-sdk to relax the pin to `<17`.

WillemJiang (approved):

PR #4295, opened 2026-07-19
fix(loop-detection): clear the per-tool frequency counter on evict/reset

chuenchen309:
## Why

The loop-detection middleware's Layer-2 (per-tool-type frequency) keeps two
structures per thread: `_tool_name_history` (a windowed deque of recent tool
names) and `_tool_name_counter` (a `Counter` mirroring the deque so `freq_count`
is O(1)). LRU eviction (`_evict_if_needed`) and `reset` dropped the deque but
left the mirror `Counter` behind. When a thread id is LRU-evicted and later
reused (the middleware is long-lived and evicts at `max_tracked_threads`, default
100), its frequency count resumes from the stale value instead of zero, so the
thread's very first fresh tool call can be force-stopped with
`[FORCED STOP] Tool <name> called N times — exceeded the per-tool safety limit`
even though that tool was just called once. `reset()` had the same gap.

## What changed

An agent whose thread id is reused after LRU eviction (or after `reset()`) no
longer gets a spurious per-tool "forced stop" on its first tool call — Layer-2
frequency counting restarts from zero, as it already did for the windowed deque.
No config, API, or default-threshold changes; the deque and its mirror `Counter`
are simply kept in sync at all three cleanup sites (evict, per-thread reset, full
reset).

## Surface area

- [x] **Agents / LangGraph** — `LoopDetectionMiddleware` per-tool frequency state cleanup in `backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py`

## Bug fix verification

- Test path that reproduces the bug: `backend/tests/test_loop_detection_middleware.py::TestToolFrequencyDetection::test_freq_counter_cleared_on_eviction` (plus `::test_freq_counter_cleared_on_reset` for the reset paths)
- Did it go red on `main` and green on this branch? **yes** — on `main` both new tests fail: after two `read_file` calls on a thread, evicting it (via `max_tracked_threads=2`) leaves `_tool_name_counter["thread-evicted"] == {"read_file": 2}`, so the reused thread's first fresh `read_file` reaches `freq_count=3` and force-stops (the test asserts `result is None` and that the counter is dropped). With the fix the counter is cleared on eviction/reset and the fresh call returns `None`. An active (non-evicted) thread still force-stops correctly at the limit (`test_freq_hard_stop_at_limit` and the existing suite stay green — the counter is only dropped, never over-cleared).

## Validation

```
cd backend
PYTHONPATH=packages/harness:. .venv/bin/pytest tests/test_loop_detection_middleware.py tests/test_loop_detection_stop_reason.py tests/test_loop_detection_config.py
# 88 passed (86 existing + 2 new)
```

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** This PR was authored by an AI coding agent (Claude Code) running on this account — the AI traced the two Layer-2 structures, reproduced the stale-counter false force-stop against `main` (evict path and both reset paths), wrote the failing regression tests first (red → green), confirmed an active thread still hard-stops so the fix doesn't over-clear, and wrote this description. The verification above is real and re-runnable from the diff. The human account holder reviews every change and is accountable for it. If this isn't the kind of contribution you want, say so and I'll close it.

### DIFF
--- a/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
+++ b/backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
@@ -357,6 +357,7 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
             evicted_id, _ = self._history.popitem(last=False)
             self._warned.pop(evicted_id, None)
             self._tool_name_history.pop(evicted_id, None)
+            self._tool_name_counter.pop(evicted_id, None)
             self._tool_freq_warned.pop(evicted_id, None)
             for key in list(self._pending_warnings):
                 if key[0] == evicted_id:
@@ -718,6 +719,7 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
                 self._history.pop(thread_id, None)
                 self._warned.pop(thread_id, None)
                 self._tool_name_history.pop(thread_id, None)
+                self._tool_name_counter.pop(thread_id, None)
                 self._tool_freq_warned.pop(thread_id, None)
                 for key in list(self._pending_warnings):
                     if key[0] == thread_id:
@@ -726,6 +728,7 @@ class LoopDetectionMiddleware(AgentMiddleware[AgentState]):
                 self._history.clear()
                 self._warned.clear()
                 self._tool_name_history.clear()
+                self._tool_name_counter.clear()
                 self._tool_freq_warned.clear()
                 self._pending_warnings.clear()
                 self._pending_warning_touch_order.clear()
--- a/backend/tests/test_loop_detection_middleware.py
+++ b/backend/tests/test_loop_detection_middleware.py
@@ -1083,6 +1083,53 @@ class TestToolFrequencyDetection:
         assert "LOOP DETECTED" in mw._pending_warnings[_pending_key("thread-A")][0]
         assert not mw._pending_warnings.get(_pending_key("thread-B"))
 
+    def test_freq_counter_cleared_on_eviction(self):
+        """LRU eviction must drop the per-tool frequency counter along with the
+        window deque. Otherwise a reused thread id resumes from a stale count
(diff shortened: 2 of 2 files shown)


backend/packages/harness/deerflow/agents/middlewares/loop_detection_middleware.py
```diff
@@ -357,6 +357,7 @@ def _evict_if_needed(self) -> None:
             evicted_id, _ = self._history.popitem(last=False)
             self._warned.pop(evicted_id, None)
             self._tool_name_history.pop(evicted_id, None)
+            self._tool_name_counter.pop(evicted_id, None)
```
willem-bd, line 360: Non-blocking, forward-looking note (not something to address in this PR): the deque + Counter mirror is now correctly synced at all three cleanup sites, but the pairing is inherently sync-prone - any future site that drops `_tool_name_history[<id>]` must remember to also drop `_tool_name_counter[<id>]`, or this exact bug recurs. The `defaultdict` auto-create on reuse is what made the desync silently dangerous (fresh empty deque vs. stale Counter).

When this state is next opened, consider encapsulating the pair behind a single helper (e.g. a `_drop_thread_locked(thread_id)` used by both `_evict_if_needed` and per-thread `reset`) so the two structures can't drift. Out of scope here; the current change is correct and complete.

(Verified empirically: reverting just these three lines on `main` makes both new tests fail with `Counter({'read_file': 2})` visible in the assertion message, and restoring them passes. Red/green claim in the description checks out.)

willem-bd (commented):
Verified the fix: red on `main` (stale `Counter({'read_file': 2})` survives `reset`/evict) and green on this branch; full suite 88 passed; the two structures are now mirrored at all five sites (init, evict, runtime append/popleft, per-thread reset, full reset). One non-blocking forward-looking note inline.

WillemJiang (approved):

PR #4297, opened 2026-07-19
fix(skills): add security_fail_closed option for moderation model outages

### LINKED ISSUE #3021: Security scanner fail-closed blocks all skills when moderation model unavailable
## Issue Description

When the security moderation model call fails in \`scan_skill_content()\`, the fallback behavior is to block ALL content (both executable and non-executable). This creates a denial-of-service scenario if the moderation model becomes unavailable.

## Current Code

\`\`\`python
except Exception:
    logger.warning("Skill security scan model call failed; using conservative fallback", exc_info=True)

if executable:
    return ScanResult("block", "Security scan unavailable for executable content; manual review required.")
return ScanResult("block", "Security scan unavailable for skill content; manual review required.")
\`\`\`

## Problem

- **DoS Risk**: If the moderation model is down, ALL skill writes are blocked
- No configuration option to control fail-open vs fail-closed behavior
- Non-executable content (documentation, config) is blocked even though risk is lower
- Manual review requirement may not be practical for all deployments

## Impact

- Development workflow blocked during model outage
- Cannot write any new skills or update existing ones
- Potential productivity loss across the team

## Suggested Fix

Consider a more nuanced fallback strategy:

1. **Configurable behavior**: Add a config flag to control fail-open vs fail-closed
2. **Risk-based fallback**: 
   - Executable content: Block (high risk)
   - Non-executable content: Warn but allow (lower risk)
3. **Circuit breaker**: After N consecutive failures, switch to warn mode

\`\`\`python
config = app_config or get_app_config()
fail_closed = config.skill_evolution.security_fail_closed  # default True

if executable or fail_closed:
    return ScanResult("block", "...")
return ScanResult("warn", "Security scan unavailable; manual review recommended")
\`\`\`

## Location

\`backend/packages/harness/deerflow/skills/security_scanner.py\`
Lines: 68-70

JOJOCrazy123:
## Summary

Fixes #3021.

When the skill security moderation model call fails in `scan_skill_content()`, the fallback currently blocks **all** content — both executable and non-executable. This turns a moderation-model outage into a denial of service: no skills can be created or updated while the model is down.

This PR adds a `skill_evolution.security_fail_closed` option to control the fallback policy:

- **`true` (default)** — preserves current behavior: block all writes when the model is unavailable.
- **`false`** — allow non-executable content with a `warn` decision during an outage, while **executable content is still always blocked**.

The default keeps the existing fail-closed behavior, so this is backward compatible.

## Changes

- `config/skill_evolution_config.py`: add `security_fail_closed: bool = True`.
- `skills/security_scanner.py`: add `_resolve_fail_closed()` helper (defaults to fail-closed if config can't be resolved) and branch the model-unavailable fallback on it. The unparseable-output path and executable-content path are unchanged (still block).
- `config.example.yaml`: document the new option.
- `tests/test_security_scanner.py`: add coverage for fail-open (non-executable → warn), fail-open executable (still block), and explicit fail-closed (block).

## Test plan

- [x] `uv run pytest tests/test_security_scanner.py -q` → all pass (incl. new cases)
- [x] `uv run ruff check` on changed files → clean
- [x] Verified `SkillEvolutionConfig` default `security_fail_closed=True` and override to `False`

### DIFF
--- a/backend/packages/harness/deerflow/skills/security_scanner.py
+++ b/backend/packages/harness/deerflow/skills/security_scanner.py
@@ -129,7 +129,7 @@ async def scan_skill_content(
                 return ScanResult(decision, str(parsed.get("reason") or "No reason provided."))
         logger.warning("Security scan produced unparseable output: %s", raw[:200])
     except Exception:
-        logger.warning("Skill security scan model call failed; using conservative fallback", exc_info=True)
+        logger.warning("Skill security scan model call failed; applying configured fail-closed/fail-open policy", exc_info=True)
 
     if model_responded:
         return ScanResult("block", "Security scan produced unparseable output; manual review required.")
@@ -137,4 +137,5 @@ async def scan_skill_content(
         return ScanResult("block", "Security scan unavailable for executable content; manual review required.")
     if _resolve_fail_closed(app_config):
         return ScanResult("block", "Security scan unavailable for skill content; manual review required.")
+    logger.warning("Security scan unavailable; failing open for non-executable skill content at %s (manual review recommended)", location)
     return ScanResult("warn", "Security scan unavailable for non-executable skill content; manual review recommended.")
--- a/backend/packages/harness/deerflow/skills/security_scanner.py
+++ b/backend/packages/harness/deerflow/skills/security_scanner.py
@@ -22,6 +22,15 @@ class ScanResult:
     reason: str
 
 
+def _resolve_fail_closed(app_config: AppConfig | None) -> bool:
+    """Resolve the fail-closed policy, defaulting to True if config is unavailable."""
+    try:
+        config = app_config or get_app_config()
+        return bool(getattr(config.skill_evolution, "security_fail_closed", True))
+    except Exception:
+        return True
+
+
 def _extract_json_object(raw: str) -> dict | None:
     raw = raw.strip()
 
@@ -126,4 +135,6 @@ async def scan_skill_content(
         return ScanResult("block", "Security scan produced unparseable output; manual review required.")
     if executable:
         return ScanResult("block", "Security scan unavailable for executable content; manual review required.")
-    return ScanResult("block", "Security scan unavailable for skill content; manual review required.")
+    if _resolve_fail_closed(app_config):
+        return ScanResult("block", "Security scan unavailable for skill content; manual review required.")
+    return ScanResult("warn", "Security scan unavailable for non-executable skill content; manual review recommended.")
--- a/backend/packages/harness/deerflow/config/skill_evolution_config.py
+++ b/backend/packages/harness/deerflow/config/skill_evolution_config.py
@@ -12,3 +12,11 @@ class SkillEvolutionConfig(BaseModel):
         default=None,
         description="Optional model name for skill security moderation. Defaults to the primary chat model.",
     )
+    security_fail_closed: bool = Field(
+        default=True,
+        description=(
+            "When the moderation model is unavailable, block skill writes if True "
+            "(fail-closed). If False, non-executable content is allowed with a warning "
+            "while executable content is still blocked."
+        ),
+    )
--- a/backend/packages/harness/deerflow/config/skill_evolution_config.py
+++ b/backend/packages/harness/deerflow/config/skill_evolution_config.py
@@ -14,9 +14,5 @@ class SkillEvolutionConfig(BaseModel):
     )
     security_fail_closed: bool = Field(
         default=True,
-        description=(
-            "When the moderation model is unavailable, block skill writes if True "
-            "(fail-closed). If False, non-executable content is allowed with a warning "
-            "while executable content is still blocked."
-        ),
+        description=("When the moderation model is unavailable, block skill writes if True (fail-closed). If False, non-executable content is allowed with a warning while executable content is still blocked."),
     )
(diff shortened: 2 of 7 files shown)


fancyboi999 (changes requested):
Thanks @JOJOCrazy123. I found two issues that should be addressed before this is ready.

### [P2] Bump the config version for the new scanner policy

- Location: `config.example.yaml:18` and `config.example.yaml:1640-1644`
- Problem: this PR adds `skill_evolution.security_fail_closed` to the configuration schema but leaves `config_version` at 26. Existing version-26 installations will not receive the outdated-config warning, so the repository's `make config-upgrade` path is not discoverable for this new policy setting. That violates the config-versioning rule in `backend/AGENTS.md`.
- Evidence: current `origin/main` and head `1903648c` both declare `config_version: 26`, while only this PR adds the new persisted field.
- Suggested fix: bump `config.example.yaml` to the next available version (27 against the current base), and resolve the number again if another config-schema PR lands first.
- Test: cover a version-26 user config against the updated example and assert it is reported as outdated and the upgrade path merges `security_fail_closed: true` without changing the user's existing `skill_evolution` values.

### [P2] Run the required backend formatter

- Location: `backend/packages/harness/deerflow/config/skill_evolution_config.py:16-22`
- Problem: the current head does not pass the repository's enforced formatting gate, so backend CI will reject it even though `ruff check` is clean.
- Evidence: in an isolated worktree on `1903648c`, `uv run ruff format --check .` reports `Would reformat: packages/harness/deerflow/config/skill_evolution_config.py`; the diff collapses the multiline `description` expression to the repository formatter's output.
- Suggested fix: run `cd backend && make format` (or the equivalent `uv run ruff format`) and commit the resulting change.
- Test: rerun `uv run ruff format --check .` and keep the full focused scanner/config suite green.

Validation on the current head used a detached worktree and isolated uv environment: the security scanner, skill-manage, installer, custom-skill router, app-config, and config-version suites passed (`124 passed`); `ruff check .` passed; only the format check above failed.

JOJOCrazy123:
Thanks @fancyboi999 — both P2 items are addressed in 9bb263d5:

- **Config version**: bumped `config.example.yaml` `config_version` 26 → 27 so existing v26 installs are flagged outdated and pick up `skill_evolution.security_fail_closed` via `make config-upgrade`. Added regression tests in `backend/tests/test_config_version.py` asserting a v26 config is reported outdated against the real example and that the upgrade merge adds `security_fail_closed: true` without changing the user's existing `skill_evolution` values.
- **Formatter**: ran `uv run ruff format` on `skill_evolution_config.py`; `uv run ruff format --check .` and `ruff check .` are both clean now.

Local: `pytest tests/test_config_version.py tests/test_security_scanner.py` → 29 passed. Please re-resolve the version number if another config-schema PR lands first.

backend/packages/harness/deerflow/skills/security_scanner.py
```diff
@@ -126,4 +135,6 @@ async def scan_skill_content(
         return ScanResult("block", "Security scan produced unparseable output; manual review required.")
     if executable:
         return ScanResult("block", "Security scan unavailable for executable content; manual review required.")
-    return ScanResult("block", "Security scan unavailable for skill content; manual review required.")
+    if _resolve_fail_closed(app_config):
+        return ScanResult("block", "Security scan unavailable for skill content; manual review required.")
+    return ScanResult("warn", "Security scan unavailable for non-executable skill content; manual review recommended.")
```
willem-bd, line 141: **Observability gap: the fail-open `warn` is silent.** When `security_fail_closed=false` and the moderation model is unavailable, this returns `warn` for non-executable content, but:

1. Nothing logs that fail-open was triggered. The only log is the pre-existing `except Exception:` at line 132 (`"Skill security scan model call failed; using conservative fallback"`), which is also now slightly inaccurate when fail-open is configured - the fallback is no longer "conservative" in that case.
2. The caller `_scan_skill_file_or_raise` in `installer.py` drops `result.reason` for non-executable `warn` decisions and returns without surfacing anything, so during a model outage with fail-open enabled, skill writes proceed with no operator-visible signal that scanning was skipped.

Suggested (small):
- Add a `logger.warning(...)` here before returning, e.g. `logger.warning("Security scan unavailable; failing open for non-executable skill content")`.
- Reword the line 132 message so it stays accurate in both modes, e.g. `"...failed; applying configured fail-closed/fail-open policy"`.
- Separately, surfacing `warn` reasons from the installer would close the gap at the caller side (that file isn't in this diff).

willem-bd (commented):
Observability note for the new fail-open path. Posting as inline comment on the `warn` return below.

JOJOCrazy123:
Thanks @willem-bd — good catch on the silent fail-open. Addressed in 707fd40d:

- Added a `logger.warning("Security scan unavailable; failing open for non-executable skill content at %s ...", location)` right before the fail-open `warn` return, so a moderation-model outage with `security_fail_closed=false` is now operator-visible instead of silent.
- Reworded the model-call-failed log (was "using conservative fallback") to "applying configured fail-closed/fail-open policy" so it stays accurate in both modes.
- Added a regression test (`test_fail_open_logs_operator_visible_warning`) asserting the fail-open path emits the warning.

`ruff format --check`, `ruff check`, and `pytest tests/test_security_scanner.py` (21 passed) are green.

Agreed that surfacing `warn` reasons from `installer.py` at the caller side would fully close the gap; that file isn't in this diff, so I've left it out of scope for this PR — happy to follow up separately.

WillemJiang (approved):

PR #4301, opened 2026-07-19
fix(gateway): handle null config.configurable in _resolve_thread_id

chuenchen309:
## Why

`_resolve_thread_id` crashes with an unhandled `AttributeError` (HTTP 500) when a client sends `config.configurable` as JSON `null` on `POST /api/runs/stream` or `/api/runs/wait`. `RunCreateRequest.config` is typed `dict[str, Any] | None`, so `{"config": {"configurable": null}}` is a valid request body that any authenticated client (e.g. SDK-style callers that build `config` programmatically and leave `configurable` unset/null) can send. It should start a fresh thread, not 500.

## What changed

`POST /api/runs/stream` and `/api/runs/wait` no longer 500 when `config.configurable` is `null` — they now generate a new thread id, exactly as they already do for `config: null` and `configurable: {}`. No change for any request that worked before.

## Surface area

- [ ] **Default behavior change** — (only the previously-crashing input changes; all working inputs are byte-for-byte identical)

## Bug fix verification

- Test path that reproduces the bug: `backend/tests/test_runs_api_endpoints.py::test_resolve_thread_id_handles_null_configurable`
- Did it go red on `main` and green on this branch? **yes** — red on `main` with `AttributeError: 'NoneType' object has no attribute 'get'` at `runs.py:29`; green after the one-line fix.

The root cause: `(body.config or {}).get("configurable", {})` only falls back to `{}` when the key is *absent*; an explicit `null` leaves the key present with value `None`, so the following `.get("thread_id")` dereferences `None`. Coalescing with `or {}` fixes it, matching the defensive idiom already used by `_checkpoint_configurable` in `thread_runs.py`.

## Validation

- `python -m pytest tests/test_runs_api_endpoints.py` → **19 passed** (incl. the new regression test).
- `ruff check app/gateway/routers/runs.py tests/test_runs_api_endpoints.py` → clean.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** An AI coding agent (Claude Code) running on this account found the bug, ran the repro, wrote the failing test and this description, end to end. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff.

### DIFF
--- a/backend/app/gateway/routers/runs.py
+++ b/backend/app/gateway/routers/runs.py
@@ -26,7 +26,7 @@ router = APIRouter(prefix="/api/runs", tags=["runs"])
 
 def _resolve_thread_id(body: RunCreateRequest) -> str:
     """Return the thread_id from the request body, or generate a new one."""
-    thread_id = (body.config or {}).get("configurable", {}).get("thread_id")
+    thread_id = ((body.config or {}).get("configurable") or {}).get("thread_id")
     if thread_id:
         return str(thread_id)
     return str(uuid.uuid4())
--- a/backend/tests/test_runs_api_endpoints.py
+++ b/backend/tests/test_runs_api_endpoints.py
@@ -321,9 +321,4 @@ def test_resolve_thread_id_handles_null_configurable():
     uuid.UUID(tid)  # a freshly generated id, not a crash
 
     # working inputs are unaffected
-    assert (
-        runs._resolve_thread_id(
-            RunCreateRequest(config={"configurable": {"thread_id": "t1"}})
-        )
-        == "t1"
-    )
+    assert runs._resolve_thread_id(RunCreateRequest(config={"configurable": {"thread_id": "t1"}})) == "t1"
--- a/backend/tests/test_runs_api_endpoints.py
+++ b/backend/tests/test_runs_api_endpoints.py
@@ -304,3 +304,26 @@ class TestRunFeedback:
         with TestClient(app) as client:
             response = client.get("/api/runs/run-fb-3/feedback")
         assert response.status_code == 503
+
+
+def test_resolve_thread_id_handles_null_configurable():
+    """A client may send ``config.configurable`` as JSON ``null``.
+
+    The key is then present with value ``None``, so the old
+    ``.get("configurable", {}).get("thread_id")`` raised ``AttributeError``
+    (an unhandled HTTP 500). Per the docstring it should generate a new id.
+    """
+    import uuid
+
+    from app.gateway.routers.thread_runs import RunCreateRequest
+
+    tid = runs._resolve_thread_id(RunCreateRequest(config={"configurable": None}))
+    uuid.UUID(tid)  # a freshly generated id, not a crash
+
+    # working inputs are unaffected
+    assert (
+        runs._resolve_thread_id(
+            RunCreateRequest(config={"configurable": {"thread_id": "t1"}})
+        )
+        == "t1"
+    )
--- a/backend/tests/test_runs_api_endpoints.py
+++ b/backend/tests/test_runs_api_endpoints.py
@@ -322,3 +322,23 @@ def test_resolve_thread_id_handles_null_configurable():
 
     # working inputs are unaffected
     assert runs._resolve_thread_id(RunCreateRequest(config={"configurable": {"thread_id": "t1"}})) == "t1"
+
+
+def test_build_run_config_handles_null_configurable():
+    """A null ``configurable`` must also survive ``build_run_config``.
+
+    ``_resolve_thread_id`` is not the only place that reads it: ``build_run_config``
+    does ``configurable.update(request_config.get("configurable", {}))`` and, in the
+    ``context`` branch, ``request_config.get("configurable", {}).keys()``. With the
+    key present and ``None``, ``.get(..., {})`` returns ``None``, so both raised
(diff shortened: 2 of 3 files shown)


fancyboi999 (changes requested):
Thanks @chuenchen309. I found one issue that should be addressed before this is ready.

[P2] Format the regression test so the backend gate can pass

- Location: `backend/tests/test_runs_api_endpoints.py:324-329`
- Problem: the bug fix and focused tests are correct, but the submitted test file fails the repository's mandatory `ruff format --check` gate.
- Evidence: isolated validation passed all 19 tests in `test_runs_api_endpoints.py`; `ruff check` also passed. `ruff format --check`, however, reports this file would be reformatted and collapses the newly added multiline assertion to one line under the repository's 240-character limit.
- Suggested fix: run `cd backend && uv run ruff format tests/test_runs_api_endpoints.py` and commit the resulting formatting-only change.
- Test: rerun `ruff format --check` for the two changed files and `pytest tests/test_runs_api_endpoints.py -q`.


Reviewed base SHA: `65792c20a592982b4632c44e39e51d15579498b9`
Reviewed head SHA: `52d00b4f091b2f435c2ada0f92c6c01f94da27f0`

backend/app/gateway/routers/runs.py
```diff
@@ -26,7 +26,7 @@
 
 def _resolve_thread_id(body: RunCreateRequest) -> str:
     """Return the thread_id from the request body, or generate a new one."""
-    thread_id = (body.config or {}).get("configurable", {}).get("thread_id")
+    thread_id = ((body.config or {}).get("configurable") or {}).get("thread_id")
```
willem-bd, line 29: This one-liner is correct for the line it touches — `or {}` is the right way to coerce an explicit `null` `configurable` to `{}`.

But it doesn't fully resolve the reported bug: `POST /api/runs/stream` and `/api/runs/wait` **still 500** with `config={"configurable": None}`. After `_resolve_thread_id` returns a fresh UUID, `body.config` flows unchanged into `build_run_config` (`backend/app/gateway/services.py:724`), whose `else` branch does at `services.py:496`:

```python
configurable.update(request_config.get("configurable", {}))
```

Here `request_config.get("configurable", {})` returns `None` (key present, so `.get` ignores its default), and `dict.update(None)` raises `TypeError: 'NoneType' object is not iterable`. So the exception just moves from `AttributeError` (here) to `TypeError` (there) — the user-facing 500 persists.

The `"context"` branch right above already coerces `context_value is None` → `{}` (`services.py:472–473`), so the symmetric fix matches existing intent:

```python
configurable.update(request_config.get("configurable") or {})
```

(Latent sibling: `services.py:469` calls `.get("configurable", {}).keys()`, which would crash the same way if a client sent both `context` and `configurable: null` — worth hardening in the same pass.)

backend/tests/test_runs_api_endpoints.py
```diff
@@ -304,3 +304,21 @@ def test_503_when_feedback_repo_not_configured(self):
...
+
+    The key is then present with value ``None``, so the old
+    ``.get("configurable", {}).get("thread_id")`` raised ``AttributeError``
+    (an unhandled HTTP 500). Per the docstring it should generate a new id.
+    """
+    import uuid
+
+    from app.gateway.routers.thread_runs import RunCreateRequest
+
+    tid = runs._resolve_thread_id(RunCreateRequest(config={"configurable": None}))
```
willem-bd, line 320: This test passes but gives false confidence: it calls `_resolve_thread_id` **in isolation**, so it can't see the second crash in `build_run_config` (`services.py:496`) noted on the `runs.py` comment. The unit goes green while the endpoint still returns 500.

Recommend an HTTP-level regression instead — POST `{"config": {"configurable": null}}` to `/api/runs/wait` (or `/stream`) via `TestClient` and assert a non-5xx status with a freshly-generated `thread_id`. This file already has the scaffolding (`_make_app`, `make_authed_test_app`). Adding `config=None` and `config={"configurable": {}}` cases would also lock in the "behaves like `config: null`" claim from the PR description.

Nit: `import uuid` and the `RunCreateRequest` import sit inside the function, but the rest of this file imports at module top (`runs` is already imported at line 12) — worth hoisting for consistency.

willem-bd (commented):
Reviewed the null-`configurable` fix. The `runs.py` change is correct in isolation, but the endpoint still 500s downstream in `build_run_config` — details inline. Two comments below.

chuenchen309:
Thanks both — pushed in ce2ac2d.

@willem-bd you're right, and thanks for catching it: the `or {}` in `runs.py` only covered `_resolve_thread_id`, but `build_run_config` reads `configurable` again and would still 500 on an explicit `null`. Extended the same coercion there:
- `services.py:496` `configurable.update(request_config.get("configurable") or {})` (was `dict.update(None)`)
- `services.py:469` `(request_config.get("configurable") or {}).keys()` in the both-`context`-and-`configurable` warning path (was `None.keys()`)

And replaced the isolated `_resolve_thread_id` assertion with a direct `build_run_config` test that exercises both branches (`{"configurable": None}` and `{"context": {}, "configurable": None}`) — it fails on the old code with `TypeError` / `AttributeError` and passes now.

@fancyboi999 `ruff format --check` and `ruff check` are green on both changed files.

WillemJiang (approved):

PR #4303, opened 2026-07-19
fix(workspace-changes): count diff body lines starting with "-- "/"++ "

chuenchen309:
## Why

The workspace-change summary (`+N -M`) under-counts edits to files whose lines start with `-- ` or `++ ` — SQL files are the common case (`-- comment`). `_count_diff_lines` drops difflib's two file headers by skipping any line starting with `+++ `/`--- `, but a deleted body line `-- get users` becomes the diff line `--- get users` (and an added `++ x` becomes `+++ x`), so those real edits get dropped from the count too. The existing comment already anticipated the `+++foo` (no-space) form and switched to a trailing-space prefix — but that left the with-space form (`-- `/`++ `) broken.

## What changed

The `+N -M` change summary now counts deletions/additions of lines that start with `-- ` or `++ ` (e.g. SQL comments). No change for any diff that didn't contain such lines.

## Surface area

- [x] **Backend API** — the counts feed `WorkspaceChangeSummary` / `WorkspaceFileChange` and the run-event summary string
- [ ] **Default behavior change** — only the previously-miscounted case changes

## Bug fix verification

- Test path: `backend/tests/test_workspace_changes.py::test_count_diff_lines_counts_content_starting_with_dashes_or_pluses`
- Red on `main`, green on this branch? **yes** — deleting `-- get users` and adding `SELECT 2` counts `+1 -0` before (the deletion is dropped) and `+1 -1` after. The existing `test_count_diff_lines_ignores_only_real_headers` still passes (its `+++added`/`---removed` body lines are counted, real headers are the first two lines).

Fix: `difflib.unified_diff` always emits the two file headers first, so skip them by position (`lines[2:]`) rather than by prefix. `@@` hunk headers start with `@` and remain uncounted.

## Validation

- `python -m pytest tests/test_workspace_changes.py -k count_diff_lines` → **2 passed**.
- `ruff check` on the two changed files → clean.

## AI assistance

**Tool(s) used:** Claude Code

**How you used it:** An AI coding agent (Claude Code) running on this account found the bug, wrote the repro and the regression test, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff.

### DIFF
--- a/backend/packages/harness/deerflow/workspace_changes/diff.py
+++ b/backend/packages/harness/deerflow/workspace_changes/diff.py
@@ -183,12 +183,12 @@ def _snapshot_text(file: FileSnapshot | None) -> str | None:
 def _count_diff_lines(lines: list[str]) -> tuple[int, int]:
     additions = 0
     deletions = 0
-    for line in lines:
-        # Unified-diff file headers are "+++ " / "--- " with a trailing space;
-        # a bare "+++"/"---" prefix would also swallow real content lines whose
-        # text begins with those sequences (e.g. an added line "+++foo").
-        if line.startswith("+++ ") or line.startswith("--- "):
-            continue
+    # difflib.unified_diff always emits the two file headers ("--- a/…" then
+    # "+++ b/…") first; skip them by position. A prefix test can't tell them
+    # apart from a hunk-body line whose content starts with "-- "/"++ " (e.g. a
+    # deleted SQL comment "-- get users" becomes the diff line "--- get users"),
+    # which would then be dropped from the count.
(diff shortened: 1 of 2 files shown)


Vanzeren:
Thanks for your contribution! The first sentence ties the positional skip to difflib's output format, which is good — could we make the contract explicit so a future caller doesn't feed this other diff sources? e.g. append: "Only valid for raw difflib.unified_diff output; sources with extra header lines (e.g. git diff) would be miscounted."

WillemJiang (approved):

PR #4306, opened 2026-07-20
feat(tui): add clear command

18062706139fcz:
## Why

The TUI already had local clear-screen behavior, but `/clear` was not registered in the slash-command registry. That made the command resolve as unknown and kept it out of the command palette.

Exposing it also needs to keep active runs stable: clearing or starting a new thread during a stream would drop the row-tracking state that incoming deltas rely on.

## How

- Register `clear` as a built-in TUI command and include it in the in-app help text.
- Guard `/clear` and `/new` while a run is streaming with the existing "Still working" feedback path.
- Keep `ClearRows` display-only by clearing transcript rows and transient streaming tracking while preserving the thread title.
- Update `README.md`, `backend/docs/TUI.md`, and `backend/AGENTS.md` for the supported command and active-run boundary.
- Add focused registry, Textual app, and reducer tests for command resolution, help text, display clearing, active-stream blocking, `/new` blocking, and title preservation.


backend/packages/harness/deerflow/tui/command_registry.py
```diff
@@ -36,6 +36,7 @@ class Resolution:
 BUILTIN_COMMANDS: tuple[Command, ...] = (
     Command("help", "Show commands and keybindings"),
     Command("new", "Start a fresh thread"),
+    Command("clear", "Clear the transcript display"),
```
willem-bd, line 39: Registering `clear` here makes the existing `ClearRows` path user-reachable, so it's worth flagging the reducer's full effect.

`ClearRows` does `replace(state, rows=(), title=None, streaming_id=None, streaming_anonymous_row_index=None)` — so `/clear` also resets the conversation **title** and the in-flight **streaming row tracking**, not just the visible rows. The help text and docs describe it as "removes rows from the current terminal display," which understates this.

Practical implication: if `/clear` is run mid-stream, the display empties but the still-running stream repopulates with a fresh assistant row on the next delta (no crash — a new row just appears). Consider either guarding mid-stream clears with the existing "still working" path used elsewhere, or noting in the docs that the title is also reset.

Minor since this is pre-existing `ClearRows` semantics — just newly reachable via this registration.

docs/superpowers/specs/2026-07-20-tui-clear-command-design.md
```diff
@@ -0,0 +1,64 @@
...
+transcript display without deleting or resetting persistent conversation
+state.
+
+## Non-goals
+
+- Do not delete or replace the active thread.
+- Do not modify checkpoints, memory, uploads, artifacts, goals, or token usage.
+- Do not invoke the model or any Gateway API.
+- Do not add a keyboard shortcut or change the existing `Ctrl+L` redraw binding.
+- Do not refactor the help system or alter `ClearRows` reducer semantics.
```
willem-bd: **Spec/plan contradict the implementation.** This non-goal (and plan Task 1 Step 5: "Do not alter the existing `_handle_builtin` `clear` branch or the `ClearRows` reducer") is violated by the implementation in this same PR: `_handle_builtin` now adds a streaming guard, and `ClearRows` (`view_state.py:206`) no longer clears `title`. Both changes are reasonable — but the committed spec/plan now read as stale and would mislead a future reader (or an agent re-running the plan). Reconcile: update the spec/plan to bless the streaming guard and title-preservation with rationale, or note the justified deviation here.

backend/packages/harness/deerflow/tui/view_state.py
```diff
@@ -203,7 +203,7 @@ def reduce(state: ViewState, action: Action) -> ViewState:
         return replace(state, title=action.title)
 
     if isinstance(action, ClearRows):
-        return replace(state, rows=(), title=None, streaming_id=None, streaming_anonymous_row_index=None)
+        return replace(state, rows=(), streaming_id=None, streaming_anonymous_row_index=None)
```
willem-bd, line 206: **PR description under-represents the scope.** This title-preservation change and the streaming guard added in `app.py` are both behavioral changes, but the PR "How" bullets list only command registration + docs + tests. A reviewer scanning the summary would miss that reducer semantics changed (title is now preserved on `/clear`) and that `/clear` is now blocked during an active run. Worth a bullet noting both so the description matches the diff.

backend/packages/harness/deerflow/tui/app.py
```diff
@@ -36,7 +36,7 @@
 )
 from .widgets.composer import ComposerInput
 
-_HELP_TEXT = "Commands:  /new  /threads  /goal  /model  /skills  /tools  /mcp  /memory  /usage  /config  /quit\nKeys:  Enter send · Ctrl+C interrupt or quit · Ctrl+L redraw · / commands · Esc close overlay"
+_HELP_TEXT = "Commands:  /new  /clear  /threads  /goal  /model  /skills  /tools  /mcp  /memory  /usage  /config  /quit\nKeys:  Enter send · Ctrl+C interrupt or quit · Ctrl+L redraw · / commands · Esc close overlay"
```
willem-bd, line 39: **`area:docs` label looks wrong.** That label is defined as "Documentation and Markdown only," but this PR ships real code changes — `command_registry.py`, `app.py` (this help text plus the new streaming guard), and `view_state.py` (reducer semantics) — plus tests. `risk:medium` is the operative label; `area:docs` should be swapped for the appropriate area label (the TUI/agents area) so it routes correctly.

backend/packages/harness/deerflow/tui/app.py
```diff
@@ -370,6 +370,9 @@ def _handle_builtin(self, name: str, args: str) -> None:
             self.state = initial_state()
             self._dispatch(SystemMessage("Started a new thread."))
         elif name == "clear":
+            if self._streaming:
+                self._dispatch(SystemMessage("Still working — wait for the current run to finish.", tone="info"))
+                return
```
willem-bd, line 381: **Asymmetry with `/new` (follow-up, not a blocker).** `/clear` now blocks during an active stream with the canonical "Still working" message — a good defensive choice (clearing mid-stream would orphan `streaming_id` so incoming deltas have nowhere to append). But `/new` (`app.py` `name == "new"` branch) does the strictly more destructive `self.state = initial_state()` mid-stream with **no** such guard. If blocking `/clear` is justified to avoid streaming desync, `/new` mid-stream is the bigger hazard. Pre-existing, not introduced here — worth a follow-up issue to give `/new` the same (or stronger) guard.

backend/packages/harness/deerflow/tui/command_registry.py
willem-bd: Follow-up: both points here were addressed in a later push — `ClearRows` now preserves `title` (`view_state.py:206`), and the `/clear` branch blocks mid-stream with the canonical "Still working" guard (`app.py:373`). Considering this resolved.

One residual: the committed spec/plan still say "do not alter `ClearRows` reducer semantics" / "do not alter the `_handle_builtin` clear branch" even though both were (reasonably) changed — flagged separately on the spec file so the docs can be reconciled with the implementation.

WillemJiang (approved):

PR #4307, opened 2026-07-20
fix(gateway): stop implying 200 webhook deliveries are unrecoverable

DaoyuanLi2816:
## Why

PR #4289 corrected a real inaccuracy in the GitHub webhook route's docs — GitHub does not auto-retry 5xx deliveries — but fancyboi999 flagged (CHANGES_REQUESTED, on the PR's sole commit `ac4e4ee9`) that the replacement wording swung too far the other way: it described a mistaken 200 response as dropping the webhook "forever with no way to recover it" / "permanently," which reads as manual recovery being impossible too. That review went unaddressed before #4289 was approved and merged about 40 minutes later. This PR addresses a comment on #4289 that got merged before it was resolved.

I verified fancyboi999's claim directly against GitHub's docs rather than taking it at face value: the manual "Redeliver" button and the REST/App redelivery endpoints place no failed-status precondition on the delivery id — any past delivery (success or failure) can be redelivered within GitHub's ~3-day window. The actual gap a mistaken 200 creates is discoverability (it won't show up in Recent Deliveries as failed, and GitHub's own recommended recovery-script pattern filters on non-OK status by convention, not platform restriction), not literal unrecoverability.

## What changed

- The webhook route's docstring and its inline fan-out comment (`backend/app/gateway/routers/github_webhooks.py`) no longer claim a 200'd delivery is unrecoverable; they now frame the 200-vs-503 difference as discoverable-and-found vs. theoretically-redeliverable-but-unsignaled, and link GitHub's redelivery docs alongside the existing failed-deliveries doc.
- `test_dispatch_failure_returns_503_not_200`'s docstring in `backend/tests/test_github_webhooks.py` gets the same correction. No assertions changed — this is a documentation-accuracy fix, not a behavior change.

## Surface area

## Validation

```
cd backend
uv run pytest tests/test_github_webhooks.py tests/test_github_dispatcher.py tests/test_github_channel.py tests/test_github_registry.py tests/test_channels.py
uv run ruff check app/gateway/routers/github_webhooks.py tests/test_github_webhooks.py
uv run ruff format --check app/gateway/routers/github_webhooks.py tests/test_github_webhooks.py
```

All 322 tests pass; ruff check and format --check are both clean.

### DIFF
--- a/backend/app/gateway/routers/github_webhooks.py
+++ b/backend/app/gateway/routers/github_webhooks.py
@@ -187,19 +187,27 @@ async def receive_github_webhook(
     automatically retry a failed delivery of any kind — 5xx, timeout, or
     connection error are all simply recorded as failed; see
     https://docs.github.com/en/webhooks/using-webhooks/handling-failed-webhook-deliveries.
-    A 200 response marks the delivery permanently successful, so
-    swallowing a transient registry filesystem error or bus publish
-    failure into a 200 would silently drop a real webhook forever with no
-    way to recover it. Returning 503 instead keeps the delivery correctly
-    recorded as failed in GitHub's Recent Deliveries / Deliveries API, so
-    an operator's manual "Redeliver" click, a REST API call, or a
-    self-hosted recovery script (the pattern GitHub's own docs recommend)
-    can still recover it — by the time that redelivery lands the
-    underlying outage is usually gone. The `is_route_enabled()` startup
-    check still handles *configuration* errors fail-closed (route absent
-    → 404); 503 is reserved for runtime failures worth making recoverable
-    this way. Permanent / non-retryable conditions (unknown event, missing
-    channel service) keep returning 200.
+    Swallowing a transient registry filesystem error or bus publish
+    failure into a 200 would mark the delivery successful, so it never
+    surfaces as failed in GitHub's Recent Deliveries / Deliveries API and
+    a scheduled recovery script filtering on non-OK status (the pattern
+    GitHub's own docs recommend) will never flag it for redelivery. That
+    is a discoverability problem, not literal unrecoverability — GitHub's
+    manual "Redeliver" button and its REST/App redelivery endpoints place
+    no failed-status precondition on the delivery id (see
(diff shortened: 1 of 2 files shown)


zhfeng (approved):
Agreed with this change - approving. The reframe from "unrecoverable" to "discoverability gap" is the accurate reading: GitHub's manual Redeliver / REST redelivery endpoints carry no failed-status precondition, so a 200'd delivery isn't lost, just un-signaled. I verified the runtime logic matches the wording - `fanout_event` propagates exceptions (no internal swallowing), the `try/except` around it converts them to 503, and the permanent pre-fanout states (no channel service, channel disabled, unhandled event) correctly stay on 200. `CancelledError` isn't masked as 503 either (it's `BaseException`, not caught by `except Exception`).

One thing worth a follow-up (out of scope for this PR, which is docs-only):

The docstring's three-way categorization - startup config errors -> 404, transient runtime -> 503, permanent -> 200 - doesn't account for the **runtime-cleared-secret 503** in the `if secret is None` defense-in-depth guard. When `GITHUB_WEBHOOK_SECRET` is unset *at runtime* (a rotation that cleared it without a restart), that guard returns 503. That's a configuration error `is_route_enabled()` didn't catch at startup, and it's permanent-until-config-restored - structurally the same as the `channels.github.enabled: false` branch, which deliberately returns 200 because retrying hits the same disabled state until the config changes.

So the same reasoning is applied inconsistently: channel-disabled (intentional off-switch) -> 200, but cleared-secret (unintentional breakage) -> 503. The 503 is defensible - it surfaces an unintended breakage in Recent Deliveries, which an operator wants to see - but the docstring doesn't mention this fourth path, so a reader mapping docs -> code will be surprised by a 503 from a config-rotation scenario the docstring implies is handled at startup as 404. Either acknowledge it in the docstring, or spell out that the 200-vs-503 split for permanent config states is intentional-vs-unintentional misconfiguration.

zhfeng (approved):

WillemJiang (approved):

